PR #1938, opened 2017-06-01
Add Logout feature to CLI

### LINKED ISSUE #1933: Question: How to logout of `gem` cli?
I'm having a problem.

My current problem is: I am unable to logout of the `gem` cli. I have logged in with one email but want to login as another now.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

shivabhusal:
# Description:
It clears the current session out by removing the locally stored credentials.
Updates the list of available commands in help document.

## Special usage
When you have multiple rubygems account. if you are already signed currently you cannot sign in using another set of credentials. But, with this addition, you now can.

closes #1933 
# Tasks:

- [x] Write tests


lib/rubygems/gemcutter_utilities.rb
```diff
@@ -165,4 +165,3 @@ def set_api_key host, key
   end
 
 end
-
```
simi: ✂️

lib/rubygems/command_manager.rb
```diff
@@ -216,4 +217,3 @@ def load_and_instantiate(command_name)
   end
 
 end
-
```
simi: ✂️

lib/rubygems/commands/logout_command.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+require 'rubygems/command'
+require 'rubygems/commands/query_command'
+require 'rubygems/config_file'
+
+##
+# An alternate to Gem::Commands::QueryCommand that searches for gems starting
+# with the supplied argument.
+
+class Gem::Commands::LogoutCommand < Gem::Commands::QueryCommand
```
simi: Isn't `Gem::Commands::Command enough`?

test/rubygems/test_gem_gemcutter_utilities.rb
```diff
@@ -207,9 +207,9 @@ def util_sign_in response, host = nil, args = []
 
     use_ui @sign_in_ui do
       if args.length > 0 then
-        @cmd.sign_in(*args)
+        @cmd.send(:sign_in, *args)
```
simi: Why is this change needed?

lib/rubygems/commands/logout_command.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def initialize
+    super 'logout', 'Logout from the current session'
```
segiddins: this should probably say that is will logout from _all_ sessions

shivabhusal:
@segiddins what next? I have made suggested changes and test passes too. Please suggest.

lib/rubygems/commands/signout_command.rb
```diff
@@ -0,0 +1,33 @@
...
+  def execute
+    if Gem.configuration.unset_api_key! then
+      $stdout.puts "You have successfully signed out out from 'RubyGems.org'."
+    else
+      $stdout.puts 'You are not currently signed in.'
+    end
+  rescue StandardError
+    $stderr.puts "File '#{Gem.configuration.credentials_path}' must have readonly permission."\
+                 " Please make sure its writeable."
+  end
```
duckinator: A few things here:

1. Having side effects in a condition (such as `if Gem.configuration.unset_api_key!` attempting to delete a file) can make reasoning about code more difficult.
2. You can use if/elsif/else instead of relying on exceptions.


Also, things inheriting from `Gem::Commands` (such as this) use `say` and `alert`, `alert_warning`, `alert_error` instead of `$stdout.puts` and `$stderr.puts`. 

The documentation for these methods is [available in lib/rubygems/user_interaction.rb](https://github.com/shivabhusal/rubygems/blob/a8fcba0260e10ec5508175dc5356c1ec75d0931e/lib/rubygems/user_interaction.rb#L349-L379).

---

So, taking those into consideration, maybe something like this?

```ruby
def execute
  credentials_path = Gem.configuration.credentials_path

  if !File.exist?(credentials_path) then
    alert_error 'You are not currently signed in.'
  elsif !File.writable?(credentials_path) then
    alert_error "File '#{Gem.configuration.credentials_path}' is read-only."\
                " Please make sure it is writable."
  else
    Gem.configuration.unset_api_key!
    say 'You have successfully signed out from all sessions.'
  end
end
```

(@segiddins: I'm not sure about my wording with "You have successfully signed out from all sessions." above, but my understanding is that that same config file can apply to other servers. Is this correct? And any thoughts on how to phrase this?)

lib/rubygems/config_file.rb
```diff
@@ -336,6 +336,10 @@ def set_api_key host, api_key
     load_api_keys # reload
   end
 
+  def unset_api_key!
+    File.exist?(credentials_path) ? File.delete(credentials_path) : false
```
shivabhusal: Agreed, but using `unless` might create confusion and hamper readability in my opinion? How about writing it in formal `if` - `else` format?

djberg96: Seems pretty clear to me. *shrug*

lib/rubygems/commands/signout_command.rb
shivabhusal: @duckinator done!

lib/rubygems/config_file.rb
bronzdoc: What @duckinator suggest is perfectly clear to me too.

duckinator:
Thanks for the PR!

Code looks good, works as expected, and has tests.

@homu r+

lib/rubygems/commands/signout_command.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+require 'rubygems/command'
+require 'rubygems/commands/query_command'
+
+class Gem::Commands::SignoutCommand < Gem::Commands::QueryCommand
```
simi: I still don't understand why `Gem::Commands::QueryCommand` is used here. instead of `Gem::Command`.

shivabhusal: um, you have point. I have copied code from another command file but forgot to change this portion of code. Thanks for point this out.

duckinator: good catch, @simi! I'm pretty sure you pointed that out earlier and I forgot to make sure it was fixed. 😓 

If nobody gets to it in the meantime, I'll make a PR fixing this either later today or tomorrow.

segiddins: it's already fixed

PR #1635, opened 2016-06-08
Creating Sign In Command

samjonester:
Closes rubygems/rubygems#1634

Sign in command signs the user into a gem's host prior to pushing a gem.
Uses the same sign in mechanism as `PushCommand`.
Can be used to sign into a gem's host prior to running automated tooling
to push gem.

PR #1943, opened 2017-06-05
Add Sign in feature to CLI

shivabhusal:
# Description:
Lets user to login to any gem-cutter compatible package host. But, the default would be rubygems.org
The command will take `--host` as a switch to let use pass the host URL

# Tasks:

- [x] Write tests


simi:
There's already PR for this - https://github.com/rubygems/rubygems/pull/1635/files. It uses GEM to determine host. I'm not sure if that will be something beneficial actually,

On the other side this PR is much simpler.

👍

shivabhusal:
@simi I accidentally closed this PR and modified the history of this branch and now I cannot re-open this.

I have created another PR though. Regarding, the already existing PR, lets see what the community thinks.

PR #1944, opened 2017-06-05
Add command `signin` to `gem` CLI

### LINKED ISSUE #1634: Create Command For Signin
I would like to suggest a feature.

I use automated tooling to push a gem to artifactory. This requires that you first manually push the gem to authenticate with the gem host. This feels like an unnecessary and error prone step, given the automated tooling to push gems in a structured and repeatable manner.

I'd like to propose the command `signin` that would use the same sign in mechanism as the `PushCommand`. This command would log into a gem's host, or host specified by the flag `--host`.

This new command could be done as a part of developer setup, and not expose the potential for an error prone manual push.

shivabhusal:
# Description:
 Lets user to login to any `gem-cutter` compatible package hosts. But, the default would be `rubygems.org`

 The command will take `--host` as a switch to let use pass the host URL

closes #1634 

**Note**: There's already PR for this - https://github.com/rubygems/rubygems/pull/1635/files.
I thought this would be much simpler. What do you guys think? 

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem_commands_signin_command.rb
+++ b/test/rubygems/test_gem_commands_signin_command.rb
@@ -19,28 +19,61 @@ class TestGemCommandsSigninCommand < Gem::TestCase
     File.delete(credentials_path)  if File.exist?(credentials_path)
     super
   end
-
+  
   def test_execute_when_not_already_signed_in
     sign_in_ui = util_capture() { @cmd.execute }
     assert_match %r{Signed in.}, sign_in_ui.output
   end
 
+  def test_execute_when_already_signed_in_with_same_host
+    host            = 'http://some-gemcutter-compatible-host.org'
+    sign_in_ui      = util_capture(nil, host) { @cmd.execute }
+    old_credentials = YAML.load_file Gem.configuration.credentials_path
+    
+    sign_in_ui      = util_capture(nil, host) { @cmd.execute }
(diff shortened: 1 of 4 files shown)


shivabhusal:
@segiddins any comment on this?

segiddins:
I haven't had a chance to look at it yet

test/rubygems/test_gem_commands_signin_command.rb
```diff
@@ -0,0 +1,60 @@
...
+    @cmd = Gem::Commands::SigninCommand.new
+  end
+
+  def teardown
+    credentials_path = Gem.configuration.credentials_path
+    File.delete(credentials_path)  if File.exist?(credentials_path)
+    super
+  end
+
+  def test_execute_when_not_already_signed_in
```
segiddins: can these tests also test that the API key is being set in the credentials file properly?
can we also test what happens when signing into a host you're already signed into?

shivabhusal: sure

shivabhusal: is it enough Samuel?

shivabhusal:
@segiddins

shivabhusal:
@segiddins is there anything else I need to take care of?

segiddins:
Looks 👍

segiddins:
@bundlerbot r+

Issue #1974, opened 2017-07-19
Provide special-cased `bundle` command that executes the correct version of Bundler


indirect:
With Bundler 2 coming, the smoothest possible transition path is to provide a `bundle` command that checks the Gemfile.lock and loads the correct version of Bundler (1.x or 2.0) depending on the version of Bundler listed in the lock file.

Over in Bundler, we spent a lot of effort trying to see if we could do this inside Bundler itself, and it turns out to be extremely problematic to remove an already-loaded library and load a different version of that library instead, from inside that library.

Instead, I'd like to propose that we ship a `bundle` command with RubyGems with the following behavior:

- if BUNDLE_GEMFILE is set, checks the lock for that file to find the bundler version
- if the current directory or parents of the current directory have a gemfile, use the lock to find the bundler version
- if no lock is found, default to the newest version of Bundler currently installed
- load the selected Bundler version and then hand off execution to the Bundler gem itself

segiddins:
Should the same happen when `Kernel.gem "bundler"` is called?

indirect:
I think it's an implementation detail and I don't care either way? Could make it easier for other binstubs that want to load the Bundle before running another gem's command, though.

PR #1977, opened 2017-07-20
Ensure that an appropriate version of Bundler is always activated

segiddins:
# Description:

Closes https://github.com/rubygems/rubygems/issues/1974.
This ensures that the `bundle` binstub, along with anything running `require "bundler/setup"` will respect, in order:

1. `ENV["BUNDLER_VERSION"]`
2. `bundle update --bundler`
3. The bundler version in the nearest `Gemfile.lock`

See https://github.com/bundler/bundler/issues/5876 for further context, and https://github.com/bundler/bundler/issues/5878 for a related change to Bundler.

# Tasks:

- [x] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -254,6 +254,7 @@ hide_lib_for_update/note.txt
 lib/rubygems.rb
 lib/rubygems/available_set.rb
 lib/rubygems/basic_specification.rb
+lib/rubygems/bundler_version_finder.rb
 lib/rubygems/command.rb
 lib/rubygems/command_manager.rb
 lib/rubygems/commands/build_command.rb
@@ -489,6 +490,7 @@ test/rubygems/test_config.rb
 test/rubygems/test_deprecate.rb
 test/rubygems/test_gem.rb
 test/rubygems/test_gem_available_set.rb
+test/rubygems/test_gem_bundler_version_finder.rb
 test/rubygems/test_gem_command.rb
 test/rubygems/test_gem_command_manager.rb
 test/rubygems/test_gem_commands_build_command.rb
--- a/Rakefile
+++ b/Rakefile
@@ -46,6 +46,7 @@ hoe = Hoe.spec 'rubygems-update' do
   spec_extras[:required_rubygems_version] = Gem::Requirement.default
   spec_extras[:required_ruby_version]     = Gem::Requirement.new '>= 1.8.7'
   spec_extras[:executables]               = ['update_rubygems']
+  spec_extras[:homepage]                  = 'https://rubygems.org'
 
   rdoc_locations <<
     'docs-push.seattlerb.org:/data/www/docs.seattlerb.org/rubygems/'
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -271,15 +271,15 @@ module Gem
 
     specs = dep.matching_specs(true)
 
-    raise Gem::GemNotFoundException,
-          "can't find gem #{dep}" if specs.empty?
-
     specs = specs.find_all { |spec|
       spec.executables.include? exec_name
     } if exec_name
 
     unless spec = specs.first
-      msg = "can't find gem #{name} (#{requirements}) with executable #{exec_name}"
+      msg = "can't find gem #{dep} with executable #{exec_name}"
+      if name == "bundler" && bundler_message = Gem::BundlerVersionFinder.missing_version_message
+        msg = bundler_message
+      end
       raise Gem::GemNotFoundException, msg
     end
 
@@ -1334,6 +1334,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
 
   MARSHAL_SPEC_DIR = "quick/Marshal.#{Gem.marshal_version}/"
 
+  autoload :BundlerVersionFinder, 'rubygems/bundler_version_finder'
   autoload :ConfigFile,         'rubygems/config_file'
   autoload :Dependency,         'rubygems/dependency'
   autoload :DependencyList,     'rubygems/dependency_list'
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -133,6 +133,7 @@ module Gem
 
   GEM_DEP_FILES = %w[
     gem.deps.rb
+    gems.rb
     Gemfile
     Isolate
   ]
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -270,7 +270,12 @@ module Gem
 
     return loaded if loaded && dep.matches_spec?(loaded)
 
-    specs = dep.matching_specs(true)
+    find_specs = proc { dep.matching_specs(true) }
+    if dep.to_s == "bundler (>= 0.a)"
+      specs = Gem::BundlerVersionFinder.without_filtering(&find_specs)
+    else
+      specs = find_specs.call
+    end
 
     specs = specs.find_all { |spec|
       spec.executables.include? exec_name
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -66,7 +66,7 @@ To install the missing version, run `gem install bundler:#{vr.first}`
     gemfile = ENV["BUNDLE_GEMFILE"]
     gemfile = nil if gemfile && gemfile.empty?
     Gem::Util.traverse_parents Dir.pwd do |directory|
-      next unless gemfile = %w[gems.rb Gemfile].find { |f| File.file?(f) }
+      next unless gemfile = Gem::GEM_DEP_FILES.find { |f| File.file?(f.untaint) }
 
       gemfile = File.join directory, gemfile
       break
@@ -77,7 +77,7 @@ To install the missing version, run `gem install bundler:#{vr.first}`
     lockfile = case gemfile
     when "gems.rb" then "gems.locked"
     else "#{gemfile}.lock"
-    end
+    end.untaint
 
     return unless File.file?(lockfile)
 
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -1,0 +1,87 @@
+module Gem::BundlerVersionFinder
+  def self.bundler_version
+    version, _reason = bundler_version_with_reason
+
+    return unless version
+
+    Gem::Version.new(version)
+  end
+
+  def self.bundler_version_with_reason
+    if v = ENV["BUNDLER_VERSION"]
+      return [v, "`$BUNDLER_VERSION`"]
+    end
+    if v = bundle_update_bundler_version
+      return if v == true
+      return [v, "`bundle update --bundler`"]
+    end
+    v, lockfile = lockfile_version
+    if v
+      return [v, "your #{lockfile}"]
+    end
+  end
+
+  def self.missing_version_message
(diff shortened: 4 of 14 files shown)


segiddins:
This is green, if anyone would like tor review it

indirect:
I'd like to do some QA testing on it myself before we merge, but this code looks super reasonable. Thanks for this!

indirect:
Hooray, I had time to test this! Here are my notes:

```
$ bundle -v
Bundler version 2.0.0.dev

$ gem list bundler

*** LOCAL GEMS ***

bundler (2.0.0.dev, 1.15.3, default: 1.15.2, 1.15.1, 1.15.0, 1.15.0.pre.4, 1.14.6)

$ cd old-project && bundle -v
Bundler version 1.15.1
The latest bundler is 1.15.3, but you are currently running 1.15.1.
To update, run `gem install bundler`
```

Good: it ran bundler 1.x! Woo! :tada:

Bad: I already have 1.15.3 installed, I shouldn't be seeing that warning.

Worrying: What happens when the newest Bundler is 2.0? Will we say "The latest is 2.0.0, but you are running 1.15.3, update with `gem install bundler`"?

After thinking about it, we probably want to do one or more of these options:

1. Warn on the latest 1.x version, and print the exact command to install that exact version.
2. Warn that 2.0 is the newest, but link to an upgrade guide instead of saying to run gem install.
3. Include instructions on how to disable the 2.0 upgrade warning for this project.

Okay, so, weird that it's refusing to use the newest 1.x Bundler I have installed, let me force it...

```
$ bundle _1.15.3_ -v
/opt/rubies/2.4.1/lib/ruby/site_ruby/2.4.0/rubygems.rb:281:in `find_spec_for_exe': Could not find 'bundler' (2.0.0.dev) required by your /Users/andre/src/rubytogether/rubytogether.org/Gemfile.lock. (Gem::GemNotFoundException)
To update to the lastest version installed on your system, run `bundle update --bundler`.
To install the missing version, run `gem install bundler:2.0.0.dev`
        from /opt/rubies/2.4.1/lib/ruby/site_ruby/2.4.0/rubygems.rb:300:in `activate_bin_path'
        from /Users/andre/.gem/ruby/2.4.1/bin/bundle:23:in `<main>'
```

Uhh... hmm. Maybe if I run `bundle update --bundler 1.15.3`?

```
$ bundle update --bundler 1.15.3
Fetching gem metadata from https://rubygems.org/..
Resolving dependencies.........
Warning: the lockfile is being updated to Bundler 2, after which you will be unable to return to Bundler 1.
Bundle updated!
Gems in the group mysql were not installed.

$ bundle -v
Bundler version 2.0.0.dev
```

Nope, I guess not. :/



So! Fixes that I think are probably needed before shipping this:

- [] don't tell people to gem install bundler for a version of bundler that is already installed
- [] when the user runs `bundle update --bundler VERSION`, install VERION and update the lockfile to VERSION
- [] do not resolve or install the bundle when a user runs `bundle update --bundler` (they might be in a state where the bundle can't be installed until they change versions, for example)
- [] use the locked bundler version to decide between bundler `~> 1.a` or `~> 2.a`, but no more specific than that
- [] load and run bundler VERSION without exceptions when a user runs `bundle _VERSION_`

segiddins:
> don't tell people to gem install bundler for a version of bundler that is already installed

requires a change in Bundler to check the installed gems instead of just looking at `Bundler::VERSION`

> when the user runs bundle update --bundler VERSION, install VERION and update the lockfile to VERSION

already worked when passed `--bundler=VERSION` or `--bundler`, but I've fixed this so it works with the separate arg as well

> do not resolve or install the bundle when a user runs bundle update --bundler (they might be in a state where the bundle can't be installed until they change versions, for example)

What is the desired behavior for `bundle update --bundler` then? Right now it just unlocks the bundler version

>  use the locked bundler version to decide between bundler ~> 1.a or ~> 2.a, but no more specific than that

So it should only restrict the _major_ version, rather than using the exact version locked in the lockfile?
I personally find that to be counter-intuitive

> load and run bundler VERSION without exceptions when a user runs bundle _VERSION_

I can probably make this happen, but only for the latest set of binstubs (the really old RG binstubs called `gem` instead of `Gem.activate_bin_path`)

matthewd:
>> use the locked bundler version to decide between bundler ~> 1.a or ~> 2.a, but no more specific than that
>
> So it should only restrict the major version, rather than using the exact version locked in the lockfile? I personally find that to be counter-intuitive

I'd suggest it's more like it's detecting the lockfile *format* [which happens to be designated by the major version locked in the lockfile], then using the release-series that suits. AIUI the goal here is to smooth the 1.x to 2.x transition without requiring those versions to be directly compatible -- not to implement "strong" versioning controls for bundler itself.


> I can probably make this happen, but only for the latest set of binstubs (the really old RG binstubs called gem instead of Gem.activate_bin_path)

Could/should a rubygems update automatically `gem pristine --all --only-executables`? I've already seen instances of someone having encountered an old-binstubs problem, upgraded rubygems, then been confused (or mislead about the cause) when observing that the problem persisted. Perhaps such a change could solve both things?

indirect:
> What is the desired behavior for bundle update --bundler then?

I think we want it to change the version of Bundler listed in the lockfile, and that's it? The problem I was having was that `bundle install` would fail until I could change the version of Bundler... but `bundle update --bundler` would try to unlock Bundler and then run `update`, which resulted in a failure.

Somewhat related, what are the chances that we can make `bundle update bundler` work, without the specific flag? I guess the problem then is that there's no way to provide a version number. :/

> So it should only restrict the major version, rather than using the exact version locked in the lockfile?

For Bundler 1.x vs 2.x, yes, it should only restrict the major version. This is consistent with the behavior of Bundler 1.x so far, and I think we need to do it to meet people's expectations about how Bundler 1.x will work.

In the future, for Bundler 2.x versions, I think we should explicitly change the behavior, and use the locking code we've got working. Bundler 2.x will use the exact version that the Bundle is locked to, with opt-in upgrades per Bundle.

> I can probably make this happen, but only for the latest set of binstubs (the really old RG binstubs called gem instead of Gem.activate_bin_path)

That's fine, let's do that. I don't think it's okay to ship a `bundle` command that stops being able to select versions at the exact moment that selecting versions becomes (possibly) critically important.

> the goal here is to smooth the 1.x to 2.x transition without requiring those versions to be directly compatible -- not to implement "strong" versioning controls for bundler itself.

Yep, pretty much that ↑ The 1.x releases need a smooth transition, and we can provide strong versioning controls for Bundler in the 2.x series.

> Could/should a rubygems update automatically gem pristine --all --only-executables?

Yeah, I think this is a really good idea. @segiddins, how do you feel about a rubygems update rebuilding all the binstubs so that we get guaranteed-good ones?

segiddins:
> how do you feel about a rubygems update rebuilding all the binstubs so that we get guaranteed-good ones?

I think that might be a breaking change? Not sure

segiddins:
@indirect think I've addressed all of your feedback, please take a look!

indirect:
@segiddins will check this out this afternoon, thanks!

indirect:
✅ load and run bundler VERSION without exceptions when a user runs bundle _VERSION_
❌ don't tell people to gem install bundler for a version of bundler that is already installed

```
$ gem list bundler

*** LOCAL GEMS ***

bundler (2.0.0.dev, 1.15.4, 1.15.3, default: 1.15.2, 1.15.1, 1.15.0, 1.15.0.pre.4, 1.14.6)

$ bundle -v
Bundler version 1.15.4

$ bundle _1.15.0_ -v
Bundler version 1.15.0
The latest bundler is 1.15.4, but you are currently running 1.15.0.
To update, run `gem install bundler`
```

✅ use the locked bundler version to decide between bundler `~> 1.a` or `~> 2.a`, but no more specific than that
❌  do not resolve or install the bundle when a user runs `bundle update --bundler` (they might be in a state where the bundle can't be installed until they change versions, for example)
❌  when the user runs `bundle update --bundler VERSION`, install VERSION and update the lockfile to VERSION

```
$ bundle update --bundler 1.15.0
To update, run `gem install bundler`
Fetching gem metadata from https://rubygems.org/............
Fetching version metadata from https://rubygems.org/..
Fetching dependency metadata from https://rubygems.org/.
Resolving dependencies...
[...]
Bundle updated!

$ git diff
diff --git a/Gemfile.lock b/Gemfile.lock
index 9d89827..62808bf 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -230,4 +230,4 @@ RUBY VERSION
    ruby 2.4.1p111

 BUNDLED WITH
-   1.15.1
+   1.15.4
```

After testing this out, I think we shouldn't actually completely remove the "you should update bundler" message. I think we should change the message from this:

> The latest bundler is 1.15.4, but you are currently running 1.15.0.
> To update, run `gem install bundler`

to this:

> The latest Bundler version is 1.15.4, but this application is locked to 1.15.3.
> To update this application, run `bundle update --bundler`.

(As you might guess, running `bundle update --bundler` should install a new version of Bundler if needed, before updating the gemfile to lock to that latest version.)

What do you think?

segiddins:
> don't tell people to gem install bundler for a version of bundler that is already installed 

Done in https://github.com/bundler/bundler/pull/5910

> do not resolve or install the bundle when a user runs bundle update --bundler (they might be in a state where the bundle can't be installed until they change versions, for example)

> when the user runs bundle update --bundler VERSION, install VERSION and update the lockfile to VERSION

Should be handled by https://github.com/bundler/bundler/pull/5956

> (As you might guess, running bundle update --bundler should install a new version of Bundler if needed, before updating the gemfile to lock to that latest version.)

Highly disagree, I don't like automatic installation, I think it makes for a worse UX when it fails

indirect:
> Highly disagree, I don't like automatic installation, I think it makes for a worse UX when it fails

Running `bundle update` downloads and installs the latest version of a gem. It doesn't seem weird (or like "automatic installation") to me for `bundle update --bundler` to download and install the latest version of Bundler?

But that aside, I think the real problem is that we are creating a new problem for updating Bundler: once the Bundler version, is locked by the lockfile, it's possible for the _installed_ version to be out of date, and it's also possible for the _locked_ version to be out of date. Do you think we should report on those two cases separately? The message above was my attempt to collapse both cases into a single prompt, with a single command to update both the installed and locked versions.

segiddins:
I think my Pr I linked above does report the two cases separately by checking the Bundler versions that are installed?

indirect:
@segiddins cool! can you show what each of the options looks like?

- outdated app & equally outdated install
- outdated app & less outdated install
- outdated app & up-to-date install
- up-to-date app & up-to-date install

segiddins:
| running version | latest released version | latest installed version | message                                                                                                                                 |
|-----------------|-------------------------|--------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|
| 1               | 2                       | 1                        | The latest bundler is 2, but you are currently running 1. To install the latest version, run `gem install bundler`                      |
| 1               | 2                       | 2                        | The latest bundler is 2, but you are currently running 1. To update to the most recent installed version, run `bundle update --bundler` |
| 1               | 2                       | 1.1                      | The latest bundler is 2, but you are currently running 1. To update to the most recent installed version, run `bundle update --bundler` |
| 2               | 2                       | 2                        |                                                                                                                                         |

indirect:
I'm not completely sold on the Bundler installation and messaging, but I think we can deal with that over in Bundler. Thanks for working on this! @bundlerbot r+

PR #1978, opened 2017-07-31
better error msg if url scheme missing for proxy

### LINKED ISSUE #1783: Error message for an invalid proxy is a bit misleading

My current problem is:

The error message for an incorrect http-proxy parametr is misleading

```
gem install rails -p=foo.bar.com
ERROR:  While executing gem ... (OptionParser::InvalidArgument)
    invalid argument: -p=foo.bar.com
```

I spent a while trying to figure out if I misspelt the parameter or something, before eventually realising it needs the http/https at the start.

It would be nicer if it could specify that 👍 

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [ ] Other

```
$ gem env version
2.6.7
```

But I assume it affects all gem versions

ajwann:
# Description:

fixes #1783

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/local_remote_options.rb
+++ b/lib/rubygems/local_remote_options.rb
@@ -24,8 +24,10 @@ module Gem::LocalRemoteOptions
         raise OptionParser::InvalidArgument, value
       end
 
-      unless ['http', 'https', 'file', 's3'].include?(uri.scheme)
-         raise OptionParser::InvalidArgument, value
+      allowed_schemes = ['http', 'https', 'file', 's3']
+      unless allowed_schemes.include?(uri.scheme)
+        msg =  "#{value}\nPreface URLs with one of #{allowed_schemes}"
+        raise OptionParser::InvalidArgument, msg
       end
 
       value
--- a/test/rubygems/test_gem_commands_install_command.rb
+++ b/test/rubygems/test_gem_commands_install_command.rb
@@ -1004,4 +1004,13 @@ ERROR:  Possible alternatives: non_existent_with_hint
     assert_equal [:test, :development], @cmd.options[:without_groups]
   end
 
+  def test_handle_proxy_with_no_scheme
+    err = assert_raises OptionParser::InvalidArgument do
+      @cmd.handle_options %w[-p=foo.bar.com]
+    end
+
+    assert_match /invalid argument: -p=foo.bar.com/ , err.message
+    assert_match /Preface URLs with one of \["http", "https", "file", "s3"\]/ , err.message
+  end
+
 end


test/rubygems/test_gem_commands_install_command.rb
```diff
@@ -1004,4 +1004,13 @@ def test_handle_options_without
     assert_equal [:test, :development], @cmd.options[:without_groups]
   end
 
+  def test_handle_proxy_with_no_scheme
+    err = assert_raises OptionParser::InvalidArgument do
+      @cmd.handle_options %w[-p=foo.bar.com]
+    end
+
+    assert_match /invalid argument: -p=foo.bar.com/ , err.message
+    assert_match /Preface URLs with one of \["http", "https", "file", "s3"\]/ , err.message
```
ajwann: I've written what I think is a straightforward test, I just want to throw an exception and test that the message I get back contains the original message + the new bit. 

I was able to test my change in ```lib/rubygems/local_remote_options.rb``` manually, and it works. However, the second assertion (line 1013) doesn't evaluate to true. It's almost as if a mock/test double is being used, but if that's the case I can't find it.

If someone could lend me a helpful hand in getting this test to run correctly, I'd really appreciate it.

segiddins:
Thanks for the PR! It looks like this is currently failing a single test: https://travis-ci.org/rubygems/rubygems/jobs/259582765#L771

lib/rubygems/local_remote_options.rb
```diff
@@ -24,8 +24,10 @@ def accept_uri_http
         raise OptionParser::InvalidArgument, value
       end
 
-      unless ['http', 'https', 'file', 's3'].include?(uri.scheme)
-         raise OptionParser::InvalidArgument, value
+      allowed_schemes = ['http', 'https', 'file', 's3']
+      unless allowed_schemes.include?(uri.scheme)
+        msg =  "#{value}\nPreface URLs with one of #{allowed_schemes}"
+        raise OptionParser::InvalidArgument, msg
```
ajwann: @segiddins It looks like I was wrong, and my fix isn't working. Upon further testing, it looks like any message that's passed on line 30 when ```OptionParser::InvalidArgument``` is raised is completely discarded, and replaced with whatever ```value``` is, since it's the iterator in the block.

I'm currently experimenting with trying to get any custom error message to be returned.

ajwann: It looks like even if I try to create my own custom ```OptionParser::InvalidArgument``` object it still returns whatever is in the ```value``` iterator.

```ruby
err = OptionParser::InvalidArgument.new('custom message')
raise err
```

The above attempt still results in the error message just being whatever ```value``` was when the exception was raised.

ajwann: I'm able to get back a custom error message if I don't use an instance of ```OptionParser::InvalidArgument```. I'll just throw a standard runtime error.

ajwann:
I'm closing this until I can get it right.

test/rubygems/test_gem_commands_install_command.rb
samir-majhi: @ajwann Have you tried printing both the exception that you've created as well as the string that you're comparing the assert with to see if there might be a typo?

PR #1979, opened 2017-08-02
Added message to uninstall command for gem that is not installed.

anantkolvankar:
When we try to uninstall a gem that's not installed, we don't get any message.

Assuming I don't have `day_greeter` gem installed in my machine.
> [anant@localhost ~]# gem uninstall day_greeter
> [anant@localhost ~]# 

with below changes it will be like this
> [anant@localhost ~]# gem uninstall day_greeter
> Gem 'day_greeter' is not installed
> [anant@localhost ~]#


segiddins:
👍 with a test

segiddins:
Awesome, thanks!
@bundlerbot r+

Issue #1980, opened 2017-08-04
A test fails when the sort order is unstable.


ko1:
I found test failure (a test included by ruby-trunk) on mswin64 (VS2017, Windows 10) environment.
and I found that if `Gem::Resolver#sort_dependencies()` returns stable result, no problem.

I fixed rubygems on ruby-trunk but I'm not sure it is common case or not. Please check by experts.

See https://svn.ruby-lang.org/cgi-bin/viewvc.cgi?view=revision&revision=59502 for details of my fix.

segiddins:
Is `Enumerable#sort_by` not stable? Adding in the extra element to the sort is probably an unnecessary performance hit

MSP-Greg:
@ko1,

The fix to rubygems on ruby-trunk fixes the issue for MinGW builds also.  This is one of a handful of failures left to be fixed.

See Test Failure - TestGemRequestSetLockfile#test_to_s_gem_dependency_non_default #1948 

Thanks ko1...

@segiddins 

> Adding in the extra element to the sort is probably an unnecessary performance hit

Like a blink of an eye vs reading your sentence...

hsbt:
I already merged @ko1 's patch in RG 2.7.

PR #1981, opened 2017-08-05
Picked commit from ruby/ruby@41e1670

hsbt:
# Description:

Picked commit from ruby/ruby@41e1670

This is patch for https://github.com/rubygems/rubygems/issues/1980

```
* lib/rubygems/resolver.rb (sort_dependencies): use stable sort.
  TestGemRequestSetLockfile#test_to_s_gem_dependency_non_default
  fails because this method return unstable results.
  Note that Enumerable#sort_by is unstable.

  I'm not sure the "stable" nature is required for RubyGems.
  The fact is that using stable sort, the test passed on
  mswin64+VS2017 where the sort results was reverse (unstable) order.
  Also using `-i` instead of `i` (it means forcing unstable sort)
  this test fails on other platform where the test successed before.
```

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/resolver.rb
+++ b/lib/rubygems/resolver.rb
@@ -275,13 +275,14 @@ class Gem::Resolver
   end
 
   def sort_dependencies(dependencies, activated, conflicts)
-    dependencies.sort_by do |dependency|
+    dependencies.sort_by.with_index do |dependency, i|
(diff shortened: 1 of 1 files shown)


segiddins:
@bundlerbot r+

PR #1985, opened 2017-08-08
Small style fixes to Installer Set

arthur-marz:
[-] Remove redundant comma from InstallerSet#local?

### DIFF
--- a/lib/rubygems/resolver/installer_set.rb
+++ b/lib/rubygems/resolver/installer_set.rb
@@ -193,7 +193,7 @@ class Gem::Resolver::InstallerSet < Gem::Resolver::Set
   # Has a local gem for +dep_name+ been added to this set?
 
   def local? dep_name # :nodoc:
-    spec, = @local[dep_name]
+    spec = @local[dep_name]
 
     spec
   end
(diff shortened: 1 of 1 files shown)


lib/rubygems/resolver/installer_set.rb
```diff
@@ -193,7 +193,7 @@ def load_spec name, ver, platform, source # :nodoc:
   # Has a local gem for +dep_name+ been added to this set?
 
   def local? dep_name # :nodoc:
-    spec, = @local[dep_name]
+    spec = @local[dep_name]
```
bronzdoc: Hey, thanks for contributing to rubygems :)

I don't think that comma is just for style purposes, that comma there ensure we get the spec part from `@local[dep_name]`
since `@local[dep_name]` can be equal to `[spec, source]`

arthur-marz: Thank you for your comment. Maybe you can point on some places in the rubygems that can be useful for the newbie contributor like me to help this gem out?

seanlinsley: If that's the case, this might be more clear:
```ruby
spec, _source = @local[dep_name]
```
Or:
```ruby
spec, _ = @local[dep_name]

bronzdoc:
Thanks!

@homu r+

bronzdoc:
@bundlerbot r+

PR #1998, opened 2017-08-29
Temporarily disable Bundler integration

segiddins:
Right now, I've been avoiding shipping `master` as RubyGems 2.7 for almost a year because (1) it relies on Bundler for `use_gemdeps` functionality (purposefully, as a part of the Bundler/RubyGems merger plan) and (2) Ruby core is not yet able to vendor a version of RubyGems that depends upon Bundler because they do not yet have they infrastructure in place to test Bundler (with rspec, etc) along with core.

This PR temporarily disables the Bundler integration, unblocking the release of 2.7. Once @hsbt gets Ruby core ready for a RubyGems that uses Bundler, we can flip the constant and ship a new minor version.

### DIFF
--- a/lib/rubygems/util.rb
+++ b/lib/rubygems/util.rb
@@ -109,26 +109,15 @@ module Gem::Util
   ##
   # Enumerates the parents of +directory+.
 
-  def self.traverse_parents directory
+  def self.traverse_parents directory, &block
     return enum_for __method__, directory unless block_given?
 
     here = File.expand_path directory
-    start = here
-
-    Dir.chdir start
-
-    begin
-      loop do
-        yield here
-
-        Dir.chdir '..'
-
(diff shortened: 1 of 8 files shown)


lib/rubygems/util.rb
```diff
@@ -109,26 +109,15 @@ def self.silent_system *command
   ##
   # Enumerates the parents of +directory+.
 
-  def self.traverse_parents directory
+  def self.traverse_parents directory, &blk
```
simi: I think `&block` is used across codebase.

indirect:
Excellent. @bundlerbot r+

hsbt:
@segiddins @indirect Confirmed. I will ping about this feature after ruby core integrates bundler.

PR #2006, opened 2017-09-05
Require digest when it is used

segiddins:
This avoids a potential issue where Gem.activate_bin_path could deadlock, since we'd attempt to activate the digest gem in the middle of activating other gems

This fixes 8 of the ten issues surfaced at https://travis-ci.org/bundler/bundler/jobs/271910850#L3531


segiddins:
It's been a week, merging since this fixes deadlocks

@bundlerbot r+

PR #2009, opened 2017-09-11
update gem yank description

dwradcliffe:
Now that gem yank permanently removes the gem we should not direct people to contact us to remove a gem.


segiddins:
👍, we just need to make trunk an allowed failure on appveyor

segiddins:
@bundlerbot r+

PR #2010, opened 2017-09-11
matches_for_glob - remove root path

ahorek:
# Description:
Hi,
during my testing I found out that this method uses a wierd glob pattern.

```
  def matches_for_glob glob # TODO: rename?
    # TODO: do we need these?? Kill it
    glob = File.join(self.lib_dirs_glob, glob)

    Dir[glob].map { |f| f.untaint } # FIX our tests are broken, run w/ SAFE=1
  end
```

example:
c:/rubyinstaller/sandbox/ruby25_mingw/lib/ruby/gems/2.5.0/gems/bigdecimal-1.3.2/{**c:/rubyinstaller/sandbox/ruby25_mingw/lib/ruby/gems/2.5.0/extensions/x64-mingw32/2.5.0/bigdecimal-1.3.2**,lib}

I think it's a bug because the absolute path in the pattern will never be matched.

I attached a PR that removes absolute paths from the pattern. What do you think?
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -275,10 +275,10 @@ class Gem::BasicSpecification
   # for this spec.
 
   def lib_dirs_glob
-    dirs = if self.require_paths.size > 1 then
-             "{#{self.require_paths.join(',')}}"
+    dirs = if self.raw_require_paths.size > 1 then
+             "{#{self.raw_require_paths.join(',')}}"
            else
-             self.require_paths.first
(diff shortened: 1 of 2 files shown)


segiddins:
Can you please add a test for this change? Thanks!

ahorek:
done

segiddins:
Cool, thanks!

hsbt:
@bundlerbot retry

PR #2011, opened 2017-09-13
Allow the require bundler specs to pass when bundler is installed as a default gem

segiddins:
This should fix the specs when running against ruby trunk, as the load path that contains bundler no longer ends in `bundler/lib`, as it did when running with `-I bundler/lib` (e.g. its in with the ruby stdlib)


hsbt:
@bundlerbot r+

PR #2012, opened 2017-09-13
Re-enable using Bundler for gemdeps

segiddins:
# Description:

Since @hsbt told us that we're ready to release v2.7 _including the use of Bundler_, we can safely re-enable using Bundler!

PR #2013, opened 2017-09-14
Renamed test fixture name on TestRequire

hsbt:
# Description:

@segiddins 

I renamed fixture name to `a.rb` from `setup.rb`.  Because Ruby 2.5 contains "lib/bundler/setup.rb" with real code of bundler. I think that it name is enough to test case.

Related with https://github.com/rubygems/rubygems/pull/2012

______________

# Tasks:

- [x] Write tests


segiddins:
I think #2011 would allow us to continue testing with the real file name?

hsbt:
Ah, #2011 resolve problem of this pull request. I think #2011 is better than my patch.

PR #2027, opened 2017-10-06
Rename `-rubygems` option. 

hsbt:
This option only needs Ruby 1.8. and It will remove at Ruby 2.5.

This commit was picked r60125 from ruby/ruby:  https://github.com/ruby/ruby/commit/9de6c712b66aad77df40661c1fc6d37e9a5c251a


segiddins:
@bundlerbot r+

PR #2028, opened 2017-10-07
ubygems.rb is unavailable on Ruby 2.5

hsbt:
Follow up with https://github.com/rubygems/rubygems/pull/2027


hsbt:
@bundlerbot r+

PR #2031, opened 2017-10-08
Do not index the doc folder in the `update_manifest` task.

colby-swandale:
# Description:

A small problem I have come across while doing work on RubyGems is when I need to update the Manifest.

When running the `rake newb` - the task generates the RubyGems docs which is an entry in `.gitignore` but is not ignored in the `rake update_manifest` task. This means that everyone has to remember to either remove the `doc` folder beforehand or remove the entries in `Manifest.txt`. We should just be ignoring the `doc` folder entirely.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -95,7 +95,7 @@ Hoe::Package.instance_method(:install_gem).tap do |existing_install_gem|
   end
 end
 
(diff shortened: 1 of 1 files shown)


hsbt:
@bundlerbot r+

indirect:
@bundlerbot retry

PR #2034, opened 2017-10-11
[SafeYAML] Avoid warning when Gem::Deprecate.skip is set

segiddins:
# Description:

This will allow Bundler to suppress the warnings, so we can get the tests passing again

# Tasks:

- [ ] Write tests


duckinator:
It'll just fail if I r+ it right now bc it relies on the Bundler PR in order to complete the fix, but this looks good to me.

olleolleolle:
@bundlerbot r+

hsbt:
@bundlerbot retry

segiddins:
@bundlerbot retry

PR #2035, opened 2017-10-11
Refactor api_endpoint to be more explicit

claudijd:
# Description:

The input validation checks on SRV api endpoint discovery are sane, but it's usage afterwards could be better.  This PR incorporates feedback from david@bamsoftware.com, to make this even safer and more explicit how we're handling this potentially unsafe input that helps with API discovery.
______________

# Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -109,9 +109,12 @@ class Gem::RemoteFetcher
       uri
     else
       target = res.target.to_s.strip
+      target_host = URI("https://" + target).host
 
-      if URI("http://" + target).host.end_with?(".#{host}")
-        return URI.parse "#{uri.scheme}://#{target}#{uri.path}"
+      if target_host.end_with?(".#{host}")
+        uri = uri.dup
+        uri.host = target_host
+        return uri
       end
 
       uri
--- a/test/rubygems/test_gem_remote_fetcher.rb
+++ b/test/rubygems/test_gem_remote_fetcher.rb
@@ -196,6 +196,21 @@ PeIQQkFng2VVot/WAQbv3ePqWq07g1BBcwIBAg==
     dns.verify
   end
 
+  def test_api_endpoint_changes_host_only
+    uri = URI.parse "http://example.com/foo"
+    target = MiniTest::Mock.new
+    target.expect :target, "user:pass@gems.example.com:8000/path?query=value#fragment"
+
+    dns = MiniTest::Mock.new
+    dns.expect :getresource, target, [String, Object]
+
+    fetch = Gem::RemoteFetcher.new nil, dns
+    assert_equal URI.parse("http://gems.example.com/foo"), fetch.api_endpoint(uri)
+
+    target.verify
+    dns.verify
+  end
+
   def test_api_endpoint_ignores_trans_domain_values
     uri = URI.parse "http://gems.example.com/foo"
     target = MiniTest::Mock.new


lib/rubygems/remote_fetcher.rb
```diff
@@ -109,9 +109,12 @@ def api_endpoint(uri)
       uri
     else
       target = res.target.to_s.strip
+      target_host = URI("https://" + target).host
 
-      if URI("http://" + target).host.end_with?(".#{host}")
-        return URI.parse "#{uri.scheme}://#{target}#{uri.path}"
```
claudijd: Note that previously we incorporated the unparsed target content without it being host trimmed.  The this PR simply makes this much more explicit and, to me, it feels safer.

lib/rubygems/remote_fetcher.rb
```diff
@@ -109,9 +109,12 @@ def api_endpoint(uri)
...
+      if target_host.end_with?(".#{host}")
+        uri = uri.dup
+        uri.host = target_host
+        return uri
```
segiddins: this `return uri` is redundant it's already returned 3 lines below

claudijd: Now that i think about it more, I'm thinking the assignment of the #dup call should be something other than uri, like modified_uri, to make it more clear that we're returning something other than the original uri.  That probably makes this more relevant and less likely to stop the uri variable name in this context.

refashioned:
It occurs to me that a better/simpler way to handle this would be to call [URI::Generic.build](https://ruby-doc.org/stdlib-2.4.3/libdoc/uri/rdoc/URI/Generic.html#method-c-build) to build the URI, rather than string interpolation. Roughly, replace this:
```
      if URI("http://" + target).host.end_with?(".#{host}")
        return URI.parse "#{uri.scheme}://#{target}#{uri.path}"
      end
```
with this (plus error handling):
```
      if target.end_with?(".#{host}")
        return URI::Generic.build :scheme=>uri.scheme, :host=>target, :path=>uri.path
      end
```
Alternatively, call `URI::Generic.build` first, and then compare its `.host` against `".#{host}"`.

`URI::Generic.build` ensures that there's no interference between different components of the URI. `target` can only go into the host component, and if it has any bytes that would make that impossible, `URI::Generic.build` raises an error.

This example program shows `URI::Generic.build` rejecting all sorts of fishy targets which were the reason for opening this pull request. Note that the current code accepts all the dodgy targets.
```
require 'uri'
scheme = "https"
path = "/path"
[
  "example.com",
  "example.com:9999",
  "user:pass@example.com",
  "example.com/bad?q=",
  "example.com/bad#frag",
].each do |target|
  uri1 = URI.parse("#{scheme}://#{target}#{path}").to_s
  begin
    uri2 = URI::Generic.build(:scheme=>scheme, :host=>target, :path=>path).to_s
  rescue URI::InvalidComponentError => err
    uri2 = "##{err.class}"
  end
  puts "%-24s  %-34s  %s" % [target.inspect, uri1, uri2]
end
```
`target`                 | string interpolation               | `URI::Generic.build`
-------------------------|------------------------------------|----------------------------
"example.com"            | https://example.com/path           | https://example.com/path
"example.com:9999"       | https://example.com:9999/path      | #URI::InvalidComponentError
"user:pass@example.com"  | https://user:pass@example.com/path | #URI::InvalidComponentError
"example.com/bad?q="     | https://example.com/bad?q=/path    | #URI::InvalidComponentError
"example.com/bad#frag"   | https://example.com/bad#frag/path  | #URI::InvalidComponentError

P.S. if someone can make https://hackerone.com/reports/274267 public, it explains more about the rationale for this change.

claudijd:
@refashioned I'll request the bug be publicly disclosed, these changes are non-exploitable to our knowledge, but more an improvement on the existing pattern to be more parnoid in how we handle these.  Please also note that we need to consider the contexts of CVE-2015-3900, CVE-2015-4020, CVE-2017-0902 to avoid a regression here, which is effectively an RCE outcome if we get it wrong.

refashioned:
#2418 would remove `api_endpoint` completely and make this pull request moot. IMO #2418 is better than this pull request, for security anyway.

claudijd:
I agree with @refashioned that removing the dynamic capability is most certainly safer/healthier IMO.  I'm going to close this PR in favor of #2418, this mimics the ask we made in 2015, when we found and disclosed CVE-2015-3900 that started this saga.

PR #2061, opened 2017-11-04
Add --[no-]check-development option to cleanup command

godfat:
Default to check so it keeps backward compatibility.

# Description:

I just realized that there's no way to force cleaning up gems without checking development dependency. I wish at least we have an option:

```
gem cleanup --no-check-development
```

This is a bit too verbose to my taste, but to make it backward compatible and consistent with the uninstall command, this is probably the way to do it.

Thanks for reviewing.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/cleanup_command.rb
+++ b/lib/rubygems/commands/cleanup_command.rb
@@ -8,11 +8,18 @@ class Gem::Commands::CleanupCommand < Gem::Command
   def initialize
     super 'cleanup',
           'Clean up old versions of installed gems',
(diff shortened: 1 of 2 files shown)


godfat:
@bronzdoc @segiddins Are there any actions I need to take? Do we need more reviews? How do I move this forward?

Thank you for looking into this!

segiddins:
@bundlerbot r+

hsbt:
@bundlerbot retry

godfat:
Thanks everyone!

Issue #2039, opened 2017-10-14
Add which gem has an invalid gem referred to (invalid == no longer existing gem, in this context here)


shevegen:
Hello rubygems maintainers,

Just now, on a new machine, I am testing my gems from a remote installation. One gem called "rbt":

> gem install rbt

The above is the command I use.

The next is the result I get:

ERROR:  Could not find a valid gem 'move_file' (>= 0) in any repository
ERROR: Possible alternatives: moveable, save_file, wavefile, zonefile, bookfile

Now, I know to some extent as to why the above error exists. I used to have
a gem called "move_file" but since then I integrated it into my larger tool-kit
gem called "roebe".

The thing is - right now I do not know which faulty package references to
the (no longer existing gem) called "move_file".

Now I come to my proposal:

- Would it be possible to show which particular gem referenced this 
gem? That way I could very easily see which gem has a broken 
dependency. It is a bit hard for me to know which gem has a stray
missing dependency in .gemspec, also because some gemspec
files are autogenerated or reference to another dependency 
file. :\

The above message does not have to change much - either a new
line that says which gems is flawed, or to integrate this into the
currently used messages, such as:

"ERROR:  Could not find a valid gem 'move_file' (>= 0) in any repository (referenced by the gem foobar"

But a new message may probably be better; if more than one gem
has this missing dependency, then all could be shown, but I am also
perfectly fine if only one gem is shown at a time. Then I can fix it
and try gem install again.

Hopefully others may find this also useful. Thanks for reading, please
feel free to close the issue at any moment in time.

bronzdoc:
https://github.com/rubygems/rubygems/pull/2067 closes this

PR #2067, opened 2017-11-06
Show which gem referenced a missing gem

akhramov:
# Description:
If gem's dependency is missing, the output just shows the missing gem,
leaving the dependent gem obscure. (#2039)

This change adds the dependent gem to the output message.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -152,14 +152,25 @@ class Gem::Command
   #--
   # TODO: replace +domain+ with a parameter to suppress suggestions
 
-  def show_lookup_failure(gem_name, version, errors, domain)
+  def show_lookup_failure(gem_name, version, errors, domain, required_by = nil)
+    gem = "'#{gem_name}' (#{version})"
+
+    msg = StringIO.new "Could not find a valid gem #{gem}".dup, "a+"
+
     if errors and !errors.empty?
-      msg = "Could not find a valid gem '#{gem_name}' (#{version}), here is why:\n".dup
+      msg << ", here is why:\n"
       errors.each { |x| msg << "          #{x.wordy}\n" }
-      alert_error msg
     else
-      alert_error "Could not find a valid gem '#{gem_name}' (#{version}) in any repository"
+      if required_by && gem != required_by then
+        msg << " (required by #{required_by}) in any repository"
+      else
+        msg << " in any repository"
+      end
     end
+
+    msg.close
+
+    alert_error msg.string
 
     unless domain == :local then # HACK
       suggestions = Gem::SpecFetcher.fetcher.suggest_gems_from_name gem_name
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -154,23 +154,20 @@ class Gem::Command
 
   def show_lookup_failure(gem_name, version, errors, domain, required_by = nil)
     gem = "'#{gem_name}' (#{version})"
-
-    msg = StringIO.new "Could not find a valid gem #{gem}".dup, "a+"
+    msg = String.new "Could not find a valid gem #{gem}"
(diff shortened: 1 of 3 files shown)


lib/rubygems/command.rb
```diff
@@ -152,13 +152,21 @@ def execute
...
     else
-      alert_error "Could not find a valid gem '#{gem_name}' (#{version}) in any repository"
+      if required_by && gem != required_by then
+        msg = "Could not find a valid gem #{gem} required by #{required_by} in any repository"
+      else
+        msg = "Could not find a valid gem #{gem} in any repository"
+      end
+
+      alert_error msg
     end
```
duckinator: I think `"Could not find a valid gem #{gem} (required by #{required_by}) in any repository"` may be a bit clearer.

Since both paths end with `alert_error msg`, would something like this work?:

```ruby
    if errors and !errors.empty?
      msg = "Could not find a valid gem #{gem}, here is why:\n".dup
      errors.each { |x| msg << "          #{x.wordy}\n" }
    elsif required_by && gem != required_by then
      msg = "Could not find a valid gem #{gem} (required by #{required_by}) in any repository"
    else
      msg = "Could not find a valid gem #{gem} in any repository"
    end
    alert_error msg
```

duckinator:
Left a comment with some suggestions! 🙂 

Also, looks like there's some tests failures, but I'm not sure what's causing those.

lib/rubygems/command.rb
akhramov: Awesome tips, thanks!

lib/rubygems/commands/install_command.rb
```diff
@@ -250,17 +250,22 @@ def install_gems # :nodoc:
...
         exit_code |= 1
-      rescue Gem::GemNotFoundException, Gem::UnsatisfiableDependencyError => e
-        domain = options[:domain]
-        domain = :local unless options[:suggest_alternate]
+      rescue Gem::GemNotFoundException => e
         show_lookup_failure e.name, e.version, e.errors, domain
 
+        exit_code |= 2
+      rescue Gem::UnsatisfiableDependencyError => e
+        show_lookup_failure(e.name, e.version, e.errors, domain,
```
colby-swandale: Can you remove the round brackets so we're consistent with everything

akhramov: Sure thing, done.

lib/rubygems/command.rb
```diff
@@ -152,15 +152,22 @@ def execute
   #--
   # TODO: replace +domain+ with a parameter to suppress suggestions
 
-  def show_lookup_failure(gem_name, version, errors, domain)
+  def show_lookup_failure(gem_name, version, errors, domain, required_by = nil)
+    gem = "'#{gem_name}' (#{version})"
+
```
colby-swandale: `msg` should be getting initialized here instead and then getting appended to. IMO this is cleaner as you're only creating `msg` once.

example: 

```
msg = StringIO.new

if errors and !errors.empty?
   msg << "Could not find a valid gem #{gem}, here is why:\n".dup
end
```

akhramov: Thanks, updated!

olleolleolle:
@colby-swandale I think this is a cool change, and it looks like your feedback has been met.

colby-swandale:
@akhramov Apologies, suggesting `StringIO` was bad advice on my part, we should just use `String` instead. I've made a commit to fix this issue and I also moved the test to where the other nonexistent install tests are in the spec file.

colby-swandale:
@duckinator r?

duckinator:
looks good to me. thanks for the PR, @akhramov!

@bundlerbot r+

PR #2072, opened 2017-11-09
Merge glob pattern using braces

znz:
Remove duplicated directory path.

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -442,7 +442,7 @@ By default, this RubyGems will install gem as:
   # for installation of bundler as default gems
   def template_files
     Dir.chdir "bundler/lib" do
-      (Dir[File.join('bundler', 'templates', '**', '*')] + Dir[File.join('bundler', 'templates', '**', '.*')]).
+      (Dir[File.join('bundler', 'templates', '**', '{*,.*}')]).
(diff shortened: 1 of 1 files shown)


lib/rubygems/commands/setup_command.rb
```diff
@@ -442,15 +442,15 @@ def rb_files_in dir
   # for installation of bundler as default gems
   def template_files
```
segiddins: Awesome! While we're at it, can you change this method to just call `template_files_in "bundler/lib"` so we're not duplicating the logic? Thanks!

hsbt: `template_files_in "bundler/lib"` is not working this method. Because installation phase needs to template files with prefix of `bundler` like `bundler/templates/newgem/...` . But uninstallation phase doesn't need it prefix.

segiddins: sorry, in that case `template_files_in "bundler/lib/bundler"` should be equivalent?

hsbt: It's good to equivalent `template_files_in` for `template_files` method.

hsbt:
I merge this. we need simplify these installer logic.

hsbt:
@bundlerbot r+

PR #2091, opened 2017-11-21
Change some strings on the Rakefile to single-quoted strings

lucasprag:
when you don't need string interpolation of special symbols.

# Description:

Cleanup the Rakefile to use single-quoted strings when you don't need string interpolation of special symbols.

# Tasks:

- [x] Write tests (does not apply)

Thank you

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,8 +3,8 @@
 require 'rubygems'
 require 'rubygems/package_task'
 
-if ENV['YAML'] == "syck"
-  ENV['TEST_SYCK'] = "1"
+if ENV['YAML'] == 'syck'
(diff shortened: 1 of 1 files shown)


bronzdoc:
Hey @lucasprag why this change is needed?

lucasprag:
Hi @bronzdoc
It's not that relevant, but double-quoted strings are a little bit slower than single-quoted strings.
Ultimately this comes down to style.

hsbt:
> a little bit slower than single-quoted strings.

No, It's very very small effect. see https://www.viget.com/articles/just-use-double-quoted-ruby-strings

lucasprag:
@hsbt ok

PR #2089, opened 2017-11-21
Avoid encoding issues by using binread in setup

mauromorales:
Fixes #2088

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2093, opened 2017-11-22
IO.binread is not provided at Ruby 1.8

hsbt:
Fixup https://github.com/rubygems/rubygems/pull/2089 at Ruby 1.8


hsbt:
@bundlerbot r+

PR #2099, opened 2017-11-24
Use setup command --regenerate-binstubs option flag

tjouan:
Test setup command binstubs regeneration
----------------------------------------

  Since 909b5fb, executable wrappers are regenerated using pristine
command if any gem are installed. New test is inspired by those for
`Gem::Commands::PristineCommand`.

Fix setup command --regenerate-binstubs option flag
---------------------------------------------------

  `--[no-]regenerate-binstubs` option flag was added in 909b5fb but is
not used yet. This change tests if the option was set before calling
`Gem::Commands::SetupCommand#regenerate_binstubs`.

  We also simplify how the option is set, since it's more similar to
`format_executable` for example (simple option flag), rather than
`document` where we use the same hash key for multiple options. This
way we can just test the value being either true or false, instead of
testing key presence (`Hash#key?`) or relying on `nil` being returned
for nonexistent hash keys with `Hash#[]`.

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -82,11 +82,7 @@ class Gem::Commands::SetupCommand < Gem::Command
 
     add_option '--[no-]regenerate-binstubs',
                'Regenerate gem binstubs' do |value, options|
-      if value then
-        options[:regenerate_binstubs] = true
-      else
-        options.delete(:regenerate_binstubs)
-      end
+      options[:regenerate_binstubs] = value
    end
 
     @verbose = nil
@@ -156,7 +152,7 @@ By default, this RubyGems will install gem as:
 
     say "RubyGems #{Gem::VERSION} installed"
 
-    regenerate_binstubs
+    regenerate_binstubs if options[:regenerate_binstubs]
 
     uninstall_old_gemcutter
 
--- a/test/rubygems/test_gem_commands_setup_command.rb
+++ b/test/rubygems/test_gem_commands_setup_command.rb
@@ -53,6 +53,30 @@ class TestGemCommandsSetupCommand < Gem::TestCase
     FileUtils.mkdir_p 'default/gems/bundler-audit-1.0.0'
   end
 
+  def gem_install name
+    gem = util_spec name do |s|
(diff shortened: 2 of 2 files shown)


tjouan:
Tasks:
------

> - [ ] Write code to solve the problem
> - [ ] Get code review from coworkers / friends

  Let me know if I should adapt those patches. To be honest the test
part is a bit of cargo cult programming inspired by
`Gem::Commands::PristineCommand` tests, so I would be happy to improve
those and receive any feedback.


### Depends on #2098

  This build will fail will following error:

    OptionParser::InvalidOption: invalid option: --silent

  Because if we run `Gem::Commands::SetupCommand` in tests, we must
also execute `Gem::Commands::PristineCommand` with
`--all --only-executables --silent` options, and the last one does not
exist anymore at this time.

tjouan:
> ### Depends on #2098
> 
>   This build will fail will following error:
> 
>     OptionParser::InvalidOption: invalid option: --silent
> 
>   Because if we run `Gem::Commands::SetupCommand` in tests, we must
> also execute `Gem::Commands::PristineCommand` with
> `--all --only-executables --silent` options, and the last one does not
> exist anymore at this time.

  #2098 was merged, build is now OK.

  Please let me know if I should improve anything in this PR
(especially the test part, as said I took inspiration from tests for
pristine command, but I'm not yet familiar with the whole test suite).

tjouan:
@segiddins, thank you for the review!

I'm still working on changes for setup command, and I'm now wondering
if `--[no-]regenerate-binstubs` shouldn't be actually named
`--[no-]regenerate-wrappers`?

Because if I'm correct, it's about RubyGems executable wrappers, and
binstubs seem like something specific to Bundler:
http://bundler.io/man/bundle-binstubs.1.html

I understand that Bundler features are merged into RubyGems, but both
concepts seem slightly different, binstups will force loading specific
versions for gems if I understand correctly.

Am I wrong? If not, should I open an issue or suggest a PR?

Thanks.

segiddins:
That's a good point, and I'm not sure what the answer is? Let's keep it the way it is for now

hsbt:
@bundlerbot r+

PR #2102, opened 2017-11-25
Call SPDX.org using HTTPS

olleolleolle:
This PR **updates the URL** to the web site which hosts all the **SPDX license definitions,** in the script which creates the Ruby file with the license names.

- fix bug
- re-generate `licenses.rb`

(This is #2100 without additional edits.)


segiddins:
Looks good!
@bundlerbot r+

PR #2103, opened 2017-11-25
Remove multi load warning from plugins documentation

tjouan:
Since c0312450, a RubyGems plugin is now only loaded from the latest
installed gem. It was released in version 2.1.0.


tjouan:
I used `[ci skip]` in commit message, maybe it was a mistake.
I'm amending the message since CI is required.

hsbt:
@bundlerbot r+

PR #2105, opened 2017-11-28
Ignored to publish rdoc documentation of rubygems for docs.seattlerb.org

hsbt:
http://docs.seattlerb.org/rubygems seems to an obsoleted site. I removed uploading task from postrelease tasks.

We need to move GitHub pages like "https://rubygems.github.com/documentation/".


hsbt:
@bundlerbot r+

PR #2107, opened 2017-12-01
Handle environment that does not have `flock` system call.

### LINKED ISSUE #2092: Unable to use gem install on Lustre filesystem mounted with noflock
I am unable to use `gem install <name>` on a computing cluster that I use.  When I issue the command I get 

```
hyades: ~$ gem install mesa_reader
ERROR:  While executing gem ... (Errno::ENOSYS)
    Function not implemented @ rb_file_flock - /pfs/home/jschwab/.gem/specs/api.rubygems.org%443/quick/Marshal.4.8/mesa_reader-0.1.0.gemspec
```

This error is not specific to this gem.  It occurs for all gems I have tried.

This computing cluster uses the Lustre file system.  Given the error message, this issue seems related to the use of the mount option
```
  noflock
	Disable posix file locking (Applications trying to use
	the functionality will get ENOSYS)
```
That option cannot be changed.

As a workaround, if I download the gem file separately (e.g. with wget), I am then able to install the gem with `gem install <file>`.

For a single gem, this isn't so bad.  But for a gem with many dependencies, downloading installing them manually becomes a chore.

This issue is related to:

  - [ ] Network problems
  - [X] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [ ] Other

```
$ gem env version
2.6.13
```

hsbt:
Fix #2092

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -161,7 +161,7 @@ module Gem
   # these are defined in Ruby 1.8.7, hence the need for this convoluted setup.
 
   READ_BINARY_ERRORS = begin
-    read_binary_errors = [Errno::EACCES, Errno::EROFS]
+    read_binary_errors = [Errno::EACCES, Errno::EROFS, Errno::ENOSYS]
     read_binary_errors << Errno::ENOTSUP if Errno.const_defined?(:ENOTSUP)
     read_binary_errors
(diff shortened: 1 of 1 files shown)


hsbt:
@bundlerbot r+

PR #2112, opened 2017-12-14
[Specification] Prefer user-installed gems to default gems

segiddins:
# Description:

@BanzaiMan I believe this will fix the issue we were discussing earlier today.

This PR will make RubyGems, when the same gem is installed both by the user and as a default gem, prefer to activate the user-installed gem. This is desirable, for example, with the default bundler gem, as it has fewer executables than the user-installed gem (i.e. the `bundler` executable).

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -812,7 +812,7 @@ class Gem::Specification < Gem::BasicSpecification
   def self.stubs
     @@stubs ||= begin
       pattern = "*.gemspec"
-      stubs = Gem.loaded_specs.values + default_stubs(pattern) + installed_stubs(dirs, pattern)
+      stubs = Gem.loaded_specs.values + installed_stubs(dirs, pattern) + default_stubs(pattern)
       stubs = uniq_by(stubs) { |stub| stub.full_name }
 
       _resort!(stubs)
@@ -843,8 +843,9 @@ class Gem::Specification < Gem::BasicSpecification
       @@stubs_by_name[name] || []
     else
       pattern = "#{name}-*.gemspec"
-      stubs = Gem.loaded_specs.values + default_stubs(pattern) +
-        installed_stubs(dirs, pattern).select { |s| Gem::Platform.match s.platform }
+      stubs = Gem.loaded_specs.values +
+        installed_stubs(dirs, pattern).select { |s| Gem::Platform.match s.platform } +
+        default_stubs(pattern)
       stubs = uniq_by(stubs) { |stub| stub.full_name }.group_by(&:name)
       stubs.each_value { |v| _resort!(v) }
 
--- a/test/rubygems/test_gem_installer.rb
+++ b/test/rubygems/test_gem_installer.rb
@@ -901,6 +901,36 @@ gem 'other', version
     assert_includes(e.message, "can't find gem a (= 3.0)")
   end
 
+  def test_install_creates_binstub_that_prefers_user_installed_gem_to_default
+    Dir.mkdir util_inst_bindir
+
+    install_default_gems new_default_spec('default', '2')
+
+    util_setup_gem do |spec|
+      spec.name = 'default'
+      spec.version = '2'
+    end
+
+    util_clear_gems
+
+    @installer.wrappers = true
+
+    @newspec = nil
+    build_rake_in do
+      use_ui @ui do
+        @newspec = @installer.install
+      end
+    end
+
+    exe = File.join @gemhome, 'bin', 'executable'
+
+    e = assert_raises RuntimeError do
+      instance_eval File.read(exe)
+    end
+
+    assert_equal(e.message, "ran executable")
+  end
+
   def test_install_creates_binstub_that_dont_trust_encoding
     Dir.mkdir util_inst_bindir
     util_setup_gem


MSP-Greg:
Having used trunk for most of the past year, I'm not sure if I agree, but there are valid reasons for any 'preference.'

Thinking forward, I believe that 2.5 has a lot more default gems than 2.4.  Also, part of the reason for their existence is so security issues can be addressed without forcing users to install a whole new ruby build.

So, we have three choices if a user installed version exists alongside a default gem:
1. Use user
2. Use default
3. Use newest

My vote is 3.

On a more general note, RubyGems and Bundler are extremely helpful packages for Ruby.  But, they also encourage package/gem authors to continue using older versions.

Every owner/maintainer will always suggest using the most recent version of their gem, but then their CI is using all sorts of older gems.  Enough on that...

Having spent almost all of my 'OSS time' the past year on:

4. Windows trunk building & testing
5. Encouraging gem owners to test with windows and/or trunk
6. A bit with frozen-string

I've seen a lot of old gem requirements, both as dependencies or development.  Neither Appveyor nor Travis force this on testing, and RubyGems/Bundler can determine what gem version is compatible with what ruby version.

Hence, I'd hope that RubyGems/Bundler do whatever it can to encourage 'use new'...

Thanks, Greg  (rant off)

segiddins:
(3) is already done. This would just swap the preference if the same version of a gem is installed, both as a default gem and as a user installed gem

MSP-Greg:
> if the same version of a gem is installed

Thanks.  Something I would have never thought about.  I guess when you stated 'the same gem', you really meant it.  I took it to mean a different version of the same gem.  Sorry.

In that case, if one wanted to have more than one copy of the same code on their system (???), I'm all for it.

Thanks, Greg

coderanger:
+1 on this from the Chef side of the world. This would fix an issue we're having with gem installation/management now that bundler is a default gem, https://github.com/chef/chef/issues/3363 and https://github.com/chef/chef/issues/6954

lib/rubygems/specification.rb
```diff
@@ -862,7 +862,7 @@ def self.stubs_for name
       @@stubs_by_name[name] || []
     else
       pattern = "#{name}-*.gemspec"
-      stubs = default_stubs(pattern) + installed_stubs(dirs, pattern)
+      stubs = installed_stubs(dirs, pattern) + default_stubs(pattern)
```
coderanger: Would also need the same change a few lines up for `self.stubs` too.

coderanger:
An additional monkey-wrench, if a default gem ends up in `Gem.loaded_specs` it will still take priority in `StubSpecification#to_spec` which might confuse some code.

segiddins:
@MSP-Greg done!

test/rubygems/test_gem_installer.rb
```diff
@@ -889,6 +889,34 @@ def test_install_creates_binstub_that_understand_version
     assert_includes(e.message, "can't find gem a (= 3.0)")
   end
 
+  def test_install_creates_binstub_that_prefers_user_installed_gem_to_default
+    Dir.mkdir util_inst_bindir
+
+    install_default_gems new_default_spec('default', '2')
+    util_setup_gem do |spec|
+      spec.name = 'default'
+      spec.version = '2.1'
```
segiddins: I think this might need to be 2 as well

deivid-rodriguez:
I rebased this PR and applied the last comment by @segiddins. Now it seems to properly catch the changes in lib in a proper regression test.

@MSP-Greg Can you double check if this fixes your problem?

MSP-Greg:
Fixes the problem.  Tried 'real-world' locally.

Thanks for finding this.  I was working on something else, and just quickly checked issues...

deivid-rodriguez:
Any reviewers here?

bronzdoc:
@bundlerbot r+

bronzdoc:
@bundlerbot r+

voxik:
Can somebody please elaborate, how the test relates to the proposed fix? I fail to see the relation.

While the change basically adjust order of loading .gemspec files, the test appears to test that under some circumstances, the installed gem overrides the executable of the previously installed default gem executable. Working on #2909, I really can't wrap my head around this.

deivid-rodriguez:
Hi @voxik.

If I recall correctly, the idea is that in case you have the same version of a gem as a default gem, and as a user-installed gem, the user-installed version is preferred.

When you install a new version of a gem with executables, rubygems will overwrite the global binstub if the new installed version of the gem is the highest. So, in this case, before these changes, installing the same version of a default version gem as a regular gem wouldn't overwrite the binstub (because of the default gem being considered "higher"), but after these change it will.

The test case seems to come from the original issue with the bundler gem shipping different set of executables according to whether it is a default gem or not.

I am not much convinced about this now that I look at it again, since I don't think gems should defer depending on whether they are default gems or not, but I think this is the explanation.

Hope it helps.

voxik:
Thank you.

> Hi @voxik.
> 
> If I recall correctly, the idea is that in case you have the same version of a gem as a default gem, and as a user-installed gem, the user-installed version is preferred.

That should be correct.

> When you install a new version of a gem with executables, rubygems will overwrite the global binstub if the new installed version of the gem is the highest. So, in this case, before these changes, installing the same version of a default version gem as a regular gem wouldn't overwrite the binstub (because of the default gem being considered "higher"), but after these change it will.

I don't think this is really correct. The `Gem.generate_bin` is a bit complex, but I don't think it prioritizes anything. It does several checks and eventually overrides the binstub, but it decides more on content then on version.

> The test case seems to come from the original issue with the bundler gem shipping different set of executables according to whether it is a default gem or not.

I think this could had been part of the issue:

https://bugs.ruby-lang.org/issues/5060

> I am not much convinced about this now that I look at it again, since I don't think gems should defer depending on whether they are default gems or not, but I think this is the explanation.

Actually, the test should be "wherever the lastest and greatest RubyGems binstub is located, it should preferably load the user installed gem". I think that the test is not wrong, but it is unobvious and better if it was split in two.

Anyway, thx for the help.

deivid-rodriguez:
> I don't think this is really correct. The Gem.generate_bin is a bit complex, but I don't think it prioritizes anything. It does several checks and eventually overrides the binstub, but it decides more on content then on version.

You're absolutely right, my bad.

> Actually, the test should be "wherever the lastest and greatest RubyGems binstub is located, it should preferably load the user installed gem". I think that the test is not wrong, but it is unobvious and better if it was split in two.

After I read it again, it seems correct to me too. How would you split it to make it more obvious?

voxik:
> How would you split it to make it more obvious?

It would be nice if there is some shorter path to test the functionality of `stubs` and `stubs_for`.

Not really sure about the rest. For example I don't think it is good idea to store the bin stub into `@gemhome`, because the default gem is stored into completely independent structure. Also, while it is nice that the binstub raises exception, it is not 100% guaranteed that it actually is the exception from the right place.

deivid-rodriguez:
I see. Feel free to contribute any improvements, I agree that working with tests should be easier.

PR #2115, opened 2017-12-19
Frozen string fix - lib/rubygems/bundler_version_finder.rb

MSP-Greg:
# Description:

When starting ruby with `RUBYOPT=--enable-frozen-string-literal`, there are a few issues in RubyGems.  This is the only issue I found in RubyGems, the others were in RDoc, which do cause RubyGems tests to fail.  Working on that.

I found this when I noticed a popluar gem was failing CI, turns out the owner had enabled the above setting.  Investigation found this...
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Gem::BundlerVersionFinder
   @without_filtering = false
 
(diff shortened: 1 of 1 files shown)


segiddins:
Can you add the frozen string literal magic comment to the file? Thanks!

MSP-Greg:
Done.

colby-swandale:
Thanks! @bundlerbot r+

PR #1644, opened 2016-06-16
Add Gem.platform_defaults to allow impls to override defaults.

headius:
# Description:

In order for implementations like JRuby to provide custom defaults for `gem` commands, we have to load `rubygems/config_file.rb` in order to patch the default value for `Gem::ConfigFile::PLATFORM_DEFAULTS. This file is not normally loaded at runtime, since it is not needed to activate gems, and the load in JRuby adds a few hundred milliseconds to our base startup. With this change, the override of `PLATFORM_DEFAULTS` is done lazily, without loading additional files or overwriting a constant.

# Tasks:

- [x] Describe the problem / feature
- [x] Write tests
- [x] Write code to solve the problem
- [x] Get code review from coworkers / friends
- [x] [Squash commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2116, opened 2017-12-19
Add Gem.operating_system_defaults to allow packagers to override defaults.

voxik:
# Description:

This change allows Ruby packagers to override defaults and lazily query
them.

This is very much the same change as #1644 to treat the
operating_system defaults the same way as platform defaults.

# Tasks:

- [x] Write tests


colby-swandale:
Thanks! @bundlerbot r+

PR #2118, opened 2017-12-20
Update for compatibilty with new minitest

MSP-Greg:
# Description:

These patches should allow RubyGems to test with both old and new minitest versions.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -84,7 +84,7 @@ end
 #
 # Tests are always run at a safe level of 1.
 
-class Gem::TestCase < MiniTest::Unit::TestCase
+class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Unit::TestCase)
 
   attr_accessor :fetcher # :nodoc:
 
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1587,7 +1587,7 @@ class TestGem < Gem::TestCase
 
     assert_equal old_style, Gem.find_unresolved_default_spec("foo.rb")
     assert_equal old_style, Gem.find_unresolved_default_spec("bar.rb")
-    assert_equal nil, Gem.find_unresolved_default_spec("baz.rb")
+    assert_nil              Gem.find_unresolved_default_spec("baz.rb")
 
     Gem.clear_default_specs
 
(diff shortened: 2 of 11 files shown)


MSP-Greg:
Why - Because it should be and can be done.

For instance, frozen-string.  How many older test dependencies work with that?  It would certainly be helpful if RubyGems passed tests with trunk and RUBYOPT with --enable-frozen-string-literal.  More helpful if trunk itself passed that way...

Anyway, if this is accepted, I also have a set of patches for windows testing that currently pass locally and on my fork. They skip hoe (an Appveyor specific test only rakefile), and use the gems that are included with Appveyor builds, etc.  I did remove 1.9.3, but I don't think it's used much on windows, as it's only available as a 32 bit build (all 2.0+ builds are 64 or 32).

So, Travis can check with the older dependencies and trunk compatibility, and Appveyor can check with new.  Best of both worlds?

duckinator:
Hi!

So, it looks like there's a few spots where you need to change it from using minitest `"~> 4.0"` to something like `[">= 4.0", "< 6.0"]`.

Here's a patch for that: https://boop.ws/191199f815c94d4894a681fcd6432dd0.txt

You can apply the patch by running the following command from the repo's top-level directory (assuming you have curl):

    curl https://boop.ws/191199f815c94d4894a681fcd6432dd0.txt | git apply -

If CI still passes after that (it works _locally_), I'm definitely okay with merging this.

Let me know if you run into any issues, and thank you for helping improve RubyGems' development environment &mdash; it's very appreciated!

MSP-Greg:
@duckinator

> If CI still passes after that (it works locally)

Thanks, that was the main thing I hoped you could verify.

>So, it looks like there's a few spots where you need to change it from using minitest "~> 4.0" to something like [">= 4.0", "< 6.0"].

I believe that ruby/ruby is still using 4.7.5 (or a derivative of).  Hence, we don't want to complicate the messy job @hsbt has to do keeping things in sync.

So, as mentioned above, what I'm proposing:

1. Travis continue testing as is currently done.  From CI on this PR, that's working.

2. Appveyor is changed to use the minitest version included with each of their builds.  I have code for that (brief description above, passing on my fork & locally with trunk), but haven't added a PR.

As mentioned above, by doing it this way, we stand a chance of getting these tests to pass on Appveyor using trunk/head with `--enable-frozen-string-literal`.

BTW, I originally started on this when I saw that flag added to a popular gem's CI.  Crashed & burned while loading the CI test environment, the crash was in RubyGems...,

Also, I haven't really checked the cases for it, but simply using the frozen magic comment will not guarantee that code will run with it set at the environment / load level.

duckinator:
@MSP-Greg ah, okay, that makes sense. In that case, this looks good to go! thanks for the PR!

@bundlerbot r+

PR #2122, opened 2017-12-22
Update appveyor.yml, add new rakefile for appveyor testing

MSP-Greg:
# Description:

Update Appveyor CI to use a simplified Rakefile, quickens testing, etc.  Also, uses minitest currently loaded in respective Appveyor builds.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -13,6 +13,28 @@ rescue ::LoadError
   require 'yaml'
 end
 
+if Gem.win_platform?
+  # For ruby < 2.0, minitest files need to copied into repo lib folder
+  if RUBY_VERSION <= '1.9.3'
+    dest = File.dirname(__FILE__).gsub(/\//, "\\")
+    src_dir = Dir.glob("#{Gem.default_dir}/gems/minitest-*").sort
+    if src_dir.last
+      src = src_dir.last.gsub(/\//, "\\")
+      `md #{dest}\\lib\\minitest`
+      `copy #{src}\\lib\\minitest.rb #{dest}\\lib`
+      `copy #{src}\\lib\\minitest\\*.* #{dest}\\lib\\minitest`
+    end
+  end
+
+  require "rake/testtask"
+
+  desc "Runs tests without hoe, typically used with windows"
+  Rake::TestTask.new(:test_no_hoe) do |t|
+    t.libs << "test"
+    t.test_files = FileList['test/**/test_*.rb']
+  end
+end
+
 begin
   require 'hoe'
 rescue Gem::ConflictError => e
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,15 +7,7 @@ branches:
 skip_tags: true
 clone_depth: 10
 
-init:
-  - mklink /d C:\git "C:\Program Files\Git"
-  - if %ruby_version%==_trunk (
-      appveyor DownloadFile https://ci.appveyor.com/api/projects/MSP-Greg/ruby-loco/artifacts/ruby_trunk.7z -FileName C:\ruby_trunk.7z &
-      7z x C:\ruby_trunk.7z -oC:\ruby_trunk )
-
 environment:
-  PATH: C:/ruby%ruby_version%/bin;C:/Program Files/7-Zip;C:/Program Files/AppVeyor/BuildAgent;C:/git/cmd;C:/Program Files (x86)/GNU/GnuPG/pub;C:/Windows/system32;C:\Windows;
-  WINDIR: C:\Windows
   matrix:
   - ruby_version: 193
   - ruby_version: 200
@@ -24,6 +16,7 @@ environment:
   - ruby_version: 21-x64
   - ruby_version: 22
   - ruby_version: 22-x64
+    GIT:  C:/git/cmd/git.exe
   - ruby_version: 23-x64
     GIT:  C:/git/cmd/git.exe
   - ruby_version: 24-x64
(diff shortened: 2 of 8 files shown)


appveyor.yml
```diff
@@ -7,62 +7,47 @@ branches:
...
-  - mklink /d C:\git "C:\Program Files\Git"
-  - if %ruby_version%==_trunk (
-      appveyor DownloadFile https://ci.appveyor.com/api/projects/MSP-Greg/ruby-loco/artifacts/ruby_trunk.7z -FileName C:\ruby_trunk.7z &
-      7z x C:\ruby_trunk.7z -oC:\ruby_trunk )
-
 environment:
-  PATH: C:/ruby%ruby_version%/bin;C:/Program Files/7-Zip;C:/Program Files/AppVeyor/BuildAgent;C:/git/cmd;C:/Program Files (x86)/GNU/GnuPG/pub;C:/Windows/system32;C:\Windows;
-  WINDIR: C:\Windows
   matrix:
-  - ruby_version: 193
```
segiddins: why is this removed?

Rakefile_av_test
```diff
@@ -0,0 +1,17 @@
+# -*- ruby -*-
```
segiddins: what does the `av` in the filename stand for?

appveyor.yml
duckinator: @MSP-Greg could you try adding 1.9.3 back and seeing how it goes, since we do still support it? if you run into issues i'im happy to help you debug it sometime next week.

MSP-Greg:
@duckinator 

> if you run into issues i'm happy to help you debug it sometime next week.

I don't know if you'll be happy soon.  Appveyor 193 log is [here](https://ci.appveyor.com/project/MSP-Greg/rubygems/build/job/kg4ixpjekml7w5k5).  No failures, but 27 errors.  As shown earlier, all other ruby versions are okay...

I'll continue to work on it.  Errors are being caused by three missing methods...

Happy Holidays, Greg

hsbt:
I against to add additional rakefile for a specific environment like appveyor.  

@MSP-Greg Can we invoke exists Rakefile in appveyor?

MSP-Greg:
@hsbt

> I'm against to add additional rakefile for a specific environment like appveyor.

Maybe it could be renamed Rakefile_windows_test

> Can we invoke exists Rakefile in appveyor?

Depends.  I did look at that.  The current Rakefile aborts if hoe isn't loaded.  But, hoe is not required for windows testing (may have to change if and when bundler testing is added to windows.)

So, if we want windows users to be able to test without hoe, but also want them to have access to the other tasks currently in the Rakefile (which I assume require hoe), I don't see any other way than using two Rakefiles.

Also, since toplevel return doesn't exist in previous versions, it adds conditional branches that aren't required with two files...

BTW - this is working fine, there are intermittent test failures, one of which is common in my ruby-loco builds (fails in parallel, passes on retry).  Have a fix for it, common issue of class properties not being reset between tests when run in random order.  Thought I'd let ruby-loco run over the weekend to make sure...

Greg

Rakefile_appveyor_test
```diff
@@ -0,0 +1,31 @@
...
+require 'rubygems/package_task'
+require "rake/testtask"
+
+begin
+  require 'psych'
+rescue ::LoadError
+  require 'yaml'
+end
+
+# Stupid hack to copy minitest files into repo folder
```
segiddins: can we not use the word `stupid`?

Rakefile_appveyor_test
```diff
@@ -0,0 +1,31 @@
...
+if RUBY_VERSION <= '1.9.3'
+  dest = File.dirname(__FILE__)
+  src_dir = Dir.glob("#{Gem.default_dir}/gems/minitest-*").sort
+  if src_dir.last
+    cmd = "copy #{src_dir.last}/lib/minitest.rb #{dest}/lib".gsub(/\//, "\\")
```
segiddins: can we use the Rake `cp` command here? And if not, can we use `sh` instead of backticks? We probably should also wrap this in a command, that the `test` task depends upon

hsbt:
@MSP-Greg I didn't understand why you refuse to install `hoe` gem. I agree to invoke test suite only `minitest`. But current master branch couldn't.

Rakefile_appveyor_test
MSP-Greg: I guess we could.  At present, it's windows only, and it's only required for 1.9.3.  All the other version pass without it.

MSP-Greg:
@hsbt 

> But current master branch couldn't

Until bundler was pulled, trunk did work with this PR, as shown above, see [here](https://ci.appveyor.com/project/MSP-Greg/rubygems/build/9) & [here](https://ci.appveyor.com/project/MSP-Greg/rubygems/build/10) (Appveyor on my fork).

Since all Appveyor builds have bundler installed, I just added the code to do so for ruby-loco, and it had one of those intermittent thread crashes in test-all.  If it doesn't pass, it doesn't get added to the 'most recent good build' set.  Next run is at 18:00 JST, I'll see whether that passes, and retest on my rubygems fork.  I'll let you know.

> I didn't understand why you refuse to install hoe gem

I prefer to not download anything that isn't needed, cached or not.  Also, I hate waiting on testing, so anything that speeds it up.  Lastly, the Appveyor people are very reasonable, so I prefer to not misuse a free service...

BTW, core RubyGems passes with frozen-string-literal.  RDoc issues cause some test failures.  Working on those...

It's 23:00 here, I've had enough.  Thanks for your work & help, Greg

MSP-Greg:
@hsbt

> I didn't understand why you refuse to install hoe gem

Because I've seen trunk and ruby-loco pass essentially the same tests hundreds of times without it.  Maybe I need a break, but I use trunk, and I'm tired of old dependencies, multiple versions of the same gem, stuff that was locked out of 2.5 for no reason, and dependencies that may have been helpful at one time, but now have no real benefit, etc,,,  [rant off]

Anyway, seeing some of the PR's, I did not try to (or want to) change Travis to minitest 5.x.  Since trunk is using something loosely based on minitest 4.x, it seemed like a good solution would be to split the testing so one could see results with both.  I don't expect it to be permanent, but converting trunk/default gems over will take time & effort.

This long weekend I ran a lot of builds on my fork, and I found several intermittent tests.  I believe what is currently here for commits works.  There maybe additional intermittent fails, but I've probably found most of them.

You might consider committing it, as you/I/segiddins/others can certainly remove it at a later time...

Thanks, Greg

MSP-Greg:
@hsbt

I'll see what can be done.  Thanks, Greg

MSP-Greg:
@hsbt

Done.  Appveyor passes, the Travis `TEST_TOOL=bundler` jobs are still running.  Travis has an intermittent failure that I've seen on 1.8.7, see [here](https://travis-ci.org/rubygems/rubygems/jobs/321931075#L1958).  The exact same code passed on my fork [here](https://travis-ci.org/MSP-Greg/rubygems/builds/321895025?utm_source=github_status&utm_medium=notification).

I suspect there may be more intermittent failures, but hopefully I've found the most common.

Thanks for all your work here and with Ruby's default gems, Greg

Rakefile
```diff
@@ -13,6 +13,28 @@ rescue ::LoadError
   require 'yaml'
 end
 
+if Gem.win_platform?
```
hsbt: 👍

MSP-Greg:
@hsbt

Funny, you mentioned 'test context'.  I originally used a 'tc_' prefix, but figured that could be confused with `TestCase`.  As you might guess, I'm not terribly comfortable with `TestCase` using instance variables that are the same as attributes used in source, but that's a larger project.

Yes, I'll prefix them with 'stub'.

As long as we're forcing another round of tests, & re the intermittent failures in 1.8.7 & [`test_gem_commands_help_command.rb`](https://github.com/rubygems/rubygems/blob/b25411fc1012c89fd1c409c7e5cc867b85247c29/test/rubygems/test_gem_commands_help_command.rb), are you okay with changing:

```ruby
class TestGemCommandsHelpCommand < Gem::TestCase
  def setup
    super

    @cmd = Gem::Commands::HelpCommand.new

    load File.expand_path('../rubygems_plugin.rb', __FILE__) unless
      Gem::Commands.const_defined? :InterruptCommand
  end
```

to:
```ruby
class TestGemCommandsHelpCommand < Gem::TestCase
  # previously this was calc'd in setup, but 1.8.7 had
  # intermittent failures, but no issues with above require
  PLUGIN = File.expand_path('../rubygems_plugin.rb', __FILE__)

  def setup
    super

    @cmd = Gem::Commands::HelpCommand.new

    load PLUGIN unless
      Gem::Commands.const_defined? :InterruptCommand
  end
```

I doubt either of us is interested in tracking down 1.8.7 fails, and most of the other intermittent fails bounced around as to versions, this issue did not.

Thanks, Greg

Rakefile
MSP-Greg: Thanks.  That's why I built https://msp-greg.github.io/, so I could relax with a tablet and find that stuff...

MSP-Greg:
Great.  Now Appveyor is humming along, but Travis is (intermittently) failing.  It's 20:00, sometime in the next day I'll look over all the Travis failures here and see if I can fix them.  Two were commands, which (I think) use the `@ui` variable that I needed to rename previously...

hsbt:
Great! Thanks for your work.

hsbt:
@bundlerbot r+

Issue #2123, opened 2017-12-22
RG 2.7 Updater broke Travis CI environment.


hsbt:
When user invoked `gem update --system` and `gem install bundler` after bundler installation like Travis environment, RG updater broke it executables of Bundler.

* https://github.com/rails/rails/pull/31541 
* https://travis-ci.org/rails/rails/builds/319911098

Reproduce instruction:

```shell
 > ruby -v
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-darwin17]
 
 > gem list

*** LOCAL GEMS ***

bigdecimal (default: 1.3.0)
did_you_mean (1.1.0)
io-console (default: 0.4.6)
json (default: 2.0.4)
minitest (5.10.1)
net-telnet (0.1.1)
openssl (default: 2.0.5)
power_assert (0.4.1)
psych (default: 2.2.2)
rake (12.0.0)
rdoc (default: 5.0.0)
test-unit (3.2.3)
xmlrpc (0.2.1)
 
 > gem i bundler
Fetching: bundler-1.16.1.gem (100%)
Successfully installed bundler-1.16.1
1 gem installed
 
 > bundle -v
Bundler version 1.16.1
 
 > gem update --system 
Updating rubygems-update
Fetching: rubygems-update-2.7.3.gem (100%)
Successfully installed rubygems-update-2.7.3
Installing RubyGems 2.7.3
Bundler 1.16.0 installed
RubyGems 2.7.3 installed
Regenerating binstubs
`/Users/hsbt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/bundler-1.16.1/exe/bundle` does not exist, maybe `gem pristine bundler` will fix it?
`/Users/hsbt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/bundler-1.16.1/exe/bundler` does not exist, maybe `gem pristine bundler` will fix it?
RubyGems system software updated

 > gem i bundler -v 1.15.4 --force
Successfully installed bundler-1.15.4
1 gem installed

 > bundle -v
/Users/hsbt/.rbenv/versions/2.4.2/bin/bundle:23:in `load': cannot load such file -- /Users/hsbt/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/bundler-1.16.1/exe/bundle (LoadError)
	from /Users/hsbt/.rbenv/versions/2.4.2/bin/bundle:23:in `<main>'
```

[...]

PR #2124, opened 2017-12-22
Fix updater with rubygems

hsbt:
Fixes https://github.com/rubygems/rubygems/issues/2123

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -359,6 +359,8 @@ By default, this RubyGems will install gem as:
     bundler_spec = Gem::Specification.load("bundler/bundler.gemspec")
     bundler_spec.files = Dir.chdir("bundler") { Dir["{*.md,{lib,exe,man}/**/*}"] }
     bundler_spec.executables -= %w[bundler bundle_ruby]
+
+    # Remove bundler-*.gemspec in default specification directory.
     Dir.entries(Gem::Specification.default_specifications_dir).
       select {|gs| gs.start_with?("bundler-") }.
       each {|gs| File.delete(File.join(Gem::Specification.default_specifications_dir, gs)) }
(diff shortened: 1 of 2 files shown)


lib/rubygems/commands/setup_command.rb
```diff
@@ -368,9 +370,16 @@ def install_default_bundler_gem
 
     bundler_spec = Gem::Specification.load(default_spec_path)
 
+    # Remove gemspec that was same version of vendoered bundler.
```
segiddins: `vendored`

segiddins: also, what is the motivation for removing the user-installed Bundler gem spec?

lib/rubygems/commands/setup_command.rb
```diff
@@ -368,9 +370,16 @@ def install_default_bundler_gem
...
+    normal_gemspec = File.join(Gem.default_dir, "specifications", "#{bundler_spec.version}.gemspec")
+    if File.file? normal_gemspec
+      File.delete normal_gemspec
+    end
+
+    # Remove gem files that were same version of vendored bundler.
     if File.directory? bundler_spec.gems_dir
       Dir.entries(bundler_spec.gems_dir).
-        select {|default_gem| File.basename(default_gem).match(/^bundler-#{Gem::Version::VERSION_PATTERN}$/) }.
+        select {|default_gem| File.basename(default_gem).match(/^bundler-#{bundler_spec.version}$/) }.
```
segiddins: you'll want to escape the version, since it will contain regexp metacharacters

hsbt:
@segiddins I added comments for a test scenario. rails/rails and other projects have following environment in Travis.

1. Travis install bundler-1.16.1
2. User invoke `gem update --system`
3. User invoke `gem install bundler`

Therefore, `update_rubygems` removed bundler-1.16.1 files and install vendored bundler-1.16.0. When a user invokes `bundle` command, rubygems activate 1.16.1 executable. So, User faced https://travis-ci.org/rails/rails/builds/319911098 .

We should not remove a `bundler-*` files. We should promote only vendored bundler and remove its version files. (In master branch, it version is 1.16.1)

hsbt:
@bundlerbot r+

segiddins:
But what happens when the user updates to a newer version of the vendored Bundler? Do we still remove the gemfile for the old version?

hsbt:
>Do we still remove the gemfile for the old version?

What mean "the gemfile"?

segiddins:
I meant the gemspec, sorry

hsbt:
I see. I reconsider it again next week.

PR #2126, opened 2017-12-23
Set whether bundler is used for gemdeps with an environmental variable

hsbt:
If we confirm to work test suite without vendored bundler and `Gem::USE_BUNDLER_FOR_GEMDEPS = false`, We got a `LoadError` of bundler files and `uninitialized constant Bundler` error. 

It's better to separate test environment until completely merging bundler.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -176,7 +176,7 @@ module Gem
     write_binary_errors
   end.freeze
 
-  USE_BUNDLER_FOR_GEMDEPS = ENV['USE_BUNDLER_FOR_GEMDEPS'] || true # :nodoc:
+  USE_BUNDLER_FOR_GEMDEPS = !ENV['DONT_USE_BUNDLER_FOR_GEMDEPS'] # :nodoc:
 
   @@win_platform = nil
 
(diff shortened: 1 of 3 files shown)


lib/rubygems.rb
```diff
@@ -176,7 +176,7 @@ module Gem
     write_binary_errors
   end.freeze
 
-  USE_BUNDLER_FOR_GEMDEPS = true # :nodoc:
+  USE_BUNDLER_FOR_GEMDEPS = ENV['USE_BUNDLER_FOR_GEMDEPS'] || true # :nodoc:
```
segiddins: This will always evaluate to true

hsbt: Ah, Thanks. I fixed a condition.

hsbt:
I postpone merging the bundler in Ruby 2.5. So, We need to invoke test suite without bundler library until merging bundler to stdlib or rubygems.

segiddins:
But it’s passing just fine for this repo?

hsbt:
When you remove `bundler` directory, You couldn't run rubygems test. It's the situation of CRuby and JRuby repository. I need to this feature.

hsbt:
@bundlerbot r+

PR #2130, opened 2017-12-26
[Travis] Test against Ruby 2.5

segiddins:
# Description:

Since Ruby 2.5 was released yesterday, we should test against it!


hsbt:
We will merge this after Travis update(It still use 2.5.0-preview1)

hsbt:
@bundlerbot r+

PR #2135, opened 2017-12-27
Test fixups for intermittent tests due to random test order

MSP-Greg:
# Description:

Many tests are still intermittently failing.  These commits are a 'first pass' at cleaning things up.  Passed three times on Travis using my fork.  Hopefully, they'll pass here...

# Tasks:

- [ ] Write tests


MSP-Greg:
As I continue to run more tests, I'm left with the idea that, although these patches represent solutions to the problem, they are not actually removing the original cause.  If the goal is to make them stable, I need to do that.  Hence, closing.

PR #2137, opened 2017-12-30
Improve the error message given when using --version with multiple gems in the install command

colby-swandale:
Set the error message to now be:

    Can't use --version with multiple gems. You can specify multiple gems with
    version requirements using `gem install 'my_gem:1.0.0' 'my_other_gem:~>2.0.0

This gives a more easily understood example of the argument format to
use in rubygems when installing multiple gems with versions constraints.

______________

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2139, opened 2018-01-03
Test fixes - fix intermittent tests, all testing done without hoe

MSP-Greg:
# Description:
I recently closed [PR 2135](https://github.com/rubygems/rubygems/pull/2135) for test updates, since I knew I was missing something.

I was incorrectly assuming that default spec files were being loaded into `File.join(@gemhome, "specifications", "default")`, but they are loaded into another directory.  So, after realizing that, I proceeded with changes & testing.

I've tested it quite a few times, found one additional issue, along with the update to Minitest 5.11 (now fixed with 5.11.1).

This PR does the following:

1. Seems to remove/fix all intermittent failures.  I assume most are due to test leaks, but I didn't set a goal of making sure I found the source of **all leaks**.    IOW, should test leaks be fixed in the source or the tests affected by them?  Some might say only the source, other more pragmatic types might say both.  Regardless, I believe I've run the tests often enough to know that a combination of cleaning leaking tests and cleaning affected tests has been done.

2. Removes hoe from Travis RG testing, matching Appveyor and ruby/ruby.

3. test_case.rb - Adds a method `util_clear_gemspecs`, which clears standard and default gemspec files.  Also, added clearing of default gemspec files to `util_clear_gems`.

4. `Binding.source_location` warnings in head/trunk are fixed by [bundler PR #6237](https://github.com/bundler/bundler/pull/6237).  On my fork, I added a patch for this to my Appveyor testing, and all warnings disappeared.

5. For example results, see my fork's [Travis builds](https://travis-ci.org/MSP-Greg/rubygems/builds), note that I bypassed the bundler testing.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_commands_setup_command.rb
+++ b/test/rubygems/test_gem_commands_setup_command.rb
@@ -10,7 +10,8 @@ class TestGemCommandsSetupCommand < Gem::TestCase
 
   def setup
     super
-
+    util_clear_gems
+    
     @install_dir = File.join @tempdir, 'install'
     @cmd = Gem::Commands::SetupCommand.new
     @cmd.options[:prefix] = @install_dir
@@ -74,6 +75,13 @@ class TestGemCommandsSetupCommand < Gem::TestCase
   end
 
   def test_execute_regenerate_binstubs
+    # Gem.ruby_version needs to be correct for passing test
+    if Gem.instance_variable_defined? :@ruby_version
+      unless Gem.ruby_version.to_s.start_with?(RUBY_VERSION)
+        Gem.send :remove_instance_variable, :@ruby_version
+      end
+    end
+
     gem_bin_path = gem_install 'a'
     write_file gem_bin_path do |io|
       io.puts 'I changed it!'
@@ -86,6 +94,13 @@ class TestGemCommandsSetupCommand < Gem::TestCase
   end
 
   def test_execute_no_regenerate_binstubs
+    # Gem.ruby_version needs to be correct for passing test
+    if Gem.instance_variable_defined? :@ruby_version
+      unless Gem.ruby_version.to_s.start_with?(RUBY_VERSION)
+        Gem.send :remove_instance_variable, :@ruby_version
+      end
+    end
+
     gem_bin_path = gem_install 'a'
     write_file gem_bin_path do |io|
       io.puts 'I changed it!'
--- a/test/rubygems/test_gem_commands_setup_command.rb
+++ b/test/rubygems/test_gem_commands_setup_command.rb
@@ -19,17 +19,17 @@ class TestGemCommandsSetupCommand < Gem::TestCase
     FileUtils.mkdir_p 'bin'
     FileUtils.mkdir_p 'lib/rubygems/ssl_certs/rubygems.org'
 
-    open 'bin/gem',                   'w' do |io| io.puts '# gem'          end
-    open 'lib/rubygems.rb',           'w' do |io| io.puts '# rubygems.rb'  end
-    open 'lib/rubygems/test_case.rb', 'w' do |io| io.puts '# test_case.rb' end
-    open 'lib/rubygems/ssl_certs/rubygems.org/foo.pem', 'w' do |io| io.puts 'PEM'       end
+    File.open 'bin/gem',                   'w' do |io| io.puts '# gem'          end
+    File.open 'lib/rubygems.rb',           'w' do |io| io.puts '# rubygems.rb'  end
+    File.open 'lib/rubygems/test_case.rb', 'w' do |io| io.puts '# test_case.rb' end
+    File.open 'lib/rubygems/ssl_certs/rubygems.org/foo.pem', 'w' do |io| io.puts 'PEM'       end
 
     FileUtils.mkdir_p 'bundler/exe'
     FileUtils.mkdir_p 'bundler/lib/bundler'
 
-    open 'bundler/exe/bundle',        'w' do |io| io.puts '# bundle'       end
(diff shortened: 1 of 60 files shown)


MSP-Greg:
Generally, I like to hold myself to 'If I break it, I fix it'.

Tests should remove any files they create.  Many of the patches in this PR do not efficiently do that, but they do seem to reliably allow the tests to pass.

In the next few weeks, I expect to go thru the code and align it with the above idea, (remove files where they're created)...

Thanks Greg

test/rubygems/test_gem_dependency_installer.rb
```diff
@@ -312,7 +313,7 @@ def test_install_doesnt_upgrade_installed_dependencies
     end
 
     assert_equal %w[a-2 b-1], Gem::Specification.map(&:full_name)
-    assert_equal %w[b-1], inst.installed_gems.map { |s| s.full_name }
+    assert_equal %w[b-1], inst.installed_gems.map(&:full_name)
```
hsbt: Please don't mix stylistic change to feature/bugfix pull request.

MSP-Greg: Sorry, I wouldn't have changed it except for the fact that the right above it uses `(&:full_name)`

I assume you would you like it changed (removed)?

segiddins:
Thanks so much for all the effort on this, @MSP-Greg !

test/rubygems/test_gem_dependency_installer.rb
hsbt: Yes. Remove/Revert it.

test/rubygems/test_gem_doctor.rb
```diff
@@ -154,6 +156,10 @@ def test_doctor_child_missing
   end
 
   def test_gem_repository_eh
+    unless (t = Dir.glob(File.join(@gemhome, "**/*.gemspec"))).empty?
```
hsbt: Why don't you use `util_clear_gemspecs` in this line?

hsbt:
@MSP-Greg I've partly merged your changes.

* https://github.com/rubygems/rubygems/pull/2158
* https://github.com/rubygems/rubygems/pull/2158

I still have some concerns. 

1. Can you describe reason why `Gem.*_*_hooks` was added in your pull request?
2. What issues resolved by `util_clear_gemspecs` and `util_clear_gems`?

test/rubygems/test_gem_commands_setup_command.rb
```diff
@@ -74,6 +75,13 @@ def gem_install name
   end
 
   def test_execute_regenerate_binstubs
+    # Gem.ruby_version needs to be correct for passing test
+    if Gem.instance_variable_defined? :@ruby_version
```
hsbt: Why don't you use `util_set_RUBY_VERSION`?

And I'm not sure why this line was required. Can you describe this?

test/rubygems/test_gem_doctor.rb
MSP-Greg: My mistake.  Early on, as I was running tests, I just had `util_clear_gems`.  As I saw more intermittent tests, I added `util_clear_gemspecs`.  I forgot to change the code here.  I'll fix, tomorrow (I'm -0600).

test/rubygems/test_gem_commands_setup_command.rb
MSP-Greg: > Why don't you use util_set_RUBY_VERSION?

I was unaware of it.  All the issue concerned was that one variable

> And I'm not sure why this line was required. Can you describe this?

There must have been another test that changed the value from default, and didn't reset it.  That's what I was speaking of in a previous message about trying to find those issues, which would require adding logging to `Gem::TestCase` and auditing all left-over artifacts...

MSP-Greg:
@hsbt 

> Can you describe reason why Gem.*_*_hooks was added in your pull request?

I assume you mean the lines like `Gem.post_build_hooks.pop 2`.  Some other tests were corrupted by them.

> What issues resolved by `util_clear_gemspecs` and `util_clear_gems`?

Left-over artifacts from other tests.  As mentioned, ideally, any tests that add files should remove them.  or, any `setup`'s adding files, their respective `teardown`'s should remove...

test/rubygems/test_gem_commands_setup_command.rb
MSP-Greg: Investigated further.  See [test_gem_request_set_gem_dependency_api.rb:L35](https://github.com/rubygems/rubygems/blob/2ea667874491eadd9fcdcb6421795bfc52252ab1/test/rubygems/test_gem_request_set_gem_dependency_api.rb#L35)?

Also, reviewing that code, it appears that if `util_set_RUBY_VERSION` was called twice, it would remove the initial/real settings.  IOW, it seems to assume that every call to `util_set_RUBY_VERSION` is followed by a call to `util_restore_RUBY_VERSION`.  Maybe I'm missing something...

MSP-Greg:
Yes. Okay to close.  It will be a few days.

As mentioned, with an example of 'sloppy test A causes issues with test B', in many cases this PR changes B.  I will take the time to write log code and find and fix all the A tests, which I think is the better solution.

I did not write all these sloppy tests.  I'm assuming that 'generic cleaning' after every test would affect test time (and maybe results), as I've seen Appveyor get 'quirky' when a lot of disk i/o was going on and testing was running parallel.

> Please remove needless changes from focus thing of this pull-request like the stylistic change.

Given ESL and the fact that you've already mentioned this, I'll ignore the phrasing of the above sentence.

Thanks, Greg

PR #2141, opened 2018-01-05
Backport ruby core changes for test fixes.

hsbt:
Ruby 2.6 will change the behavior of `$SAFE` variable. 

It's part of test fixes by @MSP-Greg like https://github.com/rubygems/rubygems/pull/2139

### DIFF
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -38,18 +38,6 @@ class TestGemRequire < Gem::TestCase
     assert require(path), "'#{path}' was already required"
   end
 
-  def append_latch spec
-    dir = spec.gem_dir
-    Dir.chdir dir do
-      spec.files.each do |file|
-        File.open file, 'a' do |fp|
-          fp.puts "FILE_ENTERED_LATCH.release"
-          fp.puts "FILE_EXIT_LATCH.await"
-        end
-      end
-    end
-  end
-
   # Providing -I on the commandline should always beat gems
   def test_dash_i_beats_gems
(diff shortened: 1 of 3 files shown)


test/rubygems/test_require.rb
```diff
@@ -422,9 +414,11 @@ def test_require_bundler_with_bundler_version
   end
 
   def silence_warnings
-    old_verbose, $VERBOSE = $VERBOSE, false
-    yield
-  ensure
-    $VERBOSE = old_verbose
+    begin
```
segiddins: the `begin` here shouldn't be necessary

test/rubygems/test_require.rb
```diff
@@ -106,8 +100,6 @@ def test_concurrent_require
     assert t1.join, "thread 1 should exit"
     assert t2.join, "thread 2 should exit"
   ensure
-    return if $! # skipping
-
     Object.send :remove_const, :FILE_ENTERED_LATCH
```
segiddins: should this now have an `if defined?`

test/rubygems/test_require.rb
hsbt: It's my mistake for syntax fixes. see  https://github.com/rubygems/rubygems/pull/2141/commits/abc00f1ac9fef0b3d0bc552bd1710561685f9621

test/rubygems/test_require.rb
hsbt: Thanks. It caused failure with Ruby 1.8 maybe. I added a condition.

hsbt:
@bundlerbot r+

PR #2142, opened 2018-01-05
Use `File.open` instead of `open`.

hsbt:
This change is not vulnerability fix. @hsbt and @shugo did audit this usage when CVE-2017-17405 was disclosed.

Because ruby core team will warn to use `Kernel#open` in standard libraries.


hsbt:
@bundlerbot r+

Issue #1967, opened 2017-07-02
gem uninstall never reports errors even if it failed to remove executables


unak:
See https://github.com/rubygems/rubygems/blob/master/lib/rubygems/uninstaller.rb#L216-L217
As you know `FileUtils.rm_f` never raises any exception if it fails to remove the specified files.
Then users cannot know the files are really removed or not, and also cannot know why the removal was failed.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.12
```
(bundled version of MRI trunk)

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

segiddins:
Can you please provide an example of what error message you were expecting but did not receive?

unak:
@segiddins I don't know.  I want to know it, really :-)
Rubygems reported nothing, so I could not know what was happened.

segiddins:
But what failed?

unak:
I don't know.  The executables were not removed.

shevegen:
I did not write the rubygems code but if I were to change it, I would probably add a standalone method "def rm_rf()" which will handle all FileUtils.rm_rf calls.

Inside that method you could also more readily fine-tune what to do; such as in the example that unak said, to also provide error-reporting to the user. At the least from the code side but people are different and nobody has to agree with me - I just feel that methods are cleaner in general than direct calls to tools whose functionality may not always be what a user wants to see. :-)

On the topic itself, I do not know rubygems internal policy but from my memory of working with rubygems, rubygems in general tends to give very little output, which is probably deliberate. My scripts output a LOT, which is often spamming stuff down, for example - not good either. But I like to see rather more than less notification, personally.

Anyway, I feel that this could quite easily be handled. If the ruby gem team decides that no change is warranted, then an ENV variable could be used. Some verbosity level such as RUBY_GEM_VERBOSE or something like that. If this is then set, then code like the above, passed through a method, could check for that variable, and if it was set, to also provide information that unak wanted to have here (e. g. his use case that was documented in the first post).

Anyway that is just one suggestion - it'll all depends on the behaviour, so perhaps the oldschool gurus like drbrain etc... can say something about the gem-policy about removing files/directories and either failing, and/or not finding such files.

colby-swandale:
[...]

PR #2154, opened 2018-01-20
Raise errors in `gem uninstall` when a file in a gem could not be removed 

colby-swandale:
A problem that i've encountered a few times and that has also been reported at #1967 is that `gem uninstall` does not report any errors if it was unable to remove any files/executables or gemspec. 

Not successfully removing all the gem's files can introduce unexpected behavior for the user, especially when the files that could not be removed are the gem's executable or gemspec. A gem could be reported as still being installed and available to execute even though the rest of the gem has been removed.

This is happening because `gem uninstall` uses `rm_f` and `rm_rf` in the FileUtils library which suppresses any errors when attempting to delete the file.

This PR changes this behavior to instead capture any errors when removing a gem's files and raise an error to the user of the problem that has been encounted.

Example error when a gem could not be successfully removed:

```
Error: unable to successfully uninstall 'a' which is located at '/private/var/folders/vg/0528djbx3r9fb1d7qlpxwzq40000gn/T/test_rubygems_633/gemhome/gems/a-2'. This is most likely because the current user does not have the appropriate permissions
```

## A Few Notes

I've changed the Gem's `gemspec` to be the last thing that is removed, as that is the point which RubyGems says if a gem is installed or not, so it should only be removed when everything else was successfully removed.

I've opted to use mocks to test this behavior. Trying to create a env that could setup the conditions for a file to be prevented from being removed using only the current user proved to be quite difficult. I'm certainly open for suggestion in testing this behavior without using mocks.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/uninstall_command.rb
+++ b/lib/rubygems/commands/uninstall_command.rb
@@ -129,11 +129,7 @@ that is a dependency of an existing gem.  You can use the
 
     specs.each do |spec|
       options[:version] = spec.version
-
-      begin
-        Gem::Uninstaller.new(spec.name, options).uninstall
-      rescue Gem::InstallError
-      end
+      uninstall_gem spec.name
     end
 
     alert "Uninstalled all gems in #{options[:install_dir]}"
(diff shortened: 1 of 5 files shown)


segiddins:
This looks cool! 👍

hsbt:
@bundlerbot r+

hsbt:
@bundlerbot retry

PR #2155, opened 2018-01-23
Remove PID from gem index directory.

hsbt:
# Description:

`gem generate_index` allows creating sequential directory used PID. It's possible to guess the directory name by other local users. It's better to use `Dir.mktmpdir` instead of PID.


djberg96:
👍

hsbt:
@bundlerbot r+

PR #2160, opened 2018-01-26
Remove hoe for test suite.

hsbt:
# Description:

This is part of https://github.com/rubygems/rubygems/pull/2139/

______________

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -2,6 +2,7 @@
 
 require 'rubygems'
 require 'rubygems/package_task'
+require "rake/testtask"
 
 if ENV['YAML'] == "syck"
   ENV['TEST_SYCK'] = "1"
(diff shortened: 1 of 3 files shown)


hsbt:
@bundlerbot r+

colby-swandale:
Why was this merged without review!?

hsbt:
> Why was this merged without review!?

Because I'm a reviewer for #2139.

colby-swandale:
ah, thanks pointing that out @hsbt

indirect:
@hsbt thank you for doing this! the way Hoe was set up has also caused many new tickets to be opened by developers who want to contribute but don't already know how Hoe works and how to use it, so I think this will be an improvement there, too.

PR #2164, opened 2018-02-01
nil guard on `Gem::Specification`

hsbt:
# Description:

the ruby core has an issue of race condition for the concurrent environment. We already added extra nil guard to `Gem::Specification`. 

I will backport its code.

______________

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

Issue #677, opened 2013-10-09
improve gem install --default


mpapis:
Forwarding problems with the new [rubinius](https://github.com/rubinius/rubinius) and [rubysl](https://github.com/rubysl) => https://github.com/wayneeseguin/rvm/issues/2260 ... basically it is about standard library gems not found in few cases, usually with bundler.

What I would like to see from rubygems side is to make `gem install --default` to fully install the gems in `$LOAD_PATH` and build native extensions (like [zlib](https://github.com/rubysl/rubysl-zlib/blob/2.0/ext/rubysl/zlib/extconf.rb)), most of the code is in place there would be needed an extra `if` to recognize `--default-stdlib` - which would install the files in stdlib and compile extensions - compared to current `--default` which only adds specifications.

I'm opening this more for discussion and asking if that change would be approved, I can prepare code for it.

cc @headius, @brixen.

copiousfreetime:
Checking in on the interest people have for how `--default` works.

voxik:
I am interested in proper solution for default gems, i.e. something like I proposed in my previous comment. I don't care about --default option, since it solves just symptoms.

enebo:
@voxik Can you uninstall those gems on Fedora?  If you do is Ruby in a broken state?

voxik:
The problem is not with installation or uninstallation, that is why we have dependencies, but with Bundler, which knows nothing about default gems and projects typically does not specifies dependencies on them

enebo:
@voxik I am not making an opinion on default gems.  I am just asking a question.  If I uninstall json in Fedora completely does Ruby stop bootstrapping?

enebo:
I meant base json ruby library not some underlying C library (I hope that was obvious but I should clarify).

headius:
There are a few issues with making things "just gems":

1. Startup time is impacted when RubyGems has to load. Normally on MRI (before 2.3) unless you use something installed as a normal gem, RubyGems does not have to do its full setup. In 2.3 they started loading the `did_you_mean` gem at boot, which impacted 2.3 startup time.
2. As @enebo mentions, if everything is "just gems" then a user can break their Ruby by uninstalling too much. Consider also that if you install multiple versions of a library that's has a default gem, you will have to uninstall each of them individually (since uninstall all would break Ruby).
[...]

PR #2166, opened 2018-02-01
Behaviour changes with default gems installer

hsbt:
# Description:

A current installer of default gems only installs executable file and `gemspec`. It didn't resolve the real world problem.

Default gems provide to upgrade standard libraries like `gem update openssl`. But current installer used by `gem install openssl --default` is inconsistency behavior with `gem update`.

I change its behavior to install gem files and build native extensions. It can upgrade instructions for the old version of Ruby like 2.3 or 2.4.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -764,7 +764,7 @@ class Gem::Installer
   def verify_default_gems
     return unless options[:install_as_default]
     return if Gem::Specification::DEFAULT_GEMS_LIST.include?(spec.name)
-    raise Gem::InstallError, "#{spec} is not a default gems"
+    raise Gem::InstallError, "#{spec} is not a default gem."
   end
 
   ##
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -321,14 +321,15 @@ class Gem::Installer
     dir_mode = options[:dir_mode]
     FileUtils.mkdir_p gem_dir, :mode => dir_mode && 0o755
 
+    extract_files
+
+    build_extensions
+    write_build_info_file
+
     if @options[:install_as_default]
       extract_bin
       write_default_spec
     else
-      extract_files
-
-      build_extensions
-      write_build_info_file
       run_post_build_hooks
     end
 
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -760,6 +760,12 @@ class Gem::Installer
     end
   end
 
+  def verify_default_gems
+    return unless options[:install_as_default]
+    return if Gem::Specification::DEFAULT_GEMS_LIST.include?(spec.name)
+    raise Gem::InstallError, "#{spec} is not a default gems"
+  end
+
   ##
   # Return the text for an application file.
 
@@ -924,6 +930,10 @@ TEXT
     verify_spec
 
     ensure_loadable_spec
+
+    verify_spec
+
+    verify_default_gems
 
     if options[:install_as_default]
       Gem.ensure_default_gem_subdirectories gem_home
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -110,6 +110,31 @@ class Gem::Specification < Gem::BasicSpecification
 
   VALID_NAME_PATTERN = /\A[a-zA-Z0-9\.\-\_]+\z/ # :nodoc:
 
+  DEFAULT_GEMS_LIST = %w[
+    bigdecimal
+    cmath
+    csv
+    date
+    dbm
+    etc
+    fcntl
+    fiddle
+    fileutils
+    gdbm
+    io-console
+    ipaddr
+    json
+    openssl
+    psych
+    rdoc
+    scanf
+    sdbm
+    stringio
+    strscan
+    webrick
+    zlib
+  ] #:nodoc:
+
   # :startdoc:
 
   ##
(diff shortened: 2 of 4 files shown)


hsbt:
Related with https://github.com/rubygems/rubygems/issues/677

lib/rubygems/specification.rb
```diff
@@ -112,6 +112,31 @@ class Gem::Specification < Gem::BasicSpecification
 
   VALID_NAME_PATTERN = /\A[a-zA-Z0-9\.\-\_]+\z/ # :nodoc:
 
+  DEFAULT_GEMS_LIST = %w[
```
segiddins: shouldn't this pull from `Gem::Specification.select(&:default_gem?).map(&:name)` ?

segiddins:
> It didn't resolve the real world problem.

Which real-world problem? Does this only manifest on `gem update`, or also `gem install`? Can you add a test case for the changed behavior?

colby-swandale:
ping @hsbt

lib/rubygems/specification.rb
hsbt: It's not working. default gems were determined without ruby version.

hsbt:
>Which real-world problem? 

If you want to use the latest version of csv with Ruby 2.4.0 via `gem install csv`, but RubyGems will activate csv on standard libraries. It needs to be promoted default gems on Ruby 2.4.

lib/rubygems/installer.rb
```diff
@@ -707,6 +708,12 @@ def verify_spec_name
     raise Gem::InstallError, "#{spec} has an invalid name"
   end
 
+  def verify_default_gems
+    return unless options[:install_as_default]
+    return if Gem::Specification::DEFAULT_GEMS_LIST.include?(spec.name)
+    raise Gem::InstallError, "#{spec} is not a default gems"
```
segiddins: `not a default gem` ?

lib/rubygems/specification.rb
segiddins: >  default gems were determined without ruby version

I'm not sure I understand what you mean by this?

hsbt: Default gems do not depend on Ruby version.

Ex. We can install default gems from listed by https://stdgems.org/  with old ruby. But is not activated in the current specification. I hope to activate them with old ruby.

lib/rubygems/specification.rb
```diff
@@ -112,6 +112,31 @@ class Gem::Specification < Gem::BasicSpecification
...
+    openssl
+    psych
+    rdoc
+    scanf
+    sdbm
+    stringio
+    strscan
+    webrick
+    zlib
+  ] #:nodoc:
```
segiddins: do we want to freeze this list, or otherwise append to it when we see more gems in the default directory?

deivid-rodriguez:
@hsbt I'm running into the following issue with `bundler`

```console
$ gem list json

*** LOCAL GEMS ***

json (default: 2.1.0)
jsonapi-renderer (0.2.0)
multi_json (1.13.1)

$ echo -e "source 'https://rubygems.org'\ngem 'json'" > Gemfile

$ bundle
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Bundled gems are installed into `./.bundle`

$ bundle pristine
Failed to pristine json (2.1.0). Cached gem /home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/cache/json-2.1.0.gem does not exist.

$ gem install json
Fetching: json-2.1.0.gem (100%)
Successfully installed json-2.1.0
1 gem installed

$ bundle pristine
Installing json 2.1.0 with native extensions

$ gem uninstall json
Successfully uninstalled json-2.1.0

$ gem list json

*** LOCAL GEMS ***

json (default: 2.1.0)
jsonapi-renderer (0.2.0)
multi_json (1.13.1)

$ bundle pristine
Failed to pristine json (2.1.0). Cached gem /home/deivid/.rbenv/versions/2.5.1/lib/ruby/gems/2.5.0/cache/json-2.1.0.gem does not exist.
```

Do you think the root problem might be what this issue is trying to address?

hsbt:
@deivid-rodriguez Yes, The current default-gems was inconsistency behavior.

deivid-rodriguez:
Great! For starters I'm going to try and confirm whether this PR fixes the problem!

deivid-rodriguez:
I tried it with this PR and I still get the same behavior. Can you confirm the issue? Maybe I didn't try it correctly.

hsbt:
@deivid-rodriguez Sorry, I didn't understand that your expected behavior. Did you hope to finish `gem pristine json`?

deivid-rodriguez:
Yeah, I hope `bundle pristine` would not crash.

But I just tried `gem pristine json` and it behaves better

```console
$ gem pristine json
Restoring gems to pristine condition...
Skipped json-2.1.0, it is a default gem
```

So, this seems like a bug in `bundler`, not in `rubygems`! Sorry for the false alarm.

hsbt:
@deivid-rodriguez Ah, I see. I missed `bundle pristine` on your first comment. It seems the issue of bundler.

deivid-rodriguez:
@hsbt I'd like to move forward the work in this PR if that's ok with you?

fallingicarus (approved):

lib/rubygems/installer.rb
Adix82: I

Adix82: I

hsbt:
I reboot this proposal for altanative approach of https://bugs.ruby-lang.org/issues/19351

mahjelan (commented):
initiation </>

hsbt:
I changed my proposal. see https://github.com/rubygems/rubygems/pull/7588

PR #2169, opened 2018-02-02
Support option for `--destdir` with upgrade installer.

### LINKED ISSUE #2134: --destdir appears to be broken for installing rubygems 2.7.x
I'm having a problem.

Attempting to package rubygems 2.7 using the --destdir feature fails; the embedded bundle installation attempts to write to the actual prefix (`/opt/ruby2.3` in my case) instead of `$DESTDIR/{prefix}`. This behavior appears with either `--destdir` or setting the environment variable `DESTDIR` before running `ruby setup.rb`.

This functionality works correctly on 2.6.14 but not on any 2.7.x versions.

This issue is related to:

  - Other

```
$ gem env version
PASTE HERE
```

not applicable; the Ruby which I am attempting to package is 2.3.6 with Rubygems 2.7.4 (latest).

The tail of running `setup.rb` during my rpmbuild can be found at https://gist.github.com/Roguelazer/0be8a4229cb14fd31006a73998296cf2.

Obviously I don't expect y'all to do anything about my RPM in particular, but you can reproduce this by attempting to run `ruby setup.rb --destdir $HOME/some/directory` and observing that it fails when it can't write to your actual ruby prefix.

hsbt:
# Description:

the original patch was created by @tjouan.

https://github.com/rubygems/rubygems/issues/2134#issuecomment-354285403

Fixes #2134

______________

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -350,7 +350,9 @@ By default, this RubyGems will install gem as:
   def install_default_bundler_gem
     return unless Gem::USE_BUNDLER_FOR_GEMDEPS
 
-    mkdir_p Gem::Specification.default_specifications_dir
+    specs_dir = File.join options[:destdir],
+      Gem::Specification.default_specifications_dir
+    mkdir_p specs_dir
 
     # Workaround for non-git environment.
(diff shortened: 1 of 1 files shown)


hsbt:
@bundlerbot r+

PR #2165, opened 2018-02-01
Backport ruby core commits for testcase

hsbt:
# Description:

I backport https://github.com/ruby/ruby/commit/b496220a1f70f8393070ffebcb883c7c5fc036d7 and https://github.com/ruby/ruby/commit/15689ed7780b06ddc14cde4f427de834177283a5 from ruby/ruby.
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2176, opened 2018-02-05
Prefer to use `Numeric#zero?` instead of `== 0`

hsbt:
Follow up with https://github.com/rubygems/rubygems/pull/2165#issuecomment-362861991


hsbt:
@bundlerbot r+

PR #2182, opened 2018-02-09
Drop to support Ruby < 2.2

hsbt:
# Description:

I know that this violates SemVer policy(But I didn't find the RubyGems adopts SemVer policy officially).

To support Ruby 1.8 and 1.9 reduces maintainers productivity. I always fix or added deprecated code like `defined?` or `respond_to?`, fix failing tests of too old ruby and preparing its environment to my box. I was burned out by its works.

# solution

Remove to support 1.8.7, 1.9.2 and 1.9.3 at RubyGems 2.8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,6 @@ branches:
     - /^[\d.]+$/
     - /.+-stable$/
 rvm:
-- 1.8.7
-- 1.9.2
-- 1.9.3
 - 2.0.0
 - 2.1.10
 - 2.2.9
@@ -31,12 +28,6 @@ script:
 - util/ci script
 matrix:
   exclude:
-    - rvm: 1.8.7
-      env: "TEST_TOOL=rubygems YAML=psych"
-    - rvm: 1.9.2
-      env: "TEST_TOOL=bundler RGV=master"
-    - rvm: 2.0.0
-      env: "TEST_TOOL=rubygems YAML=syck"
     - rvm: 2.1.10
       env: "TEST_TOOL=rubygems YAML=syck"
     - rvm: 2.2.9
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,33 +13,17 @@ branches:
     - /^[\d.]+$/
     - /.+-stable$/
 rvm:
-- 2.0.0
-- 2.1.10
 - 2.2.9
 - 2.3.6
 - 2.4.3
 - 2.5.0
 - ruby-head
 env:
-  - "TEST_TOOL=rubygems YAML=syck"
   - "TEST_TOOL=rubygems YAML=psych"
   - "TEST_TOOL=bundler RGV=master"
 script:
 - util/ci script
 matrix:
-  exclude:
-    - rvm: 2.1.10
-      env: "TEST_TOOL=rubygems YAML=syck"
-    - rvm: 2.2.9
-      env: "TEST_TOOL=rubygems YAML=syck"
-    - rvm: 2.3.6
-      env: "TEST_TOOL=rubygems YAML=syck"
-    - rvm: 2.4.3
-      env: "TEST_TOOL=rubygems YAML=syck"
-    - rvm: 2.5.0
-      env: "TEST_TOOL=rubygems YAML=syck"
-    - rvm: ruby-head
-      env: "TEST_TOOL=rubygems YAML=syck"
   allow_failures:
     # TODO: Remove 1.8.7-2.1.10 jobs after releasing bundler 1.16.2
     - rvm: 1.8.7
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,11 +9,6 @@ clone_depth: 10
 
 environment:
   matrix:
-  - ruby_version: 193
-  - ruby_version: 200
-  - ruby_version: 200-x64
-  - ruby_version: 21
-  - ruby_version: 21-x64
   - ruby_version: 22
   - ruby_version: 22-x64
     GIT:  C:/git/cmd/git.exe
@@ -32,7 +27,6 @@ init:
       7z x C:\ruby_trunk.7z -oC:\ruby_trunk )
 
 install:
-- if %ruby_version%==193 ( gem update minitest --no-document )
 - ps: >-
     git submodule update --init --recursive
 
--- a/rubygems-update.gemspec
+++ b/rubygems-update.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
   s.homepage = "https://rubygems.org".freeze
   s.licenses = ["Ruby".freeze, "MIT".freeze]
   s.rdoc_options = ["--main".freeze, "README.md".freeze, "--title=RubyGems Update Documentation".freeze]
-  s.required_ruby_version = Gem::Requirement.new(">= 1.8.7".freeze)
+  s.required_ruby_version = Gem::Requirement.new(">= 2.0.0".freeze)
   s.rubygems_version = "2.7.3".freeze
   s.summary = "".freeze
 
--- a/rubygems-update.gemspec
+++ b/rubygems-update.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
   s.homepage = "https://rubygems.org".freeze
   s.licenses = ["Ruby".freeze, "MIT".freeze]
   s.rdoc_options = ["--main".freeze, "README.md".freeze, "--title=RubyGems Update Documentation".freeze]
-  s.required_ruby_version = Gem::Requirement.new(">= 2.0.0".freeze)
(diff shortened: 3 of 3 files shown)


indirect:
I think last year we planned to release RubyGems 3.0 during this month (Feb 2018, when ruby-core drops security fixes for 2.2), and drop support 1.8, 1.9, 2.0, 2.1, and 2.2. How does that sound? /cc @segiddins @duckinator

segiddins:
It's OK with me, so long as we bump the version to 3.0

duckinator:
I'm also okay with this, as long as we bump the version to 3.0.

hsbt:
I summarized :

 * RG 2.8: (2018 Dec target): Drop to support < 2.0 version, and marked many warnings in the deprecated feature.
 * RG 3.0 ( 2019 or 2020 that was the same time of Ruby 3.0): Drop to support < 2.3 or <= 2.3, Remove deprecated feature,

@indirect How do you think about this plan? We should release 2.8 that is preparing RG 3 version.

colby-swandale:
We may not have adopted an official policy but we've been following semver in our releases for the past couple of years and dropping ruby support in 2.8 would surprise developers, so i agree that we should bump the version to 3.0.

duckinator:
To expand a bit on what I said earlier:

1. Trying to make Ruby and RubyGems version numbers match when the projects are moving at least somewhat-independently doesn't seem useful to me. They aren't tied together strongly enough right now to justify that, in my opinion.
2. Dropping support for Ruby versions in minor releases is confusing. From experience with other projects, I would expect a large, abrupt influx of confused issues and support.

So: I'm strongly in favor of dropping support for Ruby versions prior to 2.0, but I am _against_ doing it without major version bump.

duckinator:
So, my suggestion for moving forward would be:

1. Figure out a policy for versioning (semver?). Document this prominently.
2. Figure out a policy for how long RubyGems versions will be supported. Document this prominently.
3. Figure out a policy for how long RubyGems releases will support certain Ruby versions. Document this prominently.
4. Figure out what we want to make deprecated. Make a list. Be as detailed as possible about both functionality to be changed/removed _and_ reasoning. _Keep this information somewhere in case we encounter related issues later on._
5. **Publicly discuss our plans from \# 4, and solicit feedback.**
6. Start marking things as deprecated, while integrating feedback from # 5.
7. **Publicly discuss what is being deprecated,** so it's less of a surprise.
8. Make a 2.x release (or a few) _with all of the deprecation warnings_.
9. Prepare for 3.0 release &mdash; e.g., removing deprecated functionality.
10. Solicit feedback from the community on what they need.
11. Make the 3.0 release, taking into account all of the feedback gotten in prior steps.

The net result of this is that the 3.0 release won't be a special case, it'll just be a normal release under newly-decided policy. It'll be a big change, because RubyGems is long overdue for a major release, but figuring out a proper plan means any following major releases will be less difficult.

When I say "discuss," I mean getting out in the community and actively engaging with developers and companies that rely on RubyGems, etc. From prior experience with other projects, I've found that this can _significantly_ improve adoption rates of new versions.

**I am willing to take on a large chunk of the more outreach-oriented aspects of this.**

matthewd:
FWIW, this would not actually violate semver-as-specified: https://semver.org/spec/v2.0.0.html#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

hsbt:
okay, I bump current master version to 3.0 and added deprecated warnings, remove to support < Ruby 2.3 targeted Ruby 2.6 in 2018, After that, I will remove its deprecates with RG 4.0 targeted Ruby 3 in 2020.

hsbt:
I have another plan.

1.  We make the current master to 2.8 with warnings for deprecated methods. it still supports < Ruby 2.0. and We backport its warnings to RG 2.7.6+ and Ruby 2.5.1+
2. We release RG 2.8 anytime.
3a. We bump master branch to 3.0, and it targets Ruby 2.6.
3b. In the worst case, I will merge RG 2.8 into Ruby 2.6.  RG 3.0 targets Ruby 2.7 or 3.0.

I think most important things are `warnings for deprecated methods`, not Ruby versions.

duckinator:
@hsbt i'm okay with this approach.

i do think it's worthwhile to do some community outreach and documentation because 3.0 will be a huge change and we want to minimize negative impact, but that can be done in parallel to your idea.

indirect:
Thanks for the feedback, everyone!

@matthewd It's a good point that semver, as written, does not require major version bumps to drop compatibility with older dependencies. That said, I don't think "which versions of Ruby are supported" is a dependency—I think it counts as an API. Reasonable people may disagree. ¯\_(ツ)_/¯

@duckinator I think your suggestions are really good, particularly to 1) have a policy, 2) document the planned changes, and 3) make them available for discussion and announcement in advance.

@hsbt I think we will need other issues/PRs to discuss and deprecate methods, but I think that plan is a good one.

Given the status quo, where both Bundler and RubyGems have continued to support Ruby 1.8+ for many years, I think it makes sense use the major version number as a message to users: we are making a (relatively) big change by dropping support for the older, unsupported Ruby versions.

Based on all of these suggestions, I'd like to propose that we do something like this:

- [ ] Update the [POLICIES](https://github.com/rubygems/rubygems/blob/master/POLICIES.rdoc) document to make it clear that in the future, RubyGems minor releases will drop support for Ruby versions without no support from ruby-core.
  That means dropping releases of Ruby in the first minor after that Ruby version loses security support. We will re-evaluate anytime ruby-core changes their support levels or plans.
- [ ] Update `master` to 3.0 right away, and drop support for Ruby 1.8, 1.9, 2.0, and 2.1
  We can release 3.0 whenever we are ready for the next release, but this means we can stop supporting old Ruby right away.
- [ ] Open PRs for planned deprecations that include 1) what is being deprecated, 2) why we want to deprecate it, 3) when the deprecation will become an error, and 4) expected impact on Ruby and RubyGems users
  That provides opportunities for feedback, and a record of why the deprecation seems like a good idea.
- [ ] After merging deprecations, release a 3.x version of RubyGems with the deprecation warnings
  This lets us give deprecation warnings only to users on modern versions of Ruby. 👍
- [ ] Backport just the deprecation warnings to RubyGems 2.7.x, so they will be released in Ruby 2.5.x
  The backport only needs to support Ruby 2.5, not old Rubies. Releasing method deprecations in Ruby will give us the widest reach possible, so the ecosystem will be better prepared when we remove.

Some time later (we should discuss when... maybe Oct/Nov 2018?):

- [ ] Change `master` to 4.0
- [ ] Remove deprecated code
- [ ] Release 4.0
- [ ] Fix all known issues with the 4.x.x series of RubyGems
- [ ] Merge 4.x.x to ruby-trunk to be released with Ruby 2.6 in Dec 2018

I think this plan accomplishes all of the goals stated above:

- Ruby users will get a major version bump to make them notice the big Ruby version drop
- Releasing 3.0 can be easy and fast because we aren't breaking anything else
- Removing methods will include a major bump to 4.0, with many months for deprecation warnings
- @hsbt can stop supporting old rubies right away
- @duckinator will have documented policies and discussion PRs with feedback
- @matthewd will (eventually) get versions of RubyGems that drop old Ruby versions in minor releases

Please let me know if you have any other concerns or suggestions, but I think this overall plan is much better than anything that I had come up with myself. It was only possible because everyone had really great input. Thank you again, everyone. 🙇

indirect:
> and 2.2? we(ruby core team) will end to support it at 31 Mar 2018. We can release RG 3.0 after that.

To try to let you drop old Rubies right away, I would suggest changing `master` to 3.0 now and keep 2.2 support. Then, after we change POLICIES, we can drop Ruby 2.2 on 31 Mar by releasing RubyGems 3.1.

But if you think that is too much trouble, I think it is completely okay to wait and release 3.0 on 31 Mar, dropping all old Rubies at once, including 2.2. 👍🏻

hsbt:
@indirect I understand. It's no rush issue. I totally agreed your plan. Thanks for your summrize.

deivid-rodriguez:
So, dropping support for old rubies means that a user using those rubies and running `gem update --system` won't be able to upgrade, right? That doesn't seem backwards incompatible to me...

That said, using the major version to signal this big change seems sensible even if it was considered backwards compatible.

indirect:
@deivid-rodriguez we might need to make changes to RubyGems to make that happen 😬 but yes, in principle I agree it is fine to drop old versions on minor releases, and in this particular one-off case I think we need a major bump to let everyone know that we are starting to drop old versions. 👍

hsbt:
I'm going to merge this and start to work to develop RG 3 and 4. I've filed @indirect 's plan to https://github.com/rubygems/rubygems/issues/2191

hsbt:
@bundlerbot r+

PR #2195, opened 2018-02-17
remove CVEs from the rubygems repo

colby-swandale:
This PR is paired up with https://github.com/rubygems/guides/pull/209. I want to cleanup the repository a bit by moving the CVE documentation that's sitting in the root folder of the project to the guides repository instead.

These documentation are pretty old and the their value is not worth much, so moving these should be easy and undisruptive.


hsbt:
Can you remove them from https://github.com/rubygems/rubygems/blob/master/rubygems-update.gemspec#L17 too?

colby-swandale:
@hsbt Thank you for picking that up, i also need to update the `Manifest.txt` as well. 🙇

colby-swandale:
@bundlerbot r+

PR #2196, opened 2018-02-17
Add missing requires

deivid-rodriguez:
# Description:

I'm getting some crashes when running tests for my gem: https://circleci.com/gh/decidim/decidim/54469.
Adding the missing requires that the error points to seems to fix them. The require in `dependency.rb` comes from the error message, the second require comes from the error message that still appears after adding the first require.

I tried to isolate this, but I couldn't. So the minimal reproduction would be:

```
git clone https://github.com/decidim/decidim
cd decidim
git checkout rails-5.1.5-and-pg-1.0
bundle install
bundle exec rspec spec/generator_spec.rb[1:1:1:1]
```

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "rubygems/util"
+
 module Gem::BundlerVersionFinder
   @without_filtering = false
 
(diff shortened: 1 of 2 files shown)


segiddins:
Thanks! @bundlerbot r+

deivid-rodriguez:
Thanks to you! :)

PR #2198, opened 2018-02-19
Fix getting started instructions

bronzdoc:
# Description:
Fix the `Getting started` part of the documentation since we are not using the `hoe gem` anymore


colby-swandale:
Thanks! @bundlerbot r+

PR #2200, opened 2018-02-20
Update the CONTRIBUTING guide with instructions on signing up with Slack

colby-swandale:
# Description:

Most of the rubygems maintainer team now communicate over Slack rather then IRC, so we should update the docs to reflect on the best way to reach us.

I've also updated the clause on updating the `Manifest.txt` to include the rake task that can be used to generate it automatically.

# Tasks:

- [ ] Write tests


bronzdoc:
Why not keep both?

indirect:
@bronzdoc I don't think anyone from the current maintainers team idles in IRC anymore... I used to pay for IRCCloud just so I could do that, but gave up after there wasn't any activity for 6 months.

colby-swandale:
Thanks for the review everyone ❤️ 

@bundlerbot r+

PR #2201, opened 2018-02-21
Fix rubygems dev env

bronzdoc:
# Description:

This will fix the rake error when running `rake test` on ruby version < 2.5
```
File does not exist: bundler

rake aborted!
```
the reason the error was happening is that `lib/rubygems/test_case.rb` is requiring bundler even on ruby version < 2.5 and  the `rake test`  task only include `bundler/lib` when ruby is >= 2.5  

Also made the `setup` task to be dependent of the `bundler:checkout` task so the bundler submodule is initialized properly on setup.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -51,10 +51,13 @@ Rake::TestTask.new do |t|
   else
     t.ruby_opts = %w[--disable-gems]
   end
+
   t.ruby_opts << '-rdevkit' if Gem.win_platform?
 
   t.libs << "test"
-  t.libs << "bundler/lib" if RUBY_VERSION >= "2.5"
+
+  t.libs << "bundler/lib"
+
   t.test_files = FileList['test/**/test_*.rb']
 end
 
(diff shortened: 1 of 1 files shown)


bronzdoc:
this is failing in ruby 2.3.6 :scream_cat:

hsbt:
I can reproduce your issue. When we have the environment without bundler, `rake test` was failed.

Should we apply this patch?

```
diff --git Rakefile Rakefile
index 07f0f799e3..292d5da63e 100644
--- Rakefile
+++ Rakefile
@@ -53,7 +53,7 @@ Rake::TestTask.new do |t|
   t.ruby_opts << '-rdevkit' if Gem.win_platform?
 
   t.libs << "test"
-  t.libs << "bundler/lib" if RUBY_VERSION >= "2.5"
+  t.libs << "bundler/lib"
 end
```

segiddins:
I think @hsbt’s patch is the right approach, versus disabling Bundler entirely

bronzdoc:
I updated the PR, thanks for the review @hsbt @segiddins !

hsbt:
@bundlerbot r+

PR #2202, opened 2018-02-22
Update POLICIES with new EOL, and release process.

indirect:
This update to the end-of-life policy is as discussed in #2182.

The release process is a draft, and includes all of the necessary steps
that I am aware of. I'm happy to revise it further based on feedback.

### DIFF
--- a/POLICIES.rdoc
+++ b/POLICIES.rdoc
@@ -2,8 +2,8 @@
 
 Contributions to RubyGems are made via GitHub pull requests, which must be
 approved by a project committer other than the author. To approve a PR, a
-maintainer can leave a comment including the text "@bundlerbot r+", indicating that
-they have reviewed the PR and approve it. Bundlerbot will then
+maintainer can leave a comment including the text "@bundlerbot r+", indicating
+that they have reviewed the PR and approve it. Bundlerbot will then
 automatically create a merge commit, test the merge, and land the PR if the
 merge commit passes the tests.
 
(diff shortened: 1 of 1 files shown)


POLICIES.rdoc
```diff
@@ -45,14 +45,31 @@ For example, RubyGems 2.5.0 was recently released.  RubyGems does not
...
-RubyGems is 2.5.x, so the RubyGems version would become 3.x if it used
-features only available in Ruby 2.0 and later.
+For example, since Ruby 2.2 has reached end-of-life, RubyGems may use require
+Ruby 2.3 or later.  As of this writing RubyGems is at versino 3.0, so the
+RubyGems version would become 3.1 if it used features only available in Ruby
+2.3 and later.
+
+=== Security Release Process
+
+- Obtain CVE numbers as needed from HackerOne or RedHat.
```
hsbt: Why `RedHat`? I wonder why they assign CVE number of rubygems.

POLICIES.rdoc
```diff
@@ -45,14 +45,31 @@ For example, RubyGems 2.5.0 was recently released.  RubyGems does not
...
-that Ruby version.  This may result in a major version change.
+Additionally, when a Ruby version reaches end-of-life the following minor
+release of RubyGems will drop backwards compatibility with that Ruby version.
 
-For example, since Ruby 1.9.3 has reached end-of-life, RubyGems may use
-features of ruby that only exist in Ruby 2.0 and later.  As of this writing
-RubyGems is 2.5.x, so the RubyGems version would become 3.x if it used
-features only available in Ruby 2.0 and later.
+For example, since Ruby 2.2 has reached end-of-life, RubyGems may use require
+Ruby 2.3 or later.  As of this writing RubyGems is at versino 3.0, so the
```
hsbt: `s/versino/version`

POLICIES.rdoc
```diff
@@ -45,14 +45,31 @@ For example, RubyGems 2.5.0 was recently released.  RubyGems does not
...
+RubyGems version would become 3.1 if it used features only available in Ruby
+2.3 and later.
+
+=== Security Release Process
+
+- Obtain CVE numbers as needed from HackerOne or RedHat.
+- Agree on a release date with ruby-core, so patches can be backported to older
+  Ruby versions as needed.
+- Avoid releasing security updates on Fridays, so platform services don't have
```
segiddins: I think this is unfair to those who are actually doing the releases

POLICIES.rdoc
chrisarcand: Not sure I understand the question, but if you're wondering why a CVE number on RubyGems and not just Ruby itself (as RubyGems ships with Ruby, of course) it's because Red Hat RPMs packages it's own Ruby very differently than MRI and RubyGems is also a separate package. Don't ask 🙄

Disclaimer: I work for Red Hat

@indirect  s/RedHat/Red Hat/   👈  I have to include this or I'll be fired :trollface:

POLICIES.rdoc
chrisarcand: Is it? How so? Wouldn't you just wait until Monday?

POLICIES.rdoc
jrafanie: I'll add that if any help or guidance is needed in terms of CVE process for rubygems or any gem for that matter, we can certainly find someone to answer any questions you may have.  I also work for Red Hat but on the application side.

indirect: @hsbt So far, the Bundler and RubyGems projects have always gotten CVE numbers from either HackerOne or Red Hat's OSS security team. Those companies already have large blocks of CVE numbers assigned from MITRE, and they are willing to grant CVE numbers from their already-assigned blocks to Ruby open source projects. The point of the CVE system ultimately to make sure that every security issue has just one number, so it doesn't really matter whether we get our CVE numbers directly from MITRE or out of a block assigned by MITRE to another company.

indirect: (Also, it is a lot of work and takes a long time to get a block of CVE numbers assigned directly by MITRE, so it does not seem worth doing that work if we only need ~1-5 CVE numbers per year.)

POLICIES.rdoc
indirect: 👍  thanks

POLICIES.rdoc
indirect: @segiddins can you elaborate on what part of this seems unfair to the person doing the release? The goal is definitely to come up with a solution that works for the maintainers, the releaser, and the downstream developers who need to deploy those releases.

indirect:
thanks for the feedback, everyone!

@bundlerbot r+

POLICIES.rdoc
segiddins: I oftentimes only have time to do releases (which take more than just a few minutes) on weekends or holidays. “Just wait until Monday” doesn’t really work

indirect: That seems totally reasonable. We can take security releases on a case-by-case basis, and one of myself, @hsbt, @colby-swandale, or @claudijd can handle RubyGems security releases as needed. 👍

hsbt: I can work it instead of @segiddins in the day time.

indirect:
@bundlerbot retry

PR #2203, opened 2018-02-22
Fix Gem::Version.correct?

m-nakamura145:
# Description:

in ruby 2.5.0. 
```
irb(main):001:0> Gem::Version.correct?(nil)
=> true
```

I think that it was wrong.  I think that I should behave as follows.
```
irb(main):001:0> Gem::Version.correct?(nil)
=> false
```
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -170,6 +170,7 @@ class Gem::Version
   # True if the +version+ string matches RubyGems' requirements.
 
   def self.correct? version
+    return false if version.nil?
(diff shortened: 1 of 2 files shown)


hsbt:
Looks reasonable. Thanks!

hsbt:
@bundlerbot r+

m-nakamura145:
Thanks!! 👍

segiddins:
https://github.com/CocoaPods/CocoaPods/issues/7765 This change has caused issues for several CocoaPods users

Issue #1837, opened 2017-02-08
Feature request: chdir to gemspec directory when doing gem build and control output location


Rush:
My current problem is: when building a gem from outside and running `gem build ../some/path/file.gemspec` the resulting `.gem` has no contents. This is because `s.files` does not match anything. As part of my build system I'd like to run the `gem` command from a different working directory and I would also like to control the output location of the resulting `.gem` file. 

To sum up, there are two requests:
* [ ] Make sure the `gem` command chdirs to the `.gemspec` file location.
* [ ] Allow to output the `.gem` file to a different dir, or at least to the initial working dir before the chdir to the gemspec location.

This issue is related to:
  - [X] The command line `gem`

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
May be a duplicte of https://github.com/rubygems/rubygems/issues/448

segiddins:
In the time being, you can set the gemspecs files to be relative to that of __FILE__

kke:
Also ran into this. 

Currently you have to do something like this:

```console
gem build foofoo.gemspec && \
gem install `ls -t foofoo-*.gem|head -1` 
```

because there's no way to control where or what the resulting .gem will be called and `gem install foofoo-*.gem` will install any versions you have laying around in that dir.

Something like this would be nice:

```console
gem build foofoo.gemspec foofoo.gem
# and/or
gem build foofoo.gemspec --install
```

deivid-rodriguez:
This is specially confusing if there's a gemspec in the current folder.

In my case, I have this:

```
$ ls -1 **/*.gemspec
decidim-accountability/decidim-accountability.gemspec
decidim-admin/decidim-admin.gemspec
decidim-api/decidim-api.gemspec
decidim-assemblies/decidim-assemblies.gemspec
decidim-budgets/decidim-budgets.gemspec
decidim-comments/decidim-comments.gemspec
decidim-core/decidim-core.gemspec
decidim-debates/decidim-debates.gemspec
decidim-dev/decidim-dev.gemspec
decidim.gemspec
decidim-meetings/decidim-meetings.gemspec
decidim-pages/decidim-pages.gemspec
decidim-participatory_processes/decidim-participatory_processes.gemspec
decidim-proposals/decidim-proposals.gemspec
decidim-surveys/decidim-surveys.gemspec
decidim-system/decidim-system.gemspec
[...]

PR #2204, opened 2018-02-23
Allow building gems without having to be in the gem folder 

colby-swandale:
# Description:

The `gem build` command does not change into the working directory when building gems. This leads to issues when building a gem when you're not inside the same directory as the gemspec.

Here is an example of the trying to build the Bundler gem when i'm not inside the project directory.

```
› gem build bundler/bundler.gemspec
/Users/c/Projects
fatal: Not a git repository (or any of the parent directories): .git
    ["CHANGELOG.md", "LICENSE.md", "README.md", "bundler.gemspec", "exe/bundle", "exe/bundler"] are not files
```

This PR changes the `build` command to change into the same directory that the `gemspec` is located in. This makes it possible to build a gem without having to be inside it.

```
› gem build bundler/bundler.gemspec
/Users/c/Projects/bundler
  Successfully built RubyGem
  Name: bundler
  Version: 1.16.2
  File: bundler-1.16.2.gem
```

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -47,13 +47,15 @@ with gem spec:
     end
 
     if File.exist? gemspec then
-      spec = Gem::Specification.load gemspec
+      Dir.chdir(File.dirname(gemspec)) do
+        spec = Gem::Specification.load File.basename(gemspec)
 
-      if spec then
-        Gem::Package.build spec, options[:force]
-      else
-        alert_error "Error loading gemspec. Aborting."
-        terminate_interaction 1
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This is awesome @colby-swandale, thanks!

This should fix #1837, right? At least the part of "not failing", although you cannot (yet?) control the output location?

colby-swandale:
That's correct

duckinator:
This looks good to me. Is there any reason for not merging this?

segiddins:
@bundlerbot r+

PR #2211, opened 2018-02-25
Fix path checks for case insensitive filesystem

larskanis:
# Description:
On Windows paths are case insensitive. That means that path comparisons must be done case insensitive.

Different lower / upper case characters can happen, when RbConfig::TOPDIR or GEM_HOME are set to a path of different case.

This bug was introduced in commit 666ef793c . It stood out in this nokogiri issue: https://github.com/sparklemotion/nokogiri/issues/1726

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -429,6 +429,14 @@ EOM
     destination
   end
 
+  def normalize_path(pathname)
+    if Gem.win_platform?
+      pathname.downcase
+    else
+      pathname
+    end
+  end
+
   def mkdir_p_safe mkdir, mkdir_options, destination_dir, file_name
     destination_dir = realpath File.expand_path(destination_dir)
     parts = mkdir.split(File::SEPARATOR)
@@ -437,7 +445,7 @@ EOM
       path = File.expand_path(path + File::SEPARATOR + basename)
       lstat = File.lstat path rescue nil
       if !lstat || !lstat.directory?
-        unless path.start_with? destination_dir and (FileUtils.mkdir path, mkdir_options rescue false)
+        unless normalize_path(path).start_with? normalize_path(destination_dir) and (FileUtils.mkdir path, mkdir_options rescue false)
           raise Gem::Package::PathError.new(file_name, destination_dir)
         end
       end
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -524,6 +524,21 @@ class TestGemPackage < Gem::Package::TarTestCase
     assert_path_exists extracted
   end
 
+  if Gem.win_platform?
+    def test_extract_tar_gz_case_insensitive
+      package = Gem::Package.new @gem
+
+      tgz_io = util_tar_gz do |tar|
+        tar.add_file 'foo/file.rb', 0644 do |io| io.write 'hi' end
+      end
+
+      package.extract_tar_gz tgz_io, @destination.upcase
+
+      extracted = File.join @destination, 'foo/file.rb'
+      assert_path_exists extracted
+    end
+  end
+
   def test_install_location
     package = Gem::Package.new @gem
 


hsbt:
@larskanis Thanks! Can you make a this pull request without `send` method?

duckinator:
A potential approach would be to define a `normalize_path()` or similar:

```ruby
def normalize_path(pathname)
  if Gem.win_platform?
    pathname.downcase
  else
    pathname
  end
end
```

and then you can do this:

```ruby
unless normalize_path(path).start_with? normalize_path(destination_dir) and (FileUtils.mkdir path, mkdir_options rescue false)
  raise Gem::Package::PathError.new(file_name, destination_dir)
end
```

larskanis:
Thanks for reviewing! I adjusted the PR according to your proposal. And sorry for the delay!

One more note: The different path lower/uppercase spelling comes from the fact, that `destination_dir` is [expanded by one call to expand_path](https://github.com/rubygems/rubygems/blob/da7657875ebb85401e8f743c6a9af1844d6bcee5/lib/rubygems/package.rb#L433) but that `mkdir` is [expanded part by part](https://github.com/rubygems/rubygems/blob/da7657875ebb85401e8f743c6a9af1844d6bcee5/lib/rubygems/package.rb#L437). Interestingly enough both ways lead to different lower/uppercase spelling.

Since this is pretty unexpected, I opened a bug report on ruby here: https://bugs.ruby-lang.org/issues/14570

hsbt:
@bundlerbot r+

larskanis:
Ruby-2.4.4 contains the [problematic patch](https://github.com/ruby/ruby/commit/4237809aa3983fca36b520f95b48fdf9e7e88de2) to rubygems-2.6.14, but the above fix is not backported yet. Can you please make up leeway?

Also the latest commits to rubygems-2.6 are not yet in this repository.

hsbt:
@larskanis I addressed this https://bugs.ruby-lang.org/issues/14721

larskanis:
Thank you @hsbt . I patched RubyInstaller [2.4.4](https://github.com/oneclick/rubyinstaller2-packages/commit/88cf2c228b1cd85e5a112990a4d69002294e068e), [2.5.1](https://github.com/oneclick/rubyinstaller2-packages/commit/df557484bc83814ceef635bd896e5e3134126b06) and [-head](https://github.com/oneclick/rubyinstaller2-packages/commit/213fae70bc8be229f4d68018f16228e6ab79156e) before releasing these new versions. Otherwise there probably would have been a lot of complaints :-( Good to know it will be fixed with the next ruby releases.

djberg96:
> On Windows paths are case insensitive. That means that path comparisons must be done case insensitive.

Actually, HFS is as well, so we may need to revisit this. You can see this by opening up a shell on your Mac and doing `cd /USERS/YOUR_HOME` and it will work. Also, if you have a file "foo.txt" and run `File.exists?('FOO.txt')` it will return true. `File.identical?('foo.txt', 'FOO.TXT')` will also return true.

larskanis:
@djberg96 I investigated the path issue a bit deeper in the [above comment](https://github.com/rubygems/rubygems/pull/2211#issuecomment-370061792). There might be a better fix than using `if Gem.win_platform?`  - but this was the fastest and safest attempt. Did you encounter the `installing into parent path [...] is not allowed` error on HFS?

djberg96:
@larskanis I've not encountered it personally, no. I just wanted to point out that it's a potential issue elsewhere for future reference in case anyone comes across this in the future, but for now the solution is fine.

On a side note (just to make matters more confusing) the Windows FS is "case aware", meaning that you can have a "Foo.txt" and a "foo.txt" via the `FILE_FLAG_POSIX_SEMANTICS` flag to the `CreateFile` function, but that's rare in practice. Again, not something I would worry about for now.

Trying to solve it more robustly would probably require something like my `sys-filesystem` gem where you would have to check filesystem properties, but I'm guessing we don't want to go there.

PR #2217, opened 2018-03-02
Improve invalid proxy error message

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/1783

### DIFF
--- a/lib/rubygems/local_remote_options.rb
+++ b/lib/rubygems/local_remote_options.rb
@@ -7,7 +7,6 @@
 
 require 'uri'
 require 'rubygems'
-require 'rubygems/user_interaction'
 
 ##
 # Mixin methods for local and remote Gem::Command options.
--- a/lib/rubygems/local_remote_options.rb
+++ b/lib/rubygems/local_remote_options.rb
@@ -7,6 +7,7 @@
 
 require 'uri'
 require 'rubygems'
+require 'rubygems/user_interaction'
 
 ##
 # Mixin methods for local and remote Gem::Command options.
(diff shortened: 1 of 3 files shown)


lib/rubygems/local_remote_options.rb
```diff
@@ -24,8 +25,13 @@ def accept_uri_http
         raise OptionParser::InvalidArgument, value
       end
 
-      unless ['http', 'https', 'file', 's3'].include?(uri.scheme)
-         raise OptionParser::InvalidArgument, value
+      valid_uri_schemes = ["http", "https", "file", "s3"]
+      unless valid_uri_schemes.include?(uri.scheme)
+        msg =  "Invalid uri scheme for #{value}\nPreface URLs with one of #{valid_uri_schemes.map{|s| "#{s}://"}}"
+
+        ui.say(msg)
```
segiddins: is using `ui` preferable to appending to the `reason` in the `OptionParser::InvalidArgument` error?

bronzdoc: The OpttionParser.accept is overwriting the whatever y pass as a reason to `OptionParser::InvalidArgument` with the argument is passed to the block, in this case `value`
```
    OptionParser.accept URI::HTTP do |value| ...
```

I don't know if there's a way to make OptionParser.accept to stop doing that, so i came with that solution, i will investigate a little more in the evenening

bronzdoc: @segiddins one option would be to throw a different error, the problem i described only happens when we raise `OptionParser::InvalidArgument ` inside the `OptionParser.accept`, what you think?

segiddins: I'm 👍 for throwing another error

bronzdoc:
PR updated @segiddins

segiddins:
Awesome! @bundlerbot r+

PR #2227, opened 2018-03-09
prefer to use util_spec in `Gem::TestCase`

hsbt:
# Description:

`Gem::Testcase` has `new_spec` and `util_spec` for building stub specification. I migrated them to util_spec and marked new_spec to deprecated.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -816,8 +816,7 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
 
     spec
   end
-  # TODO: mark deprecate after replacing util_spec from new_spec
-  # deprecate :new_spec, :none, 2018, 12
+  deprecate :new_spec, :none, 2018, 12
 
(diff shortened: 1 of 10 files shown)


lib/rubygems/test_case.rb
```diff
@@ -866,6 +867,19 @@ def util_spec name, version = 2, deps = nil # :yields: specification
       end
     end
 
+    unless files.empty? then
```
segiddins: should this check `s.files` instead?

duckinator: @segiddins `s` isn't defined at this point.

@hsbt can you leave a comment here confirming whether or not this `unless` statement is correct?

if you confirm that's correct, lgtm.

hsbt: It is the different condition with `spec.files`. I tried to replace `spec.files`, after that some tests were failed. 

I will go current implementation as first.

hsbt:
@bundlerbot r+

PR #2230, opened 2018-03-11
 [Requirement] Treat requirements with == versions as equal

segiddins:
# Description:

Fixes https://github.com/bundler/bundler/issues/6295

Previously, Gem::Requirement.new(“= 1) != Gem::Requirement.new(“= 1.0)
We fix that by comparing equality by the (op, version) tuples, rather than string equality on the versions

Internally, we now keep requirements sorted by version, so we don’t need to re-sort each time any method that uses a sorted list of requirements is called (such as == or to_s)

# Tasks:

- [x] Write tests


hsbt:
I confirmed test-case. It's reasonable.

bronzdoc:
@bundlerbot r+

PR #2242, opened 2018-03-21
Add a note for the non-semantically versioned case

### LINKED ISSUE #2228: Semver warning is brittle
My current problem is that I continuously get this warning when I build my gem:

```
WARNING:  pessimistic dependency on rails (~> 5.1.5) may be overly strict
  if rails is semantically versioned, use:
    add_runtime_dependency 'rails', '~> 5.1', '>= 5.1.5'
```

I see some problems with it:

* What if the dependency is not semantically versioned? I don't like unfixable warnings...
* I can trick rubygems to remove the warning with `>= 5.1.5, < 5.2`. Since that's equivalent to `~> 5.1.5`, I'd say it's a bug in the detection of when to print the warning. But maybe it's intented and it's actually the expected fix when the dependency is not semantically versioned? In that case, it should be added to the message I guess. 

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
$ gem env version
2.7.6
```

deivid-rodriguez:
# Description:

Fixes #2228 by recommending another form of version specification when the overly strict dependency is not actually semantically versioned.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -144,11 +144,15 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
         _, dep_version = dep.requirement.requirements.first
 
         base = dep_version.segments.first 2
+        upper_bound = dep_version.segments.first(dep_version.segments.length - 1)
+        upper_bound[-1] += 1
 
         warning_messages << <<-WARNING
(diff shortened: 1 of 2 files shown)


bronzdoc:
Thanks! @bundlerbot r+

PR #2247, opened 2018-03-25
[TarHeader] Extract the empty header into a constant

segiddins:
# Description:

This way, we don't need to allocate new strings (and then do `String#*`) each time

# Tasks:

- [ ] Write tests


colby-swandale:
Thanks! @bundlerbot r+

PR #2250, opened 2018-03-26
Add Rubocop

colby-swandale:
# Description:

This PR introduces Rubocop to help with fixing outdated styles being used in the project and help developers stick to a style that can be tested against.

Initially, Rubocop is only checking `lib/rubygems.rb` but i plan to work on adding more files for Rubocop to check and eventually have Rubocop covering the rest of the repository.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,4 @@
+AllCops:
+  DisabledByDefault: true
+  Exclude:
+    - 'bundler/**/*'
--- a/CONTRIBUTING.rdoc
+++ b/CONTRIBUTING.rdoc
@@ -11,7 +11,7 @@ contributors to follow to reduce the time it takes to get changes merged in.
 2. Ensure that your code blends well with ours:
    * No trailing whitespace
    * Match indentation (two spaces)
-   * Match coding style (+if+, +elsif+, +when+ need trailing +then+)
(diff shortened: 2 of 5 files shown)


hsbt:
I'm only okay to add rubocop rule with `DisabledByDefault: true`. We should add cop rule step by step.

segiddins:
What if we just ran `—auto-gen-config` ?

colby-swandale:
@segiddins i looked into doing that first but there is an internal limit on the number of rules it can skip which the rubygems repo is easily exceeding.

tiagoefmoraes:
@colby-swandale you can run `rubocop --auto-gen-config --exclude-limit 0` and it will remove the limit.

It may be a good idea to `Exclude` the bundler submodule before doing this.

Also bundler has some configurations worth considering.

I used this [post](https://joanswork.com/rails-rubocop-update/) as a reference on a project recently.

colby-swandale:
I'm going to disable all cops by default and then make separate PRs that will enable/fix rubocop rules.

duckinator:
@colby-swandale do you have anything else you need to do on this PR?

your idea of gradually adding rubocop rules sounds good to me, and this PR looks okay assuming you're done with it.

colby-swandale:
@duckinator Yep, this is ready to merge.

PR #2259, opened 2018-04-01
 Fix #2153: Warn deprecated license

### LINKED ISSUE #2153: Warn for deprecated SPDX license identifiers?
Some of the current SPDX license and exception identifiers are deprecated by the SPDX (e.g. [`GPL-3.0` has been deprecated in favor of `GPL-3.0-only` in the SPDX license list v3.0][1]).  Since SPDX may eventually (year?) remove support for those deprecated identifiers, we may want to warn users who try to build packages that use the deprecated identifiers.  This would require recording deprecated identifiers in a separate location.  There's some past discussion about warning for improvable license values in #1249, and with #2152's bump to the SPDX license list v3.0 the whole GPL family of identifiers have been deprecated in favor of explicit `-only` and `-or-later` forms, so there are likely quite a few packages using deprecated identifiers.

[1]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GPL-3.0.xml#L11-L12

m1kit:
# Description:
See #2153. I added deprecated license/exception warning.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/util/licenses.rb
+++ b/lib/rubygems/util/licenses.rb
@@ -8,416 +8,420 @@ class Gem::Licenses
 
   # Software Package Data Exchange (SPDX) standard open-source software
   # license identifiers
-  LICENSE_IDENTIFIERS = %w(
-      0BSD
-      AAL
-      ADSL
-      AFL-1.1
-      AFL-1.2
-      AFL-2.0
-      AFL-2.1
-      AFL-3.0
-      AGPL-1.0
-      AGPL-3.0
-      AGPL-3.0-only
-      AGPL-3.0-or-later
-      AMDPLPA
-      AML
-      AMPAS
-      ANTLR-PD
-      APAFML
-      APL-1.0
-      APSL-1.0
-      APSL-1.1
-      APSL-1.2
-      APSL-2.0
-      Abstyles
-      Adobe-2006
-      Adobe-Glyph
-      Afmparse
-      Aladdin
-      Apache-1.0
-      Apache-1.1
-      Apache-2.0
-      Artistic-1.0
-      Artistic-1.0-Perl
-      Artistic-1.0-cl8
-      Artistic-2.0
-      BSD-1-Clause
-      BSD-2-Clause
-      BSD-2-Clause-FreeBSD
-      BSD-2-Clause-NetBSD
-      BSD-2-Clause-Patent
-      BSD-3-Clause
-      BSD-3-Clause-Attribution
-      BSD-3-Clause-Clear
-      BSD-3-Clause-LBNL
-      BSD-3-Clause-No-Nuclear-License
-      BSD-3-Clause-No-Nuclear-License-2014
-      BSD-3-Clause-No-Nuclear-Warranty
-      BSD-4-Clause
-      BSD-4-Clause-UC
-      BSD-Protection
-      BSD-Source-Code
-      BSL-1.0
-      Bahyph
-      Barr
-      Beerware
-      BitTorrent-1.0
-      BitTorrent-1.1
-      Borceux
-      CATOSL-1.1
-      CC-BY-1.0
-      CC-BY-2.0
-      CC-BY-2.5
-      CC-BY-3.0
-      CC-BY-4.0
-      CC-BY-NC-1.0
-      CC-BY-NC-2.0
-      CC-BY-NC-2.5
-      CC-BY-NC-3.0
-      CC-BY-NC-4.0
-      CC-BY-NC-ND-1.0
-      CC-BY-NC-ND-2.0
-      CC-BY-NC-ND-2.5
-      CC-BY-NC-ND-3.0
-      CC-BY-NC-ND-4.0
-      CC-BY-NC-SA-1.0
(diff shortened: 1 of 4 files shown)


m1kit:
Current warning message is just `license value 'GPL-2.0+' is deprecated.  See http://spdx.org/licenses.` I think this is not user-friendly much. Any suggestions are welcome.

wking:
> I gave up using `obsoletedBy` values added by spdx/license-list-XML#392 since currently they are only in master xml, which is unstable and not recommended.

I expect these will be available in the stable SPDX output soon, and have filed spdx/LicenseListPublisher#12 to track that.

test/rubygems/test_gem_specification.rb
```diff
@@ -2932,7 +2932,7 @@ def test_validate_license_values_plus
     util_setup_validate
 
     use_ui @ui do
-      @a1.licenses = ['GPL-2.0+']
+      @a1.licenses = ['MIT+']
```
wking: `MIT+` doen't make sense, because the MIT license is unversioned and there's no clear steward who would cut later releases.   `CDDL-1.0+` might be a better exampl (more background [here][1] and [here][2]).

[1]: https://wiki.spdx.org/view/Legal_Team/or-later-vs-unclear-disambiguation#Background_.28additional.29
[2]: https://wiki.spdx.org/view/Legal_Team/later-version-clauses

test/rubygems/test_gem_specification.rb
```diff
@@ -2983,16 +2983,29 @@ def test_validate_license_with_invalid_exception
...
 
     use_ui @ui do
-      @a1.licenses = ['GPL-2.0+ WITH Autocofn-exception-2.0']
+      @a1.licenses = ['MIT WITH Autocofn-exception-2.0']
       @a1.validate
     end
 
     assert_match <<-warning, @ui.error
-WARNING:  license value 'GPL-2.0+ WITH Autocofn-exception-2.0' is invalid.  Use a license identifier from
+WARNING:  license value 'MIT WITH Autocofn-exception-2.0' is invalid.  Use a license identifier from
```
wking: I think we should use `GPL-2.0-or-later WITH Autocofn-exception-2.0` here (and a few lines above), because the extention text only makes sense for a GPL license.  Future SPDX metadata may allow detecting license/exception mismatches automatically.  We aren't there yet, but best to stay forward-compatible.

And it may be worth changing "Use a license identifier from..." -> "use identifiers from..." to avoid folks interpreting this as a license-ID issue (in this case, it's an exception-ID issue).

m1kit:
Thank you for reviewing! I would like to wait for spdx/LicenseListPublisher#12.

test/rubygems/test_gem_specification.rb
m1kit: Fixed in f81e6b2.

test/rubygems/test_gem_specification.rb
m1kit: I avoided using `GPL-2.0-or-later WITH Autocofn-exception-2.0` because `GPL-3.0-with-autoconf-exception` would be suggested as an alternative for the invalid license. Now, I noticed that deprecated license `GPL-3.0-with-autoconf-exception` being suggested is not favorable, and fixed it in f81e6b2.

lib/rubygems/util/licenses.rb
```diff
@@ -8,416 +8,420 @@ class Gem::Licenses
...
-      libtiff
-      mpich2
-      psfrag
-      psutils
-      wxWindows
-      xinetd
-      xpp
-      zlib-acknowledgement
-  ).freeze
+  LICENSE_IDENTIFIERS = {
```
segiddins: please add documentation for what the values in this hash mean

m1kit: Thank you for reviewing! Added documentation in ca3ddd2.

hsbt:
@m1kit Is this pull request still WIP status?

m1kit:
@hsbt Currently, I am not working on this issue, so this is not WIP.
I am just waiting for https://github.com/rubygems/rubygems/pull/2259#issuecomment-377777397

hsbt:
@m1kit I see. Thanks for your updates.

simi:
@m1kit any news on this? I do not see any activity in any of linked issues.

simi:
Looking at this again, `obsoleteBy` is still not exposed. I think we can go with this one for now. @m1kit would you mind rebasing this?

m1kit:
> Looking at this again, `obsoleteBy` is still not exposed. I think we can go with this one for now. @m1kit would you mind rebasing this?

Sorry, I'm a little bit busy now. If this issue is urgent, I think I cannot do this.
If not, I would like to work on this later, maybe in next January.

m1kit:
Since I am busy with a graduation thesis, I cannot spare time this month. Sorry!

jeromedalbert:
@simi Your #6926 PR did not automatically close this issue. Can this issue be closed?

deivid-rodriguez:
Thanks for the notice, @jeromedalbert! I'll close this.

PR #2266, opened 2018-04-10
Tweak warning recommendation

deivid-rodriguez:
# Description:

In #2242 I introduced a recommendation to fix the semver warning in the case where the dependency is not semantically version. However, I think the recommendation was incorrect since, for example, `< 5.2` still allows `5.2.0.rc1`, whereas something like `~> 5.1.5` does not.

I was bitten by this when I updated my dependencies according to the recommendation and the `rails-5.2.0.rc1` gem installed on my system was unintentionally activated.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -152,7 +152,7 @@ pessimistic dependency on #{dep} may be overly strict
   if #{dep.name} is semantically versioned, use:
     add_#{dep.type}_dependency '#{dep.name}', '~> #{base.join '.'}', '>= #{dep_version}'
   if #{dep.name} is not semantically versioned, you can bypass this warning with:
-    add_#{dep.type}_dependency '#{dep.name}', '>= #{dep_version}', '< #{upper_bound.join '.'}'
+    add_#{dep.type}_dependency '#{dep.name}', '>= #{dep_version}', '< #{upper_bound.join '.'}.x'
         WARNING
       end
 
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -2614,12 +2614,12 @@ end
   if d is semantically versioned, use:
     add_runtime_dependency 'd', '~> 1.2', '>= 1.2.3'
   if d is not semantically versioned, you can bypass this warning with:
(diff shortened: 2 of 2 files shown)


lib/rubygems/specification_policy.rb
```diff
@@ -152,7 +152,7 @@ def validate_dependencies # :nodoc:
   if #{dep.name} is semantically versioned, use:
     add_#{dep.type}_dependency '#{dep.name}', '~> #{base.join '.'}', '>= #{dep_version}'
   if #{dep.name} is not semantically versioned, you can bypass this warning with:
-    add_#{dep.type}_dependency '#{dep.name}', '>= #{dep_version}', '< #{upper_bound.join '.'}'
+    add_#{dep.type}_dependency '#{dep.name}', '>= #{dep_version}', '< #{upper_bound.join '.'}.x'
```
segiddins: why `.x` as the upper bound? This will continue to allow `#{upper_bound}.a` to be resolved, I believe

deivid-rodriguez: Not sure why, but it works, I think I saw it somewhere. If I use `< 5.2.x`, it no longer allows `5.2.0.rc1` whereas `< 5.2` allows it.

indirect: is it because `x` is alphabetically before `0`? 😂

deivid-rodriguez: Funny I thought this was a feature and it seems like it's a bug  :rofl:.

If there's no "supported" equivalent of the semver operator using ">", "<=", then this advice is actually misleading...

deivid-rodriguez:
Turns out that `.a` is actually documented for usage with pre-releases (at least via tests). I've added a commit to switch to using `.a` and another one to add a test to be explicitly about how `.a` and `~>` behave together.

lib/rubygems/specification_policy.rb
deivid-rodriguez: I only now understood @seggidins original comment. I'll do some more experiments to confirm if the actual non numeric digit is actually relevant. It seems better to use `.a` anyways since it's already used in other places.

bronzdoc:
@bundlerbot r+

PR #2267, opened 2018-04-11
Simplify the code that lets us call the original, non-monkeypatched Kernel#require

sbleon:
I don't see any reason to call `remove_method` before redefining it.

### DIFF
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -11,13 +11,8 @@ module Kernel
 
   RUBYGEMS_ACTIVATION_MONITOR = Monitor.new # :nodoc:
 
-  if defined?(gem_original_require) then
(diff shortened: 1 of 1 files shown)


segiddins:
I believe it suppresses a warning about redefining an existing method?

sbleon:
Thanks for the feedback, @segiddins! I didn't see any such warnings when I ran the test suite.

According to [Matz in 2005](https://groups.google.com/d/msg/comp.lang.ruby/H0Tu7vYlt_Q/lc8mkzlMf5QJ), and [Aaron Patterson last year](https://bugs.ruby-lang.org/issues/13574), there is no warning on method redefinition if you've made an alias to the old version of the method.

duckinator:
This doesn't seem to cause any warnings for me locally or in the Travis/Appveyor logs. This may have been necessary previously, but doesn't appear to be now.

@bundlerbot r+

thanks for the PR, @sbleon!

Issue #2021, opened 2017-09-26
gemspec and `git ls-files`


MSP-Greg:
I came across this when I (mistakenly) thought I could remove git from PATH in my trunk build system.

A few gems have `git ls-files` in their gemspec file, and are packaged that way.  This caused the build system to not properly install bundled gems.  Since fixed.

If git is required for installing (some) gems, maybe `where git` could be added to `gem env`?

This issue is related to:

  - [ ] Network problems
  - [X] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:

```
ruby 2.5.0dev (2017-09-26 trunk 60025) [x64-mingw32]
gem 2.6.13
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

colby-swandale:
This seems reasonable, a PR for this issue would be appreciated 👏

lucasprag:
I could work on this @MSP-Greg @colby-swandale 
I will send a PR as soon as possible

tjouan:
Just my opinion, but I think it would be better to fix the
corresponding gem specifications, I personally made the same "mistake"
in many of my gemspec files and later regretted it.

Or gem authors may keep using a shell command, but warn about the
build dependencies in the gem documentation.

About `gem env` you might have to add hg, svn, find… Also their
versions and which shell is needed, it seems complex, and I think too
late at this point.

An alternative might be a warning when building the gem, but I see
only an ugly way to do this:

    iseq = RubyVM::InstructionSequence.compile code
    if iseq.to_a[-1].flatten.grep(Hash).any? { |e| e[:mid] == :` }
      warn "[#{file}] shell command usage."
    end
(just a quick hack in Gem::Specification.load, but not acceptable IMO)

PR #2268, opened 2018-04-13
Show git path in gem env

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2021

Adds Git path to `gem env`
```
RubyGems Environment:
  ...
  - GIT EXECUTABLE: /usr/bin/git
  ...
```

still need to add a test for this...

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -847,6 +847,21 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
+  # Git binary path
+
+  def self.git_path
+    exts = ENV["PATHEXT"] ? ENV["PATHEXT"].split(";") : [""]
+    ENV["PATH"].split(File::PATH_SEPARATOR).each do |path|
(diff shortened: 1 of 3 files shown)


lib/rubygems.rb
```diff
@@ -846,6 +846,21 @@ def self.prefix
     end
   end
 
+  ##
+  # Git binary path
+
+  def self.git_path
```
segiddins: I think we might want to make this a private method in `environment_command.rb`, since that's the only place it is being used

bronzdoc: alright :+1:

bronzdoc:
I updated the PR @segiddins let me know what you think, so i can squash the commits, thanks

lib/rubygems.rb
segiddins: could call `send` in the tests ?

bronzdoc:
@bundlerbot r+

Issue #2191, opened 2018-02-15
Release plan for RubyGems 3 & 4


hsbt:
Original post from https://github.com/rubygems/rubygems/pull/2182#issuecomment-364631805

----

Based on all of these suggestions, I'd like to propose that we do something like this:

- [x] Update the [POLICIES](https://github.com/rubygems/rubygems/blob/master/POLICIES.rdoc) document to make it clear that in the future, RubyGems minor releases will drop support for Ruby versions without no support from ruby-core. (#2202)
  That means dropping releases of Ruby in the first minor after that Ruby version loses security support. We will re-evaluate anytime ruby-core changes their support levels or plans.
- [x] Update `master` to 3.0 right away, and drop support for Ruby 1.8, 1.9, 2.0, and 2.1
  We can release 3.0 whenever we are ready for the next release, but this means we can stop supporting old Ruby right away.
- [x] Open PRs for planned deprecations that include 1) what is being deprecated, 2) why we want to deprecate it, 3) when the deprecation will become an error, and 4) expected impact on Ruby and RubyGems users
  That provides opportunities for feedback, and a record of why the deprecation seems like a good idea.
- [x] After merging deprecations, release a 3.x version of RubyGems with the deprecation warnings
  This lets us give deprecation warnings only to users on modern versions of Ruby. 👍
- [x] Backport just the deprecation warnings to RubyGems 2.7.x, so they will be released in Ruby 2.5.x
  The backport only needs to support Ruby 2.5, not old Rubies. Releasing method deprecations in Ruby will give us the widest reach possible, so the ecosystem will be better prepared when we remove.

Some time later (we should discuss when... maybe Oct/Nov 2018?):

- [ ] Change `master` to 4.0
- [ ] Remove deprecated code
- [ ] Release 4.0
- [ ] Fix all known issues with the 4.x.x series of RubyGems
- [ ] Merge 4.x.x to ruby-trunk to be released with Ruby 2.6 in Dec 2018

I think this plan accomplishes all of the goals stated above:

- Ruby users will get a major version bump to make them notice the big Ruby version drop
- Releasing 3.0 can be easy and fast because we aren't breaking anything else
- Removing methods will include a major bump to 4.0, with many months for deprecation warnings
- @hsbt can stop supporting old rubies right away
- @duckinator will have documented policies and discussion PRs with feedback
- @matthewd will (eventually) get versions of RubyGems that drop old Ruby versions in minor releases

----

shevegen:
[...]

PR #2269, opened 2018-04-17
Mark deprecation to `ubygems.rb` for RubyGems 4

hsbt:
# Description:

I'm going to mark deprecate status to obsoleted methods for RubyGems 4
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/ubygems.rb
+++ b/lib/ubygems.rb
@@ -9,6 +9,6 @@
 
 # TODO: Remove this file at RubyGems 4.0
 # Based by https://bugs.ruby-lang.org/issues/14322#note-4
-warn "`ubygems.rb' is deprecated, It will be removed on or after 2018-12-01. remove `-rubygems' from your command-line"
+warn "`ubygems.rb' is deprecated, and will be removed on or after 2018-12-01. Remove `-rubygems' from your command-line, or use `-r rubygems' instead"
 
 require 'rubygems'
--- a/lib/ubygems.rb
+++ b/lib/ubygems.rb
@@ -7,5 +7,8 @@
 # See LICENSE.txt for permissions.
 #++
 
+# TODO: Remove this file at RubyGems 4.0
+# Based by https://bugs.ruby-lang.org/issues/14322#note-4
+warn "`ubygems.rb' is deprecated, It will be removed on or after 2018-12-01. remove `-rubygems' from your command-line"
 
 require 'rubygems'


hsbt:
@segiddins I don't know. `ubygems.rb` will never use on the current supported versions of Ruby. It should be removed from this repository. 

PS. I already removed its file from ruby core at Ruby 2.5.

segiddins:
But it’s installed when I `Gem update —system`, and I don’t think we should remove it without discussing the reason for doing so

segiddins:
> I already decided to remove it in RubyGems 4

I don’t think that’s fair — it’s for the RubyGems team to collectively make decisions, not just you or me.

As to my aversion to removing this file? I run `ruby -rubygems` frequently

hsbt:
> I run ruby -rubygems frequently

with Ruby 1.8? I wonder why you add `-rubygems` in 2018.

hsbt:
1. We already adopted to remove to support old rubies: https://github.com/rubygems/rubygems/issues/2191
2. I did NOT remove its file immediately. At first, We warn deprecated message at RubyGems 3. I will remove it at RubyGems 4
3. You did accept other removable pull requests for RubyGems 3 & 4.

I think your opinion is only your experience. It's not reasonable.

segiddins:
It’s not reasonable to want a discussion, rather than just making changes, saying “a decision has been made”?

hsbt:
@segiddins I'm waiting "your reasonable reason" yet.

After Ruby 1.9, rubygems is automatically activated by the prelude of Ruby. You should not add `-rubygems` on the technical side. I do not understand your reasons for this problem

segiddins:
My problem isn't the change per-se, it's the assertion of _this has been decided_ when no discussion has taken place.

> You should not add -rubygems on the technical side

a) I run 1.8 occasionally
b) `-disable-gems`

hsbt:
> a) I run 1.8 occasionally

The current master branch was not supported Ruby 1.8. It was not related this.

> b) -disable-gems

Why you add `-rubygems` with `--disable-gems`? It's not reasonable.

segiddins:
> Why you add -rubygems with --disable-gems?

Aliases

> It was not related this.

It is related. I have scripts that are meant to work on all versions of ruby, and this feels like a breaking change for no good reason

hsbt:
> It is related. I have scripts that are meant to work on all versions of ruby,

It's your personal problem.

duckinator:
@segiddins is there any reasons your scripts can't be updated to use `-rrubygems` instead of `-rubygems`?

@hsbt _I'm in favor of deprecating this_, but i'd appreciate it if you would consider re-adding it to Ruby _until RubyGems removes it._ This was a feature for at least 12 years(!), and removing it abruptly like you did in Ruby 2.5 seems risky. We want to provide time for people to adjust their scripts.

lib/ubygems.rb
```diff
@@ -7,5 +7,8 @@
 # See LICENSE.txt for permissions.
 #++
 
+# TODO: Remove this file at RubyGems 4.0
+# Based by https://bugs.ruby-lang.org/issues/14322#note-4
+warn "`ubygems.rb' is deprecated, It will be removed on or after 2018-12-01. remove `-rubygems' from your command-line"
```
duckinator: @hsbt what do you think of this message?:

```
warn "`ubygems.rb' is deprecated, and will be removed on or after 2018-12-01. Remove `-rubygems' from your command-line, or use `-r rubygems' instead"
```

duckinator:
Left a comment with some input on the warning message.

lib/ubygems.rb
hsbt: 👍

shevegen:
Early deprecation notices in advance are good, IMO.

It gives people more time to transition (with a transition period
that should take some time too).

duckinator:
@hsbt are you going to change the warning? you've got a +1 from me on the deprecation either way, but i don't want to merge it before it's ready if you're changing it.

hsbt:
@bundlerbot r+

PR #2271, opened 2018-04-20
Add `--env-shebang` flag to setup command

jfmyers9:
# Description:

With the recent changes in ruby, installing rubygems by using `ruby setup.rb` will install the `bin/bundle` and `bin/gem` with a shebang that points at the absolute path to the ruby executable. This has proven problematic for our use case. While other `gem` commands respect the `--env-shebang` option either passed through the command line or the configuration file, the `setup_command` does not.

This change adds the `--env-shebang` flag to the setup command. When provided the flag will use the `#!/usr/bin/env ruby` shebang for the gem and bundle executable. When regenerating binstub, the `--env-shebang` flag will be passed through to the internal pristine command as well.

______________

# Tasks:

- [x] Write tests


bronzdoc:
@bundlerbot r+

PR #2278, opened 2018-05-02
support SOURCE_DATE_EPOCH to make gem spec reproducible

anthraxx:
Optionally respect the SOURCE_DATE_EPOCH environment variable to be used
instead of TODAY to allow reproducible builds of created gem specs.

In case none is specified, fall back to the current time.
______________

The problem is that using TODAY will change during time which makes created artifacts not reproducible (bit by bit identical).

Spec:
https://reproducible-builds.org/specs/source-date-epoch/

Buy-in:
https://reproducible-builds.org/docs/buy-in/

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -135,7 +135,7 @@ class Gem::Specification < Gem::BasicSpecification
     :autorequire               => nil,
     :bindir                    => 'bin',
     :cert_chain                => [],
-    :date                      => TODAY,
+    :date                      => nil,
     :dependencies              => [],
     :description               => nil,
     :email                     => nil,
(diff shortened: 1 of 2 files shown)


hsbt:
I'm not familiar with SOURCE_DATE_EPOCH environment variable. Is it major on the linux distribution like debian, fedora, archlinux?

anthraxx:
That could be a good general goal. For this PR I only targeted the generated gemspec stuff as that's what made the packages we ship in arch Linux unreproducible. We don't ship any tarballs directly just extract the final gem into a DESTDIR like folder to package it with the distro specific tools.

If you want to investigate further also on tarballs stuff, you should give "diffoscope" a shot it helps find what makes an artifact not reproducible bit by bit by analyzing its content.

anthraxx:
@hsbt @segiddins So i took a look at the TarHeader and TarWriter stuff and I think its not hard to make it fully reproducible.
I would personally still prefere to make it in a separate PR, but thats your call. Either way i would take a look into that as well.

lib/rubygems/specification.rb
```diff
@@ -1714,12 +1714,14 @@ def has_conflicts?
   end
 
   ##
-  # The date this gem was created.  Lazily defaults to the current UTC date.
+  # The date this gem was created.  Lazily defaults to the current UTC date or if
+  # SOURCE_DATE_EPOCH is set as an environment variable (unix timestamp) to
+  # support reproducible builds.
```
duckinator: @anthraxx Thoughts on changing this to something like the following?

```
# The date this gem was created.
#
# If SOURCE_DATE_EPOCH is set as an environment variable, use that to support
# reproducible builds; otherwise, default to the current UTC date.
#
# Details on SOURCE_DATE_EPOCH:
# https://reproducible-builds.org/specs/source-date-epoch/
```

hsbt:
@anthraxx Thanks for your explanation. I prefer your idea.

segiddins:
👍🏻

duckinator:
@bundlerbot r+

thanks for the PR, @anthraxx!

anthraxx:
^ tm that doesn't look like i broke it :smirk_cat:

duckinator:
@bundlerbot retry

¯\_(ツ)_/¯

PR #1135, opened 2015-01-12
Don't treat inaccessible working directories as build failures.

pete:
If the user installing the gem does not have access to the directory
where the `gem install` command is run (which is common for sudo(1),
for example) then extensions will build correctly, but will then report
a build failure when trying to return to the previous working directory
raises an exception.  This fixes the bug by returning to a fallback
directory if we can't return to the original one.

An example is probably clearer:

``` sh
mkdir /tmp/test
chmod 700 /tmp/test
cd /tmp/test
sudo -u $a_non_root_user gem install sqlite3
```

(Although a non-root user is easier to demonstrate, this happened to me
while installing a gem as root from a working directory that happed to
be inside a FUSE filesystem that root couldn't see.)

In this case, rubygems will compile sqlite3's extensions correctly, and
then try to return to the previous working directory, which will raise
an Errno::EACCES, which will trigger the rescue block that is intended
to report build failures.  This patch bounces back to dest_dir in that
case, which allows installation to proceed as normal.

PR #2282, opened 2018-05-09
chdir back to original directory when building an extension fails

segiddins:
# Description:

Fixes a regression caused by https://github.com/rubygems/rubygems/pull/1135, in which the directory wouldn't be restored when building an extension failed, causing any error handlers (such as Bundler's) to be run in an unexpected directory

# Tasks:

- [x] Write tests


segiddins:
@bundlerbot r+

PR #2303, opened 2018-05-29
Support IO.copy_stream

okkez:
Implement Gem::Package::TarReader::Entry#readpartial.
It raises EOFError when call readpartial after reach EOF.

In previous version, Gem::Package::TarReader::Entry#read does not
support 2nd argument. So we cannot run following sample code.

Sample code:

```ruby
File.open(target, "rb+") do |tar|
  Gem::Package::TarReader.new(tar) do |reader|
    reader.each do |entry|
      case
      when entry.file?
        File.open(entry.full_name, "w+") do |file|
          IO.copy_stream(entry, file)
        end
      when entry.directory?
        FileUtils.mkdir_p(entry.full_name)
      else
        # do nothing
      end
    end
  end
end
```

### DIFF
--- a/lib/rubygems/package/tar_reader/entry.rb
+++ b/lib/rubygems/package/tar_reader/entry.rb
@@ -129,7 +129,7 @@ class Gem::Package::TarReader::Entry
   # Reads +len+ bytes from the tar file entry, or the rest of the entry if
   # nil
 
-  def read(len = nil, outbuf = "".b)
+  def read(len = nil)
     check_closed
 
     return nil if @read >= @header.size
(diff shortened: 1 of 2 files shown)


okkez:
`Zlib::GzipReader#read` does not support 2nd argument for output buffer like `IO#read`.
So many test are failed.

hsbt:
Thanks. I confirmed you (and @kou ) 's usecase.

hsbt:
@bundlerbot r+

hsbt:
I gave up to backport this to 2.7 branch. Because I need to rewrite to support Ruby 1.8.

PR #2305, opened 2018-05-29
Fix warning from check_that_user_bin_dir_is_in_path on Windows

MSP-Greg:
Windows standard is that all user folders begin with a base directory of `Users`, and `PATH` is often correctly capitalized.

`Gem::Installer#check_that_user_bin_dir_is_in_path` fails with a PATH that is correctly capitazlized.  PR fixes.

Thanks, Greg

# Tasks:

- [ ] Write tests


hsbt:
I'm not sure why it is needed. Because `path = path.downcase` of L704 makes downcase to path variables.

PR #2023, opened 2017-09-28
add info command to print information about an installed gem

colby-swandale:
# Description:

Quiet often i want to know some information about a gem installed in my ruby environment such as what the gem does and a link to get more information. Currently i need to find this information via RubyGems.org which is both is taxing and unnecessary for the the user.

There is currently a `specification` command which prints the specification in YAML to the user but i find this as not very user friendly.

This PR introduces the `info` command which prints a set of basic information about the given gem, such as the homepage, path and the summary in a simple and UI friendly manner:

```
$ gem info rack
  * rack (2.0.3)
        Summary: a modular Ruby webserver interface
        Homepage: http://rack.github.io/
        Path: /Users/c/.gem/ruby/2.3.1/gems/rack-2.0.3
```
This allows me to get a quick sense of what the gem does and a link that i can follow for more information if i wanted to.

Note: This command is nearly the same thing as `bundle info` that i also implemented

The `info` command also allows specifying a version, such as:

```
$ gem info rack -v 1.6.8
  * rack (1.6.8)
        Summary: a modular Ruby webserver interface
        Homepage: http://rack.github.io/
        Path: /Users/c/.gem/ruby/2.3.1/gems/rack-1.6.8
```

It will also note if the gem is a default gem as well

```
› gem info json
  * json (1.8.3)
        Summary: This json is bundled with Ruby
        Path: /Users/c/.rubies/ruby-2.3.1/lib/ruby/gems/2.3.0/gems/json-1.8.3
        Default Gem: yes
```
______________

# Tasks:

- [x] Describe the problem / feature
- [x] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2308, opened 2018-06-01
Add alias command 'i' for 'install' command

ota42y:
# Description:

In rubygems 2.x, the command that started with `i` was `install` command only.  
So we can install gem by following command like [npm](https://docs.npmjs.com/cli/install).  
`gem i bundler`

But rubygems 3.x have `info` command.  
https://github.com/rubygems/rubygems/pull/2023  

So this command doesn't work

```bash
% gem i bundler
    Ambiguous command i matches [info, install]
```

The `i` command work well in 2.x and this command is very useful as we install lots of gem.  
So it would be nice to set an alias for this command.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -71,7 +71,7 @@ class Gem::CommandManager
     :yank,
   ]
 
-  ARIAS_COMMANDS = {
+  ALIAS_COMMANDS = {
     'i' => 'install'
   }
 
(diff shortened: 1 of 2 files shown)


hsbt:
Sounds good. I also usually use `gem i foo` every day. I will merge after It's greeen status of Travis CI.

lib/rubygems/command_manager.rb
```diff
@@ -71,6 +71,10 @@ class Gem::CommandManager
     :yank,
   ]
 
+  ARIAS_COMMANDS = {
```
qsona: typo `ALIAS` ?

ota42y: oh... I'll fix 🙇

hsbt:
@bundlerbot r+

PR #2309, opened 2018-06-02
Bundler issue

MSP-Greg:
# Description:

First of all, I don't pretend to have great knowledge of RubyGems, so, this might be totally wrong...

Testing on Appveyor seemed to be broken after a rubygems-update to 2.7.6.

Investigating locally, it seemed that `Gem::Specification.stubs` &  `Gem::Specification.stubs_for` only load gemspecs from standard locations, but don't include currently loaded gems (`Gem.loaded_specs.values`).

So, I loaded bundler in the `Rake::Testtask` (only if it's included in the repo), and added the loaded gemspecs to `stubs` and `stubs_for`.  This PR also includes a commit removing the 'temp fix' for Appveyor.

Note that in my fork I had a 'false fail' due to a leftover file artifact.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -820,7 +820,7 @@ class Gem::Specification < Gem::BasicSpecification
   def self.stubs
     @@stubs ||= begin
       pattern = "*.gemspec"
-      stubs = default_stubs(pattern).concat installed_stubs(dirs, pattern)
+      stubs = Gem.loaded_specs.values + default_stubs(pattern) + installed_stubs(dirs, pattern)
       stubs = uniq_by(stubs) { |stub| stub.full_name }
 
       _resort!(stubs)
@@ -839,7 +839,7 @@ class Gem::Specification < Gem::BasicSpecification
       @@stubs_by_name[name] || []
     else
       pattern = "#{name}-*.gemspec"
-      stubs = default_stubs(pattern) + installed_stubs(dirs, pattern)
+      stubs = Gem.loaded_specs.values + default_stubs(pattern) + installed_stubs(dirs, pattern)
       stubs = uniq_by(stubs) { |stub| stub.full_name }.group_by(&:name)
       stubs.each_value { |v| _resort!(v) }
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,18 +28,6 @@ init:
       }
 
 install:
-# Temporary fix until Appveyor installs RubyGems 2.7.7
-- ps: >-
-    if ($env:ruby_version -lt '25') {
-      gem uninstall bundler -x
-      Push-Location C:\
-      gem fetch  rubygems-update -v 2.7.7
-      gem unpack rubygems-update-2.7.7.gem
-      Push-Location C:\rubygems-update-2.7.7
-      ruby setup.rb
-      Pop-Location
-      Pop-Location
-    }
 - ps: >-
     git submodule update --init --recursive
 
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -13,6 +13,15 @@ else
   require 'rubygems'
 end
 
+# If bundler gemspec exists, add to stubs
+bundler_gemspec = File.expand_path("../../../bundler/bundler.gemspec", __FILE__)
+if File.exist?(bundler_gemspec)
+  Gem::Specification.dirs.unshift File.dirname(bundler_gemspec)
+  Gem::Specification.class_variable_set :@@stubs, nil
+  Gem::Specification.stubs
+  Gem::Specification.dirs.shift
+end
+
 begin
   gem 'minitest'
 rescue Gem::LoadError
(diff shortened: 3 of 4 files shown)


hsbt:
I wonder why this issue happens only windows specified appveyor environment. So I against to add code changes targeted environment specific problem into production code.

lib/rubygems/specification.rb
```diff
@@ -816,11 +816,11 @@ def self.each_spec(dirs) # :nodoc:
 
   ##
   # Returns a Gem::StubSpecification for every installed gem
-
+
```
segiddins: nit: whitespace

MSP-Greg:
@segiddins

I suppose so, considering I could be considered a proponent of good testing...

1. I reviewed a bit more, and `stubs_for` (sort of) contains the same code as `stubs` when `@@stubs` is nil (the code loads `@@stubs` & `@@stubs_by_name`).  So, I've changed `stubs_for` to:
```ruby
  def self.stubs_for name
    stubs unless @@stubs
    @@stubs_by_name[name] ||= EMPTY
  end
```

2. Not a lot of tests currently done on `stubs` and `stubs_for`.  I added the following, would you like more?

```ruby
  def test_self_stubs_for_includes_loaded_specs
    Gem.loaded_specs['a'] = @a1
    assert_equal [@a1], Gem::Specification.stubs_for('a')

    Gem.loaded_specs.delete 'a'
    Gem::Specification.class_variable_set(:@@stubs) = nil
  end
```

3. While I was in the test file, I noticed a test (searching for stub) that didn't have an assert.  Should I add one with this PR, or do another?

https://github.com/rubygems/rubygems/blob/ffc8940c0e54a92ff16a12abf5740a90dfa8184a/test/rubygems/test_gem_specification.rb#L103-L112

Thanks, Greg

MSP-Greg:
All done.

MSP-Greg:
@segiddins

I believe Travis is an intermittent failure.  Sorry for all the tests.

Note to self - don't work on two PR's while someone seems hell-bent on informing one of everything happening in the French Open.

Anyway, this seems to solve the issue when CI has a version on Bundler installed that is different than the one included in RubyGems.

But, I was incorrect, in Gem::Specification, `stubs_for name` will keep updating (with load_specs info) as long as `stubs` is not called.  Once `stubs` is called, both are no longer updated.  I'm not sure if that is desired, but it's working here.

Maybe they should be uncoupled and update with every call?

Thanks, Greg

segiddins:
This seems reasonable to me, but I want to let hsbt have another look

Rakefile
```diff
@@ -20,7 +20,11 @@ task :setup => ["bundler:checkout"] do
 end
 
 Rake::TestTask.new do |t|
-  t.ruby_opts = %w[--disable-gems]
+  # if bundler is in RubyGems repo, require it
```
hsbt: Do not inject bundler library on `RUBY_OPTS` layer. It can be changed with `ENV['DONT_USE_BUNDLER_FOR_GEMDEPS']`

MSP-Greg: Locally, I made the change to `ENV['DONT_USE_BUNDLER_FOR_GEMDEPS']` and I have a repeatable error:
```

  1) Error:
TestGemCommandsSetupCommand#test_env_shebang_flag:
Errno::ENOENT: No such file or directory @ rb_sysopen - C:/Greg/temp/test_rubygems_6828/install/bin/bundle
    C:/Greg/GitHub/rubygems/test/rubygems/test_gem_commands_setup_command.rb:123:in `read'
    C:/Greg/GitHub/rubygems/test/rubygems/test_gem_commands_setup_command.rb:123:in `test_env_shebang_flag'
```

MSP-Greg:
@hsbt

Maybe I'm getting the hang of this.  Currently, I was requiring bundler simply so its gemspec was visible.

So, I removed the loading in the Rakefile, and added the following to test_case.rb, after the rubygem load.

```ruby
# If bundler gemspec exists, add to stubs
bundler_gemspec = File.expand_path("../../../bundler/bundler.gemspec", __FILE__)
if File.exist?(bundler_gemspec)
  Gem::Specification.dirs.unshift File.dirname(bundler_gemspec)
  Gem::Specification.class_variable_set :@@stubs, nil
  Gem::Specification.stubs
  Gem::Specification.dirs.shift
end
```

Rebased and updated PR...

Thanks for all your work on default gems, Greg

Rakefile
hsbt: It's not related to this.

I already fixed ruby core repo: https://github.com/ruby/ruby/commit/b9212a7771a5e67f3b48bbf4c76cd087a8d87e49

MSP-Greg: Good morning.  Sorry, do you mean the problem in general or the TestGemCommandsSetupCommand#test_env_shebang_flag error?

hsbt:
@bundlerbot r+

MSP-Greg:
@hsbt

Please consider removing the 'platform: windows' label, as mentioned, this is about the ruby build having an older version of bundler that the included version under test...

Thanks, Greg

PR #2297, opened 2018-05-21
Reset hooks in test cases

segiddins:
# Description:

Closes #2296. cc @MSP-Greg 

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2311, opened 2018-06-03
test_case - fix intermittent test failures?

MSP-Greg:
# Description:

Recently in my fork, I had a failure due to the intermittent tests.  Seeing as I’m the one who did the PR that exposed the instability of the test suite, I would like to see it fixed.

Conversely, as what one might call a ‘ruby windows advocate’, I’ve been in a lot repos (esp extension gems), and I’m always leery of working on test suites, as people’s views on test code seems to vary more than the app code.

As mentioned previously, I found the intermittent tests were largely caused by two issues:

A.	Hooks added but not removed by tests.  This should be solved when PR #2297 is accepted.
B.	File artifacts created by one test that cause a subsequent test to fail.

As to issue B, I see four options of fixing it, in order of ease.

1.	Add code to `test_case.rb` to cleanup file artifacts in setup or teardown.  This may affect test time.
2.	Add methods to `test_case.rb` for cleanup, but call them only from test files that require doing so.  Again, this would be done in the test files’ setup or teardown.
3.	Add above mentioned methods to each test method that requires them.
4.	Cleanup files by name in each test method.

From a stability standpoint, I think options 1 or 2 are the best solution.  If option 1 results in to large time increase, than move to option 2…
 
One problem I had before was that the below method did not remove default gemspecs:
https://github.com/rubygems/rubygems/blob/ffc8940c0e54a92ff16a12abf5740a90dfa8184a/lib/rubygems/test_case.rb#L727-L733. I don’t recall if test re ``--user-install` were an issue.

So, this PR modifies `util_clear_gems` to account for those issues, and adds it to `setup`.  

Running locally may have increased test time a small amount (3-4%), so I thought I’d see how Travis & Appveyor are affected.

Finally, there are other locations that files can be installed to (cache?),  this doesn’t affect them.

If this works/is acceptable, I'll do another PR removing calls to `util_clear_gems` from current test files.

Thanks, Greg

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -356,6 +356,7 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
     Gem.loaded_specs.clear
     Gem.clear_default_specs
     Gem::Specification.unresolved_deps.clear
+    util_clear_gems
     if Gem::USE_BUNDLER_FOR_GEMDEPS
       Bundler.reset!
     end
@@ -722,14 +723,26 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
   end
 
   ##
-  # Removes all installed gems from +@gemhome+.
+  # Removes all installed gems from +Gem.path+.
 
   def util_clear_gems
-    FileUtils.rm_rf File.join(@gemhome, "gems") # TODO: use Gem::Dirs
-    FileUtils.mkdir File.join(@gemhome, "gems")
-    FileUtils.rm_rf File.join(@gemhome, "specifications")
-    FileUtils.mkdir File.join(@gemhome, "specifications")
-    Gem::Specification.reset
+    gem_dirs = Gem.path.map { |d| File.join d, 'gems' }
+    gem_dirs.each { |d|
+      FileUtils.rm_rf d
+      FileUtils.mkdir d
+    }
+    util_clear_specs
+  end
+
+  def util_clear_specs
+    spec_dirs = Gem.path.map { |d| File.join d, 'specifications' } +
+      [Gem::BasicSpecification.default_specifications_dir]
+    spec_dirs.each { |d|
+      FileUtils.rm_rf d
+      FileUtils.mkdir d
+    }
+    # Gem.loaded_specs.clear done in setup
+    Gem::Specification.class_variable_set(:@@stubs, nil)
   end
 
   ##


lib/rubygems/test_case.rb
```diff
@@ -722,14 +723,26 @@ def util_remove_gem(spec)
...
   ##
-  # Removes all installed gems from +@gemhome+.
+  # Removes all installed gems from +Gem.path+.
 
   def util_clear_gems
-    FileUtils.rm_rf File.join(@gemhome, "gems") # TODO: use Gem::Dirs
-    FileUtils.mkdir File.join(@gemhome, "gems")
-    FileUtils.rm_rf File.join(@gemhome, "specifications")
-    FileUtils.mkdir File.join(@gemhome, "specifications")
-    Gem::Specification.reset
```
segiddins: should we keep calling `Gem::Specification.reset` ?

MSP-Greg: First, I moved it to `util_clear_specs`, then I looked at it, and saw it called  `Gem.pre_reset_hooks` & ` Gem.post_reset_hooks`.  Given the unresolved issues with hooks, I then ran the tests locally, they passed, then here, and they passed.  Decided to leave as is and see what happens...

I don't think the stability issues are going to all disappear in a few commits.  I think your hook PR and this will get it manageable.

What would you like me to do?  Thanks, Greg

segiddins: Sounds fine, just wanted to make sure it was removed intentionally

MSP-Greg: In hindsight, I should have left in it and commented it.  Does `[skip ci]` work here, or does it retest?

duckinator:
This looks okay to me. @segiddins @MSP-Greg, are there any reasons not to merge this?

MSP-Greg:
@duckinator

Speaking for myself, back when I was running a lot of testing in my fork, there were intermittent failures, 
which also appeared here.  @segiddins fixed the 'hook' issue, and the Appveyor bundler problem got fixed (not intermittent, but it 'contaminated' the tests).

So, tests seem to be stable.  Long story short, maybe wait for more test runs and see if intermittent tests are still an issue...

hsbt:
I didn't understand what this pull request was resolved.

>So, tests seem to be stable. 

Should we still need this pull request?

MSP-Greg:
> Does we still need this pull request?

Agree that tests have been stable for a while.

A quick look showed the most recent that might involve file artifacts was https://travis-ci.org/rubygems/rubygems/jobs/378157833.  Note that that was in Travis, not Appveyor.  Back when I was running the tests, the file related errors were more common on Travis...

Hence, I'll close...

Issue #2301, opened 2018-05-26
Gem selection - Windows platform (32 vs 64) issue


MSP-Greg:
Due to the fact that I often test against many ruby trunk builds, my default is to install gems with the `--user-install` option.

While working on scripts for building fat-binary gems, and also installing them (which works correctly), it seems that if one has fat binary gems installed for both 32 & 64 bit platforms, when requiring them, the correct gem is not always chosen, and hence the require fails.

This may not be clear, so one thing this does is that 'matching' gems exist in my user gem folder. There are two entries in cache, gems, and specifications, one being `x86-mingw32` and the other `x64-mingw32`...

Since I'm mostly familiar with RubyGems test code, I'm looking thru the `require` code for where the bug is.  If anyone has any hints or tips, feel free to suggest where I should be looking.

See [Appveyor log](https://ci.appveyor.com/project/MSP-Greg/testing-appveyor/build/19)

Thanks, Greg

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [X] Other

Here are my current environment details:

```
RUBYGEMS VERSION: 2.7.7
RUBY VERSION: 2.5.1 (2018-03-29 patchlevel 57) [x64-mingw32]
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

MSP-Greg:
Fixed by PR #2339.

PR #2312, opened 2018-06-04
Fix for installing both 32 & 64 bit mingw binary gems in --user-install

MSP-Greg:
## Description:

At present, if a Windows user (mingw) installs both 32 and 64 bit binary gems in `--user-install` (assuming the same ABI version), the following occurs:

1. `gem list` shows both 32 & 64 bit gems regardless of ruby build being used.
2. `require` fails one build or the other.  I think 32 bit fails because 'x86' sorts after `x64`.

This PR fixes both issues.  Rather than using StubSpec data, platform info is derived from a regex of the `loaded_from` filename.  Fully tested locally.

See issue #2301.  Also, PR #2119 affects this (and allows it to work), as, although 32 & 64 bit binary gems have separate gem folders and spec names, they share a common bin stub.  If a bin stub has a hard coded path, it will override PATH and possibly choose the wrong ruby version...

Thanks, Greg

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -782,8 +782,18 @@ class Gem::Specification < Gem::BasicSpecification
   private_class_method :default_stubs
 
   def self.installed_stubs dirs, pattern
-    map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
-      Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
+    # if mingw, derive platform from file name (loaded_from)
+    if RUBY_PLATFORM.match('mingw')
+      map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
+        Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
+      end.select do |s|
+        (plat = s.loaded_from[/(x\d{2}-mingw32)\.gemspec/,1]) ?
+          Gem::Platform.match(plat) : true
+      end
+    else
+      map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
+        Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
+      end
     end
   end
   private_class_method :installed_stubs


MSP-Greg:
At present this PR does not have a test.  The below test fails before the PR, passes after:

```ruby
def test_installed_stubs_mingw
  skip('Windows MinGW specific') unless RUBY_PLATFORM.match('mingw')
  orig_platform = Gem::Platform.dup
  cur_plat = Gem::Platform.new RUBY_PLATFORM
  Gem.platforms = ['ruby', cur_plat]

  # create user spec
  user_spec_dir = "#{Gem.user_dir}/specifications"
  FileUtils.mkdir_p(user_spec_dir)  unless Dir.exist? user_spec_dir
  # dirs doesn't include user ?
  dirs = (Gem::Specification.dirs << user_spec_dir)

  fn = 'mingw' ; p32 = 'x86-mingw32'
  v  = '1.1.1' ; p64 = 'x64-mingw32'

  # spec_r = Gem::Specification.new(fn, v)
  spec32 = Gem::Specification.new(fn, v) { |s| s.platform = p32 }
  spec64 = Gem::Specification.new(fn, v) { |s| s.platform = p64 }

  # File.open File.join(user_spec_dir, "#{fn}-#{v}.gemspec"), 'w' do |io|
  #   io.write spec_r.to_ruby
  # end
  File.open File.join(user_spec_dir, "#{fn}-#{v}-#{p32}.gemspec"), 'w' do |io|
    io.write spec32.to_ruby
  end
  File.open File.join(user_spec_dir, "#{fn}-#{v}-#{p64}.gemspec"), 'w' do |io|
    io.write spec64.to_ruby
  end
  t = Gem::Specification.send(:installed_stubs, dirs, '*.gemspec')

  assert_equal 1, t.length
  assert_equal cur_plat, t.first.platform

  Gem.platforms = orig_platform
end
```

Thanks, Greg

lib/rubygems/specification.rb
```diff
@@ -782,9 +782,15 @@ def self.default_stubs pattern
   private_class_method :default_stubs
 
   def self.installed_stubs dirs, pattern
-    map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
+    t = map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
       Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
     end
+    # if mingw, derive platform from file name (loaded_from)
+    !RUBY_PLATFORM.match('mingw') ? t :
```
hsbt: Please rewrite without the ternary operator.

MSP-Greg: Changed & rebased

hsbt:
I didn't understand this problem.

@MSP-Greg Can you provide the reproduction instructions? and describe problem "shortly".

MSP-Greg:
@hsbt

It is an edge case.  To repo it:

1. You'll need to install 32 and 64 bit versions of the same Ruby build.

2. Find a fat binary gem that is published in both 32 and 64 bit.  I've got puma files [here](https://ci.appveyor.com/project/MSP-Greg/puma-fb/build/artifacts) for 2.0 thru 2.5

3. With each Ruby build, install the fat binary gems using `--user-install`.

4. Do a `gem list` from both the 32 & 64 bit Ruby builds.  Both will show both the 32 and the 64bit fat binaries.

5. I think the 64 bit build will load the gem, but the 32 bit will not.  I think this is because 'x64' sorts before 'x86'.

Sorry it's not simpler...  There are some Powershell files that will make things easier, if you're interested...

Greg

MSP-Greg:
@hsbt 

1. See edit in prev message.

2. Test listed above adds 32 & 64 bit fb gemspecs into `user_spec_dir`, then calls `installed_stubs`.  Before the PR, both gemspecs are returned/included.  After it, only the correct one is.  I used the regex on the file name to avoid the yaml parsing, which is the point of a stub...

FYI, I discussed 32 bit builds in Rubyinstaller2, and they will be dropped in the future, but not quite yet...

Thanks, Greg

MSP-Greg:
@hsbt

Would it be helpful if I created a script run on Appveyor to show the problem?

Greg

MSP-Greg:
I may have previously done an example on Appveyor.  Here's a new one, which also runs `gem list`.

https://ci.appveyor.com/project/MSP-Greg/testing-appveyor/build/20

Note that `gem list` shows both EventMachine 32 & 64 bit gems, regardless of which Ruby build is running.

Trying to require eventmachine in the 32 bit Ruby results in the error because, as shown in `gem list`, 'x64-mingw32' sorts before 'x86-mingw32'.  Hence, regardless of the Ruby build, the 64 bit gem is activated.

Thanks, Greg

MSP-Greg:
I updated the appveyor build, it now applies the PR.

https://ci.appveyor.com/project/MSP-Greg/testing-appveyor/build/20

Hopefully this will make the issue clear to readers unfamiliar with (or no access to) Windows.

Both 32 & 64 bit have a 'pre PR' and 'post PR' test section.  In both pre sections, `gem list` shows both 32 & 64 bit eventmachine gems in both 32 & 64 bit Ruby builds, and requiring eventmachine doesn't work in 32 bit Ruby.

After the PR, all works as it should.  This should be able to also serve as reproduction instructions, as all pertinent to the issue are shown in yellow.  Commands to switch ruby versions are not shown, but the current version is shown at the top of each section, also in yellow.

Thanks, Greg

lib/rubygems/specification.rb
```diff
@@ -782,8 +782,18 @@ def self.default_stubs pattern
   private_class_method :default_stubs
 
   def self.installed_stubs dirs, pattern
-    map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
-      Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
+    # if mingw, derive platform from file name (loaded_from)
```
segiddins: does this need to be hard-coded for `mingw`, or should it apply for any platform ?

MSP-Greg: Looking over the code and `Gem::StubSpecification::StubLine`, I think the following would work, and would be much more generic:

```ruby
  def self.installed_stubs dirs, pattern
    map_stubs(dirs, pattern) do |path, base_dir, gems_dir|
      Gem::StubSpecification.gemspec_stub(path, base_dir, gems_dir)
    end.select { |s| Gem::Platform.match s.platform }
  end
```

Thoughts?  Also, should I add the test?

segiddins: That does look a lot better to me

MSP-Greg: Ok.  Add the test?  Thanks, Greg

MSP-Greg:
My opinion is that while working in platform A, nothing in RubyGems should reference gems that are not compatible with that platform.

When I said there were two failures, I was just checking using `test_gem_specification.rb`.  Running the full suite, results are:

```
2012 runs, 5915 assertions, 25 failures, 3 errors, 39 skips
```

I've got some more tests to fix...

Greg

MSP-Greg:
@eregon,

Good day and sorry for the ping.  Due to your work with JRuby, TruffleRuby, spec, & mspec, I suspect you work with multiple platforms.  BTW, thanks for that work.

I often end up testing using multiple revisions of the same Ruby ABI version (`RbConfig::CONFIG['ruby_version']`).  Sometimes that requires an installed gem.  If the gem is installed with the `--user-install` option, I can install it once and have it shared by all the different revision builds I'm testing with.

This PR involves issues with multiple ruby platform installations sharing the same [`Gem.user_dir`](https://github.com/rubygems/rubygems/blob/d5cbe948e90c826133a94a56b8df1250f48d8fb6/lib/rubygems/defaults.rb#L76-L80) folder and the installation of platform specific gems (with the `--user-install` option).

Do JRuby and/or TruffleRuby share the same `Gem.user_dir` as an MRI installation?

If not, never mind.  If they do, have you come across issues with `--user-install` gems?

Thanks, Greg

MSP-Greg:
@eregon

Thank you for all the info.  Very helpful.

I started using `--user-install` when I needed gems for testing 'X', and somewhere, 'X' on trunk went from working to not working.  Using `--user-install` saves me the trouble of installing the gems in every trunk revision while I manually 'bi-sect'...

The issue in this PR occurs when one uses `--user-install`, and installs 32 bit & 64 bit fat-binary gems with matching versions of Ruby.  IOW, the two gems install to the same user folder, but they are different platforms.

At present, RubyGems can't resolve them.  I made the PR specific to Windows, since I suspected that Windows is the only OS where a user folder based on

```
RUBY_ENGINE + RbConfig::CONFIG['ruby_version']
```
can have two platform (32 & 64 bit) builds using it.  The issue addressed in the PR wouldn't be an issue if the user folder was based on:

```
RUBY_ENGINE + RbConfig::CONFIG['ruby_version'] + RUBY_PLATFORM
```

but changing that would be a big breaking change, and it wouldn't allow 'standard ruby' gems to be shared between 32 & 64 bit...

> I don't know if there are Ruby managers on Windows

I think there are some, for my installs matching Appveyor, I've got a PowerShell script that changes the PATH.  From my root GitHub folder, `.\rv.ps1 23` switches to 32 bit Ruby 2.3, `.\rv.ps1 26-x64` switches to 64 bit trunk.  It does pick up the user/bin folder.

Thanks again, Greg

MSP-Greg:
@segiddins, @hsbt

I have updated my branch with all mentioned changes, but I'm not sure what to do with existing tests that now have failures/errors.  Part of the problem may be some of us are not familiar with other platforms, with the exception of @hsbt?

An example is:
https://github.com/rubygems/rubygems/blob/1923fb0603208f7a4fa2ee384f454d662e114d75/test/rubygems/test_gem_commands_query_command.rb#L593-L620
This seems to expect a return with platforms `universal-Darwin` and `x86-linux`.  I would assume that both are not installable on any platform.  Is that correct?

Other tests may be failing because I'm testing on Windows, and the test gems are not compatible with my platform.

So, how should the tests be changed?  I can go thru all of them and change `Gem.platforms` to make them work, or...

This all centers around the whether gems that are not installable on one's current Ruby instance should be visible or activated.  Also, there are some tests using platform of 'unknown'.

Thoughts?

segiddins:
I dont see why `gem query` should be restricted to the current platform, but maybe that could be a flag?

MSP-Greg:
> I don't see why gem query should be restricted to the current platform

And you know the difference.  Many new users do not.

Anyway, locally it is, remotely it is not.  `gem uninstall` shows all installed gems, regardless of platform.

MSP-Greg:
@segiddins

> I don't see why gem query should be restricted to the current platform

After being AFK --

Re UI's, I adhere to the 'distracted user/new user/KISS' concept.  If I do a `gem list`, I only want to see gems that can be used on the engine/platform I'm currently using.  Anything more is unclear.  Same thing for a local `gem query`.

Advanced users know that they can inspect several folders to see what gems are installed, as distinct from what can be used on the current engine/platform.

That said, I believe that the `select` block that I added to `installed_stubs` can be moved to `stubs_for`, and it will allow the correct gem to be required, while leaving other gems commands as they are now.

What would you like me to do?

MSP-Greg:
Closed, see #2339

PR #2319, opened 2018-06-13
[WIP] add advisory command to rubygems

shlok007:
# Description:
This introduces a new `advisory` command to rubygems which can be used to mark gems as vulnerable from the CLI. It is a part of the project of adding security advisories notifications to ruby gem versions.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/advisory_command.rb
+++ b/lib/rubygems/commands/advisory_command.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+require 'rubygems/command'
+require 'rubygems/package'
+require 'rubygems/local_remote_options'
+require 'rubygems/version_option'
+require 'rubygems/gemcutter_utilities'
+
+class Gem::Commands::AdvisoryCommand < Gem::Command
+  include Gem::VersionOption
+
+  def initialize
+    super 'advisory', 'add advisories to gem versions'
+
+    add_version_option
+    add_platform_option
+  end
+
+  def arguments # :nodoc:
+    "GEM  name of the gem to be marked vulnerable"
+  end
+
+  def description # :nodoc:
+    <<-EOF
+The advisory command allows you to add advisories to a gem version.
+
+An advisory can be any vulnerability in the gem that makes the gem
+either unusable or risky to use. The risk can or cannot be a security issue.
+
+    EOF
+  end
+
+  def usage # :nodoc:
+    "gem advisory GEM [options]"
+  end
+
+  def execute
+    version = get_version_from_requirements(options[:version])
+    platform  = get_platform_from_requirements(options)
+
+    if version
+      send_advisory(version, platform)
+    else
+      say "A version argument is required: #{usage}"
+      terminate_interaction
+    end
+  end
+
+  private
+
+  def send_advisory(version, platform)
+    gem_name = get_one_gem_name
+    alert_warning "Once marked, gem advisories cannot be unmarked for that particular version."
+    title = ask "Title:"
+    description = ask "Description:"
+    url = ask "Url:"
+    ask_yes_no("Are you sure you want to proceed adding the vulnerablility for #{gem_name}-#{version}?",true)
+    say "Recording advisory for #{gem_name}-#{version} ..."
+    response = {} # replace with the request to rubygems.org API
+    say = response
+  end
+
+  def get_version_from_requirements(requirements)
+    requirements.requirements.first[1].version
+  rescue
+    nil
+  end
+
+  def get_platform_from_requirements(requirements)
+    Gem.platforms[1].to_s if requirements.key? :added_platform
+  end
+end
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -368,6 +368,10 @@ class Gem::Version
       end.reduce(&:concat)
   end
 
+  def is_vulnerable?
+    
+  end
+
   protected
 
   def _version
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -35,6 +35,7 @@ class Gem::CommandManager
   include Gem::UserInteraction
 
   BUILTIN_COMMANDS = [ # :nodoc:
+    :advisory,
     :build,
     :cert,
     :check,


lib/rubygems/version.rb
```diff
@@ -368,6 +368,10 @@ def canonical_segments
       end.reduce(&:concat)
   end
 
+  def is_vulnerable?
```
segiddins: why is this being added onto the Version object ?

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,71 @@
...
+  def send_advisory(version, platform)
+    gem_name = get_one_gem_name
+    alert_warning "Once marked, gem advisories cannot be unmarked for that particular version."
+    title = ask "Title:"
+    description = ask "Description:"
+    url = ask "Url:"
+    ask_yes_no("Are you sure you want to proceed adding the vulnerablility for #{gem_name}-#{version}?",true)
+    say "Recording advisory for #{gem_name}-#{version} ..."
+    response = {} # replace with the request to rubygems.org API
+    say = response
```
duckinator: Is this supposed to just be `say response`? It's strange to define a local variable as the last line of a method.

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  def get_version_from_requirements(requirements)
+    requirements.requirements.first[1].version
+  rescue
```
duckinator: Can you rewrite this to avoid using `rescue` for flow control, and add some documentation?

duckinator:
hi! I left some comments on what you have so far &mdash; let me know if you need me to clarify anything. :slightly_smiling_face:

duckinator:
@shlok007 are you still planning to work on this?

shlok007:
@duckinator oh yes yes! I am working on the server side code for now. Thank you so much for the review. This PR is incomplete as of right now. I'll push a bundle of changes completing the feature request in a few days.

duckinator:
looking forward to it! :slightly_smiling_face:

lib/rubygems/version_option.rb
```diff
@@ -72,6 +72,5 @@ def add_version_option(task = command, *wrap)
         options[:explicit_prerelease]
     end
   end
-
```
zoras: revert this change

duckinator:
@shlok007 this is looking good so far! :slightly_smiling_face:

lib/rubygems/advisory_option.rb
```diff
@@ -0,0 +1,17 @@
...
+##
+# Utility methods for using the RubyGems API.
+
+module Gem::AdvisoryOption
+
+  ##
+  # Option to ignore a CVE while using the gem audit command.
+
+  def add_ignore_cve_option
+    add_option('-i', '--ignore-cve CVENUMBER', 'Ignore CVE while gem audit', ) do |value, options|
```
segiddins: `during gem audit`

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,85 @@
...
+  def arguments # :nodoc:
+    "GEM  name of the gem to be marked vulnerable"
+  end
+
+  def description # :nodoc:
+    <<-EOF
+The advisory command allows you to add advisories to a gem version.
+
+An advisory can be any vulnerability in the gem that makes the gem
+either unusable or risky to use. The risk can or cannot be a security issue.
```
segiddins: `The risk does not need to be a security issue` ?

lib/rubygems/commands/audit_command.rb
```diff
@@ -0,0 +1,93 @@
...
+The audit command allows you to check advisories reported for a gem version.
+
+An advisory can be any vulnerability in the gem that makes the gem
+either unusable or risky to use. The risk can or cannot be a security issue.
+
+    EOF
+  end
+
+  def usage # :nodoc:
+    "#{program_name} GEM -v VERSION [-p PLATFORM]"
```
segiddins: should this allow being called without a gem (or a version) to audit the set of all installed gems?

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  private
+
+  def send_advisory(version, platform)
+    gem_name = get_one_gem_name
+    alert_warning "Once marked, gem advisories cannot be unmarked for that particular version."
+    title = ask "Title:"
+    description = ask "Description:"
+    url = ask "Url:"
+    cve = ask "Cve:"
```
zoras: capitalize to `CVE:`

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,85 @@
...
+      }
+      data['platform'] = platform if platform
+
+      request.set_form_data data
+    end
+    say response.body
+  end
+
+  def get_version_from_requirements(requirements)
+    requirements ? requirements.requirements.first[1].version : nil
```
zoras: It would be great to if the advisory/audit command could auto detect the installed/latest gem version here instead of supplying the gem's version every-time for auditing for each gem. 
The version can be extracted with `gem list <gemname>` and piping it to sed regex or similar.

lib/rubygems/commands/advisory_command.rb
hmistry: Do all security vulnerabilities have a CVE since you require a CVE?

Probably change the description to match current implementation and restrictions, and later change it as you add more scenarios to the feature.

lib/rubygems/commands/advisory_command.rb
```diff
@@ -0,0 +1,85 @@
...
+The advisory command allows you to add advisories to a gem version.
+
+An advisory can be any vulnerability in the gem that makes the gem
+either unusable or risky to use. The risk can or cannot be a security issue.
+
+    EOF
+  end
+
+  def usage # :nodoc:
+    "#{program_name} advisory GEM -v VERSION [-p PLATFORM]"
```
hmistry: Remove the word advisory because it prints out "advisory" twice.  See your [screenshot](https://lifeinoss.files.wordpress.com/2018/08/screen-shot-2018-08-14-at-8-58-50-am.png).

lib/rubygems/commands/advisory_command.rb
hmistry: In future, I'd consider how the warning and confirmation is presented because you'll find users don't really read such things so it'll be missed. From your screenshots, there's quite a bit of text condensed and easily buries the information. One quick thing I can think of is add new line space before/after warning and confirmation so things are grouped by context.

Since there's no way to edit an advisory for now, you may want to alert user and hopefully they check for typos, and ensure they have good descriptions.

hmistry: What happens if I don't put one of the fields like CVE or description or ? Basically what are the failure messages like?

lib/rubygems/commands/advisory_command.rb
hmistry: @zoras I think that makes sense for audit but not advisory. Advisory flags a gem's specific version with a sec issue info so it doesn't make sense for it to auto-detect all your gems.

lib/rubygems/commands/audit_command.rb
hmistry: @segiddins Would that duplicate the function of `bundle audit` except being for all gems on system and not just for bundled ones?

lib/rubygems/commands/audit_command.rb
```diff
@@ -0,0 +1,93 @@
...
+            say "URL : #{advisory[2]}"
+            say "CVE: #{advisory[3]}"
+            say "\n"
+          end
+        end
+      else
+        say details['advisories']
+      end
+    end
+  end
```
hmistry: Be consistent with your formatting (e.g. Url and CVE) between "form" (i.e. advisory form) and "display" (i.e. audit response). Choose whichever way you want and be consistent so if you choose "CVE" to be all caps, then gem advisory should ask for "CVE" not "Cve" and gem audit displays "CVE". Similar for Url, choose whether it's all caps or not and be consistent. Title makes more sense to be capitalized and that's fine because it's consistent between the 2 commands.

hmistry: A future addition to consider some day - I like "Criticality" and "Solution" information in the advisories shown by `bundle audit`. The Solution can however be limited to the vulnerable gem version and higher... I doubt people will be rolling back to lower patched major versions. 🙂

hsbt:
It's good to implrement this as rubygems plugin. I'm closing this.

PR #2323, opened 2018-06-18
Dont warn on prerelease dependencies with less than ("<")

deivid-rodriguez:
# Description:

I just saw this warning when installing my gem:

```
WARNING:  prerelease dependency on rails (< 6.0.x, >= 5.2) is not recommended
```

I can see the rationale for the "prerelease warning" but doesn't look reasonable when the requirement is using "<", since in that case we're filtering out prereleases. Right?
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -125,6 +125,8 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
 
       seen[dep.type][dep.name] = dep
 
+      dep_requirements = dep.requirement.requirements
+
       prerelease_dep = dep.requirements_list.any? do |req|
         Gem::Requirement.new(req).prerelease?
       end
@@ -132,8 +134,8 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
       warning_messages << "prerelease dependency on #{dep} is not recommended" if
           prerelease_dep && !version.prerelease?
 
-      overly_strict = dep.requirement.requirements.length == 1 &&
-          dep.requirement.requirements.any? do |op, version|
+      overly_strict = dep_requirements.length == 1 &&
+          dep_requirements.any? do |op, version|
             op == '~>' and
                 not version.prerelease? and
                 version.segments.length > 2 and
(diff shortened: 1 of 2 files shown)


segiddins:
In bundler, this sort of requirement will start allowing pre-releases

deivid-rodriguez:
I don't understand, can you elaborate?

deivid-rodriguez:
Elaborating myself, note that this is only removing a warning message in some cases where I think doesn't make sense. It's not changing behavior other than that.

segiddins:
Right, but if you have `">= 1", "< 2.pre"` as a dependency, bundler will resolve that differently than `"~> 1.0"`, even though they seem equivalent

deivid-rodriguez:
You mean that they may get different, both correct, resolutions? I think that's fine, this is just a recommendation. In any case, that seems like a comment against introducing https://github.com/rubygems/rubygems/pull/2266 and https://github.com/rubygems/rubygems/pull/2242, not this PR. Or I'm missing something?

deivid-rodriguez:
Ok, so I don't understand then. If I specify, `gem "rails", "< 6.0.x", ">= 5.2"` as a dependency of my gem, I don't think I'm depending on any pre-release version of `rails`, so the warning `WARNING:  prerelease dependency on rails (< 6.0.x, >= 5.2) is not recommended` doesn't make sense to me. Are you saying that this requirement actually allows prereleases of `rails`, i.e., that https://github.com/rubygems/rubygems/pull/2266 is incorrect? That's definitely not my experience with `bundler`.

segiddins:
> Are you saying that this requirement actually allows prereleases of rails

I believe that is the case, though I could be wrong

deivid-rodriguez:
I think you might be wrong, yeah. I just tried it and got:

```
Resolving dependencies.....
Bundler could not find compatible versions for gem "decidim":
  In snapshot (Gemfile.lock):
    decidim (= 0.13.0.pre1)

  In Gemfile:
    decidim (= 0.13.0.pre1)

    my_test_gem was resolved to 0.1.0, which depends on
      decidim (< 0.13.x, >= 0.12)

```

If I change the dependency to `>= 0.12, < 0.13`, it properly resolves and allows the prerelease.

deivid-rodriguez:
This warning might be confusing and not very easy to understand / agree with, so we might consider entirely dropping it as you suggested in https://github.com/rubygems/rubygems/issues/2228#issuecomment-372076046. But if the warning stays, I think this PR should go in.

deivid-rodriguez:
I opened #2351 as a (better, I think) alternative.

bronzdoc:
closing in favor of https://github.com/rubygems/rubygems/pull/2351

PR #2333, opened 2018-06-26
Make spec reset more informative

bronzdoc:
# Description:

By JimPanic
```
According to http://stackoverflow.com/questions/17936340/unresolved-specs-during-gemspecification-reset this warning is (also?) shown when gem doesn't know what version of a gem it should use if multiple are available. The message however does not say that.

This change finds all installed versions and displays them alongside the original dependency and its version. It also proposes a gem cleanup.
```

This add a test for it.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1280,11 +1280,17 @@ class Gem::Specification < Gem::BasicSpecification
     unresolved = unresolved_deps
     unless unresolved.empty? then
       w = "W" + "ARN"
-      warn "#{w}: Unresolved specs during Gem::Specification.reset:"
+      warn "#{w}: Unresolved or ambigious specs during Gem::Specification.reset:"
       unresolved.values.each do |dep|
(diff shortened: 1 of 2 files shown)


lib/rubygems/specification.rb
```diff
@@ -1280,11 +1280,17 @@ def self.reset
     unresolved = unresolved_deps
     unless unresolved.empty? then
       w = "W" + "ARN"
-      warn "#{w}: Unresolved specs during Gem::Specification.reset:"
+      warn "#{w}: Unresolved or ambigious specs during Gem::Specification.reset:"
       unresolved.values.each do |dep|
         warn "      #{dep}"
+        
+        versions = Gem::Specification.find_all_by_name(dep.name)
```
segiddins: can you please add a test that hits this branch as well? Thanks!

bronzdoc:
Done @segiddins

bronzdoc:
@bundlerbot r+

PR #971, opened 2014-07-11
improve performance of find_files_from_load_path

tenderlove:
In order to support case-insensitive file systems, Dir#[] is slower in
trunk Ruby:

  https://bugs.ruby-lang.org/issues/10015

This patch speeds up `find_files_from_load_path` a little on 2.1.2
(about 10%) and a lot on Ruby trunk (10x faster).  Here is my benchmark:

``` ruby
require 'benchmark/ips'

class Hello
  def self.old_find_files_from_load_path glob # :nodoc:
    $LOAD_PATH.map { |load_path|
      Dir["#{File.expand_path glob, load_path}#{Gem.suffix_pattern}"]
    }.flatten.select { |file| File.file? file.untaint }
  end

  def self.new_find_files_from_load_path glob # :nodoc:
    search = "#{glob}#{Gem.suffix_pattern}"
    $LOAD_PATH.map { |load_path|
      Dir.chdir(load_path) {
        Dir[search].map! { |f| File.join load_path, f }
      }
    }.flatten.select { |file| File.file? file.untaint }
  end
end

search = "minitest/*_plugin.rb"

Benchmark.ips do |x|
  x.report('old') do
    Hello.old_find_files_from_load_path search
  end

  x.report('new') do
    Hello.new_find_files_from_load_path search
  end
end
```

Results on 2.1.2:

```
[aaron@higgins rubygems (master)]$ ruby -v -I lib flp.rb
ruby 2.1.2p95 (2014-05-08 revision 45877) [x86_64-darwin13.0]
Calculating -------------------------------------
                 old       283 i/100ms
                 new       304 i/100ms
-------------------------------------------------
                 old     2903.8 (±4.3%) i/s -      14716 in   5.078291s
                 new     3180.5 (±4.7%) i/s -      16112 in   5.078193s
```

Results against trunk:

```
[aaron@higgins rubygems (master)]$ ruby -v -I lib flp.rb
ruby 2.2.0dev (2014-07-09 trunk 46759) [x86_64-darwin13]
Calculating -------------------------------------
                 old         8 i/100ms
                 new        85 i/100ms
-------------------------------------------------
                 old       84.6 (±7.1%) i/s -        424 in   5.046520s
                 new      858.7 (±3.3%) i/s -       4335 in   5.053763s
```

PR #2336, opened 2018-06-27
Speed up globbing relative to given directories

segiddins:
# Description:

This is picking up @tenderlove's improvements in https://github.com/rubygems/rubygems/pull/971.

Benchmark:

```ruby
class Hello
  def self.old_find_files_from_load_path glob # :nodoc:
    $LOAD_PATH.map { |load_path|
      Dir["#{File.expand_path glob, load_path}#{Gem.suffix_pattern}"]
    }.flatten.select { |file| File.file? file.untaint }
  end

  def self.aaron_find_files_from_load_path glob # :nodoc:
    search = "#{glob}#{Gem.suffix_pattern}"
    $LOAD_PATH.map { |load_path|
      Dir.chdir(load_path) {
        Dir[search].map! { |f| File.join load_path, f }
      }
    }.flatten.select { |file| File.file? file.untaint }
  end
  
  def self.new_find_files_from_load_path glob # :nodoc:
    $LOAD_PATH.map { |load_path|
      Dir.glob(glob_with_suffixes, base: load_path).map! {|f| File.join(load_path, f) }
    }.flatten.select { |file| File.file? file.untaint }
  end
end

Benchmark.ips do |x|
  x.report('old') do
    Hello.old_find_files_from_load_path search
  end

  x.report('aaron') do
    Hello.aaron_find_files_from_load_path search
  end
  
  x.report('new') do
    Hello.new_find_files_from_load_path search
  end
  
  x.compare!
end
```

Results (on Ruby 2.5):

```
                 old    94.000  i/100ms
               aaron   297.000  i/100ms
                 new   729.000  i/100ms
                 old    840.927  (±28.2%) i/s -      3.666k in   5.094465s
               aaron      2.843k (±19.2%) i/s -     13.365k in   5.012555s
                 new      7.457k (± 9.4%) i/s -     37.179k in   5.046804s

Comparison:
                 new:     7457.4 i/s
               aaron:     2842.6 i/s - 2.62x  slower
                 old:      840.9 i/s - 8.87x  slower
```

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/util.rb
+++ b/lib/rubygems/util.rb
@@ -116,4 +116,16 @@ module Gem::Util
     end
   end
 
+  ##
+  # Globs for files matching +pattern+ inside of +directory+,
+  # returning absolute paths to the matching files.
+
+  def self.glob_files_in_dir(glob, base_path)
+    if RUBY_VERSION >= "2.5"
+      Dir.glob(glob, base: base_path).map! {|f| File.join(base_path, f) }
+    else
+      Dir.glob(File.expand_path(glob, base_path))
+    end
+  end
+
 end
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -525,8 +525,9 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   def self.find_files_from_load_path glob # :nodoc:
+    glob_with_suffixes = "#{glob}#{Gem.suffix_pattern}"
     $LOAD_PATH.map { |load_path|
-      Dir["#{File.expand_path glob, load_path}#{Gem.suffix_pattern}"]
+      Gem::Util.glob_files_in_dir(glob_with_suffixes, load_path)
     }.flatten.select { |file| File.file? file.untaint }
   end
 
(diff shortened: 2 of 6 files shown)


lib/rubygems/util.rb
```diff
@@ -116,4 +116,18 @@ def self.traverse_parents directory, &block
     end
   end
 
+  ##
+  # Globs for files matching +pattern+ inside of +directory+,
+  # returning absolute paths to the matching files.
+
+  if RUBY_VERSION >= "2.5"
```
hsbt: It's better to inject this condition into `self.glob_files_in_dir` method.

segiddins: But then we incur the cost of the condition every time the method is called

hsbt: Can you show the benchmark about method call? I understand your opinion. But I didn't see that its cost is better than code style.

deivid-rodriguez: I personally agree with @hsbt here.

MSP-Greg: I suspect the cost of the condition will always be far less that the cost of reading a dir, SSD or not...

segiddins: Conditional in the method seems to be about 85% as fast as pre-computing it

MSP-Greg: Turning around, are conditionally defined methods really that bad?

MSP-Greg:
Is it a fast system with a fast SSD, or an older system with a hard drive?  How many files searched, how many files match?  Etc...

segiddins:
feel free to change it ¯\_(ツ)_/¯

segiddins:
changes made

segiddins:
@bundlerbot r+

PR #2337, opened 2018-06-27
Update gem open command help text

sonalkr132:
# Description:
-e option can be used to execute any shell command and not just editors.
Related: https://hackerone.com/bugs?subject=rubygems&report_id=298236

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/open_command.rb
+++ b/lib/rubygems/commands/open_command.rb
@@ -11,9 +11,9 @@ class Gem::Commands::OpenCommand < Gem::Command
   def initialize
     super 'open', 'Open gem sources in editor'
 
-    add_option('-e', '--editor EDITOR', String,
-               "Opens gem sources in EDITOR") do |editor, options|
-      options[:editor] = editor || get_env_editor
+    add_option('-e', '--exec COMMAND', String,
+               "Prepends COMMAND to gem path. Could be used to sepecify editor.") do |command, options|
+      options[:editor] = command || get_env_editor
     end
     add_option('-v', '--version VERSION', String,
                "Opens specific gem version") do |version|
@@ -32,14 +32,14 @@ class Gem::Commands::OpenCommand < Gem::Command
   def description # :nodoc:
     <<-EOF
         The open command opens gem in editor and changes current path
-        to gem's source directory. Editor can be specified with -e option,
-        otherwise rubygems will look for editor in $EDITOR, $VISUAL and
-        $GEM_EDITOR variables.
+        to gem's source directory.
+        Editor command can be specified with -e option, otherwise rubygems
+        will look for editor in $EDITOR, $VISUAL and $GEM_EDITOR variables.
     EOF
   end
 
   def usage # :nodoc:
-    "#{program_name} GEMNAME [-e EDITOR]"
+    "#{program_name} GEMNAME [-e COMMAND]"
   end
 
   def get_env_editor


colby-swandale:
I find this change in the documentation somewhat confusing. What if instead of updating the documentation, we instead change rubygems to only allow specifying an editor?

sonalkr132:
> we instead change rubygems to only allow specifying an editor

I am guessing, you meant creating a whitelist for editor commands. If so, I would like to highlight what @claudijd said.. we would never be able to come up with an extensive list which would covers all editors (and their flags) that exist in the wild.
IMHO, updating the documentation to correctly reflect implication of `-e` is easiest fix here. If you find my version confusing, I am open to suggestions :+1:

duckinator:
I'm confused as to the purpose of this PR. Could you provide some more information?

The main source of confusion for me is that you change the flag from `--editor` to `--exec` and change the documentation on the premise that the command isn't always an editor, but...

1. The _purpose_ of the flag is to open the source code in an editor.
2. It still defaults to a text editor.
3. You provide no examples of using non-editors with `gem open`, so it's unclear what is being solved here.
    * Is this perhaps in the HackerOne link? I can't access it &mdash; it says "this report isn't public yet."
4. You proceed to refer to it as an "editor command" anyway, later in the documentation:

https://github.com/rubygems/rubygems/blob/d8741925dcaa06bae5dda7b3cf0c0728a0a9e26e/lib/rubygems/commands/open_command.rb#L32-L39

colby-swandale:
Seeing this PR is referencing a HackerOne link (which only very few people have access to 😢 ) I'm guessing that someone has raised a ticket saying that this could be exploited. 

But I don't see how this is a problem, the documentation is already clear on what the option is doing, and changing the option to `--exec` (which breaks the public API) does not seem intuative to the user.

sonalkr132:
Yes, I am only trying to close an old HackerOne ticket. Sorry I wasn't verbose enough from start.
Please, try running, `gem open rake -e "ls / "` and it will list your root directory. `gem open -e` is a simple implementation, it works only because we call `System()` with [input to `-e` and gem path](https://github.com/sonalkr132/rubygems/blob/d8741925dcaa06bae5dda7b3cf0c0728a0a9e26e/lib/rubygems/commands/open_command.rb#L74-L78). Fixing the code so that it only works with a editors commands seems a hard problem to me. Although, I am interested in discussing any ideas you guys may have.
In my opinion, it is really not an exploit and just a documentation issue. We should be more clear about what the command does so that our hackerone reporters don't get confused. 
> changing the option to --exec (which breaks the public API) does not seem intuitive to the user.

I don't think this command is as critical or used as often, but I could be wrong. I can change my PR to remove `--editor` change if you really think it is a bad idea.

bronzdoc:
I Agree with @colby-swandale i don't see how this solve the `this could be exploited` problem, closing for now

indirect:
(Probably with the addition of keeping `--editor` as an alias, so that it is not a backwards breaking change)

sonalkr132:
Undid change of `--exec` to `--editor` on [my branch](https://github.com/sonalkr132/rubygems/blob/open-e/lib/rubygems/commands/open_command.rb). I guess closed PRs don't update commits.

Sorry about confusion guys :pray:  I have asked @indirect to give all of you access to HO. Please follow up with him if you still can't see the report.
On HackerOne, we have to walk a fine line between _valid_ and _not applicable_ reports. We may end up offending people if we start closing reports with a heavy hand. #298236, is more of nuanced one, where the report was triaged before I got to it. Personally, I would have closed it as informative with an explanation.

sonalkr132:
See #2458

PR #2339, opened 2018-06-29
Fix activation when multiple platforms installed

MSP-Greg:
# Description:

When multiple platforms of the same gem-version are installed, the correct platform may not be activated.

# Tasks:

- [X] Write tests


hsbt:
@MSP-Greg Thanks. I understood your report and the problem of #2312 .

hsbt:
@bundlerbot r+

PR #2345, opened 2018-07-05
Fix approximate recommendation with prereleases

### LINKED ISSUE #1172: Version#approximate_recommendation suggests invalid specifier for prereleases
A `Gem::Version` object returns a specifier string from the `#approximate_recommendation` method, but when the version in question is a prerelease, the approximate_recommendation requirement cannot be satisfied by the gem itself:

```
> v = Gem::Version.new("1.9.0.dev")
=> Gem::Version.new("1.9.0.dev")
> v.approximate_recommendation
=> "~> 1.9"
> Gem::Requirement.new(v.approximate_recommendation).satisfied_by?(v)
=> false
```

deivid-rodriguez:
# Description:

Fixes #1172. Not sure if this is fully correct but it at least seems to fix original @indirect's report.

Basically, when we want to find the approximate recomendation for a prerelease, we need to make sure the requirement allows prereleases so that it also satisfies the original prerelease. 
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -325,7 +325,9 @@ class Gem::Version
     segments.pop    while segments.size > 2
     segments.push 0 while segments.size < 2
 
-    "~> #{segments.join(".")}"
+    recommendation = "~> #{segments.join(".")}"
+    recommendation += ".a" if prerelease?
+    recommendation
   end
 
   ##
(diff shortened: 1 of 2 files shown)


hsbt:
@deivid-rodriguez Thanks! It's reasonable.

hsbt:
@bundlerbot r+

PR #2351, opened 2018-07-09
Remove semver gem build warning

deivid-rodriguez:
# Description:

This is an alternative to #2323.

I think this warning is too brittle at the moment because `rubygems` has no way of knowing if a dependency is semantically versioned. If a dependency is not semantically versioned, I don't see the problem with using something like `~> 1.2.3` so suggesting a constraint that does the same thing in a more complicated way just because it does not warn seems no good.

In the future maybe rubygems could provide a `spec.versioning_policy =` accessor where gems can inform of their own versioning (`:semver`, or others) but currently I think this warning should just go.
______________

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2352, opened 2018-07-09
Expand symlinks in gem path

eregon:
# Description:

The environment variable `GEM_PATH` can contain symlinks.
This is notably the case on RVM, with a typical `GEM_PATH` being
`GEM_PATH=/home/eregon/.rvm/gems/ruby-2.5.1:/home/eregon/.rvm/gems/ruby-2.5.1@global` (the `...@global` is a symlink to `/home/eregon/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0`).

Currently, RubyGems just takes the paths from `GEM_HOME` and `GEM_PATH` and store them in `Gem.dir` and `Gem.path`, without resolving symlinks.
Similarly those paths might end up in `$LOAD_PATH` and still contain symlinks.

This makes all code relying on `__FILE__`, `caller`, `$LOADED_FEATURES` or `$LOAD_PATH`, etc more brittle for paths comparisons, by having to expand paths with e.g. `File.realpath` before any path comparison involving those paths.
As an example, Bundler has a recent bug due to this, where it fails to detect its own `bundler-1.16.2/lib` directory in `$LOAD_PATH` due to the presence of symlinks in GEM_PATH: https://github.com/bundler/bundler/pull/6502 https://github.com/bundler/bundler/issues/6465

While this specific case of course should be fixed in Bundler, it's not hard to imagine many other places might have similar assumptions.
So I think it's just more robust for all gems to have RubyGems expands symlinks from the environment `GEM_HOME` and `GEM_PATH`.

Ruby implementations since Ruby 2.4.4 [resolve symlinks from `$LOAD_PATH`](https://github.com/oracle/truffleruby/commit/9a76df4287ce212fd7513baa23eefe6dfa4034a2#diff-7346c44420f7a2c1d276e7f67dcee8f8R369).
Therefore, having symlinks in `$LOAD_PATH` means an extra cost during `require` to expand each `$LOAD_PATH` with symlinks.
So it's more efficient if there are already realpaths in `$LOAD_PATH`, and less surprising for everybody (e.g., paths in `$LOADED_FEATURES` are actual concatenation of one of the `$LOAD_PATH` + `/` + required feature).
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -23,10 +23,10 @@ class Gem::PathSupport
   # hashtable, or defaults to ENV, the system environment.
   #
   def initialize(env)
-    @home     = env["GEM_HOME"] || Gem.default_dir
+    @home = env["GEM_HOME"] || Gem.default_dir
 
     if File::ALT_SEPARATOR then
-      @home   = @home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
+      @home = @home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
     end
 
     @path = split_gem_path env["GEM_PATH"], @home
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -29,6 +29,8 @@ class Gem::PathSupport
       @home = @home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
     end
 
+    @home = expand(@home)
+
     @path = split_gem_path env["GEM_PATH"], @home
 
     @spec_cache_dir = env["GEM_SPEC_CACHE"] || Gem.default_spec_cache_dir
@@ -65,7 +67,7 @@ class Gem::PathSupport
       gem_path = default_path
     end
 
-    gem_path.uniq
+    gem_path.map { |path| expand(path) }.uniq
   end
 
   # Return the default Gem path
@@ -76,5 +78,13 @@ class Gem::PathSupport
       gem_path << APPLE_GEM_HOME
     end
     gem_path
+  end
+
+  def expand(path)
+    if File.directory?(path)
+      File.realpath(path)
+    else
+      path
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I don't know the related issue that well, I was just trying to help moving forward the next bundler release :). But yeah, this PR looks like a good idea to me! :+1:

eregon:
@colby-swandale Thank you for the review!
Is there anything left to be done to merge this PR? I'm not too familiar with PRs on RubyGems.

lib/rubygems/path_support.rb
```diff
@@ -77,4 +79,12 @@ def default_path
     end
     gem_path
   end
+
+  def expand(path)
+    if File.directory?(path)
+      File.realpath(path)
```
segiddins: not all systems have `File.realpath` implemented

MSP-Greg: Any in particular?  We can probably consider @eregon well versed in most (if not all), as [spec tests on](https://github.com/ruby/spec):
```
MRI on 30 platforms and 4 versions
JRuby for both 1.7 and 9.x
TruffleRuby
Opal
```

I suspect almost all of the RubyCI.org test against it (spec).  Also, the spec file for [realpath](https://github.com/ruby/spec/blob/master/core/file/realpath_spec.rb) seems to only differentiate between windows and all else, and that may be out of date, as newer windows builds work correctly.  I know I tested it recently, but I can't remember where...

segiddins: Doesn't windows not support it?

MSP-Greg: Older versions of windows did not support symlinks, so older versions of Ruby also did not (on Windows, although Ruby lagged behind).   Anyway, after a brief AFK, I remembered that I came across it in the PR to add a few tests that were skipped (on Windows) back in.  For the versions of Ruby currently tested against, realpath & symlinks work...

eregon: @MSP-Greg Thank you for the praise :) I don't claim to know the availability of `File.realpath` offhand though.
I think it's only Ruby 1.8.7 which doesn't have `File.realpath`, does RubyGems still support such old Ruby versions?
Note that TravisCI & AppVeyor passed.

colby-swandale:
@bundlerbot r+

eregon:
@colby-swandale Thank you for merging!

PR #2356, opened 2018-07-10
Fix broken rubocop task by upgrading to 0.58.1

deivid-rodriguez:
# Description:

The problem is a broken parser release (2.5.1.1). Rubocop 0.58.1 skips
this bad release so upgrading fixes the problem. An alternative is to
explicitly install the previous version of parser, but I see no harm in
upgrading.

See https://github.com/rubocop-hq/rubocop/issues/6092.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -72,12 +72,12 @@ rescue LoadError
 end
 
 begin
-  gem "rubocop", "~> 0.54.0"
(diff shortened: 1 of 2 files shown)


colby-swandale:
Thanks! @bundlerbot r+

deivid-rodriguez:
You're welcome!

Issue #1997, opened 2017-08-29
pristine --bindir


yakimant:
Add support of `--bindir` argument for prestine tool.

For now, I can't do it with macOS environment for example:
```
$ sudo gem pristine --all --only-executables
Restoring gems to pristine condition...
...
Restored aws-sdk-2.7.15
ERROR:  While executing gem ... (Errno::EPERM)
    Operation not permitted - /usr/bin/aws.rb
```

kkavita92:
I can work on this and send a PR

colby-swandale:
@kkavita92 awesome! let us know if you have any questions.

PR #2361, opened 2018-07-18
Add bindir flag to pristine

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/1997

### DIFF
--- a/lib/rubygems/commands/pristine_command.rb
+++ b/lib/rubygems/commands/pristine_command.rb
@@ -46,6 +46,12 @@ class Gem::Commands::PristineCommand < Gem::Command
       options[:env_shebang] = value
     end
 
+    add_option('-n', '--bindir DIR',
+               'Directory where binary files are',
+               'located') do |value, options|
+      options[:bin_dir] = File.expand_path(value)
+    end
+
     add_version_option('restore to', 'pristine condition')
   end
 
@@ -160,12 +166,15 @@ extensions will be restored.
           install_defaults.to_s['--env-shebang']
         end
 
+      bin_dir = options[:bin_dir] if options[:bin_dir]
+
       installer_options = {
         :wrappers => true,
         :force => true,
         :install_dir => spec.base_dir,
         :env_shebang => env_shebang,
         :build_args => spec.build_args,
+        :bin_dir => bin_dir
       }
 
       if options[:only_executables] then
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -187,6 +187,8 @@ class Gem::Installer
     @package.prog_mode = options[:prog_mode]
     @package.data_mode = options[:data_mode]
 
+    @bin_dir = options[:bin_dir] if options[:bin_dir]
+
     if options[:user_install] and not options[:unpack] then
       @gem_home = Gem.user_dir
       @bin_dir = Gem.bindir gem_home unless options[:bin_dir]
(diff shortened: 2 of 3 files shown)


lib/rubygems/commands/pristine_command.rb
```diff
@@ -46,6 +46,12 @@ def initialize
       options[:env_shebang] = value
     end
 
+    add_option('-n', '--bindir DIR',
+               'Directory where binary files are',
```
segiddins: should this be `binstubs` instead of `binary files` ?

MSP-Greg: https://github.com/rubygems/rubygems/blob/672ddcbe14f49c48f6c5492b38480a9162680b5b/lib/rubygems/install_update_options.rb#L27-L31

segiddins: Even if it’s copy-pasted from elsewhere... the “bin” files RubyGems installs aren’t actually binary

MSP-Greg: I'd agree.  The terms 'binary' and 'binstubs' are used both as above and in docs.  What's a better name?  Maybe 'launch scripts', 'CLI scripts', 'scripts', etc.  I don't know...

tjouan:
> I'd agree.  The terms 'binary' and 'binstubs' are used both as above and in docs.  What's a better name?  Maybe 'launch scripts', 'CLI scripts', 'scripts', etc.  I don't know...

  I would say "wrappers", or "executable wrappers". This is how I'm
used to see them called in RubyGems, and how some option flags are
currently named. "scripts" alone is very good too in my opinion, but I
think that the same term should be used everywhere ideally.

  "binary scripts", "bin scripts" is a bit wrong technically as said,
but I think it's still understandable, given its frequent and
conventional usage in directory names (`/bin`, `/usr/bin`… in many
operating systems, `./bin` directory in a ruby gem, `$GEM_HOME/bin`
etc…).

  "binstubs" IMHO is confusing, the first time I saw it was with
909b5fb8 when `--[no-]regenerate-binstubs` was introduced for setup
command, and it took me a little while to understand if it was really
the same thing as the executable wrappers. Except here, I don't think
I ever saw this term.

bronzdoc:
The reason the message is what it is, is because I wanted to be consistent with the install/update ‘—bindir’ Flag description, I’ll change it to whatever we agree on.

MSP-Greg:
@bronzdoc`

> I’ll change it to whatever we agree on.

Doubt I'm part of the 'we' (not complaining), but since you've matched install/update, I think it's as it should be.  The issue of what to call these files should also account for (or fix) what's currently shown in the UI prompts, which I believe is 'executables'...

bronzdoc:
My opinion is that we should merge this and and open a new issue to discuss how this things should be named, then i could follow up with a PR that changes where appropriate. what you think? @segiddins @MSP-Greg @tjouan

MSP-Greg:
> My opinion is that we should merge this and open a new issue...

Agreed...

bronzdoc:
@bundlerbot r+

tjouan:
> My opinion is that we should merge this and and open a new issue to discuss how this things should be named, then i could follow up with a PR that changes where appropriate. what you think? @segiddins @MSP-Greg @tjouan 

  Yes, I think it should be handled separately too. I commented
because I also got confused by the different names in the past, but
this is irrelevant to changes in this branch.

Issue #2359, opened 2018-07-12
Gem::Version should handle `nil` like it used to before 2.7.7


aried3r:
My current problem is the change introduced in https://github.com/rubygems/rubygems/pull/2203. I rely on comparing versions that are `nil`, for example some gems didn't have a version in very old versions, which then correctly meant that `Gem::Version.new(nil) < Gem::Version.new('2.0')` is correctly evaluated to `true`.

An argument could be made that `nil.to_i` is `0` and thus also `< 2.0`.

CocoaPods also [faced this issue](https://github.com/CocoaPods/CocoaPods/issues/7765).

Please reconsider this change, for me it was a breaking change in a minor version.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

Here are my current environment details:

```
$ gem env version
2.7.7
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
As you mention you can do `Gem::Version.new(nil.to_i) < Gem::Version.new('2.0')` to get the behavior you want.

aried3r:
Hi! Thank you for your reply. I was looking more for a discussion about why this was changed. :)

I can't seem to find a rationale as to why handling `nil` differently than `0` or `""` makes sense and was changed in a minor version.

```Ruby
» Gem::Version.new(0) < Gem::Version.new('1.0')
=> true
Gem::Version.new('') < Gem::Version.new('1.0')
=> true
Gem::Version.new(nil) < Gem::Version.new('1.0')
ArgumentError: Malformed version number string
```

Could I propose a softer transition? Like having a "strict mode" that doesn't allow `nil` which you can opt into until it becomes the default.

Your suggestion to use `.to_i` doesn't really work for most cases I think, because usually that would come in from some variable. In my concrete case I iterate over client app versions. Older clients in the database don't have a version set, so it can happen that `client_version` is `nil`.
Calling `.to_i` works for _just nil_ because something like `Gem::Version.new('2.2.0.beta1'.to_i) < Gem::Version.new('2.1')` would return `true` because the result of `.to_i` would be just `2`.

Of course I can handle `nil` myself, but I'm wondering if there is any particular reason for RubyGems to introduce this change? It would greatly help me understand, because personally I liked that this method was very liberal in what it accepted.

[...]

PR #2363, opened 2018-07-20
Gem::Version should handle nil like it used to before

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2359
reverts https://github.com/rubygems/rubygems/pull/2203

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -170,7 +170,6 @@ class Gem::Version
   # True if the +version+ string matches RubyGems' requirements.
 
   def self.correct? version
-    return false if version.nil?
     !!(version.to_s =~ ANCHORED_VERSION_PATTERN)
   end
 
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -170,7 +170,9 @@ class Gem::Version
   # True if the +version+ string matches RubyGems' requirements.
 
   def self.correct? version
-    warn "nil versions are discouraged and will be deprecated in Rubygems 4" if version.nil?
+    unless Gem::Deprecate.skip
+      warn "nil versions are discouraged and will be deprecated in Rubygems 4" if version.nil?
+    end
 
     !!(version.to_s =~ ANCHORED_VERSION_PATTERN)
   end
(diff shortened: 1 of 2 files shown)


duckinator:
:+1: if you add deprecation warnings for `nil` and empty strings saying they won't work starting with 4.x, as mentioned in the original issue.

bronzdoc:
Oh shit i forgot that :sweat_smile: i will @duckinator thanks!

tjouan:
>   * Gem::Version should handle  like it used to before

  Shouldn't we have "nil" between those two consecutive spaces?

bronzdoc:
Since what was proposed was `I propose that we print a warning on nil in 3.x, and reject nil in 4.x. ` should we merge this and wait until 3.x to add the warning? @segiddins @duckinator

segiddins:
Isn’t master now targeting 3.x?

lib/rubygems/version.rb
```diff
@@ -170,6 +170,8 @@ def version
   # True if the +version+ string matches RubyGems' requirements.
 
   def self.correct? version
+    warn "nil versions are discouraged and will be deprecated in Rubygems 4" if version.nil?
+
     return false if version.nil?
```
hsbt: Remove this line for reverting previous behavior.

bronzdoc: done

hsbt:
Ah, a test was failed caused by https://github.com/rubygems/rubygems/pull/2363/commits/41e1bc3640d9effd2ae04d9d72b8668036c9c054

bronzdoc:
Fixed @hsbt

lib/rubygems/version.rb
```diff
@@ -170,7 +170,8 @@ def version
   # True if the +version+ string matches RubyGems' requirements.
 
   def self.correct? version
-    return false if version.nil?
+    warn "nil versions are discouraged and will be deprecated in Rubygems 4" if version.nil?
```
segiddins: should this use `Gem::Deprecate` ?

bronzdoc: Should the year and month be? "2018", "12"

segiddins: ah yeah, always forget that's how it works :/

In that case, only warn `unless Gem::Deprecate.skip?` ?

bronzdoc: done

bronzdoc:
@bundlerbot r+

Issue #1787, opened 2016-11-26
Defaulting to allowed_push_host doesn't quite work


randycoulman:
I'm having a problem.

My current problem is: When I attempt to push a private gem to a private gem server, I am still required to specify the `--host` argument, even though #1486 attempts to default the host to the `allowed_push_host` configured in the `gemspec`.

If I run:

```
gem push pkg/my-test-gem-0.1.0.gem
```

I see:

```
Enter your  credentials.
Don't have an account yet? Create one at /sign_up
   Email:   
```

According to #1486, it should default to using my configured `allowed_push_host`, but instead is prompting for my rubygems.org credentials.

I believe that the root cause of this is that at https://github.com/rubygems/rubygems/blob/0749715e4bd9e7f0fb631a352ddc649574da91c1/lib/rubygems/commands/push_command.rb#L47, the `push` command is calling `sign_in @host`, but `@host` is `nil` at this point.  It hasn't yet been defaulted to the `allowed_push_host`.

That defaulting isn't done until https://github.com/rubygems/rubygems/blob/0749715e4bd9e7f0fb631a352ddc649574da91c1/lib/rubygems/commands/push_command.rb#L76-L86 in the `send_gem` method that is called just after the `sign_in` call.

A possible fix would be to not make the `sign_in` call until after the defaulting is done, but I'm not sure if that would cause other issues.

Additional details:

In my `my-test-gem.gemspec` file, I have:

```
spec.metadata['allowed_push_host'] = "http://my-gemstash-server:9292/private"
```

In my `.gemrc` file, I have:

```
:disable_default_gem_server: true
```

And, in my `.gem/credentials` file, I have:

```
http://my-gemstash-server:9292/private: <my generated key>
```

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.8
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

segiddins:
Best fix is probably to do `@host = options[:host] || gem_data.spec.metadata['default_gem_server'] || gem_data.spec.metadata['allowed_push_host'] ` or something like that?

randycoulman:
[...]

PR #2366, opened 2018-07-26
Sign in to the correct host before push

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/1787

### DIFF
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -29,6 +29,8 @@ command.  For further discussion see the help for the yank command.
   def initialize
     super 'push', 'Push a gem up to the gem server', :host => self.host
 
+    @user_defined_host = false
+
     add_proxy_option
     add_key_option
 
@@ -36,20 +38,41 @@ command.  For further discussion see the help for the yank command.
                'Push to another gemcutter-compatible host',
                '  (e.g. https://rubygems.org)') do |value, options|
       options[:host] = value
+      @user_defined_host = true
     end
 
     @host = nil
   end
 
   def execute
-    @host = options[:host]
+    gem_name = get_one_gem_name
+    default_gem_server, push_host = get_hosts_for(gem_name)
+
+    default_host = nil
+    user_defined_host = nil
+
+    if @user_defined_host
+      user_defined_host = options[:host]
+    else
+      default_host = options[:host]
+    end
+
+    @host = if user_defined_host
+              user_defined_host
+            elsif default_gem_server
+              default_gem_server
+            elsif push_host
+              push_host
+            else
+              default_host
+            end
 
     sign_in @host
 
-    send_gem get_one_gem_name
+    send_gem(gem_name)
   end
 
-  def send_gem name
+  def send_gem(name)
     args = [:post, "api/v1/gems"]
 
     latest_rubygems_version = Gem.latest_rubygems_version
@@ -100,5 +123,15 @@ You can upgrade or downgrade to the latest release version with:
     with_response response
   end
 
+  private
+
+  def get_hosts_for(name)
+    gem_metadata = Gem::Package.new(name).spec.metadata
+
+    [
+      gem_metadata["default_gem_server"],
+      gem_metadata["allowed_push_host"]
+    ]
+  end
 end
 
--- a/test/rubygems/test_gem_commands_push_command.rb
+++ b/test/rubygems/test_gem_commands_push_command.rb
@@ -95,6 +95,26 @@ class TestGemCommandsPushCommand < Gem::TestCase
                  @fetcher.last_request["Content-Type"]
   end
 
+  def test_execute_allowed_push_host
+    @spec, @path = util_gem "freebird", "1.0.1" do |spec|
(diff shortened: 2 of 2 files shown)


lib/rubygems/commands/push_command.rb
```diff
@@ -29,27 +29,50 @@ def usage # :nodoc:
...
 
   def execute
-    @host = options[:host]
+    gem_name = get_one_gem_name
+    default_gem_server, push_host = get_hosts_for(gem_name)
+
+    default_host = nil
+    user_defined_host = nil
+
+    unless @user_defined_host
```
hsbt: Can you reverse this condition like:

```
if @user_defined_host
  user_defined_host = ...
else
  default_host = ...
end
```

It's probably easy to code-reading.

lib/rubygems/commands/push_command.rb
```diff
@@ -100,5 +123,22 @@ def send_gem name
     with_response response
   end
 
+  private
+
+  def get_hosts_for(name)
+    gem_data = Gem::Package.new(name)
+
+    default_gem_server = nil
+    if gem_data.spec.metadata.has_key?("default_gem_server")
```
tjouan: We call `Hash#has_key?` here if I'm correct?
Does RubyGems supports Ruby versions which don't implement `Hash#key?`?

lib/rubygems/commands/push_command.rb
```diff
@@ -100,5 +123,22 @@ def send_gem name
...
+      default_gem_server = gem_data.spec.metadata["default_gem_server"]
+    end
+
+    push_host = nil
+    if gem_data.spec.metadata.has_key?("allowed_push_host")
+      push_host = gem_data.spec.metadata["allowed_push_host"]
+    end
+
+    [default_gem_server, push_host]
+  end
```
tjouan: I hope I didn't misunderstood the code in this method, but wouldn't the
following be equivalent?
``` ruby
gem_metadata = Gem::Package.new(name).spec.metadata

[
  gem_metadata["default_gem_server"],
  gem_metadata["allowed_push_host"]
]
```

test/rubygems/test_gem_commands_push_command.rb
```diff
@@ -93,6 +93,28 @@ def test_execute_host
     assert_equal Gem.read_binary(@path), @fetcher.last_request.body
     assert_equal "application/octet-stream",
                  @fetcher.last_request["Content-Type"]
+
+  end
+
+  def test_execute_allowed_push_host
+    @spec, @path = util_gem "freebird", "1.0.1" do |spec|
+      spec.metadata['allowed_push_host'] = "https://privategemserver.example"
```
tjouan: Just FYI, in previous commit `"` was used even for string literals which don't
need interpolation. I would favor `'` in such case, but I don't know what is
the preferred style in RubyGems (current code on `master` uses both). I'm just
reporting the inconsistancy in those new changes.

test/rubygems/test_gem_commands_push_command.rb
```diff
@@ -93,6 +93,28 @@ def test_execute_host
     assert_equal Gem.read_binary(@path), @fetcher.last_request.body
     assert_equal "application/octet-stream",
                  @fetcher.last_request["Content-Type"]
+
```
tjouan: This adds an extra empty line in an unrelated test.

test/rubygems/test_gem_commands_push_command.rb
```diff
@@ -93,6 +93,28 @@ def test_execute_host
...
+    @cmd.options[:args] = [@path]
+
+    @cmd.execute
+
+    assert_equal Net::HTTP::Post, @fetcher.last_request.class
+    assert_equal Gem.read_binary(@path), @fetcher.last_request.body
+    assert_equal "application/octet-stream",
+                 @fetcher.last_request["Content-Type"]
+
```
tjouan: Other tests don't seem to have this extra empty line at the end of test
methods.

test/rubygems/test_gem_commands_push_command.rb
bronzdoc: why would you favor `'` in such case?

lib/rubygems/commands/push_command.rb
bronzdoc: yes, this is much better, thanks!

bronzdoc:
`It seems odd for me that tests and implementation are in separate commits, shouldn't those be squashed together?`

Agreed

lib/rubygems/commands/push_command.rb
bronzdoc: lol, specs brake with this

bronzdoc:
@bundlerbot  r+

test/rubygems/test_gem_commands_push_command.rb
tjouan: I personally find it easier/faster when reading (no need to read between two
`'` to check for code) and only little additional work when writing.
Using `"` everywhere is interesting too because it eliminates the question and
this difference. I use this last coding style too almost every day without any
problem, but I have a slight preference for first one.

However when seeing two "different" coding styles on a single line of code, I
thought it was unintentional, that's why I commented.

PR #1729, opened 2016-10-02
Deprecate Gem::InstallerTestCase#util_gem_bindir and Gem::InstallerTestCase#util_gem_dir

maclover7:
# Description:

Both of these methods are deprecated, so we should warn users if they are using them, since they will be removed in a future version of Rubygems.
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2369, opened 2018-07-27
[GSoC] Multi-factor feature for RubyGems.

ecnelises:
# Description:

Hello. This is my GSoC project, dedicated to add multifactor authentication to both RubyGems command program and Gemcutter ([RubyGems.org](https://rubygems.org)).

Work for the Gemcutter part has almost been finished (see [PR #1753](https://github.com/rubygems/rubygems.org/pull/1753), [PR #1729](https://github.com/rubygems/rubygems.org/pull/1729) and a series of [my progress reports](https://ecnelises.github.io/)).

## Content:

This PR will contain my changes to RubyGems client, adding multifactor auth for `gem push`, `gem signin` and `gem owner` commands. Since no command for editing profile, adding command for changing multifactor auth settings seems unnecessary.

## Workflow:

- User set up multifactor auth well in the site. (into `mfa_login_and_write` level)
- When user does the actions requiring MFA, an OTP prompt is shown. Or user can add `--otp` option into command, like `gem push mygem-0.0.0.gem --otp 123456`.
- If the OTP is incorrect, operation fails with failure text.
______________

# Tasks:

- [x] Add OTP requirement to `push_command`.
- [x] Add OTP requirement to `owner_command`.
- [x] Add OTP prompt to `sign_in`.
- [ ] Support for `yank_command`.
- [x] Write related tests.

### DIFF
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -30,6 +30,7 @@ permission to.
     super 'owner', 'Manage gem owners of a gem on the push server'
     add_proxy_option
     add_key_option
+    add_otp_option
     defaults.merge! :add => [], :remove => []
 
     add_option '-a', '--add EMAIL', 'Add an owner' do |value, options|
@@ -84,9 +85,10 @@ permission to.
   def manage_owners(method, name, owners)
     owners.each do |owner|
       begin
-        response = rubygems_api_request method, "api/v1/gems/#{name}/owners" do |request|
-          request.set_form_data 'email' => owner
-          request.add_field "Authorization", api_key
+        response = send_owner_request(method, name, owner)
+
+        if need_otp? response
+          response = send_owner_request(method, name, owner, true)
         end
 
         action = method == :delete ? "Removing" : "Adding"
@@ -95,6 +97,16 @@ permission to.
       rescue
         # ignore
       end
+    end
+  end
+
+  private
+
+  def send_owner_request(method, name, owner, use_otp = false)
+    rubygems_api_request method, "api/v1/gems/#{name}/owners" do |request|
+      request.set_form_data 'email' => owner
+      request.add_field "Authorization", api_key
+      request.add_field "OTP", options[:otp] if use_otp
     end
   end
 
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -33,6 +33,7 @@ command.  For further discussion see the help for the yank command.
 
     add_proxy_option
     add_key_option
+    add_otp_option
 
     add_option('--host HOST',
                'Push to another gemcutter-compatible host',
@@ -113,17 +114,26 @@ You can upgrade or downgrade to the latest release version with:
 
     say "Pushing gem to #{@host || Gem.host}..."
 
-    response = rubygems_api_request(*args) do |request|
-      request.body = Gem.read_binary name
-      request.add_field "Content-Length", request.body.size
-      request.add_field "Content-Type",   "application/octet-stream"
-      request.add_field "Authorization",  api_key
+    response = send_push_request(name, args)
+
+    if need_otp? response
+      response = send_push_request(name, args, true)
     end
 
     with_response response
   end
 
   private
+
+  def send_push_request(name, args, use_otp = false)
+    rubygems_api_request(*args) do |request|
+      request.body = Gem.read_binary name
+      request.add_field "Content-Length", request.body.size
+      request.add_field "Content-Type",   "application/octet-stream"
+      request.add_field "Authorization",  api_key
+      request.add_field "OTP", options[:otp] if use_otp
+    end
+  end
 
   def get_hosts_for(name)
     gem_metadata = Gem::Package.new(name).spec.metadata
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -177,7 +177,7 @@ module Gem::GemcutterUtilities
   # Returns true when the user has enabled multifactor authentication from
   # +response+ text.
 
-  def need_otp? response
+  def need_otp?(response)
     return unless response.kind_of?(Net::HTTPUnauthorized) &&
         response.body.start_with?('You have enabled multifactor authentication')
     return true if options[:otp]
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -25,6 +25,16 @@ module Gem::GemcutterUtilities
   end
 
   ##
+  # Add the --otp option
+
+  def add_otp_option
+    add_option('--otp CODE',
+               'Digit code for multifactor authentication') do |value, options|
+      options[:otp] = value
+    end
+  end
+
+  ##
   # The API key from the command options or from the user's configuration.
 
   def api_key
@@ -113,6 +123,13 @@ module Gem::GemcutterUtilities
       request.basic_auth email, password
     end
 
+    if need_otp? response
+      response = rubygems_api_request(:get, "api/v1/api_key", sign_in_host) do |request|
+        request.basic_auth email, password
+        request.add_field "OTP", options[:otp]
+      end
+    end
+
     with_response response do |resp|
       say "Signed in."
       set_api_key host, resp.body
@@ -154,6 +171,20 @@ module Gem::GemcutterUtilities
       say message
       terminate_interaction 1 # TODO: question this
     end
+  end
+
+  ##
+  # Returns true when the user has enabled multifactor authentication from
+  # +response+ text.
+
+  def need_otp? response
+    return unless response.kind_of?(Net::HTTPUnauthorized) &&
+        response.body.start_with?('You have enabled multifactor authentication')
+    return true if options[:otp]
+
+    say 'You have enabled multi-factor authentication. Please enter OTP code.'
+    options[:otp] = ask 'Code: '
+    true
   end
 
   def set_api_key(host, key)
--- a/lib/rubygems/commands/signin_command.rb
+++ b/lib/rubygems/commands/signin_command.rb
@@ -14,7 +14,6 @@ class Gem::Commands::SigninCommand < Gem::Command
     end
 
     add_otp_option
-
   end
 
   def description # :nodoc:
--- a/lib/rubygems/commands/signin_command.rb
+++ b/lib/rubygems/commands/signin_command.rb
@@ -12,6 +12,9 @@ class Gem::Commands::SigninCommand < Gem::Command
     add_option('--host HOST', 'Push to another gemcutter-compatible host') do |value, options|
       options[:host] = value
     end
+
+    add_otp_option
+
   end
 
   def description # :nodoc:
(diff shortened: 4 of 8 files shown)


lib/rubygems/gemcutter_utilities.rb
```diff
@@ -119,6 +119,40 @@ def sign_in sign_in_host = nil
...
+  # OTP code is needed.
+
+  def need_otp?
+    return false if options[:suppress_mfa]
+    unless instance_variable_defined? :@mfa_level
+      response = rubygems_api_request(:get, 'api/v1/multifactor_auth') do |request|
+        request.add_field 'Authorization', api_key
+      end
+
+      # For compatibility to Gemcutters without mfa support
```
hsbt: `Gemcutters` is obsoleted name. Please use `rubygems.org`

colby-swandale:
I don't like the idea of the use of `otp` and `mfa`, the code would be clearer if you stick with just the one name.

colby-swandale:
Using both `otp` and `mfa` does not seem clean, the code would be clearer if you stick with just the one name.

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -119,6 +119,40 @@ def sign_in sign_in_host = nil
...
+                   end
+    end
+
+    @mfa_level == 'mfa_login_and_write'
+  end
+
+  ##
+  # Require user for extra OTP code if multifactor authentication is enabled.
+
+  def run_mfa_check
```
colby-swandale: `run` doesn't really mean anything. Maybe just `mfa_check`?

colby-swandale:
Can you please explain how the user flows works with 2FA. Looking at the code, does the user need to sign in with 2FA first or an error will be given? 

Can we just ask the user to give their MFA token whenever they sign in to an account which has 2FA?

ecnelises:
@colby-swandale 
Thanks. _mfa_ is better known than _otp_, mixing them would confuse people not familiar with how mfa/2fa works.

Here we have a concept about _level of mfa_, inspired by npm. A user can set his level to `login_only` or `login_and_write`. `login_only` means when a user does login, a page requiring extra digit code will be shown. `login_and_write` also requires the digit code for some important operations besides login. Since I have implemented digit check for website login, add a check for MFA code to client seems more reasonable than not.

ecnelises:
A lot of `gem` commands need to be sign in. But if we require user input code first, operations next may also require digit code. That will make operation fail. Or retieving API key does not need multifactor auth as long as we require digit code for all important operations.

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -119,6 +119,40 @@ def sign_in sign_in_host = nil
...
+      # For compatibility to Gemcutters without mfa support
+      @mfa_level = case response
+                   when Net::HTTPNotFound
+                     'no_mfa'
+                   else
+                     with_response(response) { |resp| resp.body }
+                   end
+    end
+
+    @mfa_level == 'mfa_login_and_write'
```
tjouan: If there isn't a many different values for `@mfa_level`, shouldn't symbols be
favored over strings here?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -130,7 +130,7 @@ def need_otp?
         request.add_field 'Authorization', api_key
       end
 
-      # For compatibility to Gemcutters without mfa support
+      # For compatibility to servers without mfa support
```
tjouan: I think this change belongs to another commit (the one adding this comment).

lib/rubygems/commands/push_command.rb
```diff
@@ -38,13 +38,18 @@ def initialize
       options[:host] = value
     end
 
+    add_option('--mfa CODE', 'Digit code for multifactor authentication') do |value, options|
```
segiddins: can this repeated code be factored into a module?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -119,6 +119,40 @@ def sign_in sign_in_host = nil
...
   end
 
+  ##
+  # Fetch user's multifactor authentication settings and return if extra digit
+  # code is needed.
+
+  def need_mfa?
+    return false if options[:suppress_mfa]
+    unless instance_variable_defined? :@mfa_level
+      response = rubygems_api_request(:get, 'api/v1/multifactor_auth') do |request|
```
sonalkr132: Won't this be adding additional api call to all requests? For default case, we should assume that mfa on api is disabled. mfa is an opt-in feature.
From server, we can return 403 or something more appropriate for cases where otp was required but not provided. Only then, client will prompt for otp and repeat the request.

ecnelises: Here `options[:supress_mfa]` is just an internal option for keep compatibility with older test cases, since many of them relies on `last_request`. And fail message the client received is `plain/text`, should I just match it literally?

lib/rubygems/commands/push_command.rb
ecnelises: Yes. Thanks.

lib/rubygems/gemcutter_utilities.rb
sonalkr132: > for keep compatibility with older test cases, since many of them relies on last_request

Can you please explain more about this limitation? Are you sure we can't do away with `options[:suppress_mfa]`?

>  should I just match it literally?

Yeah, checking for response text would be good enough unless we come up with something more elegant.

ecnelises: If we want to remove `options[:suppress_mfa]`, I have tried that create a new module dynamically, setting `need_mfa?` always return false, and prepend it to `Gem::Commands::PushCommand` in test cases which will complain if mfa related requests are added. This method should be not so intrusive, like [here](https://paste.ofcode.org/3aBhsqT54rrcmeZUfjsBdTh).

lib/rubygems/gemcutter_utilities.rb
tjouan: No idea, I didn't even thought about performance when I commented to be honest.
It was more about "semantics", that it seemed used as an "identifier" more than
a string.
But on the other hand, as you said this data is received as string already, if
the spec pass in theory we shouldn't need more code :-)
Anyway it was a small detail I just asked about in case it would help, feel
free to ignore.

lib/rubygems/commands/owner_command.rb
```diff
@@ -87,6 +88,16 @@ def manage_owners method, name, owners
...
+          request.add_field "OTP", options[:mfa] if options[:mfa]
+        end
+
+        if need_mfa? response
+          check_mfa
+          response = rubygems_api_request method, "api/v1/gems/#{name}/owners" do |request|
+            request.set_form_data 'email' => owner
+            request.add_field "Authorization", api_key
+            request.add_field "OTP", options[:mfa]
+          end
```
sonalkr132: Please don't duplicate code. Make a method out of common code where otp is optional argument.
Also, I agree with @colby-swandale 
> the code would be clearer if you stick with just the one name.

`"OTP", options[:mfa]` :no_entry_sign: 
Please use `otp` for everything.

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -113,12 +123,31 @@ def sign_in sign_in_host = nil
       request.basic_auth email, password
     end
 
+    if need_mfa? response
+      check_mfa
```
sonalkr132: `check_mfa` and `need_mfa?` always seem to go together. Can we please call `check_mfa` from `need_mfa?` instead?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -156,6 +174,20 @@ def with_response response, error_prefix = nil
...
   end
 
+  ##
+  # Returns whether the user has enabled multifactor authentication from
+  # +response+ text.
+
+  def need_otp? response
+    return unless response.kind_of?(Net::HTTPUnauthorized) &&
+        response.body.start_with?('You have enabled multifactor authentication')
+    unless options[:otp]
```
sonalkr132: `return true if options[:otp]`

ecnelises:
😄 Is there any problems for current implementation status? I wrote in detail about what it will look like and how it's implemented in [GSoC final report](https://gist.github.com/ecnelises/9654e59877aa336977c1409ef540e2a9). Sorry for out of work these days, maybe I should add 2FA to `yank` now.

hsbt:
@indirect @sonalkr132 Can you handle this issue with rubygems.org side. I'm okay to merge this pull-request targeted RubyGems 3.0.

hsbt:
@eileencodes Do you still waiting to review this?

ChungTranThanh:
https://github.com/eileencodes

sonalkr132:
@hsbt sorry about delay with this. We did have PR on rubygems.org side blocking it which is merged now. 
Functionality wise, this PR is complete. I only need to test these changes with rubygems.org prod api, to ensure we don't break any existing functionality (which I hope to get done by EOW). You are also welcome to test it out. you can enabled mfa related featured on https://rubygems.org by running `document.cookie = 'mfa_feature=true;path=/'` in your browser console.

hsbt:
@sonalkr132 Thank you for your explanation!

RG 3.0 is a good time to merge this feature. The ruby core team will release Ruby 2.6.0-rc1 with RG 3.0 that was the feature freeze version at early December 2018.

sonalkr132:
> need to test these changes with rubygems.org prod api, to ensure we don't break any existing functionality

verified that gem `owner`, `signin` and `push` work as intended with `ui only` and `ui and api` mfa levels. Feel free to way in on aesthetic of PR, specially part we will show to users. 
@ecnelises has writes guides for [setting up mfa](https://github.com/rubygems/guides/blob/ccbd9c778490d8f596fbeda9e433daaa33b1f5df/setting-up-multifactor-authentication.md) and [cli usage](https://github.com/rubygems/guides/blob/ccbd9c778490d8f596fbeda9e433daaa33b1f5df/using-mfa-in-command-line.md), where you can find further details (PR on https://github.com/rubygems/guides/pull/218).

hsbt:
@bundlerbot r+

hsbt:
@ecnelises @sonalkr132 Thanks for your work. I'm going to release this feature with RG 3.0 at 25, Dec 2018.

sonalkr132:
> - [ ] Support for yank_command.

Not really sure why @ecnelises decided to keep it separate but PR for it is now merged on rubygems.org, I guess now he can add it here.

ecnelises:
@sonalkr132 I thought that may delay the review process and I’ve waited for `yank` related API on rubygems.org for a while. I can add it in future PRs. :-) Thanks.

邱超凡

________________________________
寄件者: Aditya Prakash <notifications@github.com>
寄件日期: 星期六, 12月 1, 2018 11:08 下午
收件者: rubygems/rubygems
副本: Qiu Chaofan; Mention
主旨: Re: [rubygems/rubygems] [GSoC] Multi-factor feature for RubyGems. (#2369)


  *    Support for yank_command.

Not really sure why @ecnelises<https://github.com/ecnelises> decided to keep it separate but PR for it is now merged on rubygems.org, I guess now he can add it here.

sonalkr132:
yeah, its cool. yank is not as often used (although very important for mfa) command.
If @hsbt really wants to see make it in one go, I hope he will be kind enough to merge it again :pray:

PR #2382, opened 2018-08-20
Allow build options to be passed to Rake

alyssais:
Rake tasks can take arguments, but unlike other extension builders, arguments were not passed through from `gem install` to Rake.

I haven't managed to write any tests because the current Rake builder tests ran neither on my machine, nor in a clean container.

This does solve the problem I was having, though.
______________

- [x] Write tests

### DIFF
--- a/lib/rubygems/ext/rake_builder.rb
+++ b/lib/rubygems/ext/rake_builder.rb
@@ -5,6 +5,8 @@
 # See LICENSE.txt for permissions.
 #++
 
+require "shellwords"
+
 class Gem::Ext::RakeBuilder < Gem::Ext::Builder
 
   def self.build(extension, dest_path, results, args=[], lib_dir=nil)
@@ -13,9 +15,6 @@ class Gem::Ext::RakeBuilder < Gem::Ext::Builder
       cmd << " #{args.join " "}" unless args.empty?
       run cmd, results
     end
-
-    # Deal with possible spaces in the path, e.g. C:/Program Files
-    dest_path = '"' + dest_path.to_s + '"' if dest_path.to_s.include?(' ')
 
     rake = ENV['rake']
 
@@ -26,9 +25,8 @@ class Gem::Ext::RakeBuilder < Gem::Ext::Builder
 
     rake ||= Gem.default_exec_format % 'rake'
 
-    cmd = "#{rake} RUBYARCHDIR=#{dest_path} RUBYLIBDIR=#{dest_path}".dup # ENV is frozen
-    cmd << " #{args.join " "}" unless args.empty?
-    run cmd, results
+    rake_args = ["RUBYARCHDIR=#{dest_path}", "RUBYLIBDIR=#{dest_path}", *args]
+    run "#{rake} #{rake_args.shelljoin}", results
 
     results
   end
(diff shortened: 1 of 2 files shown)


colby-swandale:
This will need a test before we can merge this PR

alyssais:
Have managed to get the existing tests working so will try to sort this on Monday.

lib/rubygems/ext/rake_builder.rb
```diff
@@ -26,8 +26,8 @@ def self.build(extension, dest_path, results, args=[], lib_dir=nil)
 
     rake ||= Gem.default_exec_format % 'rake'
 
-    cmd = "#{rake} RUBYARCHDIR=#{dest_path} RUBYLIBDIR=#{dest_path}" # ENV is frozen
-
+    cmd = "#{rake} RUBYARCHDIR=#{dest_path} RUBYLIBDIR=#{dest_path}".dup # ENV is frozen
+    cmd << " #{args.join " "}" unless args.empty?
```
segiddins: I don't think this will work if any of the args contain shell meta characters, such as spaces?

alyssais: I think you are probably right! Wanting to make the minimum possible change, I just copied this from the other builders. That seems to be how they do it.

segiddins: :sad_panda:

alyssais: Happy to do a followup PR that fixes this! But I think out of scope for this one since it would be a codebase-wide change.

segiddins: Yeah that’s fine, just sad to discover that we’re already doing the “wrong thing”

alyssais:
Hey, just wondering if there's anything else I need to do here that I've missed? If it's just waiting for somebody to have a look then no worries, but want to make sure it's not waiting for me to do something I've forgotten. :)

alyssais:
(The Travis failure looks to be a misconfiguration rather than a problem with this patch)

deivid-rodriguez:
@alyssais I don't think there's anything else you need to do! :smiley:

I approved the PR yesterday, but I'll wait for @segiddins to have a final look and merge since he made the initial review!

alyssais:
Oh, nice! I wasn't sure because your approval shows up grey rather than green.

deivid-rodriguez:
That's true! I think that means I don't have permissions to merge the PR, so we need to wait for @segiddins anyways :)

lib/rubygems/ext/rake_builder.rb
bronzdoc: @alyssais ^

deivid-rodriguez:
I'm sorry @alyssais if I contributed to the confusion. I agreed with @segiddins and you that this was out of the scope of this PR, that's why I added my positive review!

bronzdoc:
@bundlerbot r+

bronzdoc:
@bundlerbot r+

indirect:
@bronzdoc sorry for the trouble with bundlerbot! I think it's fixed now.

@bundlerbot r=bronzdoc

PR #2386, opened 2018-08-24
Skip permissions-dependent test when root

alyssais:
Root can cd to non-executable directories, which would cause this test to fail.

______________

- [x] Write tests


colby-swandale:
Thanks! @bundlebot r+

colby-swandale:
@bundlerbot r+

PR #2387, opened 2018-08-24
Fix test that depended on /usr/bin being in PATH

alyssais:
This isn't always going to be true. For example, /usr/bin isn't in the PATH on my NixOS system. Work around this by ensuring the test `bin_dir` is always in the PATH.

______________

- [x] Write tests


colby-swandale:
Thanks! @bundlerbot r+

PR #2394, opened 2018-09-07
Remove Trailing whitespace with rubocop.

hsbt:
# Description:

I always faced removing trailing white-space comimts on ruby core repository with our dev bot.

https://github.com/ruby/ruby/commit/73c8b1ea296ca9631f684609d62e908fd5f5d10c

I hope to detect them before merging into ruby core repository.
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2396, opened 2018-09-12
Fix broken builds using the correct rubocop version

bronzdoc:
Rubocop version to install was changed on https://github.com/rubygems/rubygems/pull/2394/files#diff-dcc988a27020494812a8376f558858a7L60

so the rake task doesn't find the correct version to use, this PR changes that.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -72,12 +72,12 @@ rescue LoadError
 end
 
 begin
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

PR #2390, opened 2018-08-31
Fixed test fail with mswin environment.

hsbt:
# Description:

This issue was found at ruby core repository. 

Co-authored-by: MSP-Greg <MSP-Greg@users.noreply.github.com>
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2399, opened 2018-09-17
Fix "interpreted as grouped expression" warning

colby-swandale:
# Description:

This PR resolves a warning which was introduced in #2390. In older versions of Ruby (1.9.3), Ruby will print a warning

    lib/rubygems/installer.rb:782: warning: (...) interpreted as grouped expression

As a result, this is breaking a spec in Bundler's test suite: https://travis-ci.org/bundler/bundler/jobs/428597922
______________

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2404, opened 2018-09-23
update rubocop and also use correct pessimistic version

colby-swandale:
# Description:

This PR has 2 parts. Firstly is just updating Rubocop to the latest version. The second part is that the constraint that is specified in the `util/ci` and the `rubygems-update.gemspec` files are not correct. 

Currently, any version that is less then '1.0' would be installed. This is bad because Rubocop could release a new version which changes a cop that will result in our CI breaking. We want to restrict the version of Rubocop to only patch releases within the given pessimistic version constraint.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/rubygems-update.gemspec
+++ b/rubygems-update.gemspec
@@ -37,5 +37,5 @@ Gem::Specification.new do |s|
   s.add_development_dependency(%q<rake>.freeze, ["~> 12.0"])
   s.add_development_dependency(%q<minitest>.freeze, ["~> 5.0"])
   s.add_development_dependency(%q<simplecov>.freeze, ["~> 0"])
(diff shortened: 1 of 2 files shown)


bronzdoc:
Shouldn't minor changes be backwards compatible API additions/changes? why it would break the bot?

deivid-rodriguez:
Rubocop is at 0.x so technically a breaking change can be made any time. After a second look though, given the strategy we follow for rubocop (disabled by default and only opt-in to selected specific cops), it's very unlikely that a rubocop upgrade will break our build.

bronzdoc:
> Rubocop is at 0.x so technically a breaking change can be made any time.

That's true! thanks for pointing that out @deivid-rodriguez

segiddins:
@bundlerbot r+

PR #2405, opened 2018-09-24
[Gem::Ext::Builder] Comments to aid future refactoring.

duckinator:
# Description:

I added various `FIXME` comments to highlight areas for future refactoring.

# Tasks:

- [x] Write ~code~ <ins>comments</ins> to solve the problem

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -148,7 +148,19 @@ EOF
   def build_extension extension, dest_path # :nodoc:
     results = []
 
+    # FIXME: Determine if this line is necessary and, if so, why.
+    # Notes:
+    # 1. As far as I can tell, this method is only called by +build_extensions+.
+    # 2. The existence of this line implies +extension+ is, or previously was,
+    #    sometimes +false+ or +nil+.
+    # 3. #1 and #2 combined suggests, but does not confirm, that
+    #    +@specs.extensions+ sometimes contained +false+ or +nil+ values.
+    # 4. Nothing seems to explicitly handle +extension+ being empty,
+    #    which makes me wonder both what it should do and what it does.
+    #
+    # - @duckinator
     extension ||= '' # I wish I knew why this line existed
+
     extension_dir =
       File.expand_path File.join(@gem_dir, File.dirname(extension))
     lib_dir = File.join @spec.full_gem_path, @spec.raw_require_paths.first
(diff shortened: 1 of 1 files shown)


lib/rubygems/ext/builder.rb
```diff
@@ -148,13 +150,30 @@ def build_error build_dir, output, backtrace = nil # :nodoc:
...
+    #    sometimes +false+ or +nil+.
+    # 3. #1 and #2 combined suggests, but does not confirm, that
+    #    +@specs.extensions+ sometimes contained +false+ or +nil+ values.
+    # 4. Nothing seems to explicitly handle +extension+ being empty,
+    #    which makes me wonder what both what it should do and what it does.
+    #
+    # - @duckinator
     extension ||= '' # I wish I knew why this line existed
+
+    # FIXME: Ambiguous-to-humans: Maybe add parens to File.join()? - @duckinator
```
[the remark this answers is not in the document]
duckinator: :+1: done.

duckinator:
@bronzdoc just pushed some changes. thoughts?

lib/rubygems/ext/builder.rb
```diff
@@ -200,6 +212,7 @@ def build_extensions
 
     FileUtils.rm_f @spec.gem_build_complete_path
 
+    # FIXME: action at a distance: @ran_rake modified deep in build_extension(). - @duckinator
```
bronzdoc: is this still needed?

lib/rubygems/ext/builder.rb
```diff
@@ -163,9 +161,8 @@ def build_extension extension, dest_path # :nodoc:
     # - @duckinator
     extension ||= '' # I wish I knew why this line existed
```
djberg96: My guess would be in case the extension is `nil`, otherwise the `File.dirname(extension)` below will fail. How that situation could arise, I'm not sure, but it must've happened at some point.

lib/rubygems/ext/builder.rb
duckinator: I feel it is. This part of the codebase is very awkward to me &mdash; there's no reason to assume that `build_extensions` calls a method which calls a second method which _then_ modifies the variable that affects the loop. But it does.

lib/rubygems/ext/builder.rb
duckinator: Yup. I'm unsure what the problem is, and it's not commented, so more poking around will probably be needed.

bronzdoc:
Looks good @duckinator! can you squash your commits to land it? thanks!

duckinator:
@bronzdoc done!

bronzdoc:
@bundlerbot r+

PR #2406, opened 2018-09-24
Move authors assigner to required attributes section of Gem::Specification

greysteil:
Currently, the [gem specification docs](https://guides.rubygems.org/specification-reference/) suggest that author / authors is a "recommended" attribute.

In fact, [author / authors is a required attribute](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification_policy.rb#L287-L291).

This PR is a code move that puts the `authors` assignment method into the required attributes section (and updates its description to explain that specifying `author` is an alternative). This should make the docs much clearer next time they're generated.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -260,22 +260,11 @@ class Gem::Specification < Gem::BasicSpecification
              ].flatten.compact.uniq.sort
   end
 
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

PR #2407, opened 2018-09-24
Enable more RuboCop rules

duckinator:
# Description:

I went through various RuboCop rules I used for other projects (and thus had at least some idea how they worked), and tried enabling ones that seemed appropriate.

I'd like some feedback from other RubyGems maintainers:

1. Did I enable any you think shouldn't be enabled?
2. For rules that let you specify adjustable values (as opposed to enabling it), do y'all have any feedback on what those should be? I mostly based it around the idea of "keeping the codebase from getting even harder to maintain."
3. Are there any RubyGems-specific coding practices I'm unintentionally going against?
4. Are there any changes I made that you think should be resolved in a different way?

This PR, aside from `.rubocop.yml`, should be entirely whitespace/code formatting changes &mdash; if you see anything else, let me know.

As of now, it returns exactly 1 warning, for a 207-character-long line.

# Tasks:

- [ ] Write tests


.rubocop.yml
```diff
@@ -3,11 +3,58 @@ AllCops:
   Exclude:
     - 'bundler/**/*'
     - 'lib/rubygems/resolver/molinillo/**/*'
+    # TODO: Remove all Exclude values after this line.
```
bronzdoc: Why?

.rubocop.yml
```diff
@@ -16,3 +63,9 @@ Layout/IndentationWidth:
 
 Layout/TrailingWhitespace:
   Enabled: true
+
+# Mixing hash styles just looks silly.
```
bronzdoc: yup lol

.rubocop.yml
```diff
@@ -3,11 +3,58 @@ AllCops:
...
+
+# FIXME: This should ideally be a lot closer to 80-ish.
+Metrics/LineLength:
+  Enabled: true
+  Max: 200
+  AllowHeredoc: true
+
+Metrics/MethodLength:
+  Enabled: true
+  Max: 200
```
bronzdoc: isn't this too long?

deivid-rodriguez:
Thanks @duckinator! I find it great to start normalizing the code base. Some people don't like introducing a lot of global git churn since it obscures history. I personally don't think this is an issue as long as you know how to use git properly. So :+1:!

I personally don't bother with `Metrics`. I find they usually get in the middle too much, and make me lose time on artificial refactorings or exclusions. I think this is very similar to the spirit with which `rubocop` is [enabled](https://github.com/bundler/bundler/blob/1bd53e3d930e4f915db5536c68b1ed7282304045/.rubocop.yml#L112) in `bundler`. So, I would not add any `Metrics` cops in here.

Regarding the other cops, I find it easiest to review if you add one cop to the config per commit and run `rubocop --auto-correct` on it. So one can see for each commit the cop being enabled and which offenses are being corrected.

bronzdoc:
LGTM @duckinator can you squash your commits please? one for the chages on files and one for the rubocop config maybe?

.rubocop.yml
duckinator: Oh! When I first worked on this, I was getting multiple thousands of RuboCop warnings from code under `test/`. Apparently other changes I made drove that down to 20.

I'll remove this line + the one below it when I continue work on this PR, and make a follow-up PR resolving those.

.rubocop.yml
duckinator: ... that was supposed to be 100 lmao

duckinator:
@bronzdoc I'll resolve the comments I replied to above + squash the commits tomorrow. :+1:

.rubocop.yml
```diff
@@ -3,11 +3,58 @@ AllCops:
   Exclude:
     - 'bundler/**/*'
     - 'lib/rubygems/resolver/molinillo/**/*'
+    # TODO: Remove all Exclude values after this line.
+    - 'test/**/*'
   TargetRubyVersion: 2.2
 
+Metrics/BlockLength:
```
segiddins: I usually disable all of the `Metrics` cops besides `LineLength`, fwiw, since I find them to just be noisy

hsbt: I'm also disabled code length of `Metrics`. It's not affecting to code quality.

lib/rubygems/defaults.rb
```diff
@@ -28,26 +28,26 @@ def self.default_spec_cache_dir
...
-             ]
-           else
-             [
-               RbConfig::CONFIG['libdir'],
-               ruby_engine,
-               'gems',
-               RbConfig::CONFIG['ruby_version']
-             ]
-           end
+    path =  if defined? RUBY_FRAMEWORK_VERSION then
```
deivid-rodriguez: This looks weird to me. Why the change? Is there a cop correcting this?

duckinator: I didn't do automated corrections. This was indented by an odd number of spaces originally, so both RuboCop and Vim were throwing a fit about it.

I personally prefer this style:

```
path =
    if ...
       ...
    end
```

But I for some reason didn't think to do that at the time.

duckinator: @deivid-rodriguez do you think I should switch to the approach of `path =` followed by the entire block indented two spaces? (See the above comment.) Alternatively, is there another way you'd format it?

duckinator:
I disabled all of the `Metrics`-related cops.

Once we figure out how I should handle @deivid-rodriguez's feedback, I'll handle that and rebase+squash.

---

Also, two questions:

1. Instead of rebasing+squashing, should I open a new PR to replace this one, which follows [the approach @deivid-rodriguez suggested](https://github.com/rubygems/rubygems/pull/2407#issuecomment-424158308)?
2. How should I limit the number of cops enabled in each PR, since I intend to continue doing this for a bit? I'm leaning towards one cop/PR after this PR, unless someone has a different suggestion.

lib/rubygems/defaults.rb
deivid-rodriguez: Hi @duckinator! I don't see the previous odd indentation that grants a reformat of the whol if-block. The only indentation problem there is the elseif block array, where the three elements of it are missing one space on their indentation.

Regarding the style you prefer, it's not my preferred one, but I don't mind if we adopted it. How common is it in the current code base? Is rubocop able to enforce it currently?

deivid-rodriguez:
@duckinator I'm sorry if my approach is too much work at this point... :sweat_smile:

I think it's best to introduce cops at least one per commit (maybe even one per PR as you suggest in the second comment). For reviewability, but also because by running `--auto-correct` instead of making the changes manually, we make sure the cop we're introducing can be autocorrected, which is quite handy, and that the way we correct our code is the way rubocop "likes it". Maybe we won't like what rubocop considers "good" and that prompts us to reconsider the introduction of the specific cop, and the specific style being enabled.

colby-swandale:
@duckinator i would be happy to merge PRs similar to this one.

deivid-rodriguez:
I didn't specifically answer the second question. I think the number of cops enabled in this PR and the size of it is reasonable, so PRs like this one in size and number of cops work for me too :+1:.

lib/rubygems/defaults.rb
duckinator: @deivid-rodriguez the apparent missing spaces were because those array elements were indented an even multiple of two spaces, whereas the `if`/`elsif`/`end` statements (and the other array elements) were not.

e.g., here's that function [as it is on master](https://raw.githubusercontent.com/rubygems/rubygems/master/lib/rubygems/defaults.rb) with the lines annotated with how they're indented:

```
  def self.default_dir # 2 spaces
    path = if defined? RUBY_FRAMEWORK_VERSION then # 4 spaces
             [ # 13 spaces
               File.dirname(RbConfig::CONFIG['sitedir']), # 15 spaces
               'Gems', #15 spaces
               RbConfig::CONFIG['ruby_version'] # 15 spaces
             ] # 13 spaces
           elsif RbConfig::CONFIG['rubylibprefix'] then # 11 spaces
             [ # 13 spaces
              RbConfig::CONFIG['rubylibprefix'], # 14 spaces
              'gems', # 14 spaces
              RbConfig::CONFIG['ruby_version'] # 14 spaces
             ] # 13 spaces
           else # 11 spaces
             [ # 13 spaces
               RbConfig::CONFIG['libdir'], # 15 spaces
               ruby_engine, # 15 spaces
               'gems', # 15 spaces
               RbConfig::CONFIG['ruby_version'] # 15 spaces
             ] # 13 spaces
           end # 11 spaces

    @default_dir ||= File.join(*path) # 4 spaces
  end # 2 spaces
```

Note that `path = ` is 7 characters (including the space after the equal sign), which means aligning things relative to that will inherently leave indentation at an odd number of spaces.

Many editors, when presented with that and set up for 2-space indentation, will simply add a single space(!) when you hit tab.

So there's two approaches folks take here:

1. indent relative to the previous line.
2. indent to even numbers of spaces.

Both are completely valid, but I lean towards the latter because it avoids things like the unintended single-space-indentation in situations like we have here.

My preferred approach of adding a newline after the equal sign means that indentation is relative to the _start_ of the line, not the end, and thus sidesteps the entire problem.

deivid-rodriguez: In this case the problematic lines are lines 10-12 that are indented only one space relative to the opening bracket of the array instead of two as they should.

I don't know about your editor configuration but it's the first time I hear of a coding style that enforces one or two spaces after the equal sign of an assignment depending on whether the resulting indentation is even or odd.

Regarding the other style (introduce a line break after the equal sign) and indent the if/case relative to the beginning of the previous line, that's something I've seen in some places but you're still going to need to fix the middle array indentation since the inconsistency will not be gone unless you fix it.

I would personally enable `Layout/IndentArray` _and_ `Layout/AlignArray` and  and auto-correct them via `rubocop --auto-correct`. I'm posting an [screenrecord](https://asciinema.org/a/isaFUwxjlupSDJWrykDHZtPua) for the motivation: enabling only one of them leaves the code in a weird style. I think this is a bug in rubocop which I'll report, but we can skip it by enabling both cops.

lib/rubygems/defaults.rb
```diff
@@ -28,26 +28,26 @@ def self.default_spec_cache_dir
...
+              [
+                File.dirname(RbConfig::CONFIG['sitedir']),
+                'Gems',
+                RbConfig::CONFIG['ruby_version']
+              ]
+            elsif RbConfig::CONFIG['rubylibprefix'] then
+              [
+                RbConfig::CONFIG['rubylibprefix'],
+                'gems',
+                RbConfig::CONFIG['ruby_version']
```
deivid-rodriguez: To try to clarify further, you can see in slightly darker green here that you're adding 2 spaces to the previous three lines, and one space to the rest of the lines. Instead it would be enough to add a single space to these three lines and leave the rest as they are. There's actually a cop checking for the offense being introduced here, I think it's `Layout/UnnecessaryWhitespace` or something like that.

lib/rubygems/defaults.rb
deivid-rodriguez: Ok, so I checked the [documentation](https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_layout.md#layoutindentarray), and it's super-clear about this. There's no bug in rubocop:

> This cop checks the indentation of the first element in an array literal where the opening bracket and the first element are on separate lines. The other elements' indentations are handled by the AlignArray cop.

duckinator: What I was trying to say is that with some editors, if e.g. you have tabs set to be 2 spaces and are indented 3 spaces, hitting tab will only take you to 4 spaces (not 5).

But if `rubocop --auto-correct` can resolve that then it's not much of a problem, I don't think.

duckinator:
Given my discussion with @deivid-rodriguez I think I'm going to make a new PR using `--auto-correct` for these, which will supersede this one.

deivid-rodriguez:
@duckinator If you're happy with doing that, I'm super happy too! Thanks so much for your work in this area!

duckinator:
This PR is replaced by #2435.

Issue #2397, opened 2018-09-13
Security#re_sign should have smarter defaults


zenspider:
`gem cert` allows you to create a new certificate to sign keys with. One option it provides is the ability to specify how many days until the certificate expires:

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/commands/cert_command.rb#L92

defaulting to 1 year:

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/commands/cert_command.rb#L152-L158

When that certificate expires and you are packaging a new gem, it will automatically `re_sign`your expired certificate. Unfortunately, that expiry is not specified in the call chain so it hits the default of 1 year:

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/security.rb#L497

So, if you specified 3 years when you created your cert, and it is now expired, the next time you package it'll become a 1 year cert. This is definitely not what the user wants. 

I have the opposite problem: I would like to see the default be able to come from `.gemrc` so that I can do the opposite: convert my 1 year certificate to be much longer.

PR #2411, opened 2018-09-27
Improve certificate expiration defaults

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/2397

This introduce a new configuration option `cert_expiration_length_days`, this option will allow a user to set the default expiration length in days of a certificate when is created or being re-signed.

`.gemrc` example:
```yaml
:verbose: true
:update_sources: true
:backtrace: true
:cert_expiration_length_days: 30
```

### DIFF
--- a/lib/rubygems/security/signer.rb
+++ b/lib/rubygems/security/signer.rb
@@ -30,6 +30,15 @@ class Gem::Security::Signer
   attr_reader :digest_name # :nodoc:
 
   ##
+  # Gem::Security::Signer options
+
+  attr_reader :options
+
+  DEFAULT_OPTIONS = {
+    expiration_length_days: 365
+  }
+
+  ##
   # Attemps to re-sign an expired cert with a given private key
   def self.re_sign_cert(expired_cert, expired_cert_path, private_key)
     return unless expired_cert.not_after < Time.now
@@ -52,10 +61,12 @@ class Gem::Security::Signer
   # +chain+ containing X509 certificates, encoding certificates or paths to
   # certificates.
 
-  def initialize key, cert_chain, passphrase = nil
+  def initialize key, cert_chain, passphrase = nil, options = {}
     @cert_chain = cert_chain
     @key        = key
     @passphrase = passphrase
+    @options = options
+    @options.merge!(DEFAULT_OPTIONS)
 
     unless @key then
       default_key  = File.join Gem.default_key_path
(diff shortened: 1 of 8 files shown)


lib/rubygems/security/signer.rb
```diff
@@ -29,6 +29,15 @@ class Gem::Security::Signer
 
   attr_reader :digest_name # :nodoc:
 
+  ##
+  # Gem::Security::Signer options
+
+  attr_reader :options
+
+  DEFAULT_OPTIONS = {
+    expiration_length_days: 365
```
colby-swandale: We're defining the default expiration time in a few places. Maybe this constant should come from `ConfigFile::DEFAULT_CERT_EXPIRATION_LENGTH_DAYS ` ?

lib/rubygems/security/signer.rb
```diff
@@ -40,7 +49,11 @@ def self.re_sign_cert(expired_cert, expired_cert_path, private_key)
 
     Gem::Security.write(expired_cert, new_expired_cert_path)
 
-    re_signed_cert = Gem::Security.re_sign(expired_cert, private_key)
+    re_signed_cert = Gem::Security.re_sign(
+      expired_cert,
+      private_key,
+      (Gem::Security::ONE_DAY * Gem.configuration.cert_expiration_length_days)
```
colby-swandale: We're calculation the date in quite a few places as well, this might be good to extract into a method somewhere.

lib/rubygems/security/signer.rb
```diff
@@ -154,7 +170,7 @@ def sign data
   # be saved as ~/.gem/gem-public_cert.pem.expired.%Y%m%d%H%M%S where the
   # expiry time (not after) is used for the timestamp.
 
-  def re_sign_key # :nodoc:
+  def re_sign_key(expiration_length: Gem::Security::ONE_YEAR) # :nodoc:
```
colby-swandale: This change is using 2 definitions for the default expiry time. Could we use `Gem::Security::ONE_YEAR` instead of having a `365` constant defined?

bronzdoc: ```Gem::Security::ONE_YEAR``` is not the same as `365` and we want to be able to control de days, does that make sense?

lib/rubygems/security/signer.rb
bronzdoc: Both calculations are not exactly the same, so i think it will be better to keep it simple

lib/rubygems/security/signer.rb
bronzdoc: I don't want to make `Gem::Security::Signer` dependant of `Gem::ConfigFile` by requiring it just for a constant. what you think?

PR #2414, opened 2018-09-27
Add version command

colby-swandale:
# Description:

I find myself often using `gem version` to check for the current version of RubyGems, but it's actually `gem --version` to get the version string.

I think it would be good to support both ways. So this PR adds the `version` command that will just print out the version string the same way that `gem --version` does.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -303,6 +303,7 @@ lib/rubygems/commands/stale_command.rb
 lib/rubygems/commands/uninstall_command.rb
 lib/rubygems/commands/unpack_command.rb
 lib/rubygems/commands/update_command.rb
+lib/rubygems/commands/version_command.rb
 lib/rubygems/commands/which_command.rb
 lib/rubygems/commands/yank_command.rb
 lib/rubygems/compatibility.rb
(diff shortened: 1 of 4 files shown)


segiddins:
I’m not sure we need to have 2 names for the same command

MSP-Greg:
#### Against

1. (segiddens) I’m not sure we need to have 2 names for the same command.
2. Most people are not checking this often.

#### For

3. Re 1, is it a command or an option?  If it's a command, allow it to also use normal command syntax.
3. Many CLI systems that have commands (as opposed to just options) use `version`.
4. Don't frustrate the people who are testing multiple versions, etc.  I've made the same mistake.

MSP-Greg:
@colby-swandale

I was just trying to make a point that it's type is somewhat undefined, as another person referred to it as a command.

As tightly coupled as Bundler & RubyGems are, practically speaking, both should support both `version` and `--version`.

bronzdoc:
Maybe if this command behaves like bundler does?

```shell
$ bundler --version 
Bundler version 1.16.2
$ bundler version  
Bundler version 1.16.2 (2018-05-16 commit 9f7bf0ac3)
```

So display extra information when people use `gem version` if not i don't see the point of this command.

MSP-Greg:
@bronzdoc

LGTM.  Commit info is certainly helpful when working with branches in Ruby.  I don't know if RG has the info...

colby-swandale:
I like the idea of having a the commit hash in the version string, but i would put that functionality into its own PR

PR #2424, opened 2018-10-05
Fix link in README.md

marcandre:


colby-swandale:
Thanks for spotting this! @bundlerbot r+

colby-swandale:
Oh the commit message needs to drop the `[ci skip]`. We need CI to pass before we can merge this change.

marcandre:
> Oh the commit message needs to drop the `[ci skip]`. We need CI to pass before we can merge this change.

Right, I imagine there's a "Require pull request reviews before merging" branch rule. I'd open a bug with Github, but I'm already 3 levels deep in the rabbit hole, so I simply opened #2428.

Issue #2373, opened 2018-07-31
Change requires to require_relative?


MSP-Greg:
Since RG is now Ruby 2.2+, does anyone object to changing appropriate `require` statements to `require_relative`?  I'll do a PR if ok...

  - [X] Other

Here are my current environment details:

```
$ gem env version
NA
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

hsbt:
What's advantage that used `require_relative`?

segiddins:
The biggest advantage is usually performance, since you dont have to search over the entire `$LOAD_PATH`, or go through RubyGems' custom `Kernel#require`, which can be rather heavy

MSP-Greg:
@segiddins

Thanks, I was just typing something similar.  Disk I/O is bad.

hsbt:
@segiddins Can you show the benchmark results?

The ruby core team has been confirming the measurement results with `require_relative` and rails. It has no biggest effect for performance.

MSP-Greg:
@hsbt

So you're saying you want a benchmark for something that obviously lowers disk i/o?  I work a lot with Appveyor, and test times vary a great deal.  I don't what exactly is the cause, but network and/or disk i/o are certainly part of it.

I'm not saying it will make a big difference, but it is good programming practice.  Why search a path for a file when you know where it is?

hsbt:
@MSP-Greg I'm not sure what your motivation about this issue. Do you want a performance improvement? or Coding Style? or other?

If It's performance improvement, I never accept performance changes without benchmark results like Disk I/O, Network, CPU time. We shouldn't relay [Cargo cult programming](https://en.wikipedia.org/wiki/Cargo_cult_programming).

MSP-Greg:
@hsbt

Your 'Cargo cult programming' link doesn't apply here.  This involves one statement, not a pattern strategy or something similar. 

> I never accept performance changes without benchmark results

It should be obvious that system configuration will greatly affect benchmark results.

> The ruby core team

I can also point to issues [8781 Use require_relative() instead of require() if possible](https://bugs.ruby-lang.org/issues/8781) and [12973 [RFC] net/http: use require_relative to reduce syscalls](https://bugs.ruby-lang.org/issues/12973).  In particular, Eric's message in 12973.

[...]

PR #2425, opened 2018-10-05
Prefer require_relative

marcandre:
There are over 300 calls to `require` that could be more efficient with `require_relative`. If I'm not mistaken, current `rubygems` supports only Ruby 2.2+, so this is a free speed gain.

### DIFF
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require "rubygems/util"
+require_relative "util"
 
 module Gem::BundlerVersionFinder
   @without_filtering = false
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -6,8 +6,8 @@
 #++
 
 require 'optparse'
-require 'rubygems/requirement'
-require 'rubygems/user_interaction'
+require_relative 'requirement'
+require_relative 'user_interaction'
 
 ##
 # Base class for all Gem commands.  When creating a new gem command, define
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -5,8 +5,8 @@
 # See LICENSE.txt for permissions.
 #++
 
-require 'rubygems/command'
-require 'rubygems/user_interaction'
+require_relative 'command'
+require_relative 'user_interaction'
 
 ##
 # The command manager registers and installs all the individual sub-commands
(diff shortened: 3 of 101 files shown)


deivid-rodriguez:
Hi @marcandre, thanks for this! This was proposed in https://github.com/rubygems/rubygems/issues/2373, but rejected by @hsbt because he wanted a benchmark to prove the speed gain.

I tried to write one so that he will accept this. You can pass an integer argument to the script that sets the LOAD_PATH size. The higher the argument, the greater the speed gain.

```ruby
require "fileutils"
require "benchmark/ips"

def load_path(n)
  $LOAD_PATH.replace([])

  (1..n.to_i).map do |dir|
    $LOAD_PATH << dir.to_s
  end

  Dir.mkdir(n)
  FileUtils.touch("#{n}/a.rb")
end

n = ARGV[0]

load_path(n)

Benchmark.ips do |x|
  x.report("require with LOAD_PATH of size #{n}:") { require "a" }
  x.report("require_relative with LOAD_PATH of size #{n}:") { require_relative "#{n}/a" }
  x.compare!
end

File.delete("#{n}/a.rb")
Dir.rmdir(n)
```

marcandre:
One thing is for sure, `require_relative` can only be faster than `require` 😆!

It's also better styling by being more explicit: makes `require`s of other gems stand out compared to files of rubygems itself.

deivid-rodriguez:
Absolutely, I would enable this just for style! But both of the times this was proposed performance was mentioned so I want to make sure it's accepted anyways :)

MSP-Greg:
@deivid-rodriguez @marcandre 

Thanks for this, and the benchmark.  Working with Windows CI as much as I do, I've seen a wide variance In test times.  Probably partly due to shared i/o on a drive.  This will certainly help...

deivid-rodriguez:
Wow, @amatsuda, thanks. I was completely unaware of these issues. Let's wait then, yeah.

MSP-Greg:
I'm not familiar with rvm, but I'm wondering about use cases where this would be an issue.

RG Is similar to a std-lib, in that normal installs or updates are always within the lib folder.  Gems (Rails) are different.

I haven't been in the bundler code for quite a while, but I rarely see code loading parts of RG.  A Gem class method or maybe some of the command files...

There maybe issues with some CI scenarios.

marcandre:
Oh, I see. So any path that might be `require`d (from another gem/app) should not be also `require_relative` until Ruby 2.5+.

Since it's hard to know for which files this might be the case, probably best to wait then 😿

PR #2426, opened 2018-10-05
Fix bundler rubygems binstub not properly looking for bundler

deivid-rodriguez:
# Description:

The problem is that when running a rubygems bundler binstub with a lock file locked to a non-installed bundler version, one would get the cryptic error message:

```
can't find gem bundler (>= 0.a) with executable bundle (Gem::GemNotFoundException)
```

where one tends to think... What? How is it not found if I just run it?!

Partially closes bundler/bundler#6595. Now the message is way more clear, but I guess it'd be better to actually run the command. Not sure if that's possible from a rubygems binstub, though.

# Tasks:

- [x] Write tests
- [x] ~Write~ Delete code to solve the problem

### DIFF
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -29,7 +29,7 @@ module Gem::BundlerVersionFinder
     return unless vr = bundler_version_with_reason
     <<-EOS
 Could not find 'bundler' (#{vr.first}) required by #{vr.last}.
-To update to the lastest version installed on your system, run `bundle update --bundler`.
+To update to the latest version installed on your system, run `bundle update --bundler`.
 To install the missing version, run `gem install bundler:#{vr.first}`
     EOS
   end
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -3,15 +3,6 @@
 require "rubygems/util"
 
 module Gem::BundlerVersionFinder
-  @without_filtering = false
-
-  def self.without_filtering
-    without_filtering, @without_filtering = true, @without_filtering
-    yield
-  ensure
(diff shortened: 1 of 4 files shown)


segiddins:
How is this fixing the issue?

deivid-rodriguez:
Because when you run the spec with this patch applied you get what you want, but when run without the patch you don't... :rofl:.

Seriously, I'm not sure. I'll comment over the code and throw some guesses.

lib/rubygems/bundler_version_finder.rb
```diff
@@ -3,15 +3,6 @@
 require "rubygems/util"
 
 module Gem::BundlerVersionFinder
-  @without_filtering = false
-
-  def self.without_filtering
-    without_filtering, @without_filtering = true, @without_filtering
```
deivid-rodriguez: So, I have no idea what this "filtering" intends to do, but I think the problem is that here the assignment is inverted (it should be `without_filtering, @without_filtering = @without_filtering, true`. That means that once this "filtering" goes back to true, it's never switched back. Since this filtering is used further below in the `bundler_version_with_reason` to prevent the method from returning an useful message, that's what it ends up happening.

So, I think this could also be fixed by inverting the assignment but I wonder what we're trying to achieve with all this untested code... :man_shrugging:.

deivid-rodriguez:
I managed to write a spec for this! :+1:

deivid-rodriguez:
Also, I think this fix might mean that `bundler 2` should require `rubygems 2.7.8` or higher?

deivid-rodriguez:
I rebased this PR and also correct a typo, and a test description.

hsbt:
@bundlerbot r+

deivid-rodriguez:
Thanks! :smiley:

Issue #2430, opened 2018-10-06
You can make hidden gems. It kinda breaks things and it's maybe a security vulnerability. Hiiiiii, everybody! 👋


JoshCheek:
Okay. A series of *very reasonable* inclinations, which were not absurd in any way whatsoever, led me to eventually wonder whether I could make a gem whose name began with a leading dot. For example, can I make a gem named `.omghi`? Surely no.

Oh!

Uhhhhhh… heh…

Yes.

Yes, I can.

Whoops! 🤷‍♀️ 

After several hours of playing with the implications, I had a wealth of amusing discoveries. For example, I couldn't require its files because the leading dot made it hidden. lol, oh how amused I was in that moment! 😖

![screen shot 2018-10-03 at 11 05 27 pm](https://user-images.githubusercontent.com/77495/46567463-e426e900-c8f8-11e8-83a8-112e7b3f0677.png)

One last curiousity led me to publish it.

And oh! What fun!

It actually published successfully!

And it installed successfully!

And it… did not… uninstall successfully!

lolol

fuck 😖

![screen shot 2018-10-04 at 8 40 46 am](https://user-images.githubusercontent.com/77495/46567458-cfe2ec00-c8f8-11e8-8502-ca1d5e2e8019.png)

I'll ignore the mounting need to apologize and just keep an eye on it. Maybe it's totally fine? Maybe nothing will happen with it 🤞

…soooooooo 😬

yeah no 😖

After being published for < 2 days, the latest version has 1k downloads.

![image](https://user-images.githubusercontent.com/77495/46568335-9ca85900-c908-11e8-9d8f-3794b5a766d7.png)

For comparison, `seeing_is_believing`, a legitimate gem that I've been maintaining for 6 years, has only 381 downloads over the last month. And that's a _legitimate_ gem that I use every day, many people know about it, and if you're the kind of person who is reading this, then you're probably the kind of person who would massively benefit from using it! (eg in place of your REPL, and absolutely in place of whatever shitty hack you're using to embed outputs into your code samples… copy & paste [#amirite](https://www.google.com/search?q=amirite))

![screen shot 2018-10-05 at 11 50 46 pm](https://user-images.githubusercontent.com/77495/46567629-86e06700-c8fb-11e8-9912-82a30fc8b5c6.png)

[...]

PR #2432, opened 2018-10-10
Restrict special chars from prefixing new gem names

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2430

A user should not be able to create a gem with a `.`, `-` or `_` as a prefix in a gem name
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -4,6 +4,8 @@ require 'uri'
 class Gem::SpecificationPolicy < SimpleDelegator
   VALID_NAME_PATTERN = /\A[a-zA-Z0-9\.\-\_]+\z/ # :nodoc:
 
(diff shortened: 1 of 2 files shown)


bronzdoc:
Thanks for the review @colby-swandale!

bronzdoc:
@bundlerbot r+

asottile:
I know it's silly, but I was depending on this behaviour for https://pre-commit.com to pick a gem name that would never conflict.  I've gone through and fixed all of my uses but unfortunately I've caused to be _a pattern_.

This wasn't called out at all in the 3.0 changelog despite being a breaking change, it might be good to call that out.

For others seeing the error message, the message is (in my case):

```
 invalid value for attribute name: "__fake_gem" can not begin with a period, dash, or underscore
```

PR #2435, opened 2018-10-12
Enable more rubocop rules

duckinator:
# Description:

I enabled more RuboCop rules, and ran `rubocop --auto-correct` with each change as @deivid-rodriguez suggested in #2407.

This PR replaces #2407.

# Tasks:

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -32,6 +32,12 @@ Layout/IndentHash:
   Enabled: true
   EnforcedStyle: consistent
 
+# NOTE: When used with --auto-correct, this only seems to auto-correct the
+#       first array item for some reason.
+Layout/IndentArray:
+  Enabled: true
+  EnforcedStyle: consistent
+
 Layout/IndentationConsistency:
   Enabled: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -14,6 +14,9 @@ Layout/BlockAlignment:
 Layout/CaseIndentation:
   Enabled: true
 
+Layout/ClosingParenthesisIndentation:
+  Enabled: true
+
 Layout/CommentIndentation:
   Enabled: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -11,6 +11,9 @@ Layout/AccessModifierIndentation:
 Layout/BlockAlignment:
   Enabled: true
 
+Layout/CaseIndentation:
+  Enabled: true
+
 Layout/CommentIndentation:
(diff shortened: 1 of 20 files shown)


.rubocop.yml
```diff
@@ -5,9 +5,42 @@ AllCops:
...
+Layout/EndOfLine:
+  Enabled: true
+  EnforcedStyle: lf
+
+Layout/IndentHash:
+  Enabled: true
+  EnforcedStyle: consistent
+
+# NOTE: When used with --auto-correct, this only seems to auto-correct the
+#       first array item for some reason.
```
deivid-rodriguez: Yeah, I noticed this when I reviewed the other PR. See https://github.com/rubygems/rubygems/pull/2407#discussion_r221091861. Given this limitation, I don't think it makes sense to use `Layout/IndentArray` but not `Layout/AlignArray`, so could we enable `Layout/AlignArray` as well?

.rubocop.yml
```diff
@@ -5,9 +5,42 @@ AllCops:
...
+Layout/IndentArray:
+  Enabled: true
+  EnforcedStyle: consistent
+  Exclude:
+    # This is a generated file.
+    - lib/rubygems/util/licenses.rb
```
deivid-rodriguez: I think we can fix this at the source by removing spaces from [here](https://github.com/rubygems/rubygems/blob/cc881a806afc90731d5a912415aa945726978b54/util/generate_spdx_license_list.rb#L28) and [here](https://github.com/rubygems/rubygems/blob/cc881a806afc90731d5a912415aa945726978b54/util/generate_spdx_license_list.rb#L33).

duckinator: Given the size of this PR, I'd like to make a second follow-up PR to address this one after this gets merged, if that's okay with other folks.

.rubocop.yml
duckinator: I thought there was something like that, but couldn't find it. Thanks for the info!

.rubocop.yml
deivid-rodriguez: Sounds good to me! :+1:

.rubocop.yml
duckinator: Attempting to enable AlignArray revealed a potential bug, or at the very least some questionable code lol https://github.com/rubygems/rubygems/issues/2437

colby-swandale:
Great! @bundlerbot r+

colby-swandale:
@bundlerbot r+

PR #2436, opened 2018-10-15
Remove rubyforge_page functionality

Schwad:
# Description:

Type: Cleanup

RubyForge was [sunsetted in 2013](https://twitter.com/evanphx/status/399552820380053505) . There are still mentions of it in the code base. For this PR, instead of looking for dead links I aimed to clear out the `rubyforge_project` attr_accessor from the rubygems specification and the places that reference it. 

Perhaps there could be a conversation here whether something could 'replace', `rubyforge_project` (`github_project`?) but to start here is the 'lite' approach that simply removes it. 

Looking forward to thoughts and feedback! :) 
______________

# Tasks:

- [ ] Write tests __(n/a)__

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -156,7 +156,6 @@ class Gem::Specification < Gem::BasicSpecification
     :required_ruby_version     => Gem::Requirement.default,
     :required_rubygems_version => Gem::Requirement.default,
     :requirements              => [],
-    :rubyforge_project         => nil,
     :rubygems_version          => Gem::VERSION,
     :signing_key               => nil,
     :specification_version     => CURRENT_SPECIFICATION_VERSION,
@@ -730,14 +729,6 @@ class Gem::Specification < Gem::BasicSpecification
   # Allows deinstallation of gems with legacy platforms.
 
   attr_writer :original_platform # :nodoc:
-
-  ##
-  # The rubyforge project this gem lives under.  i.e. RubyGems'
-  # rubyforge_project is "rubygems".
-  #
-  # This option is deprecated.
-
-  attr_accessor :rubyforge_project
 
   ##
   # The Gem::Specification version of this gemspec.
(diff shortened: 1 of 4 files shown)


bronzdoc:
There is Bundler code that is bening tested with `rubyforge_project` that needs to me modified too

hsbt:
I'm okay to merge this. But The 15000+ gems still uses `rubyforge_project`.  Like this.

```
...
active_presenter-3.2.2/lib/tasks/gem.rake:  s.rubyforge_project     = 'active_presenter'
active_press-0.1.0/active_press.gemspec:  s.rubyforge_project = "active_press"
active_reaper-0.0.2/active_reaper.gemspec:  s.rubyforge_project = "active_reaper"
...
```

We need to add the migration step that is the only `attr_writer`.

Schwad:
@hsbt thanks for the feedback and @bronzdoc thank you for the review as well. 

What I will look to do is keep the removals but utilize `attr_accessor :rubyforge_project` as suggested to keep existing gems from breaking.

Since we have access to `Gem::Deprecate`, would it be wise to implement deprecate as well? 

`deprecate :rubyforge_project, :none, 2020, 4`

Schwad:
I will see how the checks go, I have reincluded an `attr_writer` one-liner and utilized `deprecate` as well.

test/rubygems/test_gem_specification.rb
```diff
@@ -1210,14 +1206,6 @@ def test_self_stubs_for_mult_platforms
 
   DATA_PATH = File.expand_path "../data", __FILE__
 
-  def test_handles_private_null_type
-    path = File.join DATA_PATH, "null-type.gemspec.rz"
-
-    data = Marshal.load Gem::Util.inflate(Gem.read_binary(path))
-
-    assert_nil data.rubyforge_project
```
simi: Can't we just change assertion in here? What's purpose of this test?

Schwad: I went through the history (blame) on this file and the latest commits on each line weren't too illuminating: https://github.com/rubygems/rubygems/blame/master/test/rubygems/test_gem_specification.rb

However nothing about this test makes me think that `rubyforge_project` is the necessary component. We can flag up some other authors but I could preserve this test with the assertion looking at a different attribute. (say 'description'). Will commit that now.

simi: Yup, that's what I was thinking about as well. It is not clear what's the purpose of this test at all, but it is probably not directly testing `rubyforge_project`, but uses `rubyforge_project` to assert only. Changing assertion to use different attribute should work here also.

Schwad: Interestingly enough, as I just found out, `description` does not assert nil. I am wondering if this is only actually passing on an `assert_nil` because `rubyforge_project` simply isn't set anymore.....

lib/rubygems/specification.rb
```diff
@@ -1349,16 +1348,15 @@ def self._load(str)
     spec.instance_variable_set :@required_rubygems_version, array[7]
     spec.instance_variable_set :@original_platform,         array[8]
     spec.instance_variable_set :@dependencies,              array[9]
-    spec.instance_variable_set :@rubyforge_project,         array[10]
```
segiddins: I think this change would require bumping the spec version -- it might be better to keep the offsets the same, and just never use the rubyforge ivar?

duckinator: :+1: for keeping the offsets the same. You should probably also add a note explaining that it's deprecated/no longer used.

Schwad:
Okay, I'll redo the offsets. It looks like the bundler failure (referenced
in another PR) is due to it tying to an older version of `rake` which still
uses `rubyforge_project` and they won't be able to bump it until `bundler
2.0`.

What I may do to 'get this over the line' is keep offsets the same, add a
note, and save adding the `deprecate` line on the `attr_writer` until
bundler accepts my bump PR in the future.

Schwad:
Should have something up monday -just replying from email

Schwad:
Update - reference this [in bundler](https://github.com/bundler/bundler/pull/6745) and will do another PR in future with the full deprecation line so that we can move forward without breaking tests. Will make that change after bundler 2.0

Current changes restored offsets and removed explicit deprecation.

deivid-rodriguez:
So @Schwad, I take it this is ready to go now, right?

Great job here, I think there's some other attributes where we could do the same thing, for example, `test_files`.

Schwad:
@deivid-rodriguez just had a look, yes we should be good to go for this!

bronzdoc:
@bundlerbot r+

PR #2439, opened 2018-10-20
Support Keyword args with Pysch.

hsbt:
# Description:

I picked commits related Kernel#warn feature with Ruby 2.5+ and keyword args support of Pysch-3.1.0.pre1 from ruby core repository.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/safe_yaml.rb
+++ b/lib/rubygems/safe_yaml.rb
@@ -28,11 +28,19 @@ module Gem
 
     if ::YAML.respond_to? :safe_load
       def self.safe_load input
-        ::YAML.safe_load(input, WHITELISTED_CLASSES, WHITELISTED_SYMBOLS, true)
+        if Gem::Version.new(Psych::VERSION) >= Gem::Version.new('3.1.0.pre1')
+          ::YAML.safe_load(input, whitelist_classes: WHITELISTED_CLASSES, whitelist_symbols: WHITELISTED_SYMBOLS, aliases: true)
+        else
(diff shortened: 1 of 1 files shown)


segiddins:
Why are we adding in more redefinitions of methods in the stdlib?

hsbt:
@nobu Can you describe it?

nobu:
Please see https://github.com/rubygems/rubygems/pull/2442.

nobu:
Please see https://github.com/rubygems/rubygems/pull/2442.

nobu:
Please see https://github.com/rubygems/rubygems/pull/2442.

nobu:
Please see https://github.com/rubygems/rubygems/pull/2442.

nobu:
Please see https://github.com/rubygems/rubygems/pull/2442.

segiddins:
Even after looking at that PR, I can't understand why we need to redefine a method in `Kernel` inside of rubygems

duckinator:
@segiddins it looks like, due to how `require` is redefined by RubyGems, the warning shows the wrong file, so it's hard to find the true source of the warning.

MSP-Greg:
@segiddins 

> Even after looking at that PR, I can't understand why we need to redefine a method in Kernel inside of rubygems

One might say that RubyGems is inserting itself into the 'callstack' by redefining `require`.  Since most Ruby reporting (errors, warning, etc), is 'callstack' based, and some of the reporting (warnings) only show the 'first' item in the 'callstack', an adjustment should be made somewhere...

hsbt:
@bundlerbot r+

PR #2442, opened 2018-10-22
Bug/kernel#warn uplevel

nobu:
# Description:

With this script `w.rb`:
```ruby
require "Win32API"
```

The expected line is shown when rubygems is disabled.
```
$ ruby -w --disable=gems -I./lib -I./ext/win32/lib w.rb 
w.rb:1: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
```

But `kernel_require.rb` is shown when rubygems is enabled.
```
$ ruby -w --enable=gems -I./lib -I./ext/win32/lib w.rb 
/opt/local/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54: warning: Win32API is deprecated after Ruby 1.9.1; use fiddle directly instead
```

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1392,5 +1392,6 @@ Gem::Specification.load_defaults
 
 require 'rubygems/core_ext/kernel_gem'
 require 'rubygems/core_ext/kernel_require'
+require 'rubygems/core_ext/kernel_warn'
 
 Gem.use_gemdeps
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -1,28 +1,45 @@
 # frozen_string_literal: true
 
+# `uplevel` keyword argument of Kernel#warn is available since ruby 2.5.
 if RUBY_VERSION >= "2.5"
+
   module Kernel
-    path = "#{__dir__}/"
+    path = "#{__dir__}/" # Frames to be skipped start with this path.
+
+    # Suppress "method redefined" warning
     original_warn = instance_method(:warn)
     Module.new {define_method(:warn, original_warn)}
+
     original_warn = method(:warn)
 
     module_function define_method(:warn) {|*messages, uplevel: nil|
-      if uplevel
-        uplevel, = [uplevel].pack("l!").unpack("l!")
-        if uplevel >= 0
-          start = 0
-          begin
-            loc, = caller_locations(start, 1)
-            break start += uplevel unless loc
-            start += 1
-          end while (loc.path.start_with?(path) or (uplevel -= 1) >= 0)
-          uplevel = start
-        end
-        original_warn.call(*messages, uplevel: uplevel)
-      else
-        original_warn.call(*messages)
+      unless uplevel
+        return original_warn.call(*messages)
       end
+
+      # Ensure `uplevel` fits a `long`
+      uplevel, = [uplevel].pack("l!").unpack("l!")
+
+      if uplevel >= 0
+        start = 0
+        while uplevel >= 0
+          loc, = caller_locations(start, 1)
+          unless loc
+            # No more backtrace
+            start += uplevel
+            break
+          end
+
+          start += 1
+
+          unless loc.path.start_with?(path)
+            # Non-rubygems frames
+            uplevel -= 1
+          end
+        end
+        uplevel = start
+      end
+      original_warn.call(*messages, uplevel: uplevel)
     }
   end
 end
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+if RUBY_VERSION >= "2.5"
+  module Kernel
+    path = "#{__dir__}/"
+    original_warn = instance_method(:warn)
+    Module.new {define_method(:warn, original_warn)}
+    original_warn = method(:warn)
+
+    module_function define_method(:warn) {|*messages, uplevel: nil|
+      if uplevel
+        uplevel, = [uplevel].pack("l!").unpack("l!")
+        if uplevel >= 0
+          start = 0
+          begin
+            loc, = caller_locations(start, 1)
+            break start += uplevel unless loc
+            start += 1
+          end while (loc.path.start_with?(path) or (uplevel -= 1) >= 0)
+          uplevel = start
+        end
+        original_warn.call(*messages, uplevel: uplevel)
+      else
+        original_warn.call(*messages)
+      end
+    }
+  end
+end
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -412,6 +412,24 @@ class TestGemRequire < Gem::TestCase
     end
   end
 
+  if RUBY_VERSION >= "2.5"
+    def test_no_kernel_require_in_warn_with_uplevel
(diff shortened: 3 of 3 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ branches:
     - trying
     - /^[\d.]+$/
     - /.+-stable$/
+    - /^bug\//
```
segiddins: adding this shouldn't be necessary, since you're making a PR into a branch that's already tested

appveyor.yml
```diff
@@ -4,6 +4,7 @@ branches:
   - master
   - auto
   - /[\d.]+/
+  - /^bug/
```
segiddins: same as above

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+if RUBY_VERSION >= "2.5"
+  module Kernel
+    path = "#{__dir__}/"
+    original_warn = instance_method(:warn)
+    Module.new {define_method(:warn, original_warn)}
```
duckinator: This anonymous module is never used, as far as I can tell.

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
...
+    original_warn = method(:warn)
```
duckinator: This and line 6 appear to conflict. Which is it supposed to be?

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    module_function define_method(:warn) {|*messages, uplevel: nil|
+      if uplevel
+        uplevel, = [uplevel].pack("l!").unpack("l!")
```
duckinator: What is this even doing?

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
...
+        if uplevel >= 0
+          start = 0
+          begin
+            loc, = caller_locations(start, 1)
+            break start += uplevel unless loc
```
duckinator: This is a bit more readable:

```
unless loc
  start += uplevel
  break start
end
```

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
...
+            start += 1
+          end while (loc.path.start_with?(path) or (uplevel -= 1) >= 0)
```
duckinator: Two things with how this loop is done:

1. `loop do`/`end` and a conditional `break` is generally preferable over `begin`/`end while ...`.
2. Please avoid assignments in conditions, especially complicated conditions like this one.

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
...
+          uplevel = start
+        end
+        original_warn.call(*messages, uplevel: uplevel)
+      else
+        original_warn.call(*messages)
+      end
```
duckinator: Instead of:

```ruby
if uplevel
  <...>
else
  original_warn.call(*messages)
end
```

perhaps use a guard clause like this:

```ruby
unless uplevel
  original_warn.call(*messages)
end

<...>
```

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+if RUBY_VERSION >= "2.5"
```
duckinator: Please add a comment explaining why it's limited to Ruby 2.5+.

.travis.yml
MSP-Greg: Fork. Fork. Fork.  Nothing is nicer than working on a PR in one's fork and discovering that you can't run a test on it because the branch name isn't allowed.

Also, I think it's also used in another popular repo, one that Nobu and some others here occasionally contribute to...

lib/rubygems/core_ext/kernel_warn.rb
MSP-Greg: comment out 6 & 7 and see what happens...

lib/rubygems/core_ext/kernel_warn.rb
hsbt: Is it readable? I don't think so. I prefer original code-style.

nobu: @duckinator When no `uplevel`, `original_warn` will be called twice in your code.
It is not the intended behavior.

.travis.yml
nobu: Yes, the last commit is unnecessary, and I'll remove it.
It is just to run CIs before creating this PR.

lib/rubygems/core_ext/kernel_warn.rb
nobu: Converting `uplevel` to an `Integer` which fits `long`, with calling `to_int` if needed.
There is no such method right now.

lib/rubygems/core_ext/kernel_warn.rb
nobu: To suppress "method redefined" warning when `$VERBOSE`.

lib/rubygems/core_ext/kernel_warn.rb
duckinator: Sorry, I forgot a `return` in my suggestion! Looks like you figured it out, though.

lib/rubygems/core_ext/kernel_warn.rb
duckinator: Is there any reason to prefer this + `define_method` over the typical `alias`-followed-by-`def` approach along the lines of this?:

```ruby
alias original_warn warn

module_function def warn(*messages, uplevel: nil)
  # <...>
end
```

(I ask because that's more in line with how the rest of the RubyGems codebase does it.)

MSP-Greg: @duckinator consider the scope along with module_function.  Or, which method are you aliasing, class or instance?  Did you try your code?

duckinator: Ah, okay, scope explains it &mdash; my version leaks `original_warn`, which is worse inside of `Kernel` than it would be elsewhere. @nobu's approach avoids this problem.

So what @nobu did is definitely the better option here.

@MSP-Greg I did try it, and it functions as intended, but I initially missed the scope-related implications of it being in `Kernel`. Thanks for pointing that out.

duckinator:
Thanks for the PR, @nobu!

@bundlerbot r+

lib/rubygems.rb
```diff
@@ -1392,5 +1392,6 @@ def clear_default_specs
 
 require 'rubygems/core_ext/kernel_gem'
 require 'rubygems/core_ext/kernel_require'
+require 'rubygems/core_ext/kernel_warn'
```
y3i12: Facing issues with Ruby 2.4 as kernel_warn can't be loaded

MSP-Greg:
@y3i12 

I believe this PR/commit includes the file `rubygems/core_ext/kernel_warn.rb`, so the only way you could have a load error is if you're using a new version of `rubygems.rb` but an old version of the files in `rubygems/core_ext`?

Issue #2440, opened 2018-10-22
installer.rb - windows bin stubs - commit mixup (here & ruby/ruby)


MSP-Greg:
@hsbt

Apologies for not noticing this earlier.  It involves a PR to ruby/ruby for installer.rb that was overwritten by a 'merge upstream' commit.

See https://github.com/ruby/ruby/commits/trunk/lib/rubygems/installer.rb

The commit that was over-written was:

[RubyGems installer.rb - fix up my mistakes in r64582](https://github.com/ruby/ruby/commit/6853c9e1690ee9e28a3dae538dd6f02b6f69c801#diff-82bebe1a5eb4619a4b0976cbcee3c584).

As the code is currently, it works fine for installing/updating gems.  Where it does not work is when running 'make install', as the RbConfig file is in the build folder.  Old criteria was based on RbConfig info; the commit changes the criteria to whether `ruby.exe` exists in bindir.

Currently, the rake bundled gem has a bin stub of:
```
@ruby.exe "%~dpn0" %*
```

but it should be:

```
@"%~dp0ruby.exe" "%~dpn0" %*
```

The first bin stub runs the rake file using whatever ruby is in PATH.  The second runs the rake file with the Ruby in the same folder...

The commit fixes the issue.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:
trunk, master

MSP-Greg:
See 4c77a36 and #2444

PR #2444, opened 2018-10-25
Fix up installer.rb to work during make install

MSP-Greg:
# Description:

Currently, `Gem::Installer#windows_stub_script` (in `installer.rb`) works correctly when used in a Ruby installation.  When building Ruby, it does not work correctly in `make install`, since `bindir` and RbConfig path/folder information do not match up.

This commit allows the method to function correctly when used in `make install`.  It was accepted in ruby/ruby, but was overwritten by an update from this repo.

See Issue #2440.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -771,33 +771,38 @@ TEXT
   # return the stub script text used to launch the true Ruby script
 
   def windows_stub_script(bindir, bin_file_name)
-    rb_bindir = RbConfig::CONFIG["bindir"]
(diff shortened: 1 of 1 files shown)


duckinator:
@bundlerbot r+

hsbt:
@bundlerbot r+

hsbt:
@MSP-Greg Thanks. It's my failure. 

memo: the original commit of ruby/ruby is here https://github.com/ruby/ruby/commit/6853c9e1690ee9e28a3dae538dd6f02b6f69c801

PR #2446, opened 2018-10-26
Ensure platform-specific gems are installed when --ignore-dependencies is specified

camertron:
# Description:
First of all, a giant _thank you_ to everyone who maintains Rubygems - the Ruby community wouldn't be the same without you!

Now, down to business. It looks like Rubygems won't install platform-specific gems if passed the `--ignore-dependencies` option. For example, this is what happens on my computer using Rubygems 2.4.5.5 when I try to install libv8:

```
$> gem install libv8 -v 3.16.14.19 --ignore-dependencies
Building native extensions. This could take a while...
ERROR:  Error installing libv8:
	ERROR: Failed to build gem native extension.
... a bunch of GCC compiler output ...
```

Version 3.16.14.19 of libv8 exists on rubygems.org for my architecture (x86_64-darwin16) but Rubygems doesn't install it. Instead it falls back to the default platform (i.e. the "Ruby" platform) which requires me to compile v8 from source.

We can see that the `--ignore-dependencies` option is the culprit by omitting it:

```
$> gem install libv8 -v 3.16.14.19
Fetching libv8-3.16.14.19-x86_64-darwin-16.gem
Successfully installed libv8-3.16.14.19-x86_64-darwin-16
Parsing documentation for libv8-3.16.14.19-x86_64-darwin-16
Installing ri documentation for libv8-3.16.14.19-x86_64-darwin-16
Done installing documentation for libv8 after 0 seconds
1 gem installed 🎉 
```

(🎉  added for effect)

The correct, platform-specific version is installed as expected.

From what I've been able to determine from some late-night code spelunking, the `Gem::RemoteFetcher` class doesn't consider the platform when fetching the gem, which I believe is the responsibility of `Gem::DependencyInstaller`. In fact, the `install_gem_without_dependencies` method appears to almost entirely circumvent it for some reason. This PR is an attempt to use `Gem::DependencyInstaller` to install any non-local gem by moving the local-specific stuff into a separate method.

I barely know what I'm doing in this codebase, and would greatly appreciate some help writing tests. Thanks for your consideration!

# Tasks:

- [x] Describe the problem / feature
- [ ] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2449, opened 2018-10-26
update gem install for better platform specific support, --explain, --ignore-dependencies

MSP-Greg:
All pertain to `gem install`:

1. As noted in PR #2446, `--ignore-dependencies` does not work correctly with platform specific gems
2. `--explain` option is ignored when combined with `--ignore-dependencies`
3.  `--explain` option shows only gem name and version, platform information is missing

PR fixes the above.

Notes:

4. When using `--ignore-dependencies` with `--explain`, `Gem::NameTuple`'s are used instead of specs.
5. Tests passed before and after.
6. The fix for `--ignore-dependencies` is in `Gem::RemoteFetcher#download_to_cache`.  If a platform specific gem exists, but it's version is less than the generic (platform=ruby) gem, the generic gem is used.
7. Not a complicated PR, but since I can only check it on Windows, more testing would certainly be appreciated...

______________

# Tasks:

- [ ] Write tests

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -205,7 +205,12 @@ You can use `i` command instead of `install`.
         puts "Gems to install:"
 
         request_set.sorted_requests.each do |s|
-          puts "  #{s.full_name}"
+          # shows platform specific gems if used
+          if (plat = s.spec.platform) == "ruby"
+            puts "  #{s.full_name}"
+          else
+            puts "  #{s.full_name}-#{plat}"
+          end
         end
 
         return
@@ -235,6 +240,22 @@ You can use `i` command instead of `install`.
       dependency = Gem::Dependency.new name, req
       dependency.prerelease = options[:prerelease]
 
+      if options[:explain]
+        puts "Gems to install:"
+        ary = Gem::SpecFetcher.fetcher.search_for_dependency dependency
+        if Array === ary && Array === ary[0]
+          tuples = ary[0].map { |i| i[0] }
+          if p_tuple = tuples.find { |t| Gem::Platform.local.to_s == t.platform }
+            puts "  #{p_tuple.full_name}"
+          else
+            r_tuple = tuples.find { |t| 'ruby' == t.platform }
+            puts "  #{r_tuple.full_name}"
+          end
+        else
+          puts "  no matching gems found"
+        end
+        return
+      end
       fetcher = Gem::RemoteFetcher.fetcher
       gem = fetcher.download_to_cache dependency
     end
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -102,7 +102,19 @@ class Gem::RemoteFetcher
 
     return if found.empty?
 
-    spec, source = found.max_by { |(s,_)| s.version }
+    # platform Ruby spec
+    r_spec, r_source = found.max_by { |(s,_)| s.version }
+
+    # platform local spec
+    p_found = found.select { |(s,_)| s.platform != 'ruby' && Gem::Platform.match(s.platform) }
+    p_spec, p_source = p_found.max_by { |(s,_)| s.version }
+
+    # use platform specific gem if version at least equal to ruby gem
+    if p_spec && p_spec.version >= r_spec.version
+      spec = p_spec ; source = p_source
+    else
+      spec = r_spec ; source = r_source
+    end
 
     download spec, source.uri.to_s
   end


MSP-Greg:
Below are three examples of the changes from the PR.  generated on a `x64-mingw32` windows build

```
———————————— Not patched ————————————      —————————————— Patched ——————————————

C:\> gem install eventmachine --explain

Gems to install:                           Gems to install:
  eventmachine-1.2.7                         eventmachine-1.2.7-x64-mingw32


C:\> gem install eventmachine --explain --ignore-dependencies
  
** Installs                                Gems to install:
   eventmachine-1.2.7.gem                    eventmachine-1.2.7-x64-mingw32
                                             
                                             
C:\> gem install eventmachine --ignore-dependencies

** Installs                                ** Installs
   eventmachine-1.2.7.gem                     eventmachine-1.2.7-x64-mingw32.gem
```
`test_gem_commands_install_command.rb` does not have any test based on `platform`.
 `test_gem_remote_fetcher.rb` has one test (`test_download_platform_legacy`) using `platform`.  

I suspect that tests for this will be more involved that the PR code.  I personally would like tests for these so they aren't inadvertently broken in the future, but it might be a few days (or more)...

lib/rubygems/commands/install_command.rb
```diff
@@ -235,6 +240,22 @@ def install_gem_without_dependencies name, req # :nodoc:
       dependency = Gem::Dependency.new name, req
       dependency.prerelease = options[:prerelease]
 
+      if options[:explain]
+        puts "Gems to install:"
+        ary = Gem::SpecFetcher.fetcher.search_for_dependency dependency
+        if Array === ary && Array === ary[0]
```
bronzdoc: what you think of:

```ruby
ary.is_a?(Array) && ary[0].is_a?(Array)
```

lib/rubygems/commands/install_command.rb
```diff
@@ -235,6 +240,22 @@ def install_gem_without_dependencies name, req # :nodoc:
...
+          tuples = ary[0].map { |i| i[0] }
+          if p_tuple = tuples.find { |t| Gem::Platform.local.to_s == t.platform }
+            puts "  #{p_tuple.full_name}"
+          else
+            r_tuple = tuples.find { |t| 'ruby' == t.platform }
```
bronzdoc: could you use `"` instead of `'` for strings?

lib/rubygems/remote_fetcher.rb
```diff
@@ -102,7 +102,19 @@ def download_to_cache dependency
 
     return if found.empty?
 
-    spec, source = found.max_by { |(s,_)| s.version }
+    # platform Ruby spec
+    r_spec, r_source = found.max_by { |(s,_)| s.version }
+
+    # platform local spec
+    p_found = found.select { |(s,_)| s.platform != 'ruby' && Gem::Platform.match(s.platform) }
```
bronzdoc: same here `"` instead of `'` ?

bronzdoc:
can you write a test for it? thanks @MSP-Greg !

MSP-Greg:
@bronzdoc

`===` vs `is_a?` should be something any Ruby coder should be okay with, but I'll make the changes for that and the  quotes (vs apostrophe).

1. Please respond to the third comment re `gem update`.

2. So, I've got to change `update_command.rb` and add tests.  Three commits or squash them?

bronzdoc:
I totally missed your third comment @MSP-Greg sorry about that!

1. yes i'll say `eventmachine-1.2.7-x64-mingw32` in your given example

2. I just like to commit with related code, **i.e** change for update_command.rb and test for it would be a commit, but that is just my preference, i think we have no rules for that in the Rubygems codebase yet and i don't want to impose my preferences over your valuable contribution, so take it as a suggestion.

MSP-Greg:
@bronzdoc 

Thanks.  Re the commits, I'll see when I'm done, as I'm still not sure of the scope, and I'd like to finally clear up as many of the platform related issues as I can.  I may close this and do a couple of PR's, I don't know...

Another question re `gem update`, if you know the answer.  I could write a test, but I'd rather not.

Let's say I've got gem 'a' installed, and an update is available.  The updated gem has a different set of dependencies.  Currently, will that be taken into account with `gem update`?

I'm not sure yet, but it seems that `gem update` grabs a list of all installed gems and simply checks them for newer versions (with no dependency checking)?  Again, I don't know...

MSP-Greg:
@bronzdoc

I'm closing this and will submit a PR later.

1. Re commits, ok to submit a PR with 'test' commit and a 'code' commit?  I thought I'd run the test commit in my fork to verify that master is not passing most of the added tests.  Then add the code commit and create a PR.

2. You asked for a change from apostrophes to quotes (single to double).  Some of the file that I'm changing have a lot of apostrophes.  Do you want me to change them or leave them.  Also, if a type of statement typically uses apostrophes in a file, should similar statements I add follow that?

3. Add `# frozen_string_literal: true`?  I haven't added it to any files yet, but it I think RG should move towards checking it...

PR #2452, opened 2018-10-31
gem install/update, --explain, --ignore-dependencies, platform 

MSP-Greg:
`gem install` & `gem update` behavior when used with the `--explain` and/or `--ignore-dependencies` options is not consistent, along with their handling of pre-compiled or platform specific gems and the
`--platform=ruby` option..

Below are three examples of the changes from the PR.  generated on a `x64-mingw32` windows build

```
———————————— Not patched ————————————      —————————————— Patched ——————————————

C:\> gem install eventmachine --explain

Gems to install:                           Gems to install:
  eventmachine-1.2.7                         eventmachine-1.2.7-x64-mingw32


C:\> gem install eventmachine --explain --ignore-dependencies
  
** Installs                                Gems to install:
   eventmachine-1.2.7.gem                    eventmachine-1.2.7-x64-mingw32
                                             
                                             
C:\> gem install eventmachine --ignore-dependencies

** Installs                                ** Installs
   eventmachine-1.2.7.gem                     eventmachine-1.2.7-x64-mingw32.gem
```

Below is a summary of the news tests, 'F' indicates that it fails currently (before the 2nd code commit):
```
TestGemCommandsInstallCommand              test_gem_commands_install_command.rb
   test_install_gem_remote_platform_local
   test_install_gem_ignore_dependencies_local (previously both)
F  test_install_gem_ignore_dependencies_remote_platform_local
F  test_explain_platform_local
F  test_explain_platform_local_ignore_deps
F  test_explain_platform_ruby
F  test_explain_platform_ruby_ignore_deps

TestGemCommandsUpdateCommand               test_gem_commands_update_command.rb
F  test_explain_platform_local
   test_explain_platform_ruby

TestGemRemoteFetcher                       test_gem_remote_fetcher.rb
F  test_download_to_cache_local

TestGemSourceLocal                         test_gem_source_local.rb
   test_find_gem_platform_local
F  test_find_gem_platform_ruby
```

Two commits, the first changes/adds tests, the second fixes the code.  Additional discussion in PR #2446 & #2449.

**EDIT:** 2018-10-31 22:00 UTC - refactored and added `Gem::Platform.rank` method.

# Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -201,10 +201,14 @@ You can use `i` command instead of `install`.
       request_set = inst.resolve_dependencies name, req
 
       if options[:explain]
-        puts "Gems to install:"
+        say "Gems to install:"
 
         request_set.sorted_requests.each do |s|
-          puts "  #{s.full_name}"
+          # shows platform specific gems if used
+          say (plat = s.spec.platform) == Gem::Platform::RUBY ?
+            "  #{s.full_name}" :
+            "  #{s.full_name}-#{plat}"
+
         end
 
         return
@@ -234,6 +238,20 @@ You can use `i` command instead of `install`.
       dependency = Gem::Dependency.new name, req
       dependency.prerelease = options[:prerelease]
 
+      if options[:explain]
+        say "Gems to install:"
+        ary = Gem::SpecFetcher.fetcher.search_for_dependency dependency
+        if ary.is_a?(Array) && ary[0].is_a?(Array)
+          tuples = ary[0].map { |i| i[0] }
+          tuple = tuples.max_by { |s|
+            [s.version, Gem::Platform.rank(s.platform)]
+          }
+          say "  #{tuple.full_name}"
+        else
+          say "  no matching gems found"
+        end
+        return
+      end
       fetcher = Gem::RemoteFetcher.fetcher
       gem = fetcher.download_to_cache dependency
     end
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -30,6 +30,20 @@ class Gem::Platform
     end
   end
 
+  def self.rank(platform)
+    if Gem.platforms == [RUBY]     # --platform=ruby
+      if RUBY == platform then     0
+      else                        -1
+      end
+    else
+      if    local == platform then 2
+      elsif local =~ platform then 1
+      elsif RUBY  == platform then 0
+      else                        -1
+      end
+    end
+  end
+
   def self.installable?(spec)
     if spec.respond_to? :installable_platform?
       spec.installable_platform?
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -97,8 +97,10 @@ command to remove old versions.
     if options[:explain]
       say "Gems to update:"
 
-      gems_to_update.each do |(name, version)|
-        say "  #{name}-#{version}"
+      gems_to_update.each do |(name, version, platform)|
+        say platform == "ruby" ?
+          "  #{name}-#{version}" :
+          "  #{name}-#{version}-#{platform}"
       end
 
       return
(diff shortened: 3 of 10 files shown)


MSP-Greg:
Two issues:

1. I refactored once after the original PR, but missed changes to `lib/rubygems/test_utilities.rb` that the refactor allowed.

2. I believe I've gotten ruby version constraints working in `gem install`.  This would affect many of the code changes already in this PR, as it affects `gem update` and possibly others.  It will also add several tests (not even started).


**Two ruby-version constraint examples:**

1. psych 3.0.3 - on Windows, there is a pre-compiled gem, but it only works for a Ruby version up to 2.4.  Ruby 2.5.3 has 3.0.2 as a default.  When one tries to install/update psych, it tries the pre-compiled gem, and gives an error message.  It should try the `--platform=ruby` gem.

2. did_you_mean - starting with v1.1.0, it required Ruby >= 2.4, but the most recent release is v1.2.1, which only works with Ruby >= 2.5.  Using Ruby 2.4.5. `gem update`  tried 1.2.1 and failed.  It should try 1.1.2...

lib/rubygems/commands/install_command.rb
```diff
@@ -202,10 +202,15 @@ def install_gem name, version # :nodoc:
...
 
       if options[:explain]
-        puts "Gems to install:"
+        say "Gems to install:"
 
         request_set.sorted_requests.each do |s|
-          puts "  #{s.full_name}"
+          # shows platform specific gems if used
+          if (plat = s.spec.platform) == Gem::Platform::RUBY
+            say "  #{s.full_name}"
```
segiddins: `Gem::BasicSpecification#full_name` already includes the platform like this

lib/rubygems/platform.rb
```diff
@@ -30,6 +30,20 @@ def self.match(platform)
     end
   end
 
+  def self.rank(platform)
```
segiddins: is there any reason this sorts based upon the local platform, rather than being consistent across installs? See, for example, `Bundler::Resolver.sort_platforms` for a different approach

MSP-Greg: > is there any reason this sorts based upon the local platform

Yes, because it's determining what platform is selected for install/update.  Note that it's used in `max_by` selection.

Again, if you go the end of the method and insert a line `nil` at line 45, you'll see some of the new tests failing.

lib/rubygems/commands/install_command.rb
MSP-Greg: No.  Two item.  If you comment out the added conditional logic, the following test fails:
```
  2) Failure:
TestGemCommandsInstallCommand#test_explain_platform_local [C:/Greg/GitHub/rubygems/test/rubygems/test_gem_commands_install_command.rb:1154]:
Expected: "  a-2-x86-mswin32-60"
  Actual: "  a-2"
```

`RequestSet#sorted_requests` returns an array of `Gem::Resolver::ActivationRequest`.
Their `full_name` method is `@spec.full_name`.  But `@spec` is a `Gem::Resolver::IndexSpecification`, which inherits it's method from `Gem::Resolver::Specification`, and its `full_name` method's comments state:

The name and version of the specification.
Unlike `Specification#full_name`, the platform is not included.

MSP-Greg:
@segiddins

First of all, there is an error in this code.  I'm working on a better PR which finally takes into account Ruby version constraints/requirements.  But, I could push a squash to this PR with the change.

Secondly, in working on this I decided that the code comments need improvement. 

Regarding using 'Ruby version constraints/requirements', without it, I found two examples where a gem was not installed where one could be installed.  This occurred for both `gem install` & `gem update`.

But, unfortunately, the decision was made not to include 'Ruby version constraints/requirements' in the spec stub line.  So, many of the classes/methods that make use of 'simplified specs' become useless.

Since adding 'Ruby version constraints/requirements' is right back into the same code this PR is affecting, what would you like me to do?

MSP-Greg:
As I've said before, when reviewing more involved PR's, actually using/testing the patch goes a long way...

cc: @duckinator

lib/rubygems/commands/install_command.rb
segiddins: should we fix that implementation, then, rather than reimplenting what "real" specs do?

MSP-Greg: Your call.  Given that the comment mentions the difference, and it is not tagged :nodoc, is that a breaking change?

segiddins: Ah, didn't realize that was coming from the comment. In that case, maybe just extract into a method that, when given something that looks like a gem spec, does this dance, so we dont repeat it all over the place?

MSP-Greg: This messy.  There are two places where similar code exists in install and update.

But, update is using an array, install is using a `Gem::Resolver::Specification`.  So refactoring would only change install.  Hence, maybe (for the time being) leave as is?

Off topic - Apologies, as these sort of things cloud my working in RG.  Its not new, and its been worked on by a lot of people.  IMHO, some things have gotten to the point where they're just messed up.  Hopefully this weekend I'll post an issue with some thoughts on that, and what I see as possible solutions...

MSP-Greg:
Please see PR #2539 'Fix platform and ruby version issues with install/update'

It contains all the code from this PR, along with additions to correctly handle Ruby versions constraints in gemspecs.  Closing.

Issue #2395, opened 2018-09-10
Incorrect certificate error message 


Jac2NL:
I would like to suggest updating a security error message.

When I simulate an Man-in-the-Middle attack against rubygems.org and replace the SSL/TLS certificate then the Ruby application (in this case Logstash) reports the following error: 

```
# /usr/share/logstash/bin/logstash-plugin install
logstash-some-plugin
Validating logstash-some-plugin
ERROR:  SSL verification error at depth 0: unable to get local issuer certificate (20)
ERROR:  You must add /CN=jac2/OU=etc/O=etc/C=NL to your local trusted store
Unable to download data from https://rubygems.org - certificate verify failed
(https://api.rubygems.org/latest_specs.4.8.gz)
ERROR: Installation aborted, verification failed for logstash-some-plugin
```

My issue with this error is the "You must add <cert> to your local trusted store" part. i believe it should be left to the user to resolve the problem. Could be that the user forgot to add the root certificate to the truststore, could be that the server is not sending proper intermediate certificates, or it could be a MITM... In that case the user should not continue. 

For comparison, here's how wget handles the same situation:

```
Connecting to api.rubygems.org (api.rubygems.org)|192.168.x.x|:443... connected.
ERROR: cannot verify api.rubygems.org's certificate, issued by
‘C=NL,O=etc,OU=etc,CN=jac2’:
  Unable to locally verify the issuer's authority. 
```

This issue is related to:

  - [X] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.5.2.1
```

The code responsible for this error can be found here: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/request.rb#L120-L122

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

indirect:
@Jac2NL I think I agree with your reasoning, but Rubygems does not generate the text that you are complaining about. It passes through that text from the OpenSSL error message. Maybe you can file this against OpenSSL?

indirect:
Whoops, I was totally wrong about that. RubyGems _does_ print that message, and I agree that it could be misleading. Good suggestion!

bronzdoc:
[...]

PR #2454, opened 2018-10-31
Improve certificate error message

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/2395

Attempt to improve certificate error message by making less confusing when a `OpenSSL::X509::V_ERR_UNABLE_TO_GET_ISSUER_CERT_LOCALLY` or `OpenSSL::X509::V_ERR_UNABLE_TO_VERIFY_LEAF_SIGNATURE`

display a separate message for each instead of the same message.

Unfortunally we don't have access to more info in `Gem::Request.verify_certificate_message error_number` and having more info available in that method would require a lot of changes, so idk if its worth it.


hsbt:
@bundlerbot r+

djberg96:
👍

PR #2441, opened 2018-10-22
Drop ruby 1.8 support and use IO.popen

nobu:
IO.popen supports command line as an array since 1.9, as well as
system had done.  And both support redirection option since 1.9
too.  These prevents from excution by shell which could cause
unexpected expansions.

# Description:

______________

# Tasks:

- [x] Describe the problem / feature
- [ ] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2457, opened 2018-11-04
fix extension builder failure when verbose

sorah:
# Description:

Since #2441 the `command` argument at Builder.run could be an Array.

But Kernel#.system doesn't accept Array and may raise ArgumentError in such cases.

The reason behind this bug is having separate codepath to run co-process, just for really_verbose.
This patch changes the method to always use IO.popen, even in verbose.

## Steps to reproduce

```sh
echo "verbose: really" >> ~/.gemrc
gem i psych
# (or any gems you like, includes extension like nokogiri)
```

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2458, opened 2018-11-06
Update gem open command help text

sonalkr132:
# Description:
We have a hackerone report claiming `-e` is a "security vulnerability" as it can be used to execute any program and not just editors. For example, try `gem open rake -e "ls / "`, it will list files in your root directory. We call `System()` with [input to `-e` and gem path](https://github.com/sonalkr132/rubygems/blob/d8741925dcaa06bae5dda7b3cf0c0728a0a9e26e/lib/rubygems/commands/open_command.rb#L74-L78)
Ideally, this should be fixed with a whitelist of editor commands. However coming up is a whitelist which can cover different editors and their flags seems very difficult and perhaps unending task(users may want their own favorite editor whitelisted). Given the use case of the command, chances of someone getting burned by `gem open -e` command because they were using an tainted argument seems low. Should an unintended use case of gem command be in scope of our security guarantees? :thinking: 
At the very least, we should update the documentation with implementation detail, so that it doesn't come as surprise to anyone (specially, our HO reports).
This PR intends to keep advertised use case of `gem open` command the same.

# Tasks:

- [ ] Write tests


lib/rubygems/commands/open_command.rb
```diff
@@ -11,9 +11,9 @@ class Gem::Commands::OpenCommand < Gem::Command
   def initialize
     super 'open', 'Open gem sources in editor'
 
-    add_option('-e', '--editor EDITOR', String,
-               "Opens gem sources in EDITOR") do |editor, options|
-      options[:editor] = editor || get_env_editor
+    add_option('-e', '--editor COMMAND', String,
+               "Prepends COMMAND to gem path. Could be used to sepecify editor.") do |command, options|
```
indirect: s/sepecify/specify/

sonalkr132:
updated

indirect:
thanks!

(ugh I got confused about which repo this was, sorry for not using bundlerbot 😬)

PR #2465, opened 2018-11-11
Fix and lock rubocop

deivid-rodriguez:
# Description:

I was getting rubocop offenses against latest master. This is because I had `0.60.0` installed and that version was allowed by the previous requirement (`~> 0.58`), so the rake task was picking that version. 
`0.60.0` added [a bug fix](https://github.com/rubocop-hq/rubocop/pull/6333) around `Style/MutableConstant` that detects new offenses, so that's why it was failing.
 
This PR bumps the version of rubocop in use to `0.60.0`, and consistently sets the requirement to include the patch version (`~> 0.60.0`).

I would actually go as far as fully locking the dependency, because right now if the same thing happens with a patch level release (for example, `0.60.1`), the same problem will be introduced and get in the middle for contributors. But I haven't done it in this PR. 
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -72,12 +72,12 @@ rescue LoadError
 end
 
 begin
-  gem "rubocop", "~> 0.58"
+  gem "rubocop", "~> 0.59.0"
   require "rubocop/rake_task"
 
   RuboCop::RakeTask.new
 rescue LoadError
-  task(:rubocop) { abort "Install the rubocop gem (~> 0.58) to run a static analysis" }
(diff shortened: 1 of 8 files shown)


hsbt:
@deivid-rodriguez Can you rebase with the current master branch? I fixed build failure with ruby-head at https://travis-ci.org/rubygems/rubygems/builds/453671117

segiddins:
@Bundlerbot r+

PR #2468, opened 2018-11-13
Add a rubocop binstub

deivid-rodriguez:
# Description:

Currently there's no easy way to run the proper version of rubocop againt the rubygems repo. One can do `rake rubocop`, but that does not allow any rubocop options, like `--auto-correct`. One can also do, `rubocop _0.60.0_`, but it's not very handy. So I propose to replace the `rubocop` rake task with a binstub, because it's more flexible.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -71,15 +71,6 @@ rescue LoadError
   end
 end
 
-begin
-  gem "rubocop", "~> 0.60.0"
-  require "rubocop/rake_task"
-
-  RuboCop::RakeTask.new
-rescue LoadError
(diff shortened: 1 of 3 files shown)


hsbt:
Looks useful 👍

hsbt:
@bundlerbot r+

colby-swandale:
I like this change but it would be good to keep the rake task around so that we don't break anyone's workflow.

deivid-rodriguez:
@colby-swandale I'll do a follow up PR, you're right.

Issue #2388, opened 2018-08-27
OpenSSL 1.1.1 compatibility


voxik:
OpenSSL 1.1.1 have landed in Fedora Rawhide:

~~~
$ rpm -q openssl-devel
openssl-devel-1.1.1-0.pre9.fc29.x86_64
~~~

Running Ruby 2.5.1 test suite, I observe the following error:

~~~
  1) Failure:
TestGemRemoteFetcher#test_do_not_allow_invalid_client_cert_auth_connection [/builddir/build/BUILD/ruby-2.5.1/test/rubygems/test_gem_remote_fetcher.rb:845]:
[Gem::RemoteFetcher::FetchError] exception expected, not
Class: <OpenSSL::SSL::SSLError>
Message: <"SSL_read: tlsv1 alert decrypt error">
---Backtrace---
/builddir/build/BUILD/ruby-2.5.1/.ext/common/openssl/buffering.rb:182:in `sysread_nonblock'
/builddir/build/BUILD/ruby-2.5.1/.ext/common/openssl/buffering.rb:182:in `read_nonblock'
/builddir/build/BUILD/ruby-2.5.1/lib/net/protocol.rb:175:in `rbuf_fill'
/builddir/build/BUILD/ruby-2.5.1/lib/net/protocol.rb:157:in `readuntil'
/builddir/build/BUILD/ruby-2.5.1/lib/net/protocol.rb:167:in `readline'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http/response.rb:40:in `read_status_line'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http/response.rb:29:in `read_new'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http.rb:1497:in `block in transport_request'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http.rb:1494:in `catch'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http.rb:1494:in `transport_request'
/builddir/build/BUILD/ruby-2.5.1/lib/net/http.rb:1467:in `request'
/builddir/build/BUILD/ruby-2.5.1/lib/rubygems/request.rb:221:in `perform_request'
/builddir/build/BUILD/ruby-2.5.1/lib/rubygems/request.rb:156:in `fetch'
/builddir/build/BUILD/ruby-2.5.1/lib/rubygems/remote_fetcher.rb:368:in `request'
[...]

PR #2486, opened 2018-11-21
fix request.rb for intermittently raised error using TLS 1.3

MSP-Greg:
# Description:

When running tests with a Ruby built with OpenSSL 1.1.1 or later, one test has intermittently failed both here and in Ruby trunk testing.  With trunk, sometimes the failure occurs when testing parallel, but passes on retry.  The test is:
```
TestGemRemoteFetcher#test_do_not_allow_invalid_client_cert_auth_connection
```

An error should be raised in the test, but this one starts with OpenSSL, passes thru net/http, and should be caught in `request.rb`.  Since it seems to only occur with OpenSSL 1.1.1/TLSv1_3, code somewhat accounts for that...

Two examples of test failure:
https://ci.appveyor.com/project/rubygems/rubygems/builds/20388929/job/lk27qkfajexyciuj#L30
https://ci.appveyor.com/project/MSP-Greg/ruby-loco/builds/20465475#L2933

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/request.rb
+++ b/lib/rubygems/request.rb
@@ -237,6 +237,14 @@ class Gem::Request
       verbose "fatal error"
 
       raise Gem::RemoteFetcher::FetchError.new('fatal error', @uri)
+    rescue (defined?(OpenSSL::SSL) ? OpenSSL::SSL::SSLError : nil) => e
+      # As of 2018-11, error is intermittently raised when using TLS 1.3 with
+      # improper client authentication
+      if OpenSSL::SSL.const_defined? :TLS1_3_VERSION
+        raise Gem::RemoteFetcher::FetchError.new("OpenSSL error: #{e.message}", @uri)
+      else
+        raise e
+      end
     # HACK work around EOFError bug in Net::HTTP
     # NOTE Errno::ECONNABORTED raised a lot on Windows, and make impossible
     # to install gems.


MSP-Greg:
As of today 2018-11-21, the only 'standard' rubies available that are using OpenSSL 1.1.1 are windows Ruby builds 2.5.3 or greater.  The current Travis rubies (both Ubuntu & OSX/MacOS) are using earlier versions.

OpenSSL 1.1.1 is the first version with TLSv1_3, and it operates a little differently than TLSv1_2.  Some of the differences are only an issue with CI where a server and a client are both being created for testing...

bronzdoc:
@MSP-Greg can you add a test? thanks!

MSP-Greg:
See [RG issue 2388](https://github.com/rubygems/rubygems/issues/2388).

@bronzdoc

> can you add a test? thanks!

If I could consistently repo the error, yes.  Since I can't, no.

As mentioned in the first msg, while using a new version of OpenSSL/TLS, this test has intermittently failed/errored due to a particular OpenSSL error being raised and not caught.

But, an error should be raised.  I could remove the code from `lib/rubygems/request.rb` and add the error the code in `test_gem_remote_fetcher.rb`?

MSP-Greg:
@bronzdoc 

> Maybe we can mock the error with Object.send(:remove_const, :TLS1_3_VERSION) ?

The issue occurs when TLS v1.3 exists.  On Windows, I don't believe I've ever repo'd it locally.  Appveyor fails occasionally on it (both testing rugygems/rubygems & ruby/ruby).

Note that in [ruby/openssl issue 227](https://github.com/ruby/openssl/issues/227), it's stated that there are systems where the error being raised is not intermittent.  Also note the statement "TLS 1.3, which is new in OpenSSL 1.1.1, handles a client certificate differently from TLS 1.2."

This is messy, as until more repos test with more OS's and OpenSSL 1.1.1 configurations, I don't think it's well defined.  Importantly, this issue here is about OpenSSL / net/http / RubyGems correctly sensing a client security error, and stopping it.  The problem is an error is raised by OpenSSL that isn't allowed for or caught.

I was an early adopter of 1.1.1, and was building & testing prereleases.  I used it with ruby, puma, & eventmachine, and was involved in updating them & MSYS2 to use it.  By no means am I an OpenSSL expert, but I've seen what adjustments have been needed with TLS 1.3.

Long story short, I doubt this is the only change that will be needed with TLS 1.3, but it gets the CI to be consistent.  As mentioned earlier, the test code could be changed, rather than `request.rb`.

MSP-Greg:
@bronzdoc

Was looking to see about other trunk CI testing (make install before make test?), and came across the following:

http://rubyci.s3.amazonaws.com/ubuntu/ruby-trunk/log/20181129T003002Z.log.html.gz#rubyspec

Scroll up a bit, and you'll see this error on an Ubuntu 18.10 build...

lib/rubygems/request.rb
```diff
@@ -237,6 +237,14 @@ def perform_request(request) # :nodoc:
       verbose "fatal error"
 
       raise Gem::RemoteFetcher::FetchError.new('fatal error', @uri)
+    rescue (defined?(OpenSSL::SSL) ? OpenSSL::SSL::SSLError : nil) => e
+      # As of 2018-11, error is intermittently raised when using TLS 1.3 with
+      # improper client authentication
+      if OpenSSL::SSL.const_defined? :TLS1_3_VERSION
```
bronzdoc: Should this be checked earlier in https://github.com/rubygems/rubygems/blob/fccc3139d4bf0924af090e487b3e752697fe8e4c/lib/rubygems/request.rb#L134 since we already checking if `OpenSSL::SSL` is defined there?

lib/rubygems/request.rb
```diff
@@ -237,6 +237,14 @@ def perform_request(request) # :nodoc:
       verbose "fatal error"
 
       raise Gem::RemoteFetcher::FetchError.new('fatal error', @uri)
+    rescue (defined?(OpenSSL::SSL) ? OpenSSL::SSL::SSLError : nil) => e
```
bronzdoc: Why we rescuing for nil? if `OpenSSL::SSL` is not defined?

bronzdoc:
Thanks for the explanation @MSP-Greg i think i understand better now, i left some comments

lib/rubygems/request.rb
MSP-Greg: I believe all of the caller listings show hitting RG first at request.rb:222.  The rescue in 240 catches errors in line 222.

lib/rubygems/request.rb
MSP-Greg: I, possibly incorrectly, assumed that rescue was a form of `is_a?`.  It's been a few days since I did the PR, but I think I locally tested to see if it was catching other errors.

For the test to pass, it has to receive a `Gem::RemoteFetcher::FetchError` error, and if the CI passed for Ruby versions without TLS v1.3, nil is not catching other errors.

MSP-Greg:
@bronzdoc You're welcome, and thanks for the thanks.

MSP-Greg:
Intermittent test issues are difficult to test.  @hsbt did a commit that supersedes this, see ruby/ruby r66163 (GitHub [edbac1b](https://github.com/ruby/ruby/commit/edbac1b986d6cbacf191bd1fde15e5de801a4935)).

Closing.

PR #2487, opened 2018-11-22
Drop ruby 2.2 support

deivid-rodriguez:
# Description:

On #2466 I run into some ruby 2.2 specific testing issues. So I wondered why we still support MRI 2.2. The reason is that when we discussed and agreed to drop support for EOL'd rubies, MRI 2.2 had not yet been EOL'd. The original plan was to release rubygems 3.0 _before_ the 31st of march, 2018, and after that drop 2.2 support for rubygems 3.1. See https://github.com/rubygems/rubygems/pull/2182#issuecomment-364638099.

That plan didn't actually happen since rubygems 3.0 has not yet been released, so I think we can drop 2.2 support for rubygems 3.0.

This would also bring the ruby support range in line with bundler 2.0, which will only support rubies >= 2.3.0.
______________

# Tasks:

- [ ] Write tests


hsbt:
Can you update https://github.com/rubygems/rubygems/blob/master/rubygems-update.gemspec#L32 too?

deivid-rodriguez:
@hsbt That's already part of the patch: https://github.com/rubygems/rubygems/blob/2551f023bceb24f4014de19f4f90f283d71e5920/rubygems-update.gemspec#L32.

indirect:
I approve of this change 👍

hsbt:
@deivid-rodriguez Ah, Sorry. I missed it.

hsbt:
@bundlerbot r+

deivid-rodriguez:
No problem @hsbt, thanks a lot!

PR #2491, opened 2018-11-23
Update manifest

colby-swandale:
# Description:

This PR is 2 parts. The first part is that there has been a few PRs recently that have added some new files, but didn't update `Manifest.txt`.

The second part is that there is an bug in the Regex that is excluding files in `Manifest.txt` that it shouldn't be. If you run `rake update_manifest` in `master`, files that contain `doc` will be removed. I've update the regex to make sure it does not remove those files.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -309,6 +309,7 @@ lib/rubygems/compatibility.rb
 lib/rubygems/config_file.rb
 lib/rubygems/core_ext/kernel_gem.rb
 lib/rubygems/core_ext/kernel_require.rb
+lib/rubygems/core_ext/kernel_warn.rb
 lib/rubygems/defaults.rb
 lib/rubygems/dependency.rb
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
One question @colby-swandale and @bronzdoc. Should we actually ship `util/*` with rubygems? It seems to be only rubygems development scripts, so maybe that folder should be actually excluded as well?

bronzdoc:
> Should we actually ship util/* with rubygems? It seems to be only rubygems development scripts, so maybe that folder should be actually excluded as well?

@deivid-rodriguez  my opinion is that it shouldn't

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
@bronzdoc So you think we shouldn't ship the files in the `utils` folder? In that case, should we create a PR to remove the files from the manifest?

PR #2492, opened 2018-11-23
Enable cop Style/MethodCallWithArgsParentheses

bronzdoc:
# Description:
This cop checks presence of parentheses in method calls containing parameters.

for more information about this cop: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MethodCallWithArgsParentheses
______________


deivid-rodriguez:
I do agree that many times method calls with parenthesis read better, but I feel a diff of "+13,409 -13,405" is just too much. This would require a curated list of exceptions that it's going to take a lot of time to compile and review. For example, the beginning of the diff starts by putting parenthesis around `require` calls, which is a style I've never seen.

bronzdoc:
> the beginning of the diff starts by putting parenthesis around require calls, which is a style I've never seen

aha! that is true, thanks for catching it @deivid-rodriguez  

I used the `--safe-auto-correct` flag of rubocop.

Currently with this cop enabled, Rubygems has `13389 offenses`

djberg96:
I was able to reduce it to "only" 12k or so by adding:

`IgnoredMethods: ['require', 'sh', 'task', 'desc', 'puts']`

PR #2494, opened 2018-11-26
Backport ruby core repository

hsbt:
# Description:

Backport https://github.com/ruby/ruby/commit/f7e94df0d66aff47745d189ab5a7858a92233d57 and https://github.com/ruby/ruby/commit/0edc36b9f25e7b92c40d91858dfe4ef2bd06f796

______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2495, opened 2018-11-26
Backport ruby core r65992

k0kubun:
# Description:

Backport https://github.com/ruby/ruby/commit/af6adb9982b93d3e960bc6fe6fa37d01b8f9f130

This resolves an random error depending on load order, which is caused by Ruby 2.6's Bundler integration http://ci.rvm.jp/results/trunk-no-mjit@silicon-docker/1479769

We assume this happens in some load order between rubygems and bundler, but TBH we haven't figured it out yet.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/indexer.rb
+++ b/lib/rubygems/indexer.rb
@@ -4,10 +4,17 @@ require 'rubygems/package'
 require 'time'
 require 'tmpdir'
 
+rescue_exceptions = [LoadError]
+begin
+  require 'bundler/errors'
+rescue LoadError # this rubygems + old ruby
+else # this rubygems + ruby trunk with bundler
+  rescue_exceptions << Bundler::GemfileNotFound
+end
 begin
   gem 'builder'
   require 'builder/xchar'
-rescue LoadError
+rescue *rescue_exceptions
 end
 
 ##


segiddins:
Can you please give a description of what this is trying to solve, and ideally add a test case? Thanks!

k0kubun:
> Can you please give a description of what this is trying to solve

Sorry, I was suggested to follow https://github.com/rubygems/rubygems/pull/2494, but I should have prepared a better description for this case. I updated the PR's description a little (still basically it's the same as commit message's one).

> ideally add a test case

Agreed that point, but to be honest I don't know how Bundler integration of Ruby is working (only @hsbt knows), and I've never figured the exact load order to reproduce this. 

But at least Ruby's CI often fails without fixing this and at least we can catch the breakage when we merge rubygems to Ruby. As a future improvement, I want to delegate the task to write a test case to @hsbt since it's tightly related to Bundler integration.

hsbt:
@bundlerbot r+

segiddins:
@hsbt why is this merged without test coverage?

hsbt:
I know your concern. But It fixed only the test suite of ruby core repository with parallel execution. Therefore you can not see it with this repository.

segiddins:
We should still be able to write a test that ensures we don’t regress the behavior — otherwise, there’d be no reason to not deleted the added code, because no breakage would be observed.

hsbt:
I don't agree with your opinion. We need to focus to resolve the user problem, not test code.

colby-swandale:
I agree with @segiddins, It may only be fixing the tests in the ruby core repository but we still need a test to make sure we don't break this functionality in the future and disrupt the ruby project.

PR #2497, opened 2018-11-27
use '|' to preserve line breaks in welcome message

colby-swandale:
# Description:

I used the wrong block in the welcome message YAML. `>` strips the new line breaks, we want to use `|` instead.

# Tasks:

- [ ] Write tests


colby-swandale:
@bundlerbot r+

PR #2502, opened 2018-11-30
Support case insensitve file systems other than Windows when checking that user bin dir is in PATH

MSP-Greg:
# Fixes:

test_gem_ext_builder.rb - fix test skip, test now passes on Windows

test/rubygems/test_gem_installer.rb - add test `test_check_that_user_bin_dir_is_in_path_case_insens`

installer.rb - allow for ALT_SEPARATOR or SEPARATOR in Windows ENV items, update case handling

# Description:

1.  Properly handle casing issues.
2. Newer versions of Windows allows File::SEPARATOR in paths, so ENV['PATH'] & ENV['HOME'] can contain either slashes or backslashes.

# Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -695,7 +695,19 @@ class Gem::Installer
     path = ENV['PATH']
     path = path.tr(File::ALT_SEPARATOR, File::SEPARATOR) if File::ALT_SEPARATOR
 
-    if Gem.win_platform?
+    # Windows file system is case insensitive, but other OS's may also be,
+    # so we need to check
+    # test_dir folder needs to exist, try a few...
+    test_dir = if    Dir.exist? user_bin_dir  then user_bin_dir
+               elsif Dir.exist? Gem.user_home then Gem.user_home
+               else  __dir__
+               end
+
+    # fs_case_insens is true for case insensitive file systems
+    fs_case_insens =
+      test_dir != test_dir.swapcase && File.identical?(test_dir, test_dir.swapcase)
+
+    if fs_case_insens
       path = path.downcase
       user_bin_dir = user_bin_dir.downcase
     end
--- a/test/rubygems/test_gem_installer.rb
+++ b/test/rubygems/test_gem_installer.rb
@@ -219,6 +219,15 @@ gem 'other', version
     end
 
     assert_empty @ui.error
+
+    ENV['PATH'] = [orig_PATH, bin_dir.swapcase].join(File::PATH_SEPARATOR)
+
+    use_ui @ui do
+      installer.check_that_user_bin_dir_is_in_path
+    end
+
+    assert_empty @ui.error
+
   ensure
     ENV['PATH'] = orig_PATH
   end


lib/rubygems/installer.rb
```diff
@@ -698,24 +698,23 @@ def process_options # :nodoc:
 
   def check_that_user_bin_dir_is_in_path # :nodoc:
     user_bin_dir = @bin_dir || Gem.bindir(gem_home)
-    user_bin_dir = user_bin_dir.gsub(File::SEPARATOR, File::ALT_SEPARATOR) if
-      File::ALT_SEPARATOR
 
     path = ENV['PATH']
+
+    # Windows paths are case insensitive, but with either ALT_SEPARATOR
```
djberg96: HFS is case insensitive, too. Is that relevant here?

lib/rubygems/installer.rb
```diff
@@ -698,24 +698,23 @@ def process_options # :nodoc:
...
+    # or SEPARATOR
     if Gem.win_platform?
-      path = path.downcase
-      user_bin_dir = user_bin_dir.downcase
+      path = path.downcase.gsub(File::SEPARATOR, File::ALT_SEPARATOR)
```
djberg96: For single character replacement, I think `tr` is preferred (and faster).

test/rubygems/test_gem_ext_builder.rb
```diff
@@ -228,20 +228,21 @@ def test_build_extensions_extconf_bad
...
 
     refute_path_exists @spec.gem_build_complete_path
 
+    # windows Gem.ruby is SEPARATOR, gem_make.out is ALT_SEPARATOR
+    gem_ruby = !File::ALT_SEPARATOR ? Gem.ruby :
+      Gem.ruby.gsub(File::SEPARATOR, File::ALT_SEPARATOR)
+
     skip "Gem.ruby is not the name of the binary being run in the end" \
-      unless File.read(gem_make_out).include? "#{Gem.ruby}:"
+      unless make_out.include? "#{gem_ruby}:"
```
djberg96: Should we add parens here to continue the journey towards rubocop compiance?

MSP-Greg: Re parens, I recall seeing it for method defs.  Is it also intended for method calls?

lib/rubygems/installer.rb
MSP-Greg: Once other issues resolved, I'll update.

lib/rubygems/installer.rb
MSP-Greg: How can one test for HFS?

djberg96: You can test for case insensitivity in general with something like this:

```
require 'tempfile'
tempfile = Tempfile.new
p File.exists?(tempfile.path.downcase) && File.exists?(tempfile.path.upcase)
```
On my Linux laptop, this prints false. On my Mac, this prints true.

test/rubygems/test_gem_ext_builder.rb
djberg96: @MSP-Greg Not sure if there's a way to differentiate them with Rubocop, so maybe? I was just thinking of https://github.com/rubygems/rubygems/pull/2492#issuecomment-441370126 where I added exceptions to several method calls to try to cut down the warnings.

colby-swandale: params are not required for method calls. And IMO it's not something that should be enforced with rubocop

lib/rubygems/installer.rb
MSP-Greg: Rather than write a file, how about
```ruby
if __FILE__ != __FILE__.downcase
  FS_CASE_INSENS = File.exists?(__FILE__.downcase)
elsif __FILE__ != __FILE.upcase
  FS_CASE_INSENS = File.exists?(__FILE__.upcase)
else
  puts "indeterminate?"
end
p "FS_CASE_INSENS #{FS_CASE_INSENS}"
```

test/rubygems/test_gem_ext_builder.rb
MSP-Greg: I'll change them, and thanks for the reminder to check it with RuboCop.  I often forget...

MSP-Greg: @colby-swandale

Thanks for clarifying.  Sometimes I do, sometimes I don't...

lib/rubygems/installer.rb
MSP-Greg: @colby-swandale

Re the above issue, I know there is similar code elsewhere in RG (I may have done a PR?).  Not sure if it was tests or 'code'.

Might something like this be added as a constant?  I may suggest it in ruby/ruby, as I only thought it was windows...

djberg96: @MSP-Greg Ah, true, we could just use `__FILE__`, good idea!

MSP-Greg: @djberg96 - updated revised code...

MSP-Greg:
Rather than test the case sensitivity of `__FILE__`, I think it would be better to check `Gem.user_dir`.

Feedback?

djberg96:
@MSP-Greg On my Windows VM, `Gem.user_dir` does not actually exist, though.

MSP-Greg:
@djberg96 

> Gem.user_dir does not actually exist

Earlier today I was looking at that, and wondered what RG did when that was the case.

But, re using `__FILE__` for the test of case sensitivity, `check_that_user_bin_dir_is_in_path` could be testing a folder that could differ from `__FILE__` in case sensitivity?  I used to have a network with Windows, macOS/OSX, and *nix boxes on it, but not now.  So, I don't know...

I'm starting to wonder if, rather than trying to determine if the folders in question are case sensitive, whether it would be better to test for both and have a warning message that a case insensitive match was found...  I think that made sense.

BTW, I did get caught by this issue once with some CI testing...

MSP-Greg:
Just updated.  Added a test, improved case insensitive test in installer.rb.

MSP-Greg:
FYI, the test for case insensitive should use `File.identical?`, as it would currently pass in case sensitive file systems if one actually had directories like `/c/RubyGems` and  `/c/rubygems`.  Unlikely, but possible.  Pointed out by nobu.  I will change that...

djberg96:
I either didn't know about or completely forgot about that method. Yes, use that. :)

djberg96:
:+1:

MSP-Greg:
Last change - use `swapcase` instead of `upcase/downcase`.  Suggested by @nobu

I thought about putting it in a comment, and nobu mentioned about it.  If the path tested consists of all 'non-casable' characters (non letters?), the test defaults to 'case sensitive'...

lib/rubygems/installer.rb
```diff
@@ -698,25 +698,39 @@ def process_options # :nodoc:
...
+           else  __dir__
+           end
+
+    # fs_case_insens is true for case insensitive file systems
+    fs_case_insens =
+      test != test.swapcase && File.identical?(test, test.swapcase)
+
+    # Newer Windows paths can use either ALT_SEPARATOR (\) or SEPARATOR (/)
+    if File::ALT_SEPARATOR
+      include = user_bin_dir.tr(File::ALT_SEPARATOR, File::SEPARATOR)
```
olleolleolle: (Could the local variable name `include` be something else? Syntax highlighting is confused about its status.)

lib/rubygems/installer.rb
colby-swandale: @MSP-Greg sorry, i didn't see this comment. A constant sounds great!

MSP-Greg: @colby-swandale

The issue of determining whether the location in question is case insensitive was also discussed in Ruby core, and Nobu had some good suggestions which are included in the current PR (there's more than a few updates).

There are some edge cases where things get messy, like a path where `path == path.swapcase`, networked drives/symlinks, etc.  I believe the code defaults to 'case sensitive` in these cases.

lib/rubygems/installer.rb
MSP-Greg: > Syntax highlighting is confused

I hate it when that happens…  I'll update & rebase,  maybe tomorrow...

olleolleolle: (Also: here's fan-mail about your dedication to keeping Windows + Ruby good: 💌  )

MSP-Greg: @olleolleolle 

Thank you. And thanks for all your work.  Fixed...

djberg96: It's still confused because of `Kernel#test`.

MSP-Greg:
I have a lot of Ruby versions, so I use `--user-install` a lot.  I'm really tired of seeing the "you don't have **/bin" in your path.

This PR fixes that.  Any changes needed?

simi:
@MSP-Greg any chance to rebase and let it run on current CI which also runs tests on Windows?

MSP-Greg:
@simi

Rebased locally and tests pass.  I'd like to double check that it works locally with both separators and casing in my Path, I'll need to check that later, along with linting...

deivid-rodriguez:
@MSP-Greg The PR still shows conflicts though.

MSP-Greg:
All done...

simi:
:thinking: Would it make sense to move "file presence regarding case sensitivity based on system" to method in `Gem::Util`? There's already some Windows related code as well.

MSP-Greg:
@simi

I don't care.

It's not really Windows specific, as case insensitive file systems are available on other OS's.

Since a system could have access to both case sensitive and case insensitive drives/folders, etc, it is dependent on the folder tested.

I don't recall whether path comparisons are done elsewhere.  It could be left as is, and moved if another use for it is found?

deivid-rodriguez:
I find this PR hard to review because in a single commit it's introducing (that I can see):

* Changes in specs to (I guess) pass on more systems with different case sensitivity.
* Splitting a test into two different tests for case sensitive/unsensitive systems.
* Add a way inside a method in `Gem::Installer` (not in general) to detect case sensitivity of a system, instead of checking for Windows directly.
* Introduce several refactorings inside `Gem::Installer` and tests.
* Introduce several style appeasing changes inside `Gem::Installer` and tests.

I'm going to open a separate PR with only the style changes and refactorings, and then you can rebase  this PR so that it doesn't include anything other than the "casing issues" you want to fix. Ok?

deivid-rodriguez:
I extracted https://github.com/rubygems/rubygems/pull/3589 with the style changes and refactorings that I found it could be worth adding.

If you want I can force-push the rebased version of this PR to your branch so that you don't need to rebase yourself.

deivid-rodriguez:
Rebased and force-pushed, hopefully CI is OK.

Regarding the changes in here, I'm not sure about some of them.

* What other case insensitive OS do we support? HFS was mentioned, which seems to be some MacOS thing, but I guess not the default since we already test MacOS and things seem fine? We haven't had any reports about issues here outside of Windows, so I'm really not sure about introducing this check, specially when it's limited to a single place in the whole code base.

* The `ENV["PATH"] fix about it possible including the alternative File::PATH_SEPARATOR makes sense to me. I hadn't spotted that fix before removing all the code style changes by the way. I guess the extra test you added is testing that fix? If this is the case, could you extract that fix and the test to a separate PR? If it's not the case, could you extract the fix, and add a test for it, to a separate PR?

* Regarding the changes in `test/rubygems/test_gem_ext_builder.rb`, have you had any issues when testing locally or something? Tests run fine on our CI, and some contributors have run tests locally to without issues. Why is this conversion needed for the path of this particular file and not in other places?

MSP-Greg:
As to why the case insensitive code was made generic instead of Windows specific, some comments above address that.

Supporting platforms is distinct from small updates to accommodate them. See https://rubyci.org/.  There are always commits in ruby/ruby to accommodate them.

Given that the original problem involves only an incorrect nuisance warning when using `--user-install`, I should have closed this along time ago.

deivid-rodriguez:
> Supporting platforms is distinct from small updates to accommodate them. See https://rubyci.org/. There are always commits in ruby/ruby to accommodate them.

Sure, but normally we get contributions from ruby-core contributors when failures on the "expanded ruby-core CI matrix happen", and I don't think we ever heard from them about this issue?

> Given that the original problem involves only an incorrect nuisance warning when using --user-install, I should have closed this along time ago.

This is exactly what I'd like you to explain. What problem you were facing when you added all this code, and how the code fixes it? I think the issue was related to your `PATH` (or some CI you use) using `File::ALT_SEPARATOR`, and that making this warning the be displayed to you as a "false positive". I agree that that is a bug, and we should fix it, and I'd like you to extract that specific fix to a separate PR if that's possible.

Let me know if you're willing to do this, otherwise I can do it and you can perhaps verify it in your box?

MSP-Greg:
Thanks for the co-author on the commits.

> removing all the code style changes, by the way

Main change I see are the nested `unless` statements, with the outer containing nothing but the inner.  Is that what you're referring to?

I'm happy to change the code to work on Windows only and split the test changes.  I think the previous code normalized things for case, but on Windows it should also been done for the separator.

JFYI, I started coding almost fifty years ago on my father's timeshare/teletype.  Hence, I'm probably old enough to not care about how many PR's/commits I have here or there.  I also have a low tolerance for...

I'm just interesting in making OSS better.  Hence, if I see messy code in methods that I'm working in, I will simplify it.

simi:
> I'm just interesting in making OSS better. Hence, if I see messy code in methods that I'm working in, I will simplify it.

That's totally ok, but to make the review easier it is good to not mix cosmetic changes and behaviour changes. Usually it is enough to split this into two commits.

deivid-rodriguez:
I extracted the `PATH` bug fix to #3829. Then we can rebase this PR again and see what's left in here.

djberg96:
Looks like this needs a fix + rebase.

deivid-rodriguez:
I rebased this and fixed conflicts and reworked the code a bit to minimize cosmetic changes.

I'm still quite hesitant about adding this, but I don't mind introducing it if others feels strongly.

Let's see what CI says in any case.

deivid-rodriguez:
I rebased this but after a second thought, noboby has manifested a strong interest here. I'm not a fan of introducing this just in this single place and without a strong reason, given nobody using a case insensitive system other than Windows has actually complained about anything. And CI is failing anyways.

I'll just close it.

djberg96:
@deivid-rodriguez I'm curious what your concerns are with this change. Seems pretty mild to me overall.

deivid-rodriguez:
I think it's a bit pointless to introduce generic case insensitive file system support at a single place in the code base. If some users of non Windows case insensitive file systems are having issues and we want to support them, we should properly introduce full support and get it verified on such a system. But so far I haven't heard of such users or such problems.

deivid-rodriguez:
Also I said 10 months ago that even if I didn't fancy the change myself, I wouldn't oppose if others felt strongly in favor. I think 10 months of silence is enough to assume that nobody feels strongly about it :)

PR #2128, opened 2017-12-25
Support pre-release RubyGems.

hsbt:
We need to ship a pre-release version of RubyGems for evaluation.

PR #2506, opened 2018-12-03
Fix crash on certain gemspecs

deivid-rodriguez:
# Description:

If `version` was a prerelease, and was defined before `name` in the gemspec, rubygems would crash.

This commit fixes that.

This regression was introduced in https://github.com/rubygems/rubygems/pull/2128.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2693,7 +2693,7 @@ class Gem::Specification < Gem::BasicSpecification
     @version = Gem::Version.create(version)
     # skip to set required_ruby_version when pre-released rubygems.
     # It caused to raise CircularDependencyError
-    if @version.prerelease? && @name.strip != "rubygems"
+    if @version.prerelease? && (@name.nil? || @name.strip != "rubygems")
       self.required_rubygems_version = '> 1.3.1'
     end
     invalidate_memoized_attributes
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1345,6 +1345,16 @@ dependencies: []
     assert_equal '/path/to/file', e.file_path
   end
 
+  def test_initialize_prerelease_version_before_name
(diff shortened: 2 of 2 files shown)


lib/rubygems/specification.rb
```diff
@@ -2693,7 +2693,7 @@ def version=(version)
     @version = Gem::Version.create(version)
     # skip to set required_ruby_version when pre-released rubygems.
     # It caused to raise CircularDependencyError
-    if @version.prerelease? && @name.strip != "rubygems"
+    if @version.prerelease? && (@name.nil? || @name.strip != "rubygems")
```
djberg96: Should we do a casecmp here in case of "RuByGeMs"?

deivid-rodriguez: Maybe it's a good idea, because `bundler` seems to have changed the casing of this string, actually :point_right: https://github.com/bundler/bundler/commit/fcc4b607379b25f607ce8d003664902f58fefd23.

deivid-rodriguez:
I having one thought, but I'm not sure about it. Would it make sense to actually remove this check? The current version of `bundler` vendored by `rubygems` which only supports [`rubygems 1.3.6`](https://github.com/bundler/bundler/blob/a578c4ee390fadb746252b4bd496f8ee6d90add4/bundler.gemspec#L42), and this has been like that for ages (bundler 0.9.15, released back in 2010 already had this requirement).

So I'm not sure adding this check for `self.required_rubygems_version = '> 1.3.1'` is going to ever be useful?

hsbt:
@deivid-rodriguez Thanks! I confirmed to fix your test-case with this patch.

> So I'm not sure this check for `self.required_rubygems_version = '> 1.3.1'` is going to ever be useful?

I will upgrade the vendored bundler version to 2.0 after it's released. I think we can remove this code when it's time.

hsbt:
@bundlerbot r+

PR #2510, opened 2018-12-05
Move rubocop into a separate stage in travis ci

colby-swandale:
# Description:

TIL it's now possible to use multiple stages in Travis-CI while using the `rvm` matrix. With that, this PR moves rubocop into it's own stage - before any tests are executed. This should provide faster feedback to contributors and make a clear separation between linting and tests.

Here's an [example build](https://travis-ci.org/rubygems/rubygems/builds/463775265) (this failed because it was a custom build that didn't have the changes to run rubocop by itself). The build for this PR should use the separate stages as well.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,16 @@ rvm:
 env:
   - "TEST_TOOL=rubygems YAML=psych"
   - "TEST_TOOL=bundler RGV=master"
+jobs:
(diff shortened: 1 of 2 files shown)


bronzdoc:
@bundlerbot r+

Issue #1831, opened 2017-02-01
ERROR: SSL verification error at depth 2: certificate has expired (10)


paulgasbarra:
I'm having a problem or would like to suggest a feature.

My current problem is when trying to 'gem install bundle' I get the following message:

ERROR:  SSL verification error at depth 2: certificate has expired (10)
ERROR:  Certificate /C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA expired at 2014-01-28T12:00:00Z
ERROR:  SSL verification error at depth 2: certificate has expired (10)
ERROR:  Certificate /C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA expired at 2014-01-28T12:00:00Z
ERROR:  SSL verification error at depth 2: certificate has expired (10)
ERROR:  Certificate /C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA expired at 2014-01-28T12:00:00Z
ERROR:  Could not find a valid gem 'bundler' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)
ERROR:  SSL verification error at depth 2: certificate has expired (10)
ERROR:  Certificate /C=BE/O=GlobalSign nv-sa/OU=Root CA/CN=GlobalSign Root CA expired at 2014-01-28T12:00:00Z

I manually updated the file as instructed here: http://guides.rubygems.org/ssl-certificate-update/

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x ] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.10
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

colby-swandale:
I'm closing this for now. If you're still experiencing your original issue don't be afraid to re-open this ticket.

PR #2514, opened 2018-12-06
Multifactor authentication for yank command.

ecnelises:
# Description:

I added check for `otp` field, as multi-factor authentication feature to rubygems. Currently master branch of rubygems.org API has supported it. (See [#1831](https://github.com/rubygems/rubygems.org/pull/1831)) This PR will add respective support for client. 
______________

# Tasks:

- [x] Write tests

cc: @sonalkr132

### DIFF
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -88,7 +88,7 @@ permission to.
         response = send_owner_request(method, name, owner)
 
         if need_otp? response
-          response = send_owner_request(method, name, owner, true)
+          response = send_owner_request(method, name, owner)
         end
 
         action = method == :delete ? "Removing" : "Adding"
@@ -102,11 +102,11 @@ permission to.
 
   private
 
-  def send_owner_request(method, name, owner, use_otp = false)
+  def send_owner_request(method, name, owner)
     rubygems_api_request method, "api/v1/gems/#{name}/owners" do |request|
       request.set_form_data 'email' => owner
       request.add_field "Authorization", api_key
-      request.add_field "OTP", options[:otp] if use_otp
+      request.add_field "OTP", options[:otp] if options[:otp]
     end
   end
 
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -86,12 +86,6 @@ permission to.
     owners.each do |owner|
       begin
         response = send_owner_request(method, name, owner)
-
-        if need_ask_otp? response
-          ask_otp
-          response = send_owner_request(method, name, owner)
-        end
-
         action = method == :delete ? "Removing" : "Adding"
 
         with_response response, "#{action} #{owner}"
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -88,6 +88,7 @@ permission to.
         response = send_owner_request(method, name, owner)
 
         if need_ask_otp? response
+          ask_otp
           response = send_owner_request(method, name, owner)
         end
 
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -87,7 +87,7 @@ permission to.
       begin
         response = send_owner_request(method, name, owner)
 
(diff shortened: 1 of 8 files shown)


lib/rubygems/commands/yank_command.rb
```diff
@@ -61,15 +62,23 @@ def execute
 
   def yank_gem(version, platform)
     say "Yanking gem from #{self.host}..."
-    yank_api_request(:delete, version, platform, "api/v1/gems/yank")
+    args = [:delete, version, platform, "api/v1/gems/yank"]
+    response = yank_api_request(*args)
```
sonalkr132: Why are we not passing `options[:otp]` here? I don't think we need `use_otp` argument. If `options[:otp]` is empty/nil, don't add it to request.
Please fix this for other commands as well.

test/rubygems/test_gem_commands_yank_command.rb
```diff
@@ -57,6 +58,50 @@ def test_execute
     assert_equal [yank_uri], @fetcher.paths
   end
 
+  def test_execute_with_otp_success
+    response_fail = "You have enabled multifactor authentication but your request doesn't have the correct OTP code. Please check it and retry."
```
sonalkr132: single quotes here and at 86 please.

test/rubygems/test_gem_commands_yank_command.rb
```diff
@@ -57,6 +58,50 @@ def test_execute
     assert_equal [yank_uri], @fetcher.paths
   end
 
+  def test_execute_with_otp_success
+    response_fail = 'You have enabled multifactor authentication but your request doesn\'t have the correct OTP code. Please check it and retry.'
+    yank_uri = 'http://example/api/v1/gems/yank'
+    @fetcher.data[yank_uri] = proc do
+      @call_count ||= 0
+      (@call_count += 1).odd? ? [response_fail, 401, 'Unauthorized'] : ['Successfully yanked', 200, 'OK']
```
sonalkr132: what does this do again?

hsbt:
@sonalkr132 @ecnelises Please notify me when this pull-request is done status.

lib/rubygems/commands/owner_command.rb
```diff
@@ -87,8 +87,8 @@ def manage_owners(method, name, owners)
       begin
         response = send_owner_request(method, name, owner)
 
-        if need_otp? response
-          response = send_owner_request(method, name, owner, true)
+        if need_ask_otp? response
```
sonalkr132: AFAIR, I had suggested that checking if otp is needed and asking for otp should be separate function. This would cleaner when written like:
```
if need_otp? response
  ask_otp
  response = send_owner_request(method, name, owner)
end
```

ecnelises: Yes, I planned to commit this in another commit.

lib/rubygems/test_utilities.rb
```diff
@@ -48,7 +48,14 @@ def find_data(path)
       raise Gem::RemoteFetcher::FetchError.new("no data for #{path}", path)
     end
 
-    @data[path]
+    data = @data[path]
+
+    if data.respond_to?(:flatten!)
+      data.flatten!
+      data.size <= nargs ? data.last(nargs) : data.shift(nargs)
```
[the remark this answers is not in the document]
ecnelises: I mean the last 3 values should be kept regardless of how many times they are fetched. Otherwise the fake response array would be empty and cause error.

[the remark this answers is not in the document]
sonalkr132: > the last 3 values should be kept regardless of how many times they are fetched

is this a requirement for tests you have added (since `flatten` would be true only for nested array) ? if yes, explicitly declare it in test. ie
```
ok_res = [response_success, 200, 'OK']
@stub_fetcher.data["..."] = [
     [response_fail, 401, 'Unauthorized'],
     ok_res,
     ok_res,
     ...
    ]
```

IMHO, when testing, some repetition is tolerable if it makes things simpler to understand.

sonalkr132:
ping!

ecnelises:
Updated.

hsbt:
@bundlerbot r+

Issue #1881, opened 2017-03-29
symlinks containing .. get rewritten to an incorrect target by `gem build`


petertseng:
I'm **having a problem** ~~or would like to suggest a feature~~.

My current problem is that symlinks containing `..` get rewritten to an incorrect target by `gem build`.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.11
```

  - [x] I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

With apologies to those who worked on#1578 and #1457 -
I found that if I have a symlink in `dir/link1` pointing to `../file1`, when I use `gem build` and `gem install` on it, the link target gets rewritten to `dir/../file1`.
Since symlink targets are resolved relative to the symlink, that means `dir/link1` is attempting to look for `dir/dir/../file1` or `dir/file1` instead of just `../file1` or `file1`.

Here is a script I used to test this behaviour:

```bash
tmpdir=$(mktemp -d)
# if mktemp not available on system:
#tmpdir=testgem
cwd=$(pwd)
cd $tmpdir

mkdir dir
echo contents1 > file1
echo contents2 > dir/file2
ln -s ../file1 dir/link1
ln -s dir/file2 link2

if [ "$(cat dir/link1)" != "contents1" ]; then
  echo "\033[1;31mSETUP WRONG, link1 doesn't contain contents1\033[0m"
fi

if [ "$(cat link2)" != "contents2" ]; then
  echo "\033[1;31mSETUP WRONG, link2 doesn't contain contents2\033[0m"
fi

echo 'Gem::Specification.new { |s| s.name = ?a; s.version = ?1; s.summary = ?a; s.authors = ?a; s.files = %w(file1 link2 dir/link1 dir/file2) }' > a.gemspec
gem build a.gemspec
gem install a

dir=$GEM_HOME
if [ -z "$dir" ]; then
  dir=$(gem environment | grep USER | cut -d: -f2 | sed -e 's/^ *//')
fi
adir="$dir/gems/a-1"

l1="$adir/dir/link1"
if [ -L "$l1" ]; then
  if [ "$(cat "$l1")" != "contents1" ]; then
    echo "\033[1;31mINSTALLED WRONG, link1 doesn't contain contents1\033[0m"
    readlink "$l1"
[...]

Issue #2007, opened 2017-09-08
Please revert PR#1578


voxik:
In #1577 there was reported issue with symlinks and PR #1578 supposedly fixed this. But unfortunately, it probably just hides the issue and makes things worser. Look for example at redhataccess/ascii_binder/issues/64

~~~
$ gem unpack ascii_binder
Fetching: ascii_binder-0.1.10.1.gem (100%)
Unpacked gem: '/home/vondruch/ff/ascii_binder-0.1.10.1'

$ ll ascii_binder-0.1.10.1/bin/
celkem 12
-rwxr-xr-x. 1 vondruch vondruch 9864  8. zář 10.59 asciibinder
lrwxrwxrwx. 1 vondruch vondruch   15  8. zář 10.59 ascii_binder -> bin/asciibinder
~~~

Apparently, the ```ascii_binder``` should be linke to ```asciibinder```, but it points to ```bin/asciibinder```.

I'd say that the test attached to #1578 is actually wrong. Specifically the line https://github.com/rubygems/rubygems/blob/master/test/rubygems/test_gem_package.rb#L170 should be actually:

~~~
assert_equal [{'lib/code_sym.rb' => 'code.rb'}], symlinks
~~~

and the rest of the code should be adjusted to make this test case pass.

~~~
$ gem -v
2.6.11
~~~

voxik:
Just FTR, I now observe similar error as the one from #1577:

~~~
/builddir/build/BUILD/ascii_binder-0.1.10.1/usr/share/gems/gems/ascii_binder-0.1.10.1/bin/ascii_binder maybe `gem pristine ascii_binder` will fix it?
~~~

The problem here is that the link is there, but it is broken, so the File.exist? on lin https://github.com/rubygems/rubygems/blob/master/lib/rubygems/installer.rb#L474 returs false:

~~~
$ irb
irb(main):001:0> File.exist? 'bin/ascii_binder'
=> false
irb(main):002:0> File.symlink? 'bin/ascii_binder'
=> true
irb(main):003:0> File.stat 'bin/ascii_binder'
Errno::ENOENT: No such file or directory @ rb_file_s_stat - bin/ascii_binder
	from (irb):3:in `stat'
	from (irb):3
	from /usr/bin/irb:11:in `<main>'
irb(main):004:0> File.lstat 'bin/ascii_binder'
[...]

PR #1578, opened 2016-04-04
Fix broken symlink support in tar writer (+ fix broken test)

e2:
# Description:

Fix for: https://github.com/rubygems/rubygems/issues/1577

(Avoids installing symlinks before symlink targets have been copied).

# Tasks:

- [X] Describe the problem / feature
- [ ] Write tests
- [X] Write code to solve the problem
- [ ] Get code review from coworkers / friends
- [ ] [Squash commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html)

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

Issue #1577, opened 2016-04-04
installing gem with symlinked binaries fails (bug in tar writer symlink handling)


e2:
#### Problem:

The `tmuxinator` gem has `bin/mux` symlinked to `bin/tmuxinator`. 

Upon installation, `gem install tmuxinator` fails with the warning from here:

https://github.com/rubygems/rubygems/blob/089a893/lib/rubygems/installer.rb#L468

Result: symlink ends up pointing to non-existing file.

#### Why this happens:

RubyGems first tries to install the file `bin/mux`, but the `bin/tmuxinator` doesn't exist yet, so there's a warning. 

This issue is related to:
  - [X] Installing a library
  - [X] The command line `gem`

Here are my current environment details:

```
$ gem env version
2.5.1
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

#### How to reproduce reliably (using Docker):

```
$ docker run -it e2dk/rvm_user_install:mri-2.3.0                                                                                                                                                              
guest@741883a67378:~$ gem install tmuxinator
Fetching: thor-0.19.1.gem (100%)
Successfully installed thor-0.19.1
Fetching: erubis-2.7.0.gem (100%)
Successfully installed erubis-2.7.0
Fetching: tmuxinator-0.7.2.gem (100%)
/home/guest/.rvm/gems/ruby-2.3.0/gems/tmuxinator-0.7.2/bin/mux maybe `gem pristine tmuxinator` will fix it?
    __________________________________________________________
    ..........................................................
    Thank you for installing tmuxinator.
    Make sure that you've set these variables in your ENV:
      $EDITOR, $SHELL
    You can run `tmuxinator doctor` to make sure everything is set.
    Happy tmuxing with tmuxinator!
    ..........................................................
    __________________________________________________________
Successfully installed tmuxinator-0.7.2
3 gems installed
guest@741883a67378:~$ gem contents tmuxinator|grep bin
/home/guest/.rvm/gems/ruby-2.3.0/gems/tmuxinator-0.7.2/bin/mux
/home/guest/.rvm/gems/ruby-2.3.0/gems/tmuxinator-0.7.2/bin/tmuxinator
guest@741883a67378:~$ mux
bash: mux: command not found
guest@741883a67378:~$ tmuxinator version
tmuxinator 0.7.2
[...]

PR #2516, opened 2018-12-07
Fix broken symlink that points to ../*

amatsuda:
# Description:

This patch fixes a bug that `gem build` packages a broken symlink file for symlink targeting to a file or directory starting with `../`.  

This fixes a regression introduced via #1578, and addresses #1881.

______________

# Tasks:

- [x] Write tests


amatsuda:
Well, #2512 seems to be just reverting the whole code and test added in #1578. So I suppose we'll be facing the #1577 issue (with symlinks that aren't starting with `.`) again.

My patch here handles both kinds of symlinks without modifying any existing test pattern.

deivid-rodriguez:
Yeah, makes sense. So basically this PR:

Closes #2512.
Closes #2007.
Fixes #1881.

amatsuda:
Exactly, sir!

bronzdoc:
@bundlerbot r+

PR #2517, opened 2018-12-07
Use net/http, not net/https.

duckinator:
# Description:

`net/https` has been merged into `net/http` since at least 1.9.3.

The [ruby-doc.org page for net/http in Ruby 1.9.3](https://ruby-doc.org/stdlib-1.9.3/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTPSession-label-HTTPS) says `In previous versions of ruby you would need to require 'net/https' to use HTTPS. This is no longer true.`

As far as I know, we are no longer supporting Ruby <2.0, so there is no point in differentiating.

Note that `lib/rubygems/request.rb` unconditionally requires `net/http` earlier in the file, which is why I just removed the `require 'net/https'` instead of changing it.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/request.rb
+++ b/lib/rubygems/request.rb
@@ -45,7 +45,6 @@ class Gem::Request
   end
 
   def self.configure_connection_for_https(connection, cert_files)
-    require 'net/https'
     connection.use_ssl = true
     connection.verify_mode =
       Gem.configuration.ssl_verify_mode || OpenSSL::SSL::VERIFY_PEER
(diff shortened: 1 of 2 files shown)


segiddins:
Nice!
@bundlerbot r+

duckinator:
@hsbt should I make a PR reverting this change + open an issue to re-apply it after those are backported?

nurse:
RubyGems 3.0 supports Ruby2.3 or latet but 2.3 is already security patch only. Therefore it will be far future.

Issue #2437, opened 2018-10-16
questionable code in unit test


duckinator:
I'm having a problem or would like to suggest a feature.

My current problem is I found some questionable code in a test would like feedback from folks with more experience in this part of the codebase.

If you look at [test/rubygems/test_gem_remote_fetcher.rb, lines 821-826](https://github.com/rubygems/rubygems/blob/cc881a8/test/rubygems/test_gem_remote_fetcher.rb#L821-L826), you see the following code:

```ruby
    temp_ca_cert = File.join(DIR, 'ca_cert.pem'),
    with_configured_fetcher(":ssl_ca_cert: #{temp_ca_cert}") do |fetcher|
      assert_raises Gem::RemoteFetcher::FetchError do
        fetcher.fetch_path("https://localhost:#{ssl_server.config[:Port]}/insecure_redirect?to=#{@server_uri}")
      end
    end
```

The part I find questionable is the trailing comma in `temp_ca_cert = File.join(DIR, 'ca_cert.pem'),`: the comma means the following line is part of the variable assignment.

In effect, you get the equivalent of:

```ruby
    temp_ca_cert = [
      File.join(DIR, 'ca_cert.pem'),
      with_configured_fetcher(":ssl_ca_cert: #{temp_ca_cert}") { ... }
    ]
```

And, continuing to simplify the code to attempt to determine the behavior, you get this:

```
~$ cat test.rb
some_variable = [
  "a",
  "the first item should be 'a', but is actually #{some_variable.inspect}"
]
p some_variable
~$ ruby test.rb
["a", "the first item should be 'a', but is actually nil"]
~$
```

**This means that the `temp_ca_cert` variable should be nil the one time it's used in that function — but removing the comma, which resolves that, has no effect whatsoever on the test suite.** This makes me question if the test is too broad in the types of errors it accepts.

I can make a PR removing the comma, but it feels wrong that setting the value of the variable to or nil doesn't affect the test at all.

---

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2520, opened 2018-12-10
Fix remote fetcher tests

bronzdoc:
Closes https://github.com/rubygems/rubygems/issues/2437

This fix an erroneous test when testing insecure redirects...

What we wanted to test was `Gem::RemoteFetcher::FetchError` being raised by an insecure redirect, but instead we were testing for a `Gem::RemoteFetcher::FetchError` raised by a `nil` ca_cert raised in 
https://github.com/rubygems/rubygems/blob/cfa1d68962fa46cfbef624362813127a4a98a9f8/lib/rubygems/request/http_pool.rb#L38
which bubbles up to:
https://github.com/rubygems/rubygems/blob/cfa1d68962fa46cfbef624362813127a4a98a9f8/lib/rubygems/request.rb#L132


bronzdoc:
I think i broke master

bronzdoc:
@bundlerbot r+

PR #2524, opened 2018-12-12
Sort files and test_files in specifications

cbaines:
# Description:

Currently, when writing out specifications, the order of the elements
in files and test_files can differ. Even if the set of files (in the
mathematical sense) is the same, the different order causes the files
or test_files to look different.

This causes problems with the reproducibility of working with
specifications. You can install a gem on different machines, and get
gemspecs that differ.

This change sorts the files and test_files, to protect against the
specifications differing unnecessarily.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2254,6 +2254,7 @@ class Gem::Specification < Gem::BasicSpecification
 
       attributes.each do |attr_name|
         current_value = self.send attr_name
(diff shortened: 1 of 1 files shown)


colby-swandale:
Thanks for the PR! This will need a test before we consider merging this.

hsbt:
@bundlerbot r+

segiddins:
It looks like a test was never added?

PR #2529, opened 2018-12-16
Fix tests when --program-suffix and similar ruby configure options are used

jeremyevans:
Without this, the tests check for `bin/foo.cmd` instead of `bin/foo.cmd${suffix}` and fail with `Errno::ENOENT`.

### DIFF
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -5,6 +5,7 @@ require 'rubygems/command'
 require 'rubygems/installer'
 require 'pathname'
 require 'tmpdir'
+require 'rbconfig'
 
 # TODO: push this up to test_case.rb once battle tested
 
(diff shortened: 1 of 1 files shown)


hsbt:
@jeremyevans Thanks! I will merge this after passing Travis CI.

hsbt:
@bundlerbot r+

johnl:
This is causing problems for me on Ubuntu: https://bugs.ruby-lang.org/issues/15466

The installed commands don't have the expected suffixes, so the tests fail.

I'm a bit unsure whether the problem is with the behaviour or the tests. Any thoughts?

jeremyevans:
This is an issue with the tests only.  The tests should check the `--format-executable` setting, and only check for the suffix if it is set to true.  I'm guessing I hit the issue with failing tests the other way because `--format-executable` defaults to true in my environment (so that gem binaries use the same suffix as ruby itself).  I'll see if I can submit a pull request to fix this issue within a few days.

jeremyevans:
#2549 should fix this.

PR #2531, opened 2018-12-17
Fix another test when --program-suffix and similar ruby configure options are used

jeremyevans:
Sorry I missed this in #2529.


hsbt:
@bundlerbot r+

PR #2537, opened 2018-12-20
Remove dead code in Gem::Validator.

duckinator:
# Description:

In `Gem::Validator`, `verify_gem` was a function that did nothing and has had a comment saying "remove me?" since 2011. This function was only called once, in a private function.

After removing the only call to that function, I was able to refactor away `Gem::Validator#verify_gem_file` entirely.

# Tasks:

- [ ] Write tests


segiddins:
Nice!
@bundlerbot r+

PR #2538, opened 2018-12-20
Removing yaml require

lucianosousa:
# Description:
Removing dead ruby 1.x code.
Since rubygems only support Ruby 2.3 and higher versions there is no need for this anymore
______________

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

Issue #2527, opened 2018-12-14
.gitignore - add lib/rubygems/defaults ?


MSP-Greg:
Assuming there is a desire to allow people to easily contribute to RubyGems, sometimes it's helpful to use the actual repo as the source of files, especially if one wants to check against several Ruby versions.

Some builds (including MinGW) add files to `lib/rubygems/defaults` to hook build systems, etc.

Any objections to adding this to `.gitignore`?

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:

```
Multiple
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

hsbt:
What's happened? Please show the reproduce instructions and its results.

hsbt:
Please provided the reproduction instructions. I can do it with my windows environment.

MSP-Greg:
AFAIK, there's two ways to use RG git repo in an existing Ruby install.

1. Either from the command line or thru RUBYOPT, add the RG master lib folder with `-I`.  This does leave the RG files in both install & git repo accessible (they're both in $LOAD_PATH).

2. Remove the RG folders/files from the install, and symlink to the RG git repo items.

I've got lots of Ruby versions on my system, so adding more using the second option is not an issue.  JFYI, the folder names match Appveyor, which makes local testing easier.  I've also got a .ps1 file to switch PATH around...

Hence, to repo, from an install:

a) remove `rubygems.rb`, `ubygems.rb`, and the `rubygems` folder.  

b) add symlinks to the above into the RG git repo.

c) try to build an extension gem

`rubygems/defaults/operating_system` enables the build system, so adding `defaults` to `.gitignore`...

hsbt:
I also have the many of ruby versions, but I never encountered its situation.

So, It seems your environmental problem. It's out of scope from the upstream repository.

MSP-Greg:
> So, It seems your environmental problem.

No.  Been using MSFT OS’s since DOS days, configured lots of PC’s and servers…

Ruby on Windows does not require having build tools in one’s path, and the MSYS2 system may conflict with other software, like Git for Windows (which is built with it).  I prefer to set up my systems the same way normal users would.

That’s what 'defaults/operating_system.rb’ is for.

Maybe /misc and /notes.txt should be removed from .gitignore?

PR #2539, opened 2018-12-21
Fix platform and ruby version issues with install/update

MSP-Greg:
# Description:

This PR contains the code for PR #2452, but adds code to correctly account for Ruby version constraints in gemspecs.  Since the cache contains tuples without the Ruby version info, specs do need to be downloaded, but an optimization to `Gem::Resolverfind_possible` helps offset the additional time needed.

`ENV['CI']` & `ENV'['TEST_SSL']` are also added to appveyor.yml.

I believe all of the additional tests currently fail on RubyGem.  Reviewing the tests will show the issues fixed by this PR.  Feel free to install the patch for the test commit and check it...

# Tasks:

- [X] Write tests


MSP-Greg:
@hsbt

One example that this fixes, mentioned previously.  Both using Ruby 2.5.3.

Since RG is currently not using gemspec's `required_ruby_version` constraint for gem selection, it tries to install `psych-3.0.3-x64-mingw32`.  After PR (2nd example) it correctly installs `psych-3.0.3`...

Current RubyGems:
```
PS C:\> gem install psych:3.0.3 --user-install
WARNING:  You don't have c:\users\greg l\.gem\ruby\2.5.0\bin in your PATH, gem executables will not run.
ERROR:  Error installing psych:
        The last version of psych (= 3.0.3) to support your Ruby & RubyGems was 3.0.3. Try installing it with `gem install psych -v 3.0.3`
        psych requires Ruby version >= 2.2, < 2.5. The current ruby version is 2.5.3.105.
```

RubyGems after PR:
```
PS C:\> gem install psych:3.0.3 --user-install
WARNING:  You don't have c:\users\greg l\.gem\ruby\2.5.0\bin in your PATH, gem executables will not run.
Temporarily enhancing PATH for MSYS/MINGW...
Successfully installed psych-3.0.3
1 gem installed
```

Warnings are eliminated by PR #2502.  2nd example needs item mentioned in [Issue 2527](https://github.com/rubygems/rubygems/issues/2527).

Obviously, there are ways around this issue.  Conversely, RubyGems should just work, especially for CI/cloud applications.

MSP-Greg:
@deivid-rodriguez @hsbt 

Apologies.  I set my 'target' fix based on real world testing & written tests, then proceed.  Along the way, I may pick up some trivial issues.  I also have a bias towards not forcing unneeded, trivial CI.

1. This PR in particular is dealing with a long standing issue in RG.  Hence, pulling it apart for the benefit of 'it's pretty hard to review' is not a big concern on mine.  It wasn't trivial to write, so it might not be trivial to review.  **It passes the tests**.  Or, said another way, I write my tests & code to identify & fix an issue, and I try to remember to comment it, 'ease of review' is secondary.

2. Re this and #2571, is there a preference for a) multiple commits in one pr, or b) mulitiple pr's?  Note that it may not be easy to separate the code such that multiple PR's pass...

3. @deivid-rodriguez - Issue #2575 - are you planning a PR, or should I continue to work on this one?

deivid-rodriguez:
> Re this and #2571, is there a preference for a) multiple commits in one pr, or b) mulitiple pr's? Note that it may not be easy to separate the code such that multiple PR's pass...

In my opinion in this case multiple PRs is better, but multiple commits would help too.

> Issue #2575 - are you planning a PR, or should I continue to work on this one?

Not at the moment.

deivid-rodriguez:
@MSP-Greg Are you still interested in submitting a PR to fix the `required_ruby_version` issue? At this point, all the other fixes you had included here have been fixed, so you should be able to come up with an isolated PR dealing with just that single issue. What do you think?

MSP-Greg:
> Are you still interested in submitting a PR to fix the `required_ruby_version` issue?

Yes.  I've been working on something else, so I do need to look at all the recent commits. I was 'waiting for the dust to settle'...

If you have time, please have a look at issue #2637.  To me, the current behavior is inconsistent.  It's separate from the `required_ruby_version` issue, but it will involve much of the same code.

> At this point, all the other fixes you had included here have been fixed

Thank you very much for your work on that.

deivid-rodriguez:
I don't mind extracting this myself if you don't mind either. I just need to remember to set the commits as co-authored, I don't want to steal any credit!

MSP-Greg:
Part of the reason I closed this is that I found a few more places that code needed to be revised, which in itself is not a good sign.  So, I've got more work on the issue past this PR.  I'll look at it later today or tomorrow am (-600).  Thanks again.

PR #2549, opened 2018-12-26
Fix tests when ruby --program-suffix is used without rubygems --format-executable

jeremyevans:
Add a test to check for correct behavior when --format-executable
option is used.


hsbt:
@bundlerbot retry

Issue #2551, opened 2018-12-26
Setting SOURCE_DATE_EPOCH env variable breaks several tests


johnl:
If the SOURCE_DATE_EPOCH environment variable is set to any value then several tests fail.

The Debian and Ubuntu packaging build system sets this environment variable to ensure reproducible builds, and runs the test suite as part of the build.

Most of the failures are TestGemPackageTarWriter tests, checking for mtime, but some TestGemSpecification too.

The breaking tests can obviously be fixed, but I'm wondering if the rubygems test suite should also be run with SOURCE_DATE_EPOCH set to catch these problems in future.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2552, opened 2018-12-27
Fix tests broken by a SOURCE_DATE_EPOCH env var

johnl:
This fixes several tests that are broken when the SOURCE_DATE_EPOCH environment
variable is set. TestGemPackageTarWriter was a bit complicated so I just unset
the env var for the duration of the tests, and restored it.

### DIFF
--- a/test/rubygems/test_gem_package_tar_writer.rb
+++ b/test/rubygems/test_gem_package_tar_writer.rb
@@ -11,11 +11,14 @@ class TestGemPackageTarWriter < Gem::Package::TarTestCase
     @data = 'abcde12345'
     @io = TempIO.new
     @tar_writer = Gem::Package::TarWriter.new @io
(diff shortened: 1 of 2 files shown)


johnl:
#2551

hsbt:
Thanks, But I'm going to merge https://github.com/ruby/ruby/commit/8cfbaf943efc0a533d0e6e6848951aecc262a9ff later for this issue.

johnl:
ah ok, glad it's fixed though (and slightly better than mine, with the ensures). Thanks!

Any thoughts on how to avoid this in future? Can the tests also be run with SOURCE_DATE_EPOCH enabled? Or is that overkill?

hsbt:
Fixed at https://github.com/rubygems/rubygems/commit/29bdf2a29642119b20576a94b38b4c7a7e9049a6 and https://github.com/rubygems/rubygems/commit/145bc40b31272c911618c9a87468b224e6a53a3d.

PR #2553, opened 2018-12-28
Test against Ruby 2.6

segiddins:
# Description:

Since 2.6 has been released, we should make sure we support it!

# Tasks:

- [ ] Write tests


appveyor.yml
```diff
@@ -12,6 +12,7 @@ environment:
   - ruby_version: 23-x64
   - ruby_version: 24-x64
   - ruby_version: 25-x64
+  - ruby_version: 26-x64
```
segiddins: ruby didn't seem to get installed?

.travis.yml
```diff
@@ -12,14 +12,15 @@ rvm:
   - 2.3.8
   - 2.4.5
   - 2.5.3
+  - 2.6.0
```
segiddins: Bundler is still failing on 2.6, working on that over in https://github.com/bundler/bundler/pull/6856

appveyor.yml
MSP-Greg: @segiddins

With normal CI, all Ruby builds (except trunk) used on Appveyor are pre-installed.

1. The RubyInstaller2 2.6.0 build is not passing due to a last minute update to BigDecimal, which changed spec tests (added to ruby/ruby, not yet added to ruby/spec).

2. Once released, Appveyor may take a few days (or weeks) to add the build to their images.

3. Trunk/head is not affected.

4. I'll ping when it's been added...

colby-swandale:
@segiddins what are you thoughts are leaving 2.6 out of appveyor for the time being so that we can merge this PR?

MSP-Greg:
See 810671e

Appveyor has not yet updated their images with Ruby 2.6.x.  They will be updating with 2.6.1, along with installing both Bundler 1.7.x & 2.x.

MSP-Greg:
@segiddins 

Appveyor has added Ruby 2.6.1.

JFYI, Bundler versions 1.17.3 & 2.0.1 are install on Ruby versions 2.3 & later.  At present, trunk/head only has what's included in the Rubygems std-lib.

See:
https://ci.appveyor.com/project/MSP-Greg/appveyor-ruby/builds/22294299#L908

hsbt:
@bundlerbot r+

PR #2554, opened 2018-12-28
Fix Gem::Requirement equality comparison when ~> operator is used

greysteil:
# Description:

The logic to compare versions included in a requirement needs to consider precision when a `~>` operator is used, and ignore it when one isn't.

Before this patch we had the following bug:
```ruby
Gem::Requirement.new("~> 5.2.0") == Gem::Requirement.new("~> 5.2")  # => true
```

Fixes https://github.com/bundler/bundler/issues/6858.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/requirement.rb
+++ b/lib/rubygems/requirement.rb
@@ -267,7 +267,17 @@ class Gem::Requirement
 
   def ==(other) # :nodoc:
     return unless Gem::Requirement === other
-    requirements == other.requirements
+
+    # An == check is always necessary
+    return false unless requirements == other.requirements
+
+    # An == check is sufficient unless any requirements use ~>
+    return true unless requirements.any? { |r| r.first == "~>" }
+
+    # If any requirements use ~> then we also need to compare the version
+    # strings for those requirements only (to check their precision is equal)
+    requirements.select { |r| r.first == "~>" }.
+      eql?(other.requirements.select { |r| r.first == "~>" })
   end
 
   private
--- a/lib/rubygems/requirement.rb
+++ b/lib/rubygems/requirement.rb
@@ -272,12 +272,17 @@ class Gem::Requirement
     return false unless requirements == other.requirements
 
     # An == check is sufficient unless any requirements use ~>
-    return true unless requirements.any? { |r| r.first == "~>" }
+    return true unless _tilde_requirements.any?
 
-    # If any requirements use ~> then we also need to compare the version
-    # strings for those requirements only (to check their precision is equal)
-    requirements.select { |r| r.first == "~>" }.
-      eql?(other.requirements.select { |r| r.first == "~>" })
+    # If any requirements use ~> we use the stricter `#eql?` that also checks
+    # that version precision is the same
+    _tilde_requirements.eql?(other._tilde_requirements)
+  end
+
+  protected
+
+  def _tilde_requirements
+    requirements.select { |r| r.first == "~>" }
   end
 
   private
--- a/test/rubygems/test_gem_requirement.rb
+++ b/test/rubygems/test_gem_requirement.rb
@@ -20,6 +20,12 @@ class TestGemRequirement < Gem::TestCase
     refute_requirement_equal "= 1.2", "= 1.3"
     refute_requirement_equal "= 1.3", "= 1.2"
 
+    refute_requirement_equal "~> 1.3", "~> 1.3.0"
+    refute_requirement_equal "~> 1.3.0", "~> 1.3"
+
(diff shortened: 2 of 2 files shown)


lib/rubygems/requirement.rb
```diff
@@ -267,7 +267,17 @@ def to_s # :nodoc:
...
+    # An == check is always necessary
+    return false unless requirements == other.requirements
+
+    # An == check is sufficient unless any requirements use ~>
+    return true unless requirements.any? { |r| r.first == "~>" }
+
+    # If any requirements use ~> then we also need to compare the version
+    # strings for those requirements only (to check their precision is equal)
+    requirements.select { |r| r.first == "~>" }.map { |r| r.last.version_string } ==
+      other.requirements.select { |r| r.first == "~>" }.map { |r| r.last.version_string }
```
deivid-rodriguez: `version_string` is already used by the `Version` class for `eql?`, so maybe if we use `eql?` here instead of `==`, we don't need to expose `#version_string`?

greysteil: We're comparing arrays here, and they might be in a different order.

`[1, 2].eql?([2, 1]) # => false`

deivid-rodriguez: But I think [the PR]( https://github.com/rubygems/rubygems/pull/2230) that made this regression surface precisely ensures a consistent order when the class is initialized?

deivid-rodriguez: The following change still passes tests, so if it's not correct, we should add more tests to prove it

```diff
diff --git a/lib/rubygems/requirement.rb b/lib/rubygems/requirement.rb
index 32dc824e..42399b79 100644
--- a/lib/rubygems/requirement.rb
+++ b/lib/rubygems/requirement.rb
@@ -276,8 +276,7 @@ class Gem::Requirement
 
     # If any requirements use ~> then we also need to compare the version
     # strings for those requirements only (to check their precision is equal)
-    requirements.select { |r| r.first == "~>" }.map { |r| r.last.version_string } ==
-      other.requirements.select { |r| r.first == "~>" }.map { |r| r.last.version_string }
+    requirements.select { |r| r.first == "~>" }.eql?(other.requirements.select { |r| r.first == "~>" })
   end
 
   private
```

greysteil: That's a good point. So could switch to:

```
requirements.select { |r| r.first == "~>" }.eql?(
  other.requirements.select { |r| r.first == "~>" }
)
```

greysteil: Haha, our comments crossed by that's identical code - I'll update. I did write a test for the case I was concerned about there.

deivid-rodriguez: Simultaneous commenting! :smiley:

lib/rubygems/requirement.rb
```diff
@@ -267,7 +267,17 @@ def to_s # :nodoc:
...
-    requirements == other.requirements
+
+    # An == check is always necessary
+    return false unless requirements == other.requirements
+
+    # An == check is sufficient unless any requirements use ~>
+    return true unless requirements.any? { |r| r.first == "~>" }
+
+    # If any requirements use ~> then we also need to compare the version
+    # strings for those requirements only (to check their precision is equal)
```
deivid-rodriguez: We should probably update this comment now. Maybe

```ruby
      # If any requirements use ~>, use the stricter `#eql?` that also checks that
      # version precision is the same
```

lib/rubygems/requirement.rb
```diff
@@ -267,7 +267,17 @@ def to_s # :nodoc:
...
+    requirements.select { |r| r.first == "~>" }.
+      eql?(other.requirements.select { |r| r.first == "~>" })
```
deivid-rodriguez: Do you think it's worth to extract the duplicated logic to utility methods like

```ruby
  def tilde_requirements
    requirements.select { |r| r.first == "~>" }
  end

  def uses_tilde?
    tilde_requirements.any?
  end
```

greysteil:
@hsbt / @bronzdoc this should be good to merge now. Might be worth a 3.0.2 release, since it can cause unexpected Bundler results (e.g., bundler/bundler#6858 and bundler/bundler#6833). As folks upgrade to Ruby 2.6.0 we're likely to see more people run into that, and would be nice to have an easy fix (upgrade rubygems) for them.

hsbt:
I will release RG 3.0.2 with your fix at the next week or the middle of Jan.

hsbt:
@bundlerbot r+

Issue #2522, opened 2018-12-11
Allow gem push from Azure DevOps Release Pipeline


saigkill:
´The standard way for pushing a gem to Rubygems.org is:
1.) Prepare a .gem credentials directory
2.) Do a gem push

My current problem is that i want to use Azure DevOps for the test, build and release process in its pipelines. The build and release process happens in a temporary subdirectory. But it looks like it doesn't possible to place a credentials directory inside the build user home place.

This issue is related to:

  - [X] Publishing a library
  - [X] The command line `gem`

A possible solution could be to add a --apikey attribute to the gem push command. Azure provides a variablegroup with possible hidden variables.
So a `gem push mygem.gem --apikey $(apikey)` could be possible without any security breaks.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
Hi @saigkill 

> But it looks like it doesn't possible to place a credentials directory inside the build user home place.

what makes it not possible to do that? how is that a Rubygems issue?

bronzdoc:
closing due to lack of response

saigkill:
Hi @bronzdoc 
Sorry for the delay. I was sick,
In general it was just a feature request, a idea for providing a new parameter for the gem binary.
There are many CI services what are providing build pipelines. The new parameter would simplify the gem push process in that pipelines.

bronzdoc:
Ah ok @saigkill !

Maybe you can help me to understand why that can't be done with the current state of rubygems for you?

saigkill:
Hello @bronzdoc ,
thank you for your patience and your reopening of this issue. Please forgive my bad english tongue, but i will try to give my best to show my thoughts.
My current workflow is to add a .gem directory in my local user space and put there a yaml file, what contains just the api key for rubygems.org. Then i use "gem push" for pushing the gem to rubygems.org. 
If i'm using any CI with a build and release pipeline i have two options:
1. prepare a .gem directory online and download that to the pipeline and move that to workers home place
2. prepare a script what creates the needed directory and prepare a yaml file and put that to that direcotry.
Depending on the used CI it can be difficult to store a directory inside a workers home directory. In my case specially Azure DevOps allows to provide build and release pipelines, but a modification of workers home directory isn't easy. I also opened a question on Azure Help forums, but it looks, it isn't easy to use workers directory in that case, what i need.
[...]

PR #2559, opened 2018-12-30
Use env var to configure api key on push

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/2522

This PR will allow a user to set an environment var called `HOST_API_KEY` to an `api_key` that would have precedence over the `--key` flag

How differs of the current current `--key` flag of the push command is that this env var will be set to the api token value itself instead of the a key associated to an api token configured in `~/.gem/credentials` 

### DIFF
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -15,6 +15,8 @@ https://rubygems.org) and adds it to the index.
 
 The gem can be removed from the index and deleted from the server using the yank
 command.  For further discussion see the help for the yank command.
+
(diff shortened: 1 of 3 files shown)


hsbt:
Is `HOST_API_KEY` the best variable name?

I'm okay to merge this without naming concern.

bronzdoc:
@hsbt 
> Is HOST_API_KEY the best variable name?

Definitely can be improved,  any ideas of a better name?

I thought of `RUBYGEMS_HOST_API_KEY` at first but that may be confusing, people can think is  the api key for `rubygems.org`

deivid-rodriguez:
`GEM_HOST_API_KEY`?

bronzdoc:
@bundlerbot r+

PR #2560, opened 2018-12-30
Restore SOURCE_DATE_EPOCH

### LINKED ISSUE #2551: Setting SOURCE_DATE_EPOCH env variable breaks several tests

hsbt:
# Description:

Fixes #2551 and Rleated https://github.com/rubygems/rubygems/pull/2552
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -105,6 +105,7 @@ class TestGemPackage < Gem::Package::TarTestCase
   end
 
   def test_build_time_source_date_epoch
+    epoch = ENV["SOURCE_DATE_EPOCH"]
     ENV["SOURCE_DATE_EPOCH"] = "123456789"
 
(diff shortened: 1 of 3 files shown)


hsbt:
@bundlerbot r+

sorah:
I think this changes nothing because RubyGems::TestCase already has environment variable restoration logic in `setup` and `teardown`. cc @nobu

Issue #1706, opened 2016-09-03
https_proxy environment variable is ignored when empty


shaicoleman:
The https_proxy environment variable is ignored when it's empty, and it causes it to download the gem through the http proxy.

```
# this uses the http_proxy environment variable as a proxy,
# when it should not use a proxy as https_proxy is set but empty
http_proxy='http://1.2.3.4:1234' https_proxy='' gem install rake
```

This makes it impossible to use a proxy that only supports HTTP requests, but not with HTTPS (e.g. nginx as a forward proxy)

Environment: RubyGems 2.6.6, Ruby 2.3.1p112, Ubuntu 16.04 x64
I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

segiddins:
This appears to be intentional, see https://github.com/rubygems/rubygems/blob/0a99e91ebd010ddf13bdcaaf569ce07d3b826ad9/lib/rubygems/request.rb#L124

shaicoleman:
if https_proxy is undefined, then it should fallback to http_proxy, but if it's set to empty, it should not use it. That's the way almost all the application work, e.g. curl, apt, wget, etc.

bronzdoc:
@shaicoleman just to be 100% clear, so what you are asking is: if https_proxy is undefined or is set to an empty string, use http_proxy instead?

shaicoleman:
If https_proxy is undefined => fall back to http_proxy (as is currently)
if https_proxy is empty => do not use proxy

PR #2567, opened 2019-01-02
Don't use a proxy if https_proxy env var is empty

bronzdoc:
# Description:

Closes https://github.com/rubygems/rubygems/issues/1706

Do no use a proxy if `https_proxy` env var is set to empty


bronzdoc:
@bundlerbot r+

PR #2568, opened 2019-01-02
Fix intermittent test error on Appveyor & Travis

MSP-Greg:
# Description:

Intermittent test error on Appveyor, which is not restricted to particular Ruby version:

```
  1) Error:
TestGem#test_self_install_permissions_with_format_executable:
Errno::ENOENT: No such file or directory @ rb_file_s_stat - bin/foo.cmd
    C:/projects/rubygems/test/rubygems/test_gem.rb:204:in `stat'
    C:/projects/rubygems/test/rubygems/test_gem.rb:204:in `block (2 levels) in assert_self_install_permissions'
    C:/projects/rubygems/test/rubygems/test_gem.rb:203:in `each_key'
    C:/projects/rubygems/test/rubygems/test_gem.rb:203:in `block in assert_self_install_permissions'
    C:/projects/rubygems/test/rubygems/test_gem.rb:202:in `chdir'
    C:/projects/rubygems/test/rubygems/test_gem.rb:202:in `assert_self_install_permissions'
    C:/projects/rubygems/test/rubygems/test_gem.rb:155:in `test_self_install_permissions_with_format_executable'
```

From being familiar with Appveyor test issues, many of the intermittent errors in trunk parallel testing (which pass on serial retry) occur when doing 'write' then immediate 'reads'.  Probably due to multiple VM's sharing drives.  JFYI, I cannot repo locally.

This may require a longer sleep, only time will tell...

# Tasks:

- [X] Describe the problem / feature
- [ ] Write tests
- [X] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2571, opened 2019-01-03
tests - remove unneeded requires, add frozen string code

MSP-Greg:
# Description:

While working on PR #2568, I noticed that [test_gem.rb](https://github.com/rubygems/rubygems/blob/dbf4fbe50f9898e7c6cb93af6c5cf1a9b084e032/test/rubygems/test_gem.rb) did not have a 'frozen-string' comment.  Also, was wondering about unneeded require statements.  Hence:

1. Updated `test_gem.rb` to work with frozen strings.
2. Removed unneeded require statements from tests, as many autoload in `rubygems.rb`, which is required by `test_case.rb`.
3. Removed code for `Gem::QuickLoader`, which seems to be an eight year old gem with few downloads.
4. Updated minitest requirement to 5.4 from 5.0.

After this update, and using `--enable=frozen-string-literal`, there were only three test errors, all due to rdoc...

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -6,7 +6,6 @@
 #++
 
 require 'optparse'
-require 'rubygems/requirement'
 require 'rubygems/user_interaction'
 
 ##
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -2,16 +2,13 @@
 # TODO: $SAFE = 1
 
 begin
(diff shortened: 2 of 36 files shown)


hsbt:
Thanks, I will merge your changes. But This pull-request mixed multiple changes.

Please separate the 1-4 from this pull-request. It makes difficult to bisect or investigate later

PR #2579, opened 2019-01-05
Test updates - requires, old code, frozen string

MSP-Greg:
# Description:

Misc fixes, mostly for tests.  See commit descriptions.

Note that with the "test_gem.rb - update to use frozen_string_literal: true" commit, when using RUBYOPT="--enable=frozen-string-literal", the test suite completes except for a few issues caused by rdoc.

Commits do not affect each other, but separate PR's might require rebasing depending upon order of acceptance.  As suggested (and a good point), split into separate commits for bisect.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -6,7 +6,6 @@
 #++
 
 require 'optparse'
-require 'rubygems/requirement'
 require 'rubygems/user_interaction'
 
 ##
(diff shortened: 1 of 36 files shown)


hsbt:
Can you split this the multiple changes to each pull-requests?

You always mixed the many of features. It makes the code review to hard work in github.

Issue #2573, opened 2019-01-03
Implement deprecations for CLI options


deivid-rodriguez:
My current problem is that we don't really have a proper deprecation system for CLI options.

For example, we deprecated the `--no-ri` and `--no-rdoc` options [back in 2011](https://github.com/rubygems/rubygems/commit/0d22d6025ce2653ec9fb20a4e0745c79bda677da) (more than 7 years ago!) and we finally removed them very recently with the release of 3.0.0. However, many users didn't know about it and many tutorials have not yet been updated because we didn't properly inform our users.

The only thing we have in place is a "Deprecated options" group in the `gem <command> --help` text. For example, from the current `gem install --help` text

```
Usage: gem install GEMNAME [GEMNAME ...] [options] -- --build-flags [options]

  Options:
        --platform PLATFORM          Specify the platform of gem to install
    -v, --version VERSION            Specify version of gem to install
        --[no-]prerelease            Allow prerelease versions of a gem
                                     to be installed. (Only for listed gems)


  Deprecated Options:
    -u, --[no-]update-sources        Update local source cache

...
```

But that doesn't seem enough. We need standard way to mark options as deprecated so that the keep working as before but print a deprecation error, so that it informs users about the future deprecation, and encourages them to use the new preferred alternative.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
3.0.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

hsbt:
@deivid-rodriguez I prefer your idea instead of the hard corded approach like https://github.com/rubygems/rubygems/pull/2572

deivid-rodriguez:
@hsbt The approach in #2572 is specific to circumvent the removal of these options without a deprecation cycle. Another idea could be to completely revert the removal. I think that would be fine too.

What I mean is that implementing this idea does not really solve the problem for the `--no-ri` and `--no-rdoc` options, because they have already been removed.

PR #2607, opened 2019-01-23
Add deprecation warnings for cli options

bronzdoc:
# Description:
Closes https://github.com/rubygems/rubygems/issues/2573

This PR introduce deprecation warnings for cli options, now if we want to deprecate a CLI option, we call the `Gem::Command#deprecate_option`

` Gem::Command#deprecate_option` has three named parameters `short_name` and `long_name` and `version`, if a cli option has a short and a long version of it, you'll have to explicitly pass both options to `deprecate_option` to be consistent with the warning for the same option.

The `version` option will allow to set the Rubygems version for the warning message in which the option will be deprecated. 

 e.g `deprecate_option(short_name: "-f", long_name: "--force ", "3.1")` 

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -352,7 +352,7 @@ class Gem::Command
   # options in output.  See `gem help list` for an example.
 
   def add_option(*opts, &handler) # :yields: value, options
-    group_name = Symbol == opts.first.class ? opts.shift : :options
+    group_name = Symbol === opts.first ? opts.shift : :options
 
     @option_groups[group_name] << [opts, handler]
   end


hsbt:
Awesome. It's very useful.

lib/rubygems/command.rb
```diff
@@ -365,6 +366,18 @@ def remove_option(name)
     end
   end
 
+  def deprecate_option(short_name: nil, long_name: nil)
```
hsbt: Can you add a option about deprecated version like this

```
deprecate_option(long_name: "--no-ri", version: 3.1)
```

We can show with above method:

```
The \"#{option}\" option has been deprecated and will be removed in future versions of Rubygems 3.1, its use is discouraged."
```

lib/rubygems/command.rb
```diff
@@ -351,7 +352,7 @@ def when_invoked(&block)
   # options in output.  See `gem help list` for an example.
 
   def add_option(*opts, &handler) # :yields: value, options
-    group_name = Symbol === opts.first ? opts.shift : :options
+    group_name = Symbol == opts.first.class ? opts.shift : :options
```
deivid-rodriguez: This is only a stylistic change, right? I tried enabling rubocop's [`Style/CaseEquality`](https://rubocop.readthedocs.io/en/latest/cops_style/#stylecaseequality) cop, and we currently have 109 offenses in rubygems code base, so maybe we can remove the change from this PR and discuss enabling that cop separately.

test/rubygems/test_gem_command.rb
```diff
@@ -195,6 +195,64 @@ def test_option_recognition
     assert_equal ['-h', 'command'], args
   end
 
+  def test_depreacate_option_long_name
```
deivid-rodriguez: s/depreacte/deprecate/

lib/rubygems/command.rb
```diff
@@ -365,6 +366,22 @@ def remove_option(name)
...
+  def deprecate_option(short_name: nil, long_name: nil, version: nil)
+    @deprecated_options[command].merge!({ short_name => { "rg_version_to_expire" => version } }) if short_name
+    @deprecated_options[command].merge!({ long_name  => { "rg_version_to_expire" => version } }) if long_name
+  end
+
+  def check_deprecated_options(options)
+    options.each do |option|
+      if option_is_deprecated?(option)
+        version_to_expire = @deprecated_options[command][option]["rg_version_to_expire"]
+        deprecate_option_msg = "The \"#{option}\" option has been deprecated and will be removed in future versions of Rubygems #{version_to_expire}, its use is discouraged."
```
deivid-rodriguez: Perhaps

```ruby
deprecate_option_msg = if version_to_expire
                         "The \"#{option}\" option has been deprecated and will be removed in Rubygems #{version_to_expire}, its use is discouraged."
                       else
                         "The \"#{option}\" option has been deprecated and will be removed in future versions of Rubygems, its use is discouraged."
                       end
```

lib/rubygems/command.rb
bronzdoc: Sounds good 👍

hsbt:
@bundlerbot r+

lib/rubygems/command.rb
```diff
@@ -365,6 +366,28 @@ def remove_option(name)
...
+  def deprecate_option(short_name: nil, long_name: nil, version: nil)
+    @deprecated_options[command].merge!({ short_name => { "rg_version_to_expire" => version } }) if short_name
+    @deprecated_options[command].merge!({ long_name  => { "rg_version_to_expire" => version } }) if long_name
+  end
+
+  def check_deprecated_options(options)
+    options.each do |option|
+      if option_is_deprecated?(option)
+        version_to_expire = @deprecated_options[command][option]["rg_version_to_expire"]
+        #deprecate_option_msg = "The \"#{option}\" option has been deprecated and will be removed in future versions of Rubygems #{version_to_expire}, its use is discouraged."
```
segiddins: we shouldn't be shipping commented-out code

bronzdoc: Agree, sorry about that! will fix

Issue #2580, opened 2019-01-06
Introduce `gem build .` and `gem install .` 


LemonAndroid:
I'm having a problem or would like to suggest a feature.

My current problem is that it is cumbersome to retype the name of the gemspec file when building and vice-versa the .gem file when installing.

I  propose that the `build` and `install` commands also work for folder or at least for the current directory `.` by:

Bundling: Looking for the `.gemspec` file.
Installing: Choosing `.gem` file in directory that matches `.gemspec` version number or building beforehand if it's not existing.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x ] The command line `gem`
  - [ ] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

djberg96:
True, I can't think of a situation in local development where I would have more than one gemspec in the current directory. I would even say just skip the "." and build or install the first gemspec found.

voxik:
I don't mind if `gem build` pick some `.gemspec` file from current directory, either the only one which is present, the first one, the one which matches the directory name or it can fail if there is more then one `.gemspec` files, but @djberg96 please be careful with statements like "I can't think of a situation in local development where I would have more than one gemspec in the current directory". IMO, there are chances that you are wrong. Moreover, if #2587 is not resolved, there are quite high chances that we will need to place second .gemspec file into the directory to be able to package gems on Fedora.

djberg96:
@voxik Ok, I was only speaking for myself. But, this would only be a default, you could still specify a file. That, or perhaps it should build every gemspec in the current directory if none is specified.

@LemonAndroid on what the behavior should be if there are multiple gemspecs?

LemonAndroid:
@djberg96 @voxik I think it should just build or install for all gemspecs by default.

LemonAndroid:
Closing, since use-case is not understood.

Issue #2587, opened 2019-01-08
`gem build ../foo.gemspec` does not work anymore


voxik:
This does not work anymore:

~~~
$ ls
COPYING  GPLv2  GPLv3  Gemfile  LICENSE  Rakefile  lib  pdf-core.gemspec
$ ls ..
pdf-core-0.8.1  pdf-core-0.8.1.gem  pdf-core.gemspec
$ gem build ../*.gemspec
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    ["COPYING", "GPLv2", "GPLv3", "Gemfile", "LICENSE", "Rakefile"] are not files
~~~

The problem is that we are for some time using this approach in Fedora for building our packages [[1]]. Please consider reverting #2204, because it broke the functionality which was in RubyGems for years while there is not provided any alternative (the previous behavior was possible to workaround by simple `cd` command).

This was originally reported here [[2]] by @pvalena

[1]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Ruby/#_building_gems
[2]: https://bugs.ruby-lang.org/issues/15516

hsbt:
@voxik Caution: I understand you and fedora problem. Do not blame the contributors and maintainers.

voxik:
Apologies, it was not intentional. I tried to reword the original description a bit. I hope it is better now.

deivid-rodriguez:
My proposed fix for this would be:

* Restore the previous behavior.
* Add `-C` argument to `gem build` that behaves similarly to `git -C`. If given, files are picked up from the given directory, and also the gemspec is resolved from that directory if given relatively.

voxik:
> My proposed fix for this would be:
>     * Restore the previous behavior.
>     * Add `-C` argument to `gem build` that behaves similarly to `git -C`.

And similarly to `make -C`. That would be nice.

hsbt:
I'm +1 to https://github.com/rubygems/rubygems/issues/2587#issuecomment-452616438 .

Issue #448, opened 2013-02-08
$ gem install should look for dependencies in the same directory as the dependant gem


Maniacal:
If I want to install a.gem and it depends on b.gem and I want to install them from local files(`/tmp/gemstage/a.gem` and `/tmp/gemstage/b.gem`), I should be able to install them as follows:

```
gem install /tmp/gemstage/a.gem --local
```

But this will fail because it won't know where to look for b.gem.  I can cd to `/tmp/gemstage/` and then run:

```
gem install a.gem --local
```

It would be a nice feature if gem install would look for dependencies in the same folder as the gem you're installing. That way you won't have to be concerned with changing directories when scripting these installations.

copiousfreetime:
To summarize this issue:

Currently rubygems will look in the current directory for additional gems to install when you do a gem installation For example:
```
% ls -l
total 2448
-rw-r--r--  1 jeremy  staff  102400 Jan 29 14:18 addressable-2.4.0.gem
-rw-r--r--  1 jeremy  staff  220672 Jan 29 14:18 excon-0.45.4.gem
-rw-r--r--  1 jeremy  staff   31744 Jan 29 14:18 hitimes-1.2.3.gem
-rw-r--r--  1 jeremy  staff   10240 Jan 29 14:18 little-plugger-1.1.4.gem
-rw-r--r--  1 jeremy  staff   88064 Jan 29 14:18 logging-1.8.2.gem
-rw-r--r--  1 jeremy  staff   25088 Jan 29 14:18 multi_json-1.11.2.gem
-rw-r--r--  1 jeremy  staff  228864 Jan 29 14:18 rack-1.6.4.gem
-rw-r--r--  1 jeremy  staff   18432 Jan 29 14:18 rack-protection-1.5.3.gem
-rw-r--r--  1 jeremy  staff  369664 Jan 29 14:18 sinatra-1.4.7.gem
-rw-r--r--  1 jeremy  staff   65536 Jan 29 14:18 stickler-2.4.2.gem
-rw-r--r--  1 jeremy  staff   48640 Jan 29 14:18 tilt-2.0.2.gem
-rw-r--r--  1 jeremy  staff   26112 Jan 29 14:18 trollop-2.1.2.gem
 % gem install --local stickler-2.4.2.gem 
Successfully installed little-plugger-1.1.4
Successfully installed logging-1.8.2
Successfully installed trollop-2.1.2
Successfully installed excon-0.45.4
Successfully installed addressable-2.4.0
[...]

PR #2610, opened 2019-01-29
Introduce dot for install and build

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2580

This PR introduce the use of the `.` comand line argument to use when building, installing or uninstalling gems in the current directory.

The `.` argument will build all `.gemspec` files when building i.e `gem build .` and it will install/uninstall all the `.gem` files when installing or uninstalling i.e `gem install .` or `gem uninstall .`

### DIFF
--- a/test/rubygems/test_gem_commands_build_command.rb
+++ b/test/rubygems/test_gem_commands_build_command.rb
@@ -189,7 +189,7 @@ class TestGemCommandsBuildCommand < Gem::TestCase
     end
 
     assert_equal '', @ui.output
-    assert_equal "ERROR:  Gemspec file not found: some_gem\n", @ui.error
+    assert_equal "ERROR:  Gemspec file not found: some_gem.gemspec\n", @ui.error
   end
 
   def test_execute_outside_dir
@@ -228,6 +228,90 @@ class TestGemCommandsBuildCommand < Gem::TestCase
 
     assert_equal "some_gem", spec.name
     assert_equal "this is a summary", spec.summary
+  end
+
+  def test_execute_current_dir
+    gem_0 = util_spec 'some_gem' do |s|
+      s.license = 'AGPL-3.0'
+      s.files = ['README.md']
+    end
+
+    gem_1 = util_spec 'another_gem' do |s|
+      s.license = 'AGPL-3.0'
+      s.files = ['README.md']
+    end
+
+    gemspec_dir = File.join @tempdir, 'build_command_gem'
+
+    gemspec_file_0 = File.join(gemspec_dir, gem_0.spec_name)
+    readme_file_0  = File.join(gemspec_dir, 'README.md')
+
+    gemspec_file_1 = File.join(gemspec_dir, gem_1.spec_name)
+    readme_file_1  = File.join(gemspec_dir, 'README.md')
+
+    FileUtils.mkdir_p(gemspec_dir)
+
+    File.open(readme_file_0, 'w') do |f|
+      f.write "My awesome gem"
+    end
+
+    File.open(readme_file_1, 'w') do |f|
+      f.write "My awesome gem"
+    end
+
+    File.open(gemspec_file_0, 'w') do |gs|
+      gs.write gem_0.to_ruby
+    end
+
+    File.open(gemspec_file_1, 'w') do |gs|
+      gs.write gem_1.to_ruby
+    end
+
+    @cmd.options[:args] = ["."]
+
+    use_ui @ui do
+      Dir.chdir(gemspec_dir) do
+        @cmd.execute
+      end
+    end
+
+    output = @ui.output.split "\n"
+    assert_equal "  Successfully built RubyGem", output.shift
+    assert_equal "  Name: another_gem", output.shift
+    assert_equal "  Version: 2", output.shift
+    assert_equal "  File: another_gem-2.gem", output.shift
+    assert_equal "  Successfully built RubyGem", output.shift
+    assert_equal "  Name: some_gem", output.shift
+    assert_equal "  Version: 2", output.shift
+    assert_equal "  File: some_gem-2.gem", output.shift
+    assert_equal [], output
+
+    gem_file_0 = File.join(gemspec_dir, File.basename(gem_0.cache_file))
+    assert File.exist?(gem_file_0)
+
+    gem_file_1 = File.join(gemspec_dir, File.basename(gem_1.cache_file))
+    assert File.exist?(gem_file_1)
+  end
+
+  def test_execute_current_dir_no_gemspecs
+    gemspec_dir = File.join @tempdir, 'build_command_gem'
+    FileUtils.mkdir_p(gemspec_dir)
+
+    @cmd.options[:args] = ["."]
+
+    use_ui @ui do
+      Dir.chdir(gemspec_dir) do
+        assert_raises Gem::MockGemUi::TermError do
+          @cmd.execute
+        end
+      end
+    end
+
+    assert_equal "", @ui.output
+
+    Dir.chdir(gemspec_dir) do
+      assert_equal "ERROR:  Gemspec files not found: #{Dir.pwd}\n", @ui.error
+    end
   end
 
   def test_can_find_gemspecs_without_dot_gemspec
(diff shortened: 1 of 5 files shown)


MSP-Greg:
If one has gems a, b, c, & d in the 'dot' directory, and d is a dependency of a, will all four gems be installed?  Or, does `gem install .` look to rubygems.org for d?

If d is not installed from '.', should dependencies first be looked for in the 'dot' directory?

bronzdoc:
@MSP-Greg if one has gems a, b, c in the 'dot' directory, and you want to install them as of now you need to to `gem install` 3 times, this will make you just have to do `gem install` one time, i'm not changing how `gem install` works.

MSP-Greg:
@bronzdoc

Yes, I understand the change.  My question involves how dependencies of a, b, c are handled, in terms of whether they are also installed from the 'dot' folder, assuming they exist there...

MSP-Greg:
@bronzdoc

Never mind.  An edge case like this leads to a deep rabbit hole without a non-trival solution.  If a dependency in dot exists, one should use the `-​-ignore-dependencies` option...

bronzdoc:
@MSP-Greg see https://github.com/rubygems/rubygems/issues/448

MSP-Greg:
@bronzdoc 

Thanks for the issue link.  I like the idea in @deivid-rodriguez's post: https://github.com/rubygems/rubygems/issues/448#issuecomment-364673650.  I can't look at it until later, does this PR accomplish that also?

bronzdoc:
@MSP-Greg 
>  does this PR accomplish that also?

nope this is not changing `gem install` behavior :)

MSP-Greg:
@bronzdoc 

> nope this is not changing gem install behavior :)

Ok.  

1. I haven't ever needed to install more that one local gem, so I've never checked to see if one could load several from a hard code path parameter.

2. If RG `gem install` didn't previously respond to `gem install .`, then previous behavior has not been defined, so we're free to do whatever.

3. If that behavior differs from similar options (like the first point), then that is something that should be discussed.

Kind of like `gem update --explain`  :)

deivid-rodriguez:
Having several gemspecs in the same directory seems... a really weird edge case. I'm not sure whether we should support that, since nobody has requested it? @voxik said he _might_ need to do that if https://github.com/rubygems/rubygems/issues/2587 was not resolved, but #2587 has been resolved.

I think I would prefer for now to build the single gemspec in the directory and fail with a message telling the user to be explicit if there are actually multiple gemspecs?

bronzdoc:
@deivid-rodriguez 

@LemonAndroid requested it
> @djberg96 @voxik I think it should just build or install for all gemspecs by default.

@LemonAndroid can you be more explicit why would you need to build all gemspecs, trying to understand the edge case and see if it is worth keeping this PR the way it is or modify it with what @deivid-rodriguez suggested, thanks!

test/rubygems/test_gem_commands_build_command.rb
```diff
@@ -230,6 +230,90 @@ def test_execute_outside_dir
...
+  def test_execute_current_dir
+    gem_0 = util_spec 'some_gem' do |s|
+      s.license = 'AGPL-3.0'
+      s.files = ['README.md']
+    end
+
+    gem_1 = util_spec 'another_gem' do |s|
+      s.license = 'AGPL-3.0'
+      s.files = ['README.md']
+    end
```
deivid-rodriguez: My fear is that supporting this might cause confusion and it's hard to think of cases where it's useful. For example, in this test we're defining two different gems that have the same README... right? :thinking:

deivid-rodriguez:
Thanks @bronzdoc, I missed that sentence when reading the issue. :+1:

deivid-rodriguez:
The thing is... Having multiple gemspecs in the same folder is... weird to say the least, because either `spec.files` in all gemspecs will be pointing to the same files (but why are the gems diffferent then?), or `spec.files` will list disjoint sets of files (but in that case, wouldn't moving each gem to a subfolder be way clearer?).

I think we should at least have a realworld example of this before we consider supporting it.

LemonAndroid:
> The thing is... Having multiple gemspecs in the same folder is... weird to say the least, because either `spec.files` in all gemspecs will be pointing to the same files (but why are the gems diffferent then?), or `spec.files` will list disjoint sets of files (but in that case, wouldn't moving each gem to a subfolder be way clearer?).
> 
> I think we should at least have a realworld example of this before we consider supporting it.

There could be multiple folders + the .gemspec file. Each folder containing the files related to one .gemspec file.

deivid-rodriguez:
Aha, that makes sense and I have seen that case, but in my case this wouldn't work because all gems depend on each other, so the "isolated installs" would fail. I would need something like https://github.com/rubygems/rubygems/issues/448 instead.

In any case, I don't think that's what this PR is implementing? I think it's only look for gems in the root folder...

LemonAndroid:
> #448

It should basically get all gemspecs from a folder + sub folders and resolve dependencies automatically?

deivid-rodriguez:
Yes, that's what I requested there because I've seen situations where I think it could be useful. For example, https://github.com/rails/rails or https://github.com/decidim/decidim.

bronzdoc:
Ok, for now lets keep this simple, since https://github.com/rubygems/rubygems/issues/448 is not yet implemented and there's no other real world usage example,  i'll change this PR to what @deivid-rodriguez suggested.

Any objection on the way `gem install .` will work?

LemonAndroid:
Nope

voxik:
From top of my head, here are several examples of libraries with more than one .gemspec in directory:

https://github.com/ruby-concurrency/concurrent-ruby
https://github.com/flori/json/

I am pretty sure there is more. Typically it is useful for gems which are plain Ruby, but they can benefit from binary extension as well. I can imagine, that there might be cases of multiple .gemspec in one repository for better multiplatform support, but I don't have example at my hand. And the concurrent-ruby example is definitely very interesting.

And my use case is a bit specific. If #2587 was not resolve, we would need to use workflow like this [[1]]

~~~
$ gem unpack %{SOURCE0}
$ gem spec %{SOURCE0} -l --ruby > %{gem_name}.gemspec

$ cat some.patch | patch -p1
$ sed -i "s/modify/something/" somewhere/in/the/package

$ gem build %{gem_name}.gemspec
~~~

But there is quite a lot of gems shipping their upstream .gemspec, which might collide with the generated `%{gem_name}.gemspec`.

Luckily, because I can use `gem build ../foo.gemspec` again, I don't have to place two .gemspec files into one directory for my purposes and we follow a bit different set of guidelines [[2]].

[1]: https://fedoraproject.org/w/index.php?title=Packaging:Ruby&oldid=400800#Building_gems
[2]: https://docs.fedoraproject.org/en-US/packaging-guidelines/Ruby/#_building_gems

deivid-rodriguez:
Thanks @voxik for the explanation! So this is an improvement you would welcome, then? If this can be useful for gems out there, I guess we can ship it?

We could improve the patch in the future by supporting arbitrary folder instead of only ".", but this is a good start :+1:.

voxik:
I guess that `gem install some/path/*.gem` already works just fine if I am not mistaken. `gem install .` is ambiguous and magical to me, not mentioning that this PR add it just as special case for current directory.

It seems that `gem build *.gemspec` is not supported, so this could be probably nice extension, but I have never had any need for this myself.

If you are talking about cases such as Rails repository, I still think there is nice place to use shell `find`, but if you want to really support something like recursive lookup, then there should be probably some `-R` flag, which is used by `grep`, `chown` and similar.
 
Frankly, reading the #2580, I think it should be rejected as it is and this should be started over with refined scope.

deivid-rodriguez:
@voxik I agree with what you say, I think adding support for `gem build *.gemspec` satisfies the original request quite well and it's very intuitive. I support we do that instead :+1:.

bronzdoc:
I'm happy to work on this if we can all agree in the new scope :)

deivid-rodriguez:
Hahaha, right, I'm sorry @bronzdoc, but @voxik raised good points and made me change my mind.

I think the most intuitive addition is indeed supporting multiple arguments to `gem install` and `gem build`, and looping over them the same way this PR is doing over the files in the current directory.

PR #2615, opened 2019-01-31
test_gem.rb fixups & installer.rb one liner 

MSP-Greg:
# Description:

Test failure was due to random testing.  See PR #2613  d6b61c6, PR #2605  e7ab095, PR #2568  aeb353f

Failure involved setting the `format_executable` install option.  This option is uses `Gem::Installer.exec_format`, which is modified by other tests, but not reset.  Handling it correctly appears to have stopped the intermittent failures.

While debugging, the Gem::Installer issue appeared.  Windows installs have two scripts for every 'CLI gem'.  One is a cmd/bat file, which is considered an 'exe' by Windows, the other is a 'bash' style script.  From a pure Windows perspective, only the cmd/bat file is needed.  But, many Windows users have MSYS2/MinGW tools installed, and probably 'Git for Windows'.  Both have 'bash shells' available.  Also, I've seen more than a few CI bash scripts used in Appveyor testing, which allows repo's to use the same scripts for testing both on Appveyor & Travis.

Long story short, the issue with Gem::Installer is that it was using the file suffix added by `format_executable` in the bash script file name, but not the cmd/bat file name.  Fixed that and also added it to `test_self_install_permissions_with_format_executable` in `test_gem.rb`...

# Tasks:

- [X] Update tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -467,7 +467,7 @@ class Gem::Installer
 
   def generate_windows_script(filename, bindir)
     if Gem.win_platform?
-      script_name = filename + ".bat"
+      script_name = formatted_program_filename(filename) + ".bat"
       script_path = File.join bindir, File.basename(script_name)
       File.open script_path, 'w' do |file|
         file.puts windows_stub_script(bindir, filename)
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -14,8 +14,6 @@ $LOAD_PATH.map! do |path|
 end
 
 class TestGem < Gem::TestCase
-  RUBY_INSTALL_NAME = RbConfig::CONFIG['RUBY_INSTALL_NAME']
-
   PLUGINS_LOADED = [] # rubocop:disable Style/MutableConstant
 
   def setup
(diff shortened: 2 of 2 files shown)


MSP-Greg:
I added some debug output (lines 199 - 204) to this PR, only triggered if the intermittent test will fail.

See failures on [Travis](https://travis-ci.org/rubygems/rubygems/jobs/487013298#L674) & [Appveyor](https://ci.appveyor.com/project/rubygems/rubygems/builds/22030638/job/uxi2ayvj94e6g1y6#L27).

In both instances, the 'expected' key is `bin/foo.cmd`, but the file was written as `foo-foo.cmd-bar`.  I suspect this error is caused by another test creating an 'unexpected' environment, as previously the tests were not run random.

Does anyone have any idea as to the cause/bad test?  I assume it has something todo with `RUBY_INSTALL_NAME`...

deivid-rodriguez:
I reviewed this PR, @MSP-Greg, it looks good in general and the bug fix seems correct to me!

I had quite a few comments however, because there seemed to be a lot of different changes mixed together (formatting, simplifications, bug fix, state resetting). Since I had already coded my suggestions (commit splitting mostly, so the rationale for each change can be understood), I decided to open a separate reworked PR instead of commenting here. It's #2628. I would close this PR in favor of that one, if you are ok with that.

MSP-Greg:
@deivid-rodriguez

Thanks.  I'm ok with that.

> DRY ups that I consider that harm readability instead of adding to it (for example, moving the umask stuff inside the already complicated method `assert_self_install_permissions`).

Sorry. 2nd nature...

PR #2616, opened 2019-02-01
Add platform ranking method, fix --explain issue

MSP-Greg:
# Description:

Platform rankings is currently done in several places, and is a 'binary' rank.  Install/Update will not show selected platform specific gems correctly when used with the `--explain` option.

1. Add platform ranking method (`Gem::Platform.rank`), see 2nd commit, test_gem_platform.rb  additions.
2. Fix above --explain issues.

Two commits, the first changes/adds code and passes existing tests, the second adds tests. Additional discussion in PR #2446, #2449, and #2452.

# Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -201,10 +201,13 @@ You can use `i` command instead of `install`.
       request_set = inst.resolve_dependencies name, req
 
       if options[:explain]
-        puts "Gems to install:"
+        say "Gems to install:"
 
         request_set.sorted_requests.each do |s|
-          puts "  #{s.full_name}"
+          # shows platform specific gems if used
+          say (plat = s.spec.platform) == Gem::Platform::RUBY ?
+            "  #{s.full_name}" :
+            "  #{s.full_name}-#{plat}"
         end
 
         return
@@ -234,6 +237,20 @@ You can use `i` command instead of `install`.
       dependency = Gem::Dependency.new name, req
       dependency.prerelease = options[:prerelease]
 
+      if options[:explain]
+        say "Gems to install:"
+        ary = Gem::SpecFetcher.fetcher.search_for_dependency dependency
+        if ary.is_a?(Array) && ary[0].is_a?(Array)
+          tuples = ary[0].map { |i| i[0] }
+          tuple = tuples.max_by { |s|
+            [s.version, Gem::Platform.rank(s.platform)]
+          }
+          say "  #{tuple.full_name}"
+        else
+          say "  no matching gems found"
+        end
+        return
+      end
       fetcher = Gem::RemoteFetcher.fetcher
       gem = fetcher.download_to_cache dependency
     end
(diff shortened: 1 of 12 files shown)


deivid-rodriguez:
Thanks @MSP-Greg, I'll now review this PR so you can submit the next one!

MSP-Greg:
@deivid-rodriguez

Thanks again.  There are more places that `Gem::Platform.rank` should be used, I think I've picked all of then up in the `required_ruby_version` code...

MSP-Greg:
@deivid-rodriguez 

I don't know how far you've gotten into a review, but there is a bit of overlap with the `required_ruby_version` PR.  I also see that I could have used a few more commits to clarify/separate things.

Hence, should I close this, and combine the two into one PR, with several commits?

For instance, `command/install_command.rb` in the following has 'explain code' for `!options[:ignore_dependencies]`, but the `install_gem_without_dependencies` method does not.  I added 'explain code' in this PR, but the code has a minor change to work with `required_ruby_version`...

https://github.com/rubygems/rubygems/blob/a36c3d8c62bf93df386c56b9e00adb20de02e01a/lib/rubygems/commands/install_command.rb#L197-L216

MSP-Greg:
A question to all the maintainers:

Background: adding `Gem::Platform.rank` - several places in the code there are rankings (used in sorting) like below:
```
Gem::Platform.local =~ spec.platform ? 1 : 0
other.platform == Gem::Platform::RUBY ? -1 : 1
```

Gems can have both a platform 'specific' release and a platform 'generic' release.  Hence, the above rankings are non-deterministic, while `Gem::Platform.rank` ranks 'specific' above 'generic'.  Also, a centralized common method makes changes in the future easier.

On the topic of sorting, there several places where tuples/specs are sorted, but the array/set members all have a common name.  So, rather than using the `[name, version, platform]` that's available in `<=>` with NameTuple, Specification, etc, they have a `sort_by` using `[version, platform]`.

**Question:** Which if the below three options is preferred?
1. Leave the sort_by `[version, platform]`
2. Use the existing sort, which is sort of wasteful, as the the 'name' key is not needed.
3. Define something like a `sort_vers_plat` method that would do the same and also reverse the result, as the result is always reversed and used with a `find` method.

deivid-rodriguez:
@MSP-Greg So, I was reviewing this PR but it tries to fix several different issues, namely:

* Issues with the `--explain` option in both `gem install` and `gem update`.
* An issue with platform specific gems not being installed when the `--ignore-dependencies` flag is used.
* Normalize how we sort gems in the whole code base.

I prefer to kill one issue at a time, so I started by creating #2631. This uses the same approach as #2446, which only changes how the `install` command works and doesn't require to introduce global changes to how we rank gems in the whole code base, so it felt less invasive while also fixing the problem.

MSP-Greg:
@deivid-rodriguez

Thanks.

MSP-Greg:
@deivid-rodriguez

> doesn't require to introduce global changes to how we rank gems in the whole code base

So if a gem is published with both an exact match with `Gem::Platform.local` and a 'universal' gem, which should be installed?  The two 'ranking' algorithms currently used do not make that clear...

PR #2617, opened 2019-02-04
Fix sporadic CI failures

deivid-rodriguez:
# Description:

Reproducible with

```
rake test SEED=13408 TESTOPTS="--name=/test_check_executable_overwrite_format_executable\|test_self_install_permissions_with_format_executable/"
```

The problem was that this single test was not resetting
`Gem::Installer.exec_format` after modifying it.

All credit to @MSP-Greg for discovering this issue. This is also fixed by him in #2615 but this fix is better because instead of fixing the affected test, we fix the real culprit so this cannot leak to other tests in the future.

# Tasks:

- [x] Write tests


MSP-Greg:
@deivid-rodriguez

Good find.  I think both PR's are actually required...

Mine cleaned up repetitive code, and also fixed `test_self_install_permissions_with_format_executable`, as it's a **false positive**.  Without adding a prefix and/or suffix to the exe name, it tests nothing...

deivid-rodriguez:
Ok! Let's get this one merged first, then you can rebase #2615 and remove the `Gem::Installer.exec_format` resetting from it since it's no longer necessary, right?

MSP-Greg:
@deivid-rodriguez

> remove the `Gem::Installer.exec_format` resetting from it since it's no longer necessary, right?

Ah, that was the cause of the original problem, so if a test modifies something, shouldn't it reset it?  Or, I don't believe it should be removed?

MSP-Greg:
@deivid-rodriguez

Wondering whether this setting should be moved to the Gem::TestCase setup/teardown,  I briefly looked to see if the was a common item, and decided it wasn't.  Hence, any test changing it should reset it...

deivid-rodriguez:
Only six tests are changing this. I think it's fine that it's reset by each of them instead of doing it for every single test.

deivid-rodriguez:
> Ah, that was the cause of the original problem, so if a test modifies something, shouldn't it reset it? Or, I don't believe it should be removed?

Oh, sorry, then I need to check your PR then. I thought the failing spec was not changing `Gem::Installer.exec_format`, I thought it was just using a modified `Gem::Installer.exec_format` leak by another test (the one I'm fixing here).

MSP-Greg:
It's reset then changed via the change to `RbConfig::CONFIG['ruby_install_name']`, which adds a suffix so the test is a 'true positive'...

deivid-rodriguez:
I'm re-checking the code now and I think I see what you mean. I'll have a look at your PR.

MSP-Greg:
> I'm re-checking the code now and I think I see what you mean. I'll have a look at your PR.

Thank you.  Really.

RG is old code, there's certainly been a great deal of 'feature creep' (resulting in a convoluted API), and   there's also a few fundamental problems that have existed for a long time.

Hence, I'm often frustrated working in it.  That frustration sometimes causes me to not be particularly clear in my explanations...

deivid-rodriguez:
@bronzdoc r?

PR #2621, opened 2019-02-05
Downgrade bundler mismatches to a warning

### LINKED ISSUE #2592: bundle command looks for bundler 2 when not present on the system
When using rubygems 3.0.2, an error will be raised when trying to use a Gemfile.lock generated with Bundler 2.0.1 when no 2.x version is present. It was my understanding that on systems with only a bundler 1.x that the `BUNDLE WITH` was purely cosmetic and would not cause any errors.

To reproduce, first get a docker image with Ruby 2.5:

```
$ docker run -it --rm ruby:2.5 bash
```

Then make sure you're using the latest rubygems, and then generate a Gemfile.lock with a 2.0.1 `BUNDLE WITH`. Next remove that version of bundler and try to run `bundle` with a 1.x version of bundler:

```
gem update --system
gem -v 
ruby -v 
cd /tmp && mkdir rubygems-issue &&  cd rubygems-issue/
gem install bundler -v 2.0.1
touch Gemfile && bundle install
gem uninstall bundler -a -x
gem install bundler -v 1.17.3
bundle -v
```

You'll then get this error:

```
  2: from /Users/rschneeman/.gem/ruby/2.5.3/bin/bundle:23:in `<main>'
  1: from /Users/rschneeman/.rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems.rb:302:in `activate_bin_path'
/Users/rschneeman/.rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.1) required by your /private/tmp/rubygems-issue/Gemfile.lock. (Gem::GemNotFoundException)
To install the missing version, run `gem install bundler:2.0.1`
```

This behavior means that anyone who is using Ruby 2.6 and has Bundler 2.x in their Gemfile.lock will get an error on Heroku, or any other cloud provider where only one version of bundler is currently available, will get an error.

deivid-rodriguez:
# Description:

closes #2592.

The behavior described in that issue is expected, we really prefer the BUNDLED_WITH version in the lock file to be used when running a bundled application.

However, at the current time, it is too much to fail because of mismatches here, so I propose to downgrade this to a warning for the time being.

The strategy I followed is to, instead of filtering out bundler versions not matching the major version of the "BUNDLED WITH" bundler, we we put a version matching it (if it exists) in front of the matching specs, so that is the one to be picked up. Otherwise a warning is raised. By not removing any versions, we make sure we no longer fail because of not finding an appropriate bundler version.

Commit per commit review recommended, since I needed a few refactorings before implementing the fix.

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem_bundler_version_finder.rb
+++ b/test/rubygems/test_gem_bundler_version_finder.rb
@@ -100,32 +100,32 @@ class TestGemBundlerVersionFinder < Gem::TestCase
     end
   end
 
-  def test_filter
+  def test_prioritize
     versions = %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1]
     specs = versions.map { |v| util_spec("bundler", v) }
 
-    assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_filter_specs(specs)
+    assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
 
     bvf.stub(:bundler_version, v("2.1.1.1")) do
-      assert_equal %w[2 2.a 2.0 2.1.1], util_filter_specs(specs)
+      assert_equal %w[2 1 1.0 1.0.1.1 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
     end
     bvf.stub(:bundler_version, v("1.1.1.1")) do
-      assert_equal %w[1 1.0 1.0.1.1], util_filter_specs(specs)
+      assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
     end
     bvf.stub(:bundler_version, v("1")) do
-      assert_equal %w[1 1.0 1.0.1.1], util_filter_specs(specs)
+      assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
     end
     bvf.stub(:bundler_version, v("2.a")) do
-      assert_equal %w[2 2.a 2.0 2.1.1], util_filter_specs(specs)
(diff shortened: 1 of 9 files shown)


test/rubygems/test_gem_bundler_version_finder.rb
```diff
@@ -100,33 +100,33 @@ def test_compatible
...
-      assert_equal %w[1 1.0 1.0.1.1], util_filter_specs(specs).map(&:version).map(&:to_s)
+      assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
     end
     bvf.stub(:bundler_version, v("1")) do
-      assert_equal %w[1 1.0 1.0.1.1], util_filter_specs(specs).map(&:version).map(&:to_s)
+      assert_equal %w[1 1.0 1.0.1.1 2 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
     end
     bvf.stub(:bundler_version, v("2.a")) do
-      assert_equal %w[2 2.a 2.0 2.1.1], util_filter_specs(specs).map(&:version).map(&:to_s)
+      assert_equal %w[2 1 1.0 1.0.1.1 2.a 2.0 2.1.1 3 3.a 3.0 3.1.1], util_prioritize_specs(specs)
```
indirect: Shouldn't `2.a` try `2`, but then try `2.a` before falling back to `1`?

indirect: Oh, I see... it is that `2` is available, so it's moved to the front. And if `2` was not available, `2.a` would be moved to the front instead. Is that right?

deivid-rodriguez: That's exactly right.

The new strategy would be to find the first compatible version and move it to the front, but leave the array untouched if no compatible version is found. The difference with the old strategy is that no bundler versions are ever filtered out, so the array will never be left empty, and thus we'll never cause an error.

We could do further prioritization here as you suggested in the first comment, but only the first spec in this array gets actually used, so it seemed unnecessary.

indirect:
Got it. Sounds good to me!

hsbt:
I also agreed with this strategy.

hsbt:
@bundlerbot r+

hsbt:
@deivid-rodriguez Can you restore this pull-request for the future plan?

deivid-rodriguez:
@hsbt Sure! Even after restoring the branch, I wouldn't be given the option to reopen this pull request, so I opened a separate one to follow up: #2696.

PR #2625, opened 2019-02-05
Fix gem indexer tests leaking utility gems

deivid-rodriguez:
# Description:

The problem is that `SEED=26796 rake TESTOPTS="--name=/test_generate_index_modern_back_to_back\|test_doctor_dry_run/"` fails on master with

```
$ SEED=26796 rake TESTOPTS="--name=/test_generate_index_modern_back_to_back\|test_doctor_dry_run/"
Run options: "--name=/test_generate_index_modern_back_to_back|test_doctor_dry_run/" --seed 26796

# Running:

.F

Finished in 0.163319s, 12.2460 runs/s, 128.5826 assertions/s.

  1) Failure:
TestGemDoctor#test_doctor_dry_run [/home/deivid/Code/rubygems/test/rubygems/test_gem_doctor.rb:118]:
--- expected
+++ actual
@@ -6,5 +6,11 @@
 " +
 "Extra directory gems/c-2
 " +
+"Extra directory gems/d-2.0
+" +
+"Extra directory gems/d-2.0.a
+" +
+"Extra directory gems/d-2.0.b
+" +
 "
 "


2 runs, 21 assertions, 1 failures, 0 errors, 0 skips
rake aborted!
Command failed with status (1)

Tasks: TOP => default => test
(See full trace by running task with --trace)
```

This PR fixes that.

Fixes #2623.

# Tasks:

- [x] Describe the problem / feature
- [x] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2626, opened 2019-02-05
Allow to easily bisect flaky failures

deivid-rodriguez:
# Description:

I used this today to bisect some flaky test failures. I thought I should share it in case other rubygems devs find it useful.

For example,

```
SEED=26796 rake
```

currently fails with the error described in #2625.

Running

```
SEED=26796 rake bisect
```

should give you the minimal reproduction for it after a bit of investigation, namely,

```
SEED=26796 rake TESTOPTS="--name=/test_generate_index_modern_back_to_back\|test_doctor_dry_run/"
```

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -71,6 +71,17 @@ task(:rubocop) do
   sh "util/rubocop"
 end
 
+desc "Run a test suite bisection"
+task(:bisect) do
+  seed = begin
+           Integer(ENV["SEED"])
+         rescue
+           abort "Specify the failing seed as the SEED environment variable"
+         end
+
+  sh "SEED=#{seed} MTB_VERBOSE=2 util/bisect -Ilib:bundler/lib:test test"
+end
+
 # --------------------------------------------------------------------
 # Creating a release
 
--- a/Rakefile
+++ b/Rakefile
@@ -79,7 +79,8 @@ task(:bisect) do
            abort "Specify the failing seed as the SEED environment variable"
          end
 
-  sh "SEED=#{seed} MTB_VERBOSE=2 util/bisect -Ilib:bundler/lib:test test"
+  gemdir = `gem env gemdir`.chomp
+  sh "SEED=#{seed} MTB_VERBOSE=2 util/bisect -Ilib:bundler/lib:test:#{gemdir}/gems/minitest-server-1.0.5/lib test"
 end
 
 # --------------------------------------------------------------------
--- a/util/bisect
+++ b/util/bisect
@@ -2,7 +2,30 @@
 # frozen_string_literal: true
 
 begin
-  load Gem.bin_path("minitest-bisect", "minitest_bisect")
+  minitest_bisect = Gem.bin_path("minitest-bisect", "minitest_bisect")
 rescue Gem::GemNotFoundException
   abort "Install minitest-bisect gem to run a rubygems test suite bisection"
 end
+
+#
+# Need to monkeypatch the `path_expander` gem to make sure it uses the same
+# sorted set of files as `rake`. Otherwise `rake test` failures are
+# irreproducible. Can be removed once
+# https://github.com/seattlerb/path_expander/pull/4 is released.
+#
+
+module RefinedPathExpander
+  def expand_dirs_to_files(*dirs)
+    super.sort
+  end
+end
+
+require 'path_expander'
+
+class PathExpander
+
+  prepend RefinedPathExpander
+
+end
+
+load minitest_bisect
--- a/util/bisect
+++ b/util/bisect
@@ -1,0 +1,8 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+begin
+  load Gem.bin_path("minitest-bisect", "minitest_bisect")
+rescue Gem::GemNotFoundException
+  abort "Install minitest-bisect gem to run a rubygems test suite bisection"
+end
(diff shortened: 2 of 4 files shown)


util/bisect
```diff
@@ -0,0 +1,20 @@
...
+
+module RefinedPathExpander
+  def expand_dirs_to_files(*dirs)
+    super.sort
+  end
+end
+
+require 'path_expander'
+
+class PathExpander
```
simi: Is this the only change needed to successfully run `minitest_bisect`?

simi:
I don't think this should be part of rubygems itself. There will be problem maintaining this IMHO (for example there's already hardcoded path to minitest server 1.0.5).

If there's any problem running `minitest-bisect` in this codebase without monkey patching path `path_expander`, isn't that actually problem in `minitest-bisect` gem (or `path_expander` gem)?

deivid-rodriguez:
Yes! I needed both changes to get this working, but maybe there are better ways. Can you try this in your environment?

The `path_expander` changes can definitely be incorporated into the `path_expander` gem. I just wanted to try get this working without having to (yet) open PRs to external gems. Maybe I can politely ping @zenspider and ask for his opinion?

The hardcoded route to `minitest-server` is harder to avoid I think. Minitest plugin loading process looks for files with specific names (`minitest/<name_of_plugin>_plugin.rb`) in the `$LOAD_PATH`, and `minitest-server` is implemented as a `minitest` plugin. So... not sure how to do this without using `bundler`... Maybe we should just use `bundler` :sweat_smile:.

util/bisect
deivid-rodriguez: `minitest_bisect` runs fine without this. The problem is that `rake` and `path_expander` (`minitest-bisect`'s file loader) currently load files differently. So, they shuffle a different array when randomizing. Thus, failures reproducible through `rake` don't reproduce through `minitest_bisect` even if the same seed is used.

Rakefile
```diff
@@ -71,6 +71,18 @@ task(:rubocop) do
   sh "util/rubocop"
 end
 
+desc "Run a test suite bisection"
+task(:bisect) do
+  seed = begin
+           Integer(ENV["SEED"])
```
djberg96: I'd vote to let the task take an argument instead of relying on an ENV setting. Or you could do both:

```
task(:bisect, [:seed]) do
   seed ||= ENV['SEED']
   # ...
end
```
Then call it like `rake bisect[12345]`.

deivid-rodriguez: That's neat, I'll add that!

util/bisect
zenspider: Are you on Linux or something such that `Dir["*"]` isn't sorted? Is rake explicitly sorting?

deivid-rodriguez: Yes, I'm on Linux. Rake explicitly sorts [here](https://github.com/ruby/rake/blob/67ced2fab148db0ada6b0d52d321f5cc86e22960/lib/rake/file_list.rb#L404-L409), so for example in rubygems I get

```
> irb(main):001:0> FileList['test/**/test_*.rb']
=> ["test/rubygems/test_bundled_ca.rb", "test/rubygems/test_config.rb", "test/rubygems/test_deprecate.rb", "test/rubygems/test_gem.rb", "test/rubygems/test_gem_available_set.rb", ..., "test/rubygems/test_gem_version.rb", "test/rubygems/test_gem_version_option.rb", "test/rubygems/test_kernel.rb", "test/rubygems/test_remote_fetch_error.rb", "test/rubygems/test_require.rb"]

irb(main):002:0> Dir['test/**/test_*.rb']
=> ["test/rubygems/test_gem_commands_info_command.rb", "test/rubygems/test_gem_resolver_vendor_set.rb", "test/rubygems/test_gem_dependency_list.rb", "test/rubygems/test_gem_source_fetch_problem.rb", "test/rubygems/test_gem_request.rb", "test/rubygems/test_gem_resolver_best_set.rb", ..., "test/rubygems/test_gem_gem_runner.rb", "test/rubygems/test_gem_resolver_local_specification.rb", "test/rubygems/test_gem_available_set.rb", "test/rubygems/test_gem_security_policy.rb", "test/rubygems/test_gem_security.rb", "test/rubygems/test_gem_impossible_dependencies_error.rb"]
```

Rakefile
deivid-rodriguez: I was about to add this but I'm now not really sure I prefer it. Given that for running tests we do `SEED=12345 rake test`, but we can't do `rake test[12345]`. Doesn't it make more sense to keep it consistent?

djberg96: Well, as I said, you can support either approach by defaulting to ENV['SEED'] if no arg is provided, so CI would still work. But, it's your PR. :)

deivid-rodriguez:
How do other devs feel about this? I think it's very useful, I used it again in #2638.

deivid-rodriguez:
By the way, I created a PR to PathExpander, hopefully the monkeypatch can be removed in the future!

hsbt:
It seems useful for me too.

hsbt:
@bundlerbot r+

deivid-rodriguez:
@hsbt I got caught by the cops :smile:

I rebased the PR and fixed the style issues, and also added a comment about the `path_expander` monkeypatching and improved the error message when `minitest-bisect` is not installed.

I think it should be ready now!

hsbt:
Thanks a lot!

hsbt:
@bundlerbot retry

deivid-rodriguez:
Thanks to you! :heart_eyes:

PR #2628, opened 2019-02-06
Fix windows specific executables generated by `gem install`

deivid-rodriguez:
# Description:

This is a rework of #2615. While reviewing that PR I took care of splitting each change into its own commit, and added some comments in the commit messages to explain why I think the change is useful / needed / correct.

First commits are just simplifications and refactorings; the last commit, 961b514, is a bug fix.

Stuff I didn't include from #2615 are stylistic changes, inconsistent stuff, and DRY ups that I consider that harm readability instead of adding to it (for example, moving the umask stuff inside the already complicated method `assert_self_install_permissions`).

I didn't push directly to #2615 because I don't have permissions, but I did keep @MSP-Greg's authorship.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -467,7 +467,7 @@ class Gem::Installer
 
   def generate_windows_script(filename, bindir)
     if Gem.win_platform?
-      script_name = filename + ".bat"
+      script_name = formatted_program_filename(filename) + ".bat"
       script_path = File.join bindir, File.basename(script_name)
       File.open script_path, 'w' do |file|
         file.puts windows_stub_script(bindir, filename)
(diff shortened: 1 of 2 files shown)


bronzdoc:
@deivid-rodriguez thanks for taking the time to explain in detail in each commit

bronzdoc:
@bundlerbot r+

hsbt:
@deivid-rodriguez I removed to mark a milestone for 3.0.x. I'm afraid to merge this with tiny version release Because this change affects many of users.

deivid-rodriguez:
Ok! As far as I know nobody has requested this fix to be backported, so :+1:.

PR #2630, opened 2019-02-07
Enable `Layout/SpaceInsideParens` rubocop cop

deivid-rodriguez:
# Description:

One more `rubocop` cop enabled. We were pretty much consistent here already.

I can see the "alignment" argument to avoid this cop, although I don't like it since all those "pretty alignments" take effort to maintain, and maintaining them tends to obscure the real point of changes. So, if some devs doesn't like this I'll just happily close :)

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -53,6 +53,9 @@ Layout/IndentationConsistency:
 Layout/IndentationWidth:
   Enabled: true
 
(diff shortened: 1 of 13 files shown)


bronzdoc:
@bundlerbot r+

PR #2631, opened 2019-02-07
Fix `--ignore-dependencies` flag not installing platform specific gems

deivid-rodriguez:
# Description:

The problem is that Rubygems won't install platform-specific gems when the `--ignore-dependencies` option is passed.

This is a rework of #2446. It's essentially the same fix proposed by @camertron in #2446 using the same approach, but I split it into an incremental refactoring divided in small commits that always keep tests passing. So, even if the final diff can be hard to understand, a commit per commit review should be pretty straightforward.

Last commit fixes the bug by replacing the raw usage of `Gem::RemoteFetcher` with `Gem::DependencyInstaller#install` which does the proper platform filtering while also respecting the `--ignore-dependencies` flag. It also adds a regression test (borrowed from @MSP-Greg).

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem_commands_install_command.rb
+++ b/test/rubygems/test_gem_commands_install_command.rb
@@ -753,6 +753,23 @@ ERROR:  Possible alternatives: non_existent_with_hint
     assert_equal %w[a-2], @cmd.installed_specs.map { |spec| spec.full_name }
   end
 
+  def test_install_gem_ignore_dependencies_remote_platform_local
+    local = Gem::Platform.local
+    spec_fetcher do |fetcher|
+      fetcher.gem 'a', 3
+
+      fetcher.gem 'a', 3 do |s|
+        s.platform = local
+      end
+    end
+
+    @cmd.options[:ignore_dependencies] = true
+
+    @cmd.install_gem 'a', '>= 0'
+
+    assert_equal %W[a-3-#{local}], @cmd.installed_specs.map { |spec| spec.full_name }
+  end
+
   def test_install_gem_ignore_dependencies_specific_file
     spec = util_spec 'a', 2
 
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -241,7 +241,6 @@ You can use `i` command instead of `install`.
     inst = Gem::Installer.at gem, options
     inst.install
 
-    require 'rubygems/dependency_installer'
     dinst = Gem::DependencyInstaller.new options
     dinst.installed_gems.replace [inst.spec]
 
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -194,12 +194,12 @@ You can use `i` command instead of `install`.
 
     req = Gem::Requirement.create(version)
 
-    inst = Gem::DependencyInstaller.new options
+    dinst = Gem::DependencyInstaller.new options
 
     if options[:ignore_dependencies]
-      install_gem_without_dependencies inst, name, req
+      install_gem_without_dependencies dinst, name, req
     else
-      request_set = inst.resolve_dependencies name, req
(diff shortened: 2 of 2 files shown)


MSP-Greg:
@deivid-rodriguez

Thanks for the work on this.  Does `gem install <gem list> --explain --no-dependencies` work?

I'm probably missing something, but I see explain logic in [`install_gem_with_dependencies`](https://github.com/rubygems/rubygems/blob/5fbc9dc9fd9801213cc4f2009bbfb076fce794fb/lib/rubygems/commands/install_command.rb#L226-L242), but not [`install_gem_without_dependencies`](https://github.com/rubygems/rubygems/blob/5fbc9dc9fd9801213cc4f2009bbfb076fce794fb/lib/rubygems/commands/install_command.rb#L206-L224)...

deivid-rodriguez:
The work is yours, I'm just reviewing it and making sure it's well explained and divided into separate bug fixes so it's easier for other reviewers to review.

The bug you mention is also fixed in your original https://github.com/rubygems/rubygems/pull/2616, but I haven't yet extracted it to its own PR. Remember, one bite at a time :)

MSP-Greg:
Apologies.  I was taking a break and looked at the PR.  Thanks again...

deivid-rodriguez:
@colby-swandale @bronzdoc Care to review this?

bronzdoc:
@lets wait to see if @colby-swandale wants to add something.

test/rubygems/test_gem_commands_install_command.rb
```diff
@@ -753,6 +753,20 @@ def test_install_gem_ignore_dependencies_remote
     assert_equal %w[a-2], @cmd.installed_specs.map { |spec| spec.full_name }
   end
 
+  def test_install_gem_ignore_dependencies_remote_platform_local
+    local = Gem::Platform.local
+    spec_fetcher do |fetcher|
+      fetcher.gem 'a', 3
+      fetcher.gem 'a', 3 do |s| s.platform = local end
```
colby-swandale: I would wrap this in curly brackets instead of `do end` for single lines.

MSP-Greg: Have to add parens?
```ruby
      fetcher.gem 'a', 3
      fetcher.gem('a', 3) { |s| s.platform = local end }
```

deivid-rodriguez: I can also make it multiline to keep the surrounding parens style

```ruby
      fetcher.gem 'a', 3
      fetcher.gem('a', 3) do |s|
        s.platform = local
      end
```

deivid-rodriguez: I had a look at other places and the style used in this PR seems the one being most commonly used for this kind of definitions in general. I like other styles better, so I'll try to enable another style on a separate rubocop PR, but this PR can go in as is :+1:

deivid-rodriguez: #2640 has been merged, so this is good now!

deivid-rodriguez:
@bronzdoc ready!

deivid-rodriguez:
Rebased on top of #2634 that affected the same region of the code.

deivid-rodriguez:
Rebased again and also slightly tweaked the patch so that not only the commit per commit review is clear, but also the final diff is easier to understand.

deivid-rodriguez:
Merging with the previous approvals! :)

@bundlerbot r+

deivid-rodriguez:
Oh my bot! :sweat_smile:

deivid-rodriguez:
@bundlerbot r+

camertron:
Thanks for this, @deivid-rodriguez. Was this change released in an available version of Rubygems?

deivid-rodriguez:
No, not yet. We'll include this in either the next patch level release or rubygems 3.1.

camertron:
👍

PR #2634, opened 2019-02-08
Make `gem install --explain` list platforms

deivid-rodriguez:
# Description:

Extracting more isolated changes from #2616. In this case, now `gem install --explain` informs about gem platforms.

As always full credit to @MSP-Greg for the fixes and the great tests, I'm only explaining them a bit better :)

# Tasks:

- [x] Describe the problem / feature
- [x] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

Issue #2629, opened 2019-02-07
gen install --explain problems


MSP-Greg:
Two issues with `gem install`:

1. When `--explain` is used with `--ignore-dependencies`, the command proceeds to install the gem(s)

```
> gem install did_you_mean --explain --ignore-dependencies
ERROR:  Error installing did_you_mean:
        did_you_mean requires Ruby version >= 2.5.0. The current ruby version is 2.4.5.335.
```

2. When `--explain` is used, platforms are not shown:
```
> gem install jekyll --explain
Gems to install:
  public_suffix-3.0.3
  addressable-2.6.0
  colorator-1.1.0
  http_parser.rb-0.6.0
  eventmachine-1.2.7
  em-websocket-0.5.1
  concurrent-ruby-1.1.4
  i18n-0.9.5
  rb-fsevent-0.10.3
  ffi-1.10.0
  rb-inotify-0.10.0
  sass-listen-4.0.0
  sass-3.7.3
  jekyll-sass-converter-1.5.2
  ruby_dep-1.5.0
  listen-3.1.5
  jekyll-watch-2.1.2
  kramdown-1.17.0
  liquid-4.0.1
  mercenary-0.3.6
  forwardable-extended-2.6.0
  pathutil-0.16.2
  rouge-3.3.0
  safe_yaml-1.0.4
  jekyll-3.8.5
```

`eventmachine-1.2.7` is listed, it should be `eventmachine-1.2.7-x64-mingw32`, which is what would be installed.

3. The output shown with `Jekyll` is is not user friendly.  Could it be alpha sorted, or tree sorted, as in #2459?

The code for `gem install` is affected by the `required_ruby_version` issue, so the 1st & 2nd issues are fixed.  Looking for input on the third.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
RubyGems 3.0.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
closed by https://github.com/rubygems/rubygems/pull/2635 and https://github.com/rubygems/rubygems/pull/2647

PR #2635, opened 2019-02-08
Make `gem update --explain` list platforms

deivid-rodriguez:
# Description:

This is equivalent to #2634 but for `gem update`. As always, commit per commit review recommended.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -146,7 +146,7 @@ command to remove old versions.
     hig
   end
 
-  def highest_remote_version(spec) # :nodoc:
+  def highest_remote_tuple(spec) # :nodoc:
     spec_tuples = fetch_remote_gems spec
 
     matching_gems = spec_tuples.select do |g,_|
@@ -157,7 +157,7 @@ command to remove old versions.
 
     highest_remote_gem ||= [Gem::NameTuple.null]
 
-    highest_remote_gem.first.version
+    highest_remote_gem.first
   end
 
   def install_rubygems(version) # :nodoc:
@@ -271,7 +271,8 @@ command to remove old versions.
       next if not gem_names.empty? and
               gem_names.none? { |name| name == l_spec.name }
 
-      highest_remote_ver = highest_remote_version l_spec
+      highest_remote_tup = highest_remote_tuple l_spec
+      highest_remote_ver = highest_remote_tup.version
 
       if system or (l_spec.version < highest_remote_ver)
         result << [l_spec.name, [l_spec.version, highest_remote_ver].max]
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -97,8 +97,10 @@ command to remove old versions.
     if options[:explain]
       say "Gems to update:"
 
-      gems_to_update.each do |(name, version)|
-        say "  #{name}-#{version}"
+      gems_to_update.each do |(name, version, platform)|
+        say platform == "ruby" ?
+          "  #{name}-#{version}" :
+          "  #{name}-#{version}-#{platform}"
       end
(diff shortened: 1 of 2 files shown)


lib/rubygems/commands/update_command.rb
```diff
@@ -97,8 +97,10 @@ def execute
     if options[:explain]
       say "Gems to update:"
 
-      gems_to_update.each do |(name, version)|
-        say "  #{name}-#{version}"
+      gems_to_update.each do |(name, version, platform)|
+        say platform == "ruby" ?
```
bronzdoc: should the explanation here be the same as when installing?

lib/rubygems/commands/update_command.rb
```diff
@@ -146,18 +148,18 @@ def highest_installed_gems # :nodoc:
     hig
   end
 
-  def highest_remote_version(spec) # :nodoc:
+  def highest_remote_tuple(spec) # :nodoc:
```
bronzdoc: maybe specify in the name a tuple of what?

lib/rubygems/commands/update_command.rb
deivid-rodriguez: Makes sense, I'll try to reconcile that :+1:

lib/rubygems/commands/update_command.rb
deivid-rodriguez: I like it, I'll rename this helper :+1:

lib/rubygems/commands/update_command.rb
deivid-rodriguez: I reviewed this. The logic is, although similar, not the same. But the list has the same shape I think. Since there are other fixes coming to this command (number 1 in https://github.com/rubygems/rubygems/issues/2629), I think it's better not to spend time on trying to DRY this up until I see how the final version with all fixes looks like.

deivid-rodriguez: Ok, I have a refactor of this logic, but I would prefer to add it on a separate PR since it affects the install command as well, and it's not relevant to this bug fix. Is it ok?

lib/rubygems/commands/update_command.rb
deivid-rodriguez: This returns a `NameTuple` instance, probably not the best name since it also includes version and platform information. I guess I could name it `highest_remote_name_tuple`, since resolving the naming issue is probably out of the scope of this PR?

lib/rubygems/commands/update_command.rb
bronzdoc: ok, sounds good

lib/rubygems/commands/update_command.rb
bronzdoc: Since you are renaming from `highest_remote_version` to `highest_remote_tuple` i think is of the scope of this PR but i don't have a suggestion for a better name so lets keep it for now 😁

bronzdoc:
@bundlerbot r+

lib/rubygems/commands/update_command.rb
deivid-rodriguez: Yeah, what I mean is that if I name it `highest_remote_name_version_platform_tuple`, it might be more clear, but it makes it inconsistent with the class it returns (an instance of the `NameTuple` class). So I'm not sure it's worth it.

But maybe naming it `highest_remote_name_tuple` clarifies a bit that the tuple is about how we identify the gem?

deivid-rodriguez: I can at least add this rename to #2643 if you agree!

bronzdoc: Ahh i see, thanks for explaining, yeah lets add the name in #2643 👍

PR #2643, opened 2019-02-14
Refactoring install and update explanations

deivid-rodriguez:
# Description:

When formatting gem versions to be displayed by the `--explain` option of the install and update commands, we have very similar logic that lives in different places. This PR tries to unify that under the `NameTuple` class.

It tries to address https://github.com/rubygems/rubygems/pull/2635#discussion_r256501917.

@bronzdoc Is something like this what you meant by your comment?

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -203,8 +203,8 @@ You can use `i` command instead of `install`.
       if options[:explain]
         say "Gems to install:"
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bronzdoc I also added the rename here as per https://github.com/rubygems/rubygems/pull/2635#discussion_r256889907 and the following comments. :+1:

bronzdoc:
@bundlerbot r+

PR #2647, opened 2019-02-15
Fix explain with ignore-dependencies

deivid-rodriguez:
# Description:

Fixes issue number 1 in https://github.com/rubygems/rubygems/issues/2629. I'm surprised by the amount of red in this PR's diff, but I think it's not needed? :thinking:

# Tasks:

- [x] Write tests


lib/rubygems/commands/install_command.rb
```diff
@@ -196,31 +196,19 @@ def install_gem(name, version) # :nodoc:
...
-    end
-  end
+    request_set = dinst.resolve_dependencies name, req
 
-  def install_gem_without_dependencies(dinst, name, req) # :nodoc:
-    gem = nil
+    if options[:explain]
+      say "Gems to install:"
 
-    if local?
```
bronzdoc: why are we removing this check?

deivid-rodriguez: I made several checks to conclude that we can remove it:

* Removing it doesn't break any tests.
* The `:domain` option is passed and theorically handled by the request set anyways.
* This code path is currently broken anyways, and has been for a long time. In both master, 3.0.2, 2.7.8, and 2.5.2, running `gem install rails --local --ignore-dependencies` crashes with 

```
    no implicit conversion of nil into String
```

After this PR it no longer crashes but it gives a better error instead.

deivid-rodriguez: Ok, I'm now learning how you use the `--local` option. You need to specify a locally built gem with it. In that case, after this PR it seems to be still working as before. In any case, let's hold this PR and I'll first refactor this on a separate PR, adding test coverage for the refactor so we can double check we're good here.

```
```

deivid-rodriguez: I opened #2650 to address this. Once that's merged, I'll rebase this PR on top of it.

deivid-rodriguez: Marking this as solved since I justified the removal in #2650.

deivid-rodriguez:
@bronzdoc This PR should now be simpler to review after #2650 has been merged, this is ready for another round. Thanks!

bronzdoc:
@bundlerbot r+

PR #2650, opened 2019-02-15
Fix cryptic error on local and ignore-dependencies combination

deivid-rodriguez:
# Description:

Addresses https://github.com/rubygems/rubygems/pull/2647#discussion_r257277738 and successive comments by removing code. It feels great! :smile:

When combining the `--local` and the `--ignore-dependencies` flags, and specifying a nonexistent local gem, we would get the error

```
    no implicit conversion of nil into String
```

Now we get a nicer

```
ERROR:  Could not find a valid gem '<gem_name>' (>= 0) in any repository
```

I also added a bit of extra coverage to make sure that removing all this stuff does not really break anything.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -204,17 +204,7 @@ You can use `i` command instead of `install`.
   end
 
   def install_gem_without_dependencies(dinst, name, req) # :nodoc:
-    gem = nil
-
-    if local?
-      gem = fetch_local_gem name, req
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I'm going to move this forward, and rebase #2647 on top of this. I can amend that PR, or create a separate one with any feedback you may have!

deivid-rodriguez:
@bundlerbot r+

PR #2651, opened 2019-02-15
Restore transitiveness of version comparison

deivid-rodriguez:
# Description:

This is an alternative to #2597 fix to #2595.

I strongly think this is the best way to fix this, even if it _could_ create some incompatibility with some gems relying on things like "~> 5.x" being lower than _all_ 5.0.0 prereleases.

As explained in that discussion, the official way that's recommended in the docs to match all prereleases is "~> 5.a", because "a" is the first string in lexicographical order.

I created PRs to the two gems I found relying on this:

* https://github.com/rails/activemodel-serializers-xml/pull/17
* https://github.com/rails/rails-controller-testing/pull/45

I would consider this a bug fix and ship it normally on a bug fix release, but I can understand if others prefer a more conservative approach.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -344,8 +344,8 @@ class Gem::Version
     return unless Gem::Version === other
     return 0 if @version == other._version || canonical_segments == other.canonical_segments
 
-    lhsegments = _segments
-    rhsegments = other._segments
+    lhsegments = canonical_segments
+    rhsegments = other.canonical_segments
 
     lhsize = lhsegments.size
     rhsize = rhsegments.size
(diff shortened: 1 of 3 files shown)


greysteil:
I'm 👍 on shipping this as a bug fix. With the current setup:

```
Gem::Version.new("1.0.beta") >= Gem::Version.new("1.0.0.beta")
# => true
Gem::Version.new("1.0.beta") >= Gem::Version.new("1.0.0.alpha")
# => false
```

That's really surprising - it means that currently version comparison isn't transitive.

deivid-rodriguez:
Happy we're in the same page here! :tada: I should've probably pinged you about this before.

deivid-rodriguez:
Any more opinions here? It would be great to ship this with the next bugfix release!

segiddins:
👍🏻

bronzdoc:
@bundlerbot r+

PR #2652, opened 2019-02-16
Undo requirement sorting

### LINKED ISSUE #2627: Do not sort requirements
#2230 makes requirements sorted, but that is wrong IMO. Let me show you example:

~~~
$ cat testing_gem.gemspec | grep dependency
  s.add_runtime_dependency('test_runtime', ["~> 1.0", ">= 1.0.0"])
  s.add_development_dependency('test_development', ["~> 1.0", ">=1.0.0"])

$ gem buil testing_gem.gemspec 
NOTE: Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01.
Gem::Specification#has_rdoc= called from testing_gem.gemspec:36.
  Successfully built RubyGem
  Name: testing_gem
  Version: 1.0.0
  File: testing_gem-1.0.0.gem

$ gem spec --ruby testing_gem-1.0.0.gem | grep dependency
      s.add_runtime_dependency(%q<test_runtime>.freeze, [">= 1.0.0", "~> 1.0"])
      s.add_development_dependency(%q<test_development>.freeze, [">= 1.0.0", "~> 1.0"])
      s.add_dependency(%q<test_runtime>.freeze, [">= 1.0.0", "~> 1.0"])
      s.add_dependency(%q<test_development>.freeze, [">= 1.0.0", "~> 1.0"])
    s.add_dependency(%q<test_runtime>.freeze, [">= 1.0.0", "~> 1.0"])
    s.add_dependency(%q<test_development>.freeze, [">= 1.0.0", "~> 1.0"])
~~~

This roundtrip should keep the order IMO. I have the requirements specified in that order in my .gesmpec file on purpose. Ther `~> 1.0` is actually the main constraint. The `>= 1.0.0` is there just because something was broken and I am not that interested in this constraint.

And of course, such change which does not improve anything breaks my test suite [[1]]:

~~~
 1) Failure:
TestGem2Rpm#test_print_dependencies [/builddir/build/BUILD/gem2rpm-1.0.1/usr/share/gems/gems/gem2rpm-1.0.1/test/test_gem2rpm.rb:74]:
--- expected
+++ actual
@@ -1,8 +1,7 @@
-# encoding: UTF-8
-"rubygem(test_runtime) >= 1.0
+"rubygem(test_runtime) >= 1.0.0
+rubygem(test_runtime) >= 1.0
 rubygem(test_runtime) < 2
-rubygem(test_runtime) >= 1.0.0
+rubygem(test_development) >= 1.0.0
 rubygem(test_development) >= 1.0
 rubygem(test_development) < 2
-rubygem(test_development) >= 1.0.0
 "
~~~

IOW I would appreciate if only `Gem::Requirement#as_list` is sorted as it used to be.

[1]: https://github.com/fedora-ruby/gem2rpm/blob/master/test/test_gem2rpm.rb#L59

deivid-rodriguez:
# Description:

It should fix #2627 while still fixing the issue #2230 was fixing, I think.

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_requirement.rb
+++ b/test/rubygems/test_gem_requirement.rb
@@ -35,7 +35,6 @@ class TestGemRequirement < Gem::TestCase
     assert_requirement_equal "= 2", ["2"]
     assert_requirement_equal "= 2", v(2)
     assert_requirement_equal "2.0", "2"
-    assert_requirement_equal ["= 2", ">= 2"], [">= 2", "= 2"]
   end
 
   def test_create
--- a/lib/rubygems/requirement.rb
+++ b/lib/rubygems/requirement.rb
@@ -130,7 +130,6 @@ class Gem::Requirement
       @requirements = [DefaultRequirement]
     else
       @requirements = requirements.map! { |r| self.class.parse r }
-      sort_requirements!
     end
   end
 
@@ -144,7 +143,6 @@ class Gem::Requirement
     new = new.map { |r| self.class.parse r }
 
     @requirements.concat new
-    sort_requirements!
   end
 
   ##
@@ -186,7 +184,7 @@ class Gem::Requirement
   end
 
   def hash # :nodoc:
-    requirements.hash
+    requirements.sort.hash
   end
 
   def marshal_dump # :nodoc:
@@ -292,14 +290,6 @@ class Gem::Requirement
       if r[0].kind_of? Gem::SyckDefaultKey
         r[0] = "="
       end
-    end
-  end
-
-  def sort_requirements! # :nodoc:
-    @requirements.sort! do |l, r|
-      comp = l.last <=> r.last # first, sort by the requirement's version
-      next comp unless comp == 0
-      l.first <=> r.first # then, sort by the operator (for stability)
     end
   end
 
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -2544,6 +2544,14 @@ end
     assert_equal @c1, same_spec
   end
 
+  def test_to_ruby_keeps_requirements_as_originally_specified
+    spec = util_spec 'a', '1' do |s|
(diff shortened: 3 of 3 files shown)


test/rubygems/test_gem_requirement.rb
```diff
@@ -35,7 +35,6 @@ def test_initialize
     assert_requirement_equal "= 2", ["2"]
     assert_requirement_equal "= 2", v(2)
     assert_requirement_equal "2.0", "2"
-    assert_requirement_equal ["= 2", ">= 2"], [">= 2", "= 2"]
```
bronzdoc: If this PR still fixes https://github.com/rubygems/rubygems/pull/2230 why we deleting this assertion?

deivid-rodriguez: As I understand the wording of that PR, the problem was that versions `1.0.0` and `1.0` were not considered equivalent. Or similarly for [the associated bundler issue](https://github.com/bundler/bundler/issues/6295), that the requirements `>= 0.0.0` and `>= 0` were not considered equivalent.

#2230 fixed that issue and added a bunch of assertions covering that, but at the same time started sorting the requirements and added this specific test covering that.

In this PR I'm only reverting the sorting of requirements (and thus removing this associated spec), but I'm keep the fix for the above problem and all the tests covering that.

bronzdoc: @segiddins want to give feedback? ^

bronzdoc:
@bundlerbot r+

test/rubygems/test_gem_requirement.rb
segiddins: I’m not happy to see this reverted

deivid-rodriguez: Can you elaborate @segiddins? It's no big deal if we need to revert this PR.

PR #2654, opened 2019-02-17
test_gem.rb - fix intermittent test

MSP-Greg:
`test_gem.rb` has the following code (note line 158):

https://github.com/rubygems/rubygems/blob/3d27eb01f8ab89a672915ebd56c7b70768739749/test/rubygems/test_gem.rb#L156-L163

Failures are caused by `Gem.ruby`, which matches the test setting, see:

https://ci.appveyor.com/project/MSP-Greg/rubygems/build/job/wudlxafxreyk3auy#L31

# Tasks:

- [X] Fix tests

### DIFF
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -955,37 +955,35 @@ class TestGem < Gem::TestCase
 
   def test_self_ruby_escaping_spaces_in_path
     orig_bindir = RbConfig::CONFIG['bindir']
-    orig_ruby_install_name = RbConfig::CONFIG['ruby_install_name']
     orig_exe_ext = RbConfig::CONFIG['EXEEXT']
 
     RbConfig::CONFIG['bindir'] = "C:/Ruby 1.8/bin"
-    RbConfig::CONFIG['ruby_install_name'] = "ruby"
     RbConfig::CONFIG['EXEEXT'] = ".exe"
 
-    with_clean_path_to_ruby do
-      assert_equal "\"C:/Ruby 1.8/bin/ruby.exe\"", Gem.ruby
+    ruby_install_name "ruby" do
+      with_clean_path_to_ruby do
+        assert_equal "\"C:/Ruby 1.8/bin/ruby.exe\"", Gem.ruby
+      end
     end
   ensure
     RbConfig::CONFIG['bindir'] = orig_bindir
-    RbConfig::CONFIG['ruby_install_name'] = orig_ruby_install_name
     RbConfig::CONFIG['EXEEXT'] = orig_exe_ext
   end
 
   def test_self_ruby_path_without_spaces
     orig_bindir = RbConfig::CONFIG['bindir']
-    orig_ruby_install_name = RbConfig::CONFIG['ruby_install_name']
     orig_exe_ext = RbConfig::CONFIG['EXEEXT']
 
     RbConfig::CONFIG['bindir'] = "C:/Ruby18/bin"
-    RbConfig::CONFIG['ruby_install_name'] = "ruby"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I assume the problem is that `assert_self_install_permissions(format_executable: true)` is using `Gem.ruby` and thus leaving its value memoized to the "ruby27" value.

So, wrap the `assert_self_install_permissions(format_executable: true)` call with a method that saves the previous value of `Gem.ruby` and then restores it after the side-effects have happened.

```ruby
def test_self_install_permissions_with_format_executable_and_non_standard_ruby_install_name 
  Gem::Installer.exec_format = nil 
  ruby_install_name 'ruby27' do 
    restoring_path_to_ruby do
      assert_self_install_permissions(format_executable: true) 
    end
  end
end

# ....

def restoring_path_to_ruby
  orig_ruby = Gem.ruby
  yield
ensure
  Gem.instance_variable_set("@ruby", orig_ruby)
end
```

MSP-Greg:
@deivid-rodriguez

In testing, a common pattern is to use support methods that change something, then yield, then restore it.

I see no need to add another method along with another block.  One line fixes the issue.

deivid-rodriguez:
In my opinion, putting that line there makes little sense, since that method is not modifying the `@ruby` instance variable so it's super-obscure to realize what problem it is fixing.

If you want to fix this with a one liner, at least put the line at the end of the method that it's creating the side effect, namely, `assert_self_install_permissions`, or even better, around the `Gem.install` line.

deivid-rodriguez:
The same logic I'm proposing [already exists in the same file](https://github.com/rubygems/rubygems/blob/master/test/rubygems/test_gem.rb#L954-L991), so you can even refactor it to a single place.

MSP-Greg:
JFYI, I just looked for 'ruby27' as that was the string that was cuasing the error.

Makes one think about a generic method in `test_case.rb` like `RbConfigWrap` that would take a hash/kwargs, grab the current values, then set them to the passed values, then yield, then restore?

It could get messy, but it would be a 'central place' where one would also pick up things like `Gem.ruby` needing to be reset if `RbConfig::CONFIG['ruby_install_name']` was changed?

MSP-Greg:
@deivid-rodriguez

I created a branch for the issue, 1st build restricted it to Ruby 2.5 and added the seed value.  As expected, crashed and burned.  Log is [here](https://ci.appveyor.com/project/MSP-Greg/rubygems/builds/22444176).

Added another commit with the one line patch from this PR, and it passed.  Log [here](https://ci.appveyor.com/project/MSP-Greg/rubygems/builds/22444230).


Hence, the problem identified is fixed for the seed value.  If you want to clean up the tests further, I'll be happy to close this PR.

deivid-rodriguez:
@MSP-Greg I know your change fixes the issue, I tried it :). But in my opinion, it's not a good fix because it's very unclear why we reset the state at that point. I'll push an update with some related cleanups.

deivid-rodriguez:
By the way, this is a minimal reproduction of the flaky being fixed here:

```
SEED=6332 rake TESTOPTS="--name=/test_looks_for_gemdeps_files_automatically_on_start_in_parent_dir\|test_self_install_permissions_with_format_executable_and_non_standard_ruby_install_name/"
```

MSP-Greg:
@deivid-rodriguez

Thanks for the work.

>  it's very unclear

One might say the same thing about what is or isn't handled by util methods and/or TestCase setup/teardown.  I guess I didn't see it as a strong issue; I just wanted the test fixed.  I'm all for instability, but not in code or tests...

deivid-rodriguez:
@bundlerbot r+

Issue #2653, opened 2019-02-17
Fixup - required_ruby_version constraints - platform questions


MSP-Greg:
Regarding pushing the `required_ruby_version` PR, I've got a question that came up today.

Re PR #1093, I'm a little unclear as to the intent, as the PR is addressing expected behavior when
`--platform=x86-mingw32` is used with `gem install`.

RG's test setup sets the platform to `x86-mswin32-60`, so `Gem::platform.local` is not compatible with
`--platform=x86-mingw32`...

So, is the intent to allow **non-compatible gems** to be installed using the `gem install`  `--platform` option?

@kou - sorry for the ping, but I thought your opinion would be helpful.

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:

```
master
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
@MSP-Greg did you get your answers?

MSP-Greg:
@bronzdoc

No.  When one can do a fetch/unpack to look at a gem that is platform incompatible, I'm not sure why `gem install` should allow it?

bronzdoc:
> So, is the intent to allow non-compatible gems to be installed using the gem install --platform option?

The intent is to allow the user to install the gem from the platform they specify with the `--platform`  flag, if it happens that the gem they are installing is not compatible with their current platform, that's their responsibility.

PR #2655, opened 2019-02-17
Implement spec required_ruby_version constraint

MSP-Greg:
## Description:

Currently the `required_ruby_version` constraint is ignored.  This causes issues as more & more gems are using it in new releases, along the normal use in pre-compiled gems.

Fixes the issue in several places.

### Notes:

Since the tuples stored in the local cache do not contain `required_ruby_version` info, specs must be downloaded and checked.  The algorithms should minimize that.

The specs/tuples contained in `latest_specs.4.8` are no longer of use, since one needs to check the other cache files, as the most recent gem that matches `required_ruby_version` may not be the most recent.

As mentioned previously, 'spec' sorting by platform has issues. See Issue #2653.  Another posted soon.

## Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/resolver.rb
+++ b/lib/rubygems/resolver.rb
@@ -199,7 +199,8 @@ class Gem::Resolver
 
   ##
   # Extracts the specifications that may be able to fulfill +dependency+ and
-  # returns those that match the local platform and all those that match.
+  # returns an array whose first element is those that match the local platform
+  # and Ruby version.  The second element is all those that match the dependency.
 
   def find_possible(dependency) # :nodoc:
     dep_str = dependency.to_s
@@ -215,9 +216,7 @@ class Gem::Resolver
       all = matching unless matching.empty?
     end
 
-    matching_platform = select_local_platforms all
-
-    @find_possible[dep_str] = [matching_platform, all]
+    @find_possible[dep_str] = [select_local_platforms_ruby_version(all), all]
   end
 
   ##
@@ -226,6 +225,21 @@ class Gem::Resolver
   def select_local_platforms(specs) # :nodoc:
     specs.select do |spec|
       Gem::Platform.installable? spec
+    end
+  end
+
+  ##
+  # Returns the gems in +specs+ that match the local platform and also the Ruby
+  # version.
+  # TODO: can this return just the first matching, rather than all?
+
+  def select_local_platforms_ruby_version(specs) # :nodoc:
+    ruby_version = Gem.ruby_version
+    specs.select do |spec|
+      Gem::Platform.installable?(spec) &&
+      (Gem::Specification === spec ?
+        spec.required_ruby_version.satisfied_by?(ruby_version) :
+        spec.spec.required_ruby_version.satisfied_by?(ruby_version))
     end
   end
 
@@ -251,7 +265,7 @@ class Gem::Resolver
 
     sources.each do |source|
       groups[source].
-        sort_by { |spec| [spec.version, Gem::Platform.local =~ spec.platform ? 1 : 0] }.
+        sort_by { |spec| [spec.version, spec.platform == Gem::Platform::RUBY ? -1 : 1] }.
         map { |spec| ActivationRequest.new spec, dependency, [] }.
         each { |activation_request| activation_requests << activation_request }
     end
--- a/lib/rubygems/resolver.rb
+++ b/lib/rubygems/resolver.rb
@@ -111,6 +111,7 @@ class Gem::Resolver
     @skip_gems           = {}
     @soft_missing        = false
     @stats               = Gem::Resolver::Stats.new
+    @find_possible       = {}
   end
 
   def explain(stage, *data) # :nodoc:
@@ -185,12 +186,14 @@ class Gem::Resolver
   # Proceed with resolution! Returns an array of ActivationRequest objects.
 
   def resolve
+    @find_possible = {} # caches find_possible  results
     locking_dg = Molinillo::DependencyGraph.new
     Molinillo::Resolver.new(self, self).resolve(@needed.map { |d| DependencyRequest.new d, nil }, locking_dg).tsort.map(&:payload).compact
   rescue Molinillo::VersionConflict => e
     conflict = e.conflicts.values.first
     raise Gem::DependencyResolutionError, Conflict.new(conflict.requirement_trees.first.first, conflict.existing, conflict.requirement)
   ensure
+    @find_possible = {}
     @output.close if defined?(@output) and !debug?
   end
 
@@ -199,6 +202,9 @@ class Gem::Resolver
   # returns those that match the local platform and all those that match.
 
   def find_possible(dependency) # :nodoc:
+    dep_str = dependency.to_s
+    return @find_possible[dep_str] if @find_possible.key? dep_str
+
     all = @set.find_all dependency
 
     if (skip_dep_gems = skip_gems[dependency.name]) && !skip_dep_gems.empty?
@@ -211,7 +217,7 @@ class Gem::Resolver
 
     matching_platform = select_local_platforms all
 
-    return matching_platform, all
+    @find_possible[dep_str] = [matching_platform, all]
   end
 
   ##
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -125,7 +125,7 @@ command to remove old versions.
 
     fetcher = Gem::SpecFetcher.fetcher
 
-    spec_tuples, errors = fetcher.search_for_dependency dependency
+    spec_tuples, errors = fetcher.search_for_dependency dependency, best_only: true
 
     error = errors.find { |e| e.respond_to? :exception }
 
@@ -147,17 +147,13 @@ command to remove old versions.
   end
 
   def highest_remote_name_tuple(spec) # :nodoc:
-    spec_tuples = fetch_remote_gems spec
+    dependency = Gem::Dependency.new spec.name, "> #{spec.version}"
+    dependency.prerelease = options[:prerelease]
 
-    matching_gems = spec_tuples.select do |g,_|
-      g.name == spec.name and g.match_platform?
-    end
+    fetcher = Gem::SpecFetcher.fetcher
+    specs, _ = fetcher.spec_for_dependency dependency, best_only: true
 
-    highest_remote_gem = matching_gems.max
-
-    highest_remote_gem ||= [Gem::NameTuple.null]
-
-    highest_remote_gem.first
+    specs.empty? ? Gem::NameTuple.null : specs[0][0].name_tuple
   end
 
   def install_rubygems(version) # :nodoc:
--- a/lib/rubygems/dependency_installer.rb
+++ b/lib/rubygems/dependency_installer.rb
@@ -212,56 +212,26 @@ class Gem::DependencyInstaller
     end
 
     if consider_remote?
-      begin
-        # TODO this is pulled from #spec_for_dependency to allow
-        # us to filter tuples before fetching specs.
-        #
-        tuples, errors = Gem::SpecFetcher.fetcher.search_for_dependency dep
+      specs, errors = Gem::SpecFetcher.fetcher.spec_for_dependency(dep, best_only: best_only)
 
-        if best_only && !tuples.empty?
-          tuples.sort! do |a,b|
-            if b[0].version == a[0].version
-              if b[0].platform != Gem::Platform::RUBY
-                1
-              else
-                -1
-              end
-            else
-              b[0].version <=> a[0].version
-            end
-          end
-          tuples = [tuples.first]
-        end
-
-        specs = []
-        tuples.each do |tup, source|
-          begin
-            spec = source.fetch_spec(tup)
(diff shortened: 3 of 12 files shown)


lib/rubygems/resolver.rb
```diff
@@ -111,6 +111,7 @@ def initialize(needed, set = nil)
     @skip_gems           = {}
     @soft_missing        = false
     @stats               = Gem::Resolver::Stats.new
+    @find_possible       = {}
```
deivid-rodriguez: Is this memoization for performance necessary for the fix? Can you split it to a separate PR and provide a benchmark proving that it makes a difference?

bronzdoc: ping @MSP-Greg

MSP-Greg: As I stated in the comment for the commit:
Using `Gem::RequestSet.new(<rails (>= 0.0.0)>).resolve`,
`find_possible` was called 447 times, but 398 calls were previously calculated.

Maybe instead of a benchmark, try setting a breakpoint in #find_possible for debugging...

deivid-rodriguez: @MSP-Greg I can take care of checking how much of a difference the memoization makes but, can you split it out from this PR?

MSP-Greg: > can you split it out from this PR

The 2nd commit is built on top of it.  So, I'd prefer not to.

Given that the code involved will be hitting Rubygems.org, I doubt the net time change is even noticeable.  I was debugging in the method, and 447 vs 49 is an order of magnitude drop...

deivid-rodriguez: Ok, I'll do it.

deivid-rodriguez:
So, I've been having a look at this PR and I'm noticing that the performance is (understandably) really affected by this change.

As an example, `gem install rails --pre` on my machine goes from 2.30m to 8m.

Maybe the best approach here is to start using the compact index for resolution.

Thoughts?

MSP-Greg:
> As an example, gem install rails --pre on my machine goes from 2.30m to 8m

First, I renamed my `quick/Marshal.4.8` folder to clear the cached specs, then ran
```
gem install rails --pre --explain
```

1st run - 66.4 s, 2nd - 10.7 s, 3rd - 2.9 s.  Very odd that the 1st run time is much smaller than your differential.

Thanks for revisiting this, as there are too many specs downloaded into `quick/Marshal.4.8`.  I thought I set it up so it just looked at specs until it found the highest matching spec.  It appears to be downloading more.  That may be in the code that's actually fetching the specs from rubygems.org, I don't recall if I got that far into the code.  Also, issues with the time it takes to initiate an SSL connection vs downloading extra data after the initiation, keep-alive, etc...

> Maybe the best approach here is to start using the compact index for resolution.

Sorry, I'm not sue what you're referring to.  The Bundler compact index?  I suspect the only way to significantly speed it up is to add `required_ruby_version` info to the marshalled spec caches (latest_specs.4.8, prerelease_specs.4.8, specs.4.8).  That requires changes from the client to the server.

So, yes, I don't consider this an optimal solution.  I think it works, ands fixes a long standing issue that I believe will only get worse (gems with `required_ruby_version`).

deivid-rodriguez:
> 1st run - 66.4 s, 2nd - 10.7 s, 3rd - 2.9 s. Very odd that the 1st run time is much smaller than your differential.

I think my slow times are because I was running under a slow internet connection. But what matters is the comparison, right? What times do you get on master compared to this PR?

> Sorry, I'm not sure what you're referring to. The Bundler compact index? 

Yeah, sorry for not being more specific, I mean bundler's compact index indeed. I think that API already considers `required_ruby_version` and `required_rubygems_version` on the server side, so I guess should be much more efficient.

But if we go with this approach, should we instead try this solution as a fallback, when the faster resolution results in a non-installable gem? That way we would make this edge case a bit slower, but we wouldn't affect any currently correct cases. What do you think?

MSP-Greg:
> I think my slow times are because I was running under a slow internet connection. But what matters is the comparison, right?

Well, the conditions of the comparison, etc, etc.  I used `--explain` for most of my testing.  Assuming the normal use case is to install, if the 'resolution time' doubles (or more), but the 'install time' is the far larger part of the total time, I personally would prefer to have `required_ruby_version` taken into account.

> What times do you get on master compared to this PR?

1st run - 6.0 s, 2nd - 2.3 s, 3rd - 2.3 s.  Only spec that was downloaded was `rails-6.0.0.beta2.gemspec`, which is expected.

> But if we go with this approach, should we instead try this solution as a fallback, when the faster resolution results in a non-installable gem?

So if there's a 'non-installable gem', what happens if the 'installable gem' breaks the requirement of it's parent?  This could get really messy...

This kind of goes to what I may have said 'between the lines' in various issues, etc.  The current code base kind of blurs the resolution/installation boundary.  This makes `--explain` problematic, but also causes other issues.  I think 'full resolution' should be performed first, then the gems need to be installed in sorted order, as we do need child dependencies installed before their parents.

Also, I'll see if I can look thru the spec download code, as reducing the number of downloaded specs should improve the 'time to resolve'...

Thoughts?

deivid-rodriguez:
> So if there's a 'non-installable gem', what happens if the 'installable gem' breaks the requirement of it's parent? This could get really messy...

I don't understand this. This solution should always work, right? So, if we rescue the exception currently raised when the chosen gem does not respect the ruby version constraint, and try this more bulletproof solution... Shouldn't it work and not affect the performance of the majority of cases when `gem install` works fine?

> Also, I'll see if I can look thru the spec download code, as reducing the number of downloaded specs should improve the 'time to resolve'...

Thanks, that would be great!

MSP-Greg:
Confused.

> I don't understand this. This solution should always work, right?

Currently, it does.

>So, if we rescue the exception currently raised when the chosen gem does not respect the ruby version constraint, and try this more bulletproof solution...

Confused about this.  By 'exception currently raised', the only exception raised I know of is during install.  How about the below gem set?

```
Installing a
a, b ~> 1.8

b-1.9, c ~> 2.0
b-1.8, c ~> 1.9

c-2.0  requires ruby version greater than current
c-1.9  meets current ruby version
```

Make sense?  This is part of the 'resolution/installation' blur...

deivid-rodriguez:
> Confused about this. By 'exception currently raised', the only exception raised I know of is during install. How about the below gem set?

Me too :rofl:, aren't we talking about `gem install` here? (I know you also changed `gem update` in this PR, but I plan to incorporate those separately).

deivid-rodriguez:
Rescuing the `gem install` exception (and falling back to the slower but 100% correct method) would also be a great place to inform the user about what's going on, and why she will not get the latest version installed but an older one.

MSP-Greg:
>aren't we talking about gem install here?

Using my example, trying to install `c-2.0` will generate an error in `Gem::Commands::InstallCommand#install_gem`, but that's after the call to install `b-1.9`.  So that call would be in error, as `b-1.8` is what should be installed.

deivid-rodriguez:
So the problem is that the first incorrect attempt would leave some gems installed that are not necessary for the second correct attempt. Is that it?

MSP-Greg:
> So the problem is that the first incorrect attempt would leave some gems installed that are not necessary for the second correct attempt. Is that it?

Somewhat, as that can occur within the dependencies.  The main issue is backing up to install `b-1.9`.

As I think I've mentioned, RubyGems is capable of taking a set of gems (the set entered in `gem install`) and fully determining every gem that needs to be installed.  It can also generate a 'sorted list' that shows the correct installation order (so that a gem is not installed before any of its dependencies).

With that, there doesn't need to be 'gem install with dependencies' logic (and the nesting that occurs with it), as the sorted list gives the correct install order.  This is what I've referred to as 'resolve' before 'install'.

Thoughts?

MSP-Greg:
@deivid-rodriguez

Re 'resolve' before 'install', you've already done some of that with:

https://github.com/rubygems/rubygems/blob/9043e16cf7e23cc06fa5a800b276d68525b07360/lib/rubygems/commands/install_command.rb#L199

The current PR can't be merged.  I'll work on it this weekend, along with getting the spec downloads to the minimal number.  Then we can figure out how to dice & chop it...

MSP-Greg:
@deivid-rodriguez 

JFYI, resolution **does not** back track, it just fails.  Unfortunately, the only solution would be for conditionals to exist in gemspec files...

Anyway, I think that if resolution & installation are separated, the rrv constraint can be verified on a case by case basis (as you suggested).  It should require a lot less downloaded specs.

The basic flow would be to generate a request set (making sure that all options are used), then, generate `sorted_requests`, use it to check rrv, and finally, use it to install gems without dependency resolution, since it's already done by the request set.

Thoughts?

deivid-rodriguez:
So basically try the current way, verify in the rrv is satisfied on the chosen set of specs, if it is, install it, otherwise, try the new slower but always correct way.

Something like that? I think that sounds good to me, yeah.

MSP-Greg:
You're making me think about all of this again.  I was about to compose this.

I was mistaken.  The above idea is not good.

As I recall, the 'rails' requests nest 5 deep (I think the 2nd level is mostly the 'sub' action/active gems) .  

So, lets say that the an rrv failure occurs on a dependency that's at level 3.  If one finds a matching gem, it's dependencies may change.  So, we'd need to resolve both it and its dependencies again, and adjust the `sorted_requests` set accordingly.  Consider also that one of its dependencies could be shared with another gem.  I think this may not work, it will certainly be brittle, etc, etc.

Said another way, the rrv constraint needs to be considered within and while the top-level resolution happens.  So, I don't think a first pass resolution without rrv will work.

Make sense?

So, we're back to something similar to my original code.  With a clean spec cache, using rails, my system downloaded over 1k specs.  I changed the code to only download the top 10, and it decreased to about 250.  It would be less with 5.  Maybe the code could run where it starts with a level of 5, and if no errors occur, continues, but if there are errors, it increases the level to 10, and tries again?

I suspect the top 5 should work with a 'current' gem install, but it might error if one were trying to install something like rails 3.x or 4.x?

Thoughts?

MSP-Greg:
@deivid-rodriguez

Ok.  I believe the best solution is to add the rrv info to the response from:
```
https://rubygems.org/api/v1/dependencies?
```
It has all the info needed for resolution except rrv, and it's already being queried…

I'm not really keen on looking thru the rails code.  Any idea as how to get this changed?

Someone active at https://github.com/rubygems/rubygems.org...

deivid-rodriguez:
@MSP-Greg This is what I suggested in the beginning: use bundler's compact index API that already includes rrv information:

```
https://rubygems.org/info/rails
```

MSP-Greg:
@deivid-rodriguez

There's a lot of issues as to which direction to go.  I'd prefer to change the 'dependencies? api` used for RG rather than use the 'info api' used by bundler.

#### 'dependencies?` pro

* it's straight yaml/json, 'info' is proprietary string.
* minimal changes to RG if rrv is added to 'dependencies?`

#### 'info' pro

* with Bundler caching the info, it could be faster, but I haven't seen any 'why is Bundler so fast and RubyGems so slow' issues, so RG hitting the API endpoint for every query probably isn't an issue?
* where should to code go?  RG calling Bundler?  move the code to RG and change Bundler?  duplicate it?

I haven't gotten thru all the code, but it seems that **development dependencies** are not included in either, and may require downloading specs.  Any thoughts on that?

Re RG issues, a few times I have heard something similar to 'that's what Bundler is for'.  I don't know the history is between the two, but there is definitely duplicated functionality, much of which I think should exist only in RG.

I consider RG to be a package manager, but it's more complex that other package managers due to the additional issues of multiple package versions and runtime (Ruby) versions.

Bundler is a tool for apps that should aid in moving code from development to production.  Hence, one key function is 'compiling' the info in multiple gemspec files into one file.

IMO, at present RG & Bundler are still separate.  Simply adding Bundler files to the RG distribution is not the same as integrating the two.

indirect:
The dependencies API is very costly on the server-side. Because of how costly the dependencies API is to run, the RG.org server team does not plan to ever add required ruby or rubygems versions to the responses provided in the dependencies API. In addition, we want to move away from it as much as possible in the long run. I hope that information helps.

MSP-Greg:
Finally had time today to look at this.

@indirect

Thanks for the reply.  I also came across an almost five year old post:
https://andre.arko.net/2014/03/28/the-new-rubygems-index-format/

@deivid-rodriguez

I guess you're familiar with it:
https://github.com/bundler/bundler/commit/ed85fb9f7acc20c317737f3e157bcee38fbd527c

Hence, let's go with the https://rubygems.org/info endpoint.  Maybe it's time to plan for changing all of RG over to the newer endpoints used by bundler?

deivid-rodriguez:
Thanks for chiming in @indirect! I was going to reply something along those lines. It's great that we are on the same page now.

So, my idea here would be to move the new fetcher from bundler into rubygems, so rubygems can use it. Once we have that working, we can add a compatibiltiy layer to bundler so that the bundler's dependency on rubygems doesn't need to be bumped.

MSP-Greg:
> my idea here would be to move the new fetcher from bundler into rubygems

Cool.  Thanks.  I think I have to spend a bit of time looking at Bundler code...

PR #2662, opened 2019-02-28
Fallback to IPv4 when IPv6 is unreachable

### LINKED ISSUE #3463: Incorrect resolution related to required_ruby_version on macOS on Ruby 2.3
- [x] You checked all the following documents and couldn't find a solution for your issue:

- [Troubleshooting common issues](https://github.com/rubygems/bundler/blob/master/doc/TROUBLESHOOTING.md)
- [Bundler documentation site](https://bundler.io/)
- [Bundler man pages](https://bundler.io/man/bundle.1.html)
- [Bundler command line reference](https://bundler.io/v2.0/commands.html)

Please answer the following questions so we can process your issue as fast as possible:

1. Do you have a reproduction script? It would be really helpful for us if you provide one.

https://github.com/ffi/ffi/runs/545023910?check_suite_focus=true is essentially a reproduction script running in GitHub Actions.
Branch: https://github.com/eregon/ffi/tree/debug-bundler

3. What command did you run?

`bundle install`

5. What actually happened?

https://github.com/ffi/ffi/runs/545023910?check_suite_focus=true
```
$ bundle install
Fetching gem metadata from https://rubygems.org/..............
Resolving dependencies...
Fetching rake 13.0.1
Installing rake 13.0.1
Using bundler 2.1.4
Fetching diff-lcs 1.3
Installing diff-lcs 1.3
Fetching io-console 0.5.6
Installing io-console 0.5.6 with native extensions
Gem::InstallError: io-console requires Ruby version >= 2.4.0.
An error occurred while installing io-console (0.5.6), and Bundler cannot
continue.
Make sure that `gem install io-console -v '0.5.6' --source
'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  rubygems-tasks was resolved to 0.2.5, which depends on
    irb was resolved to 1.2.3, which depends on
      reline was resolved to 0.1.3, which depends on
        io-console
##[error]Process completed with exit code 5.
```

However that seems transient, sometimes it seems to work, such as in https://github.com/ffi/ffi/runs/544565919?check_suite_focus=true :
```
Resolving dependencies...
Fetching rake 13.0.1
Installing rake 13.0.1
Using bundler 2.1.4
Fetching diff-lcs 1.3
Installing diff-lcs 1.3
Fetching irb 1.1.1
Installing irb 1.1.1
Fetching kramdown 2.1.0
Installing kramdown 2.1.0
Fetching rake-compiler 1.0.9
Installing rake-compiler 1.0.9
Fetching rake-compiler-dock 1.0.1
Installing rake-compiler-dock 1.0.1
Fetching rspec-support 3.9.2
Installing rspec-support 3.9.2
Fetching rspec-core 3.9.1
Installing rspec-core 3.9.1
Fetching rspec-expectations 3.9.1
Installing rspec-expectations 3.9.1
Fetching rspec-mocks 3.9.1
Installing rspec-mocks 3.9.1
Fetching rspec 3.9.0
Installing rspec 3.9.0
Fetching rubygems-tasks 0.2.5
Installing rubygems-tasks 0.2.5
Fetching yard 0.9.24
Installing yard 0.9.24
Bundle complete! 8 Gemfile dependencies, 14 gems now installed.
```

where it picks irb 1.1.1 which doesn't have the reline dependency.

7. Please run `bundle env` and paste the output below:

This is on Ruby 2.3.8 on macOS, using Bundler 2.1.4.
There is no checked-in Gemfile.lock.
On Linux it seems to not happen.

Any idea what might happen and how to workaround this?

### LINKED ISSUE #3641: Could not find a valid gem 'bundler' (= 2.1.4), here is why: Unable to download data from https://rubygems.org/ - timed out (https://rubygems.org/specs.4.8.gz)
I'm having a problem 

My current problem is that I can't deploy a rails app into production with Docker on AWS.
Sometimes half of the gems will install but, with other builds, it's not even possible to install bundler. 

`Could not find a valid gem 'bundler' (= 2.1.4), here is why: Unable to download data from https://rubygems.org/ - timed out (https://rubygems.org/specs.4.8.gz)`

`Gem::RemoteFetcher::UnknownHostError timed out (https://rubygems.org/gems/rake-13.0.1.gem)
`

`bridge-nf-call-iptables is disabled` which I assume would overcome any of the IPV6 issues that I have been reading about. 

This issue is related to:

  - [x] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

```
$ gem env version
3.0.3
```

sonalkr132:
# Description:

Multiple users have reported that IPv6 address of rubygems.org is
unreachable[[1](https://help.rubygems.org/discussions/problems/33671-suggestion-for-easy-solution-for-the-ipv6-related-timeout-problem)][[2](https://help.rubygems.org/discussions/problems/31074-timeout-error)]. We have verified using online tools[[3](https://www.ssllabs.com/ssltest/analyze.html?d=rubygems.org)][[4](https://ready.chair6.net/?url=rubygems.org)] as well
as all of aws regions with ipv6 support that there is no misconfig issue
on our part.
IPv6 route being broken is a bit expected and hence, RFC8305[[5](https://tools.ietf.org/html/rfc8305)]
recommends starting connection to both IPv4 and IPv6 address in parallel
and use whichever returns first (this is TLDR explanation, check RFC for
details). This patch of TCPSocket.initialize does something similar, as
in, it starts connection to all resolved IPs in parallel and uses first
successful connection's IP to initalize the TCPSocket. We use `Net:HTTP`
as our remote featcher and it internally uses TCPSocket.open[[6](https://github.com/ruby/ruby/blob/4444025d16ae1a586eee6a0ac9bdd09e33833f3c/lib/net/http.rb#L950)].

Even tho support of IPv6 fallback is spotty, I have open a ruby issue
for it[[7](https://bugs.ruby-lang.org/issues/15628)].

About test for this.. we would need to simulate a ipv6 broken route, only way I know of is using `ip6tables` and that needs `sudo`.

Fixes #3641.
Closes #3463.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -354,6 +354,7 @@ lib/rubygems/config_file.rb
 lib/rubygems/core_ext/kernel_gem.rb
 lib/rubygems/core_ext/kernel_require.rb
 lib/rubygems/core_ext/kernel_warn.rb
+lib/rubygems/core_ext/tcpsocket_init.rb
 lib/rubygems/defaults.rb
 lib/rubygems/dependency.rb
 lib/rubygems/dependency_installer.rb
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -45,6 +45,7 @@ class Gem::ConfigFile
   DEFAULT_UPDATE_SOURCES = true
   DEFAULT_CONCURRENT_DOWNLOADS = 8
   DEFAULT_CERT_EXPIRATION_LENGTH_DAYS = 365
+  DEFAULT_IPv4_FALLBACK_ENABLED = false
 
   ##
   # For Ruby packagers to set configuration defaults.  Set in
@@ -141,6 +142,11 @@ class Gem::ConfigFile
   attr_accessor :cert_expiration_length_days
 
   ##
+  # Fallback to IPv4 when IPv6 is not reachable or slow (default: false)
+
+  attr_accessor :ipv4_fallback_enabled
+
+  ##
   # Path name of directory or file of openssl client certificate, used for remote https connection with client authentication
 
   attr_reader :ssl_client_cert
@@ -175,6 +181,7 @@ class Gem::ConfigFile
     @update_sources = DEFAULT_UPDATE_SOURCES
     @concurrent_downloads = DEFAULT_CONCURRENT_DOWNLOADS
     @cert_expiration_length_days = DEFAULT_CERT_EXPIRATION_LENGTH_DAYS
+    @ipv4_fallback_enabled = DEFAULT_IPv4_FALLBACK_ENABLED
 
     operating_system_config = Marshal.load Marshal.dump(OPERATING_SYSTEM_DEFAULTS)
     platform_config = Marshal.load Marshal.dump(PLATFORM_DEFAULTS)
@@ -203,6 +210,7 @@ class Gem::ConfigFile
     @disable_default_gem_server  = @hash[:disable_default_gem_server]  if @hash.key? :disable_default_gem_server
     @sources                     = @hash[:sources]                     if @hash.key? :sources
     @cert_expiration_length_days = @hash[:cert_expiration_length_days] if @hash.key? :cert_expiration_length_days
+    @ipv4_fallback_enabled       = @hash[:ipv4_fallback_enabled]       if @hash.key? :ipv4_fallback_enabled
 
     @ssl_verify_mode  = @hash[:ssl_verify_mode]  if @hash.key? :ssl_verify_mode
     @ssl_ca_cert      = @hash[:ssl_ca_cert]      if @hash.key? :ssl_ca_cert
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -45,7 +45,7 @@ class Gem::ConfigFile
   DEFAULT_UPDATE_SOURCES = true
   DEFAULT_CONCURRENT_DOWNLOADS = 8
   DEFAULT_CERT_EXPIRATION_LENGTH_DAYS = 365
-  DEFAULT_IPv4_FALLBACK_ENABLED = false
+  DEFAULT_IPV4_FALLBACK_ENABLED = false
 
   ##
   # For Ruby packagers to set configuration defaults.  Set in
@@ -142,6 +142,7 @@ class Gem::ConfigFile
   attr_accessor :cert_expiration_length_days
 
   ##
+  # == Experimental ==
   # Fallback to IPv4 when IPv6 is not reachable or slow (default: false)
 
   attr_accessor :ipv4_fallback_enabled
@@ -181,7 +182,7 @@ class Gem::ConfigFile
     @update_sources = DEFAULT_UPDATE_SOURCES
     @concurrent_downloads = DEFAULT_CONCURRENT_DOWNLOADS
     @cert_expiration_length_days = DEFAULT_CERT_EXPIRATION_LENGTH_DAYS
-    @ipv4_fallback_enabled = DEFAULT_IPv4_FALLBACK_ENABLED
+    @ipv4_fallback_enabled = ENV['IPV4_FALLBACK_ENABLED'] == 'true' || DEFAULT_IPV4_FALLBACK_ENABLED
 
     operating_system_config = Marshal.load Marshal.dump(OPERATING_SYSTEM_DEFAULTS)
     platform_config = Marshal.load Marshal.dump(PLATFORM_DEFAULTS)
--- a/lib/rubygems/core_ext/tcpsocket_init.rb
+++ b/lib/rubygems/core_ext/tcpsocket_init.rb
@@ -1,0 +1,49 @@
+require 'socket'
+
+module CoreExtensions
+  module TCPSocketExt
+    def self.prepended(base)
+      base.prepend Initializer
+    end
+
+    module Initializer
+      CONNECTION_TIMEOUT = 5
+      IPV4_DELAY_SECONDS = 0.3
+
+      def initialize(host, serv, *rest)
+        mutex = Mutex.new
+        addrs = []
+        cond_var = ConditionVariable.new
+
+        Addrinfo.foreach(host, serv, nil, :STREAM) do |addr|
+          Thread.report_on_exception = false if defined? Thread.report_on_exception = ()
+
+          Thread.new(addr) do
+            # give head start to ipv6 addresses
+            sleep IPV4_DELAY_SECONDS if addr.ipv4?
+
+            # raises Errno::ECONNREFUSED when ip:port is unreachable
+            Socket.tcp(addr.ip_address, serv, connect_timeout: CONNECTION_TIMEOUT).close
+            mutex.synchronize do
+              addrs << addr.ip_address
+              cond_var.signal
+            end
+          end
+        end
+
+        mutex.synchronize do
+          timeout_time = CONNECTION_TIMEOUT + Time.now.to_f
+          while addrs.empty? && (remaining_time = timeout_time - Time.now.to_f) > 0
+            cond_var.wait(mutex, remaining_time)
+          end
+
+          host = addrs.shift unless addrs.empty?
+        end
+
+        super(host, serv, *rest)
+      end
+    end
+  end
+end
+
+TCPSocket.prepend CoreExtensions::TCPSocketExt
--- a/lib/rubygems/core_ext/tcpsocket_init.rb
+++ b/lib/rubygems/core_ext/tcpsocket_init.rb
@@ -8,7 +8,7 @@ module CoreExtensions
 
     module Initializer
       CONNECTION_TIMEOUT = 5
-      IPV4_DELAY_SECONDS = 0.3
+      IPV4_DELAY_SECONDS = 0.1
 
       def initialize(host, serv, *rest)
         mutex = Mutex.new
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -18,12 +18,6 @@ module Gem::InstallUpdateOptions
   # Add the install/update options to the option parser.
 
   def add_install_update_options
-    add_option(:"Install/Update", '--ipv4-fallback-enabled',
-          'Gem repository directory to get installed',
-          'gems') do |value, options|
-      Gem.configuration.ipv4_fallback_enabled = value
-    end
-
     add_option(:"Install/Update", '-i', '--install-dir DIR',
                'Gem repository directory to get installed',
                'gems') do |value, options|
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -18,6 +18,12 @@ module Gem::InstallUpdateOptions
   # Add the install/update options to the option parser.
 
   def add_install_update_options
+    add_option(:"Install/Update", '--ipv4-fallback-enabled',
+          'Gem repository directory to get installed',
+          'gems') do |value, options|
+      Gem.configuration.ipv4_fallback_enabled = value
+    end
+
     add_option(:"Install/Update", '-i', '--install-dir DIR',
                'Gem repository directory to get installed',
                'gems') do |value, options|
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -6,7 +6,6 @@ require 'rubygems/s3_uri_signer'
 require 'rubygems/uri_formatter'
 require 'rubygems/uri_parsing'
 require 'rubygems/user_interaction'
-require 'rubygems/core_ext/tcpsocket_init'
 require 'resolv'
 
 ##
@@ -79,6 +78,7 @@ class Gem::RemoteFetcher
   #            fetching the gem.
 
   def initialize(proxy=nil, dns=nil, headers={})
+    require 'rubygems/core_ext/tcpsocket_init' if Gem.configuration.ipv4_fallback_enabled
     require 'net/http'
     require 'stringio'
     require 'uri'
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -6,6 +6,7 @@ require 'rubygems/s3_uri_signer'
 require 'rubygems/uri_formatter'
 require 'rubygems/uri_parsing'
 require 'rubygems/user_interaction'
+require 'rubygems/core_ext/tcpsocket_init'
 require 'resolv'
 
 ##
--- a/test/rubygems/test_gem_config_file.rb
+++ b/test/rubygems/test_gem_config_file.rb
@@ -41,6 +41,7 @@ class TestGemConfigFile < Gem::TestCase
     assert_equal true, @cfg.verbose
     assert_equal [@gem_repo], Gem.sources
     assert_equal 365, @cfg.cert_expiration_length_days
+    assert_equal false, @cfg.ipv4_fallback_enabled
 
     File.open @temp_conf, 'w' do |fp|
       fp.puts ":backtrace: true"
@@ -56,6 +57,7 @@ class TestGemConfigFile < Gem::TestCase
       fp.puts ":ssl_verify_mode: 0"
       fp.puts ":ssl_ca_cert: /etc/ssl/certs"
       fp.puts ":cert_expiration_length_days: 28"
+      fp.puts ":ipv4_fallback_enabled: true"
     end
 
     util_config_file
@@ -70,6 +72,14 @@ class TestGemConfigFile < Gem::TestCase
     assert_equal 0, @cfg.ssl_verify_mode
     assert_equal '/etc/ssl/certs', @cfg.ssl_ca_cert
     assert_equal 28, @cfg.cert_expiration_length_days
+    assert_equal true, @cfg.ipv4_fallback_enabled
+  end
+
+  def test_initialize_ipv4_fallback_enabled_env
+    ENV['IPV4_FALLBACK_ENABLED'] = 'true'
+    util_config_file %W[--config-file #{@temp_conf}]
+
+    assert_equal true, @cfg.ipv4_fallback_enabled
   end
 
   def test_initialize_handle_arguments_config_file
--- a/test/rubygems/test_gem_remote_fetcher.rb
+++ b/test/rubygems/test_gem_remote_fetcher.rb
@@ -962,6 +962,12 @@ PeIQQkFng2VVot/WAQbv3ePqWq07g1BBcwIBAg==
     end
   end
 
+  def test_tcpsocketext_require
+    with_configured_fetcher(":ipv4_fallback_enabled: true") do |fetcher|
+      refute require('rubygems/core_ext/tcpsocket_init')
+    end
+  end
+
   def with_configured_fetcher(config_str = nil, &block)
     if config_str
       temp_conf = File.join @tempdir, '.gemrc'


sonalkr132:
Bundler tests are failing. How do I run bundler tests in context of my rubygems changes?

deivid-rodriguez:
I think if you switch to the `bundler` folder, run `bin/rake spec:deps`, and then `bin/rspec <spec_thats_failing>`, it should properly detecting that bundler is a submodule of the rubygems repo, and run tests using the rubygems from the checked out repo. If it doesn't let me know!

bronzdoc:
You need to set the `RGV ` environment variable to the branch of rubygems you want to test against inside the bundler submodule @sonalkr132 and run `bin/rspec` ...

deivid-rodriguez:
I think it should just work because bundler's Rakefile already detects the rubygems checkout: https://github.com/bundler/bundler/blob/master/Rakefile#L9-L14. No need to set the `RGV` varible I think, because that will download rubygems, and in this case it's already checked out.

bronzdoc:
@deivid-rodriguez it didn't for me, i just tried it since i'm testing a patch in bundler against a patch in RG

sonalkr132:
Thanks for help guys. I have my tests passing now, but this implementation is adding 4-6 minutes to bundler test suit (has redudant connection checking).
I am trying a slightly different [implmentation](https://github.com/sonalkr132/rubygems/commit/3984f5e00649e8b42d826c5aaec4512240868ef0) and for that tests are failing in [build](https://travis-ci.org/sonalkr132/rubygems/jobs/504507837#L385) but not locally. Did you ever face this (some `-I...` stuff?) or is this a secret feature discouraging people from patching ruby classes? :sweat_smile: 

We can probably also use `resolve-replace` [here](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/request/http_pool.rb#L40) (start `make_connection` for each IP, with resolve-replace controlling dns) but using that seems more fragile to me than this patch.

deivid-rodriguez:
I'm so sorry I didn't follow up on the bundler tests troubleshooting, I forgot about this!

Happily, @sonalkr132 found his way through and also [documented it :heart:](https://github.com/rubygems/rubygems/pull/2680).

deivid-rodriguez:
> I am trying a slightly different implmentation and for that tests are failing in build but not locally. Did you ever face this (some -I... stuff?) or is this a secret feature discouraging people from patching ruby classes? 

Sorry for not following up, @sonalkr132. I'll have a look and see if I understand what's going on :+1:

deivid-rodriguez:
@sonalkr132 The test that are failing in your alternative patch seem to be failing only in CI because... [they only run in CI](https://github.com/rubygems/rubygems/blob/fa6e547330e97b417ed11f262d68d03c57abeeda/test/rubygems/test_bundled_ca.rb#L11), because of [being slow](https://github.com/rubygems/rubygems/commit/0cf5e2a802a808bfbf4d2def6a176e8fc7a9a154).

deivid-rodriguez:
> @deivid-rodriguez it didn't for me, i just tried it since i'm testing a patch in bundler against a patch in RG

@bronzdoc I looked into this and you're right. I noticed this when reopening #2696 and CI passed (it should've failed when being tested against `bundler`'s master as per the job added in #2695).

The problem is that currently we only test against a remote branch or tag, not against the checked out code of `rubygems`. So basically when we open any PR to rubygems, bundler's tests are not running against that PR, but against master.

As per the lines of code in bundler's Rakefile that I pointed out earlier: https://github.com/bundler/bundler/blob/master/Rakefile#L9-L14, this seems unintentional, so I'll fix it to make our specs against bundler a proper regression run and not just a sanity check against master.

deivid-rodriguez:
I created https://github.com/bundler/bundler/pull/7070 to fix this. If it's correct, we can start specifying `RGV=..` in rubygems to really test bundler against the proper copy of rubygems.

duckinator:
hey, what's the status of this? It'd be awesome to have IPv4 fallback in place.

florrain:
Hi @sonalkr132 👋  I second the previous comment, the fallback to ipv4 would be very helpful. Would you have an idea on the next steps here?

florrain:
I'm able to reproduce the IPv6 error using my company VPN that is for some reason not able to resolve rubygems.org. Rubygems.org using IPv4 works though. So I can assist with the testing if that can help getting this out of the door.

sonalkr132:
Hi, sorry about the delay in response here. 
The issue with this as of now is that monkey patching `TCPSocketExt` in rubygems is not far from what (or where) the fix should be. A significant number of bundler `real-world` tests (which are way too slow to run/debug) are failing with this implementation. I was also getting different failures on Travis and my local setup.

Now we are using GitHub ci and I have been meaning to update this with a slightly less intrusive change (which could pass the test suit), however, haven't been able to find time for it.

IPv6 not working for some users (particularly in Europe) is also a Fastly issue, where only sites hosted on Fastly are unreachable. We have filed a support ticket about this (so have other users) but no progress so far.
@florrain it would be helpful you can open a support ticket on help.rubygems.org and I can forward debugging details I need, which I will forward on our support ticket.

deivid-rodriguez:
Hello @sonalkr132!

Today I spent some time getting a ipv6 tunnel (ipv6 adoption in Spain is not really a thing) to that I can have ipv6, and then breaking it through `ip6tables` as you suggested. Timeouts on `gem install` started happening :+1:.

Then I rebased & installed this PR and timeouts were fixed :+1:. I also tried some `bundle install`'s and run `bundler` & `rubygems` specs, and all seems good.

Can you share the realworld bundler stuff that was not playing well with this?

sonalkr132:
This seems to be working, both on CI and locally. We may have fixed real-world tests in unrelated fixes.

As I hinted above, patching `TCPSocket` is not really an ideal fix. Fallback support should be inbuilt in `Net::HTTP`. Another downside of redundant connection checking (`Socket.tcp(addr.ip_address, serv, connect_timeout: CONNECTION_TIMEOUT).close` part) is that now timeout is 60 (default) + 5 seconds.

I am open to suggestions for better ways of doing this. If the current implementation seems acceptable, I can look into adding some bundler sudo tests.

deivid-rodriguez:
Nice!

I have one question. As per the RFC, should we also start parallel name resolutions per family, in case it's ipv6 name resolution that's faulty,? Possibly useful for this, in ruby 2.7.1, `Addrinfo` methods accept a `timeout` parameter: https://github.com/ruby/ruby/commit/0e9d56f5e73ed2fd8e7c858fdea7b7d5b905bb64.

Also, [a `:resolv_timeout` parameter might be coming the `TCPSocket.new`](https://github.com/ruby/ruby/pull/3472) but I don't think that one would be useful for us here.

deivid-rodriguez:
Regarding the right place to add this, should we instead patch `Net::HTTP` then? I mean, if `Net::HTTP` feels better, why did you go with monkeypatching `TCPSocket.new`. Was it easier?

deivid-rodriguez:
Oh, I now see what you mean by re-reading your ruby-core ticket, this should be patched in, you mean this should be patched in https://github.com/ruby/ruby/blob/4444025d16ae1a586eee6a0ac9bdd09e33833f3c/ext/socket/ipsocket.c#L42. That makes sense and I agree it'd be the cleanest place.

Also, regarding my comment about async DNS, your implementation is a good step forward anyways, it's just something I noticed. It'd be nice to get people having this issue in real life to try this to see if this version fixes their issues already.

I'll comment on your ruby-core ticket, maybe we'll get some feedback.

sonalkr132:
> should we also start parallel name resolutions per family, in case it's ipv6 name resolution that's faulty,? 

If we were an HTTP client library, we should definitely do this.
I think we should try to introduce the smallest possible change to fix our specific issue. Adding parallel DNS record resolution would mean we are taking over ruby's `Addrinfo` responsibility as well. As I mentioned in the other ticket, clients not able to resolve `AAAA` records specifically is not an issue we have. Our issue about the IPv6 route being broken is either because of misbehaving Fastly PoP or misconfiguration of IPv6 routes on the user's host/network.
Both `AAAA` and `A` records are served by the same set of nameservers and either it would respond to both queries or none.

I am sure you are aware that most RFCs are to be taken as suggestions on best practice and are not followed word for word in real implementation. [Here](https://github.com/golang/go/blob/master/src/net/dial.go#L447) is the golang implementation of the same, if you follow the trail of addr resolution you will end up [here](https://github.com/golang/go/blob/b60ec4cc4b230f4d0787acf82057947b8bf80cea/src/net/cgo_unix.go#L161), where they are just calling `getaddrinfo` of C (which returns all IP addresses sorted by class).

deivid-rodriguez:
Sure, please don't take this as a blocker from me, it was just something I noticed as I said previously.

sonalkr132:
Thank you for pointing it out. If the increase in timeout and patching a core ruby class seems like an acceptable fix for this, I will try to add tests for this a couple of days.

> I'll comment on your ruby-core ticket, maybe we'll get some feedback.

A review from someone who is familiar with all this would be very valuable.

deivid-rodriguez:
I'm fine with patching a core class and with the increase in the global timeout that you mentioned, as long as we fix the issue.

A test would be great too, and also a review from someone familiar with all this.

Something I would really like though if we're all good with doing this, is someone having this issue in real life to install this patch and confirm that it really fixes the problem and causes no issues whatsoever. Another approach we could take to get this battle tested before making it generally available is to ship this experimentally behind an environment variable, and point to users having this issue from now on to use it.

sonalkr132:
> ship this experimentally behind an environment variable, and point to users having this issue from now on to use it.

good idea. I will add a flag for this.

> , is someone having this issue in real life to install this patch and confirm that it really fixes the problem

Disabling ipv6 is much easier for the user than manually patching their installed gem. Perhaps they may do it if we release an rc with this.

deivid-rodriguez:
Great, let's ship this as opt-in first :+1:.

> Disabling ipv6 is much easier for the user than manually patching their installed gem. Perhaps they may do it if we release an rc with this.

Yeah, if I recall correctly, the instructions for disabling IPv6 where pretty simple. I guess this depends too on how clear and concise the instructions (and rollback instructions) for applying the patch are, and how much the users who report these issues are willing to help us :smiley:. In any case, `gem update --system --pre` is definitely more clear and concise than the alternative.

sonalkr132:
I have added `--ipv4-fallback-enabled` flag and config and tests for the config. Let me know if you can think of a better name. I don't want to call it happy eyeballs as reusing the TCP connection seems like a requirement to qualify. I had tried reusing the connection, and my attempt exists [here](https://github.com/rubygems/rubygems/commit/e1ecb3598dc57f65b81408f0ae139635cd66b4b5). I didn't pursue it as errors were beyond my comprehension and it felt extremely rickety. 

[Github action doesn't support IPv6 enabled hosts](https://github.com/actions/virtual-environments/issues/668), so a test for the functionality won't be possible.

sonalkr132:
I tried using the localhost server for testing this (lo has `::1` on github actions hosts), it didn't go that well either. 
Making requests to `localhost` always uses ipv4 for some reason. This means we have to use something other than `localhost` and point that to `::1`. We can't use `resolv-replace` as it overrides `TCPSocket.initialize` as well. I [tried using `/etc/host`](https://github.com/sonalkr132/rubygems/pull/4/commits/3e0630b7dcc101f9362e9ca4b6bd2b7669d4a955) and the test passes locally but not on CI.

deivid-rodriguez:
@sonalkr132 Thanks for all your efforts here. I can try to dedicate some time to those tests issues you're having, but if you didn't manage to figure this out, it's unlikely that I will. So I'd say it's fine to ship this with out "realworld tests".

Only thing I'm hesitant about is a new CLI flag. I take CLI flags as part of the public interface of our CLI, so I'd rather add new CLI flags only for conditional behaviour that it's put in there to stay. Since in this case we just want to easily enable experimenting with some behaviour that we would want to be the default in the future, I was suggesting an environment variable instead. However, maybe under some systems, passing environment variables to a command, or setting them permanently is not as easy/common as it is on UNIX? In that case, I'd be fine with adding an experimental CLI flag, with a clear note that the intention is to abruptly remove in the future without deprecation and make its behaviour the default, without possibility of opt-in out.

leonardo-jesus:
This is really useful, I would like this merged... I'm having problems with IPV6 fallback

deivid-rodriguez:
Great! Since you're here, could you install this branch and confirm whether try this PR works for you? You should:

```
git clone https://github.com/rubygems/rubygems
cd rubygems
git checkout -b sonalkr132-ipv6-fallback master
git pull https://github.com/sonalkr132/rubygems.git ipv6-fallback
ruby setup.rb
```

Then run `gem install` or whatever `gem`command you're having issues with with the `DEFAULT_IPV4_FALLBACK_ENABLED` set to `true`.

@sonalkr132 This is looking good. Comments that come to mind:

* Like `CLI` flags, I also tend to consider configuration keys as public interface. However, while rubygems will bail out on an unknown CLI  flag, it will happily proceed if unknow configuration settings are present in the gemrc file. So a setting marked as experimental that we can remove whenever we consider the new behaviour stable sounds great :+1:.

* Someone self-assigned https://bugs.ruby-lang.org/issues/15628, so that's some movement there :)

leonardo-jesus:
Hey @deivid-rodriguez, I'm having problems in the last step:
image

deivid-rodriguez:
You don't have the ruby executable in your PATH, that's weird. I guess you need to either install `ruby` (although if you're having issues with `rubygems` I assume you have it installed already), or to find where it is installed, and add that folder to your `PATH`.

leonardo-jesus:
Oooh sorry, I was in the wrong terminal, I was using git bash, but for Ruby, I use bash on CMD.

I repeat the process, and this is the results:
image

deivid-rodriguez:
Uh, that's bad! Which OS are you using? How did you install ruby?

leonardo-jesus:
It's Windows 10, I followed this tutorial: https://develop.zendesk.com/hc/en-us/community/posts/360037023114--Tutorial-ZAT-installation-on-Windows-10?page=1#community_comment_360013338254

But I had problems when need to start to use Ruby...

deivid-rodriguez:
Hei! Thanks for the information. Could you check which version of the `rubygems-integration` apt package are you using? `apt policy rubygems-integration` should give you that information. I think the error you're running into was [fixed in version `1.17.2` of that package](https://salsa.debian.org/ruby-team/rubygems-integration/-/commit/0f2c7dd60f3452a5c9a7f00388e5c735a18a4326), which has probably not yet been backported to ubuntu. Could you verify if this is the case, and if it is, we can patch your version of `rubygems-integration` to fix the issue and let you try this thing :sweat_smile:.

leonardo-jesus:
Thanks for the commands, yes I'm at 1.16 ;-;

image

What I have to do now?

deivid-rodriguez:
I'm not sure how you're supposed to use ruby after that tutorial, because on one hand it installs `rbenv` which is supposed to let you install and switch between several rubies under the same user account, but on the other hand it installs some gems using `sudo gem install <blabla>` using system ruby. Which `gem install` command is giving you errors? Is it using `sudo`, or without `sudo`.

If it's using `sudo`, then you'll need to uprade the `rubygems-integration` apt package to get past the error. I see that ubuntu 20.10 already has `rubygems-integration` 1.17.2, so you could upgrade the OS altogether if you wanted to with `do-release-upgrade`. Otherwise you could manually download the `.deb` package and install it manually and I think it should work. You can find links to download it [here](https://pkgs.org/download/rubygems-integration) and then you could run `sudo dpkg -i <downloaded_file>.deb` to install it.

All that said, we generally recommend installing and running rubies through rbenv rather than the OS version, so you might want to follow [this tutorial](https://pkgs.org/download/rubygems-integration) instead, and you wouldn't need any `rubygems-integration` hacky upgrades in that case.

Good luck and let me know how it goes!

leonardo-jesus:
The problem is when I try to use with sudo:
image

I will follow your tutorial, but I have some doubts, I send message at Bundler Slack, at your DM, can you answer me there, or you prefer here?

deivid-rodriguez:
Better here @leonardo-jesus.

akostadinov:
Hi, what is stopping this PR from progressing? Seems to be a long awaited capability.

sonalkr132:
> how do I know if issue is IPv4 fallback? 

are you able to access rubygems.org in the browser or with cli tools other than `gem`? You can try the following
```
# works
$ wget -4 https://www.rubygems.org

# fails. means ipv6 issue
$ wget -6 https://www.rubygems.org

# fails. ipv6 issue specific to your ISP or network
$ wget -6 https://google.com
```

Please feel free to mail at support@rubygems.org is you need further help.

akostadinov:
@sonalkr132 , see my message, there is no way I can access anything over IPv6 and it is disabled on all interfaces to make sure nothing can bind to an IPv6 address. Still since yesterday I'm hitting a lot of timeouts. I also don't see any IPv6 attempts on the network with Wireshark (but I still see AAAA DNS queries). I think unlikely to be the IPv6 issue so I guess I can try support@rubygems.org as you suggested.

Dan33l:
Hi,

We are very very interested by this fallback to IPv4 behavior.  It permits to ensure that rubygems.org can be always joined. Many others CLI tools have this behavior. We have some continuous integration jobs that are nightmare and that could be smoothly solved with the fallback to IPv4 when can not join rubygems.org via IPv6.

deivid-rodriguez:
Hi @Dan33l! This PR is waiting for someone to confirm that it fixes this situation. If I recall correctly, I tried it with @leonardo-jesus, and it didn't fix the timeouts for home although we were not able to confirm 100% whether the patch had been installed correctly, and in the end he lost access to the environment where this was reproducing.

akostadinov:
@deivid-rodriguez , I just tested this and it fixes the issue for me. I can't test my full builds because I somehow can't figure out how to build the gem and install on the CI system. Could you advise?

deivid-rodriguez:
Hi @akostadinov, I posted some instructions here: https://github.com/rubygems/rubygems/pull/2662#issuecomment-713435472. But I thought your issue was not related to ipv6 in the end? How did you verify this patch locally?

akostadinov:
@deivid-rodriguez , I did a local test with `ruby -Ilib bin/gem install` with single gems but apparently it due to luck or lack of attention I got confused. Now with full tests it became apparent that it doesn't help.

So yes - my problem doesn't seem to be IPv6 related and is not affected by this PR.

For anybody with container builds, here are the commands I added to Dockerfile to update rubygems before .

```
RUN git config --global user.name "Your Name" ; git config --global user.email "you@example.com"
RUN export GIT_MERGE_AUTOEDIT=no ; set -x ; git clone --single-branch https://github.com/rubygems/rubygems.git && cd rubygems && git checkout -b sonalkr132-ipv6-fallback master && git pull https://github.com/sonalkr132/rubygems.git ipv6-fallback
RUN set -x ; cd rubygems && gem --version && ruby setup.rb && gem --version
```

I'm sorry for the noise.

deivid-rodriguez:
Don't be sorry! On the contrary, thanks for reporting back so thoroughly!

deivid-rodriguez:
There's some movement here on the ruby-core side: https://github.com/ruby/ruby/pull/4038. :tada:

Let's see how that PR works out, and in case it's merged, we can start recommending monkeypatching `socket.rb` with the merged code instead of this PR I guess?

sonalkr132:
> we can start recommending monkeypatching socket.rb with the merged code instead of this PR I guess?

I was going to suggest that we can ship rubygems with the said patch, however two issues:
- the patch doesn't fix Net:HTTP
- given the nuances of the reviews the said patch is receiving, it seems like no simple fix.

deivid-rodriguez:
I was also planning to test the said patch with https://github.com/ruby/net-http/pull/10/ on top. Given that PR seems like a good idea and switches from ` TCPSocket.open` to `Socket.tcp`, it could fix `Net::HTTP` in this regard?

sonalkr132:
>  Given that PR seems like a good idea and switches from TCPSocket.open to Socket.tcp

Probably there is a good reason for TCPSocket.open and Socket.tcp being different and use of former in net:http. mmasaki said that he will fix `TCPSocket.open` separately. https://github.com/ruby/ruby/pull/4038#issuecomment-774398960

deivid-rodriguez:
https://github.com/ruby/net-http/pull/10 seems a very good idea to me, since it also fixes very noticiable inefficiencies in `net/http` usages, as reported in https://bugs.ruby-lang.org/issues/16381. Same fix is also proposed separately at https://github.com/ruby/ruby/pull/1806.

Anyways, yeah, I saw the comment that `TCPSocket.open` will also be fixed, was just thinking that if those proposals get attended first we won't need that. Also if this gets completely fixed in C land, that makes it easier for us to backport the fixes to rubygems.

deivid-rodriguez:
https://github.com/ruby/net-http/pull/10 was merged by @hsbt, so the currently proposed ruby-core implementation of happy eyeballs (which was also just updated to address feedback) might be enough now.

ddfreyne:
My IPv6 connectivity broke again today, which gave me the opportunity to test this out,

Initially:

```
% cat ~/.gemrc
gem: --no-document⏎

% gem update
ERROR:  While executing gem ... (Gem::RemoteFetcher::UnknownHostError)
    timed out (https://rubygems.org/specs.4.8.gz)
```

I added `:ipv4_fallback_enabled: true`:

```
% cat ~/.gemrc
gem: --no-document
:ipv4_fallback_enabled: true

% gem update
Updating installed gems
Updating aws-partitions
Fetching aws-partitions-1.427.0.gem
Successfully installed aws-partitions-1.427.0
Updating aws-sdk-core
Fetching aws-sdk-core-3.112.0.gem
Successfully installed aws-sdk-core-3.112.0
Updating aws-sdk-rails
Fetching aws-sdk-dynamodb-1.59.0.gem
Fetching aws-record-2.5.0.gem
Fetching aws-sessionstore-dynamodb-2.0.1.gem
Fetching aws-sdk-rails-3.6.0.gem
Fetching aws-sdk-sqs-1.36.0.gem
Successfully installed aws-sdk-dynamodb-1.59.0
Successfully installed aws-record-2.5.0
Successfully installed aws-sdk-sqs-1.36.0
Successfully installed aws-sessionstore-dynamodb-2.0.1
Successfully installed aws-sdk-rails-3.6.0
Updating aws-sdk-ses
Fetching aws-sdk-ses-1.37.0.gem
Successfully installed aws-sdk-ses-1.37.0
Updating brakeman
Fetching brakeman-5.0.0.gem
Successfully installed brakeman-5.0.0
Updating bundler
Fetching bundler-2.2.10.gem
Successfully installed bundler-2.2.10
Updating capybara
Fetching capybara-3.35.3.gem
Successfully installed capybara-3.35.3
Updating childprocess
Fetching childprocess-4.0.0.gem
Successfully installed childprocess-4.0.0
Updating devise
Fetching devise-4.7.3.gem
Successfully installed devise-4.7.3
Updating fugit
Fetching fugit-1.4.2.gem
Fetching raabro-1.4.0.gem
Successfully installed raabro-1.4.0
Successfully installed fugit-1.4.2
Updating puma
Fetching puma-5.2.1.gem
Successfully installed puma-5.2.1
Updating rack-attack
Fetching rack-attack-6.5.0.gem
Successfully installed rack-attack-6.5.0
Updating redis
Fetching redis-4.2.5.gem
Successfully installed redis-4.2.5
Updating rspec-rails
Fetching rspec-rails-4.0.2.gem
Successfully installed rspec-rails-4.0.2
Updating rubocop
Fetching rubocop-1.10.0.gem
Successfully installed rubocop-1.10.0
Updating rubygems-update
Fetching rubygems-update-3.2.10.gem
Successfully installed rubygems-update-3.2.10
Updating semantic_range
Fetching semantic_range-2.3.1.gem
Successfully installed semantic_range-2.3.1
Updating sentry-raven
Fetching sentry-raven-3.1.1.gem
Successfully installed sentry-raven-3.1.1
Updating sidekiq
Fetching sidekiq-6.1.3.gem
Successfully installed sidekiq-6.1.3
Updating web-console
Fetching web-console-4.1.0.gem
Successfully installed web-console-4.1.0
Gems updated: aws-partitions aws-sdk-core aws-record aws-sdk-dynamodb aws-sdk-rails aws-sdk-sqs aws-sessionstore-dynamodb aws-sdk-ses brakeman bundler capybara childprocess devise fugit raabro puma rack-attack redis rspec-rails rubocop rubygems-update semantic_range sentry-raven sidekiq web-console
```

I removed `:ipv4_fallback_enabled: true`:

```
% cat ~/.gemrc
gem: --no-document

% gem update
ERROR:  While executing gem ... (Gem::RemoteFetcher::UnknownHostError)
    timed out (https://rubygems.org/specs.4.8.gz)
```

So it looks like this works as intended!

deivid-rodriguez:
Nice!!!

deivid-rodriguez:
Should we move this forward then @sonalkr132?

sonalkr132:
Yeah, I think so. This is disabled by default, so it can hardly do any harm. We can document it as a workaround for now and eventually migrate to fix in https://github.com/ruby/ruby/pull/4038, which is much more comprehensive. I am not sure if we would ever enable current flag by default.

@ddfreyne please do open a ticket on https://support.fastly.com and add support@rubygems.org or my email in cc. Make sure that your connection issue is only for fastly IPs or any IPv6 address. 
```
wget -6 rubygems.org # fails
wget -6 www.google.com # works
```

ddfreyne:
Sadly the IPv6 problem lies entirely on my ISP’s side, not Fastly’s. I get an IPv6 address, which works intermittently: occasionally all IPv6 traffic, whether to Google or Fastly or anything else, gets dropped and thus yields timeouts.

deivid-rodriguez:
Can you rebase it @sonalkr132?

sonalkr132:
rebased.

deivid-rodriguez (approved):
I'm merging this after CI is it looks ok to you @sonalkr132.

sonalkr132:
yes, please go ahead. should we add a guide/blog for this? seems like it can use a page for context around the issue in Net:HTTP and error message.

simi (approved):

akostadinov:
Which version will this fall into?

simi:
@akostadinov most likely 3.2.11

deivid-rodriguez:
Yes, I'll release it tomorrow hopefully. And yes, a guide/blog would be cool, an easy place to link to so we don't need to repeat the same instructions again and again when people run into this.

benlangfeld (commented):
Would it not be better to fix net/http to use `Socket.connect_nonblock` than to spawn all these threads?

deivid-rodriguez:
It sounds like you want to report an issue/suggestion to https://github.com/ruby/net-http?

benlangfeld:
@deivid-rodriguez I was asking a question.

PR #2672, opened 2019-03-07
Ignore to handle the different platform

hsbt:
# Description:

I set `GEM_HOME` to `~/.gem` and invoke `gem i jruby-launcher` with JRuby. After that, I swicthed CRuby and invoke `gem -v` or `gem pristine --extension`, I got the warnings and error message about `jruby-launcher`.

The current RubyGems do handle the extensions of different platform. It should be ignore or skip.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -71,8 +71,11 @@ class Gem::BasicSpecification
     elsif missing_extensions?
       @ignored = true
 
-      warn "Ignoring #{full_name} because its extensions are not built. " +
-        "Try: gem pristine #{name} --version #{version}"
+      if platform == RUBY_ENGINE
+        warn "Ignoring #{full_name} because its extensions are not built. " +
+          "Try: gem pristine #{name} --version #{version}"
+      end
+
       return false
     end
 
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -71,7 +71,7 @@ class Gem::BasicSpecification
     elsif missing_extensions?
       @ignored = true
 
-      if platform == RUBY_ENGINE
+      if RUBY_ENGINE == platform || Gem::Platform.local === platform
         warn "Ignoring #{full_name} because its extensions are not built. " +
           "Try: gem pristine #{name} --version #{version}"
       end
--- a/lib/rubygems/commands/pristine_command.rb
+++ b/lib/rubygems/commands/pristine_command.rb
@@ -104,7 +104,7 @@ extensions will be restored.
               end.flatten
             end
 
-    specs = specs.select{|spec| spec.platform == RUBY_ENGINE }
+    specs = specs.select{|spec| RUBY_ENGINE == spec.platform || Gem::Platform.local === spec.platform }
 
     if specs.to_a.empty?
       raise Gem::Exception,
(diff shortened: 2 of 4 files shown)


lib/rubygems/commands/pristine_command.rb
```diff
@@ -104,6 +104,8 @@ def execute
               end.flatten
             end
 
+    specs = specs.select{|spec| spec.platform == RUBY_ENGINE }
```
deivid-rodriguez: I see that normally the check for the current platform is done using `Platform.local`. For example, 

https://github.com/rubygems/rubygems/blob/8d5490364ac3a50eecefe880cbf1201ba355dfb0/lib/rubygems/resolver/installer_set.rb#L65

or

https://github.com/rubygems/rubygems/blob/8d5490364ac3a50eecefe880cbf1201ba355dfb0/lib/rubygems/resolver/installer_set.rb#L65

Should the same check be applied here or it's better to use `RUBY_ENGINE`?

test/rubygems/test_gem_commands_pristine_command.rb
```diff
@@ -514,6 +514,47 @@ def test_execute_default_gem
...
+    install_gem a
+    install_gem b
+
+    @cmd.options[:extensions]     = true
+    @cmd.options[:extensions_set] = true
+    @cmd.options[:args] = []
+
+    use_ui @ui do
+      @cmd.execute
+    end
```
deivid-rodriguez: You can probably wrap the above block with a `util_set_arch` to a platform that's different from `java`, so that this test doesn't accidentally break if we ever add a `jruby` CI.

lib/rubygems/commands/pristine_command.rb
MSP-Greg: Comparisons of `s.platform` also involves the issue of how the `--platform` option affects gem selection.  There seems to be an assumption that `--platform=ruby` is the only valid use, probably based on the fact that platform specific gems are chosen before 'ruby' gems.

test/rubygems/test_gem_commands_pristine_command.rb
hsbt: Good catch, I did update it.

hsbt:
@bundlerbot r+

lib/rubygems/basic_specification.rb
```diff
@@ -71,8 +71,11 @@ def contains_requirable_file?(file)
     elsif missing_extensions?
       @ignored = true
 
-      warn "Ignoring #{full_name} because its extensions are not built. " +
-        "Try: gem pristine #{name} --version #{version}"
+      if RUBY_ENGINE == platform || Gem::Platform.local === platform
```
eregon, line 74: This `RUBY_ENGINE == platform` check doesn't seem right, and causes a failure when running the tests on TruffleRuby.
It should be `Gem::Platform::RUBY == platform`, right?
(as in the RUBY platform, the universal platform independent of the local platform)

cc @bjfish

deivid-rodriguez: I suggested to replace this condition with `Gem::Platform.local === platform` in https://github.com/rubygems/rubygems/pull/2672#discussion_r263358460, but I think @hsbt misunderstood my suggestion and added the condition instead.

eregon: So it should be just `if Gem::Platform.local === platform` then?
Will a not-precompiled locally installed and compiled gem report a non-RUBY platform?

eregon: https://github.com/rubygems/rubygems/pull/3495

PR #2674, opened 2019-03-07
Deprecate a few unused methods

deivid-rodriguez:
# Description:

These three methods in `DependencyInstaller` don't seem used anwhere, and two of them are also marked as `:nodoc:`. I'd like to deprecate them, so we can safely remove the code in the future, and also to mark that we don't really need to maintain or add fixes to these methods.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/dependency_installer.rb
+++ b/lib/rubygems/dependency_installer.rb
@@ -175,6 +175,7 @@ class Gem::DependencyInstaller
 
     @available.pick_best!
   end
+  deprecate :available_set_for, :none, 2019, 12
 
   ##
   # Indicated, based on the requested domain, if local
@@ -266,6 +267,7 @@ class Gem::DependencyInstaller
 
     set
   end
+  deprecate :find_gems_with_sources, :none, 2019, 12
 
   ##
   # Finds a spec and the source_uri it came from for gem +gem_name+ and
@@ -312,6 +314,7 @@ class Gem::DependencyInstaller
 
     @available = set
   end
+  deprecate :find_spec_by_name_and_version, :none, 2019, 12
 
   ##
   # Gathers all dependencies necessary for the installation from local and
--- a/lib/rubygems/dependency_installer.rb
+++ b/lib/rubygems/dependency_installer.rb
@@ -143,7 +143,9 @@ class Gem::DependencyInstaller
                   end
         end
 
-        results = find_gems_with_sources(dep)
+        results = Gem::Deprecate.skip_during do
+          find_gems_with_sources(dep)
+        end
 
         results.sorted.each do |t|
           to_do.push t.spec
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Flaky! On it!

lib/rubygems/dependency_installer.rb
```diff
@@ -266,6 +267,7 @@ def find_gems_with_sources(dep, best_only=false) # :nodoc:
 
     set
   end
+  deprecate :find_gems_with_sources, :none, 2019, 12
```
bronzdoc: this is used in `add_found_dependencies` method

deivid-rodriguez:
Did someone restart the build? I forgot to note down the flaky seed, so I'll have a look next time it appears.

bronzdoc:
> Did someone restart the build? I forgot to note down the flaky seed, so I'll have a look next time it appears.

it was me, the build failure seemed it wasn't related to the changes...

deivid-rodriguez:
Yeah, it was not related at all.

I was planning to fix it on a separate PR anyways. But it's no problem, I'll take care of it the next time the flaky appears :+1:.

bronzdoc:
this will brake bundler test against rubygems tho, isn't it?

deivid-rodriguez:
mmmm... why? Is bundler using this code?

bronzdoc:
> I was planning to fix it on a separate PR anyways. But it's no problem, I'll take care of it the next time the flaky appears 👍.

Ahh my bad!

deivid-rodriguez:
No problem at all!

bronzdoc:
> mmmm... why? Is bundler using this code?

Ahh nvm i gave it a quick look at the `ruby-head` failure and thought it was related, but i just gave it a closer look at it doesn't 👍

lib/rubygems/dependency_installer.rb
deivid-rodriguez: Yeah, but `add_found_dependencies` is deprecated too!

deivid-rodriguez: We should skip deprecations when `add_found_dependencies` calls this method though. I'll add that change.

MSP-Greg: It's also called on line 173 of:
https://github.com/rubygems/rubygems/blob/8d5490364ac3a50eecefe880cbf1201ba355dfb0/lib/rubygems/dependency_installer.rb#L167-L177

But that method (`available_set_for`) seems to only be called by `test_gem_dependency_installer.rb`...

deivid-rodriguez: Yeah, I didn't count the unit tests for these methods as actual usages of them.

I added the change to not print deprecations to users for methods they are not using directly :point_right: 04719726cba0d2840956d30ff2901f462a69f0a7.

deivid-rodriguez:
This one should be ready now I think!

bronzdoc:
@bundlerbot r+

PR #2676, opened 2019-03-08
Fix metadata description docs

sonalkr132:
# Description:

> settable by gem authors without requiring an update to the rubygems software.

this has never been true. Not sure why it said that.
Using G in the gem is converting it to link on using `rake spec_guide` ([broken here](https://github.com/rubygems/guides/commit/2ca58b0a29791567d3e997bf3c5c7ebb73b54364#diff-a0ed24268b853e4412f766326125de00L12)).

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -17,7 +17,7 @@ require 'rubygems/util/list'
 require 'stringio'
 
 ##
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

PR #2679, opened 2019-03-10
Removed circular require

nobu:
# Description:
rubygems/text.rb requires rubygems.rb but also is required by rubygems.rb via rubygems/user_interaction.rb.
It doesn't make sense to require rubygems/text.rb alone, `require` in it seems useless.
______________

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2687, opened 2019-03-19
Simplify Rakefile

hsbt:
# Description:

I removed some duplicated tasks in the current Rakefile.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,7 +10,7 @@ task :setup => ["bundler:checkout"] do
   gemspec = Gem::Specification.load(File.expand_path("../rubygems-update.gemspec", __FILE__))
 
   gemspec.dependencies.each do |dep|
-    sh "gem install '#{dep.name}:#{dep.requirement.to_s}'"
+    sh "gem install '#{dep.name}:#{dep.requirement.to_s}' --conservative --no-document"
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -46,11 +46,6 @@ RDoc::Task.new :rdoc => 'docs', :clobber_rdoc => 'clobber_docs' do |doc|
   doc.rdoc_dir = 'doc'
 end
 
-desc "Install gems needed to run the tests"
-task :install_test_deps => :clean do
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -46,11 +46,6 @@ RDoc::Task.new :rdoc => 'docs', :clobber_rdoc => 'clobber_docs' do |doc|
   doc.rdoc_dir = 'doc'
 end
 
-desc "Install gems needed to run the tests"
-task :install_test_deps => :clean do
-  sh "gem install minitest -v '~> 5.0'"
```
deivid-rodriguez: The other place where this is defined is 

https://github.com/rubygems/rubygems/blob/a0b3d351043e824819b3d66ce3e7d6958b4ba252/util/ci#L55

Right?

Maybe we can install move that line here and call the rake task from there?

Rakefile
```diff
@@ -352,11 +347,6 @@ end
 
 # Misc Tasks ---------------------------------------------------------
 
-desc "Cleanup trailing whitespace"
-task :whitespace do
-  system 'find . -not \( -name .svn -prune -o -name .git -prune \) -type f -print0 | xargs -0 sed -i "" -E "s/[[:space:]]*$//"'
-end
-
```
deivid-rodriguez: Why is this task duplicated? You mean by rubocop? Technically this task works on all project files, so if we add some traling whitespace to the README, this task will remove it.

hsbt: Fmm, Your comment is right. But I never invoke this task in the past releases.

Rakefile
hsbt: I update it for using `rake setup` in `util/ci`

Rakefile
deivid-rodriguez: Right. We should either remove it or use it, yeah. Maybe add it to CI? I don't have a strong opinion, whatever you think it's best works for me :+1:.

hsbt:
@bundlerbot r+

PR #2694, opened 2019-03-23
Revert downgrading bundler mismatches to a warning

deivid-rodriguez:
This reverts commit dad8d252f2b1562c4517fa6d73eb39dfbd0c4f37, reversing changes made to b4063c6028fe5f31277ad8b9a4e6f652ca453edd.

# Description:

#2621 broke bundler's build. I'm going to investigate why but in the meantime I don't want to have a red master in bundler so I'm reverting this.

I wonder if we should test rubygems against a vendored bundler that contains the code in bundler's master. Probably too much overhead :man_shrugging:.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -268,10 +268,17 @@ module Gem
 
     return loaded if loaded && dep.matches_spec?(loaded)
 
-    spec = dep.spec_for_exe(exec_name)
+    specs = dep.matching_specs(true)
 
-    unless spec
+    specs = specs.find_all do |spec|
+      spec.executables.include? exec_name
+    end if exec_name
+
+    unless spec = specs.first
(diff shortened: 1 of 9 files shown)


hsbt:
👌 

We should unify the rubygems and bundler like rails/rails repository and use the master branches both.

deivid-rodriguez:
Yes. We definitely should...

deivid-rodriguez:
Maybe we can add a single job to test this for now like in here https://github.com/rubygems/rubygems/commit/d9a44412f6a00ef0bb8fcc70347d76e9f2b8b6bd. A bit of overhead but not too much.

deivid-rodriguez:
@bundlerbot r+

PR #2695, opened 2019-03-23
Add a job to test bundler's master

deivid-rodriguez:
# Description:

#2621 broke our bundler's master build without us noticing in advance. I'd like to add one job that would've allowed us to catch this.

# Tasks:

- [ ] Write tests

### DIFF
--- a/util/ci
+++ b/util/ci
@@ -57,6 +57,8 @@ when %w(before_script)
     run('gem', %w(list --details))
     run('gem', %w(env))
   else
+    run('git', %W(reset --hard origin/#{ENV["BDV"]})) if ENV["BDV"]
+
     # Fix incorrect default gem specifications on ruby 2.6.1. Can be removed
     # when 2.6.2 is released and we start testing against it. See
     # https://bugs.ruby-lang.org/issues/15582 for more information
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ rvm:
 jobs:
   include:
     - stage: linting
-      rvm: 2.6
+      rvm: 2.6.1
       env:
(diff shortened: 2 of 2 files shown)


util/ci
```diff
@@ -57,6 +57,8 @@ when %w(before_script)
     run('gem', %w(list --details))
     run('gem', %w(env))
   else
+    run('git', %W(reset --hard origin/#{ENV["BDV"]})) if ENV["BDV"]
```
bronzdoc: What `BDV` stands for?

deivid-rodriguez: `BUNDLER_VERSION`, wanted to be consistent with the `RGV` variable. But being consistent in crypticness is not very important :rofl:.

We can't use `BUNDLER_VERSION` though since that variable is used by `bundler` and `rubygems` for other things.

deivid-rodriguez:
Great, build failed, that means it's working :)

util/ci
bronzdoc: haha, yeah 🤔  ok lets keep it like that, i guess naming it `BV` make as much sense as `BDV` 🤷‍♂️

deivid-rodriguez: Yeah, I don't really have a strong opinion. Maybe we can also use something more clear and verbose like `BUNDLER_CHECKOUT`?

bronzdoc: Now that i know what `BDV` stands for  i'm ok with it for consistency with `RGV. Maybe we can make both more clear and verbose in the future.

deivid-rodriguez: Sounds good!

deivid-rodriguez:
@bundlerbot r=bronzdoc

PR #2700, opened 2019-03-29
Added TruffleRuby to Travis CI

hsbt:
# Description:

/cc @eregon 
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,9 @@ jobs:
       rvm: jruby-9.2.7.0
       env: "TEST_TOOL=rubygems"
     - stage: test
+      rvm: truffleruby-19.0.0
+      env: "TEST_TOOL=rubygems"
+    - stage: test
       rvm: ruby-head
       env: "TEST_TOOL=rubygems"
     - stage: test


eregon:
https://travis-ci.org/rubygems/rubygems/jobs/512963782 shows an early failure, I'll debug it.
Currently, TruffleRuby tests the in-tree RubyGems from MRI 2.6.2.

eregon:
I'm working on this branch to debug: https://github.com/rubygems/rubygems/compare/master...eregon:test-truffleruby

eregon:
Interesting, it looks like even on MRI, `bundler-unload` is still installed after the `gem uninstall` in `tool/ci`: https://travis-ci.org/eregon/rubygems/jobs/512971591

eregon:
Currently the failure is due to TruffleRuby not yet supporting `Dir.glob(glob, base: base)`.

deivid-rodriguez:
@eregon Where is that used? I recall wanting to use that feature on https://github.com/bundler/bundler/pull/6985, but realizing that it didn't have the support range we needed (I think it's 2.5+ only).

eregon:
@deivid-rodriguez https://github.com/rubygems/rubygems/blob/78022d6a1136007dab6d8e4dadc8b0f229d0158c/lib/rubygems/util.rb#L127

deivid-rodriguez:
Aha, it has a version guard, that makes sense.

eregon:
I wish it could be a feature check, so it would work even if implementations don't have `Dir.glob(base:)` and yet are 2.5+, but that's probably not that easy or nice to do.

deivid-rodriguez:
I guess we can check the arity, or maybe add an explicit check for truffle ruby for the time being? I understand if supporting this feature is not on your immediate roadmap but you still want truffle ruby tested here for regressions, so I'm :+1: for doing something about it.

eregon:
> I guess we can check the arity, or maybe add an explicit check for truffle ruby for the time being? I understand if supporting this feature is not on your immediate roadmap but you still want truffle ruby tested here for regressions, so I'm  for doing something about it.

Checking arity is not enough unfortunately, as this is a core method. I think one way would be to call it and see if it raises on, e.g., `Dir.glob(__FILE__, base: Dir.pwd)`. But even then, that doesn't trigger an error in TruffleRuby because there is a fast path if the `glob` pattern exists as a file. `Dir.glob(__FILE__[0...-1] + "*", base: Dir.pwd)` triggers it but becomes a bit silly.

We plan to implement it soon in TruffleRuby, but in general it's always better to do feature checks unless really hard like here, so support for other implementations doesn't need to way for all features to be implemented and the next release of the implementation.

deivid-rodriguez:
Yeah, that makes sense. Up to you really, maybe we can add an explicit check for truffle ruby in this case, or do that cumbersome feature check, or wait till you implemente the feature. I don't have a strong opinion either way, whatever works best for you I'd say.

eregon:
> maybe we can add an explicit check for truffle ruby in this case

I try to avoid that, because this becomes then obsolete as soon as the bug is fixed.

> I don't have a strong opinion either way, whatever works best for you I'd say.

Thanks, in this case we'll just implement it in TruffleRuby, that's the cleanest and we'll have a release soon.

eregon:
TruffleRuby 1.0.0-rc15 contains the fix for `Dir.glob(base:)`, so tests now start:
https://travis-ci.org/eregon/rubygems/jobs/517971345

hsbt:
@eregon I rebased & pushed with the current master. But it still failed.

https://travis-ci.org/rubygems/rubygems/jobs/519844085

deivid-rodriguez:
Looks better though, at least there's a fair amount of dots in there :)

eregon:
@hsbt Thanks. Yes, I saw it as well, I need to investigate why running the tests results in an `OutOfMemoryError`.

eregon:
Locally, with latest TruffleRuby, I could run the RubyGems tests with `rake test TESTOPTS="-s0 -v"` and got:
```
2097 runs, 6487 assertions, 14 failures, 10 errors, 5 skips
```
So that's kind of good, only 24 tests fail out of 2097.

However, on TravisCI it's still failing with out-of-memory errors, I'm going to debug that further.

hsbt:
@eregon I updated this branch with `truffleruby-19.0.0`. But It's still failing status.

eregon:
@hsbt I opened https://github.com/rubygems/rubygems/pull/2797 where I have a few fixes that allow to execute more tests, and I can easily push more commits.
So I propose to close this PR and continue work there.

PR #2702, opened 2019-03-29
Use CI fix in bundler's master

deivid-rodriguez:
# Description:

So that PRs get tested against the checked out rubygems copy instead of rubygems master.

Note that we can only do this on the bundler master build, because the required changes are not available in the vendored copy of bundler. I think that's good enough for now.

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2706, opened 2019-03-29
Forbid `find_spec_for_exe` without an `exec_name`

deivid-rodriguez:
# Description:

With this we can make `Gem.bin_path` and `Gem.activate_bin_path` behave exactly the same regarding this parameter, and it allows us to simplify bundler's integration since the current version handles the case where no executable is passed, defaulting to `Gem::Specification.default_executable` in that case, which is a deprecated rubygems method anyways.

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

Issue #2701, opened 2019-03-29
RubyGems is silent when trying to uninstall a gem outside of the gem home


eregon:
```
$ gem env                                                                     
RubyGems Environment:
  - RUBYGEMS VERSION: 3.0.3
  - RUBY VERSION: 2.6.2 (2019-03-13 patchlevel 47) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/eregon/.gem/ruby/2.6.2
  - USER INSTALLATION DIRECTORY: /home/eregon/.gem/ruby/2.6.0
  - RUBY EXECUTABLE: /home/eregon/.rubies/ruby-2.6.2/bin/ruby
  - GIT EXECUTABLE: /usr/bin/git
  - EXECUTABLE DIRECTORY: /home/eregon/.gem/ruby/2.6.2/bin
  - SPEC CACHE DIRECTORY: /home/eregon/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /home/eregon/.rubies/ruby-2.6.2/etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /home/eregon/.gem/ruby/2.6.2
     - /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - "gem" => "--no-document"
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /home/eregon/.gem/ruby/2.6.2/bin
     - /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0/bin
     - /home/eregon/.rubies/ruby-2.6.2/bin
     - /home/eregon/bin
     - /usr/local/bin
     - /usr/local/sbin
     - /usr/bin
     - /usr/sbin
     - /home/eregon/code/mx
     - /home/eregon/code/mspec/bin
     - /opt/crystal/bin
$ gem install --install-dir /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0 path
Successfully installed path-2.0.1
1 gem installed
$ gem uninstall path
# BUG: Nothing printed and it didn't remove the gem!
$ gem uninstall --debug path
NOTE:  Debugging mode prints all exceptions even when rescued
[...]

PR #2707, opened 2019-03-29
Unswallow uninstall error

deivid-rodriguez:
# Description:

Sometimes `gem uninstall <gem_name>` fails to uninstall the gem, but does not give any errors. See #2701.

This error swallowing seems intentionally introduced specifically for the `--all` option in https://github.com/rubygems/rubygems/commit/564ffe277b7604304794b1eb008fe6e6d107743b, but I don't think this is good. In that case, `--all` is leaving gems around so the user should be informed too?

Maybe `gem uninstall` should try all paths in `GEM_PATH` instead?

Not sure, but the current behavior is no good.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/uninstall_command.rb
+++ b/lib/rubygems/commands/uninstall_command.rb
@@ -180,8 +180,6 @@ that is a dependency of an existing gem.  You can use the
 
   def uninstall_gem(gem_name)
     uninstall(gem_name)
-  rescue Gem::InstallError
-    nil
   rescue Gem::GemNotInHomeException => e
     spec = e.spec
     alert("In order to remove #{spec.name}, please execute:\n" +
--- a/test/rubygems/test_gem_commands_uninstall_command.rb
+++ b/test/rubygems/test_gem_commands_uninstall_command.rb
@@ -307,7 +307,7 @@ class TestGemCommandsUninstallCommand < Gem::InstallerTestCase
     gemhome2 = "#{@gemhome}2"
 
     a_4, = util_gem 'a', 4
-    install_gem a_4, :install_dir => gemhome2
+    install_gem a_4
 
     Gem::Specification.dirs = [@gemhome, gemhome2]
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
If we want to get this in quickly without changing any behavior, I can also scope the swallowing to the `--all` option only.

deivid-rodriguez:
We should fix the culprit of the error though, I'll try to figure out why this is succeeding under some version manager and failing on others.

deivid-rodriguez:
Ok. So the explanation is that by default gem uninstall tries to uninstall from `GEM_HOME` _and_ `Gem.user_dir`.

On a bare ruby installation, `Gem.paths.path - [Gem.paths.home] == [Gem.user_dir]` is true, so the uninstall succeeds. But on `rvm` these are three different directories for some reason.

eregon:
@deivid-rodriguez It's not urgent, I have a workaround, let's find a proper fix.

eregon:
I think in all cases we'd want error messages to at least be printed to let the user know, so this fix looks good to me.

eregon:
> Ok. So the explanation is that by default `gem uninstall` tries to uninstall from `GEM_HOME` _and_ `Gem.user_dir`.
> 
> On a bare ruby installation, `Gem.paths.path - [Gem.paths.home] == [Gem.user_dir]` is true, so the uninstall succeeds. But on `rvm`, `Gem.user_dir` is a different directory not included in `GEM_PATH`, so uninstall fails because the uninstall command never looks in the proper location.

That's surprising. Maybe `Gem.user_dir` shouldn't be considered here, because we want to install/uninstall from some directory of the `GEM_PATH`/`Gem.path`, but not outside of that, isn't it? (except maybe if specifically requested with `--install-dir`)
Or alternatively, if `Gem.user_dir` maybe it should always be part of `Gem.path` then?

deivid-rodriguez:
Yeah, apparently this is optional

https://github.com/rubygems/rubygems/blob/0fa3165af080d432a2f879a097ecbe82c8f83293/lib/rubygems/commands/uninstall_command.rb#L55-L60

And defaults to true

https://github.com/rubygems/rubygems/blob/0fa3165af080d432a2f879a097ecbe82c8f83293/lib/rubygems/commands/uninstall_command.rb#L18

I'm not really sure how `Gem.user_dir` is used in other parts of rubygems. I was definitely surprised too when I found out about it being considered, year. But I was not surprised when `gem uninstall` succeeded while trying to reproduce this issue. So, I guess my naive expectation is that all paths in `GEM_PATH` are considered, not only `GEM_HOME`.

deivid-rodriguez:
Let's do this then!

@bundlerbot r=hsbt

eregon:
Thanks for the fix!

aried3r:
Hey, we're currently facing CI issues in https://github.com/rails/webpacker/issues/2243. I know this is because we cannot (yet) support bundler 2.0 in that repo in all cases and thus is not directly a bug, but it's still backwards incompatible behavior, in case the CLI is considered public API.

I guess we're hitting this new behavior [here](https://travis-ci.org/rails/webpacker/jobs/573902360#L748-L751) and can just `|| true` in this case? I don't necessarily advocate for rolling back this PR, since it actually fixes bugs, we were just caught off-guard by the change.

deivid-rodriguez:
Yeah, @aried3r, I'm sorry we broke your CI...

There's always a fine line between backwards incompatibility and bug fixes, because every change in the code, even bug fixes, change the behaviour of programs. In this case, we've chosen to consider this a bug fix, and thus ship it in a patch release.

If you want your CI to be locked, and only upgrade whenever you want to / need to, you can lock the rubygems version in your CI with `gem update --system 3.0.6`.

Hope it helps!

aried3r:
No worries, just wanted to make sure we actually know _why_ it happened. Thanks for your quick reply and your and everybody else's work on rubygems!

PR #1239, opened 2015-05-01
[Specification] Also reset specs_by_name when setting all

segiddins:
\c @tenderlove @indirect

PR #2711, opened 2019-04-02
Let `@@stubs_by_name` to be incrementally populated again.

voxik:
Originally, the call to `.stubs_for` allowed to incrementally populate the `@@stubs_by_name` (especially see the `"#{name}-*.gemspec"` pattern in 4fa03bb7aac9f25f44394e818433fdda9962ae8d). Now it looks like it expects that all stubs are loaded, but the `.stubs_for` still matches the .gemspec files by the `name` pattern:

https://github.com/rubygems/rubygems/blob/6d45e0f7ac1caca22900e39f703e226c4cfdebf7/lib/rubygems/specification.rb#L845

I think this was done by mistake incrementally by PR #1239 and 4cee8ca9199ac7b3ab8647e0b78615f55d3eb02b. I think the best option is to get back to the original implementation, to let RubyGems incrementally populate the array. Other option would be to replace the logic in `.stub_for` by call to `.stubs`, but the means the performance improvement from the original commit was lost.

@tenderlove @segiddins WDYT?

BTW it would be nice if the PRs and commit messages gave more explanation why they were done. Now it is just guesswork what was the intention.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -830,8 +830,8 @@ class Gem::Specification < Gem::BasicSpecification
   # only returns stubs that match Gem.platforms
 
   def self.stubs_for(name)
-    if @@stubs
-      @@stubs_by_name[name] || []
+    if @@stubs_by_name[name]
+      @@stubs_by_name[name]
     else
       pattern = "#{name}-*.gemspec"
       stubs = Gem.loaded_specs.values +
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1193,6 +1193,11 @@ dependencies: []
     assert_equal ['b-1'], Gem::Specification.stubs_for('b').map { |s| s.full_name }
     assert_equal 2, Gem::Specification.class_variable_get(:@@stubs_by_name).length
 
+    assert_equal(
+      Gem::Specification.stubs_for('a').map { |s| s.object_id },
+      Gem::Specification.stubs_for('a').map { |s| s.object_id }
+    )
+
     Gem.loaded_specs.delete 'a'
     Gem.loaded_specs.delete 'b'
     Gem::Specification.class_variable_set(:@@stubs, nil)
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1179,6 +1179,25 @@ dependencies: []
     Gem::Specification.class_variable_set(:@@stubs, nil)
   end
 
+  def test_self_stubs_for_lazy_loading
+    Gem.loaded_specs.clear
+    Gem::Specification.class_variable_set(:@@stubs, nil)
+
+    dir_standard_specs = File.join Gem.dir, 'specifications'
+
+    save_gemspec('a-1', '1', dir_standard_specs){|s| s.name = 'a' }
+    save_gemspec('b-1', '1', dir_standard_specs){|s| s.name = 'b' }
+
+    assert_equal ['a-1'], Gem::Specification.stubs_for('a').map { |s| s.full_name }
+    assert_equal 1, Gem::Specification.class_variable_get(:@@stubs_by_name).length
+    assert_equal ['b-1'], Gem::Specification.stubs_for('b').map { |s| s.full_name }
+    assert_equal 2, Gem::Specification.class_variable_get(:@@stubs_by_name).length
+
+    Gem.loaded_specs.delete 'a'
+    Gem.loaded_specs.delete 'b'
+    Gem::Specification.class_variable_set(:@@stubs, nil)
+  end
+
   def test_self_stubs_for_mult_platforms
     # gems for two different platforms are installed with --user-install
     # the correct one should be returned in the array


voxik:
Not sure about the test errors. They don't seems to be related, but I might be wrong.

segiddins:
This looks like it is breaking `Gem.refresh` behavior bundler was depending on? Additionally, could you please look into adding a spec that covers incrementally populating the array so we don’t regress again in the future?

deivid-rodriguez:
I had a look at this and the failure is expected. The failing test ensures that `Gem.refresh` does nothing in `bundler`. Previously the test would call it before and after finding stubs for a gem, and it would return `[]` before and after. Now it will return the gem before and after, because the array of stubs is now populated incrementally.

I think I would instead deprecate or remove `Gem.refresh`, and remove the test in bundler?

deivid-rodriguez:
@voxik In my opinion, this PR makes `Gem.refresh` unnecessary. So my plan would be to deprecate it, and remove the bundler code to monkeypatch it to a noop and the corresponding spec. After all, users could still call `Gem::Specification.reset` and break bundler in the same way... :man_shrugging:. Thoughts?

deivid-rodriguez:
By the way, this was introduced into bundler here: https://github.com/bundler/bundler/commit/f0a65fea6ba4907a181eae5ede77bc1d74fa7c78, to fix a bug in unicorn because it uses `Gem.refresh` since https://github.com/defunkt/unicorn/commit/8fc898a63e742190903d331958d3b6627f819624.

voxik:
I don't understand why the `Gem.refresh` should be unnecessary? In RubyGems, it does what it is supposed to do, doesn't it? I.e. if forces reload of everything. Please note that gems might be not just added, but also removed.

Or are you speaking about the Bundler monkeypatch? It probably makes sense it is no-op, not sure. But I can't say I really understand the test. It should probably test that the `Gem::Specifications.stubs` is still the same before and after the `refresh` call.

deivid-rodriguez:
> I don't understand why the Gem.refresh should be unnecessary? In RubyGems, it does what it is supposed to do, doesn't it? I.e. if forces reload of everything. Please note that gems might be not just added, but also removed.

You're totally right, I failed to consider _removing_ gems :+1:

> Or are you speaking about the Bundler monkeypatch? It probably makes sense it is no-op, not sure. But I can't say I really understand the test. It should probably test that the Gem::Specifications.stubs is still the same before and after the refresh call.

I investigated a bit more, and I noticed that making `Gem.refresh` a noop is not necessary. Also, I wrote the test different to still test the same thing, but not affect this PR. See bundler/bundler#7246.

voxik:
I have updated the PR to test the lazy loading. I also added the fix for #2827 (could be separate PR, but it was on the same place and I was lazy :see_no_evil: sorry).

voxik:
The test suite looks good this time :+1:

deivid-rodriguez:
Hi again @voxik! Sorry it took me so long to get back here.

I tried several things:

* Dust off the script in the original commit to check the performance improvement. Result: Once I managed to get the script working, I found the performance is constant with or without this change. But maybe I did something wrong.

* I tried reverting the change in lib. Result: the new test you added still passes.

So, I feel like we're missing something here? :thinking:

Code wise this patch looks perfectly fine for me and also, it fixes [one spec failure](https://travis-ci.org/bundler/bundler/jobs/569278965) I'm getting when running specs against ruby-head on [a branch I'm working on](https://github.com/bundler/bundler/compare/revert_bin_activation_rubygems_monkeypatches) in bundler (the branch intends to remove some rubygems monkeypatches that I don't think we need). But I would like to understand this a bit better if we can.

voxik:
>     * Dust off the script in the original commit to check the performance improvement. Result: Once I managed to get the script working, I found the performance is constant with or without this change. But maybe I did something wrong.

I have not digged into the performance nor it is my real concern. My real concern is correct code and this one is broken.

>     * I tried reverting the change in lib. Result: the new test you added still passes.

Well, yes. The issue is that the original code check for `@@stubs` value which is always `nil` in the test case (and it is also `nil` prior any `Gem::Specification.stubs` call), so it goes through the second branch. This branch never populates the `@@stubs`, but it populates `@@stubs_by_name`. Not sure how to really test the cache hit though.

Just FTR, this is the code I use to demonstrate the cache behavior:

~~~
$ git diff
diff --git a/lib/rubygems/specification.rb b/lib/rubygems/specification.rb
index 72fbaa5f..89e7b4e3 100644
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -840,8 +840,10 @@ class Gem::Specification < Gem::BasicSpecification
 
   def self.stubs_for(name)
     if @@stubs_by_name[name]
+      puts "♥ cache"
       @@stubs_by_name[name]
     else
+      puts "! load"
       pattern = "#{name}-*.gemspec"
       stubs = Gem.loaded_specs.values +
         installed_stubs(dirs, pattern).select { |s| Gem::Platform.match s.platform } +
diff --git a/test/rubygems/test_gem_specification.rb b/test/rubygems/test_gem_specification.rb
index 7d73fa2d..9616742d 100644
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1175,8 +1175,12 @@ dependencies: []
     save_gemspec('a-1', '1', dir_standard_specs){|s| s.name = 'a' }
     save_gemspec('b-1', '1', dir_standard_specs){|s| s.name = 'b' }
 
+    puts 'a-1'
+    assert_equal ['a-1'], Gem::Specification.stubs_for('a').map { |s| s.full_name }
+    puts 'a-1'
     assert_equal ['a-1'], Gem::Specification.stubs_for('a').map { |s| s.full_name }
     assert_equal 1, Gem::Specification.class_variable_get(:@@stubs_by_name).length
+    puts 'b-1'
     assert_equal ['b-1'], Gem::Specification.stubs_for('b').map { |s| s.full_name }
     assert_equal 2, Gem::Specification.class_variable_get(:@@stubs_by_name).length
 
~~~

And this is the output:

~~~
a-1
! load
a-1
♥ cache
b-1
! load
~~~

And here is the original unmodified code path:

~~~
$ git diff
diff --git a/lib/rubygems/specification.rb b/lib/rubygems/specification.rb
index 72fbaa5f..2c1cf82d 100644
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -839,9 +839,11 @@ class Gem::Specification < Gem::BasicSpecification
   # only returns stubs that match Gem.platforms
 
   def self.stubs_for(name)
-    if @@stubs_by_name[name]
-      @@stubs_by_name[name]
+    if @@stubs
+      puts "♥ cache"
+      @@stubs_by_name[name] || []
     else
+      puts "! load"
       pattern = "#{name}-*.gemspec"
       stubs = Gem.loaded_specs.values +
         installed_stubs(dirs, pattern).select { |s| Gem::Platform.match s.platform } +
diff --git a/test/rubygems/test_gem_specification.rb b/test/rubygems/test_gem_specification.rb
index 7d73fa2d..9616742d 100644
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1175,8 +1175,12 @@ dependencies: []
     save_gemspec('a-1', '1', dir_standard_specs){|s| s.name = 'a' }
     save_gemspec('b-1', '1', dir_standard_specs){|s| s.name = 'b' }
 
+    puts 'a-1'
+    assert_equal ['a-1'], Gem::Specification.stubs_for('a').map { |s| s.full_name }
+    puts 'a-1'
     assert_equal ['a-1'], Gem::Specification.stubs_for('a').map { |s| s.full_name }
     assert_equal 1, Gem::Specification.class_variable_get(:@@stubs_by_name).length
+    puts 'b-1'
     assert_equal ['b-1'], Gem::Specification.stubs_for('b').map { |s| s.full_name }
     assert_equal 2, Gem::Specification.class_variable_get(:@@stubs_by_name).length
 
~~~


and here is the output with the code path from the patch:

~~~
a-1
! load
a-1
! load
b-1
! load
~~~

As you can see, the .spec is reloaded.

voxik:
Also, maybe this 4cee8ca9199ac7b3ab8647e0b78615f55d3eb02b should be reverted for better performance, because the commit message is quite right, but forgets that the `@@stubs_by_name[name]` is cached independently from the `@@stubs` by this method itself. But hard to say what may be the impact ...

voxik:
I have updated the test for lazy loading to also test that two subsequent calls of `stub_for` returns the same spec objects, checked by their `object_id`. This makes the test fail with the original code.

voxik:
BTW I was thinking about better message for the latest `assert_equal` but couldn't come with anything meaningful :blush: I am open for suggestions (of course people can dig through the git changelog for explanation 😇)

voxik:
> BTW I was thinking about better message for the latest `assert_equal` but couldn't come with anything meaningful blush I am open for suggestions

Well, separate test case would probably resolve this, but there is so much boilerplate ...

voxik:
Rebased, because PR #2970 extracted one commit out of this PR.

deivid-rodriguez:
Oh, indeed, thanks @voxik! Wanted to ask you to do just that, but forgot!

deivid-rodriguez:
Reviews appreciated for the rest of this PR :)

deivid-rodriguez:
There's been no feedback here whatsoever, so this is going in with my approval! :)

PR #2715, opened 2019-04-03
Deprecate unused Gem::Installer#unpack method.

voxik:
Deprecate unused `Gem::Installer#unpack` method. This method is useless for past seven years (f0dcf9f).
    
The `options[:unpack]` was used on various places for some tweaks (164a642 and 41f7ffd), it is unused nowadays.

Please note that `unpack` option from `Gem::Installer#verify_gem_home` method is removed in separate commit, because it breaks API. It should be considered separately. But I would prefer to break the API, because I really doubt somebody is using the method with non-default option.

### DIFF
--- a/lib/rubygems/commands/unpack_command.rb
+++ b/lib/rubygems/commands/unpack_command.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 require 'rubygems/command'
-require 'rubygems/installer'
 require 'rubygems/version_option'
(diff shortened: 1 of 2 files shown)


hsbt:
It seems okay when passing test suite.

deivid-rodriguez:
Seems fine to me too.

bronzdoc:
@bundlerbot r+

PR #2718, opened 2019-04-06
Skip deprecation warning during specs

deivid-rodriguez:
# Description:

Since #2715, specs print a deprecation warning "NOTE: Gem::Installer#unpack is deprecated with no replacement. It will be removed on or after 2020-04-01.
Gem::Installer#unpack called from /home/deivid/Code/rubygems/test/rubygems/test_gem_installer.rb:1727."

This PR skips that during the spec run.

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2720, opened 2019-04-08
Allow uninstall from symlinked GEM_HOME

### LINKED ISSUE #2705: gem uninstall -i path doesn't work if path is a symlink
```
$ gem env                                                                     
RubyGems Environment:
  - RUBYGEMS VERSION: 3.0.3
  - RUBY VERSION: 2.6.2 (2019-03-13 patchlevel 47) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/eregon/.gem/ruby/2.6.2
  - USER INSTALLATION DIRECTORY: /home/eregon/.gem/ruby/2.6.0
  - RUBY EXECUTABLE: /home/eregon/.rubies/ruby-2.6.2/bin/ruby
  - GIT EXECUTABLE: /usr/bin/git
  - EXECUTABLE DIRECTORY: /home/eregon/.gem/ruby/2.6.2/bin
  - SPEC CACHE DIRECTORY: /home/eregon/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /home/eregon/.rubies/ruby-2.6.2/etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /home/eregon/.gem/ruby/2.6.2
     - /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - "gem" => "--no-document"
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /home/eregon/.gem/ruby/2.6.2/bin
     - /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0/bin
     - /home/eregon/.rubies/ruby-2.6.2/bin
     - /home/eregon/bin
     - /usr/local/bin
     - /usr/local/sbin
     - /usr/bin
     - /usr/sbin
     - /home/eregon/code/mx
     - /home/eregon/code/mspec/bin
     - /opt/crystal/bin
$ gem install --install-dir /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0 path
Successfully installed path-2.0.1
1 gem installed
$ ln -s /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0 sym
$ gem uninstall -i sym path
# No output, and the gem is still there
$ gem uninstall -i /home/eregon/.rubies/ruby-2.6.2/lib/ruby/gems/2.6.0 path
# Works
```

This notably happens if one gives the `..@global` directory for removing gems when using RVM, such as what the CI does currently:
https://github.com/rubygems/rubygems/blob/78022d6a1136007dab6d8e4dadc8b0f229d0158c/util/ci#L52

Related to #2701.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
$ gem env version
3.0.3
```

Relates to #2700.

deivid-rodriguez:
# Description:

Currently, if GEM_HOME is a symlink (for example, in rvm's ruby-head), `gem uninstall` fails. I hope to fix that in this PR.

Fixes #2705.

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem_uninstaller.rb
+++ b/test/rubygems/test_gem_uninstaller.rb
@@ -22,9 +22,10 @@ class TestGemUninstaller < Gem::InstallerTestCase
   end
 
   def test_initialize_expand_path
-    uninstaller = Gem::Uninstaller.new nil, :install_dir => '/foo//bar'
+    FileUtils.mkdir_p 'foo/bar'
+    uninstaller = Gem::Uninstaller.new nil, :install_dir => 'foo//bar'
 
-    assert_match %r|/foo/bar$|, uninstaller.instance_variable_get(:@gem_home)
+    assert_match %r|foo/bar$|, uninstaller.instance_variable_get(:@gem_home)
   end
 
   def test_ask_if_ok
@@ -133,6 +134,7 @@ class TestGemUninstaller < Gem::InstallerTestCase
   end
 
   def test_remove_not_in_home
+    Dir.mkdir "#{@gemhome}2"
     uninstaller = Gem::Uninstaller.new nil, :install_dir => "#{@gemhome}2"
 
     e = assert_raises Gem::GemNotInHomeException do
@@ -147,6 +149,22 @@ class TestGemUninstaller < Gem::InstallerTestCase
     assert_equal expected, e.message
 
     assert_path_exists @spec.gem_dir
+  end
+
+  def test_remove_symlinked_gem_home
+    Dir.mktmpdir("gem_home") do |dir|
+      symlinked_gem_home = "#{dir}/#{File.basename(@gemhome)}"
+
+      FileUtils.ln_s(@gemhome, dir)
+
+      uninstaller = Gem::Uninstaller.new nil, :install_dir => symlinked_gem_home
+
+      use_ui ui do
+        uninstaller.remove @spec
+      end
+
+      refute_path_exists @spec.gem_dir
+    end
   end
 
   def test_path_ok_eh
@@ -313,6 +331,7 @@ create_makefile '#{@spec.name}'
   end
 
   def test_uninstall_wrong_repo
+    Dir.mkdir "#{@gemhome}2"
     Gem.use_paths "#{@gemhome}2", [@gemhome]
 
     uninstaller = Gem::Uninstaller.new @spec.name, :executables => true
--- a/lib/rubygems/uninstaller.rb
+++ b/lib/rubygems/uninstaller.rb
@@ -46,7 +46,7 @@ class Gem::Uninstaller
     # TODO document the valid options
     @gem                = gem
     @version            = options[:version] || Gem::Requirement.default
-    @gem_home           = File.expand_path(options[:install_dir] || Gem.dir)
+    @gem_home           = File.realpath(options[:install_dir] || Gem.dir)
     @force_executables  = options[:executables]
     @force_all          = options[:all]
     @force_ignore       = options[:ignore]


deivid-rodriguez:
It'd be great if someone with a Windows machine, or with some Windows knowledge could have a look at Appveyor :sweat_smile:.

test/rubygems/test_gem_uninstaller.rb
```diff
@@ -149,6 +151,20 @@ def test_remove_not_in_home
     assert_path_exists @spec.gem_dir
   end
 
+  def test_remove_symlinked_gem_home
+    symlinked_gem_home = Tempfile.new("gem_home").path
+
+    FileUtils.ln_sf(@gemhome, symlinked_gem_home)
```
p-mongo: I imagine at this point the file created by Tempfile is still open, and this line tries to remove it and replace it with a symlink. This works on unix but not on windows.

My suggestion is to create a temporary directory instead of a temporary file and use `ln_s` instead of `ln_sf` to create a symlink within that directory.

MSP-Greg:
@deivid-rodriguez 

> It'd be great if someone with a Windows machine, or with some Windows knowledge

1st - yes, 2nd - questionable...

Windows won't make a symlink from an existing file.  Something like the following:

```ruby
  def test_remove_symlinked_gem_home
    symlinked_gem_home = (tmp = Tempfile.new("gem_home")).path
    tmp.close(false) if windows?            # windows won't close an open file

    FileUtils.ln_sf(@gemhome, symlinked_gem_home)
```

p-mongo:
That pattern is a temporary file race (see https://dwheeler.com/secure-programs/Secure-Programs-HOWTO/avoid-race.html section 7.11.1.2. Temporary Files).

MSP-Greg:
Well, if the intent is to get a safe name, not many other options.  I ran it in Appveyor and it passed.  Ruby Core runs tests parallel, I haven't tried it there.

Since Windows will not allow one to create a symlink from an existing file system name, and `Tempfile.new` returns a object that should be open, so I'm not sure I see the problem...

p-mongo:
Creating a temporary directory, and working within it, is the safe alternative.

https://ruby-doc.org/stdlib-1.9.3/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

deivid-rodriguez:
Thanks both for your help! It should be fixed now.

deivid-rodriguez:
As can be seen in the tests I had to modify, the uninstaller now complains if you pass it a non-existent directory. Sounds like a good thing.

deivid-rodriguez:
I also took the chance to fix our CI against ruby-head, since it was hitting precisely this bug as pointed out by @eregon in #2705.

deivid-rodriguez:
Can we get this in for a fully green CI matrix, including ruby-head? :smiley:

util/ci
```diff
@@ -49,7 +49,7 @@ end
 case ARGV
 when %w(before_script)
   if TOOL.rubygems?
-    run('gem', %W(uninstall executable-hooks gem-wrappers bundler-unload -x --force -i #{`gem env home`.strip}@global))
+    run('ruby', %W(-I lib bin/gem uninstall executable-hooks gem-wrappers bundler-unload -x --force -i #{`gem env home`.strip}@global))
```
bronzdoc: Why is this change needed here?

deivid-rodriguez: Not really "needed" but our build against `ruby-head` is [hitting precisely this bug](https://travis-ci.org/rubygems/rubygems/jobs/519464636#L236-L241). So by changing this, we get to use this fix, and thus get it green.

bronzdoc: Maybe have this in a separate PR? just so this change doesn't get obscure in the purpose of this PR, what you think?

deivid-rodriguez: Sure.

deivid-rodriguez: I removed the commit fixing the ruby-head build and will open a separate PR once this one is merged.

bronzdoc: Maybe we can merge the other one first and then rebase this one? so this is green in ruby-head too :)

deivid-rodriguez: We need the fix in this PR as a prerequisite to fix ruby-head!

deivid-rodriguez: If we want this PR to be green in ruby-head, I need to restore the commit I removed.

deivid-rodriguez: That's why I put both commits in this PR, because they are very related.

bronzdoc:
@bundlerbot r+

PR #2721, opened 2019-04-09
Test ruby-head against bundler's master

deivid-rodriguez:
# Description:

I'd like to get our ruby-head build green. The current failure of bundler's build is fixed on bundler's master. To pick up this fix, I'd like to change ruby-head's build to test bundler's master instead of the currently vendored one.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,21 @@ rvm:
   - 2.4.5
   - 2.5.5
   - 2.6.2
-  - ruby-head
 jobs:
   include:
     - stage: linting
       rvm: 2.6.2
-      env:
-        - "TEST_TOOL=rubygems"
+      env: "TEST_TOOL=rubygems"
       script: util/ci rubocop
     - stage: test
       rvm: 2.6.2
-      env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,18 +15,21 @@ rvm:
...
-      env:
-        - "TEST_TOOL=rubygems"
+      env: "TEST_TOOL=rubygems"
       script: util/ci rubocop
     - stage: test
       rvm: 2.6.2
-      env:
-        - "TEST_TOOL=bundler RGV=.. BDV=master"
+      env: "TEST_TOOL=bundler RGV=.. BDV=master"
+    - stage: test
```
bronzdoc: This stage will test Rubygems against vendored bundler right?

deivid-rodriguez: Yes. The regular test run for rubygems adds the vendored bundler to the LOAD_PATH, so the few specs that use bundler should use the vendored one. This change affects only bundler's specs, that run against the rubygems checkout instead of against rubygems master. This change is similar to #2695, but for ruby-head.

bronzdoc:
@bundlerbot r+

PR #2725, opened 2019-04-10
Refactor push command host branching

thomas-mcdonald:
# Description:

I started looking into notifying users when making non-mfa pushes to RubyGems.org, but discovered the host decision logic was branch heavy and (I believe) can be simplified to reduce complexity.

I hand traced the code while making these changes, so probably needs a careful eye in review. I'll post a couple of explanatory comments alongside.

There is an outstanding failing test as the test suite seems to heavily rely on the internal state of the class and certain logic happening in `send_gem` rather than in `execute`. I'm hoping to dig into that a bit more on a plane tomorrow.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -70,13 +70,7 @@ The push command will use ~/.gem/credentials to authenticate to a server, but yo
   def send_gem(name)
     args = [:post, "api/v1/gems"]
 
-    gem_data = Gem::Package.new(name)
-
-    push_host = nil
-
-    if gem_data.spec.metadata.has_key?('allowed_push_host')
-      push_host = gem_data.spec.metadata['allowed_push_host']
-    end
+    _, push_host = get_hosts_for(name)
 
     @host ||= push_host
 
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -72,10 +72,6 @@ The push command will use ~/.gem/credentials to authenticate to a server, but yo
 
     gem_data = Gem::Package.new(name)
 
-    unless @host
-      @host = gem_data.spec.metadata['default_gem_server']
-    end
-
     push_host = nil
 
     if gem_data.spec.metadata.has_key?('allowed_push_host')
--- a/lib/rubygems/commands/push_command.rb
+++ b/lib/rubygems/commands/push_command.rb
@@ -52,23 +52,14 @@ The push command will use ~/.gem/credentials to authenticate to a server, but yo
     gem_name = get_one_gem_name
     default_gem_server, push_host = get_hosts_for(gem_name)
 
-    default_host = nil
-    user_defined_host = nil
-
-    if @user_defined_host
-      user_defined_host = options[:host]
-    else
-      default_host = options[:host]
-    end
-
-    @host = if user_defined_host
-              user_defined_host
+    @host = if @user_defined_host
+              options[:host]
             elsif default_gem_server
               default_gem_server
             elsif push_host
               push_host
             else
-              default_host
+              options[:host]
             end
 
     sign_in @host


lib/rubygems/commands/push_command.rb
```diff
@@ -52,23 +52,14 @@ def execute
     gem_name = get_one_gem_name
     default_gem_server, push_host = get_hosts_for(gem_name)
```
thomas-mcdonald: This block was a branch and variable heavy way of saying use the user-provided host if available, or fall-back to the default host if none of the other options match. IMO the resultant single-if condition is clearer in intentions

lib/rubygems/commands/push_command.rb
```diff
@@ -98,19 +89,7 @@ def send_gem(name)
       terminate_interaction 1
     end
 
-    gem_data = Gem::Package.new(name)
-
-    unless @host
-      @host = gem_data.spec.metadata['default_gem_server']
-    end
```
thomas-mcdonald: With the if branch above, host is not nil (I think? or it would take some serious effort to make it thus), and default_gem_server is already handled

lib/rubygems/commands/push_command.rb
```diff
@@ -98,19 +89,7 @@ def send_gem(name)
...
-
-    push_host = nil
-
-    if gem_data.spec.metadata.has_key?('allowed_push_host')
-      push_host = gem_data.spec.metadata['allowed_push_host']
-    end
-
-    @host ||= push_host
```
thomas-mcdonald: host will not be nil if `push_host` is present so assignment is not useful, use the same mechanism as in `execute` to standardise access to gem metadata

thomas-mcdonald:
Investigating tests, I believe that `test_execute_allowed_push_host` is functionally equivalent to `test_sending_gem_defaulting_to_allowed_push_host` while also requiring less knowledge of the implementation internals. I've removed the defaulting test, and I think this is good for review.

bronzdoc:
Hey @thomas-mcdonald thanks for the PR! i just have a little doubt :)

> Investigating tests, I believe that test_execute_allowed_push_host is functionally equivalent to test_sending_gem_defaulting_to_allowed_push_host

if the test are equivalent why does `TestGemCommandsPushCommand#test_sending_gem_defaulting_to_allowed_push_host` fails with the changes?

thomas-mcdonald:
The test is equivalent from the perspective of the overall behaviour of the command. However if directly calling `Gem::Commands::PushCommand#send_gem` is a part of the public API, `test_sending_gem_defaulting_to_allowed_push_host` is a valid test that ensures given a blank state of PushComand there is no regression in the body of the send_gem method in isolation. I guess the validity of this change (and my other changes to `send_gem`) depend on if we consider the API for `Gem::Command`s to be the command as a whole or each public method in the class.

lib/rubygems/commands/push_command.rb
deivid-rodriguez: Yup. I understand the same thing. Can we extract this specific part of the PR to a separate commit and include your commit in the commit message?

deivid-rodriguez: Yup, I understand the same thing. Can you extract the refactoring of this block to a separate commit and include your explanation in the commit description?

lib/rubygems/commands/push_command.rb
```diff
@@ -98,24 +89,11 @@ def send_gem(name)
...
+    _, push_host = get_hosts_for(name)
 
-    # Always include @host, even if it's nil
     args += [ @host, push_host ]
 
-    say "Pushing gem to #{@host || Gem.host}..."
+    say "Pushing gem to #{@host}..."
```
deivid-rodriguez: I think `@host` could actually be `nil` if... everything is `nil`. If that's the case, it _does_ default to `Gem.host` unless `Gem.configuration.disable_default_gem_server` is true.

I think I would keep this message and the previous comment (`# Always include @host, even if it's nil`) as they are now, and we can iterate the refactoring further later.

lib/rubygems/commands/push_command.rb
deivid-rodriguez: Not sure wtf I'm doing but I seem to be deleting my own comments unintentionally. This is the third time I write this :rofl:.

So, I was saying that I agree with this and that it would be good if we could extract the refactoring of this block to a separate commit, and use your comment as the commit message.

thomas-mcdonald:
Sounds good! I'll push the requested changes through at some point this week

thomas-mcdonald:
Fell out with git while rebasing, but should be good to review again @deivid-rodriguez! Sorry for the delay

lib/rubygems/commands/push_command.rb
deivid-rodriguez: To make the refactoring even easier to follow, I extracted this particular refactoring and explanation to its own commit at https://github.com/rubygems/rubygems/pull/2725/commits/e3ccff3c5cd19f56dacadc9a82c9c4ccb89383c7.

deivid-rodriguez:
Sorry this one fell off my radar, thanks for working on this! I only added a couple of minor tweaks and history edits, it looks great! I love deleting code.

thomas-mcdonald:
Ha yes, I discovered it earlier on my GitHub PR list and figured I'd give it a rebase! Thanks

deivid-rodriguez:
Thank you so much for this refactoring! :purple_heart:

PR #2727, opened 2019-04-14
Fix pre-release version comparison across versions with different non-canonical segments

greysteil:
# Description:

Currently, the following happens:

```
Gem::Version.new("1.0.beta") >= Gem::Version.new("1.0.0.beta")
# => true
Gem::Version.new("1.0.beta") >= Gem::Version.new("1.0.0.alpha")
# => false
```

That is inconsistent. I propose that the second statement should be true. (Alternatively we could make the first statement false, but that is arguably a breaking change.)

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -344,8 +344,8 @@ class Gem::Version
     return unless Gem::Version === other
     return 0 if @version == other._version || canonical_segments == other.canonical_segments
 
-    lhsegments = _segments
-    rhsegments = other._segments
+    lhsegments = canonical_segments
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@greysteil See https://github.com/rubygems/rubygems/pull/2651 :smiley:

greysteil:
Ha! Make sense. I think the current situation is definitely a bug, and we should ship this as a bug fix. The case I put in the description of this PR isn't really defensible as a feature.

I'll close here and add a 👍 to #2651.

PR #2729, opened 2019-04-15
Use current checkout in CI to uninstall RVM related gems

deivid-rodriguez:
# Description:

Using the latest rubygems is better because we can pickup the latest bug
fixes present in master, but non on whatever rubygems version Travis is
using.

For example, once we merge #2720, this change will allow to get the `ruby-head` build fixed.

# Tasks:

- [ ] Write tests

### DIFF
--- a/util/ci
+++ b/util/ci
@@ -49,7 +49,7 @@ end
 case ARGV
 when %w(before_script)
   if TOOL.rubygems?
-    run('gem', %W(uninstall executable-hooks gem-wrappers bundler-unload -x --force -i #{`gem env home`.strip}@global))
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Funny. So I think what the failure is telling us here is that #2720 is not a pure bug fix, but a regression fix, and that this was actually working on whatever rubygems version the rubies in Travis have installed by default?

Anyways, it sounds like we need to merge #2720 first.

bronzdoc:
@deivid-rodriguez ok lets try that

deivid-rodriguez:
@bundlerbot r+

PR #1868, opened 2017-03-14
[Require] Ensure -I beats a default gem

segiddins:
# Description:

Fixes #1866.
Definitely a WIP, just wanted to let everyone have a look at this.

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2730, opened 2019-04-15
Vendorize latest bundler

deivid-rodriguez:
# Description:

I'd like to propose that we start vendorizing the latest bundler master.

The prerequisite for this would be that we cut a bundler release _before_ cutting a rubygems release, so that we can get the submodule back to a released version before releasing rubygems.

The motivation is that it is simpler to keep CI green if we do this. For example, fixing #1868 made surface some issues in bundler's specs that are only fixed in bundler's master. This has happened before, as can be seen by the list of cherry-pick we already need for the build to be green. Cherry-picking is not always easy, and not very clear why it's for the reader of the CI config, so I think I would prefer to do this?

# Tasks:

- [ ] Write tests

### DIFF
--- a/util/ci
+++ b/util/ci
@@ -59,14 +59,6 @@ when %w(before_script)
   else
     if ENV["BDV"]
       run('git', %W(reset --hard origin/#{ENV["BDV"]}))
-    elsif RUBY_VERSION == "2.6.2" || RUBY_VERSION == "2.5.5"
-      # Manually cherry-pick changes needed so that bundler does not load
-      # default gems by default, and thus its specs pass.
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
By the way, I'm not sure if I changed the submodule the right way. I run `git submodule update --remote`, and committed the result.

hsbt:
👍 We start to consider a repository unification about RubyGems and Bundler in this year.

hsbt:
@bundlerbot r+

bronzdoc:
@bundlerbot r+

Issue #2692, opened 2019-03-20
It's not clear how to use regexps with `gem dependency` without looking into the source


x-yuri:
I'm having a problem.

My current problem is... I was trying to find out when `rails` stopped depending on `bundler < 2.0`, so I did:

```
$ gem dependency -rv 5.1.2 rails
Gem rails-5.1.2
  actioncable (= 5.1.2)
  actionmailer (= 5.1.2)
  actionpack (= 5.1.2)
  actionview (= 5.1.2)
  activejob (= 5.1.2)
  activemodel (= 5.1.2)
  activerecord (= 5.1.2)
  activesupport (= 5.1.2)
  bundler (>= 1.3.0, < 2.0)
  railties (= 5.1.2)
  sprockets-rails (>= 2.0.0)

Gem rails-i18n-5.1.2
  i18n (>= 0.7, < 2)
  i18n-spec (~> 0.6.0, development)
  i18n-tasks (~> 0.9.20, development)
  railties (>= 5.0, < 6)
  rspec-rails (~> 3.7, development)

Gem rails_jskit-5.1.2
  appraisal (~> 0, development)
  fuubar (~> 2.0, development)
  pry (~> 0.10, development)
  pry-doc (~> 0.6, development)
  pry-nav (~> 0.2, development)
  rails (~> 4.0)
  rspec-rails (~> 3.1, development)
  sqlite3 (~> 1.3, development)
```

But then I saw that `gem dependency` takes regexp, so I did:

```
$ gem dependency -rv 5.1.2 ^rails$
No gems found matching ^rails$ (= 5.1.2)
```

But it works with `gem search`:

```
$ gem search ^rails$

*** REMOTE GEMS ***

rails (5.2.2.1)
```

Only looking at the [source][1] revealed that I've got to do like so:

[1]: https://github.com/rubygems/rubygems/blob/v3.0.3/lib/rubygems/commands/dependency_command.rb#L211-L216

```
$ gem dependency -rv 5.1.2 /^rails$/
Gem rails-5.1.2
  actioncable (= 5.1.2)
  actionmailer (= 5.1.2)
  actionpack (= 5.1.2)
  actionview (= 5.1.2)
  activejob (= 5.1.2)
  activemodel (= 5.1.2)
  activerecord (= 5.1.2)
  activesupport (= 5.1.2)
  bundler (>= 1.3.0, < 2.0)
  railties (= 5.1.2)
  sprockets-rails (>= 2.0.0)
```

[...]

PR #2738, opened 2019-04-22
Don't use regex delimiters when searching for a dependency

bronzdoc:
# Description:

closes https://github.com/rubygems/rubygems/issues/2692

This PR will make `gem search [REGEXP]` and `gem dependency [REGEXP]` use the same [REGEXP] pattern for consistency.

so now you can do `gem dependency ^rails$` instead of `gem dependency /^rails$/` 


bronzdoc:
@bundlerbot r+

PR #2739, opened 2019-04-24
Refactor query command

bronzdoc:
# Description:
Attempt to make query command code clearer

### DIFF
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -137,37 +137,8 @@ is too hard to use.
 
   #Guts of original execute
   def show_gems(name, prerelease)
-    req = Gem::Requirement.default
-
-    show_local_gems(name) if local?
-
-    if remote?
-      display_header 'REMOTE'
-
-      fetcher = Gem::SpecFetcher.fetcher
-
-      type = if options[:all]
-               if options[:prerelease]
-                 :complete
-               else
-                 :released
-               end
-             elsif options[:prerelease]
-               :prerelease
-             else
-               :latest
-             end
-
-      if name.respond_to?(:source) && name.source.empty?
-        spec_tuples = fetcher.detect(type) { true }
-      else
-        spec_tuples = fetcher.detect(type) do |name_tuple|
-          name === name_tuple.name
-        end
-      end
-
-      output_query_results spec_tuples
-    end
+    show_local_gems(name)  if local?
+    show_remote_gems(name) if remote?
   end
 
   def show_local_gems(name, req = Gem::Requirement.default)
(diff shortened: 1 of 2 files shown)


lib/rubygems/commands/query_command.rb
```diff
@@ -129,55 +137,55 @@ def display_header(type)
 
   #Guts of original execute
   def show_gems(name, prerelease)
```
deivid-rodriguez: You have extracted `prerelease` to a method, so this method argument is now unused.

lib/rubygems/commands/query_command.rb
```diff
@@ -129,55 +137,55 @@ def display_header(type)
...
-    req = Gem::Requirement.default
-    # TODO: deprecate for real
-    dep = Gem::Deprecate.skip_during { Gem::Dependency.new name, req }
-    dep.prerelease = prerelease
```
deivid-rodriguez: I'm unsure about this deprecation but the current command is broken for local gems and the `--no-prerelease` option.

If we keep the above three lines and add `specs.select! { |spec| dep.match?(spec.name, spec.version) }` after the initial filtering of local specs, namely, where we do:

```ruby
specs = Gem::Specification.find_all do |s|
  s.name =~ name and req =~ s.version
end
```

Then I think that would fix that bug and we can remove the warning about it

```ruby
if prerelease? && !both?
  alert_warning("prereleases are always shown locally")
end
```

bronzdoc: @deivid-rodriguez how can i reproduce the bug?

deivid-rodriguez:
```
$ ruby -Ilib bin/gem list bund

*** LOCAL GEMS ***

bundler (2.1.0.pre.1, 2.0.1, default: 1.17.3)

$ ruby -Ilib bin/gem query bund --prerelease
WARNING:  prereleases are always shown locally

*** LOCAL GEMS ***

bundler (2.1.0.pre.1, 2.0.1, default: 1.17.3)

$ ruby -Ilib bin/gem query bund --no-prerelease

*** LOCAL GEMS ***

bundler (2.1.0.pre.1, 2.0.1, default: 1.17.3)
```

If I pass the `--no-prerelease` parameter, prereleases should be excluded.

bronzdoc:
This is ready if you have any more feedback i appreciate it @deivid-rodriguez 👀  
I fixed the `--no-prerelease` bug and removed the warning, since is not needed anymore.

lib/rubygems/commands/query_command.rb
```diff
@@ -123,6 +123,10 @@ def prerelease?
     options[:prerelease]
   end
 
+  def show_prereleases?
+    options[:prerelease].nil? || options[:prerelease]
```
deivid-rodriguez: You can reuse the `prerelease?` method here :)

bronzdoc: how i missed that 😂 thanks!

bronzdoc:
> I would instead show them too by default for --remote too, but we can do that separately 
I'll do a followup PR after this is merged for that 👍

lib/rubygems/commands/query_command.rb
```diff
@@ -78,46 +78,58 @@ def description # :nodoc:
...
-      terminate_interaction exit_code
-    end
+  def prerelease?
+    options[:prerelease]
+  end
 
-    names = Array(name)
-    names.each { |n| show_gems n, prerelease }
+  def show_prereleases?
+    options[:prerelease].nil? || prerelease?
```
deivid-rodriguez: You can use the `prerelease` method in the first part of the condition as well :)

bronzdoc: isn't `prerelease?.nil?` to weird?

deivid-rodriguez: Hahaha, you're right! Maybe we can call the method `prerelease` instead, since it's not strictly boolean anyways.

bronzdoc: Fair 👍

bronzdoc:
@bundlerbot r+

Issue #2681, opened 2019-03-11
Gem.remove_unresolved_default_spec doesn't delete spec's files from @path_to_default_spec_map


x-yuri:
I'm having a problem.

My current problem is... I've been trying to figure out what this line of code is [for][1], and noticed that [`Gem.remove_unresolved_default_spec`][2] doesn't necessarily undo what [`Gem.register_default_spec`][3] does. Doesn't undo in case of the new format. I believe it was broken by this [commit][4].

[1]: https://github.com/rubygems/rubygems/blob/v3.0.3/lib/rubygems/core_ext/kernel_require.rb#L40
[2]: https://github.com/rubygems/rubygems/blob/v3.0.3/lib/rubygems.rb#L1283-L1287
[3]: https://github.com/rubygems/rubygems/blob/v3.0.3/lib/rubygems.rb#L1251-L1270
[4]: https://github.com/rubygems/rubygems/commit/cc6d4c31

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

Here are my current environment details:

```
$ gem env version
3.0.1
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

On a side note, could you by any chance explain [the line][1]? I believe the method was added in the following [commit][5]. Do we really need to undo registering a default spec? Also judging from the commit, it seems there's no way to require a file from standard library if there's a file by that name in a default gem. Is it not an issue because standard library and default gems are both maintained by language developers?

[5]: https://github.com/rubygems/rubygems/commit/02627781

**UPD** Also, for one reason or another some of the default gems (`bundler`, `irb`, `rdoc` in `ruby-2.6.0`)—supposedly, those with executables—have gem dirs (`lib/ruby/gems/<version>/gems`), but gem dirs contain only executables. And corresponding specifications list only the executables. As such requiring such gem's files resolves to the default gem version, not the lastest as is the case with the rest of the gems. That for one makes `puma` fail to restart under `ruby-2.6.0`.

deivid-rodriguez:
Hello @x-yuri!

[...]

PR #2741, opened 2019-04-25
Don't remove default spec files from mapping after require

deivid-rodriguez:
# Description:

I discovered a bug in our monkey patched `require` where default gems would not be properly removed from our "default spec file mapping" after being required the first time. This doesn't have any effect at the correctness level because `Kernel.gem` will detect that the gem has already been activated and return `false` in that case.

However, it results in `require` doing more work than it's supposed since it will loop through all files in the default gem and unsuccessfully try to remove its default mappings, _every_ time a default gem is required.

This patch fixes that bug.

As per the explanation of the fix, the previous method to remove the default spec mappings would only try to remove keys like `logger.rb`, but not `logger`. And the `logger` form is what's actually used with `require` most of the times.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1303,6 +1303,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     def remove_unresolved_default_spec(spec)
       spec.files.each do |file|
         @path_to_default_spec_map.delete(file)
+        @path_to_default_spec_map.delete(file.sub(suffix_regexp, ""))
       end
     end
 
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1293,18 +1293,8 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     # Find a Gem::Specification of default gem from +path+
 
     def find_unresolved_default_spec(path)
-      @path_to_default_spec_map[path]
-    end
-
-    ##
-    # Remove needless Gem::Specification of default gem from
-    # unresolved default gem list
-
-    def remove_unresolved_default_spec(spec)
-      spec.files.each do |file|
-        @path_to_default_spec_map.delete(file)
-        @path_to_default_spec_map.delete(file.sub(suffix_regexp, ""))
-      end
+      default_spec = @path_to_default_spec_map[path]
+      return default_spec if default_spec && loaded_specs[default_spec.name] != default_spec
     end
 
     ##
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1024,6 +1024,10 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     @suffix_pattern ||= "{#{suffixes.join(',')}}"
   end
 
+  def self.suffix_regexp
+    @suffix_regexp ||= /#{Regexp.union(suffixes)}\z/
+  end
+
   ##
   # Suffixes for require-able paths.
 
@@ -1274,8 +1278,6 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
         prefix_pattern = /^(#{prefix_group})/
       end
 
-      suffix_pattern = /#{Regexp.union(Gem.suffixes)}\z/
-
       spec.files.each do |file|
         if new_format
           file = file.sub(prefix_pattern, "")
@@ -1283,7 +1285,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
         end
 
         @path_to_default_spec_map[file] = spec
-        @path_to_default_spec_map[file.sub(suffix_pattern, "")] = spec
+        @path_to_default_spec_map[file.sub(suffix_regexp, "")] = spec
       end
     end
 
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -41,6 +41,10 @@ class TestGemRequire < Gem::TestCase
     assert require(path), "'#{path}' was already required"
   end
 
+  def refute_require(path)
+    refute require(path), "'#{path}' was not yet required"
+  end
+
   # Providing -I on the commandline should always beat gems
   def test_dash_i_beats_gems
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
While replying to @x-yuri in #2681 I noticed that we can probably kill this logic instead... Thoughts?

deivid-rodriguez:
I don't know much about benchmarking but I tried to compare both versions, the one fixing the bug, and the one avoid the removal of the default spec mapping altogether:

```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.report("require-with-default-spec-map-removal") do |times|
    load "lib/rubygems/core_ext/kernel_require.rb"
    times.times { require "json" }
  end

  x.report("require-without-default-spec-map-removal") do |times|
    load "lib/rubygems/core_ext/kernel_require_2.rb"
    times.times { require "json" }
  end

  x.compare!
end
```

It says avoiding the default spec mapping removal is 1.43x faster.

```
$ ruby test.rb 
require-with-default-spec-map-removal
                       213.000  i/100ms
require-without-default-spec-map-removal
                       219.000  i/100ms
require-with-default-spec-map-removal
                         23.700k (± 5.7%) i/s -    118.002k in   4.999749s
require-without-default-spec-map-removal
                         33.854k (± 4.9%) i/s -    168.849k in   5.002759s

Comparison:
require-without-default-spec-map-removal:    33853.7 i/s
require-with-default-spec-map-removal:    23700.0 i/s - 1.43x  slower
```

x-yuri:
So, every time a file is required, `ruby` is bound to check if `path` is in `@path_to_default_spec_map`. And reducing the map can theoretically lighten the burden.

I'm no benchmarking expert either. And as you've said yourself it's not clear if `remove_unresolved_default_spec` does any good. In practice. So I'd rather err on the side of not making a premature optimization. That is, I'd rather not have code that is not necessarily needed. Unless someone smarter can prove otherwise.

deivid-rodriguez:
> So, every time a file is required, ruby is bound to check if path is in @path_to_default_spec_map. And reducing the map can theoretically lighten the burden.

Mmmm, well, being a hash, accessing it should be o(1). I think the idea of this method was to avoid "reactivating" default gems every time they are required, since after the first require the files would no longer be in the `@path_to_default_spec_map` hash. However, this is not an issue in practice, since `Kernel.gem` already checks for this.

https://github.com/rubygems/rubygems/blob/b7b0cde49d29250f7c7852e3e9f651dff1b3a6a9/lib/rubygems/core_ext/kernel_gem.rb#L61-L63

So, yeah, it's better to not use the `remove_unresolved_default_spec` method since it's not needed and it might make a performance difference since it loops through all the files in default gems.

deivid-rodriguez:
The diff is somewhat messy, since it includes three different changes:

* Removal of the `remove_unresolved_default_spec` method. This is what should make the performance difference.

* Logic to avoid calling `Kernel.gem` a second time. I think this is not very important and it's only there to make the test I added initially pass, but I don't think it saves us much, and I'm tempted to just remove it (together with the test).  The only thing preventing me from removing it is that I'm unsure how to test the removal otherwise...

* Some refactoring to move a local variable (`suffix_pattern`) to a method and give it a better name (`suffix_regexp`). I made this change while working on the initial fix, I think it improves the clarity of the code.

x-yuri:
> Mmmm, well, being a hash, accessing it should be o(1). I think the idea of this method was to avoid "reactivating" default gems every time they are required, since after the first require the files would no longer be in the @path_to_default_spec_map hash.

Not that it matters much probably, but my idea was that there are not that many default gems. So most of the time it would hit [`Gem.find_unresolved_default_spec`][1] call, but won't "step in." But now that I think about it, it's not about default gems, but rather about their files. Every required file belonging to a default gem would pass. But like you said there's a check down the road (`return if loaded`). So the impact is not that big as it could be.

P.S. And good point about it being o(1).

[1]: https://github.com/rubygems/rubygems/blob/v3.0.3/lib/rubygems/core_ext/kernel_require.rb#L39

lib/rubygems.rb
```diff
@@ -1012,6 +1012,10 @@ def self.suffix_pattern
     @suffix_pattern ||= "{#{suffixes.join(',')}}"
   end
 
+  def self.suffix_regexp
```
segiddins: document and mark private?

test/rubygems/test_require.rb
```diff
@@ -296,6 +300,22 @@ def test_default_gem_only
...
+
+    assert_require "default/gem"
+
+    times_called = 0
+
+    Kernel.stub(:gem, ->(name, requirement) { times_called += 1 }) do
+      refute_require "default/gem"
+    end
+
+    assert_equal 0, times_called
```
segiddins: im not entirely understanding what we expect to be called 0 times?

lib/rubygems.rb
deivid-rodriguez: I'd like to keep it inline with the about `suffix_pattern` method, so I'll make both private?

test/rubygems/test_require.rb
deivid-rodriguez: I expect `Kernel.gem` to no longer be called after the first required. Currently it's called every time you require a default gem.

deivid-rodriguez:
Thanks for having a look @segiddins. I replied to your comments, but this is not ready yet.

My idea is to remove the current test and replace it with a performance test that makes sure that the time to require a default gem does not increase linearly with the number of files in the default gem, but it takes a constant time.

deivid-rodriguez:
Sorry for requesting the review when this is not ready, by the way... :pray:.

When I requested it, I thought I was done... :sweat_smile:

deivid-rodriguez:
I wrote the performance test, but it was too dependent on the system running the test, so I decided this is ok as it is.

Reviews appreciated!

deivid-rodriguez:
Thanks! :heart:

@bundlerbot r=bronzdoc

PR #2742, opened 2019-04-25
Cleanup base test case

deivid-rodriguez:
# Description:

I was working on adding a proper performance test for #2741, and since I needed a separate base test class, I noticed that the current base test class is a bit of a mess. In this PR I tried to improve it a bit.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -537,22 +537,6 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
   end
 
   ##
-  # creates a temporary directory with hax
-  # TODO: deprecate and remove
-
-  def create_tmpdir
-    tmpdir = nil
-
-    Dir.chdir Dir.tmpdir do
-      tmpdir = Dir.pwd
-    end # HACK OSX /private/tmp
-
-    tmpdir = File.join tmpdir, "test_rubygems_#{$$}"
-    FileUtils.mkdir_p tmpdir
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@deivid-rodriguez

Re `Remove references to old minitest', that may still be needed for use in Ruby Core testing, which does not use the bundled gems for testing.

deivid-rodriguez:
Good point, I'll wait for @hsbt to confirm and also take the chance to actually run specs from ruby-core :+1:.

MSP-Greg:
> actually run specs from ruby-core

That might be tricky.  I've got patches in ruby-loco to allow tests to run from the install directory (without the build directory).  Things like loading cert files break...  I did the patches a long time ago, so I'd have to look thru them to see all the issues...

deivid-rodriguez:
I checked ruby-core. It has a strange situation where minitest 5 is a bundled gem, but they use sort of a fork of minitest 4 internally. I could see where bundler specs are run, but I could find where rubygems tests are run. Anyways, I removed the commit just to be sure it's all good.

deivid-rodriguez:
I run rubygems tests from ruby-core and it's all good :+1:

deivid-rodriguez:
@hsbt Thoughts about this?

hsbt:
I confirmed to work this on ruby-core repo. Thanks!

hsbt:
@bundlerbot r+

deivid-rodriguez:
Thanks @hsbt!

PR #2748, opened 2019-04-26
Fix TODOs

deivid-rodriguez:
# Description:

Never seen these exceptions not they have been reported, so I think we're happy now?

# Tasks:

- [x] Write tests


bronzdoc:
@bundlerbot r+

PR #2749, opened 2019-04-27
Extract an `add_to_load_path` method

deivid-rodriguez:
# Description:

This logic is duplicated in several places in bundler, so I'd like to extract it to a public method, so that `bundler` can reuse it. 

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2755, opened 2019-04-30
Query command: Default to show prerelease gems for a remote domain

bronzdoc:
# Description:
Following up with: https://github.com/rubygems/rubygems/pull/2739#issuecomment-487142764

This PR will make whenever we query for a gem in a remote domain show the prereleases of that gem as default, this is to  make the output consistent with querying a gem in a local domain.

______________

### DIFF
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -190,7 +190,7 @@ is too hard to use.
     elsif options[:prerelease]
       :prerelease
     else
-      :latest
+      :abs_latest
     end
   end
 
--- a/test/rubygems/test_gem_commands_query_command.rb
+++ b/test/rubygems/test_gem_commands_query_command.rb
@@ -37,7 +37,7 @@ class TestGemCommandsQueryCommandWithInstalledGems < Gem::TestCase
 
 *** REMOTE GEMS ***
 
-a (2)
+a (3.a, 2)
 pl (1 i386-linux)
     EOF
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Thoughts on whether we should be doing this as part of the regular release process, or as part of the next major?

indirect:
This will change how `gem list` and `gem search` work, right? Do we have any sense of whether this is a change people want or expect for those commands?

deivid-rodriguez:
Not actually, but if we change this, we should change it everywhere. The current default is that when searching for local gems, prereleases are included, but when searching for remote gems, they are ignored. That means that:

* `gem list` lists prereleases by default because it considers only local gems.
* `gem search` ignores prereleases by default because it considers remote gems.
* `gem query` ignores prereleases by default because it considers remote gems.

By the way, I'm now reading the help of the `gem query` command:

>   Description:
>
>    The query command is the basis for the list and search commands.    
>
>    You should really use the list and search commands instead.  This command is too hard to use.

deivid-rodriguez:
Actually, I'm wrong. `gem list` and `gem search` inherit from this command, so yeah. This will change the behavior of those commands, I didn't know that.

deivid-rodriguez:
I think I would deprecate `gem query`?

bronzdoc:
Since this is just Adding prereleases to the already expected output we could shipit as a feature?

> I think I would deprecate gem query

I guess we could do that

deivid-rodriguez:
Hei @bronzdoc! I think the main confusion with this came with the `query` command, but since we are deprecating that in #2967, I don't think changing this is worth it. I think people will consider reasonable that `gem search` does not list prereleases by default, but `gem list` does, so I would close this PR I think.

bronzdoc:
@deivid-rodriguez I agree

PR #2756, opened 2019-04-30
Added supported versions of Ruby.

### LINKED ISSUE #2586: Documenting incompatibility of Ruby 2.5 with RubyGems < 2.6.10?
I don't know the best way to handle the following, since we're talking about older versions of RubyGems. If there's no good action to be taken, I understand.

It appears that Ruby 2.5+ is incompatible with RubyGems < 2.6.10, as any gem installation will lead to the error:

```
ERROR:  While executing gem ... (ArgumentError)
    wrong number of arguments (given 1, expected 0)
```

This is due to a `require 'fileutils'` [call](https://github.com/rubygems/rubygems/blob/v2.6.9/lib/rubygems/installer.rb#L162) in the initializer for `Gem::Installer`. Before RubyGems 2.6.10, `require` had a [call to `gem`](https://github.com/rubygems/rubygems/blob/v2.6.9/lib/rubygems/core_ext/kernel_require.rb#L44-L48) that was intended to call `Kernel.gem` but was actually calling a `gem` attribute method defined in `Gem::Installer` itself in this situation, since `require` was called as a method on the installer object itself. A change to calling `Kernel.gem` directly in #1822 fixed this issue. This particular issue does not occur in Ruby < 2.5 because the call to `gem` only occurs when trying to require a default gem, and `fileutils` only became a default gem in Ruby 2.5.

For reference, I have a slightly longer version of this explanation in a Stack Overflow answer (https://stackoverflow.com/a/54046354/7151673). It looks like this issue was also diagnosed in https://github.com/rubygems/rubygems/issues/2224#issuecomment-371020699, though I didn't see it until after I did all of my own debugging. 🤷‍♂️ 

Is there anything more that can be done here to document this? In my own case, I had an old installation of RVM that by default was installing the latest version of RubyGems 2.2 even when I installed Ruby 2.5 or 2.6 recently. That could be considered an RVM issue for sure, but the Stack Overflow questions I found indicate that something similar might be happening with rbenv.

Thanks for all your work here!

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

hsbt:
# Description:

Fixes #2586

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,9 @@ For more information about how to use RubyGems, see our RubyGems basics guide at
 
 ## Requirements
 
-* RubyGems 2.6 support Ruby 2.4 or lower version of Ruby.
-* RubyGems 2.7 support Ruby 2.5 or lower version of Ruby.
-* RubyGems 3.0 support Ruby 2.3+.
+* RubyGems 2.6 supports Ruby 2.4 or lower.
+* RubyGems 2.7 supports Ruby 2.5 or lower.
+* RubyGems 3.0 supports Ruby 2.3 or later.
 
 ## Installation
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ For more information about how to use RubyGems, see our RubyGems basics guide at
 
 * RubyGems 2.6 support Ruby 2.4 or lower version of Ruby.
 * RubyGems 2.7 support Ruby 2.5 or lower version of Ruby.
-* RubyGems 3.0 support Ruby 2.6 or lower version of Ruby.
+* RubyGems 3.0 support Ruby 2.3+.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@hsbt

One could read this to mean that RG 3.0 supports any version of Ruby < 2.7.

Hence, should the lower constraint also be mentioned?

hsbt:
@MSP-Greg Good catch. I update it.

README.md
```diff
@@ -25,6 +25,12 @@ Finally, inside your Ruby program, load the Nokogiri gem and start parsing your
 
 For more information about how to use RubyGems, see our RubyGems basics guide at [guides.rubygems.org](http://guides.rubygems.org/rubygems-basics/)
 
+## Requirements
+
+* RubyGems 2.6 support Ruby 2.4 or lower version of Ruby.
+* RubyGems 2.7 support Ruby 2.5 or lower version of Ruby.
+* RubyGems 3.0 support Ruby 2.3+.
```
deivid-rodriguez: I'm not great at English but I think "RubyGems 2.6 supports Ruby 2.4 or lower" reads better.

hsbt:
@bundlerbot r+

PR #2757, opened 2019-05-01
Refactor tests

deivid-rodriguez:
# Description:

I was working on adding some tests for installing defafult gems, and I noticed that it's pretty hard to work with installer tests, because the base setup does too much, so sometimes you need to reset state at the beginning of each test to go back to a clean slate.

Instead, I've removed all the common setup, and make each test explicitly do only the stuff it needs to do.

This allows to remove a lot of state resetting from many tests. 

I also included a little follow up to the addition of jruby in https://github.com/rubygems/rubygems/pull/2757/commits/309de0ed2376c8efb248e030c9847e0feaf18cfa by undoing some skips that are not really necessary.
 
# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_commands_uninstall_command.rb
+++ b/test/rubygems/test_gem_commands_uninstall_command.rb
@@ -73,15 +73,6 @@ class TestGemCommandsUninstallCommand < Gem::InstallerTestCase
 
   def test_execute_removes_executable
     initial_install
-    ui = Gem::MockGemUi.new
-
-    installer = util_setup_gem ui
-
-    build_rake_in do
-      use_ui ui do
-        installer.install
-      end
-    end
 
     if win_platform?
       assert File.exist?(@executable)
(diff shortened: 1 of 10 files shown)


test/rubygems/test_gem_commands_uninstall_command.rb
```diff
@@ -78,7 +79,7 @@ def test_execute_dependency_order
   def test_execute_removes_executable
     ui = Gem::MockGemUi.new
 
-    util_setup_gem ui
+    @installer = util_setup_gem ui
```
bronzdoc: Why we need to assign to an instance variable?

deivid-rodriguez: No particular reason.

I migrated everything from the `setup` method to be defined locally to each test method but kept the instance variables because the helper methods still defined some instance variables that were still referenced in the tests. But the `@installer` variable in particular can be safely changed to be a local variable. I can do that in a follow up commit :+1:.

deivid-rodriguez: I addressed your feedback, and also fixed a state thas was surfaced by this refactoring in  https://github.com/rubygems/rubygems/pull/2757/commits/aa14595f38f2f3424d2960db4463aedbba6f2e78.

bronzdoc:
@bundlerbot r+

PR #2764, opened 2019-05-14
Properly fetch `Gem#latest_spec_for` with multiple sources

kevlogan90:
# Description:
There are times when getting the latest spec for a gem is returning the wrong result. If multiple sources are used, and a gem exists in both, then [spec_fetcher](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/spec_fetcher.rb#L86) will return a spec for each source the gem is found in and [sort it by oldest to newest](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/spec_fetcher.rb#L136).

 Returning the last item from the list of specs resolves this issue.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -888,6 +888,27 @@ class TestGem < Gem::TestCase
     assert_equal gems['a-2'], spec
   end
 
+  def test_self_latest_spec_for_multiple_sources
+    uri = 'https://example.sample.com/'
+    source = Gem::Source.new(uri)
+    source_list = Gem::SourceList.new
+    source_list << Gem::Source.new(@uri)
+    source_list << source
+    Gem.sources.replace source_list
+
+    spec_fetcher(uri) do |fetcher|
+      fetcher.spec 'a', 1.1
+    end
+
+    gems = spec_fetcher do |fetcher|
+      fetcher.spec 'a', 1
+      fetcher.spec 'a', '3.a'
+      fetcher.spec 'a', 2
+    end
+    spec = Gem.latest_spec_for 'a'
+    assert_equal gems['a-2'], spec
+  end
+
   def test_self_latest_rubygems_version
     spec_fetcher do |fetcher|
       fetcher.spec 'rubygems-update', '1.8.23'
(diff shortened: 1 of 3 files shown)


bronzdoc:
Hey @kevlogan90 thanks for your contribution!

Could you add test for this change? thanks!

kevlogan90:
Hey @bronzdoc I've added some tests. Please let me know if there are any changes that need to be made.

bronzdoc:
@kevlogan90 i ran your tests without the code modifications you made and your tests still pass, so i'm not sure what is this fixing exactly.

kevlogan90:
Sorry @bronzdoc , I see what you mean. I updated the test. It should fail for you now using the unmodified code.

test/rubygems/test_gem.rb
```diff
@@ -764,6 +785,29 @@ def test_self_latest_version_for
...
+
+    spec_fetcher do |fetcher|
+      fetcher.spec 'a', 1
+      fetcher.spec 'a', 2
+      fetcher.spec 'a', '3.a'
+    end
+
+    version = Gem.latest_version_for 'a'
+
+    assert_equal Gem::Version.new(2), version
```
bronzdoc: Same here, It seems that `a-2` is  only in one resource not both

test/rubygems/test_gem.rb
```diff
@@ -740,6 +740,27 @@ def test_self_latest_spec_for
...
+      fetcher.spec 'a', 1.1
+    end
+
+    gems = spec_fetcher do |fetcher|
+      fetcher.spec 'a', 1
+      fetcher.spec 'a', '3.a'
+      fetcher.spec 'a', 2
+    end
+    spec = Gem.latest_spec_for 'a'
+    assert_equal gems['a-2'], spec
```
bronzdoc: the problem happens when `multiple sources are used, and a gem exists in both`
but this test is not testing that, it seems that `a-2` is  only in one of the resources not both, i'm missing something?

kevlogan90: Am I misunderstanding? The gem is `a`, with different versions `1`, `1.1`, `2`, `3.a`. It's possible for not all versions to be in all sources. In my case, which is likely an edge case, the gem was forked to an internal hosted rubygems source, and isn't updated regularly.

deivid-rodriguez:
@kevlogan90 I'm sorry for the lack of feedback here? The PR looks good to me in principle, are you still interested? If so, please rebase it so we trigger the latest CI again? If not, please let us know so that we take it over.

deivid-rodriguez:
No further updates, so I'll take over this PR. Thanks @kevlogan90!

PR #2766, opened 2019-05-16
Fix setup command if format_executable is true by default

jeremyevans:
# Description:

Some operating systems that support multiple versions of
ruby set `:format_executable` by default.  Previously, the
setup command handled things for `gem` but not for `bundle`,
resulting in the following test failure:

```
TestGemCommandsSetupCommand#test_env_shebang_flag:
Errno::ENOENT: No such file or directory @ rb_sysopen - /tmp/test_rubygems_30662/install/bin/bundle26
    test/rubygems/test_gem_commands_setup_command.rb:149:in `read'
    test/rubygems/test_gem_commands_setup_command.rb:149:in `test_env_shebang_flag'
```

This patch fixes the issue by passing the `:format_executable`
option when installing the bundler gem during setup.

# Tasks:

- [X] Write tests (fixes existing test failure)

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -429,7 +429,7 @@ By default, this RubyGems will install gem as:
     Dir.chdir("bundler") do
       built_gem = Gem::Package.build(bundler_spec)
       begin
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks @jeremyevans!

@bundlerbot r+

PR #2767, opened 2019-05-18
Expose windows path normalization utility

deivid-rodriguez:
# Description:

I need this exact piece of logic in `bundler` (https://github.com/bundler/bundler/pull/7161#discussion_r284914104), so I guess we should move it to some place where it makes more sense to reuse it.

I don't have a strong opinion though, maybe it's better to not bother and duplicate the logic.

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2178, opened 2018-02-05
Fix regression of destdir on Windows platform

hsbt:
# Description:

Followup https://github.com/rubygems/rubygems/issues/2134 and https://github.com/rubygems/rubygems/pull/2169

`Gem::Specification.default_specifications_dir` returns full path contained "C:/..." string on Windows platform. Windows couldn't handle C:/... suffix string like "C:/foo/C:/bar".

I ignored destdir options with upgrade installer when it invoked on Windows.
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2768, opened 2019-05-18
Don't apply `--destdir` twice when running `setup.rb`

alyssais:
# Description:

Prior to this patch, if I ran:

    ruby setup.rb --destdir /foo

Then Bundler files would be written into /foo/foo, because destdir was being prepended, even though bundler_spec.bin_dir already included destdir.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -385,6 +385,18 @@ By default, this RubyGems will install gem as:
 
     bundler_spec = Gem::Specification.load(default_spec_path)
 
+    # The base_dir value for a specification is inferred by walking up from the
+    # folder where the spec was `loaded_from`. In the case of default gems, we
+    # walk up two levels, because they live at `specifications/default/`, whereas
+    # in the case of regular gems we walk up just one level because they live at
+    # `specifications/`. However, in this case, the gem we are installing is
+    # misdetected as a regular gem, when it's a default gem in reality. This is
+    # because when there's a `:destdir`, the `loaded_from` path has changed and
+    # doesn't match `Gem.default_specifications_dir` which is the criteria to
+    # tag a gem as a default gem. So, in that case, write the correct
+    # `@base_dir` directly.
+    bundler_spec.instance_variable_set(:@base_dir, File.dirname(File.dirname(specs_dir)))
+
     # Remove gemspec that was same version of vendored bundler.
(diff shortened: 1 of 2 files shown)


bronzdoc:
ok with a test 👍

hsbt:
Note: I added this statement at https://github.com/rubygems/rubygems/pull/2178. 

> because destdir was being prepended

Can you show the above code? I'm okay to merge this after confirming it.

deivid-rodriguez:
Hi!

I tried to add a test for this fix, and the test made another issue surface. I think it should be correct now. The issue was not easy to fix so I added a hack which I tried to explain with a very long comment.

To compare, the result of `ruby setup.rb --destdir /tmp/foo`:

#### Without my extra fix:

```
$ tree /tmp/foo/home/deivid/.rbenv/versions/2.6.6/lib/ruby/gems/
/tmp/foo/home/deivid/.rbenv/versions/2.6.6/lib/ruby/gems/
└── 2.6.0
    └── specifications
        ├── default
        │   └── bundler-2.2.0.rc.1.gemspec
        └── gems
            └── bundler-2.2.0.rc.1
                └── exe
                    └── bundle

6 directories, 2 files
```

#### With my extra fix

```
$ tree /tmp/foo/home/deivid/.rbenv/versions/2.6.6/lib/ruby/gems/
/tmp/foo/home/deivid/.rbenv/versions/2.6.6/lib/ruby/gems/
└── 2.6.0
    ├── gems
    │   └── bundler-2.2.0.rc.1
    │       └── exe
    │           └── bundle
    └── specifications
        └── default
            └── bundler-2.2.0.rc.1.gemspec

6 directories, 2 files
```

alyssais:
> Hi!
>
> I tried to add a test for this fix, and the test made another issue
> surface. I think it should be correct now. The issue was not easy to
> fix so I added a hack which I tried to explain with a very long
> comment.

I'm afraid it's been so long I've basically lost all the context for
this change, so I don't think I can usefully comment, other than I do
remember trying to add a test for this, and running into some problem,
which might have been the same one.

deivid-rodriguez:
I forgot about this PR but it seems all good. We might be able to get rid of the hack when we move #2909 forward. We'll have the regression test in place to verify that we don't break this particular functionality.

deivid-rodriguez:
Actually I'm doing some real world testing and getting some issues. Will update this.

deivid-rodriguez:
I'm stupid. I was doing the realworld testing with the fix reverted 😂.

PR #2777, opened 2019-05-27
Update the certificate files to make the test pass on Debian 10

mame:
# Description:

The old certificate files (for example, test/rubygems/ca_cert.pem) were
signed by SHA1.  This message digest is considered too weak and rejected
by OpenSSL 1.1.1 or later.  Because of this, the test suite does not
pass on Debian 10.

https://rubyci.org/logs/rubyci.s3.amazonaws.com/debian/ruby-master/log/20190527T123003Z.fail.html.gz#test%2Frubygems

This change regenerates the files.
A shell script for the regeneration (util/create_certs.sh) is also
added.
______________

# Tasks:

- [x] Write tests


hsbt:
@bundlerbot r+

util/create_certs.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# cp /etc/ssl/openssl.cnf . # copied from OpenSSL 1.1.1b source
```
deivid-rodriguez: Should we vendor this file (or whatever part is needed for generating our certs), so that the certificate generation works by default?

mame: In my opinion, it is not necessary because this script won't be used in near future.

The old cert files were created in 2012.  Because SHA1 is weak nowadays, we need to regenerate the files.  So I guess that the next time to regenerate will be mid or late-2020s.  When we do regenerate, we should fetch the latest version of openssl.cnf at that time.  Moreover, we may need to modify the script itself.  This script is just a memorandum for the work.

deivid-rodriguez: I see.

I wouldn't ship the script in that case and copy it in the commit message instead since it's about tracking work and not about providing a script that developers are meant to use.

In any case, no big deal, thanks so much for fixing this! :smiley:

Issue #2773, opened 2019-05-22
Managing gem sources via CLI leaks sources defined outside of ~/.gemrc into ~/.gemrc


soulcutter:
I'm having a problem or would like to suggest a feature.

Some context: I wrote [a blog post suggesting a way of managing gem sources for people who commit their dotfiles to public repositories](http://www.soulcutter.com/articles/hiding-gemrc-credentials-in-dotfiles.html).

My current problem is that any time a `gem` command is used mutates gem sources, all sources are written to `~/.gemrc` . This is unexpected behavior since I expect that sources I define outside of `~/.gemrc` to remain outside of `~/.gemrc` so that they do not accidentally get published to a public dotfiles repository.

Steps to reproduce:

1. Create a file `~/.gemrc.local` with contents such as:
```yaml
---
:sources:
- https://rubygems.org/
- https://asdf@gemfury.io/me/
```
2. `$ export GEMRC=~/.gemrc.local`
3. `$ gem sources --remove https://rubygems.org/`
4. `$ cat ~/.gemrc` will show `https://asdf@gemfury.io/me/`

I can think of a couple of approaches that would be less-unexpected. As a side-note `$ gem sources` will still list `https://rubygems.org/` as in-effect because `~/.gemrc.local` is not edited at all and therefor the intent of the command was not fulfilled

1. The merging behavior of the gem command's configuration keeps track of where `sources` are defined and edits that location. This might be difficult (impossible?) to achieve.
2. The merging behavior of the gem command's configuration keeps track of where `sources` are defined and if that location is not `~/.gemrc` it errors / warns you that it is unable to mutate sources because sources are defined elsewhere (hinting that you will have to manually resolve that problem).

There may be a larger approach to examine, which is considering `sources` as a special configuration key that doesn't have the "shallow-merge" behavior that I describe in my blog post, and instead has a different form of configuration behavior. That could be heavy-handed for resolving this specific issue, but figured I'd throw that out there.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
3.0.3

$ echo $GEMRC
/Users/bradleyschaefer/.gemrc.local
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

soulcutter:
[...]

PR #2779, opened 2019-05-29
Write to the correct config file(.gemrc)

bronzdoc:
# Description:
Closes https://github.com/rubygems/rubygems/issues/2773

Currently Rubygems config file(`.gemrc`) is not set to the value of `GEMRC` environment variable if set, leading to unexpected behaviour as described in https://github.com/rubygems/rubygems/issues/2773 this PR will make the config file to be set to the value of `GEMRC` if provided.

______________

### DIFF
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -472,6 +472,7 @@ if you believe they were disclosed to a third party.
   protected :hash
 
   private
+
   def set_config_file_name(args)
     @config_file_name = ENV["GEMRC"]
     need_config_file_name = false
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -473,7 +473,7 @@ if you believe they were disclosed to a third party.
 
   private
   def set_config_file_name(args)
-    @config_file_name = nil
+    @config_file_name = ENV["GEMRC"] || nil
     need_config_file_name = false
 
     args.each do |arg|
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -334,7 +334,7 @@ if you believe they were disclosed to a third party.
     yaml_errors = [ArgumentError]
     yaml_errors << Psych::SyntaxError if defined?(Psych::SyntaxError)
 
-    return {} unless filename and File.exist? filename
+    return {} unless filename && !filename.empty? && File.exist?(filename)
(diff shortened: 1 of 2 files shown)


soulcutter:
This seems like a good solution. Thanks!

lib/rubygems/config_file.rb
```diff
@@ -486,4 +471,21 @@ def ==(other) # :nodoc:
   attr_reader :hash
   protected :hash
 
+  private
+  def set_config_file_name(args)
+    @config_file_name = ENV["GEMRC"] || nil
```
deivid-rodriguez: I think you can simplify this to

```suggestion
    @config_file_name = ENV["GEMRC"]
```

The only case when I think it would make a difference is when `ENV["GEMRC"]` is explicitly set to `false`, but even in that case, the `load_file` method should shortcircuit in the same way:

https://github.com/rubygems/rubygems/blob/bc6d07d0d9690003c86664b76fb0d8eee5206492/lib/rubygems/config_file.rb#L352

bronzdoc: Done 👍

test/rubygems/test_gem_config_file.rb
```diff
@@ -19,7 +19,7 @@ def setup
     Gem::ConfigFile::PLATFORM_DEFAULTS.clear
 
     @env_gemrc = ENV['GEMRC']
-    ENV['GEMRC'] = ''
+    ENV['GEMRC'] = nil
```
deivid-rodriguez: What's the behavior now with `ENV["GEMRC"] = ''`. I guess we could keep the current behaviour by checking for empty too in the `load_file` early return.

bronzdoc: Done :)

lib/rubygems/config_file.rb
```diff
@@ -486,4 +471,21 @@ def ==(other) # :nodoc:
   attr_reader :hash
   protected :hash
 
+  private
```
deivid-rodriguez: Our majority style is to add a blank line after private. I created #2785 to enable the corresponding cop.

bronzdoc: Done

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
Awesome!

PR #2780, opened 2019-05-29
Fix for large values in UID/GID fields in tar archives

conf:
# Description:
Fixes https://github.com/rubygems/rubygems/issues/2213.

Current check for octal values is too strict for some fields.
For example, on Windows workstations UID/GID values can be quite
large for domain members and cannot fit into 8 bytes when written
as octal values, so we need to allow 256-based values.

Both GNU tar and BSD tar implement this behavior, see
http://git.savannah.gnu.org/cgit/tar.git/tree/src/list.c#n861
and `tar_atol` implementation in
https://opensource.apple.com/source/libarchive/libarchive-41.20.1/libarchive/libarchive/archive_read_support_format_tar.c.auto.html
______________

- [x] Write tests


bronzdoc:
@bundlerbot r+

conf:
@bronzdoc Thank you for a quick merge ❤️ ! Any idea when this gets released?

jmartin-r7:
Adding another vote for getting this out in a release. 🙏

hsbt:
Released 3.0.5 with this fix: https://blog.rubygems.org/2019/08/16/3.0.5-released.html

Issue #2772, opened 2019-05-21
stringio is required eagerly by RubyGems


eregon:
Currently, RubyGems requires `stringio` eagerly:
https://github.com/rubygems/rubygems/blob/d470f1787a94f751c907e8299ec44f61b7689432/lib/rubygems/specification.rb#L17

The only usage in that file though seems in
https://github.com/rubygems/rubygems/blob/d470f1787a94f751c907e8299ec44f61b7689432/lib/rubygems/specification.rb#L2538
where it could be required on demand.

The effect of this is that `StringIO` is defined in MRI, without `require`, which makes some gems accidentally rely on RubyGems loading it:
```
$ ruby -e 'p StringIO'                                                
StringIO
```

How about requiring `stringio` lazily in RubyGems, so users of `stringio` must require it explicitly like other stdlibs, and notice easily the missing `require` if `--disable-gems` is used?

This issue is related to:
  - [x] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

eregon:
FWIW, there are quite a few files loaded eagerly by RubyGems on startup, making startup slower:
```
$ ruby --disable-gems -ve 'puts $"'         
ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-linux]
enumerator.so
thread.rb
rational.so
complex.so
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/x86_64-linux/enc/encdb.so
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/x86_64-linux/enc/trans/transdb.so
```

```
$ ruby -ve 'puts $"' | grep -v rubygems | grep -v did_you_mean
ruby 2.6.2p47 (2019-03-13 revision 67232) [x86_64-linux]
enumerator.so
thread.rb
rational.so
complex.so
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/x86_64-linux/enc/encdb.so
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/x86_64-linux/enc/trans/transdb.so
Required by RubyGems:
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/x86_64-linux/rbconfig.rb
/home/eregon/.rubies/ruby-2.6.2/lib/ruby/2.6.0/delegate.rb
[...]

PR #2781, opened 2019-05-29
Lazy require stringio

bronzdoc:
# Description:
close https://github.com/rubygems/rubygems/issues/2772
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/mock_gem_ui.rb
+++ b/lib/rubygems/mock_gem_ui.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-require 'stringio'
 require 'rubygems/user_interaction'
 
 ##
@@ -46,6 +45,7 @@ class Gem::MockGemUi < Gem::StreamUI
   end
 
   def initialize(input = "")
+    require 'stringio'
     ins = StringIO.new input
     outs = StringIO.new
     errs = StringIO.new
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Nice! Perhaps we should have a test that boots rubygems and makes sure `stringio` (and with time other unnecessary default gems) are not required?

bronzdoc:
> Perhaps we should have a test that boots rubygems and makes sure stringio
What would be your approach to such test @deivid-rodriguez ?

deivid-rodriguez:
Perhaps running something like `ruby --disable-gems -Ilib -rrubygems -e 'exit require("stringio")'` in TravisCI would be enough for now?

bronzdoc:
> Perhaps running something like ruby --disable-gems -Ilib -rrubygems -e 'exit require("stringio")' in TravisCI would be enough for now?

@deivid-rodriguez you mean as part of our `util/ci` script?

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
> @deivid-rodriguez you mean as part of our util/ci script?

Yeah, that's what I meant. But your PR was good I agree with @hsbt, we can add the check on a separate PR. Thanks for doing this!

Issue #2776, opened 2019-05-26
#to_ruby isn't handling OpenSSL::PKey::RSA instances


zenspider:
If I add the following to any of my gems' Rakefiles:

```ruby
pp hoe.spec
pp hoe.spec.to_ruby
```

I get the following output:

```
10003 % rake wtf
Gem::Specification.new do |s|
  s.name = "autotest"
  s.version = Gem::Version.new("5.0.0")
  s.installed_by_version = Gem::Version.new("0")
  s.authors = ["Ryan Davis"]
  s.cert_chain = ["/Users/ryan/.gem/gem-public_cert.pem"]
  s.date = Time.utc(2019, 5, 26)
  s.dependencies = [Gem::Dependency.new("minitest-autotest",
    Gem::Requirement.new(["~> 1.0"]),
    :runtime),
   Gem::Dependency.new("rdoc",
    Gem::Requirement.new([">= 4.0", "< 7"]),
    :development),
   Gem::Dependency.new("hoe", Gem::Requirement.new(["~> 3.17"]), :development)]
  s.description = "This is a stub gem to fix the confusion caused by autotest being part\n" +
  "of the ZenTest suite."
  s.email = ["ryand-ruby@zenspider.com"]
  s.extra_rdoc_files = ["History.txt", "Manifest.txt", "README.txt"]
  s.files = ["History.txt", "Manifest.txt", "README.txt", "Rakefile"]
  s.homepage = "http://www.zenspider.com/ZSS/Products/ZenTest/"
  s.licenses = ["MIT"]
  s.rdoc_options = ["--main", "README.txt"]
  s.require_paths = ["lib"]
  s.rubygems_version = "3.0.2"
  s.signing_key = #<OpenSSL::PKey::RSA:0x00007fc27405da50>
  s.specification_version = 4
  s.summary = "This is a stub gem to fix the confusion caused by autotest being part of the ZenTest suite."
  end
rake aborted!
Gem::Exception: ruby_code case not handled: OpenSSL::PKey::RSA
/Users/ryan/Work/p4/zss/src/autotest/dev/Rakefile:19:in `<top (required)>'
(See full trace by running task with --trace)
```

Using rubygems 3.0.2 currently.

The code in question:

```ruby
  def ruby_code(obj)
    case obj
    when String            then obj.dump + ".freeze"
    when Array             then '[' + obj.map { |x| ruby_code x }.join(", ") + ']'
    when Hash              then
[...]

PR #2782, opened 2019-05-30
Make Gem::Specification#ruby_code handle OpenSSL::PKey::RSA objects

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2776

@zenspider what you think?
______________

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2303,6 +2303,7 @@ class Gem::Specification < Gem::BasicSpecification
     when Time              then obj.strftime('%Y-%m-%d').dump
     when Numeric           then obj.inspect
     when true, false, nil  then obj.inspect
+    when OpenSSL::PKey::RSA then obj.class
     when Gem::Platform     then "Gem::Platform.new(#{obj.to_a.inspect})"
     when Gem::Requirement  then
       list = obj.as_list
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2293,19 +2293,19 @@ class Gem::Specification < Gem::BasicSpecification
 
   def ruby_code(obj)
     case obj
-    when String            then obj.dump + ".freeze"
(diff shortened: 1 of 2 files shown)


zenspider:
That seems fine to me, but I suspect this is used to serialize specs still. Skipping outright might be a better option... I'd want someone more in the know to weigh in on that. I haven't maintained rubygems in ages.

But for MY purposes, as long as it doesn't crash I'm happy.

zenspider:
Also, thank you. That was very timely.

deivid-rodriguez:
I would vote for completely skipping this field too. As far as I know, this is only used at `gem build` time, and it's no longer needed once the deserialized spec file is installed to the final user system?

bronzdoc:
Removing the signing key trigger a couple of errors, and made me notice we are printing the `cert_chain` too but since it's an array it's handled by:

```ruby
when Array              then '[' + obj.map { |x| ruby_code x }.join(", ") + ']'¬
```

so kinda the output we want to avoid by not doing `to_s` or to `to_text` from a OpenSSL::PKey::RSA object, should we skip the `cert_chain` attribute too?

bronzdoc:
Build failures are not related to this change, I don't understand why travis is having issues with bundler in this particular PR 🤔

zenspider:
The tests know how much I hate bundler and are getting back at me. 🙁

MSP-Greg:
I think this may be the result of the older version of bundler currently being used.

I believe the changes in https://github.com/bundler/bundler/commit/e742c3d5f458a4a59cf0eaab2567eca844f956d1 / https://github.com/bundler/bundler/pull/7100 will fix this.

But, it breaks Ruby master/trunk/ruby-head builds when they are used for CI testing. That fix is at https://github.com/bundler/bundler/pull/7248

zenspider:
... status? There's been 4 releases since this came up...

bronzdoc:
@bundlerbot r+

PR #2789, opened 2019-06-06
Synchronize access to the Gem::Specification::LOAD_CACHE Hash

eregon:
The `Gem::Specification::LOAD_CACHE` Hash is accessed concurrently, notably when installing a gem with C extensions but does not use synchronization. This PR adds synchronization so that it works fine on implementations which do not (yet) have a thread-safe `Hash`, such as TruffleRuby and JRuby.

Furthermore, it ensures the cached `Gem::Specification` object is always the same for a given path, which was not guaranteed before. That seems a good guaranty to have.

This code is already applied in TruffleRuby's copy of RubyGems in the stdlib, but it needs to be upstreamed for updating RubyGems to keep working with TruffleRuby.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -109,6 +109,7 @@ class Gem::Specification < Gem::BasicSpecification
   # rubocop:disable Style/MutableConstant
   LOAD_CACHE = {} # :nodoc:
   # rubocop:enable Style/MutableConstant
+  LOAD_CACHE_MUTEX = Mutex.new
 
   private_constant :LOAD_CACHE if defined? private_constant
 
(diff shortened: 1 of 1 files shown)


eregon:
@hsbt Could you review this?

hsbt:
Seems fine. Thanks!

hsbt:
@bundlerbot r+

eregon:
@hsbt `jruby-head` hangs, which seems unrelated and seems to have already happened on master in https://travis-ci.org/rubygems/rubygems/jobs/538597876.
Could you retry the CI?

deivid-rodriguez:
@bundlerbot retry

eregon:
@hsbt @deivid-rodriguez Thank you for merging!

PR #2795, opened 2019-06-18
Task to install rubygems to local system

deivid-rodriguez:
# Description:

This PR adds a `rake install` task that installs the copy of rubygems to the users system. This is better than the `ruby setup.rb` alternative, because it's more similar to what actually happens in real life when users install `rubygems`.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -93,6 +93,11 @@ task :check_deprecations do
   end
 end
 
+desc "Install rubygems to local system"
+task :install => :package do
+  sh "gem install pkg/rubygems-update-#{v}.gem && update_rubygems"
+end
+
 desc "Release rubygems-#{v}"
 task :release => :prerelease do
   Rake::Task["package"].invoke


Rakefile
```diff
@@ -93,6 +93,11 @@ task :check_deprecations do
   end
 end
 
+desc "Install rubygems to local system"
+task :install => :package do
+  sh "gem install pkg/rubygems-update-#{v}.gem && update_rubygems"
```
bronzdoc: Should we use our own executable instead of the system executable?

deivid-rodriguez: @bronzdoc I did some work in https://github.com/rubygems/rubygems/compare/move_gem_to_exe to make it easier to always use the local version of the gem CLI for development tasks, but for now I would like to keep it consistent with the rest of the `Rakefile`, where system rubygems is used for all tasks.

bronzdoc: Fair 👍

bronzdoc:
@bundlerbot r+

PR #2796, opened 2019-06-19
Add an attr_reader to Gem::Installer for the package instance variable

djberg96:
A simple PR that adds an attr_reader to the `@package` attribute. I wanted this recently for a rubygems plugin where I wanted to get at the underlying package. I can do it currently via `instance_variable_get`, but this is nicer.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -65,6 +65,11 @@ class Gem::Installer
 
   attr_reader :options
 
+  ##
+  # The gem package instance.
+
+  attr_reader :package
+
   @path_warning = false
 
   @install_lock = Mutex.new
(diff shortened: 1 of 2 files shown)


djberg96:
Weird, specs passed locally for me on mac, is there some reason `@installer` is nil on windows?

djberg96:
@deivid-rodriguez Ok, fixed the tests. Look ok?

lib/rubygems/installer.rb
```diff
@@ -65,6 +65,11 @@ class Gem::Installer
 
   attr_reader :options
 
+  ##
+  # The gem package instance.
+
+  attr_reader :package
```
bronzdoc: Where this attribute will be set?

djberg96: It's set in the constructor.

bronzdoc:
@bundlerbot r+

PR #2798, opened 2019-06-20
Explicitly deprecate `rubyforge_project`

deivid-rodriguez:
# Description:

In https://github.com/rubygems/rubygems/pull/2436, we removed all the `rubyforge_project` functionality, but kept the gemspec writer as a noop, so that people's gemspecs won't break. However, we didn't deprecate the writer. We should do it, so that we can actually remove it in the future.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -732,6 +732,7 @@ class Gem::Specification < Gem::BasicSpecification
   # Formerly used to set rubyforge project.
 
   attr_writer :rubyforge_project
+  deprecate :rubyforge_project=, :none,       2019, 12
 
   ##
   # The Gem::Specification version of this gemspec.
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1786,8 +1786,10 @@ You may need to `gem install -g` to install missing gems
 
     EXPECTED
 
-    assert_output nil, expected do
-      Gem.use_gemdeps
+    Gem::Deprecate.skip_during do
+      assert_output nil, expected do
+        Gem.use_gemdeps
+      end
     end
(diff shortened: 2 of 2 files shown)


bronzdoc:
@bundlerbot r+

simi:
:wave: 

I tried to live on latest rubygems for a while but this change actually makes it really hard. Wouldn't be possible to relax a little on this deprecation, revert back warning for now and just deny rubygems.org pushes having this in gemspec to make ecosystem handle it over the time on its own?

Once we will ensure top 100/1000/??? gems are updated, we can roll this out again.

I have plan to check top x gems and if hosted on github to open pull requested by script. Anyway it would take some time to let gem maintainers merge those and release new versions.

deivid-rodriguez:
IMO the problem is not this deprecation itself, but that it is showing up for end users. Deprecations for gemspec attributes should only affect gem authors in my opinion.

This is what https://github.com/rubygems/rubygems/pull/3047 intends to fix and it is my plan to review it soon.

Also, note that #3108 alleviated this problem since gemspecs are not read for every command.

simi:
@deivid-rodriguez thanks for info. Yes, I share the same idea. Gem users can't do anything about that and they just need to wait until this is fixed in gem. Would be nice to move this warning for gem maintainers. Any plan to also reject invalid gemspec (having deprecated options) at rubygems.org?

deivid-rodriguez:
We have a `strict` option to `gem build` that makes `gem build` fail if the gemspec has warnings. But if the gem author is not using this option, and has built a valid package, I don't think we should prevent pushing the gem? I think we should prevent pushing once we actually _remove_ the attribute, not only deprecate it.

PR #2804, opened 2019-06-23
Remove missleading TODO comment in security_option.rb

bronzdoc:
# Description:
Remove missleading TODO comment. Can't use `@parser.accept` since not every class where this is used has a parser available. i.e `lib/rubygems/install_update_options.rb`
______________


bronzdoc:
@bundlerbot r+

Issue #2793, opened 2019-06-17
AWS SigV2 deprecation in s3 source


apakulov-stripe:
Would like to suggest a feature.

[S3 source](https://guides.rubygems.org/using-s3-source) has been added a part of Rubygems 3.0.

AWS announced that they are planning to deprecated SigV2 soon  https://aws.amazon.com/blogs/aws/amazon-s3-update-sigv2-deprecation-period-extended-modified/ and current S3-Source functionality relies on SigV2 implementation and does it [without](https://github.com/rubygems/rubygems/blob/15706d3e853f9d1fd7b6ddbdcf38cf9a83a345b9/lib/rubygems/remote_fetcher.rb#L348:L361) `aws-sdk` dependency.

Also, current implementation does not support temporary credentials: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

This issue is related to:
  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

apakulov-stripe:
It's possible to minimize the dependency list to these two gems:
- https://rubygems.org/gems/aws-sigv4/versions/1.1.0
- https://rubygems.org/gems/aws-sdk-core/versions/3.56.0

bronzdoc:
closed by https://github.com/rubygems/rubygems/pull/2807

PR #2807, opened 2019-06-25
Upgrading S3 source signature to AWS SigV4

apakulov-stripe:
# Description:
Initial problem description can be found in https://github.com/rubygems/rubygems/issues/2793

The following PR adopts AWS SigV4 implementation with query params according this spec: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

As a part this PR I also added the following functionality:
- Explicitly define a region in `.gemrc` for a give S3 bucket.
- Session-Token support for [temporary credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html).
- Added an option to define a `provider` in `.gemrc` to extract S3 credentials from different sources:
  - `env` extracts credentials from environment variable according to the AWS [env-vars ](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) naming convention
  - `instance_profile` uses EC2 metadata endpoint to extract unique EC2 instance credentials.

Supported `s3_source` options:
```
s3_source: {
  bucket-default: {
    id: "...",
    secret: "..."
  },
  bucket-with-region: {
    id: "...",
    secret: "...",
    region: "us-east-1"
  },
  bucket-with-region-and-token: {
    id: "...",
    secret: "...",
    security_token: "...",
    region: "us-west-2"
  },
  bucket-with-env-provider-and-region: {
    provider: "env",
    region: "us-west-2"
  },
  bucket-with-env-instance-profile-and-region: {
    provider: "instance_profile",
    region: "us-west-2"
  }
}
```
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -343,30 +343,70 @@ class Gem::RemoteFetcher
 
   protected
 
+  S3Config = Struct.new :access_key_id, :secret_access_key, :security_token, :region
+
   # we have our own signing code here to avoid a dependency on the aws-sdk gem
   # fortunately, a simple GET request isn't too complex to sign properly
   def sign_s3_url(uri, expiration = nil)
     require 'base64'
+    require 'digest'
     require 'openssl'
 
-    id, secret = s3_source_auth uri
+    s3_config = s3_source_auth uri
+    expiration ||= 3600
 
-    expiration ||= s3_expiration
-    canonical_path = "/#{uri.host}#{uri.path}"
-    payload = "GET\n\n\n#{expiration}\n#{canonical_path}"
-    digest = OpenSSL::HMAC.digest('sha1', secret, payload)
-    # URI.escape is deprecated, and there isn't yet a replacement that does quite what we want
-    signature = Base64.encode64(digest).gsub("\n", '').gsub(/[\+\/=]/) { |c| BASE64_URI_TRANSLATE[c] }
-    URI.parse("https://#{uri.host}.s3.amazonaws.com#{uri.path}?AWSAccessKeyId=#{id}&Expires=#{expiration}&Signature=#{signature}")
-  end
+    current_time = Time.now.utc
+    date_time = current_time.strftime("%Y%m%dT%H%m%SZ")
+    date = date_time[0,8]
 
-  def s3_expiration
-    (Time.now + 3600).to_i # one hour from now
+    credential_info = "#{date}/#{s3_config.region}/s3/aws4_request"
+    canonical_host = "#{uri.host}.s3.#{s3_config.region}.amazonaws.com"
+
+    canonical_params = {}
+    canonical_params['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256'
+    canonical_params['X-Amz-Credential'] = "#{s3_config.access_key_id}/#{credential_info}"
+    canonical_params['X-Amz-Date'] = date_time
+    canonical_params['X-Amz-Expires'] = expiration.to_s
+    canonical_params['X-Amz-SignedHeaders'] = 'host'
+    canonical_params['X-Amz-Security-Token'] = s3_config.security_token if s3_config.security_token
+
+    query_params = canonical_params.sort.to_h.map do |key, value|
+      "#{base64_uri_escape(key)}=#{base64_uri_escape(value)}"
+    end.join('&')
+
+    canonical_request = [
+      'GET',
+      uri.path,
+      query_params,
+      "host:#{canonical_host}",
+      '', # empty params
+      'host',
+      'UNSIGNED-PAYLOAD',
+    ].join("\n")
+
+    string_to_sign = [
+      "AWS4-HMAC-SHA256",
+      date_time,
+      credential_info,
+      Digest::SHA256.hexdigest(canonical_request)
+    ].join("\n")
+
+    date_key = OpenSSL::HMAC.digest('sha256', "AWS4" + s3_config.secret_access_key, date)
+    date_region_key = OpenSSL::HMAC.digest('sha256', date_key, s3_config.region)
+    date_region_service_key = OpenSSL::HMAC.digest('sha256', date_region_key, "s3")
+    signing_key = OpenSSL::HMAC.digest('sha256', date_region_service_key, "aws4_request")
+    signature = OpenSSL::HMAC.hexdigest('sha256', signing_key, string_to_sign)
+
+    URI.parse("https://#{canonical_host}#{uri.path}?#{query_params}&X-Amz-Signature=#{signature}")
   end
 
   BASE64_URI_TRANSLATE = { '+' => '%2B', '/' => '%2F', '=' => '%3D' }.freeze
 
   private
+
+  def base64_uri_escape(str)
+    str.gsub("\n", '').gsub(/[\+\/=]/) { |c| BASE64_URI_TRANSLATE[c] }
+  end
 
   def proxy_for(proxy, uri)
     Gem::Request.proxy_uri(proxy || Gem::Request.get_proxy_from_env(uri.scheme))
@@ -379,7 +419,7 @@ class Gem::RemoteFetcher
   end
 
   def s3_source_auth(uri)
-    return [uri.user, uri.password] if uri.user && uri.password
+    return S3Config.new(uri.user, uri.password, nil, 'us-east-1') if uri.user && uri.password
 
     s3_source = Gem.configuration[:s3_source] || Gem.configuration['s3_source']
     host = uri.host
@@ -388,11 +428,31 @@ class Gem::RemoteFetcher
     auth = s3_source[host] || s3_source[host.to_sym]
     raise FetchError.new("no key for host #{host} in s3_source in .gemrc", "s3://#{host}") unless auth
 
-    id = auth[:id] || auth['id']
-    secret = auth[:secret] || auth['secret']
-    raise FetchError.new("s3_source for #{host} missing id or secret", "s3://#{host}") unless id and secret
+    provider = auth[:provider] || auth['provider']
+    case provider
+    when 'env'
+      id = ENV['AWS_ACCESS_KEY_ID']
+      secret = ENV['AWS_SECRET_ACCESS_KEY']
+      security_token = ENV['AWS_SESSION_TOKEN']
+    when 'instance_profile'
+      require 'json'
+      credentials_response = fetch_http URI(EC2_METADATA_CREDENTIALS)
+      credentials = JSON.parse(credentials_response)
+      id = credentials['AccessKeyId']
+      secret = credentials['SecretAccessKey']
+      security_token = credentials['Token']
+    else
+      id = auth[:id] || auth['id']
+      secret = auth[:secret] || auth['secret']
+      raise FetchError.new("s3_source for #{host} missing id or secret", "s3://#{host}") unless id and secret
 
-    [id, secret]
+      security_token = auth[:security_token] || auth['security_token']
+    end
+
+    region = auth[:region] || auth['region'] || 'us-east-1'
+    S3Config.new(id, secret, security_token, region)
   end
+
+  EC2_METADATA_CREDENTIALS = "http://169.254.169.254/latest/meta-data/identity-credentials/ec2/security-credentials/ec2-instance"
 
 end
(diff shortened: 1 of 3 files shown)


apakulov-stripe:
@bronzdoc any ETAs on reviewing this PR?

bronzdoc:
Thanks for the PR @apakulov-stripe
Will give it a look tonight, if anyone else in the team has not looked at it first

lib/rubygems/remote_fetcher.rb
```diff
@@ -343,31 +343,71 @@ def close_all
 
   protected
 
+  S3Config = Struct.new :access_key_id, :secret_access_key, :security_token, :region
+
   # we have our own signing code here to avoid a dependency on the aws-sdk gem
```
simi: :thinking: Is this comment about simplicity still valid? This looks more complex on first sight. :eyes:

lib/rubygems/remote_fetcher.rb
```diff
@@ -343,31 +343,71 @@ def close_all
...
-  def s3_expiration
-    (Time.now + 3600).to_i # one hour from now
+    s3_config = s3_source_auth uri
+    expiration ||= 3600
+
+    current_time = Time.now.utc
+    date_time = current_time.strftime("%Y%m%dT%H%m%SZ")
+    date = date_time[0,8]
+
+    credential_info = "#{date}/#{s3_config.region}/s3/aws4_request"
```
bronzdoc: ```credential_info = "#{s3_config.access_key_id}/#{date}/#{s3_config.region}/s3/aws4_request"``` ?

lib/rubygems/remote_fetcher.rb
```diff
@@ -343,31 +343,71 @@ def close_all
...
-    digest = OpenSSL::HMAC.digest('sha1', secret, payload)
-    # URI.escape is deprecated, and there isn't yet a replacement that does quite what we want
-    signature = Base64.encode64(digest).gsub("\n", '').gsub(/[\+\/=]/) { |c| BASE64_URI_TRANSLATE[c] }
-    URI.parse("https://#{uri.host}.s3.amazonaws.com#{uri.path}?AWSAccessKeyId=#{id}&Expires=#{expiration}&Signature=#{signature}")
-  end
-
-  def s3_expiration
-    (Time.now + 3600).to_i # one hour from now
+    s3_config = s3_source_auth uri
+    expiration ||= 3600
```
bronzdoc: should this be  ` expiration ||= 86400` ?

lib/rubygems/remote_fetcher.rb
```diff
@@ -343,31 +343,71 @@ def close_all
...
+
+    canonical_params = {}
+    canonical_params['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256'
+    canonical_params['X-Amz-Credential'] = "#{s3_config.access_key_id}/#{credential_info}"
+    canonical_params['X-Amz-Date'] = date_time
+    canonical_params['X-Amz-Expires'] = expiration.to_s
+    canonical_params['X-Amz-SignedHeaders'] = 'host'
+    canonical_params['X-Amz-Security-Token'] = s3_config.security_token if s3_config.security_token
+
+    query_params = canonical_params.sort.to_h.map do |key, value|
```
bronzdoc: canonical_params need to be sorted?

lib/rubygems/remote_fetcher.rb
```diff
@@ -388,11 +428,31 @@ def s3_source_auth(uri)
...
 
-    [id, secret]
+      security_token = auth[:security_token] || auth['security_token']
+    end
+
+    region = auth[:region] || auth['region'] || 'us-east-1'
+    S3Config.new(id, secret, security_token, region)
   end
 
+  EC2_METADATA_CREDENTIALS = "http://169.254.169.254/latest/meta-data/identity-credentials/ec2/security-credentials/ec2-instance"
```
bronzdoc: This url(http://169.254.169.254/latest/meta-data/identity-credentials/ec2/security-credentials/ec2-instance) is unreachable

lib/rubygems/remote_fetcher.rb
```diff
@@ -343,31 +343,71 @@ def close_all
...
+
+    current_time = Time.now.utc
+    date_time = current_time.strftime("%Y%m%dT%H%m%SZ")
+    date = date_time[0,8]
+
+    credential_info = "#{date}/#{s3_config.region}/s3/aws4_request"
+    canonical_host = "#{uri.host}.s3.#{s3_config.region}.amazonaws.com"
+
+    canonical_params = {}
+    canonical_params['X-Amz-Algorithm'] = 'AWS4-HMAC-SHA256'
```
bronzdoc: Can you use double quotes for strings?

lib/rubygems/remote_fetcher.rb
```diff
@@ -388,11 +428,31 @@ def s3_source_auth(uri)
...
+      require 'json'
+      credentials_response = fetch_http URI(EC2_METADATA_CREDENTIALS)
+      credentials = JSON.parse(credentials_response)
+      id = credentials['AccessKeyId']
+      secret = credentials['SecretAccessKey']
+      security_token = credentials['Token']
+    else
+      id = auth[:id] || auth['id']
+      secret = auth[:secret] || auth['secret']
+      raise FetchError.new("s3_source for #{host} missing id or secret", "s3://#{host}") unless id and secret
```
bronzdoc: Can you use  `&&` instead of `and`?

bronzdoc:
@sonalkr132 want to give it a look?

lib/rubygems/remote_fetcher.rb
apakulov-stripe: `credential_info` is later used in `string_to_sign`

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Yes, it's not mentioned in the doc I shared, but it's mentioned in this one: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html

> CanonicalQueryString specifies the URI-encoded query string parameters. You URI-encode name and values individually. You must also sort the parameters in the canonical query string alphabetically by key name. The sorting occurs after encoding. The query string in the following URI example is prefix=somePrefix&marker=someMarker&max-keys=20

lib/rubygems/remote_fetcher.rb
apakulov-stripe: This endpoint is only available on EC2 instances. For more details checkout: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Removed this line in fb62d30.

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Fixed in fb62d30.

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Fixed in fb62d30.

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Fixed in fb62d30.

lib/rubygems/remote_fetcher.rb
bronzdoc: Sorry for not being specific, i mean in the whole PR 🙇

lib/rubygems/remote_fetcher.rb
bronzdoc: Ahh i see 👍

simi:
:thinking: Is there any chance to move all S3 only related logic to some separated class and use it from fetcher to make fetcher less complex? Now it mixes fetcher and S3 logic IMHO.

bronzdoc:
> Is there any chance to move all S3 only related logic to some separated class and use it from fetcher to make fetcher less complex?

IMO that is a good idea @simi

lib/rubygems/remote_fetcher.rb
apakulov-stripe: Addressed in c30d21ec

apakulov-stripe:
@simi @bronzdoc I moved S3 signature logic into a separate class - `S3URISigner`. Please take another look.

lib/rubygems/s3_uri_signer.rb
```diff
@@ -0,0 +1,166 @@
...
+
+  attr_accessor :uri
+
+  def initialize(uri)
+    @uri = uri
+  end
+
+  ##
+  # Signs S3 URI using query-params according to the reference: https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html
+  def sign(expiration = nil)
```
bronzdoc: Why not make this default to `86400` instead of doing `expiration ||= 86400` ?

lib/rubygems/remote_fetcher.rb
```diff
@@ -284,10 +285,19 @@ def fetch_path(uri, mtime = nil, head = false)
   end
 
   def fetch_s3(uri, mtime = nil, head = false)
-    public_uri = sign_s3_url(uri)
+    begin
+      public_uri = s3_uri_signer(uri).sign
```
simi: Do we need `s3_uri_signer` method? Isn't it possible to use `Gem::S3URISigner.new(uri).sign` or make shortcut as a `Gem::S3URISigner.sign(uri)` here?

simi:
:thinking: Since the only needed S3 related method in fetcher is `fetch_s3` now, it will make sense for me to add some API to register custom uri scheme handler and inject it into list of known schemes at https://github.com/rubygems/rubygems/blob/6a1856517fd1455b4626e4c57b2403d1d352bde5/lib/rubygems/remote_fetcher.rb#L152.

Then I think it will be really easy to write custom fetcher as plugin (and we can leave S3 fetcher default part of rubygems still).

I'm thinking about something like this.

```ruby
# lib/rubygems_plugin.rb

class Gem::RemoteFetcher
  def fetch_s3(uri, mtime = nil, head = false)
    # ...
  end
end

Gem::RemoteFetcher.register_fetcher('s3') # to add it to known supported uri schemes.
```

`fetch_s3` will be called by https://github.com/rubygems/rubygems/blob/6a1856517fd1455b4626e4c57b2403d1d352bde5/lib/rubygems/remote_fetcher.rb#L263

If you'll miss the plugin, we can extend unsupported scheme message at https://github.com/rubygems/rubygems/blob/6a1856517fd1455b4626e4c57b2403d1d352bde5/lib/rubygems/remote_fetcher.rb#L204 as well.

It will allow a refactoring and separating remote_fetcher to multiple classes like http_fetcher, https_fetcher, ... as well to reduce complexity of remote fetcher itself leaving it responsible just to fetch needed data, not handling various protocols used to fetch data anymore.

Is it worth it separated issue or initial PR?

lib/rubygems/remote_fetcher.rb
apakulov-stripe: It's used to override `Gem::S3URISigner` in the unit-test. Checkout `assert_fetch_s3` in `test/rubygems/test_gem_remote_fetcher.rb`:

```
def fetcher.s3_uri_signer(uri)
  s3_uri_signer = Gem::S3URISigner.new(uri)
  def s3_uri_signer.ec2_metadata
    $instance_profile
  end
  s3_uri_signer
end
```

lib/rubygems/s3_uri_signer.rb
apakulov-stripe: Fixed in `1372e50c`.

apakulov-stripe:
@simi it's a good suggestion and I think it's worth doing it as a separate PR.

apakulov-stripe:
@simi @bronzdoc is there anything else needed to get this PR merged into master?

simi:
I still think this is really complex and out of scope for rubygems actually, but codewise it is ok for me for now. I think `Gem::S3URISigner` can be designed a little better to not have long methods, but I can take a look for possible refactoring on my own.

Is there any [documentation change](https://guides.rubygems.org/using-s3-source/) needed as well?

apakulov-stripe:
@simi I can move some code around in `Gem::S3URISigner.sign` to make it smaller. 
We need a documentation change. Will create a follow-up PR in https://github.com/rubygems/guides

bronzdoc:
> I still think this is really complex and out of scope for rubygems actually
I think this is just improving S3-Source functionality which is something we already support

bronzdoc:
@bundlerbot r+

PR #2808, opened 2019-06-25
Remove useless comment in exceptions.rb

bronzdoc:
# Description:

Remove useless comment in exceptions.rb
______________

### DIFF
--- a/lib/rubygems/exceptions.rb
+++ b/lib/rubygems/exceptions.rb
@@ -1,8 +1,4 @@
 # frozen_string_literal: true
-# TODO: the documentation in here is terrible.
-#
-# Each exception needs a brief description and the scenarios where it is
-# likely to be raised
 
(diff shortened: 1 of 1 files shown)


lib/rubygems/exceptions.rb
```diff
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
-# TODO: the documentation in here is terrible.
-#
+
 # Each exception needs a brief description and the scenarios where it is
```
simi: :thinking: I think those two lines are explaining the original `TODO` problem and should be removed as well if `TODO` line is going to be removed.

bronzdoc: Thanks @simi, you are right  👍

bronzdoc:
@bundlerbot r+

PR #2811, opened 2019-06-25
Remove missleading comment, no reason to move Gem.host to Gem::Util

bronzdoc:
# Description:
The comment suggests moving `Gem.host` and `Gem.host=` but i don't see a good reason to do it, since the logic behind both methods is simple enough
______________


bronzdoc:
@bundlerbot r+

PR #2813, opened 2019-06-27
Drop support for 'gem env packageversion'

bronzdoc:
# Description:
`gem env packageversion` support should have dropped in Rubygems 3.0
______________


bronzdoc:
🤔 i'm not sure why this still failing

simi:
@bronzdoc the problem in here is that `bundler/man/bundle-add.1` file (and all added to manifest in this pull request) are created during `rake man:build` in bundler repo. The failure is complaining about those missing files. I think those files should be removed from Manifest.txt or being versioned in bundler repository.

deivid-rodriguez:
Yeah, you need to remove those files (`git clean -fx` from bundler's folder). and regenerate the manifest.

I think we could improve this in `bundler` by removing those generated files automatically after tests run (they do some sanity checks on `rake man:build`), and after `rake build` runs. I'll open a PR on bundler to do that.

bronzdoc:
Thanks for the review folks :)

@bundlerbot r+

PR #2815, opened 2019-06-27
Improve `rake package` test error message

deivid-rodriguez:
# Description:

When you add or remove new files, this test will fail unless you run `rake update_manifest` and commit the result. The new error message makes that more clear.

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2818, opened 2019-06-30
Remove useless TODO comment

bronzdoc:
# Description:
Without context this comment is useless, so just removing it...

______________

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1530,7 +1530,7 @@ class Gem::Specification < Gem::BasicSpecification
   # a full path.
 
   def bin_dir
-    @bin_dir ||= File.join gem_dir, bindir # TODO: this is unfortunate
+    @bin_dir ||= File.join gem_dir, bindir
   end
 
   ##


lib/rubygems/specification.rb
```diff
@@ -1530,7 +1530,7 @@ def authors
   # a full path.
 
   def bin_dir
-    @bin_dir ||= File.join gem_dir, bindir # TODO: this is unfortunate
```
simi: Do you think that TODO it related to the method NOTE and possible confusion?

bronzdoc: I thik it is!, thanks for pointing that out @simi 

Maybe rename `bin_dir` to `bin_dir_path` ?

simi: Uhh, I think changing public specification method signature is not good idea. Anyway `NOTE` is probably enough in here. There's no need for this TODO.

bronzdoc:
@bundlerbot r+

lib/rubygems/specification.rb
bronzdoc: I was not clear... Yeah, I meant as suggestion for the future

PR #2819, opened 2019-06-30
Remove TODO comment, there's no Gem::Dirs constant

bronzdoc:
# Description:
Remove TODO comment, there's no Gem::Dirs constant
______________

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -675,7 +675,7 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
   # Removes all installed gems from +@gemhome+.
 
   def util_clear_gems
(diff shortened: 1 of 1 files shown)


lib/rubygems/test_case.rb
```diff
@@ -675,7 +675,7 @@ def util_remove_gem(spec)
   # Removes all installed gems from +@gemhome+.
 
   def util_clear_gems
-    FileUtils.rm_rf File.join(@gemhome, "gems") # TODO: use Gem::Dirs
```
simi: Isn't `@gemhome` same as `Gem.paths.home` now? Maybe that was the original idea.

bronzdoc: Maybe 🤔 will check

bronzdoc: They are not @simi

PR #2820, opened 2019-06-30
Remove unused 'raise' from test_case

bronzdoc:
# Description:
Seems like this `raise` is useless now?

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -807,9 +807,6 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
   # location are returned.
 
   def util_gem(name, version, deps = nil, &block)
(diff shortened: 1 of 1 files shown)


lib/rubygems/test_case.rb
```diff
@@ -807,9 +807,6 @@ def util_spec(name, version = 2, deps = nil, *files) # :yields: specification
   # location are returned.
 
   def util_gem(name, version, deps = nil, &block)
-    # TODO: deprecate
```
simi: Any idea why this was marked to deprecate?

bronzdoc: I have no idea tbh, but it seems it's not raised across the entire test suite anymore

bronzdoc:
@bundlerbot

bronzdoc:
@bundlerbot r+

PR #2823, opened 2019-06-30
Remove TODO comment about warning on setting instead of pushing

bronzdoc:
# Description:
Remove TODO comment about warning on setting instead of pushing.

This methods are heavily used across specs and i personally think those warning at this point will not add much value, thoughts?
______________

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1847,29 +1847,23 @@ class Gem::Specification < Gem::BasicSpecification
   end
 
   ##
-  # Sets executables to +value+, ensuring it is an array. Don't
-  # use this, push onto the array instead.
+  # Sets executables to +value+, ensuring it is an array.
 
   def executables=(value)
-    # TODO: warn about setting instead of pushing
     @executables = Array(value)
   end
 
   ##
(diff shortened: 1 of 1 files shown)


simi:
If only specs are related to this, I think we should think about removing/deprecating those methods.

bronzdoc:
@simi if we deprecate/remove those methods people will have to modify their gemspecs to use `<<` instead of `=` I'm not sure that change is something we want?

duckinator:
This looks good to me, although I initially interpreted "specs" as referring to tests instead of gemspecs, which confused me for a moment.

Realistically, I don't think these can be deprecated at this point. It'd break a ton of gemspecs in the wild.

@bronzdoc is there a reason this is marked as wip, or is it good to go?

simi:
@duckinator I did the same mistake, sorry.

duckinator:
@simi it's all good. :slightly_smiling_face: 

@bronzdoc since it's still marked WIP, I'm leaving this alone. If you're done with it, go ahead and merge it. :+1:

bronzdoc:
`specs` is used across the Rubygems codebase to denote `specifications`. Sorry for the confusion!

bronzdoc:
@bundlerbot r+

PR #2825, opened 2019-07-11
Simplify #to_ruby code

nobu:
# Description:

Separate #add_runtime_dependency and #add_development_dependency
availability condition from #specification_version availability,
which is not related to directly.
Also check if the former method is available, instead of comparing
the version.

# Tasks:

- [x] Write tests


hsbt:
@bundlerbot r+

Issue #2750, opened 2019-04-27
Is Gem::Installer#gem actually set?


djberg96:
While fiddling around with gem plugins and tinkering with the `Gem::Installer` class in general, I noticed that there's an attr_reader for `:gem`. However, near as I can tell, this value is never set. I don't see `@gem` set anywhere.

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/installer.rb#L49

With Gem.pre_install, this value always comes back nil. You can try with the following code:

```
# Put this in 'lib/rubygems_plugin.rb' somewhere
# Run RUBYOPT=-Ilib gem install any_gem
#
Gem.pre_install do |installer|
  p installer.gem # nil
end
```

djberg96:
@bronzdoc @segiddins Any idea if this is supposed to be set, and what it's supposed to be set to?

segiddins:
Zero clue, sorry

bronzdoc:
@djberg96 
well... it's suposed to be set to `Filename of the gem being installed.` but i don't see where is set or if it set at all as you mentioned, i ran the test suite with out it to see if there was something testing for it, but there is no test failures.

djberg96:
@bronzdoc I wasn't entirely sure what the "filename of the gem" is supposed to be. Is it referring to the `.tar.gz` filename? Or just the gem name? Or perhaps the output of `gem which`?

Maybe I'm overthinking it.

bronzdoc:
@djberg96 that's a good question, but i don't know the answer.

Could have been any of the tree but since it never was setup we will never know, but of the 3 options what would be the best to be in your opinion?

djberg96:
My gut reaction is that it was meant to be the name of the `.tar.gz` file, but I'm not married to the idea or anything.

@drbrain Don't suppose you remember, do you?

bronzdoc:
Can this be closed @djberg96 ?

djberg96:
@bronzdoc was it fixed?

djberg96:
@bronzdoc I spent some more time looking at this. I think we ultimately want the `package.gem.path`, but that means two accessors: the `package` reader for `Gem::Package` and, once that's in place, a method like so:

```
def gem
  @package.gem.path
rescue Gem::Package::Error => e
  raise Gem::InstallError, "invalid gem: #{e.message}"
end
```

Part 1 is at https://github.com/rubygems/rubygems/pull/2796

djberg96:
[...]

PR #2828, opened 2019-07-11
Add a gem attr to the Gem::Package class

djberg96:
This PR just adds a simple attr reader to the `Gem::Package` class that lets you access the gem object passed to the constructor. Potentially useful in plugins so that you don't have to use `instance_variable_get`.

This is part 2 of https://github.com/rubygems/rubygems/issues/2750.

### DIFF
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -1093,10 +1093,10 @@ class TestGemPackage < Gem::Package::TarTestCase
     assert_equal @spec, package.spec
   end
 
-  def test_gem_attr
-    package = Gem::Package.new(@gem)
-    assert_equal(package.gem, @gem)
-  end
+ def test_gem_attr
+   package = Gem::Package.new(@gem)
+   assert_equal(package.gem.path, @gem)
+ end
 
   def test_spec_from_io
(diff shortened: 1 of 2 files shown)


test/rubygems/test_gem_package.rb
```diff
@@ -1093,6 +1093,11 @@ def test_spec
     assert_equal @spec, package.spec
   end
 
+  def test_gem_attr
+    package = Gem::Package.new(@gem)
+    assert_equal(package.gem, @gem)
```
duckinator: Did you mean `package.gem.path`, here?

Asking due to the nature of the error on Travis CI: https://travis-ci.org/rubygems/rubygems/jobs/557334428#L278

djberg96: I'm confused, why is it expecting a path? It's just a attribute referencing the argument passed to `initialize`.

bronzdoc: @djberg96 Its because we are calling `self.new` first and then`initialize` https://github.com/rubygems/rubygems/blob/02f8c4b416a62361e721c973a5a23742e0dabbbd/lib/rubygems/package.rb#L141 first and then`initialize`

bronzdoc:
@bundlerbot r+

PR #2834, opened 2019-07-17
Use File#open instead of Kernel#open in stub_specification.rb

bronzdoc:
# Description:
Use `File#open` instead of `Kernel#open`


bronzdoc:
@bundlerbot r+

PR #2835, opened 2019-07-17
Adds web command

### LINKED ISSUE #2784: Integrate functionality from gem-web into RubyGems gem CLI
As a selected project on Google Summer of Code 2019, I have to integrate functionality from [gem-web](https://github.com/bitboxer/gem-web) into RubyGems gem CLI.

I thought I should create an issue so it can be addressed when I made the PR. Besides that others contributors can make suggestions or comments here too.

_(This is a bit different from common issues, so I filled this template the best as I could)_
My current problem is [to integrate functionality from gem-web into RubyGems gem CLI](https://summerofcode.withgoogle.com/projects/#5260793296191488).

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
$ gem env version
2.7.8
```

MatheusRich:
# Description:

Closes #2784.

Ruby claims to be "a programmer's best friend", and focus on "simplicity and productivity". As rubygems is now part of Ruby, I think it should follow those principles too. 

I'm integrating [gem-web](https://github.com/bitboxer/gem-web) into rubygems as part of [GSoC 2019](https://summerofcode.withgoogle.com/projects/#5260793296191488). This new command allows user to quickly find documentation of a gem.

I think this is specially good when one's not sure about what a gem does, or is confused with its name. Maybe the user wants to quickly check the rubygems' page of some gem (to check versions, dependencies, etc). With this command the access will be faster and easier with just one command on the terminal. Other use case could be opening the documentation for a gem, which improves productivity.

The usage would be like this:
```
gem web rails # Opens github
```

    Options:
      -c, --sourcecode                 Open sourcecode gem
      -d, --doc                        Open documentation of gem
      -w, --webpage                    Open webpage of gem. This is the default.
      -r, --rubygems                   Open the rubygems page for that gem

Let me hear some feedback about this.

# Tasks:

- [x] Write tests


simi:
I think this is too complex and also too much bounded to 3rd party services.

IMHO it is enough to print out back URL so you can use it to open in browser for example in my system I will be able to do `chromium-browser $(gem web rails)`. 

Also we should respect gemspec homepage only, since it is the only URL we should believe. Looking up on public services by search or parsing metadata is dangerous. What will happen when you have gem from private gem server sharing same name as one from public server? If I understand it well, you'll end up on wrong page.

To make it really simple (at least as a starting point) I think `gem web rails` should just print out `http://rubyonrails.org/` to STDOUT since it is set as a homepage in gemspec (https://guides.rubygems.org/specification-reference/#homepage).

example:
```
$ gem web rails
https://rubyonrails.org/
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+require 'open-uri'
+require 'json'
+require 'rubygems/platform'
+
+module Gem
+  module Web
+    class Executor
+
+      def open_default_browser_cmd(local_os, version, uri)
```
bronzdoc: This would be really hard to maintain, would be better if it reads the `browser` from an environment variable

MatheusRich: @bronzdoc That's a nice option. `gem-web` used the gem `Launchy` to open links, so it was kinda hard to think a good way to reproduce this.

MatheusRich:
The tests are failing in Veyor because some tests aren't mocking the OS. I'll wait until we reach a consensus about the `web` command output before correcting them.

indirect:
I think maybe printing the URL by default, but if `BROWSER` is set opening the URL using that command? I'm definitely happy to set `BROWSER=open` to get automatic opening.

MatheusRich:
@indirect @bronzdoc I've updated the code following your suggestion

bronzdoc:
Hey @MatheusRich! thanks! :)

I think this should be simpler -  try to open the url defined in `gemspec homepage` if the `BROWSER` env var is defined. If not just print the url defined in `gemspec homepage` as @simi mentioned.

Bounding this functionality to third party services will make Rubygems harder to maintain, for the reasons @simi mentioned. So it will be better to keep it simpler at the beginning and gradually add functionality to this command if it is worth it.

MatheusRich:
@bronzdoc I've been studying how to implement this without the API request. I will use `Gem::Specification.find_by_name(gem)` to get the spec, and then use its data. Btw, I saw this approach too:  `Gem::Package.new(gem).spec`, but I think I'll go with the first one.

Of course this limits the `web` command to installed gems.

indirect:
Is it possible to do `find_by_name` with a fallback on the API request? For example, I don't like to install Rails system-wide with `gem install rails`. Instead, I only install Rails for specific applications using `bundle install`. If this changes to require the gem to be installed, I won't be able to `gem web rails` anymore 😬

MatheusRich:
@indirect I think it's possible too. Once I finish updating the code, I'll propose a solution with the `-r --remote` option

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,78 @@
...
+        open_default_browser("https://rubygems.org/gems/#{gem}")
+      end
+
+      def open_rubytoolbox(gem)
+        open_default_browser("https://www.ruby-toolbox.com/projects/#{gem}")
+      end
+
+      def open_default_browser(uri)
+        open_browser_cmd = ENV['BROWSER']
+        if open_browser_cmd.nil?
```
duckinator: This should probably be something like `if open_browser_cmd.nil? || open_browser_cmd.empty?`

That would let you set `$BROWSER` to an empty string to force it to be printed to stdout:

```
$ echo $BROWSER
firefox
$ BROWSER= gem web rails
https://github.com/rails/rails
```

MatheusRich: Ok! I'll add it

MatheusRich:
@bronzdoc @duckinator @indirect What do you guys think of this solution? Once you guys agree, I'll add the `--remote` option that @indirect asked.

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+      rescue Gem::MissingSpecError => e
+        puts e.message
+      end
+
+      def get_info_from_metadata(spec, info)
+        return unless supported_options_from_metadata.include?(info)
+
+        uri = spec.metadata[info]
+        if !uri.nil? && !uri.empty?
```
duckinator: 1. It may be worth swapping this if/else statement around, to be consistent with `open_default_browser` below.
2. No need for parenthesis in the `puts(...)` call.

```ruby
if uri.nil? || uri.empty?
  puts "This gem does not have this information."
else
  open_default_browser(uri)
end
```

duckinator:
@MatheusRich I left a comment about some minor refactoring. Aside from that, I think it's be fine to go ahead and add `--remote`. :+1:

lib/rubygems/executor.rb
MatheusRich: Yeah, I thought about make it like this, but I was unsure. I'll update it!

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+        else
+          puts("This gem does not have this information.")
+        end
+      end
+
+      def open_rubygems(gem)
+        open_default_browser("https://rubygems.org/gems/#{gem}")
+      end
+
+      def open_default_browser(uri)
```
bronzdoc: rename this to `open_browser` ?

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      def get_info_from_metadata(spec, info)
+        return unless supported_options_from_metadata.include?(info)
+
+        uri = spec.metadata[info]
+        if !uri.nil? && !uri.empty?
+          open_default_browser(uri)
+        else
+          puts("This gem does not have this information.")
```
bronzdoc:
```ruby
puts "#{spec.name} does not specify #{info}. "
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+module Gem
+  module Web
+    class Executor
+
+      def supported_options_from_metadata
+        ["source_code_uri", "documentation_uri"]
+      end
+
+      def open_page(gem, options)
+        spec = Gem::Specification.find_by_name(gem)
```
bronzdoc:
```ruby
begin       
  spec = Gem::Specification.find_by_name(gem)
rescue Gem::MissingSpecError => e
   puts e.message
end
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+        elsif options[:doc]
+          get_info_from_metadata(spec, "documentation_uri")
+        elsif options[:webpage]
+          open_default_browser(spec.homepage)
+        elsif options[:rubygems]
+          open_rubygems(gem)
+        else # The default option is homepage
+          open_default_browser(spec.homepage)
+        end
+      rescue Gem::MissingSpecError => e
```
bronzdoc: This shouldn't be down here

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+        puts e.message
+      end
+
+      def get_info_from_metadata(spec, info)
+        return unless supported_options_from_metadata.include?(info)
```
bronzdoc: You need this?

what about:
```ruby
uri = spec.metadata[info]
return unless uri
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+          puts("This gem does not have this information.")
+        end
+      end
+
+      def open_rubygems(gem)
+        open_default_browser("https://rubygems.org/gems/#{gem}")
+      end
+
+      def open_default_browser(uri)
+        open_browser_cmd = ENV['BROWSER']
```
bronzdoc:
```ruby
browser = ENV["BROWSER"]
```

bronzdoc:
@MatheusRich 
I don't think adding a `--remote` flag is ideal to a `web` command.

What @indirect is trying to say(I think) is that we should use `find_by_name` to look for the gem locally, but in the case that the gem is missing make a request to the rubygems api looking for the info. This would happen by default without specifying any flag.

indirect:
Yup, that’s what I was trying to say. :)

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,58 @@
...
+      end
+
+      def open_page(gem, options)
+        spec = Gem::Specification.find_by_name(gem)
+
+        if options[:sourcecode]
+          get_info_from_metadata(spec, "source_code_uri")
+        elsif options[:doc]
+          get_info_from_metadata(spec, "documentation_uri")
+        elsif options[:webpage]
```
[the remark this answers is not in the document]
MatheusRich: That's a good point. I only left it 'cuz you guys could want to change the default option. Since we're agreeing with homepage being the default, I'll remove this `elsif`

lib/rubygems/executor.rb
MatheusRich: How should I prevent running the remaining code? Adding a `exit 1` after the error message?

lib/rubygems/executor.rb
MatheusRich: I can't return, because this would prevent printing out the error message `Gem '#{spec.name}' does not specify #{info}.`. So I only removed the check

MatheusRich:
@bronzdoc @indirect What do you guys think of the implementation?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+require 'rubygems/command'
+require 'rubygems/executor'
+require 'rubygems/version_option'
+
+class Gem::Commands::WebCommand < Gem::Command
+
+  include Gem::VersionOption
+  attr_reader :executor
```
simi: Is this line used?

MatheusRich: Nope, removing it!

MatheusRich: Whoops. Actually I do use this. I use it in `execute` method.

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,45 @@
...
+      :version => Gem::Requirement.default,
+      :latest => false
+
+    add_option("-c", "--sourcecode", "Open sourcecode gem") do |v|
+      options[:sourcecode] = v
+    end
+    add_option("-d", "--doc", "Open documentation of gem") do |v|
+      options[:doc] = v
+    end
+    add_option("-w", "--webpage", "Open webpage of gem") do |v|
```
bronzdoc: Do we need this flag?

lib/rubygems/commands/web_command.rb
bronzdoc: @MatheusRich you don't need that `attr_reader` to use `@executor` in the `execute` method

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,71 @@
...
+
+      def open_page(gem, options)
+        begin
+          spec = Gem::Specification.find_by_name(gem)
+        rescue Gem::MissingSpecError => e
+          spec = fetch_remote_spec(gem)
+
+          if spec.nil?
+            puts e.message
+            puts "Could not find '#{gem}' in rubygems.org too."
```
bronzdoc: Maybe don't print the `e.message` ? 

change to `Could not find '#{gem}' in rubygems.org`

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,71 @@
...
+        elsif options[:rubygems]
+          open_rubygems(gem)
+        else # The default option is homepage
+          open_browser(spec.homepage)
+        end
+      end
+
+      def fetch_remote_spec(gem)
+        dep = Gem::Dependency.new gem
+        found, _ = Gem::SpecFetcher.fetcher.spec_for_dependency dep
```
bronzdoc: use parenthesis `Gem::SpecFetcher.fetcher.spec_for_dependency(dep)`

lib/rubygems/commands/web_command.rb
MatheusRich: Removing it!

lib/rubygems/executor.rb
MatheusRich: Ok!

lib/rubygems/commands/web_command.rb
MatheusRich: @bronzdoc Right. Now how can I stub `fetch_remote_spec`, because I was doing

```ruby
@cmd.executor.stub :fetch_remote_spec, nil
```

Since there's no access to this instance anymore, is there any other way to stub this? I think minitest doesn't have a `stub_any_instance` method.

lib/rubygems/executor.rb
MatheusRich: Done!

MatheusRich:
@bronzdoc Any other suggestion?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'rubygems/command'
+require 'rubygems/executor'
+require 'rubygems/version_option'
+
+class Gem::Commands::WebCommand < Gem::Command
+
+  include Gem::VersionOption
+
+  def initialize
+    super 'web', "Open the gem's homepage",
```
bronzdoc: Be consistent with the quotes, use double quotes `"`

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+require 'rubygems/command'
+require 'rubygems/executor'
+require 'rubygems/version_option'
+
+class Gem::Commands::WebCommand < Gem::Command
+
+  include Gem::VersionOption
```
bronzdoc: Why we need this?

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+require 'open-uri'
+require 'json'
+require 'rubygems/platform'
```
bronzdoc: Why we need this also?

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
...
+          get_info_from_metadata(spec, "documentation_uri")
+        elsif options[:rubygems]
+          open_rubygems(gem)
+        else # The default option is homepage
+          open_browser(spec.homepage)
+        end
+      end
+
+      def fetch_remote_spec(gem)
+        dep = Gem::Dependency.new gem
```
bronzdoc: Use  parentheses when calling a method with params

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
...
+        found, _ = Gem::SpecFetcher.fetcher.spec_for_dependency(dep)
+        spec_tuple = found.first
+
+        if spec_tuple.nil? || spec_tuple.empty?
+          nil
+        else
+          spec_tuple.first
+        end
```
bronzdoc: You can rewrite it like:
```ruby
spec_tuple.first unless spec_tuple.nil? || spec_tuple.empty?
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
...
+          open_browser(spec.homepage)
+        end
+      end
+
+      def fetch_remote_spec(gem)
+        dep = Gem::Dependency.new gem
+        found, _ = Gem::SpecFetcher.fetcher.spec_for_dependency(dep)
+        spec_tuple = found.first
+
+        if spec_tuple.nil? || spec_tuple.empty?
```
bronzdoc: Maybe you want to `strip` `spec_tuple`before checking for `empty?`
```ruby
"".empty? #=> true
" ".empty? #=> false
" ".strip.empty? #=> true
```

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
...
+          end
+        end
+
+        if options[:sourcecode]
+          get_info_from_metadata(spec, "source_code_uri")
+        elsif options[:doc]
+          get_info_from_metadata(spec, "documentation_uri")
+        elsif options[:rubygems]
+          open_rubygems(gem)
+        else # The default option is homepage
```
bronzdoc: Do we need this comment?

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,70 @@
...
+          nil
+        else
+          spec_tuple.first
+        end
+      end
+
+      def get_info_from_metadata(spec, info)
+        uri = spec.metadata[info]
+
+        if uri.nil? || uri.empty?
```
bronzdoc: `strip` uri?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def initialize
+    super 'web', "Open the gem's homepage",
+      :command => nil,
+      :version => Gem::Requirement.default,
+      :latest => false
+
+    add_option("-c", "--sourcecode", "Open sourcecode gem") do |v|
```
duckinator: Possibly change this to say "Open source code for the gem"?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,41 @@
...
+      options[:sourcecode] = v
+    end
+    add_option("-d", "--doc", "Open documentation of gem") do |v|
```
duckinator: Possibly change this to "Open documentation for the gem"?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,41 @@
...
+      options[:doc] = v
+    end
+    add_option("-r", "--rubygems", "Open the rubygems page of a gem") do |v|
```
duckinator: Possibly change this to "Open the rubygems.org page for the gem"?

duckinator:
@MatheusRich I see that @bronzdoc left some suggestions/questions, and I added a few as well. You're making fantastic progress on this! 👍

lib/rubygems/executor.rb
MatheusRich: I'll have to add `uri = spec.metadata[info]&.strip`, because this could return `nil`.

lib/rubygems/executor.rb
MatheusRich: It would be `spec_tuple.first.strip unless spec_tuple.nil? || spec_tuple.empty?`.

Maybe `spec_tuple.first&.strip unless spec_tuple.nil? || spec_tuple.empty?`, but I don't know if there's a way to spec_tuple.first to be `nil`

lib/rubygems/executor.rb
MatheusRich: Seems right.

MatheusRich:
@bronzdoc What do you think?

lib/rubygems/executor.rb
```diff
@@ -0,0 +1,65 @@
...
+          puts "Gem '#{spec.name}' does not specify #{info}."
+        else
+          open_browser(uri)
+        end
+      end
+
+      def open_rubygems(gem)
+        open_browser("https://rubygems.org/gems/#{gem}")
```
simi: :thinking: 

This is reliable address only for gems coming from `rubygems.org` source. I think to make this command bulletproof it will be safer to remove this auto-generated link and rely only on gemspec info provided by gem author.

Alternatively we should track the origin of gem and use this address only for gems coming from `rubygems.org`.

duckinator: Looking through everything again, I think this is the last remaining problem preventing this from being merged.

It may make sense to rename the `-r`/--rubygems` flag once this is changed, but I'm not sure what it would be changed to.

I'll poke at this a bit and see if I can find anything about getting the origin for the gem.

MatheusRich: @duckinator Ok! If needed, I can remove the option at all.

duckinator: @MatheusRich Unfortunately, it looks the `.gemspec` files for installed gems don't include any information about their origin. I think it may be best to just remove the `-r`/`--rubygems` flags for the time being.

Someone can always add an equivalent later if a way to get the upstream information is found/added.

simi: I think we do not store gem origin. It would be great new feature IMHO, but for now since the option is called `rubygems` I think it makes sense to keep it as is. Anyway I think we can extend option help message to mention this problem.

duckinator: @simi trying to make sure I understood correctly -- are you saying to just keep it `-r`/`--rubygems` but add a note to `gem web --help` saying that it won't work right for gems not hosted on rubygems.org?

simi: Yup, my idea is to keep `-r/--rubygems` and explain at `gem web --help` that it just trying to blidly find that gem on `rubygems.org` by name.

duckinator: @MatheusRich I think following @simi's approach (just adding a note to `gem web --help` that `-r`/`--rubygems` always assumes things are hosted on rubygems.org) should be fine. If anyone comes up with a better approach, someone can always make a PR later on.

MatheusRich: Ok! I'll update it

MatheusRich: @duckinator What do you think?

simi: I think we should mention if gem is not hosted on rubygems it can fail or find another gem with same name hosted on rubygems. Anyway I have no idea how to write that in nice short way (I'm not native english speaker). Anyone any idea?

lib/rubygems/commands/web_command.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def description # :nodoc:
+    <<~HEREDOC
+      The web command allows you to get/open webpages for a given gem.
+
+      It's possible to configure it to automatically open links:
+
+        $ BROWSER=firefox gem web rails
+
+      The --rubygems option only fetches gems from rubygems.org
```
duckinator: Perhaps this?

```suggestion
      If the gem is not hosted on rubygems.org, using --rubygems will result
      in a broken or incorrect link.
```

lib/rubygems/executor.rb
duckinator: I made a suggestion here: https://github.com/rubygems/rubygems/pull/2835/files#r344308330 :+1:

lib/rubygems/commands/web_command.rb
MatheusRich: Done!

duckinator:
Hey, sorry for the slow response on my part. I think this is good to go — @bronzdoc could you give this a quick review again, when you have a moment?

duckinator:
I just realized that everything related to CI has disappeared from this PR, for some reason. It seems to have happened with a few of them. I'm looking into it now.

duckinator:
Actually... since it's only two PRs that are affected, and this one may be ready, let me just see if this works:

@bundlerbot try

duckinator:
Okay, looks like that trick will work for the affected PRs, even if it's awkward/annoying.

@MatheusRich it looks like there's still some failing tests: https://github.com/rubygems/rubygems/runs/314777260

bronzdoc:
Hey @MatheusRich!

I was testing this and I'm not getting the expected behavior
```shell
$ gem web -c rails
Gem 'rails' does not specify source_code_uri.
```

```shell
$ gem web -d rails
Gem 'rails' does not specify documentation_uri.
```

But rails do have that metadata

`rails.gemspec:`

```ruby
s.metadata = {
    "bug_tracker_uri"   => "https://github.com/rails/rails/issues",
    "documentation_uri" => "https://api.rubyonrails.org/v#{version}/",
    "mailing_list_uri"  => "https://groups.google.com/forum/#!forum/rubyonrails-talk",
    "source_code_uri"   => "https://github.com/rails/rails/tree/v#{version}",
  }
```

I think the local spec we read doesn't include the `spec.metadata` I'll investigate and get back or if someone else can confirm the behavior I'm getting that would be helpful too.

duckinator:
@bronzdoc how were you running it? I was able to do this just now:

```
~/dev/ruby/matheusrich$ git clone -b adds-web-command https://github.com/matheusrich/rubygems.git
Cloning into 'rubygems'...
remote: Enumerating objects: 58740, done.
remote: Total 58740 (delta 0), reused 0 (delta 0), pack-reused 58740
Receiving objects: 100% (58740/58740), 15.86 MiB | 41.96 MiB/s, done.
Resolving deltas: 100% (37535/37535), done.
~/dev/ruby/matheusrich$ cd rubygems/
~/dev/ruby/matheusrich/rubygems$ export BROWSER=
~/dev/ruby/matheusrich/rubygems$ ruby -Ilib bin/gem web -c rails
https://github.com/rails/rails/tree/v6.0.1
~/dev/ruby/matheusrich/rubygems$ ruby -Ilib bin/gem web -d rails
https://api.rubyonrails.org/v6.0.1/
~/dev/ruby/matheusrich/rubygems$
```

MatheusRich:
@duckinator I saw that the repo changed quite a bit since my last updates. I'm going to merge master on this and see if tests fail and update what is necessary

MatheusRich:
@duckinator I updated it, but it seems that those tests are failing in other PRs too. Ruby-head is allowed to fail?

bronzdoc:
@duckinator It's weird

My local version of Rails is 6.0.0 and the spec of that version is missing the metadata, so the web command fails with that version. It works when I install Rails 6.0.1 since rails spec for that version do contains metadata. 🤔 

`Rails 6.0.0`
```
# -*- encoding: utf-8 -*-
# stub: rails 6.0.0 ruby lib

Gem::Specification.new do |s|
  s.name = "rails".freeze
  s.version = "6.0.0"

  s.required_rubygems_version = Gem::Requirement.new(">= 1.8.11".freeze) if s.respond_to? :required_rubygems_version=
  s.require_paths = ["lib".freeze]
  s.authors = ["David Heinemeier Hansson".freeze]
  s.date = "2019-08-16"
  s.description = "Ruby on Rails is a full-stack web framework optimized for programmer happiness and sustainable productivity. It encourages beautiful code by favoring convention over configuration.".freeze
  s.email = "david@loudthinking.com".freeze
  s.homepage = "https://rubyonrails.org".freeze
  s.licenses = ["MIT".freeze]
  s.required_ruby_version = Gem::Requirement.new(">= 2.5.0".freeze)
  s.rubygems_version = "3.1.0.pre1".freeze
  s.summary = "Full-stack web application framework.".freeze

  s.installed_by_version = "3.1.0.pre1" if s.respond_to? :installed_by_version

  if s.respond_to? :specification_version then
    s.specification_version = 4

    if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
      s.add_runtime_dependency(%q<activesupport>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actionpack>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actionview>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<activemodel>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<activerecord>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actionmailer>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<activejob>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actioncable>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<activestorage>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actionmailbox>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<actiontext>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<railties>.freeze, ["= 6.0.0"])
      s.add_runtime_dependency(%q<bundler>.freeze, [">= 1.3.0"])
      s.add_runtime_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
    else
      s.add_dependency(%q<activesupport>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actionpack>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actionview>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<activemodel>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<activerecord>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actionmailer>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<activejob>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actioncable>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<activestorage>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actionmailbox>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<actiontext>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<railties>.freeze, ["= 6.0.0"])
      s.add_dependency(%q<bundler>.freeze, [">= 1.3.0"])
      s.add_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
    end
  else
    s.add_dependency(%q<activesupport>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actionpack>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actionview>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<activemodel>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<activerecord>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actionmailer>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<activejob>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actioncable>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<activestorage>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actionmailbox>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<actiontext>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<railties>.freeze, ["= 6.0.0"])
    s.add_dependency(%q<bundler>.freeze, [">= 1.3.0"])
    s.add_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
  end
end
```

`Rails 6.0.1`
```
# -*- encoding: utf-8 -*-
# stub: rails 6.0.1 ruby lib

Gem::Specification.new do |s|
  s.name = "rails".freeze
  s.version = "6.0.1"

  s.required_rubygems_version = Gem::Requirement.new(">= 1.8.11".freeze) if s.respond_to? :required_rubygems_version=
  s.metadata = { "bug_tracker_uri" => "https://github.com/rails/rails/issues", "documentation_uri" => "https://api.rubyonrails.org/v6.0.1/", "mailing_list_uri" => "https://groups.google.com/forum/#!forum/rubyonrails-talk", "source_code_uri" => "https://github.com/rails/rails/tree/v6.0.1" } if s.respond_to? :metadata=
  s.require_paths = ["lib".freeze]
  s.authors = ["David Heinemeier Hansson".freeze]
  s.date = "2019-11-05"
  s.description = "Ruby on Rails is a full-stack web framework optimized for programmer happiness and sustainable productivity. It encourages beautiful code by favoring convention over configuration.".freeze
  s.email = "david@loudthinking.com".freeze
  s.homepage = "https://rubyonrails.org".freeze
  s.licenses = ["MIT".freeze]
  s.required_ruby_version = Gem::Requirement.new(">= 2.5.0".freeze)
  s.rubygems_version = "3.0.6".freeze
  s.summary = "Full-stack web application framework.".freeze

  s.installed_by_version = "3.0.6" if s.respond_to? :installed_by_version

  if s.respond_to? :specification_version then
    s.specification_version = 4

    if Gem::Version.new(Gem::VERSION) >= Gem::Version.new('1.2.0') then
      s.add_runtime_dependency(%q<activesupport>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actionpack>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actionview>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<activemodel>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<activerecord>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actionmailer>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<activejob>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actioncable>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<activestorage>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actionmailbox>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<actiontext>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<railties>.freeze, ["= 6.0.1"])
      s.add_runtime_dependency(%q<bundler>.freeze, [">= 1.3.0"])
      s.add_runtime_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
    else
      s.add_dependency(%q<activesupport>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actionpack>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actionview>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<activemodel>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<activerecord>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actionmailer>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<activejob>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actioncable>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<activestorage>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actionmailbox>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<actiontext>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<railties>.freeze, ["= 6.0.1"])
      s.add_dependency(%q<bundler>.freeze, [">= 1.3.0"])
      s.add_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
    end
  else
    s.add_dependency(%q<activesupport>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actionpack>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actionview>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<activemodel>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<activerecord>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actionmailer>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<activejob>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actioncable>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<activestorage>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actionmailbox>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<actiontext>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<railties>.freeze, ["= 6.0.1"])
    s.add_dependency(%q<bundler>.freeze, [">= 1.3.0"])
    s.add_dependency(%q<sprockets-rails>.freeze, [">= 2.0.0"])
  end
end

```

duckinator:
@MatheusRich it looks like [in the old Travis CI config ruby-head _was_ allowed to fail](https://github.com/rubygems/rubygems/commit/d2d2ee6dbf6ae20b2b8321e15433b884a0c42cda#diff-354f30a63fb0907d4ad57269548329e3L44-L48). [As best I can tell](https://github.com/rubygems/rubygems/blob/master/.github/workflows/ubuntu-rvm.yml) it isn't set up that way. I'm not sure if this was deliberate.

---

@bronzdoc it looks like they actually added that only on October 11 of this year, and this change was included as part of 6.0.1: https://github.com/rails/rails/commit/493edf044ff4c025d2874e957693e78b96a260d7#diff-73ac7236f7a8247a9cd61d338ef49cda

Do you think it'd make sense to have it fall back to checking remotely both if the gem isn't installed _or_ if the local gemspec doesn't contain the metadata?

duckinator:
@bundlerbot try

deivid-rodriguez:
I'm confused about `bundlerbot` continuously retrying this? :grimacing:

Failure is known and I'm looking into it (TL;DR rvm ruby-head image is outdated and that creates some failures when testing against bundler master).

bronzdoc:
@deivid-rodriguez maybe if bundlerbot tries enough times it will pass 😛

deivid-rodriguez:
This is terribly annoying, what's going on?

deivid-rodriguez:
I manually canceled every running workflow to try stop this.

deivid-rodriguez:
That didn't work. Now I tried manually rebasing and force-pushing.

deivid-rodriguez:
Yeah, good call @bronzdoc :sweat_smile:

duckinator:
@deivid-rodriguez I have no idea why bundlerbot was freaking out, but thank you for handling that + confirming that the failing tests is a known problem. :+1:

duckinator:
@bronzdoc which part of the gem-web functionality is missing that you want?

Here's all of the features I see from reading gem-web's source, and a check means this PR has it:

- [x] `--sourcecode`
- [x] `--doc`
- [x] `--rubygems`
- [ ] `--webpage` &mdash; kind of? This is the default behavior, but there's no option. Should it be added for compatibility reasons?
- [ ] `--rubytoolbox` &mdash; missing; I'd somehow missed this aspect of it before.
- [x] opens the page in browser &mdash; I honestly feel this PR handles it better than gem-web, by falling back to just printing it.

Are you looking for `--rubytoolbox`, or have I missed something? I think it would be worth adding `--rubytoolbox`, but I'm okay with that being deferred to a follow-up PR.

---

I'm also in favor of `gem web` being a separate command as it is here, primarily because `gem info` is exclusively non-interactive but `gem web` is exclusively interactive (albeit indirectly, in that it opens a web browser).

duckinator:
@bronzdoc also: I do think `gem info` should be updated to include all project URLs it knows of, but I still think it's okay for the two commands to be separate since they wouldn't be duplicating any logic (only data sources).

bronzdoc:
@duckinator 

The `gem-web` plugin opens the pages by default, this implementation prints the url and just open a web page if you have set `BROWSER`

So we would add this command and use it like `BROWSER=open gem web rails`  instead of installing the plugin and just do `gem web rails` ?

If we are going to do `BROWSER=open gem web rails --sourcecode` why we just don't do `open $(gem info rails --sourcecode)` for example.

MatheusRich:
@bronzdoc you could export BROWSER and the command will automatically open the browser when running `gem web rails`

duckinator:
@bronzdoc all you need to do is add `export BROWSER=open` in your `.bash_profile` (or equivalent file for your shell of choice), and then you can just do `gem web rails`.

`$BROWSER` is a fairly common environment variable to already have set, at least on *nix systems, which is why it was used. It's actually one of the two methods used by launchy (the gem the original gem-web uses to open a browser). The other way is hard-coding things for multiple platforms based on what executables exist, which of a bunch of destop-environment-specific environment variables are set, etc.

This PR made the call to simplify the design by _only_ using `$BROWSER` and falling back to printing it. I feel like this is a reasonable approach, especially given that we try to avoid pulling in extra dependencies. I feel like, at least for the initial fully-integrated implementation, anything farther is just complicating the matter. We can always make it more clever later.

duckinator:
To clarify re: "anything farther is just complicating the matter", this PR already adds 283 lines of code. The code to make it detect the browser reliably is actually pretty complicated. I'm 100% okay with that being added if folks want it, but I think the complexity outweighs the urgency & it should be left for a follow-up PR.

duckinator:
@bronzdoc does having this PR stick with `$BROWSER` (to get the core functionality in place) and adding automatic detection of things in a follow-up PR work for you? I'd be okay implementing that part of it myself.

bronzdoc:
I know I can do `export BROWSER=open`. My point is that the main functionality of the command will depend on an external factor like setting the `BROWSER` env var.

Also, what is the benefit of adding this to the Rubygems codebase?

If we add this we will have two commands(`info`, `web`) that try to do the same thing(provide information of a gem to the user), In my opinion that is confusing.

I think functionality like this fits perfectly into a plugin and I don't see any issues with this being a plugin that could make us want to add it to the Rubygems codebase.

If our concern is that people could find documentation of gems easily, we should improve the `info` command which is something users are already familiarized with and add the urls printing functionality to it. If people feel like they want to open browsers instead of just seeing the url they can install `gem-web`.

duckinator:
@bronzdoc since `gem info` builds on `gem query`, and `gem query` is deprecated, what are your thoughts on me rewriting/refactoring `gem info` to be independent of `gem query` and then reusing @MatheusRich' work here to add `gem web`-like functionality to it?

I would also be okay with putting in the work to add a separate class that determines the command to open the browser, so it works without needing `$BROWSER` set (like the original).

deivid-rodriguez:
> @bronzdoc since gem info builds on gem query, and gem query is deprecated, what are your thoughts on me rewriting/refactoring gem info to be independent of gem query

@duckinator Just so you know, I did just that in https://github.com/rubygems/rubygems/pull/3014.

duckinator:
@deivid-rodriguez nice, thanks for doing that! :slightly_smiling_face: 

@bronzdoc are you still okay with this functionality being added to `gem info` as you suggested [in your latest review of this PR](https://github.com/rubygems/rubygems/pull/2835#pullrequestreview-323391784)?

bronzdoc:
@duckinator

>  @bronzdoc are you still okay with this functionality being added to gem info as you suggested in your latest review of this PR?

"print info of the metadata urls", yes :)

MatheusRich:
@bronzdoc Do you think it's better to open another PR and close this one, or just revert everything and start again here?

duckinator:
@MatheusRich it may be best to start a new PR, but you can probably reuse a lot of the code from this one.

A rough idea on how I'd personally approach it, but you don't need to strictly follow this:

- add the `add_option()` calls and `@executor = ...` line from `lib/rubygems/commands/web_command.rb` to `lib/rubygems/commands/info_command.rb`
- define a `execute` method in `lib/rubygems/commands/info_command.rb` which calls `@executor.open_page(...)` when appropriate, and otherwise calls the version of `execute()` pulled in by the `include Gem::QueryUtils` line (I'm not sure how to do this off the top of my head &mdash; if you need help let me know)
- move `Gem::Web::Executor` into `lib/rubygems/commands/info_command.rb`, renamed to something like `Gem::Commands::InfoCommand::Executor` (and adjust `@executor = ...` line accordingly)
- the tests in `test/rubygems/test_gem_commands_web_command.rb` can be combined with the now-existing `test/rubygems/test_gem_commands_info_command.rb`; adjust as needed due to things being renamed
- the `lib/rubygems/command_manager.rb` and `Manifest.txt` changes become unnecessary

And then at that point you can address things like e.g. reworking it to have an explicit `--webpage` option and a `--rubytoolbox` option, and general cleanup as needed due to shuffling things around.

bronzdoc:
@MatheusRich 

> @bronzdoc Do you think it's better to open another PR and close this one, or just revert everything and start again here?

A new PR thanks! 🙇

PR #2836, opened 2019-07-17
Correctly handle symlinks.

### LINKED ISSUE #2826: Fix broken symlinks in tar
This is mostly copy of #2007, because it was not resolved at all :/ This was the issue and still is the issue:

~~~
$ gem unpack ascii_binder
Fetching: ascii_binder-0.1.10.1.gem (100%)
Unpacked gem: '/home/vondruch/ff/ascii_binder-0.1.10.1'

$ ll ascii_binder-0.1.10.1/bin/
celkem 12
-rwxr-xr-x. 1 vondruch vondruch 9864  8. zář 10.59 asciibinder
lrwxrwxrwx. 1 vondruch vondruch   15  8. zář 10.59 ascii_binder -> bin/asciibinder
~~~

Apparently, the ascii_binder should be link to asciibinder, but it points to bin/asciibinder.

I also pointed out that this test condition https://github.com/rubygems/rubygems/blob/master/test/rubygems/test_gem_package.rb#L172 should actually be:

~~~
assert_equal [{'lib/code_sym.rb' => 'code.rb'}], symlinks
~~~

Which was not fixed by #2516

voxik:
IOW don't touch symlink references. If they are wrong, it is upstream failure (there could be printed some warning, but this is out of the scope of this PR).

Also, avoid issues reported in #1577 and basically reverts #1578. The issues were caused by `File.exist?` failing on symlinks, although they were correctly available.

This resolves #2826 as well as #2007 and should not reintroduce #1881.


voxik:
I should note that I was considering to use `File.lstat` instead of checking for symlink and file existence, but that fails with exception and it appeared uglier to me. But I might rework the PR if it should be preferred.

voxik:
Also, I am not convinced that the warning suggesting `gem pristine` is useful. May be we should drop it and problem solved. It was introduced with this b524c653431f7394dffafef8689dbe67133a8392 commit without too much explanation. It seems that the intention was just to safeguard the `File.stat` called later.

voxik:
This would be the `lstat` version.

~~~
$ git diff
diff --git a/lib/rubygems/installer.rb b/lib/rubygems/installer.rb
index 0676dcea..f2012cb5 100644
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -505,15 +505,19 @@ class Gem::Installer
       filename.untaint
       bin_path = File.join gem_dir, spec.bindir, filename
 
-      unless File.exist? bin_path
+      begin
+        stat = File.lstat bin_path
+      rescue Errno::ENOENT
         # TODO change this to a more useful warning
         warn "`#{bin_path}` does not exist, maybe `gem pristine #{spec.name}` will fix it?"
         next
       end
 
-      mode = File.stat(bin_path).mode
-      dir_mode = options[:prog_mode] || (mode | 0111)
-      FileUtils.chmod dir_mode, bin_path unless dir_mode == mode
+      if stat.file?
+        mode = stat.mode
+        dir_mode = options[:prog_mode] || (mode | 0111)
+        FileUtils.chmod dir_mode, bin_path unless dir_mode == mode
+      end
 
       check_executable_overwrite filename
 
~~~

voxik:
> Also, I am not convinced that the warning suggesting `gem pristine` is useful. May be we should drop it and problem solved.

This might be problem on Windows, when the symlink is probably not created. Suggesting `gem pristine` is more like joke in this context.

deivid-rodriguez:
I guess for symlinked executables the correct solution would be to create a symlinked binstub, right?

deivid-rodriguez:
Are you still interested on this, @voxik?

voxik:
Sorry, it got buried in my TODO list :disappointed: Looking at it right now.

voxik:
So I swapped the conditions. If the file does not exist, the old warning message is used or information about dangling symlink is provided (that happens for the old ascii_binder, where the link was possibly broken in the archive). Looking more into history, the `gem pristine` might be actually useful and the message originates in this feature [here](https://github.com/rubygems/rubygems/commit/771f9e4d6c9713ae5f3d701f5048ee00b928a547). Considering this, I would leave the warning as they are now.

deivid-rodriguez:
Thanks @voxik, that all makes sense. I'll be offline until next Wednesday, then I'll come back to this PR :+1:.

lib/rubygems/installer.rb
```diff
@@ -484,16 +484,22 @@ def generate_bin # :nodoc:
...
+          dir_mode = options[:prog_mode] || (mode | 0111)
+          FileUtils.chmod dir_mode, bin_path unless dir_mode == mode
+        end
+      else
+        if File.symlink? bin_path
+          warn "`#{bin_path}` is dangling symlink pointing to `#{File.readlink bin_path}`"
+        else
+          warn "`#{bin_path}` does not exist, maybe `gem pristine #{spec.name}` will fix it?"
+        end
+
```
simi, line 499: :scissors:

deivid-rodriguez:
Alright, I finally got to this.

I rebased the PR, removed some irrelevant changes, added a test, and splitted it into two improvements as separate commits:

* Correctly handle symlinks when building gem packages by not adding any special handling.
* When gems created with previous rubygems versions include dangling symlinks (due to the bug we are fixing now), show a more accurate warning. 

I believe this should be ready now!

deivid-rodriguez:
Alright, going in!

PR #2841, opened 2019-07-23
Move default specifications dir definition out of BasicSpecification.

voxik:
This was never the right place. The method got there just by evolution, not by design. Move it within default methods, where it suits better.

Since this method is presumably used just internally, it should be safe to deprecate it and remove later.

### DIFF
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -35,7 +35,14 @@ class Gem::BasicSpecification
   end
 
   def self.default_specifications_dir
-    File.join(Gem.default_dir, "specifications", "default")
+    Gem.default_specifications_dir
+  end
+
+  class << self
+
+    extend Gem::Deprecate
+    deprecate :default_specifications_dir, "Gem::BasicSpecification.default_specifications_dir", 2020, 02
+
   end
 
   ##
@@ -84,7 +91,7 @@ class Gem::BasicSpecification
 
   def default_gem?
     loaded_from &&
-      File.dirname(loaded_from) == self.class.default_specifications_dir
+      File.dirname(loaded_from) == Gem.default_specifications_dir
   end
 
   ##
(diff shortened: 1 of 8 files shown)


lib/rubygems/basic_specification.rb
```diff
@@ -35,7 +35,14 @@ def initialize
...
 
   def self.default_specifications_dir
-    File.join(Gem.default_dir, "specifications", "default")
+    Gem.default_specifications_dir
+  end
+
+  class << self
+
+    extend Gem::Deprecate
+    deprecate :default_specifications_dir, "Gem::BasicSpecification.default_specifications_dir", 2020, 2
```
bronzdoc: 2019, 12?

lib/rubygems/basic_specification.rb
```diff
@@ -35,7 +35,14 @@ def initialize
   end
 
   def self.default_specifications_dir
```
hsbt: Can move this to under the `class << self` scope?

voxik: Well, looking at 021d303debfba923dc24dbfa99152c8c708e60c6, it appeared this is acceptable. Because otherwise, it should be probably modified everywhere.

lib/rubygems/basic_specification.rb
voxik: I guess `2019, 02` should be more readable, if that is acceptable.

lib/rubygems/basic_specification.rb
voxik: But if your proposal is the preferred way, then you are right that it could set bad example in the `basic_specification.rb`

hsbt: I see. I will update them with this patch.

hsbt:
Thanks! good catch.

@bundlerbot r+

PR #2846, opened 2019-07-26
Replace domain parameter with a parameter to suppress suggestions

bronzdoc:
# Description:
Replace domain parameter in Gem::Command#show_lookup_failure with a parameter to suppress suggestions

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -155,9 +155,8 @@ class Gem::Command
   ##
   # Display to the user that a gem couldn't be found and reasons why
   #--
-  # TODO: replace +domain+ with a parameter to suppress suggestions
 
-  def show_lookup_failure(gem_name, version, errors, domain, required_by = nil)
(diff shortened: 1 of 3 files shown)


bronzdoc:
bundler specs failing 🤔

deivid-rodriguez:
I think the failure is actually the `jruby` entry against `rubygems`. Bundler specs are failing when using `ruby-head` + `bundler`'s master, but that's already happening, so it's not a problem with this PR.

I think restarting the `jruby` build should fix things.

bronzdoc:
@hsbt @deivid-rodriguez r+ ?

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
I'm sorry I'm late for the review here!

This is a non private method whose signature we are changing, but I doubt anybody other than us is using it, or expect to be able to use it, so I think this is good :+1:

PR #2850, opened 2019-07-28
Fix @ran_rake assignment in builder.rb

bronzdoc:
# Description:
Initialize `@ran_rake` correctly, so we don't have to set it up later and cause confusion
______________


bronzdoc:
@bundlerbot r+

bronzdoc:
@bundlerbot retry

PR #2851, opened 2019-07-28
Move find_home to PathSupport

bronzdoc:
# Description:

Move `find_home` to `PathSupport`
______________

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -552,35 +552,6 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     return files
   end
 
-  ##
-  # Finds the user's home directory.
-  #--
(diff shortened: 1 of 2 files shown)


simi:
What is the reason for this? Thanks to this it must be public and I'm not sure this is something we need public.

bronzdoc:
@simi well... the reason was that it made sense to me that a class named `PathSupport` would be responsible of a path related method like `find_home`. But you have a point... maybe we can get rid of `find_home` completely? and add the logic into `user_home` since **A**.  `find_home` is not public and **B.** the only place used is in `user_home`.

simi:
I think we can leave it as is just remove the `FIXME` comment.

bronzdoc:
Will close this then @simi

PR #2849, opened 2019-07-28
ext/builder.rb cleanup

bronzdoc:
# Description:

`ext/builder.rb` cleanup
______________

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2852, opened 2019-07-28
Refactor Builder to lookup for builder before build_extension is called.

simi:
followup to https://github.com/rubygems/rubygems/pull/2850 @duckinator @bronzdoc 

It is possible to remove `@ran_rake` at all. If we find builder before we try to construct `extension_dir` in `build_extension` method, it will raise on invalid names like empty string or nil.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -109,8 +109,6 @@ class Gem::Ext::Builder
     @spec       = spec
     @build_args = build_args
     @gem_dir    = spec.full_gem_path
-
-    @ran_rake   = false
   end
 
   ##
@@ -123,12 +121,11 @@ class Gem::Ext::Builder
     when /configure/ then
       Gem::Ext::ConfigureBuilder
     when /rakefile/i, /mkrf_conf/i then
-      @ran_rake = true
       Gem::Ext::RakeBuilder
     when /CMakeLists.txt/ then
       Gem::Ext::CmakeBuilder
     else
-      extension_dir = File.join @gem_dir, File.dirname(extension)
+      extension_dir = File.join @gem_dir, File.dirname(extension.to_s)
 
       message = "No builder for extension '#{extension}'"
       build_error extension_dir, message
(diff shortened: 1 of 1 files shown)


lib/rubygems/ext/builder.rb
```diff
@@ -123,12 +121,11 @@ def builder_for(extension) # :nodoc:
     when /configure/ then
       Gem::Ext::ConfigureBuilder
     when /rakefile/i, /mkrf_conf/i then
-      @ran_rake = true
       Gem::Ext::RakeBuilder
     when /CMakeLists.txt/ then
       Gem::Ext::CmakeBuilder
     else
-      extension_dir = File.join @gem_dir, File.dirname(extension)
+      extension_dir = File.join @gem_dir, File.dirname(extension.to_s)
```
simi: This is the only place we need valid string.

lib/rubygems/ext/builder.rb
```diff
@@ -221,9 +203,10 @@ def build_extensions
     FileUtils.rm_f @spec.gem_build_complete_path
 
     @spec.extensions.each do |extension|
-      break if @ran_rake
+      builder = builder_for extension
+      build_extension extension, builder, dest_path
 
-      build_extension extension, dest_path
+      break if builder.is_a? Gem::Ext::RakeBuilder
```
simi: Code breaks after rake builder is triggered for some reason. I have no idea why this is needed. All spec passes also without this. It was added at https://github.com/rubygems/rubygems/commit/20cdf22b844a63d163cc80eff819b99330bd6fa1.

In theory if you have multiple extensions defined in gemspec, if rake one is specified, no other extension one after rake will be build.

:thinking: I don't see any reason for this. But for now I have kept this to not modify any behaviour.

MSP-Greg: Haven't been following that closely, but...

> In theory if you have multiple extensions defined in gemspec

An example would be EventMachine, two extconf.rb files, and two *.so files compiled.

simi: As I see https://github.com/eventmachine/eventmachine/blob/d761d2a6916b854b5d62aeaa4775be359e9cf6ce/eventmachine.gemspec#L16 there are two extensions defined, but both using `extconf` builder. But if first one will use rake builder, the second one will not be built and I have no idea what's the reason for this behaviour.

bronzdoc:
This PR would be easier to understand after we merge https://github.com/rubygems/rubygems/pull/2849?

what you think @simi @MSP-Greg ?

MSP-Greg:
@bronzdoc

Earlier today I saw the conversation, and it seemed to mention issues with multiple extensions.  I recalled that EM had two, so I mentioned it.  I've never had issues with building on Windows (at least related to RG), so I've never looked at the RG code.  Hence, right now I have no opinion.

I'll have a look at the code tomorrow (-0500 from GMT).

simi:
@bronzdoc #2849 does a little another approach, but I can combine the best of both PRs into this one. I have not noticed that one before, I'm sorry.

@MSP-Greg I'm just sharing my confusion about this behaviour, anyway this PR doesn't change it.

duckinator:
This looks like it should be good, but I agree that rebasing it will make it easier to understand (just because the diff will be smaller), so I'd be a bit more confident in my assessment at that point.

@simi when you have a moment, could you rebase this? (since #2849 was merged.)

Thanks for the PR, by the way!

bronzdoc:
Ah I missed your comment @simi 😞

bronzdoc:
@simi do you want me to pick this up?

PR #2860, opened 2019-08-01
Use gsub with Hash

znz:
# Description:

gsub with Hash is better than gsub with block.

______________

# Tasks:

- [ ] Write tests


duckinator:
Thanks for the PR, @znz!

Using `gsub` with a Hash has been around since 1.9.3 (which is older than the earliest supported by RubyGems 3.x), and the code is equivalent, so this looks good to me.

@bundlerbot r+

PR #2863, opened 2019-08-02
Update the latest version of bundler

hsbt:
# Description:

Update to e1c518363641208429f397170354054b3d28effd

We need to this before merging https://github.com/rubygems/rubygems/pull/2857
______________

# Tasks:

- [ ] Write tests


hsbt:
These test fails are caused by https://github.com/bundler/bundler/pull/7272

deivid-rodriguez:
Hi @hsbt!

I had a closer look at this, and bundler/bundler#7274 won't fix it :(

We fix it, we could merge #1868, which solves precisely the problem we're suffering from here. If we specify `-I bundler` during testing, we don't want the default copy of bundler to be activated, but want to use the locally vendored copy instead. This is what that PR is fixing.

hsbt:
@deivid-rodriguez Thanks, I confirmed to resolve to activate issue of bundler version with #1868

deivid-rodriguez:
@hsbt I merged #1868, so you can move this forward :+1:

hsbt:
@deivid-rodriguez Thanks! I did squash this branch, I will merge this after CI was finished.

hsbt:
@bundlerbot r+

PR #2874, opened 2019-08-14
Do not mutate uri.query during s3 signature creation

apakulov-stripe:
# Description:
My previous PR https://github.com/rubygems/rubygems/pull/2807 introduced a small regression. 

When it fetches a new file using `Gem::RemoteFetcher.fetch_path` it unpacks it if an URI ends with `.gz`: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/remote_fetcher.rb#L267:L273

In my PR above I mutated `uri` object and added params to the `uri.query`. As a result of that `uri` no longer ends with`.gz` and breaks `gem install` with s3 source.

- It fetches `prerelease_specs.4.8.gz` and stores as `prerelease_specs.4.8` on the file system. Keeping it unzipped.
- During `gem install` it will fail because of incompatible file-format:
```
    incompatible marshal file format (can't be read)
	format version 4.8 required; 31.139 given
```

The reason it didn't fail in previous iterations, because it relied on already cached file on a disk that was properly unzipped.

P.S. I also moved `ConfigurationError` message to cover all mechanism of fetching credentials.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/s3_uri_signer.rb
+++ b/lib/rubygems/s3_uri_signer.rb
@@ -49,12 +49,12 @@ class Gem::S3URISigner
     credential_info = "#{date}/#{s3_config.region}/s3/aws4_request"
     canonical_host = "#{uri.host}.s3.#{s3_config.region}.amazonaws.com"
 
-    uri.query = generate_canonical_query_params(s3_config, date_time, credential_info, expiration)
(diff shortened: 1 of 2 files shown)


hsbt:
Thanks, I got it.

@bundlerbot r+

PR #2877, opened 2019-08-15
Upgrade TravisCI to use bionic

deivid-rodriguez:
# Description:

Not sure if we'll change our CI to use github actions or something, but for now I need to upgrade to bionic, because I did that in the bundler repo, and now bundler CI fails because it fails to install a `groff` version only present in bionic.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: ruby
-dist: xenial
+dist: bionic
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

PR #2882, opened 2019-08-17
Set SOURCE_DATE_EPOCH env var if not provided.

### LINKED ISSUE #2290: reproducible builds proposal: make gem define SOURCE_DATE_EPOCH itself
I would like to suggest to make gem itself a potential SOURCE_DATE_EPOCH declarer instead of "only" making reproducible artifacts whenever the outside world defines the SOURCE_DATE_EPOCH environment variable.

While above works perfectly for all distros, as they and teir build tools and pipelines itself define SOURCE_DATE_EPOCH it would be awesome if the gem tool/script could define SOURCE_DATE_EPOCH itself.
This proposal would allow every gem aquired from the rubygems repository purely build with ```gem``` instead of any distro or other packaging related tool defining SOURCE_DATE_EPOCH to be independently reproduced.

This would need to check SOURCE_DATE_EPOCH in the gem command line tool, and if it is ***not*** yet define, it should define it to the current utc timestamp.

Example like it is done in Arch Linux's ```makepkg```:
https://git.archlinux.org/pacman.git/tree/scripts/makepkg.sh.in#n93

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [x] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Related to pull requests:
#2289 #2278

Spec:
https://reproducible-builds.org/specs/source-date-epoch/

Buy-in:
https://reproducible-builds.org/docs/buy-in/

duckinator:
# Description:

Set SOURCE_DATE_EPOCH env var if not provided.

Fixes #2290.

1. `Gem::Specification.date` returns SOURCE_DATE_EPOCH when defined,
2. this commit makes RubyGems set it _persistently_ when not provided.

This combination means that you can build a gem, check the build time,
and use that value to generate a new build -- and then verify they're
the same.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1242,6 +1242,23 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
 
   end
 
+  ##
+  # The SOURCE_DATE_EPOCH environment variable (or, if that's not set, the current time), converted to Time object.
+  # This is used throughout RubyGems for enabling reproducible builds.
+  #
+  # If it is not set as an environment variable already, this also sets it.
+  #
+  # Details on SOURCE_DATE_EPOCH:
+  # https://reproducible-builds.org/specs/source-date-epoch/
+
+  def self.source_date_epoch
+    if ENV["SOURCE_DATE_EPOCH"].nil? || ENV["SOURCE_DATE_EPOCH"].empty?
+      ENV["SOURCE_DATE_EPOCH"] = Time.now.to_i.to_s
+    end
+
+    Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.freeze
+  end
+
   # FIX: Almost everywhere else we use the `def self.` way of defining class
   # methods, and then we switch over to `class << self` here. Pick one or the
   # other.
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1667,7 +1667,7 @@ class Gem::Specification < Gem::BasicSpecification
   # https://reproducible-builds.org/specs/source-date-epoch/
 
   def date
-    @date ||= ENV["SOURCE_DATE_EPOCH"] ? Time.utc(*Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.to_a[3..5].reverse) : TODAY
+    @date ||= Time.utc(*Gem.source_date_epoch.utc.to_a[3..5].reverse)
   end
 
   DateLike = Object.new # :nodoc:
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -188,7 +188,7 @@ class Gem::Package
   def initialize(gem, security_policy) # :notnew:
     @gem = gem
 
-    @build_time      = ENV["SOURCE_DATE_EPOCH"] ? Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc : Time.now
+    @build_time      = Gem.source_date_epoch
     @checksums       = {}
     @contents        = nil
     @digests         = Hash.new { |h, algorithm| h[algorithm] = {} }
--- a/lib/rubygems/package/tar_writer.rb
+++ b/lib/rubygems/package/tar_writer.rb
@@ -123,7 +123,7 @@ class Gem::Package::TarWriter
 
     header = Gem::Package::TarHeader.new :name => name, :mode => mode,
                                          :size => size, :prefix => prefix,
-                                         :mtime => ENV["SOURCE_DATE_EPOCH"] ? Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc : Time.now
+                                         :mtime => Gem.source_date_epoch
 
     @io.write header
     @io.pos = final_pos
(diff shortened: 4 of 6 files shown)


duckinator:
Looks like the test is wrong and there are contents that don't match. It seems like all of the problems are related to the mtimes for various things in the final `.gem` file.

duckinator:
It now has a passing test for reproducibility that can be made to fail by changing:

```ruby
ENV["SOURCE_DATE_EPOCH"] = new_epoch
```

to:

```ruby
ENV["SOURCE_DATE_EPOCH"] = (new_epoch.to_i + 1).to_s
```

So I am a lot more confident that it actually works properly, now.

duckinator:
So, assuming CI passes, the current state of this is: there is a test proving generated .gem files for the same source with the same SOURCE_DATE_EPOCH are equivalent, but I have had difficulty doing this manually for some reason.

By screwing with SOURCE_DATE_EPOCH on the second build in the test, I can make the test fail, so I don't think it's the test that's wrong this time.

lib/rubygems.rb
```diff
@@ -1242,6 +1242,23 @@ class << self
...
+  # https://reproducible-builds.org/specs/source-date-epoch/
+
+  def self.source_date_epoch
+    if ENV["SOURCE_DATE_EPOCH"].to_s.empty?
+      ENV["SOURCE_DATE_EPOCH"] = Time.now.to_i.to_s
+    end
+
+    Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.freeze
+  end
+
```
bronzdoc: What you think of:
```ruby
def self.source_date_epoch
  source_date_epoch = ENV["SOURCE_DATE_EPOCH"]
  source_date_epoch = Time.now.to_i if source_date_epoch.strip.empty?
 
  Time.at(source_date_epoch).utc.freeze
end
```

lib/rubygems/specification.rb
```diff
@@ -1667,7 +1667,7 @@ def has_conflicts?
   # https://reproducible-builds.org/specs/source-date-epoch/
 
   def date
-    @date ||= ENV["SOURCE_DATE_EPOCH"] ? Time.utc(*Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.to_a[3..5].reverse) : TODAY
+    @date ||= ENV["SOURCE_DATE_EPOCH"] ? Time.utc(*Gem.source_date_epoch.utc.to_a[3..5].reverse) : TODAY
```
bronzdoc: Shouldn't this be:
```ruby
@date ||= Gem.source_date_epoch ? Time.utc(*Gem.source_date_epoch.to_a[3..5].reverse) : TODAY
```

lib/rubygems.rb
```diff
@@ -1242,6 +1242,23 @@ class << self
...
+  # The SOURCE_DATE_EPOCH environment variable (or, if that's not set, the current time), converted to Time object.
+  # This is used throughout RubyGems for enabling reproducible builds.
+  #
+  # If it is not set as an environment variable already, this also sets it.
+  #
+  # Details on SOURCE_DATE_EPOCH:
+  # https://reproducible-builds.org/specs/source-date-epoch/
+
+  def self.source_date_epoch
+    if ENV["SOURCE_DATE_EPOCH"].to_s.empty?
```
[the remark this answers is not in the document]
duckinator: env vars can also be nil. I could switch this to `ENV["SOURCE_DATE_EPOCH"].nil? || ENV["SOURCE_DATE_EPOCH"].empty?`

I think that'd probably be clearer.

lib/rubygems.rb
duckinator: The entire purpose of this PR is to set `ENV["SOURCE_DATE_EPOCH"]` if it hasn't been set, which your variant doesn't do.

lib/rubygems/specification.rb
duckinator: No, because `Gem.source_date_epoch` always returns a value.

However, since `Gem.source_date_epoch` always returns a value, I think that means it _can_ become this:

```ruby
@date ||= Time.utc(*Gem.source_date_epoch.utc.to_a[3..5].reverse)
```

duckinator: It can probably be able to be simplified more than that, but I'd need to dig into the actual result of that function call to determine that.

duckinator:
Thanks for the feedback @bronzdoc. I'll go through and update the code in a bit. :+1:

lib/rubygems/specification.rb
duckinator: It looks like what that code does is take a UTC time and return that date, but at midnight UTC. So I can't think of a way off-hand to simplify it farther than what I said above, but that's still a nice improvement.

```
>> Time.now.utc
=> 2019-08-20 20:44:18 UTC
>> Time.utc(*Time.now.utc.to_a[3..5].reverse)
=> 2019-08-20 00:00:00 UTC
```

duckinator:
@bronzdoc I just pushed changes which should resolve most of what you mentioned. Let me know what you think. :slightly_smiling_face:

duckinator:
Just rebased it off latest master.

djberg96:
LGTM 👍

duckinator:
@bundlerbot r=djberg96

Issue #2859, opened 2019-07-31
Let `gem build` without additional arguments build the only gemspec in a directory


ghost:
I'm having a problem or would like to suggest a feature.

My current problem is that it's an annoyance to always type out the full command (i.e. `gem build gemname.gemspec`), especially when working on multiple gems.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
`3.0.0.beta3`
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

**I suggest to only allow this to work, when there is only one `gemspec` file in the current directory.**

bronzdoc:
Maybe we can revisit https://github.com/rubygems/rubygems/pull/2610

LemonAndroid:
pls no :D 

i mean yes, ofc, tho

duckinator:
I'm very strongly in favor of this, explicitly with the condition of only having it work if there's exactly one gemspec in the current directory (as the original issue suggests).

@bronzdoc the approach taken in #2610 causes a pretty drastic increase in complexity, so I'd like to redirect back to the issue as it was opened here. It feels like a more streamlined and manageable subset of your idea, to me.

bronzdoc:
@duckinator 
> explicitly with the condition of only having it work if there's exactly one gemspec in the current directory

Yeah, i'm working on something of that nature. 

Definitely https://github.com/rubygems/rubygems/pull/2610 brings more complexity

PR #2887, opened 2019-08-18
Build the first gemspec we found if no arguments are passed to gem build

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/2859

Build the first gemspec we found if no arguments are passed to `gem build`.

```shell
 $ ls
total 72
-rw-r--r--  1 bronzdoc  staff    93 Aug 18 14:52 Gemfile
-rw-r--r--  1 bronzdoc  staff  1167 Aug 18 14:52 README.md
-rw-r--r--  1 bronzdoc  staff    51 Aug 18 14:52 Rakefile
drwxr-xr-x  4 bronzdoc  staff   128 Aug 18 14:52 bin
drwxr-xr-x  4 bronzdoc  staff   128 Aug 18 14:52 lib
-rw-r--r--  1 bronzdoc  staff  1206 Aug 18 14:53 test_gem.gemspec
```

```shell
$ gem build
  Successfully built RubyGem
  Name: test_gem
  Version: 0.1.0
  File: test_gem-0.1.0.gem
```
______________

### DIFF
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -62,7 +62,7 @@ Gems can be saved to a specified filename with the output option:
 
   private
 
-  def build_gem(gemspec = get_one_gem_name)
+  def build_gem(gemspec = get_one_optional_argument)
     unless File.exist?(gemspec)
       gemspec += ".gemspec" if File.exist?(gemspec + ".gemspec")
     end
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -82,7 +82,7 @@ Gems can be saved to a specified filename with the output option:
 
       if options[:build_path]
         Dir.chdir(File.dirname(gemspec)) do
-          spec = Gem::Specification.load File.basename(gemspec)
+          spec = Gem::Specification.load(File.basename(gemspec))
           build_package(spec)
         end
       else
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -57,10 +57,18 @@ Gems can be saved to a specified filename with the output option:
   end
 
   def execute
-    build_gem
+    build_gem(gem_name)
   end
 
   private
+
+  def gem_name
+    get_one_optional_argument || find_gemspecs.first
+  end
+
+  def find_gemspecs
+    Dir.glob("*.gemspec").sort
+  end
 
   def build_gem(gem_name = get_one_optional_argument)
     gemspec = File.exist?(gem_name) ? gem_name : "#{gem_name}.gemspec"
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -57,14 +57,11 @@ Gems can be saved to a specified filename with the output option:
   end
 
   def execute
+    gem_name = get_one_optional_argument || find_gemspec
     build_gem(gem_name)
   end
 
   private
-
-  def gem_name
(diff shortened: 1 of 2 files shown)


duckinator:
@bronzdoc I'd recommend raising an error if >1 gemspecs are found and one isn't specified, but aside from that this looks good! :slightly_smiling_face:

bronzdoc:
Thanks for the review @duckinator!

duckinator:
@bronzdoc this looks good to me! I'm not sure what Travis CI is unhappy, though.

bronzdoc:
@duckinator it seems it was an rvm issue in tavis ci, i just restarded the build

test/rubygems/test_gem_commands_build_command.rb
```diff
@@ -242,6 +242,94 @@ def test_can_find_gemspecs_without_dot_gemspec
...
+
+    gemspecs = ["another_gem-2.gemspec", "some_gem-2.gemspec"]
+    assert_equal "", @ui.output
+    assert_equal @ui.error, "ERROR:  Multiple gemspecs found: #{gemspecs}, please specify one\n"
+
+    expected_gem = File.join(gemspec_dir, File.basename(another_gem.cache_file))
+    refute File.exist?(expected_gem)
+  end
+
+  def util_test_build_gem(gem)
```
deivid-rodriguez: This method seems duplicated.

bronzdoc: Weird, just removed it.

lib/rubygems/commands/build_command.rb
```diff
@@ -57,13 +57,30 @@ def usage # :nodoc:
...
+    if gemspecs.size > 1
+      alert_error "Multiple gemspecs found: #{gemspecs}, please specify one"
+      terminate_interaction(1)
     end
 
-    if File.exist? gemspec
+    gemspecs.first
+  end
+
+  def build_gem(gem_name = get_one_optional_argument)
```
deivid-rodriguez: This method is never called without an argument, so the optional value is unnecessary. Also, the parameter (`gem_name`) is called with the same name as a method, that confuses me a bit, personally.

I would remove the `build_gem` and `gem_name` methods and inline the whole logic inside the `execute` method, because the execute method now just delegates to `build_gem`, which seems like an unnecessary indirection to me that doesn't help readability.

Something like

```ruby
def execute
  gem_name = get_one_optional_argument || find_gemspec
  gemspec = File.exist?(gem_name) ? gem_name : "#{gem_name}.gemspec"

  # ... and the rest of the current `build_gem` logic ...
end
```

bronzdoc: I removed the `gem_name` method.

But i do prefer:

```ruby
def execute
  gem_name = get_one_optional_argument || find_gemspec
  build_gem(gem_name)
end
```

over 

```ruby
def execute
  gem_name = get_one_optional_argument || find_gemspec
  gemspec = File.exist?(gem_name) ? gem_name : "#{gem_name}.gemspec"

    if File.exist?(gemspec)
      spec = Gem::Specification.load(gemspec)

      if options[:build_path]
        Dir.chdir(File.dirname(gemspec)) do
          spec = Gem::Specification.load(File.basename(gemspec))
          build_package(spec)
        end
      else
        build_package(spec)
      end

    else
      alert_error "Gemspec file not found: #{gemspec}"
      terminate_interaction(1)
    end
end
```
In the former I don't have a doubt what `execute` is exactly doing

bronzdoc:
@bundlerbot r+

bronzdoc:
@bundlerbot r+

PR #2893, opened 2019-08-21
Improve `gem uninstall --all`

### LINKED ISSUE #2892: uninstall a gem then crash on default (Gem::InstallError)
Hello 😊

My current problem is that when I try to remove a gem, it say that the gem is uninstalled but then crash.

On RSpec we have to do some weird things to properly handle multiple Ruby and Rails version on Travis. https://github.com/rspec/rspec-rails/pull/2158/files#diff-b262826f376893c7c9a30ff4b124d72aR12

We do `gem uninstall -aIx bundler` but we get two information that seems to be the opposite (?)

```
Removing bundler
Successfully uninstalled bundler-2.0.2
ERROR:  While executing gem ... (Gem::InstallError)
    gem "bundler" cannot be uninstalled because it is a default gem
```

* Does it means the second error is for another bundler gem version?
* How could we avoid that crash on default gems?

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
ruby --version
ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-linux]

$ rvm --version
rvm 1.29.3 (latest) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]

$ bundle --version
Bundler version 1.16.2

$ gem --version
2.7.7
```

deivid-rodriguez:
# Description:

After we started unswallowing errors of `gem uninstall` in #2707, some people got failing CI runs because they were doing things like `gem uninstall <default_gem> --all --force`. Previously that would work just fine, because the error while trying to uninstall a default gem was being swallowed.

In my opinion, if you try to do `gem uninstall bundler:1.17.2` (on a ruby where this is the default bundler version), then it is fine to fail hard, but if you do `gem uninstall bundler --all`, then the command should uninstall all bundler versions that can be uninstalled (and skip the default version), but not fail just because bundler is a default gem.

It can be argued though that the command should indeed fail hard, because it didn't really succeeded on trying to do what the user asked for: the user tried to uninstall _all_versions, but that wasn't achieved because the default version was left around.

In order to find a compromise between these two visions, and maximize usability, I propose to not fail hard, but log a line for each default version that couldn't be removed.

Note that the current behavior is completely inconsistent (in my opinion), because it fails if the default copy is the single copy in the system, but succeeds if it's not:

```
$ gem uninstall rdoc --all --force 
ERROR:  While executing gem ... (Gem::InstallError)
    gem "rdoc" cannot be uninstalled because it is a default gem

$ gem install rdoc
Fetching rdoc-6.1.1.gem
Successfully installed rdoc-6.1.1
1 gem installed

$ gem uninstall rdoc --all --force 
Successfully uninstalled rdoc-6.1.1
```

After this PR, the messages are more consistent and informative, in my opinion, and the status code is always success (as it was before #2707).

```
$ ruby -Ilib bin/gem uninstall rdoc --all --force
Ignored rdoc-6.1.0 because it is a default gem

$ gem install rdoc
Fetching rdoc-6.1.1.gem
Successfully installed rdoc-6.1.1
1 gem installed

$ ruby -Ilib bin/gem uninstall rdoc --all --force
Ignored rdoc-6.1.0 because it is a default gem
Successfully uninstalled rdoc-6.1.1
```

Thoughts?

# Tasks:

- [x] Write tests

Closes #2892.

### DIFF
--- a/lib/rubygems/uninstaller.rb
+++ b/lib/rubygems/uninstaller.rb
@@ -88,6 +88,10 @@ class Gem::Uninstaller
       list << spec
     end
 
+    if list.empty?
+      raise Gem::InstallError, "gem #{@gem.inspect} is not installed"
+    end
+
     default_specs, list = list.partition do |spec|
       spec.default_gem?
     end
@@ -101,9 +105,7 @@ class Gem::Uninstaller
 
     if list.empty?
       if other_repo_specs.empty?
-        if default_specs.empty?
-          raise Gem::InstallError, "gem #{@gem.inspect} is not installed"
-        else
+        if default_specs.any?
           message =
             "gem #{@gem.inspect} cannot be uninstalled " +
             "because it is a default gem"
--- a/lib/rubygems/uninstaller.rb
+++ b/lib/rubygems/uninstaller.rb
@@ -96,6 +96,10 @@ class Gem::Uninstaller
       spec.default_gem?
     end
 
+    default_specs.each do |default_spec|
+      say "Gem #{default_spec.full_name} cannot be uninstalled because it is a default gem"
+    end
+
     list, other_repo_specs = list.partition do |spec|
       @gem_home == spec.base_dir or
         (@user_install and spec.base_dir == Gem.user_dir)
@@ -104,14 +108,7 @@ class Gem::Uninstaller
     list.sort!
 
     if list.empty?
-      if other_repo_specs.empty?
-        if default_specs.any?
-          message =
-            "gem #{@gem.inspect} cannot be uninstalled " +
-            "because it is a default gem"
-          raise Gem::InstallError, message
-        end
-      end
+      return unless other_repo_specs.any?
 
       other_repos = other_repo_specs.map { |spec| spec.base_dir }.uniq
 
--- a/test/rubygems/test_gem_uninstaller.rb
+++ b/test/rubygems/test_gem_uninstaller.rb
@@ -221,13 +221,13 @@ class TestGemUninstaller < Gem::InstallerTestCase
 
     uninstaller = Gem::Uninstaller.new spec.name, :executables => true
 
-    e = assert_raises Gem::InstallError do
+    use_ui @ui do
       uninstaller.uninstall
     end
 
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
If we did this, and released it with 3.0.7, we'll probably save some people from having to change some CI scripts.

lib/rubygems/uninstaller.rb
```diff
@@ -88,10 +88,18 @@ def uninstall
       list << spec
     end
 
+    if list.empty?
+      raise Gem::InstallError, "gem #{@gem.inspect} is not installed"
```
benoittgt: Don't you think we could instead have an error class `Gem::UninstallError` when we do uninstall? 

It is probably a breaking change... :/

benoittgt:
Thanks a lot for the quick PR @deivid-rodriguez and @hsbt for the review. 🙏

lib/rubygems/uninstaller.rb
deivid-rodriguez: Absolutely, it was very confusing the first time I saw this. I'll try to change this in the future.

test/rubygems/test_gem_uninstaller.rb
```diff
@@ -221,13 +221,13 @@ def test_uninstall_default_gem
...
+    use_ui @ui do
       uninstaller.uninstall
     end
 
-    assert_equal 'gem "default" cannot be uninstalled ' +
-                 'because it is a default gem',
-                 e.message
+    lines = @ui.output.split("\n")
+
+    assert_equal 'Ignored default-2 because it is a default gem', lines.shift
```
bronzdoc: What is wrong with the original error message?

deivid-rodriguez: You're right, the previous message was probably as good as this one. I'll revert to the one in place.

deivid-rodriguez: @bronzdoc updated!

PR #2894, opened 2019-08-21
Use IAM role to extract security-credentials for EC2 instance

apakulov-stripe:
# Description:
Accidentally used the wrong metadata endpoint that is reserved for the internal use. If we want to use proper EC2 instance-profile credentials we need to extract the role-name first by calling `iam/info` and use it to construct `iam/security-credentials/role-name`.

If `iam_role` is not attached to the instance it will raise an exception `ec2_metadata_request` since `iam/info` will return 404.

See metadata documentation for more details: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

Is there a chance we can have a new release for this? (I've been waiting for `3.0.5`/`3.0.6`, but unfortunately notice this issue just now).

Thank you!

### DIFF
--- a/lib/rubygems/s3_uri_signer.rb
+++ b/lib/rubygems/s3_uri_signer.rb
@@ -150,16 +150,23 @@ class Gem::S3URISigner
     require 'rubygems/request/connection_pools'
     require 'json'
 
-    metadata_uri = URI(EC2_METADATA_CREDENTIALS)
-    @request_pool ||= create_request_pool(metadata_uri)
-    request = Gem::Request.new(metadata_uri, Net::HTTP::Get, nil, @request_pool)
+    iam_info = ec2_metadata_request(EC2_IAM_INFO)
+    # Expected format: arn:aws:iam::<id>:instance-profile/<role_name>
+    role_name = iam_info['InstanceProfileArn'].split('/')[1]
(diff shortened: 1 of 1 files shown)


lib/rubygems/s3_uri_signer.rb
```diff
@@ -150,16 +150,23 @@ def ec2_metadata_credentials_json
     require 'rubygems/request/connection_pools'
     require 'json'
 
-    metadata_uri = URI(EC2_METADATA_CREDENTIALS)
-    @request_pool ||= create_request_pool(metadata_uri)
-    request = Gem::Request.new(metadata_uri, Net::HTTP::Get, nil, @request_pool)
+    iam_info = ec2_metadata_request(EC2_IAM_INFO)
+    # Expected format: arn:aws:iam::<id>:instance-profile/<role_name>
+    role_name = iam_info['InstanceProfileArn'].split('/')[1]
```
bronzdoc: `role_name = iam_info['InstanceProfileArn'].split('/').last` ?

apakulov-stripe:
@bronzdoc PTAL

bronzdoc:
Thanks @apakulov-stripe!

@bundlerbot r+

apakulov-stripe:
@bronzdoc what is the ETA for the next release?

Issue #2886, opened 2019-08-18
gem update installs system-wide updates in home


exploide:
This is a follow up of #2273 which was closed due to inactivity.

When `gem update` is invoked with `--user-install` option (e.g. set in `.gemrc`), it installs updates to the home directory, as expected. But it also installs updates for system-wide gems there. Like the reporter of #2273, I also think this is usually not wanted.

I propose to only update user-installed gems when `gem update` is invoked with `--user-install`.

Alternatively, I'm interested in any other way to update exclusively user-installed gems as a workaround when this gets rejected. 

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
> gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.0.3
  - RUBY VERSION: 2.6.3 (2019-04-16 patchlevel 62) [x86_64-linux]
  - INSTALLATION DIRECTORY: /usr/share/gems
  - USER INSTALLATION DIRECTORY: /home/jannik/.gem/ruby
  - RUBY EXECUTABLE: /usr/bin/ruby
  - GIT EXECUTABLE: /usr/bin/git
  - EXECUTABLE DIRECTORY: /usr/bin
  - SPEC CACHE DIRECTORY: /home/jannik/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /usr/share/gems
     - /home/jannik/.gem/ruby
     - /usr/local/share/gems
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - "gem" => "--no-document --user-install --bindir /home/jannik/.gem/bin"
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /home/jannik/.gem/bin
     - /home/jannik/.local/bin
     - /usr/local/sbin
     - /usr/sbin
     - /usr/local/bin
     - /usr/bin
     - /bin
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
@exploide  can you provide the contents of your `.gemrc` or how are you setting the `--user-install` ? 

[...]

PR #2901, opened 2019-08-31
Fix gem update --user-install

bronzdoc:
# Description:

Closes https://github.com/rubygems/rubygems/issues/2886

Update only user insatalled gems when the `--user-install` flag is passed to `gem update`

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -136,6 +136,9 @@ command to remove old versions.
   def highest_installed_gems # :nodoc:
     hig = {} # highest installed gems
 
+    # Get only gem specifications installed as --user-install
+    Gem::Specification.dirs = Gem.user_dir if options[:user_install]
+
     Gem::Specification.each do |spec|
       if hig[spec.name].nil? or hig[spec.name].version < spec.version
         hig[spec.name] = spec
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -73,8 +73,6 @@ command to remove old versions.
       say "Latest version already installed. Done."
       terminate_interaction
     end
-
-    options[:user_install] = false
   end
 
   def check_update_arguments # :nodoc:
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -90,9 +90,10 @@ command to remove old versions.
       return
     end
 
-    hig = highest_installed_gems
-
-    gems_to_update = which_to_update hig, options[:args].uniq
+    gems_to_update = which_to_update(
+      highest_installed_gems,
+      options[:args].uniq
+    )
 
     if options[:explain]
       say "Gems to update:"
--- a/test/rubygems/test_gem_commands_update_command.rb
+++ b/test/rubygems/test_gem_commands_update_command.rb
@@ -359,10 +359,10 @@ class TestGemCommandsUpdateCommand < Gem::TestCase
   end
 
   def test_execute_user_install
-    spec_fetcher do |fetcher|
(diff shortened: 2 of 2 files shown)


duckinator:
@bronzdoc should this perhaps print a message with a quick explanation that non-user gems are being skipped + how to update those?

Other than that, this looks good to me overall, but would definitely be a breaking change so I'm not sure when it should be merged.

bronzdoc:
@duckinator  I don't  know 🤔  `gem update --user-install` updating system-wide gems seems like a bug to me.

MSP-Greg:
I agree with both of you.  But, I'd prefer the change.  I'm always updating my local ruby-head build, so all my gems used for testing are in --user-install...

duckinator:
@bronzdoc thinking on it a bit more, I can see arguments in either direction. I'm cool with it being merged and treated as a bugfix if other folks are. :+1:

deivid-rodriguez:
I'll review this tomorrow, :+1:. I do agree in principle with considering this a bug.

test/rubygems/test_gem_commands_update_command.rb
```diff
@@ -373,7 +373,13 @@ def test_execute_user_install
     installer = @cmd.installer
     user_install = installer.instance_variable_get :@user_install
 
-    assert user_install, 'user_install must be set on the installer'
+    assert user_install, "user_install must be set on the installer"
```
deivid-rodriguez: Usual pickiness of mine of not changing unnecessary code just for style.

lib/rubygems/commands/update_command.rb
```diff
@@ -137,6 +136,9 @@ def fetch_remote_gems(spec) # :nodoc:
   def highest_installed_gems # :nodoc:
     hig = {} # highest installed gems
 
+    # Get only gem specifications installed as --user-install
+    Gem::Specification.dirs = Gem.user_dir if options[:user_install]
```
deivid-rodriguez: I think this is the actual fix needed and none of the changes in `lib/rubygems/specification.rb` are actually needed?

deivid-rodriguez: I don't know where the new methods are being used? :thinking: I see what you're trying to do, but I don't think you finished the implementation? You might need to loop only through the newly added `Specification.user_installed_stubs` in the line below this one.

Also, we would need a new test to make sure that default gems are not updated.

bronzdoc: Hmm 🤔 you are right, I mixed things up.

I'll move the changes of `specification.rb` to another/followup PR

test/rubygems/test_gem_commands_update_command.rb
bronzdoc: 🤔 I usually don't, unless is related  code that I'm modifying any way 🙈

bronzdoc:
@bundlerbot try

bronzdoc:
Build is not running after a force push 🙄

test/rubygems/test_gem_commands_update_command.rb
deivid-rodriguez: Right. Although I prefer to never touch style unless explicitly normalizing it, I'm also aware of that approach of "fix style as you go".

I won't bug you with this kind of comments any more if the code you're changing contains minor style tweaks next to it :+1:.

lib/rubygems/commands/update_command.rb
deivid-rodriguez: Yeah. This doesn't completely fix the bug, but it improves things over the current situation in my opinion, so let's do it. I'll open a separate ticket for the remaining issue.

deivid-rodriguez:
Let's see how our bot is feeling now :)

@bundlerbot merge

PR #2904, opened 2019-09-03
Fix spurious chdir warnings while inside other chdir blocks

steakknife:
This solves spurious warnings like the following:

```
{{SITE_RUBY}}/rubygems/ext/builder.rb:183: warning: conflicting chdir during another chdir block
{{SITE_RUBY}}/rubygems/ext/builder.rb:191: warning: conflicting chdir during another chdir block
Successfully installed
```

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -161,18 +161,20 @@ EOF
       FileUtils.mkdir_p dest_path
 
       CHDIR_MUTEX.synchronize do
-        pwd = Dir.getwd
-        Dir.chdir extension_dir
+        maybe_failed_to_chdir_back_to_pwd = false
         begin
-          results = builder.build(extension, dest_path,
-                                  results, @build_args, lib_dir)
+          Dir.chdir extension_dir do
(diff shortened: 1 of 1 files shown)


steakknife:
So it doesn't re-`raise` as in the previous code's expected failure case.

```diff
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -171,8 +171,11 @@ EOF
             maybe_failed_to_chdir_back_to_pwd = true
           end
         rescue SystemCallError
-          Dir.chdir dest_path if maybe_failed_to_chdir_back_to_pwd
-          raise
+          if maybe_failed_to_chdir_back_to_pwd
+            Dir.chdir dest_path
+          else
+            raise
+          end
         end
       end
```

deivid-rodriguez:
@steakknife Can you clarify what you mean by your last comment? Does it mean that the extra portion of the diff is not necessary to eliminate the warnings, but for not re-raising as we did before?

deivid-rodriguez:
Ping @steakknife

PR #2908, opened 2019-09-10
Fix underscore version selection for bundler

deivid-rodriguez:
# Description:

When rubygems' underscore version selection is used with bundler itself, it doesn't work. This PR should fix it.

# Tasks:

- [x] Write tests

Fixes bundler/bundler#7345.
Closes bundler/bundler#7349.

### DIFF
--- a/lib/rubygems/dependency.rb
+++ b/lib/rubygems/dependency.rb
@@ -281,7 +281,7 @@ class Gem::Dependency
       requirement.satisfied_by?(spec.version) && env_req.satisfied_by?(spec.version)
     end.map(&:to_spec)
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I added a test for this fix. "MacOS 2.3.x" check failed initially but I think it was unrelated, so I restarted it and it's now green.

The main benefit of this fix is that it allows users to easily skip the "BUNDLED WITH" automatic bundler version locking by specifying the desired version surrounded by underscores (`bundle _1.17.3_ install`, or whatever), just like with any other rubygems binstub.

hsbt:
@bundlerbot retry

PR #2909, opened 2019-09-11
Use different directory structure for "default" gems. 

voxik:
Currently, all gems are installed into "gems" directory under the Ruby direcotory structure. The patch moves the "default" gems out of that structure and puts them into separate "default" structure.

This makes the "default" gems more similar to the regular gems, allowing to simplify the codebase and avoiding issues with overriding `Gem.default_dir` via operating_system.rb.

This aims to resolve https://bugs.ruby-lang.org/issues/14737.


voxik:
My biggest concern with this PR is actually naming, because naming is hard. I think this demonstrates the issue a bit:

https://github.com/rubygems/rubygems/pull/2909/commits/9c315143ea178a0626532883b9e8b6166e9fd257#diff-0c4b3e9996ebec38f191e68264665984R322

The commit 9c315143ea178a0626532883b9e8b6166e9fd257 moves the default gems into "default" directory structure, but as you can see, the naming is really confusing. With this PR, it is no more obvious, what should be the directory structures. Using "gems" and "default" appears unfortunate. So I am looking for alternatives. For some time, I was considering something along the lines of "stdlib", but stdlib is currently split to "default" and "bundled" gems, but this does not leave any room for user installed gems.

OTOH, this can be seen as first step move the user installed gems out of the ruby directory structure, where they does not belong.

voxik:
Chm, linting errors introduced long time ago preventing me from seeing CI results? That is disappointing :disappointed:

deivid-rodriguez:
Sorry about that, and thanks for your work. :muscle: 

The lint errors are being introduced by this PR though. Rubocop's default is to forbid redundant whitespace except for alignment. In this case, by introducing the `if` clause, `rubocop` considers the new assigments as separate blocks of code and thus complains about the extra spaces.

I understand this is misleading, since you obviously tried to keep the existing style, so I'll remove the "except for alignment" thing on a separate PR.

voxik:
> so I'll remove the "except for alignment" thing on a separate PR.

This makes my day much better, thx :wink:

deivid-rodriguez:
Well, this seems like a bug in `rubocop` to me, actually.

I tried enabling

```yaml
Layout/SpaceAroundOperators:
  AllowForAlignment: false
  Enabled: true
```

And that didn't complain about the offending block of code, so I don't know why it would complain about your changes. :thinking:.

Anyways, sorry about this. Could you change this to use a ternary operator or something for now? Or feel free to temporarily modify the `.travis.yml` file to do ` util/ci.sh rubocop || true` or whatever so you can see more CI feedback.

deivid-rodriguez:
Great, I opened [an issue at rubocop](https://github.com/rubocop-hq/rubocop/issues/7344) to try figure this out.

voxik:
Testing this with Ruby, it seems I'll need to submit as PR against Ruby repository, or at least convince rbinstall to use Gem::Installer instead of the custom code used ATM ...

deivid-rodriguez:
@voxik Is https://github.com/ruby/ruby/pull/2545 what you're talking about?

voxik:
Yes, thx for referencing it here. I wanted to open a ticket in Redmine, but I still have not done all the testing I would love to do :see_no_evil: especially the tests together with this PR. But if some reviews/approves/merges, I won't object :blush:

voxik:
BTW I still think that something like the `UnpackedInstaller` should belong to Rubygems/Bundler. But since I don't think I'll be able to work on such feature anytime soon, I just leave the idea noted in this ticket :wink:

deivid-rodriguez:
Just to make sure, both PRs have currently a red CI status, are they ready or do they need more work?

voxik:
The ruby/ruby#2545 is red just because of some random GC failure on Windows. I cannot imagine there would be any relation to my PR.

This one is red due to some Ruby 2.7 issue and not sure about the other reason. I would need to investigate.

swrobel:
@voxik maybe rebase off master and see how CI fares now that Ruby 2.7 is officially released?

voxik:
Rebased, but there is not one issue I need to investigate :disappointed: 

~~~
   1) Failure:
TestGemInstaller#test_install_creates_binstub_that_prefers_user_installed_gem_to_default [/home/runner/work/rubygems/rubygems/test/rubygems/test_gem_installer.rb:1013]:
default gem's executable not installed.
Expected path '/tmp/test_rubygems_3049/gemhome/bin/executable' to exist.
~~~

voxik:
Not sure what are the test failures. They do not seems to be related :man_shrugging:

deivid-rodriguez:
I'll try to have a look at what's going on. Thanks for working on this :purple_heart:.

voxik:
I think that the `TestGemCommandsInstallCommand#test_execute_rdoc_with_path` issue is due to RDoc 6.2.0 and https://github.com/ruby/rdoc/commit/6e78ab0eba2ebbe4ed2bc4f4e8b3efc9823a452e or RDoc 6.2.1 should fix it. IOW `Gemfile.lock` should be refreshed.

Nevertheless, why is the `Gemfile.lock` stored in the repo anyway? I always thought it is bad practice to store Gemfile.lock in the repo [1].


[1]: https://stackoverflow.com/questions/4151495/should-gemfile-lock-be-included-in-gitignore

simi:
@voxik IMHO it is stored in repo to get dev. dependencies frozen. That linked SO question is more than 10 years old. It was recommended to not store `Gemfile.lock` for libraries before, but I think this changed over the time and it is best practice now (for example to get reproducible builds on CI).

voxik:
Given the #5043, maybe I should dustoff this attempt ...

deivid-rodriguez:
Mmmm, do you think this would fix that? In any case, I'm definitely 👍 on this direction, I think it should also help with #5098.

lib/rubygems/defaults.rb
```diff
@@ -45,6 +45,27 @@ def self.default_dir
...
   end
 
+  ##
+  # Directory to StdLib default gems
+
+  def self.default_gems_dir
+    path = if defined? RUBY_FRAMEWORK_VERSION
+             [
+               File.dirname(RbConfig::CONFIG['sitedir']),
+               'Default',
```
pvalena, line 55: Is the capital `D` intentional?

pvalena:
LGTM.

lib/rubygems/defaults.rb
voxik: Yes, similarly to `Gems` in `.default_dir` above. But this was refactored and dropped, so there is no point to discuss this anymore.

voxik:
I have pushed rebase. However, there are some test failures. Mainly these two:

~~~
.F
===============================================================================
Failure: test_install_default_bundler_gem(TestGemCommandsSetupCommand): <"/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230413-1951-1y5leo/gemhome/specifications/bundler-2.5.0.dev.gemspec"> was expected to not exist
/home/runner/work/rubygems/rubygems/test/rubygems/test_gem_commands_setup_command.rb:210:in `test_install_default_bundler_gem'
     207:     assert_path_exist File.join(Gem.dir, "specifications", "bundler-audit-1.0.0.gemspec")
     208: 
     209:     # expect to remove normal gem that was same version. because it's promoted default gems.
  => 210:     assert_path_not_exist File.join(Gem.dir, "specifications", "bundler-#{bundler_version}.gemspec")
     211: 
     212:     assert_path_exist "#{Gem.dir}/gems/bundler-#{bundler_version}"
     213:     assert_path_exist "#{Gem.dir}/gems/bundler-1.15.4"
===============================================================================
..F
===============================================================================
Failure: test_install_default_bundler_gem_with_destdir_flag(TestGemCommandsSetupCommand): <"/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230413-1951-karu5u/foo/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230413-1951-karu5u/defaultgemhome/specifications/bundler-2.5.0.dev.gemspec"> was expected to exist
/home/runner/work/rubygems/rubygems/test/rubygems/test_gem_commands_setup_command.rb:279:in `test_install_default_bundler_gem_with_destdir_flag'
     276: 
     277:     # installs the updated bundler gemspec to destdir
     278:     assert_path_not_exist prepend_destdir(destdir, previous_bundler_specification_path)
  => 279:     assert_path_exist prepend_destdir(destdir, new_bundler_specification_path)
     280: 
     281:     bundler_spec.executables.each do |e|
     282:       assert_path_exist prepend_destdir(destdir, File.join(@gemhome, "gems", bundler_spec.full_name, bundler_spec.bindir, e))
===============================================================================
~~~

I have already spend quite some time trying to address these, but the main struggle is that I am not clear what the behavior should really be (but also the implementation is quite quirky). Thinking about this, the best option IMHO would probably be if installation of new version of default gem touched only the "default" directory, keeping all the rest as it is. Would this be acceptable?

voxik:
And not sure what are the `test_realworld_*` test failures :thinking:  I hoped that is just my environment. I'll need to investigate

voxik:
> And not sure what are the `test_realworld_*` test failures thinking I hoped that is just my environment. I'll need to investigate

I have addressed these two, but I cannot say I am in love with those test cases. I don't think that relying on the system setup is fine. As this demonstrates, it also makes refactoring harder.

hsbt:
I decided to reject this. We will not make that structure to complicated and complex.

PR #2911, opened 2019-09-11
Fix Kernel#warn override

jeremyevans:
# Description:

This makes the Kernel#warn override behave the same as the default
Kernel#warn.  Previously, behavior could be different, because an
empty hash or empty keyword arguments would be swallowed by the
override, and the second to last argument could be treated as the
keyword argument by the default Kernel#warn.

For example:

    warn({x:1}, {y:2}, {})

    # C version outputs:
    {:x=>1}
    {:y=>2}

    # Rubygems version raises:
    unknown keyword: y (ArgumentError)

By using a keyword splat, and passing the keyword splat to the
default version, we can make behavior the same.

This issue was originally reported on the Ruby bug tracker:
https://bugs.ruby-lang.org/issues/16157#note-8

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -14,7 +14,11 @@ if RUBY_VERSION >= "2.5"
 
     module_function define_method(:warn) {|*messages, **kw|
       unless uplevel = kw[:uplevel]
-        return original_warn.call(*messages, **kw)
+        if Gem.java_platform?
+          return original_warn.call(*messages)
+        else
+          return original_warn.call(*messages, **kw)
+        end
       end
 
       # Ensure `uplevel` fits a `long`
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Hi @jeremyevans! Thanks for the fix.

I added a second commit that should fix the failing test and a regression test for your fix. What do you think?

deivid-rodriguez:
I rebased this PR, precisely to check how it is doing against ruby-head (since it was passing for me locally against ruby 2.7.0dev (2019-09-09T12:27:40Z master 89c5d5a64e), and our CI no longer depends on TravisCI upgrading its ruby-head installation.

But CI doesn't seem to have been triggered.

@hbst Any ideas?

hsbt:
Umm, We couldn't care the ruby-head version on Travis. If we hope to test the latest head version, We can use https://github.com/ruby/ruby-docker-images with GitHub Actions.

I try to switch it from RVM later.

hsbt:
@bundlerbot r+

Issue #1968, opened 2017-07-05
gem pristine doesn't seem to account for --user-install


mpalmer:
Having somehow managed to get a pile of extensions miscompiled, I'm now being prompted to run `gem pristine <gem> --version <something>` for rather a lot of things.  However, it isn't working so well:

    $ gem pristine unicorn --version 5.3.0
    ERROR:  While executing gem ... (Gem::FilePermissionError)
        You don't have write permissions for the /var/lib/gems/2.3.0 directory.
    $ sudo gem pristine unicorn --version 5.3.0
    ERROR:  While executing gem ... (Gem::Exception)
        Failed to find gems ["unicorn"] = 5.3.0

I install all my gems with `--user-install` by default (my `~/.gemrc` has `install: --user-install`), so it's not surprising that running `gem pristine` as `root` gets confused, however I'm somewhat perturbed that running `gem pristine` as me is erroring out.

If I explicitly set `GEM_HOME=~/.gem` and then run `gem pristine`, everything seems to work, but my understanding was that doing that in general was considered poor form, and that `--user-install` was the recommended solution.

Here are my current environment details:

```
$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 2.5.2
  - RUBY VERSION: 2.3.3 (2016-11-21 patchlevel 222) [x86_64-linux-gnu]
  - INSTALLATION DIRECTORY: /var/lib/gems/2.3.0
  - USER INSTALLATION DIRECTORY: /home/mpalmer/.gem/ruby/2.3.0
  - RUBY EXECUTABLE: /usr/bin/ruby2.3
  - EXECUTABLE DIRECTORY: /usr/local/bin
  - SPEC CACHE DIRECTORY: /home/mpalmer/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-linux
  - GEM PATHS:
     - /var/lib/gems/2.3.0
     - /home/mpalmer/.gem/ruby/2.3.0
     - /usr/lib/x86_64-linux-gnu/rubygems-integration/2.3.0
     - /usr/share/rubygems-integration/2.3.0
     - /usr/share/rubygems-integration/all
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - :benchmark => false
     - :sources => ["https://rubygems.org/"]
[...]

PR #2914, opened 2019-09-14
Fix gem pristine not accounting for user installed gems

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/1968

The problem is gem pristine is always checking for the `INSTALLATION DIRECTORY` instead of the `USER INSTALLATION DIRECTORY` even if a gem is installed using the `--user-install` flag.

If a `INSTALLATION DIRECTORY` is not writable it would cause an error like:

```
ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /var/lib/gems/2.3.0 directory.
```

This check is not needed in the pristine command, writable gem_home directory is already being checked at installation time by `Gem::Installer#pre_install_checks`
______________

### DIFF
--- a/lib/rubygems/commands/pristine_command.rb
+++ b/lib/rubygems/commands/pristine_command.rb
@@ -111,11 +111,6 @@ extensions will be restored.
             "Failed to find gems #{options[:args]} #{options[:version]}"
     end
 
-    install_dir = Gem.dir # TODO use installer option
-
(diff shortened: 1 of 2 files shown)


bronzdoc:
@bundlerbot r+

bronzdoc:
Agree @deivid-rodriguez 

Changes in `test_execute_all` were a mistake, fixed. Thanks!

PR #2915, opened 2019-09-15
Try to use GitHub Actions

hsbt:
# Description:

Added the initial workflow file.
______________

# Tasks:

- [ ] Write tests


MSP-Greg:
@hsbt

In all files:

```
      - name: Setup ruby
        uses: actions/setup-ruby@v1
        with:
          version: ${{ matrix.ruby }}
          architecture: 'x64'
```

should be:
```
      - name: Setup ruby
      uses: actions/setup-ruby@v1
      with:
```

`version:` is deprecated, use `ruby-version:`.  `architecture:` is shown in a code example, but actually isn't defined as a valid input (see https://github.com/actions/setup-ruby/blob/master/action.yml).

Ruby 2.3.x isn't available on Windows, may not be available in ubuntu-18.04.  Can't test till later..

hsbt:
@deivid-rodriguez Can you review this? After that, I will remove appveyor and Travis CI configurations.

MSP-Greg:
JFYI, Actions Windows CI is showing 67 skips, where AppVeyor CI is showing 30.

My memory is vague, but this is either the RG test suite needing a 'no spaces' `GIT` environment setting, or issues with the 'DevKit' path not being enabled in Actions.  If this is accepted (LGTM), I'll fix the issue...

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          ruby-version: ${{ matrix.ruby }}
+      - name: Install Dependencies
+        run: util/ci.sh before_script
+        env:
+          TEST_TOOL: ${{ matrix.test_tool }}
+      - name: Run Lint
+        run: util/ci.sh rubocop
+        env:
+          TEST_TOOL: rubygems
```
deivid-rodriguez: Haven't yet digged into the new Github CI system, but if I undestand correctly, "jobs" would be TravisCI "stages". In our previous configuration, we considered enough to limit the "lint" stage to just the latest ruby, but here we're running it against all matrix entries. So perhaps add a separate "job" the run rubocop against a single ruby?

.github/workflows/ubuntu-rvm.yml
```diff
@@ -0,0 +1,47 @@
...
+        TEST_TOOL: ${{ matrix.test_tool }}
+    - name: Run test
+      run: |
+        source $HOME/.rvm/scripts/rvm
+        util/ci.sh script
+      continue-on-error: true
+      env:
+        TEST_TOOL: ${{ matrix.test_tool }}
+        RGV: ${{ matrix.rgv }}
+        BDV: ${{ matrix.bdv }}
```
deivid-rodriguez: The reason for this separate workflow is that there's currently no way to use `ruby-head` in github actions without using `rvm`. Right?

util/ci.sh
```diff
@@ -12,10 +12,6 @@ case $1 in
   before_script)
     if [ "$TEST_TOOL" = "rubygems" ]
     then
-      gemhome=$(gem env home)
-
-      ruby -I lib bin/gem uninstall executable-hooks gem-wrappers bundler-unload -x --force -i "$gemhome@global"
-
```
deivid-rodriguez: Should we move this to the rvm workflow, or is it not necessary at all? I tend to prefer to actually remove it, because a ruby installation with these tweaks is actually what rvm users get, so it feels best to test against more real world installations :+1:.

util/ci.sh
```diff
@@ -42,7 +38,7 @@ case $1 in
       exec rake test
     else
       cd bundler
-      exec rake spec:travis -t
+      exec rake spec -t
```
deivid-rodriguez: This runs only the standard bundler tests, without "sudo" tests and "realworld" tests. Is this to make what's run by rubygems "lighter", or is there any problem with those tests? In case we just want to make this run faster, maybe is better to run only "realworld" tests? It would definitely be way faster, but I'm not sure it's thorough enough.

Rakefile
```diff
@@ -10,7 +10,7 @@ task :setup => ["bundler:checkout"] do
   gemspec = Gem::Specification.load(File.expand_path("../rubygems-update.gemspec", __FILE__))
 
   gemspec.dependencies.each do |dep|
-    sh "gem install '#{dep.name}:#{dep.requirement.to_s}' --conservative --no-document"
+    sh "gem install '#{dep.name}:#{dep.requirement.to_s}' --conservative --no-document --force"
```
deivid-rodriguez: Not sure what's the reason for this, but I needed this when trying out the [`ronn-ng`](https://github.com/apjanke/ronn-ng) gem. Since I already had the [`ronn`](https://github.com/rtomayko/ronn) gem installed, without `--force` the `ronn`  executable (provided by both gems) wouldn't be updated. So, :+1:.

.github/workflows/ubuntu.yml
```diff
@@ -0,0 +1,42 @@
+name: ubuntu
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: [ '2.3.x', '2.4.x', '2.5.x', '2.6.x' ]
```
deivid-rodriguez: Is there a way to also migrate our jruby matrix entry?

bronzdoc:
This is great, thanks @hsbt 👍

.github/workflows/ubuntu-rvm.yml
MSP-Greg: At present, ruby-head & JRuby are not 'natively' available on any Actions platform/OS.

.github/workflows/ubuntu.yml
hsbt: GitHub Actions don't have the `stage` feature of Travis CI yet. If we want to separate the lint stage from the multiple ruby versions, we can add yaml file for only lint stage.

deivid-rodriguez: So you can't add another top level job? If that's the case, a separate file is fine, or if you believe running rubocop for every ruby version is not too much, then we can leave it as it is.

util/ci.sh
hsbt: We should add sudo and realworld to GitHub Actions. But I'm not sure how to handle them on Actions yet.

.github/workflows/ubuntu.yml
hsbt: I will add jruby to ubuntu-rvm matrix.

hsbt:
@bundlerbot r+

PR #2922, opened 2019-09-20
Detect libc version, closes #2918

### LINKED ISSUE #2918: Gem.platforms doesn't include musl
Hi! I was building a binary gems repository for my project (with [gem-compiler](https://github.com/luislavena/gem-compiler)), which is deployed to Alpine Linux, and while installing them I noticed they would never pull the platform-specific gems.  So I went the rabbit hole and found out that `RbConfig::CONFIG['arch']` would correctly return `x86_64-linux-musl` but it would not be included into `Gem.platforms`...

Finally I found out that applying the following patch would find the `-musl` prefix and add it to `Gem.platforms`.  It seems that Alpine packagers have found out this because otherwise Rubygems would pull x86_64-linux gems built for glibc so they patched it to only contain `['ruby']`.  That's why I was always getting the source gem instead.

I tested on my system and `RbConfig::CONFIG['arch']` returns `x86_64-linux` while in Alpine it returns `x86_64-linux-musl` so I think it wouldn't break anything, though I'm not sure about other Linuxes.

```diff
--- platform.rb.orig
+++ platform.rb
@@ -89,7 +89,7 @@
                       when /^dalvik(\d+)?$/ then        [ 'dalvik',    $1  ]
                       when /^dotnet$/ then              [ 'dotnet',    nil ]
                       when /^dotnet([\d.]*)/ then       [ 'dotnet',    $1  ]
-                      when /linux/ then                 [ 'linux',     $1  ]
+                      when /linux-?(\w+)?/ then         [ 'linux',     $1  ]
                       when /mingw32/ then               [ 'mingw32',   nil ]
                       when /(mswin\d+)(\_(\d+))?/ then
                         os, version = $1, $3
```

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

```
$ gem env version
2.7.6.2
```

fauno:
# Description:

Detect libc version on Linux, makes it possible to distinguish between distributions compiled with alternative libc

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -88,7 +88,7 @@ class Gem::Platform
                       when /^dalvik(\d+)?$/ then        [ 'dalvik',    $1  ]
                       when /^dotnet$/ then              [ 'dotnet',    nil ]
                       when /^dotnet([\d.]*)/ then       [ 'dotnet',    $1  ]
-                      when /linux-?(\w+)?/ then         [ 'linux',     $1  ]
+                      when /linux-?((?!gnu)\w+)?/ then  [ 'linux',     $1  ]
                       when /mingw32/ then               [ 'mingw32',   nil ]
                       when /(mswin\d+)(\_(\d+))?/ then
                         os, version = $1, $3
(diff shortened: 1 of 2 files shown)


hsbt:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

PR #2924, opened 2019-09-24
Use binstubs provided by the bundler repo

deivid-rodriguez:
# Description:

With this PR I hope to fix the CI.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ branches:
     - /.+-stable$/
 env:
   - "TEST_TOOL=rubygems"
-  - "TEST_TOOL=bundler RGV=master"
+  - "TEST_TOOL=bundler RGV=.."
 rvm:
   - 2.6.4
 jobs:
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This needs this PR in bundler. I tried it against it and it works! :+1:

hsbt:
👍 Thanks! I will update this after merging https://github.com/bundler/bundler/pull/7364

deivid-rodriguez:
@hsbt Turns out this PR is no longer necessary after bundler/bundler#7364. Still it has some interesting changes, let me update what I think we can keep.

hsbt:
ah, I see. Should we close this?

deivid-rodriguez:
I updated the PR with the changes I think we can still include.

deivid-rodriguez:
The first commit in here is good and I created [a separate PR for it](https://github.com/rubygems/rubygems/pull/2926). For the second one we'll need until we vendor the next version of bundler, so I'll rebase this PR once that happens.

deivid-rodriguez:
Closing since these changes were incorporated with #2981.

PR #2925, opened 2019-09-24
Introduce default prerelease requirement

deivid-rodriguez:
# Description:

This PR unifies a bit of logic by introducing a `Gem::Requirement.default_prerelease` analogous to `Gem::Requirement.default`, the difference between them being that the latter satisfies any non-prerelease version while the former allows any version.

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2930, opened 2019-09-26
filter dependency type and name strictly.

hsbt:
# Description:

`Gem::Specification#type` has only `:runtime` and `:development` and Also `Gem::Specification#name` is only one-word. 

We should filter them

Co-authored-by: Yusuke Endoh <mame@ruby-lang.org>
______________

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot r+

PR #2933, opened 2019-09-26
Always pass an encoding option to Zlib::GzipReader.wrap

nobu:
# Description:

The arity of this method has been -1 since the import, so the
option has been passed always, even if `Zlib::GzipReader#initialize`
does not take the option.  Actually it takes the option since 1.9.

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot r+

PR #2929, opened 2019-09-25
Support `--enable-load-relative` inside binstubs 

deivid-rodriguez:
# Description:

~`ruby setup.rb` was broken probably since #2782. Requiring `openssl` fixes it.~

~Closes #2928.~

When ruby has been compiled with the `--enable-load-relative` configuration, ruby core generates binstub files with some bash code on top, making `gem install` not detect them as rubygems binstubs.

Closes #5245.

# Tasks:

- [x] Describe the problem / feature
- [x] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2937, opened 2019-09-30
Gem::Specification#to_ruby needs OpenSSL

nobu:
# Description:

Since b1d825ab3a25a12e2602c13555b22809c93bc12e, lib/rubygems/specification.rb should require openssl.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2423,6 +2423,7 @@ class Gem::Specification < Gem::BasicSpecification
   # still have their default values are omitted.
 
   def to_ruby
+    require 'openssl'
     mark_version
     result = []
     result << "# -*- encoding: utf-8 -*-"


lib/rubygems/specification.rb
```diff
@@ -2423,6 +2423,7 @@ def test_files # :nodoc:
   # still have their default values are omitted.
 
   def to_ruby
+    require 'openssl'
```
bronzdoc: I think we can require it where it's needed.
https://github.com/rubygems/rubygems/blob/0dd5b8856dbecfffdb6be5b192f08804ef5ae3f4/lib/rubygems/specification.rb#L2465

deivid-rodriguez:
This is a duplicate of #2929, but the other PR doesn't have passing tests because of an unrelated issue, so I'm fine merging this as it is. I think the place where the `require` has been put is good. I placed it right were it's used, but I don't think it matters much?

bronzdoc:
@bundlerbot r+

PR #2944, opened 2019-10-06
Time comparison around date boundary

nobu:
# Description:

More failures around date boundary in UTC.

https://gist.github.com/ko1/a88834d744a0682711dab53fd9015a04#file-brlog-trunk-nopara-20191005-235153-L393-L472

The date got at the loading/initialization time may not equal the date when the test methods run, as time advances.
`to_ruby` and `metadata` test should assert that the results contain the corresponding `date` values, not `TODAY` which is not concerned directly.
The assertion about `date` attribute and `TODAY` should have done in `test_date`.

# Tasks:

- [x] Write tests

### DIFF
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1695,8 +1695,9 @@ dependencies: []
 
   def test_date
     today = Gem::Specification::TODAY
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

hsbt:
@bundlerbot r+

PR #2946, opened 2019-10-10
Use docker for ruby master

hsbt:
# Description:

RVM is too heavy for us. I try to use the docker environment in GitHub Actions.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -16,7 +16,7 @@ status = [
   "ubuntu (2.6.x, bundler)",
   "windows (2.6.x)",
   "ubuntu_bundler_master (2.6.x)",
-  "ubuntu_rvm (ruby-head, rubygems)",
+  "ruby_master",
   "ubuntu_rvm (ruby-head, bundler)",
   "ubuntu_rvm (jruby-9.2.8.0, rubygems)",
 ]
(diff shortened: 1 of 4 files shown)


bronzdoc:
@hsbt how RVM is to heavy to us compared to Docker?

deivid-rodriguez:
I agree using docker is better. It doesn't require workarounds like the ones I added in #2957, and we don't need to download RVM, install it, install ruby-head, and enable rvm. With docker with have ruby's master branch ready to use.

deivid-rodriguez:
Rebased and improved :)

deivid-rodriguez:
Rebased and improved :)

deivid-rodriguez:
I'm going to squash and merge this PR now :+1:

deivid-rodriguez:
@bundlerbot merge

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
I found the culprit, I had not removed the github check we're moving away from the RVM workflow, that was the reason why bors was timing out (since the workflow had been removed, the check was never being received).

Fixed that so I'll try again.

@bundlerbot retry

deivid-rodriguez:
Yay! :tada:

PR #2951, opened 2019-10-18
Remove taint usage on Ruby 2.7+

jeremyevans:
# Description:

Ruby 2.7 deprecates taint and it no longer has an effect.
This attempts to leave the behavior the same on older
Ruby versions, but avoid the use of deprecated methods
related to taint on Ruby 2.7+.

See https://bugs.ruby-lang.org/issues/16131 for details.

# Tasks:

- [X] Update tests


hsbt:
@bundlerbot r+

PR #2953, opened 2019-10-21
Check Manifest.txt is up to date

deivid-rodriguez:
# Description:

This PR improves our project checks to make sure that the Manifest.txt file is always kept up to date. And updates it, since the `bundler/UPGRADING.md` file was missing.

# Tasks:

- [x] Write tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@
 /misc
 /notes.txt
 /pkg
-/tmp
+/tmp/*
+!/tmp/.keep
 /coverage
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -23,6 +23,14 @@ class TestProjectSanity < Minitest::Test
     end
   end
 
+  def test_manifest_is_up_to_date
+    skip unless File.exist?(File.expand_path("../../../Rakefile", __FILE__))
+
+    _, status = Open3.capture2e("rake check_manifest")
+
+    assert status.success?, "Expected Manifest.txt to be up to date, but it's not. Run `rake update_manifest` to sync it."
+  end
+
   private
 
   def with_empty_pkg_folder
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -3,9 +3,9 @@
 require "rubygems/test_case"
 require "open3"
 
-class TestRakePackage < Minitest::Test
+class TestProjectSanity < Minitest::Test
 
-  def test_builds_ok
+  def test_rake_package_builds_ok
     skip unless File.exist?(File.expand_path("../../../Rakefile", __FILE__))
 
     with_empty_pkg_folder do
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -10,5 +10,6 @@
 /misc
 /notes.txt
 /pkg
-/tmp
+/tmp/*
+!/tmp/.keep
```
bronzdoc: > !/tmp/.keep

What this do?

test/rubygems/test_project_sanity.rb
```diff
@@ -0,0 +1,58 @@
...
+    _, status = Open3.capture2e("rake check_manifest")
+
+    assert status.success?, "Expected Manifest.txt to be up to date, but it's not. Run `rake update_manifest` to sync it."
+  end
+
+  private
+
+  def with_empty_pkg_folder
+    if File.exist?("pkg")
+      FileUtils.cp_r "pkg", "tmp"
```
bronzdoc: Any specific reason why these kind of methods don't have parenthesis but others do?

.gitignore
deivid-rodriguez: "Unignores" tmp/.keep, which was ignored by the previous line.

test/rubygems/test_project_sanity.rb
deivid-rodriguez: No. Do you want me to add parentheses or something?

djberg96: @deivid-rodriguez I think that's the preferred style now, though the current rubocop.yml doesn't seem to enforce it. But, I will defer to @bronzdoc and the rest of the rubygems community. :)

deivid-rodriguez: Yeah, we talked about it in the past, but the preferred style is not consistent. We prefer no parenthesis for many methods: most methods in `Kernel`, most Minitest assertions, and more. So I don't usually care much about it.

bronzdoc: > We prefer no parenthesis for many methods: most methods in Kernel, most Minitest assertions

I would not put parenthesis in those either since is idiomatic and the lack of a specific receiver make it read more like english. (My personal opinion)

> No. Do you want me to add parentheses or something?

If is not much of a problem for you. Yes, for the ones with a specific receiver.  🙇

deivid-rodriguez: Not at all, and I agree with your opinion. I'll give it another try at enforcing that style through rubocop.

deivid-rodriguez: Thought I had commented here... Anyways, I did a quick try to enforce this through rubocop and I don't think it's worth it since even with a configuration very flexible in terms of exceptions, we still have more than 7K offenses.

Anyways, fixed the specific styling here.

deivid-rodriguez:
Thanks for having a look @bronzdoc!

@bundlerbot r=bronzdoc

PR #1418, opened 2015-12-19
Try symlink on Windows

nobu:
Try symlink regardless platform name.  Now File.symlink is
implemented on Windows.

PR #2961, opened 2019-10-23
Remove symlink support conditionals

deivid-rodriguez:
# Description:

I assumed this code was there to help ruby-core testing rubygems against platforms that do not support symlinks, since the platforms rubygems tests against all support symlinks (according to @nobu's comment in https://github.com/rubygems/rubygems/pull/1418#issuecomment-166056170, symlinks are supported on Windows since ruby 2.3, which is our minimum required ruby).

However, some of the tests were still being skipped only on Windows, and we haven't received any failures reported from ruby-core about this, so I think ruby-core is only testing against platforms that support symlinks.

So, I think we can get rid of all this code.

This bit of code:

https://github.com/nobu/rubygems/blob/558c8fa165d501935b2409ac4d07bb6a57eff13f/lib/rubygems/installer.rb#L525-L527

is uncovered, but it was uncovered before as well for the reasons mentioned below.

# Tasks:

- [ ] Write tests


nobu:
Symlink on Windows is **not always** available.
For non-administrator, it needs an additional privilege setting which is disable by the default.

djberg96:
Unfortunately, @nobu is correct. You either have to run with admin privileges and/or have the right to create symlinks in your local security policy.

I'm not entirely sure why Microsoft did that, but there you go.

deivid-rodriguez:
I see. So it's not needed in our CI because it's probably run with admin privileges but it's needed for developers running under Windows (that have not enabled this setting).

I'm going to close this then, thanks for the information :+1:.

djberg96:
It might be worth adding some comments there explaining why we still need the check. It is possible to check for proper privileges programatically, but that would require some special win32api or ffi wrappers.

deivid-rodriguez:
I'll add some comments on a separate PR :+1:.

PR #2964, opened 2019-10-24
Update Actions windows.yml - GIT, DevKit, cmd -> ps

MSP-Greg:
# Description:

1. GitHub Actions changed the Windows default script engine on Actions from cmd to ps. Updated.

2. I use `git` commands daily in Ruby code.  But, the test suite has always required a git path without a space.  I think this was previously in the Appveyor script.  This, along with adding DevKit, lowers the skip count from 67 to 30.

Currently, no updates to DevKit/MSYS2 are done.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -16,7 +16,13 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - name: Install Dependencies
-        run: gem --% install "rake:>=10.4.2" "minitest:>=5.4.3" --conservative --no-document
+        # updated to match gems in oldest Actions Ruby version, currently 2.4
+        run: gem --% install "rake:>=12.0.0" "minitest:>=5.10.1" --conservative --no-document
       - name: Run Test
-        run: rake test
+        # create a symlink for git to a folder without a space, needed for CI only
+        run: |
+          New-Item -Path "C:\git" -ItemType Junction -Value "C:\Program Files\Git" 1> $null
+          $env:GIT = "C:\git\cmd\git.exe"
+          ridk enable
+          rake test
     timeout-minutes: 15


.github/workflows/windows.yml
```diff
@@ -15,8 +15,13 @@ jobs:
         uses: actions/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+      - name: DevKit, ENV['PATH']
+        uses: MSP-Greg/msys2-action@master
```
bronzdoc: What is this?

MSP-Greg:
All CI has a concept of 'steps'.  In GitHub Actions (GHA), steps are isolated from other steps in ways that are different from other CI systems.  One way is ENV.  Also, some operations may involve a lot of code.  GHA allows 'packages' to be used for difficult CI Operations.  One example would be:

https://github.com/ruby/ruby/blob/master/.github/workflows/macos.yml#L58.

Some RG tests require an available set of build tools, and also a few *nix CLI commands.  These are all available thru the MSYS2 build tools, but they are not available by default in GHA.  Also, the MSYS2 build system is frequently updated, and and GHA hasn't even committed to updating their images with current Ruby versions.  Same for other software.  There are also multiple copies of tools available from the current ENV['path'] settings.

So, the action in question can be viewed at:

https://github.com/MSP-Greg/msys2-action

It's a general purpose tool for performing operations related to MSYS2, and using MSYS2 with Ruby.  Its main purpose is setting up MSYS2/Ruby when building/testing extension gems.

It, and the adjustments for `GIT` are the reason 37 tests are no longer skipped on MinGW CI.  JFYI, I've worked with the RG test suite in ruby/ruby...

.github/workflows/windows.yml
MSP-Greg: @bronzdoc

Given that the requirements here are minor, removed the custom action and replaced with `ridk enable` in the test script.

bronzdoc:
I think this just need a rebase, sorry for the trouble @MSP-Greg and thanks!

MSP-Greg:
The vast majority of PR's I test in my fork first.  As to the gem versions changes, they are the versions installed on Windows Ruby 2.4.x.  See https://github.com/MSP-Greg/github-actions-ruby-info/actions and inspect the latest test.

Somewhat off topic, but regarding the action mentioned above.  Ruby 2.4.6 is the current version installed on Actions Windows.  It was compiled with gcc 8.3.0.  The DevKit included with it has gcc 7.3.0. The current MSYS2 gcc compiler is 9.2.0, which I think the Windows Installer for Ruby will prompt one to update.  That's why I created the Action.

I'll rebase...

bronzdoc:
@bundlerbot r+

MSP-Greg:
@duckinator

Just thought I'd ask, do you work with Windows?  Maybe not necessarily your main platform...

duckinator:
@MSP-Greg not consistently, but yeah. why do you ask?

MSP-Greg:
@duckinator In case I'm having a "what am I missing?" moment that might be Windows specific.  Or, "I don't know about this, can you test/repo it?".   Working with ruby-head, there were a few times where something worked locally, but did not work in CI...  Anyway, thanks.

bronzdoc:
@bundlerbot retry

MSP-Greg:
@bronzdoc

This seems frozen?  I can resubmit...

bronzdoc:
@MSP-Greg please. Thank you!

MSP-Greg:
Closing so I can rebase & open new

PR #2966, opened 2019-10-26
Update command line parsing to work under ps

deivid-rodriguez:
# Description:

This is a minimal set of changes extracted from #2964 to fix our Windows CI. We can merge the other changes separately since they don't require inmediate introduccion to get the CI green.


bronzdoc:
@bundlerbot r+

PR #2967, opened 2019-10-26
Set deprecation warning on query command

bronzdoc:
# Description:
We have been discouraging user for the use of the query command for a long time.
```
Summary:
    Query gem information in local or remote repositories

  Description:
    The query command is the basis for the list and search commands.

    You should really use the list and search commands instead.  This command
    is too hard to use.
```

 It's time to deprecate it.

This will setup a deprecation warning whenever is used.

______________

### DIFF
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -4,8 +4,12 @@ require 'rubygems/local_remote_options'
 require 'rubygems/spec_fetcher'
 require 'rubygems/version_option'
 require 'rubygems/text'
+require 'rubygems/deprecate'
 
 class Gem::Commands::QueryCommand < Gem::Command
+
+  extend Gem::Deprecate
+  deprecate_command(2019, 12)
 
   include Gem::Text
   include Gem::LocalRemoteOptions
--- a/test/rubygems/test_gem_commands_query_command.rb
+++ b/test/rubygems/test_gem_commands_query_command.rb
@@ -844,6 +844,10 @@ othergem (1.2.3)
     assert_equal expected, @stub_ui.output
   end
 
+  def test_depprecated
+    assert @cmd.deprecated?
+  end
+
   private
 
   def add_gems_to_fetcher


bronzdoc:
@bundlerbot try

deivid-rodriguez:
@bundlerbot try

bronzdoc:
@deivid-rodriguez I think I need to rebase this

deivid-rodriguez:
`@bundlerbot try` should test the PR against latest master, so in principle I don't think it needs rebasing?

Not sure if you planned to add some tests for the deprecation message.

bronzdoc:
Your are right @deivid-rodriguez I forgot 🙄 


> Not sure if you planned to add some tests for the deprecation message.
Was not on my plans 😬 but I could add one :)

deivid-rodriguez:
Up to you, if it's not too hard I think we can add them.

bronzdoc:
@deivid-rodriguez since `command_manager` is the responsible for printing the command deprecation warning I added a test [here](https://github.com/rubygems/rubygems/blob/3d14cd2b26a9e343e5b375ceb88e87ca48cea6af/test/rubygems/test_gem_command_manager.rb#L271
) when I introduced the `deprecate_command` method.

Said that, I understand what you mean. Maybe we want each command to handle it's own deprecation warning?

> Something I would like that we improved in the future is to be able to give the user a more specific message.

That would be nice 👍

deivid-rodriguez:
I see.

> Said that, I understand what you mean. Maybe we want each command to handle it's own deprecation warning?

If this is easy to do and leads to better testability and similarly complex code, I'm all for it. That said, I'm happy with this PR being merged as is.

bronzdoc:
> If this is easy to do and leads to better testability and similarly complex code, I'm all for it. That said, I'm happy with this PR being merged as is.

@deivid-rodriguez I'll look into this, thanks looking at this

bronzdoc:
@bundlerbot r+

PR #2935, opened 2019-09-28
Add deprecate command method

bronzdoc:
# Description:

This PR adds a `deprecate_command` method to deprecate RubyGems commands.

### Example
```ruby            
class Gem::Commands::QueryCommand
  extend Gem::Deprecate

  deprecate_command(2019, 12)

  def execute
    gem_names = Array(options[:name])

    if !args.empty?
      gem_names = options[:exact] ? args.map{|arg| /\A#{Regexp.escape(arg)}\Z/ } : args.map{|arg| /#{arg}/i }
    end

    terminate_interaction(check_installed_gems(gem_names)) if check_installed_gems?

    gem_names.each { |n| show_gems(n) }
  end
 ...
end
```

```shell
$ gem query
ebsocket-driver (0.7.1, 0.7.0, 0.6.5)
websocket-extensions (0.1.4, 0.1.3)
will_paginate (3.0.12)
xml-simple (1.1.5)
xmlrpc (0.3.0)
xpath (3.2.0, 2.1.0)
yajl-ruby (1.4.1)
yard (0.9.20, 0.8.7.6)
yell (2.2.0)
zeitwerk (2.1.10, 2.1.9, 2.1.8, 2.1.6, 2.1.2)
zlib (default: 1.0.0)

NOTE: query command is deprecated. It will be removed on or after 2019-12-01.
```

______________

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #2974, opened 2019-10-31
Simplify deprecate command method

bronzdoc:
# Description:
 
Simplify `Gem::Deprecate#deprecate_command` that was introduced here https://github.com/rubygems/rubygems/pull/2935
______________

### DIFF
--- a/test/rubygems/test_deprecate.rb
+++ b/test/rubygems/test_deprecate.rb
@@ -95,13 +95,13 @@ class TestDeprecate < Gem::TestCase
   end
 
   def test_deprecate_command
+    foo = FooCommand.new
+
     out, err = capture_io do
-      foo = FooCommand.new
       foo.execute
     end
 
-    assert_equal "pew pew!\n", out
-    assert_match(/NOTE: foo command is deprecated. It will be removed on or after 2099-04-01.\n/, err)
+    assert foo.deprecated?
   end
 
 end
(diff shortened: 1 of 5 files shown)


test/rubygems/test_deprecate.rb
```diff
@@ -95,13 +95,13 @@ def execute
...
   def test_deprecate_command
+    foo = FooCommand.new
+
     out, err = capture_io do
-      foo = FooCommand.new
       foo.execute
     end
 
-    assert_equal "pew pew!\n", out
-    assert_match(/NOTE: foo command is deprecated. It will be removed on or after 2099-04-01.\n/, err)
```
deivid-rodriguez: Can we keep this assertion, or add a separate test to check it? I'd like the following line to be covered:

https://github.com/rubygems/rubygems/blob/3235920c22a416c95a27f5332e0b39cf72bd6532/lib/rubygems/command_manager.rb#L179

duckinator:
@bronzdoc woah, this is a _great_ improvement, nicely done! After you add the test @deivid-rodriguez mentioned, I think this is good to go!

test/rubygems/test_deprecate.rb
bronzdoc: Done @deivid-rodriguez

deivid-rodriguez:
@bundlerbot r+

PR #2982, opened 2019-11-08
Remove unnecessary executable flags

deivid-rodriguez:
# Description:

These files had the executable flag set, but they're not meant to be run
directly.

# Tasks:

- [ ] Write tests


duckinator:
@deivid-rodriguez good catch! Looks like it's been that way for over 13 years lmao

bors r+

duckinator:
err, let's try that again.

@bundlerbot r+

deivid-rodriguez:
Thanks @duckinator! Merit goes to [fd](https://github.com/sharkdp/fd) and the clear syntax highlighting of its results :)

PR #2992, opened 2019-11-19
Deprecate `gem generate_index --modern` and `gem generate_index --no-modern`

deivid-rodriguez:
# Description:

In https://github.com/rubygems/rubygems/pull/2607 we added an API to deprecate option switches to rubygems commands. The original motivation was to prevent abrupt removals like we did for `--no-ri` and `--no-rdoc`, but since the removal of those options was not reverted, we never started using `deprecate_option`.

This PR officially deprecates `gem generate_index --modern` and `gem generate_index --no-modern`, that have been "noops" for a long time, since rubygems generates only "modern" indexes.

While implementing this, I noticed that the previous definition of `deprecate_option` was overly complicated, and also that it could use the possibility of an extra message to further explain each deprecation. So I removed the unnecessary options, and added the extra feature.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -397,9 +397,9 @@ class Gem::Command
         version_to_expire = @deprecated_options[command][option]["rg_version_to_expire"]
 
         deprecate_option_msg = if version_to_expire
-                                 "The \"#{option}\" option has been deprecated and will be removed in Rubygems #{version_to_expire}, its use is discouraged."
+                                 "The \"#{option}\" option has been deprecated and will be removed in Rubygems #{version_to_expire}."
                                else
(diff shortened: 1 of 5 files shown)


bronzdoc:
> I noticed that the previous definition of `deprecate_command` was overly complicated, and also that it could use the possibility of an extra message to further explain each deprecation.

You mean `deprecate_option`right ?

deivid-rodriguez:
Yes, edited!

bronzdoc:
@bundlerbot r+

deivid-rodriguez:
@bundlerbot r=bronzdoc

No idea why those tests failed... :grimacing:

deivid-rodriguez:
@bundlerbot retry

deivid-rodriguez:
@bundlerbot retry

PR #3001, opened 2019-11-27
Lazily load `open3`

deivid-rodriguez:
# Description:

Since open3 will be gemified in ruby 2.7, it's better to load it as late as possible to avoid version conflicts with user chosen versions.

This should get us closer to a green CI against ruby-head.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Self-merging to try get ruby-head build working as soon as possible.

@bundlerbot merge

PR #3002, opened 2019-11-27
Remove `delegate` dependency

deivid-rodriguez:
# Description:

The `delegate` library will be gemified in ruby 2.7, so we should try to remove our dependency on it because otherwise we can run into gem activation conflicts with user chosen versions.

We were only depending on it in one place, but our `Kernel.warn` modification actually depended on this dependency because the `delegate` library [performs some customizations on `Kernel` methods](https://github.com/ruby/ruby/blob/cdcaf041127430b30ee4b32452762666495a732a/lib/delegate.rb#L40-L53) and that was accidentally making our customizations work. 

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -12,7 +12,13 @@ if RUBY_VERSION >= "2.5"
 
     original_warn = method(:warn)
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks for having a look and milestone'ing this!

@bundlerbot r=bronzdoc

PR #3005, opened 2019-12-01
Lazily load `uri`

deivid-rodriguez:
# Description:

The `uri` library will be a default gem on ruby 2.7. Because of this, requiring `uri` at the top of this file will cause the default `uri` gem to be activated during `bundler/setup`, and that will result in gem activation conflicts.

So, it's better to lazily load `uri` here.

I'm using this patch in bundler/bundler#7460 to get bundler specs passing against ruby 2.7.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -1,4 +1,3 @@
-require 'uri'
 require 'rubygems/user_interaction'
 
 class Gem::SpecificationPolicy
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Merging this quickly to be able to move bundler/bundler#7460 forward and get full ruby 2.7 support ready as soon as possible.

@bundlerbot merge

hsbt:
:+1:

PR #3012, opened 2019-12-02
Use bundler to manage development dependencies

deivid-rodriguez:
# Description:

This is a experiment to start using `bundler` to manage our development dependencies.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,8 @@
+source "https://rubygems.org"
+
+gem "builder", "~> 3.0"
+gem "rdoc", "~> 6.0"
+gem "rake", "~> 12.0"
+gem "minitest", "~> 5.0"
+gem "simplecov", "~> 0"
+gem "rubocop", "~> 0.74.0"
--- a/Rakefile
+++ b/Rakefile
@@ -7,11 +7,8 @@ require 'psych'
 
 desc "Setup Rubygems dev environment"
 task :setup => ["bundler:checkout"] do
-  gemspec = Gem::Specification.load(File.expand_path("../rubygems-update.gemspec", __FILE__))
-
-  gemspec.dependencies.each do |dep|
-    sh "gem install '#{dep.name}:#{dep.requirement.to_s}' --conservative --no-document --force"
-  end
+  sh "gem install bundler:2.0.2"
+  sh "bundle install"
 end
 
 desc "Setup git hooks"
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,0 +1,2 @@
+---
+BUNDLE_PATH__SYSTEM: "true"
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -22,7 +22,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - name: Install Dependencies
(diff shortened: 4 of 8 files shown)


hsbt:
@deivid-rodriguez Does this ready to merge? I prefer to use `Gemfile` instead of `Gem::Specification#add_development_dependency`.

deivid-rodriguez:
It should be ready now, yeah!

CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ here: http://guides.rubygems.org/contributing/
 
 ## Getting Started
 
-    $ rake setup
+    $ bundle install
     $ rake test
```
bronzdoc: Should we change this instruction to `bundle exec rake test` ?

deivid-rodriguez: No, since I don't think that's going to work... We want tests to run on a fully unbundled environment. I now realize this is confusing so I should at least add a note about it.

Rakefile
```diff
@@ -409,11 +400,6 @@ task :check_manifest do
 end
 
 namespace :bundler do
-  desc "Initialize bundler submodule"
-  task :checkout do
-    sh "git submodule update --init"
```
bronzdoc: This now would have to be documented and explicitly called whenever a new developer would want to contribute to Rubygems?

Rakefile
```diff
@@ -5,15 +5,6 @@ require 'rubygems/package_task'
 require "rake/testtask"
 require 'psych'
 
-desc "Setup Rubygems dev environment"
-task :setup => ["bundler:checkout"] do
```
bronzdoc: Could this just call `bundle install` instead of deleting it?

Gemfile
```diff
@@ -0,0 +1,8 @@
+source "https://rubygems.org"
+
```
bronzdoc: what if we just `gempsec` here?

CONTRIBUTING.md
bronzdoc: What is the benefit of using bundler if we want to run things in a unbundled environment?

deivid-rodriguez: My original use case is that I wanted to vendor [the `uri` gem](https://github.com/ruby/uri) inside `rubygems`, so I wanted to include [the `automatiek` gem](https://github.com/segiddins/automatiek) to help me with that. But, I needed an unreleased version of automatiek, so I wanted to do something like `gem "automatiek", git: "https://github.com/deivid-rodriguez/automatiek", branch: "my_additional_unreleased_changes"`. Using a gemspec doesn't allow that.

Rakefile
deivid-rodriguez: Ooops, I thought `git clone` would initialize the submodule by default, but it doesn't. Let me restore this.

Rakefile
deivid-rodriguez: Sure, that way we can keep the `bundler:checkout` prerequisite.

Gemfile
deivid-rodriguez: Since `rubygems-update` does not have any runtime dependencies, I don't think it's necesary?

bronzdoc:
@bundlerbot r+

PR #3017, opened 2019-12-04
Refactor remote fetcher

deivid-rodriguez:
# Description:

In bundler/bundler#7460 I'm vendoring the `uri` library inside `bundler` to add support for using the `uri` version inside Gemfile's just like any other gem. However, that was not the only thing needed since `bundler/setup` also touches some code from this class in rubygems.

So the initial approach I took was to duplicate all the code inside bundler, replacing all reference to `URI` with `Bundler::URI` (the vendored namespace). Like in https://github.com/bundler/bundler/pull/7460/commits/3bcc57980b27fc98a9cb44c7853261c206fdc109.

However, this is ugly and hard to maintain so in this PR I'm refactoring this class so that it is enough to override only one method in the `bundler` subclass, like this: https://github.com/bundler/bundler/blob/aa3a25d30383a16e002dd94c209a6078746204a7/lib/bundler/gem_remote_fetcher.rb 

# Tasks:

- [ ] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -493,6 +493,7 @@ lib/rubygems/test_utilities.rb
 lib/rubygems/text.rb
 lib/rubygems/uninstaller.rb
 lib/rubygems/uri_formatter.rb
+lib/rubygems/uri_parser.rb
 lib/rubygems/user_interaction.rb
 lib/rubygems/util.rb
 lib/rubygems/util/licenses.rb
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -494,6 +494,7 @@ lib/rubygems/text.rb
 lib/rubygems/uninstaller.rb
 lib/rubygems/uri_formatter.rb
 lib/rubygems/uri_parser.rb
+lib/rubygems/uri_parsing.rb
 lib/rubygems/user_interaction.rb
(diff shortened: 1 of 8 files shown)


bronzdoc:
@bundlerbot r+

deivid-rodriguez:
I'm actually doing some more testing on the bundler side, and it seems that this refactoring is not actually necessary... :man_shrugging:

hsbt:
@bundlerbot merge

deivid-rodriguez:
As I mentioned in the previous comment, this PR wasn't necessary in the end, but it doesn't hurt either I think, so thanks for reviewing and merging :)

bronzdoc:
@hsbt please don't merge PR's tagged as `wip` 😬

deivid-rodriguez:
Right, I marked it as WIP because I wanted to tweak something else before setting it for review. Let me update what I wanted to improve on a separate PR.

deivid-rodriguez:
Forget about it, I think it's fine as it is.

hsbt:
ah sorry. I believe we start to use [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) instead of wip label of wip title. It's great to see wip status for everyone.

PR #3020, opened 2019-12-06
Remove regular 2.3 testing

deivid-rodriguez:
# Description:

From the failures at #3017, it seems that `actions/setup-ruby` dropped support for testing against ruby 2.3.

So, I'm removing 2.3.x from our matrix. To ensure support until we actually drop it, I replaced it with an entry in the `rvm` file.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,7 +1,5 @@
 status = [
   "ubuntu_lint",
-  "ubuntu (2.3.x, rubygems)",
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@bundlerbot r=hsbt

deivid-rodriguez:
@bundlerbot retry

PR #3024, opened 2019-12-13
Remove unnecessary CI step

deivid-rodriguez:
# Description:

Currently, the `apt install` step in our CI causes some stability some times. However, only a single test relies on the `zip` package being present. That test was added on https://github.com/rubygems/rubygems/pull/2787 and essentially checked that the `Manifest.txt` file was in sync. However, in https://github.com/rubygems/rubygems/pull/2953 a more robust check for that was added, so I think we can remove the original test and thus remove the `apt install` step too.

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -1,27 +1,8 @@
 # frozen_string_literal: true
 
-require "rubygems/test_case"
 require "open3"
 
 class TestProjectSanity < Minitest::Test
-
-  def test_rake_package_builds_ok
-    skip unless File.exist?(File.expand_path("../../../Rakefile", __FILE__))
-
-    with_empty_pkg_folder do
-      output, status = Open3.capture2e("rake package")
-
-      assert_equal true, status.success?, <<~MSG.chomp
-        Expected `rake package` to work, but got errors:
-
-        ```
-        #{output}
-        ```
-
-        If you have added or removed files, make sure you run `rake update_manifest` to update the `Manifest.txt` accordingly
-      MSG
-    end
-  end
 
   def test_manifest_is_up_to_date
     skip unless File.exist?(File.expand_path("../../../Rakefile", __FILE__))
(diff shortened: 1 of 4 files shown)


util/ci.sh
```diff
@@ -16,6 +16,9 @@ case $1 in
       ruby -I lib bin/gem uninstall executable-hooks gem-wrappers bundler-unload -x --force -i "$gemhome@global"
     fi
 
+    # --force flag is a workaround for Windows executable conflict, root cause should be fixed
+    gem install rake --no-document --force
```
bronzdoc: I don't understand this. If is only for windows why will be installed for everything  and if the root cause "should be fixed" why is needed?

test/rubygems/test_project_sanity.rb
```diff
@@ -1,28 +1,9 @@
 # frozen_string_literal: true
 
-require "rubygems/test_case"
 require "open3"
 
 class TestProjectSanity < Minitest::Test
 
-  def test_rake_package_builds_ok
```
bronzdoc: Isn't this the `more robust test` that you point to in your issue comment?

util/ci.sh
deivid-rodriguez: I was writing an explanation but I decided to remove the commit instead since my goal here is to fix CI, and it was not necessary.

test/rubygems/test_project_sanity.rb
deivid-rodriguez: No, please double check the references. The test I'm referring to `test_manifest_is_up_to_date`.

The "more robust test" is not a very accurate expression though, I meant something like the "more focused test" better. When I added `test_rake_package_builds_ok` the idea was to check that the manifest was up to date and `test_manifest_is_up_to_date` does a better job at that.

bronzdoc:
@bundlerbot merge

deivid-rodriguez:
@bundlerbot retry

Issue #2686, opened 2019-03-18
Missing Wrapper


drcapulet:
### Behavior

When Bundler is installed via a `gem update --system`, the wrapper files normally generated by `gem install` are missing. Notably the `bundle` command is the `exe/bundle` file from the Bundler gem.

### Reproduction

Start image:

```
$ docker run -it --rm ruby:2.5 bash
```

```
root@911c503c3cc4:/# which bundle
/usr/local/bin/bundle
root@911c503c3cc4:/# bundle _1.17.3_ version
Could not find command "_1.17.3_".
root@911c503c3cc4:/# gem install bundler --version 2.0.1
Fetching bundler-2.0.1.gem
Successfully installed bundler-2.0.1
1 gem installed
root@911c503c3cc4:/# hash -r
root@911c503c3cc4:/# which bundle
/usr/local/bundle/bin/bundle
root@911c503c3cc4:/# bundle _1.17.3_ version
Bundler version 1.17.3 (2019-03-04 commit d7089abb6)
root@911c503c3cc4:/#
```

Note that the `hash -r` is required otherwise this will happen:

```
root@e86383a3c9fa:/# which bundle
/usr/local/bin/bundle
root@e86383a3c9fa:/# type bundle
bundle is /usr/local/bin/bundle
root@e86383a3c9fa:/# bundle _1.17.3_ version
Could not find command "_1.17.3_".
root@e86383a3c9fa:/# type bundle
bundle is hashed (/usr/local/bin/bundle)
root@e86383a3c9fa:/# gem install bundler --version 2.0.1
Fetching bundler-2.0.1.gem
Successfully installed bundler-2.0.1
1 gem installed
root@e86383a3c9fa:/# which bundle
/usr/local/bundle/bin/bundle
root@e86383a3c9fa:/# type bundle
bundle is hashed (/usr/local/bin/bundle)
root@e86383a3c9fa:/# bundle _1.17.3_ version
Could not find command "_1.17.3_".
```

deivid-rodriguez:
@drcapulet Thanks for the quality report! :heart:

I agree this is unexpected and I tried to fix it in #2690.

drcapulet:
@deivid-rodriguez thanks for the quick fix!

PR #3035, opened 2019-12-16
Fix `gem update --system --force`

### LINKED ISSUE #3030: The `--force` flag is ignored when installing the bundled bundler using `update --system`
I'm having a problem.

My current problem is installing rubygems to environment which already installed bundler .

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

docker image `ruby:2.5`.

```
$ docker run -it --rm ruby:2.5 gem env version
3.0.3
```

The installation pauses by asked whether to overwrite.

```
$ docker run -it --rm ruby:2.5 gem update --system --force
Updating rubygems-update
Fetching rubygems-update-3.1.0.gem
Successfully installed rubygems-update-3.1.0
Installing RubyGems 3.1.0
  Successfully built RubyGem
  Name: bundler
  Version: 2.1.0.pre.3
  File: bundler-2.1.0.pre.3.gem
bundler's executable "bundle" conflicts with /usr/local/bin/bundle
Overwrite the executable? [yN]  

```

deivid-rodriguez:
# Description:

The `--force` option is not implemented by `gem update --system`, but it should because `gem update --system` also installs binstubs (`bundler` and `gem` at the least)

Note that in the case of https://github.com/rubygems/rubygems/issues/3030, `rubygems` is prompting the user even though the `bundler` binstub has actually been generated by `rubygems`. The reason is that the `ruby.2.5` docker image ships with rubygems 3.0.3 which has [this bug](https://github.com/rubygems/rubygems/issues/2686). That means it generates an incorrect bundler binstub, and when you try to update it, user gets prompted because rubygems sees the binstub as "modified".

The `--force` option is useful here, since it's a good way for `rubygems` to be able to "auto-heal" its own installation in cases like this one.

# Tasks:

- [x] Write tests

Closes #3030.

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -17,6 +17,7 @@ class Gem::Commands::SetupCommand < Gem::Command
 
     super 'setup', 'Install RubyGems',
           :format_executable => true, :document => %w[ri],
+          :force => false,
           :site_or_vendor => 'sitelibdir',
           :destdir => '', :prefix => '', :previous_version => '',
           :regenerate_binstubs => true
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
@hsbt If you're ok with it, could you review and release this as 3.1.2?

hsbt:
👍 I confirmed to work `--force` flag in my box too.

@bundlerbot merge

PR #3040, opened 2019-12-18
Restore non prompting `gem update --system` behavior

deivid-rodriguez:
# Description:

After the upgrade to rubygems 3.1, CIs of people who had `gem update --system` in their CI scripts broke, because `gem update --system` would prompt the user for confirmation to overwrite the `bundler` binstub that had apparently been changed. However, these are pristine installations, nobody has edited this file other than rubygems, so this is a bug.

This wrong binstub used to get created by ruby (through `make install`, prior to version 2.6) or by rubygems (through `gem update --system`, prior to version 3.1). Those bugs are now fixed, but fixing them sadly made this issue appear because the correct binstub that the newest rubygems wants to create no longer matches the old one (that might have been generated by previous versions of rubygems or ruby), and rubygems refuses to overwrite it.

The problem with the fix to the wrong binstub is that it changed rubygems behavior, because whereas in rubygems 3.0 bundler binstubs were overwritten manually without confirmation, in the new version they are not, since we use the default gem installer, which uses `--force=false` by default.

So, to fix the issue and preserve rubygems 3.0 behavior, we default to `--force=true` in this particular case.

To protect ourselves from future similar breakage, I added some integration tests to check that the current version of rubygems can always upgrade older installations of all supported rubies.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -17,7 +17,7 @@ class Gem::Commands::SetupCommand < Gem::Command
 
     super 'setup', 'Install RubyGems',
           :format_executable => true, :document => %w[ri],
-          :force => false,
+          :force => true,
           :site_or_vendor => 'sitelibdir',
           :destdir => '', :prefix => '', :previous_version => '',
           :regenerate_binstubs => true
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -1,5 +1,10 @@
 status = [
   "ubuntu_lint",
+  "install (2.3.8)",
+  "install (2.4.9)",
+  "install (2.5.7)",
+  "install (2.6.5)",
+  "install (jruby-9.2.9.0)",
   "macos (2.4.x)",
   "ubuntu (2.4.x, rubygems)",
   "ubuntu (2.4.x, bundler)",
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -1,0 +1,44 @@
+name: install
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - staging
+      - trying
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        ruby: [ '2.3.8', '2.4.9', '2.5.7', '2.6.5', 'jruby-9.2.9.0' ]
+
+    steps:
+      - uses: actions/checkout@v1
+
+      - run: git submodule update -i
+
+      - name: Fix world writable dirs
+        run: |
+          chmod go-w $HOME
+          sudo chmod -R go-w /usr/share
+          sudo chmod -R go-w /opt
+
+      - name: Set up RVM
+        run: curl -sSL https://get.rvm.io | bash
+
+      - name: Set up Ruby
+        run: |
+          source $HOME/.rvm/scripts/rvm
+          rvm install ${{ matrix.ruby }} --binary --autolibs=disable
+          rvm --default use ${{ matrix.ruby }}
+
+      - name: Install rubygems
+        run: |
+          source $HOME/.rvm/scripts/rvm
+          ruby -Ilib -S rake install
+
+    timeout-minutes: 10
--- a/Rakefile
+++ b/Rakefile
@@ -97,7 +97,7 @@ end
 
 desc "Install rubygems to local system"
 task :install => :package do
-  sh "gem install pkg/rubygems-update-#{v}.gem && update_rubygems"
+  sh "ruby -Ilib bin/gem install pkg/rubygems-update-#{v}.gem && update_rubygems"
 end
 
 desc "Release rubygems-#{v}"
--- a/bin/update_rubygems
+++ b/bin/update_rubygems
@@ -32,5 +32,5 @@ else
   update_dir = File.dirname(update_dir)
   Dir.chdir update_dir
   ENV["GEM_PREV_VER"] = Gem::VERSION
-  system(Gem.ruby, 'setup.rb', *ARGV)
+  abort unless system(Gem.ruby, 'setup.rb', *ARGV)
 end


deivid-rodriguez:
Also, while developing this fix I noticed other things:

* As noticed by https://github.com/rubygems/rubygems/issues/3030#issuecomment-567021416, our TTY detection code is not properly detecting that TravisCI is not a TTY. It does work as expected on Github Actions. The overwrite check is prepared to _fail_ without overwriting, but not prompt the user, if it finds an executable conflict.

* However, the rubygems updater swallows `ruby setup.rb` errors, so unless the TTY detection fails and prompts the user, a CI should never will never fail because of this. I fixed this in be4c79030d78e48b25cdb17df8fd9145c5cb93e6, partly because it's a bug in my opinion, but also because I can't get `gem update --system` to fail in our CI otherwise, and I really want to make sure we detect this kind of problems before they hit users in the future.

deivid-rodriguez:
Actually, I think there might be a better fix for this. I'll get to it later, marking it as WIP again.

indirect:
Thanks for working on this! I think having an update process that doesn’t suddenly start to require user input is really important 👍🏻

deivid-rodriguez:
I edited the PR description to describe the final (way simpler) solution.

deivid-rodriguez:
Yeah, I agree it's important, and I was very unhappy when I saw this breaking. So I want to make sure it doesn't happen again.

lib/rubygems/commands/setup_command.rb
```diff
@@ -17,7 +17,7 @@ def initialize
 
     super 'setup', 'Install RubyGems',
           :format_executable => true, :document => %w[ri],
-          :force => false,
+          :force => true,
```
hsbt: I'm still against to set the default as `--force` option. The destructive action should be set by end-user explicity.

Can you add the warning message with target executables? like this:

```
WARNING: rubygems-update will overwrite the following files:
/path/to/bundle
```

deivid-rodriguez: @hsbt I understand your concern but the point of this PR is to not change the previous behavior. We can think of changing the default in the future (4.0?) but for now the main priority IMO is to restore the previous behavior.

See how things work in rubygems 3.0:

```
$ gem --version
3.0.5
$ echo "puts 'EDIT'" >> /home/deivid/.rbenv/versions/2.6.5/bin/bundle
$ cat /home/deivid/.rbenv/versions/2.6.5/bin/bundle
#!/home/deivid/.rbenv/versions/2.6.5/bin/ruby
#
# This file was generated by RubyGems.
#
# The application 'bundler' is installed as part of a gem, and
# this file is here to facilitate running it.
#

require 'rubygems'

version = ">= 0.a"

str = ARGV.first
if str
  str = str.b[/\A_(.*)_\z/, 1]
  if str and Gem::Version.correct?(str)
    version = str
    ARGV.shift
  end
end

if Gem.respond_to?(:activate_bin_path)
load Gem.activate_bin_path('bundler', 'bundle', version)
else
gem "bundler", version
load Gem.bin_path("bundler", "bundle", version)
end
puts 'EDIT'
deivid@chaba ~ $ gem update --system 3.0.6
Updating rubygems-update
Successfully installed rubygems-update-3.0.6
Installing RubyGems 3.0.6
Bundler 1.17.3 installed
RubyGems 3.0.6 installed
Regenerating binstubs

=== 3.0.6 / 2019-08-17

Bug fixes:

* Revert #2813. It broke the compatibility with 3.0.x versions.

=== 3.0.5 / 2019-08-16

Minor enhancements:

* Use env var to configure api key on push. Pull request #2559 by Luis
  Sagastume.
* Unswallow uninstall error. Pull request #2707 by David Rodríguez.
* Expose windows path normalization utility. Pull request #2767 by David
  Rodríguez.
* Clean which command. Pull request #2801 by Luis Sagastume.
* Upgrading S3 source signature to AWS SigV4. Pull request #2807 by
  Alexander Pakulov.
* Remove missleading comment, no reason to move Gem.host to Gem::Util.
  Pull request #2811 by Luis Sagastume.
* Drop support for 'gem env packageversion'. Pull request #2813 by Luis
  Sagastume.
* Take into account just git tracked files in update_manifest rake task.
  Pull request #2816 by Luis Sagastume.
* Remove TODO comment, there's no Gem::Dirs constant. Pull request #2819
  by Luis Sagastume.
* Remove unused 'raise' from test_case. Pull request #2820 by Luis
  Sagastume.
* Move TODO comment to an information comment. Pull request #2821 by Luis
  Sagastume.
* Use File#open instead of Kernel#open in stub_specification.rb. Pull
  request #2834 by Luis Sagastume.
* Make error code a gemcutter_utilities a constant. Pull request #2844 by
  Luis Sagastume.
* Remove FIXME comment related to PathSupport. Pull request #2854 by Luis
  Sagastume.
* Use gsub with Hash. Pull request #2860 by Kazuhiro NISHIYAMA.
* Use the standard RUBY_ENGINE_VERSION instead of JRUBY_VERSION. Pull
  request #2864 by Benoit Daloze.
* Do not mutate uri.query during s3 signature creation. Pull request #2874
  by Alexander Pakulov.
* Fixup #2844. Pull request #2878 by SHIBATA Hiroshi.

Bug fixes:

* Fix intermittent test error on Appveyor & Travis. Pull request #2568 by
  MSP-Greg.
* Extend timeout on assert_self_install_permissions. Pull request #2605 by
  SHIBATA Hiroshi.
* Better folder assertions. Pull request #2644 by David Rodríguez.
* Fix default gem executable installation when folder is not `bin/`. Pull
  request #2649 by David Rodríguez.
* Fix gem uninstall behavior. Pull request #2663 by Luis Sagastume.
* Fix for large values in UID/GID fields in tar archives. Pull request
  #2780 by Alexey Shein.
* Fixed task order for release. Pull request #2792 by SHIBATA Hiroshi.
* Ignore GEMRC variable for test suite. Pull request #2837 by SHIBATA
  Hiroshi.


------------------------------------------------------------------------------

RubyGems installed the following executables:
	/home/deivid/.rbenv/versions/2.6.5/bin/gem
	/home/deivid/.rbenv/versions/2.6.5/bin/bundle

RubyGems system software updated
$ cat /home/deivid/.rbenv/versions/2.6.5/bin/bundle
#!/home/deivid/.rbenv/versions/2.6.5/bin/ruby
#
# This file was generated by RubyGems.
#
# The application 'bundler' is installed as part of a gem, and
# this file is here to facilitate running it.
#

require 'rubygems'

version = ">= 0.a"

str = ARGV.first
if str
  str = str.b[/\A_(.*)_\z/, 1]
  if str and Gem::Version.correct?(str)
    version = str
    ARGV.shift
  end
end

if Gem.respond_to?(:activate_bin_path)
load Gem.activate_bin_path('bundler', 'bundle', version)
else
gem "bundler", version
load Gem.bin_path("bundler", "bundle", version)
end
```

So, rubygems 3.0 forcefully overwrites executables and already informs clearly about the executables it installed.

However, 3.1, even if with this PR also overwrites executables, it no longer informs about the `bundler` executables being written. I think that's an issue and I will fix it as you suggested :+1:.

hsbt: I understand to important to keep the previous behavior. But also the previous behavior is not expected specification.

Ok, I will merge this for your suggestion.

hsbt:
@bundlerbot merge

lib/rubygems/commands/setup_command.rb
deivid-rodriguez: I was fixing the bug you found: Installed `bundle` executable is no longer informed in rubygems 3.1. I will open a separate PR since we are already merging this.

deivid-rodriguez: See https://github.com/rubygems/rubygems/pull/3042.

hsbt: 👍

hsbt:
@bundlerbot retry

Issue #2984, opened 2019-11-08
All `gem` commands print warnings on master+3.1.0pre2, if certain gems are installed


duckinator:
My current problem is that running any `gem` command results in warnings about `Gem::Specification#rubyforge_project=` for _every gem you have installed that uses `rubyforge_project = ...` in their gemspec.

Unfortunately, two such gems are on my system because I use Vagrant for some things (erubis and a 3+ year old version of childprocess).

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details, which conveniently demonstrates the problem:

```
$ gem env version
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /usr/share/gems/specifications/childprocess-0.5.9.gemspec:16.
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /usr/share/gems/specifications/erubis-2.7.0.gemspec:17.
3.1.0.pre2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
@duckinator How is that a bug?

They should remove that from their gemspec. `rubyforge_project=` is deprecated and not encouraged to use isn't it?

bronzdoc:
@duckinator I poked around this and i'm not able to reproduce it.

```
➜  rubygems git:(97e97686) ✗ gem list | grep erubis
erubis (2.7.0)
➜  rubygems git:(97e97686) ✗ gem env version
3.1.0.pre2
➜  rubygems git:(97e97686) ✗
```

Can you share your `gem env`?

duckinator:
<details>
<summary>Output of `gem env`</summary>

```
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /usr/share/gems/specifications/childprocess-0.5.9.gemspec:16.
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
[...]

PR #3047, opened 2019-12-22
Fix rubyforge deprecated warnings

MSP-Greg:
# Description:

`Specification#rubyforge_project=` is deprecated, and issues warning when loading any specs containing it.  Hence, the warning is shown when many gem commands are used.

See https://github.com/rubygems/rubygems/issues/2984

Add a `deprecate_spec` method that only warns when `gem build` is used.

I haven't looked at RG tests for quite a while, so considering this a draft.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -78,6 +78,11 @@ class Gem::CommandManager
   }.freeze
 
   ##
+  # The current command name as a symbol
+
+  attr_reader :current_command
+
+  ##
   # Return the authoritative instance of the command manager.
 
   def self.instance
@@ -175,6 +180,7 @@ class Gem::CommandManager
     else
       cmd_name = args.shift.downcase
       cmd = find_command cmd_name
+      @current_command = cmd.command.to_sym
       cmd.invoke_with_build_args args, build_args
       cmd.deprecation_warning if cmd.deprecated?
     end
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -81,6 +81,31 @@ module Gem::Deprecate
       end
     end
   end
-  module_function :deprecate, :deprecate_command, :skip_during
+
+  ##
+  # Same as `deprecate`, but only warns during BuildCommand
+  # Used for deprecated specification properties
+
+  def deprecate_spec(name, repl, year, month)
+    class_eval do
+      old = "_deprecated_#{name}"
+      alias_method old, name
+      define_method name do |*args, &block|
+        if Gem::CommandManager.instance.current_command == :build and
+          !Gem::Deprecate.skip
+          klass = self.kind_of? Module
+          target = klass ? "#{self}." : "#{self.class}#"
+          msg = [ "NOTE: #{target}#{name} is deprecated",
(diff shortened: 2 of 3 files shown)


simi:
Can't we remove `rubyforge_project` at all in next release?

MSP-Greg:
'Remove' might not be the best word.  For all operations other than gem build, it should be ignored, which is the main reason for this PR.  That needs to remain as long as existing gems still use it.

What happens if you try to build a gem with `rubyforge_project` in the gemspec, I don't know.

It should be ignored, but it could additionally raise an error or show a deprecate warning.  I haven't checked to see if currently it's ignored during build.

MSP-Greg:
See PR https://github.com/rubygems/rubygems/pull/3051 for a much simpler solution.  The concept of a `deprecate_spec` method is something that should be used for deprecated Spec items, but may not be needed for this issue...

deivid-rodriguez:
> What happens if you try to build a gem with rubyforge_project in the gemspec, I don't know.

Isn't this what this PR should be implementing? Are you saying you don't know because of the lack of tests?

MSP-Greg:
Ever written something when you were trying to get out the door?  I got up this morning intending to delete that or edit it.  But I wanted to get https://github.com/rubygems/rubygems/pull/3051 written and submitted.

> Isn't this what this PR should be implementing?

Technically, no.  This PR is affecting only when the warning is shown, it's not affecting the actually 'build'.  Or, doesn't change the build, just the console output during the build.

> Are you saying you don't know

Yesterday I saw the code you wrote that changed how `rubyforge_project=` was handled...

deivid-rodriguez:
> Technically, no. This PR is affecting only when the warning is shown, it's not affecting the actual 'build'. Or, doesn't change the build, just the console output during the build.

Yeah, sure, so what this PR does is show the warning only on `gem build` when such a gemspec is being built, not on every command where such a gemspec is installed. Right?

MSP-Greg:
Exactly.

duckinator:
@MSP-Greg this PR looks really good imo. I'm a bit bothered by the duplicated error message, but I think it's minor enough that I'll defer to you on whether you think it's worth dealing with.

Let me know when you have this in a state you're happy with and have tests and all that, and I'll give it another look. :+1:

madlew:
Fingers crossed for merge and release asap :)

MSP-Greg:
@madlew

Apologies, I've been busy with Ruby/Actions/Windows things, and I think the proper tests for this are messy, and I haven't been in the RG test suite for quite a while.

One should probably add a deprecated property to Gem::Specification and then test that it properly raises errors/warnings in the different commands...

deivid-rodriguez:
I plan to get to this soon. Note though that this issue is already partially alleviated in master since #3108.

itay-grudev:
Guys! This is developer torture. Please get rid of these warnings!

simi:
We're working on that @itay-grudev to get this to new 3.1. release. Stay tuned.

deivid-rodriguez:
Sorry, forgot to update this. This was fixed in https://github.com/rubygems/rubygems/pull/3506 using a similar approach to the one in this PR.

The fix will be shipped with either rubygems 3.1.4, rubygems 3.2.0, or both.

Closing this PR, by the way!

PR #3049, opened 2019-12-23
Add tests to check if Gem.ruby_version works with ruby git master

mame:
# Description:

Now Ruby source code is managed under git, and `ruby -v` was slightly changed.
This PR adds two tests to check if Gem.ruby_version works with the new format of RUBY_VERSION.

https://bugs.ruby-lang.org/issues/14632
https://bugs.ruby-lang.org/issues/15843
______________

# Tasks:

- [x] Write tests


hsbt:
@bundlerbot merge

PR #3051, opened 2019-12-24
Specification - add REMOVED_METHODS, remove rubyforge_project=

MSP-Greg:
# Description:

Currently, `rubyforge_project=` is a deprecated Specification property/method.  Also, if a repo has the property in its *.gemspec file, it will not be added to a built gem's metadata.

But, if an older version of RubyGems is used to install a gem with the property in its metadata, and then a `gem update --system` is performed, almost all gem operations will emit a warning about the deprecated property.

This PR adds an array constant that contains the names of removed Specification properties/methods.   The array is then used in `Specification#method_missing` to ignore the method call.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -734,14 +734,6 @@ class Gem::Specification < Gem::BasicSpecification
   attr_writer :original_platform # :nodoc:
 
   ##
-  # Deprecated and ignored.
-  #
-  # Formerly used to set rubyforge project.
-
-  attr_writer :rubyforge_project
-  deprecate :rubyforge_project=, :none,       2019, 12
-
-  ##
   # The Gem::Specification version of this gemspec.
   #
   # Do not set this, it is set automatically when the gem is packaged.
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -103,6 +103,13 @@ class Gem::Specification < Gem::BasicSpecification
      4 => 18,
   }.freeze
 
+  ##
+  # Some Specification methods/properties are deprecated and then removed.
+  # Gems installed with older RubyGems versions may have these items
+  # in their .gemspec files.  Used in #method_missing to ignore them.
+
+  REMOVED_METHODS = [:rubyforge_project=].freeze
+
   today = Time.now.utc
   TODAY = Time.utc(today.year, today.month, today.day) # :nodoc:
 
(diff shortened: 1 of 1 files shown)


lib/rubygems/specification.rb
```diff
@@ -103,6 +103,13 @@ class Gem::Specification < Gem::BasicSpecification
      4 => 18,
   }.freeze
 
+  ##
+  # Some Specification methods/properties are deprecated and then removed.
+  # Gems installed with older RubyGems versions may have these items
+  # in their .gemspec files.  Used in #method_missing to ignore them.
+
+  REMOVED_METHODS = [:rubyforge_project=].freeze
```
bronzdoc: I'm not sure if we want to maintain a list of these.

Maybe we should just remove it in the next release since we already warned that it will be deprecated at `2019, 12`  🤔

deivid-rodriguez: But then things will crash in the case @MSP-Greg mentioned, right? I don't think we want to exchange annoying warnings with more annoying crashes, that can't really be fixed easily.

I believe the approach in #3047 is better though, because if the author of the offending gemspec ever rebuilds the gem, she will notice the deprecation needs to be addressed?

MSP-Greg: Was awake anyway... 

> But then things will crash

Even to the point where one can't uninstall the offending gem.

Re https://github.com/rubygems/rubygems/pull/3047, I left that open because I personally wondered about a combination of the two. and also https://github.com/rubygems/rubygems/pull/3047 doesn't allow for actually removing the code for the setter method/attribute.

One question re https://github.com/rubygems/rubygems/pull/3047, it based the 'build' logic on actually using BuildCommand.  Is there a better way to detect that?

> the author of the offending gemspec ever rebuilds the gem, they will notice the deprecation needs to be addressed?

I also wonder about whether it should be a 'stop' rather than a notice/warning.

Thanks, & AFK...

deivid-rodriguez: Yeah. At some point we should raise instead of warn. And at some point after changing the code to raising, we'll be able to completely remove the code.

But for now we just want to fix the issue where deprecation are being printed where not intented, and I think the right way to do that is #3047 better than this PR.

MSP-Greg: What I was trying to say above is (once a determination is made about how to determine if a build is occurring) that the code here could incorporate that into the `#method_missing` logic, and allow removing code.

Something like changing:
```ruby
return if REMOVED_METHODS.include? sym
```
to
```ruby
return if REMOVED_METHODS.include?(sym) and
  Gem::CommandManager.instance.current_command != :build
```

And then it will raise only on build.  Obviously, #3047 needs to be finished first...

MSP-Greg:
Closing in favor of PR https://github.com/rubygems/rubygems/pull/3506

PR #3059, opened 2019-12-29
Fix Gem::Specification#to_ruby without OpenSSL

alyssais:
# Description:

OpenSSL is an optional dependency of Ruby.  Requiring openssl
unconditionally means that Rubies that don't have the OpenSSL library
won't be able to use RubyGems at all.

In such a Ruby, an OpenSSL::PKey::RSA object can't exist, so we can
just skip the check for those objects if there is no OpenSSL.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2414,7 +2414,10 @@ class Gem::Specification < Gem::BasicSpecification
   # still have their default values are omitted.
 
   def to_ruby
-    require 'openssl'
+    begin
+      require 'openssl'
+    rescue LoadError
+    end
     mark_version
     result = []
     result << "# -*- encoding: utf-8 -*-"
@@ -2454,7 +2457,9 @@ class Gem::Specification < Gem::BasicSpecification
       next if handled.include? attr_name
       current_value = self.send(attr_name)
       if current_value != default_value(attr_name) || self.class.required_attribute?(attr_name)
-        result << "  s.#{attr_name} = #{ruby_code current_value}" unless current_value.is_a?(OpenSSL::PKey::RSA)
+        unless defined?(OpenSSL) && current_value.is_a?(OpenSSL::PKey::RSA)
+          result << "  s.#{attr_name} = #{ruby_code current_value}"
+        end
       end
     end
 


simi:
This require was added at https://github.com/rubygems/rubygems/pull/2937. Is Ruby without openssl  supported by rubygems? Simple search for openssl at github (https://github.com/rubygems/rubygems/search?p=1&q=OpenSSL&unscoped_q=OpenSSL) shows different approaches how to handle OpenSSL presence.

lib/rubygems/specification.rb
```diff
@@ -2414,7 +2414,10 @@ def test_files # :nodoc:
   # still have their default values are omitted.
 
   def to_ruby
-    require 'openssl'
+    begin
+      require 'openssl'
```
zimbatm:
```suggestion
      require 'openssl'
      has_openssl = true
```

lib/rubygems/specification.rb
```diff
@@ -2414,7 +2414,10 @@ def test_files # :nodoc:
...
+    rescue LoadError
```
zimbatm:
```suggestion
    rescue LoadError
      has_openssl = false
```

lib/rubygems/specification.rb
```diff
@@ -2454,7 +2457,9 @@ def to_ruby
       next if handled.include? attr_name
       current_value = self.send(attr_name)
       if current_value != default_value(attr_name) || self.class.required_attribute?(attr_name)
-        result << "  s.#{attr_name} = #{ruby_code current_value}" unless current_value.is_a?(OpenSSL::PKey::RSA)
+        unless defined?(OpenSSL) && current_value.is_a?(OpenSSL::PKey::RSA)
```
zimbatm:
```suggestion
        unless has_openssl && current_value.is_a?(OpenSSL::PKey::RSA)
```

zimbatm:
@simi ruby is shipping with rubygems right? If ruby can have openssl disabled then rubygems should probably also support it at least for basic functionality.

alyssais:
> @simi ruby is shipping with rubygems right? If ruby can have openssl
> disabled then rubygems should probably also support it at least for
> basic functionality.

Correct.  Because of this method, installing Ruby 2.7 at all without
OpenSSL is broken.

lib/rubygems/specification.rb
bronzdoc: I like this approach

colby-swandale: i prefer keeping consistent with the library by checking for the constant

mame:
See https://bugs.ruby-lang.org/issues/16475

I think that rubygems does not have to seriously support an environment without openssl, but I want `make test-all` of ruby to work.

zimbatm:
Just to give a use-case: a small ruby script that I want to distribute with a minimal installation of ruby. The script just uses the ruby stdlib and doesn't need network access so OpenSSL is not needed.

deivid-rodriguez:
I'm introducing this fix as part of #3816, so closing this!

Thanks @alyssais and sorry for the delay!

PR #3061, opened 2019-12-31
Gem metadata URLs don't support subdomains with underscore

wintersolutions:
A metadata URL with underscores in the subdomain in `*.gemspec` is considered invalid by `rubygems`:

```ruby
spec.metadata = {'documentation_uri' => 'https://example_workspace.gitlab.io/my_example' }
```
is rejected with:
```
The gemspec at ~/projects/my_example.gemspec is not valid. Please fix this gemspec. (Gem::InvalidSpecificationException)
The validation error was 'metadata['documentation_uri'] has invalid link: "https://example_workspace.gitlab.io/my_example"'
```
A slightly different subdomain like `example-workspace` would be accepted by the regex:

```ruby
%r{\Ahttps?:\/\/([^\s:@]+:[^\s:@]*@)?[A-Za-z\d\-]+(\.[A-Za-z\d\-]+)+\.?(:\d{1,5})?([\/?]\S*)?\z}
```

I therefore propose to change the validation regex into:

```ruby
%r{\Ahttps?:\/\/([^\s:@]+:[^\s:@]*@)?[A-Za-z\d\-\_]+(\.[A-Za-z\d\-\_]+)+\.?(:\d{1,5})?([\/?]\S*)?\z}
```
To allow for underscores in subdomains/domains.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -389,7 +389,7 @@ class Gem::Specification < Gem::BasicSpecification
   # These links will be used on your gem's page on rubygems.org and must pass
   # validation against following regex.
   #
-  #   %r{\Ahttps?:\/\/([^\s:@]+:[^\s:@]*@)?[A-Za-z\d\-]+(\.[A-Za-z\d\-]+)+\.?(:\d{1,5})?([\/?]\S*)?\z}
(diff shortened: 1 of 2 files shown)


znz:
I think normal domains must not include `_`.
It cannot access by some users. (For example: via squid proxy)
I use DNS records including `_` for special purpose only. (For example: `_acme-challenge`)

wintersolutions:
Yeah I did not know that the underscore was considered a invalid character but it seems its not in the standard and just supported by modern browsers. I'll open a ticket with gitlab.

PR #3063, opened 2020-01-01
Updates to some old gem-signing docs.

tiegz:
# Description:

Some of the gem-signing docs are out-of-date since most of them were updated in 2012:

* `gem build` includes a single checksum file now, relpacing the old *.sum files, as of 3f2e05972c85d4f4d9cd5e56e5b033bfb4d11b84
* updates the results of the `gem help cert` command in the docs with newer flags
* replaces an old YAML method that was removed from Psych

Additional question: I almost replaced ~/.gem in these docs with ~/.ssh for storing gem signing keys, since the [Rubygems guides use ~/.ssh](https://guides.rubygems.org/security/), for consistency between docs + guides. But then I wondered "why would you store gem signing keys in an ssh folder?" and couldn't find an explanation in https://github.com/rubygems/guides/pull/70 . Curious which one would be best. 

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -8,7 +8,8 @@
 # Example using a Gem::Package
 #
 # Builds a .gem file given a Gem::Specification. A .gem file is a tarball
-# which contains a data.tar.gz and metadata.gz, and possibly signatures.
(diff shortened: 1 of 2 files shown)


bronzdoc:
@bundlerbot r+

PR #3066, opened 2020-01-02
Vendor bundler 2.1.3

deivid-rodriguez:
# Description:

I think this should fix the current CI issues in master.

PR #3067, opened 2020-01-02
Reduce bors timeout to one hour

deivid-rodriguez:
# Description:

The timeout for bors to report a timeout is excessive in my opinion. For example, #3066 has already failed due to a flaky test issue, but bors won't notify until three hours from now.

### DIFF
--- a/.github/bors.toml
+++ b/.github/bors.toml
@@ -24,7 +24,7 @@ status = [
   "ubuntu_rvm (jruby-9.2.9.0)",
 ]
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
It sounds like the failure in #3066 is not related, actually. Bors simply takes a very long time since a failure happen until it is reported, not sure why.

deivid-rodriguez:
@bundlerbot r=colby-swandale

Let's try it!

PR #3070, opened 2020-01-04
Add .bundle/gems to .gitignore.

duckinator:
# Description:

Since we can use bundler for at least some RubyGems-related stuff now, it seems useful to ignore things installed using it.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.rbc
 *.swp
 .DS_Store
+/.bundle/gems
 /.idea
 /.rdoc
 /.rvmrc


.gitignore
```diff
@@ -1,6 +1,7 @@
 *.rbc
 *.swp
 .DS_Store
+/.bundle/gems
```
deivid-rodriguez: I don't think bundler generates this folder by default. At least it doesn't for me. :thinking:

deivid-rodriguez: Actually, I recently removed this from the official docker images since I thought that folder doesn't exist by default: https://github.com/docker-library/ruby/pull/306/commits/8f3650c2eaa81946c8bf878e3ed55efb0af7790f. I might be missing something here... :/

duckinator: d'oh, poked around and it looks like i configured it to do that 3 years ago(!) and forgot lmao

i'll close this and go add it to my global gitignore. :sweat_smile:

duckinator: d'oh, poked around and it looks like i configured it to do that 3 years ago(!) and forgot lmao

i'll close this and go add it to my global gitignore. :sweat_smile:

duckinator: Huh. I wonder if it's a config option that something suggests, perhaps?

deivid-rodriguez: Ok, I was a bit worried that my commit to the docker library could break things for people. Good to know that it was a personal, old configuration and not some default :sweat_smile:.

deivid-rodriguez: > Huh. I wonder if it's a config option that something suggests, perhaps?

Yeah, that was my concern, since it's too coincidental that the folders match... :man_shrugging:

duckinator: Specifically, this is my `.bundle/config`, and I think the last two lines are the relevant ones:

```
---
BUNDLE_GEM__COC: "true"
BUNDLE_GEM__MIT: "true"
BUNDLE_GEM__TEST: "rspec"
BUNDLE_PATH: ".bundle/gems"
BUNDLE_DISABLE_SHARED_GEMS: "true"
```

deivid-rodriguez: Yeah... Well, this sounds like it belongs to you global gitignore indeed. But I'll investigate if maybe this is a side effect of the `disable_shared_gems` setting or somehing.

duckinator: Alright, I'll close this PR. :+1:

PR #3071, opened 2020-01-07
Make Gem::Specification work even if openssl is unavailable

mame:
# Description:

Currently, to_ruby always require OpenSSL, which makes "make install" of
ruby fail.

See https://bugs.ruby-lang.org/issues/16475

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2414,7 +2414,10 @@ class Gem::Specification < Gem::BasicSpecification
   # still have their default values are omitted.
 
   def to_ruby
-    require 'openssl'
+    begin
+      require 'openssl'
+    rescue LoadError
+    end
     mark_version
     result = []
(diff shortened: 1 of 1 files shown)


mame:
@nobu inserted `require "openssl"` at 0dd5b8856dbecfffdb6be5b192f08804ef5ae3f4.

I think this `require` was needed just for suppressing an NameError during access to the constant `OpenSSL`.  Otherwise, we never get an instance of `OpenSSL::PKey::RSA`, so `current_value.is_a?(OpenSSL::PKey::RSA)` does not make sense at all.  So, what we needed here was not `require "openssl"`, but the check `defined?(OpenSSL) &&`.

If so, it may be good to completely remove `begin require "openssl"; rescue LoadError; end`.

mame:
Oops, @hsbt told me that this duplicates https://github.com/rubygems/rubygems/pull/3071.  Sorry for the noise.

PR #3073, opened 2020-01-08
Remove the bundler version selector

### LINKED ISSUE #2671: Updating rubygems triggers irreversible bundler upgrade
## Behavior

The `gem update --system` causes bundler to be updated to 1.17.3 and it cannot be uninstalled. This prevents people from using an older version of bundler on their system if they have a need for it.

Originally reported in https://github.com/heroku/heroku-buildpack-ruby/pull/835#issuecomment-469878345

## Expected Behavior

Having the rubygems update command also modify the version of bundler was unexpected to me. I expected it to only bump the rubygems version installed.

## Reproduction

Start docker

```
$  docker run -it --rm ruby:2.3.7 bash
```

Update the system

```
root@e4bcc426e52d:/# gem list | grep bundler
bundler (1.16.6, default: 1.16.2)
root@e4bcc426e52d:/# gem update --system
Updating rubygems-update
Fetching: rubygems-update-3.0.3.gem (100%)
Successfully installed rubygems-update-3.0.3
Installing RubyGems 3.0.3
Bundler 1.17.3 installed
RubyGems 3.0.3 installed
root@e4bcc426e52d:/# gem list | grep bundler
bundler (default: 1.17.3, 1.16.6)
```

Uninstall bundler, but it's still there:

```
root@e4bcc426e52d:/# gem uninstall bundler -v 1.17.3
root@e4bcc426e52d:/# gem list | grep bundler
bundler (default: 1.17.3, 1.16.6)
```

### LINKED ISSUE #2800: Ability to disable all Bundler version checks
Can we add an escape valve into Rubygems 2.7 and 3.0 that prevents it from firing the version check, or at least tells it to emit a warning instead of an error?

Maybe an environment variable like `RUBYGEMS_BUNDLER_VERSION_CHECK=warn` ? 

This type of feature would be very useful to work around bugs that we've not yet found, similar to how a bug was not found until bundler 2.0.3 was released. https://github.com/bundler/bundler/issues/7205

### LINKED ISSUE #3275: Bundler 2.1.0 raising error when it doesn't need to
I thought this behavior was removed from bundler:

```
$ bundle
	2: from /Users/rschneeman/.rubies/ruby-2.7.0/bin/bundle:23:in `<main>'
	1: from /Users/rschneeman/.rubies/ruby-2.7.0/lib/ruby/2.7.0/rubygems.rb:294:in `activate_bin_path'
/Users/rschneeman/.rubies/ruby-2.7.0/lib/ruby/2.7.0/rubygems.rb:275:in `find_spec_for_exe': Could not find 'bundler' (1.16.2) required by your /Users/rschneeman/Documents/projects/barnes/Gemfile.lock. (Gem::GemNotFoundException)
To install the missing version, run `gem install bundler:1.16.2`
⛄ 2.7.0 🚀  ~/Documents/projects/barnes (master)
$ ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-darwin17]
⛄ 2.7.0 🚀  ~/Documents/projects/barnes (master)
$ gem -v
3.1.2
```

Here's the Gemfile.lock https://gist.github.com/schneems/751ea79c9886d397e076cf5389575796

I've been asking for some kind of a killswitch or "escape valve" and the message I got was that it wasn't needed because the behavior was being removed. Is that still happening or is that work only going to be targeting 3.0? If it's not going to happen for a while can we add an env var that skips this check ah-la "escape valve"?

### LINKED ISSUE #3283: Bundler 2.1.0 breaks bundle install in projects with Gemfile.lock or older Rubies
The new version of bundler released yesterday, seems to break `bundle install` all over the place at our company.
This seems to be the case when there is a pre-existing `Gemfile.lock` in the project that has a `BUNDLED WITH` section. In our case the version in the Gemfile.lock `1.17.3` in most cases. Which leads to the error:

```
/lib/ruby/site_ruby/2.5.0/rubygems.rb:275:in `find_spec_for_exe': Could not find 'bundler' (1.17.3) required by your /vagrant/mpx-db-stats/Gemfile.lock. (Gem::GemNotFoundException)
To install the missing version, run `gem install bundler:1.17.3
```

The problem can be solved by installing either 1.17.3 with `gem install bundler:1.17.3`, or by removing Gemfile.lock (but not really a good option in some projects).

Also some Travis builds started failing because it uses ruby 2.5.3 by default. Somehow those can be fixed by updating to 2.5.7. However that only helps if Gemfile.lock is already at 2.0.2 I think, as we still had problems with deploys that already build on ruby 2.5.7.

Sorry for this confusing report, but s.th. is weird with this new version and I can't really put my finger on it yet.

### LINKED ISSUE #3334: "bundle --help" not working when Gemfile.lock contains "BUNDLED WITH    2.0.0" but I don't have it installed
Hmm, feels like it applies to all commands and switches, is this the intended behaviour? Does `--version` and `--help` actually read the lockfile?

```
$ bundle --version
	2: from /Users/dentarg/.gem/ruby/2.6.1/bin/bundle:23:in `<main>'
	1: from /Users/dentarg/.rubies/ruby-2.6.1/lib/ruby/site_ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/Users/dentarg/.rubies/ruby-2.6.1/lib/ruby/site_ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.0) required by your /Users/dentarg/Dropbox/code/gists/gists/ruby-bug-15499/Gemfile.lock. (Gem::GemNotFoundException)
```

```
$ bundle --help
	2: from /Users/dentarg/.gem/ruby/2.6.1/bin/bundle:23:in `<main>'
	1: from /Users/dentarg/.rubies/ruby-2.6.1/lib/ruby/site_ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/Users/dentarg/.rubies/ruby-2.6.1/lib/ruby/site_ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.0) required by your /Users/dentarg/Dropbox/code/gists/gists/ruby-bug-15499/Gemfile.lock. (Gem::GemNotFoundException)
```

I'm using:

```
$ bundle --version
Bundler version 1.17.3
```

### LINKED ISSUE #3879: Remove the BundlerVersionFinder / "Bundler version switcher"
I think we've discussed this several times and agreed on it, but I couldn't find an issue on the tracker.

Bundler 1 seems no longer maintained, the [last release of Bundler 1](https://rubygems.org/gems/bundler/versions) was in 2018.
Therefore it seems important, notably for security reasons and to increase Bundler 2 adoption, that Bundler 2 is used for all cases if installed, even if the Gemfile.lock still has `BUNDLED WITH 1.x.y`.

There are many reasons why a Gemfile.lock might still have `BUNDLED WITH 1.x.y`, either it wasn't updated in some time, or maintainers didn't want to force people using Bundler 2, or the I think the main reason: due to `BundlerVersionFinder`, if the Gemfile.lock was `BUNDLED WITH 1.x.y` it would always remain so, because `BundlerVersionFinder` would force using Bundler 1 even if you did not want to.

Also I think the general Bundler version autoswitcher mechanism [has confused more than helped](https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html).
`bundler` would again act like a regular gem in that if you run the executable, you use the latest version again.
That's intuitive, and it's good for security and to get the latest release's bug fixes.

Also the current situation makes it hard or potentially annoying to deprecate APIs that Bundler 1 might still use:
https://github.com/rubygems/rubygems/pull/3817/files#r453293031

I don't know what's the current status regarding unification of RubyGems and Bundler, but it seems for now Bundler is still a separate gem? Any place to read the vision and progress on that?

Regarding multiple lockfile parsers, etc, my understand is Bundler 2's Gemfile.lock syntax is a strict superset of Bundler 1's Gemfile.lock syntax. They both use `lockfile v1` according to https://github.com/rubygems/rfcs/blob/master/text/0000-lockfile-versions.md.
So I think there is no need for versioning the parser yet, and it doesn't seem a prerequisite to remove the BundlerVersionFinder.

Relates to https://github.com/rubygems/rfcs/blob/master/text/0000-lockfile-versions.md
Related to #3275.

cc @indirect @deivid-rodriguez @colby-swandale @schneems @duckinator

deivid-rodriguez:
# Description:

I do like the idea of bundler being able to lock its own version, but this implementation is not the right way to do it in my opinion.

I aim to reimplement this inside bundler in the near future, but it needs to be done in a more user friendly way that properly informs the user about what's going on, and never raises when not necessary.

But for now I think we should remove this code from rubygems because this is causing more problems than benefits and it's being counterproductive.

Closes #2671.
Closes #2800.
Closes #3275.
Closes #3283.
Closes #3334.
Closes #3879.

# Tasks:

- [ ] Write tests


colby-swandale:
The bundler/rubygems team has been communicating to lots of people in the community that we would first propose an RFC of what the exact changes would be in relation to the Bundler Version Finder/Bundler handling of major versions. 

I would prefer to stick to that plan before we commit anything and make sure everyone is on the same page and offer an opportunity for any feedback.

\cc @indirect

colby-swandale:
\cc @segiddins

bronzdoc:
@deivid-rodriguez What problems is this causing you?

bronzdoc:
> The bundler/rubygems team has been communicating to lots of people in the community that we would first propose an RFC 

@colby-swandale Can you point to where that communication is happening, would like to know.

colby-swandale:
> Can you point to where that communication is happening, would like to know.

The Bundler team had a big meeting during RubyKaigi where we discussed the Bundler version Finder and ideas on how to solve the confusion lots of people are having. We came up with a fairly solid idea but wanted to offer people a chance to give feedback before committing to anything.

After RubyKaigi, people online and in-person have asked us what our plans are and  i've just been relaying our plans from RubyKaigi.

deivid-rodriguez:
> @deivid-rodriguez What problems is this causing you?

Hei @bronzdoc, I skimmed through the list of open issues in bundler and rubygems, and filled the PR description with a tentative list of issues that would be closed by this PR (there might be more). I've also seen this issue mentioned in OS packages integration mailing lists, stackoverflow questions, etc. 

> The bundler/rubygems team has been communicating to lots of people in the community that we would first propose an RFC of what the exact changes would be in relation to the Bundler Version Finder/Bundler handling of major versions.

> I would prefer to stick to that plan before we commit anything to make sure everyone is on the same page, and offer an opportunity for any feedback.

I am planning to prepare a RFC for the new version switcher I plan to implement inside bundler. It will be based on the old trampolining code, but will be more user friendly in the sense that it never raises unnecessarily and transparently informs the user about what's going on.

But I really think we should do this first and quick, because this makes bundler very user unfriendly for no good reason (no lockfile format changes between v1 and v2, nor major incompatibilites), and I believe it discourages adoption as @eregon pointed out not long ago.

eregon:
I agree it's better to remove the current and confusing version switcher for Bundler.
Having it in RubyGems makes it very inconvenient to change or adapt, I think the logic should be done in Bundler.

Here is my blog post where I argue we should drop the version autoswitching mechanism in RubyGems: https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html

bronzdoc:
@bundlerbot merge

hsbt:
@deivid-rodriguez I also agree with your proposal. Can you describe the your idea on rfcs or bundler repo? We should merge this after that.

segiddins:
We had it in bundler before and it worked even less than the RubyGems implementation. I’d like to see some evidence that moving it back will be an improvement before just ripping out all this work

eregon:
IMHO the way to fix this is to not have a magical Bundler version switcher but instead:
* Have a format version inside the Gemfile.lock. That format version would be about the Gemfile.lock format, *not* the Bundler gem version.
* All Bundler versions know the most recent format version they support
* Newer Bundler versions can read older format versions
* Older Bundler can still use Gemfile.lock as long as format version <= supported format version (so e.g., Bundler 1 can be used on a Gemfile.lock generated by Bundler 2.0.1)
* Keep the format version in the Gemfile.lock unmodified at least for `bundle install`. Maybe for `bundle update` too. Updating the format should likely be done only when not possible otherwise, explicitly requested, or when there is no `Gemfile.lock`.

That way people can use newest Bundler on older Gemfiles, don't need many different versions of Bundler installed, and only need a newer Bundler when interacting with a newer Gemfile.lock format. Not everyone has to migrate at the same time, reducing frustration if anything doesn't work as expected, or if people would rather migrate later.

benbuckman:
Hi, any news on this since January 16? We're currently fighting the auto-switcher (with multiple versions of Bundler competing and deprecation warnings inconsistently appearing), and the ability to turn it off would help a lot. Thanks!

eregon:
Related merged RFC: https://github.com/rubygems/rfcs/pull/24

Should we proceed with this then?

I'd like Bundler 1 to be able to install Gemfile.lock generated by Bundler 2 (since they use the same lockfile version). Is it something likely to happen?

A workaround for now is to remove the `BUNDLED WITH` line in `Gemfile.lock`, but that's a bit of a hack.

maximeg:
Hi,

What's the status on this?

The strict pinning is quite pain in the ***. I ran on a lot of issues due to this.

Example: 
For a project having a Gemfile.lock `BUNDLED_WITH 2.2.16`
If on the servers, `gem update --system` is ran and it upgrades bundler from 2.6.16 to 2.2.17, suddendly the app would refuse to be restarted, every `bundle exec` commands would fail due to this:

```
May  8 11:19:26 web-1-prod puma[29493]: Activating bundler (2.2.16) failed:
May  8 11:19:26 web-1-prod puma[29493]: Could not find 'bundler' (2.2.16) required by your /home/{project}/current/Gemfile.lock
```

This cause a lot of "synchronize the watches" issues between local env / code / infra.

Can you point me to where those issues are being discussed? 
I really want to have things moving forward by either have pinning by MAJOR.MINOR or the pinning being removed and have a pinned lockfile version or even an opt-out command line option or ...

deivid-rodriguez:
There's no strict pinning at the moment, only major version pinning. I think you must be running into some other issue, please report it and we'll try to help!

I think the current mechanism of preferring the locked version if present but falling back if not is working fine for most people. The current plan for improving this is to make sure that `bundle install` also installs the locked bundler version, so that version switching always ends up selecting the exact locked version.

Anyways, I'm closing this since we don't plan to proceed with it.

PR #3076, opened 2020-01-09
Make sure to reset Gem.ruby when changed

### LINKED ISSUE #3075: Tests fail with SEED=60547

My current problem is that tests fail with SEED=60547 at [current master](https://github.com/rubygems/rubygems/commit/778b7440d78f06bebdaef3cb67f4ab7f97635f51).

Will have a look.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

```
$ gem env version
PASTE HERE
```

deivid-rodriguez:
# Description:

Every time our specs modify either `RbConfig::CONFIG["ruby_install_name"]` or `RbConfig::CONFIG["bindir"]`, we should save the value of `Gem.ruby` prior to modifying it and restore it afterwards since the value of `Gem.ruby` is inferred from these values and memoized.

Fixes #3075.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -169,20 +169,24 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
   # original value when the block ends
   #
   def bindir(value)
-    bindir = RbConfig::CONFIG['bindir']
+    with_clean_path_to_ruby do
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@bundlerbot merge

Issue #3013, opened 2019-12-02
Deprecations should have a version horizon, not a date horizon


deivid-rodriguez:
My current problem is that I think our deprecations should have a version horizon like: "this thing will be removed in rubygems 4", instead of a date horizon like: "this thing will be removed on or after december 2020".

Does anyone know what's the rationale of setting a specific date instead of a future version? Setting a date seems very arbitrary to me.

bronzdoc:
I Think that would be better.

rubyFeedback:
It may be due to xmas release of stable ruby, which rubygems might (or might not) want to sync on (but is at the least partially necessary to sync, since otherwise you have the ruby release cycle become at odds with rubygems, which would be bad for the larger ecosystem IMO)..

I would actually consider to include both. Something like:

"This code part xyz will be removed in rubygems 4" (or 4.0).
"The expected time for this to happen may be after December 2020,
but the exact date may vary."

Or something like this.

That way, people will know at which specific version it will be changed;
AND they would at least know some rough estimate as to when this may
happen - the latter may be useful for people who are not as much aware
for how ruby works, such as for newcomers or something.

The above is just a suggestion; the message could be more terse of course).

duckinator:
I think defaulting to specifying the version _may_ be more useful than dates.

Related but distinct from this, I just opened #3022 ("Large number of methods where planned deprecation dates are 1+ year ago").

Schwad:
Specifying the date is useful for those who rely on the dependency, when forecasting timelines when they ought to resolve to a newer version. 

I understand the sentiments here, but practically dates do have use to teams as well. I don't see switching from it as harmful however.

bronzdoc:
@Schwad The problem with dates is that for whatever reason we cannot remove/deprecate the thing being deprecated at the date we initially specified, we are stuck with that date in the codebase until we decide is appropriate to remove it entirely. Like is happening now, i.e https://github.com/rubygems/rubygems/issues/3022

Right now `Gem::Util.gunzip` will trigger a deprecation warning that it would be removed at `12/2018`. 
We want to avoid that.

bronzdoc:
Closed with https://github.com/rubygems/rubygems/pull/3414

PR #3085, opened 2020-01-16
Only print month & year in deprecation messages

Schwad:
Currently `deprecate` only supports four arguments, allowing for deprecation year and month. Attempting to supply a 'day' fails. 

```
class MyClass
  extend Gem::Deprecate
  
  def the_dying_method
  end
  deprecate :the_dying_method, :the_new_method, 2020, 5, 10
  
  def the_new_method
  end
end

#=> ArgumentError (wrong number of arguments (given 5, expected 4))
```

This supports prior implementation (day is optional) but can support more specific deprecation dates:

`NOTE: MyClass#the_dying_method is deprecated; use the_new_method instead. It will be removed on or after 2020-05-10.`

# Description:

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -60,7 +60,7 @@ module Gem::Deprecate
         target = klass ? "#{self}." : "#{self.class}#"
         msg = [ "NOTE: #{target}#{name} is deprecated",
                 repl == :none ? " with no replacement" : "; use #{repl} instead",
-                ". It will be removed on or after %4d-%02d-01." % [year, month],
+                ". It will be removed on or after %4d-%02d." % [year, month],
                 "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
         ]
         warn "#{msg.join}." unless Gem::Deprecate.skip
--- a/test/rubygems/test_deprecate.rb
+++ b/test/rubygems/test_deprecate.rb
@@ -152,6 +152,6 @@ class TestDeprecate < Gem::TestCase
     assert_match(/OtherThing#foo is deprecated; use bar instead\./, err)
     assert_match(/OtherThing#foo_arg is deprecated; use bar_arg instead\./, err)
     assert_match(/OtherThing#foo_kwarg is deprecated; use bar_kwarg instead\./, err)
-    assert_match(/on or after 2099-03-01/, err)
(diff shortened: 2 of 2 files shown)


bronzdoc:
Hey @Schwad thanks for the PR!

I don't see the benefit of adding a day to the deprecate method. We would rather have a full month to deprecate something than having a day, so I don't see how adding a specific day would improve things?  

We also want to move into `version horizon` instead of  a `date horizon` see: https://github.com/rubygems/rubygems/issues/3013

Schwad:
Hey @bronzdoc - thanks for the kind words!

I completely appreciate your sentiment - and from a pragmatic perspective can see why it specifies the year/month/day in the output on deprecation. 

However, I feel that specifying the "day" in the output but leaving it locked in as the "1st" day of the month is not beneficial. 

> We would rather have a full month to deprecate something than having a day, so I don't see how adding a > specific day would improve things?

Perhaps instead we should change the output to "It will be removed on or after 2020-05" instead of "2020-05-01" if we go along with your viewpoint?

Thanks for your feedback! :)

deivid-rodriguez:
Hi @Schwad, sorry we dropped the ball here. I think completely removing the `01` from the output is the best idea. Would you be willing to update this PR or should we take it over? Thanks! :)

simi:
AFAIK `Gem::Deprecate` is internal API not meant to be used from the outside (there were some troubles with that already). @Schwad I do recommend to stop using `Gem::Deprecate` and bring different deprecation process to your project/gem.

deivid-rodriguez:
I don't like that people are using `Gem::Deprecate`, but it seems too late for that now.  We couldn't come up with a good strategy to deprecate it when we had a look at it recently, so I guess we have to leave with it. If we accept that, I wouldn't mind shipping this tiny improvement.

deivid-rodriguez:
@simi Should I close this, then?

deivid-rodriguez:
Alright, let's do this. I rebased and changed this PR to implement the small suggestion of harcoding the "01" day from the deprecation message.

simi (approved):

bronzdoc (approved):

Schwad:
Thanks, @deivid-rodriguez ! 🙏

deivid-rodriguez:
You're welcome!

PR #3088, opened 2020-01-17
Avoid changing $SOURCE_DATE_EPOCH.

duckinator:
# Description:

Avoid changing $SOURCE_DATE_EPOCH.

- Gem.source_date_epoch no longer sets any environment variables.
- Gem::Ext::Builder sets $SOURCE_DATE_EPOCH for subprocesses.
- If $SOURCE_DATE_EPOCH is unset while the program is running, it reverts
  to using the time when `Gem.source_date_epoch_string` was first called.
- Added test to confirm that `Gem.source_date_epoch` stays stable over time,
  even when $SOURCE_DATE_EPOCH is not set.
- Updated tests that made now-invalid assumptions.
- Add slight kludge to part of the test suite so `Gem.source_date_epoch`
  is regenerated for each test.

# Tasks:

- [x] Write tests
- [ ] code review from coworkers / friends

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -189,6 +189,8 @@ module Gem
   @pre_reset_hooks      ||= []
   @post_reset_hooks     ||= []
 
+  @default_source_date_epoch = nil
+
   ##
   # Try to activate a gem containing +path+. Returns true if
   # activation succeeded or wasn't needed because it was already
@@ -1164,20 +1166,43 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
-  # The SOURCE_DATE_EPOCH environment variable (or, if that's not set, the current time), converted to Time object.
-  # This is used throughout RubyGems for enabling reproducible builds.
+  # If the SOURCE_DATE_EPOCH environment variable is set, returns it's value.
+  # Otherwise, returns the time that `Gem.source_date_epoch_string` was
+  # first called in the same format as SOURCE_DATE_EPOCH.
   #
-  # If it is not set as an environment variable already, this also sets it.
+  # NOTE(@duckinator): The implementation is a tad weird because we want to:
+  #   1. Make builds reproducible by default, by having this function always
+  #      return the same result during a given run.
+  #   2. Allow changing ENV['SOURCE_DATE_EPOCH'] at runtime, since multiple
+  #      tests that set this variable will be run in a single process.
+  #
+  # If you simplify this function and a lot of tests fail, that is likely
+  # due to #2 above.
   #
   # Details on SOURCE_DATE_EPOCH:
   # https://reproducible-builds.org/specs/source-date-epoch/
 
-  def self.source_date_epoch
-    if ENV["SOURCE_DATE_EPOCH"].nil? || ENV["SOURCE_DATE_EPOCH"].empty?
-      ENV["SOURCE_DATE_EPOCH"] = Time.now.to_i.to_s
-    end
+  def self.source_date_epoch_string
+    # The value used if $SOURCE_DATE_EPOCH is not set.
+    @default_source_date_epoch ||= Time.now.to_i.to_s
 
-    Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.freeze
+    specified_epoch = ENV["SOURCE_DATE_EPOCH"]
+
+    # If it's empty or just whitespace, treat it like it wasn't set at all.
+    specified_epoch = nil if !specified_epoch.nil? && specified_epoch.strip.empty?
+
+    epoch = specified_epoch || @default_source_date_epoch
+
+    epoch.strip
+  end
+
+  ##
+  # Returns the value of Gem.source_date_epoch_string, as a Time object.
+  #
+  # This is used throughout RubyGems for enabling reproducible builds.
+
+  def self.source_date_epoch
+    Time.at(self.source_date_epoch_string.to_i).utc.freeze
   end
 
   # FIX: Almost everywhere else we use the `def self.` way of defining class
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1906,6 +1906,24 @@ You may need to `gem install -g` to install missing gems
     assert platform_defaults.is_a? Hash
   end
 
+  # Ensure that `Gem.source_date_epoch` is consistent even if
+  # $SOURCE_DATE_EPOCH has not been set.
+  def test_default_source_date_epoch_doesnt_change
+    old_epoch = ENV['SOURCE_DATE_EPOCH']
+    ENV['SOURCE_DATE_EPOCH'] = nil
+
+    # Unfortunately, there is no real way to test this aside from waiting
+    # enough for `Time.now.to_i` to change -- which is a whole second.
+    #
+    # Fortunately, we only need to do this once.
+    a = Gem.source_date_epoch
+    sleep 1
+    b = Gem.source_date_epoch
+    assert_equal a, b
+  ensure
+    ENV['SOURCE_DATE_EPOCH'] = old_epoch
+  end
+
   def ruby_install_name(name)
     with_clean_path_to_ruby do
       orig_RUBY_INSTALL_NAME = RbConfig::CONFIG['ruby_install_name']
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -106,7 +106,7 @@ class TestGemPackage < Gem::Package::TarTestCase
     assert_equal expected, YAML.load(checksums)
   end
 
-  def test_build_time_source_date_epoch
+  def test_build_time_uses_source_date_epoch
     epoch = ENV["SOURCE_DATE_EPOCH"]
     ENV["SOURCE_DATE_EPOCH"] = "123456789"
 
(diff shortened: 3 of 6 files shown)


duckinator:
I'd like to make sure we're testing that reproducible builds works right with C extensions and the like, but I'm not sure how to do that. There may already be tests for it, but I couldn't find any when poking around real quick.

duckinator:
@bundlerbot try

lib/rubygems.rb
```diff
@@ -1236,20 +1238,40 @@ class << self
...
 
-    Time.at(ENV["SOURCE_DATE_EPOCH"].to_i).utc.freeze
+  ##
+  # Sets the SOURCE_DATE_EPOCH environment variable, runs +block+,
+  # resets SOURCE_DATE_EPOCH, and then returns the result from running +block+.
+  def self.with_source_date_epoch(&block)
+    old_epoch = ENV["SOURCE_DATE_EPOCH"]
+    new_epoch = self.raw_source_date_epoch
+
+    ENV["SOURCE_DATE_EPOCH"] = new_epoch unless old_epoch == new_epoch
```
bronzdoc: It is necessary that we still set this environment variable? could we manage globally the state of the build time internally(using `SOURCE_DATE_EPOCH`) in Rubygems instead?

duckinator: So, I'm not entirely sure. For now, I settled for setting it only when needed. As far as I know, the only way it would be a problem is if it spawns any subprocesses. Whenever RubyGems spawns processes where we don't control all of the code, like if it's building a C extension, then it's completely unavoidable.

This is partially related to what I mentioned above, about wanting tests for reproducible builds + C extensions and such.

duckinator: I basically approached this first attempt by tightening up the timeframe during which we set the environment variable. If it can be tightened even farther, I would definitely like to. But I think reproducible builds for C extensions would be a prerequisite. (Well, assuming that even works right now? I'm not sure if it does.)

duckinator: Also: I _think_ the places I use `Gem.with_source_date_epoch { ... }` are only called during the `build`/`install` subcommands, but I've not confirmed this.

duckinator: Actually, I just missed something very obvious &mdash; if we can find any places where subprocesses would be started during `gem build`/`rake build` and the like, I think we could just put it into the subprocesses environment and avoid it entirely! So, just need to find out if and where that happens, and find a way to tie it in. :slightly_smiling_face:

duckinator:
I'm not sure why this failure is happening: https://github.com/rubygems/rubygems/runs/400125050#step:7:61

And this one appears to be an rvm problem: https://github.com/rubygems/rubygems/runs/400125049#step:6:1

duckinator:
There is also a problem with this PR where, if you don't explicitly set `$SOURCE_DATE_EPOCH`, it doesn't cache the value. This means that, without setting that variable, it winds up not being reproducible.

I do like this behavior because it means that if we save the value of `Gem.source_date_epoch` somewhere we have reproducible-by-default builds, which is super nice. I'll try to fix this in the next few days.

deivid-rodriguez:
> I'm not sure why this failure is happening: https://github.com/rubygems/rubygems/runs/400125050#step:7:61

This one is a problem with bundler's tests. It will be fixed once we version the next version, sorry about that. Let me know if that's blocking you too much and I'll provide a workaround.

> And this one appears to be an rvm problem: https://github.com/rubygems/rubygems/runs/400125049#step:6:1

Yes, it seems like a problem with rvm indeed. I found it yesterday on a different repo. I'll look into it.

duckinator:
@deivid-rodriguez thanks for the info. :+1: atm it's not blocking me, so no rush.

duckinator:
Rebased based on master, and that fixed the unexpected problems I mentioned earlier.

All of the tests are passing. So it's just a matter of resolving this problem, which I'm working on right now:

> There is also a problem with this PR where, if you don't explicitly set $SOURCE_DATE_EPOCH, it doesn't cache the value. This means that, without setting that variable, it winds up not being reproducible.
>
> I do like this behavior because it means that if we save the value of Gem.source_date_epoch somewhere we have reproducible-by-default builds, which is super nice. I'll try to fix this in the next few days.

indirect:
In the original implementation of $SOURCE_DATE_EPOCH, it was a deliberate design decision to only activate reproducible builds if the env var was set, and to keep perfect backwards compatibility if it was not yet. I would be careful about enabling reproducible build logic if the env var is not set, since there is the possibility we could potentially break something that depends on builds being unique every time they are run.

duckinator:
@indirect should the behavior that the last release had be considered a bug, and thus changed now, in that case?

indirect:
@duckinator sorry, I have no idea what behavior the last release had, so I don't know how to answer that question.

duckinator:
@indirect sorry, that was a bit vague. The entire 3.1.x branch has defaulted to builds being reproducible, due to how I implemented it in #2882. Looking at it, though, it seems reproducible-by-default builds was the entire point of that PR.

Also, I think you might've misunderstood what I meant by "reproducible builds by default": If `SOURCE_DATE_EPOCH` is not set, `Gem.source_date_epoch` will always be the current time when the build starts. If you run a build without setting `SOURCE_DATE_EPOCH`, then run it again, you will get unique files.

However, if you build it once, find out what `Gem.source_date_epoch_string` was and then _set `SOURCE_DATE_EPOCH` to that value_ for a second build, the should be the same.

indirect:
oh! I get it now. Yes, that seems great. :)

duckinator:
Added a (currently failing) test for ensuring that `Gem.source_date_epoch` doesn't change over a single run of RubyGems. Unfortunately, my current attempt at fixing it breaks a bunch of other things, so I've not included that just yet. :stuck_out_tongue:

duckinator:
Just committed a first pass at it, since I'm going AFK for a bit. I'm expecting failures in `test_gem_package_tar_writer.rb`, and I think they're caused by uses of `Time.stub :now`. I'll need to think a bit on how to handle that.

A potential kludgey-but-should-be-reliable approach would be to set `@default_source_date_epoch` (in the `Gem` module) to `nil` before every test, as this would trigger `Gem.source_date_epoch`'s default value being recalculated. I may do that as a first attempt to verify the problem, even if I go with a different approach later on.

test/rubygems/test_gem.rb
```diff
@@ -1906,6 +1906,24 @@ def test_platform_defaults
     assert platform_defaults.is_a? Hash
   end
 
+  # Ensure that `Gem.source_date_epoch` is consistent even if
+  # $SOURCE_DATE_EPOCH has not been set.
+  def test_default_source_date_epoch_doesnt_change
+    old_epoch = ENV['SOURCE_DATE_EPOCH']
+    ENV['SOURCE_DATE_EPOCH'] = nil
+
+    # Unfortunately, there is no real way to test this aside from waiting
```
simi: There's `Time.now` being stubbed in `test/rubygems/test_gem_package_tar_writer.rb`. Any reason to not use this approach as well in here? Would be test too coupled with implementation?

duckinator: I suspect you may have commented on the wrong part of the code. Did you intend to comment on [`test/rubygems/test_gem.rb`](https://github.com/rubygems/rubygems/pull/3088/files#diff-6d1d09cbeb26123575a4efe6dc0ddab2R86)?

If you were referring to that, I did add that bit which resets `@default_source_date_epoch` to `test_gem_package_tar_writer.rb` for the same reason.

simi: I mean instead of using `sleep 1` stub `Time.now`.

duckinator: Ah, okay. I think that would be coupling it a bit too closely to the implementation, yeah.

duckinator:
Okay! I think this is done and ready for a proper review, now. :3

duckinator:
@bronzdoc @simi if y'all could take another look at this when you have a moment, I'd appreciate it.

test/rubygems/test_gem_package.rb
```diff
@@ -124,12 +124,10 @@ def test_build_time_source_date_epoch
     ENV["SOURCE_DATE_EPOCH"] = epoch
   end
 
-  def test_build_time_source_date_epoch_automatically_set
+  def test_build_time_without_source_date_epoch
```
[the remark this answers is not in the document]
duckinator: The reason I changed it is because the old one assumed that the default value is set between `start_time = ...` and `end_time = ...`. It being set at-or-after `start_time` is not guaranteed anymore and was never an actual requirement. In retrospect, I think that assertion was accidentally testing an implementation detail.

However, looking at it again, the `end_time = ...` + associated assertion should be able to be re-added without issue. Does adding those two lines back work for you, @bronzdoc? If not, lmk what you'd prefer.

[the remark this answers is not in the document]
bronzdoc: Ah I see. I think is ok as it is @duckinator

bronzdoc:
@bundlerbot merge

bronzdoc:
@bundlerbot retry

duckinator:
That seems unrelated to my work. Not sure what's up.

deivid-rodriguez:
I opened https://github.com/rubygems/rubygems/issues/3140 to track dealing with that flaky test. It seems unrelated indeed.

bronzdoc:
@bundlerbot retry

Issue #3087, opened 2020-01-17
Bundler version finder breaks when run in deleted current working dir


asf-stripe:
I'm having a problem or would like to suggest a feature.

My current problem is that `Gem::Specification.find_by_name('bundler')` raises an exception if the program running it is currently in a deleted directory

This issue is related to:

  - [x] Other

Here are my current environment details:

```
$ gem env version
3.1.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

# Reproducing this issue
Here's a repro case:

```ruby
require 'tmpdir'

Dir.mktmpdir do |dir|
  Dir.chdir(dir)
  puts "pwd is now #{Dir.pwd}"
end

# our current dir is now gone
bundler_gem = Gem::Specification.find_by_name('bundler') # ...and this breaks.
puts("Using Bundler #{bundler_gem.version.to_s}")
```

<details>
  <summary>output and backtrace</summary>

```
pwd is now /private/var/folders/nt/f37r9kbx1n73b4sj48rr4tsw0000gn/T/d20200117-32602-19akv9a
Traceback (most recent call last):
        10: from main.rb:9:in `<main>'
         9: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/specification.rb:988:in `find_by_name'
         8: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/dependency.rb:323:in `to_spec'
         7: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/dependency.rb:303:in `to_specs'
         6: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/dependency.rb:284:in `matching_specs'
         5: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/bundler_version_finder.rb:45:in `filter!'
         4: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/bundler_version_finder.rb:7:in `bundler_version'
         3: from /Users/asf/.rbenv/versions/2.6.3/lib/ruby/site_ruby/2.6.0/rubygems/bundler_version_finder.rb:22:in `bundler_version_with_reason'
[...]

PR #3090, opened 2020-01-21
Fix error when trying to find bundler with a deleted "working directo…

bronzdoc:
closes https://github.com/rubygems/rubygems/issues/3087

# Description:
Fix and error in the Bundler version finder when run in a deleted `current working directory` 
______________

### DIFF
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -83,19 +83,18 @@ To install the missing version, run `gem install bundler:#{vr.first}`
     gemfile = ENV["BUNDLE_GEMFILE"]
     gemfile = nil if gemfile && gemfile.empty?
 
-    pwd = begin
-            Dir.pwd
-          rescue Errno::ENOENT
-            Dir.chdir
-            Dir.pwd
-          end
+    unless gemfile
+      begin
+        Gem::Util.traverse_parents(Dir.pwd) do |directory|
+          next unless gemfile = Gem::GEM_DEP_FILES.find { |f| File.file?(f.tap(&Gem::UNTAINT)) }
 
-    Gem::Util.traverse_parents(pwd) do |directory|
-      next unless gemfile = Gem::GEM_DEP_FILES.find { |f| File.file?(f.tap(&Gem::UNTAINT)) }
-
-      gemfile = File.join directory, gemfile
-      break
-    end unless gemfile
+          gemfile = File.join directory, gemfile
+          break
+        end
+      rescue Errno::ENOENT
+        return
+      end
+    end
 
     return unless gemfile
 
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -82,7 +82,15 @@ To install the missing version, run `gem install bundler:#{vr.first}`
   def self.lockfile_contents
     gemfile = ENV["BUNDLE_GEMFILE"]
     gemfile = nil if gemfile && gemfile.empty?
-    Gem::Util.traverse_parents Dir.pwd do |directory|
+
+    pwd = begin
+            Dir.pwd
+          rescue Errno::ENOENT
+            Dir.chdir
+            Dir.pwd
+          end
+
+    Gem::Util.traverse_parents(pwd) do |directory|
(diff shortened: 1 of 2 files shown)


test/rubygems/test_gem_bundler_version_finder.rb
```diff
@@ -78,6 +78,16 @@ def test_bundler_version_with_reason
...
   end
 
+  def test_deleted_directory
+    require "tmpdir"
+
+    Dir.mktmpdir do |dir|
+      Dir.chdir(dir)
+    end
+
+    assert_nil bvf.bundler_version_with_reason
```
asf-stripe: I think this test probably should chdir back to its old dir, otherwise the other tests are going to have a hard time / fail in an order-dependent way.

lib/rubygems/bundler_version_finder.rb
```diff
@@ -82,7 +82,15 @@ def self.lockfile_version
...
     gemfile = ENV["BUNDLE_GEMFILE"]
     gemfile = nil if gemfile && gemfile.empty?
-    Gem::Util.traverse_parents Dir.pwd do |directory|
+
+    pwd = begin
+            Dir.pwd
+          rescue Errno::ENOENT
+            Dir.chdir
+            Dir.pwd
+          end
```
asf-stripe: I was wondering if there should be a common wrapper for `Dir.pwd`, given its unsafe-ish nature regarding non-existent directories; but I think this should do for now.

Reading up on `Dir.chdir` with no args though, will this work if no `$HOME` env variable is set? It might be safer to not traverse anything in case the current dir doesn't exist (i.e., there's never going to be a Gemfile in the hierarchy, since we can no longer be sure about the hierarchy)

bronzdoc: Re: `Dir.chdir` good point 🤔

bronzdoc: The thing is that even if we don't traverse anything, if we don't restore the dir with `Dir.chdir` and `Dir.pwd` gets called in another part of the code it will raise the same error.

bronzdoc:
For some reason the CI Windows environment can't delete the tmpdir of the test I added for this change - `Errno::EACCES: Permission denied @ dir_s_rmdir - C:/Users/RUNNER~1/AppData/Local/Temp/d20200125-5128-70ojio`

@MSP-Greg Have this happened to you in the past? maybe you can point me to the right track 🙇

MSP-Greg:
@bronzdoc

> Have this happened to you in the past?

Similar.  I've debated whether to track down the issue or not.

Windows uses TEMP and TMP for ENV settings to get the temp folder.  The path is `C:/Users/runneradmin/AppData/Local/Temp`, but since `runneradmin` is a 'long path name', it's 'short name' of `RUNNER~1` is used, probably for compatibility with legacy apps.

What I've done when it's an issue is set TMPDIR to a folder.  TMPDIR is the first ENV value checked by Ruby.  From years ago, I've also set up my local systems that way...

Maybe something like:

```yaml
      - name: Run Test
        run: |
          mkdir temp
          $env:TMPDIR = "$pwd/temp"
          ridk enable
          rake test
```

MSP-Greg:
@bronzdoc

I just looked at the test that's causing the issue.

On Windows, one can't delete the dir that is `$pwd` / `Dir.pwd`.  Nothing to do with temp files or directories...

bronzdoc:
Thanks so much for looking into this @MSP-Greg! 
I guess I would have to explicitly create and remove a dir to test this then. 🤷‍♂

MSP-Greg:
@bronzdoc

I'm probably missing something, but I think the test won't work on Windows and, hence, needs to be skipped.

As one can't delete `$pwd`, one also can't `Dir.chdir` into a non-existent directory...

test/rubygems/test_gem_bundler_version_finder.rb
duckinator: @bronzdoc can you elaborate on why you marked this as resolved? I have the same concern as @asf-stripe, but I may be missing some context. (E.g., does the setup process run before each test call `Dir.chdir`, or something like that?

bronzdoc: Thanks both of you! I fixed it. It will restore to the original PWD, somehow I thought I was passing that block to `Dir.chdir`...

bronzdoc:
@bundlerbot r+

PR #3096, opened 2020-01-24
Warn when gem is outdated

### LINKED ISSUE #1620: gem command should notify the user when it is out of date

This issue is related to:
The command line `gem`

I noticed that `pip` provides feedback as follows when it is out of date:

```
You are using pip version 7.1.2, however version 8.1.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

Similar behavior from rubygems would likely encourage many more users to update their client. I use ruby on a daily basis and often forget to run `gem update --system` for long periods of time.

soapy1:
# Description:

Addresses (this very old) issue: https://github.com/rubygems/rubygems/issues/1620
Will warn when the version of gem is outdated

Closes #1620.

Disclaimer: I'm new to ruby, using this as a way to learn

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -9,6 +9,7 @@ require 'rubygems'
 require 'rubygems/command_manager'
 require 'rubygems/deprecate'
 require 'rubygems/user_interaction'
+require 'rubygems/openssl'
 
 ##
 # Load additional plugins from $LOAD_PATH
@@ -83,7 +84,7 @@ class Gem::GemRunner
   end
 
   def warn_on_outdated
-    if Gem.latest_rubygems_version > Gem.rubygems_version
+    if defined?(OpenSSL::SSL) && (Gem.latest_rubygems_version > Gem.rubygems_version)
       say "You are currently using gem #{Gem.rubygems_version}, however gem #{Gem.latest_rubygems_version} is availble.\nConsider upgrading using the command `gem update --system`\n"
     end
   end
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -33,6 +33,7 @@ class Gem::GemRunner
   # Run the gem command with the following arguments.
 
   def run(args)
+    warn_on_outdated
     build_args = extract_build_args args
 
     do_configuration args
@@ -74,6 +75,15 @@ class Gem::GemRunner
     Gem.use_paths Gem.configuration[:gemhome], Gem.configuration[:gempath]
     Gem::Command.extra_args = Gem.configuration[:gem]
   end
+
+  def warn_on_outdated
+    current_version = Gem.rubygems_version
+    rubygems_latest = Gem.latest_rubygems_version
+    if rubygems_latest > current_version
+      alert_warning "You are currently using gem #{current_version}, however gem #{rubygems_latest} is availble.\nConsider upgrading using the command `gem update --system`\n"
+    end
+  end
+
 end
 
 Gem.load_plugins
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -25,6 +25,9 @@ Gem.load_env_plugins rescue nil
 # classes they call directly.
 
 class Gem::GemRunner
+
+  include Gem::UserInteraction
+
   def initialize
     @command_manager_class = Gem::CommandManager
     @config_file_class = Gem::ConfigFile
@@ -34,11 +37,6 @@ class Gem::GemRunner
   # Run the gem command with the following arguments.
 
   def run(args)
-    begin
-      warn_on_outdated
-    rescue
-      alert_warning "Could not determine if rubygems is out of date"
-    end
     build_args = extract_build_args args
 
     do_configuration args
@@ -57,6 +55,8 @@ class Gem::GemRunner
     end
 
     cmd.run Gem.configuration.args, build_args
+
+    warn_on_outdated
   end
 
   ##
@@ -82,10 +82,8 @@ class Gem::GemRunner
   end
 
   def warn_on_outdated
-    current_version = Gem.rubygems_version
-    rubygems_latest = Gem.latest_rubygems_version
-    if rubygems_latest > current_version
-      say "You are currently using gem #{current_version}, however gem #{rubygems_latest} is availble.\nConsider upgrading using the command `gem update --system`\n"
+    if Gem.latest_rubygems_version > Gem.rubygems_version
+      say "You are currently using gem #{Gem.rubygems_version}, however gem #{Gem.latest_rubygems_version} is availble.\nConsider upgrading using the command `gem update --system`\n"
     end
   end
 
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -25,7 +25,6 @@ Gem.load_env_plugins rescue nil
 # classes they call directly.
 
 class Gem::GemRunner
-
   include Gem::UserInteraction
 
   def initialize
@@ -56,7 +55,9 @@ class Gem::GemRunner
 
     cmd.run Gem.configuration.args, build_args
 
-    warn_on_outdated
+    if !Gem.disable_system_update_message
+      warn_on_outdated
+    end
   end
 
   ##
@@ -86,7 +87,6 @@ class Gem::GemRunner
       say "You are currently using gem #{Gem.rubygems_version}, however gem #{Gem.latest_rubygems_version} is availble.\nConsider upgrading using the command `gem update --system`\n"
     end
   end
-
 end
 
 Gem.load_plugins
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -8,6 +8,7 @@
 require 'rubygems'
 require 'rubygems/command_manager'
 require 'rubygems/deprecate'
+require 'rubygems/user_interaction'
 
 ##
 # Load additional plugins from $LOAD_PATH
@@ -80,7 +81,7 @@ class Gem::GemRunner
     current_version = Gem.rubygems_version
     rubygems_latest = Gem.latest_rubygems_version
     if rubygems_latest > current_version
-      alert_warning "You are currently using gem #{current_version}, however gem #{rubygems_latest} is availble.\nConsider upgrading using the command `gem update --system`\n"
+      say "You are currently using gem #{current_version}, however gem #{rubygems_latest} is availble.\nConsider upgrading using the command `gem update --system`\n"
     end
   end
 
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -34,7 +34,11 @@ class Gem::GemRunner
   # Run the gem command with the following arguments.
 
   def run(args)
-    warn_on_outdated
+    begin
+      warn_on_outdated
+    rescue
+      alert_warning "Could not determine if rubygems is out of date"
+    end
     build_args = extract_build_args args
 
     do_configuration args
--- a/lib/rubygems/source.rb
+++ b/lib/rubygems/source.rb
@@ -179,6 +179,7 @@ class Gem::Source
     local_file = File.join(cache_dir, file_name)
     retried    = false
 
+    require "fileutils"
     FileUtils.mkdir_p cache_dir if update_cache?
 
     spec_dump = fetcher.cache_update_path spec_path, local_file, update_cache?
--- a/test/rubygems/test_gem_gem_runner.rb
+++ b/test/rubygems/test_gem_gem_runner.rb
@@ -133,8 +133,6 @@ class TestGemGemRunner < Gem::TestCase
     use_ui @ui do
       @runner.run(args)
     end
-
-    assert_match /You are currently using gem/, @ui.output
   end
 
   private
--- a/test/rubygems/test_gem_gem_runner.rb
+++ b/test/rubygems/test_gem_gem_runner.rb
@@ -140,5 +140,4 @@ class TestGemGemRunner < Gem::TestCase
   def singleton_gem_class
     class << Gem; self; end
   end
-
 end
--- a/test/rubygems/test_gem_gem_runner.rb
+++ b/test/rubygems/test_gem_gem_runner.rb
@@ -121,4 +121,26 @@ class TestGemGemRunner < Gem::TestCase
 
     assert_empty @ui.error
   end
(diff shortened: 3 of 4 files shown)


util/ci.sh
```diff
@@ -13,7 +13,7 @@ case $1 in
     if [ -d "$HOME/.rvm" ]
     then
       gemhome=$(gem env home)
-      ruby -I lib bin/gem uninstall executable-hooks gem-wrappers bundler-unload -x --force -i "$gemhome@global"
+      ruby -I lib bin/gem uninstall executable-hooks gem-wrappers bundler-unload rubygems-update -x --force -i "$gemhome@global"
```
bronzdoc: I think this is out of the scope of this PR

test/rubygems/test_gem_gem_runner.rb
```diff
@@ -11,12 +11,25 @@ def setup
...
+      end
+    end
   end
 
   def teardown
     super
 
     Gem::Command.build_args = @orig_args
+
+    singleton_gem_class.class_eval do
```
bronzdoc: Maybe we can include this logic in `test_warn_on_outdated` since we just need it for that test

lib/rubygems/gem_runner.rb
```diff
@@ -39,6 +42,11 @@ def initialize(options={})
   # Run the gem command with the following arguments.
 
   def run(args)
+    begin
+      warn_on_outdated
```
bronzdoc: I think we would want to run this message at the end

lib/rubygems/gem_runner.rb
```diff
@@ -81,6 +89,14 @@ def do_configuration(args)
     Gem::Command.extra_args = Gem.configuration[:gem]
   end
 
+  def warn_on_outdated
+    current_version = Gem.rubygems_version
+    rubygems_latest = Gem.latest_rubygems_version
+    if rubygems_latest > current_version
```
bronzdoc: What you think of removing the variables and just use `Gem.rubygems_version` and `Gem.latest_rubygems_version`?

test/rubygems/test_gem_gem_runner.rb
```diff
@@ -97,4 +110,25 @@ def test_search_succeeds
...
+    use_ui @ui do
+      @runner.run(args)
+    end
+
+    assert_match /You are currently using gem/, @ui.output
+  end
+
+  private
+
+  def singleton_gem_class
```
bronzdoc: Maybe we can remove this method and add it's logic to `test_warn_on_outdated?`

bronzdoc:
Thanks so much @soapy1! I apologize for the late response!

I left some comments :)

duckinator:
Hey! I agree with @bronzdoc's review, and don't have anything to add beyond that. When that review is addressed I'll go through the PR again.

Thanks for working on this, @soapy1! :slightly_smiling_face:

lib/rubygems/gem_runner.rb
```diff
@@ -39,6 +42,11 @@ def initialize(options={})
   # Run the gem command with the following arguments.
 
   def run(args)
+    begin
+      warn_on_outdated
+    rescue
```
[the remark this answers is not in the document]
soapy1: iirc if the rubgems-update package is not available getting the latest rubygems versions (`Gem.latest_rubygems_version`) will throw an error. This happens in bundler tests. Will dig into this a bit more

test/rubygems/test_gem_gem_runner.rb
soapy1: I gave that a go but doing it caused tests to fail
```
TestGemGemRunner#test_query_is_deprecated:
RuntimeError: Can't find 'rubygems-update' in any repo. Check `gem source list`.
    /__w/rubygems/rubygems/lib/rubygems.rb:909:in `latest_rubygems_version'
    /__w/rubygems/rubygems/lib/rubygems/gem_runner.rb:86:in `warn_on_outdated'
    /__w/rubygems/rubygems/lib/rubygems/gem_runner.rb:60:in `run'
    /__w/rubygems/rubygems/test/rubygems/test_gem_gem_runner.rb:74:in `block in test_query_is_deprecated'
    /__w/rubygems/rubygems/lib/rubygems/user_interaction.rb:47:in `use_ui'
    /__w/rubygems/rubygems/lib/rubygems/user_interaction.rb:70:in `use_ui'
    /__w/rubygems/rubygems/test/rubygems/test_gem_gem_runner.rb:73:in `test_query_is_deprecated'

```

It seems like the gem `rubygems-update` needs to be around in order to successfully determine `Gem.latest_rubygems_version`. eg.
https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb#L907-L910 

Adding this to the setup/teardown of the tests get around this problem. Maybe there is a nicer way to solve this though? Could also have a try/catch type thing so that it does not fail tragically when `rubygems-update is not available`?

deivid-rodriguez:
I'm sorry we dropped the ball here @soapy1. Would you be willing to revive this PR or should we take it over at this point?

soapy1:
Hey, ya no worries! I'm a bit slammed the next couple of days so I'll probably get it up closer to the end of the week. If that's too late feel free to drive this home 😃

deivid-rodriguez:
That's great @soapy1, we can definitely wait :). Thanks!!

simi:
:thinking: We should disable this when Rubygems is maintained by system package (aka `Gem.disable_system_update_message` is `false or nil`. Otherwise users would get warning, but `gem update --system` would not fix it.

deivid-rodriguez:
Indeed, good catch!

lib/rubygems/source.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+autoload :FileUtils, 'fileutils'
```
soapy1, line 2: I saw that there has recently been some changes with loading fileutils. I'm not sure what the context is there, so I'm not very confident that this is correct. However, without loading fileutils here, the bundler tests fail and checking for updates does not work. The traceback looks something like:
```
 Invoking `gem list` failed with output:
       /home/runner/work/rubygems/rubygems/lib/rubygems/source.rb:182:in `load_specs': uninitialized constant Gem::Source::FileUtils (NameError)
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:256:in `tuples_for'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:220:in `block in available_specs'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/source_list.rb:98:in `each'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/source_list.rb:98:in `each_source'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:216:in `available_specs'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:90:in `search_for_dependency'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:153:in `spec_for_dependency'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems.rb:864:in `latest_spec_for'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems.rb:883:in `latest_version_for'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems.rb:875:in `latest_rubygems_version'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/gem_runner.rb:86:in `warn_on_outdated'
       	from /home/runner/work/rubygems/rubygems/lib/rubygems/gem_runner.rb:59:in `run'
       	from /home/runner/.rubies/ruby-2.4.10/bin/gem:21:in `<main>'
```

soapy1:
@deivid-rodriguez I think this is ready for another review 🎉

lib/rubygems/source.rb
deivid-rodriguez: Yes, the context was that I removed a global fileutils `autoload` like that one in other to workaround some `jruby` code loading bugs (where things apparently sporadically crash if you both `autoload` a constant and also directly `require` the code defining it). So, if the changes in this PR make a missing `FileUtils` constant error surface, I'd rather add the `require` inline right before the error, to keep it consistent and don't hit the same bug in jruby again.

soapy1: ah, I see! Thanks for the context

deivid-rodriguez:
Thanks @soapy1 :heart:. I'll review this as soon as I can.

My mind concern now is whether we should make this compulsory, opt-out, or opt-in. If the second or the latter, it would require us to introduce a setting for it or something. Not sure if someone has any thoughts on this. I'd like to promote upgrading as much as possible, but also I don't want to be annoying, that's why I'm thinking maybe we need a configuration setting.

indirect:
My personal leaning is default on with opt-out via a setting. That’s how (for example) npm and bundler both work today, so it feels fairly standard to me.

lib/rubygems/gem_runner.rb
```diff
@@ -24,6 +25,15 @@
...
 
 class Gem::GemRunner
+  include Gem::UserInteraction
+
+  HAVE_OPENSSL = begin
+    require 'openssl'
+    true
+  rescue LoadError
+    false
+  end
```
deivid-rodriguez, line 35: You can instead add a toplevel `require 'rubygems/openssl'`, and then check `defined?(OpenSSL)` to keep things consistent with what we do in other places.

soapy1: It seems like the checks are not a fan of `defined?(OpenSSL)` 😭 . But maybe I'm doing this wrong?

deivid-rodriguez: I believe you might need to rebase. The file `rubygems/openssl` I only introduced it very recently to unify all the `openssl` loading code. You shouldn't need to wrap the require with any `LoadError` rescuing because that's handled inside that file.

soapy1:
@deivid-rodriguez sorry for the delay with getting all the tests passing. This is ready for another review 🙏

deivid-rodriguez:
@soapy1 It looks very good. Only thing is we'd like to provide a way out of this warning through a `.gemrc` setting, like `warn_when_out_of_date` or something like that. If you're up for adding it that'd be great, otherwise this is a great starting point so that we can wrap up this feature. So thanks a lot in any case! :heart_eyes:

test/rubygems/test_gem_gem_runner.rb
```diff
@@ -109,4 +121,23 @@ def test_search_succeeds
...
+    singleton_gem_class.class_eval do
+      def latest_rubygems_version
+        Gem::Version.new "99999999.0.0"
+      end
+    end
+
+    use_ui @ui do
+      @runner.run(args)
+    end
+  end
```
duckinator, line 136: It looks like this is basically just testing that it successfully runs. Should this be checking the output at all?

duckinator:
Aside from the one comment I left about the test, this looks really good. I'm excited it's finally being added!

soapy1:
> Only thing is we'd like to provide a way out of this warning through a .gemrc setting, like warn_when_out_of_date or something like that. If you're up for adding it that'd be great, otherwise this is a great starting point so that we can wrap up this feature

I think I'll leave that to you folks if that's alright. I think you'll have a much quicker time sorting that all out 😸

deivid-rodriguez:
No problem, we'll wrap it up. Thanks for all the work :heart:.

lib/rubygems/gem_runner.rb
```diff
@@ -74,6 +82,12 @@ def do_configuration(args)
     Gem.use_paths Gem.configuration[:gemhome], Gem.configuration[:gempath]
     Gem::Command.extra_args = Gem.configuration[:gem]
   end
+
+  def warn_on_outdated
+    if defined?(OpenSSL::SSL) && (Gem.latest_rubygems_version > Gem.rubygems_version)
+      say "You are currently using gem #{Gem.rubygems_version}, however gem #{Gem.latest_rubygems_version} is availble.\nConsider upgrading using the command `gem update --system`\n"
```
simi, line 88: I suggest to use `RubyGems` instead of `gem`. `gem` is command name, but common name for whole package manager is `RubyGems`.

"You are currently using RubyGems 3.0.4, however RubyGems 3.1.4 is available."

There's also currently typo in available (`availble`).

lib/rubygems/gem_runner.rb
```diff
@@ -74,6 +82,12 @@ def do_configuration(args)
     Gem.use_paths Gem.configuration[:gemhome], Gem.configuration[:gempath]
     Gem::Command.extra_args = Gem.configuration[:gem]
   end
+
+  def warn_on_outdated
+    if defined?(OpenSSL::SSL) && (Gem.latest_rubygems_version > Gem.rubygems_version)
```
simi, line 87: `Gem.latest_rubygems_version` does remote request, which can fail and raise.

```ruby
irb(main):001:0> Gem.latest_rubygems_version                                                                                                                                                  
        5: from /home/retro/.rubies/ruby-2.7.2/bin/irb:23:in `<main>'
        4: from /home/retro/.rubies/ruby-2.7.2/bin/irb:23:in `load'                            
        3: from /home/retro/.rubies/ruby-2.7.2/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        2: from (irb):1                                                                                                                                                                       
        1: from /home/retro/.rubies/ruby-2.7.2/lib/ruby/2.7.0/rubygems.rb:952:in `latest_rubygems_version'
RuntimeError (Can't find 'rubygems-update' in any repo. Check `gem source list`.)
```

With current implementation, any gem command (even `gem help`) would fail for example when offline. I think we should do this check only for various commands (like gem install, update, ...) where it is common to hit the remote. Also we should handle gracefully when `Gem.latest_rubygems_version ` fails or it doesn't respond in reasonable time.

simi: I really like `yarn` warning about the update. It is not annoying (I had no idea until I looked into code).

https://github.com/yarnpkg/yarn/blob/a4708b29ac74df97bac45365cba4f1d62537ceb7/src/cli/commands/install.js#L1067-L1126

1. don't show upgrade dialog on CI or non-TTY terminals
2. don't check if disabled in user config
3. only check for updates once a day
4. don't bug for updates on tagged releases (don't warn if version includes `-`, I think that makes it friendly for rubygems devs)
5. swallow errors (if anything goes wrong, just ignore and swallow all exceptions)

deivid-rodriguez:
This was superseded by #5922, already merged. Thanks for providing the starting point of this @soapy1, it was really useful to get this moving forward ❤️.

PR #3104, opened 2020-01-26
Remove Gem::Commands::QueryCommand

bronzdoc:
# Description:

Remove deprecated command `Gem::Commands::QueryCommand`
______________


hsbt:
@bundlerbot merge

deivid-rodriguez:
This PR is most likely a mistake.

We only deprecated this command [2 months ago](https://github.com/rubygems/rubygems/pull/2967), and we haven't yet released the deprecation to our end users, so nobody knows that we deprecated the command.

This is the reason why I opened #3013, because with the current system it's very easy to deprecate something and remove it within the same release, which is the exact same thing as abruptly removing it.

hsbt:
Ah, It's reasonable. We should revert this and reapply in the next year.

PR #3108, opened 2020-01-27
Rework plugins system and speed up rubygems

### LINKED ISSUE #2984: All `gem` commands print warnings on master+3.1.0pre2, if certain gems are installed

deivid-rodriguez:
# Description:

I'm pretty excited about this work. The idea is to speed up rubygems require time. This is the improvement in my computer:

#### Before

```
$ time gem env version
NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed on or after 2019-12-01.
Gem::Specification#rubyforge_project= called from /home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/specifications/erubis-2.7.0.gemspec:17.
3.1.2

real	0m0,235s
user	0m0,230s
sys	0m0,007s
```

#### After

```
$ time gem env version
3.2.0.pre1

real	0m0,103s
user	0m0,099s
sys	0m0,007s
```

### The current (slow) rubygems plugin system

The idea comes from noticing that the rubygems plugin system is quite heavy because everytime rubygems is required, it looks in _all_ installed gems for files named `rubygems_plugin.rb`, so that it can load them. This is not just a simple system file search, rubygems also needs to evaluate every gemspec, to know what the `require_paths` for each gem are, and look for a `rubygems_plugin.rb` file in there.

As you can imagine, this is quite inefficient. And also, it explains why simple commands such as `gem env version` print warnings about deprecated gemspec syntax (#2984).

### The new (faster) rubygems plugin system

Instead, I propose to keep rubygems plugins in a well known location, so when rubygems is required, it simply needs to require whatever plugins are in there. This is, of course, much faster and does not require to look for files or evaluate gemspecs.

The tricky part is to keep this new plugins location up to date. To achieve that, we do something similar to what's currently done with binstubs. When we install a gem, we also install a plugin wrapper in there that simply requires the plugin inside the gem. When we uninstall it, we also delete that wrapper. `gem update --system` regenerates plugins, automatically migrating rubygems to the new layout.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -43,10 +43,10 @@ require 'rubygems/errors'
 #
 # == RubyGems Plugins
 #
-# As of RubyGems 1.3.2, RubyGems will load plugins installed in gems or
+# RubyGems will load plugins in the latest version of each installed gem or
 # $LOAD_PATH.  Plugins must be named 'rubygems_plugin' (.rb, .so, etc) and
-# placed at the root of your gem's #require_path.  Plugins are discovered via
-# Gem::find_files and then loaded.
+# placed at the root of your gem's #require_path.  Plugins are installed at a
+# special location and loaded on boot.
 #
 # For an example plugin, see the {Graph gem}[https://github.com/seattlerb/graph]
 # which adds a `gem graph` command.
@@ -148,6 +148,7 @@ module Gem
     doc
     extensions
     gems
+    plugins
     specifications
   ].freeze
 
@@ -420,6 +421,10 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
 
   def self.spec_cache_dir
     paths.spec_cache_dir
+  end
+
+  def self.plugins_dir
+    File.join(dir, "plugins")
   end
 
   ##
@@ -980,6 +985,20 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
+  # Glob pattern for require-able plugin suffixes.
+
+  def self.plugin_suffix_pattern
+    @plugin_suffix_pattern ||= "_plugin#{suffix_pattern}"
+  end
+
+  ##
+  # Regexp for require-able plugin suffixes.
+
+  def self.plugin_suffix_regexp
+    @plugin_suffix_regexp ||= /_plugin#{suffix_regexp}\z/
+  end
+
+  ##
   # Suffixes for require-able paths.
 
   def self.suffixes
@@ -1077,11 +1096,10 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
-  # Find the 'rubygems_plugin' files in the latest installed gems and load
-  # them
+  # Find rubygems plugin files in the standard location and load them
 
   def self.load_plugins
-    load_plugin_files find_latest_files('rubygems_plugin', false)
+    load_plugin_files Gem::Util.glob_files_in_dir("*#{Gem.plugin_suffix_pattern}", plugins_dir)
   end
 
   ##
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -398,6 +398,7 @@ lib/rubygems/install_message.rb
 lib/rubygems/install_update_options.rb
 lib/rubygems/installer.rb
 lib/rubygems/installer_test_case.rb
+lib/rubygems/installer_uninstaller_utils.rb
 lib/rubygems/local_remote_options.rb
 lib/rubygems/mock_gem_ui.rb
 lib/rubygems/name_tuple.rb
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -280,6 +280,13 @@ class Gem::BasicSpecification
   end
 
   ##
+  # Returns the list of plugins in this spec.
+
+  def plugins
+    matches_for_glob("rubygems#{Gem.plugin_suffix_pattern}")
+  end
+
+  ##
   # Returns a string usable in Dir.glob to match all requirable paths
   # for this spec.
 
--- a/lib/rubygems/commands/pristine_command.rb
+++ b/lib/rubygems/commands/pristine_command.rb
@@ -40,6 +40,11 @@ class Gem::Commands::PristineCommand < Gem::Command
       options[:only_executables] = value
     end
 
+    add_option('--only-plugins',
+               'Only restore plugins') do |value, options|
+      options[:only_plugins] = value
+    end
+
     add_option('-E', '--[no-]env-shebang',
                'Rewrite executables with a shebang',
                'of /usr/bin/env') do |value, options|
@@ -126,14 +131,14 @@ extensions will be restored.
         end
       end
 
-      unless spec.extensions.empty? or options[:extensions] or options[:only_executables]
+      unless spec.extensions.empty? or options[:extensions] or options[:only_executables] or options[:only_plugins]
         say "Skipped #{spec.full_name}, it needs to compile an extension"
         next
       end
 
       gem = spec.cache_file
 
-      unless File.exist? gem or options[:only_executables]
+      unless File.exist? gem or options[:only_executables] or options[:only_plugins]
         require 'rubygems/remote_fetcher'
 
         say "Cached gem for #{spec.full_name} not found, attempting to fetch..."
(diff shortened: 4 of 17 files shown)


deivid-rodriguez:
I updated the PR description, and the implementation should be ready.

I'm still thinking of adding some higher level tests, but I'm not sure whether to add them to `test/rubygems/test_gem_commands_install_command.rb` and `test/rubygems/test_gem_commands_uninstall_command.rb`, or to add a realworld tests similar to the one I've been running locally:

```bash
rm -f ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/plugins/yard_plugin.rb && rake install && gem install yard:0.9.23 && cat ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/plugins/yard_plugin.rb && gem install yard:0.9.24 && cat ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/plugins/yard_plugin.rb && gem uninstall yard:0.24.0 --executables --backtrace && cat ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/plugins/yard_plugin.rb && gem uninstall yard:0.23.0 --executables --backtrace && ls ~/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/plugins/ -la
```

Essentially, install and uninstall gems with plugins and make sure the plugins location is properly updated.

deivid-rodriguez:
Oh, also something to be noticed, if you try this locally and don't see a great improvement, it might be because you have etiher the `executable-hooks` or the `gem-wrappers` gems installed. These gems have rubygems plugins that are quite slow, so that's the bottleneck in that case, so the improvement is not apparent. But if you have `yard` installed, which also includes a rubygems plugin, it should still be much faster.

deivid-rodriguez:
Ok, so I found and fixed some of bugs, and added a bunch of tests, so I'm happy with this now! Removing the WIP tag :)

lib/rubygems.rb
```diff
@@ -1077,11 +1096,10 @@ def self.load_plugin_files(plugins) # :nodoc:
   end
 
   ##
-  # Find the 'rubygems_plugin' files in the latest installed gems and load
-  # them
+  # Find rubygems plugin files in the standard location and load them
 
   def self.load_plugins
-    load_plugin_files find_latest_files('rubygems_plugin', false)
+    load_plugin_files Gem::Util.glob_files_in_dir("*#{Gem.plugin_suffix_pattern}", plugins_dir)
```
deivid-rodriguez: This is the most important change in this PR. We go from:

* Look for requirable files named 'rubygems_plugin' in all installed gems, which means looking in every installed gem's folder, evaluate its gemspec to find its 'require_paths' (usually `lib/`) and actually look in that folder.

to: 

* Require any files in the standard rubygems plugins folder.

The rest of the code in this PR is essentially code to be able to do this particular change safely and in a backwards compatible way.

deivid-rodriguez:
Thanks both for having a look, let's do it!

@bundlerbot merge

lib/rubygems.rb
dgholz: Will this respect $LOAD_PATH? If I activate a gem, I'd expect its plugin to be added to `gem`, and if I deactivate it, I'd expect its plugin not to be added.

deivid-rodriguez: What do you mean exactly by "I expect its plugin to be added to `gem`"? In any case, the end user behavior of `rubygems` should be unchanged by this PR.

dgholz: No problem! So the current way to load plugins is to scan for all active gems that have a `rubygems_require.rb` in the `lib` folder. So, if a gem isn't activated, its plugins aren't added, and when it is activated (by having its `lib` folder listed in the `$LOAD_PATH`), its plugins are added.

What I mean by "I expect its plugin to be added to `gem`" is just the first part, the normal, regular expected behaviour. I mentioned it to be symmetrical with the deactivation case.

I think, with this refactor, if I install a gem, its plugin will be added to `gem` regardless of if the gem is activated or not. With the pre-refactor implementation, gems that weren't activated didn't have their plugins added to `gem`, because activation meant "had their gem `lib` folder added to `$LOAD_PATH`", and that's where `Gem::Specification.latest_specs` looks.

deivid-rodriguez: I'm pretty sure it doesn't work the way you think, although I agree it would probably make more sense the way you explain, but I didn't want to change the behaviour just now.

Rubygems loads plugins for the latest version of every installed gem (in addition to plugins present in the $LOAD_PATH), no matter whether it has yet been activated or not.

dgholz: As an example, if I have the following `Gemfile`s:
```
source 'https://rubygems.org'

gem 'gem-compiler'
```

```
source 'https://rubygems.org'
```

Then, when I run `bundle exec gem compile --help` with the first one, I'd expect to see the help docs for `gem-compiler`'s plugin, and for the second one I'd expect to see `Unknown command compile`.

deivid-rodriguez: I see. Everything I said applies for a "non-bundled context".

But `bundler` further modifies `rubygems` plugins system to adapt everything `rubygems` does to apply to just the gems in the current bundle. So you're right, maybe whatever monkeypatches `bundler` is doing to adapt to `rubygems` plugins system need to be modified.

I'll have a look and get back to you.

deivid-rodriguez: I just tried this and it seems to work as expected? Did you try it?

dgholz: I tried it, but I don't have any new-style plugin gems to install. With `gem-compile`, it works because of the call to [`load_env_plugins` in `Gem::GemRunner`](https://github.com/rubygems/rubygems/blob/b933fb6b103f7b8aae7881e5f80fbf5f9d3e26dc/lib/rubygems/gem_runner.rb#L16). Is there a gem you are using that uses the reworked plugin system?

dgholz: Also thanks for following up on this!

deivid-rodriguez: No problem. What do you mean by "new style plugins"?

dgholz: The ones that register themselves by creating a file in `Gem.plugins_dir`.

deivid-rodriguez: That is _all_ rubygems plugins actually. If your gem includes a `rubygems_plugin.rb` file, it will be registered like that in the next rubygems version. In other words, `gem-compiler` is an example of such a plugin, and it works as expected for me.

dgholz: ah, because the `load plugin` statement in `load_plugin_files` catches the exception when the `require` fails, because the gem is not in the `$LOAD_PATH`. So, the answer to my original question is 'yes, it does respect `$LOAD_PATH`'.

Will you be removing the call to `load_env_plugins` in the future?

deivid-rodriguez: I don't think that rescue is actually involved here, since the `require` uses absolute paths, so it does not involve `$LOAD_PATH`. The reason this "works" is that even if the "compile" command is actually registered, it can never be loaded when running through `bundler` since the command file is not on the `$LOAD_PATH`.

See the error against current rubygems:

```
$ bundle exec gem compile
    Unknown command compile
``` 

vs the new error

```
$ bundle exec gem compile
ERROR:  Loading command: compile (LoadError)
	cannot load such file -- rubygems/commands/compile_command
    undefined method `invoke_with_build_args' for nil:NilClass
```

This means we _do_ need to adapt `bundler` to work with the new plugin system.

I'm working on a fix for this [here](https://github.com/rubygems/bundler/compare/catch_up_with_rubygems_plugin_loading_changes).

deivid-rodriguez: Ok, so I spent a fair bit of time on this today. Turns out the different error messages are unrelated to this change in rubygems, but depend on the fact whether `gem-compiler` is installed globally (in addition to through the `Gemfile` or not).

Even if not caused by this particular change in rubygems, I think that is a bug since the global environment should never "pollute" a `bundle exec` context like that, so I'm fixing it in the branch I pointed to.

PR #3113, opened 2020-01-28
Remove functionality to create an Gem::Installer object with a string

bronzdoc:
# Description:

This deprecation warning has been showing since `2015`
```shell
de3f8d0de lib/rubygems/installer.rb          (Aaron Patterson       2015-04-10 14:20:44 -0700 187)         warn "constructing an Installer object with a string is deprecated. Please use Gem::Installer.at (called from: #{caller.first})"
```
 Maybe it's time to actually remove functionality to create `Gem::Installer` objects with a string.
______________

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -180,15 +180,7 @@ class Gem::Installer
     require 'fileutils'
 
     @options = options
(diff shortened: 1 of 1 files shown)


hsbt:
@bundlerbot merge

PR #3115, opened 2020-01-30
Fix incorrect test to test what it wants to test

deivid-rodriguez:
# Description:

This PR fixes an incorrect test that I found while working on #3108.

Previously `util_setup_installer` was ignoring the block passed to it, the user installed gem was not even named "default". However, since it installed an executable with the same name as the default gem, the test still passed.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
@bundlerbot merge

bronzdoc:
:shipit:

PR #3122, opened 2020-02-03
Allow releasing with a rubygems pre version

deivid-rodriguez:
# Description:

This condition is too restrictive in my opinion. If a user has a pre version of rubygems installed, she should be fully responsible for it, and we shouldn't restrict any functionality.

Also, why is a new prerelease disallowed but an old prelease allowed, or why is 2.0.0.rc2 explictly whitelisted? I believe this kind of exceptions are one more reason to actually permit this.

# Tasks:

- [ ] Write tests


bronzdoc:
@bundlerbot merge

Issue #1829, opened 2017-01-27
Sources without trailing slash not resolved correctly


randycoulman:
I'm having a problem or would like to suggest a feature.

My current problem is when attempting to run any `gem` command with a gem source that doesn't have a trailing slash, the last segment of the path is removed.  For example, following the [documentation for a gemstash private gem repository](https://github.com/bundler/gemstash/blob/master/docs/gemstash-private-gems.7.md), I might configure a gem source like `http://localhost:9292/private`.  When I attempt to install a gem (or list or search) from that source, the `gem` command line will attempt to `GET` `/api/v1/dependencies?gems=my_private_gem` instead of `/private/api/v1/dependencies?gems=my_private_gem`.

See the discussion starting with this comment: https://github.com/bundler/gemstash/pull/134#issuecomment-275400757

Also, see this skipped test on the gemstash repository: https://github.com/bundler/gemstash/blob/master/spec/integration_spec.rb#L140-L147

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.8
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

smellsblue:
I've investigated this before, and I think I found where the problem was. This line (https://github.com/rubygems/rubygems/blob/master/lib/rubygems/source.rb#L187):
```ruby
spec_path  = api_uri + "#{file_name}.gz"
```
is concatenating a URI with another string, which ends up dropping the path of the url. Here is a demonstration in IRB:
```
$ irb
irb(main):001:0> require "uri"
=> true
irb(main):002:0> uri = URI.parse("https://some-gem.server/custom-path")
=> #<URI::HTTPS https://some-gem.server/custom-path>
irb(main):003:0> puts(uri + "latest_specs.4.8.gz")
https://some-gem.server/latest_specs.4.8.gz
=> nil
```

I don't know if there are other places with the same problem, but it seems to me that this should instead result in `https://some-gem.server/custom-path/latest_specs.4.8.gz`... but maybe there are cases where the existing behavior is more correct?

djberg96:
http://blog.honeybadger.io/why-is-uri-join-so-counterintuitive/

I wonder if File.join would be appropriate here.

indirect:
[...]

PR #3123, opened 2020-02-04
Handle unexpected behavior with URI#merge and subpaths missing trailing slashes

drcapulet:
Due to how `URI#merge` works (which `URI#+` is aliased to), when using a source under a subpath, it can get overwritten:

```ruby
> URI.parse('https://gems.example.com/private') + './api/v1/dependencies'
=> #<URI::HTTPS https://gems.example.com/api/v1/dependencies>
> URI.parse('https://gems.example.com/private/') + './api/v1/dependencies'
=> #<URI::HTTPS https://gems.example.com/private/api/v1/dependencies>
```

It's not immediately obvious that errors fetching would be caused by a missing slash, especially if you have a server mounted at both `/` and `/private`.

This PR makes sure that calls to `URI#+` are more obvious and subpaths aren't overwritten and instead appended.

Fixes https://github.com/rubygems/rubygems/issues/1829.

### DIFF
--- a/lib/rubygems/source.rb
+++ b/lib/rubygems/source.rb
@@ -80,7 +80,7 @@ class Gem::Source
   def dependency_resolver_set # :nodoc:
     return Gem::Resolver::IndexSet.new self if 'file' == uri.scheme
 
-    bundler_api_uri = uri + './api/v1/dependencies'
+    bundler_api_uri = enforce_trailing_slash(uri) + './api/v1/dependencies'
 
     begin
       fetcher = Gem::RemoteFetcher.fetcher
@@ -132,7 +132,7 @@ class Gem::Source
 
     spec_file_name = name_tuple.spec_name
 
-    source_uri = uri + "#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}"
+    source_uri = enforce_trailing_slash(uri) + "#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}"
 
     cache_dir = cache_dir source_uri
 
@@ -176,7 +176,7 @@ class Gem::Source
     file       = FILES[type]
     fetcher    = Gem::RemoteFetcher.fetcher
     file_name  = "#{file}.#{Gem.marshal_version}"
-    spec_path  = uri + "#{file_name}.gz"
+    spec_path  = enforce_trailing_slash(uri) + "#{file_name}.gz"
     cache_dir  = cache_dir spec_path
     local_file = File.join(cache_dir, file_name)
     retried    = false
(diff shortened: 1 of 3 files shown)


bronzdoc:
Hey @drcapulet! could you add the test into `test/rubygems/test_gem_source.rb`? thank you!

drcapulet:
@bronzdoc Was there something you'd want me to cover in `test/rubygems/test_gem_source.rb` that isn't in `test/rubygems/test_gem_source_subpath_problem.rb`? Or do you mean try merging them together? If so, I wasn't sure on a good way to achieve what I wanted with `#setup`.

bronzdoc:
@drcapulet sorry for the confusion. I meant "move" not "add"

hsbt:
@drcapulet Hi, Can you move the your test into the `test_gem_source.rb`?

After that, I will merge this.

drcapulet:
@hsbt As I mentioned earlier, I wasn't sure of a good way to set `@gem_repo` - do you just want me to copy the setup across the three tests?

deivid-rodriguez:
@hsbt Should @drcapulet copy the test setup across each of the tests or do you have another preferred way? In my opinion, a separate test file like it is coded currently is fine.

By the way, does this fix or is related to https://github.com/rubygems/rubygems/issues/1829?

drcapulet:
@deivid-rodriguez Yes, this should fix #1829.

deivid-rodriguez:
Great! Could you rebase this PR to make sure everything is still good, and so that the PR is ready to merge in case @hsbt is fine with it now?

lib/rubygems/source.rb
```diff
@@ -80,7 +80,7 @@ def ==(other) # :nodoc:
   def dependency_resolver_set # :nodoc:
     return Gem::Resolver::IndexSet.new self if 'file' == uri.scheme
 
-    bundler_api_uri = uri + './api/v1/dependencies'
+    bundler_api_uri = uri.merge(uri.path.gsub(/\/+$/, '') + '/').merge('api/v1/dependencies')
```
deivid-rodriguez, line 83: To make the actual fix here more apparent, could we extract this logic to a `enforce_trailing_slash` helper, and keep the previous code untouched otherwise?

```suggestion
    bundler_api_uri = enforce_trailing_slash(uri) + './api/v1/dependencies'
```

deivid-rodriguez (approved):
I rebased this PR and added my suggestions. I'll merge in a couple of days unless there's further feedback.

deivid-rodriguez:
Thanks @drcapulet!

PR #3124, opened 2020-02-04
Fix -I require priority

deivid-rodriguez:
# Description:

If `require "a"` is run when two folders have been specified in the -I option including a "a.rb" file and a "a.so" file respectively, [the ruby spec says](https://github.com/ruby/spec/blob/d80a6e2b221d4f17a8cadcac75ef950c59cba901/core/kernel/shared/require.rb#L234-L246) that the ".rb" file should always be preferred. However, the logic we added in https://github.com/rubygems/rubygems/commit/6b81076d9 to make the -I option always beat default gems does not respect
this spec, creating a difference from the original ruby-core's require.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -43,18 +43,18 @@ module Kernel
     # https://github.com/rubygems/rubygems/pull/1868
     resolved_path = begin
       rp = nil
-      $LOAD_PATH[0...Gem.load_path_insert_index || -1].each do |lp|
-        safe_lp = lp.dup.tap(&Gem::UNTAINT)
-        begin
-          if File.symlink? safe_lp # for backward compatibility
-            next
+      Gem.suffixes.each do |s|
+        $LOAD_PATH[0...Gem.load_path_insert_index || -1].each do |lp|
+          safe_lp = lp.dup.tap(&Gem::UNTAINT)
+          begin
+            if File.symlink? safe_lp # for backward compatibility
+              next
+            end
+          rescue SecurityError
+            RUBYGEMS_ACTIVATION_MONITOR.exit
+            raise
           end
-        rescue SecurityError
-          RUBYGEMS_ACTIVATION_MONITOR.exit
-          raise
-        end
 
-        Gem.suffixes.each do |s|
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@hsbt In order to reproduce this issue with a test, I had to revert 9325078c63e7d48013059c2e5844dcfdd26bb092. I think the issue you were having when running tests inside ruby-core back then could actually be this same issue, so that's why I need to revert it. Could you check this PR against ruby-core?

deivid-rodriguez:
This issue was found by the Debian ruby team when trying to package psych for ruby2.7, and this PR should close the related issue they opened to the psych project: https://github.com/ruby/psych/issues/431.

deivid-rodriguez:
I need to figure out the Windows CI issue, but other than that I think it should be ready.

deivid-rodriguez:
Windows CI is passing now too :+1:.

lib/rubygems/core_ext/kernel_require.rb
```diff
@@ -43,18 +43,18 @@ def require(path)
     # https://github.com/rubygems/rubygems/pull/1868
     resolved_path = begin
       rp = nil
-      $LOAD_PATH[0...Gem.load_path_insert_index || -1].each do |lp|
-        safe_lp = lp.dup.tap(&Gem::UNTAINT)
-        begin
-          if File.symlink? safe_lp # for backward compatibility
-            next
+      Gem.suffixes.each do |s|
```
hsbt: Note: I confirmed `Gem.suffixes` returned `["", ".rb", ".bundle"]` on macOS.

deivid-rodriguez: Thanks! I had two doubts here:

* Is the "empty extension" actually needed?
* Should I explicitly loop first through ".rb", and then through the rest of the extensions, so that we don't depend on the order of `Gem.suffixes`?

deivid-rodriguez: Both of the issues are actually present in the current code, so I'll move this forward for now.

deivid-rodriguez:
@bundlerbot r=hsbt

PR #3130, opened 2020-02-12
Prefer to use String#match

hsbt:
# Description:

`String#match?` is faster 10% than `=~` .

see https://github.com/ruby/csv/pull/30
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/compatibility.rb
+++ b/lib/rubygems/compatibility.rb
@@ -38,3 +38,23 @@ module Gem
 
   RubyGemsPackageVersion = VERSION
 end
+
+# https://github.com/ruby/csv/pull/30
+unless String.method_defined?(:match?)
+  module Gem
+    module MatchP
+      refine String do
+        def match?(pattern)
+          pattern === self
+        end
+      end
+    end
+
+    refine Regexp do
+      def match?(string)
+        self === string
+      end
+    end
+  end
+
+end
--- a/lib/rubygems/package/tar_header.rb
+++ b/lib/rubygems/package/tar_header.rb
@@ -30,6 +30,8 @@
 
 class Gem::Package::TarHeader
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Can I close this one? I don't think we need this.

hsbt:
@deivid-rodriguez I added the motivation about this.

deivid-rodriguez:
Yeah, I mean, I'm good with this but is there a specific code path that gets much faster with this change and that justifies introducing compatibility code for this instead of waiting until we drop ruby 2.3 support? I mean there's a lot of micro optimizations we could introduce like.. anything from [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance) or [https://github.com/DamirSvrtan/fasterer](fasterer), but... why this specific one and not others?

Also, regarding this particular change, I think `String#ends_with?` might be faster and supports ruby 2.3.

olleolleolle:
@deivid-rodriguez [String#end_with?](https://rubyapi.org/2.3/o/string#method-i-end_with-3F) would be useful to check "suffixes", but in this string, it's all of the String. From the \A to the \z.

Would it help to drag out the pattern into a constant? `/\A[0-7]*\z/`

deivid-rodriguez:
Right, I misread the code, thank you!

Then forget about my `end_with?` suggestion.

My first concern is still valid, I don't see a reason for this particular optimization given the compatibility code we need to add. I prefer to wait until we drop ruby 2.3 support.

hsbt:
Closing favor from https://github.com/rubygems/rubygems/pull/3398/commits/ef2dcef07533163a5da2e3efd9faf6ee5aa4c3cc

hsbt:
ref. https://github.com/rubygems/rubygems/pull/3398

Issue #3131, opened 2020-02-13
(default) gem conflicts with StdLib


voxik:
The libraries required via RubyGems conflict with StdLib libraries (and possibly also with libraries loaded from different load paths). Consider this:

~~~
$ ruby -v
ruby 2.5.7p206 (2019-10-01 revision 67816) [x86_64-linux]

$ gem update psych --user-install
Updating installed gems
Updating psych
WARNING:  You don't have /builddir/.gem/ruby/2.5.0/bin in your PATH,
	  gem executables will not run.
Building native extensions. This could take a while...
Successfully installed psych-3.1.0
Parsing documentation for psych-3.1.0
Installing ri documentation for psych-3.1.0
Installing darkfish documentation for psych-3.1.0
Done installing documentation for psych after 1 seconds
Parsing documentation for psych-3.1.0
Done installing documentation for psych after 0 seconds
Gems updated: psych

$ RUBYOPT="--disable-gems -rpsych" ruby -e 'require "rubygems"; Gem.load_yaml'
/builddir/.gem/ruby/2.5.0/gems/psych-3.1.0/lib/psych.rb:237: warning: already initialized constant Psych::LIBYAML_VERSION
/usr/local/lib/ruby/2.5.0/psych.rb:232: warning: previous definition of LIBYAML_VERSION was here

$ RUBYOPT="-rpsych" ruby -e '$LOAD_PATH.unshift "/usr/local/lib/ruby/2.5.0"; require "psych"'
/usr/local/lib/ruby/2.5.0/psych.rb:232: warning: already initialized constant Psych::LIBYAML_VERSION
/builddir/.gem/ruby/2.5.0/gems/psych-3.1.0/lib/psych.rb:237: warning: previous definition of LIBYAML_VERSION was here
~~~

So the issue is that RubyGems are loading libraries which were already loaded. I think that if `require "psych"` was successful once, it should never trigger load once again. But as you can see that is not the case.

Please note that the example is using Ruby 2.5.7 just because there is newer Psych available then the one in the StdLib, but it would be the same issue for Ruby 2.7.

deivid-rodriguez:
@voxik I proposed a fix for the first case you explained since it seems reasonable to me that if a file from a default gem was loaded prior to rubygems, rubygems should respect that when loading the same thing through rubygems "extended require".

[...]

PR #3132, opened 2020-02-13
Little cleanup in test suite

deivid-rodriguez:
# Description:

Just some cleanups in the test suite. I made this while working on #3131.

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -78,8 +78,6 @@ class TestGemRequire < Gem::TestCase
     FileUtils.mkdir_p File.dirname c_rb
     File.open(c_rb, 'w') {|f| f.write "class Object; HELLO = 'world' end" }
 
-    lp = $LOAD_PATH.dup
-
     # Pretend to provide a commandline argument that overrides a file in gem b
     $LOAD_PATH.unshift dash_i_arg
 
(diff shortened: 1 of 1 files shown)


bronzdoc:
@bundlerbot r+

deivid-rodriguez:
This is currently introducing a flaky test failure, so I prefer to investigate that before merging, ok?

@bundlerbot r-

deivid-rodriguez:
This PR should be ready now. The idea was to kill some complicated logic with `$LOADED_FEATURES` restoring.

Previously, the tests would assume that a "meaningful feature", like say `rubygems/gem_runner` could be loaded in he middle of a test. That means not all "dummy features" loaded by a test (like test gems and things like that) need to be removed from the `$LOADED_FEATURES` global variable after the test. These "meaningful features" need to be preserved because if they are not, other tests might load them, and since they are not in `$LOADED_FEATURES`, `require` will think they haven't yet been loaded and will load them again, causing redefinition warnings.

So we have code like:

https://github.com/rubygems/rubygems/blob/39fae05861bb71fb9d0aa12b50c16d9b0362fbbe/lib/rubygems/test_case.rb#L427-L439

This strategy is overcomplicated and also brittle. For example, under the ruby-core setup of tests as proposed in https://github.com/ruby/ruby/pull/3213, where `.` is in the `$LOAD_PATH`, and `tmp` is contained inside `.`, this logic fails to work properly and that's what's causing the failures there.

As an alternative, we make sure tests only load "dummy feaures" and every thing else is preloaded. And with that prerequisite, we can kill all this logic.

deivid-rodriguez:
Most of the changes had been introduced separately. I rebased this and left the changes the still make sense to me, I removed the cleanup I mentioned in the previous message because it didn't seem worth it now.

PR #3133, opened 2020-02-14
Improve require

deivid-rodriguez:
# Description:

Some changes to make our custom require do less work and to get tests passing against ruby-core.

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -567,18 +567,20 @@ class TestGemRequire < Gem::TestCase
     write_file File.join(@tempdir, "extconf.rb") do |io|
       io.write <<-RUBY
         require "mkmf"
+        CONFIG['LDSHARED'] = '$(TOUCH) $@ ||'
         create_makefile("#{name}")
       RUBY
     end
 
     write_file File.join(@tempdir, "#{name}.c") do |io|
       io.write <<-C
-        #include <ruby.h>
         void Init_#{name}() { }
       C
     end
 
-    spec.files += ["extconf.rb", "#{name}.c"]
+    write_file File.join(@tempdir, "depend")
+
+    spec.files += ["extconf.rb", "depend", "#{name}.c"]
 
     so = File.join(spec.gem_dir, "#{name}.#{RbConfig::CONFIG["DLEXT"]}")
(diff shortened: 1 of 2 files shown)


test/rubygems/test_require.rb
```diff
@@ -567,18 +567,20 @@ def util_install_extension_file(name)
...
       io.write <<-C
-        #include <ruby.h>
         void Init_#{name}() { }
       C
     end
 
-    spec.files += ["extconf.rb", "#{name}.c"]
+    write_file File.join(@tempdir, "depend")
+
+    spec.files += ["extconf.rb", "depend", "#{name}.c"]
```
bronzdoc: why we adding the "depend" string?

deivid-rodriguez: It's an empty file, actually.

This is related to the compilation of the extension, ruby-core runs against several different systems, and on some systems, an unavailable header was being added to the compilation and causing the extension to not compile. I think this was probably an issue in ruby-core, and I think it's probably fixed by removing the `<ruby.h>` header (I will check), but the idea of the empty `depend` file is that compiling the `.c` file does not depend on any other files, preventing this kind of issues by making the extension "more dummy". I copied it from another dummy extension in some other test.

Anyways, I will review this an follow up.

deivid-rodriguez: @bronzdoc So if I revert this, compiling the extension tries to pull the `config.h` header which is missing. See for example, this ruby-core test run after I reverted these changes: https://github.com/ruby/ruby/pull/2889/checks?check_run_id=467008623.

I'm not really sure why, but making sure nothing gets pulled with an empty `depend` file fixes it, so I'm setting this PR as ready.

bronzdoc:
@bundlerbot merge

deivid-rodriguez:
Thanks for merging @bronzdoc. After a second thought though, I'd like to revert 4fc0ab21c0f7713829abb522ce3b6d8e24c126b3.

Technically, extensionless ruby files are valid ruby files that can be required. For example, `bin/bundle` is sometimes required from other binstubs even if it's also runnable directly.

So, we should technically consider this kind of files too.

Have a look at #3153.

PR #3137, opened 2020-02-17
Update the gem method for Gem::Installer

djberg96:
Currently the `gem` method on the `Gem::Installer` class method is always nil, since it's an `attr_reader` but `@gem` is never set anywhere.

This PR updates it so that it returns the filename of the gem package.

Followup to https://github.com/rubygems/rubygems/issues/2750


bronzdoc:
@bundlerbot merge

djberg96:
@bronzdoc Thanks!

PR #3138, opened 2020-02-18
Use modern Ruby idioms

voxik:
The `Specification::uniq_by` and `Specification::sort_by!` are not required since PR #2255. On top of that, `Specification::sort_by!` is not used at all. Since both methods are private, it should be safe to remove them immediately.


bronzdoc:
@bundlerbot merge

PR #3143, opened 2020-02-20
Make `util/bisect` script more useful

deivid-rodriguez:
# Description:

Sometimes when I work on order dependent test failures, I forget to run bisections through `rake bisect` and run the `util/bisect` script directly. But that does not currently work.

This PR moves changes inside `util/bisect` so that it works regardless of how it's run.

# Tasks:

- [ ] Write tests

### DIFF
--- a/util/bisect
+++ b/util/bisect
@@ -7,8 +7,7 @@ seed = begin
          abort "Specify the failing seed as the SEED environment variable"
        end
 
-gemdir = `gem env gemdir`.chomp
-ENV["RUBYOPT"] = "-Ilib:bundler/lib:test:#{gemdir}/gems/minitest-server-1.0.5/lib"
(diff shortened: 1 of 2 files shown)


util/bisect
```diff
@@ -1,6 +1,16 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
+seed = begin
+         Integer(ENV["SEED"])
+       rescue
+         abort "Specify the failing seed as the SEED environment variable"
+       end
+
+gemdir = `gem env gemdir`.chomp
```
simi: Does `gemdir = Gem.dir` do the same?

deivid-rodriguez: Absolutely, not sure why I did that in the first place. Will update, thanks! :+1:

deivid-rodriguez:
Thanks for the review @simi :purple_heart:

@bundlerbot merge

Issue #3050, opened 2019-12-23
Issues unpacking tar files "is not an octal string"


lacostej:
I am unable to unpack some tar files that tar handles properly. It fails at parsing the `size`

```
	2: from /Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_reader.rb:61:in `each'
	1: from /Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_header.rb:112:in `from'
/Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_header.rb:130:in `strict_oct': "          0" is not an octal string (ArgumentError)
```
and `checksum`
```
	2: from /Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_reader.rb:61:in `each'
	1: from /Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_header.rb:114:in `from'
/Users/lacostej/.rvm/rubies/ruby-2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/package/tar_header.rb:130:in `strict_oct': " 15423" is not an octal string (ArgumentError)
```

Stripping to the `size` and `checksum` fields make the problem go away:
```
        :size     => strict_oct(fields.shift.strip),
        :mtime    => strict_oct(fields.shift),
        :checksum => strict_oct(fields.shift.strip),
```

Here are my current environment details:

```
$ gem env version
3.1.2
```

(I was on 2.7.6 when I encountered the issue, and updated to see if that issue had been fixed)

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
Hey @lacostej can you provide steps to reproduce? thank you

lacostej:
@bronzdoc I cannot distribute the zip file I encountered the problem with: it is a unity package I purchased on Unity.com. So far I only have one file I reproduced it on.

Here's the output of the execution of the code that shows the regression between ruby 2.4.1 and 2.5.3

[...]

PR #3146, opened 2020-02-20
Allow spaces in file headers during octal check

trizes:
Fixes https://github.com/rubygems/rubygems/issues/3050.

# Description:
I've encountered archives with files that contain spaces in uid, gid, size and other fields. Decompression of such archives returns an error due a `self.strict_oct` check in tar_header.rb file:
`ArgumentError: "     0" is not an octal string`

This pull request strips spaces inside `strict_oct` check. I can't reproduce a file with space in the header unfortunately and can't share the files that caused an error, files do not belong to me and headers are overwritten when I clear the files. I suspect files are created on a very old system or hex editor is used to clear headers for security.

______________

# Tasks:

- [x] Write tests


deivid-rodriguez:
I read [the wikipedia page](https://en.wikipedia.org/wiki/Tar_(computing)#Header) about the TAR format and it says:

> Numeric values are encoded in octal numbers using ASCII digits, with leading zeroes. For historical reasons, a final NUL or space character should also be used. Thus although there are 12 bytes reserved for storing the file size, only 11 octal digits can be stored. This gives a maximum file size of 8 gigabytes on archived files. To overcome this limitation, star in 2001 introduced a base-256 coding that is indicated by setting the high-order bit of the leftmost byte of a numeric field. GNU-tar and BSD-tar followed this idea. Additionally, versions of tar from before the first POSIX standard from 1988 pad the values with spaces instead of zeroes.

So, it sounds like this tar file you're having trouble with was created by a version of tar that might have been released before I was born :rofl:.

The change looks simple enough, but I'm a bit hesistant to support such an ancient format. Could you to talk to the owner of the file so that they regenerate the file using the new format. I wonder if a current version of `tar` will be able to untar the file? In that case, fixing the file would be as simple as untar'ing and tar'ing back again.

jmartin-r7:
Can this be backported to 3.0 and 3.1 streams?

deivid-rodriguez:
Not sure if there's any change of backporting this. Are you having any issues with upgrading?

deivid-rodriguez:
To 3.2.0.rc.1 I mean. `gem update --system --pre` should do the trick.

jmartin-r7:
@deivid-rodriguez, thanks for taking a look. I have another workaround for now. I just happen to use tar processing and have run into this issue with an archive. I can wait to remove my workaround after 3.2 has a stable released.

PR #3147, opened 2020-02-21
Clarify gem install --bindir purpose

olivierlacan:
The sentence "Directory where executables are located" suggests that
"executables" are already "located" (prior to running the command) there.

I know this is obvious from the perspective of implementors but this is
confusing, especially from the perspective of someone who doesn't realize there
is a difference between a gem and its binary executable.

This propose alternative suggest by the use of the future tense that this is a
prospective binary executable installation path and not a present one:

> Directory where executables will be placed when the gem is installed


bronzdoc:
@bundlerbot merge

PR #3149, opened 2020-02-21
Resolve latest version via `gem contents`

dnrce:
# Description:
## Issue
Scenario: I have two versions of gem `foo` installed.
Current behavior: `gem contents foo` returns information about the older version.
Expected behavior: `gem contents foo` returns information about the newer version.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/commands/contents_command.rb
+++ b/lib/rubygems/commands/contents_command.rb
@@ -167,7 +167,7 @@ prefix or only the files that are requireable.
   end
 
   def spec_for(name)
-    spec = Gem::Specification.find_all_by_name(name, @version).last
+    spec = Gem::Specification.find_all_by_name(name, @version).first
 
     return spec if spec
 
--- a/test/rubygems/test_gem_commands_contents_command.rb
+++ b/test/rubygems/test_gem_commands_contents_command.rb
@@ -152,6 +152,23 @@ class TestGemCommandsContentsCommand < Gem::TestCase
     assert_equal "", @ui.error
   end
 
+  def test_execute_show_install_dir_latest_version
+    @cmd.options[:args] = %w[foo]
+    @cmd.options[:show_install_dir] = true
+
+    gem 'foo', 1
+    gem 'foo', 2
+
+    use_ui @ui do
+      @cmd.execute
+    end
+
+    expected = File.join @gemhome, 'gems', 'foo-2'
+
+    assert_equal "#{expected}\n", @ui.output
+    assert_equal "", @ui.error
+  end
+
   def test_execute_show_install_dir_version
     @cmd.options[:args] = %w[foo]
     @cmd.options[:show_install_dir] = true
--- a/test/rubygems/test_gem_commands_contents_command.rb
+++ b/test/rubygems/test_gem_commands_contents_command.rb
@@ -105,6 +105,22 @@ class TestGemCommandsContentsCommand < Gem::TestCase
     assert_empty @ui.error
   end
 
+  def test_execute_missing_version
+    @cmd.options[:args] = %w[foo]
+    @cmd.options[:version] = Gem::Requirement.new '= 2'
+
+    gem 'foo', 1
+
+    assert_raises Gem::MockGemUi::TermError do
+      use_ui @ui do
+        @cmd.execute
+      end
+    end
+
+    assert_match "Unable to find gem 'foo'", @ui.output
+    assert_empty @ui.error
+  end
+
   def test_execute_missing_multiple
     @cmd.options[:args] = %w[foo bar]
 


lib/rubygems/commands/contents_command.rb
```diff
@@ -167,18 +167,16 @@ def show_files(files)
...
+  rescue Gem::MissingSpecError
     say "Unable to find gem '#{name}' in #{@path_kind}"
 
     if Gem.configuration.verbose
       say "\nDirectories searched:"
       @spec_dirs.sort.each { |dir| say dir }
     end
 
-    return nil
+    nil
```
bronzdoc: I prefer to explicitly call `return nil`,  just `nil` at the end seems weird

lib/rubygems/commands/contents_command.rb
```diff
@@ -167,18 +167,16 @@ def show_files(files)
   end
 
   def spec_for(name)
-    spec = Gem::Specification.find_all_by_name(name, @version).last
-
-    return spec if spec
-
+    Gem::Specification.find_by_name(name, @version)
```
bronzdoc: `spec = Gem::Specification.find_all_by_name(name, @version).first` ?

lib/rubygems/commands/contents_command.rb
dnrce: I was just following https://github.com/rubocop-hq/ruby-style-guide#no-explicit-return and matching the way `true` is returned on L127 and L137 but I'm happy to remove this change.

lib/rubygems/commands/contents_command.rb
dnrce: This method currently has bad logic (with `.last`) so I was trying to get away from it having _any_ of its own logic for choosing which of the gems to use, because that logic already exists elsewhere:
 * `Gem::Specification.find_all_by_name` is documented as `Returns every spec that matches +name+`, which then requires an additional logical step in this method to limit it to one spec.
 * `Gem::Specification.find_by_name` is documented as `Find the best specification matching a +name+` and already incorporates the necessary logic to find a single spec.

Does that thinking make sense?

But I'm certainly not very familiar with the RubyGems code so if you think simply switching from `.last` to `.first` is better, I can update.

lib/rubygems/commands/contents_command.rb
simi: IMHO it's good practice to not touch lines which do not need to be touched.

lib/rubygems/commands/contents_command.rb
deivid-rodriguez: I prefer `.first` too for this fix, since it makes the fix more explicit and consistent with what we do in other commands, for example:

https://github.com/rubygems/rubygems/blob/98ca0d40f1f961a44c55695139ac1d7df7c087d8/lib/rubygems/commands/open_command.rb#L81

I'm happy to discuss improving/unifying this, but I think it should be a separate PR.

lib/rubygems/commands/contents_command.rb
deivid-rodriguez: Neutral to this style change, but if we decide to enable that cop, it should be done as part of a separate PR and discussion.

dnrce: Agreed on leaving other lines untouched as a general rule -- in this case it was not to enforce the style cop by itself so much as it was to make this half of the block parallel the other half after the addition of the `rescue`. (This line itself wasn't touched but its block context did change.)

I've updated the PR though and this is no longer in scope.

lib/rubygems/commands/contents_command.rb
dnrce: Thank you, that other example is super helpful to see. I'm totally happy to stay consistent with that so have updated the PR accordingly.

test/rubygems/test_gem_commands_contents_command.rb
```diff
@@ -105,6 +105,22 @@ def test_execute_missing_single
     assert_empty @ui.error
   end
 
+  def test_execute_missing_version
```
dnrce: After having revised the PR this test isn't strictly necessary to the functionality I'm updating, but I have left it in anyway because it seems to be a missing case. Can remove if it feels out of scope though.

deivid-rodriguez: I'm happy with including that test, but I would split it into a separate commit to make it clear that it's not related to the bug fix.

dnrce: Done.

bronzdoc:
@bundlerbot merge

dnrce:
Thanks, all, for the review!

Issue #3018, opened 2019-12-04
typoing or deprecated/removed gems should find a bit more, by relaxing the search (and allowing '-')


rubyFeedback:
I am updating some of my old gems, and noticed that the gem suggestion
on the commandline is a bit ... off, in my opinion.

Specifically take a gem named "taglib". In my old code I am using
it.

I then do this on the commandlie:

    gem install taglib

I find no useful result and instead gem displays this:

ERROR:  Could not find a valid gem 'taglib' (>= 0) in any repository
ERROR:  Possible alternatives: rtaglib, taglib2, taglob, agilib, alglib

Ok so far so fine - but the new release is actually:

    https://rubygems.org/gems/taglib-ruby

So the name is similar, same starting name "taglib", just that the
author added "-ruby" as string. (On a side note: none of my own
gems ever has any "-" character, but I do use "_").

I am not totally sure this is the new one, but the name is more
similar than the other variants. Yet it does not show up in the
suggestion offered here.

Not sure if the "Possible alternatives" listing is limited, but
I would also consider to include gems that

a) .start_with? the same input name, int his case "taglib"

and, more importantly

b) to consider ignoring "-" tokens if they happen past the
input given by the user.

That way the suggestion could also show "taglib-ruby",
which I think is the real variant that I want. At the least
I can not find another explanation why rtaglib would be
suggested, or taglob, but not taglib-ruby. (In the long
run perhaps we could have some additional meta-info or so
to make the message-response by gem better, but for the
time being this small change might help a little. The
only change would be to include it in the above listing.
It's even more logical than "alglib" IMO, since I'd have
to type quite a lot to come from "taglib" to "alglib".)

PS: I managed to typo twice in the title ... :\

duckinator:
Hey!

I'm not entirely sure what the best way to handle this is, but figured I'd provide some background on how this works currently, for context. :slightly_smiling_face: 

---

These suggestions ultimately come from [`Gem::SpecFetcher#suggest_gems_from_name()`](https://github.com/rubygems/rubygems/blob/bde88f0/lib/rubygems/spec_fetcher.rb#L174-L198).

It's implemented using the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) to determine how similar a string is. My understanding of how this works is that two strings are considered more similar if fewer changes are required to get from one to the other, and less similar if it requires more changes.

[...]

PR #3151, opened 2020-02-21
Install suggestions

soapy1:
# Description:
addresses the issue: https://github.com/rubygems/rubygems/issues/3018
Makes the suggested gems when install fails more similar to the results of using `gem search`.
Just throwing an idea out there, maybe not the direction that this should be going in.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/spec_fetcher.rb
+++ b/lib/rubygems/spec_fetcher.rb
@@ -171,7 +171,7 @@ class Gem::SpecFetcher
   # Suggests gems based on the supplied +gem_name+. Returns an array of
   # alternative gem names.
 
-  def suggest_gems_from_name(gem_name, type = :latest)
+  def suggest_gems_from_name(gem_name, type = :latest, num_results = 5)
     gem_name        = gem_name.downcase.tr('_-', '')
     max             = gem_name.size / 2
     names           = available_specs(type).first.values.flatten(1)
@@ -194,10 +194,10 @@ class Gem::SpecFetcher
                 matches.uniq.sort_by { |name, dist| dist }
               end
 
-    matches.first(5).map { |name, dist| name }
+    matches.first(num_results).map { |name, dist| name }
   end
 
-  def suggest_gems_from_name_regex(gem_name, type = :latest)
+  def suggest_gems_from_name_regex(gem_name, type = :latest, num_results = 5)
     name = Regexp.new gem_name.downcase.tr('_-', '')
     names = available_specs(type).first.values.flatten(1)
 
@@ -212,7 +212,7 @@ class Gem::SpecFetcher
                 matches.uniq
               end
 
-    return matches
+    return matches.first(num_results)
   end
 
   ##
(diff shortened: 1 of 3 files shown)


lib/rubygems/spec_fetcher.rb
```diff
@@ -197,6 +197,24 @@ def suggest_gems_from_name(gem_name, type = :latest)
     matches.first(5).map { |name, dist| name }
   end
 
+  def suggest_gems_from_name_regex(gem_name, type = :latest)
```
bronzdoc: I noticed this is almost the same as `Gem::SpecFetcher.fetcher.suggest_gems_from_name` but without the `levenshtein_distance` filter. I'm not sure if this is what we want, in theory the `levenshtein_distance` should give us the closest match but I checked this functionality and I see the regex results are useful too. Maybe we can include the two results? 🤔

soapy1: That seems good to me. How do you envision the 2 results getting combined? Currently, it looks like the top 5 results are taken from the levenshtein_distance filter and all the results are returned form the regex filter. I think mashing the two results together might be ok. That might also result in having a really big list of suggestions.

bronzdoc: Maybe we could show the 10 first results and show first the regex matches that match the name of the gem at the start and then the `levenshtein_distance filetered` ones ?

soapy1: Gave this a crack. Now it will return 10 results. It will try to fill that 10 with the regex filter first, then fill in the rest with the levenshtein filter

bronzdoc:
There's some problem with the bundler build, will give it a closer look tommorow

lib/rubygems/spec_fetcher.rb
```diff
@@ -171,30 +171,34 @@ def spec_for_dependency(dependency, matching_platform=true)
...
       next unless n.match_platform?
-
-      distance = levenshtein_distance gem_name, n.name.downcase.tr('_-', '')
-
-      next if distance >= max
-
-      return [n.name] if distance == 0
-
-      [n.name, distance]
+      [n.name, 0] if name === n.name.downcase.tr('_-', '')
```
duckinator: Any reason to prefer regular expressions and `===` over this? To the best of my knowledge, they're equivalent.

```suggestion
      [n.name, 0] if n.name.downcase.tr('_-', '').include?(gem_name)
```

lib/rubygems/spec_fetcher.rb
```diff
@@ -171,30 +171,34 @@ def spec_for_dependency(dependency, matching_platform=true)
   # Suggests gems based on the supplied +gem_name+. Returns an array of
   # alternative gem names.
 
-  def suggest_gems_from_name(gem_name, type = :latest)
+  def suggest_gems_from_name(gem_name, type = :latest, num_results = 5)
     gem_name        = gem_name.downcase.tr('_-', '')
+    name            = Regexp.new gem_name
```
duckinator: If you apply my suggestion to use `.include?(...)` below, be sure to remove this line since it will be unused. :slightly_smiling_face: 

```suggestion
```

soapy1:
@duckinator good catch, using not regex looks way nicer!

duckinator:
@bundlerbot r+

PR #3153, opened 2020-02-25
Revert "Exclude empty suffix from `-I` require loop"

deivid-rodriguez:
# Description:

This reverts commit 4fc0ab21c0f7713829abb522ce3b6d8e24c126b3.

Technically, extensionless ruby files are valid ruby files that can be required. For example, `bin/bundle` is sometimes required from other binstubs even if it's also runnable directly.

So, we should technically consider this kind of files too.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
@bundlerbot r=bronzdoc

PR #3152, opened 2020-02-24
Run unstable workflow as a daily scheduled build

deivid-rodriguez:
# Description:

We want our CI in PRs to be as deterministic as possible, while still find breakage with ruby-head as it changes reasonably early. I think this is a good compromise for that.

# Tasks:

- [x] Describe the problem / feature
- [ ] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3155, opened 2020-02-26
Try fix scheduled workflow

deivid-rodriguez:
# Description:

The scheduled workflow added in #3152 doesn't seem to be running. The only different with the [example in the documentation](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onschedule) is the quoting so I'd like to double check whether it makes a difference.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/unstable.yml
+++ b/.github/workflows/unstable.yml
@@ -2,7 +2,7 @@ name: unstable
 
 on:
   schedule:
-    - cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Yeah, I would be surprised too. And yeah, I'm not sure how else we can try this since I think scheduled workflows only run on master.

deivid-rodriguez:
@bundlerbot r=bronzdoc

PR #3156, opened 2020-02-26
Only rescue the errors we actually want to rescue

deivid-rodriguez:
# Description:

Ruby's `rbinstaller.rb` is incorrectly calling `Gem.ensure_gem_subdirectories` with keyword arguments [here](https://github.com/ruby/ruby/blob/83705c42cedd9489596859827d7201c59feccebd/tool/rbinstall.rb#L822) and [here](https://github.com/ruby/ruby/blob/83705c42cedd9489596859827d7201c59feccebd/tool/rbinstall.rb#L869).

But this method does not take keyword args :grimacing:.

Because of this blind rescue, this error goes unnoticed when `rbinstaller.rb` is run. 

This PR fixes this problem by rescuing only the specific error we intend to rescue, and let any other errors happen.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -468,7 +468,10 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     subdirs.each do |name|
       subdir = File.join dir, name
       next if File.exist? subdir
-      FileUtils.mkdir_p subdir, **options rescue nil
+      begin
+        FileUtils.mkdir_p subdir, **options
+      rescue Errno::EACCES
+      end
     end
   ensure
     File.umask old_umask


deivid-rodriguez:
I'm thinking that this fixes a bug, but fixing this breaks ruby's `rbinstall.rb`, which is used by `make install` to install default gems and bundled gems. The call is being corrected at ruby/ruby#2930 thought.

I _think_ this is fine because older rubies will still ship with older rubygems versions, where neither the `Gem.ensure_gem_subdirectories` call nor the implementation will be fixed so things will work as they have always worked. If user upgrades rubygems on an older ruby, that will break `rbinstall.rb`, but at that point ruby is already installed, so I believe that's fine... :thinking:

deivid-rodriguez:
I'll go ahead and do this.

@bundlerbot merge

PR #3160, opened 2020-02-27
Reenable skipped test on jruby

deivid-rodriguez:
# Description:

We used to have some flaky test failures on tests similar to this one, but it seems like they are gone. This test doesn't look too special compared to the others in this file, and the "skip comment" talks about openssl in general, not about the specific test which sounds work since `openssl` _does_ work on `jruby`.

So I'll tentatively reenable it.

# Tasks:

- [ ] Write tests


hsbt:
@bundlerbot merge

PR #3163, opened 2020-03-05
Fix new plugins layout edge cases

deivid-rodriguez:
# Description:

I run into some issues with the new plugins layout while working on an unrelated bundler PR. The problem in a few words is that the rubygems plugin management system does not respect the `:install_dir` parameter to `Gem::Installer` and `Gem::Uninstaller`.

This PR fixes that.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -319,6 +319,13 @@ module Gem
   end
 
   ##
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
@bundlerbot merge

PR #3164, opened 2020-03-05
Rename scheduled build

deivid-rodriguez:
# Description:

The current scheduled build is not currently working, and I'm not sure why. A potential cause is that this specific workflow has already run as a "regular workflow" in the past, so that somehow could be preventing it from running now as a scheduled workflow? :man_shrugging:

As a blind guess, I'd like to check whether renaming it fixes the issue.

If this doesn't work, I'll ask at Github support.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1,11 +1,11 @@
-name: unstable
+name: daily
 
 on:
   schedule:
     - cron: '0 0 * * *'
 
 jobs:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Jruby flaky is back :(

deivid-rodriguez:
I'll try this just in case.

@bundlerbot merge

PR #3172, opened 2020-03-13
CmakeBuilder seems to fail on JRuby regardless of CI

eregon:
* See https://github.com/jruby/jruby/issues/6113
* Also, note that $CI is not set in GitHub Actions at the time of writing:
  https://github.com/actions/virtual-environments/issues/520

cc @deivid-rodriguez


deivid-rodriguez:
I'm actually observing the opposite. These tests seem to consistently pass locally on jruby for me, so my guess is that we added the CI check due to some issue with the old Travis CI.

So I'm trying to remove this skips in #3383.

eregon:
That's even better, I'll close this then.

PR #2026, opened 2017-10-06
Update Molinillo 0.6.6 same as bundler-1.16.4

hsbt:
I hope to use the same version of Molinillo on rubygems and bundler on ruby core. In my plan, Ruby 2.5.0 is going to bundle rubygems 2.7.x and bundler 1.16.x.

PR #3388, opened 2020-03-15
Bump vendored molinillo of rubygems to 0.7.0

deivid-rodriguez:
# Description:

The problem is that rubygems uses a different version of molinillo than bundler.

So should get those in sync so that we can eventually use a single resolver.

This PR upgrades rubygems molinillo to the latest molinillo's master, which is what bundler currently uses.

Supersedes https://github.com/rubygems/rubygems/pull/2026.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -53,6 +53,7 @@ begin
   require "automatiek"
 
   Automatiek::RakeTask.new("molinillo") do |lib|
+    lib.version = "0.7.0"
     lib.download = { :github => "https://github.com/CocoaPods/Molinillo" }
     lib.namespace = "Molinillo"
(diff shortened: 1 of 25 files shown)


deivid-rodriguez:
I managed to get specs passing on my system, but there still seems to be some ruby version related issues. I'll investigate it.

voxik:
@deivid-rodriguez you are my hero. Now I wonder, when will Bundler start to use the `Gem::Resolver::Molinillo`?

deivid-rodriguez:
I guess we could start now :shrug:. Like "if rubygems is using the molinillo version bundler needs, use that, otherwise, use the vendored copy".

PR #3390, opened 2020-03-16
Add `bundle fund` command 

gjtorikian:
Migrated from https://github.com/rubygems/bundler/pull/7668

* * *

This PR implements the `bundle fund` proposal in [this accepted RFC](https://github.com/rubygems/rfcs/blob/aa031d3774a4ce6cb06d317ab03ea5cd161da498/text/0000-add-funding-uri.md).

The rationale and technical details for the `bundle fund` command are explained in the RFC, but roughly:

* A new metadata key, `funding_uri`, can be specified, that points to a funding page for a gem
* On a successful `bundle install` or `bundle update`, a message will be printed out that indicates gems which require support: `4 gems you depend on are looking for funding`
* `bundle fund` will list out all those gems, with their funding links
* `bundle info` will include the `funding_uri` in its printout
* You can restrict `bundle fund` to only list information for a specific group
* Dependencies of dependencies are currently not listed in `bundle fund`; this is [by design](https://github.com/rubygems/rfcs/blob/aa031d3774a4ce6cb06d317ab03ea5cd161da498/text/0000-add-funding-uri.md#unresolved-questions)

### DIFF
--- a/bundler/lib/bundler/cli/common.rb
+++ b/bundler/lib/bundler/cli/common.rb
@@ -14,6 +14,20 @@ module Bundler
       Bundler.ui.info msg
     end
 
+    def self.output_fund_metadata_summary
+      definition = Bundler.definition
+      current_dependencies = definition.requested_dependencies
+      current_specs = definition.specs
+
+      count = current_dependencies.count {|dep| current_specs[dep.name].first.metadata.key?("funding_uri") }
+
+      return if count.zero?
+
+      intro = count > 1 ? "#{count} installed gems you directly depend on are" : "#{count} installed gem you directly depend on is"
+      message = "#{intro} looking for funding.\n  Run `bundle fund` for details"
+      Bundler.ui.info message
+    end
+
     def self.output_without_groups_message(command)
       return if Bundler.settings[:without].empty?
       Bundler.ui.confirm without_groups_message(command)
--- a/bundler/lib/bundler/cli/fund.rb
+++ b/bundler/lib/bundler/cli/fund.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Bundler
+  class CLI::Fund
+    attr_reader :options
+
+    def initialize(options)
+      @options = options
+    end
+
+    def run
+      Bundler.definition.validate_runtime!
+
+      groups = Array(options[:group]).map(&:to_sym)
+
+      deps = if groups.any?
+        Bundler.definition.dependencies_for(groups)
+      else
+        Bundler.definition.current_dependencies
+      end
+
+      fund_info = deps.each_with_object([]) do |dep, arr|
+        spec = Bundler.definition.specs[dep.name].first
+        if spec.metadata.key?("funding_uri")
+          arr << "* #{spec.name} (#{spec.version})\n  Funding: #{spec.metadata["funding_uri"]}"
+        end
+      end
+
+      if fund_info.empty?
+        Bundler.ui.info "None of the installed gems you directly depend on are looking for funding."
+      else
+        Bundler.ui.info fund_info.join("\n")
+      end
+    end
+  end
+end
--- a/bundler/spec/commands/fund_spec.rb
+++ b/bundler/spec/commands/fund_spec.rb
@@ -1,0 +1,55 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle fund" do
+  it "prints fund information for all gems in the bundle" do
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'has_metadata'
+      gem 'has_funding'
+      gem 'rack-obama'
+    G
+
+    bundle "fund"
+
+    expect(out).to include("* has_metadata (1.0)\n  Funding: https://example.com/has_metadata/funding")
+    expect(out).to include("* has_funding (1.2.3)\n  Funding: https://example.com/has_funding/funding")
+    expect(out).to_not include("rack-obama")
+  end
+
+  it "does not consider fund information for gem dependencies" do
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'gem_with_dependent_funding'
+    G
+
+    bundle "fund"
+
+    expect(out).to_not include("* has_funding (1.2.3)\n  Funding: https://example.com/has_funding/funding")
+    expect(out).to_not include("gem_with_dependent_funding")
+  end
+
+  it "prints message if none of the gems have fund information" do
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'rack-obama'
+    G
+
+    bundle "fund"
+
+    expect(out).to include("None of the installed gems you directly depend on are looking for funding.")
+  end
+
+  describe "with --group option" do
+    it "prints fund message for only specified group gems" do
+      install_gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+        gem 'has_metadata', :group => :development
+        gem 'has_funding'
+      G
+
+      bundle "fund --group development"
+      expect(out).to include("* has_metadata (1.0)\n  Funding: https://example.com/has_metadata/funding")
+      expect(out).to_not include("* has_funding (1.2.3)\n  Funding: https://example.com/has_funding/funding")
+    end
+  end
+end
--- a/bundler/spec/install/gems/fund_spec.rb
+++ b/bundler/spec/install/gems/fund_spec.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle install" do
+  context "with gem sources" do
+    context "when gems include a fund URI" do
+      it "displays the plural fund message after installing" do
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'has_metadata'
+          gem 'has_funding'
+          gem 'rack-obama'
+        G
+
+        expect(out).to include("2 installed gems you directly depend on are looking for funding.")
+      end
+
+      it "displays the singular fund message after installing" do
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'has_funding'
+          gem 'rack-obama'
+        G
+
+        expect(out).to include("1 installed gem you directly depend on is looking for funding.")
+      end
+    end
+
+    context "when gems do not include fund messages" do
+      it "does not display any fund messages" do
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem "activesupport"
+        G
+
+        expect(out).not_to include("gem you depend on")
+      end
+    end
+
+    context "when a dependency includes a fund message" do
+      it "does not display the fund message" do
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'gem_with_dependent_funding'
+        G
+
+        expect(out).not_to include("gem you depend on")
+      end
+    end
+  end
+
+  context "with git sources" do
+    context "when gems include fund URI" do
+      it "displays the fund message after installing" do
+        build_git "also_has_funding" do |s|
+          s.metadata = {
+            "funding_uri" => "https://example.com/also_has_funding/funding",
+          }
+        end
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.0")}'
+        G
+
+        expect(out).to include("1 installed gem you directly depend on is looking for funding.")
+      end
+
+      it "displays the fund message if repo is updated" do
+        build_git "also_has_funding" do |s|
+          s.metadata = {
+            "funding_uri" => "https://example.com/also_has_funding/funding",
+          }
+        end
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.0")}'
+        G
+
+        build_git "also_has_funding", "1.1" do |s|
+          s.metadata = {
+            "funding_uri" => "https://example.com/also_has_funding/funding",
+          }
+        end
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.1")}'
+        G
+
+        expect(out).to include("1 installed gem you directly depend on is looking for funding.")
+      end
+
+      it "displays the fund message if repo is not updated" do
+        build_git "also_has_funding" do |s|
+          s.metadata = {
+            "funding_uri" => "https://example.com/also_has_funding/funding",
+          }
+        end
+        gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.0")}'
+        G
+
+        bundle :install
+        expect(out).to include("1 installed gem you directly depend on is looking for funding.")
+
+        bundle :install
+        expect(out).to include("1 installed gem you directly depend on is looking for funding.")
+      end
+    end
+  end
+end
--- a/bundler/spec/update/gems/fund_spec.rb
+++ b/bundler/spec/update/gems/fund_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle update" do
+  before do
+    gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'has_metadata'
+      gem 'has_funding', '< 2.0'
+    G
+
+    bundle :install
+  end
+
+  context "when listed gems are updated" do
+    before do
+      gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem 'has_metadata'
+        gem 'has_funding'
+      G
+
+      bundle :update, :all => true
+    end
+
+    it "displays fund message" do
+      expect(out).to include("2 installed gems you directly depend on are looking for funding.")
+    end
+  end
+end
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -32,6 +32,7 @@ bundler/lib/bundler/cli/config.rb
 bundler/lib/bundler/cli/console.rb
 bundler/lib/bundler/cli/doctor.rb
 bundler/lib/bundler/cli/exec.rb
+bundler/lib/bundler/cli/fund.rb
 bundler/lib/bundler/cli/gem.rb
 bundler/lib/bundler/cli/info.rb
 bundler/lib/bundler/cli/init.rb
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -439,6 +439,14 @@ module Bundler
       Outdated.new(options, gems).run
     end
 
+    desc "fund [OPTIONS]", "Lists information about gems seeking funding assistance"
+    method_option "group", :aliases => "-g", :type => :array, :banner =>
+      "Fetch funding information for a specific group"
+    def fund
+      require_relative "cli/fund"
+      Fund.new(options).run
+    end
+
     desc "cache [OPTIONS]", "Locks and then caches all of the gems into vendor/cache"
     method_option "all",  :type => :boolean,
                           :default => Bundler.feature_flag.cache_all?,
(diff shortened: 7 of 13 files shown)


gjtorikian:
Just bumping this to hope to get some attention to it. ✌️

simi:
Overall looks good to me. I have two suggestions:

1. Once merged, promote `funding_uri` at https://guides.rubygems.org/specification-reference/#metadata as well.
2. What about to make it part of `bundle gem` gemspec template?

gjtorikian:
>     1. Once merged, promote `funding_uri` at https://guides.rubygems.org/specification-reference/#metadata as well.

Sure, and individual `rubygems.org/gems` pages too.

>     2\. What about to make it part of `bundle gem` gemspec template?

From [the RFC](https://github.com/rubygems/rfcs/pull/23), I didn't want to do this. Maybe in a future iteration?

gjtorikian:
Hey, just bumping this...again.

deivid-rodriguez:
@gjtorikian I'm so sorry nobody stepped in to review this before. Thanks so much for being so patient.

I'm going to put this on my list for this week. If I haven't reviewed this by next Monday, please ping me again.

deivid-rodriguez:
@gjtorikian Could you rebase this PR, it's 1289 commits behind current HEAD, I'd like to make sure it's green before reviewing. I can also do that for you as well if you're busy.

gjtorikian:
@deivid-rodriguez Ok, I've rebased.

deivid-rodriguez:
Tests are failing. It's because we removed the `bundle!` test helper. You should use `bundle` now.

gjtorikian:
Ah, ok. Done!

deivid-rodriguez:
Hello @gjtorikian!

I just read the RFC, it looks really really good :heart:. I agree with everything, including the potentially most controversial parts like not adding an `opt-out` flag to skip the `<n> gems you depend on are looking for funding` message. It's just a small unintrusive line, I don't think people will be bothered by it.

Regarding the message, what do you think about tweaking it to "<n> gems you **directly** depend on are looking for funding". That way we give some clue that they might be the only code you depend on that it's looking for funding.

Regarding the functionality, it looks all good other than that little tweak. I'll go ahead with reviewing the implementation now.

gjtorikian:
@deivid-rodriguez That makes complete sense to me, and I've tweaked the message. I also removed the `!` as it felt a bit too enthusiastic.

bundler/spec/update/gems/fund_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:config) {}
+
+  before do
+    gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'has_metadata'
+      gem 'has_funding', '< 2.0'
+    G
+
+    bundle "config set #{config}" if config
```
deivid-rodriguez, line 13: This `config` seems unused? Maybe initially you meant to make this configurable?

bundler/spec/update/gems/fund_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    bundle :install
+  end
+
+  shared_examples "a fund message outputter" do
+    it "should display fund message for updated gems" do
+      expect(out).to include("2 gems you directly depend on are looking for funding.")
+    end
+  end
```
deivid-rodriguez, line 22: It seems like it's better to move this shared spec inline since there's only one spec using it?

deivid-rodriguez: So, this will repeat the same message every time `bundle update` is run, just like if `bundle install` had been run, right? I was thinking if this should only inform about funding of gems that were explicitly updated. But it sounds tricky and potentially misleading. This seems fine.

bundler/spec/install/gems/fund_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+
+        bundle :install
+        expect(out).not_to include("gem you depend on")
+      end
+    end
+  end
+
+  context "with git sources" do
+    context "when gems include fund URI" do
+      it "should display the fund URI after installing" do
```
deivid-rodriguez, line 57:
```suggestion
      it "should display the fund message after installing" do
```

bundler/spec/install/gems/fund_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.0")}'
+        G
+
+        bundle :install
+        expect(out).to include("1 gem you directly depend on is looking for funding.")
+      end
+
+      it "should display the fund URI if repo is updated" do
```
deivid-rodriguez, line 72:
```suggestion
      it "should display the fund message if repo is updated" do
```

bundler/spec/install/gems/fund_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'also_has_funding', :git => '#{lib_path("also_has_funding-1.1")}'
+        G
+        bundle :install
+
+        expect(out).to include("1 gem you directly depend on is looking for funding.")
+      end
+
+      it "should still display the fund URI if repo is not updated" do
```
deivid-rodriguez, line 98:
```suggestion
      it "should still display the fund message if repo is not updated" do
```

deivid-rodriguez (commented):
Made a first pass. Looking good!

gjtorikian (commented):
Thanks @deivid-rodriguez ! Everything made sense and I've made the changes.

gjtorikian:
So many test issues 🙈! But all good now.

bundler/lib/bundler/cli/common.rb
```diff
@@ -14,6 +14,22 @@ def self.print_post_install_message(name, msg)
       Bundler.ui.info msg
     end
 
+    def self.output_fund_metadata_summary(gems_seeking_funding)
```
deivid-rodriguez, line 17: Actually what I meant is that this parameter is not really needed, so I think we can remove it plus all the changes in `bundler/lib/bundler/installer.rb`, and `bundler/lib/bundler/paralell_installer.rb`.

bundler/lib/bundler/cli/common.rb
```diff
@@ -14,6 +14,22 @@ def self.print_post_install_message(name, msg)
...
+      current_dependencies = Bundler.definition.current_dependencies.map(&:name)
+      current_specs = Bundler.definition.specs
+
+      count = current_dependencies.count do |dep|
+        return 0 if (dep = current_specs[dep].first).nil?
```
deivid-rodriguez, line 22: Can this `nil?` actually happen? Also, even if it could happen, I guess you wouldn't want to early `return 0`, but instead make the result of the block invokation to be `false`? If we can't find a test case where this is needed, I'd rather to make the `nil?` check at all.

bundler/lib/bundler/cli/fund.rb
```diff
@@ -0,0 +1,36 @@
...
+      groups = Array(options[:group]).map(&:to_sym)
+
+      deps = if groups.any?
+        Bundler.definition.current_dependencies.select {|d| (d.groups & groups).any? }
+      else
+        Bundler.definition.current_dependencies
+      end
+
+      fund_info = deps.each_with_object([]) do |dep, arr|
+        spec = Bundler.definition.specs[dep.name].first
```
deivid-rodriguez, line 23: I copied my `count` block above from here, and you're not checking for `spec.nil?` here. I guess it can't be `nil`.

bundler/lib/bundler/cli/fund.rb
```diff
@@ -0,0 +1,36 @@
...
+      @options = options
+    end
+
+    def run
+      Bundler.definition.validate_runtime!
+
+      groups = Array(options[:group]).map(&:to_sym)
+
+      deps = if groups.any?
+        Bundler.definition.current_dependencies.select {|d| (d.groups & groups).any? }
```
deivid-rodriguez, line 17: I think this is equivalent to `Bundler.definition.dependencies_for(groups)`.

bundler/spec/commands/fund_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      gem 'rack-obama'
+    G
+
+    bundle "fund"
+
+    expect(out).to include("None of the gems you directly depend on are looking for funding.")
+  end
+
+  describe "with --group option" do
+    it "should print fund message for only specified group gems" do
```
deivid-rodriguez, line 43: Super picky comment, but I prefer to use a "present tense" style like the other specs in this file rather than a "should style". So I'd go with

```suggestion
    it "prints fund message for only specified group gems" do
```

bundler/spec/install/gems/fund_spec.rb
```diff
@@ -0,0 +1,117 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle install" do
+  context "with gem sources" do
+    context "when gems include a fund URI" do
+      it "should display the plural fund message after installing" do
```
deivid-rodriguez, line 6: Same thing about should vs present style for this spec and the others in this file.

bundler/spec/install/gems/fund_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+        gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem 'has_metadata'
+          gem 'has_funding'
+          gem 'rack-obama'
+        G
+
+        bundle :install
```
deivid-rodriguez, line 14: Unless you want to be explicit for a readabilty reason (maybe in the last spec in this file it makes sense to keep them separate), use `install_gemfile` over `gemfile` +  `bundle :install`.

bundler/spec/update/gems/fund_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle update" do
+  let(:config) {}
```
deivid-rodriguez, line 4: This `let` can go too.

deivid-rodriguez (commented):
I made another pass, mainly little style issues now.

bundler/spec/update/gems/fund_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+      gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem 'has_metadata'
+        gem 'has_funding'
+      G
+
+      bundle :update, :all => true
+    end
+
+    it "should display fund message for updated gems" do
```
deivid-rodriguez, line 27: Let's say

```suggestion
    it "displays fund message" do
```

since the current wording makes it unclear whether information is displayed for all gems or only the ones explicitly updated.

bundler/spec/update/gems/fund_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem 'has_metadata'
+      gem 'has_funding', '< 2.0'
+    G
+
+    bundle :install
+  end
+
+  context "when listed gem is updated" do
```
deivid-rodriguez, line 16: I see this is also coming from `spec/update/gems/post_install_spec.rb` but it doesn't make sense since `bundle update` is being run for all gems, not with a specific gem argument. So we could say

```suggestion
  context "when gems are updated" do
```

or something like that.

bundler/lib/bundler/cli/common.rb
gjtorikian: Whoops!

bundler/lib/bundler/cli/common.rb
gjtorikian: It can `nil`, but only in the test suites—I had tests failing without this check. I no longer have access to those runs, but perhaps they've been kept in the GitHub Actions logs? I guess what I could/should have done was to just fix the test suite. What do you think?

gjtorikian:
@deivid-rodriguez Ok, I have figured out what was going on with that `nil` check. Here is the test which was causing the issue:

https://github.com/rubygems/rubygems/blob/5d2b50b02b73938611754602798522a8eba71447/bundler/spec/commands/exec_spec.rb#L271-L286

Well, it turns out, `current_dependencies` will _not_ respect that `--without`, while `specs` does: https://github.com/gjtorikian/rubygems/blob/2517d795f61057b64f7daa391e94b91c2fe6ca4f/bundler/lib/bundler/cli/common.rb#L18-L21

So the `current_dependencies` array we iterate over has more values than the `current_specs` list, because of the groups not excluded.

To fix this, I looked around the code, and honestly the easiest solution was to move `requested_dependencies` out of `private` and into `public`. `requested_specs` was already public, so I see no reason for `requested_dependencies` to have been kept public (other than perhaps its accidental inclusion under `private`--thanks Ruby!). [Here is the blame where `requested_dependencies` was made `private`](https://github.com/gjtorikian/rubygems/blame/1cdc4f03818275840eb4d575e25c95642a91dc1f/bundler/lib/bundler/definition.rb#L907) and [the commit](8307823d80925a1393a19dcf9c4eebf7f9554208) made over 10 years ago doesn't appear to have any reason for it to be private, let alone a strong one.

If this is unacceptable (changes to the API for example) I could do some sort of `groups  = Array(options[:group]).map(&:to_sym)` manipulation but it would probably be messier and not DRY.

gjtorikian:
Ah yeah, seems that the `older-rubygems-bundler / older_rubygems_bundler (2.3.8, v2.6.14)` suite didn't like that. I'll pause here and wait for some guidance on how to handle it, then. Too many force pushes already. 🤪

bundler/lib/bundler/cli/common.rb
deivid-rodriguez: Mmm, not sure, maybe undo the change and let's see what fails now.

deivid-rodriguez:
Ok, I'll pick this up tomorrow again and check this!

bundler/lib/bundler/installer.rb
```diff
@@ -15,7 +15,7 @@ class << self
       Installer.ambiguous_gems = []
     end
 
-    attr_reader :post_install_messages
+    attr_reader :post_install_messages, :fund_metadata
```
deivid-rodriguez, line 18: All the changes in this file can go now, right?

bundler/lib/bundler/installer/parallel_installer.rb
```diff
@@ -6,12 +6,13 @@
 module Bundler
   class ParallelInstaller
     class SpecInstallation
-      attr_accessor :spec, :name, :post_install_message, :state, :error
+      attr_accessor :spec, :name, :post_install_message, :fund_metadata, :state, :error
```
deivid-rodriguez, line 9: Same with all the changes in this file.

bundler/lib/bundler/cli/fund.rb
```diff
@@ -0,0 +1,36 @@
...
+      @options = options
+    end
+
+    def run
+      Bundler.definition.validate_runtime!
+
+      groups = Array(options[:group]).map(&:to_sym)
+
+      deps = if groups.any?
+        Bundler.definition.dependencies_for(groups)
```
deivid-rodriguez, line 17: Tests failed because I suggested this method which is also private, I guess it's fine to make it public too.

deivid-rodriguez (commented):
This feels like almost ready now!

bundler/lib/bundler/installer.rb
gjtorikian: Whoops, yes, that was lazy of me. 🙈

bundler/lib/bundler/cli/fund.rb
gjtorikian: I think so too; `specs_for` is already also public.

deivid-rodriguez:
That `nil` is still causing issues. I'll clone the repo later and figure it out.

deivid-rodriguez:
Hello again @gjtorikian! I had a look at this. The problem is that `Bundler.definition.requested_specs` returns "not materialized" specs. That essentially means specs not relevant for the current platform are not filtered out. Your initial implementation indirectly used `Bundler.definition.specs` which materializes all the specs relevant to the requested groups:

https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/definition.rb#L180

So I think changing `Bundler.definition.requested_specs` to `Bundler.definition.specs` should get things back to working.

gjtorikian:
@deivid-rodriguez Looks like that did it!

bundler/lib/bundler/cli/common.rb
```diff
@@ -14,6 +14,19 @@ def self.print_post_install_message(name, msg)
       Bundler.ui.info msg
     end
 
+    def self.output_fund_metadata_summary
+      current_dependencies = Bundler.definition.requested_dependencies.map(&:name)
```
deivid-rodriguez, line 18: Another nit, but for simmetry, could we avoid the `map` here, and instead do `dep.name` inside the `count` block?

deivid-rodriguez: Also, I'm wondering whether we should actually use `requested_dependencies` here, since it kind of makes the message inaccurate because of taking groups into account. So, if your current environment is using, say, `without production`, it will omit to include the `production` gems in the count. But technically you still depend on those gems (they're pretty critical in this case, actually :sweat_smile:).

deivid-rodriguez (commented):
Another two minor comments.

bundler/lib/bundler/cli/common.rb
gjtorikian: > Also, I'm wondering whether we should actually use `requested_dependencies` here, since it kind of makes the message inaccurate because of taking groups into account.

I think I agree with you, and made that change!

gjtorikian:
I have no idea why Windows builds are now broken. 😞

deivid-rodriguez:
It sounds like a similar problem as before of not filtering out specs not relevant to the platform. If instead of 928591a5fe31bba05122d660f82cea6235679b8c, you change `current_dependencies = Bundler.definition.requested_dependecies` with `current_dependencies = Bundler.definition.current_dependencies`, does it work?

Also, could you rebase again since you are at it?

gjtorikian:
> does it work?

It does not. I can reproduce the same `undefined method `metadata' for nil:NilClass` locally (on macOS). Perhaps `specs` does not take platforms into consideration? 

> Also, could you rebase again since you are at it?

Yes, I'll do this once we're all green.

deivid-rodriguez:
mmm. Interesting. I cloned the PR, made the modification and run the relevant specs on Linux and they passed. Theorically `definition.specs` should filter out specs not relevant for the current platform, and avoid `nil` results. Can you still push the modification so that I can see the failures?

deivid-rodriguez:
What if using the same reasoning as with groups before, we don't do any platform filtering either? Maybe

```
current_dependencies = Bundler.definition.dependencies
current_specs = Bundler.definition.resolve
```

As you can see, this is all a bit of a mess :sweat_smile:.

gjtorikian:
> Can you still push the modification so that I can see the failures?

Ok, I pushed it!

deivid-rodriguez:
Oh wow, is there anything my suggestion doesn't break? :laughing:

Ok, let me check this.

deivid-rodriguez:
Ok @gjtorikian, sorry for the back and forth here.

The problem is that `specs` already does the filtering of excluding not configured groups, so we can't get the funding information for those from there. We could try `resolve.materialize(current_dependencies)` instead to get _all_ materialized specs (without any group filtering), but in order for that to work, it turns out we'd need to actually install those specs, and that didn't happen because groups were filtered out in the first place.

So, since I don't see an easy fix for this, I propose we go back to https://github.com/rubygems/rubygems/pull/3390#discussion_r495939587 and address that instead by further tweaking the message to

> \<n\> _installed_ gems you directly depend on are looking for funding.

to make it more clear that any gem exclusions happening before gem installation, for example `without production`, are not included in the count.

gjtorikian:
🎆

deivid-rodriguez (approved):
Hi!

I'm aware that at this point I'm probably just being annoying, but technically the commit that needs `requested_dependencies` to be made public is the second, so the change making that method public belongs in there :sweat_smile:.

Other than that, I'd like to thank you so much for your infinite patience and for your polite well-spaced pings along these past months :purple_heart:.

gjtorikian:
Heh, no worries. It's fixed up!

I am just excited to be able to add this and get some visibility for support. I have another change pending for visualizing this metadata on rubygems.org next. 😄

deivid-rodriguez:
Hei, that's awesome to hear. This will make a better ecosystem overall.

Regarding my last request, I don't think it's actually fixed because `bundle fund` requires `dependencies_for` to be made public, so that part should go with the first commit, while the post_install method requires `requested_dependencies` to be made public, so that part should be included in the second commit. The reason I ask for this is not only for a better git history, but also if I ever need to bisect something, I'd like each individual commit to pass specs individually.

gjtorikian:
> The reason I ask for this is not only for a better git history, but also if I ever need to bisect something, I'd like each individual commit to pass specs individually

It's all good! It's your project, I'm just committing to it. 😆 

Re-rebased and pushed.

deivid-rodriguez:
Awesome, thanks so much @gjtorikian!! :tada:

simi:
:partying_face:

gjtorikian:
As a maintainer I know how problematic the next question is but I am compelled to ask: when can we expect a new release with these changes? 🙈

deivid-rodriguez:
We hope to release bundler 2.2.0.rc.2 next week including this feature.

PR #3387, opened 2020-03-15
Make `bundler config --local` docs clearer

kevinburkemeter:
The previous references to "app" could easily be confused with the
"app" directory of a Rails project, leading people to look in the
wrong place for Bundler configuration. Instead use a variable that
should make it clearer that we mean the top level of the project.

Fixes #3385.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3391, opened 2020-03-16
Deterministic man txt pages

deivid-rodriguez:
# Description:

Sometimes contributors get discrepancies in the generated txt man output, depending on the `groff` version being used, and that makes our man pages checks fail.

This PR tries to make the output more consistent.

As a side effect, it actually reads better.

References https://github.com/rubygems/rubygems/pull/3387#issuecomment-599261526.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -120,7 +120,7 @@ namespace :man do
         end
 
         file "#{roff}.txt" => roff do
-          sh "groff -wall -mtty-char -mandoc -Tascii #{roff} | col -xb > #{roff}.txt"
+          sh "groff -Wbreak -mtty-char -mandoc -Tascii -rHY=0 #{roff} | col -xb > #{roff}.txt"
         end
 
         task :build_all_pages => "#{roff}.txt"
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -116,7 +116,7 @@ namespace :man do
         index << [ronn, File.basename(roff)]
 
         file roff => ["man", ronn] do
-          sh "bin/ronn --roff --pipe --date #{Time.now.strftime("%Y-%m-%d")} #{ronn} > #{roff}"
+          sh "bin/ronn --warnings --roff --pipe --date #{Time.now.strftime("%Y-%m-%d")} #{ronn} > #{roff}"
         end
 
         file "#{roff}.txt" => roff do
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -120,7 +120,7 @@ namespace :man do
         end
 
         file "#{roff}.txt" => roff do
-          sh "groff -Wall -mtty-char -mandoc -Tascii #{roff} | col -b > #{roff}.txt"
+          sh "groff -wall -mtty-char -mandoc -Tascii #{roff} | col -b > #{roff}.txt"
         end
 
         task :build_all_pages => "#{roff}.txt"
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -120,7 +120,7 @@ namespace :man do
         end
 
         file "#{roff}.txt" => roff do
-          sh "groff -wall -mtty-char -mandoc -Tascii #{roff} | col -b > #{roff}.txt"
+          sh "groff -wall -mtty-char -mandoc -Tascii #{roff} | col -xb > #{roff}.txt"
         end
 
         task :build_all_pages => "#{roff}.txt"
--- a/bundler/man/bundle-add.1.txt
+++ b/bundler/man/bundle-add.1.txt
@@ -1,4 +1,4 @@
-BUNDLE-ADD(1)							 BUNDLE-ADD(1)
+BUNDLE-ADD(1)                                                    BUNDLE-ADD(1)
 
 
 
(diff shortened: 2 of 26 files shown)


kevinburkemeter:
FYI, I'm using GNU groff on Mac and get this diff. Not clear why the escape sequences are being printed, it's frustrating.

```
-NAME
-       bundle-add - Add gem to the Gemfile and run bundle install
+1mNAME0m
+       1mbundle-add 22m- Add gem to the Gemfile and run bundle install

-SYNOPSIS
-       bundle     add     GEM_NAME     [--group=GROUP]     [--version=VERSION]
+1mSYNOPSIS0m
+       1mbundle     add     4m22mGEM_NAME24m     [--group=GROUP]     [--version=VERSION]
        [--source=SOURCE]   [--git=GIT]   [--branch=BRANCH]    [--skip-install]
        [--strict] [--optimistic]
```

kevinburkemeter:
Also not sure whether that's coming from "col" or "groff" - I tried compiling "col" on my Mac from the linux-utils source to remove the diff and no dice.

deivid-rodriguez:
Those look like color escape sequences, could you edit the `rake` task to pass the `-c` flag to `groff` (which according to my man page) and see if it helps?

deivid-rodriguez:
That difference doesn't show up in https://github.com/rubygems/rubygems/pull/3387/files though, so I should assume this PR is fixing the whitespace discrepancies?

kevinburkemeter:
I can get it closer to what you have by setting GROFF_NO_SGR=true in the environment before running. Still I'm confused by this diff. Maybe I need to pass an option that says don't hyphenate words.

```
-BUNDLE-ADD(1)							 BUNDLE-ADD(1)
+BUNDLE-ADD(1)                                                    BUNDLE-ADD(1)



@@ -6,13 +6,13 @@ NAME
        bundle-add - Add gem to the Gemfile and run bundle install

 SYNOPSIS
-       bundle	  add	  GEM_NAME     [--group=GROUP]	   [--version=VERSION]
-       [--source=SOURCE]   [--git=GIT]	 [--branch=BRANCH]    [--skip-install]
+       bundle     add     GEM_NAME     [--group=GROUP]     [--version=VERSION]
+       [--source=SOURCE]   [--git=GIT]   [--branch=BRANCH]    [--skip-install]
        [--strict] [--optimistic]

 DESCRIPTION
-       Adds  the  named  gem  to  the  Gemfile	and run bundle install. bundle
-       install can be avoided by using the flag --skip-install.
+       Adds  the  named  gem to the Gemfile and run bundle install. bundle in-
+       stall can be avoided by using the flag --skip-install.

        Example:

@@ -29,30 +29,30 @@ DESCRIPTION

 OPTIONS
        --version, -v
-	      Specify version requirements(s) for the added gem.
+              Specify version requirements(s) for the added gem.

        --group, -g
-	      Specify  the  group(s) for the added gem. Multiple groups should
-	      be separated by commas.
+              Specify  the  group(s) for the added gem. Multiple groups should
+              be separated by commas.

        --source, , -s
-	      Specify the source for the added gem.
+              Specify the source for the added gem.
```

kevinburkemeter:
The difference comes because my program is replacing a tab with a space character, and the checked in output deletes the tab characters altogther.

deivid-rodriguez:
> The difference comes because my program is replacing a tab with a space character, 

That is exactly what this PR is doing, so it should fix the problem in that case! :)

kevinburkemeter:
OK. Now the only difference is my program is inserting hyphens and yours isn't. I can't see why that's happening. I asked https://stackoverflow.com/q/60712000/329700 in the hopes of finding an answer.

kevinburkemeter:
Here's what I'm running

```
$ groff --version
GNU groff version 1.22.4
Copyright (C) 2018 Free Software Foundation, Inc.
GNU groff comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of groff and its subprograms
under the terms of the GNU General Public License.
For more information about these matters, see the file
named COPYING.

called subprograms:

GNU grops (groff) version 1.22.4
GNU troff (groff) version 1.22.4
```

kevinburkemeter:
Aha, it's entirely the difference between groff 1.22.3 (the default on Ubuntu) and 1.22.4. I compiled Groff 1.22.3 on my Mac and it's not generating a diff anymore.

kevinburkemeter:
ok, LGTM

kevinburkemeter:
oh, I still think it is a good idea to enable `--warnings` in ronn, this is so finicky that it seems better to be able to see things than to not see them.

deivid-rodriguez:
Nice research! Thanks for documenting all that :purple_heart:.

> oh, I still think it is a good idea to enable --warnings in ronn, this is so finicky that it seems better to be able to see things than to not see them.

Yep, that's what I'm doing in this PR too. Previously they were ignored.

kevinburkemeter:
I don't see it? looks like this line is unchanged

```ruby
          sh "bin/ronn --roff --pipe --date #{Time.now.strftime("%Y-%m-%d")} #{ronn} > #{roff}"
```

deivid-rodriguez:
Oh, sorry, I meant in `groff` (what converts `.roff` to `.txt`), where I changed `-Wall` (ignore all warnings) to `-wall` (show all warnings).

We can definitely enable warnings in `ronn` too, I'll add that here :+1:.

deivid-rodriguez:
I think the hyphenation discrepancy could be fixed by adding a line with `.nh` right after the `.TH` line on every man page. Could be added as a `ronn` feature or do it ourselves.

deivid-rodriguez:
I _think_ I figured it out thanks to https://stackoverflow.com/questions/22712319/run-aspell-on-manpage#22713606.

With this, I think we might fully eliminate the discrepancies between `groff` versions?

Also, the new output is much better because hyphenation was also applied to actual shell commands, and that could confuse some people.

deivid-rodriguez:
Let's do it! Hopefully this makes txt man pages more readable and eases documentation contributions.

PR #3396, opened 2020-03-17
Removed self-hosted YAML Serializer and PsychedYAML

hsbt:
# Description:

Migrate from https://github.com/rubygems/bundler/pull/6536

I hope to unify the specification of YAML serializer in Rubygems. But We have a some of the activated issues related to the default gems.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -641,12 +641,11 @@ EOF
     private
 
     def eval_yaml_gemspec(path, contents)
-      require_relative "bundler/psyched_yaml"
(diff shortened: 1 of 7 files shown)


simi:
@hsbt I have rebased this (also to restart CI since old logs were not available anymore). I hope that is fine.

Would it be possible to remove `lib/bundler/psyched_yaml.rb` as well?

hsbt:
I gave up to migrate self-hosted YAML serializer to `psych` gem. Because we keep to specify any version with Gemfile. I have no idea to do it with psych as the default gems.

Issue #3089, opened 2020-01-19
Better handling of the `required_ruby_version` constraint in gemspecs


deivid-rodriguez:
This issue is completely equivalent to #2575, but for the `required_ruby_version` constraint.

As an example, using the latest version of `rubygems`, running `gem install rails` on `ruby-2.4.9` will fail like this:

```console
$ gem install rails
Fetching sprockets-4.0.0.gem
Fetching rack-test-1.1.0.gem
ERROR:  Error installing rails:
	There are no versions of activesupport (= 6.0.2.1) compatible with your Ruby & RubyGems. Maybe try installing an older version of the gem you're looking for?
	activesupport requires Ruby version >= 2.5.0. The current ruby version is 2.4.9.362.
```

Instead, `rubygems` should install the latest version of `rails` compatible with ruby 2.4.9.

There's a WIP PR for this at https://github.com/rubygems/rubygems/pull/2655, but still needs work.

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
3.1.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

MSP-Greg:
@deivid-rodriguez

I thought there was agreement that the best way to implement this was to change RG over to the 'API' data used by Bundler, and drop the older (inefficient) API used by RG.  Especially with Bundler now integrated.

Am I mistaken?  It's been a while...

deivid-rodriguez:
Mmmm, what do you mean? I just opened a ticket about a rubygems issue, I didn't suggest any implementations for the solution.

deivid-rodriguez:
I do agree that's the best way to fix it, yeah.

deivid-rodriguez:
Oh, I think I understand what you mean, you ask because I linked to your PR, right? Well, I think it's best to switch to the API used by bundler, but I think we can probably still reuse some of your work?

MSP-Greg:
> you ask because I linked to your PR, right?

Yes.

> I think it's best to switch to the API used by bundler

Totally agree.

> I think we can probably still reuse some of your work?

I think when I did the PR, I was unaware of the Bundler API.

[...]

Issue #2575, opened 2019-01-04
Better handling of the `required_rubygems_version` constraint in gemspecs


deivid-rodriguez:
My current problem is that the `required_rubygems_version` constraint in gemspec does not have a user friendly fallback behavior.

Currently the `gem` CLI will try to install the latest available version, and fail if the running rubygems version does not satisfy the `required_rubygems_version` constraint set by that version's gemspec.

A better behavior would be that rubygems automatically selects the newest version of the gem that satisfies the running rubygems version, and does not fail at all.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
PASTE HERE
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

MSP-Greg:
As mentioned in PR #2616, I have code that partially addresses this issue.  Today I noticed there are two remaining issues, `gem update` & `gem outdated`.  I don't think `gem install` requires changes, as I used for real-world testing.

The current  `gem update` makes use of NameTuples, I assume in an attempt to speed up the process.  

This isn't helpful with the `required_rubygems_version` constraint, so the code needs to be rewritten.

1. So, assuming I'm going to be the one fixing `Gem::Commands::UpdateCommand`, does anyone a have any issues with making it a subclass of `Gem::Commands::InstallCommand` with changes as needed?   The changes hopefully only  involve the 'initial dependencies', along with filtering the result set.  I think...

2. Currently console output includes info about gems that are not updated. Should that info be included when using the `--explain` option?

3. Past issues have mentioned the lack of platform info when using `--explain`.  This is also an issue with 'update' if a gem changes platforms with update.  IOW, it may change from Gem::Platform.local to Gem::Platform::RUBY, or vice-versa.  Should `--explain` output be changed to make a platform change apparent?

Haven't looked at `Gem::Command::Outdated`, but it may only need changes to `Gem::Specification.outdated_and_latest_version`...

MSP-Greg:
Interesting that there has been very few comments about an issue with such a large effect on Ruby.

I've done a bit of work in this (both code & tests), and it's expanded quite a bit beyond the earlier PR.  

Some issues:

[...]

PR #3398, opened 2020-03-18
Drop Ruby 2.3 and RubyGems 2.5

hsbt:
# Description:

This pr is migrated from https://github.com/rubygems/bundler/pull/7588

We will merge this after 2019-03-31.
______________

# Tasks:

- [x] Write tests


deivid-rodriguez:
We should try to deal with #2575 and #3089 first, if possible.

deivid-rodriguez:
As I said, I believe we should try to address #2575 and #3089 first, so I'm :-1: on doing this right now. Something we could do as a middle ground to reduce the number of jobs is to only test against the default rubygems version when testing bundler against ruby 2.3 and ruby 2.4.

hsbt:
I understood the current behavior of `required_ruby_version` is unfriendly for users. 

But we couldn't backport the feature of `required_ruby_version` handling for Ruby 2.3 and its rubygems version. So, Ruby 2.3 users couldn't use it in the feature if we fixed its issues.

Do you have an any idea about this?

deivid-rodriguez:
Yeah, you are right. Even if we fix this, the fix won't be backported to the rubies that currently need it. So the fix will only be effective for users who are currently doing `gem update --system` in CI. Also, I expect [ruby/setup-ruby](https://github.com/ruby/setup-ruby] to have alleviated this problem.

Still, I feel this is more of a philosophical issue. When we drop support and this issue affects people, I'd like to be able to tell people: "we have fixed this in rubygems. Make sure you update rubygems before running `gem install bundler` and the issue we'll be gone" rather than "you need to drop support for ruby 2.3" or "you need to stay on an old bundler version".

Because the first alternative encourges adoption of newer versions of bundler & rubygems while the latter discourages it.

That said, I understand that if we don't manage to fix the `required_ruby_version` issue soon, we'll need to drop support at some point. My take to dropping support for old rubies requires two conditions:

* The version being dropped has reached its end of life.
* Supporting the version is getting in the middle and hurting development.

So the question is whether the answer to the second question is already "Yes". I haven't personally been bothered by supporting ruby 2.3 and ruby 2.4 (yet). The thing that might start bothering me is the excessive number of CI jobs, but the bottleneck for CI is currently Windows + ruby 2.6, and we could stop testing for _some_ rubygems+bundler combinations without needing to drop full support.

So, to sum up my opinion: I'd like to wait a few more months and reevaluate this again by then.

hsbt:
I see. This changes is not rush to merge. I hope to merge in this year at least.

deivid-rodriguez:
Alright, so I was considering whether it makes sense to drop support now, and I believe it's best to merge this PR next December right before releasing the next minor version. We could merge it now, but I think it would become counterproductive, because:

* It will complicate backports to the stable branch, that still needs to support old rubies.
* It might allow incompatible changes into master that will later have to be addressed in the stable branch at release time.

I will keep rebasing this PR once in a while, and we can move on with this right before cutting the next stable branch. How does that sound?

deivid-rodriguez:
Also, I want to proposed, since we are at it, to also drop support for 2.5 and 2.6 at the same time since they will be already unsupported by the end of this year.

hsbt:
Ruby 2.6 is EOL today. I submit another PR to remove supporting under the Ruby 2.6.

PR #3402, opened 2020-03-18
Bump vendored molinillo of bundler to 0.7.0

hsbt:
# Description:

Migrated from https://github.com/rubygems/bundler/pull/7210

/cc @deivid-rodriguez 
______________

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Thanks for migrating all these PRs @hsbt! :purple_heart:

Quoting from the previous PR, in case someone has answers:

> What's the state of current molinillo's master? Would it be possible to make a new release that the next version of bundler can use? As I recall, there were only some changes for better error reporting.

I don't like the current situation because we're using a released version of molinillo with some cherry-picks on top, which seems not ideal. That's why I proposed this PR, to at least use something that matches the master branch.

deivid-rodriguez:
This, and https://github.com/rubygems/rubygems/pull/3388, are ready now since Molinillo finally made a release :tada:.

PR #3412, opened 2020-03-22
Revert "Remove Gem::DependencyInstaller#find_gems_with_sources"

### LINKED ISSUE #3411: ruby-debug-ide won't install with the next rubygems version
The `ruby-debug-ide` gem uses `Gem::DependencyInstaller#find_gem_with_sources` in its extension:

https://github.com/ruby-debug/ruby-debug-ide/blob/02737f9d8e1c093d77c11dc1f839cd0c44f5df99/ext/mkrf_conf.rb#L5-L9

But we have completely removed this method in rubygems master in 04c79d3eb9d9803d9fae78575b125b325b97206e.

We should do something like:

* Restore the method so that ruby-debug-ide keeps installing with the next rubygems version.
* Figure out what they are doing, and what API rubygems should provide to do that.
* Implement whatever we need to implement and provide a patch to ruby-debug-ide to start using it.
* Eventually remove `Gem::DependencyInstaller#find_gem_with_sources` once all supported rubies already provide the new APIs by default.

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

```
$ gem env version
PASTE HERE
```

deivid-rodriguez:
# Description:

I noticed that a pretty popular gem, `ruby-debug-ide`, is using `Gem::DependencyInstaller#find_gems_with_sources`, so removing this method means that installation of any `Gemfile` including `ruby-debug-ide` will break in the next rubygems version.

We obviously don't want that, so I'm reverting the removal for the moment.

Final removal is postponed until next year until we find a better way to manage deprecations (#3013).

Note that the check `ruby-debug-ide` is actually broken, but we still don't want to break installing it. Hopefully, they will fix this and stop using this method soon. See https://github.com/ruby-debug/ruby-debug-ide/pull/142.

Closes #3411.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/dependency_installer.rb
+++ b/lib/rubygems/dependency_installer.rb
@@ -7,6 +7,7 @@ require 'rubygems/spec_fetcher'
 require 'rubygems/user_interaction'
 require 'rubygems/source'
 require 'rubygems/available_set'
+require 'rubygems/deprecate'
 
 ##
 # Installs a gem along with all its dependencies from local and remote gems.
@@ -14,6 +15,7 @@ require 'rubygems/available_set'
 class Gem::DependencyInstaller
 
   include Gem::UserInteraction
+  extend Gem::Deprecate
 
   DEFAULT_OPTIONS = { # :nodoc:
     :env_shebang         => false,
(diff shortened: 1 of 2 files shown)


bronzdoc:
Hey @deivid-rodriguez 😃 

Just to understand this correctly, we are reverting this because we didn't give enough time to people to be aware of the deprecation warning?

deivid-rodriguez:
I believe the gem is popular enough to justify reverting. Note that even if they fix the extension and release a new gem, any `Gemfile` locked to an older version of the extension will crash on `bundle install`.

I think @hsbt had some tools to grep for usages of code inside gems, we could run that to see if there's more usages.

deivid-rodriguez:
As an alternative, we could keep it as a "noop" returning `[]` with a warning about it not doing anything useful and just being kept for compatibility.

That would allow `ruby-debug-ide` to still work and us to remove the code.

bronzdoc:
Maybe we just should not include the commit that removed the method in the next release?

Also if we put a deprecation warning in Rubygems code and people decide to ignore it we can't really do much about it.

deivid-rodriguez:
> Maybe we just should not include the commit that removed the method in the next release?

That is what I'm doing here, reverting that commit.

> Also if we put a deprecation warning in Rubygems code and people decide to ignore it we can't really do much about it.

I don't think that a gem maintainer being a little slow (define "slow" anyways) when updating their code to be warning free on newer rubygems should affect end users that have nothing to do with maintenance of a single specific gem. This will not be breaking a single gem's maintainer workflow, this will be breaking `bundle install` for everybody using this gem.

This warning have only been out there for like, 3 months, anyways. I don't think it's enough time to just let this break. And I believe it's actually easier for us to keep this method as a noop for now, than to deal with all the issues about `bundle install` breaking after people upgraded rubygems.

hsbt:
This is the list used with `Gem::DependencyInstaller#find_gems_with_sources`. 

https://gist.github.com/hsbt/1565e70bfc41b5e6c64db791cd8df14d

I thought we should revert this because this list includes `chef` that is the famous provisioning tool.

deivid-rodriguez:
Thanks for double checking this @hsbt!

PR #3414, opened 2020-03-23
Version horizon deprecations

bronzdoc:
# Description:
First step of https://github.com/rubygems/rubygems/issues/3013

This PR makes `deprecate` and `deprecate_command` accept a date horizon and instead of a date, 
______________

### DIFF
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -65,6 +65,23 @@ module Gem::Deprecate
     end
   end
 
+  # Deprecation method to deprecate Rubygems commands
+  def deprecate_command(rubygems_version:)
+    class_eval do
+      define_method "deprecated?" do
+        true
+      end
+
+      define_method "deprecation_warning" do
+        msg = [ "#{self.command} command is deprecated",
+                ". It will be removed in Rubygems #{rubygems_version}.\n",
+        ]
+
+        alert_warning "#{msg.join}" unless Gem::Deprecate.skip
+      end
+    end
+  end
+
   ##
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -41,6 +41,10 @@ module Gem::Deprecate
     Gem::Deprecate.skip = original
   end
 
+  def self.next_rubygems_major_version # :nodoc:
+    Gem::Version.new(Gem.rubygems_version.segments.first).bump
+  end
+
   ##
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
@@ -48,9 +52,6 @@ module Gem::Deprecate
   # Rubygems version that it is planned to go away.
 
   def deprecate(name, replacement=:none)
-    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
-    next_rubygems_major_version = current_major.bump
-
     class_eval do
       old = "_deprecated_#{name}"
       alias_method old, name
@@ -59,7 +60,7 @@ module Gem::Deprecate
         target = klass ? "#{self}." : "#{self.class}#"
         msg = [ "NOTE: #{target}#{name} is deprecated",
                 replacement == :none ? " with no replacement" : "; use #{replacement} instead",
-                ". It will be removed in Rubygems #{next_rubygems_major_version}",
+                ". It will be removed in Rubygems #{Gem::Deprecate.next_rubygems_major_version}",
                 "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
         ]
         warn "#{msg.join}." unless Gem::Deprecate.skip
@@ -70,9 +71,6 @@ module Gem::Deprecate
 
   # Deprecation method to deprecate Rubygems commands
   def deprecate_command
-    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
-    next_rubygems_major_version = current_major.bump
-
     class_eval do
       define_method "deprecated?" do
         true
@@ -80,7 +78,7 @@ module Gem::Deprecate
 
       define_method "deprecation_warning" do
         msg = [ "#{self.command} command is deprecated",
-                ". It will be removed in Rubygems #{next_rubygems_major_version}.\n",
+                ". It will be removed in Rubygems #{Gem::Deprecate.next_rubygems_major_version}.\n",
         ]
 
         alert_warning "#{msg.join}" unless Gem::Deprecate.skip
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -45,6 +45,30 @@ module Gem::Deprecate
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
   # telling the user of +repl+ (unless +repl+ is :none) and the
+  #Rubygems version that it is planned to go away.
+
+  def deprecate(name:, replacement:, rubygems_version:)
+    class_eval do
+      old = "_deprecated_#{name}"
+      alias_method old, name
+      define_method name do |*args, &block|
+        klass = self.kind_of? Module
+        target = klass ? "#{self}." : "#{self.class}#"
+        msg = [ "NOTE: #{target}#{name} is deprecated",
+                repl == :none ? " with no replacement" : "; use #{replacement} instead",
+                ". It will be removed in Rubygems #{rubygems_version}",
+                "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
+        ]
+        warn "#{msg.join}." unless Gem::Deprecate.skip
+        send old, *args, &block
+      end
+    end
+  end
+
+  ##
+  # Simple deprecation method that deprecates +name+ by wrapping it up
+  # in a dummy method. It warns on each call to the dummy method
+  # telling the user of +repl+ (unless +repl+ is :none) and the
   # year/month that it is planned to go away.
 
   def deprecate(name, repl, year, month)
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -47,7 +47,10 @@ module Gem::Deprecate
   # telling the user of +repl+ (unless +repl+ is :none) and the
   # Rubygems version that it is planned to go away.
 
-  def deprecate(name:, replacement:)
+  def deprecate(name, replacement=:none)
+    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
+    next_rubygems_major_version = current_major.bump
+
     class_eval do
       old = "_deprecated_#{name}"
       alias_method old, name
@@ -55,8 +58,8 @@ module Gem::Deprecate
         klass = self.kind_of? Module
         target = klass ? "#{self}." : "#{self.class}#"
         msg = [ "NOTE: #{target}#{name} is deprecated",
-                repl == :none ? " with no replacement" : "; use #{replacement} instead",
-                ". It will be removed in Rubygems #{rubygems_version}",
+                replacement == :none ? " with no replacement" : "; use #{replacement} instead",
+                ". It will be removed in Rubygems #{next_rubygems_major_version}",
                 "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
         ]
         warn "#{msg.join}." unless Gem::Deprecate.skip
@@ -67,7 +70,9 @@ module Gem::Deprecate
 
   # Deprecation method to deprecate Rubygems commands
   def deprecate_command
-    next_rubygems_major_version = Gem.rubygems_version + 1
+    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
+    next_rubygems_major_version = current_major.bump
+
     class_eval do
       define_method "deprecated?" do
         true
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
@bronzdoc Loving this! This is WIP, right? I mean, next steps would be to change the calls to use the new implemenation of `deprecate` and `deprecate_command`.

Also, can we default to set removal date for the next major is no explicit version is passed? Actually, we might want to remove the parameter altogether because I don't think it makes sense to deprecate something two major versions in advance?

bronzdoc:
@deivid-rodriguez 

> might want to remove the parameter altogether because I don't think it makes sense to deprecate something two major versions in advance?

Good point, makes sense.

> next steps would be to change the calls to use the new implemenation of deprecate and deprecate_command.

Yeah, my idea to implement it like this was that we might want to discuss when those methods were going to be deprecated(which version), but if we are going to set the deprecation to the next major always I think I should just get rid of the "date" methods and use the new ones.

What you think?

deivid-rodriguez:
Yeah, I think removal on the next major is the safest thing to do?

lib/rubygems/deprecate.rb
```diff
@@ -41,13 +41,54 @@ def skip_during
     Gem::Deprecate.skip = original
   end
 
+  ##
+  # Simple deprecation method that deprecates +name+ by wrapping it up
+  # in a dummy method. It warns on each call to the dummy method
+  # telling the user of +repl+ (unless +repl+ is :none) and the
+  #Rubygems version that it is planned to go away.
```
olleolleolle:
```suggestion
  # Rubygems version that it is planned to go away.
```

Was that spacing intentional or a typo?

bronzdoc: Not intentional, thanks @olleolleolle !

bronzdoc:
@deivid-rodriguez so if we set the default to be always the next major(calculated with the current major) without passing a major version explicitly, we might run into the same problem we have today. If we deprecate a method/command and we forget to deprecate it in the next major release we would be stuck with it for the next release.

So:
 A: We pass the version to deprecate it explicitly and check that the version that we pass is not two majors ahead

 B: We calculate it to always be the next major but we will have to build tooling to check for deprecations before a major release

C:?

deivid-rodriguez:
Indeed!

I was actually going to propose B) as a follow up for this PR.

Isn't A) subject to the same problem? I we do `deprecate :foo, :bar, version: 4` (or whatever signature we choose), aren't we subject to releasing rubygems 4 and forgetting to remove the code, and users getting messages about "foo will be deprecated in rubygems 4" when they're already using rubygems 4? I think we'll need to build tooling anyways?

deivid-rodriguez:
Actually, I built some tooling to check the current date based deprecations:

https://github.com/rubygems/rubygems/blob/2bbcdcde08b90d4ef03da8fb1f7a8a3313e13bb8/Rakefile#L89-L100

But I set it to only be activated when releasing a major version, so it's never actually been run yet :sweat_smile:.

So we need to adapt to custom `rubocop` cop to the new deprecations. Since we're killing the date based deprecations, I think it actually makes sense to adapt that code inside this PR.

Thoughts?

bronzdoc:
@deivid-rodriguez 
> Isn't A) subject to the same problem? I we do deprecate :foo, :bar, version: 4
True, I think I was thinking on checking something like `:version >= current_rubygems_major`  inside deprecate and do just `return` in the method override instead of alerting but i guess we don't get rid of the method and it will just make things more confusing lol

> Thoughts?
Sounds good, we can adapt that to the new version horizon deprecation. Is there a benefit in making it a rubocop custom cop instead of a rake task?

deivid-rodriguez:
It's actually a `rake` task implemented under the hood as a custom cop. When I implemented it, using rubocop looked more neat to me than "regexp'ing the source code", that why I used a custom cop. Since now it's already there, I think it shouldn't be hard to adapt it. The method name will be the same, only the argument checking logic will change.

bronzdoc:
Errors are not related, seems like the bundler container can connect to `azure.archive.ubuntu.com` to 
 install `graphviz` 😬

deivid-rodriguez:
Nice @bronzdoc, it looks good to me now!

After a second though I'm not sure we even need the `rubocop` check. What if upon release, we decide that we don't want to remove yet one or two deprecations and stick with them for one more major cycle? In that case, the solution would be to leave the deprecations there, so that they also warn users after the release. So, there would be now way to get the rubocop check green and make the release in the case :sweat_smile:. 

We definitely need to add "Remove pending deprecations" to our release playbook, in the case of releasing a major, but maybe this automated check is unnecessary and might get in the middle.

In any case, this is a strict improvement over what we have now, so I'm happy to merge this and discuss the removal of the rubocop check separately.

util/cops/deprecations.rb
```diff
@@ -8,36 +8,24 @@ module Rubygems
       #
       # @example
       #
-      #   As of March, 2019
```
deivid-rodriguez: The cop is no longer about dates, so this sentence should just go away.

util/cops/deprecations.rb
```diff
@@ -8,36 +8,24 @@ module Rubygems
...
+      #   As of March, 2020
       #
       #   # bad
-      #   deprecate :safdfa, nil, 2018, 12
-      #   deprecate :safdfa, nil, 2019, 03
+      #   deprecate :safdfa, :none
       #
       #   # good
-      #   deprecate :safdfa, nil, 2019, 04
+      #   deprecate :safdfa
```
deivid-rodriguez: In this case, the "good" alternative would be that the code is removed, but leaving an empty line could be confusing so maybe add a comment like:

```suggestion
      #   the `deprecate` call is fully removed
```

bronzdoc:
@deivid-rodriguez sounds good, lets keep it and decide if we have to remove it in the future.

Meanwhile I added a check for `deprecate_command` to the cop. Thanks for the review and let me know if i'm missed something 🙂

lib/rubygems/deprecate.rb
```diff
@@ -45,18 +45,21 @@ def skip_during
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
   # telling the user of +repl+ (unless +repl+ is :none) and the
-  # year/month that it is planned to go away.
+  # Rubygems version that it is planned to go away.
+
+  def deprecate(name, replacement=:none)
+    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
+    next_rubygems_major_version = current_major.bump
```
deivid-rodriguez: Should we extract this common logic to a private method in the module?

util/cops/deprecations.rb
```diff
@@ -8,42 +8,26 @@ module Rubygems
...
-      #   As of March, 2019
-      #
       #   # bad
-      #   deprecate :safdfa, nil, 2018, 12
-      #   deprecate :safdfa, nil, 2019, 03
+      #   deprecate :safdfa, :none
       #
       #   # good
-      #   deprecate :safdfa, nil, 2019, 04
+      #   the `deprecate` call is fully removed
```
deivid-rodriguez: I think generally in rubocop's good and bad examples, the code right below the `# good` and `# bad` headers, at the same indentation level, should be valid ruby. So I'd suggest:

```suggestion
      #   # the `deprecate` call is fully removed
```

deivid-rodriguez: I know that's not what I previously suggested, I just noticed the missing comment when I saw the updated code.

lib/rubygems/deprecate.rb
bronzdoc: I tried to, but it seems there's not a trivial way of doing it and I don't want do some obscure complicated ruby thing just for that 😝

deivid-rodriguez: What about:

```diff
diff --git a/lib/rubygems/deprecate.rb b/lib/rubygems/deprecate.rb
index 57501e1a2b..9dd042c2ff 100644
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -31,6 +31,10 @@ module Gem::Deprecate
     @skip = v
   end
 
+  def self.next_rubygems_major_version # :nodoc:
+    Gem::Version.new(Gem.rubygems_version.segments.first).bump
+  end
+
   ##
   # Temporarily turn off warnings. Intended for tests only.
 
@@ -48,9 +52,6 @@ module Gem::Deprecate
   # Rubygems version that it is planned to go away.
 
   def deprecate(name, replacement=:none)
-    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
-    next_rubygems_major_version = current_major.bump
-
     class_eval do
       old = "_deprecated_#{name}"
       alias_method old, name
@@ -59,7 +60,7 @@ module Gem::Deprecate
         target = klass ? "#{self}." : "#{self.class}#"
         msg = [ "NOTE: #{target}#{name} is deprecated",
                 replacement == :none ? " with no replacement" : "; use #{replacement} instead",
-                ". It will be removed in Rubygems #{next_rubygems_major_version}",
+                ". It will be removed in Rubygems #{Gem::Deprecate.next_rubygems_major_version}",
                 "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
         ]
         warn "#{msg.join}." unless Gem::Deprecate.skip
@@ -70,9 +71,6 @@ module Gem::Deprecate
 
   # Deprecation method to deprecate Rubygems commands
   def deprecate_command
-    current_major = Gem::Version.new(Gem.rubygems_version.segments.first)
-    next_rubygems_major_version = current_major.bump
-
     class_eval do
       define_method "deprecated?" do
         true
@@ -80,7 +78,7 @@ module Gem::Deprecate
 
       define_method "deprecation_warning" do
         msg = [ "#{self.command} command is deprecated",
-                ". It will be removed in Rubygems #{next_rubygems_major_version}.\n",
+                ". It will be removed in Rubygems #{Gem::Deprecate.next_rubygems_major_version}.\n",
         ]
 
         alert_warning "#{msg.join}" unless Gem::Deprecate.skip
```

I don't think it's too magical.

test/rubygems/test_gem_gem_runner.rb
```diff
@@ -76,7 +76,7 @@ def test_query_is_deprecated
       assert_nil @runner.run(args)
     end
 
-    assert_equal "WARNING:  query command is deprecated. It will be removed on or after 2020-12-01.\n", @ui.error
+    assert_match /WARNING:  query command is deprecated. It will be removed in Rubygems [0-9]+/, @ui.error
```
deivid-rodriguez: You're missing adding parentheses here. Without them, ruby warns:

```
/home/deivid/Code/rubygems/test/rubygems/test_gem_gem_runner.rb:79: warning: ambiguous first argument; put parentheses or a space even after `/' operator
```

lib/rubygems/deprecate.rb
bronzdoc: That works 🤷‍♂

bronzdoc:
Thanks for the Review @deivid-rodriguez I address your changes, going to merge this 🚀

deivid-rodriguez:
Thanks @bronzdoc!

simi:
:information_source: I have seen this deprecation mechanismus used in wild by gems extending `Gem::Deprecate` and those libs are broken now. For example https://github.com/appsignal/appsignal-ruby/pull/599.

Since that method is not `nodoc`, would it be possible to get back original signature and postpone this change to next major version? Or do rubygems not care about this problem and should that be used in libs one by one for now? Is there any way to prevent extending of this outside of rubygems itself than?

deivid-rodriguez:
Hi @simi, thanks for noticing this!

I guess we should keep the signature and deprecate the date and month parameters, but... deprecating a deprecation mechanism... makes my head ache :rofl:.

Not sure, let's think about it a bit more. :thinking:

simi:
:thinking: Alternatively we can keep the old `deprecate` method and rename the current one to `deprecate_soon` or something similar and use it where needed.

deivid-rodriguez:
Yeah, I was thinking of something like that too. But it seems weird to give our new preferred deprecation mechanism an artificial name.

Something else I'm noticing is that, we've turned an (apparently public, not sure if that was ever intended) mechanism for deprecating generic methods into something rubygems specific, because the message now mentions rubygems. So that's another compatibility problem on top of that.

deivid-rodriguez:
There's an accepted feature request in ruby-core to make this builtin, by the way: https://bugs.ruby-lang.org/issues/16018.

Currently I tend to think we should deprecate passing date and month, and explicitly pass a version. I believe that was @bronzdoc's initial implementation, and I told him that an explicit version parameter was not necessary. However, if we want to keep this method generic, it sounds like the easiest way. 

I'll think of it a bit more.

simi:
> Yeah, I was thinking of something like that too. But it seems weird to give our new preferred deprecation mechanism an artificial name.

Right name is the key here. Maybe `rubygems_deprecate` can work better.

> Something else I'm noticing is that, we've turned an (apparently public, not sure if that was ever intended) mechanism for deprecating generic methods into something rubygems specific, because the message now mentions rubygems.

The question in here is if rubygems is going to accept this fact and continue supporting `Gem::Deprecation` as is now (now = latest release). IMHO that would be the best to not get tons of angry library maintainers :).

> There's an accepted feature request in ruby-core to make this builtin, by the way: https://bugs.ruby-lang.org/issues/16018.

It is not even developed and would take a long time unless it will be publicly used.

:thinking: FYI I have spotted the same wild usage of rubygems tar reader/writer as well.

deivid-rodriguez:
Regarding the ruby-core feature request, yeah, I only pointed to it as a reference, but it's not something useful at the moment.

Anyways, I definitely appreciate that you spotted this, this early, and I think we should definitely restore support for this at least for the next minor, since I'm sure it is indeed widely used.

> Currently I tend to think we should deprecate passing date and month, and explicitly pass a version.

Actually, deprecations are too tricky. We probably should do this and instead restore the original implementation for now and add our new _private_ mechanism for deprecating stuff ourselves, as you suggested.

simi:
> Regarding the ruby-core feature request, yeah, I only pointed to it as a reference, but it's not something useful at the moment.

Yeah, thanks for that. I had no idea about that.

PR #3417, opened 2020-03-23
Adding semver operator

### LINKED ISSUE #1919: Gem::Requirement add SemVer operator

My current problem is SemVer is too hard to implement.

This issue is related to:

  - [x] Publishing a library
  - [x] The command line `gem`

I will abide by the [code of conduct] (https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md). ✅ 

---

This is regarding the Gem::Requirement class, documented at https://ruby-doc.org/stdlib-2.1.0/libdoc/rubygems/rdoc/Gem/Requirement.html

This issue proposes to add a new operator to compare Semantic Versions. See http://semver.org/

Currently we can express:

> rack ~> 1.2.1 (which means >= 1.2.1 and < 1.3.0)

But it will be more valuable if we can express

> rack ~~> 1.2.1 (which means >= 1.2.1 and < 2.0.0)

zachrab:
# Description:
This is not a final commit; I'd like to get maintainer feedback on the path I'm taking to implement this feature.

Feature- Adding `^` character to signify the gem version can be greater than equal to specified version but less than the major version bump. This is done with the introduction of `major_bump`

Testing- Wrote some testing scenarios but I would like to verify the extent of the testing (e.g. type of test cases, etc.)

Code- WIP 

Closes #1919.
______________

# Tasks:

- [X] Write tests


simi:
:thinking: Is `^ 5.0.0` same as `~> 5.0`?

MSP-Greg:
I think the main benefit with `^` is one can do things like `^5.1.2`, where as with current code, that requires both a `~>` and a `>=` requirement...

simi:
@MSP-Greg sure, I was looking into possibility to just rewrite `^` to `~>` internally (as you mentioned by combination of `~>` a `>=`).

hsbt:
@zachrab Can you describe the use-case with this feature? I also agree with @simi 's concern.

deivid-rodriguez:
I think this is a good feature, and I missed it many times

This is the feature request ticket for it with more discussion: https://github.com/rubygems/rubygems/issues/1919.

PR #3420, opened 2020-03-24
Add Github CI action to automatically label Bundler PRs

colby-swandale:
Per the title, something we can easily automate.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,2 @@
+Bundler:
+  - bundler/**/*
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,11 @@
+name: "Pull Request Labeler"
+on:
+- pull_request
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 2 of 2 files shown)


hsbt:
Can we replace @bronzdoc's labeler to this for the issues too?

colby-swandale:
@hsbt do we still want the triage label? We don't seem to be using it all that much.

deivid-rodriguez:
Nice! :heart_eyes:

In my opinion, @bronzdoc's labeler in its current form is not very useful because it blindly tags every single PR or issue, so essentially the "triage" label is equivalent to no label at all. In fact, I feel it even gets in the middle a bit because once you do the triaging for a PR or issue, you have to remove the label.

So, my vote would go to stop using it.

hsbt:
I'm also +1 to remove/stop it. We should re-arrange to label guideline in this repo.

colby-swandale:
I will merge this now, but raise a new issue to propose dropping the automated triage label.

bronzdoc:
This is nice! I don't think we use the labeler in PR's that much. I recently added functionality to label PR's with the bundler label based on the files modified but I think this is better.

PR #3422, opened 2020-03-24
Accept not only /usr/bin/env but also /bin/env in some tests

mame:
RubyGems chooses available one from the two: /usr/bin/env or /bin/env

https://github.com/rubygems/rubygems/blob/20b0d609484df2b514954ba9ef890a7cbdd01d18/lib/rubygems/installer.rb#L38

So, it is good not to hard-code /usr/bin/env in tests.

Actually, these tests fail on Android because it provides /bin/env
available but not /usr/bin/env.

https://rubyci.org/logs/rubyci.s3.amazonaws.com/android28-x86_64/ruby-master/log/20200324T074312Z.fail.html.gz

### DIFF
--- a/test/rubygems/test_gem_commands_pristine_command.rb
+++ b/test/rubygems/test_gem_commands_pristine_command.rb
@@ -160,11 +160,9 @@ class TestGemCommandsPristineCommand < Gem::TestCase
 
     ruby_exec = sprintf Gem.default_exec_format, 'ruby'
 
-    if win_platform?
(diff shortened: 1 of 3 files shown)


mame:
This change had been introduced in ruby/ruby but (maybe accidentally) removed by https://github.com/ruby/ruby/pull/2889 :-)  @deivid-rodriguez

deivid-rodriguez:
Thank you @mame! Yes, it was unintentional. Since they changes had not been contributed upstream, they got unintentionally reverted when syncing when upstream master.

mame:
Thanks!

PR #3424, opened 2020-03-24
Specify explicit separator not to be affected by $;

nobu:
# Description:

Get rid to be affected by `$;`.

Usually nothing.
When using rubygems as a library, the global variable can be set non-nil value.

Specify separator argument to `String#split` explicitly.

______________

# Tasks:

- [x] Write tests


deivid-rodriguez:
Thanks @nobu!!

PR #3429, opened 2020-03-25
Respect already loaded files from default gems

### LINKED ISSUE #3131: (default) gem conflicts with StdLib

deivid-rodriguez:
# Description:

This PR fixes the first issue described in #3131, by considering files already loaded and activating their corresponding specifications during default gem registration.

Closes #3131.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1230,6 +1230,8 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
           next unless $~
         end
 
+        spec.activate if already_loaded?(file)
+
         @path_to_default_spec_map[file] = spec
         @path_to_default_spec_map[file.sub(suffix_regexp, "")] = spec
       end
@@ -1294,6 +1296,18 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
     # work
 
     attr_reader :pre_uninstall_hooks
+
+    private
+
+    def already_loaded?(file)
+      default_gem_load_paths.find do |load_path_entry|
+        $LOADED_FEATURES.include?("#{load_path_entry}/#{file}")
+      end
+    end
+
+    def default_gem_load_paths
+      @default_gem_load_paths ||= $LOAD_PATH[load_path_insert_index..-1]
+    end
 
   end
 
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -802,6 +802,7 @@ class Gem::TestCase < Minitest::Test
 
     lib_dir = File.join(@tempdir, "default_gems", "lib")
     lib_dir.instance_variable_set(:@gem_prelude_index, lib_dir)
+    Gem.instance_variable_set(:@default_gem_load_paths, [*Gem.instance_variable_get(:@default_gem_load_paths), lib_dir])
     $LOAD_PATH.unshift(lib_dir)
     files.each do |file|
       rb_path = File.join(lib_dir, file)
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -271,6 +271,24 @@ class TestGemRequire < Gem::TestCase
     $LOAD_PATH.replace lp if load_path_changed
   end
 
+  def test_activate_via_require_respects_loaded_default_from_default_gems
+    a1 = new_default_spec "a", "1", nil, "a.rb"
+
+    # simulate requiring a default gem before rubygems is loaded
+    Kernel.send(:gem_original_require, "a")
+
+    # simulate registering default specs on loading rubygems
+    install_default_gems a1
+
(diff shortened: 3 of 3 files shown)


lib/rubygems/test_case.rb
```diff
@@ -808,8 +808,7 @@ def new_default_spec(name, version, deps = nil, *files)
     spec.files = files
 
     lib_dir = File.join(@tempdir, "default_gems", "lib")
-    lib_dir.instance_variable_set(:@gem_prelude_index, lib_dir)
-    $LOAD_PATH.unshift(lib_dir)
+    $LOAD_PATH.push(lib_dir)
```
voxik: Could you please elaborate, how these could be equivalent (possibly better in the commit message :angel:)? Because gems always have precedence over StdLib, so how could append be the same as prepend? This IMO gives precedence to StdLib stuff instead of the RubyGems test suite stuff. But I might missing something.

deivid-rodriguez: Default gems normally live at the end of the default $LOAD_PATH. So adding them in the end makes more sense in my opinion. If we add them at the front, we need to hack into the internals of the $LOAD_PATH so that the result of `Gem.load_path_insert_index` is still correct.

In any case, this is just a test simplification independent from the fix in this PR. I can change it back if you prefer.

deivid-rodriguez: >  This IMO gives precedence to StdLib stuff instead of the RubyGems test suite stuff

Yes, this is correct, but it doesn't change a thing, because everything needed from the rubygems code base should've been required by then, and because we don't use dummy gems with file names conflicting with rubygems files.

But if you feel this is worse or incorrect, I'm fine reverting this change. As I said, it's not an important part of this PR.

deivid-rodriguez: I'll revert the change though so that this PR focuses more on the bug fix.

deivid-rodriguez: Actually, I'm not really sure, it feels more realistic to me to add default test gems at the end of the $LOAD_PATH rather than at the front.

So I'm not sure if it's worth reverting.

voxik: From my POV, it'd be much better if this was separate PR. As you can see, tweaking `$LOAD_PATH` is delicate thing, which might have unexpected results.

lib/rubygems.rb
```diff
@@ -1270,6 +1270,14 @@ def register_default_spec(spec)
           next unless $~
         end
 
+        expanded_candidates = spec.full_require_paths.map do |path|
```
deivid-rodriguez: I just noticed that this outer loop is unnecessary for gems with "new format gemspecs", and that my patch this doesn't properly fix the issue for gems with "old format gemspecs". I'll work on fixing this.

deivid-rodriguez:
@voxik I removed the commit you suggested and simplified the implementation. I think the code is easier to understand now. Let me know what you think.

PR #3432, opened 2020-03-26
Report GitLab CI within bundler user-agent string

brodock:
# Description:

Bundler user-agent is being used to generate content for rubygems stats
https://stats.rubygems.org 

This adds GitLab CI specific ENV variable.

There is a limited amount of CI providers being listed in https://stats.rubygems.org/

By including GitLab we expect that a portion of the current "Ci" item should include GitLab CI and therefore report correctly.

Added specific ENV to the list of known providers. This will allow bundler to report that information, which can later on be parsed by https://github.com/rubytogether/kirby

______________

# Tasks:

- [x] Write tests


colby-swandale:
Thanks!

PR #3438, opened 2020-03-26
Add ability to uninstall plugins

hsbt:
# Description:

Migrated from https://github.com/rubygems/bundler/pull/7409

/cc @fatkodima
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -47,6 +47,36 @@ module Bundler
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # Uninstalls plugins by the given names
+    #
+    # @param [Array<String>] names the names of plugins to be uninstalled
+    def uninstall(names, options)
+      if names.empty? && !options[:all]
+        Bundler.ui.error "No plugins to uninstall. Specify at least 1 plugin to uninstall.\n"\
+          "Use --all option to uninstall all the installed plugins."
+        return
+      end
+
+      names = index.installed_plugins if options[:all]
+      if names.any?
+        names.each do |name|
+          begin
+            if index.installed?(name)
+              Bundler.rm_rf(index.plugin_path(name))
+              index.remove_plugin(name)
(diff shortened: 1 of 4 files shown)


bundler/lib/bundler/plugin/index.rb
```diff
@@ -71,6 +71,15 @@ def register_plugin(name, path, load_paths, commands, sources, hooks)
         raise
       end
 
+      def remove_plugin(name)
+        @commands.delete_if {|_, v| v == name }
```
deivid-rodriguez, line 75: `@commands` looks like the same kind of structure as `@sources`, so it should be removed with the following I guess?
 
```suggestion
        @commands.delete(name)
```

bundler/lib/bundler/plugin/index.rb
```diff
@@ -71,6 +71,15 @@ def register_plugin(name, path, load_paths, commands, sources, hooks)
         raise
       end
 
+      def remove_plugin(name)
```
deivid-rodriguez, line 74: I think `unregister_plugin` fits better?

bundler/lib/bundler/plugin.rb
```diff
@@ -47,6 +47,36 @@ def install(names, options)
...
+        names.each do |name|
+          begin
+            if index.installed?(name)
+              Bundler.rm_rf(index.plugin_path(name))
+              index.remove_plugin(name)
+              Bundler.ui.info "Uninstalled plugin #{name}"
+            else
+              Bundler.ui.error "Plugin #{name} is not installed \n"
+            end
+          rescue PluginError => e
```
deivid-rodriguez, line 71: I don't think this error can actually be raised?

bundler/lib/bundler/plugin/index.rb
fatkodima: I do not remember the exact structure of plugins index file right now, but here we should remove by matching value (as opposed to `delete` method, which deletes matching key).

hsbt:
Ah, Sorry for the unfriendly situation. #3482 is good to continue this.

@fatkodima Thanks!

PR #3440, opened 2020-03-26
Fix race condition on bundler's parallel installer

deivid-rodriguez:
# Description:

We're getting several crashes in our specs where `Enumerable` methods are called on `nil`, resulting in crashes.

See https://github.com/rubygems/rubygems/runs/537767938 for a recent example.

Note that this issue has been aggravated since we enabled parallel installation by default in #3393.

When installing in parallel, `bundler` creates several `Gem::Installer`
instances that run in parallel. These installers access the `@@all` class
variable of `Gem::Specification` concurrently.

If a concurrent thread calls `Gem::Specification.reset` (resetting `@all` to `nil`) while another thread is running `Gem::Specification._all` or another method that expects `@@all` to be
loaded and not `nil`, that can result in `Enumerable` methods being called on `nil`, resulting in crashes.

I fix it by protecting the other concurrent access to the `@all` variable.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -509,7 +509,7 @@ class Gem::Installer
   end
 
   def generate_plugins # :nodoc:
-    latest = Gem::Specification.latest_spec_for(spec.name)
+    latest = Gem::Installer.install_lock.synchronize { Gem::Specification.latest_spec_for(spec.name) }
     return if latest && latest.version > spec.version
 
     ensure_writable_dir @plugins_dir


deivid-rodriguez:
Note that this race condition was introduced with the new [rubygems plugin layout](https://github.com/rubygems/rubygems/pull/3108), so it does not affect any released version and can only be noticed in matrix CI entries that run bundler's specs against rubygems master.

deivid-rodriguez:
Merging this hoping that it will stabilize CI.

PR #3445, opened 2020-03-27
Minitest cleanup

deivid-rodriguez:
# Description:

Some small code cleanups related to minitest.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -12,7 +12,7 @@ if File.exist?(bundler_gemspec)
 end
 
 begin
-  gem 'minitest', '~> 5.0'
+  gem 'minitest', '~> 5.13'
 rescue Gem::LoadError
 end
 
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -127,12 +127,6 @@ class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Uni
     assert_equal expected.sort, loaded.sort if expected
   end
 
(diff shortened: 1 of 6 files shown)


lib/rubygems/test_case.rb
```diff
@@ -11,11 +11,6 @@
   Gem::Specification.dirs.shift
 end
 
-begin
-  gem 'minitest', '~> 5.0'
```
colby-swandale: Are we suppose to be using `bundle exec rake test` from now on? I'm lost why we don't need this anymore

deivid-rodriguez: I can see how you're lost because removing this is just wrong :laughing:. Nice catch, removing the commit.

deivid-rodriguez: @colby-swandale I ended up restoring this but removing the exception swallowing code. Since after these changes our test suite requires minitest `5.13` or higher because I removed the compatibility code, I think the exception is useful to tell users not satisfying the requirement to install a matching version. Let me know what you think.

colby-swandale: That sounds reasonable. Would it be worth catching the exception anyway and add a friendly error message to say you need to run `bundle install`?

deivid-rodriguez: @colby-swandale Since our tests are run under different environments, I think I'd rather the raw exception in place? For example, if our tests are run from `ruby-core`, I don't think a `bundle install` suggestion would apply there. Plus, `gem install minitest` will fix the issue anyways without a full `bundle install`.

PR #3446, opened 2020-03-27
Improvements on Jruby's CI

deivid-rodriguez:
# Description:

This code is not necessary and has the bad side effect of causing test failures in cascade once a single test inside the file fails.
    
See https://github.com/rubygems/rubygems/runs/538978838 for an example.

# Tasks:

- [ ] Write tests


Gemfile
```diff
@@ -7,3 +7,4 @@ gem "minitest", "~> 5.0"
 gem "simplecov", "~> 0.17.0"
 gem "rubocop", "~> 0.80.1"
 gem "rubocop-performance", "~> 1.5.2"
+gem "webrick", "~> 1.6"
```
bronzdoc: Why we adding webrick?

deivid-rodriguez: We're not really adding it, but making the dependency explicit. The reason for doing it is explained in the commit message. I'm happy to extract that change to a separate PR, or to rename this PR to "Improvements on jruby's CI" to better match the current changes.

bronzdoc: Oh I was just seeing the title `Force a newer version of `webrick`` didn't notice there was more ha!
> rename this PR to "Improvements on jruby's CI" to better match the current changes.

Maybe this 👍

PR #3457, opened 2020-03-29
Convert `simulate_ruby_engine` to `ruby_engine_is` guard

mame:
`simulate_ruby_engine` replaces `RUBY_ENGINE` to make MRI pretend JRuby.
This is a completely unreasonable and impossible idea because the
constant `RUBY_ENGINE` is used to insert a code fragment for a
particular engine, like this:

https://github.com/ruby/ruby/commit/2b6848af0e
```
require "jruby" if RUBY_ENGINE == "jruby"
```

And a bundler spec fails like this:

https://github.com/ruby/ruby/runs/542345866?check_suite_focus=true#step:14:106
```
/home/runner/work/ruby/ruby/src/lib/irb/ruby-lex.rb:14:in `require': cannot load such file -- jruby (LoadError)
```

This changeset converts all `simulate_ruby_engine` to `ruby_engine_is`
guard that executes a given block only on the specified engine.
If you want to test it with JRuby, you must use JRuby, not make MRI
pretend JRuby.

cc/ @headius @hsbt

### DIFF
--- a/bundler/spec/support/platforms.rb
+++ b/bundler/spec/support/platforms.rb
@@ -65,12 +65,10 @@ module Spec
     end
 
     def local_ruby_engine
-      ENV["BUNDLER_SPEC_RUBY_ENGINE"] || RUBY_ENGINE
+      RUBY_ENGINE
     end
 
     def local_engine_version
-      return ENV["BUNDLER_SPEC_RUBY_ENGINE_VERSION"] if ENV["BUNDLER_SPEC_RUBY_ENGINE_VERSION"]
-
       RUBY_ENGINE_VERSION
     end
 
--- a/bundler/spec/commands/outdated_spec.rb
+++ b/bundler/spec/commands/outdated_spec.rb
@@ -606,7 +606,7 @@ RSpec.describe "bundle outdated" do
     end
 
     it "reports that updates are available if the JRuby platform is used" do
-      simulate_ruby_engine "jruby", "1.6.7" do
+      ruby_engine_is "jruby", "1.6.7" do
         simulate_platform "jruby" do
           install_gemfile <<-G
             source "#{file_uri_for(gem_repo2)}"
--- a/bundler/spec/install/gemfile/gemspec_spec.rb
+++ b/bundler/spec/install/gemfile/gemspec_spec.rb
@@ -169,20 +169,20 @@ RSpec.describe "bundle install from an existing gemspec" do
 
   it "should match a lockfile on non-ruby platforms with a transitive platform dependency" do
     simulate_platform java
-    simulate_ruby_engine "jruby"
+    ruby_engine_is "jruby" do
+      build_lib("foo", :path => tmp.join("foo")) do |s|
+        s.add_dependency "platform_specific"
+      end
 
-    build_lib("foo", :path => tmp.join("foo")) do |s|
-      s.add_dependency "platform_specific"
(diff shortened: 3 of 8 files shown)


nobu:
I think that the first name `simualte_ruby_engine` is too misleading.
It should be a name expected **not** to work, e.g.,`with_unexpceted_ruby_engine`.
And the other misused cases should use `ruby_engine_is` guard.

headius:
:+1: This is good. We want to keep diffs to a minimum for JRuby, but there are places where any Ruby engine might want to have specific code. The only way to test those situations is to actually run those engines... not to try to fake out the test harness into thinking it's running on that engine.

There's also a lot of JRuby-specific stuff here that should be audited. Some have very old JRuby versions (1.6.7 is from 2012... 9.1.x is no longer supported by us). Others may be behavioral differences we don't want to maintain, but we JRuby folks don't know about them.

eregon:
This issue also exists in RubyGems:
https://github.com/rubygems/rubygems/search?q=RUBY_ENGINE&unscoped_q=RUBY_ENGINE
For example:
https://github.com/rubygems/rubygems/blob/8a5dfa6cff25444fd1cbb1ed6705e75f39c5f888/test/rubygems/test_gem_request_set_gem_dependency_api.rb#L22

I think it's useful to test with a changed local platform, that e.g., version parsing comparison still works. But it should be done in a way that doesn't require to change `RUBY_*` constants, as indeed that might break many things.

Maybe Bundler/RubyGems could read RUBY_ENGINE/RUBY_VERSION into their own constants/class ivars, and then change those for tests?

deivid-rodriguez:
Yeah, I don't like that we do this either... :grimacing:.

However, I believe that simply skipping these specs if we're not on jruby is not a good solution because we don't yet have a jruby CI, so the method even if hacky and brittle, at least gives us some coverage.

My proposal for now would be:

* Keep the current method where it makes sense, and try to improve it in the future to not change `RUBY_*` constants like @eregon mentioned. In particular, without having gone through all the specs, I think we should probably keep those cases where a particular `jruby` version is passed.

* For the rest of the specs that only require a "generic jruby", remove the method and tag them instead with a `:jruby` tag. Then configure RSpec to only run those only on `jruby`, and run `bin/rspec --tag jruby` on a new workflow that runs on `jruby`. Our specs are too slow at the moment for a full run on jruby, but I think we can definitely afford to run only this narrow set of specs.

I'm happy to take this one.

bundler/spec/support/platforms.rb
```diff
@@ -65,12 +65,10 @@ def not_local_tag
     end
 
     def local_ruby_engine
-      ENV["BUNDLER_SPEC_RUBY_ENGINE"] || RUBY_ENGINE
+      RUBY_ENGINE
     end
 
     def local_engine_version
-      return ENV["BUNDLER_SPEC_RUBY_ENGINE_VERSION"] if ENV["BUNDLER_SPEC_RUBY_ENGINE_VERSION"]
```
eregon: This and the above are already a mechanism to simulate another Ruby engine without changing the RUBY_* constants.
It could also be done more directly by stubbing those methods during the tests needing it.

hsbt:
@mame Can you rebase with the current master branch? I update the step of `actions/checkout@v2` now.

https://github.com/rubygems/rubygems/pull/3460

deivid-rodriguez:
@hsbt approved #3464 so closing here in favor of that PR.

Thanks @mame for your work and sorry for the trouble :pray:.

PR #3462, opened 2020-03-30
Revert "Minitest cleanup"

hsbt:
# Description:

Unfortunately,  https://github.com/rubygems/rubygems/pull/3445 completely broke the test suite on the ruby core repository.

Reverted the related changes for the ruby core.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -98,7 +98,7 @@ end
 # and uninstall gems, fetch remote gems through a stub fetcher and be assured
 # your normal set of gems is not affected.
 
-class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Unit::TestCase)
+class Gem::TestCase < Minitest::Test
 
   extend Gem::Deprecate
 
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -11,7 +11,10 @@ if File.exist?(bundler_gemspec)
   Gem::Specification.dirs.shift
 end
 
-gem 'minitest', '~> 5.13'
+begin
+  gem 'minitest', '~> 5.13'
+rescue Gem::LoadError
+end
 
 begin
   require 'simplecov'
@@ -95,7 +98,7 @@ end
 # and uninstall gems, fetch remote gems through a stub fetcher and be assured
 # your normal set of gems is not affected.
 
-class Gem::TestCase < Minitest::Test
+class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Unit::TestCase)
 
   extend Gem::Deprecate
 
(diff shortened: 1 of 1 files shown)


hsbt:
The API of minitest is unstable. We should move to `test-unit` instead of `minitest` 

https://github.com/rubygems/rubygems/pull/3141

bronzdoc:
@hsbt aren't we testing against ruby-head to avoid this issues?

MSP-Greg:
> aren't we testing against ruby-head to avoid this issues?

ruby/ruby essentially uses a test system that is based on old versions of minitest & test-unit with a fair number of changes.  Tests run here on ruby-head are not running on that test system.  It could be done in Actions, I think...

MSP-Greg:
@hsbt

I just looked at #3445, and the change in `lib/rubygems/test_case.rb` around line 100 would have totally broken things at one point in time. I know because I wrote the original line.  Reverting that may solve some of the issues.

From:
```ruby
class Gem::TestCase < (defined?(Minitest::Test) ? Minitest::Test : MiniTest::Unit::TestCase)
```
To:
```ruby
class Gem::TestCase < Minitest::Test
```

hsbt:
@MSP-Greg Thanks. your suggestion is correct. I confirmed to work `Minitest::Test` on the ruby core repository.

MSP-Greg:
Re setting up a CI job that tests as in ruby/ruby:

Ruby's tests run in the build folder using make.  A while ago I decided to run them against the install folder, and hence, without make.  I think the main RG issue was changing Gem::TestCase to find support files, maybe pems, certs, and good/bad rake?

So, to run CI, it would b something like copying RG code into ruby-head, downloading the ruby repo and copying current RG tests into the location they are currently are in, and running 'test-all' on that folder.

Make sense?  Would that be helpful?

deivid-rodriguez:
@MSP-Greg If we setup the structure of ruby-core repository and we can run rubygems tests the way they are run in ruby-core, I think it could be worth to avoid these issues, yeah. So your idea of "not using make" would be to avoid having to compile ruby and everything, right, but simply test the different structure of the repo, but using a "standard" ruby installation. Is that your idea?

hsbt:
We may remove `assert_path_exists` and `refute_path_exists` from this. I try to migrate them to the assertion class of ruby-core test suite.

deivid-rodriguez:
Nice, thanks!

MSP-Greg:
@deivid-rodriguez

> So your idea of "not using make"

For various reasons, all the ruby/ruby tests run with make, and they run based on a 'build/check/install' flow.  The issue I have is that the install is what's important.  Or, running tests before install is created isn't a good check of whether install works.

Anyway, let me do some work with it.

Another option is what's done in reline, as the file [download-test_readline.rb](https://github.com/ruby/reline/blob/9fff7b5738006db5df05f0533b7d0b9b1cf69f4d/download-test_readline.rb) is used to download much of the ruby/ruby 'test system', and those files are used for some of the CI.  The download time isn't an issue.

deivid-rodriguez:
I think a good indicator for whether whatever we implement is a good check or not is whether it would've caught this issue.

PR #3464, opened 2020-03-30
Real jruby

deivid-rodriguez:
# Description:

This is my alternative proposal to #3457.

The `simulate_ruby_engine` approach of overwriting constants such as `RUBY_ENGINE` is brittle and has caused some specs to break in ruby-core, so we should get rid of it.

As an alternative, I'm migrating the specs using it to a real jruby CI.

I had to add a few workarounds to get the specs passing on jruby, due to [a bug in jruby](https://github.com/jruby/jruby/issues/6153).

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/jruby-bundler.yml
+++ b/.github/workflows/jruby-bundler.yml
@@ -8,7 +8,7 @@ on:
       - master
 
 jobs:
-  warbler:
+  jruby_bundler:
     runs-on: ubuntu-latest
 
     env:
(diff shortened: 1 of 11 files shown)


simi:
:+1: I like this approach better @deivid-rodriguez. Feel free to ping me if any help is needed.

deivid-rodriguez:
Thanks @simi, let's see if CI is ok.

.github/workflows/jruby-bundler.yml
```diff
@@ -11,12 +11,23 @@ jobs:
   warbler:
```
hsbt: We should use `jruby_bundler` instead of `warbler`

deivid-rodriguez:
Going in, thanks @hsbt!

PR #3435, opened 2020-03-26
Include gem name in native extension build directory path

hsbt:
# Description:

Migrated from https://github.com/rubygems/bundler/pull/6934
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3475, opened 2020-04-02
Multiple gems in git repo bugfix

joelnordell:
# Description:

This change is an important fix for when a Gemfile specifies multiple gems that are contained in one git repository (where each gem is in a subdirectory within the repo) and they all have native extensions.

Without this change, in certain situations bundler would only build one of the native extensions but not the others.

______________

The fix is to separate the extensions build path for each gem inside the repository, so that the `gem.build_complete` file exists independently for each separate gem.

______________

# Tasks:

- [X] Write tests

______________

This PR replaces https://github.com/rubygems/rubygems/pull/3435 and adds a test.

### DIFF
--- a/bundler/spec/install/gems/native_extensions_spec.rb
+++ b/bundler/spec/install/gems/native_extensions_spec.rb
@@ -87,6 +87,58 @@ RSpec.describe "installing a gem with native extensions", :ruby_repo do
     expect(out).to eq("true")
   end
 
+  it "installs correctly from git when multiple gems with extensions share one repository" do
+    build_repo2 do
+      ["one", "two"].each do |n|
+        build_lib "c_extension_#{n}", "1.0", :path => lib_path("gems/c_extension_#{n}") do |s|
+          s.extensions = ["ext/extconf.rb"]
+          s.write "ext/extconf.rb", <<-E
+            require "mkmf"
+            name = "c_extension_bundle_#{n}"
+            dir_config(name)
+            raise "OMG" unless with_config("c_extension_#{n}") == "#{n}"
+            create_makefile(name)
+          E
+
+          s.write "ext/c_extension_#{n}.c", <<-C
+            #include "ruby.h"
+
+            VALUE c_extension_#{n}_value(VALUE self) {
+              return rb_str_new_cstr("#{n}");
+            }
+
+            void Init_c_extension_bundle_#{n}() {
+              VALUE c_Extension = rb_define_class("CExtension_#{n}", rb_cObject);
+              rb_define_method(c_Extension, "value", c_extension_#{n}_value, 0);
+            }
+          C
+
+          s.write "lib/c_extension_#{n}.rb", <<-C
+            require "c_extension_bundle_#{n}"
+          C
+        end
+      end
+      build_git "gems", :path => lib_path("gems"), :gemspec => false
+    end
+
+    bundle! "config set build.c_extension_one --with-c_extension_one=one"
+    bundle! "config set build.c_extension_two --with-c_extension_two=two"
+
+    # 1st time, require only one gem -- only one of the extensions gets built.
+    install_gemfile! <<-G
+      gem "c_extension_one", :git => #{lib_path("gems").to_s.dump}
+    G
+
+    # 2nd time, require both gems -- we need both extensions to be built now.
+    install_gemfile! <<-G
+      gem "c_extension_one", :git => #{lib_path("gems").to_s.dump}
+      gem "c_extension_two", :git => #{lib_path("gems").to_s.dump}
+    G
+
+    run! "Bundler.require; puts CExtension_one.new.value; puts CExtension_two.new.value"
+    expect(out).to eq("one\ntwo")
+  end
+
   it "install with multiple build flags" do
     build_git "c_extension" do |s|
       s.extensions = ["ext/extconf.rb"]
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -51,7 +51,8 @@ module Gem
       alias_method :rg_extension_dir, :extension_dir
       def extension_dir
         @bundler_extension_dir ||= if source.respond_to?(:extension_dir_name)
-          File.expand_path(File.join(extensions_dir, source.extension_dir_name))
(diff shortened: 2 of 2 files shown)


joelnordell:
@crugg -- FYI, here is the rspec test for the bundler fix.

bundler/spec/install/gems/native_extensions_spec.rb
```diff
@@ -88,6 +88,62 @@
...
+
+    bundle! "config set build.c_extension_one --with-c_extension_one=one"
+    bundle! "config set build.c_extension_two --with-c_extension_two=two"
+
+    # 1st time, require only one gem -- only one of the extensions gets built.
+    install_gemfile! <<-G
+      gem "c_extension_one", :git => #{lib_path("gems").to_s.dump}
+    G
+
+    expect(out).not_to include("extconf.rb failed")
```
deivid-rodriguez, line 133: I know this is probably copied from other specs in this file that do the same, but this line is not necessary, since the `install_gemfile!` will fail earlier in case of an error when building the extension.

bundler/spec/install/gems/native_extensions_spec.rb
```diff
@@ -88,6 +88,62 @@
...
+
+    # 2nd time, require both gems -- we need both extensions to be built now.
+    install_gemfile! <<-G
+      gem "c_extension_one", :git => #{lib_path("gems").to_s.dump}
+      gem "c_extension_two", :git => #{lib_path("gems").to_s.dump}
+    G
+
+    expect(out).not_to include("extconf.rb failed")
```
deivid-rodriguez, line 141: Same as the other comment here.

bundler/spec/install/gems/native_extensions_spec.rb
joelnordell: Yep, that was taken from the other adjacent tests to be consistent with them. I'm fine with removing it since the final expectation can only pass if everything succeeds with the build - should I remove it from all the other tests in that file too?

joelnordell:
> While reviewing it, I noticed a warning being printed when bundle installing this kind of setups (not sure if you noticed it). Anyways, since you've been going over this area of the code, I appreciate if you can have a look at it, it's #3479.

Sure, I'll take a look.

bundler/spec/install/gems/native_extensions_spec.rb
deivid-rodriguez: I'm doing that in #3479, so not necessary!

joelnordell:
@deivid-rodriguez two of the checks here failed but seem to be unrelated to the changes in this PR. Is there any way to re-run those checks?

deivid-rodriguez:
Can you rebase this PR to make sure it runs of top of the current master? Just in case.

bundler/spec/install/gems/native_extensions_spec.rb
```diff
@@ -87,6 +87,58 @@
     expect(out).to eq("true")
   end
 
+  it "installs from git when multiple gems share one repository" do
```
deivid-rodriguez, line 90: To make the wording of the spec more exact, I'd say:

```suggestion
  it "installs correctly from git when multiple gems with extensions share one repository" do
```

Just pointing out that the test is about extensions, and adding "correctly" to try to emphasize that the install won't actually fail, but will install "incorrectly" (extensions artifacts overwriting each other).

joelnordell:
> I finished to review this. Just a little tweak to the spec wording, and also to anwer [@segiddins' question in the original PR](https://github.com/rubygems/bundler/pull/6934#issuecomment-460010015):
> 
> > Will this require rebuilding all existing extensions?
> 
> I don't think so. This PR should be fully "backwards compatible" in that sense, since the target folder name is preserved _unless_ there's a conflicting situation like the one in this issue. Please @joelnordell correct me if I'm wrong.

That is correct. To demonstrate this, here are some results showing where bundler builds native extensions for two of the test cases (one gem in one git repo, the previously existing test; and multiple gems in one repo, the new test) - both with and without this change:

---

### One gem in one repo (existing test case - most typical scenario):
Running from `master`:
```
$ rspec spec/install/gems/native_extensions_spec.rb:49
$ find tmp/1/gems/system/bundler/gems/ -name gem.build_complete
tmp/1/gems/system/bundler/gems/extensions/x86_64-linux/2.7.0/c_extension-1.0-f5582e00bb79/gem.build_complete
```

Running from this branch:
```
$ rspec spec/install/gems/native_extensions_spec.rb:49
$ find tmp/1/gems/system/bundler/gems/ -name gem.build_complete
tmp/1/gems/system/bundler/gems/extensions/x86_64-linux/2.7.0/c_extension-1.0-ed2ed6b79d5a/gem.build_complete
```

You can see here that the path where the extension is built is unchanged (aside from the git commit hash, which is different every time the test runs).

---

### Multiple gems in one repo (new test case - scenario that previously did not work):
Running from `master` (with the new test cherry-picked):
```sh
$ rspec spec/install/gems/native_extensions_spec.rb:138 # NOTE the test fails here
$ find tmp/1/gems/system/bundler/gems/ -name gem.build_complete
tmp/1/gems/system/bundler/gems/gems-bb0b90940942/extensions/x86_64-linux/2.7.0/gems-bb0b90940942/gem.build_complete
```

Running from this branch:
```
$ rspec spec/install/gems/native_extensions_spec.rb:138
$ find tmp/1/gems/system/bundler/gems/ -name gem.build_complete
tmp/1/gems/system/bundler/gems/gems-63da39124084/extensions/x86_64-linux/2.7.0/gems-63da39124084-c_extension_one/gem.build_complete
tmp/1/gems/system/bundler/gems/gems-63da39124084/extensions/x86_64-linux/2.7.0/gems-63da39124084-c_extension_two/gem.build_complete
```

In this case, however, with this change the extensions are built in different paths so that `gem.build_complete` exists independently for each extension.

---

The bottom line is this change is backwards compatible for the most common case of a repository containing only one gem which is located in the root of the repository.

For the less common case of a gem being contained in a subdirectory of a repository (regardless of how many there are), when upgrading bundler this would result in the previously-built extension no longer being recognized as having been built, and it would then get rebuilt.

joelnordell:
@deivid-rodriguez - this should be ready to go now - updated the test description and squashed to one commit.

PR #3477, opened 2020-04-02
[Bundler] Definition: initialize an instance variable to avoid a Ruby warning

olleolleolle:
# Description:

This PR avoids a Ruby warning "instance variable `@new_platform` not initialized".

A warning was emitted on using Bundler:

```
/home/travis/.rvm/rubies/ruby-head/lib/ruby/2.8.0/bundler/definition.rb:569: warning: instance variable @new_platform not initialized
/home/travis/.rvm/rubies/ruby-head/lib/ruby/2.8.0/bundler/definition.rb:537: warning: instance variable @new_platform not initialized
```

I initialized the instance variable along with some other instance variables which were set to nil in the constructor.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -77,6 +77,7 @@ module Bundler
       @locked_bundler_version = nil
       @locked_ruby_version    = nil
       @locked_specs_incomplete_for_platform = false
+      @new_platform = nil
 
       if lockfile && File.exist?(lockfile)
         @lockfile_contents = Bundler.read_file(lockfile)


deivid-rodriguez:
@olleolleolle Can you show the full backtrace? I'm curious how you run into this since this warning doesn't seem to show up in regular usage.

olleolleolle:
It was in a test run using ruby-head. I'll see what I can dig up.

deivid-rodriguez:
Aha, thanks! And also for testing against ruby-head :)

We do have a daily ruby-head CI which currently shows no warnings: https://github.com/rubygems/rubygems/runs/556606269?check_suite_focus=true. But test environments differ a lot between them.

olleolleolle:
https://travis-ci.org/github/taf2/curb/jobs/670201825#L889

Just after compile of the extension

deivid-rodriguez:
Thanks @olleolleolle, that's super helpful!

Your fix should be good in any case, but I like to find out about these "edge case usages".

PR #3479, opened 2020-04-03
Fix double chdir warning

deivid-rodriguez:
Closes #2904.

When bundler installs a git gem with extensions, the following warning is printed:

```
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/site_ruby/2.7.0/rubygems/ext/builder.rb:175: warning: conflicting chdir during another chdir block
```

My fix is to remove the outermost directory change which turned out to be unnecessary.

I found this issue while reviewing #3475.

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/source/path/installer.rb
+++ b/bundler/lib/bundler/source/path/installer.rb
@@ -26,18 +26,16 @@ module Bundler
         end
 
         def post_install
-          SharedHelpers.chdir(@gem_dir) do
-            run_hooks(:pre_install)
+          run_hooks(:pre_install)
 
-            unless @disable_extensions
-              build_extensions
-              run_hooks(:post_build)
-            end
-
-            generate_bin unless spec.executables.nil? || spec.executables.empty?
-
-            run_hooks(:post_install)
+          unless @disable_extensions
+            build_extensions
+            run_hooks(:post_build)
           end
+
+          generate_bin unless spec.executables.nil? || spec.executables.empty?
+
+          run_hooks(:post_install)
         ensure
           Bundler.rm_rf(@tmp_dir) if Bundler.requires_sudo?
         end
--- a/bundler/spec/install/gems/native_extensions_spec.rb
+++ b/bundler/spec/install/gems/native_extensions_spec.rb
@@ -38,9 +38,8 @@ RSpec.describe "installing a gem with native extensions", :ruby_repo do
     G
 
     bundle "config set build.c_extension --with-c_extension=hello"
-    bundle "install"
+    bundle! "install"
 
-    expect(out).not_to include("extconf.rb failed")
     expect(out).to include("Installing c_extension 1.0 with native extensions")
 
     run "Bundler.require; puts CExtension.new.its_true"
@@ -82,8 +81,6 @@ RSpec.describe "installing a gem with native extensions", :ruby_repo do
       gem "c_extension", :git => #{lib_path("c_extension-1.0").to_s.dump}
     G
 
-    expect(out).not_to include("extconf.rb failed")
-
     run! "Bundler.require; puts CExtension.new.its_true"
     expect(out).to eq("true")
   end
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/source/path/installer.rb
```diff
@@ -26,18 +26,16 @@ def initialize(spec, options = {})
         end
 
         def post_install
-          SharedHelpers.chdir(@gem_dir) do
```
joelnordell, line 29: Removing this `chdir` slightly changes the semantics of the `:pre_install`, `:post_build`, and `:post_install` hooks -- if anyone else is using those hooks and expecting the current directory while running the hook to be `@gem_dir` this could break compatibility.

Are these hooks considered part of a public API, or are they only used internally by `bundler` itself?

deivid-rodriguez: These are hooks originally run and registered by `rubygems` and the thing is, they don't actually switch to the installed gem folder when run by `rubygems` on a regular `gem install <gem_name>` command, or when run through `Gem::Installer` on other `bundler` sources, like the standard `source "https://rubygems.org`.

In this case, we run them manually because for the `git` source we don't use `Gem::Installer` to install the gem, but I don't see a reason why we should switch to the gem folder for this particular source.

deivid-rodriguez: In order words, in my eyes this as a bug fix, not a breaking change.

If some gem `foo` has a gem hook registered that depends on being on the gem's installation directory, the hook wouldn't be working at all, because `gem install` does not switch folders before running gem hooks.

joelnordell: Makes sense!

FWIW, in researching this warning message I ran across this:

https://bugs.ruby-lang.org/issues/15661

That issue points out that `Dir.chdir` has a _process-global_ effect. This has interesting implications for Bundler::Worker and parallel installs.

I'm sure that's been considered already, since I see some integration with the rubygems `CHDIR_MUTEX` - but nonetheless I plan to do a quick audit of the bundler code to make sure that all use of `chdir` is safe and won't have unexpected side effects if multiple workers are running in parallel.

deivid-rodriguez: Yeah, globally changing the PWD for compiling extensions is the biggest thread safety concern we have. It should be considered as you point out, but I'm not sure if everything is fully thread safe since we have some flaky test failures that seem to indicate the opposite.

I started some work on rubygems in relation to that which consists of no longer changing folders inside the ruby process when compiling extensions, but use `make -C`, `ruby -C` and so on instead.

deivid-rodriguez: Your audit will be super appreciated by the way :heart_eyes:.

joelnordell: Another possible way to address it could be to fork before building native extensions - the child process could then chdir all it wants without affecting anything else.

deivid-rodriguez: I guess, yeah.

So this PR looks good to you, right?

joelnordell: Yep, looks good

deivid-rodriguez:
@joelnordell I shared the work I had done regarding changing directories in: https://github.com/rubygems/rubygems/pull/3498.

casperisfine:
👋 this warning became a hard failure on ruby 3.0 just a couple days ago: https://bugs.ruby-lang.org/issues/15661.

So right now ruby-head testing is fairly hard. Any change we could get a new bundler version out with this fix?

deivid-rodriguez:
This PR was released with 2.2.0.rc.1, and it's also bundled with ruby-core now since it vendors 2.2.0.rc.1. Is that enough?

casperisfine:
My bad, I misread the date of 2.2.0.rc.1, I somehow saw it as released in January. Sorry!

deivid-rodriguez:
No problem!

PR #3481, opened 2020-04-03
Do not pre-install Bundler for RubyGems and Bundler tests

eregon:
It looks like `rake setup` already installs Bundler, so there is no need for `ruby/setup-ruby` to also install Bundler.


deivid-rodriguez:
Thanks @eregon!

PR #3482, opened 2020-04-03
Add ability to uninstall plugins

fatkodima:
Reopened #3438 

@deivid-rodriguez You are right (https://github.com/rubygems/rubygems/pull/3438#discussion_r403256897). Updated with your suggestion.

Closes #3438

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -47,6 +47,32 @@ module Bundler
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
     end
 
+    # Uninstalls plugins by the given names
+    #
+    # @param [Array<String>] names the names of plugins to be uninstalled
+    def uninstall(names, options)
+      if names.empty? && !options[:all]
+        Bundler.ui.error "No plugins to uninstall. Specify at least 1 plugin to uninstall.\n"\
+          "Use --all option to uninstall all the installed plugins."
+        return
+      end
+
+      names = index.installed_plugins if options[:all]
+      if names.any?
+        names.each do |name|
+          if index.installed?(name)
+            Bundler.rm_rf(index.plugin_path(name))
+            index.unregister_plugin(name)
+            Bundler.ui.info "Uninstalled plugin #{name}"
+          else
+            Bundler.ui.error "Plugin #{name} is not installed \n"
+          end
+        end
+      else
+        Bundler.ui.info "No plugins installed"
+      end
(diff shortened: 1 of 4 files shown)


bundler/lib/bundler/plugin.rb
```diff
@@ -47,6 +47,32 @@ def install(names, options)
...
+            Bundler.rm_rf(index.plugin_path(name))
+            index.unregister_plugin(name)
+            Bundler.ui.info "Uninstalled plugin #{name}"
+          else
+            Bundler.ui.error "Plugin #{name} is not installed \n"
+          end
+        end
+      else
+        Bundler.ui.info "No plugins installed"
+      end
```
deivid-rodriguez, line 73: I wonder if we should simplify this method to do something symmetric to what the `install` method:

* Extract an `Plugin:Uninstaller` class which takes care of the actual file removal and returns the list of uninstalled plugins. Maybe this class could be reused for the `rescue` block logic of the `install` method?

* Move the unregistration of the uninstalled plugins returned by the `Uninstaller` to an `unsave_plugins` helper.

Do you think this refactoring is worth doing?

bundler/lib/bundler/plugin.rb
```diff
@@ -47,6 +47,32 @@ def install(names, options)
       Bundler.ui.error "Failed to install plugin #{name}: #{e.message}\n  #{e.backtrace.join("\n ")}"
```
deivid-rodriguez, line 47: I was reading this block to see if some of the "uninstallation logic" could/should be reused here, and I'm wondering... where is the `name` variable being defined? Is this actually working/neeed? :thinking:

I know this is not your code, but if you're willing to double check this, I appreciate it.

fatkodima: This is needed, but not working. In eb5873fc0e method signature was changed, but this line stayed the same.
I think it is better to fix this in a separate pr.

bundler/lib/bundler/plugin.rb
fatkodima: Imo, this method is pretty simple right now and and we have better control with logging messages. I'd left it as it is for now.

deivid-rodriguez: Ok, let's leave it for now.

bundler/lib/bundler/plugin.rb
deivid-rodriguez: I opened #3484 to track it.

deivid-rodriguez:
Let's do it!

Thanks @fatkodima!

Issue #3140, opened 2020-02-18
Flaky `TestGemSource#test_dependency_resolver_set_file_uri` test failure


deivid-rodriguez:
I'm having a problem or would like to suggest a feature.

My current problem is a flaky test failure. See:

* https://github.com/rubygems/rubygems/runs/453291193.
* https://github.com/rubygems/rubygems/runs/459504331.
* https://github.com/rubygems/rubygems/pull/3165/checks?check_run_id=489809509.
* https://github.com/rubygems/rubygems/pull/3483/checks?check_run_id=561932495.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

deivid-rodriguez:
An initial clue is that this has happened twice so far, both times under Windows + MRI 2.5.

deivid-rodriguez:
It happened again on Windows + MRI 2.4.

So it does seem Windows specific, but not ruby version specific.

deivid-rodriguez:
One more: https://github.com/rubygems/rubygems/pull/3561/checks?check_run_id=629062286.

deivid-rodriguez:
Once again, https://github.com/rubygems/rubygems/pull/3562/checks?check_run_id=634233669. New failures include some more debugging information:

```
TestGemSource#test_dependency_resolver_set_file_uri:
Zlib::GzipFile::Error: #<Zlib::GzipFile::Error: not in gzip format, input="\x1F\x8B\b\x00\x11">
    D:/a/rubygems/rubygems/lib/rubygems/util.rb:15:in `initialize'
    D:/a/rubygems/rubygems/lib/rubygems/util.rb:15:in `new'
    D:/a/rubygems/rubygems/lib/rubygems/util.rb:15:in `gunzip'
    D:/a/rubygems/rubygems/lib/rubygems/test_utilities.rb:71:in `fetch_path'
    D:/a/rubygems/rubygems/lib/rubygems/test_utilities.rb:78:in `cache_update_path'
    D:/a/rubygems/rubygems/lib/rubygems/source.rb:186:in `load_specs'
    D:/a/rubygems/rubygems/lib/rubygems/spec_fetcher.rb:257:in `tuples_for'
[...]

PR #3485, opened 2020-04-05
Improve gzip errors logging

deivid-rodriguez:
# Description:

By default, the `Zlib::GzipFile::Error` does not include the actual data that was not in gzip format that caused the error.

However, its `#inspect` method includes it.

I think this can be helpful to troubleshoot errors and give us more information to debug #3140, which we keep getting intermittently.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/util.rb
+++ b/lib/rubygems/util.rb
@@ -11,7 +11,13 @@ module Gem::Util
     require 'stringio'
     data = StringIO.new(data, 'r')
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
We [hit this issue again on master today](https://github.com/rubygems/rubygems/runs/585656118), so I'm going to tentatively merge this to see more information the next time it happens.

PR #3487, opened 2020-04-05
Fix broken exception recovery code when installing plugins

### LINKED ISSUE #3484: Exception recovery code when installing plugins is broken
The code to recover from issues while installing plugins and undo any side effects made at that point seems broken.

In eb5873fc0e method signature was changed, but https://github.com/rubygems/rubygems/blob/2a9bae4c38231c5ee930be042edbc0dcec122085/bundler/lib/bundler/plugin.rb#L47 was not updated.

So, code at the moment seems both broken and not covered by any specs.

_Originally noticed at https://github.com/rubygems/rubygems/pull/3482_

fatkodima:
Fixes #3484

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -39,12 +39,11 @@ module Bundler
 
       save_plugins names, specs
     rescue PluginError => e
-      if specs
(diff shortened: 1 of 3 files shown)


fatkodima:
@deivid-rodriguez Updated with your suggestion.

deivid-rodriguez:
Thanks so much, I'll merge this tomorrow!

PR #3473, opened 2020-04-01
Use bundler to manage bundler development dependencies

deivid-rodriguez:
# Description:

This is an experiment I made today to try use bundler to manage bundler development dependencies. Still WIP.

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3488, opened 2020-04-05
Windows Actions - export TMPDIR=$RUNNER_TEMP

MSP-Greg:
# Description:

1. Actions Windows CI currently uses VM's hosted on an 8 core system.  There is a hard drive (C:) and an SSD (D:).  ENV['TEMP'] & ENV["TMP'] both point to the hard drive.  From various Actions work/testing, the hard drive can be much slower than the SSD.

2. For legacy reasons, Windows supports what are referred to as 'long' and 'short' (8.3) file names.  ENV['TEMP'] & ENV["TMP'] use 'short' names.  Never having researched every possible method call in Ruby, I can't say for sure, but string comparisons could cause issues.

Odd issues with Windows Actions CI

For the 'temp path', I believe Ruby uses [ ENV['TMPDIR'], ENV["TEMP'], ENV["TMP'] ] to select the folder.  Set ENV['TMPDIR'] = ENV['RUNNER_TEMP'], which is a long file name on the SSD.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/windows-bundler.yml
+++ b/.github/workflows/windows-bundler.yml
@@ -29,6 +29,7 @@ jobs:
         shell: bash
       - name: Run specs
         run: |
+          export TMPDIR=$RUNNER_TEMP
           bin/parallel_rspec spec
         working-directory: ./bundler
         shell: bash
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -25,6 +25,7 @@ jobs:
         shell: bash
       - name: Run Test
         run: |
+          $env:TMPDIR = $env:RUNNER_TEMP
           ridk enable
           rake test
     timeout-minutes: 15


MSP-Greg:
@deivid-rodriguez

I particularly liked the [screen grab](https://user-images.githubusercontent.com/2887858/78188599-cc3e1000-7470-11ea-9c35-6ed761429c75.png) in https://github.com/rubygems/rubygems/pull/3473#issuecomment-607500369, but I think https://github.com/rubygems/rubygems/pull/3488/checks?check_run_id=562758765 may be even better.

Especially considering that the two files changed in this PR were Window Actions yaml files...

deivid-rodriguez:
@MSP-Greg I was very excited about making our Windows CI faster, but I don't see any speed up? :thinking:

simi:
> For the 'temp path', I believe Ruby uses [ ENV['TMPDIR'], ENV["TEMP'], ENV["TMP'] ] to select the folder.

I was looking at the same recently. You're right (it is just in different order) https://github.com/ruby/ruby/blob/98f6c74b429f9e8afccb000da4a50920479dffd6/lib/tmpdir.rb#L23.

deivid-rodriguez:
Where does the checked out code run?

I think that this shows no improvement, because it doesn't actually get used, since bundler specs reset it to the local `tmp/` folder:

https://github.com/rubygems/rubygems/blob/137aa0201ebc398b4f42fc3bda5944acf191a49e/bundler/spec/support/rubygems_ext.rb#L37

MSP-Greg:
@simi Thanks.  I knew TMPDIR was first, didn't check the remaining two...

@deivid-rodriguez

1. I didn't say significant time savings.  There was a significant difference when downloading and extracting a large compressed file (containing several thousand files).  YMMV.  Disk I/O speed may or may not affect test times.

2. I didn't check if Bundler changes TMPDIR.  If it uses its own, never mind.

Closing.

deivid-rodriguez:
I see, thanks.

PR #3498, opened 2020-04-08
Don't change ruby process CWD when building extensions

deivid-rodriguez:
# Description:

This leads to simpler code because we don't need to maintain a mutex to ensure thread safety.

# Tasks:

- [ ] Write tests


eregon:
This looks great!
It should enable to build C extensions in parallel which could lead to very nice speedups for large Gemfiles.

deivid-rodriguez:
I had forgotten about this. I rebased it!

eregon (approved):
Looks good to me!

deivid-rodriguez:
Let's do it!

PR #2797, opened 2019-06-19
Test TruffleRuby in CI

eregon:
# Description:

We'd like to test TruffleRuby in RubyGems' CI, but it's not working yet.
This is related to #2700, but with additional fixes to run tests.

cc @hsbt 

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3499, opened 2020-04-08
Update test_realworld_default_gem to use rubygems project sources

bjfish:
# Description:

The current test uses `popen` and appears to load `rubygems` from the ruby implementation instead of the `rubygems` from the project directory under test.

This was found while test spec failures for https://github.com/rubygems/rubygems/pull/2797.

The fix is to pass the `rubygems` project `lib` dir to the `popen` command to ensure correct files are loaded.

______________


bjfish:
Pushed a fix to make the path relative to `__FILE__`.

PR #3506, opened 2020-04-10
Deprecate rubyforge_project attribute only during build time.

simi:
:information_source: initial idea as discussed on Slack with @deivid-rodriguez

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -6,6 +6,7 @@
 # See LICENSE.txt for permissions.
 #++
 
+require 'set'
 require 'rubygems/version'
 require 'rubygems/requirement'
 require 'rubygems/platform'
@@ -192,6 +193,12 @@ class Gem::Specification < Gem::BasicSpecification
   NOT_FOUND = Struct.new(:to_spec, :this).new # :nodoc:
   @@spec_with_requirable_file = {}
   @@active_stub_with_requirable_file = {}
+
+  # Tracking removed method calls to warn users during build time.
+  REMOVED_METHODS = [:rubyforge_project=].freeze # :nodoc:
+  def removed_method_calls
+    @removed_method_calls ||= Set.new
+  end
 
   ######################################################################
   # :section: Required gemspec attributes
@@ -726,13 +733,6 @@ class Gem::Specification < Gem::BasicSpecification
   # Allows deinstallation of gems with legacy platforms.
 
   attr_writer :original_platform # :nodoc:
-
-  ##
-  # Deprecated via specification policy and ignored during runtime.
-  #
-  # Formerly used to set rubyforge project.
-
-  attr_accessor :rubyforge_project
 
   ##
   # The Gem::Specification version of this gemspec.
@@ -2097,9 +2097,15 @@ class Gem::Specification < Gem::BasicSpecification
   end
 
   ##
+  # Track removed method calls to warn about during build time.
   # Warn about unknown attributes while loading a spec.
 
   def method_missing(sym, *a, &b) # :nodoc:
+    if REMOVED_METHODS.include?(sym)
+      removed_method_calls << sym
+      return
+    end
+
     if @specification_version > CURRENT_SPECIFICATION_VERSION and
       sym.to_s =~ /=$/
       warn "ignoring #{sym} loading #{full_name}" if $DEBUG
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -728,12 +728,11 @@ class Gem::Specification < Gem::BasicSpecification
   attr_writer :original_platform # :nodoc:
 
   ##
-  # Deprecated and ignored.
+  # Deprecated via specification policy and ignored during runtime.
   #
   # Formerly used to set rubyforge project.
 
-  attr_writer :rubyforge_project
-  rubygems_deprecate :rubyforge_project=
+  attr_accessor :rubyforge_project
 
   ##
   # The Gem::Specification version of this gemspec.
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -6,7 +6,6 @@
 # See LICENSE.txt for permissions.
 #++
 
-require 'set'
 require 'rubygems/version'
 require 'rubygems/requirement'
 require 'rubygems/platform'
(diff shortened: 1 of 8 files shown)


lib/rubygems/specification.rb
```diff
@@ -728,12 +728,11 @@ def test_files=(files) # :nodoc:
...
 
   ##
-  # Deprecated and ignored.
+  # Deprecated via specification policy and ignored during runtime.
   #
   # Formerly used to set rubyforge project.
 
-  attr_writer :rubyforge_project
-  deprecate :rubyforge_project=
+  attr_accessor :rubyforge_project
```
simi, line 735: This makes `rubyforge_project` accessible again. Is that problem for now?

lib/rubygems/specification_policy.rb
```diff
@@ -409,6 +413,12 @@ def validate_shebang_line_in(executable)
     warning "#{executable_path} is missing #! line"
   end
 
+  def validate_deprecated_attributes
+    DEPRECATED_ATTRIBUTES.each do |attr|
+      warning("#{attr} is deprecated") unless @specification.send(attr).nil?
```
simi, line 418: Any idea for better deprecation warning here?

simi:
:information_source: initial testing went well. Gem build of gemspec including `rubyforge_project` warns for now:

```
[retro@retro  rubygems (warn-about-rubyforge-during-build *%=)]❤ gem build rubygems-update.gemspec 
WARNING:  rubyforge_project is deprecated
  Successfully built RubyGem
  Name: rubygems-update
  Version: 3.2.0.pre1.pre.test
  File: rubygems-update-3.2.0.pre1.pre.test.gem
```

And application (after local `rake install`) doesn't warn about `rubyforge_project` deprecations.

lib/rubygems/specification_policy.rb
simi: Also it would be possible to `error` later in here.

simi:
:information_source: :red_circle: JRuby build seems unrelated.

olleolleolle:
Nice!

deivid-rodriguez:
Thanks @simi, I plan to get to this very soon.

lib/rubygems/specification_policy.rb
deivid-rodriguez: Should we follow the same pattern as in `validate_required_attributes`? Defining the list in `Gem::Specification` in that same way would allow us to reuse https://github.com/rubygems/rubygems/pull/3051 and completely get rid of the `rubyforge_project` accessor.

deivid-rodriguez: Thinking of a good message to give to users, I believe we should probably call this `validate_removed_attributes` since what they have in common is that the fix is to remove the attribute from the gemspec. We could say something like:

```suggestion
      warning("#{attr} is deprecated and ignored. Please remove this attribute from your gemspec to ensure that your gem continues to build in the future.") unless @specification.send(attr).nil?
```

Or a version of that communicating the same message but maybe a bit less threatening :sweat_smile:

simi: Sure, that message was just a placeholder for now.

deivid-rodriguez: Just to make sure @simi, you're planning to continue the work here, right?

simi: on it now

simi:
@deivid-rodriguez feel free to check latest commit with little different approach per your comment at https://github.com/rubygems/rubygems/pull/3506#discussion_r410288112.

lib/rubygems/specification.rb
```diff
@@ -193,6 +194,12 @@ class Gem::Specification < Gem::BasicSpecification
   @@spec_with_requirable_file = {}
   @@active_stub_with_requirable_file = {}
 
+  # Tracking removed method calls to warn users during build time.
+  REMOVED_METHODS = [:rubyforge_project=].freeze #:nodoc:
```
olleolleolle, line 198:
```suggestion
  REMOVED_METHODS = [:rubyforge_project=].freeze # :nodoc:
```

Is that space there significant?

test/rubygems/test_gem_specification.rb
```diff
@@ -3110,6 +3110,21 @@ def test_validate_license
     warning
   end
 
+  def test_removed_methods
+    assert_equal Gem::Specification::REMOVED_METHODS, [:rubyforge_project=]
```
olleolleolle, line 3114: Perhaps this could be a check that the list _includes_ `:rubyforge_project=`?

lib/rubygems/specification.rb
simi: Thanks, I'll fix! This change is not intentional.

simi:
@bronzdoc ok, I'll add.

simi:
@bronzdoc test added. I had to use gemspec file since `rubyforge_project` do not survive `to_ruby` "serialization" used for preparing gemspecs in other build tests (which is actually great).

deivid-rodriguez:
I'll do another review pass tomorrow.

Just note that to reproduce the issue you need to downgrade rubygems to pre-3.0, install a gem with `rubyforge_project`, and then upgrade rubygems back.

Installing a gem with the `rubyforge_forge` attribute with a modern rubygems does not suffer from this issue because as @simi pointed out, `rubyforge_project` does not survive `to_ruby` "serialization".

By the way, an idea to be able to properly get rid of deprecated attributes like this would be to "rewrite" all stubbed gemspecs during rubygems upgrade. Not for this PR, but just an idea I had.

lib/rubygems/specification.rb
```diff
@@ -193,6 +194,12 @@ class Gem::Specification < Gem::BasicSpecification
   @@spec_with_requirable_file = {}
   @@active_stub_with_requirable_file = {}
 
+  # Tracking removed method calls to warn users during build time.
+  REMOVED_METHODS = [:rubyforge_project=].freeze # :nodoc:
+  def removed_method_calls
```
deivid-rodriguez, line 199: Could we use a plain Array + `uniq`? I could imagine Hiroshi making `set` a default gem, and us having trouble here because of using it... :sweat_smile: Bundler already uses `set` but `rubygems` doesn't I believe so I guess if we can avoid it we should.

simi: sure, np

simi: I have removed set and moved to array. `rubyforge_project=` should be usually called only once during gemspec evaluation.

PR #3507, opened 2020-04-10
Skip flaky tests on jruby

deivid-rodriguez:
# Description:

This is a temporary workaround for https://github.com/rubygems/rubygems/issues/3451.

It seems most likely a problem in either jruby or webrick, so until fixed upstream let's skip it so that it doesn't get in the middle.

Related upstream tickets:

https://github.com/ruby/webrick/pull/40
https://github.com/jruby/jruby/issues/6171

# Tasks:

- [x] Describe the problem / feature
- [ ] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3508, opened 2020-04-10
Cleanup remote fetcher tests

deivid-rodriguez:
# Description:

While working on #3507, I noticed that test servers are created at the class level of each test (so code is "reused" across tests), and I don't think that's necessary. I actually wonder if that could actually be leaving some server threads around that might be interfering somehow.

So I moved everything to the instance level since it's simpler.

Since I was at it, I also remove some unnecessary `DIR` constant. Builtin `__dir__` does the same thing.

The patch is very simple if you look at it [excluding whitespace](https://github.com/rubygems/rubygems/pull/3508/files?w=1).

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_remote_fetcher.rb
+++ b/test/rubygems/test_gem_remote_fetcher.rb
@@ -78,8 +78,6 @@ gems:
 
   PROXY_DATA = SERVER_DATA.gsub(/0.4.11/, '0.4.2')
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks for having a look!

simi:
:+1:

PR #3504, opened 2020-04-09
Run ruby/ruby testing

MSP-Greg:
# Description:

## What was the end-user or developer problem that led to this PR?

ruby/ruby CI uses vendored test libraries.  When changes from this repo are pushed to ruby/ruby, failures occasionally occur.  PR downloads the ruby/ruby test files and uses them to run CI here.

## What is your fix for the problem, implemented in this PR?

1. Create util/ruby_core_test.rb, which downloads the files from ruby/ruby.
2. A few files require small fixes.
3. Add a workflow file to run the CI

Note: the 'builder' gem is enabled for CI in this PR, but the 'builder' gem is not used in ruby/ruby CI.

______________

# Tasks:

- [x] Describe the problem / feature
- [ ] Write tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3510, opened 2020-04-11
Try to run ruby core test with rubygems repo

hsbt:
# Description:

Followed up https://github.com/rubygems/rubygems/pull/3504

I sometimes faced the test failure with the code of this repository on ruby core test suite. I try to run rubygems test with ruby core test suite with HEAD commit of ruby repo.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/.github/workflows/ruby-core-test-suite.yml
+++ b/.github/workflows/ruby-core-test-suite.yml
@@ -25,7 +25,7 @@ jobs:
           mkdir rubygems
       - uses: actions/checkout@v2
         with:
-          repository: "ruby/ruby"
+          repository: ruby/ruby
           path: ruby/ruby
       - name: Build Ruby
         run: |
--- a/.github/workflows/ruby-core-test-suite.yml
+++ b/.github/workflows/ruby-core-test-suite.yml
@@ -1,4 +1,4 @@
-name: ruby-core-test
+name: ruby-core
 
 on:
   pull_request:
@@ -8,7 +8,7 @@ on:
       - master
 
 jobs:
-  ruby_core_test:
+  ruby_core:
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/ruby-core-test-suite.yml
+++ b/.github/workflows/ruby-core-test-suite.yml
@@ -1,0 +1,52 @@
+name: ruby-core-test
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - master
+
+jobs:
+  ruby_core_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        target: [rubygems, bundler]
+    steps:
+      - name: Install libraries
+        run: |
+          set -x
+          sudo apt-get update -q || :
+          sudo apt-get install --no-install-recommends -q -y build-essential libssl-dev libyaml-dev libreadline6-dev zlib1g-dev libncurses5-dev libffi-dev libgdbm-dev bison autoconf ruby
+      - name: Prepare the working directory
+        run: |
+          mkdir ruby
+          mkdir rubygems
+      - uses: actions/checkout@v2
+        with:
+          repository: "ruby/ruby"
+          path: ruby/ruby
+      - name: Build Ruby
+        run: |
+          autoconf
+          ./configure -C --disable-install-doc
+          make -j2
+        working-directory: ruby/ruby
+      - uses: actions/checkout@v2
+        with:
+          path: rubygems/rubygems
+      - name: Test RubyGems
+        run: |
+          ruby tool/sync_default_gems.rb rubygems
+          make test-all TESTS="rubygems -j2"
+        working-directory: ruby/ruby
+        continue-on-error: true
+        if: matrix.target == 'rubygems'
+      - name: Test Bundler
+        run: |
+          ruby tool/sync_default_gems.rb bundler
+          make test-bundler
+        working-directory: ruby/ruby
+        continue-on-error: true
+        if: matrix.target == 'bundler'


hsbt:
@deivid-rodriguez Should we move it to daily jobs?

.github/workflows/ruby-core-test-suite.yml
```diff
@@ -0,0 +1,52 @@
...
+        working-directory: ruby/ruby
+      - uses: actions/checkout@v2
+        with:
+          path: rubygems/rubygems
+      - name: Test RubyGems
+        run: |
+          ruby tool/sync_default_gems.rb rubygems
+          make test-all TESTS="rubygems -j2"
+        working-directory: ruby/ruby
+        continue-on-error: true
```
deivid-rodriguez, line 44: I see that the check is actually broken at the moment (I seem to have broken it again when moving our setup to use bundler to manage development dependencies?).

So I guess the idea is to start with this as an allowed failure, and once we fix the current situation, we start enforcing it to be green, right?

That approach sounds good to me! :+1:

.github/workflows/ruby-core-test-suite.yml
```diff
@@ -0,0 +1,52 @@
+name: ruby-core-test
```
deivid-rodriguez, line 1: I would call the workflows and the job simply "ruby-core", and "ruby_core" respectively!

MSP-Greg:
One problem with with this type of CI (vs #3504) is that there is no means of determining if ruby/ruby is currently failing its own CI.  It may generate false failures when that is the case, depending on whether the failure affects code in RG.

[ruby/setup-ruby](https://github.com/ruby/setup-ruby) (and hence, [MSP-Greg/setup-ruby-pkgs](https://github.com/MSP-Greg/setup-ruby-pkgs)) will only provide passing ruby/ruby builds.  The full set of ruby/ruby tests suites is ran before they replace the 'current build'.

This same issue affects Bundler CI.  Since it uses its own test system, it's not coupled to head like RG is, and hence, using the builds provided by [ruby/setup-ruby](https://github.com/ruby/setup-ruby) provides more stable CI.

deivid-rodriguez:
You're right @MSP-Greg, unfortunately your check is superior to this one in that sense, and now that you pointed this out I'm reluctant to introduce something like this for every PR (and it kind of defeats the point it we do it on a scheduled basis).

However, I agree with the problems pointed out by @hsbt in #3504, specially in terms of maintainability. Your PR is very brittle since it included a patch that might no longer apply, a hardcoded list of files that might be renamed, no longer be enough, or whatever. Also, a bunch of code that we'd need to start maintaining include some powershell logic that we don't really want to maintain.

Ideally we run rubygems & bundler tests using the 'head' ruby of `ruby/setup-ruby`, but against the structure/setup of the checked out ruby-core repo. Can we set the `RUBY` environment variable or something like that to achieve that and avoid having to compile ruby, for example?

Also, is not pushing to master directly without green CI something the ruby-core has actually considered?

deivid-rodriguez:
I think we should maybe focus on #3141 which would reduce differences and thus make this kind of issues less frequent?

hsbt:
I only run `make (ruby) -j` for building ruby interpreter and test rubygems and bundler. The ruby core team always keep `make ruby`. We faced the test failures for `make test-all` not `make ruby`.

MSP-Greg:
@hsbt

Simply because `make ruby` passes doesn't mean it works, it juts means it compiles.  It only works if all the test suites pass.

MSP-Greg:
@deivid-rodriguez

> Your PR is very brittle since it included a patch

That patch may not be needed, but it 'bullet-proofs' the accidental loading of files contained in the build.  The patch is short, the file the patch operates on is short, and the file is stable.

> start maintaining include some powershell logic

Both of you make it sound like some screwy thing I pulled from a distant hidden cave full of IT renegades.  I've seen articles mentioning how much adoption there's been on macOS & *nix platforms.  Again, it's not that much code, and unlikely to need changes.  It might be able to be converted to bash for Windows, but may be brittle for a lot of reasons I don't want to get into.

> Also, is not pushing to master directly without green CI something the ruby-core has actually considered?

It is master.  But, reminder that some of the failures happen due to external repos (like RG, Bundler, and the std-lib gems) pushing changes.  Which is why we talking about this CI issue...

deivid-rodriguez:
> I only run make (ruby) -j for building ruby interpreter and test rubygems and bundler. The ruby core team always keep make ruby. We faced the test failures for make test-all not make ruby.

If compiling ruby is considered stable enough, I'm happy to try this. I think @hsbt knows this better than myself, so ok with any decision.

deivid-rodriguez:
@MSP-Greg Thank you for your work in #3504, but both @hsbt and I agree of not following that approach. It's not just the powershell script, or the `.patch` file, or the hardcoded list of files to pull from the ruby-core repo, it's the combination of everything. So I don't think we should discuss that further, since I think that's why @hsbt locked #3504 :sweat_smile:.

By the way, I just opened https://bugs.ruby-lang.org/projects/ruby-master/activity to try to address a potential root cause of all this.

MSP-Greg:
Main ruby/ruby tests on Ubuntu, macOS, mingw, and mswin.  Additionally, JRuby & TruffleRuby might be given consideration.

.github/workflows/ruby-core-test-suite.yml
hsbt: Yes, The test of current master branch with ruby master failed now. I will work after merging this.

hsbt:
The scope of CI is difficult. The mainly motivation is finding to breakage changes of rubygems for ruby core like this https://github.com/rubygems/rubygems/runs/578285555#step:7:28

deivid-rodriguez:
Thanks for working on this @hsbt! Hopefully we can catch ruby-core testing issues earlier from now on! :purple_heart:

PR #3514, opened 2020-04-12
CmakeLists.txt eruby annotation

golirev:
Currently, CMakeLists.txt can be specified in the spec file instead of extconf.rb.
However, unlike extconf.rb, the method of reflecting the information of the Ruby interpreter was unknown.

With this change, the information of the Ruby interpreter can be reflected using eRuby.
eg: s.extensions = %w[ext/CMakeLists.txt.erb]


hsbt:
It's out of scope for specification of RubyGems. You can add this workflow into your local Rakefile.

Issue #3448, opened 2020-03-27
Random test failure on Windows (output syncronization issue?)


deivid-rodriguez:
Sometimes, we get the following spec failure under Windows:

```
Failures:

  1) bundle install with git sources with an extension does reinstall the extension when changing refs

     Failure/Error: raise "Invoking #{method}!(#{args.map(&:inspect).join(", ")}) failed:\n#{last_command.stdboth}"

     RuntimeError:
       Invoking run!("        require 'foo'\n        puts FOO\n") failed:
       d:/a/rubygems/rubygems/bundler/lib/bundler/source/git/git_proxy.rb:241:in `allowed_in_path': The git source d:/a/rubygems/rubygems/bundler/tmp/1/libs/foo-1.0 is not yet checked out. Please run `bundle install` before trying to start your application (Bundler::GitError)
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git/git_proxy.rb:198:in `find_local_revision'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git/git_proxy.rb:68:in `revision'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git.rb:226:in `revision'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git.rb:93:in `install_path'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/path.rb:131:in `expanded_path'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/path.rb:172:in `load_spec_files'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git.rb:201:in `load_spec_files'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/path.rb:105:in `local_specs'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/source/git.rb:168:in `specs'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/definition.rb:768:in `block in converge_locked_specs'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/definition.rb:754:in `each'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/definition.rb:754:in `converge_locked_specs'
       	from d:/a/rubygems/rubygems/bundler/lib/bundler/definition.rb:252:in `resolve'
[...]

PR #3397, opened 2020-03-17
Added Ruby 2.7 and RubyGems 3.1.2 to CI

hsbt:
# Description:

Fixed https://github.com/rubygems/rubygems/issues/3261

Related with https://github.com/rubygems/bundler/pull/7593
______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3516, opened 2020-04-14
No chdir on git operations

deivid-rodriguez:
# Description:

While working on #3397, I was getting a lot of thread safety issues in tests that perform git operations. Bundler is supposed to reuse the mutex rubygems uses for building extensions that need to change folders for compilation in a thread safe manner. I couldn't stop the issue, but something in there seems not thread safe.

My solution, since I'm bad at spotting thread safety issues, is to fix the root cause instead and not change the CWD when performing `git` operations.

With these changes, I managed to get a Windows passing CI for bundler under ruby 2.7.

I'm extracting the change here though to make the change more focused and explain it.

This PR also includes some refactoring I made in the `GitProxy` class to be able to remove the dependency on the CWD. 

This PR should also close https://github.com/rubygems/rubygems/issues/3448.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -169,7 +169,7 @@ module Bundler
           end
         end
 
-        def git(command, check_errors = true, error_msg = nil)
+        def git(command, check_errors = true)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
#3397 passed again with this changes! :tada: I plan to merge this tomorrow.

PR #3517, opened 2020-04-14
Improve specs debugability

deivid-rodriguez:
# Description:

While working on #3397, I made several changes to improve the debugability of bundler specs. In particular:

* Fail earlier when a `bundle install` command fails, so that the culprit is more apparent.
* When building test git repos with git subcommands, make sure those commands are properly recorded and their results are logged in case of errors.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This should make it easier to debug problems in other PR, so I'll merge it!

PR #3525, opened 2020-04-17
Fix environment backup and restoring on Windows

deivid-rodriguez:
# Description:

This PR fixes several failing specs about backing & restoring the environment on Windows. It also simplifies the specs because they were failing as well due to complicated command lines that were not properly shellescaped.

Environment backup and restore specs are failing on Windows.
 
`ENV` is case insensitive on Windows, but on other platforms is not. However, our `EnvironmentPreserver` class operates on ruby hashes, so it's case sensitive.
    
So whenever we use `ENV.to_hash`, we need to make sure to canonalize the keys, so they are always UPCASED.
    
In addition to that, due to [a bug in ruby](https://bugs.ruby-lang.org/issues/16798), `ENV.replace` doesn't work when using the canonical name for an environment variable does not match the case being passed to it. So we workaround that by using `ENV.clear` + manual setting on each key instead.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -34,9 +34,9 @@ require_relative "bundler/build_metadata"
 # of loaded and required modules.
 #
 module Bundler
-  environment_preserver = EnvironmentPreserver.new(ENV, EnvironmentPreserver::BUNDLER_KEYS)
+  environment_preserver = EnvironmentPreserver.from_env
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
The bug was fixed in ruby-core, so we'll be able to get rid of some of this code at some point.

PR #3526, opened 2020-04-18
Fix a bunch of test issues on Windows

deivid-rodriguez:
# Description:

Some specs where reporting "corrupt test gem" because the `.gem` was
being read in non-binary mode, and that made the header size of the gem
not match the actual size and caused tar entry reading to crash.

I changed the specific calls from the fake servers classes, but also
changed all the other `File.read` calls to `File.binread` to prevent
similar future issues.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This PR brings down the number of test failures on Windows from 287 to 183! Slowly getting there :tada:

PR #3528, opened 2020-04-18
Fix symlink RubyGems test problems for non-admin user.

simi:
👋

I was about to check current test suite status on Windows as well and I have spotted RubyGems tests fail when started as non-admin user because of symlinks support.

I have added skip for one more test and added branched assert for second one since it fallbacks to different code when symlink fails to create. I have chosen this approach instead of mostly duplicating test (with two different asserts in the end).

I had to move symlink_supported? method to shared installer test case and I have enhanced comment a little to reflect current state.


deivid-rodriguez:
It'd be nice to ensure that our tests pass on Windows under a non-admin user, to prevent Windows developer from running into these issues in the future. I first learnt about these issues not currently being catched by our CI here: https://github.com/rubygems/rubygems/pull/2961.

simi:
@deivid-rodriguez on CI? With this change rubygems works well. I'll take a look at bundler at some point as well.

deivid-rodriguez:
Yeah, on CI.

I mean, CI was passing on Windows without this changes, but the tests were actually broken for a developer running ruby on windows locally without admin privileges.

It'd be nice to be able to ensure that we don't break it again by for example not adding a needed skip for a test that needs symlinks. Anyways, no big deal.

simi:
Yup, I understand. I have raised question if there's any easy way on CI - https://github.community/t5/GitHub-Actions/Windows-runner-with-non-admin-user/td-p/54648. It could be probably scripted as well, but I'm not sure how that works.

Issue #3512, opened 2020-04-11
Restore `Gem::Deprecate` to the way it worked in the previous release


deivid-rodriguez:
We changed `Gem::Deprecate` recently to consider versions instead of dates & months when deprecating stuff.

I think this is a better way of deprecating things, but unfortunately, libraries out there extend their classes with `Gem::Deprecate` and use it to deprecate methods this way.

So we should probably restore `Gem::Deprecate` to work in the same way as before, and add our new _private_ mechanism for deprecating rubygems things using version horizons instead of dates.

_Originally found by @simi in https://github.com/rubygems/rubygems/pull/3414#issuecomment-612507160_

simi:
Seems like release blocker for me, is there any label for this kind of problem?

deivid-rodriguez:
I added it to the next minor's milestone, so that we don't release it without closing this ticket first. Sounds good?

deivid-rodriguez:
We don't have such a label, but I guess we could add it as a new `type: release blocker` :shrug:

simi:
> Sounds good?

Yes!

bronzdoc:
Ah bummer 😒

bronzdoc:
I guess we will have to put deprecations in our deprecation method 😄

simi:
@bronzdoc or accept fact what has happened with `Gem::Deprecate` over time and keep it as is and roll internal deprecation for internal usage. IMHO we need to rollback this change back an prepare proper (maybe RFC based) approach how to change/deprecate `Gem::Deprecate` over the time.

simi:
:information_source: just got into this trouble again today using `rqrcode_core` gem. Should I prepare rollback for now and can we discuss the other part later in separated case? We can also sync on Slack if needed.

bronzdoc:
I'll submit a PR soon @simi

PR #3530, opened 2020-04-19
Rollback gem deprecate

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/3512

This PR does a few things:
- Restore old depreciation method 
- Deprecates old deprecation method
- Rename new version horizon deprecation methods

### DIFF
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -51,7 +51,7 @@ module Gem::Deprecate
   # telling the user of +repl+ (unless +repl+ is :none) and the
   # Rubygems version that it is planned to go away.
 
-  def deprecate(name, replacement=:none)
+  def rubygems_deprecate(name, replacement=:none)
     class_eval do
       old = "_deprecated_#{name}"
       alias_method old, name
@@ -70,7 +70,7 @@ module Gem::Deprecate
   end
 
   # Deprecation method to deprecate Rubygems commands
-  def deprecate_command
+  def rubygems_deprecate_command
     class_eval do
       define_method "deprecated?" do
         true
@@ -86,6 +86,6 @@ module Gem::Deprecate
     end
   end
 
-  module_function :deprecate, :deprecate_command, :skip_during
+  module_function :rubygems_deprecate, :rubygems_deprecate_command, :skip_during
 
 end
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -49,6 +49,31 @@ module Gem::Deprecate
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
   # telling the user of +repl+ (unless +repl+ is :none) and the
+  # year/month that it is planned to go away.
+
+  def deprecate(name, repl, year, month)
+    class_eval do
+      old = "_deprecated_#{name}"
+      alias_method old, name
+      define_method name do |*args, &block|
+        klass = self.kind_of? Module
+        target = klass ? "#{self}." : "#{self.class}#"
+        msg = [ "NOTE: #{target}#{name} is deprecated",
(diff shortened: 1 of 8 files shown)


simi:
I think this is good enough for now. We can take a look at internal deprecation improvements later.

lib/rubygems/deprecate.rb
```diff
@@ -45,13 +45,38 @@ def self.next_rubygems_major_version # :nodoc:
...
+      old = "_deprecated_#{name}"
+      alias_method old, name
+      define_method name do |*args, &block|
+        klass = self.kind_of? Module
+        target = klass ? "#{self}." : "#{self.class}#"
+        msg = [ "NOTE: #{target}#{name} is deprecated",
+                repl == :none ? " with no replacement" : "; use #{repl} instead",
+                ". It will be removed on or after %4d-%02d-01." % [year, month],
+                "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
+        ]
```
deivid-rodriguez, line 65: I believe we are introducing here the same kind of issue we're trying to solve in #3506: we're showing a deprecation message only meant for gem authors to all users. And we're showing this deprecation message about the deprecation system right next to the deprecation of a method that end users _do_ care about.

I feel we might cause brain damage to our users :laughing:.

So, I believe we should revert the deprecation of `Gem::Deprecate` for now until we figure out good strategies to be smart about whether we are in a end user context or in a gem development context, and apply them to this particular case.

simi: I have missed that. I agree on removing this deprecation warning for now.

deivid-rodriguez: FWIW, I placed this comment at the wrong place, my comments refers to the line _below_ this one, just to clarify.

lib/rubygems/deprecate.rb
```diff
@@ -70,7 +95,7 @@ def deprecate(name, replacement=:none)
   end
 
   # Deprecation method to deprecate Rubygems commands
-  def deprecate_command
```
deivid-rodriguez, line 73: I think the names convey it quite well, but still we should make it super clear that this is an internal module, not a public one that can be used by anyone. Not sure which strategies are tipically used for this.

lib/rubygems/deprecate.rb
```diff
@@ -45,13 +45,38 @@ def self.next_rubygems_major_version # :nodoc:
...
+  end
+
   ##
   # Simple deprecation method that deprecates +name+ by wrapping it up
   # in a dummy method. It warns on each call to the dummy method
   # telling the user of +repl+ (unless +repl+ is :none) and the
   # Rubygems version that it is planned to go away.
 
-  def deprecate(name, replacement=:none)
+  def rubygems_deprecate(name, replacement=:none)
```
deivid-rodriguez, line 79: I think the new name conveys it quite well, but we should make it super clear that this is an internal module, not a public one for everyone to use. Not sure which strategies are tipically used for this.

lib/rubygems/deprecate.rb
bronzdoc: I'm not sure if there's other way to do this without end users seeing the deprecation warning. 🤔 

This is not something that we could check at build time...

deivid-rodriguez: I think we can try to approximate it. I first idea that came to my mind was whether there's a ".gemspec" file at the CWD or up in the directory hierarchy.

simi: Can we just remove it for now (should I provide PR?) and discuss this as a follow-up in new issue?

deivid-rodriguez: Yes, yes, that's exactly what I was suggesting.

simi: https://github.com/rubygems/rubygems/pull/3536

PR #3531, opened 2020-04-19
When possible, prefer String#start_with? and String#end_with? over regex.

olleolleolle:
# Description:

This PR replaces a few regular expressions with String method usage.

- In one of the cases, filenames were checked for ending with "gz" - this is changed to check for ending with ".gz"
- The change was made to make it even easier to read the code, and to match only from the start of the input (as opposed to start of the line)

I read a PR, noted that in unrelated code there was a regex where we didn't need it. I made sure end_with? & start_with? were implemented on String in the lowest-supported version. According to https://rubyapi.org/2.3/o/String#method-i-start_with-3F it was present in 2.3.

Use `String#start_with?` and `String#end_with?` over regular expressions.

I picked this because I guessed they read easier and were quicker to run.

**Update**: this PR also enables the `rubocop-performance` cop `Performance/StartWith` to monitor this over time.

______________

# Tasks:

- [ ] Write tests


olleolleolle:
👋 @simi Hi! When I read the code near the changes you made, I saw these.

simi:
:wave: @olleolleolle personally I'm ok with this change, but I have no idea what's rubygem policy on those changes (not really changing behaviour).

deivid-rodriguez:
I'm happy with changes making the style more consistent, and I normally apply them just like @olleolleolle did here: as I notice the inconsistencies during code reviews. But I normally like to enable a corresponding rubocop cop, to ensure that we stay consistent in the future. Is there a cop for this?

lib/rubygems/command.rb
```diff
@@ -456,7 +456,7 @@ def add_extra_args(args)
     until extra.empty? do
       ex = []
       ex << extra.shift
-      ex << extra.shift if extra.first.to_s =~ /^[^-]/
+      ex << extra.shift unless extra.first.to_s.start_with?("-")
```
olleolleolle, line 459: This changed the behavior! Apologies.

olleolleolle: This _would_ do the right thing, but at the cost of looking sad:

```ruby
      if (first_character = extra.first.to_s && first_character && !first_character.empty? && first_character.start_with?("-"))
        ex << extra.shift
      end
```

bronzdoc: Let's stick with `extra.shift if extra.first.to_s =~ /^[^-]/` for this then 😁

olleolleolle: @bronzdoc The sad was reduced (a little bit) in the above (I edited it in-place)..

olleolleolle:
> I'm happy with changes making the style more consistent, and I normally apply them just like @olleolleolle did here: as I notice the inconsistencies during code reviews. But I normally like to enable a corresponding rubocop cop, to ensure that we stay consistent in the future. Is there a cop for this?

One such cop is in the rubocop-performance gem: [Performance/StartWith](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Performance/StartWith).

Enabling it, I found more instances.

lib/rubygems/installer.rb
```diff
@@ -586,7 +586,7 @@ def shebang(bin_file_name)
     path = File.join gem_dir, spec.bindir, bin_file_name
     first_line = File.open(path, "rb") {|file| file.gets}
 
-    if /\A#!/ =~ first_line
+    if first_line&.start_with?("#!")
```
simi, line 589: This will bring first occurence of this syntax to RubyGems. Is that intentional? Wouldn't be `first_line.to_s.start_with` safer at least for now?

olleolleolle: I can make that change, to be less controversial in this PR.

lib/rubygems/installer.rb
```diff
@@ -586,7 +586,7 @@ def shebang(bin_file_name)
     path = File.join gem_dir, spec.bindir, bin_file_name
     first_line = File.open(path, "rb") {|file| file.gets}
 
-    if /\A#!/ =~ first_line
+    if first_line.to_s.start_with?("#!")
```
simi, line 589: alt you we can "type-cast" only if needed by changing line above to 

```
first_line = File.open(path, "rb") {|file| file.gets} || ""
```

lib/rubygems/command.rb
```diff
@@ -456,7 +456,9 @@ def add_extra_args(args)
     until extra.empty? do
       ex = []
       ex << extra.shift
-      ex << extra.shift if extra.first.to_s =~ /^[^-]/
+      if (!extra.first.to_s.empty? && !extra.first.to_s.start_with?("-"))
```
deivid-rodriguez, line 459: Do you feel this code is better than the previous code? I think the previous code was easier to grasp, so perhaps we should ignore the cop here? Or at least invert the condition so it reads a bit simpler? (`!A && !B = A || B`).

lib/rubygems/platform.rb
```diff
@@ -150,7 +150,7 @@ def ===(other)
 
     # cpu
     ([nil,'universal'].include?(@cpu) or [nil, 'universal'].include?(other.cpu) or @cpu == other.cpu or
-    (@cpu == 'arm' and other.cpu =~ /\Aarm/)) and
+     (@cpu == 'arm' and other.cpu.start_with?("arm"))) and
```
deivid-rodriguez, line 153: Is the indentation change intentional?

lib/rubygems/specification.rb
```diff
@@ -2102,7 +2102,7 @@ def mark_version
 
   def method_missing(sym, *a, &b) # :nodoc:
     if @specification_version > CURRENT_SPECIFICATION_VERSION and
-      sym.to_s =~ /=$/
+        sym.to_s.end_with?("=")
```
deivid-rodriguez, line 2105: Same here about indentation? If the second indendation is better, can we fix it separately by applying an indentation cop :)?

lib/rubygems/platform.rb
olleolleolle: Unintended, fixing.

lib/rubygems/command.rb
olleolleolle: (`nil` and `""` must not enter the conditional.) I reverted, added a disable.

lib/rubygems/specification.rb
```diff
@@ -1153,7 +1153,7 @@ def self.non_nil_attributes
 
   def self.normalize_yaml_input(input)
     result = input.respond_to?(:read) ? input.read : input
-    result = "--- " + result unless result =~ /\A--- /
+    result = "--- " + result unless result.start_with?("--- ")
```
simi, line 1156: This could potentially explode on `nil` input as well (called from `from_yaml`, which will explode later anyway on `nil` input).

simi:
After looking into this a little more deeper, I'm more "sceptic" in here. Is there any real performance benefit for users in here we can trade for this "dangerous" default? If not, I think it would be better as is for now.

lib/rubygems/specification.rb
deivid-rodriguez: I checked this one and it can't be `nil` since `gsub!` is being called on it two lines later.

deivid-rodriguez: I mean, we're not introducing any issue that's not already there, and since we haven't had any crash reports here, I think we're fine.

olleolleolle:
One of the readability things about start_with: The reader can know that the rest of the code isn't attempting to use `$'` or any other side effect.

deivid-rodriguez:
What should we do here? Should we go ahead with this? I definitely agree with this being more readable.

deivid-rodriguez:
I wasn't totally sold since I also get @simi's point of staying on the safe side, but let's do it!

PR #3533, opened 2020-04-20
Normalize heredoc delimiters

deivid-rodriguez:
# Description:

On another PR, I noticed that while normally we use `RUBY` as the delimiter for heredocs including ruby code, sometimes we use `RB`.

`RUBY` is better so I'm replacing `RB` occurrences with `RUBY`.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -401,11 +401,11 @@ module Gem
           target[k] = v
         when Array
           unless Gem::Deprecate.skip
-            warn <<-eowarn
+            warn <<-EOWARN
 Array values in the parameter to `Gem.paths=` are deprecated.
 Please use a String or nil.
 An Array (#{env.inspect}) was passed in from #{caller[3]}
-            eowarn
+            EOWARN
           end
           target[k] = v.join File::PATH_SEPARATOR
         end
--- a/test/rubygems/test_gem_commands_setup_command.rb
+++ b/test/rubygems/test_gem_commands_setup_command.rb
@@ -359,7 +359,7 @@ class TestGemCommandsSetupCommand < Gem::TestCase
     @cmd.options[:previous_version] = Gem::Version.new '2.0.2'
 
     File.open 'History.txt', 'w' do |io|
-      io.puts <<-History_txt
+      io.puts <<-HISTORY_TXT
 # coding: UTF-8
 
 === #{Gem::VERSION} / 2013-03-26
@@ -377,7 +377,7 @@ class TestGemCommandsSetupCommand < Gem::TestCase
 
 * Bug fixes:
   * Yet more bugs fixed
-      History_txt
+      HISTORY_TXT
     end
 
     use_ui @ui do
--- a/test/rubygems/test_gem_ext_cmake_builder.rb
+++ b/test/rubygems/test_gem_ext_cmake_builder.rb
@@ -23,11 +23,11 @@ class TestGemExtCmakeBuilder < Gem::TestCase
 
   def test_self_build
     File.open File.join(@ext, 'CMakeLists.txt'), 'w' do |cmakelists|
-      cmakelists.write <<-eo_cmake
+      cmakelists.write <<-EO_CMAKE
 cmake_minimum_required(VERSION 2.6)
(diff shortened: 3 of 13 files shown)


simi:
:thinking: I see there are various `HEREDOC` names used across the code. Some are lower case, some upper case, some have meaningful name like `RUBY`, some have less meaningful name like `D` or `EOF`.

There are two related cops we can consider to use as well if you're doing `HEREDOC` cleanup.

https://rubocop.readthedocs.io/en/latest/cops_naming/#namingheredocdelimitercase
https://rubocop.readthedocs.io/en/latest/cops_naming/#namingheredocdelimiternaming

deivid-rodriguez:
Good point @simi. I enabled `Naming/HeredocLimiterNaming` to check for the specific issue I wanted to normalize, and also enabled the delimiter case cop.

deivid-rodriguez:
@simi You ok with this?

lib/rubygems.rb
```diff
@@ -401,11 +401,11 @@ def self.paths=(env)
           target[k] = v
         when Array
           unless Gem::Deprecate.skip
-            warn <<-eowarn
+            warn <<-EOWARN
```
simi, line 404: Can we unify this to use `WARNING` as well?

test/rubygems/test_gem_ext_cmake_builder.rb
```diff
@@ -23,11 +23,11 @@ def setup
 
   def test_self_build
     File.open File.join(@ext, 'CMakeLists.txt'), 'w' do |cmakelists|
-      cmakelists.write <<-eo_cmake
+      cmakelists.write <<-EO_CMAKE
```
simi, line 26: Would `CMAKE` be enough here?

test/rubygems/test_gem_remote_fetcher.rb
```diff
@@ -81,7 +81,7 @@ class TestGemRemoteFetcher < Gem::TestCase
   # Generated via:
   #   x = OpenSSL::PKey::DH.new(2048) # wait a while...
   #   x.to_s => pem
-  TEST_KEY_DH2048 = OpenSSL::PKey::DH.new <<-_end_of_pem_
+  TEST_KEY_DH2048 = OpenSSL::PKey::DH.new <<-_END_OF_PEM_
```
simi, line 84: `PEM`?

test/rubygems/test_gem_request_set.rb
```diff
@@ -396,9 +396,9 @@ def test_resolve_git
     rs = Gem::RequestSet.new
 
     tf = Tempfile.open 'gem.deps.rb' do |io|
-      io.puts <<-gems_deps_rb
+      io.puts <<-GEMS_DEPS_RB
```
simi, line 399: If I understand it well, this is also just code ruby block. Any reason to not use `RUBY` as well?

test/rubygems/test_gem_commands_setup_command.rb
```diff
@@ -359,7 +359,7 @@ def test_show_release_notes
     @cmd.options[:previous_version] = Gem::Version.new '2.0.2'
 
     File.open 'History.txt', 'w' do |io|
-      io.puts <<-History_txt
+      io.puts <<-HISTORY_TXT
```
simi, line 362: Would `TEXT` be enough here?

simi: Also can't we use 

```ruby
File.write('History.txt', <<-TEXT)
  ...
TEXT

here?
```

simi:
@deivid-rodriguez I left some minor non-blocking comments from my side.

deivid-rodriguez:
@simi I only made the necessary changes to get the `Naming/HeredocDelimiterCase` cop green since you mentioned it, but I didn't intend to make a whole review of heredoc naming. There's for example a lot of heredocs named `G`, which could be renamed to `GEMFILE` or `RUBY` as well. So much stuff that can be improved in the code base...

Let me get this in and iterate on your suggestions next time I work on a style PR, ok? :)

PR #3534, opened 2020-04-20
Load rubygems plugins from RUBYLIB during `bundle install` and `bundle update`

djmarcin:
# Description:
Restore load_env_plugins so that plugins stored in RUBYLIB are loaded by bundler.

# What was the end-user or developer problem that led to this PR?

https://github.com/rubygems/bundler/pull/4954 broke plugins that are loaded by setting RUBYLIB. The comments in the PR make clear that the intention was to augment the set of plugins being loaded, not remove ENV plugins and replace with only plugins in the standard location. This hook does work when loaded by `gem install` because of https://github.com/rubygems/rubygems/blob/0ab446f46556e83a36d9e026acae26468bc08b91/lib/rubygems/gem_runner.rb#L15.

This explains why when rbenv attempts to hook into bundler like in the following link, it doesn't work -- this code is never actually executed during `bundle install`. https://github.com/rbenv/rbenv/blob/0843745be9267de0296de8725c5b7bfcbbb6bddf/rbenv.d/exec/gem-rehash/rubygems_plugin.rb#L14-L35

This PR restores the load from environment variables that was erroneously removed in https://github.com/rubygems/bundler/pull/4954

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/installer.rb
+++ b/bundler/lib/bundler/installer.rb
@@ -244,6 +244,7 @@ module Bundler
         end
       end.flatten
       Bundler.rubygems.load_plugin_files(path_plugin_files)
+      Bundler.rubygems.load_env_plugins
     end
 
     def ensure_specs_are_compatible!
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -227,6 +227,10 @@ module Bundler
       Gem.load_plugin_files(files) if Gem.respond_to?(:load_plugin_files)
     end
 
+    def load_env_plugins
+      Gem.load_env_plugins if Gem.respond_to?(:load_env_plugins)
+    end
+
     def ui=(obj)
       Gem::DefaultUserInteraction.ui = obj
     end
--- a/bundler/spec/commands/install_spec.rb
+++ b/bundler/spec/commands/install_spec.rb
@@ -222,6 +222,18 @@ RSpec.describe "bundle install with gem sources" do
       expect(the_bundle).to include_gems "rack 1.0.0", "activesupport 2.3.5"
     end
 
+    it "loads env plugins" do
+      plugin_msg = "hello from an env plugin!"
+      create_file "plugins/rubygems_plugin.rb", "puts '#{plugin_msg}'"
+      rubylib = ENV["RUBYLIB"].to_s.split(File::PATH_SEPARATOR).unshift(bundled_app("plugins").to_s).join(File::PATH_SEPARATOR)
+      install_gemfile <<-G, :env => { "RUBYLIB" => rubylib }
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack"
+      G
+
+      expect(last_command.stdboth).to include(plugin_msg)
+    end
+
     describe "with a gem that installs multiple platforms" do
       it "installs gems for the local platform as first choice" do
         skip "version is 1.0, not 1.0.0" if Gem.win_platform?


deivid-rodriguez:
I'm so sorry for the delay getting to this. Wow, this has been broken for such a long time! Can you add a test to ensure we don't regress again in the future?

djmarcin:
I took a stab today, but I’m not familiar with the tests so I was
unsuccessful. I’ll work more on it soon, but I’m prepping for a move so
might be a few days.

deivid-rodriguez:
Thanks, no rush!

Another option would be to write a high level test in a workflow. Like install `rbenv` and `ruby-build`, install `ruby`, install a `Gemfile`, make sure rehashing or whatever needs to happen in the `rbenv` plugin actually happens.

djmarcin:
@deivid-rodriguez Thanks for the pointer to workflows, but I think I managed to get a good lightweight test using the install command suite. It creates a simple plugin that emits a message to stdout and verifies that the bundle install command runs the plugin and the message appears in the output. I verified that the test does not pass before this PR and passes after.

The only test that did not pass was `ruby-core / ruby_core (bundler) (pull_request)` workflow and I don't really know why (https://github.com/rubygems/rubygems/runs/1069643478?check_suite_focus=true). I expect it has something to do with this workflow having a unique test environment as I see several tests skipped due to being in ruby-core, but I'm not sure that's the right solution here. Do you have any insight?

djmarcin:
I figured it out, I needed to be more careful setting the `RUBYLIB` var for the test and handle the platform specific separator.

djmarcin:
Hi @deivid-rodriguez, hoping you've had some time to think about this and get this landed. It would be nice to get automatic reshimming working again for tools like rbenv and asdf. Hopefully any larger refactor of plugins would not need to block this relatively targeted fix.

deivid-rodriguez:
Hei @djmarcin! Thanks a lot for your patience here. I'm putting this PR on my review queue and plan to get to this in the next couple of weeks. The fix itself looks fine from what I remember, so it will land, don't worry :)

Other than reviewing the fix itself, I also wanted to try out `rbenv` and `asdf` to verify what end user behaviour there is being fixed. Could you explain what's actually being fixed in both version managers? I use `rbenv` on a daily basis and it's been literally years without running into any issues, so I wonder whether their rubygems plugin is actually necessary at all.

djmarcin:
Thanks @deivid-rodriguez, appreciate your time.

> Other than reviewing the fix itself, I also wanted to try out `rbenv` and `asdf` to verify what end user behaviour there is being fixed. Could you explain what's actually being fixed in both version managers? I use `rbenv` on a daily basis and it's been literally years without running into any issues, so I wonder whether their rubygems plugin is actually necessary at all.

Both version managers use similar ["shim"](https://github.com/rbenv/rbenv#understanding-shims) strategies for making installed binaries available on your PATH. They both attempt to use the plugin mechanism to "reshim" (asdf) or "rehash" (rbenv) to pick up new binaries installed by new gems. When you use `gem install` this works via the expected plugin mechanism, but when you use `bundle install` it does not (due to the issue fixed by this PR). One reason that you may not see issues is that once a shim exists it does not need to be recreated, even when switching to new versions of ruby or new projects. As long as you don't `rm` your shims directory, the issue may remain hidden after the first resolution for a given gem.

rbenv works around this by always running a rehash during `rbenv init` which it encourages you to put in your shell startup script, so each time you open a new shell it fixes any missing shims at the cost of making your new shells open more slowly. This workaround narrows the window of the issue to between `bundle install` and opening a new shell (which fixes the problem in all shells). 

asdf does not have a corresponding workaround, meaning that the issue persists until users run a reshim operation manually.

With this PR, the need for a manual rehash/reshim goes away in both version managers as new gems always trigger the plugin mechanism, whether they come via `gem install` or `bundle install`.

deivid-rodriguez (approved):
@djmarcin I'm really sorry for the delay. I had a look at this and looks great. Thanks for the great explanation of how this improves things for version managers too!

deivid-rodriguez:
Could you rebase on top of the latest default branch so that CI runs one more time and I'll merge after it's green :+1:.

deivid-rodriguez:
No worries, I rebased it for you, and also added a super tiny tweak to the spec just to make the style a bit more consistent with other specs manipulating RUBYLIB.

deivid-rodriguez:
Thanks @djmarcin!! :tada:

PR #3536, opened 2020-04-21
Revert deprecation of deprecate method for now.

simi:
follow-up of #3530, https://github.com/rubygems/rubygems/pull/3530#discussion_r411419721

### DIFF
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -63,7 +63,6 @@ module Gem::Deprecate
                 ". It will be removed on or after %4d-%02d-01." % [year, month],
                 "\n#{target}#{name} called from #{Gem.location_of_caller.join(":")}",
         ]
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I'll let @bronzdoc merge this to make sure that he's inline with our plan :+1:.

bronzdoc:
What's the plan? 

@deivid-rodriguez can you elaborate on https://github.com/rubygems/rubygems/pull/3530#discussion_r412015428?

deivid-rodriguez:
The plan is to leave things unchanged our users for now until we have a better alternative. The comment you linked to is just a rough idea that needs to be properly explored.

bronzdoc:
I don't see an easier way than this right now, but ok let's think about it a little more before affecting the end users...

I'll try to come up with something.

deivid-rodriguez:
I opened #3539 to follow up.

Issue #3454, opened 2020-03-28
bundle gem always uses RSpec, even if configured for minitest


mperham:
Hi, I'm creating a new gem with Bundler 2.1.4:

```
# From bundle-gem man page:
# Note that these options can also be specified via Bundler's global configuration file using the
# following names:
#       o   gem.test

$ bundle config gem.test minitest
$ bundle config gem.test
Settings for `gem.test` in order of priority. The top value will be used
Set for the current user (/Users/mikeperham/.bundle/config): "minitest"
$ bundle gem somegem -t
Creating gem 'somegem'...
      [elided]
      create  somegem/.rspec
      create  somegem/spec/spec_helper.rb
      create  somegem/spec/somegem_spec.rb
```

Bundler creates an RSpec test suite for me, rather than using minitest as configured.

deivid-rodriguez:
Interesting. I can reproduce.

It happens when, even if you have configured `gem.test`, you specify the `-t` option. So the bug is that the `-t` option without a value always defaults to `RSpec` instead of respecting the `gem.test` setting.

The workaround is better than the problematic command, though. SInce you have already configured your preference, don't pass the `-t` option and use `bundle gem somegem` instead. :smile:

deivid-rodriguez:
Actually, after reading `bundle gem --help`

>     -t, --test=minitest, --test=rspec
>          Specify the test framework that Bundler should use when generating the project. Acceptable values are minitest and rspec. The GEM_NAME.gemspec will be configured and a skeleton test/spec directory will be created based
>          on this option. If this option is unspecified, an interactive prompt will be displayed and the answer will be saved in Bundler´s global config for future bundle gem use. If no option is specified, the  default  testing
>          framework is RSpec.

I think the current behavior is exactly what's documented :thinking:.

Except maybe "If no option is specified, the default testing framework is RSpec" could be clarified to "If no value is given for the option, the default testing framework is RSpec".

mperham:
From my POV, it makes no sense for -t to ignore the gem.test config. That’s literally why I didn’t give a value, because I had globally config’d it.

deivid-rodriguez:
I mean, I don't necessarily disagree with you, I'm just saying that the current behavior is expected and documented. But we can change it.

The way it currently works is:
 
| command line | result |
| ---| --- |
| No `-t` option given | Use whatever is configured or prompt and configure if nothing configured |
| `-t` option given without a value | Use RSpec |
[...]

PR #3394, opened 2020-03-16
Add notes describing why groff 1.22.3 must be used

kevinburkemeter:
Updates #3387.
Updates #3391.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3544, opened 2020-04-24
Fix bundle gem always using RSpec, even if configured for another test framework

FTLam11:
# Description:

The problem is described in detail [here](https://github.com/rubygems/rubygems/issues/3454).

* Default to `gem.test` in global config instead of RSpec when passed empty `-t` flag
* Additionally ask the user to choose the test framework when empty `-t` flag is passed and `gem.test` is set to `false` (The user has previously configured no test framework, but wants to use a test framework for a new gem)

Fixes https://github.com/rubygems/rubygems/issues/3454.
______________

# Tasks:

- [X] Write tests

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -571,7 +571,7 @@ module Bundler
     method_option :git, :type => :boolean, :default => true, :desc => "Initialize a git repo inside your library."
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config set gem.mit true`."
     method_option :rubocop, :type => :boolean, :desc => "Add rubocop to the generated Rakefile and gemspec. Set a default with `bundle config set gem.rubocop true`."
-    method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
+    method_option :test, :type => :string, :lazy_default => Bundler.settings["gem.test"] || "", :aliases => "-t", :banner => "Use the specified test framework for your library",
                          :desc => "Generate a test directory for your library, either rspec, minitest or test-unit. Set a default with `bundle config set gem.test rspec`."
     def gem(name)
     end
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -177,6 +177,12 @@ module Bundler
 
       Bundler.ui.info "Gem '#{name}' was successfully created. " \
         "For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html"
+
+      if options[:test] == Bundler.settings["gem.test"]
+        Bundler.ui.info "Bundler is configured to generate test files for #{Bundler.settings["gem.test"]}, "\
+                        "so -t is not needed if you want to continue using it. " \
+                        "This setting can be changed anytime with bundle config gem.test."
+      end
     rescue Errno::EEXIST => e
       raise GenericSystemCallError.new(e, "There was a conflict while creating the new gem.")
     end
@@ -216,9 +222,7 @@ module Bundler
       if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
         result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. \n" \
-                                "If Bundler is configured to not generate test files, your choice will only be applied to this instance. \n" \
-                                "Otherwise, future bundle gem calls will use your choice, so -t is not needed if your choice will be the same. \n" \
-                                "This setting can be changed anytime with bundle config gem.test <value>. rspec/minitest/test-unit/(none):"
+                                "#{test_framework_help_info} rspec/minitest/test-unit/(none):"
         if result =~ /rspec|minitest|test-unit/
           test_framework = result
         else
@@ -231,6 +235,15 @@ module Bundler
       end
 
       test_framework
+    end
+
+    def test_framework_help_info
+      if Bundler.settings["gem.test"] == false
+        "Your choice will only be applied to this gem."
+      else
+        "Future bundle gem calls will use your choice. " \
+        "This setting can be changed anytime with bundle config gem.test."
+      end
     end
 
     def bundler_dependency_version
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -221,8 +221,9 @@ module Bundler
 
       if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. \n" \
-                                "#{test_framework_hint} rspec/minitest/test-unit/(none):"
+        Bundler.ui.info test_framework_hint
+        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. " \
+                                "rspec/minitest/test-unit/(none):"
         if result =~ /rspec|minitest|test-unit/
           test_framework = result
         else
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -213,10 +213,12 @@ module Bundler
     def ask_and_set_test_framework
       test_framework = options[:test] || Bundler.settings["gem.test"]
 
-      if test_framework.nil?
+      if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now and " \
-          "in the future. rspec/minitest/test-unit/(none):"
+        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. \n" \
+                                "If Bundler is configured to not generate test files, your choice will only be applied to this instance. \n" \
+                                "Otherwise, future bundle gem calls will use your choice, so -t is not needed if your choice will be the same. \n" \
+                                "This setting can be changed anytime with bundle config gem.test <value>. rspec/minitest/test-unit/(none):"
         if result =~ /rspec|minitest|test-unit/
           test_framework = result
         else
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -181,7 +181,7 @@ module Bundler
       if options[:test] == Bundler.settings["gem.test"]
         Bundler.ui.info "Bundler is configured to generate test files for #{Bundler.settings["gem.test"]}, "\
                         "so -t is not needed if you want to continue using it. " \
-                        "This setting can be changed anytime with bundle config gem.test."
+                        "This setting can be changed anytime with `bundle config gem.test`."
       end
     rescue Errno::EEXIST => e
       raise GenericSystemCallError.new(e, "There was a conflict while creating the new gem.")
@@ -222,7 +222,7 @@ module Bundler
       if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
         result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. \n" \
-                                "#{test_framework_help_info} rspec/minitest/test-unit/(none):"
+                                "#{test_framework_hint} rspec/minitest/test-unit/(none):"
         if result =~ /rspec|minitest|test-unit/
           test_framework = result
         else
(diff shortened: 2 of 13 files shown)


bundler/lib/bundler/cli.rb
```diff
@@ -570,7 +570,7 @@ def viz
     method_option :git, :type => :boolean, :default => true, :desc => "Initialize a git repo inside your library."
     method_option :mit, :type => :boolean, :desc => "Generate an MIT license file. Set a default with `bundle config set gem.mit true`."
     method_option :rubocop, :type => :boolean, :desc => "Add rubocop to the generated Rakefile and gemspec. Set a default with `bundle config set gem.rubocop true`."
-    method_option :test, :type => :string, :lazy_default => "rspec", :aliases => "-t", :banner => "rspec",
```
bronzdoc, line 573: I think we still want to default to Rspec. But if it's a test framework configured use that instead

FTLam11: Defaulting to RSpec makes passing empty -t ambiguous, this is the main problem encountered in the original issue. A test framework was configured, but ignored with `lazy_default` defaulting to `rspec`.

If `lazy_default` is set to any test framework, then which of `options[:test]` and `Bundler.settings["gem.test"]` should take precedence?

bronzdoc: `Bundler.settings["gem.test"]` if it is true else fallback to the default `options[:test]`

With this implementation if you have configured something in `Bundler.settings["gem.test"]` and you pass `-t` when creating a gem no tests are generated.

FTLam11: Thanks for catching that error with the current implementation!

bundler/lib/bundler/cli/gem.rb
```diff
@@ -79,13 +79,12 @@ def run
       ]
 
       templates.merge!("gitignore.tt" => ".gitignore") if Bundler.git_present?
+      templates.merge!("travis.yml.tt" => ".travis.yml")
 
       if test_framework = ask_and_set_test_framework
         config[:test] = test_framework
         config[:test_framework_version] = TEST_FRAMEWORK_VERSIONS[test_framework]
 
-        templates.merge!("travis.yml.tt" => ".travis.yml")
```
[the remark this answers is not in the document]
FTLam11: Cool, do I need to also open a corresponding issue? I think I want to address the above change first before finishing up this PR.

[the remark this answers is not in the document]
bronzdoc: A PR explaining you change will be fine :)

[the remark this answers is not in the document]
FTLam11: I ended up dropping this change after seeing [some discussion about it in a separate PR](https://github.com/rubygems/rubygems/pull/3584).

bundler/lib/bundler/cli.rb
FTLam11: I incorporated your suggestion of lazy defaulting to `Bundler.settings["gem.test"]`. In addition, if this value is falsy then fallback to `"`. `"` is used to check if we want to reconfirm setting the test framework.

FTLam11:
@bronzdoc Would you do another review? Thanks!

bronzdoc:
@FTLam11 Will give this a look tomorrow. sorry for the delay!

bundler/man/bundle-gem.1
```diff
@@ -65,7 +65,7 @@ Do not create a \fBLICENSE\.txt\fR (overrides \fB\-\-mit\fR specified in the glo
 .
 .TP
 \fB\-t\fR, \fB\-\-test=minitest\fR, \fB\-\-test=rspec\fR, \fB\-\-test=test\-unit\fR
-Specify the test framework that Bundler should use when generating the project\. Acceptable values are \fBminitest\fR, \fBrspec\fR and \fBtest\-unit\fR\. The \fBGEM_NAME\.gemspec\fR will be configured and a skeleton test/spec directory will be created based on this option\. If this option is unspecified, an interactive prompt will be displayed and the answer will be saved in Bundler\'s global config for future \fBbundle gem\fR use\. If no option is specified, the default testing framework is RSpec\.
+Specify the test framework that Bundler should use when generating the project\.  Acceptable values are \fBminitest\fR, \fBrspec\fR and \fBtest\-unit\fR\. The \fBGEM_NAME\.gemspec\fR will be configured and a skeleton test/spec directory will be created based on this option\. If this option is unspecified, an interactive prompt will be displayed and the answer will be saved in Bundler\'s global config for future \fBbundle gem\fR use\. If no option is specified, this defaults to the global config setting\.
 .
```
bronzdoc, line 69: Maybe be more clear about the complete functionality here, what you think of something like:

`this defaults to the global config setting or in case that nothing is configured it will prompt for a test framework and configure it`

bronzdoc:
Just as a reference. This is what this PR is doing

| command line | result |
| ---| --- |
| No `-t` option given | Use whatever is configured or prompt and configure if nothing configured |
| `-t` option given without a value | Use whatever is configured or prompt and configure if nothing configured |
| `-t` option given with a value | Use the value |

deivid-rodriguez:
Just to make sure, what is the PR doing when `-t` is given and `false` is configured? It sounds pretty tricky and I'm not sure what the best behaviour is. Maybe ask for a test framework but not persist the choice in configuration.

bundler/lib/bundler/cli.rb
```diff
@@ -570,7 +570,7 @@ def viz
...
+    method_option :test, :type => :string, :lazy_default => Bundler.settings["GEM.TEST"] || "", :aliases => "-t", :banner => "Use the specified test framework for your library",
```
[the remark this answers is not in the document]
FTLam11: Noted, will do. Thanks!

FTLam11:
> Just to make sure, what is the PR doing when `-t` is given and `false` is configured? It sounds pretty tricky and I'm not sure what the best behaviour is. Maybe ask for a test framework but not persist the choice in configuration.

With the current PR and the above scenario, the user will be asked to choose a test framework and the choice will be saved in their configuration.

It's clear what we should do if the user only wants tests for one gem (ask and do not save the choice) or tests for all future gems (ask and save the choice). What's not so clear is how we infer the user's intent from the flag and configuration combination. I unfortunately don't have a better suggestion right now.

bundler/man/bundle-gem.1
FTLam11: This might slightly change depending on: https://github.com/rubygems/rubygems/pull/3544#issuecomment-630700501.

deivid-rodriguez:
Right, it's tricky.

My proposal would be to make saving configuration a one-off thing that's only done if there's no prior configuration, but never overwrite it if already present.

I think if we properly inform the user when saving the configuration the first time that:

* The setting will be applied by default on future `bundle gem` calls, so no `-t` will be needed anymore if the choice will be the same.

* The setting can be changed at any time with `bundle config gem.test <value>`.

Then I think it would be fine and make everyone happy?

FTLam11:
@deivid-rodriguez Sorry, I keep associating asking/setting the test framework as a single action. This PR does not change the behavior of setting `gem.test`, so `gem.test` is only set once and never overwritten when present.

I'll look into incorporating the help text suggestions. Thanks!

deivid-rodriguez:
Nice, I just tried this PR and it does what we believe is the most sensible thing :+1:.

bundler/man/bundle-gem.1
FTLam11: I updated the man pages, tried to keep similar language with the rest of the files

deivid-rodriguez:
@FTLam11 This works in a nice user friendly way now, but since now sometimes `bundler` will ask you to enter your test framework choice, but _will not_ save that choice and only use it for the current invokation, I believe we need to tweak the message depending on this behaviour.

deivid-rodriguez:
@FTLam11 Now I understand you slack question. The problem is that your PR is out of date with master. The month is already updated in master, so you would need to rebase the PR.

FTLam11:
@deivid-rodriguez The test framework prompt has been updated to hint how the user's selection will be applied. The man pages have also been updated with a more verbose explanation.

deivid-rodriguez:
Nice, it's looking good. It doesn't seem fully rebased though, since the changes in the month are still showing up. Can you fully rebase it so that CI runs against the latest master branch? If your upstream remote is named "upstream", `git pull --rebase upstream master` from your "FTLam11-#3454-bundle-gem-test-framework-default-setting" branch and then force push should do the trick.

FTLam11:
@deivid-rodriguez Got it, thanks for the git help and detailed explanation! I see my first rebase was just cleaning up my own branch history, I didn't actually rebase onto master.

deivid-rodriguez:
No problem! I'm sorry getting this PR ready is being harder than usual :sweat_smile: 

Now I see some whitespace changes to man pages where `.ronn` sources have not been changed? I thought it would be because the repo HEAD you used to regenerate man source didn't include https://github.com/rubygems/rubygems/pull/3391, but I'm afraid that PR is older than your previous local HEAD, so I'm unsure what's going on. Can you undo all changes to `txt` files (`git show bundler/man/*.txt|git apply -R`) and then regenerate man pages again `bin/rake man:build` to see if the differences creep up again?

FTLam11:
Turns out when I first rebuilt the man pages, my groff was v1.19.2 which is the built-in macOS default. I tried installing v1.22.4 via homebrew, but I got even weirder output from groff. I looked through #3394, and was eventually able compile groff from source (v1.22.3). Hopefully the whitespace issue will get resolved now.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -213,10 +213,12 @@ def validate_ext_name
...
 
-      if test_framework.nil?
+      if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now and " \
-          "in the future. rspec/minitest/test-unit/(none):"
+        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. \n" \
+                                "If Bundler is configured to not generate test files, your choice will only be applied to this instance. \n" \
+                                "Otherwise, future bundle gem calls will use your choice, so -t is not needed if your choice will be the same. \n" \
+                                "This setting can be changed anytime with bundle config gem.test <value>. rspec/minitest/test-unit/(none):"
```
deivid-rodriguez, line 221: I know it's tricky and I might just being annoying, but we have the knowledge about this, so I think we can skip all the extra output when not needed.

So I would check:

* If bundler is configured to not generate test files, only add "Your choice will only be applied to this instance" to the output.

* Otherwise, check whether the `-t` flag has been passed.

  * If `-t` has been passed, add the output "Future bundle gem calls will use your choice, so -t is not needed if your choice will be the same. This setting can be changed anytime with bundle config gem.test <value>."

  * Otherwise, add the output "Future bundle gem calls will use your choice. This setting can be changed anytime with bundle config gem.test <value>.".

This way I think we can educate users about the stuff that's relevant to them, while keeping the output minimal and not too overwhelming.

What do you think?

bundler/man/bundle-gem.ronn
```diff
@@ -64,10 +64,17 @@ configuration file using the following names:
...
+
+  When Bundler is configured to generate tests, this defaults to Bundler's
+  global config setting `gem.test`.
+
+  When Bundler is configured to not generate tests, an interactive prompt will
+  be displayed and the answer will be used for the current rubygem project.
+
+  When Bundler is unconfigured, an interactive prompt will be displayed and
+  the answer will be saved in Bundler's global config for future `bundle gem`
+  use.
```
deivid-rodriguez, line 77: New docs look awesome! :star_struck:

bundler/lib/bundler/cli/gem.rb
FTLam11: You bring up great points about user facing hints/output. I tweaked the display messages a bit. When `-t` is passed and isn't necessary, I ordered the message last after the gem is created to make it easier to notice. I was debating using `Bundler.ui.warn` but for now am sticking with `Bundler.ui.info`.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -231,6 +237,15 @@ def ask_and_set_test_framework
       test_framework
     end
 
+    def test_framework_help_info
+      if Bundler.settings["gem.test"] == false
+        "Your choice will only be applied to this gem."
+      else
+        "Future bundle gem calls will use your choice. " \
```
deivid-rodriguez, line 244:
```suggestion
        "Future `bundle gem` calls will use your choice. " \
```

bundler/lib/bundler/cli/gem.rb
```diff
@@ -231,6 +237,15 @@ def ask_and_set_test_framework
...
+        "This setting can be changed anytime with bundle config gem.test."
```
deivid-rodriguez, line 245:
```suggestion
        "This setting can be changed anytime with `bundle config gem.test <value>`."
```

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Is it possible to add output assertions for each of the branches to some existing tests?

FTLam11: Yep, added specs for both branches above and for when `-t` isn't needed.

FTLam11:
@deivid-rodriguez @bronzdoc Thank you both for guidance and patience! 🎉

PR #3545, opened 2020-04-24
Support XDG specification

hsbt:
# Description:

RubyGems still uses the hard-coded file location like `~/.gem/credentials`. The users couldn't change them.

I added to support XDG specification.

The original commits are   https://github.com/ruby/ruby/pull/2174
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/spec/bundler/env_spec.rb
+++ b/bundler/spec/bundler/env_spec.rb
@@ -45,9 +45,15 @@ RSpec.describe Bundler::Env do
       it "prints user path" do
         skip "needs to use a valid HOME" if Gem.win_platform? && RUBY_VERSION < "2.6.0"
 
-        with_clear_paths("HOME", "/a/b/c") do
+        if Gem::VERSION >= "3.2.0.pre.1"
+          allow(Gem).to receive(:data_home) { "/a/b/c/.local/share" }
           out = described_class.report
-          expect(out).to include("User Path   /a/b/c/.gem")
+          expect(out).to include("User Path   /a/b/c/.local/share/gem")
+        else
+          with_clear_paths("HOME", "/a/b/c") do
+            out = described_class.report
+            expect(out).to include("User Path   /a/b/c/.gem")
+          end
         end
       end
 
--- a/bundler/spec/bundler/env_spec.rb
+++ b/bundler/spec/bundler/env_spec.rb
@@ -45,15 +45,11 @@ RSpec.describe Bundler::Env do
       it "prints user path" do
         skip "needs to use a valid HOME" if Gem.win_platform? && RUBY_VERSION < "2.6.0"
 
-        if Gem::VERSION >= "3.2.0.pre.1"
-          allow(Gem).to receive(:data_home) { "/a/b/c/.local/share" }
+        with_clear_paths("HOME", "/a/b/c") do
+          allow(File).to receive(:exist?)
+          allow(File).to receive(:exist?).with("/a/b/c/.gem").and_return(true)
           out = described_class.report
-          expect(out).to include("User Path   /a/b/c/.local/share/gem")
-        else
-          with_clear_paths("HOME", "/a/b/c") do
-            out = described_class.report
-            expect(out).to include("User Path   /a/b/c/.gem")
-          end
+          expect(out).to include("User Path   /a/b/c/.gem")
         end
       end
 
--- a/bundler/spec/bundler/friendly_errors_spec.rb
+++ b/bundler/spec/bundler/friendly_errors_spec.rb
@@ -6,8 +6,11 @@ require "cgi"
 
 RSpec.describe Bundler, "friendly errors" do
   context "with invalid YAML in .gemrc" do
+    let(:config_home) { File.dirname(Gem.configuration.config_file_name) }
+
     before do
-      File.open(home(".gemrc"), "w") do |f|
+      FileUtils.mkdir_p config_home
+      File.open(Gem.configuration.config_file_name, "w") do |f|
         f.write "invalid: yaml: hah"
       end
     end
@@ -24,7 +27,11 @@ RSpec.describe Bundler, "friendly errors" do
 
       bundle :install, :env => { "DEBUG" => "true" }
 
-      expect(err).to include("Failed to load #{home(".gemrc")}")
+      if Gem::VERSION >= "3.2.0.pre.1"
+        expect(err).to include("Failed to load #{File.join(config_home, "gemrc")}")
+      else
+        expect(err).to include("Failed to load #{home(".gemrc")}")
+      end
       expect(exitstatus).to eq(0) if exitstatus
     end
   end
--- a/bundler/spec/bundler/friendly_errors_spec.rb
+++ b/bundler/spec/bundler/friendly_errors_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Bundler, "friendly errors" do
     end
 
     after do
-      FileUtils.rm(home(".gemrc"))
+      FileUtils.rm(Gem.configuration.config_file_name)
     end
 
     it "reports a relevant friendly error message" do
--- a/bundler/spec/bundler/friendly_errors_spec.rb
+++ b/bundler/spec/bundler/friendly_errors_spec.rb
@@ -6,33 +6,59 @@ require "cgi"
 
 RSpec.describe Bundler, "friendly errors" do
   context "with invalid YAML in .gemrc" do
-    let(:config_home) { File.dirname(Gem.configuration.config_file_name) }
-
-    before do
-      FileUtils.mkdir_p config_home
-      File.open(Gem.configuration.config_file_name, "w") do |f|
-        f.write "invalid: yaml: hah"
+    context "with the old ~/.gemrc" do
+      before do
+        File.open(home(".gemrc"), "w") do |f|
+          f.write "invalid: yaml: hah"
+        end
       end
-    end
 
-    after do
-      FileUtils.rm(Gem.configuration.config_file_name)
-    end
+      after do
+        FileUtils.rm(home(".gemrc"))
+      end
 
-    it "reports a relevant friendly error message" do
-      gemfile <<-G
-        source "#{file_uri_for(gem_repo1)}"
-        gem "rack"
-      G
+      it "reports a relevant friendly error message" do
+        gemfile <<-G
+          source "#{file_uri_for(gem_repo1)}"
+          gem "rack"
+        G
 
-      bundle :install, :env => { "DEBUG" => "true" }
+        bundle :install, :env => { "DEBUG" => "true" }
 
-      if Gem::VERSION >= "3.2.0.pre.1"
-        expect(err).to include("Failed to load #{File.join(config_home, "gemrc")}")
-      else
         expect(err).to include("Failed to load #{home(".gemrc")}")
+        expect(exitstatus).to eq(0) if exitstatus
       end
-      expect(exitstatus).to eq(0) if exitstatus
+    end
+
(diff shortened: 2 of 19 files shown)


lib/rubygems/test_case.rb
```diff
@@ -355,6 +355,13 @@ def setup
     Dir.chdir @tempdir
 
     ENV['HOME'] = @userhome
+    FileUtils.mkdir_p File.join(@userhome, ".gem")
+    File.write File.join(@userhome, ".gemrc"), "--- {}"
+    temp_cred = File.join(@userhome, '.gem', 'credentials')
+    FileUtils.mkdir_p File.dirname(temp_cred)
+    File.write temp_cred, ':rubygems_api_key: 701229f217cdf23b1344c7b4b54ca97'
+    File.chmod 0600, temp_cred
```
deivid-rodriguez, line 363: In my opinion, only the tests that need this should create these files, and they should clean them up after themselves. That makes much more clear the side effects each test has, and also avoids hacks like the one you had to add to `test/rubygems/test_gem_commands_signout_command.rb` where you delete the files if they exists.

hsbt: It affects with the following tests.

* test_gem_config_file.rb
* test_gem_commands_push_command.rb
* test_gem_commands_owner_command.rb
* test_gem_commands_signin_command.rb:

We should avoid to use copy&paste code.

deivid-rodriguez: I totally agree with avoid copy-pasting around large chunks of code. I was suggesting to extract all this logic (and the corresponding teardown part if any), and run those methods on the setup callbacks of the specific tests that need it.

This tests probably make direct use of the specific paths being created here, so having the setup more explicit inside these tests makes them easier to follow and understand. And it also avoids all the unnecessary work of creating all these files for every single test file where only 4 of them need that.

If it was most of the tests needing this, then I would understand putting this in the global setup method (which is already a monster by the way), but that's not the case here.

hsbt: I did split them from global setup scope at 7b360d9

lib/rubygems/test_case.rb
```diff
@@ -357,7 +357,12 @@ def setup
     Dir.chdir @tempdir
 
     ENV['HOME'] = @userhome
+    FileUtils.mkdir_p File.join(@userhome, ".gem")
+    File.write File.join(@userhome, ".gemrc"), "--- {}"
```
deivid-rodriguez, line 361: Following the same criteria as in my other comment. Why are we creating this file after every test? Can we remove this and instead restore the lines in `test/rubygems/test_gem_commands_build_command.rb`, `test/rubygems/test_gem_commands_cert_command.rb` and so on that were creating this file and/or folder only when needed?

bundler/spec/bundler/env_spec.rb
```diff
@@ -45,9 +45,15 @@
       it "prints user path" do
         skip "needs to use a valid HOME" if Gem.win_platform? && RUBY_VERSION < "2.6.0"
 
-        with_clear_paths("HOME", "/a/b/c") do
+        if Gem::VERSION >= "3.2.0.pre.1"
+          allow(Gem).to receive(:data_home) { "/a/b/c/.local/share" }
```
deivid-rodriguez, line 49: I assume `data_home` will only be picked up if it exists? Or this PR is actually changing behavior and a credentials file in `~/.gem/credentials` will no longer be picked up?

bundler/spec/bundler/friendly_errors_spec.rb
```diff
@@ -24,7 +27,11 @@
 
       bundle :install, :env => { "DEBUG" => "true" }
 
-      expect(err).to include("Failed to load #{home(".gemrc")}")
+      if Gem::VERSION >= "3.2.0.pre.1"
+        expect(err).to include("Failed to load #{File.join(config_home, "gemrc")}")
```
deivid-rodriguez, line 31: Same kind of question about the default location of `gemrc`, is it changing? If it's only being picked up if the file exists, I don't think we need all this because we are no longer creating this file for every test.

lib/rubygems/test_case.rb
```diff
@@ -470,6 +475,13 @@ def teardown
     @back_ui.close
   end
 
+  def credential_setup
+    @temp_cred = File.join(@userhome, '.gem', 'credentials')
+    FileUtils.mkdir_p File.dirname(@temp_cred)
+    File.write @temp_cred, ':rubygems_api_key: 701229f217cdf23b1344c7b4b54ca97'
+    File.chmod 0600, @temp_cred
```
deivid-rodriguez, line 482: Are the tests using `credential_setup` cleaning up after themselves? Ideally we would do this in an `around` hook but minitest doesn't have these so we need a separate `crendential_teardown` method.

test/rubygems/test_gem_commands_signout_command.rb
```diff
@@ -8,6 +8,7 @@ class TestGemCommandsSignoutCommand < Gem::TestCase
 
   def setup
     super
+    File.delete Gem.configuration.credentials_path if File.exist?(Gem.configuration.credentials_path)
```
deivid-rodriguez, line 11: We no longer need to do this since now we're only creating this file for the tests that need it, and they should be cleaning it up after themselves.

lib/rubygems/test_case.rb
hsbt: I did try to use `credential_teardown`. But it caused the many of test failures because the some of tests relied on the credential file implicitly.

test/rubygems/test_gem_commands_signout_command.rb
hsbt: Right. Thanks.

bundler/spec/bundler/env_spec.rb
hsbt: > I assume `data_home` will only be picked up if it exists?

Your expectation is correct. RG 3.2.0+ will pick `~/.gem/credentials` only if it exists.

lib/rubygems/test_case.rb
deivid-rodriguez: Do you mind if I have a look and try to fix it?

simi:
I did some local testing and everything seems ok for now. Old files were picked when present (for example gem credentials),and new files were created at new place (supporting XDG specification).

:information_source: Got an idea to also show those new "base dirs" at `gem env` output including credentials path. But I can contribute this once this is merged in different PR.

bundler/spec/bundler/env_spec.rb
deivid-rodriguez: Right. So, the fact that other tests are leaving side effects is creating the false impression that the default has changed here. I'd like to try fix that.

bundler/spec/bundler/friendly_errors_spec.rb
hsbt: Sorry, I'm not sure what you expected. We need to this for the old versions of RubyGems because they use only the old default location.

Can I remove this example?

hsbt:
@simi Thanks for your work. I appreciate it.

lib/rubygems/test_case.rb
hsbt: It's super helpful for me. Because you know I'm always busy with the other Ruby works 😇

bundler/spec/bundler/friendly_errors_spec.rb
deivid-rodriguez: My expectation is that, since we should only create the new file structure (`~/.local/share/whatever') for the tests that need it, and those files should be cleaned up after those tests are done, this test should keep working just like it worked before in a way that is independent from the rubygems version being run.

Otherwise, when I read this test, the first thing that comes to mind is that the default `gemrc` home has changed in rubuygems 3.2, and that my gemrc file will no longer be read when I upgrade.

lib/rubygems/test_case.rb
deivid-rodriguez: Yes, I'm aware :) I'll look at it later today :+1:

bundler/spec/bundler/friendly_errors_spec.rb
hsbt: I see. I will split this example to the XDG_CONFIG_HOME and old configuration.

deivid-rodriguez:
By the way, a lot of tests fail on my system when I use this branch, no idea why yet.

bundler/spec/bundler/friendly_errors_spec.rb
hsbt: I added test for the old gemrc at https://github.com/rubygems/rubygems/pull/3545/commits/b27eefd51f9279fae25b64c85889472098bad4d9

But I'm still not sure what your requested. Can you show the code for this? Thanks.

bundler/spec/bundler/env_spec.rb
hsbt: Same as https://github.com/rubygems/rubygems/pull/3545#discussion_r417241889

I felt your request as test structure. It's helpful for me if you submit the code about this. Thanks.

bundler/spec/bundler/friendly_errors_spec.rb
deivid-rodriguez: I will work on it, sure.

deivid-rodriguez: But so that maybe it's more clear, my expectation is that this PR shouldn't change this test file at all. This PR deals with error reporting, and the default `.gemrc` location hasn't changed, so this PR shouldn't touch this file.

lib/rubygems/test_case.rb
hsbt: I understood your concerns. But if I removed them, over the 50+ tests was failed because it depends on the current `.gem` location or others. I choose the default fallback way on this time. 

I'm happy to fix with https://github.com/rubygems/rubygems/pull/3545#discussion_r417232784

hsbt:
@deivid-rodriguez Can we merget this?

simi:
I can test quickly again.

deivid-rodriguez:
Oh, sorry, forgot to update this. It's ready from my side, yeah. :+1:.

Note that I made no changes to `lib` code whatsoever, only touched tests, so there should be no need to retest locally unless you want to do it for some other reason.

deivid-rodriguez:
With the commits I added, only a single change is needed to bundler specs (on a test that mocks Gem.user_dir, so now needs to mock its existance as well so that the new default doesn't apply). This simplifies the diff for this PR a lot and also makes it more clear that this PR is backwards compatible.

hsbt:
@deivid-rodriguez Your additional commits seems the simple and better than my first commits. Thanks.

deivid-rodriguez:
Thanks for your work!

lgarron:
I'm finding that `gem push` will still create credentials in `~/.gem/credentials`. Is there something I may be missing?

simi:
https://github.com/rubygems/rubygems/blob/957dadcb09f2ea21929183c3af089dfe2a1d1bfe/lib/rubygems/config_file.rb#L268-L278

Even if you're using version able to follow XDG dir specs, it respects `~/.gem/credentials` if already present for smooth transition.

lgarron:
Ah, it seems that this device has only the macOS system default `gem` version, which is `3.0.3.1`.

I've tried installing `gem` v3.3.7 via `rbenv` and that seems to have this fix!

simi:
:thinking: It landed little earlier.

image

lgarron:
> 🤔 It landed little earlier.

Ah, I see, that's just the latest, I needed to expand. Thanks for tip!

PR #3546, opened 2020-04-24
Refactor gemfile-lockfile source checks

deivid-rodriguez:
# Description:

I was trying to understand why some tests regarding gemfile <-> lockfile source equivalence specs were failing while making some changes, and had to go through this code.

In order to make it a bit simpler, I applied some basic prepositional logic rules to it.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -456,7 +456,7 @@ module Bundler
       @locked_deps.each  {|name, d| both_sources[name][1] = d.source }
 
       both_sources.each do |name, (dep, lock_source)|
-        next unless !lock_source.nil? && (dep.nil? || (!dep.nil? && !lock_source.can_lock?(dep)))
+        next if lock_source.nil? || (!dep.nil? && (dep.nil? || lock_source.can_lock?(dep)))
         gemfile_source_name = (dep && dep.source) || "no specified source"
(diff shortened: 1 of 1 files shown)


bundler/lib/bundler/definition.rb
```diff
@@ -456,9 +456,9 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
       @locked_deps.each  {|name, d| both_sources[name][1] = d.source }
 
       both_sources.each do |name, (dep, lock_source)|
-        next unless (dep.nil? && !lock_source.nil?) || (!dep.nil? && !lock_source.nil? && !lock_source.can_lock?(dep))
+        next if lock_source.nil? || (dep && lock_source.can_lock?(dep))
         gemfile_source_name = (dep && dep.source) || "no specified source"
-        lockfile_source_name = lock_source || "no specified source"
```
simi, line 461: If I understand this well, `lock_source` was already checked for presence before and this removed part was never reached thus it is ok to remove. Can you confirm?

deivid-rodriguez: Yes, that's my reasoning. The previous `next` guard skips to the next iteration of `lock_source` is `nil`, so this could only be reached if `lock_source` was `false`. But `false` is not a valid value for a source.

PR #3547, opened 2020-04-24
Spare some unneeded rescue blocks

deivid-rodriguez:
# Description:

The `kernel_exec` method wraps `Kernel.exec` with some rescues to make sure that the file we're exec'ing to exists and is executable. In this branch of the code, though, we already used `Bundler.which` to find a proper executable, so those precautions are not necessary.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Not really sure if this PR is worth it, I think I'll just close it.

PR #3548, opened 2020-04-24
Show RSpec filter annoucements when running plain rspec

deivid-rodriguez:
I think never showing them was unintentional.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/.rspec_parallel
+++ b/bundler/.rspec_parallel
@@ -1,4 +1,3 @@
 --format progress
 --format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
 --require spec_helper
---require support/parallel.rb
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -23,7 +23,6 @@ require_relative "support/filters"
 require_relative "support/helpers"
 require_relative "support/indexes"
 require_relative "support/matchers"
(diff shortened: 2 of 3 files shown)


simi:
:information_source: seems it was introduced by @hsbt at https://github.com/rubygems/bundler/commit/faccc522d11f633ba402082ccc220fa0fd89b78f

deivid-rodriguez:
Indeed.

The explicit introduction of `require 'support/parallel.rb'` inside `.rspec_parallel` inside that commit seems to confirm that it was unintentional. At that point in time we had the following code inside `spec/spec_helper.rb`, which made the helper file also be picked up by regular RSpec runs:

https://github.com/rubygems/bundler/blob/faccc522d11f633ba402082ccc220fa0fd89b78f/spec/spec_helper.rb#L18-L22

```ruby
Dir["#{File.expand_path("../support", __FILE__)}/*.rb"].each do |file|
  file = file.gsub(%r{\A#{Regexp.escape File.expand_path("..", __FILE__)}/}, "")
  require file unless file.end_with?("hax.rb")
end
```

hsbt:
I forgot why I enabled this configuration. Maybe, I faced the some of issues of parallel tests.

deivid-rodriguez:
Yes, it makes sense for parallel tests because the filter announcements get printed one time per thread making them pretty annoying. But for regular RSpec they are ok.

PR #3550, opened 2020-04-25
Rewrite test to verify an actually supported scenario

deivid-rodriguez:
# Description:

This test was editing the lockfile directly, which is not something we explictly support. I changed the test to verify the real scenario we _do_ support.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Thanks @simi!

Issue #3392, opened 2020-03-16
Flaky failure related to parallel installation


deivid-rodriguez:
We're getting the following flaky test failure: https://github.com/rubygems/rubygems/pull/3389/checks?check_run_id=510768406

From a quick look, parallel installation the `rails` gem is failing during a specific `bundle install` command. Also, even if `rails` sometimes fails to be installed, the parallel installer fails to report any errors.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

deivid-rodriguez:
We got another instance of these, not sure if it's exactly the same thing, but I'll post it here:

It happens on jruby: https://github.com/rubygems/rubygems/pull/3471/checks?check_run_id=551960534

```
In particular, this is the bundler subprocess crashing:

        $  /home/runner/.rubies/jruby-9.2.11.0/bin/jruby -I/home/runner/work/rubygems/rubygems/bundler/lib:/home/runner/work/rubygems/rubygems/bundler/spec -rsupport/hax -rsupport/artifice/fail /home/runner/work/rubygems/rubygems/bundler/exe/bundle update --all
       Fetching source index from file:///home/runner/work/rubygems/rubygems/bundler/tmp/2/gems/remote2/
       Resolving dependencies...
       Using bundler 2.2.0.dev
       Fetching rack 1.2 (was 1.0.0)
       Installing rack 1.2 (was 1.0.0)
       NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed in Rubygems 4
       Gem::Specification#rubyforge_project= called from /home/runner/.rubies/jruby-9.2.11.0/lib/ruby/gems/shared/specifications/default/json-2.2.0-java.gemspec:18.
       NOTE: Gem::Specification#rubyforge_project= is deprecated with no replacement. It will be removed in Rubygems 4
       Gem::Specification#rubyforge_project= called from /home/runner/.rubies/jruby-9.2.11.0/lib/ruby/gems/shared/specifications/default/json-2.2.0-java.gemspec:18.
       NameError: uninitialized constant Gem::Platform
       An error occurred while installing activesupport (3.0), and Bundler cannot continue.
       Make sure that `gem install activesupport -v '3.0' --source 'file:///home/runner/work/rubygems/rubygems/bundler/tmp/2/gems/remote2/'` succeeds before bundling.

       In Gemfile:
         activesupport
       # $? => 5
```

[...]

PR #3555, opened 2020-04-25
Debug parallel install error

deivid-rodriguez:
# Description:

This is a refactoring PR to make it easier to figure out https://github.com/rubygems/rubygems/issues/3392.

What we currently know is that a `Bundler::InstallError` with empty message is being raised at:

https://github.com/rubygems/rubygems/blob/7a8048a54ae2304c94b99043609ea6497e7c969d/bundler/lib/bundler/installer/parallel_installer.rb#L197

That error is presumably generated by one of the parallel worker installation threads, and since it's not a `Bundler::BundlerError` because in that case it would be raised in the previous line, I believe it's being passed to the parent thread in the catch-all rescue here:

https://github.com/rubygems/rubygems/blob/7a8048a54ae2304c94b99043609ea6497e7c969d/bundler/lib/bundler/installer/gem_installer.rb#L26-L27

I believe this is too much "error swallowing" since it makes it difficult to identify the real culprit of unexpected errors.

So the end goal of this PR is to replace that catch-all rescue with a more concrete list of exceptions without changing any end user behaviour, so that next time the error appears, it will be more apparent what the cause is.

While working on this I added other minor refactorings too.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/errors.rb
+++ b/bundler/lib/bundler/errors.rb
@@ -56,6 +56,7 @@ module Bundler
   class SudoNotPermittedError < BundlerError; status_code(30); end
   class ThreadCreationError < BundlerError; status_code(33); end
   class APIResponseMismatchError < BundlerError; status_code(34); end
+  class APIResponseInvalidDependenciesError < BundlerError; status_code(35); end
   class GemfileEvalError < GemfileError; end
   class MarshalError < StandardError; end
 
--- a/bundler/lib/bundler/installer/gem_installer.rb
+++ b/bundler/lib/bundler/installer/gem_installer.rb
@@ -23,7 +23,7 @@ module Bundler
       raise
     rescue Errno::ENOSPC
       return false, out_of_space_message
-    rescue StandardError => e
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
Note that in this PR I'm adding a specific error to deal with malformed gemspec dependencies, so that we can handle that specific error but not swallow the rest of the potential errors.

But I'm considering the alternative of completely reverting https://github.com/rubygems/bundler/pull/5819. That would achieve the same thing and allow us to simplify the code. The error (if it's still possible to trigger it) would still happen but it would lead to a bug report template instead of to a regular gem installation error, which in my opinion is a good thing.

Related: https://github.com/rubygems/rubygems.org/issues/2329.

Opinions welcome!

deivid-rodriguez:
We might want to go ahead and fully revert rubygems/bundler#5819 in the future, but let's do this for now!

Thanks both for having a look :)

deivid-rodriguez:
We run into the error again, without better information :( https://github.com/rubygems/rubygems/pull/3562/checks?check_run_id=629143241.

But I'm preparing another patch to try figure it out.

Issue #3521, opened 2020-04-15
Please improve `Gem::MissingSpecError` error message


voxik:
Consider this backtrace:

~~~
/usr/share/rubygems/rubygems/dependency.rb:311:in `to_specs': Could not find 'ipaddr' (>= 0) among 12 total gem(s) (Gem::MissingSpecError)
Checked in 'GEM_PATH=/builddir/.gem/ruby:/usr/share/gems:/usr/local/share/gems', execute `gem env` for more information
	from /usr/share/rubygems/rubygems/specification.rb:1402:in `block in activate_dependencies'
	from /usr/share/rubygems/rubygems/specification.rb:1391:in `each'
	from /usr/share/rubygems/rubygems/specification.rb:1391:in `activate_dependencies'
	from /usr/share/rubygems/rubygems/specification.rb:1373:in `activate'
	from /usr/share/rubygems/rubygems/specification.rb:1405:in `block in activate_dependencies'
	from /usr/share/rubygems/rubygems/specification.rb:1391:in `each'
	from /usr/share/rubygems/rubygems/specification.rb:1391:in `activate_dependencies'
	from /usr/share/rubygems/rubygems/specification.rb:1373:in `activate'
	from /usr/share/rubygems/rubygems.rb:209:in `try_activate'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:161:in `rescue in require'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:156:in `require'
	from /builddir/build/BUILD/rubygem-net-ssh-gateway-1.2.0/usr/share/gems/gems/net-ssh-gateway-1.2.0/lib/net/ssh/gateway.rb:2:in `<top (required)>'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:92:in `require'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:92:in `require'
	from /builddir/build/BUILD/rubygem-net-ssh-gateway-1.2.0/usr/share/gems/gems/net-ssh-gateway-1.2.0/test/gateway_test.rb:4:in `<top (required)>'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:92:in `require'
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:92:in `require'
	from -e:1:in `glob'
	from -e:1:in `<main>'
~~~

[...]

PR #3559, opened 2020-04-26
Improve missing spec error

bronzdoc:
Closes https://github.com/rubygems/rubygems/issues/3521

Show gemspec location when a `Gem::MissingSpecError` is raised while trying to activate a gem

---

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1393,7 +1393,11 @@ class Gem::Specification < Gem::BasicSpecification
         raise e
       end
 
-      specs = spec_dep.to_specs
+      begin
+        specs = spec_dep.to_specs
+      rescue Gem::MissingSpecError => e
+        raise Gem::MissingSpecError.new(e.name, e.requirement, "at: #{self.spec_file}")
+      end
 
       if specs.size == 1
         specs.first.activate
--- a/lib/rubygems/errors.rb
+++ b/lib/rubygems/errors.rb
@@ -28,14 +28,15 @@ module Gem
   # superclass Gem::LoadError to catch all types of load errors.
   class MissingSpecError < Gem::LoadError
 
-    def initialize(name, requirement)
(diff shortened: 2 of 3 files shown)


lib/rubygems/specification.rb
```diff
@@ -1393,7 +1393,11 @@ def activate_dependencies
         raise e
       end
 
-      specs = spec_dep.to_specs
+      begin
+        specs = spec_dep.to_specs
+      rescue Gem::MissingSpecError => e
+        raise Gem::MissingSpecError.new(e.name, e.requirement, "at: #{self.spec_file}")
```
deivid-rodriguez, line 1399: Would it be possible the `raise` with the proper message from the beginning?

bronzdoc: https://github.com/rubygems/rubygems/blob/492496af849f564e96374671f3f3261452268e3d/lib/rubygems/dependency.rb#L311
Unfortunately we don't have access to the `runtime spec` trying to resolve it's dependencies inside `depency.rb` where the original error is raised.

deivid-rodriguez: I see. Then I would raise a separate error like that represents that the runtime dependency `spec_dep` inside the `spec_file` gemspec couldn't be satisfied by any installed specifications, which I belieive is what's happening here, right?

Rescuing an exception and reraising the same exception, but with different arguments feels wrong to me.

deivid-rodriguez: Also, does this same logic would apply to other `to_specs` calls in this file? What about `to_spec` calls?, which internally call `to_specs`?

I think this PR can work to fix the reported problem, but if we merge it as is, we could open a ticket to make a deeper refactoring of this situation.

bronzdoc: Definitively this can be improved! maybe we can optionally pass more context when creating a dependency so we can use spec information when needed in a dependency? 🤔

Issue #3524, opened 2020-04-16
Deprecations for CLI flags are not displayed if they use `--flag=value` format


deivid-rodriguez:
1. Do you have a reproduction script? It would be really helpful for us if you provide one.

```bash
mkdir test
cd test
bundle init
bundle install --path=vendor/bundle
```

2. What are you trying to accomplish?

Deprecation messages that are displayed every time they are expected.

3. What command did you run?

```
bundle install --path=vendor/bundle
```

4. What were you expecting to happen?

```
$ bundle install --path=vendor/bundle
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag
Fetching gem metadata from https://rubygems.org/..............
```

5. What actually happened?

```
bundle install --jobs=3 --retry=3 --path=${BUNDLE_PATH:-vendor/bundle}
Fetching gem metadata from https://rubygems.org/..............
(...)
```

PR #3561, opened 2020-04-27
Show deprecation warning regardles of how you specify a flag

bronzdoc:
Closes https://github.com/rubygems/rubygems/issues/3524

Show deprecations for CLI flags if they use `--flag=value` format

----

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -815,7 +815,7 @@ module Bundler
       option = current_command.options[name]
       flag_name = option.switch_name
 
-      name_index = ARGV.find {|arg| flag_name == arg }
+      name_index = ARGV.find {|arg| flag_name == arg.split("=")[0] }
       return unless name_index
 
       value = options[name]
--- a/bundler/spec/other/major_deprecation_spec.rb
+++ b/bundler/spec/other/major_deprecation_spec.rb
@@ -121,6 +121,28 @@ RSpec.describe "major deprecations" do
     pending "should fail with a helpful error", :bundler => "3"
   end
 
+  context "bundle check --path=" do
+    before do
+      install_gemfile <<-G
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/cli.rb
```diff
@@ -815,7 +815,10 @@ def remembered_flag_deprecation(name)
       option = current_command.options[name]
       flag_name = option.switch_name
 
-      name_index = ARGV.find {|arg| flag_name == arg }
+      name_index = ARGV.find do |arg|
+        flag_name == arg || arg.split("=").find {|str| flag_name == str }
+      end
+
```
deivid-rodriguez, line 821: This implementation is subject to pathological false positives like

```
$ mkdir ./--path
$ bundle install --binstubs=--path
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'path'`, and stop using this flag
[DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs`
Could not locate Gemfile
```

I believe you can go with the simpler

```ruby
      name_index = ARGV.find {|arg| flag_name == arg.split("=")[0] }
```

bronzdoc:
Thanks @deivid-rodriguez! 🙂 

Just for curiosity, why you force pushed to the branch?

deivid-rodriguez:
I fixed a style nit that was some minor that I didn't want to bother you about :sweat_smile:

deivid-rodriguez:
And then I run into like three flaky test failures in a row :rofl:

deivid-rodriguez:
The nitpick was that in bundler's code normally when we create a local variable and use it right after on a guard clause, we don't leave an empty line between the local variable assignment and the guard clause. I see it as part of the same "logical block" of code.

bronzdoc:
Got it!

PR #3566, opened 2020-04-28
Disallow downgrades to too old versions

### LINKED ISSUE #3004: Prevent downgrades to an older version than the one shipped with the current ruby
My current problem is that (don't ask me why, please :rofl:) I run `gem update --system 1.3.0`, and while the downgrade itself worked, that broke my `rubygems` & `bundler` installation because modern rubies don't work with such old `rubygems` versions.

Since, for each ruby version, we only test `bundler` against the `rubygems` versions higher than or equal to the `rubygems` version that shipped with that ruby, I propose that we forbid `gem update --system <target_version>` if `<target_version>` is lower than the `rubygems` version the currently running ruby shipped with.

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

deivid-rodriguez:

The problem is that for each ruby, we only test rubygems versions higher than or equal to the rubygems version that came with that ruby. So we don't really know if older versions actually work under that ruby.

My fix is to prevent downgrading past the original version shipped with each ruby.

For the implementation, I considered different alternatives like:

* Keeping a static list of supported versions.
* Parsing the `rubygems.rb` inside `RbConfig::CONFIG["rubylibdir"]` to figure out the original ruby.

But finally went with the current approach of finding the version through a subprocess that uses that ruby.

Fixes #3004.

# Tasks:

- [x] Write tests


deivid-rodriguez:
I switched to a simpler approach motivated by https://github.com/rubygems/rubygems/issues/3004#issuecomment-620706716.

I also [saved the initial implementation to a separate branch](https://github.com/rubygems/rubygems/tree/minimum_versions_per_ruby) so that we can pick it up again in the future if we decide to try it.

lib/rubygems/commands/update_command.rb
```diff
@@ -75,6 +75,13 @@ def check_latest_rubygems(version) # :nodoc:
     end
   end
 
+  def check_oldest_rubygems(version) # :nodoc:
+    if oldest_supported_version > version
+      alert_error "rubygems #{version} is not supported. The oldest supported version is #{oldest_supported_version}"
```
simi, line 80: Do we need to fail here? Isn't warning enough here?

lib/rubygems/commands/update_command.rb
```diff
@@ -306,4 +315,10 @@ def which_to_update(highest_installed_gems, gem_names, system = false)
     result
   end
 
+  private
+
+  def oldest_supported_version
+    @oldest_supported_version ||= Gem::Version.new("2.5.2")
```
simi, line 321: What's the rationale behind deciding to support `>= 2.5.2` for now?

lib/rubygems/commands/update_command.rb
deivid-rodriguez: The related ticket proposed to prevent downgrades, warnings were not consider.

As I explained in the related ticket, when I downgraded to a very old rubygems, my ruby installation was completely messed up (rubygems is specially sensible to these issues since it's required by ruby by default). A warning wouln't have saved me from anything.

So, yeah, if we don't test against a specific rubygems version, I prefer to call it not supported and prevent downgrades.

lib/rubygems/commands/update_command.rb
deivid-rodriguez: It's the rubygems version that comes by default with the oldest ruby that we support (2.3).

simi: I'm just thinking about the way to not forget about this to move forward once 2.3 support would be dropped.

lib/rubygems/commands/update_command.rb
deivid-rodriguez: Also note that if you really want to downgrade to an ancient version and mess up your installation, you can still do it: downgrade to a rubygems version without this feature, and from there downgrade to wherever :)

lib/rubygems/commands/update_command.rb
simi: maybe some kind of `Ruby 2.3` related comment would be enough

deivid-rodriguez: Aha. I'll add a more useful test that will break if we forget to do that :+1:.

hsbt, line 321: Can you add the comment about the Ruby version? It's hard to understand the Ruby version from RubyGems version.

deivid-rodriguez: Yes, I will, good point :+1:.

bronzdoc: Maybe instead of a comment put it in a constant?

deivid-rodriguez: What should I put in a constant?

deivid-rodriguez: Oh, I see. Yiu mean something like `RUBYGEMS_VERSION_PROVIDED_BY_RUBY_2_3` or something like that?

bronzdoc: Yes

bronzdoc: `RUBYGEMS_VERSION_PROVIDED_BY_OLDEST_RUBY_SUPPORTED`? :shipit:

Issue #3557, opened 2020-04-26
Send notifications to maintainers slack channel when "head" builds break


deivid-rodriguez:
A while ago we configured builds against ruby-head to not run for every PR but daily.

However, I was expecting Github to email or notify somehow when these builds fail, but it does not do that, actually.

So we should enhance scheduled builds to let us know when they break.

I think ideally we would:

* Save the result of the previous run to an artifact. Ideally including number and location of failures.
* When the scheduled run finishes, compare the result with the previous run, and if it changed, send a slack notification with the changes.

For a first version though it would be enough to notify the channel on every fail, although perhaps it would better be a separate channel to avoid too much noise on the maintainers channel.

I found https://github.com/rtCamp/action-slack-notify which could be useful.

deivid-rodriguez:
This actions sounds more appropriate: https://github.com/8398a7/action-slack

deivid-rodriguez:
Yesterday, our daily `truffleruby-head` build [started failing](https://github.com/rubygems/rubygems/actions/runs/153444893) and we didn't get notified.

And I have no idea why :disappointed:

deivid-rodriguez:
Fixed by https://github.com/rubygems/rubygems/pull/3786.

PR #3573, opened 2020-04-30
Notify ruby-head build in slack

bronzdoc:
https://github.com/rubygems/rubygems/issues/3557

-----

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -30,4 +30,30 @@ jobs:
         run: |
           bin/rake spec:all
         working-directory: ./bundler
+
+      - uses: actions/download-artifact@v2
+        with:
+          name: status
+          path: previous_status.txt
+      - name: Save fail status
+        if: failure()
+        run: echo "FAIL" > status.txt
+      - name: Save pass status
+        if: success()
+        run: echo "PASS" > status.txt
+      - name: Compare with previous run
+        run: echo "::set-env name=STATUS_CHANGED::$(diff status.txt previous_status.txt; echo $?)"
+
+      - uses: 8398a7/action-slack@v3
+        with:
+          status: ${{ job.status }}
+          fields: repo,message,commit,author,action,eventName,ref,workflow
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
My idea here would be to save the status of the previous head build to an artifact, and use that to only send notifications when the status has changed.

The rationale is that we don't want to get overwhelmed with useless notifications. We only want to know when the build gets broken, and if someone fixes it, when it starts being green again. Once it's broken we don't need daily notifications to remind us the same thing, and once it's green we don't need daily notifications to remind us that it's green.

Completely untested, but the gist would be something like:

```yaml
- uses: actions/download-artifact@v2
  with:
    name: status
    path: previous_status.txt

- name: Save fail status
  if: failure()
  run: echo "FAIL" > status.txt

- name: Save pass status
  if: success()
  run: echo "PASS" > status.txt

- name: Compare with previous run
  run: echo "::set-env name=STATUS_CHANGED::$(diff status.txt previous_status.txt; echo $?)"

- uses: 8398a7/action-slack@v3
  with:
    status: ${{ job.status }}
    fields: repo,message,commit,author,action,eventName,ref,workflow
    env:
      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
    if: env.STATUS_CHANGED != '0'

- uses: actions/upload-artifact@v2
  with:
    name: status
    path: status.txt
```

deivid-rodriguez:
This is awesome. We should add the same kind of thing to `daily-bundler.yml` as well.

bronzdoc:
@deivid-rodriguez 👀

bronzdoc:
@deivid-rodriguez yeah, I was going to ask that. Let's merge it then :shipit:

deivid-rodriguez:
Actually, I just realized that unless there's a way to manually upload an artifact, we're going to need to do this in two steps.

First change the workflow to only upload the artifact with the build status, and then add the rest of the stuff. Because otherwise the initial artifact download is going to fail because there will be no such artifact, right?

deivid-rodriguez:
Ok, so there should be a "fun" but easy fix:

* Push a branch named "foo" that modifies the relevant workflows to run on `push` events on branch "foo", and remove the all the changes you added here except for the final "upload_artifact" part.

That should result in the current status properly uploaded as an artifact and should make the future scheduled run work I believe.

PR #3575, opened 2020-04-30
Speed up some terribly slow tests

deivid-rodriguez:

The problem was that some tests, 4 to be precise, were really really slow, because each of them created 500 dummy gems in 500 separate subprocesses.

My fix is to, instead of making a very big index so that the API request limit is reached, we should mock the API request limit, so that creating dummy gems is not needed.

This results in a 99% improvement on the sequential running time of these tests, which brings it down from 4 minutes to 4 seconds on my machine, and I believe it could make specs actually runnable under jruby.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
I'm going to merge this since it's quite noticeable, so it helps me with my compulsive running of bundler specs.

simi:
:shipit:

Issue #3578, opened 2020-04-30
Improve error message if `rake` is not found when building a "Rakefile" extension


deivid-rodriguez:
If one of the dependencies uses `rake` to compile its extension, but doesn't declare `rake` as a dependency, one get an error like the one in https://github.com/rubygems/rubygems/pull/2326.

We could probably improve the error message to suggest potential remediation steps.

simi:
Should we maybe warn on build time when `Rakefile` extension is provided, but `rake` is not added as a dependency?

deivid-rodriguez:
I think that's a very nice idea.

PR #3580, opened 2020-04-30
Add build warning when rake based extension is present without rake dep...

simi:
potentially closes https://github.com/rubygems/rubygems/issues/3578

### DIFF
--- a/lib/rubygems/ext/build_error.rb
+++ b/lib/rubygems/ext/build_error.rb
@@ -2,5 +2,7 @@
 ##
 # Raised when there is an error while building extensions.
 
+require 'rubygems/exceptions'
+
 class Gem::Ext::BuildError < Gem::InstallError
 end
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -1,4 +1,3 @@
-require 'rubygems/ext'
 require 'rubygems/user_interaction'
 
 class Gem::SpecificationPolicy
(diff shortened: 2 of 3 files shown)


lib/rubygems/specification_policy.rb
```diff
@@ -417,6 +420,15 @@ def validate_removed_attributes # :nodoc:
     end
   end
 
+  def validate_extensions # :nodoc:
+    builder = Gem::Ext::Builder.new(@specification)
+
+    rake_extension = @specification.extensions.any? {|s| builder.builder_for(s) == Gem::Ext::RakeBuilder }
```
simi, line 426: This can be moved to Gem`::Ext::Builder` as well and we can just call `builder.rake_extension?` in here. WDYT?

simi:
:information_source: not sure why some builds were cancelled. CI failures look unrelated to this change.

MSP-Greg:
>  not sure why some builds were cancelled

```yaml
    strategy:
      fail-fast: false
```

lib/rubygems/ext/build_error.rb
```diff
@@ -2,5 +2,7 @@
 ##
 # Raised when there is an error while building extensions.
 
+require 'rubygems/exceptions'
```
deivid-rodriguez, line 5: Why this new require?

lib/rubygems/specification_policy.rb
deivid-rodriguez: Looks good to me either as is or moving the method, whatever you prefer.

lib/rubygems/specification_policy.rb
```diff
@@ -417,6 +420,17 @@ def validate_removed_attributes # :nodoc:
...
+    rake_dependency = @specification.dependencies.any? {|d| d.name == 'rake'}
+
+    warning <<-WARNING if rake_extension && !rake_dependency
+You have specified rake based extension, but rake is not added as dependency. It is recommended to add rake as a dependency since there's no guarantee rake will be already installed.
```
deivid-rodriguez, line 430: Normally `rake` is put in the `Gemfile` by the default `bundle gem` generator. To prevent any confusion, maybe it'd be good to specify "added as a dependency in your gemspec file", or mention the gemspec in some other way.

lib/rubygems/ext/build_error.rb
simi: otherwise I got error running test, but I can take a look if that's not test problem

deivid-rodriguez: Oh, I see the "require chain" now and how the error happens.

This specific require makes sense to me since a constant it defines is used right after.

What I would maybe do is moving the `rubygems/ext` require in `lib/rubygems/specification_policy.rb` to the first line inside the new `validate_extensions` method to make sure we don't load more code than necessary upfront.

simi: done

lib/rubygems/specification_policy.rb
simi: ok, let's keep it simple for now as is

lib/rubygems/specification_policy.rb
simi: I have mentioned gemspec. Can you check again?

PR #3583, opened 2020-05-01
Delay removal of git shortcut DSL additions

### LINKED ISSUE #3579: Git based shortcuts are broken when running in "bundler 3 mode"
As per https://github.com/rubygems/bundler/pull/7000, they should only get deprecated, but not stop working.
Please answer the following questions so we can process your issue as fast as possible:

1. Do you have a reproduction script?

Run `bundle install` on a Gemfile using a `:github` source under a "bundler 3 build" (edit `lib/bundler/version.rb` on a local clone and run `bin/rake install`).

2. What are you trying to accomplish?

A smooth deprecation path.

3. What command did you run?

`bundle install`.

4. What were you expecting to happen?

A deprecation message about the `:github` source.

5. What actually happened?

```
[!] There was an error parsing `Gemfile`: You passed :github as an option for gem 'rails', but it is invalid. Valid options are: group, groups, git, path, glob, name, branch, ref, tag, require, submodules, platform, platforms, type, source, install_if, gemfile. You may be able to resolve this by upgrading Bundler to the newest version.. Bundler cannot continue.

 #  from /home/deivid/Code/playground/bundler/shortcuts/Gemfile:5
 #  
 >  gem "rails", github: "rails/rails"
```

I found this issue while looking at https://github.com/rubygems/bundler-site/pull/513.

deivid-rodriguez:
# Description:

Fixes #3579.

We delayed the deprecation of git based DSL shortcuts in rubygems/bundler#7000, but didn't delay their removal.

This PR does that.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/feature_flag.rb
+++ b/bundler/lib/bundler/feature_flag.rb
@@ -41,7 +41,7 @@ module Bundler
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
     settings_flag(:print_only_version_number) { bundler_3_mode? }
     settings_flag(:setup_makes_kernel_gem_public) { !bundler_3_mode? }
-    settings_flag(:skip_default_git_sources) { bundler_3_mode? }
+    settings_flag(:skip_default_git_sources) { bundler_4_mode? }
     settings_flag(:specific_platform) { bundler_3_mode? }
     settings_flag(:suppress_install_using_messages) { bundler_3_mode? }
     settings_flag(:unlock_source_unlocks_spec) { !bundler_3_mode? }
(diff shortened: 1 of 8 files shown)


bundler/lib/bundler/feature_flag.rb
```diff
@@ -41,7 +41,7 @@ def self.settings_method(name, key, &default)
     settings_flag(:plugins) { @bundler_version >= Gem::Version.new("1.14") }
     settings_flag(:print_only_version_number) { bundler_3_mode? }
     settings_flag(:setup_makes_kernel_gem_public) { !bundler_3_mode? }
-    settings_flag(:skip_default_git_sources) { bundler_3_mode? }
+    settings_flag(:skip_default_git_sources) { bundler_4_mode? }
```
deivid-rodriguez, line 44: In this version I delayed the deprecation, but I believe there's no use of maintaining code that never gets run by either end users or tests, so I would rather actually remove this code until we release bundler 3, and only then restore it.

PR #3584, opened 2020-05-01
Make bundle gem create travis.yml by default

FTLam11:
# Description:

While working on a different PR to update selecting a test framework for `bundle gem`, I found that the creation of `travis.yml` is dependent on the presence of a test framework being configured (via `--test` or `Bundler.settings["gem.test"]`) when `bundle gem` is executed.

Testing is one of multiple CI responsibilities, so perhaps it makes more sense to generate `travis.yml` by default, even if a user has explicitly configured no test framework?

This PR will allow `bundle gem` to create `travis.yml` by default instead of depending on whether a test framework has been configured.

______________

# Tasks:

- [X] Write tests

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -79,12 +79,11 @@ module Bundler
       ]
 
       templates.merge!("gitignore.tt" => ".gitignore") if Bundler.git_present?
+      templates.merge!("travis.yml.tt" => ".travis.yml")
 
       if test_framework = ask_and_set_test_framework
         config[:test] = test_framework
         config[:test_framework_version] = TEST_FRAMEWORK_VERSIONS[test_framework]
-
-        templates.merge!("travis.yml.tt" => ".travis.yml")
 
         case test_framework
         when "rspec"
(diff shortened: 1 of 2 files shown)


simi:
I think having useful files only generated by default make sense. In this configuration travis would run just `rake` (per https://docs.travis-ci.com/user/languages/ruby/). How is that useful? Instead we should provide travis.yml preconfigured to actually do the build per testing framework selected :thinking: .

deivid-rodriguez:
The current generator configures the default `rake` task to run tests for the specified framework, so I think this is fine.

In any case, if we wanted to change this, it's a different issue from what this PR is fixing, right?

simi:
If I understand it well if you do not select any test framework, `Rakefile` will not be enhanced to run test suite by default, `rake` will not run any tests and fail instead and thus this `.travis.yml would result into failed build by default.

deivid-rodriguez:
Oh, you're totally right! :facepalm:

 A TravisCI file like the one we're generating is only useful for running tests. If a user doesn't want tests, it's very likely that she won't want any CI configuration either.

@FTLam11 Are we missing something?

FTLam11:
No, I don't have anything to add at the moment. With no testing framework, it looks like the default task is configured to do nothing (instead of `rake spec` or `rake test`). Closing this PR, thanks for the review!

PR #3595, opened 2020-05-04
Don't load system rubygems at all when `RGV` specified for tests

deivid-rodriguez:
# Description:

In #3592 I noticed that "system rubygems" is being loaded by our tests. I think this is fine if we don't specify `ENV["RGV"]` but if we do, rubygems should always be picked up from there.

This PR tries to implement that.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/ubuntu-bundler.yml
+++ b/.github/workflows/ubuntu-bundler.yml
@@ -27,6 +27,7 @@ jobs:
           - { ruby: 2.7.1, rgv: v3.0.6  }
     env:
       RGV: ${{ matrix.rgv }}
+      RUBYOPT: --disable-gems
     steps:
       - uses: actions/checkout@v2
       - name: Setup ruby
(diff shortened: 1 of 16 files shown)


deivid-rodriguez:
Yay!

Bundler tested against a specific rubgems version (setting `ENV["RGV"]`) no longer load system rubygems whatsover. This has the following advantages:

* Makes tests more isolated from the environment where they are run.
* Hopefully gets #3592 green, since the failures there seem to indicate leakages to the system `bundler`.
* Checking for this uncovered at least two places that were leaking to either system bundler or system rubygems. This code is at the least unneeded, and maybe causes bugs, so I'm killing it in https://github.com/rubygems/rubygems/pull/3595/commits/4eb4ebea2837336364ade457f9d3a73a4b7856e1 and https://github.com/rubygems/rubygems/pull/3595/commits/6e71e7be6789d40add645e924c0024dd2f7d3d2c.

The test I added involve an `at_exit` hook, which is not super pretty but much better than the initial solution I tried: `rm -rf <system_rubygems>` :laughing:.

deivid-rodriguez:
Merging this. Happy to address comments/concerns on followup PRs!

Issue #3226, opened 2018-08-30
Bundler connection to rubygems.org failed; Ruby & RubyGems connection succeeded


EvanGrillo:
## input
```
ruby -ropen-uri -e 'eval open("https://git.io/vQhWq").read'
```
## output
```
Here's your Ruby and OpenSSL environment:

Ruby:           2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin16]
RubyGems:       2.7.7
Bundler:        1.16.2
Compiled with:  OpenSSL 1.0.2p  14 Aug 2018
Loaded version: OpenSSL 1.0.2p  14 Aug 2018
SSL_CERT_FILE:  /usr/local/etc/openssl/cert.pem
SSL_CERT_DIR:   /usr/local/etc/openssl/certs

With that out of the way, let's see if you can connect to rubygems.org...

Bundler connection to rubygems.org:       failed  ❌  (execution expired)
RubyGems connection to rubygems.org:      success ✅
Ruby net/http connection to rubygems.org: success ✅

Although your Ruby installation and RubyGems can both connect to rubygems.org, Bundler is having trouble. The most likely way to fix this is to upgrade Bundler by running `gem install bundler`. Run this script again after doing that to make sure everything is all set. If you're still having trouble, check out the troubleshooting guide at http://ruby.to/ssl-check-failed 📦
(eval):136: warning: constant OpenSSL::SSL::SSLContext::METHODS is deprecated
```

deivid-rodriguez:
Anyone who participated in this thread is using ipv6? You can check https://ipv6-test.com/, for example. I'm doing some triaging for issues that could get solved by #2662.

deivid-rodriguez:
Nobody answered here, so I'll assume this is no longer an issue.

Issue #2677, opened 2019-03-08
 Latency or Timeout Reaching to: https://rubygems.global.ssl.fastly.net/test in China Network


Dragoncell:
I'm having a problem 

Impact Time: 05:45-09:00 PDT 03/08/2019
Latency or Timeout Reaching to: https://rubygems.global.ssl.fastly.net/test 

This issue is related to:

  - [ X] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

colby-swandale:
I doubt this is an issue with the RubyGems library, i suggest moving this issue to https://github.com/rubygems/rubygems.org

Issue #1701, opened 2016-08-27
Constant timeouts with rubygems.global.ssl.fastly.net - Gitlab.com CI / DigitalOcean


Nowaker:
I've recently migrated from Codeship to GitLab.com's internal CI and most of the time the build won't pass due to `Errno::ETIMEDOUT` from rubygems.global.ssl.fastly.net. The application I build consists of multiple git submodules, each having their own Gemfile, so there's a lot of deps to download. So many that my chances for a successful build are quite small. There's no particular gem that doesn't download - there's always *some* gem that won't download. Examples from three different builds:

```
Gem::RemoteFetcher::FetchError: Errno::ETIMEDOUT: Connection timed out - connect(2) for "rubygems.global.ssl.fastly.net" port 443 (https://rubygems.org/gems/web-console-2.2.1.gem)
```
```
Gem::RemoteFetcher::FetchError: Errno::ETIMEDOUT: Connection timed out - connect(2) for "rubygems.global.ssl.fastly.net" port 443 (https://rubygems.org/gems/tilt-2.0.1.gem)
```
```
Gem::RemoteFetcher::FetchError: Errno::ETIMEDOUT: Connection timed out - connect(2) for "rubygems.global.ssl.fastly.net" port 443 (https://rubygems.org/gems/rack-protection-1.5.3.gem)
```

Quote from #557 by @drbrain:

> ETIMEDOUT indicates there is a problem with your network connection, probably with your local machine. Please check for a firewall or other configuration change you made.

[GitLab.com CI uses DigitalOcean for builds](https://about.gitlab.com/2016/04/19/gitlab-partners-with-digitalocean-to-make-continuous-integration-faster-safer-and-more-affordable/), as seen in the build log below, so "problem with local network connection" is out of question. Everyone uses DigitalOcean. Docker image used for the build is Docker's official [ruby:2.2.2](https://hub.docker.com/_/ruby/) - here's the [Dockerfile](https://github.com/docker-library/ruby/blob/2d6449f03976ededa14be5cac1e9e070b74e4de4/2.2/Dockerfile).

```
Running on runner-30dcea4b-project-1582107-concurrent-0 via runner-30dcea4b-machine-1472325066-7915ab81-digital-ocean-4gb...
```

Here's some debug info:
```
$ gem env version
2.4.8
```

Anything else you need?

Nowaker:
[...]

Issue #585, opened 2013-07-03
gem 2.0.3 Unable to download data from https://rubygems.org/ - 


sri85:
I am getting the following error while tryin to install watir- web driver

```
ERROR:  Could not find a valid gem 'watir-webdriver' (>= 0), here is why:
          Unable to download data from https://rubygems.org/ - SSL_connect SYSCALL returned=5 errno=0 state=SSLv2/v3 read server hello A (https://rubygems.org/latest_specs.4.8.gz)
```

gem install watir-webdriver --no-ri --no-rdoc

PR #3605, opened 2020-05-07
Add max_retries, replace_resolv, and timeout for slow internet connections

esotericpig:
# Description:

Possible related issues: #2677 #3226 #1701 #585

In short, my internet sucks, which means, I have to call `gem install <some gem>` and `gem update` multiple times until they finally work.

By adding a `max_retries` and `timeout` setting, through my testing, I now only need to call these commands once.

I think this is important for people that live in countries/cities with poor quality of internet (especially developing countries).

When running `gem install <some gem>` or `gem update` with bad/slow internet, you'll constantly get this error:

`Failed to open TCP connection to rubygems.org:443 (getaddrinfo: Name or service not known) (SocketError)`

because the DNS resolver times out, or you'll get general timeout errors (Open/Read/SSL connect timeout errors).

By default, nothing is set/changed for users.

The user must change their `~/.gemrc` file:

```yaml
:max_retries: 3
:replace_resolv: true
:timeout: 300
```

`max_retries` must be an integer.

`timeout` is the number of seconds, which can be `nil`, a float, or an int. It sets the Open, Read, SSL, and Continue timeouts (via [Net::HTTP](https://ruby-doc.org/stdlib-2.7.0/libdoc/net/http/rdoc/Net/HTTP.html)).

`replace_resolv` will do:

```ruby
require 'resolv'
require 'resolv-replace'
```

In my testing, this didn't improve anything for me, but many Stackoverflow answers cite this as a way to improve DNS issues in Ruby with slow internet. I would recommend not setting this, and instead just using this:

```yaml
:max_retries: 3
:timeout: 300
```

Here are the default values:

```yaml
:max_retries: 0
:replace_resolv: false
:timeout: null # nil
```

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -28,6 +28,9 @@ require 'rbconfig'
 # +:sources+:: Sets Gem::sources
 # +:verbose+:: See #verbose
 # +:concurrent_downloads+:: See #concurrent_downloads
+# +:max_retries+:: See #max_retries
+# +:replace_resolv+:: See #replace_resolv
+# +:timeout+:: See #timeout
 #
 # gemrc files may exist in various locations and are read and merged in
 # the following order:
@@ -46,6 +49,9 @@ class Gem::ConfigFile
   DEFAULT_UPDATE_SOURCES = true
   DEFAULT_CONCURRENT_DOWNLOADS = 8
   DEFAULT_CERT_EXPIRATION_LENGTH_DAYS = 365
+  DEFAULT_MAX_RETRIES = 0
+  DEFAULT_REPLACE_RESOLV = false
+  DEFAULT_TIMEOUT = nil
 
   ##
   # For Ruby packagers to set configuration defaults.  Set in
@@ -113,6 +119,34 @@ class Gem::ConfigFile
   attr_accessor :concurrent_downloads
 
   ##
+  # Number of times to retry downloads for slow internet connections.
+  # Must be an integer.
+  #
+  # See Gem::Request
+  # See Gem::RemoteFetcher
+
+  attr_accessor :max_retries
+
+  ##
+  # +true+ to require +'resolv-replace'+ for slow internet connections,
+  # else, +false+ to not.
+  #
+  # See Gem::Request
+  # See Gem::RemoteFetcher
+
+  attr_accessor :replace_resolv
+
+  ##
+  # Open, Read, SSL, and Continue timeouts in seconds for slow internet connections.
+  # +nil+ will use the default values (i.e., not set anything).
+  # Can be an integer, float, or +nil+.
+  #
+  # See Gem::Request
+  # See Gem::RemoteFetcher
+
+  attr_accessor :timeout
+
+  ##
   # True if we want to update the SourceInfoCache every time, false otherwise
 
   attr_accessor :update_sources
@@ -176,6 +210,9 @@ class Gem::ConfigFile
     @update_sources = DEFAULT_UPDATE_SOURCES
     @concurrent_downloads = DEFAULT_CONCURRENT_DOWNLOADS
     @cert_expiration_length_days = DEFAULT_CERT_EXPIRATION_LENGTH_DAYS
+    @max_retries = DEFAULT_MAX_RETRIES
+    @replace_resolv = DEFAULT_REPLACE_RESOLV
+    @timeout = DEFAULT_TIMEOUT
 
     operating_system_config = Marshal.load Marshal.dump(OPERATING_SYSTEM_DEFAULTS)
     platform_config = Marshal.load Marshal.dump(PLATFORM_DEFAULTS)
@@ -204,6 +241,15 @@ class Gem::ConfigFile
     @disable_default_gem_server  = @hash[:disable_default_gem_server]  if @hash.key? :disable_default_gem_server
     @sources                     = @hash[:sources]                     if @hash.key? :sources
     @cert_expiration_length_days = @hash[:cert_expiration_length_days] if @hash.key? :cert_expiration_length_days
+    @timeout                     = @hash[:timeout]                     if @hash.key? :timeout
+
+    if @hash.key? :max_retries
+      @max_retries = @hash[:max_retries].to_i() # Must be an int
+    end
+
+    if @hash.key? :replace_resolv
+      @replace_resolv = @hash[:replace_resolv] ? true : false
+    end
 
     @ssl_verify_mode  = @hash[:ssl_verify_mode]  if @hash.key? :ssl_verify_mode
     @ssl_ca_cert      = @hash[:ssl_ca_cert]      if @hash.key? :ssl_ca_cert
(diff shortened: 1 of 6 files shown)


simi:
@esotericpig looks amazing! I have never heard of `resolv-replace`. I'll take a look what it actually do.

If I understand it well, it doesn't change anything by default and you need to enable it in `.gemrc`. :heavy_check_mark: 

:thinking: Later we can also set default retry to non-zero.

esotericpig:
Hi, thanks. I don't know why the tests fail on windows-bundler. Is that due to my code or something else? The failing tests are cryptic to me.

deivid-rodriguez:
@esotericpig It's [a known random test failure](https://github.com/rubygems/rubygems/issues/3448), don't worry about it :+1:

lib/rubygems/remote_fetcher.rb
```diff
@@ -92,6 +114,10 @@ def initialize(proxy=nil, dns=nil, headers={})
     @cert_files = Gem::Request.get_cert_files
 
     @headers = headers
+
+    @max_retries = max_retries.to_i() # Must be an int
```
bronzdoc, line 118: Not need to use parens at the end of `to_i`

lib/rubygems/config_file.rb
```diff
@@ -204,6 +241,15 @@ def initialize(args)
     @disable_default_gem_server  = @hash[:disable_default_gem_server]  if @hash.key? :disable_default_gem_server
     @sources                     = @hash[:sources]                     if @hash.key? :sources
     @cert_expiration_length_days = @hash[:cert_expiration_length_days] if @hash.key? :cert_expiration_length_days
+    @timeout                     = @hash[:timeout]                     if @hash.key? :timeout
+
+    if @hash.key? :max_retries
+      @max_retries = @hash[:max_retries].to_i() # Must be an int
```
bronzdoc, line 247: No need for parens at the end of `to_i`

lib/rubygems/request.rb
```diff
@@ -27,14 +33,27 @@ def self.proxy_uri(proxy) # :nodoc:
...
+
     @uri = uri
     @request_class = request_class
     @last_modified = last_modified
     @requests = Hash.new 0
     @user_agent = user_agent
 
     @connection_pool = pool
+
+    @max_retries = max_retries.to_i() # Must be an int
```
bronzdoc, line 55: Not need for parens at the end of `to_i`

lib/rubygems/request.rb
```diff
@@ -191,8 +210,24 @@ def self.get_proxy_from_env(scheme = 'http')
   def perform_request(request) # :nodoc:
     connection = connection_for @uri
 
+    if @max_retries > 0
+      connection.max_retries = @max_retries
+
+      verbose "using max retries of #{@max_retries}"
+    end
+
+    if !@timeout.nil?()
```
bronzdoc, line 219: Not need for parens at the end of `nil?`

lib/rubygems/remote_fetcher.rb
esotericpig: Okay, I changed it. I know it's not necessary, just stylistic choice. If possible, this should be added to rubocop I think, if this is rubygem's style convention.

esotericpig:
simi mentioned that I should change the default value to `nil`. Is this necessary?

There are other values in `config_file.rb` that don't default to `nil`, so I thought it was okay.

If it's necessary, I'll change the default value to `nil`.

lib/rubygems/remote_fetcher.rb
bronzdoc: Agree. Thanks for making the changes and sorry for the hassle.

David just added a cop for this https://github.com/rubygems/rubygems/pull/3682

deivid-rodriguez:
This sounds like it could be related to #2662 and might not be needed if we introduce that PR? Can you try that PR and see if your connection behaviour improves?

deivid-rodriguez:
So, I should assume you don't have ipv6?

> Hi thanks. I don't think falling back to IPv4 will fix it, as per the issue, the errors were related to rubygems.org being unreachable, but I never got an error stating it was unreachable. Also, it would eventually download/update after multiple tries on my end.

@sonalkr132 will know best but this sounds very similar to the ipv6 problems people are having.

In any case, I'm not opposed to introducing this (in particular `:timeout` and `:max_retries`, not sure about `:replace_resolve`) I was just curious whether you had ipv6 and whether trying out the linked PR made things better for you.

sonalkr132:
> getaddrinfo: Name or service not known

This doesn't seem like IPv6 issue to me. As the user mentioned, the inability to resolve IP addresses for rubygems.org can occur when they are on a slow network.
IPv6 connection issue is either because of misconfigured IPv6 routes on the user's part or because of misbehaving Fastly PoP servers. `Name or service not known` means the client was not able to resolve rubygems.org IPs.

deivid-rodriguez:
Ah, ok. I thought this case could also be due to IPv6 issues on the client in case there's a IPv6 address configured as the DNS server.

deivid-rodriguez:
If that was the case though, the linked PR wouldn't fix it since it doesn't implement Happy Eyeballs on the DNS resolution itself.

sonalkr132:
Our DNS provider is separate from the CDN provider. I am yet to come across a case where there was a connection issue because the DNS query for the `AAAA` record failed but the `A` record worked.

deivid-rodriguez:
Ok, that's ruled out then! I was just talking from what I found out from reading the RFC, but I definitely lack hands on experience with this issue.

deivid-rodriguez:
> I am yet to come across a case where there was a connection issue because the DNS query for the AAAA record failed but the A record worked.

Technically (if I'm understanding the theory of this right), both DNS queries would fail in the (potential) case I'm talking about here, since both queries would be sent to a IPv6 host. The issue would be using the IPv6 name server without fallback, instead of falling back to a IPv4 name server in case there are issues when resolving over IPv6.

In any case, I totally trust you that this is not an issue in practice, just wanted to clarify what I mean. It would be this section of the RFC: https://tools.ietf.org/html/rfc8305#section-3.1.

sonalkr132:
>  both queries would be sent to a IPv6 host. The issue would be using the IPv6 name server without fallback, instead of falling back to a IPv4 name server in case there are issues when resolving over IPv6.

yes, this is technically possible. 
In my experience, it is very uncommon for anyone to use IPv6 nameserver on their host. An IPv4 nameserver (ex: 8.8.8.8) can respond to both `AAAA` and `A` queries.

deivid-rodriguez:
>  In my experience, it is very uncommon for anyone to use IPv6 nameserver on their host. An IPv4 nameserver (ex: 8.8.8.8) can respond to both AAAA and A queries.

Makes sense, yeah. The little adoption of IPv6 so far seems a bit like a chicken and egg problem to me: people don't support it because it gives trouble, and it gives trouble because people don't support it :laughing:.

esotericpig:
It's been 4 years. Closing my pull request.

simi:
@esotericpig by any chance, have your internet connection situation improved?

esotericpig:
I moved to another place, so it's fine now. However, I do think it'd be good to have these or similar changes to help people who may live in an area with poor internet or far away from the RubyGems servers.

PR #3606, opened 2020-05-07
Update links from rubygems/bundler to rubygems/rubygems

sonicdoe:
Updates links from [github.com/rubygems/bundler](https://github.com/rubygems/bundler) to [github.com/rubygems/rubygems](https://github.com/rubygems/rubygems) as the former repository has been archived on GitHub.

### DIFF
--- a/bundler/README.md
+++ b/bundler/README.md
@@ -55,8 +55,8 @@ While some Bundler contributors are compensated by Ruby Together, the project ma
 
 ### Code of Conduct
 
-Everyone interacting in the Bundler project's codebases, issue trackers, chat rooms, and mailing lists is expected to follow the [Bundler code of conduct](https://github.com/rubygems/bundler/blob/master/CODE_OF_CONDUCT.md).
+Everyone interacting in the Bundler project's codebases, issue trackers, chat rooms, and mailing lists is expected to follow the [Bundler code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).
 
 ### License
 
-Bundler is available under an [MIT License](https://github.com/rubygems/bundler/blob/master/LICENSE.md).
+Bundler is available under an [MIT License](https://github.com/rubygems/rubygems/blob/master/bundler/LICENSE.md).
--- a/bundler/bundler.gemspec
+++ b/bundler/bundler.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |s|
 
   if s.respond_to?(:metadata=)
     s.metadata = {
-      "bug_tracker_uri" => "https://github.com/rubygems/rubygems/issues",
+      "bug_tracker_uri" => "https://github.com/rubygems/rubygems/issues?q=is%3Aopen+is%3Aissue+label%3ABundler",
       "changelog_uri" => "https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md",
       "homepage_uri" => "https://bundler.io/",
       "source_code_uri" => "https://github.com/rubygems/rubygems/",
--- a/bundler/bundler.gemspec
+++ b/bundler/bundler.gemspec
@@ -24,10 +24,10 @@ Gem::Specification.new do |s|
 
   if s.respond_to?(:metadata=)
     s.metadata = {
-      "bug_tracker_uri" => "https://github.com/rubygems/bundler/issues",
-      "changelog_uri" => "https://github.com/rubygems/bundler/blob/master/CHANGELOG.md",
+      "bug_tracker_uri" => "https://github.com/rubygems/rubygems/issues",
+      "changelog_uri" => "https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md",
       "homepage_uri" => "https://bundler.io/",
-      "source_code_uri" => "https://github.com/rubygems/bundler/",
+      "source_code_uri" => "https://github.com/rubygems/rubygems/",
     }
   end
 
--- a/bundler/doc/contributing/ISSUES.md
+++ b/bundler/doc/contributing/ISSUES.md
@@ -46,6 +46,6 @@ If your version of Bundler does not have the `bundle env` command, then please i
 
 If you have either `rubygems-bundler` or `open_gem` installed, please try removing them and then following the troubleshooting steps above before opening a new ticket.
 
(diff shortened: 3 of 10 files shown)


bundler/README.md
```diff
@@ -1,5 +1,5 @@
 [![Version     ](https://img.shields.io/gem/v/bundler.svg?style=flat)](https://rubygems.org/gems/bundler)
-[![Inline docs ](https://inch-ci.org/github/rubygems/bundler.svg?style=flat)](https://inch-ci.org/github/rubygems/bundler)
+[![Inline docs ](https://inch-ci.org/github/rubygems/rubygems.svg?style=flat)](https://inch-ci.org/github/rubygems/rubygems)
```
sonicdoe, line 2: This link may not be as relevant as before because it now applies to the whole RubyGems project, however, I’ve decided to just update it in this pull request.

bundler/doc/contributing/ISSUES.md
```diff
@@ -46,6 +46,6 @@ If your version of Bundler does not have the `bundle env` command, then please i
 
 If you have either `rubygems-bundler` or `open_gem` installed, please try removing them and then following the troubleshooting steps above before opening a new ticket.
 
-[Create a gist](https://gist.github.com) containing all of that information, then visit the [Bundler issue tracker](https://github.com/rubygems/bundler/issues) and [create a ticket](https://github.com/rubygems/bundler/issues/new) describing your problem and linking to your gist.
+[Create a gist](https://gist.github.com) containing all of that information, then visit the [Bundler issue tracker](https://github.com/rubygems/rubygems/issues) and [create a ticket](https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md) describing your problem and linking to your gist.
```
sonicdoe, line 49: I’ve added two parameters so that the issue is automatically labeled “Bundler” and the [Bundler-related template](https://github.com/rubygems/rubygems/blob/master/.github/ISSUE_TEMPLATE/bundler-related-issue.md) is presented.

bundler/lib/bundler/friendly_errors.rb
```diff
@@ -100,15 +100,15 @@ def request_issue_report_for(e)
         #{issues_url(e)}
 
         If there aren't any reports for this error yet, please create copy and paste the report template above into a new issue. Don't forget to anonymize any private data! The new issue form is located at:
-        https://github.com/rubygems/bundler/issues/new
+        https://github.com/rubygems/rubygems/issues/new?labels=Bundler
```
sonicdoe, line 103: I didn’t put the [Bundler-related template](https://github.com/rubygems/rubygems/blob/master/.github/ISSUE_TEMPLATE/bundler-related-issue.md) here because the message already asks one to use the report template above.

bundler/lib/bundler/friendly_errors.rb
```diff
@@ -100,15 +100,15 @@ def request_issue_report_for(e)
...
       EOS
     end
 
     def issues_url(exception)
       message = exception.message.lines.first.tr(":", " ").chomp
       message = message.split("-").first if exception.is_a?(Errno)
       require "cgi"
-      "https://github.com/rubygems/bundler/search?q=" \
+      "https://github.com/rubygems/rubygems/search?q=" \
```
sonicdoe, line 111: We could filter by the “Bundler” label but it might be better to just search for _all_ relevant issues (even if they haven’t been properly tagged yet).

bundler/README.md
deivid-rodriguez: I'm not sure if anybody cares about inch-ci. And I'm also not sure whether we could configure it to manage docs for both projects in this repo.

But for now I wouldn't touch this since at least it makes it clear that the docs page is not updated since the repo merge. There's indeed a rubygems inch-ci page too, but [it hasn't been built since 4 years ago](https://inch-ci.org/github/rubygems/rubygems/history?branch=master) so that one I'm pretty sure nobody cares about.

bundler/bundler.gemspec
```diff
@@ -24,10 +24,10 @@ Gem::Specification.new do |s|
 
   if s.respond_to?(:metadata=)
     s.metadata = {
-      "bug_tracker_uri" => "https://github.com/rubygems/bundler/issues",
-      "changelog_uri" => "https://github.com/rubygems/bundler/blob/master/CHANGELOG.md",
+      "bug_tracker_uri" => "https://github.com/rubygems/rubygems/issues",
```
olleolleolle, line 27: Is it possible to use query string parameters to point out Issues with the Bundler Label?

olleolleolle:
PS: this is super! Thanks for the precision!

bundler/README.md
sonicdoe: I’ve opened https://github.com/rubygems/rubygems/pull/3617 to remove the Inch CI badge.

I’d rather simply update the link in this pull request because keeping the old one could be confusing as it would always show an outdated evaluation. However, if you prefer, I can also omit this change.

deivid-rodriguez: Ok, fair enough!

bundler/bundler.gemspec
sonicdoe: We could, but similar to https://github.com/rubygems/rubygems/pull/3606#discussion_r421697145, I’m not sure whether this would be better. If you prefer, though, we can update the link to [include a filter](https://github.com/rubygems/rubygems/issues?q=is%3Aopen+is%3Aissue+label%3ABundler).

olleolleolle: Right, yep, _that_'s better.

sonicdoe: Should we also change other instances to filter by the “Bundler” label or should we just update this `bug_tracker_uri`?

olleolleolle: I think the filter is pretty smart for this. So, I'd use it for all "used to be Bundler Issues" places..

sonicdoe: Changed in 96dc57de0069ab82b47126c8c9729eed1db383fe. Analogously, there are some existing links which should now filter by the “RubyGems” label but I’d rather do that in a separate pull request as to focus on updating [rubygems/bundler](https://github.com/rubygems/bundler) links in this one.

deivid-rodriguez:
Thanks so much!

PR #3611, opened 2020-05-08
3.1 backport - Remove failing ubuntu-rvm CI flow.

simi:
This one is failing on 3.1 branch and it was removed on master CI actually.


simi:
hmm, unfortunately this doesn't make CI green :'(

deivid-rodriguez:
The other failures being hit here sound like https://github.com/rubygems/rubygems/pull/3076, so you could try backporting that.

simi:
@hsbt I did my best to bring back 3.1 CI. If are going to merge this, I can rebase all my 3.1 backporting cases to check they do not break anything before merge.

Manifest.txt
```diff
@@ -158,8 +158,10 @@ bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
 bundler/lib/bundler/templates/newgem/rspec.tt
 bundler/lib/bundler/templates/newgem/spec/newgem_spec.rb.tt
 bundler/lib/bundler/templates/newgem/spec/spec_helper.rb.tt
-bundler/lib/bundler/templates/newgem/test/newgem_test.rb.tt
-bundler/lib/bundler/templates/newgem/test/test_helper.rb.tt
+bundler/lib/bundler/templates/newgem/test/minitest/newgem_test.rb.tt
```
hsbt, line 161: L161-L164 are accidentally additions?

hsbt: I got it. But We shouldn't bundle without Bundler 2.1 branch.

deivid-rodriguez: I don't think so, but there's maybe an issue here.

@simi bumped the vendored version of bundler to fix a testing issue, but after a second thought, we don't want to do that because that would mean including an undefined bundler version together with rubygems.

So, I'm not really sure, perhaps we can skip the bundler test suite on the stable branch.

deivid-rodriguez: Or update the submodule to the proper ref dinamically during CI, not in source control.

simi: I can also prepare patch to apply on CI only to vendored bundler .

deivid-rodriguez: Yeah, that too. Although @hsbt doesn't like `.patch` files :sweat_smile:

hsbt: Or remove `head` from CI of RG 3.1 branch. RG 3.1 support under the Ruby 2.7, not 2.8/3.0.

deivid-rodriguez: I think it was not head failing, but the regular suite against the vendored bundler. But I agree with removing head anyways.

deivid-rodriguez: @simi Are you still planning to work on this?

simi: yup, if this is the way for new 3.1 release, I'll try to make this green

deivid-rodriguez: I'm not sure how @hsbt wants to handle it, but having a green 3.1 stable branch feels good anyways :man_shrugging:

simi:
OK, I did some progress, removed ruby head related testing and also I have removed bundler for now (since it is not vendored yet and in theory it has own CI). Is this acceptable?

deivid-rodriguez:
I would rather checkout the commit in the submodule that fixes the issue. Same thing as you did in https://github.com/rubygems/rubygems/pull/3611/commits/bb5e77c80c8f1ac6135e94486bc2a48bd3f982f6, but checkout the reference inside the CI script instead of commting it.

simi:
That actually doesn't make sense to me.

Original bundle reference is to https://github.com/rubygems/bundler/commit/971d39 which is still having own CI (which failed for that commit). IMHO it would make sense to skip bundler CI at all and move to actual reference having green build on its own.

deivid-rodriguez:
The point of testing bundler from rubygems is (I believe) to make sure that the specific version of bundler that `gem update --system` will install works properly against the rubygems version you are upgrading to.

We've been doing this in both master and stable branches for a long time, so in my opinion changing this would grant a separate discussion. Here your intention is to fix CI, not to change the way we test the libraries.

simi:
I'm really lost here and I'm going to leave this. Feel free to reuse all work done in here. Once 3.1 will have green CI, I'll rebase my other pull request to make merge smoother and I'll :pray: for new 3.1 release removing that annoying rubyforge warning. That's my whole motivation.

deivid-rodriguez:
To try to further clarify what I mean, the CI you linked to run against several versions of rubygems, but not against the _current_ 3.1 stable branch of rubygems. That would be the point of this workflow.

Another alternative is to make the bundler CI an allowed failure. Since commits to 3.1 stable land very unfrequently, I think it's fine to manually check the CI when you're about to release. In this case, there's no real issues with bundler, it's just a "testing issue".

deivid-rodriguez:
Sure, I just wanted try to explain my opinion better. Thanks for working on this and trying to make a new release easier and thus speed it up :heart:.

deivid-rodriguez:
I set bundler as an allowed failure. Will now cleanup the messy git history and leave it for @hsbt to evaluate and merge :+1:.

simi:
ok, thanks for help :pray:

deivid-rodriguez:
@hsbt If you're too busy for a release right now, I'm happy to do it myself.

In any case, please point me to a release playbook so that more than one person has the knowledge to do a release.

hsbt:
I can release it in this week.

PR #3613, opened 2020-05-09
Make sure `install_if` specs are run and pass

deivid-rodriguez:
# Description:

As discovered and fixed by @hsbt at https://github.com/rubygems/rubygems/pull/3612, this test file was unrunnable.

We hadn't noticed because it was not being run by default due to its unconventional naming.

This PR fixes the name so that the file is run by default by RSpec.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/spec/install/gemfile/install_if_spec.rb
+++ b/bundler/spec/install/gemfile/install_if_spec.rb
@@ -29,7 +29,7 @@ RSpec.describe "bundle install with install_if conditionals" do
             rack
 
       PLATFORMS
-        ruby
+        #{lockfile_platforms}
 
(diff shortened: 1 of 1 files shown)


simi:
I did a quick check and there's no other file with same problem. :+1:

deivid-rodriguez:
They are actually failing on bundler 3 and Windows. Not too surprising since this test was actually never run by default since introduced at https://github.com/rubygems/bundler/pull/3611.

I'll eventually work on this. Anybody feel free to pick it up.

deivid-rodriguez:
Actually, I think I might've fixed it.

Issue #3570, opened 2020-04-29
LoadError: cannot load such file -- bundler/gem_tasks, incorrect entry for Bundler in $LOAD_PATH


eregon:
This issue happens when using Bundler 2.1.4 on Ruby 2.3.8.
It doesn't happen on Ruby 2.4+, nor with Bundler 1 on Ruby 2.3.8.

See https://github.com/ruby/setup-ruby/issues/51

It can be reproduced easily with:
```
git clone https://github.com/anirbanmu/str_metrics.git 
cd str_metrics
git checkout 7e4645e3919ca85a7eac87146ba6797271bb3dbd

chruby 2.3.8
gem install bundler
bundle config --local path vendor/bundle
bundle install
bundle exec rake spec
```

That gives:
```
rake aborted!
LoadError: cannot load such file -- bundler/gem_tasks
/home/eregon/tmp/str_metrics/Rakefile:7:in `require'
/home/eregon/tmp/str_metrics/Rakefile:7:in `<top (required)>'
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/home/eregon/.rubies/ruby-2.3.8/bin/bundle:22:in `load'
/home/eregon/.rubies/ruby-2.3.8/bin/bundle:22:in `<main>'
(See full trace by running task with --trace)
zsh: exit 1     bundle exec rake spec
```

And the reason is that `$LOAD_PATH` is incorrect inside the Rakefile, adding `puts $LOAD_PATH` as the first line of the Rakefile gives:
```
/home/eregon/tmp/str_metrics/lib
/home/eregon/tmp/extensions/x86_64-linux/2.3.0-static/str_metrics-0.1.1
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/rubocop-rspec-1.38.1/lib
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/rubocop-performance-1.5.2/lib
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/rubocop-0.81.0/lib
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/unicode-display_width-1.7.0/lib
/home/eregon/tmp/str_metrics/vendor/bundle/ruby/2.3.0/gems/ruby-progressbar-1.10.1/lib
[...]

PR #3614, opened 2020-05-09
Run specs against an installed bundler

deivid-rodriguez:
# Description:

This PR is an attempt to run specs against an installed version of bundler, not directly against the sources.

This is good because it resembles real usage more closely and that should make it easier to reproduce realworld issues.

For starters, it revealed one bug, which is that I thought that `bundler` supported the `uri` gem inside `gems.rb` files, but it doesn't. There's a test for this which previously passed, but with the new setup it was revealed that `uri` is actually loaded by rubygems during the `active_bin_path` call inside the `bundler` binstub. So `bundler` can't respect the user-choice.

Also, with this PR, https://github.com/rubygems/rubygems/issues/3570 might reproduce in CI on ruby 2.3, we'll see.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This one is ready now and includes a fix for #3570.

But I'll extract the fix to a separate PR and leave the refactoring of the specs to run installed bundler's for this PR.

PR #3615, opened 2020-05-09
Bump compact_index to 0.12.1

deivid-rodriguez:
# Description:

Just upgrading our tests to use the new `compact_index` release.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
@sonalkr132 This one should be ok now. I updated bundler to pass strings to the `compact_index` gem. I considered restoring support for `Gem::Requirement`'s as in the previous version, but I believe it was just working as an accident, since for example the `dependencies` field in `CompactIndex::Version` doesn't support `Gem::Requirement`'s either.

PR #3622, opened 2020-05-13
Bump vendored libraries

deivid-rodriguez:
# Description:

Just a review pass of our vendored code.

We're currently vendoring `uri`'s master branch because the [uri gem has no associated tag](https://github.com/ruby/uri/issues/7) and automatiek needs that.

For now I upgraded to latest master, but it's probably best to create the git tag, and revert to 0.10.0.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -213,6 +213,10 @@ else
     lib.vendor_lib = "lib/bundler/vendor/fileutils"
   end
 
+  # We currently include changes over the official version to avoid requiring
+  # the optional `net-http-pipeline` dependency, so that its version can be
+  # selected by end users. We also include changes to require the vendored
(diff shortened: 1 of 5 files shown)


eregon:
https://github.com/rubygems/rubygems/pull/3622/checks?check_run_id=674481111 seems a transient failure, I'll make a PR to fix.

deivid-rodriguez:
@hsbt tagged `uri-v0.10.0` so I went back to that. The changes are minimal now.

The situation with `automatiek` is quite inconvenient at the moment by the way, since there's [some unreleased changes](https://github.com/segiddins/automatiek/issues/10) that are needed to properly work with the vendoring `rake` tasks.

Given that the gem is essentially one ruby file, what do you think about moving the code here and abandoning it?

PR #3624, opened 2020-05-13
Backport the ruby core changes for Bundler 2.2.0dev merge

hsbt:
# Description:

I did merge the current master branch into ruby core repository. We need to add the some changes for the following changes:

* Use relative path for the hard-coded path like `lib` and `gemspec`
* Mask the some examples used `ruby_repo` label.

The first failing case was caused by `bundle 'exec foo'` changes. I'm not sure why these examples was failed. It failed with `rake` and `rackup`. 

The second one is `FileUtils.rm_rf ".bundle"` in before block. In the ruby core repository located a test dependencies like rspec into `.bundle` directory. It was passed with Bundler 2.1 with Ruby 2.7. I'm also not sure why it was failed with the current master branch.

I'm happy to remove ruby_repo label before releasing Bundler 2.2.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -336,7 +336,7 @@ module Spec
 
         replace_build_metadata(build_metadata, dir: build_path) # rubocop:disable Style/HashSyntax
 
-        gem_command! "build bundler.gemspec", :dir => build_path
+        gem_command! "build #{relative_gemspec}", :dir => build_path
 
         yield(bundler_path)
       ensure
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -576,5 +576,11 @@ module Spec
       end
       port
     end
+
+  private
+
+    def git_root_dir?
+      root.to_s == `git rev-parse --show-toplevel`.chomp
+    end
   end
 end
--- a/bundler/spec/commands/clean_spec.rb
+++ b/bundler/spec/commands/clean_spec.rb
@@ -765,7 +765,7 @@ RSpec.describe "bundle clean" do
     should_not_have_gems "foo-1.0"
   end
 
-  it "doesn't remove extensions artifacts from bundled git gems after clean", :ruby_repo do
+  it "doesn't remove extensions artifacts from bundled git gems after clean" do
     build_git "very_simple_git_binary", &:add_c_extension
 
     revision = revision_for(lib_path("very_simple_git_binary-1.0"))
@@ -788,7 +788,7 @@ RSpec.describe "bundle clean" do
     expect(vendored_gems("bundler/gems/very_simple_git_binary-1.0-#{revision[0..11]}")).to exist
   end
 
-  it "removes extension directories", :ruby_repo do
+  it "removes extension directories" do
     gemfile <<-G
       source "#{file_uri_for(gem_repo1)}"
 
@@ -824,7 +824,7 @@ RSpec.describe "bundle clean" do
     expect(simple_binary_extensions_dir).to exist
   end
 
-  it "removes git extension directories", :ruby_repo do
+  it "removes git extension directories" do
     build_git "very_simple_git_binary", &:add_c_extension
 
     revision = revision_for(lib_path("very_simple_git_binary-1.0"))
@@ -864,7 +864,7 @@ RSpec.describe "bundle clean" do
     expect(very_simple_binary_extensions_dir).not_to exist
   end
 
-  it "keeps git extension directories when excluded by group", :ruby_repo do
+  it "keeps git extension directories when excluded by group" do
     build_git "very_simple_git_binary", &:add_c_extension
 
     revision = revision_for(lib_path("very_simple_git_binary-1.0"))
--- a/bundler/spec/commands/clean_spec.rb
+++ b/bundler/spec/commands/clean_spec.rb
@@ -765,7 +765,7 @@ RSpec.describe "bundle clean" do
     should_not_have_gems "foo-1.0"
   end
 
-  it "doesn't remove extensions artifacts from bundled git gems after clean" do
+  it "doesn't remove extensions artifacts from bundled git gems after clean", :ruby_repo do
     build_git "very_simple_git_binary", &:add_c_extension
 
     revision = revision_for(lib_path("very_simple_git_binary-1.0"))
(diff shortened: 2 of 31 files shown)


bundler/spec/support/helpers.rb
```diff
@@ -330,7 +330,7 @@ def with_built_bundler(version = nil)
 
         replace_version_file(version, dir: build_path) # rubocop:disable Style/HashSyntax
 
-        gem_command! "build bundler.gemspec", :dir => build_path
+        gem_command! "build #{shipped_gemspec}", :dir => build_path
```
deivid-rodriguez, line 333: This doesn't make sense to me. We must be in the same folder as the gemspec in order to build the gem correctly, and we have just copied the file in there. We want the gemspec to be to copied one (where we have changed the version), not the one from the sources.

deivid-rodriguez:
> The second one is FileUtils.rm_rf ".bundle" in before block. 

This one is a mistake in our specs, a leftover from when we stopped `chdir`'ing to the dummy app folder for every spec. The correct code should be `FileUtils.rm_rf bundled_app(".bundle")`.

bundler/spec/support/helpers.rb
hsbt: I understood. But It's hard to rewite the filelist on only path.rb. Can I add the `ruby_core?` condition in helpers.rb?

deivid-rodriguez: I don't understand what you mean. The rewritten file list comes from `shipped_files` in `path.rb`  so it should be correct for the core repo, shouldn't it?

deivid-rodriguez:
Let me run the ruby-core specs tomorrow and I'll figure out why some specs are no longer working.

bundler/spec/support/helpers.rb
hsbt: gemspec in ruby repo was located in `lib/bundler/bundler.gemspec`, not `bundler.gemspec` in `shipped_files`. so, build_path + "/bundler.gemspec" was failed with `bundle build bundler.gemspec`.

We should move it to the top of `build_path`

hsbt: Like this https://github.com/ruby/ruby/pull/3105/files

deivid-rodriguez: I understand now. The patch you linked to makes sense :+1:.

deivid-rodriguez:
I'm going to update ruby-core and figure out the failures now :+1:.

deivid-rodriguez:
If figured out the issue with `bundle exec`. Will update this PR.

deivid-rodriguez:
@hsbt I pushed some fixes and now all ruby-core specs pass on my machine (except for some exeptions about default gem activation). I still left the `:ruby_repo` tag in case you find some issues when integrating this and need to re-skip some specs.

hsbt:
I try to run this branch on ruby repo at https://github.com/ruby/ruby/pull/3114

deivid-rodriguez:
It also requires some changes on the ruby-core side, in particular, fixing the glob in the gemspec for the man pages, like it's done in this PR.

deivid-rodriguez:
Like this:

```diff
diff --git a/lib/bundler/bundler.gemspec b/lib/bundler/bundler.gemspec
index 1ef46c8e6e..d29f800c3a 100644
--- a/lib/bundler/bundler.gemspec
+++ b/lib/bundler/bundler.gemspec
@@ -34,7 +34,7 @@
   s.required_ruby_version     = ">= 2.3.0"
   s.required_rubygems_version = ">= 2.5.2"
 
-  s.files = (Dir.glob("lib/bundler/**/*", File::FNM_DOTMATCH) + Dir.glob("man/bundler*") + Dir.glob("libexec/bundle*")).reject {|f| File.directory?(f) }
+  s.files = (Dir.glob("lib/bundler/**/*", File::FNM_DOTMATCH) + Dir.glob("man/bundle*") + Dir.glob("man/gemfile*") + Dir.glob("libexec/bundle*")).reject {|f| File.directory?(f) }
 
   s.files += ["lib/bundler.rb"]
 
```

hsbt:
@deivid-rodriguez Thanks, But the current HEAD commit was failing with build the native extension like this .

>make[1]: *** No rule to make target '/include/x86_64-linux/ruby/config.h', needed by 'very_simple_git_binary.o'.  Stop.

see https://github.com/ruby/ruby/pull/3114/checks?check_run_id=679939703

deivid-rodriguez:
Sorry, all tests pass for me locally under ruby-core :(

Is there an actual regression here or all the failures are previous skips? If they are, let's keep them skipped.

hsbt:
It's hard things 😇 

In my investigation. This failures caused by 2 cases:

1. The harf of failures are previous skips that are couldn't find the executables like `rackup`
2. Another failures are build error with native extensions.

deivid-rodriguez:
What I mean is: if we readd all the `:ruby_core` tags that we previously had, will all tests pass, or do we have new failures?

hsbt:
> or do we have new failures?

That is. the native extension failures affected by https://github.com/rubygems/rubygems/pull/3614/

deivid-rodriguez:
Ah, ok. I will check those then.

hsbt:
I could reproduce it with `make test-bundler BUNDLER_SPECS=commands/pristine_spec.rb` on https://github.com/ruby/ruby/pull/3114

deivid-rodriguez:
But all of pristine specs were previously skipped in ruby_core (they had the `:ruby_repo`) tag. Not saying it's not an issue, just saying it's not new.

hsbt:
Ah, I confirmed. I misunderstand you mean skipped examples only executables issues.

hsbt:
I'm trying with https://github.com/ruby/ruby/pull/3114/commits/60ccdc43781eaafc101f32e28c8dd6f9c29adda3

hsbt:
My macbook has a swollen battery issue. So, it's currently under repair status. I will re-work this PR after my macbook back to my home.

hsbt:
@deivid-rodriguez Finally, the current HEAD commit was passed with the test suite of ruby core repository. Do you have an additional concerns?

deivid-rodriguez:
No, go ahead!

hsbt:
Thanks!

PR #3630, opened 2020-05-15
Increase timeouts in test_gem_stream_ui.rb

eregon:
* 0.1s is really short and fails in CI: #3622

cc @deivid-rodriguez


deivid-rodriguez:
Thanks!

Issue #3365, opened 2018-05-14
`SystemStackError` after upgrade to bundler 1.16 


nelhage-stripe:
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/usr/lib/rbenv/versions/2.4/bin/bundle install`

- **What did you expect to happen?**

`bundle install` would complete successfully.

- **What happened instead?**

We (Stripe) maintain a large Ruby application internally. After upgrading to bundler 1.16.1 from 1.15.4 we see reproducible `SystemStackError`s from infinite recursion in our CI environment (but only in CI; I have not been able to reproduce locally yet, so it depends on some combination of the environment and our CI settings).

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

I have searched but haven't found anything obviously similar.

I'm able to (e.g.) bisect or capture a CI build and drop in a `pry`, but because the issue only reproduces in CI so far doing so is slightly involved, so I've opened an issue first in the hopes that you'll be able to point us at productive things to try.

I've redacted our `Gemfile` and `Gemfile.lock` since they are large and contain various internal references; I'd be happy to provide them to the `bundler` team privately if useful.

- **Have you read our issues document, https://github.com/bundler/bundler/blob/master/doc/contributing/ISSUES.md?**

 Yes.

## Backtrace

```
SystemStackError: stack level too deep
  /usr/lib/rbenv/versions/2.4/lib/ruby/2.4.0/rubygems/stub_specification.rb:172:in `extensions'
  /usr/lib/rbenv/versions/2.4/lib/ruby/2.4.0/rubygems/stub_specification.rb:148:in `missing_extensions?'
  [ following two lines repeat many times ]
  /usr/lib/rbenv/versions/2.4/lib/ruby/2.4.0/rubygems/stub_specification.rb:151:in `missing_extensions?'
  /usr/lib/rbenv/versions/2.4/lib/ruby/gems/2.4.0/gems/bundler-1.16.1/lib/bundler/stub_specification.rb:38:in `missing_extensions?'
  /usr/lib/rbenv/versions/2.4/lib/ruby/2.4.0/rubygems/stub_specification.rb:151:in `missing_extensions?'
  /usr/lib/rbenv/versions/2.4/lib/ruby/gems/2.4.0/gems/bundler-1.16.1/lib/bundler/stub_specification.rb:38:in `missing_extensions?'
[...]

PR #3635, opened 2020-05-17
Avoid stack overflow inside `StubSpecification` on some edge cases

deivid-rodriguez:
# Description:

This is a port of https://github.com/rubygems/bundler/pull/6653 from the old repo, which closes https://github.com/rubygems/rubygems/issues/3365.

# Tasks:

- [x] Write tests


deivid-rodriguez:
I ended up changing the strategy from the original PR to make sure that we don't load the full specification even in the edge case of the referenced issue. Also added some other improvements while investigating the issue. I believe it's ready.

Issue #3629, opened 2020-05-15
Gem.load_yaml: psych: `<class:Parser>': superclass mismatch for class Mark (TypeError)


junaruga:
I'm having a problem.

My current problem is the following Ruby script shows the following `TypeError` with non-default gem: psych.

```
$ cat test.rb 
require 'psych'
Gem.load_yaml # <= The error happens.
```

```
$ ruby test.rb
/usr/lib64/gems/ruby/psych-3.1.0/psych.so: warning: already initialized constant Psych::Parser::ANY
...
/usr/share/gems/gems/psych-3.1.0/lib/psych/parser.rb:34:in `<class:Parser>': superclass mismatch for class Mark (TypeError)
```

I prepared the reproducer script here with Travis CI. The 3 cases that have the issue.

https://github.com/junaruga/report-rubygems-load-yaml
https://travis-ci.org/github/junaruga/report-rubygems-load-yaml

The issue comes from [this ticket in Fedora project](https://bugzilla.redhat.com/show_bug.cgi?id=1835836). And it is related to https://github.com/rubygems/rubygems/issues/3131 and https://github.com/ruby/psych/issues/386 .
I want to see that `Gem.load_yaml` will be fixed for now, for example by removing `gem 'psych', '>= 2.0.0'` part that I mentioned [here](https://bugzilla.redhat.com/show_bug.cgi?id=1835836#c5).

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:

```
$ gem env version
3.1.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

deivid-rodriguez:
Does https://github.com/rubygems/rubygems/pull/3429 fix the issue?

junaruga:
Thanks for the info. Let me check it.

junaruga:
I will check it on Monday 18th May.

junaruga:
> Does #3429 fix the issue?

It does not work on Fedora rawhide. It's a same error with the case without the patch.
I confirmed `test_activate_via_require_respects_loaded_default_from_default_gems` passes.

[...]

PR #3638, opened 2020-05-19
bundler: Remove explicit `psych` activation.

junaruga:
# Description:

On the commit: https://github.com/rubygems/rubygems/commit/1ccf0912a161d20e0c4a7b139fd76e8739a411ba , we removed `gem 'psych'` line to fix the similar error at https://github.com/rubygems/rubygems/issues/3629 .

There is one more part to remove in bundler code. And it's not harmful to remove it. See https://github.com/rubygems/rubygems/issues/3629#issuecomment-630697387 .

In the Fedora RPM Ruby environment where there is no default gem psych but there is regular gem psych, the following error happens.

```
$ cat test3.rb 
require 'psych'
require 'bundler/psyched_yaml' # <= The error happens.
```

```
$ ruby test3.rb
/usr/lib64/gems/ruby/psych-3.1.0/psych.so: warning: already initialized constant Psych::Parser::ANY
...
/usr/share/gems/gems/psych-3.1.0/lib/psych/parser.rb:34:in `<class:Parser>': superclass mismatch for class Mark (TypeError)
```

Remove `gem 'psych'` line causing the error. It's same with https://github.com/rubygems/rubygems/pull/3636 .

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/psyched_yaml.rb
+++ b/bundler/lib/bundler/psyched_yaml.rb
@@ -1,11 +1,5 @@
 # frozen_string_literal: true
 
-# Psych could be a gem, so try to ask for it
-begin
-  gem "psych"
-rescue LoadError
-end if defined?(gem)
-
 # Psych could be in the stdlib
 # but it's too late if Syck is already loaded
 begin


hsbt:
👍

e2002e:
Hi @junaruga, I'm having an issue which looks similar when trying to run ruby gem on arm.
I did cross compile the version 2.7.4 and all dependencies are satisfied, but upon loading psych.so in rubygems.rb line 701:
```
begin
      # Try requiring the gem version *or* stdlib version of psych.
      require 'psych'
rescue ::LoadError
```
I have a superclass mismatch for Psych::SyntaxError which should be Psych::Exception
I erased the line you mentionned "gem 'psych', '>= 2.0.0'" but it makes no differences.
Can you clarify what is happening here and how to fix it ? Thanks.

Issue #1866, opened 2017-03-08
-I should take precedence over default gems


segiddins:
I'm having a problem or would like to suggest a feature.

My current problem is that the following test fails by activating `c`

```diff
diff --git a/test/rubygems/test_require.rb b/test/rubygems/test_require.rb
index dd606e44..63bc3461 100644
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -80,6 +80,37 @@ class TestGemRequire < Gem::TestCase
     Object.send :remove_const, :HELLO if Object.const_defined? :HELLO
   end
 
+  # Providing -I on the commandline should always beat gems
+  def test_dash_i_beats_default_gems
+    a1 = new_default_spec "a", "1", {"b" => "= 1"}, "test_gem_require_a.rb"
+    b1 = new_default_spec "b", "1", {"c" => "> 0"}, "b/c.rb"
+    c1 = new_default_spec "c", "1", nil, "c/c.rb"
+    c2 = new_default_spec "c", "2", nil, "c/c.rb"
+
+    install_default_specs c1, c2, b1, a1
+
+    dir = Dir.mktmpdir("test_require", @tempdir)
+    dash_i_arg = File.join dir, 'lib'
+
+    c_rb = File.join dash_i_arg, 'b', 'c.rb'
+
+    FileUtils.mkdir_p File.dirname c_rb
+    File.open(c_rb, 'w') { |f| f.write "class Object; HELLO = 'world' end" }
+
+    lp = $LOAD_PATH.dup
+
+    # Pretend to provide a commandline argument that overrides a file in gem b
+    $LOAD_PATH.unshift dash_i_arg
+
+    assert_require 'test_gem_require_a'
+    assert_require 'b/c' # this should be required from -I
+    assert_equal "world", ::Object::HELLO
+    assert_equal %w(a-1), loaded_spec_names
+  ensure
+    $LOAD_PATH.replace lp
+    Object.send :remove_const, :HELLO if Object.const_defined? :HELLO
+  end
+
   def test_concurrent_require
     skip 'deadlock' if /^1\.8\./ =~ RUBY_VERSION
 
```

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

Here are my current environment details:

```
$ gem env version
2.6.10
```

[...]

PR #3639, opened 2020-05-19
Fix correctness and performance regression in `require`

### LINKED ISSUE #3491: Edge case difference in behavior from the standard `Kernel.require`
There's a slight difference in behavior between our extended `require` and the standard `require` provided by ruby by default. Sometimes, `$LOADED_FEATURES` cache is not respected when `$LOAD_PATH` priority has changed.

The issue can be reproduced as follows:

```
$ ruby -v
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]

$ gem list reline

*** LOCAL GEMS ***

reline (0.1.3, default: 0.1.2)

$ ruby --disable-gems -e 'require "reline"; puts Reline::VERSION; $LOAD_PATH.unshift "/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib"; puts require("reline")'
0.1.2
false

$ ruby -e 'require "reline"; puts Reline::VERSION; $LOAD_PATH.unshift "/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0"; puts require("reline")'
0.1.3
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/version.rb:2: warning: already initialized constant Reline::VERSION
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/version.rb:2: warning: previous definition of VERSION was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/config.rb:6: warning: already initialized constant Reline::Config::DEFAULT_PATH
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/config.rb:6: warning: previous definition of DEFAULT_PATH was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/config.rb:8: warning: already initialized constant Reline::Config::KEYSEQ_PATTERN
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/config.rb:8: warning: previous definition of KEYSEQ_PATTERN was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/config.rb:14: warning: already initialized constant Reline::Config::VARIABLE_NAMES
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/config.rb:14: warning: previous definition of VARIABLE_NAMES was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/config.rb:41: warning: already initialized constant Reline::Config::VARIABLE_NAME_SYMBOLS
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/config.rb:41: warning: previous definition of VARIABLE_NAME_SYMBOLS was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/key_actor/base.rb:2: warning: already initialized constant Reline::KeyActor::Base::MAPPING
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/key_actor/base.rb:2: warning: previous definition of MAPPING was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/key_actor/emacs.rb:2: warning: already initialized constant Reline::KeyActor::Emacs::MAPPING
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/key_actor/emacs.rb:2: warning: previous definition of MAPPING was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/key_actor/vi_command.rb:2: warning: already initialized constant Reline::KeyActor::ViCommand::MAPPING
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/key_actor/vi_command.rb:2: warning: previous definition of MAPPING was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/key_actor/vi_insert.rb:2: warning: already initialized constant Reline::KeyActor::ViInsert::MAPPING
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/key_actor/vi_insert.rb:2: warning: previous definition of MAPPING was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/kill_ring.rb:3: warning: already initialized constant Reline::KillRing::State::FRESH
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/kill_ring.rb:3: warning: previous definition of FRESH was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/kill_ring.rb:4: warning: already initialized constant Reline::KillRing::State::CONTINUED
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/kill_ring.rb:4: warning: previous definition of CONTINUED was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/kill_ring.rb:5: warning: already initialized constant Reline::KillRing::State::PROCESSED
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/kill_ring.rb:5: warning: previous definition of PROCESSED was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/kill_ring.rb:6: warning: already initialized constant Reline::KillRing::State::YANK
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/kill_ring.rb:6: warning: previous definition of YANK was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/kill_ring.rb:9: warning: already initialized constant Reline::KillRing::RingPoint
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/kill_ring.rb:9: warning: previous definition of RingPoint was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode.rb:2: warning: already initialized constant Reline::Unicode::EscapedPairs
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode.rb:2: warning: previous definition of EscapedPairs was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode.rb:36: warning: already initialized constant Reline::Unicode::EscapedChars
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode.rb:36: warning: previous definition of EscapedChars was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:6: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_F
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:6: warning: previous definition of TYPE_F was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:13: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_H
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:13: warning: previous definition of TYPE_H was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:24: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_W
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:24: warning: previous definition of TYPE_W was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:139: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_NA
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:139: warning: previous definition of TYPE_NA was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:150: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_A
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:150: warning: previous definition of TYPE_A was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/unicode/east_asian_width.rb:333: warning: already initialized constant Reline::Unicode::EastAsianWidth::TYPE_N
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/unicode/east_asian_width.rb:333: warning: previous definition of TYPE_N was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:21: warning: already initialized constant Reline::LineEditor::VI_MOTIONS
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:21: warning: previous definition of VI_MOTIONS was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:43: warning: already initialized constant Reline::LineEditor::CompletionState::NORMAL
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:43: warning: previous definition of NORMAL was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:44: warning: already initialized constant Reline::LineEditor::CompletionState::COMPLETION
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:44: warning: previous definition of COMPLETION was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:45: warning: already initialized constant Reline::LineEditor::CompletionState::MENU
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:45: warning: previous definition of MENU was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:46: warning: already initialized constant Reline::LineEditor::CompletionState::JOURNEY
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:46: warning: previous definition of JOURNEY was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:47: warning: already initialized constant Reline::LineEditor::CompletionState::MENU_WITH_PERFECT_MATCH
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:47: warning: previous definition of MENU_WITH_PERFECT_MATCH was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:48: warning: already initialized constant Reline::LineEditor::CompletionState::PERFECT_MATCH
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:48: warning: previous definition of PERFECT_MATCH was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:51: warning: redefining constant Struct::CompletionJourneyData
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:51: warning: already initialized constant Reline::LineEditor::CompletionJourneyData
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:51: warning: previous definition of CompletionJourneyData was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:52: warning: redefining constant Struct::MenuInfo
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:52: warning: already initialized constant Reline::LineEditor::MenuInfo
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:52: warning: previous definition of MenuInfo was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:54: warning: already initialized constant Reline::LineEditor::CSI_REGEXP
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:54: warning: previous definition of CSI_REGEXP was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:55: warning: already initialized constant Reline::LineEditor::OSC_REGEXP
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:55: warning: previous definition of OSC_REGEXP was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:56: warning: already initialized constant Reline::LineEditor::NON_PRINTING_START
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:56: warning: previous definition of NON_PRINTING_START was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:57: warning: already initialized constant Reline::LineEditor::NON_PRINTING_END
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:57: warning: previous definition of NON_PRINTING_END was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/line_editor.rb:58: warning: already initialized constant Reline::LineEditor::WIDTH_SCANNER
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/line_editor.rb:58: warning: previous definition of WIDTH_SCANNER was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:12: warning: already initialized constant Reline::FILENAME_COMPLETION_PROC
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:12: warning: previous definition of FILENAME_COMPLETION_PROC was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:13: warning: already initialized constant Reline::USERNAME_COMPLETION_PROC
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:13: warning: previous definition of USERNAME_COMPLETION_PROC was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:15: warning: redefining constant Struct::Key
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:15: warning: already initialized constant Reline::Key
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:15: warning: previous definition of Key was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:16: warning: already initialized constant Reline::CursorPos
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:16: warning: previous definition of CursorPos was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:19: warning: already initialized constant Reline::Core::ATTR_READER_NAMES
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:19: warning: previous definition of ATTR_READER_NAMES was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:409: warning: already initialized constant Reline::HISTORY
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:430: warning: previous definition of HISTORY was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/ansi.rb:4: warning: already initialized constant Reline::ANSI::RAW_KEYSTROKE_CONFIG
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/ansi.rb:12: warning: previous definition of RAW_KEYSTROKE_CONFIG was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline.rb:423: warning: already initialized constant Reline::IOGate
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline.rb:428: warning: previous definition of IOGate was here
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/2.7.0/reline/general_io.rb:4: warning: already initialized constant Reline::GeneralIO::RAW_KEYSTROKE_CONFIG
/home/deivid/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/reline-0.1.3/lib/reline/general_io.rb:12: warning: previous definition of RAW_KEYSTROKE_CONFIG was here
true
```

Ruby's require will respect a previously required file, not matter what changes are made to the `$LOAD_PATH`. Our monkeypatched require will not.

This issue was originally reported by @voxik at #3131 but since there are two separate bugs in there, I'm splitting the second one into its own ticket.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

### LINKED ISSUE #3631: Rubygems 3.1.x performance regression due to #1868
#1868 introduced code to scan through the $LOAD_PATH to test for the presence of files in the -I args of the ruby cli.

What I think was missed there is that the the $LOAD_PATH includes the lib directories of all of the activated gems:

e.g. from some debugging output of `puts $LOAD_PATH[0...load_path_insert_index]`:

```
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/test-kitchen-2.5.0/bin/../lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/concurrent-ruby-1.1.6/lib/concurrent-ruby
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/i18n-1.8.2/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/thread_safe-0.3.6/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/tzinfo-1.2.7/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/minitest-5.14.0/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/activesupport-5.2.4.2/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/public_suffix-4.0.5/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/addressable-2.7.0/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/mixlib-shellout-3.0.9/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/mixlib-cli-2.1.6/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/appbundler-0.13.2/lib
[...]
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-byebug-3.9.0/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/slop-3.6.0/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-remote-0.1.8/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/pry-stack_explorer-0.4.9.3/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/rexml-3.2.4/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/ruby-prof-1.3.2/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/extensions/x86_64-linux/2.7.0/ruby-prof-1.3.2
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/ruby-shadow-2.5.0/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/extensions/x86_64-linux/2.7.0/ruby-shadow-2.5.0
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/stove-7.1.6/lib
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/yard-0.9.25/lib
```

It looks like what is happening is this is now an extra scan through every single activated gem, times the three suffixes which every require call has to go through in addition to all the processing that was there in rubygems 2.x

The difference is fairly substantial, this is rubygems 3.1.2 performance:

```
# time /opt/chef-workstation/bin/kitchen list
/opt/chef-workstation/bin/kitchen list  13.02s user 5.64s system 49% cpu 37.690 total
```

versus commenting out the code block from that PR (in its current form):

```
# time /opt/chef-workstation/bin/kitchen list
/opt/chef-workstation/bin/kitchen list  2.42s user 0.61s system 48% cpu 6.256 total
```

I think that if the file to be required is an internal gem that it is particularly badly performant because it has to fail this loop first going through all the gems, then it has to go through all the same logic as before until finally falling back to scanning through the internal ruby gems.

One solution might be to delete the now-duplicate second scan of activated gems, but it also seems like this loop is particularly slow for some reason (the 600% speedup after deleting it suggests something else is going on).

Another solution would be to tweak the loop so that it was only looping through the command line -I flags (as I think was intended, but I don't know enough about how rubygems handles the $LOAD_PATH).

Additionally, this is on ruby 2.7.1 so the UNTAINT call should be hitting the empty-proc side of that, so it shouldn't be untainting all the strings.

### LINKED ISSUE #3647: `require` changes behavior comparing to plain Ruby
~~~
$ for psych in psych{1..3}; do mkdir $psych; echo "puts '$psych'" > $psych/psych.rb; done

$ find .
.
./psych2
./psych2/psych.rb
./psych1
./psych1/psych.rb
./psych3
./psych3/psych.rb

$ ruby -e '(1..3).each {|psych| $:.unshift("psych#{psych}"); require "psych" }'
psych1
psych2
psych3

$ ruby --disable-gems -e '(1..3).each {|psych| $:.unshift("psych#{psych}"); require "psych" }'
psych1
~~~

I have mentioned this test first time in \[[1]\] and this might be very well duplicate or resolved by #3131 or #3491. I just think this is nice and clean example which would be worth of its test case, because RubyGems should not modify the behavior of original Ruby `require` method in this way.

[1]: https://bugs.ruby-lang.org/issues/16904#note-8

deivid-rodriguez:
# Description:

A recent fix in rubygems, #1868, caused both a perfomance and a functionality regression in our custom `require`.

On one hand, `require` was made slower because it traverse all activated gems by default.

On the other hand, the unintended manual traversing of those paths also caused divergences in functionality from the standard `require` since the `$LOADED_FEATURES` cache would no longer be hit.

This PR keeps track of the activated gem paths in order to traverse only directories prepended to the `$LOAD_PATH` by ruby `CLI` flags or environment variables.

Fixes #3491.
Fixes #3631.
Fixes #3647.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -594,9 +594,19 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
+  # The number of paths in the `$LOAD_PATH` from activated gems. Used to
+  # prioritize `-I` and `ENV['RUBYLIB`]` entries during `require`.
+
+  def self.activated_gem_paths
+    @activated_gem_paths ||= 0
+  end
+
+  ##
   # Add a list of paths to the $LOAD_PATH at the proper place.
 
   def self.add_to_load_path(*paths)
+    @activated_gem_paths = activated_gem_paths + paths.size
+
     insert_index = load_path_insert_index
 
     if insert_index
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -590,7 +590,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
 
     index = $LOAD_PATH.index RbConfig::CONFIG['sitelibdir']
 
-    index
+    index || 0
   end
 
   ##
@@ -607,15 +607,8 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   def self.add_to_load_path(*paths)
     @activated_gem_paths = activated_gem_paths + paths.size
 
-    insert_index = load_path_insert_index
-
-    if insert_index
-      # gem directories must come after -I and ENV['RUBYLIB']
-      $LOAD_PATH.insert(insert_index, *paths)
-    else
-      # we are probably testing in core, -I and RUBYLIB don't apply
-      $LOAD_PATH.unshift(*paths)
-    end
+    # gem directories must come after -I and ENV['RUBYLIB']
+    $LOAD_PATH.insert(Gem.load_path_insert_index, *paths)
   end
 
   @yaml_loaded = false
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -43,8 +43,9 @@ module Kernel
     # https://github.com/rubygems/rubygems/pull/1868
     resolved_path = begin
       rp = nil
+      load_path_check_index = Gem.load_path_insert_index - Gem.activated_gem_paths
       Gem.suffixes.each do |s|
-        $LOAD_PATH[0...Gem.load_path_insert_index - Gem.activated_gem_paths].each do |lp|
+        $LOAD_PATH[0...load_path_check_index].each do |lp|
           safe_lp = lp.dup.tap(&Gem::UNTAINT)
           begin
             if File.symlink? safe_lp # for backward compatibility
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -47,7 +47,7 @@ module Kernel
         load_path_insert_index = Gem.load_path_insert_index
         break unless load_path_insert_index
 
-        $LOAD_PATH[0...load_path_insert_index].each do |lp|
+        $LOAD_PATH[0...load_path_insert_index - Gem.activated_gem_paths].each do |lp|
           safe_lp = lp.dup.tap(&Gem::UNTAINT)
           begin
             if File.symlink? safe_lp # for backward compatibility
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -41,7 +41,6 @@ module Kernel
 
     if spec = Gem.find_unresolved_default_spec(path)
       # Ensure -I beats a default gem
-      # https://github.com/rubygems/rubygems/pull/1868
       resolved_path = begin
         rp = nil
         load_path_check_index = Gem.load_path_insert_index - Gem.activated_gem_paths
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -39,46 +39,41 @@ module Kernel
 
     path = path.to_path if path.respond_to? :to_path
 
-    # Ensure -I beats a default gem
-    # https://github.com/rubygems/rubygems/pull/1868
-    resolved_path = begin
-      rp = nil
-      load_path_check_index = Gem.load_path_insert_index - Gem.activated_gem_paths
-      Gem.suffixes.each do |s|
-        $LOAD_PATH[0...load_path_check_index].each do |lp|
-          safe_lp = lp.dup.tap(&Gem::UNTAINT)
-          begin
-            if File.symlink? safe_lp # for backward compatibility
-              next
-            end
-          rescue SecurityError
-            RUBYGEMS_ACTIVATION_MONITOR.exit
-            raise
-          end
-
-          full_path = File.expand_path(File.join(safe_lp, "#{path}#{s}"))
-          if File.file?(full_path)
-            rp = full_path
-            break
-          end
-        end
-        break if rp
-      end
-      rp
-    end
-
-    if resolved_path
-      RUBYGEMS_ACTIVATION_MONITOR.exit
-      return gem_original_require(resolved_path)
-    end
-
     if spec = Gem.find_unresolved_default_spec(path)
+      # Ensure -I beats a default gem
+      # https://github.com/rubygems/rubygems/pull/1868
+      resolved_path = begin
+        rp = nil
+        load_path_check_index = Gem.load_path_insert_index - Gem.activated_gem_paths
+        Gem.suffixes.each do |s|
+          $LOAD_PATH[0...load_path_check_index].each do |lp|
+            safe_lp = lp.dup.tap(&Gem::UNTAINT)
+            begin
+              if File.symlink? safe_lp # for backward compatibility
+                next
+              end
+            rescue SecurityError
+              RUBYGEMS_ACTIVATION_MONITOR.exit
+              raise
+            end
+
+            full_path = File.expand_path(File.join(safe_lp, "#{path}#{s}"))
+            if File.file?(full_path)
+              rp = full_path
+              break
+            end
+          end
+          break if rp
+        end
+        rp
+      end
+
       begin
         Kernel.send(:gem, spec.name, Gem::Requirement.default_prerelease)
       rescue Exception
         RUBYGEMS_ACTIVATION_MONITOR.exit
         raise
-      end
+      end unless resolved_path
     end
 
     # If there are no unresolved deps, then we can use just try
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -44,10 +44,7 @@ module Kernel
     resolved_path = begin
       rp = nil
       Gem.suffixes.each do |s|
-        load_path_insert_index = Gem.load_path_insert_index
-        break unless load_path_insert_index
-
-        $LOAD_PATH[0...load_path_insert_index - Gem.activated_gem_paths].each do |lp|
+        $LOAD_PATH[0...Gem.load_path_insert_index - Gem.activated_gem_paths].each do |lp|
           safe_lp = lp.dup.tap(&Gem::UNTAINT)
           begin
             if File.symlink? safe_lp # for backward compatibility
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -384,6 +384,7 @@ class Gem::TestCase < Minitest::Test
     Gem::Security.reset
 
     Gem.loaded_specs.clear
+    Gem.instance_variable_set(:@activated_gem_paths, 0)
     Gem.clear_default_specs
     Bundler.reset!
 
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -45,6 +45,35 @@ class TestGemRequire < Gem::TestCase
     refute require(path), "'#{path}' was not yet required"
   end
 
+  def test_respect_loaded_features_caching_like_standard_require
+    dir = Dir.mktmpdir("test_require", @tempdir)
+
+    lp1 = File.join dir, 'foo1'
+    foo1 = File.join lp1, 'foo.rb'
+
+    FileUtils.mkdir_p lp1
+    File.open(foo1, 'w') { |f| f.write "class Object; HELLO = 'foo1' end" }
+
+    lp = $LOAD_PATH.dup
+
+    $LOAD_PATH.unshift lp1
+    assert_require 'foo'
+    assert_equal "foo1", ::Object::HELLO
+
+    lp2 = File.join dir, 'foo2'
+    foo2 = File.join lp2, 'foo.rb'
+
+    FileUtils.mkdir_p lp2
+    File.open(foo2, 'w') { |f| f.write "class Object; HELLO = 'foo2' end" }
+
+    $LOAD_PATH.unshift lp2
+    refute_require 'foo'
+    assert_equal "foo1", ::Object::HELLO
+  ensure
+    $LOAD_PATH.replace lp
+    Object.send :remove_const, :HELLO if Object.const_defined? :HELLO
+  end
+
   # Providing -I on the commandline should always beat gems
   def test_dash_i_beats_gems
     a1 = util_spec "a", "1", {"b" => "= 1"}, "lib/test_gem_require_a.rb"
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -420,6 +420,17 @@ class TestGemRequire < Gem::TestCase
     assert_equal 0, times_called
   end
 
+  def test_second_gem_require_does_not_resolve_path_manually_before_going_through_standard_require
+    a1 = util_spec "a", "1", nil, "lib/test_gem_require_a.rb"
+    install_gem a1
+
+    assert_require "test_gem_require_a"
+
+    stub(:gem_original_require, ->(path) { assert_equal "test_gem_require_a", path }) do
+      require "test_gem_require_a"
+    end
+  end
+
   def test_realworld_default_gem
     testing_ruby_repo = !ENV["GEM_COMMAND"].nil?
     skip "this test can't work under ruby-core setup" if testing_ruby_repo || java_platform?


lamont-granquist:
Timing tests:

current 3.1 behavior:  /opt/chef-workstation/bin/kitchen list  11.60s user 4.53s system 54% cpu 29.468 total
completely removed behavior:  /opt/chef-workstation/bin/kitchen list  2.58s user 0.82s system 48% cpu 7.046 total
this patch's behavior:  /opt/chef-workstation/bin/kitchen list  2.57s user 0.82s system 48% cpu 7.053 total

So that looks good.

I'm also seeing that it accurate:

Throwing an `-I/foo/bar` in the binstub, I'm only seeing these two being looped over first:

```
/opt/chef-workstation/embedded/lib/ruby/gems/2.7.0/gems/test-kitchen-2.5.1/bin/../lib
/foo/bar
```

The code reads correct to me as well.

lib/rubygems.rb
```diff
@@ -590,22 +590,25 @@ def self.load_path_insert_index
...
   def self.add_to_load_path(*paths)
-    insert_index = load_path_insert_index
+    @activated_gem_paths += paths.size
 
-    if insert_index
-      # gem directories must come after -I and ENV['RUBYLIB']
-      $LOAD_PATH.insert(insert_index, *paths)
-    else
-      # we are probably testing in core, -I and RUBYLIB don't apply
-      $LOAD_PATH.unshift(*paths)
```
voxik, line 607: Honestly, the `$LOAD_PATH.unshift(*paths)` should be the only logic which should be used here, because this is wrong:

~~~
$ ruby -Ifoo -e 'gem "psych"; puts $LOAD_PATH'
/foo
/builddir/.gem/ruby/2.7.0/gems/psych-3.1.0/lib
/builddir/.gem/ruby/2.7.0/extensions/x86_64-linux/2.7.0-static/psych-3.1.0
/usr/local/lib/ruby/site_ruby/2.7.0
/usr/local/lib/ruby/site_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/site_ruby
/usr/local/lib/ruby/vendor_ruby/2.7.0
/usr/local/lib/ruby/vendor_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/vendor_ruby
/usr/local/lib/ruby/2.7.0
/usr/local/lib/ruby/2.7.0/x86_64-linux
~~~

the `-Ifoo` was specified earlier, therefore it should have lower priority then the `gem "psych"` call.

voxik: Also, I am not convinced that the original reason as #1866 is even correct, because default gems are simply loaded from the ruby stdlib and therefore every `-I` option has *always* precedence, so I think the original test case must be flawed.

voxik: Sorry, my initial example was probably not complete enough because it is not obvious that the activated Psych is user installed. Here is more complete example:

~~~
$ gem list | grep psych
psych (3.1.0)

$ ruby -Ifoo -e 'gem "psych"; puts $LOAD_PATH'
/foo
/builddir/.gem/ruby/2.7.0/gems/psych-3.1.0/lib
/builddir/.gem/ruby/2.7.0/extensions/x86_64-linux/2.7.0-static/psych-3.1.0
/usr/local/lib/ruby/site_ruby/2.7.0
/usr/local/lib/ruby/site_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/site_ruby
/usr/local/lib/ruby/vendor_ruby/2.7.0
/usr/local/lib/ruby/vendor_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/vendor_ruby
/usr/local/lib/ruby/2.7.0
/usr/local/lib/ruby/2.7.0/x86_64-linux

$ rm -rf ~/.gem/

$ ruby -Ifoo -e 'gem "psych"; puts $LOAD_PATH'
/foo
/usr/local/lib/ruby/site_ruby/2.7.0
/usr/local/lib/ruby/site_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/site_ruby
/usr/local/lib/ruby/vendor_ruby/2.7.0
/usr/local/lib/ruby/vendor_ruby/2.7.0/x86_64-linux
/usr/local/lib/ruby/vendor_ruby
/usr/local/lib/ruby/2.7.0
/usr/local/lib/ruby/2.7.0/x86_64-linux
~~~

The first example is the same as the one above with locally installed psych overriding the default gem. Then I delete user the `.gem` directory, which removes the installed gem and go with defaults. As you can see, there are no paths inserted anywhere for the default gem, because there are not paths to be inserted and therefore I believe the whole logic is there just to satisfy the wrong test case not reflecting the reality.

deivid-rodriguez: Regarding your first comment, that's how rubygems require has been working for years, gems have always been added right before the default `$LOAD_PATH`, not at the front, and the reason has always been that `-I` CLI flags should take precedence. If you feel that historical behavior should be changed, feel free to open an issue to discuss.

> Also, I am not convinced that the original reason as #1866 is even correct, because default gems are simply loaded from the ruby stdlib and therefore every -I option has always precedence, so I think the original test case must be flawed.

Before #1866, if you had `fileutils.rb` file in your current folder and specified `-I.`, the file in your current folder would be ignored because the `fileutils` default gem would get activated and the file in the default `$LOAD_PATH` would get loaded.

deivid-rodriguez: Sorry, our comments crossed, but I don't see how the last comment you posted is related to this ticket or any of the related tickets. Can you elaborate a bit? What you're doing there (removing a gem directly without `gem uninstall` and thus not removing its gemspec) is something I had never tried and we just learnt how rubygems behaves there :laughing:. But I don't quite see the relation.

deivid-rodriguez: Sorry, my comment was not accurate. The file in your currrent folder will be loaded, but the default gem with the same name will also be activated, and from the point of view of #1866, thar is unexpected. I understand that it is debatable though and I'm open to discussing switching back to the old behavior which I believe is what you're proposing?

voxik: > that's how rubygems require has been working for years, gems have always been added right before the default $LOAD_PATH, not at the front, and the reason has always been that -I CLI flags should take precedence.

Ah, you are right. Sorry, I have been mistaken.

My entire point is that if you take high level look at what RubyGems do, the `gem` call is essentially just modification of `$LOAD_PATH`. From that POV call of `gem 'psych'` for default gem is no-op (second example) while it modifies the `$LOAD_PATH` for non default gem (first example). The #1866 talks just about "default" gems. If the gem activation for default gem is no-op, then the `-I` option has to automatically take the precedence.

I understand the situation is a bit more complex behind curtain, but maybe it is overly complex. Considering that the test case just tries to emulate what is the real situation, I have just low confidence that the test case is correct. But I have not done any deep investigation.

> (removing a gem directly without gem uninstall and thus not removing its gemspec)

Here I remove the whole directory from user home, therefore I remove everything, the gem, the gemspec, the documentation, the gem cache, etc, making sure I reset my environment to pristine conditions. I could do `gem uninstall` of course, but that would still left some leftovers behind.

voxik: Looking close into #1868, I think that call to `gem_original_require(resolved_path)` means that the `Kernel#require` is called with completely different path then originally requested, so `require "psych"` might RubyGems transform to `require "some/path/psych"` and that is not acceptable IMO.

deivid-rodriguez: Sorry, I missed that your example was removing the whole gem directory. Then I'm not sure why activating psych is not failing, I would've expected it to fail? :thinking:. Anyways, I feel it's only tangentially related to this issue?

 > My entire point is that if you take high level look at what RubyGems do, the gem call is essentially just modification of $LOAD_PATH. From that POV call of gem 'psych' for default gem is no-op (second example) while it modifies the $LOAD_PATH for non default gem (first example). The #1866 talks just about "default" gems. If the gem activation for default gem is no-op, then the -I option has to automatically take the precedence.

> I understand the situation is a bit more complex behind curtain, but maybe it is overly complex. Considering that the test case just tries to emulate what is the real situation, I have just low confidence that the test case is correct. But I have not done any deep investigation.

Yes, it is more complex. And yes, maybe it's overly complex. Gem activation for default gems does not add paths to the `$LOAD_PATH` but it's definitely not a no-op. As I said, I'm happy to try simplify/change things so that they are simpler. But again, separate discussion. Here I'm just fixing two bugs, and one of those were reported by you, so I would guess you are ok with fixing it?

> Looking close into #1868, I think that call to gem_original_require(resolved_path) means that the Kernel#require is called with completely different path then originally requested, so require "psych" might RubyGems transform to require "some/path/psych" and that is not acceptable IMO.

This is what ruby's C `require` does internally when you pass a string to it, instead of a path.I wouldn't think that a reason per se for not being acceptable, but I'm happy to discuss it. Again, I feel it fits a separate discussion better.

Can you give me your opinion om whether you would accept, reject, or propose any specific modifications to this change set?

deivid-rodriguez: I'll hold this PR until I have a look at your #3647 report. Thanks, neat report! :)

deivid-rodriguez: With the case presented by @voxik in #3647 it becomes more clear that he's right about

> Looking close into #1868, I think that call to gem_original_require(resolved_path) means that the Kernel#require is called with completely different path then originally requested, so require "psych" might RubyGems transform to require "some/path/psych" and that is not acceptable IMO.

not being acceptable. Because with that code, the `$LOADED_FEATURES` cache of standard require is no longer hit. And there's no way around that unless we reimplement that kind of caching ourselves which seems like... Not the way to go.

I'm trying to revert #1868 in #3648 as an alternative.

deivid-rodriguez: Ok, so I had an idea and ended up adding a couple of extra commits here to also fix #3647. Now there's no manual `$LOAD_PATH` traversal whatsoever for every require, so this should definitely fix #3631, and also there's no more `gem_original_require` calls with a "modified path".

lamont-granquist:
Looks even faster:

/opt/chef-workstation/bin/kitchen list  2.41s user 0.62s system 48% cpu 6.201 total

deivid-rodriguez:
Yeah! Now we no longer do the `-I` paths traversal for every require, but only when a name present on a default gem is required :)

lib/rubygems.rb
```diff
@@ -590,22 +590,25 @@ def self.load_path_insert_index
 
     index = $LOAD_PATH.index RbConfig::CONFIG['sitelibdir']
 
-    index
+    index || 0
+  end
+
+  ##
+  # The number of paths in the `$LOAD_PATH` from activated gems. Used to
+  # proritize `-I` and `ENV['RUBYLIB`]` entries during `require`.
```
olleolleolle, line 598:
```suggestion
  # prioritize `-I` and `ENV['RUBYLIB`]` entries during `require`.
```

Typo.

deivid-rodriguez: Thanks!

lib/rubygems.rb
deivid-rodriguez: Concerns should've been addressed now.

PR #3655, opened 2020-05-23
Fix source priority for transitive dependencies and split lockfile rubygems source sections

### LINKED ISSUE #3643: Bundler: BUNDLE_DISABLE_MULTISOURCE changes the Gemfile.lock + breaks deployment mode
without BUNDLE_DISABLE_MULTISOURCE all gems are under a single source
with BUNDLE_DISABLE_MULTISOURCE gems are split into different sources

... prefer to always use the same approach so users do not run into merge conflicts and deployment mode install does not fail

workaround is using this at the top of the Gemfile:
```ruby
# https://github.com/rubygems/rubygems/issues/3643
Bundler::FeatureFlag.send(:settings_flag, :disable_multisource) { true }
```

### LINKED ISSUE #3982: Global vs block gem server source priority doesn't work as expected
In a situation like the following:

```ruby
# my-private-gem.gemspec
# ...
gem.dependency("my-another-private-gem")

# Gemfile
source 'https://my-private-server' do
  gem 'my-private-gem'
end
```

Bundler should try to resolve using the private source for my-private-gem and all of its indirect dependencies. Only if a valid resolution is not found, it should also consider `rubygems.org` for those dependencies.

Insted, Bundle will still install `my-another-private-gem` from rubygems.org if the gem exists there. We should probably search first in the same source as `my-private-gem` before falling back.

_Originally posted by @rafaelfranca in https://github.com/rubygems/rubygems/issues/3948#issuecomment-692920465_

deivid-rodriguez:

We have several source priority bug fixes in master that we have never released because they need the `disable_multisource` setting toggled, and that will only happen by default in Bundler 3.

The reason for this is that this setting drops support for multiple global sources in the `gems.rb/Gemfile` file, and also enables separate section for each rubygems source in the lockfile. We considered this to be backwards incompatible, although I'm not sure that was the right call after looking into this.

This commit enables the functionality by default, and completely removes the `disable_multisource` setting. The only backwards compatible concern I'm having is when some is using a lokfile in the old format in
frozen mode. In that case, to not break any workflows, we print a warning and still use that lockfile. Also, plugins still need the
previous functionality. I think this can probably be fixed later, but I'm not doing that here.

In every other case, we will run in the new secure mode and update the `Gemfile.lock` file with the new format.

In addition, this PR also fixes source priority for transitive dependencies, which was not fully fixed by enabling the `disable_multisource` mode.

Fixes #3643.
Fixes #3982.

# Tasks:

- [x] Write tests

### DIFF
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -46,15 +46,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           path: rubygems/rubygems
-      - name: Test RubyGems
+      - name: Sync tools
         run: |
           ruby tool/sync_default_gems.rb rubygems
-          make -j2 -s test-all TESTS="rubygems --no-retry"
+          ruby tool/sync_default_gems.rb bundler
+        working-directory: ruby/ruby
+      - name: Test RubyGems
+        run: make -j2 -s test-all TESTS="rubygems --no-retry"
         working-directory: ruby/ruby
         if: matrix.target == 'Rubygems'
       - name: Test Bundler
         run: |
-          ruby tool/sync_default_gems.rb bundler
           git checkout lib/bundler/bundler.gemspec
           git add .
           make test-bundler-parallel
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -900,13 +900,7 @@ module Bundler
       # Record the specs available in each gem's source, so that those
       # specs will be available later when the resolver knows where to
       # look for that gemspec (or its dependencies)
-      source_requirements = { :default => sources.default_source }
-      default = Bundler.feature_flag.disable_multisource? && sources.default_source
-      dependencies.each do |dep|
-        dep_source = dep.source || default
-        next unless dep_source
-        source_requirements[dep.name] = dep_source
-      end
+      source_requirements = { :default => sources.default_source }.merge(dependency_source_requirements)
       metadata_dependencies.each do |dep|
         source_requirements[dep.name] = sources.metadata_source
       end
@@ -916,15 +910,7 @@ module Bundler
     end
 
     def pinned_spec_names(skip = nil)
-      pinned_names = []
-      default = Bundler.feature_flag.disable_multisource? && sources.default_source
-      dependencies.each do |dep|
-        dep_source = dep.source || default
-        next unless dep_source
-        next if dep_source == skip
-        pinned_names << dep.name
-      end
-      pinned_names
+      dependency_source_requirements.reject {|_, source| source == skip }.keys
     end
 
     def requested_groups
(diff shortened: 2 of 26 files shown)


RoMi67:
Idk

deivid-rodriguez:
I finally got the CI green here. It should be ready for a review now!

deivid-rodriguez:
Just to clarify something else.

Lockfiles who currently have multiple rubygems sources (not top level, just more than one), and that are run in frozen mode will start to fail because bundler won't like that kind of lockfile with all rubygems sources merged in a single section.

I think that's fine, since it is what this issue is about. Those people will have to regenerate their lockfiles in non frozen mode so that each gem is unambiguously recorded as coming from a specific rubygems source, thus fixing the potential security problem.

deivid-rodriguez:
I guess I could try respecting the bad lockfile if in frozen mode and print a warning instead... :thinking:

deivid-rodriguez:
I think CI failed here because of #4366, so I rebased this PR on top of [the fix for that](#4371) to see if that fixes the CI.

aaronjensen:
Will this allow us to work around the attacks described in https://www.bleepingcomputer.com/news/security/researcher-hacks-over-35-tech-firms-in-novel-supply-chain-attack/ ?

As things are today, it doesn't seem to be possible to say *only* get these dependencies from this gem server

aaronjensen:
Related: I know this has been discussed before and decided against, but `gemspec`s **really** should be allowed to specify the source for their gem dependencies.

deivid-rodriguez:
> Will this allow us to work around the attacks described in https://www.bleepingcomputer.com/news/security/researcher-hacks-over-35-tech-firms-in-novel-supply-chain-attack/ ?

Yes, this PR should fix that issue. I will release it today.

aaronjensen:
That's great, thanks. Can you clarify how it works with transitive dependencies? i.e. if I have:

```rb
source "my-source.com" do
  gem "my-gem" # depends on my-other-gem
end
```

Will it prefer to get `my-other-gem` from `my-source.com`?

aaronjensen:
I reread the description, and it seems like the answer is yes

deivid-rodriguez:
It is!

bquorning:
Just FYI, it looks like this PR added some of the performance problems back that were fixed in 2.2.8.

On 145bc3e62027fa67b83887e6d5997a88d703800d:
```
~/Code/rubygems/bundler/bin/bundle check  0.44s user 0.09s system 99% cpu 0.531 total
```

On da7d605d0738eb96897999b9cd0dbb9c1ec5b5fd:
```
~/Code/rubygems/bundler/bin/bundle check  2.76s user 0.16s system 99% cpu 2.928 total
```

deivid-rodriguez:
Thanks @bquorning, can you open an issue? Best if reproducible of course!

PR #3660, opened 2020-05-26
Incorporate a more reliable path check in Kernel#require

yuki24:
This pull request replaces the `load_error.message.end_with?` call with `load_error.path`, which I believe is more reliable (unnoticeably faster.)

A few weeks ago, I added [a "Did you mean?" feature to `LoadError`](https://github.com/ruby/did_you_mean/pull/143) so users would be notified of a typo with possible suggestions. It seemed safe to do so as most of the tests were passing in [the Ruby CI](https://github.com/ruby/ruby/pull/3135). However I have found one issue where some gems could not be loaded with `require "gem_name"` any more even though it is clearly installed:

```sh
$ gem list | grep pry
pry (0.13.1)
pry-byebug (3.9.0)
pry-rails (0.3.9)

$ ruby -e 'require "pry"'
        2: from -e:1:in `<main>'
        1: from ~/.rbenv/versions/2.7.1/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require'
~/.rbenv/versions/2.7.1/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:92:in `require': cannot load such file -- pry (LoadError)
Did you mean?  pty
```

It turns out that the problem is in the `require` method in Rubygems, where we call `end_with?` on the error message of the load error. Rubygems depends on the error message, and `did_you_mean` overrides it with something unexpected for Rubygems, hence breaking `require`.

In this case, we can simply use the `LoadError#path` method. It's been available [at least Ruby 2.0.0](https://docs.ruby-lang.org/en/2.0.0/LoadError.html) and LoadError's error message is [generated based off of the same `path` value in Ruby](https://github.com/ruby/ruby/blob/073cc5e815fcf5178fe4e515fcde74dc3597adeb/error.c#L1974-L1976), so I feel pretty good about it.

In terms of testing, I'm not sure if it's worth it. I could write a test for it, but it does look a bit too contrived and it could even cause an issue when the test suite is parallelized:

```ruby
def test_normal_gems_with_overridden_load_error_message
  normal_gem_spec = util_spec("default", "3.0", nil, "lib/default/gem.rb")
  install_specs(normal_gem_spec)

  LoadError.define_method :message do
    "Overridden message"
  end

  assert_require "default/gem"
  assert_equal %w[default-3.0], loaded_spec_names
ensure
  LoadError.class_eval do
    undef message
    def message
      to_s
    end
  end
end
```

I also can't image anyone would ever go back to `message.end_with?`, so I'll just leave the test there and lease the decision up to the maintainers here.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -509,6 +509,33 @@ class TestGemRequire < Gem::TestCase
     assert_equal %w[default-3.0], loaded_spec_names
   end
 
+  def test_normal_gems_with_overridden_load_error_message
+    normal_gem_spec = util_spec("normal", "3.0", nil, "lib/normal/gem.rb")
+
+    install_specs(normal_gem_spec)
+
+    File.write("require_with_overridden_load_error_message.rb", <<-RUBY)
+      LoadError.class_eval do
+        def message
+          "Overridden message"
+        end
+      end
+
+      require 'normal/gem'
+    RUBY
+
+    require "open3"
+
+    output, exit_status = Open3.capture2e(
+      { "GEM_HOME" => Gem.paths.home },
+      *ruby_with_rubygems_in_load_path,
+      "-r",
+      "./require_with_overridden_load_error_message.rb"
+    )
+
+    assert exit_status.success?, "Require failed due to #{output}"
+  end
+
   def test_default_gem_prerelease
     default_gem_spec = new_default_spec("default", "2.0.0",
                                         nil, "default/gem.rb")
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -148,7 +148,7 @@ module Kernel
     RUBYGEMS_ACTIVATION_MONITOR.enter
 
     begin
-      if load_error.message.end_with?(path) and Gem.try_activate(path)
+      if load_error.path == path and Gem.try_activate(path)
         require_again = true
       end
     ensure


yuki24:
Alright, added a test and the CI is ✅ let me know if there is anything else I should do.

test/rubygems/test_require.rb
```diff
@@ -509,6 +509,33 @@ def test_default_gem_and_normal_gem
     assert_equal %w[default-3.0], loaded_spec_names
   end
 
+  def test_normal_gems_with_overridden_load_error_message
+    normal_gem_spec = util_spec("default", "3.0", nil, "lib/default/gem.rb")
```
deivid-rodriguez, line 513: Could you use names and paths different than "default" just to make it clear that default gems are not involved in this test? The "normal gem_spec" name for the variable already emphasizes this, but still.

test/rubygems/test_require.rb
```diff
@@ -509,6 +509,33 @@ def test_default_gem_and_normal_gem
...
+
+    install_specs(normal_gem_spec)
+
+    File.write("require_with_overridden_loead_error_message.rb", <<-RUBY)
```
deivid-rodriguez, line 517:
```suggestion
    File.write("require_with_overridden_load_error_message.rb", <<-RUBY)
```

test/rubygems/test_require.rb
```diff
@@ -509,6 +509,33 @@ def test_default_gem_and_normal_gem
...
+      require 'default/gem'
+    RUBY
+
+    require "open3"
+
+    output, exit_status = Open3.capture2e(
+      { "GEM_HOME" => Gem.paths.home },
+      *ruby_with_rubygems_in_load_path,
+      "-r",
+      "./require_with_overridden_loead_error_message.rb"
```
deivid-rodriguez, line 533:
```suggestion
      "./require_with_overridden_load_error_message.rb"
```

test/rubygems/test_require.rb
yuki24:
```suggestion
    normal_gem_spec = util_spec("normal", "3.0", nil, "lib/normal/gem.rb")
```

test/rubygems/test_require.rb
```diff
@@ -509,6 +509,33 @@ def test_default_gem_and_normal_gem
...
+    install_specs(normal_gem_spec)
+
+    File.write("require_with_overridden_load_error_message.rb", <<-RUBY)
+      LoadError.class_eval do
+        def message
+          "Overridden message"
+        end
+      end
+
+      require 'default/gem'
```
yuki24, line 524:
```suggestion
      require 'normal/gem'
```

test/rubygems/test_require.rb
yuki24: All done :+1:

PR #3403, opened 2020-03-18
Add new option to `bundle gem` for choosing a CI sevice

hsbt:
# Description:

Migrated from https://github.com/rubygems/bundler/pull/7627

/cc @colby-swandale @indirect 

______________

# Tasks:

- [ ] Describe the problem / feature
- [ ] Write tests
- [ ] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3667, opened 2020-05-29
Follow-up #3403: Add new option to `bundle gem` for choosing a CI service

FTLam11:
Please refer to #3403 for full details. This PR is more of a follow-up and mainly incorporates naming/grammar suggestions while addressing some test suite failures.

One thing that may warrant discussion: Test framework selection/configuration does not currently affect the generation of CI files. So a user can choose to not generate test framework files, while still being able to generate CI files.
______________

# Tasks:

- [x] Write tests


bundler/spec/commands/newgem_spec.rb
```diff
@@ -668,6 +651,94 @@ def create_temporary_dir(dir)
       it_behaves_like "test framework is absent"
     end
 
+    context "--ci with no arugment" do
```
olleolleolle, line 654:
```suggestion
    context "--ci with no argument" do
```

bundler/lib/bundler/templates/newgem/.github/workflows/main.yml.tt
```diff
@@ -0,0 +1,18 @@
+name: Ruby
+
+on: [push,pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby <%= ::Gem::Version.new(RUBY_VERSION).segments[0..1].join(".") %>
```
deivid-rodriguez, line 10: I think `Set up Ruby" is enough. Easier to keep up to date.

bundler/lib/bundler/templates/newgem/.github/workflows/main.yml.tt
```diff
@@ -0,0 +1,18 @@
...
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: <%= ::Gem::Version.new(RUBY_VERSION).segments[0..1].join(".") %>.x
+    - name: Build and test with Rake
+      run: |
+        gem install bundler -v <%= Bundler::VERSION %>
+        bundle install --jobs 4 --retry 3
```
deivid-rodriguez, line 17: I would remove the `--jobs 4`, since the next version of `bundler` will autochoose jobs by default. We can also remove `--retry 3` since it's the default.

bundler/lib/bundler/templates/newgem/.gitlab-ci.yml.tt
```diff
@@ -0,0 +1,9 @@
+image: ruby:<%= RUBY_VERSION %>
+
+before_script:
+  - gem install bundler
+  - bundle install -v <%= Bundler::VERSION %>
```
deivid-rodriguez, line 5: This command is incorrect, the `-v <%= Bundler::VERSION %>` should go in the line above.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -247,6 +257,35 @@ def test_framework_hint
...
+          ci_template = false
+        end
+      end
+
+      if Bundler.settings["gem.ci"].nil?
+        Bundler.settings.set_global("gem.ci", ci_template)
+      end
+
+      ci_template
+    end
```
deivid-rodriguez, line 287: In this case, the `--ci` flag requires a value or what happens if I pass just `--ci`. Do I get asked?

bundler/lib/bundler/cli.rb
```diff
@@ -575,6 +575,7 @@ def viz
     method_option :rubocop, :type => :boolean, :desc => "Add rubocop to the generated Rakefile and gemspec. Set a default with `bundle config set gem.rubocop true`."
     method_option :test, :type => :string, :lazy_default => Bundler.settings["gem.test"] || "", :aliases => "-t", :banner => "Use the specified test framework for your library",
                          :desc => "Generate a test directory for your library, either rspec, minitest or test-unit. Set a default with `bundle config set gem.test rspec`."
+    method_option :ci,   :type => :string, :desc => "Generate CI configuration, either GitHub Actions, Travis CI, GitLab CI or CircleCI. Set a default with `bundle config set gem.ci (github|travis|gitlab|circle)`"
```
deivid-rodriguez, line 578: Style nit, remove the extra spacing between `:ci,`, and `:type`.

bundler/lib/bundler/cli/gem.rb
FTLam11: Bundler configuration of test framework/CI is now consistent. Hints will be displayed depending on the global config/user passed flag. Just passing `--ci` will bring up the prompt to ask for user input.

FTLam11:
I just remembered to also update `bundle gem` man pages for the new `--ci` option.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -183,6 +193,12 @@ def run
                         "so -t is not needed if you want to continue using it. " \
                         "This setting can be changed anytime with `bundle config gem.test`."
       end
+
+      if options[:ci] == Bundler.settings["gem.ci"]
```
bronzdoc, line 197: What about:
```ruby
"#{options[:ci]} is already configured, ignoring --ci flag"
```

bundler/lib/bundler/cli/gem.rb
```diff
@@ -221,8 +237,9 @@ def ask_and_set_test_framework
 
       if test_framework.to_s.empty?
         Bundler.ui.confirm "Do you want to generate tests with your gem?"
-        Bundler.ui.info test_framework_hint
-        result = Bundler.ui.ask "Type 'rspec', 'minitest' or 'test-unit' to generate those test files now. " \
+        Bundler.ui.info hint_text("test")
+
+        result = Bundler.ui.ask "Enter a framework name to generate those test files now. " \
```
bronzdoc, line 242: Maybe just:

"Enter a test framework." ?

bundler/lib/bundler/cli/gem.rb
```diff
@@ -238,15 +255,45 @@ def ask_and_set_test_framework
...
-        "This setting can be changed anytime with `bundle config gem.test`."
+        "This setting can be changed anytime with `bundle config gem.#{setting}`."
       end
     end
 
+    def ask_and_set_ci
+      ci_template = options[:ci] || Bundler.settings["gem.ci"]
+
+      if ci_template.to_s.empty?
+        Bundler.ui.confirm "Do you want to set up automated testing for your gem? " \
```
bronzdoc, line 271: I feel like we are saying to much, we don't need to tell them what is CI used for

Maybe just:

```
      "Do you want to configure a continuous integration service?\n"\

      "Supported services:"\
      "* CircleCI:       https://circleci.com/\n" \
      "* GitHub Actions: https://github.com/features/actions\n" \
      "* GitLab CI:      https://docs.gitlab.com/ee/ci/\n" \
       "* Travis CI:      https://travis-ci.org/\n" \
          "\n"
```

bundler/man/bundle-gem.1.txt
```diff
@@ -83,9 +83,27 @@ OPTIONS
               displayed  and  the  answer  will  be  saved in Bundler's global
               config for future bundle gem use.
 
+       --ci, --test=github, --test=travis, --test=gitlab, --test=circle
```
bronzdoc, line 86: Why we pass the ci value to the `--test` flag instead of the `--ci` flag?

FTLam11: Sorry, that was a bad copy pasta.

bundler/lib/bundler/cli/gem.rb
FTLam11: The current reminder text appears after the gem has been generated. I put it towards the end so that it would be easier to notice once all the generator output is shown.

If we moved this hint to before any templates are generated, I think your suggestion would work better. What do you think?

bundler/lib/bundler/templates/newgem/.github/workflows/main.yml.tt
```diff
@@ -0,0 +1,18 @@
...
+
+on: [push,pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Ruby
+      uses: actions/setup-ruby@v1
```
deivid-rodriguez, line 11: I just noticed that this is no longer the recommended action for setting up ruby, it's `ruby/setup-ruby` now.

bundler/lib/bundler/templates/newgem/.github/workflows/main.yml.tt
```diff
@@ -0,0 +1,18 @@
...
+      with:
+        ruby-version: <%= ::Gem::Version.new(RUBY_VERSION).segments[0..1].join(".") %>.x
```
deivid-rodriguez, line 13: The new action supports full ruby version locking, so you can say `RUBY_VERSION` here like in the other templates.

bundler/lib/bundler/templates/newgem/.github/workflows/main.yml.tt
```diff
@@ -0,0 +1,18 @@
...
+    - name: Build and test with Rake
```
deivid-rodriguez, line 14: I'd say "Run tests" like in the `CircleCI` configuration.

bundler/lib/bundler/cli/gem.rb
bronzdoc: Sounds good 👍

bundler/lib/bundler/cli/gem.rb
```diff
@@ -238,13 +248,45 @@ def ask_and_set_test_framework
...
+          ci_template = false
+        end
+      end
+
+      if Bundler.settings["gem.ci"].nil?
+        Bundler.settings.set_global("gem.ci", ci_template)
+      end
+
+      if options[:ci] == Bundler.settings["gem.ci"]
+        Bundler.ui.info "#{options[:ci]} is already configured, ignoring --ci flag."
```
deivid-rodriguez, line 286: This is cute, I like it! :+1: Does it make sense to make this consistent with the `--test` flag?

bundler/lib/bundler/templates/newgem/.circleci/config.yml.tt
```diff
@@ -0,0 +1,13 @@
...
+    docker:
+      - image: ruby:<%= RUBY_VERSION %>
+    steps:
+      - checkout
+      - run:
+          name: Run tests
+          command: |
+            gem install bundler -v <%= Bundler::Version %>
+            bundle install
+            bundle exec rake test
```
deivid-rodriguez, line 13: Sorry for going back and forth here

Since now `--test` and `--ci` flags are completely independent, let's not make the CI configuration "test specific". Let's name the default step "Run the default task", and run the `bundle exec rake` command, whatever that is. Same kind of thing would apply for all CI configurations.

FTLam11: Generic templates are good! 3/4 CI templates were updated.

bundler/lib/bundler/templates/newgem/.circleci/config.yml.tt
```diff
@@ -0,0 +1,13 @@
...
+jobs:
+  build:
+    docker:
+      - image: ruby:<%= RUBY_VERSION %>
+    steps:
+      - checkout
+      - run:
+          name: Run the default task
+          command: |
+            gem install bundler -v <%= Bundler::Version %>
```
deivid-rodriguez, line 11:
```suggestion
            gem install bundler -v <%= Bundler::VERSION %>
```

The `Bundler::Version` constant doesn't exist.

This error actually makes `bundle gem <foo> --ci circle` crash. The current specs didn't catch it because some use `bundle "gem ...` instead of `bundle! "gem...`, and the former doesn't raise when errors happen.

Could you change those to always use the `bundle!` helper?

This is our test suite's fault in my opinion, and I plan to fix it: the "raising" version should be the default unless you explicitly opt-out.

FTLam11: Ohhhh, so `bundle` should be used when we are testing for an error to be raised? Are there other situations where we should prefer to use `bundle`? I did a rough count of each used in the rubygems repo.

```
ag -Q 'bundle "' | wc -l
902
ag -Q 'bundle! "' | wc -l
386
```

I'll check `newgem_spec.rb` and make the appropriate changes.

deivid-rodriguez: > Ohhhh, so bundle should be used when we are testing for an error to be raised? Are there other situations where we should prefer to use bundle?

Yep, I believe testing errors is the only use case, otherwise always use `bundle!`. My plan is to change `bundle` to start raising on error, and change the specs that test for errors to pass a `raise: false` option to opt out.

FTLam11:
> Only thing I wonder is whether it'd be better to introduce #3685 first, and then you rebase and remove the last commit from this PR, so that we don't introduce changes that are about to be reverted.

Sounds good, I'll be on standby. Thanks for giving really thorough reviews through this entire PR, it has been very helpful.

deivid-rodriguez:
Hi @FTLam11, as you can see from the merge, we ended up doing this in a different order than I had planned, so you no longer need to mess with rebasing this PR :)

Just wanted to thank you for your hard work on improving our gem generator :muscle:.

PR #3668, opened 2020-05-29
Skip gemspec validations that warn except when on a gem authoring context

### LINKED ISSUE #3664: Programs using "bundle ruby" start slower with proprietary gemspec licenses
We have discovered that specifying a proprietary license in a gemspec causes `bundle ruby` programs to start slower. This happens independently of platform, bundler or ruby version.

This issue is related to:
  - [x ] Other

```
$ gem env version
3.0.3
$ bundler -v
Bundler version 1.17.3
```

## Problem description

We have a large monorepo with 50+ projects. All have gemspecs. We specified a proprietary license "Internal source code", as our application is closed source. We had noticed for some time that our invocation times were not great doing even simple commands.

1. We invoke a `bundle rake` command that does nothing
1. The project Gemfile refers to ~30+ other projects using `path` attributes
1. The invocation takes ~ 2.5 seconds

## Unintentional debugging

In unrelated work, we added [license_finder](https://github.com/pivotal/LicenseFinder) to our project, and as part of cleanup for that we saw that the [gemspec guidelines](https://guides.rubygems.org/specification-reference/#license=) say that "Not specifying a license means all rights are reserved; others have no rights to use the code for any purpose."

We therefore removed our proprietary license form all of our internal gemspecs. Immediately all of our command execution times improved. In the example above it fell to 0.7 seconds.

## The penny drops

Digging deeper, we found that during startup, the `rubygems` gem performs [validate_licenses](https://github.com/rubygems/rubygems/blob/82a440aea93909f0d191e4256486589aa946414b/lib/rubygems/specification_policy.rb#L69). As part of this, it tries to see if the [license is known](https://github.com/rubygems/rubygems/blob/82a440aea93909f0d191e4256486589aa946414b/lib/rubygems/specification_policy.rb#L329), and if it isn't it tries to find the best match by [levenshtein distance](https://github.com/rubygems/rubygems/blob/82a440aea93909f0d191e4256486589aa946414b/lib/rubygems/util/licenses.rb#L432).

This last algorithmic lookup is not cheap, and it is where the app was spending ~1.8 seconds during every startup, as every internal gem had an unknown license. It also isn't necessary for us when invoking `bundle ruby` commands, as the warning is never printed, as `bundler` [silences them](https://github.com/rubygems/bundler/blob/35be6d9a603084f719fec4f4028c18860def07f6/lib/bundler/rubygems_integration.rb#L70).

## Next steps

This work yielded a number of threads, and overlaps slightly between `bundler` and `rubygems`, so I'm sure maintainers from each may be interested. I was starting down the process of creating a PR, but realised that since there a couple of ways of fixing this I should confer first.

At the very least, I'd like to clarify internal mono-repo license best practice. I will open a new issue for that if I may.

Possible actions:

1. Change the lookup to cache levenshtein distance results for an unknown license.
1. Change the validation to not do the levenshtein distance lookup if the result won't be used
1. Something else

## Try this at home

If you have a large mono repo with internal project gemspecs, get your `sed` on and replace all the licenses in all the gemspecs with "FooTest". 

Everything will then start slower.

deivid-rodriguez:
# Description:

As highlighted by #3664, gemspec validations can be slow.

Also, gemspec validations are targeted to gem authors, not final users. That's (I guess) why bundler [silences them](https://github.com/rubygems/bundler/blob/35be6d9a603084f719fec4f4028c18860def07f6/lib/bundler/rubygems_integration.rb#L70) when using `path` gems. It needs to validate them to make sure they are functional, but it's not interested in validation warnings.

So, my proposal is to limit the "optional linting" than emits warnings to "gem packaging" contexts.

As per the implementation, I "reused" the `packaging` parameter to `Gem::Specification#validate` for this. The [introduction of this parameter](https://github.com/rubygems/rubygems/commit/e95ff9a2d52f65e25b5e4a60c3f64d751c9e6df4) is not very explicit about its motivation, but I believe the intention was precisely to do something like this.

Closes #3664.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -37,6 +37,9 @@ class Gem::SpecificationPolicy
   #
   # Raises InvalidSpecificationException if the spec does not pass the
   # checks.
+  #
+  # It also performs some validations that do not raise but print warning
+  # messages instead.
 
   def validate(strict = false)
     validate_nil_attributes
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -44,7 +44,7 @@ class Gem::SpecificationPolicy
   def validate(strict = false)
     validate_required!
 
-    validate_optional(strict)
+    validate_optional(strict) if packaging || strict
 
     true
   end
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -309,8 +309,7 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
             end
 
     unless Array === val and val.all? {|x| x.kind_of?(klass)}
-      raise(Gem::InvalidSpecificationException,
-            "#{field} must be an Array of #{klass}")
+      error "#{field} must be an Array of #{klass}"
     end
   end
 
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -127,7 +127,9 @@ class Gem::SpecificationPolicy
   end
 
   ##
-  # Implementation for Specification#validate_dependencies
+  # Checks that dependencies use requirements as we recommend.  Warnings are
+  # issued when dependencies are open-ended or overly strict for semantic
+  # versioning.
 
   def validate_dependencies # :nodoc:
     # NOTE: see REFACTOR note in Gem::Dependency about types - this might be brittle
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -33,8 +33,7 @@ class Gem::SpecificationPolicy
   attr_accessor :packaging
 
   ##
-  # Checks that the specification contains all required fields, and does a
-  # very basic sanity check.
+  # Does a sanity check on the specification.
   #
   # Raises InvalidSpecificationException if the spec does not pass the
   # checks.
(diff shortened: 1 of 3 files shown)


simi:
Looks good to me! I'm a little worried about the test coverage of this change, but I have currently no idea how to make it better. I can take a look and rethink this later if needed.

deivid-rodriguez:
If there's no objections to do this, I think the coverage is fine. All these validations are quite thoroughly tested I believe.

lib/rubygems/specification_policy.rb
```diff
@@ -33,13 +33,32 @@ def initialize(specification)
...
+  # Does a sanity check on the specification.
   #
   # Raises InvalidSpecificationException if the spec does not pass the
   # checks.
+  #
+  # It also performs some validations that do not raise but print warning
+  # messages instead.
 
   def validate(strict = false)
+    validate_strict
```
bronzdoc, line 45: we will `validate_strict` even if `strict` is false?

deivid-rodriguez: Nice catch, the method was very poorly named!

I'll rename this method to `validate_required!` to avoid any confusion with the `strict` flag.

deivid-rodriguez: The answer to your question is "yes", by the way :)

deivid-rodriguez: To expand on the answer, the `strict` flag means that validations that normally lead to warnings should be considered as hard errors, but validations that always lead to hard errors should still be run even if `strict` is `false`.

deivid-rodriguez: Code updated!

deivid-rodriguez:
@bronzdoc Could you review again :recycle: ?

deivid-rodriguez:
Ok, let's do this!

Issue #3650, opened 2020-05-21
Making require "rubygems/package" work reliably


eregon:
```
$ gem env version
3.0.3
```

On Ruby 2.6.6:
```
$ ruby -v --disable-gems -e 'require "rubygems/package"' |& cat
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]
/home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/deprecate.rb:24:in `<top (required)>': uninitialized constant Gem (NameError)
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/exceptions.rb:7:in `require'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/exceptions.rb:7:in `<top (required)>'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/security.rb:8:in `require'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/security.rb:8:in `<top (required)>'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/package.rb:44:in `require'
	from /home/eregon/.rubies/ruby-2.6.6/lib/ruby/2.6.0/rubygems/package.rb:44:in `<top (required)>'
	from -e:1:in `require'
	from -e:1:in `<main>'
```
Same on 2.7.1 (RubyGems 3.1.2).

Some few gems directly `require "rubygems/package"` (~78 gems with `csearch 'require "rubygems/package"' | wc -l`).
For example https://github.com/appsignal/appsignal-ruby/blob/v2.3.4/ext/extconf.rb

This works on MRI without the `--disable-gems` flag by "luck" due to MRI defining an empty `Gem` module as a way to remember "I should `require 'rubygems'` at startup".

To support this correctly, I think we should do this:
```ruby
module Gem; end
```

In every such file that can be required directly,
or alternatively always require first a common file (which would define `Gem`) in all such files.
(We could also use nesting definitions like `module Gem; module Deprecate; everything indented 2 spaces more`, but that would change a lot of lines, and cause many conflicts)

This causes an issue for TruffleRuby since that error happens even without `--disable-gems`:
https://github.com/oracle/truffleruby/issues/2014
[...]

PR #3670, opened 2020-05-31
Make sure rubygems/package can be directly required reliably

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/3650

______________

### DIFF
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -12,4 +12,10 @@ class TestProjectSanity < Minitest::Test
     assert status.success?, "Expected Manifest.txt to be up to date, but it's not. Run `rake update_manifest` to sync it."
   end
 
+  def test_require_rubygems_package
+    _, status = Open3.capture2e("ruby -v --disable-gems -I 'lib' -e 'require \"rubygems/package\"'")
+
+    assert status.success?
+  end
+
 end
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -2,7 +2,7 @@
 
 require "open3"
 
-class TestProjectSanity < Minitest::Test
+class TestProjectSanity < Gem::TestCase
 
   def test_manifest_is_up_to_date
     skip unless File.exist?(File.expand_path("../../../Rakefile", __FILE__))
@@ -13,9 +13,11 @@ class TestProjectSanity < Minitest::Test
   end
 
   def test_require_rubygems_package
-    _, status = Open3.capture2e("ruby -v --disable-gems -I 'lib' -e 'require \"rubygems/package\"'")
+    require "rubygems/test_case"
 
-    assert status.success?
+    err, status = Open3.capture2e(*ruby_with_rubygems_in_load_path, "--disable-gems", "-e", "'require \"rubygems/package\"'")
+
+    assert status.success?, err
   end
 
 end
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "rubygems/test_case"
 require "open3"
 
 class TestProjectSanity < Gem::TestCase
@@ -13,8 +14,6 @@ class TestProjectSanity < Gem::TestCase
   end
 
   def test_require_rubygems_package
-    require "rubygems/test_case"
-
     err, status = Open3.capture2e(*ruby_with_rubygems_in_load_path, "--disable-gems", "-e", "'require \"rubygems/package\"'")
 
     assert status.success?, err
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -42,6 +42,7 @@
 # #files are the files in the .gem tar file, not the Ruby files in the gem
 # #extract_files and #contents automatically call #verify
 
+require "rubygems"
 require 'rubygems/security'
 require 'rubygems/specification'
 require 'rubygems/user_interaction'


deivid-rodriguez:
Maybe we can add a little test of requiring `rubygems/package` on a `ruby` started with the `--disable-gems` flag.

bronzdoc:
I just tested this and `module Gem; end` Is not fixing anything in this case. I still see the same error reported.

If I put it before every require I hit:
```
        5: from -e:1:in `<main>'
        4: from -e:1:in `require'
        3: from /Users/bronzdoc/projects/rubygems/lib/rubygems/package.rb:49:in `<top (required)>'
        2: from /Users/bronzdoc/projects/rubygems/lib/rubygems/package.rb:49:in `require'
        1: from /Users/bronzdoc/projects/rubygems/lib/rubygems/specification.rb:39:in `<top (required)>'
/Users/bronzdoc/projects/rubygems/lib/rubygems/specification.rb:158:in `<class:Specification>': uninitialized constant Gem::VERSION (NameError)
```

But doing `require "rubygems"` fixes it as expected...

So maybe we should `require "rubygems"` do you see any problem with this @deivid-rodriguez ?

deivid-rodriguez:
That's weird because `specification.rb` requires `rubygems/version` at the top, so I think your `$LOAD_PATH` setup might be incorrect.

Either way, we should change our internal requires to use `require_relative`.

bronzdoc:
does this works for you `ruby -v --disable-gems -I "lib" -e 'require "rubygems/package"'
` ?
With the change in this PR?

deivid-rodriguez:
Oh, right! I always get confused with this. In the case of `rubygems`, `Gem::VERSION` is not defined in `rubygems/version.rb` but in `rubygems.rb`. So, yeah, in this case, it sounds like we should `require "rubygems"` indeed.

deivid-rodriguez:
@eregon Thoughts?

eregon:
Interesting.
`ruby -v --disable-gems -e 'require "rubygems/package"'` works for me with the `module Gem; end` in deprecate.rb with Ruby 2.6.6 and default RubyGems 3.0.3.
And it indeed fails with RubyGems master + `module Gem; end`:
```
$ ruby -v --disable-gems -I "lib" -e 'require "rubygems/package"'
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]
	5: from -e:1:in `<main>'
	4: from -e:1:in `require'
	3: from /home/eregon/code/rubygems/lib/rubygems/package.rb:46:in `<top (required)>'
	2: from /home/eregon/code/rubygems/lib/rubygems/package.rb:46:in `require'
	1: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:39:in `<top (required)>'
/home/eregon/code/rubygems/lib/rubygems/specification.rb:158:in `<class:Specification>': uninitialized constant Gem::VERSION (NameError)
```

I agree we should just require the full RubyGems here, but that introduces a cycle because rubygems.rb requires deprecate.rb:
https://github.com/rubygems/rubygems/blob/d841d769027382fe80ec1f49711311355a2a575b/lib/rubygems.rb#L19
And so it fails too by adding `require 'rubygems'` in `deprecate.rb`:
```
ruby -v --disable-gems -I "lib" -e 'require "rubygems/package"'
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]
/home/eregon/code/rubygems/lib/rubygems.rb:19: warning: /home/eregon/code/rubygems/lib/rubygems.rb:19: warning: loading in progress, circular require considered harmful - /home/eregon/code/rubygems/lib/rubygems/deprecate.rb
	from -e:1:in  `<main>'
	from -e:1:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems.rb:19:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems.rb:19:in  `require'

/home/eregon/code/rubygems/lib/rubygems/requirement.rb:3: warning: /home/eregon/code/rubygems/lib/rubygems/requirement.rb:3: warning: loading in progress, circular require considered harmful - /home/eregon/code/rubygems/lib/rubygems/deprecate.rb
	from -e:1:in  `<main>'
	from -e:1:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems.rb:115:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems.rb:1334:in  `<module:Gem>'
	from /home/eregon/code/rubygems/lib/rubygems.rb:1334:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/specification.rb:10:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/specification.rb:10:in  `require'
	from /home/eregon/code/rubygems/lib/rubygems/requirement.rb:3:in  `<top (required)>'
	from /home/eregon/code/rubygems/lib/rubygems/requirement.rb:3:in  `require'

	20: from -e:1:in `<main>'
	19: from -e:1:in `require'
	18: from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in `<top (required)>'
	17: from /home/eregon/code/rubygems/lib/rubygems/package.rb:45:in `require'
	16: from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in `<top (required)>'
	15: from /home/eregon/code/rubygems/lib/rubygems/security.rb:8:in `require'
	14: from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in `<top (required)>'
	13: from /home/eregon/code/rubygems/lib/rubygems/exceptions.rb:3:in `require'
	12: from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in `<top (required)>'
	11: from /home/eregon/code/rubygems/lib/rubygems/deprecate.rb:24:in `require'
	10: from /home/eregon/code/rubygems/lib/rubygems.rb:115:in `<top (required)>'
	 9: from /home/eregon/code/rubygems/lib/rubygems.rb:1334:in `<module:Gem>'
	 8: from /home/eregon/code/rubygems/lib/rubygems.rb:1334:in `require'
	 7: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:10:in `<top (required)>'
	 6: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:10:in `require'
	 5: from /home/eregon/code/rubygems/lib/rubygems/requirement.rb:12:in `<top (required)>'
	 4: from /home/eregon/code/rubygems/lib/rubygems/requirement.rb:37:in `<class:Requirement>'
	 3: from /home/eregon/code/rubygems/lib/rubygems/version.rb:206:in `new'
	 2: from /home/eregon/code/rubygems/lib/rubygems/version.rb:206:in `new'
	 1: from /home/eregon/code/rubygems/lib/rubygems/version.rb:214:in `initialize'
/home/eregon/code/rubygems/lib/rubygems/version.rb:174:in `correct?': uninitialized constant Gem::Deprecate (NameError)
```

So I think we need to move `package.rb` to `package_impl.rb` (or some other "private" name), replace all references inside RubyGems requiring `package.rb` to require `package_impl.rb` instead (to avoid cyclic require), and then package.rb would then just be `require 'rubygems'`.
As a bonus, that would also work on TruffleRuby where Gem is an autoload.

eregon:
I see, just `require 'rubygems'` in `package.rb` actually solves it for `require "rubygems/package"`, because that's not required by `rubygems.rb`.
I think that's a good and simple solution :+1: 

---

Note that `require "rubygems/specification"` doesn't work though:
```
$ ruby --disable-gems -I "lib" -e 'require "rubygems/specification"'
	4: from -e:1:in `<main>'
	3: from -e:1:in `require'
	2: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:9:in `<top (required)>'
	1: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:9:in `require'
/home/eregon/code/rubygems/lib/rubygems/version.rb:152:in `<top (required)>': uninitialized constant Gem (NameError)
```
Not sure if that's a proper entry point, but it's used in [47 gems](https://gist.github.com/eregon/efb9d42ba6a66b343d1ecc1dfc39b775).
`rubygems/specification.rb` is required by `rubygems.rb` so we cannot just `require "rubygems"` there:
```
$ ruby --disable-gems -I "lib" -e 'require "rubygems/specification"'
	4: from -e:1:in `<main>'
	3: from -e:1:in `require'
	2: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:9:in `<top (required)>'
	1: from /home/eregon/code/rubygems/lib/rubygems/specification.rb:9:in `require'
/home/eregon/code/rubygems/lib/rubygems.rb:1358:in `<top (required)>': uninitialized constant Gem::Specification (NameError)
```

`module Gem; VERSION = "3.2.0.pre1".freeze; end` in version.rb (just to try) gets further but then:
```
$ ruby --disable-gems -I "lib" -e 'require "rubygems/specification"'
	2: from -e:1:in `<main>'
	1: from -e:1:in `require'
/home/eregon/code/rubygems/lib/rubygems/specification.rb:2673:in `<top (required)>': undefined method `clear_paths' for Gem:Module (NoMethodError)
```
And `clear_paths` is defined in `rubygems.rb`.

So that one seems hard and I'd tend to say `require "rubygems/specification"` is not a supported entry point (without first a `require 'rubygems'` in user code, and then it's useless), because it's already required by `require "rubygems"` and so it is cyclic.

A possible fix could be to split rubygems.rb in smaller 2 parts, and require the first part in specification.rb, and the second part requires specification.rb + code depending on it, and rubygems.rb requires both parts. But (!) `Gem.clear_paths` [is called in specification.rb](https://github.com/rubygems/rubygems/blob/d841d769027382fe80ec1f49711311355a2a575b/lib/rubygems/specification.rb#L2672-L2673) and it does [`Gem::Specification.reset`](https://github.com/rubygems/rubygems/blob/d841d769027382fe80ec1f49711311355a2a575b/lib/rubygems.rb#L336) so this seems highly cyclic and problematic, and we'd likely need to move or remove that call to `Gem.clear_paths` (it's suspicious anyway, would probably be better in rubygems.rb). Anyway I give up for now on that one, it's not a good idea to `require "rubygems/specification"` in user code.

eregon:
I updated the above, I only searched for `require "rubygems/specification"` but there are a lot more `require 'rubygems/specification'` and it's even used in Bundler.

deivid-rodriguez:
Let's fix this one for now then by `require`'ing `rubygems` on top of `rubygems/package`. We can deal with the other potential issues @eregon mentioned when they bite someone.

eregon:
Agreed 👍

bronzdoc:
:shipit:

test/rubygems/test_project_sanity.rb
```diff
@@ -12,4 +12,10 @@ def test_manifest_is_up_to_date
     assert status.success?, "Expected Manifest.txt to be up to date, but it's not. Run `rake update_manifest` to sync it."
   end
 
+  def test_require_rubygems_package
+    _, status = Open3.capture2e("ruby -v --disable-gems -I 'lib' -e 'require \"rubygems/package\"'")
```
deivid-rodriguez, line 16: I created this helper `ruby_with_rubygems_in_load_path` to simplify command line building when creating ruby subprocesses. You could use it here.

Also, I would capture the output into a variable and pass it as a second argument to the assert below so that it is displayed in case of error.

test/rubygems/test_project_sanity.rb
```diff
@@ -12,4 +12,12 @@ def test_manifest_is_up_to_date
     assert status.success?, "Expected Manifest.txt to be up to date, but it's not. Run `rake update_manifest` to sync it."
   end
 
+  def test_require_rubygems_package
+    require "rubygems/test_case"
```
deivid-rodriguez, line 16: This require needs to go at the top of the file, right before inheriting from `Gem::TestCase`.

bronzdoc: good catch. Did it before changing the parent class, why is this working?

deivid-rodriguez: Because `rubygems/test_case` is being required by other tests. If you run the test file in isolation (`rake TEST=test/rubygems/test_project_sanity.tb`) it will fail though.

bronzdoc: done

PR #3672, opened 2020-06-01
Follow up to addition of head build slack notifications

deivid-rodriguez:
# Description:

This is a followup to #3573.

* I'm adding some missing `if: always()` because some steps need to run unconditionally.
* Also doing some trickery so that initial artifacts are uploaded.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -55,7 +55,7 @@ jobs:
           fields: repo,message,commit,author,action,eventName,ref,workflow
         env:
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
-        if: env.STATUS_CHANGED != '0'
+        if: failure() && env.STATUS_CHANGED != '0'
 
       - name: Rotate status
         run: mv new_status.txt status.txt
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -38,16 +38,16 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: bundler_status
-          path: previous_status.txt
+          path: status.txt
         if: always()
 
       - name: Save status
         if: always()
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Initial artifacts have been uploaded. Now I'm going to test the notification once on push.

deivid-rodriguez:
I'm going nuts with this :rofl:. @bronzdoc Have you added the `SLACK_WEBHOOK_URL` secret?

bronzdoc:
@deivid-rodriguez lol yes

deivid-rodriguez:
@bronzdoc There's one remaining issue here that I couldn't figure out where the initial artifact download wouldn't find any artifacts, while a previous upload had been successful. I'm not sure what's going on, but maybe it will be fixed once the task is run on the default bracnh :man_shrugging:.

So I'm going to merge this to try.

Something else I decided to change is to only notify when status changes from "passing" to "broken" because the opposite case (from "broken" to "passing") should be verified from the PR that fixes it, so I think it's not necessary.

bronzdoc:
Lets try this

deivid-rodriguez:
Ok, so unfortunately, I misunderstood how Github Actions artifacts work, and it turns out we can't use them for this.

Artifacts are used to shared data between jobs in the same workflow, not between different workflows, which is what we're trying to do here :(

To do what we want to achieve here we can:

* Use an external service, like s3.
* Use github API to check the latest status: https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs-for-a-repository.

I'll revert these changes for now and reopen the associated ticket.

deivid-rodriguez:
Made a bit of digging into the API. I think we can get the last status with something like:

```
curl -S "https://api.github.com/repos/rubygems/rubygems/actions/runs?event=schedule&branch=master" |jq '.workflow_runs | map(select(.workflow_id == 716807)) | .[0].conclusion'
```

Samething for `bundler` but changing the id.

PR #3677, opened 2020-06-02
Run CI at 3.1 branch head as well.

simi:
Hello, I was about to check 3.1 head branch CI status after recent merges and I have spotted it doesn't start build on push. This change should change it.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -5,8 +5,7 @@ on:
 
   push:
     branches:
-      - staging
-      - trying
+      - 3.1
 
 jobs:
   install:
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
Looks like the https change needed some updates.

simi:
@deivid-rodriguez yes, this fail was expected. Both branches merged were green on its own, but red merged. That was the reason for being curious about the homu ability to update pull requests after merge in automated way.

deivid-rodriguez:
Yes, this is something bors-ng would've detected indeed. Not sure if they will have fixed the issues that made us removing it, but I'm happy with not using it, honestly.

hsbt:
Thanks!

PR #3685, opened 2020-06-03
Raise on errors

deivid-rodriguez:
# Description:

Some work to make bundler specs raise on errors by default, which should improve bug prevention.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This one should be ready now. 

These changes detected a bunch of specs that were not testing what they were supposed to due to some commands failing silently. I rewrote all of them to hopefully test the right thing, except for [one which I decided to remove after reviewing its history](https://github.com/rubygems/rubygems/pull/3685/commits/12445d80fb6477f85048c84aa2d3573e8b97fec4).

bundler/spec/support/helpers.rb
```diff
@@ -25,11 +25,7 @@ def reset_paths!
 
     def self.bang(method)
       define_method("#{method}!") do |*args, &blk|
-        send(method, *args, &blk).tap do
-          unless last_command.success?
-            raise "Invoking #{method}!(#{args.map(&:inspect).join(", ")}) failed:\n#{last_command.stdboth}"
-          end
-        end
+        send(method, *args, &blk)
```
simi, line 28: What's purpose of this method now?

deivid-rodriguez: Oh, this one used to implement the raising `!` behaviour. Now it does nothing other than delegating, so that `bundle!` and `bundle` now do the same thing.

I will remove this method in the next PR and remove all usages of "hang helpers". But at the last minute I decided to leave that part for the next PR because the diff was becoming huge.

deivid-rodriguez: Do you prefer that I remove it in this PR?

deivid-rodriguez: I'll merge this and remove this method in the next PR as I planned to :+1:.

PR #3687, opened 2020-06-04
Add `--github-username` option and config to `bundle gem`

okuramasafumi:
# Description:

This option allows to fill in `[USERNAME]` in README with explicit value from user input.

Before this change, the value came only from `git config github.user`.
This can be a good way for some users, but there are a few problems.
Firstly, it's not documented that we can fill in `[USERNAME]` this way.
Secondly, the config value is set by some tools implicitly, and is not documented in `git-config` manpage.

Here are some changes in this Pull Request to address these issues.

* Add `--github-username` option to `bundle gem` command
* Save config value if a user gives `--github-username`
* Address the fact that we first look to `git config github.user` for GitHub username in command option description

Note:
This Pull Request actually doesn't change its current behavior.
If a user has `github.user` in their git config, it'll be used.
It's described in command description, but maybe it's not enough.

I sometimes create a new gem and there are always `[USERNAME]` in its generated README.
I didn't know that I can fill in by setting `github.user` in git config, and after knowing it I don't like that way since that value in config is not documented anywhere. I wanted more explicit way of doing so.

I added `--github-username` option to `bungle gem` command.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -599,6 +599,7 @@ module Bundler
                        :desc => "Generate CI configuration, either GitHub Actions, Travis CI, GitLab CI or CircleCI. Set a default with `bundle config set --global gem.ci (github|travis|gitlab|circle)`"
     method_option :linter, :type => :string, :lazy_default => Bundler.settings["gem.linter"] || "",
                            :desc => "Add a linter and code formatter, either RuboCop or Standard. Set a default with `bundle config set --global gem.linter (rubocop|standard)`"
+    method_option :github_username, :type => :string, :default => Bundler.settings["gem.github_username"], :banner => "Set your username on GitHub", :desc => "Fill in GitHub username on README so that you don't have to do it manually. Set a default with `bundle config set --global gem.github_username <your_username>`."
 
     def gem(name)
     end
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -42,8 +42,16 @@ module Bundler
       use_git = Bundler.git_present? && options[:git]
 
       git_author_name = use_git ? `git config user.name`.chomp : ""
-      github_username = use_git ? `git config github.user`.chomp : ""
+      git_username = use_git ? `git config github.user`.chomp : ""
       git_user_email = use_git ? `git config user.email`.chomp : ""
+
+      github_username = if options[:github_username].nil?
+        git_username
+      elsif options[:github_username] == false
+        ""
+      else
+        options[:github_username]
+      end
 
       config = {
         :name             => name,
--- a/bundler/lib/bundler/man/bundle-add.1
+++ b/bundler/lib/bundler/man/bundle-add.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-ADD" "1" "March 2021" "" ""
+.TH "BUNDLE\-ADD" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-add\fR \- Add gem to the Gemfile and run bundle install
--- a/bundler/lib/bundler/man/bundle-binstubs.1
+++ b/bundler/lib/bundler/man/bundle-binstubs.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-BINSTUBS" "1" "March 2021" "" ""
+.TH "BUNDLE\-BINSTUBS" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-binstubs\fR \- Install the binstubs of the listed gems
--- a/bundler/lib/bundler/man/bundle-cache.1
+++ b/bundler/lib/bundler/man/bundle-cache.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CACHE" "1" "March 2021" "" ""
+.TH "BUNDLE\-CACHE" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-cache\fR \- Package your needed \fB\.gem\fR files into your application
--- a/bundler/lib/bundler/man/bundle-check.1
+++ b/bundler/lib/bundler/man/bundle-check.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CHECK" "1" "March 2021" "" ""
+.TH "BUNDLE\-CHECK" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-check\fR \- Verifies if dependencies are satisfied by installed gems
--- a/bundler/lib/bundler/man/bundle-clean.1
+++ b/bundler/lib/bundler/man/bundle-clean.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CLEAN" "1" "March 2021" "" ""
+.TH "BUNDLE\-CLEAN" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-clean\fR \- Cleans up unused gems in your bundler directory
--- a/bundler/lib/bundler/man/bundle-config.1
+++ b/bundler/lib/bundler/man/bundle-config.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CONFIG" "1" "March 2021" "" ""
+.TH "BUNDLE\-CONFIG" "1" "April 2021" "" ""
 .
 .SH "NAME"
 \fBbundle\-config\fR \- Set bundler configuration options
@@ -197,6 +197,9 @@ The following is a list of all configuration keys and their purpose\. You can le
 .
 .IP "\(bu" 4
 \fBfrozen\fR (\fBBUNDLE_FROZEN\fR): Disallow changes to the \fBGemfile\fR\. When the \fBGemfile\fR is changed and the lockfile has not been updated, running Bundler commands will be blocked\. Defaults to \fBtrue\fR when \fB\-\-deployment\fR is used\.
+.
+.IP "\(bu" 4
+\fBgem\.github_username\fR (\fBBUNDLE_GEM__GITHUB_USERNAME\fR): Sets a GitHub username or organization to be used in \fBREADME\fR file when you create a new gem via \fBbundle gem\fR command\. It can be overriden by passing an explicit \fB\-\-github\-username\fR flag to \fBbundle gem\fR\.
 .
 .IP "\(bu" 4
 \fBgem\.push_key\fR (\fBBUNDLE_GEM__PUSH_KEY\fR): Sets the \fB\-\-key\fR parameter for \fBgem push\fR when using the \fBrake release\fR command with a private gemstash server\.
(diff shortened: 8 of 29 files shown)


deivid-rodriguez:
> Firstly, it's not documented that we can fill in [USERNAME] this way.

I would say that the solution to this would be to document it?

> Secondly, the config value is set by some tools implicitly, and is not documented in git-config manpage.

Which tools set this? Maybe using a custom git config is not the best option here, but I don't like a CLI flag either since users will always want the same value for it. So configuration is better for this than a CLI flag that you have to specifiy every time.

I listen to ideas here, but my initial take on this is to document this and leave things as they are.

okuramasafumi:
> I would say that the solution to this would be to document it?

Yes it should be a quick solution. I personally prefer putting settings about rubygems/bundler in its own config file over putting them in git config.

> Which tools set this?

I heard `hub` command sets this.

> since users will always want the same value for it. So configuration is better for this than a CLI flag that you have to specifiy every time.

I agree, so this PR adds the command line option but doesn't force users to fill in the value, and lets users store the value in bundler config after they fill it in for the first time so that they don't have to specify everytime.

deivid-rodriguez:
@okuramasafumi I believe you have a point, and also you're right that this is not only a CLI flag but also a permanent configuration. So I'm ok with this in general.

My remaining doubt is whether it makes sense to "dirty" the CLI that will be only used once, and whether it's better to provide this as a configuration setting only. What do you think?

deivid-rodriguez:
Something else I'm noticing, although not directly related to this PR, is that our gem generator assumes the code will be hosted on GitHub, so I believe we could provide a more useful default template by prefilling gemspec fields such as `homepage`, `homepage_uri`, `source_code_uri` and `changelog_uri` with "github values".

hsbt:
> My remaining doubt is whether it makes sense to "dirty" the CLI that will be only used once, and whether it's better to provide this as a configuration setting only. What do you think?

I agree with this. It's better to add the alternative options to bundle config.

okuramasafumi:
> My remaining doubt is whether it makes sense to "dirty" the CLI that will be only used once, and whether it's better to provide this as a configuration setting only. What do you think?

I see that although we might want to change our test framework/CI tool sometimes, we might never change our GitHub username. Adding a new flag for this option can be "dirty".

I'll make my PR like:

* Remove CLI flag
* Fetch config value from git config first not to break backward compatibility
* When config value in git config is missing, try to fetch the value from bundle config
* When they are all missing, default to `[USERNAME]`

deivid-rodriguez:
Yeah, but I'm still in doubt because looking at where this is used (code hosting URLs essentially), you might want to create a gem under your employer's github organization, for example. So that would be a use case.

okuramasafumi:
Ah that would be a use case. I think we can leave the CLI unchanged and if users want a CLI flag we can add it later.

okuramasafumi:
Some specs fail in CI but not in my local env (I tried reusing seed but the result was the same).
It says `bundle!` command is not found, both in Ubuntu and Windows platform.
They happen in specs I added. However, there are the same `bundle!` method used in similar specs and they are all green.
I inspected and the only difference is that they have `readline` flag in its `describe`.
I don't know how to fix it. What can I do?

deivid-rodriguez:
You need to rebase your PR on top of the latest master and replace `bundle!` with `bundle`, since we have removed the former.

okuramasafumi:
@deivid-rodriguez Thanks, now all specs are green!
Should I squash three commits into one?

deivid-rodriguez:
Don't squash commits yet, since I'm still thinking whether it's better to introduce `--github-username` as a flag directly. I plan to create a gem under the `rubygems` organization soon, and I think I would use the flag :sweat_smile:. Also, I believe this would be the first `gem.<configuration>` that doesn't have a corresponding CLI flag, so adding the floag might save users from some surprises. @hsbt Any thoughts?

bundler/lib/bundler/cli.rb
```diff
@@ -577,7 +577,6 @@ def viz
                          :desc => "Generate a test directory for your library, either rspec, minitest or test-unit. Set a default with `bundle config set gem.test (rspec|minitest|test-unit)`."
     method_option :ci, :type => :string, :lazy_default => Bundler.settings["gem.ci"] || "",
                        :desc => "Generate CI configuration, either GitHub Actions, Travis CI, GitLab CI or CircleCI. Set a default with `bundle config set gem.ci (github|travis|gitlab|circle)`"
-
```
deivid-rodriguez, line 580: Can you restore this blank line?

bundler/man/bundle-check.1.txt
```diff
@@ -9,8 +9,8 @@ SYNOPSIS
        bundle check [--dry-run] [--gemfile=FILE] [--path=PATH]
 
 DESCRIPTION
-       check  searches the local machine for each of the gems requested in the
-       Gemfile. If all gems are found, Bundler prints a  success  message  and
+       check searches the local machine for each of the gems requested in  the
+       Gemfile.  If  all  gems are found, Bundler prints a success message and
```
deivid-rodriguez, line 13: I hate when man pages updates do this... I should fix that.

bundler/man/bundle-config.ronn
```diff
@@ -193,6 +193,9 @@ learn more about their operation in [bundle install(1)](bundle-install.1.html).
    Disallow changes to the `Gemfile`. When the `Gemfile` is changed and the
    lockfile has not been updated, running Bundler commands will be blocked.
    Defaults to `true` when `--deployment` is used.
+* `gem.github_username` (`BUNDLE_GEM__GITHUB_USERNAME`):
+   Sets a username for GitHub. This will be used in `README` file when you
```
deivid-rodriguez, line 197: I would say 
```suggestion
   Sets a GitHub username or organization to be used in `README` file when you
```

deivid-rodriguez:
Any feedback on my latest comment? Should we go with the original `--github-username` flag approach?

okuramasafumi:
@deivid-rodriguez Sorry for the late reply, I noticed that there are so many files to be rebased so I'll do that later.

For the spec of the flag, I'm not sure how many people add the flag just once.
If people need to add this flag many times to switch from personal to organization and vice versa, it doesn't make sense anymore.

deivid-rodriguez:
> @deivid-rodriguez Sorry for the late reply, I noticed that there are so many files to be rebased so I'll do that later.

I can also do that for you, no problem.

> For the spec of the flag, I'm not sure how many people add the flag just once.
> If people need to add this flag many times to switch from personal to organization and vice versa, it doesn't make sense anymore.

Even if people need to add the flag many times, it's still useful, no? I mean, if you create 80% of gems under your organization's account, and 20% under your personal account, it still sounds like useful to have both a configuration and a CLI flag, so that you can configure `bundle gem` to normally use your organization, but also be able to pass `--github-username` when you want to create a gem under your personal account.

okuramasafumi:
> if you create 80% of gems under your organization's account, and 20% under your personal account, it still sounds like useful to have both a configuration and a CLI flag, so that you can configure `bundle gem` to normally use your organization, but also be able to pass `--github-username` when you want to create a gem under your personal account.

I consider this situation and suspect that whether we should overwrite configuration value when users give the flag second time.
Maybe we should save the value from the flag to the config value only when the config value is empty. In other words, when users already have the value in config, the flag changes it temporarily, not parmanently.

deivid-rodriguez:
Mmmmm, is that no how it currently works? If it's not, I think that's definitely how it should work!

deivid-rodriguez:
I rebased the PR. If we decide to keep the flag, it'd be a matter of reverting the second commit.

okuramasafumi:
Thank you!
One if the checks failed due to something related to RuboCop, but I could not resolve it :(

deivid-rodriguez:
Sorry about that, it should be fixed by https://github.com/rubygems/rubygems/pull/3946. We can rebase again after that's merged.

okuramasafumi:
Cool, and I'll add one more commit to make the flag to change the config temporarily if there's already the config value.

deivid-rodriguez:
@okuramasafumi I double checked and this is the current behavior, so I don't think there's anything to change.

```
$ bundle config
Settings are listed in order of priority. The top value will be used.

$ bundle gem foo
Creating gem 'foo'...
Do you want to generate tests with your gem?
Future `bundle gem` calls will use your choice. This setting can be changed anytime with `bundle config gem.test`.
Enter a test framework. rspec/minitest/test-unit/(none): rspec
Do you want to set up continuous integration for your gem? Supported services:
* CircleCI:       https://circleci.com/
* GitHub Actions: https://github.com/features/actions
* GitLab CI:      https://docs.gitlab.com/ee/ci/
* Travis CI:      https://travis-ci.org/

Future `bundle gem` calls will use your choice. This setting can be changed anytime with `bundle config gem.ci`.
Enter a CI service. github/travis/gitlab/circle/(none): github
Do you want to license your code permissively under the MIT license?
This means that any other developer or company will be legally allowed to use your code for free as long as they admit you created it. You can read more about the MIT license at https://choosealicense.com/licenses/mit. y/(n): 
Do you want to include a code of conduct in gems you generate?
Codes of conduct can increase contributions to your project by contributors who prefer collaborative, safe spaces. You can read more about the code of conduct at contributor-covenant.org. Having a code of conduct means agreeing to the responsibility of enforcing it, so be sure that you are prepared to do that. Be sure that your email address is specified as a contact in the generated code of conduct so that people know who to contact in case of a violation. For suggestions about how to enforce codes of conduct, see https://bit.ly/coc-enforcement. y/(n): 
Do you want to add rubocop as a dependency for gems you generate?
RuboCop is a static code analyzer that has out-of-the-box rules for many of the guidelines in the community style guide. For more information, see the RuboCop docs (https://docs.rubocop.org/en/stable/) and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide). y/(n): 
      create  foo/Gemfile
      create  foo/lib/foo.rb
      create  foo/lib/foo/version.rb
      create  foo/foo.gemspec
      create  foo/Rakefile
      create  foo/README.md
      create  foo/bin/console
      create  foo/bin/setup
      create  foo/.gitignore
      create  foo/.rspec
      create  foo/spec/spec_helper.rb
      create  foo/spec/foo_spec.rb
      create  foo/.github/workflows/main.yml
Initializing git repo in /home/deivid/Code/playground/foo
Gem 'foo' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html

$ bundle config
Settings are listed in order of priority. The top value will be used.
gem.test
Set for the current user (/home/deivid/.bundle/config): "rspec"

gem.ci
Set for the current user (/home/deivid/.bundle/config): "github"

gem.mit
Set for the current user (/home/deivid/.bundle/config): false

gem.coc
Set for the current user (/home/deivid/.bundle/config): false

gem.rubocop
Set for the current user (/home/deivid/.bundle/config): false

$ bundle gem foo --test=minitest
Creating gem 'foo'...
    conflict  foo/Gemfile
Overwrite /home/deivid/Code/playground/foo/Gemfile? (enter "h" for help) [Ynaqdhm] deivid@chaba ~/Code/playground $ bundle gem foo2 --test=minitest
Creating gem 'foo2'...
      create  foo2/Gemfile
      create  foo2/lib/foo2.rb
      create  foo2/lib/foo2/version.rb
      create  foo2/foo2.gemspec
      create  foo2/Rakefile
      create  foo2/README.md
      create  foo2/bin/console
      create  foo2/bin/setup
      create  foo2/.gitignore
      create  foo2/test/test_helper.rb
      create  foo2/test/test_foo2.rb
      create  foo2/.github/workflows/main.yml
Initializing git repo in /home/deivid/Code/playground/foo2
Gem 'foo2' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html

$ bundle config
Settings are listed in order of priority. The top value will be used.
gem.test
Set for the current user (/home/deivid/.bundle/config): "rspec"

gem.ci
Set for the current user (/home/deivid/.bundle/config): "github"

gem.mit
Set for the current user (/home/deivid/.bundle/config): false

gem.coc
Set for the current user (/home/deivid/.bundle/config): false

gem.rubocop
Set for the current user (/home/deivid/.bundle/config): false
```

okuramasafumi:
Now I see. Then, should I revert the second commit?

deivid-rodriguez:
Yeah, I think it's a valid use case after all, and it keeps things more consistent. Sorry for the extra work :pray:.

deivid-rodriguez:
Just merged the PR I mentioned, so rebasing again should get this PR green.

deivid-rodriguez:
There's still some issues in the specs, since we removed the `bundle!` helper. You should use `bundle` now.

deivid-rodriguez:
Also, I'm reading the specs and I wonder. Shouldn't the new option/configuration take precedence over the old `github.user` git configuration?

deivid-rodriguez:
@okuramasafumi Are you still interested in this PR?

deivid-rodriguez:
Ok, I addressed all concerns myself, and this one should be ready now!

okuramasafumi:
@deivid-rodriguez Thank you for your hard work!

PR #3698, opened 2020-06-07
Fix ruby-head failing bundler tests

deivid-rodriguez:
# Description:

This PR fixes bundler CI run against ruby-head, which had been failing for a while, [like this](https://github.com/rubygems/rubygems/actions/runs/127258555). The problem was that the `sudo` version installed on ubuntu 16.04 doesn't not accept `--preserve-env` with an argument, and our `sudo` specs use that.

I fixed it by upgrading the workflow to ubuntu 18.04.

In addition to that, I pinned all base OS versions of our workflows to avoid further inconsistencies and to prevent underlying changes in the github actions "latest" alias breaking us.

Proof of fixed build is: https://github.com/rubygems/rubygems/runs/747745016?check_suite_focus=true.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   daily_bundler:
-    runs-on: ubuntu-16.04
+    runs-on: ubuntu-18.04
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


hsbt:
👍

PR #3700, opened 2020-06-08
Remove vendored fileutils

deivid-rodriguez:
# Description:

Just trigerring CI to double check whether this is something we can actually do safely.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Ok, so my initial investigation here:

* `fileutils` does not seem loaded on `bundler/setup`.
* If we stop vendoring it though, there will be some redefinition warnings when you point to git gems that load `FileUtils` inside their gemspecs.

I'd like to think for a bit about how to fix the second issue.

martinemde:
Closing old PR. We still vendor fileutils.

PR #3702, opened 2020-06-09
Remove all "bang helpers" from our specs

deivid-rodriguez:
# Description:

After #3685, all of our test helpers that shell out raise by default, so we no longer need a different set of helpers to achieve that.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -292,7 +292,7 @@ module Spec
     def install_gem(path)
       raise "OMG `#{path}` does not exist!" unless File.exist?(path)
 
-      gem_command! "install --no-document --ignore-dependencies '#{path}'"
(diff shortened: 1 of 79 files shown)


bundler/spec/support/helpers.rb
```diff
@@ -23,12 +23,6 @@ def reset_paths!
       Gem.clear_paths
     end
 
-    def self.bang(method)
-      define_method("#{method}!") do |*args, &blk|
-        send(method, *args, &blk)
-      end
-    end
-
```
deivid-rodriguez, line 31: @simi With a quite a few changes to only use the "non bang methods", we can now kill this (now dummy) helper.

simi: :+1:

PR #3703, opened 2020-06-10
Adapt release tasks to new repo

deivid-rodriguez:
# Description:

This is some initial work to adapt bundler release tasks to the new repository layout. My idea is that bundler releases reuse rubygems stable branches, so from the "3.2" stable branch, we will release rubygems 3.2.0 and bundler 2.2.0. Any further patch level releases of bundler and rubygems are also done from this stable branch.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
I updated the release playbooks to explain the new branch naming convention, and to explain the release process of a minor release of bundler as I did it last time.

deivid-rodriguez:
Pushed a few fixes and more improvements to the playbook.

I wasn't sure how to use a specific convention for `bundler` tags. Finally went with adding `Bundler::GemHelper.tag_prefix=` that let's you add a prefix for the tags created by `rake release`, and set it to `Bundler::GemHelper.tag_prefix = "bundler-"`.

Maybe I should extract that to a separate PR since it's actually a new feature.

PR #3704, opened 2020-06-10
Use space before block parameters everywhere

deivid-rodriguez:
# Description:

This commit changes our internal style and also the style used in the
generated `Gemfile` by our gem generator.

This style is used everywhere, end users have opened PRs in the past to
tweak the generated style, and I keep doing the same style mistake that
doesn't appease our current configuration, so this change also maximizes
my own happiness.

This is a migration of https://github.com/rubygems/bundler/pull/7645 into the new repo.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -68,6 +68,10 @@ Layout/IndentationWidth:
 Layout/SpaceAroundOperators:
   Enabled: true
 
+Layout/SpaceInsideBlockBraces:
+  Enabled: true
+  SpaceBeforeBlockParameters: true
+
 Layout/SpaceInsideParens:
   Enabled: true
 
--- a/Rakefile
+++ b/Rakefile
@@ -405,7 +405,7 @@ end
 
 desc "Update the manifest to reflect what's on disk"
 task :update_manifest do
-  File.open('Manifest.txt', 'w') {|f| f.puts(Rubygems::ProjectFiles.all.sort) }
+  File.open('Manifest.txt', 'w') { |f| f.puts(Rubygems::ProjectFiles.all.sort) }
 end
 
 desc "Check the manifest is up to date"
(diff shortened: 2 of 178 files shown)


hsbt:
I never use `SpaceBeforeBlockParameters: true` style over the 20+ years in my Ruby life. I approved `Layout/SpaceInsideBlockBraces` with space of end block.

I disagree to apply with `SpaceBeforeBlockParameters: true`.

Note: Rails also don't use this https://github.com/rails/rails/blob/master/.rubocop.yml#L159

indirect:
For whatever it's worth, I also personally don't put spaces before block parameters. That said, I don't spend much time writing code anymore, so I think active committers should have more of a vote than me.

More than any particular rule, though, I would love for us to set up editor or git hooks for the project that automatically apply the configuration on commit so I never have to think about it again.

deivid-rodriguez:
Sorry, I misunderstood from the absence of opinions from other team members in the referenced PR (except for @indirect, who said he doesn't use it, but he could live with it) that you'd be ok with this.

I said this: https://github.com/rubygems/bundler/pull/7645#issuecomment-592034257, then the PR sat there for a month without any objections, and then @hsbt invited the OP to reopen it in the new repo. Hence my misunderstanding.

> Note: Rails also don't use this https://github.com/rails/rails/blob/master/.rubocop.yml#L159

Just a small clarification, to prevent future readers from confusion: the above is incorrect. Rails uses a space before block parameters. They don't configure it explicitly because once you enable this cop, it's the default. So you only need the configuration if you want it false.

> More than any particular rule, though, I would love for us to set up editor or git hooks for the project that automatically apply the configuration on commit so I never have to think about it again.

Yeah, I have my editor configured for rubocop, but for some reason it broke for bundler after the repo merge, I need to investigate why :grimacing:.

Anyways, closing this!

hsbt:
> and then @hsbt invited the OP to reopen it in the new repo.

I apologize for my confusing behavior. I only continue to discuss it to new repo.

> the above is incorrect. Rails uses a space before block parameters.

Ah, Your comments is right. I misunderstood.

Thank you for applying to my request.

deivid-rodriguez:
No problem @hsbt, it's all good!

PR #3709, opened 2020-06-11
Use space inside block braces everywhere

deivid-rodriguez:
# Description:

In #3704 @hsbt refused to switch to using a space before block parameters in bundler's code base.

I'm sad about that, but I still prefer any style that's consistent over my own preferences.

So, instead of changing `bundler` to use the style that I like, I propose to change `rubygems` to use the style that I don't like, so that at least we have a consistent code base in this regard.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -68,6 +68,10 @@ Layout/IndentationWidth:
 Layout/SpaceAroundOperators:
   Enabled: true
 
+Layout/SpaceInsideBlockBraces:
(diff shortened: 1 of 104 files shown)


deivid-rodriguez:
I got @hsbt and @indirect on my side here, so this should be a no brainer :stuck_out_tongue:.

deivid-rodriguez:
No problem!

PR #3713, opened 2020-06-12
Remove random code

deivid-rodriguez:
# Description:

There's some random code at the end of the `specification.rb` file that doesn't make any sense to me, and it's confusing, so I'm going to kill it.
 
# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2665,6 +2665,3 @@ class Gem::Specification < Gem::BasicSpecification
   end
 
 end
-
-# DOC: What is this and why is it here, randomly, at the end of this file?
-Gem.clear_paths


deivid-rodriguez:
For what it's worth, the call was introduced 10 years ago [here](https://github.com/rubygems/rubygems/commit/10f1094f81fdb6dfb680655b684d73e226de5400), and the intention was to fix a performance regression: https://bugs.ruby-lang.org/issues/4962.

I don't see how removing this call could make things slower, so I think this should be fine is CI is ok.

simi:
:eyes: It seems there's some benchmark code you can test at related issue - https://bugs.ruby-lang.org/issues/4962.

deivid-rodriguez:
Yep, tried those and they show no significant differences.

PR #3715, opened 2020-06-12
Remove encoding magic comments

deivid-rodriguez:
# Description:

This PR removes magic encoding comments from rubygems code base in another small step to reconcile bundler & rubygems code styles.

These comments are no longer needed since ruby 2.0.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -101,6 +101,11 @@ Performance/StartWith:
 Performance/StringReplacement:
   Enabled: true
 
+Style/Encoding:
+  Enabled: true
+  Exclude:
+    - test/rubygems/specifications/foo-0.0.1-x86-mswin32.gemspec
+
 Style/MethodCallWithoutArgsParentheses:
   Enabled: true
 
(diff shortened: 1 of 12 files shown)


deivid-rodriguez:
I'm very confused that the diff is showing that I'm removing the `test/rubygems/specifications/foo-0.0.1-x86-mswin32.gemspec` file, but I don't think I intentionally removed it :thinking:.

deivid-rodriguez:
Aha! So it's not removing that file, but removing the encoding comment from it. And it sounds like it's necessary in that case according to our tests, will need to look in more detail.

simi:
I think you can leave that file as is or unpack the gem, remove the comment and pack again.

deivid-rodriguez:
That's not a gem, it's a gemspec. It's just detected as binary because it has a binary character inside. I guess I can add an exclusion for that file and forget about it.

simi:
Ahh, sorry. :+1:

PR #3718, opened 2020-06-12
Restore sequential install by default on Windows

deivid-rodriguez:
# Description:

Since we [enabled parallel installation by default](https://github.com/rubygems/rubygems/pull/3393), we've fixed all the issues related to it that have come up, except for [a Windows issue](https://github.com/rubygems/rubygems/issues/3448) that we haven't yet figured out. This issue is hit by our specs on a daily basis and there's no reason to believe that it won't be hit by end users in a similar way.

So, both to stop the testing flakyness and to prevent regressions in the default behavior on Windows, I'd rather leave the default as it was before on Windows for now.

Closes https://github.com/rubygems/rubygems/issues/3448.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/installer.rb
+++ b/bundler/lib/bundler/installer.rb
@@ -202,8 +202,6 @@ module Bundler
         return jobs
       end
 
-      return 1 unless can_install_in_parallel?
-
       Bundler.settings[:jobs] || processor_count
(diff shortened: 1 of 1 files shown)


simi:
:+1:

PR #3720, opened 2020-06-13
Add Gem.disable_system_update_message to disable gem update --system if needed.

simi:
This is initial take on possibility to disable `gem update --system` useful for distribution maintainers.

For Debian, this can be set at https://salsa.debian.org/ruby-team/rubygems-integration/-/blob/39d3a082f568ec0b6c9acd10e54a4a31a08bd8b5/lib/rubygems/defaults/operating_system.rb#L56. End-users can disable this by providing `DEBIAN_DISABLE_RUBYGEMS_INTEGRATION` as they can do that already for the rest of integration patches.

For Fedora, this can be set at https://src.fedoraproject.org/rpms/rubygems/blob/master/f/operating_system.rb. Not sure if any env variable is needed to make users able to disable this.

@terceiro (Debian ruby maintainer) @voxik (Fedora ruby maintainer), would this be useful for you :question: 

### example

```ruby
# operating_system.rb
if Gem.respond_to?(:disable_system_update_message=)
  Gem.disable_system_update_message = 'RubyGems is updated by system package manager. Please use "abc update rubygems" instead.'
end
```

```
[retro@retro  rubygems (disable-update *%=)]❤ gem update --system
ERROR:  RubyGems is updated by system package manager. Please use "abc update rubygems" instead.
```

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -264,7 +264,7 @@ command to remove old versions.
   def update_rubygems
     if Gem.disable_system_update_message
       alert_error Gem.disable_system_update_message
-      return
+      terminate_interaction 1
     end
 
     check_update_arguments
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -262,6 +262,11 @@ command to remove old versions.
   # Update RubyGems software to the latest version.
 
   def update_rubygems
+    if Gem.disable_system_update_message
+      alert_error Gem.disable_system_update_message
+      return
+    end
+
     check_update_arguments
 
     version, requirement = rubygems_target_version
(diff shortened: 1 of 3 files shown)


terceiro:
>  would this be useful for you ?

yes

lib/rubygems/commands/update_command.rb
```diff
@@ -262,6 +262,11 @@ def update_gems(gems_to_update)
   # Update RubyGems software to the latest version.
 
   def update_rubygems
+    if Gem.disable_system_update_message
+      alert_error Gem.disable_system_update_message
+      return
```
terceiro, line 267: I would expect the update attempt to fail (i.e. exit non-zero)

simi: Ahh, right. Do you have any preferred exit code for this?

voxik:
LGTM

lib/rubygems/commands/update_command.rb
simi: fixed @terceiro, it will return 1 now

PR #3722, opened 2020-06-14
Make a spec helper more portable and improve some skip messages

deivid-rodriguez:
# Description:

I found this commit buried in my local git repo, and it claims to fix a couple of spec failures on Windows.

The commit makes sense to me, so let's see if it actually fixes the failures.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Both tests still fail, but the change in the test helper seems more portable, right?

PR #3725, opened 2020-06-15
Enforce ruby-core integration workflow

deivid-rodriguez:
# Description:

Keeping in sync with ruby-core is tough :sweat_smile:, because we not only need to backport stuff periodically, but also to forward-port the stuff that the core team changes directly in their vendored code to keep their CI green.

This PR tries to minimize the number of changes the ruby-core does directly inside our code, by running the specs the way they run them and ensuring they are green.

A problem that was mentioned in 2de34ac14691f95ab2472fd479b9a60d0b2c7e4b when this was introduced was that sometimes ruby will fail to compile. To fix that issue, I build the latest reference provided by https://github.com/ruby/setup-ruby, which should always build fine.

Build should fail now, but should pass after https://github.com/ruby/ruby/pull/3230 and https://github.com/ruby/ruby/pull/3213 are merged.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This is green now! :tada:

I'd expect this to save us some trouble with future bundler & rubygems imports into ruby-core.

PR #3728, opened 2020-06-16
Rename hidden templates

deivid-rodriguez:
# Description:

Because they don't play nice with ruby-core's gitignore and cause issues with ruby-core integration. See https://github.com/ruby/ruby/pull/3230.

Also, because it's consistent with other templates such as `gitignore.tt`.

This PR should fix CI for bundler at #3725.

# Tasks:

- [ ] Write tests


hsbt:
👍 It's better changes for human too.

deivid-rodriguez:
Thanks both! :heart:

PR #3729, opened 2020-06-16
Allow tests without rake

deivid-rodriguez:
# Description:

While integrating the latest rubygems HEAD commits into ruby-core I noticed an error like the following:

```
/home/runner/work/ruby/ruby/src/test/rubygems/test_gem_package_task.rb: cannot load such file -- rake/packagetask (Test::Unit::ProxyError)
```

I believe this is because ruby-core doesn't run tests through `rake` but directly through `ruby`, and the environment where tests run usually does not even have `rake` installed.

So I changed test code to only `warn` about `rake` not being installed in this case, just like we do for `openssl` in `test_gem_remote_fetcher` tests.

Also took the chance to simplify a bunch of other related stuff.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -1534,24 +1534,4 @@ Also, a list:
 
 end
 
-# require dependencies that are not discoverable once GEM_HOME and GEM_PATH
-# are wiped
-begin
-  gem 'rake'
-rescue Gem::LoadError
-end
-
-begin
-  require 'rake/packagetask'
-rescue LoadError
-end
-
-begin
-  gem 'rdoc'
-  require 'rdoc'
-
-  require 'rubygems/rdoc'
-rescue LoadError, Gem::LoadError
-end
-
 require 'rubygems/test_utilities'
--- a/test/rubygems/test_gem_remote_fetcher.rb
+++ b/test/rubygems/test_gem_remote_fetcher.rb
@@ -5,8 +5,7 @@ require 'webrick'
 begin
   require 'webrick/https'
 rescue LoadError => e
-  raise unless (e.respond_to?(:path) && e.path == 'openssl') ||
-               e.message =~ / -- openssl$/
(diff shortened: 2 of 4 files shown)


lib/rubygems/test_case.rb
```diff
@@ -1534,24 +1534,4 @@ def self.key_path(key_name)
...
-  gem 'rake'
-rescue Gem::LoadError
-end
-
-begin
-  require 'rake/packagetask'
-rescue LoadError
-end
-
-begin
```
bronzdoc, line 1549: What this removal have to do with rake?

deivid-rodriguez: Nothing, they were other unnecessary gem activations and went ahead and removed them too. I'll remove this part from the PR.

test/rubygems/test_gem_remote_fetcher.rb
```diff
@@ -5,8 +5,7 @@
 begin
   require 'webrick/https'
 rescue LoadError => e
-  raise unless (e.respond_to?(:path) && e.path == 'openssl') ||
-               e.message =~ / -- openssl$/
+  raise unless e.path == 'openssl'
```
deivid-rodriguez, line 8: Do you want me to also remove this change, since it's also unrelated to `rake`?

lib/rubygems/test_case.rb
bronzdoc: I was just curious, no big deal

test/rubygems/test_gem_remote_fetcher.rb
bronzdoc: No big deal

lib/rubygems/test_case.rb
deivid-rodriguez: Ok, thanks! I'll merge this as is then!

PR #3731, opened 2020-06-16
Ship default .rubocop.yml

utkarsh2102:
### Description:

Currently, there is no `.rubocop.yml` shipped by default.
So when a user runs `rubocop` after creating a new gem via `bundle gem foo`, it throws a bunch of offenses.

With the default `.rubocop.yml` present, the number of those offenses significantly reduce by 25.

### What was the end-user or developer problem that led to this PR?

A bunch of offenses thrown by rubocop since there's no default `.rubocop.yml` file.

### What is your fix for the problem, implemented in this PR?

This PR adds a default `.rubocop.yml` file. 
______________

### Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -79,7 +79,6 @@ module Bundler
       ]
 
       templates.merge!("gitignore.tt" => ".gitignore") if Bundler.git_present?
-      templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
 
       if test_framework = ask_and_set_test_framework
         config[:test] = test_framework
@@ -149,6 +148,7 @@ module Bundler
         "and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide).")
         config[:rubocop] = true
         Bundler.ui.info "RuboCop enabled in config"
+        templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
       end
 
       templates.merge!("exe/newgem.tt" => "exe/#{name}") if config[:exe]
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -79,6 +79,7 @@ module Bundler
       ]
 
       templates.merge!("gitignore.tt" => ".gitignore") if Bundler.git_present?
+      templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
 
       if test_framework = ask_and_set_test_framework
         config[:test] = test_framework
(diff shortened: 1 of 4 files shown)


bronzdoc:
Thanks for the PR @utkarsh2102!

What other cops are need to reduce it to 0?

utkarsh2102:
Hi @bronzdoc,

> What other cops are need to reduce it to 0?

The other things that can bring down the offenses are mostly add frozen string literal comment, using double-quotes in some places, freezing a constant, and some other small things.
I can create another PR if you'd like to get them fixed, too? :smile:

utkarsh2102:
Hi @deivid-rodriguez,
I see your #3728 to remove hidden files, d'you want me to do the same here?
I suppose so? :see_no_evil:

deivid-rodriguez:
Yes, please!

bundler/lib/bundler/templates/newgem/.rubocop.yml.tt
```diff
@@ -0,0 +1,9 @@
+# This is the default RuboCop configuration.
+
```
deivid-rodriguez, line 2: I would remove the above comment since it might have the effect of leading users into thinking that the actual default rubocop configuration is not being applied.

bronzdoc:
@utkarsh2102 I think if we are adding a default its better if we add all the cops that will reduce the offenses to 0 in this PR.

bundler/lib/bundler/templates/newgem/.rubocop.yml.tt
utkarsh2102: Aye, aye. Shall fix.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -79,6 +79,7 @@ def run
       ]
 
       templates.merge!("gitignore.tt" => ".gitignore") if Bundler.git_present?
+      templates.merge!(".rubocop.yml.tt" => ".rubocop.yml")
```
deivid-rodriguez, line 82: This file should only be added if the `--rubocop` flag is passed or enabled in configuration. So, move this line under the relevant condition below and adapt the specs accordingly.

utkarsh2102: :sob: 
Shall do!

deivid-rodriguez: Thanks! :purple_heart:

utkarsh2102: Should be done! :smile:

utkarsh2102:
Thanks!
@bronzdoc, I can create another PR (soon!) to address those issues as well :smile:

deivid-rodriguez:
Thanks @utkarsh2102!!

Issue #3381, opened 2016-05-03
binstubs support on windows


lynncyrin:
From: https://github.com/bundler/bundler-features/issues/92

Moving this issue from the bundler-featured repo, cc @elovelan in case you still wanted to submit a PR

deivid-rodriguez:
Copying here the original issue text, since it's helpful:

> Unless you're running under cygwin, msys or another POSIX compatibility layer, Windows doesn't know how to interpret `#!/usr/bin/env ruby`.

> I would be happy to implement this but want to get some feedback first. I should be able to borrow most of [the code from RubyGems](https://github.com/rubygems/rubygems/blob/40e59d5619ac176cb7c0d76829f817b63fe145e1/lib/rubygems/installer.rb#L436).

Marking this as "help wanted" :)

deivid-rodriguez:
Fixing this would also allow us to unskip most (if not all) of the Windows skips in `spec/commands/binstubs_spec.rb` and also fix the spec failures that I expected to fix in https://github.com/rubygems/rubygems/pull/3722.

deivid-rodriguez:
Also, an alternative idea suggested by @janpio in #3353:

> Although #3381 suggested using the Rubygems implementation as a template, it was much easier to use how e.g. Cordova CLI does it: Just create a command.cmd file with content @ruby "%~dpn0" %* (for us with ruby here). Fixed the tests and has minimal code.

deivid-rodriguez:
And yet another idea from @janpio coming from #3351:

> Indeed, a POC that adds `ruby` if the command can not be found "solves" the problem here:
> ```
>   rescue Errno::ENOENT
>      if Gem.win_platform? then
>        puts "§§§ add ruby in front of command for Windows"
>        args.unshift('ruby')
>        kernel_exec(*args)
>      end
>      Bundler.ui = ui
>```
>(line 62, exec.rb)

MSP-Greg:
@deivid-rodriguez

At some point, the idea was to have a 'dual purpose' binstub script that would run from either a bash shell or a Windows shell.  Ruby master (and maybe 2.7) use this, and it has an `.cmd` extension.  So, what happens is a bash shell works with `bundle.cmd`, but not `bundle`.  It's a good idea, except for cross-platform testing.

[...]

PR #3736, opened 2020-06-17
Bundler - fix executable binstubs (project binstubs) on Windows

MSP-Greg:
# Description:

Generate Windows specific project binstubs, and remove related spec skips.

After removing global skip, one additional skip is needed, see 2nd commit.  This adds 33 additional specs, so test time increases...

Discussion in https://github.com/rubygems/rubygems/issues/3381

Fix `Bundler::Installer#generate_bundler_executable_stubs` & `Bundler::Installer#generate_standalone_bundler_executable_stubs`, write files with 'LF` for EOL.
______________

# Tasks:

- [x]  Fix tests

### DIFF
--- a/bundler/spec/commands/binstubs_spec.rb
+++ b/bundler/spec/commands/binstubs_spec.rb
@@ -1,10 +1,6 @@
 # frozen_string_literal: true
 
 RSpec.describe "bundle binstubs <gem>" do
-  before do
-    skip "https://github.com/rubygems/bundler/issues/6894" if Gem.win_platform?
-  end
-
   context "when the gem exists in the lockfile" do
     it "sets up the binstub" do
       install_gemfile <<-G
@@ -79,6 +75,7 @@ RSpec.describe "bundle binstubs <gem>" do
 
     context "when generating bundle binstub outside bundler" do
       it "should abort" do
+        skip "Unknown issue" if Gem.win_platform?
         install_gemfile <<-G
           source "#{file_uri_for(gem_repo1)}"
           gem "rack"
--- a/bundler/spec/commands/binstubs_spec.rb
+++ b/bundler/spec/commands/binstubs_spec.rb
@@ -75,7 +75,6 @@ RSpec.describe "bundle binstubs <gem>" do
 
     context "when generating bundle binstub outside bundler" do
       it "should abort" do
-        skip "Unknown issue" if Gem.win_platform?
         install_gemfile <<-G
           source "#{file_uri_for(gem_repo1)}"
           gem "rack"
@@ -285,8 +284,7 @@ RSpec.describe "bundle binstubs <gem>" do
         G
 
         bundle "binstubs rack --shebang jruby"
-
-        expect(File.open(bundled_app("bin/rackup")).gets).to eq("#!/usr/bin/env jruby\n")
+        expect(File.readlines(bundled_app("bin/rackup")).first).to eq("#!/usr/bin/env jruby\n")
       end
     end
   end
--- a/bundler/lib/bundler/installer.rb
+++ b/bundler/lib/bundler/installer.rb
@@ -135,12 +135,17 @@ module Bundler
           next
         end
 
-        File.open(binstub_path, "w", 0o777 & ~File.umask) do |f|
-          if RUBY_VERSION >= "2.6"
-            f.puts ERB.new(template, :trim_mode => "-").result(binding)
-          else
-            f.puts ERB.new(template, nil, "-").result(binding)
-          end
+        mode = Bundler::WINDOWS ? "wb:UTF-8" : "w"
+        content = if RUBY_VERSION >= "2.6"
+          ERB.new(template, :trim_mode => "-").result(binding)
+        else
+          ERB.new(template, nil, "-").result(binding)
+        end
+
+        File.write(binstub_path, content, :mode => mode, :perm => 0o777 & ~File.umask)
+        if Bundler::WINDOWS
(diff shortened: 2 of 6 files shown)


deivid-rodriguez:
Thanks @MSP-Greg, this makes sense and I think it's ok to generate the extra file if `bundle binstubs` is run on Windows.

One of the two places you changed fixes stuff for `bundle install --standalone`, so I believe you can unskip stuff also in `bundler/spec/install/gems/standalone_spec.rb`. Actually, everywhere where `bundle binstubs` is used in the specs should be fixed, I believe you can unskip all the specs tagged as `skip "exec format error" if Gem.win_platform?`.

MSP-Greg:
JFYI, bash uses the shebang line.  Windows uses file extensions for the equivalent to that, and also for argument passing, an example being launching Excel with an *.xls file name passed as a argument.

Hence, without the extension, Windows doesn't understand what to do with a binstub.  That's the reason Windows binstubs have a .cmd or .bat extension.  Most Windows Rubies also have similar files without an extension that allow it to launch in a bash shell.

Ruby master has changed that, in that there is one binstub file with a cmd extension.  From either bash or cmd shells, `bundle.cmd --version` works.  The problem with that approach is that it breaks any cross-platform scripts that people may use, a good example being CI.

As mentioned elsewhere, I believe any Ruby commands that 'shell out' always use the Windows cmd shell.  So, even if Bundler specs are started from a bash shell, any specs that involve 'shelling out' will use the cmd shell.

This PR addresses the issue in Bundler's running code, and can/should be merged regardless of whether it allows more specs to pass.

As to checking the other skips, I'll see what I can do,  That might quickly get very messy, and could expose more issues in runtime code.  Then again, the fix in this PR was relatively painless, so maybe I should be more optimistic...

deivid-rodriguez:
Thanks, that makes sense.

> As mentioned elsewhere, I believe any Ruby commands that 'shell out' always use the Windows cmd shell. So, even if Bundler specs are started from a bash shell, any specs that involve 'shelling out' will use the cmd shell.

Yeah, I remember reading at some point when try to setup our Windows CI that the `COMSPEC` environment variable could be used to select the default shell on Windows?

> As to checking the other skips, I'll see what I can do, That might quickly get very messy, and could expose more issues in runtime code. Then again, the fix in this PR was relatively painless, so maybe I should be more optimistic...

This should be fairly easy. Remove those skips, and if things fail add then back. I'm not asking to fix any extra issue here, just make sure we start enforcing all tests that get fixed by these changes.

MSP-Greg:
> Yeah, I remember reading at some point when try to setup our Windows CI that the COMSPEC environment variable could be used to select the default shell on Windows?

Yes, I've tried to set that to bash, couldn't find a way to make it work.  The error reporting isn't very helpful...

I'll check on the skipped specs a bit later.

MSP-Greg:
Some gems have their own binstubs, like rack has rackup.  Windows needs them all installed, then it needs to create .cmd files if they (or .bat files) don't exist.

Where would be the best place to add that code?

I had a failure in spec/runtime/executable_spec.rb due to that, I modified installer.rb, and it worked, but I doubt that's the correct place for the code...

deivid-rodriguez:
I'm not sure, maybe I can give you feedback over code after we fix this issue?

Let's remove the specs that you can and keep the ones still needed, then let's merge this, and then let's take care of the remaining failures in `spec/runtime/executable_spec.rb`, how about that?

MSP-Greg:
Done.  The three spec files that had skips removed were also the only spec files using `File.open` without a block, which leaves an open file reference.  Changed that also...

bundler/spec/commands/binstubs_spec.rb
```diff
@@ -289,7 +285,7 @@
 
         bundle "binstubs rack --shebang jruby"
 
-        expect(File.open(bundled_app("bin/rackup")).gets).to eq("#!/usr/bin/env jruby\n")
+        expect(File.read(bundled_app("bin/rackup"), 64)).to start_with("#!/usr/bin/env jruby\n")
```
deivid-rodriguez, line 288: Can you use `File.readlines(...).first' instead? That "64" looks ugly and brittle.

MSP-Greg: I can, but I'd prefer not.  The main reason I used it is that the whole file doesn't need to be read, which `File.readlines` does.  One could use intermediate values, or lose a set of paren's, but that's not allowed.

I'll have a look at the other test files.

deivid-rodriguez: I know, but this file has like 30 lines, it's not going to make any difference. Reading it entirely is fine. Please change it.

MSP-Greg: Done.

bundler/spec/commands/binstubs_spec.rb
```diff
@@ -288,8 +284,7 @@
         G
 
         bundle "binstubs rack --shebang jruby"
-
-        expect(File.open(bundled_app("bin/rackup")).gets).to eq("#!/usr/bin/env jruby\n")
+        expect(File.readlines(bundled_app("bin/rackup")).first).to start_with("#!/usr/bin/env jruby\n")
```
deivid-rodriguez, line 287: Sorry, last tiny comment. Since we're using `File.readlines` now, can you keep using `eq`?

MSP-Greg: Sorry, I get robotic with reviews.  Done.

kaycoinsUSA:
{.patch -bundler - fix on Windows. "ruby gems"}``

PR #3740, opened 2020-06-19
Fix RuboCop offenses

utkarsh2102:
These offenses appear when you create a gem with
`bundle gem foo` and run `rubocop` over it.

Initially, there were around 45 offenses detected,
but with #3731 and this, the number of offenses
have been reduced to 0.

@bronzdoc, as promised, here's this :tada: 

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -24,7 +24,9 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  spec.files = Dir.chdir(File.expand_path(__dir__)) do
+    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  end
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require_relative "lib/<%=config[:namespaced_path]%>/version"
 
 Gem::Specification.new do |spec|
@@ -6,8 +8,8 @@ Gem::Specification.new do |spec|
   spec.authors       = [<%= config[:author].inspect %>]
   spec.email         = [<%= config[:email].inspect %>]
 
-  spec.summary       = %q{TODO: Write a short summary, because RubyGems requires one.}
-  spec.description   = %q{TODO: Write a longer description or delete this line.}
+  spec.summary       = "TODO: Write a short summary, because RubyGems requires one."
+  spec.description   = "TODO: Write a longer description or delete this line."
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
 <%- if config[:mit] -%>
   spec.license       = "MIT"
@@ -22,9 +24,7 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files         = Dir.chdir(File.expand_path("..", __FILE__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
-  end
+  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   spec.bindir        = "exe"
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -155,9 +155,10 @@ RSpec.describe "bundle gem" do
     it "generates a gem skeleton with rubocop" do
       gem_skeleton_assertions
       expect(bundled_app("test-gem/Rakefile")).to read_as(
-        include('require "rubocop/rake_task"').
+        include('# frozen_string_literal: true').
+        and(include('require "rubocop/rake_task"').
         and(include("RuboCop::RakeTask.new").
-        and(match(/:default.+:rubocop/)))
+        and(match(/default:.+:rubocop/))))
       )
     end
 
@@ -182,7 +183,6 @@ RSpec.describe "bundle gem" do
       realworld_system_gems gems, :path => path
       bundle "install", :dir => bundled_app("#{gem_name}")
       bundle "exec rubocop", :dir => bundled_app("#{gem_name}")
-
       expect($?.exitstatus).to eq(0) if exitstatus
     end
   end
@@ -555,7 +555,7 @@ RSpec.describe "bundle gem" do
             t.test_files = FileList["test/**/*_test.rb"]
           end
 
-          task :default => :test
+          task default: :test
         RAKEFILE
 
         expect(bundled_app("#{gem_name}/Rakefile").read).to eq(rakefile)
@@ -613,7 +613,7 @@ RSpec.describe "bundle gem" do
             t.test_files = FileList["test/**/*_test.rb"]
           end
 
-          task :default => :test
+          task default: :test
         RAKEFILE
 
         expect(bundled_app("#{gem_name}/Rakefile").read).to eq(rakefile)
@@ -914,7 +914,7 @@ RSpec.describe "bundle gem" do
             ext.lib_dir = "lib/#{gem_name}"
           end
 
-          task :default => [:clobber, :compile]
+          task default: [:clobber, :compile]
         RAKEFILE
 
         expect(bundled_app("#{gem_name}/Rakefile").read).to eq(rakefile)
@@ -1010,7 +1010,7 @@ Usage: "bundle gem NAME [OPTIONS]"
 
         RSpec::Core::RakeTask.new(:spec)
 
-        task :default => :spec
+        task default: :spec
       RAKEFILE
 
       expect(bundled_app("foobar/Rakefile").read).to eq(rakefile)
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -155,7 +155,7 @@ RSpec.describe "bundle gem" do
     it "generates a gem skeleton with rubocop" do
       gem_skeleton_assertions
       expect(bundled_app("test-gem/Rakefile")).to read_as(
-        include('# frozen_string_literal: true').
+        include("# frozen_string_literal: true").
         and(include('require "rubocop/rake_task"').
         and(include("RuboCop::RakeTask.new").
         and(match(/default:.+:rubocop/))))
@@ -176,11 +176,11 @@ RSpec.describe "bundle gem" do
     end
 
     it "run rubocop inside the generated gem with no offenses" do
-      prepare_gemspec(bundled_app("#{gem_name}", "#{gem_name}.gemspec"))
+      prepare_gemspec(bundled_app(gem_name, "#{gem_name}.gemspec"))
       gems = ["rake", "rubocop"]
-      path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app("#{gem_name}")) : system_gem_path
+      path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app(gem_name)) : system_gem_path
       realworld_system_gems gems, :path => path
-      bundle "exec rubocop --ignore-parent-exclusion", :dir => bundled_app("#{gem_name}")
+      bundle "exec rubocop --ignore-parent-exclusion", :dir => bundled_app(gem_name)
       expect($?.exitstatus).to eq(0) if exitstatus
     end
   end
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -177,12 +177,10 @@ RSpec.describe "bundle gem" do
 
     it "run rubocop inside the generated gem with no offenses" do
       prepare_gemspec(bundled_app("#{gem_name}", "#{gem_name}.gemspec"))
-
-      gems = ["rubocop"]
+      gems = ["rake", "rubocop"]
       path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app("#{gem_name}")) : system_gem_path
       realworld_system_gems gems, :path => path
-      bundle "install", :dir => bundled_app("#{gem_name}")
-      bundle "exec rubocop", :dir => bundled_app("#{gem_name}")
+      bundle "exec rubocop --ignore-parent-exclusion", :dir => bundled_app("#{gem_name}")
       expect($?.exitstatus).to eq(0) if exitstatus
     end
   end
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -180,7 +180,7 @@ RSpec.describe "bundle gem" do
       gems = ["rake", "rubocop -v 0.80.1"]
       path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app(gem_name)) : system_gem_path
       realworld_system_gems gems, :path => path
-      bundle "exec rubocop --ignore-parent-exclusion", :dir => bundled_app(gem_name)
+      bundle "exec rubocop --config .rubocop.yml", :dir => bundled_app(gem_name)
       expect($?.exitstatus).to eq(0) if exitstatus
     end
   end
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -173,6 +173,18 @@ RSpec.describe "bundle gem" do
     it "generates a default .rubocop.yml" do
       expect(bundled_app("#{gem_name}/.rubocop.yml")).to exist
     end
+
+    it "run rubocop inside the generated gem with no offenses" do
+      prepare_gemspec(bundled_app("#{gem_name}", "#{gem_name}.gemspec"))
+
+      gems = ["rubocop"]
+      path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app("#{gem_name}")) : system_gem_path
+      realworld_system_gems gems, :path => path
+      bundle "install", :dir => bundled_app("#{gem_name}")
+      bundle "exec rubocop", :dir => bundled_app("#{gem_name}")
+
+      expect($?.exitstatus).to eq(0) if exitstatus
+    end
   end
 
   shared_examples_for "--no-rubocop flag" do
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -532,6 +532,8 @@ RSpec.describe "bundle gem" do
 
       it "creates a default rake task to run the test suite" do
         rakefile = strip_whitespace <<-RAKEFILE
+          # frozen_string_literal: true
+
           require "bundler/gem_tasks"
           require "rake/testtask"
 
@@ -588,6 +590,8 @@ RSpec.describe "bundle gem" do
 
       it "creates a default rake task to run the test suite" do
         rakefile = strip_whitespace <<-RAKEFILE
+          # frozen_string_literal: true
+
           require "bundler/gem_tasks"
           require "rake/testtask"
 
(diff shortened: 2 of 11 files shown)


bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -22,7 +24,7 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files         = Dir.chdir(File.expand_path("..", __FILE__)) do
+  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
```
utkarsh2102, line 27: The remaining 2 offenses that I mentioned in the description comes from here:
```
Offenses:

foo.gemspec:22:13: C: Layout/ExtraSpacing: Unnecessary spacing detected.
  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
            ^^^^^^^^
foo.gemspec:22:22: C: Layout/SpaceAroundOperators: Operator = should be surrounded by a single space.
  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
                     ^

8 files inspected, 2 offenses detected
```

I don't know what to do with it, because it doesn't look nice if we remove those spaces.
The one possible option here is to add the following lines to `.rubocop.yml`:
```
Layout/ExtraSpacing:
  Enabled: true
  Exclude:
    - newgem.gemspec.tt

Layout/SpaceAroundOperators:
  Enabled: true
  Exclude:
    - newgem.gemspec.tt
```

But I am not sure how right would it be to do this.

deivid-rodriguez: I personally dislike the spaces, so I would remove then.

utkarsh2102: Ah, sure! I'd love that myself, but it would be inconsistent with other things in the gemspec:
```
--- a/foo.gemspec
+++ b/foo.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
 
-  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
+  spec.files = Dir.chdir(File.expand_path(__dir__)) do
     `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   end
   spec.bindir        = "exe"
```

Notice the spaces b/w `spec.bindir` and `= "exe"`. Similarly, the entire gemspec is nicely formatted and `spec.files` would be odd one out, I feel.

Personally, I don't mind doing this, but it's just that I'd want to let you know this so we don't have any unexpected surprises :smile:

deivid-rodriguez: Oh, kill them all! I believe `rubocop` should actually start complaining about the `spec.bindir` line as soon as you remove the spaces on the `spec.files` line.

utkarsh2102: Aye, aye, cap'n!
I'll do this and also fix the specs :sweat_smile:

utkarsh2102: > Oh, kill them all! I believe `rubocop` should actually start complaining about the `spec.bindir` line as soon as you remove the spaces on the `spec.files` line.

Just btw, `rubocop` is happy just after removing the spaces from `spec.files` :P
It doesn't care about anything else :joy:

deivid-rodriguez: Oh, I see, it allows those spaces "for alignment" with the other assignments. So to completely fix this you need to either remove spaces in all assignments, or change the `spec.files` into a one-liner, so that `rubocop` detects that the spaces are there for alignment.

I don't care much about which solution you choose, so go ahead with the one you prefer.

simi:
I'm sorry I have missed https://github.com/rubygems/rubygems/pull/3731/files, but I think we should ship no rule overrides, just follow the default ones (to not bring any another opinion in there actually) and create on CI new test to create gem and run rubocop on top of that to ensure no offences are present on skeleton.

bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
utkarsh2102: Ah, this is interesting! I found out why:
```
--- a/foo.gemspec
+++ b/foo.gemspec
@@ -9,6 +9,10 @@ Gem::Specification.new do |spec|
   spec.summary       = "TODO: Write a short summary, because RubyGems requires one."
   spec.description   = "TODO: Write a longer description or delete this line."
   spec.homepage      = "TODO: Put your gem's website or public repo URL here."
+  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
+    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
+  end
+
   spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
 
   spec.metadata["allowed_push_host"] = "TODO: Set to 'http://mygemserver.com'"
```

If the spaces are being done in a block of lines, then it doesn't care.
`rubocop` registers no offense here and is happy!
On the other hand, this:
```
  spec.files         = Dir.chdir(File.expand_path(__dir__)) do
    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
  end
  spec.bindir        = "exe"
  spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
  spec.require_paths = ["lib"]
```
is being registered as an offense. Interesting!

deivid-rodriguez:
@simi I tend to recall that we discussed this in the past and we want to encourage using double quotes by overriding this default in the generated gem skeleton. That's my opinion anyways :)

deivid-rodriguez:
I'm happy with adding a CI test to ensure that rubocop is green on generated gems.

utkarsh2102:
> I'm sorry I have missed https://github.com/rubygems/rubygems/pull/3731/files, but I think we should ship no rule overrides, just follow the default ones (to not bring any another opinion in there actually) and create on CI new test to create gem and run rubocop on top of that to ensure no offences are present on skeleton.

The team prefers the use of double quotes so in case there's a case of string interpolation, there'll be no inconsistency with both sorts of "StringLiterals". And I tend to agree with this.

deivid-rodriguez:
@utkarsh2102 Yeah, I believe that's why the `bundler` team uses `double_quotes` internally. The question is whether we want to promote that also on new gems. My opinion is yes to that as well :)

utkarsh2102:
> The question is whether we want to promote that also on new gems. My opinion is yes to that as well :)

I wholeheartedly agree with this as well :smile:

deivid-rodriguez:
@simi Well, my recollection was wrong. I double checked https://github.com/rubygems/bundler/pull/6455, and the only opinions there are:

* @colby-swandale's follow community defaults.
* Mine: ship a `.rubocop.yml` with the overrides we want to ship.

So, it seems that I'm losing this battle too :sweat_smile:

deivid-rodriguez:
@indirect @hsbt Any opinions here?

simi:
> AFAICU, the team prefers the use of double quotes so in case there's a case of string interpolation, there'll be no inconsistency with both sorts of "StringLiterals". And I tend to agree with this.

I do not fight for any rule and actually bundler/rubygems standards are not related here since it is new gem skeleton, not actual part of runtime.

I'm just trying to say that if we would like to integrate rubocop in new gem skeleton, we should not do any decision on configuration. The same we do for rspec and minitest, just minimal default integration needed to be able to start testing without bringing any opinions.

deivid-rodriguez:
Yeah, that's the debate here. My opinion is that it is fine that we bring some opinions, specially with double_quotes, since a relevant part of the community actively changes this default. And it also provides a sample configuration, which is nice to get started in my opinion.

utkarsh2102:
> I do not fight for any rule and actually bundler/rubygems standards are not related here since it is new gem skeleton, not actual part of runtime.
> I'm just trying to say that if we would like to integrate rubocop in new gem skeleton, we should not do any decision on configuration. The same we do for rspec and minitest, just minimal default integration needed to be able to start testing without bringing any opinions.

Of course. It's just opinion-based. #3731 was raised because I didn't like `rubocop` throwing around 45 offense to a very new (skeleton) gem.
I personally would prefer double_quotes (and understood that team would prefer that too because of inconsistencies caused by the string interpolation thingy). So raised that and now this.

I am perfectly fine with whatever the outcome of this discussion is, but I think it's important that we, at least, discuss this.

And obviously, all in good faith! :heart::heart:

deivid-rodriguez:
Yeah, as long as we ship an offense free rubocop setup, I'm fine with any outcome here.

Let's wait for some more opinions, and if the current balance (2 to 1) doesn't change, let's remove the generated `.rubocop.yml` file and switch the templates to whatever `rubocop` likes.

simi:
>  I didn't like rubocop throwing around 45 offense to a very new (skeleton) gem.

Yes, I think we all have the same goal. Provide skeleton without any rubocop violations. The question is how to get to this state :) My opinion was written already, anyway I'm ok with any solution.

:thinking: Next step should be actually pull request to integrate rubocop check in bundled CI scripts as well and have CI green by default. :partying_face:

deivid-rodriguez:
@utkarsh2102 Do you want to give that test a try? In the new gem specs, add a spec that runs rubocop inside the generated gem, and checks that there are no errors.

utkarsh2102:
> @utkarsh2102 Do you want to give that test a try? In the new gem specs, add a spec that runs rubocop inside the generated gem, and checks that there are no errors.

Sure, I'd love to do that! :smile:

hsbt:
I don't like rubocop's default rule. In fact, many of projects disable it rule.

* https://github.com/rails/rails/blob/master/.rubocop.yml#L9
* https://github.com/github/rubocop-github/blob/master/config/default_deprecated.yml#L4
* https://github.com/discourse/rubocop-discourse/blob/master/default.yml#L10

In my opinion, we put `DisabledByDefault: true` to default rubocop configuration on new gem. But I'm okay to this pull request without `FrozenStringLiteralComment` style. Because the default of frozen string literal is canceled at Ruby 3.0. There is no reason to enforce it.

deivid-rodriguez:
Yeah, have a look at https://github.com/rubocop-hq/rubocop/issues/5306 for a discussion on what the default quote style is for ruby. It's about 50/50, although the most popular projects usually use double quotes.

I don't mind about `FrozenStringLiteralComment`. I think it doesn't hurt even if it won't be the default, but I don't mind if we turn it off.

However, I believe `DisabledByDefault: true` limits the usefulness of adding `rubocop` too much. I use `DisabledByDefault: true` all the time for big projects with very inconsistent style, so that I can gradually enforce consistency, but for new projects I think using all the power of `rubocop` is better.

utkarsh2102:
> But I'm okay to this pull request without FrozenStringLiteralComment style. Because the default of frozen string literal is canceled at Ruby 3.0. There is no reason to enforce it.

I don't mind, **but** how about, we consider this road:
-> Ruby 3.0 is released with the default frozen string literal canceled.
-> FrozenStringLiteralComment cop in RuboCop if fixed to not enforce this behavior.
-> And finally then we remove this here.

This might be sounding crazy, but the whole idea of doing this is to not show any offenses by running `rubocop` in the first place.
If we don't enfore `FrozenStringLiteralComment` now, the skeleton gem will have around 10 offenses.

deivid-rodriguez:
@utkarsh2102 What @hsbt is suggesting is to add

```yaml
Style/FrozenStringLiteralComment:
  Enabled: false
```

to the generated gem `rubocop` configuration and not add any `frozen_string_literal: true` comment to the templates. That approach would also have no offenses.

utkarsh2102:
> @utkarsh2102 What @hsbt is suggesting is to add
> 
> ```yaml
> Style/FrozenStringLiteralComment:
>   Enabled: false
> ```
> 
> to the generated gem `rubocop` configuration and not add any `frozen_string_literal: true` comment to the templates. That approach would also have no offenses.

Ah, sure. This means that we are definitely shipping a `.rubocop.yml`? :smile:

deivid-rodriguez:
Haha, good question :laughing:.

It seems that we have a tie (2-2), although I'm still giving this some more thought and I might even reconsider my vote :sweat_smile:

Let's hold on a bit on that, but you can definitely add the spec already since everybody wants that.

bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -22,9 +24,7 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files         = Dir.chdir(File.expand_path("..", __FILE__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
-  end
+  spec.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```
deivid-rodriguez, line 27: Please add back the `Dir.chdir` block. Without that, evaluating the gemspec from outside `File.expand_path("..", __FILE__)` won't work.

utkarsh2102: Yikes!
Then I shall just remove the spaces :confused:

deivid-rodriguez: You can do this

```ruby
spec.files         = Dir.chdir(File.expand_path("..", __FILE__)) { `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) } }
```

utkarsh2102: The line length exceeds the usual limit. So since there's no way can we wrap this within 125 characters, I'll remove the spaces for now \o/

bundler/spec/commands/newgem_spec.rb
```diff
@@ -173,6 +173,18 @@ def gem_skeleton_assertions
...
+      prepare_gemspec(bundled_app("#{gem_name}", "#{gem_name}.gemspec"))
+
+      gems = ["rubocop"]
+      path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app("#{gem_name}")) : system_gem_path
+      realworld_system_gems gems, :path => path
+      bundle "install", :dir => bundled_app("#{gem_name}")
+      bundle "exec rubocop", :dir => bundled_app("#{gem_name}")
+
+      expect($?.exitstatus).to eq(0) if exitstatus
+    end
```
utkarsh2102, line 187: Right now, this doesn't work. It needs more work & debugging.
I am not entirely sure how to pull this off, so pushed this WIP commit :sweat_smile: 

@simi, any ideas or suggestions?

deivid-rodriguez:
> `@indirect` `@hsbt` Any opinions here?

Also forgot to ask @bronzdoc!

indirect:
I am strongly in favor of having tests for this, good suggestion @utkarsh2102!

Personally, I would like my new gems to use double-quotes. 😅 I am happy with removing frozen string literals, to get us ahead of the Ruby 3 curve, and I would agree with @deivid-rodriguez to leave the rules on by default since this is a new project.

That said, I am in favor of this work even if the consensus ends up not agreeing with my personal preferences. 😄

bundler/spec/commands/newgem_spec.rb
simi: Ahh, I had different idea, to actually create just CI script to install bundler from git and run few commands.

Anyway I'll take a look to this, since ti is probably better.

deivid-rodriguez: I created #3742, which should fix at least part of the problems that you are having with this spec (currently `realworld_system_gems` does not support a `:path` option).

utkarsh2102: Thanks for this, @deivid-rodriguez! :heart: 
I got the test working :rocket:

utkarsh2102:
D'oh! It looks like I need to set the rubocop version to `0.80.1`, right?
And I don't understand this failure: https://github.com/rubygems/rubygems/runs/791381071#step:10:335? Would locking the version fix that, too?

deivid-rodriguez:
Me neither... I'm not sure if looking will fix it :man_shrugging:

bronzdoc:
Is it possible that we make the skeleton to comply with the default rubocop rules? 

That way we ship an unopinionated config and solve the main problem which is an skeleton free of rubocop failures

utkarsh2102:
> Me neither... I'm not sure if looking will fix it :man_shrugging: 

https://github.com/rubygems/rubygems/pull/3740/checks?check_run_id=791396843#step:5:171 --> this tells me that we should definitely lock it to the version you ship. Unless we want to drop ruby2.3 support? :smile: 

Also, when we lock it to 0.80.1, we'll need to comply with 80 characters in a line. This means we need to fix two more offenses (the LineLength ones).

> Is it possible that we make the skeleton to comply with the default rubocop rules?
> That way we ship an unopinionated config and solve the main problem which is an skeleton free of rubocop failures

@bronzdoc, hello,
For that to happen, we need to agree on two things here:
- Use single_quotes instead of double_quotes (as what some of us here prefers)
- We should be okay with `FrozenStringLiteralComment` style for now (as opposed [here](https://github.com/rubygems/rubygems/pull/3740#issuecomment-646613980))

We can only make the skeleton to comply with the default rubocop rules if we have some agreement on the above two things^ :smile:

deivid-rodriguez:
Thanks @utkarsh2102, that's a very accurate description of the situation...

My opinion regarding the line that's too long is

```yaml
Metrics:
  Enabled: false
```

:laughing: 

But since we have a tie regarding being opinionated, I guess let's not ship a `.rubocop.yml` file and add an inline disable instead (`# rubocop:disable Metrics/LineLength` at the end of the long line)?

And yes, lock the version!

And yes, we need to figure out that error :(

bronzdoc:
@utkarsh2102 Ahh I see the problem is that we don't like rubocop defaults... (can't believe Style/StringLiterals default is single quotes)

I don't like opinionated configs but I do prefer double quoting and would like to encourage new gem creators to use it.

And I am ok with removing frozen string literals

deivid-rodriguez:
So, now I don't know if we have a tie anymore, or not.... or what. Too many moving parts :laughing:

What's sure is we must be driving @utkarsh2102 crazy lol

simi:
Let's get any version ready for now. We can discuss default rubocop.yml on Slack at any time before following release and prepare changes if needed.

deivid-rodriguez:
Agreed @simi :+1:

@utkarsh2102 Go with whatever option you like best, and we can tweak things after merge.

> And I don't understand this failure: https://github.com/rubygems/rubygems/runs/791381071#step:10:335? Would locking the version fix that, too?

I believe since the `rubocop-ast` gem was only extracting recently, and it's in the backtrace, locking the version might indeed fix that too :crossed_fingers:.

utkarsh2102:
Okay, it's getting craizer now.. :see_no_evil: 
I tried both the ways of locking the version:
- ["rake", "rubocop -v 0.80.1"], and
- ["rake", "rubocop-0.80.1"]

but the CI fails weirdly?

utkarsh2102:
Argh, the CI is already broken! So it's not because of my changes :man_facepalming: 
@deivid-rodriguez, your last merge broke something \o/

deivid-rodriguez:
@utkarsh2102 Interesting, thanks for noticing. I'll revert that PR now.

utkarsh2102:
Finallyyy! The CI is greeeen! :sob::tada:

utkarsh2102:
> I can't believe this is green finally :sob: 
> Great job :muscle: 

Thanks to you for helping and being patient with me :heart:

deivid-rodriguez:
Since this PR fulfills our main goal of having a clean `rubocop` run on new gems by default, I'll merge this unless anybody has objections.

As @simi suggested, we can discuss further tweaks on the default configuration later.

indirect:
Ah, I just noticed the change from %q to double quotes in the gemspec—this isn't blocking, but the reason I did that is so that gem descriptions can contain both double and single quotes (like `%q{Helps users' "friends" meet each other}` can just be typed directly into the string. It's super frustrating to break a string by typing an english sentence that describes the gem. :)

utkarsh2102:
Hi @indirect,

> Ah, I just noticed the change from %q to double quotes in the gemspec—this isn't blocking, but the reason I did that is so that gem descriptions can contain both double and single quotes (like `%q{Helps users' "friends" meet each other}` can just be typed directly into the string. It's super frustrating to break a string by typing an english sentence that describes the gem. :)

I hear you, I really do!
And I agree that it's frustrating indeed to have things broken by a mere sentence. And I'll be even happy to revert that back to using `%q`, but in order to do so and at the same time make sure we've 0 offenses shown by RuboCop, we need to put this into `.rubocop.yml` and this will be yet another "opinionated" config that we ship --- which was the whole discussion we had above :P

Personally, I am happy to do so but again, we will need to discuss more about whether we want to ship a default `.rubocop.yml` or not.

The best way forward, I think, is
- We get this merged so we at least have an offenseless skeleton gem, which will be very good!
- Simultaneously, I can start a thread about this on Slack and we can discuss more about this and decide how the things will going to be by the next release.

That said, I can even revert back to `%q` and disable `Style/PercentLiteralDelimiters` and `Style/RedundantPercentQ` by default, but we will go in the same loop again.. :P

So let me know what you think? :sweat_smile:

indirect:
Oh I think we should merge first! I called that out as not blocking on purpose. 😄 Thanks for trying to work with us.

utkarsh2102:
> Thanks for trying to work with us.

It's both, an honor and a pleasure to do so ❤️

deivid-rodriguez:
Awesome, merging this pending further discussions about magic string literals comments, or specific exclusions to the default string quotes rule in the gemspec :+1:.

Thanks so much for your patience @utkarsh2102!

PR #3743, opened 2020-06-20
.gitattributes - diff=ruby, eol, rubocop_bundler.yml - eol

MSP-Greg:
## Description:

.gitattributes- text eol=lf, add diff=ruby for several directories

rubocop_bundler.yml - set Layout/EndOfLine to match RubyGems (EOL should be lf)

Running RuboCop on Bundler showed errors for **every.file.in.bundler**.

see Desc

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,9 @@
+# git log -L:<method>:<file>
+*.rb diff=ruby
+*.gemspec diff=ruby
+bin/* diff=ruby
+bundler/bin/* diff=ruby
+bundler/exe/* diff=ruby
+
+# Auto detect text files and perform LF normalization
+* text eol=lf
--- a/.rubocop_bundler.yml
+++ b/.rubocop_bundler.yml
@@ -297,6 +297,7 @@ Layout/EndAlignment:
 
 Layout/EndOfLine:
   Enabled: true
+  EnforcedStyle: lf
 
 Layout/ExtraSpacing:
   Enabled: true


.gitattributes
```diff
@@ -0,0 +1,9 @@
+# git log -L:<method>:<file>
+*.rb diff=ruby
+*.gemspec diff=ruby
+bin/* diff=ruby
+bundler/bin/* diff=ruby
+bundler/exe/* diff=ruby
```
deivid-rodriguez, line 6: All the above seems completely unrelated to this changeset? If it's indeed unrelated, please drop it and explain why you need this on a separate PR.

Or at least split it into a different commit, and explain in the commit message why you need this. I've never seen these configuration on any ruby gem.

.gitattributes
```diff
@@ -0,0 +1,9 @@
...
+
+# Auto detect text files and perform LF normalization
+* text eol=lf
```
deivid-rodriguez, line 9: I guess you need this for windows, sounds fine :+1:.

.rubocop_bundler.yml
```diff
@@ -297,6 +297,7 @@ Layout/EndAlignment:
 
 Layout/EndOfLine:
   Enabled: true
+  EnforcedStyle: lf
```
deivid-rodriguez, line 300: :+1:

deivid-rodriguez:
Also, for future PRs, could you give them a title with "human words"? The files being edited, and the changes being made are already part of the commits, so repeating that doesn't add much value. Thanks :pray:.

.gitattributes
MSP-Greg: https://github.com/ruby/ruby/blob/master/.gitattributes

https://github.com/rails/rails/blob/master/.gitattributes

deivid-rodriguez: That's nice, I didn't know about it. I'm putting 

```
*.rb diff=ruby
*.gemspec diff=ruby
```

in my global `~/.git/config/attributes` so I can get the candy without having to add PRs to every project :)

Anyways, I'm good with this :+1:. Will you split this to a separate commit or PR?

MSP-Greg: > Will you split this to a separate commit or PR?

Given that this has no affect on the anything but local use (git & RuboCop), I really don't see a need to run CI on it again.  If you really want, I can split it into two commits...  It's six lines excluding comments...

deivid-rodriguez: It seems contradictory to me that you want to add changes so that we get a better git history, but refuse to split your commits so that we get a better git history. Either you care about git history, or you don't. So please, either remove the 6 lines, or split them to a separate commit.

MSP-Greg: Done - two commits

deivid-rodriguez: @MSP-Greg Thanks for splitting the commits. Part of the motivation of splitting unrelated changes into separate commits is to explain the reasons for the changes in the commit messages, so that a future reader can understand the _why_. The commit itself already contains the file being changed, and the changes being made, so please in the future add meaningful commit messages. Thanks!

PR #3747, opened 2020-06-21
Revert part of the new `.gitattributes` file

deivid-rodriguez:
# Description:

Since it broke CI.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,6 +4,3 @@
 bin/* diff=ruby
 bundler/bin/* diff=ruby
 bundler/exe/* diff=ruby
-
-# Auto detect text files and perform LF normalization
-* text eol=lf
--- a/.github/workflows/ubuntu-bundler.yml
+++ b/.github/workflows/ubuntu-bundler.yml
@@ -30,6 +30,11 @@ jobs:
       RUBYOPT: --disable-gems
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          path: bundler/tmp/rubygems
+          ref: ${{ matrix.rgv }}
+        if: matrix.rgv != '..'
       - name: Setup ruby
         uses: ruby/setup-ruby@v1
         with:
--- a/.github/workflows/ubuntu-bundler3.yml
+++ b/.github/workflows/ubuntu-bundler3.yml
@@ -18,6 +18,11 @@ jobs:
       RGV: ${{ matrix.rgv }}
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+        with:
+          path: bundler/tmp/rubygems
+          ref: ${{ matrix.rgv }}
+        if: matrix.rgv != '..'
       - name: Setup ruby
         uses: ruby/setup-ruby@v1
         with:


.gitattributes
```diff
@@ -4,6 +4,3 @@
 bin/* diff=ruby
 bundler/bin/* diff=ruby
 bundler/exe/* diff=ruby
-
-# Auto detect text files and perform LF normalization
-* text eol=lf
```
utkarsh2102, line 9: I don't think if this was the real reason for the CI to break? :thinking:

deivid-rodriguez: Yes, it is. But it's a mess because our CI checks out `rubygems` master branch, which doesn't have the updated `.gitattributes` file, so it's still broken :laughing:.

utkarsh2102: Oh wow, then I think we're good to merge? 🔥

deivid-rodriguez: I'm doing some experimentation with solutions that will allow us to actually catch these issues before merge, not after :sweat_smile:.

deivid-rodriguez: It worked! Merging this to go back to green.

utkarsh2102:
YOU ARE AWESOME! :sob::heart:

MSP-Greg:
@deivid-rodriguez

You should revert the revert of .gitattributes https://github.com/rubygems/rubygems/commit/fc762d9d899f29b872826a80bdb44b904e0c24fd, as it is not the cause of any issues.

As currently shown in https://github.com/MSP-Greg/rubygems/commits/master, I reverted the commit, pushed it to my fork's master, then added another dummy commit that forced CI.  See See https://github.com/MSP-Greg/rubygems/actions, and look at all the jobs from `ubuntu-bundler #13` to `install #8`. They're currently the top set of jobs listed, and all workflows passed.

The jobs immediately below those I canceled, as I had not updated the tags in my fork for a while.

There seems to be a tendency for a few people involved in this repo to have critical knee-jerk reactions to anything they're unfamiliar with.  There also seems to be a lack of concern for how easy it is for contributors to work with PR's in their forks.

I come here to make RubyGems/Bundler better, not deal with wannabe autocrats...

@deivid-rodriguez

IMO, you've found a lot of bugs here, and you've written a lot of good code to fix them.  Conversely, you revert a commit with no discussion because you're unfamiliar with it, but when I do a PR that adds a few lines of helpful code, you're off and running.

JFYI, I haven't really looked at the cause of the issue.  Actions like `actions/checkout@v2` are pinned to the major release (they can be pinned further), and they may have changed something with a minor/patch.  Using the action, then checking out another ref is probably not considered normal use, as allowing it can affect the time it takes to clone it...

deivid-rodriguez:
@MSP-Greg Believe it or not, your PR [broke our master branch CI](https://github.com/rubygems/rubygems/runs/791566901). Not only there, but consistently in every branch. Please don't use your fork as an example, your fork's results are meaningless if the canonical repo is broken.

> IMO, you've found a lot of bugs here, and you've written a lot of good code to fix them. Conversely, you revert a commit with no discussion because you're unfamiliar with it, but when I do a PR that adds a few lines of helpful code, you're off and running.

I'm sorry but I tested my PR locally and made sure that it was unbreaking the repo, and that I understood why it was broken in the first place. If you want the actual explanation of why the PR works in your fork, it's because of several reasons:

* You don't reverted the full PR, only the `.gitattributes` change.
* Even if you revert the full PR, it would still work, because our scripts clone the master branch of the canonical repo, which still have the `.gitattributes` file without your changes:

https://github.com/rubygems/rubygems/blob/3c7f2498d3a8bac6c83548a4cc434afcdf864a1a/bundler/spec/support/rubygems_version_manager.rb#L111

If you want a clue of where the errors are about, you can run this:

```
git clone --no-checkout https://github.com/rubygems/rubygems test_repo
cd test_repo
git checkout fa2723f66
git status
```
and observe all the source control changes creating by the lf conversion that shouldn't have been applied to a bunch of files.

Please, don't pretend that you know everything, and don't expect the projects you contribute to always do things the way you do them. And please moderate your tone.

deivid-rodriguez:
By the way, I'm happy to reintroduce a fixed version of your changes. Something like this should do the trick:

```diff
+
+# Auto detect text files and perform LF normalization
+* text eol=lf
+bundler/spec/support/artifice/vcr_cassettes/realworld/**/response binary
```

MSP-Greg:
Re .gitattributes, I apologize, I was unaware of the binary files, and this issue is also affected by git config's `[core] autocrlf` setting.  I can't look at that for a bit, my main windows issue was the RuboCop EOL, and that's fixed.

Re `rubygems_version_manager.rb`, you've added workflow code that creates tmp/rubygems for any RG versions other than master, so rubygems/rubygems is only used for that?

Regardless, with a PR that affects both Bundler and RubyGems (yes, probably uncommon), this would be a an issue?  Likewise, working in a fork doesn't work.  And, I'm sure everyone's favorite, Bundler CI in Ruby Core.

Not the best day for me coding wise.  Any thoughts?  As to forks and Actions CI, one could test for the actual repo under test (using ENV), but that doesn't help with local or Ruby Core.  Not sure if a simple copy could be done?

https://github.com/rubygems/rubygems/blob/3c7f2498d3a8bac6c83548a4cc434afcdf864a1a/bundler/spec/support/rubygems_version_manager.rb#L104-L115

deivid-rodriguez:
> Re .gitattributes, I apologize, I was unaware of the binary files, and this issue is also affected by git config's [core] autocrlf setting. I can't look at that for a bit, my main windows issue was the RuboCop EOL, and that's fixed.

Thanks for apologizing, accepted. I'm happy you got to fix your main issue (in the future, remember, separate PRs, they would've also simplified this revert :wink:).

> Re rubygems_version_manager.rb, you've added workflow code that creates tmp/rubygems for any RG versions other than master, so rubygems/rubygems is only used for that?

When a workflow is testing bundler against a rubygems version different than the one at "..", then we clone rubygems at a temporary path, checkout the proper tag, and use that rubygems to test bundler against. So all the `git clone` and subsequent `git checkout` issues only affected this case. When you test against `..`, you don't need to add any `git` magic, since the code you want to test against is already there.

> Regardless, with a PR that affects both Bundler and RubyGems (yes, probably uncommon), this would be a an issue? Likewise, working in a fork doesn't work. And, I'm sure everyone's favorite, Bundler CI in Ruby Core.

I don't think there would be any issues. Changing both at the same time is getting more common with the repo merge, actually. These cases are one of the main improvements of the repo merge at the moment.

> Not the best day for me coding wise. Any thoughts? As to forks and Actions CI, one could test for the actual repo under test (using ENV), but that doesn't help with local or Ruby Core. Not sure if a simple copy could be done?

I believe, since we already have rubygems fetched locally, we might be able to actually use the local repository to clone from, I'll look into it.

MSP-Greg:
> I believe, since we already have rubygems fetched locally, we might be able to actually use the local repository to clone from, I'll look into it.

Cool, but Ruby core?  Never tried to clone a sub-directory.  I haven't looked into how the Bundler specs are run there...

Any other places that you're aware that would affect CI in a fork?

Issue #3582, opened 2020-05-01
Man pages check break CI every month


deivid-rodriguez:
We have a CI check that makes sure that man pages stay in sync with their sources. This check is good, however, the check refreshes the month in the man pages header and fails when the month changes, because it detects changes like this:

```diff

-.TH "BUNDLE\-ADD" "1" "April 2020" "" ""
+.TH "BUNDLE\-ADD" "1" "May 2020" "" ""
```

Potential solutions:

* Exclude "month header" regeneration from the `bin/rake man:build` task.
* Exclude "month header" checking from the `bin/rake man:check` task.
* Automatically commit the result of `bin/rake man:build` to master once a month.

My preferred solution would be 3, but it sounds like tricky github actions stuff.

PR #3748, opened 2020-06-21
Add actions to automatically bump man page month

utkarsh2102:
At the beginning of every month, the CI breaks as it has a check that makes sure that man pages stay in sync with their sources.

Therefore, someone (David) has to manually update the man pages every month by running `bin/rake man:build`.

With this, having actions doing this can help save manual work.

Closes: #3582

______________

# Tasks:

### DIFF
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -1,10 +1,6 @@
 name: monthly
 
 on:
-  push
-    branches:
-      - master
-
   schedule:
     - cron: '0 0 1 * *'
 
@@ -18,24 +14,25 @@ jobs:
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: ${{ matrix.ruby }}
+          ruby-version: 2.7.1
           bundler: none
 
-      - name: Install dependencies
-        run: rake setup
+      - name: Install ronn
+        run: gem install ronn:'~> 0.7.3'
 
       - name: Update month
         run: bin/rake man:build
+        working-directory: ./bundler
 
-      - name: Commit changes
-        uses: EndBug/add-and-commit@v4
-        with:
-          author_name: Utkarsh Gupta
-          author_email: utkarsh@debian.org
-          message: "Bump man pages month"
-          add: "bundler/man/*"
-        env:
-          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
-        if: failure() && env.OLD_STATUS == '"success"'
+      - name: Set up git config
+        run: |
+          git config user.name "The Bundler Bot"
+          git config user.email "bot@bundler.io"
 
-    timeout-minutes: 60
+      - name: Commit and push the changes
+        run: |
+          git add bundler/man/*
+          git commit -m "Bump man pages month"
+          git push origin master
+
+    timeout-minutes: 15
--- a/.github/workflows/monthly.yml
+++ b/.github/workflows/monthly.yml
@@ -1,0 +1,41 @@
+name: monthly
+
+on:
+  push
+    branches:
+      - master
+
+  schedule:
+    - cron: '0 0 1 * *'
+
+jobs:
+  run:
+    name: Refresh month of man pages
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler: none
+
+      - name: Install dependencies
+        run: rake setup
+
+      - name: Update month
+        run: bin/rake man:build
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v4
+        with:
+          author_name: Utkarsh Gupta
+          author_email: utkarsh@debian.org
+          message: "Bump man pages month"
+          add: "bundler/man/*"
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
+        if: failure() && env.OLD_STATUS == '"success"'
+
+    timeout-minutes: 60


utkarsh2102:
> Potential solutions:
> 
> - Exclude "month header" regeneration from the bin/rake man:build task.
> - Exclude "month header" checking from the bin/rake man:check task.
> - Automatically commit the result of bin/rake man:build to master once a month.
> 
> My preferred solution would be 3, but it sounds like tricky github actions stuff.

Hi @deivid-rodriguez,
I took a stab at this and did it in the way as you preferred :smile: 

I am not sure, but I hope this helps :sweat_smile:

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Update month
+        run: bin/rake man:build
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v4
+        with:
+          author_name: Utkarsh Gupta
+          author_email: utkarsh@debian.org
+          message: "Bump man pages month"
+          add: "bundler/man/*"
```
utkarsh2102, line 36: Another way of doing this is:

```
      - name: Set up git config
        run: |
          git config user.name "Utkarsh Gupta"
          git config user.email "utkarsh@debian.org"
      - name: Commit and push the changes
        run: |
          git add bundler/man/*
          git commit -m "Bump man pages month"
          git push origin master
```

deivid-rodriguez: Mmmm, don't really have a strong opinion, I guess since it's so simple we might do it ourselves instead of relying on an external action?

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
+        if: failure() && env.OLD_STATUS == '"success"'
```
deivid-rodriguez, line 39: I believe this is bad copy-pasta from other workflow? It should be `if: success()` I believe.

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler: none
+
+      - name: Install dependencies
+        run: rake setup
+
+      - name: Update month
+        run: bin/rake man:build
```
deivid-rodriguez, line 28: This should be run inside the `bundler` folder, see other bundler actions that use `working_directory:`.

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler: none
+
+      - name: Install dependencies
+        run: rake setup
```
deivid-rodriguez, line 25: Use `working_directory: ./bundler` so that this runs on the proper folder. Also, I think you want `bin/rake spec:deps` here? Or actually maybe `gem install ronn:'~> 0.7.3'` is enough?

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+  run:
+    name: Refresh month of man pages
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
```
deivid-rodriguez, line 21: You don't have a `matrix` here, use the latest ruby directly: `ruby-version: 2.7.1`.

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          author_name: Utkarsh Gupta
+          author_email: utkarsh@debian.org
+          message: "Bump man pages month"
+          add: "bundler/man/*"
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }} # required
+        if: failure() && env.OLD_STATUS == '"success"'
+
+    timeout-minutes: 60
```
deivid-rodriguez, line 41: I think `timeout-minutes: 15` should be more than enough.

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
+name: monthly
+
+on:
+  push
+    branches:
+      - master
```
deivid-rodriguez, line 6: I think this should be removed, we only want the scheduled part.

.github/workflows/monthly.yml
utkarsh2102: Sure, I'll change it to this other way I mentioned above^ :smile:

.github/workflows/monthly.yml
utkarsh2102: Yeah, I took this from `daily-bundler.yml` and `daily.yml`.
Since they were doing this, I copied the same \o/
Shall I change it to `if: success()`?

.github/workflows/monthly.yml
utkarsh2102: Ah, yeah! Shall do, thanks!

.github/workflows/monthly.yml
utkarsh2102: Okay, I can check this in a fresh container, I'll change it to whatever works! :smile:

.github/workflows/monthly.yml
utkarsh2102: Eeks! My bad.
Okay, I  can use 2.7.1, of course. Thanks! :smile:

.github/workflows/monthly.yml
utkarsh2102: Aye, aye :boat:

.github/workflows/monthly.yml
deivid-rodriguez: Yes! Actually, `if: success()` is the default, so just removing this line should do the trick.

.github/workflows/monthly.yml
utkarsh2102: Ooh, okay. I am not sure about this but since you're way more experienced, I'll drop this :smile:

.github/workflows/monthly.yml
```diff
@@ -0,0 +1,41 @@
...
+        run: rake setup
+
+      - name: Update month
+        run: bin/rake man:build
+
+      - name: Commit changes
+        uses: EndBug/add-and-commit@v4
+        with:
+          author_name: Utkarsh Gupta
+          author_email: utkarsh@debian.org
```
hsbt, line 34:
```suggestion
          author_name: The Bundler Bot
          author_email: bot@bundler.io
```

utkarsh2102: Thanks, @hsbt! :fire: 
Since I am fixing this entire `name` section, so I'll accommodate this in the next commit :smile:

utkarsh2102:
Of course. Shall do! :)

utkarsh2102:
@deivid-rodriguez, works like a charm:
https://github.com/utkarsh2102/rubygems/commits/check-for-man-page-action :D

indirect:
@deivid-rodriguez sorry, I think your wrong permissions were my fault. Fixed! There is nothing blocking pushes to master, just a requirement that tests pass to allow merges.

deivid-rodriguez:
No problem, thanks! I have proper permissions now :)

So, you're ok with this @indirect?

By the way, I noticed that the "require status checks to pass" is indeed checked, but no specific checks are selected, so that's essentially the same as not being checked :sweat_smile:. I'll fix it!

indirect:
@deivid-rodriguez yes, this seems good! I thought "require checks to pass" would allow if tests never ran, but block if a test failed? But maybe not. 😅

deivid-rodriguez:
@indirect I just merged a PR with failing status checks: https://github.com/rubygems/rubygems/pull/3762, just to verify this (and because the CI failure was clearly unrelated, and on an entry that shouldn't actually be required). It indeed merged without issues. May I go ahead and check the proper boxes in our configuration to prevent merging PRs with failed required checks?

indirect:
@deivid-rodriguez yes! go for it. we probably should have checked those boxes when we stopped using Bors 😬

deivid-rodriguez:
Indeed! Better late than never though :)

deivid-rodriguez:
Done! :+1:

deivid-rodriguez:
Funny enough, enabling these checks yesterday broke the first run of this automated task today: https://github.com/rubygems/rubygems/runs/824703487?check_suite_focus=true.

```
2020-07-01T00:36:34.0538173Z remote: error: GH006: Protected branch update failed for refs/heads/master.        
2020-07-01T00:36:34.0538683Z remote: error: 56 of 56 required status checks are expected.        
2020-07-01T00:36:34.0553069Z To https://github.com/rubygems/rubygems
2020-07-01T00:36:34.0553786Z  ! [remote rejected] master -> master (protected branch hook declined)
2020-07-01T00:36:34.0554113Z error: failed to push some refs to 'https://github.com/rubygems/rubygems'
```

I believe you can whitelist certain users to be able to commit directly to protected branches, so it probably makes sense to use that feature with `bundlerbot`.

deivid-rodriguez:
Actually, from reading what the option does, I think it's not going to work:

> * [ ] Restrict who can push to matching branches.
> 
> _Specify people, teams or apps allowed to push to matching branches. Required status checks will still prevent these people, teams and apps from merging if the checks fail._ 

So we'll want to either disable these checks again or do something different like creating a PR instead of pushing directly to master.

utkarsh2102:
> So we'll want to either disable these checks again or do something different like creating a PR instead of pushing directly to master.

Aw, too bad :/
But we could do something like:
- Actions to create a PR
- Actions to merge that (maybe after the checks have passed or something of that sort!)

Of course, this will be a single actions, but probably we'll have to rely on some external actions.
D'you think that would suffice?

utkarsh2102:
Hahaha, only today I was rebasing my branches and I found this: https://github.com/utkarsh2102/rubygems/commit/fda72733da279bd7e1f42f867e936d2d19cb24ee

The Bundler Bot at least pushed to my master for bumping the man-page month and it worked alright :P
(nice to know!) :smile:

deivid-rodriguez:
Hi @utkarsh2102!

That would be awesome indeed. As a matter of fact, having some sort of "Merge after CI passes" through an action would be useful in general, maybe by tagging the PR with a label, or commenting something specific. We used to have something similar back when we used [bors-ng](https://github.com/bors-ng/bors-ng), so that commenting `@bundlerbot merge` would run CI and merge the PR. However, we stopped using it because it had many issues.

utkarsh2102:
Aha, I see!
Okay, I am going to take a stab at this in the coming days and hopefully come up with something! :smile:

PR #3749, opened 2020-06-21
Clone from local

deivid-rodriguez:
# Description:

Experimenting.

# Tasks:

- [ ] Write tests


MSP-Greg:
So my commit messages are too detailed, maybe we can meet in the middle?  'Should fix it now'?

So it's either the first commit or fetch-depth 0?  Either way, LGTM.  Solves several problems.

Since we're on the issues of Bundler CI and version hell, could the following be added to the bottom of bundler/spec/spec_helper.rb?  Sometimes it's helpful to see in the log...
```ruby
puts "\nBundler version:  #{Bundler::VERSION}\nRubyGems version: #{Gem::VERSION}", ""
```

deivid-rodriguez:
This PR is a draft, that's why I marked it as a draft. You should use it too.

MSP-Greg:
> You should use it too.

Never tried that.  What do I push to?

Thumbs down is for the code idea or otherwise?

PR #3750, opened 2020-06-22
Move `automatiek` to the main repo

deivid-rodriguez:
# Description:

Currently the process of vendoring dependencies is quite tedious because either it requires manual changes, or `automatiek` needs to be refined and tweaked constantly to accommodate new dependencies we vendor. 

Since `automatiek` lives as a external gem, when we need to change it, we need to wait for a release before using it in `bundler`, sice our current `Gemfile` setup does not support "git sourced gems". So this process is not very agile.

My fix is to move automatiek to the rubygems repo, since it's just one ruby file.

See https://github.com/segiddins/automatiek/issues/10.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -252,4 +252,6 @@ end
 
 task :default => :spec
 
-Dir["task/*.rake"].each(&method(:load))
+load "task/build_metadata.rake"
+load "task/bundler_3.rake"
+load "task/release.rake"
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -168,80 +168,62 @@ namespace :man do
   end
 end
 
-begin
-  Spec::Rubygems.gem_require("automatiek")
-rescue Gem::LoadError => e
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I had this idea [a few days ago here](https://github.com/rubygems/rubygems/pull/3622#issuecomment-646805446), and I believe it's the simplest thing to do.

deivid-rodriguez:
Great, let's do it!

segiddins:
This is a violation of the automatiek license — given a substantial piece of the software was copied, the license & copyright notice must accompany the copied code

deivid-rodriguez:
Given that we use the same license (MIT), and distribute that license with the code, what other actions should we take? Never heard of such an issue when talking about one of the most permissive license.

In any case, we should definitely write a notice somewhere saying big thanks to you for the original code :raised_hands:, since that information is no longer available in `git`.

segiddins:
>The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

deivid-rodriguez:
Oh, I think I see what you mean, let me open a PR and I'll tag you.

PR #3751, opened 2020-06-22
remove explicit require for auto-loaded constant

kares:
I do not think RGs needs:
`autoload :Specification, 'rubygems/specification'` 
as well as 
`require 'rubygems/specification'`

maybe for compatibility we should only keep the `require` and remove the `autoload` instead?
the intention seems to have been to only have one, when the require was added but got lost: https://github.com/rubygems/rubygems/commit/0962032ce05d64fd4b87edf748c1e8e93a44f6b7 (https://github.com/rubygems/rubygems/commit/10f1094f81fdb6dfb680655b684d73e226de5400)

This is revealing a rare JRuby 9.2 auto-loading bug: https://github.com/jruby/jruby/issues/6293

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1335,8 +1335,6 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   autoload :Specification,      File.expand_path('rubygems/specification', __dir__)
   autoload :Util,               File.expand_path('rubygems/util', __dir__)
   autoload :Version,            File.expand_path('rubygems/version', __dir__)
-
-  require "rubygems/specification"
 end
 
 require 'rubygems/exceptions'
--- a/lib/rubygems/dependency.rb
+++ b/lib/rubygems/dependency.rb
@@ -2,9 +2,6 @@
 ##
 # The Dependency class holds a Gem name and a Gem::Requirement.
 
-require "rubygems/bundler_version_finder"
-require "rubygems/requirement"
-
 class Gem::Dependency
 
   ##
--- a/lib/rubygems/gem_runner.rb
+++ b/lib/rubygems/gem_runner.rb
@@ -7,7 +7,6 @@
 
 require 'rubygems'
 require 'rubygems/command_manager'
-require 'rubygems/config_file'
 require 'rubygems/deprecate'
 
 ##
--- a/lib/rubygems/name_tuple.rb
+++ b/lib/rubygems/name_tuple.rb
@@ -4,8 +4,6 @@
 # Represents a gem of name +name+ at +version+ of +platform+. These
 # wrap the data returned from the indexes.
 
-require 'rubygems/platform'
-
 class Gem::NameTuple
 
   def initialize(name, version, platform="ruby")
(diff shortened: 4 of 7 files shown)


kares:
Thanks David, appreciate your feedback. Indeed it's been there for 8+ years.
Also made sense to me to just remove the `require` as it should act pretty much the same for consumers. 
Maybe except if someone is checking `$LOADED_FEATURES` for *rubygems/specification* which sounds silly.

kares:
curious how risky this seems, would this end up in RGs 3.2 or could the change also lend in the 3.1 branch for next release?

deivid-rodriguez:
Only 3.2.

deivid-rodriguez:
I mean, do you need us to backport this to 3.1 for jruby?

kares:
> This sounds like an equivalent issue? Gem::Platform is also autoloaded in lib/rubygems.rb, and explicitly required from rubygems/specification.

Nice catch, thanks David. Seems like so, let me review all requires and update the PR.
Users likely use `Gem::Specification` API more so that might be the reason why we've only seen those, so far (from reports).

> I mean, do you need us to backport this to 3.1 for jruby?

JRuby's plan was to update RGs to latest 3.1 and patch (for 9.2.12) only if next RGs accepts this PR.
Since users might upgrade RGs not having the patch in next RGs likely makes it not worthy to work-around by patching RGs JRuby ships. And as to why it's not being fixed for real, there's a JRuby load-service rewrite pending and we did not want to spent too much time poking around the current impl, as we might lend a new one in 9.3.

deivid-rodriguez:
I'm merging this patch and it will be part of rubygems 3.2, that's for sure. 

I would also be fine with backporting it to 3.1. We could consider backporting the inverse patch to be 100% sure that nothing breaks, but I think it should be fine as it is. But backporting depends on @hsbt, which is the person who currently releases rubygems.

simi:
:information_source: I can prepare backport patch if needed.

deivid-rodriguez:
After reading @kares message again, I understand that as long as we include this with both 3.2, and the next 3.1 patch, it should be fine, right? No matter when we actually release those.

kares:
> I understand that as long as we include this with both 3.2, and the next 3.1 patch, it should be fine, right? No matter when we actually release those.

Yes, that would have been the plan. 

I did a review NOT to require autoload-ed dependencies and the changeset has grown.
This could be sold as a startup time improvement as well ... for all Rubies booting RGs.

Let me know what you guys think, does this seem breaking in any ways for the 3.1 backport?

simi:
I thought those requires are in place to support requiring only specific files if needed, but for example `rubygems/specification` is not loadable on its own. I do not see any regression in here.

:thinking: _Any idea why some files are required explicitly and some are required by autoload? Would it be possible to unify this actually?_

MSP-Greg:
> Any idea why some files are required explicitly and some are required by autoload?

Maybe required files should only be what's needed for RG's `Kernel#reguire` to work?  Everything else could be 'load on demand'?  Unfortunately, this can also affect CI.  Having CI/tests require a minimal number of lib files is a good goal, but often impractical...

deivid-rodriguez:
Yeah, I can only imagine really weird use cases where this would be breaking. I'm fine with merging this as is, and even backporting it.

PR #3762, opened 2020-06-28
Avoid calling LoadError#message to not trigger DidYouMean

casperisfine:
Ref: https://github.com/ruby/did_you_mean/commit/078e9b625c887cabb9d88284a73084f260316dcd
Ref: https://github.com/rubygems/rubygems/pull/3660

While profiling our application boot against `ruby-head` I noticed it spend a lot of time in `DidYouMean`:

```
==================================
  Mode: cpu(1000)
  Samples: 59251 (14.73% miss rate)
  GC: 18583 (31.36%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
...
       929   (1.6%)         929   (1.6%)     DidYouMean::Jaro.distance
```

```
==================================
  Mode: wall(1000)
  Samples: 61040 (14.92% miss rate)
  GC: 18959 (31.06%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
...
       939   (1.5%)         939   (1.5%)     DidYouMean::Jaro.distance
```

Overall if I look at `DidYouMean::RequirePathChecker#corrections` specifically, it's `1.7%` of boot time
```
DidYouMean::RequirePathChecker#corrections (/usr/lib/ruby-shopify/ruby-shopify-head/lib/ruby/2.8.0/did_you_mean/spell_checkers/require_path_checker.rb:25)
  samples:     6 self (0.0%)  /    982 total (1.7%)
```

So the idea here is to do like Active Support when it parses `NameError` message, use `original_message` to skip the correction lookup.

cc @yuki24 @deivid-rodriguez

### DIFF
--- a/bundler/lib/bundler/runtime.rb
+++ b/bundler/lib/bundler/runtime.rb
@@ -43,14 +43,6 @@ module Bundler
       self
     end
 
-    REQUIRE_ERRORS = [
-      /^no such file to load -- (.+)$/i,
-      /^Missing \w+ (?:file\s*)?([^\s]+.rb)$/i,
-      /^Missing API definition file in (.+)$/i,
-      /^cannot load such file -- (.+)$/i,
-      /^dlopen\([^)]*\): Library not loaded: (.+)$/i,
-    ].freeze
-
     def require(*groups)
       groups.map!(&:to_sym)
       groups = [:default] if groups.empty?
@@ -79,16 +71,14 @@ module Bundler
             end
           end
         rescue LoadError => e
-          REQUIRE_ERRORS.find {|r| r =~ e.message }
-          raise if dep.autorequire || $1 != required_file
+          raise if dep.autorequire || e.path != required_file
 
           if dep.autorequire.nil? && dep.name.include?("-")
             begin
               namespaced_file = dep.name.tr("-", "/")
               Kernel.require namespaced_file
             rescue LoadError => e
-              REQUIRE_ERRORS.find {|r| r =~ e.message }
-              raise if $1 != namespaced_file
+              raise if e.path != namespaced_file
             end
           end
         end


yuki24:
@casperisfine Great catch! You might have tried it already but I wonder if it's possible to leverage the `LoadError#path` here as well. This seems like 10-year old code that's probably very hard to maintain... cc @deivid-rodriguez

deivid-rodriguez:
Yeah, I was having the same thought. Ideally we would completely get rid of the parsing of the specific wording of errors, so we could do something like 

```diff
diff --git a/bundler/lib/bundler/runtime.rb b/bundler/lib/bundler/runtime.rb
index 93a801eb6c..a58a8386e9 100644
--- a/bundler/lib/bundler/runtime.rb
+++ b/bundler/lib/bundler/runtime.rb
@@ -43,14 +43,6 @@ def setup(*groups)
       self
     end
 
-    REQUIRE_ERRORS = [
-      /^no such file to load -- (.+)$/i,
-      /^Missing \w+ (?:file\s*)?([^\s]+.rb)$/i,
-      /^Missing API definition file in (.+)$/i,
-      /^cannot load such file -- (.+)$/i,
-      /^dlopen\([^)]*\): Library not loaded: (.+)$/i,
-    ].freeze
-
     def require(*groups)
       groups.map!(&:to_sym)
       groups = [:default] if groups.empty?
@@ -79,16 +71,14 @@ def require(*groups)
             end
           end
         rescue LoadError => e
-          REQUIRE_ERRORS.find {|r| r =~ e.message }
-          raise if dep.autorequire || $1 != required_file
+          raise if dep.autorequire || e.path != required_file
 
           if dep.autorequire.nil? && dep.name.include?("-")
             begin
               namespaced_file = dep.name.tr("-", "/")
               Kernel.require namespaced_file
             rescue LoadError => e
-              REQUIRE_ERRORS.find {|r| r =~ e.message }
-              raise if $1 != namespaced_file
+              raise if e.path != namespaced_file
             end
           end
         end
```

Only concern would be if someone is directly using directly the `REQUIRE_ERRORS` constant for some reason (since it's at the same level as public methods such as `Bundler.setup`), so we could be more conservative and use `deprecate_constant` and nodoc it. But not sure if it's really needed.

casperisfine:
Yeah, I figured if it was possible when I edited the code, but opted for the most straightforward fix, because I didn't have much context.

I updated the PR to use `#path` instead.

yuki24:
Code LGTM, and the test failure doesn't to be related to the change. I wonder if one of you could restart the build to see if it fixes itself?

deivid-rodriguez:
Yeah, it wasn't related and I actually removed that entry from CIs in PRs in #3764 to avoid these issues. So rebasing should do the trick.

deivid-rodriguez:
Anyways, this should be fine, no need to rebase!

PR #3763, opened 2020-06-28
Stop using deprecated OpenSSL::Digest constants

bdewater:

Prepare for the deprecations planned in https://github.com/ruby/openssl/pull/366

Refactoring and moving a few things around, making a `create_digest` helper method to account for the cases where OpenSSL is not available.

### DIFF
--- a/lib/rubygems/security.rb
+++ b/lib/rubygems/security.rb
@@ -341,14 +341,7 @@ module Gem::Security
   ##
   # Used internally to select the signing digest from all computed digests
 
-  DIGEST_NAME = # :nodoc:
-    if defined?(OpenSSL::Digest::SHA256)
-      'SHA256'
-    elsif defined?(OpenSSL::Digest::SHA1)
-      'SHA1'
-    else
-      'SHA512'
-    end
+  DIGEST_NAME = 'SHA256' # :nodoc:
 
   ##
   # Algorithm for creating the key pair used to sign gems
--- a/lib/rubygems/security.rb
+++ b/lib/rubygems/security.rb
@@ -339,26 +339,15 @@ module Gem::Security
   class Exception < Gem::Exception; end
 
   ##
-  # Digest algorithm used to sign gems
-
-  DIGEST_ALGORITHM =
-    if defined?(OpenSSL::Digest::SHA256)
-      OpenSSL::Digest::SHA256
-    elsif defined?(OpenSSL::Digest::SHA1)
-      OpenSSL::Digest::SHA1
-    else
-      require 'digest'
-      Digest::SHA512
-    end
-
-  ##
   # Used internally to select the signing digest from all computed digests
 
   DIGEST_NAME = # :nodoc:
-    if DIGEST_ALGORITHM.method_defined? :name
-      DIGEST_ALGORITHM.new.name
+    if defined?(OpenSSL::Digest::SHA256)
+      'SHA256'
+    elsif defined?(OpenSSL::Digest::SHA1)
+      'SHA1'
     else
-      DIGEST_ALGORITHM.name[/::([^:]+)\z/, 1]
+      'SHA512'
     end
 
   ##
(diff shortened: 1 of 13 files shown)


bdewater:
The `ubuntu / ubuntu (truffleruby-head)` test timed out on installing dependencies :(

lib/rubygems/security.rb
```diff
@@ -338,27 +338,16 @@ module Gem::Security
...
-    if DIGEST_ALGORITHM.method_defined? :name
-      DIGEST_ALGORITHM.new.name
+    if defined?(OpenSSL::Digest::SHA256)
+      'SHA256'
+    elsif defined?(OpenSSL::Digest::SHA1)
+      'SHA1'
     else
-      DIGEST_ALGORITHM.name[/::([^:]+)\z/, 1]
+      'SHA512'
     end
```
rhenium, line 351: All code paths using Gem::Security::DIGEST_{ALGORITHM,NAME}, or Gem::Security in general, seem to have required OpenSSL anyway - Gem::Security::Signer#sign wouldn't work if DIGEST_ALGORITHM is ::Digest::SHA512.

Since `defined?(OpenSSL::Digest::SHA256)` is never false when OpenSSL is available, I think this can simply be

```
DIGEST_NAME = 'SHA256'
```

bdewater: I wasn't entirely sure of the original logic but this is the closest I could come up with while fixing the deprecations and keeping the behaviour intact.

I dug a little deeper in the history:
- [Allow RubyGems tests to run without OpenSSL (2013)](https://github.com/rubygems/rubygems/commit/0bb225a6063df0448a28f8bfe147422f0d7d1c24)
- [Gem::Security::DIGEST_ALGORITHM updated to SHA-256. (2016)](https://github.com/rubygems/rubygems/commit/d35a6c3b02f5b1d020c27b8ef2d412fdbcafde4b)
- [Fallback to Digest::SHA512 (2018)](https://github.com/rubygems/rubygems/commit/9471f8ed2bdc12248d2619bbbce6e53cd6c16cb6) - this one surprised me, why not `Digest::SHA256` instead?

I think your suggestion of `DIGEST_NAME = 'SHA256'` is safe to do since the `create_digest` method now takes care of situations where OpenSSL is unavailable. It also decouples algorithm choice from implementation availability - SHA256 is supported in both `Digest` and `OpenSSL::Digest`, and keeps the TarWriter specific fallback contained to that use case. I will add another commit with this refactor.

eregon:
FYI it seems this PR broke TruffleRuby: https://github.com/oracle/truffleruby/issues/2040
I'll investigate.

deivid-rodriguez:
Just to clarify, this PR didn't "break TruffleRuby". It helped detecting a issue in truffleruby, completely independent from rubygems. This is the exact reason why I moved this job to a daily build. We want to help implementations early detect issues that could affect us, but we don't want external bugs in unreleased rubies to affect our PRs.

eregon:
To clarify, I meant it purely as a factual thing. Latest RubyGems worked on TruffleRuby before this PR, not after.
Upcoming fix in TruffleRuby in https://github.com/oracle/truffleruby/issues/2040.
It's unfortunate there [was a timeout](https://github.com/rubygems/rubygems/runs/816434717?check_suite_focus=true) (maybe in GH, maybe TruffleRuby, I'm sorry in any case) and as a result truffleruby-head was moved to daily, which means we only detected later that this PR was causing a failure when running latest RubyGems on TruffleRuby.
IMHO release/unreleased doesn't matter, the latest TruffleRuby release would have the exact same issue (missing Digest::Instance#new).
Let's discuss in https://github.com/oracle/truffleruby/issues/2040 on what to do next, this particular issue will be fixed soon.

deivid-rodriguez:
> To clarify, I meant it purely as a factual thing.

:+1:

>  Latest RubyGems worked on TruffleRuby before this PR, not after.

Correct, although it'd be more accurate to say "Latest RubyGems worked on TruffleRuby _head as of today_ before this PR, not after.

> IMHO release/unreleased doesn't matter, the latest TruffleRuby release would have the exact same issue (missing Digest::Instance#new).

If we were officially supporting any truffleruby release (and by official support I mean "testing any official truffleruby release in CI", and don't merge PRs unless they are green), then yeah, a failure like this would've prevented this PR from being merged and we would've find a workaround to play nice with a ruby we already support. However, that's not the case here, I don't want to commit to support a random truffleruby-head build.

Hope this makes sense.

deivid-rodriguez:
> It's unfortunate there was a timeout (maybe in GH, maybe TruffleRuby, I'm sorry in any case) and as a result truffleruby-head was moved to daily, which means we only detected later that this PR was causing a failure when running latest RubyGems on TruffleRuby.

Also to clarify this point, this timeout only reminded me to move the matrix entry to the daily build, but it was not the motivation for it. I wanted to do that for a while since it's our current policy for "head testing" but I had forgotten to do it until now.

PR #3765, opened 2020-06-30
Fix remaining RuboCop issues

utkarsh2102:
With #3731 and #3740 merged, this covers up the
remaining part of the issues.
This was discovered when one tries to create a gem
with a different framework.
Could be reproduced with:
`bundle gem foo --ext --test=minitest --rubocop`

______________

# Tasks:

- [x] Write tests


utkarsh2102:
> I ideally we would extract the spec you added in the previous PR to a shared example, and reuse it for the different flag combinations. This is already done for several things in `bundle gem` specs. Would you like to give it a try?

Hi @deivid-rodriguez,
I already have and I think I have messed up something else in turn :P
(that's why I didn't push :sweat_smile:)

PR #3766, opened 2020-06-30
Allow setting a tag prefix to be used by gem release tasks

deivid-rodriguez:
# Description:

In order to be able to release rubygems & bundler from the same repo, we need to make sure that the generated tags don't overwrite each other. For example, we don't want the tag for the bundler 2.2.0 release to overwrite the tag for the rubygems 2.2.0 release.

My fix is to provide a `Bundler::GemHelper.tag_prefix=` setter to be able to have some control over the tag names. I plan to use it [like here](https://github.com/rubygems/rubygems/pull/3703/commits/71bd2487725d184124eb6648f977e9f7255c7602).

Honestly, I don't specially like this solution, but it fixes the issue and unblocks the releases. I'm happy to hear other proposals though.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -15,6 +15,10 @@ module Bundler
         new(opts[:dir], opts[:name]).install
       end
 
+      def tag_prefix=(prefix)
+        instance.tag_prefix = prefix
+      end
+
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Great! Merging this then! :heart:

PR #3745, opened 2020-06-20
Filter Actions workflows by path

MSP-Greg:
# Description:

Very few of the Actions CI jobs need to run if a commit/PR affects either Bundler or RubyGems files.  Most jobs need to be run only if one or the other changes.

## What was the end-user or developer problem that led to this PR?

CI takes too long, with unneeded jobs being run.

## What is your fix for the problem, implemented in this PR?

Using Actions `paths:` and `paths-ignore:`, workflows can be ran based on which files have changed in a PR/commit.  Use filters `"bundler/**"` and `".github/workflows/*bundler*.yml"` to positive filter for Bundler jobs, and use them with negative filter for RubyGems jobs.

Added a push branch filter of `"pr-**"`, so contributors can run CI in the forks, as long as the branch starts with `pr-`. Presently, the only way CI runs in a fork is from the `master` branch.

Closes #3735
______________

# Tasks:

- [x] Describe the problem / feature
- [x] Ran tests
- [x] Write code to solve the problem
- [ ] Get code review from coworkers / friends

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #3776, opened 2020-07-01
Optimize CI

### LINKED ISSUE #3735: GitHub Actions jobs - filter by path?
Actions allows filtering jobs by path.  See:
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths

There are an awful lot of jobs being run, and the workflows don't seem to indicate that all RG jobs need to run for a Bundler change, and vice versa.

So, I see three types of jobs:
1. Jobs that only need to run with a RubyGems change
2. Jobs that only need to run with a Bundler change
3. Jobs that need to run with any change

With some reorganization, this could be corrected.  It may require splitting jobs in workflows, or adding additional workflows, but it would quicken CI time quite a bit...

Thoughts?

deivid-rodriguez:
# Description:

Many of our CI matrix entries take care of testing bundler's code against different (old) versions of rubygems. In these cases, if only code in rubygems changes, it's pointless to run all these builds, since since the code being changed is not exercised by any of them (they only run old rubygems, not the code that we have changed).

So, we can skip these builds in the case where only rubygems files have been changed by a PR.

Supersedes #3745.

Closes #3735.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/older-rubygems-bundler.yml
+++ b/.github/workflows/older-rubygems-bundler.yml
@@ -1,0 +1,62 @@
+name: older-rubygems-bundler
+
+on:
+  pull_request:
+
+  push:
+    branches:
+      - master
+      - 3.2
+
+jobs:
+  older_rubygems_bundler:
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        ruby: [ 2.3.8, 2.4.10, 2.5.8, 2.6.6, 2.7.1 ]
+        rgv: [ v2.5.2, v2.6.14, v2.7.10, v3.0.8, v3.1.4 ]
+        bundler: [ '' ]
+        exclude:
+          - { bundler: '', ruby: 2.4.10, rgv: v2.5.2  }
+          - { bundler: '', ruby: 2.5.8, rgv: v2.5.2  }
+          - { bundler: '', ruby: 2.5.8, rgv: v2.6.14 }
+          - { bundler: '', ruby: 2.6.6, rgv: v2.5.2  }
+          - { bundler: '', ruby: 2.6.6, rgv: v2.6.14 }
+          - { bundler: '', ruby: 2.6.6, rgv: v2.7.10 }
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I set this as a draft since I want to do some tests to make sure it works as expected.

deivid-rodriguez:
Ok, so this works but there's another problem derivated from having made all checks "required" (in addition to the one detected in https://github.com/rubygems/rubygems/pull/3748#issuecomment-652223902, where man pages can't be synced automatically).

The problem is that since sometimes some workflows are skipped, some required checks always show as pending, as the current status checks in this PR show.

I think all maintainers are responsible enough to not merge PRs with failing statuses, and a bad merge to master wouldn't be the end of the world anyways, so I would propose to uncheck the boxes again and go back to how we were working before I cheked them a few days ago.

deivid-rodriguez:
@hsbt This is ready. I hope it is now more clear than #3745. 

The idea is that if a PR only changes files inside `rubygems` code, we need to run everything _except for_ `bundler` specs that run against older versions of `rubygems`, because in that case, the proper version of `rubygems` is checked out at `tmp/rubygems`, and the changed files in rubygems are not actually run.

Given how heavy the `bundler` specs are, this can improve CI feedback in certain situations.

To demostrate how this works, I added several dummy changes to this PR. After the last commit, since now the PR only edits `lib/rubygems.rb`, all bundler specs against older versions were skipped, so 37 jobs were run instead of the standard 56.

Let me know what you think.

deivid-rodriguez:
I removed all the dummy commits testing that this was working properly. Setting it as ready.

deivid-rodriguez:
I'm going to merge this because it should skip ~20 unnecessary jobs (of ~20' each) for each push on PRs only changing code in `rubygems`. We can resolve concerns (or fully revert if I'm missing something) on followup PRs.

PR #3778, opened 2020-07-02
Fix setting build metadata in development

deivid-rodriguez:
# Description:

This code should've changed when we merged the repos, but we forgot to do it.

Before this PR:

```
$ bin/bundle version
Bundler version 2.2.0.dev (2020-07-02 commit 54d9c28b)
(master) deivid@chaba ~/Code/rubygems/rubygems/bundler $ git show 54d9c28b
tree 54d9c28b

.gitignore
.rspec
CHANGELOG.md
LICENSE.md
README.md
Rakefile
UPGRADING.md
bin/
bundler.gemspec
dev_gems.rb
dev_gems.rb.lock
doc/
exe/
lib/
man/
spec/
task/
test_gems.rb
test_gems.rb.lock
tool/
```

After this PR:

```
$ bin/bundle version
Bundler version 2.2.0.dev (2020-07-02 commit 378d38a5c1)
$ git show 378d38a5c1 
commit 378d38a5c17dd710728513155f713e215798496a (HEAD -> build_metadata, origin/build_metadata)
Author: David Rodríguez <deivid.rodriguez@riseup.net>
Date:   Thu Jul 2 13:10:45 2020 +0200

    More refactoring of setting build metadata

diff --git a/bundler/spec/support/build_metadata.rb b/bundler/spec/support/build_metadata.rb
index e8994226bd..b49729fa33 100644
--- a/bundler/spec/support/build_metadata.rb
+++ b/bundler/spec/support/build_metadata.rb
@@ -9,7 +9,7 @@ module BuildMetadata
     include Spec::Helpers
 
     def write_build_metadata(build_metadata, dir: source_root)
-      replace_build_metadata(build_metadata.merge(:git_commit_sha => git_commit_sha))
+      replace_build_metadata(build_metadata.merge(:git_commit_sha => git_commit_sha, :built_at => loaded_gemspec.date.utc.strftime("%Y-%m-%d")))
     end
 
     def replace_build_metadata(build_metadata, dir: source_root)
diff --git a/bundler/spec/support/helpers.rb b/bundler/spec/support/helpers.rb
index b3239a3974..668526412d 100644
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -330,11 +330,7 @@ def with_built_bundler(version = nil)
 
         replace_version_file(version, dir: build_path) # rubocop:disable Style/HashSyntax
 
-        build_metadata = {
[...]

PR #3781, opened 2020-07-02
Remove fileutils autoload

deivid-rodriguez:
# Description:

We got another crash on jruby due to the issue they have where sometimes both explicitly autoloading and requiring something fails: https://github.com/rubygems/rubygems/pull/3778/checks?check_run_id=830626276.

In this case, I'd like to try removing the top level `autoload`, since this `autoload` keeps coming up and surprising me. I don't recall the situation, but I was definitely bitten by this inside `bundler` when I was expecting `Bundler`'s own copy of `fileutils` to be loaded, and instead getting the default version.

Let's run tests and see what fails.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Thanks @colby-swandale!

PR #3782, opened 2020-07-02
Don't authenticate github API through query parameters

deivid-rodriguez:
When releasing, I got an email from Github that this is deprecated.

This PR fixes it, while also improving some related stuff.

This is a follow up to https://github.com/rubygems/rubygems/pull/3678.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/task/release.rake
+++ b/bundler/task/release.rake
@@ -32,7 +32,7 @@ task "release:rubygem_push" => ["release:verify_docs", "release:verify_github", 
 namespace :release do
   task :verify_docs => :"man:check"
 
-  def gh_api_post(opts)
+  def gh_api_authenticated_request(opts)
     require "netrc"
     require "net/http"
     require "json"
@@ -70,7 +70,7 @@ namespace :release do
 
   task :verify_github do
     require "pp"
-    gh_api_post :path => "/user"
+    gh_api_authenticated_request :path => "/user"
   end
 
   def confirm(prompt = "")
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
In ruby 2.8, [this will be much easier](https://bugs.ruby-lang.org/issues/16686), by the way :smile:.

bundler/task/release.rake
```diff
@@ -65,9 +68,10 @@ namespace :release do
     JSON.parse(response.body)
   end
 
+  desc "Make sure github API is ready to be used"
   task :verify_github do
     require "pp"
-    gh_api_post :path => "/user"
+    PP.pp gh_api :path => "/user"
```
colby-swandale, line 74: Is this `PP.pp`for debugging or intentional?

deivid-rodriguez: It's intentional, just wanted to give a hint that the authentication went fine, but it's not really necessary since if there's an error, I guess it will be displayed.

colby-swandale: You're the only person ever using this, so you're free to basically do whatever you want ¯\_(ツ)_/¯. I was just checking

deivid-rodriguez: Thanks @colby-swandale! This has actually been used for a quite a few releases now, since https://github.com/rubygems/bundler/commit/8e8b8142c98254d42c710f06627624fe0126b104.

Anyways, I'm not really fond of the print and I don't think it adds much value, so I will remove it.

deivid-rodriguez: I updated the PR to remove the `require` instead. Thanks for the feedback!

PR #3785, opened 2020-07-03
Push only the release tag

deivid-rodriguez:

Currently `rake release` pushes all local tags by default, which might lead to unexpected results sometimes, for example in the case where you have a custom local remote with more tags than the remote that you're pushing to.

I believe what everyone expects here is the only the release tag to be pushed.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Making changes to bundler without running specs locally never works for me :laughing:. I'll fix this.

deivid-rodriguez:
Thanks for your reviews @colby-swandale :purple_heart:

PR #3786, opened 2020-07-04
Fix daily notifications

deivid-rodriguez:
# Description:

This should definitely fix #3689, I hope.

The problem with the previous approach is that we were checking the last workflow run status, but since at the time we check this, the workflow itself is running, we were getting that run (with status "in_progress" and conclusion "null"). Instead, we want the run previous to the current one.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Merging before daily CI starts failing again and we miss it :smiley:

Issue #2907, opened 2019-09-09
CI Use - stop multi-line output of gem update --system


MSP-Greg:
Some CI providers collapse output in their logs, others do not.  Some update RG & Bundler for older Ruby releases, some do not.

While running `gem update --system` in CI, it outputs over 1,000 lines of very helpful version history.

I tried `--no-post-install-message` which didn't do anything.  I tried `--silent`, which threw a `invalid option: --silent` error and stopped the update, even though it appears in `gem update --help` as a common option.

When one is wrestling with intermittent tests, differing results by CI provider and OS, everything sets one off.

Please tell me there is some way to stop the output?

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
@MSP-Greg what version of rubygems is CI using?

MSP-Greg:
It varies.  I believe both Travis & AppVeyor update RubyGems & Bundler in their Ruby builds.

GitHub Actions does not.  An example at https://github.com/puma/puma/pull/1918/checks?check_run_id=218917392.  Every 'RubyGems, Bundler Update' step has some output.

I apologize, as the output varies depending on the version one upgrades from.

Regardless, it would be nice to be able to silence it...

MSP-Greg:
Maybe silence the output if `ENV['CI']` exists?

bronzdoc:
@MSP-Greg  does `gem update --system --silent` doesn't silence output with the last Rubygems version for you?

MSP-Greg:
@bronzdoc

Here's a GtiHub Actions build: https://github.com/MSP-Greg/puma/runs/219228112 with `--silent` added.

Beerzaa304:
> Will give it a look @MSP-Greg thanks

duckinator:
I did some poking around about this.

`gem update --system` basically installs the rubygems-update gem, `cd`s to the directory containing that gem, and runs `ruby --disable-gems setup.rb <args>` where `<args>` are determined by `Gem::Commands::UpdateCommand#update_rubygems_arguments`.

However, if you look at that function, it does not preserve `--silent`: https://github.com/rubygems/rubygems/blob/master/lib/rubygems/commands/update_command.rb#L256-L265

But it looks like setup.rb, when run manually, does support it.

So I think, basically, that function may just need something along the lines of `args << '--silent' if options[:silent]`?

---

[...]

PR #3789, opened 2020-07-04
Have "gem update --system" pass through the --silent flag.

duckinator:
# Description:

Attempt at fixing #2907.

`gem update --system --silent` did not work.

If `gem update --system` gets `--silent`, pass it through to `setup.rb`.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -283,6 +283,7 @@ command to remove old versions.
 
   def update_rubygems_arguments # :nodoc:
     args = []
+    args << '--silent' if options[:silent]
     args << '--prefix' << Gem.prefix if Gem.prefix
     args << '--no-document' unless options[:document].include?('rdoc') || options[:document].include?('ri')
     args << '--no-format-executable' if options[:no_format_executable]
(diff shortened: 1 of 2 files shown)


duckinator:
I'm not sure how to test this, even manually, so I haven't confirmed this works.

cc @deivid-rodriguez @MSP-Greg

MSP-Greg:
> I'm not sure how to test this, even manually

Apply the patch to your current RG install, then `gem update --system --silent`.

It works, and it's really silent, as in nothing.  I have not looked at the code, something might be helpful, as in:
```
Bundler 2.1.4 installed
RubyGems 3.1.4 installed
RubyGems installed the following executables:
        C:/Greg/Ruby25-x64/bin/gem
        C:/Greg/Ruby25-x64/bin/bundle
```

But, one can always see that by not using `--silent`, so not important...  Thanks.

MSP-Greg:
This is off-topic, but instead of all the standard output that is essentially a change log, might that just be replaced with a link/URI?  Just output the info shown above and the link...

duckinator:
@MSP-Greg I would like something between the changelog-dump and silence, too, but yeah that's a separate issue.

bronzdoc:
@duckinator you might want to add a test in `test_gem_commands_update_command.rb`, alot of the examples there set the `system` flag so as David mentioned is a matter of test for an empty output.

Thanks so much for this ❤️

duckinator:
I think this is ready now. :+1:

PR #3792, opened 2020-07-06
Automatic management of the changelog

deivid-rodriguez:
# Description:

One of the biggest pain points of releasing bundler is having to go through each unreleased PR and manually write the changelog if applicable. This is really tedious and takes precious time.

In this PR I propose a change in our workflows to automatically manage the changelog.

The workflow is as follow:

* PR is opened.
* PR is created by default with the `skip-bundler-changelog` label (this should be changed at https://github.com/rubygems/issue-triage and the label created). This is so that if no actions are taken, the PR is not included in the changelog by default.
* Review process happens until is ready to merge.
* Maintainers evaluate the PR and replace the `skip-bundler-changelog` label with any of the following labels if it makes sense to do so. The following is a mapping of PR labels to changelog sections (a couple of them would need to be created), so that if a PR with one of the labels is merged, a corresponding changelog entry is created:

    | Label | Changelog section |
    | ------- | ------------------------- |
    | type: security | Security: |
    |  type: major bump | Breaking Changes: |
    |  type: highlight | Highlights: |
    |  type: deprecation | Deprecations: |
    |  type: feature implementation | Features: |
    |  type: performance | Performance: |
    |  type: documentation | Documentation: |
    |  type: improvement | Improvements: |
    |  type: bug fix | Bugfixes: |

* Maintainer edits the PR title to properly describe the PR if necessary.
* Maintaner merges the PR.

After the PR is merged, the following should happen automatically:

* A new release draft is created and pushed to github if the merged PR doesn't include the `skip-bundler-changelog` label.
* The unreleased section of the changelog is modified to include the title of the recently merged PR and committed back to the default branch. 

By doing it this way, we reduce a lot the amount of work needed to release bundler, because the changelog will always be ready.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/release-drafter-bundler.yml
+++ b/.github/release-drafter-bundler.yml
@@ -1,0 +1,75 @@
+---
+
+template: |
+  ## (Unreleased)
+
+  $CHANGES
+
+name-template: "bundler-v$RESOLVED_VERSION"
+
+tag-template: "bundler-v$RESOLVED_VERSION"
+
+categories:
+  - title: "Security fixes:"
+    labels:
+      - "bundler: security fix"
+
+  - title: "Breaking Changes:"
+    labels:
+      - "bundler: breaking change"
+
+  - title: "Major enhancements:"
+    labels:
+      - "bundler: major enhancement"
+
+  - title: "Deprecations:"
+    labels:
+      - "bundler: deprecation"
+
+  - title: "Features:"
+    labels:
+      - "bundler: feature"
+
+  - title: "Performance:"
+    labels:
+      - "bundler: performance"
+
+  - title: "Documentation:"
+    labels:
+      - "bundler: documentation"
+
+  - title: "Minor enhancements:"
+    labels:
+      - "bundler: minor enhancement"
+
+  - title: "Bug fixes:"
+    labels:
+      - "bundler: bug fix"
+
+change-template: "  - $TITLE [#$NUMBER]($URL)"
+
+no-changes-template: "  No changes."
+
+include-labels:
+  - "bundler: security fix"
+  - "bundler: breaking change"
+  - "bundler: major enhancement"
+  - "bundler: deprecation"
+  - "bundler: feature"
+  - "bundler: performance"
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
Yeah!

This PR will both keep the `CHANGELOG.md` file up to date as PRs are merged, and keep a Github Release Draft in sync with it using [release-drafter](https://github.com/release-drafter/release-drafter).

Now I need to figure out how to try all this stuff out :sweat_smile:.

deivid-rodriguez:
Oh, also (something I added in the last minute), it includes a separate workflow that gets run in case we manually update the `CHANGELOG.md` file. For example, because we forgot to tag a PR. In that case, the workflow should sync the release draft with the changes made in the changelog file.

deivid-rodriguez:
Docs added!

bronzdoc:
In Gemstash we use citrus `https://github.com/rubygems/gemstash/blob/master/rake/changelog.citrus` for this if you feel like exploring it 👀

deivid-rodriguez:
Thanks @bronzdoc! I didn't know gemstash could autogenerate its changelog, nice! The difference with my approach is that instead of a one-off task to be run manually before releases, with this approach the task is run automatically by our CI's so that the changelog is always up to date.

The thing I don't like about this PR is the need to label _every_ PR with `skip-bundler-changelog` by default. It feels weird. I want to provide an alternative which is:

* Our label criteria is: `bundler: <issue_type>` or `rubygems: <issue_type>`, where the possible issue types are the ones defined in #3807.
* By default all PRs are labelled with either `bundler: triage` or `rubygems: triage`.
* The changelog CI scripts defined in this PR only consider PRs labeled with `bundler: <issue_type>`.

Thoughts?

deivid-rodriguez:
I'm going to convert the set of tags we have now to follow the mentioned criteria and then rebase this PR to catch up with that.

deivid-rodriguez:
Closing since I'm introducing these changes through separate smaller PRs.

PR #3794, opened 2020-07-06
Revert the glob_files_in_dir optimization

casperisfine:
It was introduced in https://github.com/rubygems/rubygems/pull/2336

At that time it was using `File.join` and might have been faster.

But since https://github.com/rubygems/rubygems/pull/2536 it now use
`File.expand_path` and it bench slower on my machine:

```ruby
require 'benchmark/ips'

def old_glob_files_in_dir(glob, base_path)
  Dir.glob(File.expand_path(glob, base_path))
end

def new_glob_files_in_dir(glob, base_path)
  Dir.glob(glob, base: base_path).map! {|f| File.expand_path(f, base_path) }
end

Benchmark.ips do |x|
  x.report('old') { old_glob_files_in_dir("*.gemspec", "/opt/rubies/2.8.0-dev/lib/ruby/gems/2.8.0/specifications/default") }
  x.report('new') { new_glob_files_in_dir("*.gemspec", "/opt/rubies/2.8.0-dev/lib/ruby/gems/2.8.0/specifications/default") }
  x.compare!
end
```

```
                 old   679.000  i/100ms
                 new   507.000  i/100ms
                 old      6.830k (± 2.1%) i/s -     34.629k in   5.072317s
                 new      5.099k (± 1.5%) i/s -     25.857k in   5.072127s

Comparison:
                 old:     6830.0 i/s
                 new:     5099.1 i/s - 1.34x  slower
```

That's on MRI 2.6.5 but 2.7.1 and 2.8.0-dev give similar results.

### DIFF
--- a/lib/rubygems/util.rb
+++ b/lib/rubygems/util.rb
@@ -99,11 +99,7 @@ module Gem::Util
   # returning absolute paths to the matching files.
 
   def self.glob_files_in_dir(glob, base_path)
-    if RUBY_VERSION >= "2.5"
-      Dir.glob(glob, base: base_path).map! {|f| File.expand_path(f, base_path) }
-    else
-      Dir.glob(File.expand_path(glob, base_path))
-    end
+    Dir.glob(File.expand_path(glob, base_path))
   end
 
   ##


casperisfine:
Looking at the original PR (https://github.com/rubygems/rubygems/pull/971), it's possible that there's something missing in my benchmark.

I used theses calling values because that's what I saw in the boot profile, sounds like the original intent was to parse the entire load path for minitest plugins, the different pattern might change the performance profile. I'll try to see if I can reproduce the old 10X difference.

deivid-rodriguez:
I adapted the benchmark to the original pattern, and it seems about 50% faster the way it's done now, so I'll close this.

```ruby
require "benchmark/ips"

class Hello
  def self.find_files_from_load_path_with_base_arg glob # :nodoc:
    $LOAD_PATH.map { |load_path|
      Dir.glob(glob_with_suffixes, base: load_path).map! {|f| File.expand_path(f, load_path) }
    }
  end

  def self.find_files_from_load_path_without_base_arg glob # :nodoc:
    $LOAD_PATH.map { |load_path|
      Dir.glob(File.expand_path(glob_with_suffixes, load_path))
    }
  end
end

Benchmark.ips do |x|
  x.report('with base arg') do
    Hello.find_files_from_load_path_with_base_arg search
  end
  
  x.report('without base arg') do
    Hello.find_files_from_load_path_without_base_arg search
  end
  
  x.compare!
end
```

```
       with base arg     1.371k i/100ms
    without base arg   925.000  i/100ms
       with base arg     13.510k (± 1.7%) i/s -     68.550k in   5.075447s
    without base arg      9.298k (± 1.6%) i/s -     47.175k in   5.074872s

Comparison:
       with base arg:    13510.1 i/s
    without base arg:     9298.3 i/s - 1.45x  (± 0.00) slower
```

PR #3795, opened 2020-07-07
Remove unnecessary `Gem.clear_paths` from spec setup

deivid-rodriguez:
# Description:

Bundler specs only rely on a modified rubygems environment where everything is under `tmp` for specs that shell out. For those, modifying the `GEM_PATH` and `GEM_HOME` environment variables so that they are properly setup when rubygems is loaded inside the subprocess is enough.

For unit specs, modifying the `rubygems` environment on the fly through `Gem.clear_paths` is not really needed, and can actually cause problems. For example, I had two versions of `diff-lcs` installed of my system and I was getting warnings like this:

```
WARN: Unresolved or ambiguous specs during Gem::Specification.reset:
      diff-lcs (>= 1.2.0, < 2.0)
WARN: Clearing out unresolved specs. Try 'gem cleanup <gem>'
Please report a bug if this causes problems.
```

The warnings were there for a good reason. If I ignored them and introduce a test failure on a unit spec, I would've got errors like the following where failing to require `diff/lcs` would mask the actual test failures:

```
Failures:

  1) Bundler::Fetcher::Index error handling when a OpenSSL::SSL::SSLError occurs behaves like the error is properly handled when a 401 response occurs and there was userinfo should raise a Bundler::Fetcher::BadAuthenticationError
     Failure/Error:
       expect { subject.specs(gem_names) }.to raise_error(Bundler::Fetcher::BadAuthenticationError,
         %r{Baad username or password for http://remote-uri.org})

     LoadError:
       cannot load such file -- diff/lcs
     Shared Example Group: "the error is properly handled" called from ./spec/bundler/fetcher/index_spec.rb:107
     # ./spec/bundler/fetcher/index_spec.rb:46:in `block (6 levels) in <top (required)>'
     # ./spec/spec_helper.rb:102:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:102:in `block (3 levels) in <top (required)>'
     # ./spec/support/helpers.rb:353:in `block in with_gem_path_as'
     # ./spec/support/helpers.rb:367:in `without_env_side_effects'
     # ./spec/support/helpers.rb:349:in `with_gem_path_as'
     # ./spec/spec_helper.rb:101:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:73:in `block (2 levels) in <top (required)>'
     # ./spec/support/rubygems_ext.rb:90:in `load'
     # ./spec/support/rubygems_ext.rb:90:in `gem_load_and_activate'
```

By removing this method call, I believe we open the door to eventually be able to run `bundler` specs through `bundler`, but I'll try that out another time.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -77,81 +77,80 @@ namespace :man do
   if RUBY_ENGINE == "jruby"
     task(:build) {}
   else
-    begin
-      Spec::Rubygems.gem_require("ronn")
-    rescue Gem::LoadError => e
-      desc "Build the man pages"
-      task(:build) { abort "We couldn't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
+    directory "man"
 
-      desc "Verify man pages are in sync"
-      task(:check) { abort "We couldn't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
-    else
-      directory "man"
+    index = Dir["man/*.ronn"].map do |source|
+      ronn = "man/#{File.basename(source)}"
+      roff = "man/#{File.basename(source, ".ronn")}"
 
-      index = Dir["man/*.ronn"].map do |source|
-        ronn = "man/#{File.basename(source)}"
-        roff = "man/#{File.basename(source, ".ronn")}"
-
-        file roff => ["man", ronn] do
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
It was not as easy as it seemed :sweat_smile:.

deivid-rodriguez:
The problem mentioned in this PR was actually fixed by #3981. Some changes in this PR still can be useful, but I'll propose them separately.

PR #3797, opened 2020-07-08
Remove all `syck` references

deivid-rodriguez:
# Description:

While working on a performance improvement, I stumbled upon an issue when (un)marshalling an object very early in the `ruby` process, and getting some errors about `Syck` constants not being defined.

After reading [this blog post](https://blog.rubygems.org/2011/08/31/shaving-the-yaml-yak.html), published almost 10 years ago already, my understanding is that this problem could come up in two ways:

* Rubygems.org serving "corrupted gemspecs". As far as I understand this was fixed in rubygems.org a lot time ago, since
https://github.com/rubygems/rubygems.org/pull/331.

* Clients having a ten years old gemspec cache with some of these bad gemspecs. In this case, there's no easy solution but I think ten years is enough and rebuilding the cache should do the trick.

So, I think it's time we remove this.

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/endpoint_specification.rb
+++ b/bundler/lib/bundler/endpoint_specification.rb
@@ -3,7 +3,6 @@
 module Bundler
   # used for Creating Specifications from the Gemcutter Endpoint
   class EndpointSpecification < Gem::Specification
-    ILLFORMED_MESSAGE = 'Ill-formed requirement ["#<YAML::Syck::DefaultKey'.freeze
     include MatchPlatform
 
     attr_reader :name, :version, :platform, :required_rubygems_version, :required_ruby_version, :checksum
@@ -129,13 +128,6 @@ module Bundler
 
     def build_dependency(name, requirements)
       Gem::Dependency.new(name, requirements)
-    rescue ArgumentError => e
-      raise unless e.message.include?(ILLFORMED_MESSAGE)
-      puts # we shouldn't print the error message on the "fetching info" status line
-      raise GemspecError,
-        "Unfortunately, the gem #{name} (#{version}) has an invalid " \
-        "gemspec.\nPlease ask the gem author to yank the bad version to fix " \
-        "this issue. For more information, see http://bit.ly/syck-defaultkey."
     end
   end
 end
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -641,10 +641,8 @@ EOF
     def eval_yaml_gemspec(path, contents)
       require_relative "bundler/psyched_yaml"
 
-      # If the YAML is invalid, Syck raises an ArgumentError, and Psych
-      # raises a Psych::SyntaxError. See psyched_yaml.rb for more info.
       Gem::Specification.from_yaml(contents)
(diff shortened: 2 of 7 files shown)


hsbt:
I have a bit of worries remove it. But I agree to remove it with the next major version up like RubyGems 4.0

deivid-rodriguez:
What are you specific worries? Is it bullet point number 2 or something else? If I understood this correctly none of the marsahlled data coming from rubygems.org should contain `Syck::DefaultKey` since https://github.com/rubygems/rubygems.org/pull/331 around 10 years ago.

hsbt:
In my understanding, After removing `Syck` hack, RubyGems can't unpack the old gem created by syck library.

If We can unpack the all of gems in rubygems without `Syck::DefaultKey`, I have no worries about this.

deivid-rodriguez:
I think you might be right. Old gems that were packaged with `syck` will still contain this `Syck::DefaultKey` thingy and rubygems won't be able to unpack them. I'll double check this to figure out what we can remove safely and what not.

hsbt:
👍 I'm positive to remove it if we have a safe transition with the old gems.

simi:
:thinking: seems similar to https://github.com/rubygems/rubygems/pull/3396

Would it be possible to mock `Syck::DefaultKey` if not present or remove it from gemspec (normalize) before reading?

deivid-rodriguez:
@simi I think that's what we've been doing for years :thinking:...

I never got back to this, but thinking of it again now, it seems to me that the only way of getting rid of all these workaround codes (what this PR is proposing) would be to deprecate installing "broken gems" by adding some deprecation messages in the places where we are workarounding `Syck::DefaultKey` inside gemspecs from corrupted gems.

bundler/lib/bundler/endpoint_specification.rb
```diff
@@ -129,13 +128,6 @@ def parse_metadata(data)
 
     def build_dependency(name, requirements)
       Gem::Dependency.new(name, requirements)
-    rescue ArgumentError => e
-      raise unless e.message.include?(ILLFORMED_MESSAGE)
-      puts # we shouldn't print the error message on the "fetching info" status line
-      raise GemspecError,
-        "Unfortunately, the gem #{name} (#{version}) has an invalid " \
-        "gemspec.\nPlease ask the gem author to yank the bad version to fix " \
-        "this issue. For more information, see http://bit.ly/syck-defaultkey."
```
deivid-rodriguez, line 138: Another observation about this is that although we seem to have been warning this situation for a long time, as far as I can see, this code is only reached from parsing the compact index "info" endpoint, which as far as I understand, can't serve this "weird dependencies".

Only situation where I think we could still be bitten by this is when unpacking a gem package and reading a corrupted gemspec inside it. It'd be nice to find such a corrupted gem package and run some experiments on it.

deivid-rodriguez:
We removed all `syck` references from `rubygems` and released it and still no reports of anything broken. I think we can move this one forward too.

hsbt (approved):

simi (approved):

PR #3807, opened 2020-07-10
Unify changelog sections in both libraries

deivid-rodriguez:
This PR intends to keep unifying workflows in the new rubygems + bundler monorepo.

In this case, I standarize changelogs to use the same set of sections.

The result is the union of the labels rubygems currently uses with the labels
bundler current uses, just to be able to make the unification without triggering
any discussion of which the sections should actually be.

The only exception I made to this is the remove the "Style changes" section from
the rubygems changelog, since in my opinion those changes should clearly not be
included since they don't provide any value to end users.

But if others disagree I'm happy to remove that from this PR and keep that
section for now.

On top of that there's some naming unification, like renaming the section I just
made up in bundler in the last rc release, "Hightlights", to match rubygems
"Major enhancements", which conveys the same intention.

The final set of sections for both repositories is now:

   * Breaking changes
   * Bug fixes
   * ~Compatibility changes~
   * Deprecations
   * Documentation
   * Features
   * Major enhancements
   * Minor enhancements
   * Performance
   * Security fixes

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -572,7 +572,6 @@ Bug fixes:
 
 Security fixes:
 
-* Fixed following vulnerabilities:
   * CVE-2019-8320: Delete directory using symlink when decompressing tar
   * CVE-2019-8321: Escape sequence injection vulnerability in `verbose`
   * CVE-2019-8322: Escape sequence injection vulnerability in `gem owner`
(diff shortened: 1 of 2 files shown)


hsbt:
The basically idea is fine. I have a question. What difference `Breaking changes` and `Compatibility changes`?

deivid-rodriguez:
No difference really. I only kept it because in the `rubygems` changelog they are listed as different sections but I'm happy to unify those as "Breaking changes".

deivid-rodriguez:
I made a quick audit of previous usages of "Compatibility changes" and moved them to more appropriate sections :+1:.

deivid-rodriguez:
Could you have a look?

PR #3808, opened 2020-07-10
Remove "style changes" section from `History.txt`

deivid-rodriguez:
# Description:

I proposed this inside #3807, but since it's a change that contradicts what @hsbt is currently doing when releasing `rubygems`, I think it's fair to propose it separately and have a separate discussion.

My opinion is that including this kind of changes in the changelog doesn't provide any value to end users. So we should not include them.

@hsbt What do you think?

# Tasks:

- [ ] Write tests

### DIFF
--- a/History.txt
+++ b/History.txt
@@ -329,20 +329,6 @@ Compatibility changes:
 * Add deprecation warnings for cli options. Pull request #2607 by Luis
   Sagastume.
 
-Style changes:
-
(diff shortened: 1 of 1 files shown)


hsbt:
👍 I'm okay to remove it. It's not related the user's benefit.

deivid-rodriguez:
Great!

PR #3811, opened 2020-07-10
Don't format executables on `gem update --system`

### LINKED ISSUE #3564: Bundler is installed as `jbundler` on `jruby` by rubygems installer
My current problem is that if you run rubygems installer (`gem update --system`) on `jruby`, it installs the `bundler` executable as `jbundler`, not as `bundler`. That means `bundler` is not accessible in the usual way via `bundle` and you need to run `jbundle` instead. This is unexpected, because when you run `gem install bundler`, this does not happen and a `bundle` binstub is installed, so `jruby` users are used to running `bundle`, just like they run `ruby` or `gem`, instead of `jruby` or `jgem`.

Note that this is only a problem since https://github.com/rubygems/rubygems/pull/2766 (shipped with rubygems 3.1), because previously no `:format_executable` option would be passed to the default `bundler` installer. So the behaviour in previous rubygems versions is the expected.

The behaviour has been always unexpected for the `gem` binstub (`jgem` is always installed). However, `jruby` workarounds the issue by [providing a `gem` binstub that delegates to `jgem`](https://github.com/jruby/jruby/blob/248a1d32cfacf59e9834a4bd0071c5ed46524b36/bin/gem).

The `:format_executable` makes sense for operating systems so that they can have several rubies installed at that same time, and executables for each of them installed as, for example, `gem27`, `bundler27`, and so on. So we don't want to revert https://github.com/rubygems/rubygems/pull/2766.

The question is, what should we do, then? There's some options:

* Change [the `setup` command default value to the `:format_executable`](https://github.com/rubygems/rubygems/blob/84e32f8c0c7e90a485c93cd61e8f11f7daf7a182/lib/rubygems/commands/setup_command.rb#L19) to `false`. That would match [the default value for the option in the `install` command](https://github.com/rubygems/rubygems/blob/84e32f8c0c7e90a485c93cd61e8f11f7daf7a182/lib/rubygems/commands/install_command.rb#L24) and it seems like a clean solution. However:
  - ~Since `jruby` already provides a "workaround binstub", changing this would mean overwriting this binstub on `gem update --system`, and rubygems will normally refuse to do that unless you pass the `--force` option.~ EDIT: The `--force` option is `true` by default for the `setup` command, so this shouldn't be a problem in principle.
  - Operating systems are most likely already configuring `install: --format-executable` in their `gemrc` files (or in the `operating_system.rb` script), but since `true` is the default for the `setup` command, they might not be doing the same thing for the `setup` command, and this change could break things.

* Do nothing and wait for `jruby` to provide a similar binstub for `jbundler` or to configure the proper `--format-executable` option for the setup command. 

In my opinion, the former option is cleaner and makes more sense, but requires more careful coordination with OS packagers and the `jruby` team. The latter option is obviously easier for us since it requires no changes on our side. So... :man_shrugging:.

Pinging @enebo from `jruby`, @terceiro from Debian, and @voxik from Fedora to gather some opinions.

deivid-rodriguez:
# Description:

This restores 3.0 behaviour and goes back to not formatting executables by default on `gem update --system`. This is friendlier for jruby and doesn't really affect OS packagers. And having changed this is actually a regression because it was never intended.

This is actually already tested by our `install-rubygems.yml` workflow, but we currently include a workaround in our "development `gem update --system` simulator task" to avoid the issue on jruby. This PR also removes the workaround.

Fixes #3564.

# Tasks:

- [ ] Write tests

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -100,7 +100,7 @@ end
 
 desc "Install rubygems to local system"
 task :install => [:clear_package, :package] do
-  sh "ruby -Ilib bin/gem install --no-document pkg/rubygems-update-#{v}.gem && update_rubygems --no-document --no-format-executable"
+  sh "ruby -Ilib bin/gem install --no-document pkg/rubygems-update-#{v}.gem && update_rubygems --no-document"
 end
 
 desc "Clears previously built package"
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -15,7 +15,7 @@ class Gem::Commands::SetupCommand < Gem::Command
     require 'tmpdir'
 
     super 'setup', 'Install RubyGems',
-          :format_executable => true, :document => %w[ri],
+          :format_executable => false, :document => %w[ri],
           :force => true,
           :site_or_vendor => 'sitelibdir',
           :destdir => '', :prefix => '', :previous_version => '',


deivid-rodriguez:
Merging this because it was consensuated with OS packagers and jruby.

Issue #3173, opened 2020-03-13
[Feature request: behaviour change for gem unpack] Less stringent requirement for openssl


rubyFeedback:
When doing:

    gem unpack foobar-0.1.2.gem 

I get this failure:

ERROR:  Loading command: unpack (LoadError)
        cannot load such file -- openssl
ERROR:  While executing gem ... (NoMethodError)
    undefined method `invoke_with_build_args' for nil:NilClass

So it fails because openssl is not installed.

Allow me to explain the situation. The above error happens NOT on my home computer.

My home computer works perfectly fine, but I am in a restricted university environment
too, with some old ubuntu version running.

Getting openssl to work on that machine, and the ruby bindings to openssl, is
annoying. I also have lots of local gems which I wanted to batch-install via
oldschool setup.rb. This also works; I have this in a standalone .rb file too
that does the job for me.

On that machine, though, I was unable to extract .gem files due to the openssl
requirement, and to me I do not quite understand why I need openssl if
I am to want to extract a local .gem file?

I worked around the above by just using "tar", moving the resulting .gz
into the directory, extracting it too, then using setup.rb. So perhaps openssl
should become not a hard dependency if a user wants to install a local
gem.

(Not sure whether the command should have been **gem unpack ./foobar-0.1.2.gem** 
instead but the guess could first check for a local gem anyway - I am not even
sure why we need openssl support for unpacking any gem, ever The tar command
already works as-is. This is what I do when gem does not work, so I don't 
quite understand why we need openssl at all times. It's ok for downloading a
new remote gem and stuff, but for local gems I just don't see the logic behind
it. Note that the above is not really a big problem, just inconvenient; I managed
to install all my local gems anyway, but it did take me a few extra steps - would
have been easier if "gem unpack *.gem" would have worked without the
dependency on openssl as-is)

PR #3816, opened 2020-07-11
Audit `openssl` requiring code

deivid-rodriguez:
# Description:

This PR reviews all code requiring openssl in both repos, and centralizes it as much as possible.

As a result, this PR fixes https://github.com/rubygems/rubygems/issues/3173, and closes https://github.com/rubygems/rubygems/pull/3059.

It also fixes https://bugs.ruby-lang.org/issues/16475.

I added a small test (just run `gem list bundler`) to make sure that bare functionality of `rubygems` still works without openssl.

# Tasks:

- [x] Write tests

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -14,6 +14,7 @@ jobs:
     strategy:
       matrix:
         ruby: [ 2.3.8, 2.4.10, 2.5.8, 2.6.6, 2.7.1, jruby-9.2.11.1 ]
+        openssl: [true, false]
     steps:
       - uses: actions/checkout@v2
       - name: Setup ruby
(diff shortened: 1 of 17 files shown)


hsbt:
This pull-request is useful.

But I'm still not sure why we should support non-openssl environment. Is there real use-case it?

deivid-rodriguez:
Looking at the issues that came up, very few people, but still _some_ people, have such environments.

I think we could consider fully dropping support for non-openssl environments on the next major version. But the way this got broken was completely unintentional and we still have code around (what I'm unifying here) to actually support non-openssl environments.

So, I think we can restore bare support for these environments for now (note that without this fix, as per the ruby-core issue, `ruby` won't even install without `openssl`).

Also, note that fixing this was only a side effect, my original intention was to unify the code that requires `openssl` to try to figure out the `jruby` failures we are currently experiencing (like https://github.com/rubygems/rubygems/runs/864749290?check_suite_focus=true). But since I was at it, I decided to fix this.

deivid-rodriguez:
This PR makes the code dealing with loading `openssl` more unified and clear, and also fixes some (admittedly rare) usages as a side effect. So I'll merge this after CI passes. We can consider fully dropping all usages without `openssl` for rubygems 4.

Issue #2945, opened 2019-10-08
Rubygems is installing the precompiled C extensions for MRI on TruffleRuby


rafaelfranca:
When installing a gem that provides precompiled binaries, as for example [google-protobuf](https://rubygems.org/gems/google-protobuf/versions/3.9.2-universal-darwin), RubyGems is installing the MRI binary even though it only works for MRI and not TruffleRuby.

It seems that it should be aware of the Ruby engine being used and not download the compiled gem if the engine used to compile it is not the same as the local engine.

I'm willing to fix the problem but I could not find the place where RubyGems detect if the platform the gem was built is the same as the one it is being installed, so any pointers would be appreciated.

Also, the `Gem::Platform` object lacks information about the `RUBY_ENGINE` being used in the local platform and in the spec platform, so I'm not sure if we will need a change to rubygems.org to provide this information on the spec that were already pushed in order to fix this issue.

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
RubyGems Environment:
  - RUBYGEMS VERSION: 3.0.3
  - RUBY VERSION: 2.6.2 (2019-10-03 patchlevel 0) [x86_64-darwin]
  - INSTALLATION DIRECTORY: /Users/rafaelfranca/.gem/truffleruby/2.6.2
  - USER INSTALLATION DIRECTORY: /Users/rafaelfranca/.gem/truffleruby/19.3.0-dev-7d425d48
  - RUBY EXECUTABLE: /Users/rafaelfranca/src/truffleruby/mxbuild/truffleruby-jvm/jre/languages/ruby/bin/truffleruby
  - GIT EXECUTABLE: /opt/dev/bin/user/git
  - EXECUTABLE DIRECTORY: /Users/rafaelfranca/.gem/truffleruby/2.6.2/bin
  - SPEC CACHE DIRECTORY: /Users/rafaelfranca/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /Users/rafaelfranca/src/truffleruby/etc
  - RUBYGEMS PLATFORMS:
    - ruby
    - x86_64-darwin
  - GEM PATHS:
     - /Users/rafaelfranca/.gem/truffleruby/2.6.2
     - /Users/rafaelfranca/src/truffleruby/mxbuild/truffleruby-jvm/jre/languages/ruby/lib/gems
[...]

PR #3817, opened 2020-07-11
Pass more information when comparing platforms

eregon:
* Passing the gem specification enables more flexibility
  when checking whether a gem is compatible
  with the various aspects of the "current platform"
  including OS, architecture, RUBY_ENGINE, etc.
* Relates to https://github.com/rubygems/rubygems/issues/2945
* Based on https://github.com/oracle/truffleruby/commit/ae83667c220f52cbea4e83c44d2b9f7027688a3a

As mentioned in #2945 and in other issues (notably about including the libc in the "platform"), the current Gem::Platform captures too little and is sometimes not enough to check if a precompiled extension is actually compatible with the local machine installing the gem.

One example is the `sassc` and `libv8` gems both have extensions which do not use the Ruby C API and are therefore independent of the Ruby implementation. So such precompiled "extensions" for such gems can be reused on TruffleRuby, JRuby, etc, while typical precompiled C extensions cannot due to having a different ABI.

Also consider that while `sassc` can be compiled from source, it's "just" rather slow.
But for `libv8` it's both hard to build (needs dependencies typically not installed by Rubyists) and it takes >15 minutes to compile (https://github.com/oracle/truffleruby/issues/1827#issuecomment-607386676).

This PR makes a step forwards by giving more information and therefore flexibility when checking whether a Gem's platform is compatible with the local machine.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -18,18 +18,38 @@ class Gem::Platform
   end
 
   def self.match(platform)
-    Gem.platforms.any? do |local_platform|
+    match_platforms?(platform, Gem.platforms)
+  end
+
+  class << self
+    extend Gem::Deprecate
+    rubygems_deprecate :match, "Gem::Platform.match_spec?"
+  end
+
+  def self.match_platforms?(platform, platforms)
+    platforms.any? do |local_platform|
       platform.nil? or
         local_platform == platform or
         (local_platform != Gem::Platform::RUBY and local_platform =~ platform)
     end
+  end
+  private_class_method :match_platforms?
+
+  def self.match_spec?(spec)
+    match_gem?(spec.platform, spec.name)
+  end
+
+  def self.match_gem?(platform, gem_name)
+    # Note: this method might be redefined by Ruby implementations to
+    # customize behavior per RUBY_ENGINE, gem_name or other criteria.
+    match_platforms?(platform, Gem.platforms)
   end
 
   def self.installable?(spec)
     if spec.respond_to? :installable_platform?
       spec.installable_platform?
     else
-      match spec.platform
+      match_spec? spec
     end
   end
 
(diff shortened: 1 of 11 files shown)


lib/rubygems/platform.rb
```diff
@@ -23,18 +19,40 @@ def self.local
...
 
   def self.match(platform)
-    Gem.platforms.any? do |local_platform|
+    match_platforms?(platform, Gem.platforms)
+  end
+
+  class << self
+
+    extend Gem::Deprecate
+    rubygems_deprecate :match, "Gem::Platform.match_spec?"
```
eregon, line 28: This might be annoying when using Bundler 1, and as long as there is the RubyGems Bundler version switcher (when removed, then Bundler 2+ would always be used and so just having a recent Bundler would avoid the issue).
Any idea how to "soft deprecate" this?

eregon:
@deivid-rodriguez Could you review this? Do you think we could merge it?

lib/rubygems/platform.rb
indirect: IMO if you are on a RubyGems new enough to get this deprecation it's ok to tell you to upgrade to Bundler 2? But open to hearing from people that would be a problem for. :)

eregon:
Thanks for merging!

lib/rubygems/platform.rb
eregon: One case I can imagine is using e.g. Ruby 2.8/3.0 or update RubyGems (which would include this code),
and then `bundle install` on a project with `Gemfile.lock` `BUNDLED WITH 1.x.y`.
Then Bundler 1 would be used (due to the version switcher) and the warning would be shown.

I think a good way to fix that is to remove the version switcher, so Bundler 2 would be used in that case.
BTW, since it seems Bundler 1 seems no longer maintained ([last release of Bundler 1](https://rubygems.org/gems/bundler/versions) in 2018), it seems important to use Bundler 2 for all cases if installed, even if the Gemfile.lock still has `BUNDLED WITH 1.x.y`.

eregon: I filed https://github.com/rubygems/rubygems/issues/3879 about this.

PR #3819, opened 2020-07-12
Let more exceptions flow

deivid-rodriguez:
# Description:

If any error happens while verifying a package entry, it doesn't mean
that the package is corrupt. It could be a bug in rubygems, for example.
This in fact happened in CI and the current error doesn't make it easy
to troubleshoot the root cause, since it doesn't provide a backtrace.

See
https://github.com/rubygems/rubygems/pull/3807/checks?check_run_id=862526615.

So I propose to let the exception happens. There was something useful
about the previous message, which is the file entry where the error
happened, so I'm keeping that information in a warning message.

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -674,10 +674,9 @@ EOM
     when 'data.tar.gz' then
       verify_gz entry
     end
-  rescue => e
-    message = "package is corrupt, exception while verifying: " +
-              "#{e.message} (#{e.class})"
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This issue keeps coming up in jruby's CI (https://github.com/rubygems/rubygems/pull/3820/checks?check_run_id=864749290), so I'm going to merge this for better troubleshooting.

PR #3820, opened 2020-07-13
Rubocop normalization

deivid-rodriguez:
# Description:

Another small step towards normalizing `rubygems` and `bundler` code style.

In this case:

* I enforced no empty lines around class body inside `rubygems` to normalize the code style with the one `bundler` uses.
* ~I enabled `Style/IfInsideElse` inside `rubygems` to normalize the code style with the one `bundler` uses.~ Ended up disabling the cop.

# Tasks:

- [ ] Write tests

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -108,6 +108,9 @@ Style/Encoding:
   Exclude:
     - test/rubygems/specifications/foo-0.0.1-x86-mswin32.gemspec
 
+Style/IfInsideElse:
+  Enabled: false
+
 Style/MethodCallWithoutArgsParentheses:
   Enabled: true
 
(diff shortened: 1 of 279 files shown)


.rubocop.yml
```diff
@@ -109,6 +108,9 @@ Style/Encoding:
   Exclude:
     - test/rubygems/specifications/foo-0.0.1-x86-mswin32.gemspec
 
+Style/IfInsideElse:
```
deivid-rodriguez, line 111: Alternatively, I'm also fine with disabling this rule in `bundler`. I feel this rule sometimes leads to better code, but sometimes it doesn't, so it's probably not worth bothering.

deivid-rodriguez: I ended up disabling the rule, but I don't have a strong opinion either way.

hsbt: 👍 I agree to disable it on bundler. `IfInsideElse` is too strict for me.

deivid-rodriguez:
Thanks @hsbt!

eregon:
Thanks for this! :)

PR #3822, opened 2020-07-14
Explicitly pass target version & tag to release drafter

deivid-rodriguez:
# Description:

Release drafter is not smart enough to figure out the version we want to release next. At least, not in the case of prereleases. Pass it explicitly instead.

In addtion to that, I only removed the "(Unreleased)" header from the release draft, since it doesn't provide any value.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Let's see if this works better.

PR #3826, opened 2020-07-14
Wrap up automatic changelog management

deivid-rodriguez:
# Description:

This PR is to wrap up automatic changelog management, fixing the remaining issues. These are:

*  Automatic commits to master are creating changelog categories as second level sections, where as in our current changelog they are not even headings. Since I think it looks better that way, I normalized the changelog to follow that pattern.

* One of the sections had a casing typo. It's "Breaking changes", no "Breaking changes".

* This PR adds an alternative workflow that syncronizes the github release draft with the changelog that will run on pushes to master that include direct changes to the change log. In those cases, the inverse case, namely syncronizing the changelog file from the github release draft is skipped. __There's an issue here because the current task is creating a different release draft with the same name instead of updating the current one_.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
All issues should be fixed now. I'll merge this after CI, and everything should be working fine.

Only remaining issue I think will happen when the synchronization of the changelog with the release draft leads to no changes at all an tries to create/push an empty commit. I'll fix that later.

deivid-rodriguez:
Let's see how it goes.

PR #3827, opened 2020-07-14
Migrate Github tasks to use octokit

deivid-rodriguez:
# Description:

Updating the prerelease draft in #3826 gets tricky because if you're not authenticated, you only get 15 releases by default, and we don't have a pagination implementation for authenticated requests. Instead of implemented this, I migrated the tasks to use [octokit](/github.com/octokit/octokit.rb) since it seemed  simpler.

# Tasks:

- [ ] Write tests


hsbt:
👍

Issue #3355, opened 2018-12-20
Using a plugin as a path doesn't work


Fryguy:
Application Gemfile has

```ruby
plugin 'bundler-inject', :path => File.expand_path("~/dev/bundler-inject")
```

`bundle update` gives me

```
Could not find gem 'bundler-inject' in any of the gem sources listed in your Gemfile.
```

Note that if I push a git repo somewhere and use the `:git` option, then it works fine.  Also note that the following alternate path notation doesn't work either.

```
path File.expand_path("~/dev/bundler-inject") do
  plugin 'bundler-inject'
end
```

Using bundler 1.17.2, but also didn't work with bundler 1.17.1

PR #3828, opened 2020-07-14
bring back the possibility to install a plugin from path

nijikon:
# Description:

Documentation states that it is possible to install a plugin from a `path`, but it's not working.

```ruby
plugin 'my_plugin', path: '/path/to/plugin' # Installs from a path
```

I was not able to install the plugin from a local path.

It allows us to use the `path` option again.

Fixes https://github.com/rubygems/rubygems/issues/3355.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/CHANGELOG.md
+++ b/bundler/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Bug fixes:
 
   - Fix broken exception recovery code when installing plugins [#3487](https://github.com/rubygems/rubygems/pull/3487)
+  - Bring back the possibility to install a plugin from path
 
 ## 2.2.0.rc.1 (Jul 02, 2020)
 
--- a/bundler/lib/bundler/plugin/dsl.rb
+++ b/bundler/lib/bundler/plugin/dsl.rb
@@ -10,7 +10,7 @@ module Bundler
       # So that we don't have to override all there methods to dummy ones
       # explicitly.
       # They will be handled by method_missing
-      [:gemspec, :gem, :path, :install_if, :platforms, :env].each {|m| undef_method m }
+      [:gemspec, :gem, :install_if, :platforms, :env].each {|m| undef_method m }
 
       # This lists the plugins that was added automatically and not specified by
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Hi, thanks!

Not saying that we don't want to introduce this (it makes sense to me), but I don't think this has ever been supported, so I wouldn't say "bring back". It sounds like the documentation is wrong.

Anyways, have you tried this? I was expecting it to be harder, at least requiring you to define a new plugin `Path` source. We'll definitely need tests here.

deivid-rodriguez:
I just saw this is a draft PR, sorry if my feedback comes too early :sweat_smile:

nijikon:
Didn't check the backlog to see if it was ever supported, I was convinced that it was based on the documentation.

I did check if it works, I really on this heavily when I develop our [diffend-ruby](https://github.com/diffend-io/diffend-ruby) plugin. I would not waste your time by pushing a change that was not tested manually.

Really liked that you gave feedback on this already, the only reason why it is a draft is that I didn't write the specs yet 😅

colby-swandale:
I'm 95% sure using `path` was unintentionally not working when the plugin system was shipped.

deivid-rodriguez:
That's good to know, let's call this a bug fix then :)

colby-swandale:
@nijikon can you just update the Changelog to say that we've fixed a bug of plugins not being installed via `path` and turn this into a PR.

ngan:
This would be awesome to land. We're trying to make a bundler plugin but it's excruciatingly painful to test and debug when you can only use a local git source.  The [docs](https://bundler.io/guides/bundler_plugins.html#install_from_command) state that you can use `path` and I was excited about not having to publish a plugin I'd like to use privately.

deivid-rodriguez:
@nijikon Do you still plan to work on this?

deivid-rodriguez:
@nijikon Went ahead and added a spec! Closing in favor of #4020.

PR #3830, opened 2020-07-14
add plugin hooks for bundle check

nijikon:
# Description:

I want to be able to execute a plugin when `bundle check` is run.

I'm not able to execute a plugin when `bundle check` is run.

Introduce `before-check-all` and `after-check-all` hooks when `bundle check` is run.

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/CHANGELOG.md
+++ b/bundler/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Bug fixes:
 
   - Fix broken exception recovery code when installing plugins [#3487](https://github.com/rubygems/rubygems/pull/3487)
+  - Add plugin hooks for `bundle check` command, `before-check-all` and `after-check-all`
 
 ## 2.2.0.rc.1 (Jul 02, 2020)
 
--- a/bundler/lib/bundler/cli/check.rb
+++ b/bundler/lib/bundler/cli/check.rb
@@ -14,10 +14,15 @@ module Bundler
       begin
         definition = Bundler.definition
         definition.validate_runtime!
+
+        Plugin.hook(Plugin::Events::GEM_BEFORE_CHECK_ALL, definition.dependencies)
+
         not_installed = definition.missing_specs
       rescue GemNotFound, VersionConflict
         Bundler.ui.error "Bundler can't satisfy your Gemfile's dependencies."
         Bundler.ui.warn "Install missing gems with `bundle install`."
+
+        Plugin.hook(Plugin::Events::GEM_AFTER_CHECK_ALL, definition.dependencies)
         exit 1
       end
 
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
Hi @nijikon! :wave:

Could you explain your use case? As I understand it, plugin hooks are to allow hooking into internal events, like `bundler` installing or requiring a gem, not for hooking into the execution of specific CLI commands. If we were to add something like this, wouldn't we need to add plugin hooks to signal the execution of every single CLI command? That doesn't sound right to me. Maybe your use case is valid, but there's a better place for the hooks you want to add that's not inside the implementation of a particular CLI command. Or maybe you need to do whatever you need to do, and then call `bundle check`, or viceversa, and there's no need to change `bundler` at all.

nijikon:
I'm sorry that I didn't describe my use case at all. Let me fix that.

Let's say that you have a project in which you don't update gems. When you run your CI build, you are using `bundle check || bundle install` to check if all the dependencies are satisfied, so you don't have to run `bundle install`.

What we do in our [diffend-ruby](https://github.com/diffend-io/diffend-ruby) plugin, we use `before-install-all` hook, to be able to protect you so you won't install a malicious gem to your local/staging/production environment.

The problem starts when you don't update your gems and use `bundle check` command. There is no way to hook up to the runtime to run the security checks. That's why I wanted to introduce those hooks. I already tested it, and it works just didn't write specs, hence the draft PR.

colby-swandale:
@nijikon reading into your use-case, i get the feeling that adding a [plugin command](https://bundler.io/guides/bundler_plugins.html#developing_your_plugin_commands) for your gem might be the better solution. ie: `bundler check-security` or something alike.

I agree with the points @deivid-rodriguez raised, i don't think adding hooks into CLI commands is something we want to support. But reading your use-case made me think lots about an alternative solution. What if we add a feature to allow plugins to add custom checks to `bundler check`? It would work the same as plugin commands. When the user runs `bundler check`, it will run the default set of bundler checks, but then a plugin could register extra checks.

nijikon:
Adding a plugin command is a bit too much, in my opinion. It would require the user to do an extra step, modify their CI, capistrano deployment etc. I don't want that. I want them to add the plugin and without any additional action.

To be honest the only reason why I put it in `CLI::Check` is because a lack of proper wrapper to handle the check. If you look into how it is implemented in the `CLI::Install`, it executes the `Installer` here, https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/cli/install.rb#L64. Which is a wrapper for plugin hooks in my opinion. https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/installer.rb#L22-L28. We could easily implement it in `CLI::Install` like this without the wrapper
```ruby
Plugin.hook(Plugin::Events::GEM_BEFORE_INSTALL_ALL, definition.dependencies)
Bundler::Installer.new(Bundler.root, definition).run(options)
Plugin.hook(Plugin::Events::GEM_AFTER_INSTALL_ALL, definition.dependencies)
```

I'm happy to jump on a call and brainstorm this.

deivid-rodriguez:
> What we do in our [diffend-ruby](https://github.com/diffend-io/diffend-ruby) plugin, we use before-install-all hook, to be able to protect you so you won't install a malicious gem to your local/staging/production environment.

That links goes no where for me. Maybe you still need to publish the sources or the link is wrong?

> @nijikon reading into your use-case, i get the feeling that adding a plugin command for your gem might be the better solution. ie: bundler check-security or something alike.

I was thinking exactly the same thing as @colby-swandale, that seems to be calling for a custom `bundle` command to me. In fact, it seems quite similar to [`bundle audit`](https://github.com/rubysec/bundler-audit)? But I guess you have a point that it is an extra steps for your users.

Something I didn't quite get from your explanation is in which case these new hooks would be useful and the install hooks won't do the trick. Like do some of your users run `bundle check` but don't try to install gems at all?

hsbt:
The existing plugin hooks are designed around internal processing phases rather than CLI command boundaries, so I don't think adding command-level hooks is the right approach.

I'm planning to add `after-eval` hooks that fire around Gemfile evaluation in `Definition.build` (see https://github.com/ruby/rubygems/pull/6961). Since `bundle check` goes through `Definition.build` internally, the `after-eval` hook will also fire during `bundle check` and should cover this use case.

Closing this as won't merge.

PR #3833, opened 2020-07-15
Revert "Automatic changelog update"

deivid-rodriguez:
# Description:

This reverts commit a69fb0f56116e14da3c8747a0b70951a1b90f9ea.

Latest push to master triggered an incorrect revert of my changelog updates. My hypothesis of why this happened, is because the `paths-ignore` configuration setting that was supposed to prevent it was introduced with that same push.

If this hypothesis is correct, merging this PR should not be automatically reverted :crossed_fingers: 

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Thanks to [the github community forum](https://github.community/t/cant-get-ignore-paths-to-do-what-i-want/122819), I found I was using the `paths-ignore` option incorrectly. So I pushed an alternative. Hopefully it will work this time :crossed_fingers:.

PR #3835, opened 2020-07-16
Rewrite changelog management in ruby

deivid-rodriguez:
# Description:

Unfortunately (as can be seen by recent automatic commits to master), release-drafter is not really working as expected. It seems to have [a](https://github.com/release-drafter/release-drafter/issues/592) [bug](https://github.com/release-drafter/release-drafter/issues/493) that leads to "random behaviour". 

Since we were never interested in release drafts in the first place, only in automating updating our changelog, I think we can drop `release-drafter` in favor of a custom implementation.

This is what this PR does.

For the first push, I'm adding this branch to the changelog worklow and showing what would be committed by the workflow triggered by pushes to the default branch, just to verify that this works as expected.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Working fine!

imagen

Will get this ready to merge now.

PR #3836, opened 2020-07-17
Only update information in the changelog from last merged PR

deivid-rodriguez:
I'm not confortable with the fact that all PRs get rechecked every time, and if there's any changes, an automatic commit is made. I think it's more predictable and friendly if the syncronization simply adds a changelog commit on top of every merged PR that's relevant to end users.

This PR implements that.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
Due to several issues we're seeing coming, we've decided to give it a try to use https://github.com/peter-evans/slash-command-dispatch to instead add the changelog commit to each PR before merging instead of directly to master. I'll give this a try tomorrow. For now I'm closing this.

PR #3840, opened 2020-07-19
Update endpoint of gem signin command

sonalkr132:
# Description:

Updates `gem signin` command to use a new endpoint which requires an API key name and scopes the key should have access to. Each invocation of `gem signin` will create a new API key.

A single API key with access to everything was a security risk.

Majority of changes related to this is on the server-side and can be seen here: https://github.com/rubygems/rubygems.org/pull/1962

A brief summary of the changes:
- adds support for multiple API keys per account and split access into scopes. `show_dashabord` is an exclusive scope.
- stores API key to DB after hashing
- legacy `user.api_key` will be migrated to the new table `api_keys` and will have all scopes enabled (full access).
- `GET api/v1/api_keys` will continue to return `user.api_key` so that `gem signin` doesn't break for old clients. (may be deleted in future)
- new API keys will have `rubygems_` prefix enabling us to sign up for the GitHub program for notifications of key pushed in public repos. 

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -69,16 +69,12 @@ module Gem::GemcutterUtilities
       end
   end
 
-  def scope
-    @scope ||= nil
-  end
-
   ##
   # Creates an RubyGems API to +host+ and +path+ with the given HTTP +method+.
   #
   # If +allowed_push_host+ metadata is present, then it will only allow that host.
 
-  def rubygems_api_request(method, path, host = nil, allowed_push_host = nil, &block)
+  def rubygems_api_request(method, path, host = nil, allowed_push_host = nil, scope: nil, &block)
     require 'net/http'
 
     self.host = host if host
@@ -110,7 +106,7 @@ module Gem::GemcutterUtilities
     end
 
     if api_key_forbidden?(response)
-      update_scope
+      update_scope(scope)
       Gem::RemoteFetcher.fetcher.request(uri, request_method, &block)
     else
       response
@@ -126,25 +122,25 @@ module Gem::GemcutterUtilities
     ask 'Code: '
   end
 
-  def update_scope
+  def update_scope(scope)
     sign_in_host        = self.host
     pretty_host         = pretty_host(sign_in_host)
-    update_scope_params = { self.scope => true }
+    update_scope_params = { scope => true }
 
-    say "The existing key doesn't have access of #{@scope} on #{pretty_host}. Please sign in to update access."
+    say "The existing key doesn't have access of #{scope} on #{pretty_host}. Please sign in to update access."
 
-    email = ask "   Email: "
+    email    = ask "   Email: "
     password = ask_for_password "Password: "
 
     response = rubygems_api_request(:put, "api/v1/api_key",
-                                    sign_in_host) do |request|
+                                    sign_in_host, scope: scope) do |request|
       request.basic_auth email, password
       request.add_field "OTP", options[:otp] if options[:otp]
       request.body = URI.encode_www_form({:api_key => api_key }.merge(update_scope_params))
     end
 
     with_response response do |resp|
-      say "Added #{@scope} scope to the API key"
+      say "Added #{scope} scope to the existing API key"
     end
   end
 
@@ -152,7 +148,7 @@ module Gem::GemcutterUtilities
   # Signs in with the RubyGems API at +sign_in_host+ and sets the rubygems API
   # key.
 
-  def sign_in(sign_in_host = nil)
+  def sign_in(sign_in_host = nil, scope: nil)
     sign_in_host ||= self.host
     return if api_key
 
@@ -164,19 +160,20 @@ module Gem::GemcutterUtilities
 
     email = ask "   Email: "
     password = ask_for_password "Password: "
+    say "\n"
 
-    key_name     = get_key_name
-    scope_params = get_scope_params
+    key_name     = get_key_name(scope)
+    scope_params = get_scope_params(scope)
 
     response = rubygems_api_request(:post, "api/v1/api_key",
-                                    sign_in_host) do |request|
+                                    sign_in_host, scope: scope) do |request|
       request.basic_auth email, password
       request.add_field "OTP", options[:otp] if options[:otp]
       request.body = URI.encode_www_form({ name: key_name }.merge(scope_params))
     end
 
     with_response response do |resp|
-      say "Signed in."
+      say "Signed in with API key: #{key_name}."
       set_api_key host, resp.body
     end
   end
@@ -240,11 +237,11 @@ module Gem::GemcutterUtilities
     end
   end
 
-  def get_scope_params
+  def get_scope_params(scope)
     scope_params = {}
 
-    if self.scope
-      scope_params = { self.scope => true }
+    if scope
+      scope_params = { scope => true }
     else
       say "Please select scopes you want to enable for the API key (y/n)"
       API_SCOPES.each do |scope|
@@ -257,12 +254,14 @@ module Gem::GemcutterUtilities
     scope_params
   end
 
-  def get_key_name
-    hostname = Socket.gethostname
-    default_key_name = "#{hostname}" || "unkown-host-#{Time.now.to_i}"
+  def get_key_name(scope)
+    hostname = Socket.gethostname || "unkown-host"
+    user = ENV["USER"] || ENV["USERNAME"] || "unkown-user"
+    ts = Time.now.strftime("%Y%m%d%H%M%S")
+    default_key_name = "#{hostname}-#{user}-#{ts}"
 
-    key_name = ask "API Key name [#{default_key_name}]: "
-    if key_name.empty?
+    key_name = ask "API Key name [#{default_key_name}]: " unless scope
+    if key_name.nil? || key_name.empty?
       default_key_name
     else
       key_name
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -8,11 +8,12 @@ require 'rubygems/text'
 module Gem::GemcutterUtilities
 
   ERROR_CODE = 1
-  API_SCOPES = %i[index_rubygems push_rubygem yank_rubygem add_owner remove_owner webhook_actions show_dashboard].freeze
+  API_SCOPES = %i[index_rubygems push_rubygem yank_rubygem add_owner remove_owner access_webhooks show_dashboard].freeze
 
   include Gem::Text
 
   attr_writer :host
+  attr_writer :scope
 
   ##
   # Add the --key option
@@ -68,6 +69,10 @@ module Gem::GemcutterUtilities
       end
   end
 
+  def scope
+    @scope ||= nil
+  end
+
   ##
   # Creates an RubyGems API to +host+ and +path+ with the given HTTP +method+.
   #
@@ -96,11 +101,19 @@ module Gem::GemcutterUtilities
 
     request_method = Net::HTTP.const_get method.to_s.capitalize
     response = Gem::RemoteFetcher.fetcher.request(uri, request_method, &block)
-    return response unless mfa_unauthorized?(response)
 
-    Gem::RemoteFetcher.fetcher.request(uri, request_method) do |req|
-      req.add_field "OTP", get_otp
-      block.call(req)
+    if mfa_unauthorized?(response)
+      response = Gem::RemoteFetcher.fetcher.request(uri, request_method) do |req|
+        req.add_field "OTP", get_otp
+        block.call(req)
+      end
+    end
+
+    if api_key_forbidden?(response)
+      update_scope
+      Gem::RemoteFetcher.fetcher.request(uri, request_method, &block)
+    else
+      response
     end
   end
 
@@ -113,6 +126,28 @@ module Gem::GemcutterUtilities
     ask 'Code: '
   end
 
+  def update_scope
+    sign_in_host        = self.host
+    pretty_host         = pretty_host(sign_in_host)
+    update_scope_params = { self.scope => true }
+
+    say "The existing key doesn't have access of #{@scope} on #{pretty_host}. Please sign in to update access."
+
+    email = ask "   Email: "
+    password = ask_for_password "Password: "
+
+    response = rubygems_api_request(:put, "api/v1/api_key",
+                                    sign_in_host) do |request|
+      request.basic_auth email, password
+      request.add_field "OTP", options[:otp] if options[:otp]
+      request.body = URI.encode_www_form({:api_key => api_key }.merge(update_scope_params))
+    end
+
+    with_response response do |resp|
+      say "Added #{@scope} scope to the API key"
+    end
+  end
+
   ##
   # Signs in with the RubyGems API at +sign_in_host+ and sets the rubygems API
   # key.
@@ -121,11 +156,7 @@ module Gem::GemcutterUtilities
     sign_in_host ||= self.host
     return if api_key
 
-    pretty_host = if Gem::DEFAULT_HOST == sign_in_host
-                    'RubyGems.org'
-                  else
-                    sign_in_host
-                  end
+    pretty_host = pretty_host(sign_in_host)
 
     say "Enter your #{pretty_host} credentials."
     say "Don't have an account yet? " +
@@ -134,19 +165,8 @@ module Gem::GemcutterUtilities
     email = ask "   Email: "
     password = ask_for_password "Password: "
 
-    hostname = Socket.gethostname
-    default_key_name = "#{hostname}"
-    key_name = ask "API Key name [#{default_key_name}]: "
-    key_name = default_key_name if key_name.empty?
-
-    scope_params = {}
-
-    say "Please select scopes you want to enable for the API key (y/n)"
-    API_SCOPES.each do |scope|
-      selected = ask "#{scope} [n]: "
-      scope_params[scope] = true if selected == "y"
-    end
-    say "\n"
+    key_name     = get_key_name
+    scope_params = get_scope_params
 
     response = rubygems_api_request(:post, "api/v1/api_key",
                                     sign_in_host) do |request|
@@ -210,4 +230,46 @@ module Gem::GemcutterUtilities
     end
   end
 
+  private
+
+  def pretty_host(host)
+    if Gem::DEFAULT_HOST == host
+      'RubyGems.org'
+    else
+      host
+    end
+  end
+
+  def get_scope_params
+    scope_params = {}
+
+    if self.scope
+      scope_params = { self.scope => true }
+    else
+      say "Please select scopes you want to enable for the API key (y/n)"
+      API_SCOPES.each do |scope|
+        selected = ask "#{scope} [y/N]: "
+        scope_params[scope] = true if selected =~ /^[yY](es)?$/
+      end
+      say "\n"
+    end
+
+    scope_params
+  end
+
+  def get_key_name
+    hostname = Socket.gethostname
+    default_key_name = "#{hostname}" || "unkown-host-#{Time.now.to_i}"
+
+    key_name = ask "API Key name [#{default_key_name}]: "
+    if key_name.empty?
+      default_key_name
+    else
+      key_name
+    end
+  end
+
+  def api_key_forbidden?(response)
+    response.kind_of?(Net::HTTPForbidden) && response.body.start_with?("The API key doesn't have access")
+  end
 end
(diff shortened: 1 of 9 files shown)


sonalkr132:
> gem signin happens as part of a different command, like gem push, or gem owner should rubygems ask all this questions? I think ideally it would try to use a token with the proper scope if present, and create one otherwise without asking all these extra questions?

We can update the default answers but I think it should always ask all questions. IMHO, this would lead to least surprise for most users. Even if we were to make it clear to users about the scope we assumed (though text output of the command), at a later time the user may be confused about why a different command is not working when they remember doing `signin`. 
I believe a flag that would allow the client to assume scope (or a flag for each scope) would be a better user experience than assuming scope and not asking for user interaction by default.

> the request will eventually fail with an authorization error?

As of now, the user will just `Access Denied. Please sign up for an account at https://rubygems.org`. Are you trying to suggest we should explicitly say access was denied because of incorrect scope? if yes, this change will be done on the server-side.

>  each command should try to use a properly scoped token if present?

I would think if a host needed access to multiple scopes, the user would allow multiple scopes on the same key. Can you please explain the advantage/usecase of having multiple keys with separate scopes and configuring them on the same host?

deivid-rodriguez:
Consider this:

* I run `gem push` and I currently don't have any local credentials.
* I get ask my login credentials by the client, plus all the questions about the scopes I want.
* Since I only want to push a gem, and I want to be conservative, I only say yes to the push scope.
* My token gets saved, and `gem push` (my usual and only authenticated server operation) works flawlessly for me from now on.
* At some point in the future, I happen to need to run `gem owner`. Rubygems detects that I have a local token and tries to use it. I get an error because the token doesn't have the proper scopes.

The prior scenario is not nice IMO, because the user went with the most secure option (just ask for permissions for what she needs), and got an error as a result. A better outcome would've been that the client detects that the user does have a token, but not with the proper credentials, and saves either a new token with the proper "aggregated scopes", or a separate token that future `gem owner` invokations will use. Hence my suggestion of possibly keeping multple scopes locally, but now that I think about it a better alternative is maybe to overwrite the old token with a new one all the "accumulated scopes"? If we did that, the user wouldn't get any errors, and we wouldn't bother her with "token questions".

What I'm saying is that we can infer the scopes a user wants from the command she has run, so I don't think we need to ask all those questions.

deivid-rodriguez:
To further clarify, the method we are changing here gets called when running `gem signin` direclty, but also when running `gem push`, `gem owner` or other operations, as long as there's no local credentials. So in my opinion it makes sense to ask all the questions for `gem signin`, but I don't think it makes sense to ask them on `gem push` or `gem owner`, because we already know what the user wants to do.

sonalkr132:
> my suggestion of possibly keeping multple scopes locally, but now that I think about it a better alternative is maybe to overwrite the old token with a new one with all the "accumulated scopes"

Overwriting old token without user interaction may lead to an increase in the number of API tokens a rubygems.org account has (with a good number of them being unused). Does asking the user to delete the key we are overwriting seem reasonable to you? Users may want not to delete the key if they have used it on other hosts.

> we can infer the scopes a user wants from the command she has run

I would recommend making users confirm the scope we are inferring as they may not know about the addition of scopes to API keys and it may seem confusing to ask for sign in again. I am guessing you don't think this is an issue. I will update this to not ask for the scope questions on `push`, `yank` and `owner` command.

deivid-rodriguez:
> Overwriting old token without user interaction may lead to an increase in the number of API tokens a rubygems.org account has (with a good number of them being unused). Does asking the user to delete the key we are overwriting seem reasonable to you? Users may not want to delete the key if they have used it on other hosts.

That's a very good point, I missed that. What about keeping separate keys in `~/.gem/credentials.yml`, each with the proper scope? Like

```yaml
---
:rubygems_api_key: xxxxxx # the legacy one with all scopes
:rubygems_push_api_key: yyyyyy # with only the push scope
:rubygems_yank_api_key: zzzzzz # with only the yank scope
...
```

If user runs `gem singin` explicitly, then we save the key with the multiple scope choice as `:rubygems_api_key`, and use that as a fallback if there's no command specific key?

> I would recommend making users confirm the scope we are inferring as they may not know about the addition of scopes to API keys and it may seem confusing to ask for sign in again. I am guessing you don't think this is an issue. I will update this to not ask for the scope questions on `push`, `yank` and `owner` command.

mmmmm, yeah, that could surprise some folks that were used to how this used to work and are getting set up from scratch on a new host. Although only on the first invokation of every command, but still. Maybe rather than asking for confirmation (since I can't think of an scenario where someone would say "no"), we could show a CLI message when saving the scoped API key, and what to expect on future invokations?

deivid-rodriguez:
Actually, since we need sign in the user again in these scenarios, and prompt for email and password again, we can take that chance to communicate why we're asking for credentials again. What do you think?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -132,12 +133,26 @@ def sign_in(sign_in_host = nil)
 
     email = ask "   Email: "
     password = ask_for_password "Password: "
+
+    hostname = Socket.gethostname
+    default_key_name = "#{hostname}"
+    key_name = ask "Key name [#{default_key_name}]: "
```
colby-swandale, line 139: I think the suggested key name should instead be a safe version of the host that the user is signing into. Ie: `rubygems_org`

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -132,12 +133,26 @@ def sign_in(sign_in_host = nil)
...
+    key_name = default_key_name if key_name.empty?
+
+    scope_params = {}
+
+    say "Please select scopes you want to enable for the key (y/n)"
+    API_SCOPES.each do |scope|
```
colby-swandale, line 145: As i've mentioned in Slack - i think asking the user to answer 7 different questions is not preferable for UX. Could we have a default set of scopes that's assumed, which the user can add/remove with provided options/arguments to `gem signin`

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -132,12 +133,26 @@ def sign_in(sign_in_host = nil)
...
+      selected = ask "#{scope} [n]: "
+      scope_params[scope] = true if selected == "y"
```
colby-swandale, line 147: You will need to cover other different variants, ie: `y|Y|yes` etc. IIRC rubygems handles this already in some places.

lib/rubygems/gemcutter_utilities.rb
sonalkr132: In my mind, the purpose of the name of the key is to identify where the key is being used in the client. I am sorry but I don't understand how naming the key `rubysgems_org` would help in this or perhaps serve a greater purpose? please elaborate.

lib/rubygems/gemcutter_utilities.rb
sonalkr132: > asking the user to answer 7 different questions is not preferable for UX. 

I agree that it is not as convenient as the current UX. IMHO, making the user explicitly pick scopes they want to enable when they run `gem signin` is the most secure option. I would recommend against having a default scope as it would lead to privilege escalation in the cases where user may not be using the default we pick for them. I am sure you are familiar with `the principle of least privilege`. Enabling a presumed limited default defeats the very purpose of separating scopes if we are going to group a certain number of them in the client anyway.

sonalkr132: I would also like to add that security and convenience are often at odds with each other. If you can think of way we could have both, it would be gold.

lib/rubygems/gemcutter_utilities.rb
sonalkr132: I have updated this to a regex.

sonalkr132:
I have changed this to select scopes as per command and update when necessary. Following is what it looks like as of now:
```
# No creds found. first invocation
$ ruby -Ilib bin/gem push ~/hola/botanicus-merb-etag-4.1.50.gem
Don't have an account yet? Create one at http://rubygems.org/sign_up
   Email:   mail@mail.com
Password:
API Key name [hostname]:
Signed in.
Pushing gem to http://rubygems.org...
Successfully registered gem: botanicus-merb-etag (4.1.50)

# key exists but doesn't have access to add owner
$ ruby -Ilib bin/gem owner botanicus-merb-etag --add aditya.prakash132@gmail.com
The existing key doesn't have access of add_owner on RubyGems.org. Please sign in to update access.
   Email:   apps-team@media.net
Password:
Added add_owner scope to the API key
sonalkr1321w was added as an unconfirmed owner. Ownership access will be enabled after the user clicks on the confirmation mail sent to their email.
Owners for gem: botanicus-merb-etag
```

Please let me know if this seems acceptable and I can work on adding tests.

deivid-rodriguez:
@sonalkr132 That seems like a nice UX to me. If I understand correctly `gem signin` still asks all the scope questions, right? Since most (all?) commands that need api keys already signin "on the fly", I think it's perfectly fine to reserve `gem signin` for a finer grained management of scopes, and thus ask all the questions there :+1:.

sonalkr132:
> gem signin still asks all the scope questions, right?

Yes, `gem signin` will ask for all scopes with the default answer being `N` (disabled) on all option.

> most (all?) commands that need api keys already signin "on the fly"

AFAIK, gem `owner`, `push` and `yank` are the only commands which use the API key. The correct scope will be picked for the creation/update of API keys when any of them are invoked. user will need to confirm api-key name we picked. We would appreciate your opinion not this as well.. what would a good default API key name and whether we should allow duplicate names. The current implementation uses hostname as if for default and we don't enforce unique name checks on the server side.

deivid-rodriguez:
Ok, I'll look into this more closely and leave more feedback here as soon as I can :+1:.

sonalkr132:
Server-side PR incase you want to check that https://github.com/rubygems/rubygems.org/pull/1962

In addition to separating scopes, other major changes include storage of keys as encrypted hash, option to check when was the last access time of the key, and update of legacy `GET /api/v1/api_keys` to create new API key (with all but `show_dashboard` scope enabled) on every request.

sonalkr132:
Hi, any updates on this? Please let me know if there is anything I can do to move this forward.

deivid-rodriguez:
Hei! Sorry @sonalkr132, I haven't managed to get to this yet. Will try to give it a shot this week or the next one.

deivid-rodriguez:
@sonalkr132 I started reviewing this PR, it's looking great :). I hope to finish by Monday.

What's the expected order for the server side and client side PRs. I'm guessing it doesn't really matter since both PR are backwards compatible, right?

sonalkr132:
server-side PR has to be merged/released first. it supports all rubygems versions (besides the implementation in this PR). I haven't merged it already as I am not sure about what kind of changes I may have to make to accommodate requests from client-side PR.

It would also be nicer to have an RC in the client so that we can make an announcement to the users. "check out our new API keys implementation. install rubygems x.y.z.rc".

deivid-rodriguez:
Oh, I see. This PR uses a new endpoint for API keys that's provided by the server side PR, right?

sonalkr132:
> This PR uses a new endpoint for API keys that's provided by the server side PR, right?

yes, correct. old API endpoint will be supported for almost a year.

deivid-rodriguez (commented):
@sonalkr132 This looks good to me!

I only have one comment on the implementation.

I would instead change the `sign_in` helper to explicitly receive the scope. I believe that would remove the need keep the scope on an instance variable per command, and would make things more explicit and easier to read. What do you think?

Regarding

> user will need to confirm api-key name we picked. We would appreciate your opinion on this as well.. what would a good default API key name and whether we should allow duplicate names. The current implementation uses hostname as-is for default and we don't enforce unique name checks on the server side.

I'm not sure. It sounds like enforcing uniqueness doesn't buy us much, but then should we bother the user at all with a "choose a key name" question? Maybe we should "resolve the name" automatically to a clear and unique enough name that the user can understand we she sees it in the rubygems.org side, and log it to screen for transparency, but not ask anything?

deivid-rodriguez:
I was really unsure about whether to tag this as "minor" or "major". Since I'm most of the times unsure about this and I believe this it is something super subjective and not too useful, I plan to merge "major enhancement" and "minor enhancement" into "enhancement".

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -112,6 +126,28 @@ def get_otp
     ask 'Code: '
   end
 
+  def update_scope
+    sign_in_host        = self.host
+    pretty_host         = pretty_host(sign_in_host)
+    update_scope_params = { self.scope => true }
```
sonalkr132, line 132: >  I believe that would remove the need keep the scope on an instance variable per command, and would make things more explicit and easier to read

`scope` is used not only in `sign_in` but also to update scope in `rubygems_api_request`. scope would have to be passed twice in each action, which didn't seem as pleasant to me.
it would definitely make thinks more explicit. let me know if you think we should change it to explicit arg.

sonalkr132:
> the user at all with a "choose a key name" question? Maybe we should "resolve the name" automatically to a clear and unique enough name that the user can understand we she sees it in the rubygems.org side

just to clarify, the user will see the following and they can press enter to accept the default keyname:
> API Key name [example-hostname] :

we can use `#{hostname}-#{whoami}-#{date}` for commands which assume the scope. I am not sure if this will be enough for users to identify keys, especially when they can't see the plain text value of the key in rubygems.org.
I have noticed that some sites show a few characters in plain text (like credit card numbers), ex: xxxx9c3393. which will mean we will need to store the last six characters in a separate column as plain text just for display.

deivid-rodriguez (commented):
Yes, I had noticed that the question is optional and can be bypassed with ENTER.

So, to make sure I understand, with this PR, every time I log in to rubygems on a new machine, or after running `gem signout` on the same machine, a new API key with no scopes will get automatically created. Then scopes get added automatically as we "do things" using this key.

If this is the case, I think it maybe makes sense to keep the question and actually enforce key uniqueness.That way, we allow the user to enter the key name that they want to use and to reuse it in case it already exists?

My main concern is that users understand what we are asking, and why.

sonalkr132:
> every time I log in to rubygems on a new machine, or after running gem signout on the same machine, a new API key with no scopes will get automatically created

no, this is not the case. when you run `gem sigin` all questions related to scope will be asked and the new key will have scope enabled where the user entered `Y`. 
This was previously discussed [here](https://github.com/rubygems/rubygems/pull/3840#issuecomment-697441166):

>> gem signin still asks all the scope questions, right?

> Yes, gem signin will ask for all scopes with the default answer being N (disabled) on all option.

>> most (all?) commands that need api keys already signin "on the fly"

> AFAIK, gem owner, push and yank are the only commands which use the API key. The correct scope will be picked for the creation/update of API keys when any of them are invoked.


scopes get automatically added if your existing key doesn't have permission to a specific action you are trying to perform. the existing key will always have some scopes enabled. We won't ask for a key name as we are updating an existing key.

> My main concern is that users understand what we are asking, and why.

perhaps instead of just using `API Key name [hostname]:` we can be a bit more descriptive. `Creating a new API key as one doesn't exist on your host. please enter a name you want to use to identify this key in the rubygems.org UI`. 
IMHO, this is unnecessarily verbose and users would have an intuitive sense to figure out why we are asking for a key name.
Let me know if you have ideas about making things clearer.

deivid-rodriguez:
> no, this is not the case. when you run gem sigin all questions related to scope will be asked and the new key will have scope enabled where the user entered Y.

Sorry, yes, I know this but I worded my question badly. Let me try word it better now:

The first time any CLI command uses the `sign_in` helper method (i.e. requests rubygems.org credentials) on a fresh machine, a new API key with necessary scopes for the CLI command being run will be created on the rubygems.org side and saved locally to a `~/.local/share/gem/credentials` file. The next time I run a command that needs rubygems.org authentication, that API key will be picked from the local file and scopes on it will be updated for the running CLI command.

If, however, I run the gem CLI on a different machine, there will be presumably no `~/.local/share/gem/credentials` file, so a _new_ API will be created. Is that right? And I assume the same happens if I run `gem signout` on the first machine since that command deletes the local `~/.local/share/gem/credentials`.

Assuming the above is correct, I think the users might want to reuse the same API key in the new machine, so if keys are unique, he'd only need to specify the unique name she gave to the key the first time. I think that's something I would do, because I presume I'd end up making the same operations on a new server, so leaking the key would have the same bad consequences so I would "refuse" the little extra security of temporarily having an API with less scopes in favour of keeping a single API key.

So, asking for an API key name makes sense to me in this context so that users effectively choose whether to reuse their API key or create a new one.

> perhaps instead of just using API Key name [hostname]: we can be a bit more descriptive. Creating a new API key as one doesn't exist on your host. please enter a name you want to use to identify this key in the rubygems.org UI.
IMHO, this is unnecessarily verbose and users would have an intuitive sense to figure out why we are asking for a key name.
Let me know if you have ideas about making things clearer.

I'm definitely in favour of being more descriptive, yeah.

deivid-rodriguez:
If the above is something we don't want to do (allowing the CLI to reuse a key by name) and we prefer to create new keys everytime we sign in on a new host, I think I'd rather go with your idea of using `#{hostname}-#{whoami}-#{date}` for commands which assume the scope, and don't ask any questions. Maybe add "last time used" timestamp information in the rubygems.org side so users can easily do "old key cleanup".

sonalkr132:
> no ~/.local/share/gem/credentials file, so a new API will be created. Is that right?

yes this is correct.

>  I think the users might want to reuse the same API key in the new machine, so if keys are unique, she'd only need to specify the unique name she gave to the key the first time.

correct if I am wrong, but you seem to be assuming that if the user entered the same API key name, we will be able to plain text key of the existing key from the serverside. This won't be possible because we are going to hash the plain text keys with one-way encryption and we will be to show the plain text key to the user only once, the first time they created it. It is a security risk to store API keys in plain text and with "new API keys" we are trying to migrate away from it. 

>  I think I'd rather go with your idea of using #{hostname}-#{whoami}-#{date} for commands which assume the scope

please confirm that this is what you want to do given the new information.

> Maybe add "last time used" timestamp information in the rubygems.org side

this is already implemented. users will be able to see when was an API key successfully used for authentication in rubygems.org UI.

deivid-rodriguez:
> correct me if I am wrong, but you seem to be assuming that if the user entered the same API key name, we will be able to show the plain text version of the existing key from the serverside. This won't be possible because we are going to hash the keys with one-way encryption (just like passwords) and we will be to show the plain text key to the user only once, the first time they created it. It is a security risk to store API keys in plain text and with "new API keys" we are trying to migrate away from it.

Oh, you're totally right! I was actually not assuming that _intentionally_ but I realized now that my proposal of being able to "reuse keys" implies being able to do that, since the plain text key would indeed need to be returned from the server.

> please confirm that this is what you want to do given the new information.

Yes, given that "last time used" information is given on the server side, I believe creating a key named `#{hostname}-#{whoami}-#{date}` is clear enough so that we don't need to ask any questions. Instead, we could log to screen something like "an API key named #{hostname}-#{whoami}-#{date} was created on the server side and was saved to `~/.local/share/gems/credentials` for future passwordless logins`. But I'm not sure if that's even necessary? What do you think?

Other question I have is whether you think `gem signout` should delete the key on the server side. It doesn't seem necessary but maybe nice to have?

sonalkr132:
> Instead, we could log to screen something like "an API key named #{hostname}-#{whoami}-#{date} was created on the server side and was saved to ~/.local/share/gems/credentials for future passwordless logins`. 

As of now we just say `signed in`, which presumably implies that a key was saved in `~/.local/share/gems/credentials `. I suppose we can update the success text to include key name. something like `Signed in with API key #{hostname}-#{whoami}-#{date}`.
> Enter your RubyGems.org credentials.
Don't have an account yet? Create one at http://rubygems.org/sign_up
...
Signed in.

> gem signout should delete the key on the server side.

we can probably provide a flag to do add this behavior. I don't think we can use this for default, given how `sign out` works currently and the fact that the key may be reused in multiple env.

deivid-rodriguez:
> As of now we just say signed in, which presumably implies that a key was saved in ~/.local/share/gems/credentials . I suppose we can update the success text to include key name. something like Signed in with API key #{hostname}-#{whoami}-#{date}.

That seems like a nice combination of conciseness and transparency :+1:.

> we can probably provide a flag to do add this behavior. I don't think we can use this for default, given how sign out works currently and the fact that the key may be reused in multiple env.

You mean if a user copies its `~/.local/share/gems/credentials` file to a different machine for example, right? Well, you're right again. After a second thought, I don't think it's worth doing this at all, not even under a flag. Users can always delete keys on the server side, right?

sonalkr132:
Updated success text and removed the prompt for API key name.

> You mean if a user copies its ~/.local/share/gems/credentials file to a different machine for example, right? Well, you're right again. After a second thought, I don't think it's worth doing this at all, not even under a flag. Users can always delete keys on the server side, right?

API keys can be created from the web UI as well, user can use the same key on as many hosts as they chose to do so. Yes, the user can delete keys from the web UI.

deivid-rodriguez (approved):
Looks good to me! :+1:

deivid-rodriguez:
Thanks for your work @sonalkr132!

PR #3850, opened 2020-07-24
Try to fix openssl availability issues on jruby

deivid-rodriguez:
# Description:

This is a followup to https://github.com/rubygems/rubygems/pull/3809, and all the other PRs linked to that trying to deal with jruby load service issues.

To recap the issue, as first mentioned [here](https://github.com/rubygems/rubygems/pull/3751), apparently on jruby sometimes autoloading a constant and making it available through `require` can lead to weird concurrency issues that result in the constant not being available in the end.

So the solution we've been proposing for these issue is to not do that, and either require or `autoload`, but not both.

Everything seems good now, except for `OpenSSL` constants, where we sometimes still get errors. In https://github.com/rubygems/rubygems/pull/3809 I unified our code to always require instead of autoloading `openssl`. However, the problem still reproduces on our latest code. See https://github.com/rubygems/rubygems/pull/3867/checks?check_run_id=1059447242, for example:

```
 NameError: uninitialized constant Gem::Security::OpenSSL
         org/jruby/RubyModule.java:3760:in `const_missing'
         /home/runner/work/rubygems/rubygems/lib/rubygems/security.rb:452:in `create_digest'
         /home/runner/work/rubygems/rubygems/lib/rubygems/package.rb:361:in `block in digest'
         org/jruby/RubyArray.java:1809:in `each'
         (...)
```

After closer inspection, I noticed that the constant was still getting autoloaded `here`: https://github.com/ruby/ruby/blob/93b78abd774109d1333d59eaf439b2e69ed0fe00/lib/net/http.rb#L25. And we can't really avoid that unless we stop requiring `net/http` and vendor it instead.

So my approach was to instead of requiring `openssl` always, autoload it always. Other than potentially fixing the bug, this has two extra benefits:

* We can unde one monkey-patch to the `net-http-persistent` gem.
* It should be more efficient.

So I'd like to try this approach.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
I filled the description of this PR and marked it as ready!

deivid-rodriguez:
I didn't see the jruby crashes again since the last time, but since this has other benefits I decided to go ahead with this. If it causes any issues, I'll revert it.

PR #3856, opened 2020-07-26
Vendors DidYouMean::Levenshtein.distance from did_you_mean @ 1.4.0

austinpray:
- Similar to https://github.com/rubygems/rubygems/pull/3857

* * *

# Description:

_See edit history and thread for full story_

- Significant performance boost

## Benchmarks

:chart: [Benchmark code](https://github.com/austinpray/rubygems_levenshtein_distance/blob/master/text_gem_bench.rb) [(permalink)](https://github.com/austinpray/rubygems_levenshtein_distance/blob/f76e1cdcab98a40b2aef3e49206761f2dee896f7/text_gem_bench.rb)

Looks like this could be a **2x performance boost**

```
# Single Insertion

              Before     3.844k i/100ms
               After     8.307k i/100ms
              Before     38.015k (± 1.4%) i/s -    192.200k in   5.056880s
               After     83.084k (± 3.8%) i/s -    415.350k in   5.007893s
# Complex

              Before     4.481k i/100ms
               After     9.399k i/100ms
              Before     44.281k (± 2.9%) i/s -    224.050k in   5.064414s
               After     97.084k (± 4.2%) i/s -    488.748k in   5.044751s
```

While looking for reference implementations of the edit distance algorithm I saw that the `DidYouMean::Levenshtein.distance` function is highly optimized. I figured it's a good idea to consolidate implementations of this method!

`GEM::Text.levenshtein_distance` is now a vendored version of `DidYouMean::Levenshtein.distance`

## Old history before the squash

For posterity the old history is here:
- https://github.com/austinpray/rubygems/tree/levenshtein-dym-b4squash
- https://github.com/austinpray/rubygems/tree/9b8bca135491f0dcc5c53f0326874239818b899d

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/text.rb
+++ b/lib/rubygems/text.rb
@@ -49,37 +49,38 @@ module Gem::Text
     end
   end
 
-  # This code is based directly on the Text gem implementation
   # Returns a value representing the "cost" of transforming str1 into str2
+  # Vendored version of DidYouMean::Levenshtein.distance from the ruby/did_you_mean gem @ 1.4.0
+  # https://git.io/JJgZI
   def levenshtein_distance(str1, str2)
-    s = str1
-    t = str2
-    n = s.length
-    m = t.length
-
-    return m if (0 == n)
-    return n if (0 == m)
+    n = str1.length
+    m = str2.length
+    return m if n.zero?
+    return n if m.zero?
 
     d = (0..m).to_a
     x = nil
 
-    str1.each_char.each_with_index do |char1,i|
-      e = i + 1
+    # to avoid duplicating an enumerable object, create it outside of the loop
+    str2_codepoints = str2.codepoints
 
-      str2.each_char.each_with_index do |char2,j|
-        cost = (char1 == char2) ? 0 : 1
+    str1.each_codepoint.with_index(1) do |char1, i|
+      j = 0
+      while j < m
+        cost = (char1 == str2_codepoints[j]) ? 0 : 1
         x = min3(
-             d[j + 1] + 1, # insertion
-             e + 1,      # deletion
-             d[j] + cost # substitution
-           )
-        d[j] = e
-        e = x
-      end
+          d[j + 1] + 1, # insertion
+          i + 1,      # deletion
+          d[j] + cost # substitution
+        )
+        d[j] = i
+        i = x
 
+        j += 1
+      end
       d[m] = x
     end
 
-    return x
+    x
   end
 end
--- a/test/rubygems/test_gem_text.rb
+++ b/test/rubygems/test_gem_text.rb
@@ -83,6 +83,12 @@ Without the wrapping, the text might not look good in the RSS feed.
     assert_equal 7, levenshtein_distance("zentest", "xxxxxxx")
   end
 
+  def test_levenshtein_distance_all
+    assert_equal 6, levenshtein_distance("algorithm", "altruistic")
+    assert_equal 3, levenshtein_distance("saturday", "sunday")
+    assert_equal 3, levenshtein_distance("kitten", "sitting")
+  end
+
   def test_truncate_text
     assert_equal "abc", truncate_text("abc", "desc")
     assert_equal "Truncating desc to 2 characters:\nab", truncate_text("abc", "desc", 2)


austinpray:
I don't think the tests failing are my fault :thinking:

hsbt:
@austinpray We can uninstall the bundled gems after the Ruby installation. After that, this patch will not work. 

Can you add the fallback like this?

```ruby
begin
  require 'did_you_mean/levenshtein'
  (snip)
rescue LoadError
  # this part is the original code
end
```

austinpray:
@hsbt I added a fallback. I couldn't use `begin ... rescue LoadError` inside the function body because it slowed everything down. Here's the benchmark of trying both methods: https://github.com/austinpray/rubygems_levenshtein_distance/pull/1

Changes:
1. Update `GEM::Text.levenshtein_distance` to use the latest optimized code from `DidYouMean::Levenshtein.distance`
2. If `DidYouMean::Levenshtein.distance` is available (Ruby 2.7+), use that directly.

lib/rubygems/text.rb
```diff
@@ -49,37 +49,54 @@ def min3(a, b, c) # :nodoc:
...
+    def levenshtein_distance(str1, str2)
+      DidYouMean::Levenshtein.distance(str1, str2)
+    end
+  end
+
+  Gem::Text.prepend(Gem::Text::LevenshteinExt)
+
+rescue LoadError
+  # ignored
+end
```
eregon, line 102: Can you wrap this in `defined?(DidYouMean)`?
If the user did `--disable-did_you_mean` then RubyGems should not load it.

Also maybe only `defined?(DidYouMean::Levenshtein)` is enough (no `require`, no `rescue`) since did_you_mean.rb loads did_you_mean/levenshtein.rb eagerly.

lib/rubygems/text.rb
```diff
@@ -86,7 +86,7 @@ def levenshtein_distance(str1, str2)
 end
 
 # Use the latest and greatest DidYouMean::Levenshtein if it is available
-begin
+if RUBY_VERSION >= '2.7' && defined?(DidYouMean)
   require 'did_you_mean/levenshtein'
```
austinpray, line 90: My instincts are telling me that we should have this explicit require in here. 

1. `did_you_mean/levenshtein` is an implicit dependency of some of the did_you_mean spellcheckers. I don't like the idea of ruby retiring the use of levenshtein in the spell checker required by default and then we lose this implicit dependency. I like the idea of this failing quickly if in the future the environment is not what we expect. Avoid silent fallbacks and whatnot.

lib/rubygems/text.rb
```diff
@@ -86,7 +86,7 @@ def levenshtein_distance(str1, str2)
 end
 
 # Use the latest and greatest DidYouMean::Levenshtein if it is available
-begin
+if RUBY_VERSION >= '2.7' && defined?(DidYouMean)
```
austinpray, line 89: In my testing in `irb` as well as `ruby -e` on ruby 2.3 to 2.7 I get

`defined?(DidYouMean)` | `defined?(DidYouMean::Levenshtein)`
--- | ---
"constant" | "constant"

However, in the context of the rubygems runner as tested with `rake install` or `ruby -Ilib bin/gem ....` I get 

`defined?(DidYouMean)` | `defined?(DidYouMean::Levenshtein)`
--- | ---
"constant" | nil

I didn't get to the root of it, but I can definitely see that rubygems tries to isolate it's environment for good reason. On ruby 2.6 downwards in the context of the rubygems runner: the bundled did_you_mean gem is not in the load path so we cannot require `did_you_mean/levenshtein`. On ruby 2.7 upwards `did_you_mean/levenshtein` is always available for requiring.

**I like this better anyway:** it simplifies the logic here:

_note that the vendored version is based on did_you_mean 1.4.0_


. | default | --disable-did_you_mean
--- | --- | ---
**ruby >=2.7** | did_youmean >= 1.4.0 | use vendored
**ruby <2.7** | use vendored | use vendored

**:heavy_check_mark: Simple**

Compare this with if we were to somehow successfully use the bundled version on ruby <2.7:

. | default | --disable-did_you_mean
--- | --- | ---
**ruby >=2.7** | did_youmean >= 1.4.0 | use vendored
**ruby 2.6** | did_you_mean 1.3.0 | use vendored
**ruby 2.5** | did_you_mean 1.2.0 | use vendored
**ruby 2.4** | did_you_mean 1.1.0 | use vendored
**ruby 2.3** | did_you_mean 1.0.0 | use vendored

**:unamused: Complicated, inconsistent performance optimizations**

lib/rubygems/text.rb
```diff
@@ -49,37 +49,51 @@ def min3(a, b, c) # :nodoc:
     end
   end
 
-  # This code is based directly on the Text gem implementation
   # Returns a value representing the "cost" of transforming str1 into str2
+  # Vendored version of DidYouMean::Levenshtein.distance from the did_you_mean gem
```
austinpray, line 53:
```suggestion
  # Vendored version of DidYouMean::Levenshtein.distance from the ruby/did_you_mean gem @ 1.4.0
```

lib/rubygems/text.rb
austinpray: Here's some code to test that `defined?(DidYouMean::Levenshtein)` matrix. Requires docker

https://gist.github.com/austinpray/31972f4364d3269d511e1b478147492c

eregon: I see, that makes sense.

I wonder though, is it so valuable to use `did_you_mean`'s version and not just always use the vendored version?
I think that would simplify things a lot more and make it fully reliable (e.g., `did_you_mean` renames the file in a future release and RubyGems will break (!), bad and unexpected).
I think it's best RubyGems avoids any unneeded dependency.

austinpray: @eregon Yeah, that sounds fine to me. I would rather not cause headaches down the line. I went ahead and dropped the DidYouMean dependency. I can take any further ideas of using DidYouMean to improve this feature into a different PR.

The benchmarks still hold that the updated function is a 2x speedup raw iterations-wise. In unscientific testing it shaves about a second off of the suggestion time on my machine:
Screen Capture_select-area_20200801102036

duckinator:
I think always vendoring it is the right call. This looks good to me.

@austinpray are you finished working on this? If so, I'll go ahead and merge it. :+1:

deivid-rodriguez:
I also like the approach that was finally taken here :+1:.

austinpray:
@duckinator Yep everything looks good to me! I was expecting a squash and merge on this one since the history is a wild ride.👍

Thank you everyone for working with me on this one! I learned a lot in the process.

deivid-rodriguez:
@austinpray We don't have "squash and merge" option enabled (one reason is some rake tasks we have rely on parsing PR merge commit messages). But I do like to have a clean git history, so if you can rebase and squash yourself, that'd be nice :).

austinpray:
@deivid-rodriguez squashed the history :+1: 

For posterity the old history is here:
- https://github.com/austinpray/rubygems/tree/levenshtein-dym-b4squash
- https://github.com/austinpray/rubygems/tree/9b8bca135491f0dcc5c53f0326874239818b899d

duckinator:
Thanks for the PR, @austinpray! :tada:

PR #3857, opened 2020-07-26
Use DidYouMean::SpellChecker for gem suggestions in Bundler

austinpray:
Similar to https://github.com/rubygems/rubygems/pull/3856

* * *

replaces `Bundler::SimilarityDetector` with [`DidYouMean::SpellChecker`](https://github.com/ruby/did_you_mean/blob/2ddf39b874808685965dbc47d344cf6c7651807c/lib/did_you_mean/spell_checker.rb)

1. Higher quality spelling suggestions in my testing
1. Benefit from upstream improvements, less code to maintain in this repo. The current `Bundler::SimilarityDetector.levenshtein_distance` dynamic programming implementation could be optimized quite a bit. 
1. Should see a significant performance improvement, under the hood `DidYouMean::SpellChecker` [is highly optimized](https://github.com/ruby/did_you_mean/commit/bf64b25501c07ee8475997e94ca51db67a7f31ed#diff-6ed35dc5fb06924d97b2a98adb0bc285).

RubyGems 3.0 supports Ruby 2.3 or later. Ruby 2.3 and later ships with the `did_you_mean` gem.

# Description:

## Notes for reviewer

- `Bundler::SimilarityDetector` is not used in Bundler anymore. I would like to delete it, but I haven't haven't because I am not sure if we should pull the rug out from people that might be using for their own purposes. **What should we do to phase out this module?**

Was improving performance over in https://github.com/rubygems/rubygems/pull/3856 and saw this as an opportunity as well.

Use `DidYouMean::SpellChecker` as a drop-in replacement for `Bundler::SimilarityDetector`

______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/lib/bundler/cli/common.rb
+++ b/bundler/lib/bundler/cli/common.rb
@@ -60,11 +60,11 @@ module Bundler
     end
 
     def self.gem_not_found_message(missing_gem_name, alternatives)
-      require_relative "../similarity_detector"
+      require "did_you_mean/spell_checker"
       message = "Could not find gem '#{missing_gem_name}'."
-      alternate_names = alternatives.map {|a| a.respond_to?(:name) ? a.name : a }
-      suggestions = SimilarityDetector.new(alternate_names).similar_word_list(missing_gem_name)
-      message += "\nDid you mean #{suggestions}?" if suggestions
+      alternate_names = alternatives.map { |a| a.respond_to?(:name) ? a.name : a }
+      suggestions = DidYouMean::SpellChecker.new(dictionary: alternate_names).correct(missing_gem_name)
+      message += "\nDid you mean #{word_list(suggestions)}?" unless suggestions.empty?
       message
     end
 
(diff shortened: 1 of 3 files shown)


bundler/lib/bundler/cli/common.rb
```diff
@@ -97,5 +97,21 @@ def self.clean_after_install?
...
+        return ''
+      end
+
+      words = words.map { |word| "'#{word}'" }
+
+      if words.length == 1
+        return words[0]
+      end
+
+      [words[0..-2].join(", "), words[-1]].join(" or ")
```
austinpray, line 114:
```suggestion
      [words[0..-2].join(", "), words[-1]].join(", or ")
```

Personally I like oxford commas ("apple, orange, or pear"). I went without it so the output is similar to how it has always been ("apple, orange or pear").

bundler/lib/bundler/similarity_detector.rb
```diff
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+# This module is not used anywhere in Bundler
+# It is included for backwards compatibility in-case someone is relying on it
```
austinpray, line 4: I left this module in here so I don't pull the rug out from someone who might be using it in their own project. I would like to delete it, is there a deprecation path for stuff like this?

hsbt:
@austinpray Can you update this Pr by `rubocop -a bundler/lib/bundler/cli/common.rb`?

hsbt:
@austinpray Sorry to my late action. I'm happy to introduce this feature at Bundler 4 and Ruby 4.0. Thank you!

PR #3865, opened 2020-07-29
Accept responses with no etag header

indirect:
If the server does not provide an etag header, treat all responses as a
cache miss. This should allow 3rd party gem servers to more easily
provide compact_index results. /cc @mperham

See also https://github.com/mperham/sidekiq/issues/4158.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -66,8 +66,8 @@ module Bundler
             end
           end
 
-          response_etag = (response["ETag"] || "").gsub(%r{\AW/}, "")
-          if etag_for(local_temp_path) == response_etag
+          etag = (response["ETag"] || "").gsub(%r{\AW/}, "")
+          if etag.length.zero? || etag_for(local_temp_path) == etag
             SharedHelpers.filesystem_access(local_path) do
               FileUtils.mv(local_temp_path, local_path)
             end
@@ -75,7 +75,7 @@ module Bundler
           end
 
           if retrying
-            raise MisMatchedChecksumError.new(remote_path, response_etag, etag_for(local_temp_path))
+            raise MisMatchedChecksumError.new(remote_path, etag, etag_for(local_temp_path))
           end
 
           update(local_path, remote_path, :retrying)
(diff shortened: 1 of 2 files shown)


indirect:
@deivid-rodriguez have you seen this kind of failure on windows before? I'm confused about why it passes everywhere but windows 😬

deivid-rodriguez:
Yes, I've seen those in the past in other places. I think it would happen randomly only on Windows when trying to rename a folder while being inside it.

deivid-rodriguez:
I'll have a look at this tomorrow, see if I can figure it out.

indirect:
Oh no, I think I figured it out. The existing test hard-codes `/tmp/localpath`, which doesn't exist in Windows. How did this ever pass on Windows before?? 😬 Hopefully this will fix it...

duckinator:
This looks good to me. :+1:

fschwahn:
@indirect this patch is not in `2.2.0.rc.2` - with which version will this be released? This fixes a critical bug which makes it impossible to run `bundle update` for me (which incapacitates eg. dependabot). It's a bit hard to follow development now that the repositories are merged with rubygems (eg. no branches / no tags).

deivid-rodriguez:
I forgot to backport this, sorry. Will set it for the next release.

fschwahn:
awesome, thanks!

deivid-rodriguez:
> It's a bit hard to follow development now that the repositories are merged with rubygems (eg. no branches / no tags).

Stable branch for bundler-2.2 is that same as for rubygems-3.2 (the [3.2 branch](https://github.com/rubygems/rubygems/tree/3.2)), and tags are prefixed with [`bundler-`](https://github.com/rubygems/rubygems/releases/tag/bundler-v2.2.0.rc.2). I know it's currently very weird, will get less weird as we improve things.

deivid-rodriguez:
@indirect I wasn't sure of how to tag this PR. I went with `bundler: minor improvements`, but based on https://github.com/rubygems/rubygems/pull/3865#issuecomment-709483977 it could well be tagged as `bundler: bug fix` too? :thinking:

fschwahn:
> Stable branch for bundler-2.2 is that same as for rubygems-3.2 (the 3.2 branch), and tags are prefixed with bundler-. I know it's currently very weird, will get less weird as we improve things.

Thanks, that helps. I didn't find the tags because github truncates the list, but searching works fine 👍

indirect:
@deivid-rodriguez yeah, I would categorize this as a bugfix. It was the intended behavior, and this was fixing the implementation to match what we intended. 👍

deivid-rodriguez:
Thanks @indirect!

Issue #3866, opened 2020-07-29
dry-run options


nobu:
I'm surprised a little by the followings.

```sh
$ gem check --dryrun
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --dryrun
Did you mean?  dry-run

$ gem cleanup --dry-run
ERROR:  While executing gem ... (OptionParser::InvalidOption)
    invalid option: --dry-run
Did you mean?  dryrun
```

Why are these different?

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
3.2.0.pre1
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

duckinator:
I think it would be best to change `gem check` to use `--dry-run`, since in both RubyGems and Bundler that's the only time that option shows up without the hyphen. The current `--dryrun` can be added as an alias for `--dry-run`, to preserve backwards compatibility.

austinpray:
+1 for `--dry-run`. I would not expect `--dryrun`. `git clean --dry-run ...` for instance.

PR #3867, opened 2020-07-31
Make --dry-run flag consistent across rubygems commands

bronzdoc:
# Description:
closes https://github.com/rubygems/rubygems/issues/3866

Make `--dryrun` flag consistent across Rubygems commands
______________

### DIFF
--- a/lib/rubygems/commands/cleanup_command.rb
+++ b/lib/rubygems/commands/cleanup_command.rb
@@ -15,6 +15,12 @@ class Gem::Commands::CleanupCommand < Gem::Command
       options[:dryrun] = true
     end
 
+    add_option('-n', '-d', '--dryrun',
+               'Do not uninstall gems') do |value, options|
+      options[:dryrun] = true
+    end
+    deprecate_option('--dryrun', extra_msg: 'Use --dry-run instead')
+
     add_option('-D', '--[no-]check-development',
                'Check development dependencies while uninstalling',
                '(default: true)') do |value, options|
@@ -41,7 +47,7 @@ class Gem::Commands::CleanupCommand < Gem::Command
   end
 
   def defaults_str # :nodoc:
-    "--no-dryrun"
+    "--no-dry-run"
   end
 
   def description # :nodoc:
--- a/lib/rubygems/commands/cleanup_command.rb
+++ b/lib/rubygems/commands/cleanup_command.rb
@@ -10,7 +10,7 @@ class Gem::Commands::CleanupCommand < Gem::Command
           :force => false, :install_dir => Gem.dir,
           :check_dev => true
 
-    add_option('-n', '-d', '--dryrun',
+    add_option('-n', '-d', '--dry-run',
                'Do not uninstall gems') do |value, options|
       options[:dryrun] = true
     end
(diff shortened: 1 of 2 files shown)


nobu:
GNU make also provides `-n`, `--just-print`, `--dry-run`, `--recon`.

duckinator:
@bronzdoc on master, `--dryrun` is only used once across Bundler and RubyGems, whereas `--dry-run` is used twice and seems to be the norm elsewhere. Given that, I'd rather see `gem cleanup` switched to use `--dry-run` with `--dryrun` as a backwards-compatible alias, because it would make the whole RubyGems+Bundler pair more internally consistent and more consistent with other software.

deivid-rodriguez:
I agree with @bronzdoc!

deivid-rodriguez:
@bronzdoc Do you agree with @duckinator's suggestion? Are you planning to update the PR with that approach?

bronzdoc:
Dang, I totally forgot about this... I agree!

Will Work on Rubygems later today and will update this PR sorry for taking that long fellas

lib/rubygems/commands/cleanup_command.rb
```diff
@@ -10,10 +10,16 @@ def initialize
           :force => false, :install_dir => Gem.dir,
           :check_dev => true
 
+    add_option('-n', '-d', '--dry-run',
+               'Do not uninstall gems') do |value, options|
+      options[:dryrun] = true
+    end
+
```
deivid-rodriguez, line 17: What happens with the `-n` and `-d` shortcuts? Do they warn? I think they shouldn't.

Also (althought unrelated to this PR but since we're here), why two shortcuts? Seems like an incorrect usage of `add_option` :thinking:.

bronzdoc: `-d` and `-n` will not warn.

> Also (althought unrelated to this PR but since we're here), why two shortcuts? Seems like an incorrect usage of add_option

🤔  I don't know what was the reasoning for those two shortcuts, but I agree with you I think it should be just one

deivid-rodriguez (commented):
I saw that the `--update-sources` option to `gem install` passes a `:Deprecated` first argument to `add_option` and the consequence of this is that the option appears under a separate group called `Deprecated Options` in `gem install --help` output. Maybe it's a good idea to do it here too?

bronzdoc:
Should I deprecate `-d` to and stick to just `-n` since im already touching those flags in this PR?

deivid-rodriguez:
Not sure :thinking:. I think we can leave it as it is, and remove it if it confuses someone, but I don't have an issue with deprecating it either.

bronzdoc:
Ok, sounds good

lib/rubygems/commands/cleanup_command.rb
```diff
@@ -10,11 +10,17 @@ def initialize
           :force => false, :install_dir => Gem.dir,
           :check_dev => true
 
-    add_option('-n', '-d', '--dryrun',
+    add_option('-n', '-d', '--dry-run',
                'Do not uninstall gems') do |value, options|
       options[:dryrun] = true
     end
 
+    add_option(:deprecated, '--dryrun',
```
deivid-rodriguez, line 18: Could you fix the casing to `:Deprecated`? Output looks better that way ("Deprecated Options" instead of "deprecated Options").

deivid-rodriguez (approved):

deivid-rodriguez:
CI failure is unrelated (possibly fixed by #3850), so merging.

Thanks @bronzdoc!

PR #3868, opened 2020-07-31
Deprecate server command

bronzdoc:
# Description:

Deprecate Server command since is not used and is currently unmaintained. 
______________

### DIFF
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -174,8 +174,13 @@ class Gem::CommandManager
     else
       cmd_name = args.shift.downcase
       cmd = find_command cmd_name
-      cmd.invoke_with_build_args args, build_args
-      cmd.deprecation_warning if cmd.deprecated?
+      if cmd_name == "server"
+        cmd.deprecation_warning if cmd.deprecated?
+        cmd.invoke_with_build_args args, build_args
+      else
+        cmd.invoke_with_build_args args, build_args
+        cmd.deprecation_warning if cmd.deprecated?
+      end
     end
   end
 
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -174,13 +174,8 @@ class Gem::CommandManager
     else
       cmd_name = args.shift.downcase
       cmd = find_command cmd_name
-      if cmd_name == "server"
-        cmd.deprecation_warning if cmd.deprecated?
-        cmd.invoke_with_build_args args, build_args
-      else
-        cmd.invoke_with_build_args args, build_args
-        cmd.deprecation_warning if cmd.deprecated?
-      end
+      cmd.deprecation_warning if cmd.deprecated?
+      cmd.invoke_with_build_args args, build_args
     end
   end
 
--- a/lib/rubygems/commands/server_command.rb
+++ b/lib/rubygems/commands/server_command.rb
@@ -1,8 +1,12 @@
 # frozen_string_literal: true
 require 'rubygems/command'
 require 'rubygems/server'
+require 'rubygems/deprecate'
 
 class Gem::Commands::ServerCommand < Gem::Command
+  extend Gem::Deprecate
+  rubygems_deprecate_command
+
   def initialize
     super 'server', 'Documentation and gem repository HTTP server',
           :port => 8808, :gemdir => [], :daemon => false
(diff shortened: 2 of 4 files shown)


lib/rubygems/command_manager.rb
```diff
@@ -174,8 +174,13 @@ def process_args(args, build_args=nil)
     else
       cmd_name = args.shift.downcase
       cmd = find_command cmd_name
-      cmd.invoke_with_build_args args, build_args
-      cmd.deprecation_warning if cmd.deprecated?
+      if cmd_name == "server"
+        cmd.deprecation_warning if cmd.deprecated?
+        cmd.invoke_with_build_args args, build_args
```
deivid-rodriguez, line 179: I think this change is a good thing in general, we could completely remove the check for the server command and always do this?

duckinator: I agree -- I think this is a good change in general. I'd be making warnings always show up first, regardless of the command. :slightly_smiling_face:

bronzdoc:
@deivid-rodriguez @duckinator Addressed your comments, thanks!

duckinator (approved):
This looks good to me, as long as all the checks pass. :+1:

deivid-rodriguez:
I had some hesitations about doing this after seeing https://github.com/rubygems/rubygems/issues/4016, which shows how having some "toy gem server tool" cab be quite useful for educational/demonstration purposes.

Since the original reason for deprecating it was "we don't have bandwidth to address security issues on it", I wonder if, instead of deprecating it, we should show a note about it on usage and recommend that it should never be used in production. :thinking:

Also somewhat related https://bugs.ruby-lang.org/issues/17303, because `webrick` is used by `gem server` and because of reason of removal is the same as in here.

duckinator:
@deivid-rodriguez I am okay with a tool akin to `gem server`, but the current one seems to cause more problems than it solves. If someone is willing and able to take up the work of either rewriting it from scratch or cleaning it up to the point it doesn't have gaping security holes, I am okay with re-adding it. But, if it is not vastly improved, I am strongly in favor of working towards removing it.

deivid-rodriguez:
Well, I haven't look into the issues reported about `gem server` and in general I can't really make a judgement on the "the current one seems to cause more problems than it solves" statement. I just figured, since `gem server` seems useful for educational/demo purposes, that adding a disclaimer on usage could be enough. But you're better qualified than me to make this decision, since you have actually looked into this, so I'm fine with it :+1:.

sonalkr132:
> Since the original reason for deprecating it was "we don't have bandwidth to address security issues on it"
> which shows how having some "toy gem server tool" cab be quite useful for educational/demonstration purposes.

I would argue that gemstash is just as easy to install and use as `gem server`. we shouldn't have to duplicate our effort for this, gemstash is a project under rubygems organization.
I am also not sure if the usecase of easy to access "toy" implementation justifies potentially insecure hosts everywhere ruby is installed. The open HO reports are not the only reports we received for gem server over the year. I am afraid unmaintained/half baked implementation is doing more harm than good.

deivid-rodriguez:
Alright! :+1:

simi:
:thinking: I was using `gem server` for documentation server when i was offline few times. From my point of view I'm ok with deprecating gem server command, but it will be nice to also point out to users what to use instead to have the same experience. Currently using `gemstash start` doesn't provide the same experience as `gem server`. It runs in daemonized mode not exposing which port is used, by using `gemstash start --no-daemonize` it points out there's `localhost:9292` ready to server, but if you reach that URL it just redirects you to `rubygems.org`. IMHO overall experience is totally different.

Also notice there are some blog posts mentioning `gem server` like https://www.mikeperham.com/2010/12/16/using-rdoc/.

Would it be possible to at least extract `gem server` to plugin and mention that in deprecation message? Alternatively I can try to contribute server command to rdoc itself to be able to just mention this in deprecation message.

sonalkr132:
Yeah, gemstash doesn't have anything for doc. To be fair, "gem server" and a local documentation viewer are two very different responsibilities and probably should have never been part of the same command. 
I am not really sure about the recommendation of `yard server -g`. Probably will good clarification that it is only for seeing docs and not an alternative for "gem server".

simi:
@sonalkr132 To be honest, until going thru this discussion, I had no idea about the gemserver purpose of `gem server` and I was really confused why it is not called gem docs or something since in my eyes it was serving just docs. ¯\_(ツ)_/¯

> Probably will good clarification that it is only for seeing docs and not an alternative for "gem server".

Yes, that's my idea. For deprecation message provide something like:

```
Hello, gem server was deprecated and it will be removed in next major version.

For simple gem server:
- please take a look at official rubygems gemstash project (gem install gemstash & gemstash start)
For local docs server:
- install yard and start `yarn server -g` to browse locally installed gems docs
```

PR #3872, opened 2020-08-02
Update man page and deprecation warning for binstubs --all

### LINKED ISSUE #3251: Using binstubs for all gems with the deprecated --binstubs
I am trying to figure out how to properly set up bundler in Github actions, and bumped into what seems to be a missing or at least confusing functionality.

### `bundle install --binstubs` is deprecated

```
$ bundle install --binstubs ./bin
[DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs`
```

### `bundle binstubs` does not have a way to stub all gems

And, in addition it points to the deprecated `--binstubs` flag, which I am guessing is a documentation issue.

```
$ bundle binstubs --help
SYNOPSIS
       bundle binstubs GEM_NAME [--force] [--path PATH] [--standalone]

# ... snip ...

BUNDLE INSTALL --BINSTUBS
       To  create  binstubs  for  all  the  gems in the bundle you can use the
       --binstubs flag in bundle install(1) bundle-install.1.html.
```

So - what is the proper way to binstub all gems?

---

To add some context: When using a vendor path for the bundle (for caching), then any installed gem that comes with a binary, is not found in the path. This is why I started looking at this binstubs feature (which I never used before).

I would much rather **avoid creating these binstubs**, and instead, just add the `vendor/bundle/.../bin` folder to the path, but I was unable to find the proper command to obtain this path - as it seems to differ from system to system.

I was hoping there is some `bundler env get bin_path` or similar so I can do this in GitHub Actions:

```yaml
- name: Bundle install
  run: |
    gem install bundler
    bundle config path vendor/bundle
    bundle install --jobs 4 --retry 3
    echo "::add-path::$(bundle env get bin_path)"   
```

rreinhardt9:
Resolves #3251

Update the deprecation warning to point the user to the `--all` option
for bundle binstubs as the preferred way to install binstubs for all
gems in a bundle.

Update `bundle binstubs` man page to include the `--all` option :tada:
and remove the note pointing to the now deprecated method of installing
all binstubs.

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -81,10 +81,10 @@ namespace :man do
       Spec::Rubygems.gem_require("ronn")
     rescue Gem::LoadError => e
       desc "Build the man pages"
-      task(:build) { abort "We couln't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
+      task(:build) { abort "We couldn't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
 
       desc "Verify man pages are in sync"
-      task(:check) { abort "We couln't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
+      task(:check) { abort "We couldn't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
     else
       directory "man"
 
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -53,7 +53,7 @@ module Bundler
 
       if options["binstubs"]
         Bundler::SharedHelpers.major_deprecation 2,
-          "The --binstubs option will be removed in favor of `bundle binstubs`"
+          "The --binstubs option will be removed in favor of `bundle binstubs --all`"
       end
 
       Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.feature_flag.plugins?
--- a/bundler/man/bundle-binstubs.1
+++ b/bundler/man/bundle-binstubs.1
@@ -36,5 +36,7 @@ Makes binstubs that can work without depending on Rubygems or Bundler at runtime
 \fB\-\-shebang\fR
 Specify a different shebang executable name than the default (default \'ruby\')
 .
-.SH "BUNDLE INSTALL \-\-BINSTUBS"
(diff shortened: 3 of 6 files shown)


rreinhardt9:
I'm super new to contributing period and to rubygems specifically, so do let me know if there is something I can do to be more helpful because I probably just don't realize it 😆 

One question that popped up when working on this; when running `bin/rake man:build` it creates a bunch of diffs in other places besides just the man page I changed. I get diffs on these output files:
Screen Shot 2020-08-02 at 2 35 51 PM

But the weird part is that it's just adding/removing whitespace to balance out the words in the row 🤔 
Screen Shot 2020-08-02 at 2 36 01 PM

I didn't commit those updates because it seemed like that was probably something with my configuration because otherwise the CI check would have not passed for other branches I would figure. Is that a correct assumption or is should I look more into that?

bundler/Rakefile
```diff
@@ -81,10 +81,10 @@ namespace :man do
       Spec::Rubygems.gem_require("ronn")
     rescue Gem::LoadError => e
       desc "Build the man pages"
-      task(:build) { abort "We couln't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
+      task(:build) { abort "We couldn't activate ronn (#{e.requirement}). Try `gem install ronn:'#{e.requirement}'` to be able to build the help pages" }
```
rreinhardt9, line 84: While working on getting set up locally, I just stumbled on and fixed this typo.

bundler/lib/bundler/cli/install.rb
```diff
@@ -53,7 +53,7 @@ def run
 
       if options["binstubs"]
         Bundler::SharedHelpers.major_deprecation 2,
-          "The --binstubs option will be removed in favor of `bundle binstubs`"
+          "The --binstubs option will be removed in favor of `bundle binstubs --all`"
```
rreinhardt9, line 56: From what I understand if someone is looking for the behavior they had before using `bundle install --binstubs` which was to create binstubs for the whole bundle, `bundle binstubs --all` would be the command/ flag combination to get that. Is that a correct understanding?

colby-swandale: That's correct

bundler/Rakefile
deivid-rodriguez: Thanks! In general, I try to split unrelated changes to separate commits, but it's no big deal :)

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks so much @rreinhardt9!!

PR #3876, opened 2020-08-04
Add writable check for cache dir

xndcn:
Sometimes "install_dir/cache" directory is not writable although "install_dir" is writable.

# Description:

Sometimes `install_dir` is writable but `install_dir`/cache may not. Under this condition, download gem will fail. A screenshot is also attached in following comment.
Some issues can also be found in history with keywords like "rb_sysopen cache"

When `install_dir`/cache is not writable, it will fallthrough to `Gem.home_dir`/cache directory.

# Tasks:

- [ ] Write tests


hsbt:
It's reasonable. Thanks.

PR #3877, opened 2020-08-04
Look for absolute path when removing bundler/setup from RUBYOPT in Bundler.unbundled_env method

### LINKED ISSUE #3705: 'RUBYOPT' still contain "-r bundler/setup" when using `unbundled_env`
Hi everyone 👋 , thank you very much for the work on Rubygems/Bundler.

I found out an issue related to `bundle exec` which set the `RUBYOPT` env variables to something like: `-r/opt/rubies/2.7.1/lib/ruby/2.7.0/bundler/setup`

When using `Bundler.with_unbundled_env`, bundler is supposed is to strip out the `-r bundler/setup` part but the regex is incorrect and assume the path is relative whereas it was set as an absolute path https://github.com/rubygems/rubygems/blob/e2a4895aab3225652d5946dc34623fb4177a2678/bundler/lib/bundler.rb#L356

Concretely, this creates a bug if you try to `bundle install` programatically after running `bundle exec`.

```ruby
# Gemfile

sources 'https://rubygems.org'
gem 'activesupport'

# --------------

# Gemfile_bis

sources 'https://rubygems.org'
gem 'activerecord'

# --------------

# script.rb

Bundler.with_unbundled_env do
  system('BUNDLE_GEMFILE=Gemfile_bis bundle install')
end

# --------------

# bundle exec script.rb
# Could not find gem 'activerecord' in any of the gem sources listed in your Gemfile.
# Run `bundle install` to install missing gems.
```

Tested on Ruby 2.7.1 , Bundler 2.1.4. The bug seems present on Bundler edge.

rreinhardt9:
# Why?

Resolves #3705

Bundler no longer supplies the require in RUBYOPT as a relative path but as an absolute path (see https://github.com/rubygems/rubygems/commit/afc00b1e74466cf2b66cac652acc731f8c713a76 ) 

# What?

This updates `unbundled_env` to match that behavior and looks for a require with an absolute path to `bundler/setup` when removing traces of bundler from the environment variable RUBYOPT. It will also continue to look for relative paths to remove as well in order to not create issues with existing scripts people might have that sets it as a relative path.

# Testing Notes

In the examples below, to follow along you'll have to change the absolute paths in the commands based on where your ruby gems and bundler repo are located on your computer. I found `$ gem list -d bundler` helpful to see the path bundler is installed at.

I would expect this issue to appear only on 2.2.0.rc.1 and later because that is the release the commit was in changing RUBYOPT to have an absolute and not relative path and it was:

```
$ RUBYOPT=-r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.2.0.rc.1/lib/bundler/setup bundler/bin/bundle exec ruby -rbundler -e'puts Bundler::VERSION'
=> 2.2.0.rc.1

$ RUBYOPT=-r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.2.0.rc.1/lib/bundler/setup bundle exec ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
=> -r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.2.0.rc.1/lib/bundler/setup
```

I then confirmed that the -r statement in RUBYOPT was removed when running the bundler code from this feature branch:

```
$ RUBYOPT=-r/Users/ross/dev/rubygems/bundler/lib/bundler/setup bundle exec ruby -rbundler -e'puts Bundler::VERSION'
=> 2.2.0.rc.2

$ RUBYOPT=-r/Users/ross/dev/rubygems/bundler/lib/bundler/setup bundle exec ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
=> # Nothing returned
```

I went back to version 2.1.2 to see what the behavior was before:
```
$ bundle exec ruby -rbundler -e'puts Bundler::VERSION'
=> 2.1.2

$ bundle exec ruby -rbundler -e'puts ENV["RUBYOPT"]'
=> -r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/lib/bundler/setup

$ bundle exec ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
=> # Nothing

$ RUBYOPT=-rbundler/setup bundle exec ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
=> # Still nothing

# Doesn't remove an absolute path
$ RUBYOPT=-r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/lib/bundler/setup bundle exec ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
=> -r/Users/ross/.rbenv/versions/2.7.0/lib/ruby/gems/2.7.0/gems/bundler-2.1.4/lib/bundler/setup
```

This gives me confidence that:
- The issue was only present in and after 2.2.0.rc.1
- The change in this behavior addresses the issues seen in 2.2.0.rc.1

# Caveats

One caveat to mention is that before using `unbundled_env` would always remove `bundler/setup` from RUBYOPT regardless of the version of bundler the entry in RUBYOPT was targeted at. After this change, `unbundled_env` only removes the absolute path for `bundler/setup` that points to the currently activated version of bundler. We discussed making a broader regex for trying to clean up any -r for `bundler/setup` but were concerned that might be too broad and might lead to trouble.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -353,7 +353,9 @@ EOF
       env.delete_if {|k, _| k[0, 7] == "BUNDLE_" }
 
       if env.key?("RUBYOPT")
-        env["RUBYOPT"] = env["RUBYOPT"].sub "-rbundler/setup", ""
+        rubyopt = env["RUBYOPT"].split(" ")
+        rubyopt.delete("-rbundler/setup")
+        env["RUBYOPT"] = rubyopt.join(" ")
       end
 
       if env.key?("RUBYLIB")
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -354,6 +354,7 @@ EOF
 
       if env.key?("RUBYOPT")
         rubyopt = env["RUBYOPT"].split(" ")
+        rubyopt.delete("-r#{File.expand_path("bundler/setup", __dir__)}")
         rubyopt.delete("-rbundler/setup")
         env["RUBYOPT"] = rubyopt.join(" ")
       end
--- a/bundler/spec/runtime/with_unbundled_env_spec.rb
+++ b/bundler/spec/runtime/with_unbundled_env_spec.rb
@@ -84,11 +84,11 @@ RSpec.describe "Bundler.with_env helpers" do
       expect(last_command.stdboth).to include "false"
     end
 
-    it "should remove '-rbundler/setup' from RUBYOPT" do
+    it "should remove '-rbundler/setup' from RUBYOPT even if it was present in original env" do
       create_file("source.rb", <<-RUBY)
         print #{modified_env}['RUBYOPT']
       RUBY
-      ENV["RUBYOPT"] = "-W2 -rbundler/setup #{ENV["RUBYOPT"]}"
+      ENV["BUNDLER_ORIG_RUBYOPT"] = "-W2 -rbundler/setup #{ENV["RUBYOPT"]}"
       simulate_bundler_version_when_missing_prerelease_default_gem_activation do
         bundle_exec_ruby bundled_app("source.rb")
       end
--- a/bundler/spec/runtime/with_unbundled_env_spec.rb
+++ b/bundler/spec/runtime/with_unbundled_env_spec.rb
@@ -84,7 +84,19 @@ RSpec.describe "Bundler.with_env helpers" do
       expect(last_command.stdboth).to include "false"
     end
 
-    it "should remove '-rbundler/setup' from RUBYOPT even if it was present in original env" do
+    it "should remove absolute path to 'bundler/setup' from RUBYOPT even if it was present in original env" do
+      create_file("source.rb", <<-RUBY)
+        print #{modified_env}['RUBYOPT']
+      RUBY
+      setup_require = "-r#{lib_dir}/bundler/setup"
+      ENV["BUNDLER_ORIG_RUBYOPT"] = "-W2 #{setup_require} #{ENV["RUBYOPT"]}"
+      simulate_bundler_version_when_missing_prerelease_default_gem_activation do
+        bundle_exec_ruby bundled_app("source.rb")
+      end
+      expect(last_command.stdboth).not_to include(setup_require)
+    end
+
+    it "should remove relative path to 'bundler/setup' from RUBYOPT even if it was present in original env" do
       create_file("source.rb", <<-RUBY)
         print #{modified_env}['RUBYOPT']
       RUBY


bundler/lib/bundler.rb
```diff
@@ -344,25 +344,7 @@ def clean_env
...
-      end
-
-      if env.key?("RUBYLIB")
-        rubylib = env["RUBYLIB"].split(File::PATH_SEPARATOR)
-        rubylib.delete(File.expand_path("..", __FILE__))
-        env["RUBYLIB"] = rubylib.join(File::PATH_SEPARATOR)
-      end
-
-      env
+      original_env.reject {|k, _| k[0, 7] == "BUNDLE_" }
```
rreinhardt9, line 347: [Hash#reject](https://ruby-doc.org/core-2.7.1/Hash.html#method-i-reject) is used here in a similar way to `delete_if` was used before, the difference being reject does not change the env hash in place, but will just return a new hash containing entries that the block returned false for.

bundler/spec/runtime/with_unbundled_env_spec.rb
```diff
@@ -72,32 +72,22 @@ def run_bundler_script(env, script)
       bundle_exec_ruby bundled_app("source.rb")
       expect(out).to eq original
     end
-  end
-
-  shared_examples_for "an unbundling helper" do
```
rreinhardt9, line 77: It's a little confusing in the diff, but what I did here was move all the tests for RUBYOPT, MANPATH, and RUBYLIB out of the shared "unbundling helper" and made them tests for `original_env`. The shared "unbundling helper" is now much smaller and appears beneath them.

bundler/spec/runtime/with_unbundled_env_spec.rb
```diff
@@ -72,32 +72,22 @@ def run_bundler_script(env, script)
...
-      ENV["BUNDLE_PATH"] = "./foo"
-      bundle_exec_ruby bundled_app("source.rb")
-      expect(last_command.stdboth).to include "false"
-    end
 
-    it "should remove '-rbundler/setup' from RUBYOPT" do
+    it "should remove 'bundler/setup' from RUBYOPT" do
       create_file("source.rb", <<-RUBY)
-        print #{modified_env}['RUBYOPT']
+        print Bundler.original_env['RUBYOPT']
```
rreinhardt9, line 78: the print statement here was dynamic before because it was in the shared example; now it's just testing specifically for `original_env` and is no longer in the shared example.

bundler/spec/runtime/with_unbundled_env_spec.rb
```diff
@@ -72,32 +72,22 @@ def run_bundler_script(env, script)
...
       RUBY
-      ENV["RUBYOPT"] = "-W2 -rbundler/setup #{ENV["RUBYOPT"]}"
+      setup_require = "-r#{lib_dir}/bundler/setup"
+      ENV["RUBYOPT"] = "-W2 #{setup_require} #{ENV["RUBYOPT"]}"
```
rreinhardt9, line 81: using `setup_require` here attempts to make the `bundler/setup` path more realistic since it's now explicitly in the lib_dir and not unqualified when it's set for real by bundler.

bundler/spec/runtime/with_unbundled_env_spec.rb
```diff
@@ -116,6 +106,17 @@ def run_bundler_script(env, script)
     end
   end
 
+  shared_examples_for "an unbundling helper" do
```
rreinhardt9, line 109: This is what is left of the shared example that the other specs were moved out of.

bundler/lib/bundler.rb
```diff
@@ -344,25 +344,7 @@ def clean_env
 
     # @return [Hash] Environment with all bundler-related variables removed
     def unbundled_env
-      env = original_env
-
-      if env.key?("BUNDLER_ORIG_MANPATH")
-        env["MANPATH"] = env["BUNDLER_ORIG_MANPATH"]
```
rreinhardt9, line 350: `BUNDLER_ORIG_` appears to have been an old prefix that isn't used anymore https://github.com/rubygems/rubygems/commit/8e3b925413e1ed5f721e2f7d90dfe6f0e3c43af5#diff-83eb5e43a306cb4526be4c5e5a790479L9

deivid-rodriguez:
Hi @rreinhardt9. After a second look, I don't think this is correct.

There's two helpers here, meant to be used for different things:

* `Bundler.with_unbundled_env` is meant to cleanup every bundler related variable, not matter what the original environment is.
* `Bundler.with_original_env` is meant to restore the environment as it was before bundler was loaded.

So, for example,

```
$ RUBYOPT=-rbundler/setup ruby -rbundler -e'puts Bundler.unbundled_env["RUBYOPT"]'
```

should print nothing, as opposed to `-rbundler/setup`. I believe your changes essentially remove the distinction between these two helpers.

rreinhardt9:
@deivid-rodriguez Ah, gotcha thanks!!! That helps give me clarity... so the intention of `Bundler.with_unbundled_env` is to go even further than `Bundler.with_original_env` and try to clear out any bundler related environment variables that were set, even if they were present in the original environment.

I'll look at it again from that perspective, I'm thinking in that case we for RUBYOPT we change:
```Ruby
      if env.key?("RUBYOPT")
        env["RUBYOPT"] = env["RUBYOPT"].sub "-rbundler/setup", ""
      end
```
so that the regex matches any path ending in bundler/setup (something like `env["RUBYOPT"].sub /-r.*bundler\/setup/, "")

I'll:
- [ ] Put back check for RUBYOPT and update the regex
- [ ] Undo the changes to the RUBYLIB and MANPATH code
- [ ] move the tests back to where they were/ update RUBYOPT test for the new regex check

deivid-rodriguez:
Yes, exactly.

Only thing I'm not sure about is whether to use a regexp to detect `bundler/setup` like you suggested, or add a separate substitution to look for an exact match of the expanded `bundler/setup` path that [we previously added](https://github.com/rubygems/rubygems/blob/1a70d323d9e866a550491623544e7258a8616889/bundler/lib/bundler/shared_helpers.rb#L306).

rreinhardt9:
I'm wondering about that too...

On one hand, if the method is intended to remove any traces of bundler that were in the original environment the regex would catch bundler/setup entries that might not be the same as that from the currently running version of bundler. I'm a little concerned the regex could carry some risk because its effects would be more broad. 

Additionally, while dealing with RUBYLIB in the same method it looks for an exact expanded match to remove; so doing the same for RUBYOPT would be consistent with that behavior.

Based on that removing the exact expanded path for RUBYOPT seems like the way to go. 🤔

RoMi67:
its a ruby gem ball not oath

￼

Robert Mizani

BioBlockchain Eco-Trust, Honorary Biologist/Creator/Contractor/G.E.O.
￼
1-818-621-3200
5550 Buffalo Ave
Unit RM
Los Angeles, South California91403-2525
https://www.BiologicalScienceConsulting.COM

________________________________

RoMi67:
raw

￼

Robert Mizani

BioBlockchain Eco-Trust, Honorary Biologist/Creator/Contractor/G.E.O.
￼
1-818-621-3200
5550 Buffalo Ave
Unit RM
Los Angeles, South California91403-2525
https://www.BiologicalScienceConsulting.COM

________________________________

RoMi67:
[Image]
________________________________

RoMi67:
# # #[Image]


________________________________

RoMi67:
carry on

￼

Robert Mizani

BioBlockchain Eco-Trust, Honorary Biologist/Creator/Contractor/G.E.O.
￼
1-818-621-3200
5550 Buffalo Ave
Unit RM
Los Angeles, South California91403-2525
https://www.BiologicalScienceConsulting.COM

________________________________

deivid-rodriguez:
Agreed!

deivid-rodriguez:
If you remember, when you update the PR, could you also update the title to reflect what it's finally doing?

bundler/lib/bundler.rb
```diff
@@ -353,7 +353,9 @@ def unbundled_env
       env.delete_if {|k, _| k[0, 7] == "BUNDLE_" }
 
       if env.key?("RUBYOPT")
-        env["RUBYOPT"] = env["RUBYOPT"].sub "-rbundler/setup", ""
+        rubyopt = env["RUBYOPT"].split(" ")
+        rubyopt.delete("-r#{File.expand_path("bundler/setup", __dir__)}")
+        env["RUBYOPT"] = rubyopt.join(" ")
```
rreinhardt9, line 358: I followed the pattern in RUBYLIB below and split the string into an Array on " ", then deleted the `bundler/setup` entry, then joined the RUBYOPT array back together with " ". I did that to avoid having to deal with white space when deleting the entry.

For example, if RUBYOPT contained`-r/path/to/bundler/setup -r/something/else` and I just subbed out `-r/path/to/bundler/setup` for `""` RUBYOPT would now have a leading space left over.

But, if I tried to remove `-r/path/to/bundler/setup ` (with trailing space) then it wouldn't work when it was the ONLY entry because there would be no trailing space. It would also not work if it was the last entry.

Splitting on " " helps us avoid all of that because it will "just work" for any of the above cases and it won't leave the leading space behind in RUBYOPT.

The default for `String#split` is " " so I don't technically need to specify it, but I went ahead and specified it explicitly here for clarity (so that it's clearer what is happening)

bundler/spec/runtime/with_unbundled_env_spec.rb
```diff
@@ -84,15 +84,16 @@ def run_bundler_script(env, script)
...
     end
 
-    it "should remove '-rbundler/setup' from RUBYOPT" do
+    it "should remove '-rbundler/setup' from RUBYOPT even if it was present in original env" do
       create_file("source.rb", <<-RUBY)
         print #{modified_env}['RUBYOPT']
       RUBY
-      ENV["RUBYOPT"] = "-W2 -rbundler/setup #{ENV["RUBYOPT"]}"
+      setup_require = "-r#{lib_dir}/bundler/setup"
+      ENV["BUNDLER_ORIG_RUBYOPT"] = "-W2 #{setup_require} #{ENV["RUBYOPT"]}"
```
rreinhardt9, line 92: Because we are trying to simulate that the RUBYOPT contained this `bundler/setup` value IN the original env, we actually have to set it in `BUNDLER_ORIG_RUBYOPT` for the test. We are simulating that this `bundler/setup` entry in RUBYOPT existed BEFORE bundler was activated (even though it didn't).

Otherwise, if we set it in `RUBYOPT` it will already be removed even without the code in `unbundled_env` (The way the spec was written before it still passed when I commented out the code in `unbundled_env` for handling RUBYOPT).

That is because `unbundled_env` calls `original_env` as it's starting point, which will remove whatever we set here in RUBYOPT because it was not in the original env when the test suite first activated bundler.

rreinhardt9:
@deivid-rodriguez Thanks so much for your patience helping me through this! I've made updates we talked about and updated the PR description/title to reflect the new approach and it's ready for another look 🎉

deivid-rodriguez:
Looks good, but I feel we should still keep the old `-rbundler/setup` check as well. I believe people putting `-rbundler/setup` in `RUBYOPT` on their scripts will still expect `Bundler.with_unbundled_env` to remove that.

rreinhardt9:
That is a great point 😬 🤦 thanks for catching that!!!  I just added a commit handling relative in addition to absolute paths. Once you get a chance to check it out let me know if you want that squashed before the merge and I can do that 🎉

deivid-rodriguez (approved):
Looks good!

If you wanted to curate the commit history, I'd split the three different improvements you ended up making into separate commits (and move your comments on the code to the commit message of each individual commit). Namely:

* Improve space handling when removing entries from `RUBYOPT`, as you [explained here](https://github.com/rubygems/rubygems/pull/3877/files#r479708565).
* Improve coverage of existing specs to better simulate the original presence of `-rbundler/setup` in `RUBYOPT`, as you [explained here](https://github.com/rubygems/rubygems/pull/3877/files#r479708949).
* The bug fix itself with the new spec.

But I'm aware of my overpickyness about git history, so this is also fine as is :sweat_smile:.

rreinhardt9:
I love it! I'm a fan of an informative git history as well and don't mind splitting those 🎉 I'll do that!

deivid-rodriguez (approved):
Looks great!

deivid-rodriguez:
Thanks!!!

PR #3886, opened 2020-08-14
Add `--all-platforms` flag to `bundle binstubs` to generate binstubs for all platforms

### LINKED ISSUE #2120: Give option to generate Windows bin stubs regardless of host OS
I stumbled across the fact that when working with jruby on a Linux system, no Windows stub files are generated for installed gems. This is reasonable for a native ruby installation, but it would be nice to also generate windows stubs in a jruby environment because this would make the jruby folder portable also across OS types as long as there is a JRE available.

Please see the discussion over at https://github.com/jruby/jruby/issues/4907

Being on a Windows host, currently also shebang scripts are generated, so the situation is a little bit assymetrical. I also noticed that when being on a Windows system the setting for shebang style seems to be ignored because I then get a `#! jruby` shebang even though jruby configures to be using `/usr/bin/env`. Should I open another issue for this?

SeekingMeaning:
# Description:

This closes #2120

> I stumbled across the fact that when working with jruby on a Linux system, no Windows stub files are generated for installed gems. This is reasonable for a native ruby installation, but it would be nice to also generate windows stubs in a jruby environment because this would make the jruby folder portable also across OS types as long as there is a JRE available.

This adds `--all-platforms` option suggested by segiddins

I think the manpages also should be updated but I'm not quite sure how to go about doing that

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -381,6 +381,8 @@ module Bundler
       "Make binstubs that can work without the Bundler runtime"
     method_option "all", :type => :boolean, :banner =>
       "Install binstubs for all gems"
+    method_option "all-platforms", :type => :boolean, :default => false, :banner =>
+      "Install binstubs for all platforms"
     def binstubs(*gems)
       require_relative "cli/binstubs"
       Binstubs.new(options, gems).run
--- a/bundler/lib/bundler/cli/binstubs.rb
+++ b/bundler/lib/bundler/cli/binstubs.rb
@@ -16,7 +16,11 @@ module Bundler
       Bundler.settings.set_command_option_if_given :shebang, options["shebang"]
       installer = Installer.new(Bundler.root, Bundler.definition)
 
-      installer_opts = { :force => options[:force], :binstubs_cmd => true }
+      installer_opts = {
+        :force => options[:force],
+        :binstubs_cmd => true,
+        :all_platforms => options["all-platforms"],
+      }
 
       if options[:all]
         raise InvalidOption, "Cannot specify --all with specific gems" unless gems.empty?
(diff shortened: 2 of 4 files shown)


deivid-rodriguez (changes requested):
Hi @SeekingMeaning! I didn't review this because I thought it needed work due to some specs being broken. But now I looked at the failing specs and I'm not 100% sure they failures are related to this PR. They seem to break only on Windows, so that might indicate a relationship, but still I'm not fully sure.

Could you rebase and force-push this PR just to make sure?

deivid-rodriguez:
And thanks for this, by the way! :smiley:

bundler/lib/bundler/cli.rb
```diff
@@ -381,6 +381,8 @@ def info(gem_name)
       "Make binstubs that can work without the Bundler runtime"
     method_option "all", :type => :boolean, :banner =>
```
SeekingMeaning, line 382: Perhaps this should be renamed to `all-gems` (and have `--all` added as an alias)?

deivid-rodriguez: Not sure, I think `--all` is mostly fine?

deivid-rodriguez:
@SeekingMeaning This looks good, but could you add some tests for it?

SeekingMeaning:
Would adding this to [`binstubs_spec.rb`](https://github.com/rubygems/rubygems/blob/master/bundler/spec/commands/binstubs_spec.rb) work?

```rb
    it "allows installing binstubs for all platforms" do
      install_gemfile <<-G
        gem "rack"
      G

      bundle "binstubs rack --all-platforms"

      expect(bundled_app("bin/rackup")).to exist
      expect(bundled_app("bin/rackup.cmd")).to exist
    end
```

deivid-rodriguez:
That seems enough to me, yeah :+1:. But since there's two different places where we're adding the check for the new option to generate the extra binstubs, I'd add a separate spec that covers the remaining place (I think you need to also run `bundle binstubs --standalone --all-platforms`, or something like that).

deivid-rodriguez:
@SeekingMeaning This looks great to me, thanks! Could you rebase on top of latest master and squash both commits into one?

deivid-rodriguez (approved):
Thanks so much @SeekingMeaning!! :purple_heart:

PR #3887, opened 2020-08-17
only stat user home once when path is missing

kares:
noticed several unnecessary stat calls when `Gem.user_home` was set to non existent path.

just poking around RGs

treat `false` as an end state - do not `stat` again when looking for cache and `false` already

--- 

### DIFF
--- a/lib/rubygems/source.rb
+++ b/lib/rubygems/source.rb
@@ -35,6 +35,7 @@ class Gem::Source
     end
 
     @uri = uri
+    @update_cache = nil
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I think the CI failure was workarounded recently and not this PR's fault. Could you rebase to verify?

deivid-rodriguez (approved):
Thanks for the rebase and the PR @kares!

This looks good to me provided CI passes :+1:.

PR #3888, opened 2020-08-18
Re-use install update default options

hsbt:
# Description:

I found the current `install_update_default_str` is not used in RubyGems at https://github.com/rubygems/rubygems/pull/2847#issuecomment-670486536

I re-use the its method and extract the current default options to `install_update_options.rb` from ` install_command.rb` and `update_command.rb`.

And I removed `rdoc` document option from `gem update`. It's breaking changes with RubyGems 3.1. But It's inconsistency with `install` command for 9 years ago.  see https://github.com/rubygems/rubygems/blob/master/lib/rubygems/dependency_installer.rb#L20

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -186,7 +186,6 @@ module Gem::InstallUpdateOptions
   def install_update_options
     {
       :document => %w[ri],
-      :wrappers => true,
     }
   end
 
@@ -194,7 +193,7 @@ module Gem::InstallUpdateOptions
   # Default description for the gem install and update commands.
 
   def install_update_defaults_str
-    '--document=ri --wrappers'
+    '--document=ri'
   end
 
 end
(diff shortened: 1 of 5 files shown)


hsbt:
I only integrated the document option. and removing `rdoc` option from `gem update` to fix inconsistency with `gem install`.

lib/rubygems/install_update_options.rb
```diff
@@ -181,10 +181,19 @@ def add_install_update_options
...
+      :document => %w[ri],
+    }
+  end
+
+  ##
+  # Default description for the gem install and update commands.
 
   def install_update_defaults_str
-    '--document=rdoc,ri --wrappers'
+    '--document=ri'
```
deivid-rodriguez, line 196: Do you know what this `--wrappers` thing is?

deivid-rodriguez:
Seems fine! Hopefully it is :smiley:

lib/rubygems/install_update_options.rb
hsbt: The default value of `--wrappers` is `true` on `DependencyInstaller`. I remove it with the other options.

deivid-rodriguez: Yeah, I understood, I was just asking if you now what it does. I think it's about creating binstubs wrappers instead of symlinks, but not sure. Anyways, I was just curious.

PR #3889, opened 2020-08-18
Sort requirements in Gem::Requirement to succeed comparison with different order

toy:
# Description:

This PR ensures that comparison of requirements works properly when order or requirements is different.
Before this PR following two dependencies are considered different:
```ruby
<Gem::Dependency type=:runtime name="fspath" requirements="< 4, >= 2.1">
<Gem::Dependency type=:runtime name="fspath" requirements=">= 2.1, < 4">
```

When debugging an issue toy/image_optim_pack#20 I've found that in certain conditions platform-specific gem is not selected and an invalid warning is displayed due to comparison of equal dependencies failing. I was able to reproduce it by running following commands:

```shell
gem uninstall image_optim_pack --all

gem install image_optim_pack --platform ruby

cat <<GEMFILE > Gemfile
source 'https://rubygems.org'
gem 'image_optim'
GEMFILE
bundle

cat <<GEMFILE > Gemfile
source 'https://rubygems.org'
gem 'image_optim'
gem 'image_optim_pack'
GEMFILE
bundle
```

Debugging the issue I found that inside `Bundler::LazySpecification#__materialize__` the comparison of dependencies failed due to different order of requirements which led to the warning suggesting that ruby and platform specific gems have different dependencies and selected ruby version of the gem. Suggested change fixes the behaviour.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -129,6 +129,34 @@ module Gem
     end
   end
 
+  unless Gem::Requirement.new("> 1", "< 2") == Gem::Requirement.new("< 2", "> 1")
+    class Requirement
+      module OrderIndependentComparison
+        def ==(other)
+          if _requirements_sorted? && other._requirements_sorted?
+            super
+          else
+            _with_sorted_requirements == other._with_sorted_requirements
+          end
+        end
+
+        protected
+
+        def _requirements_sorted?
+          return @_are_requirements_sorted if defined?(@_are_requirements_sorted)
+          strings = as_list
+          @_are_requirements_sorted = strings == strings.sort
+        end
+
+        def _with_sorted_requirements
+          @_with_sorted_requirements ||= _requirements_sorted? ? self : self.class.new(as_list.sort)
+        end
+      end
+
+      prepend OrderIndependentComparison
+    end
+  end
+
   class Platform
     JAVA  = Gem::Platform.new("java") unless defined?(JAVA)
     MSWIN = Gem::Platform.new("mswin32") unless defined?(MSWIN)
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -129,6 +129,7 @@ module Gem
     end
   end
 
+  # comparison is done order independently since rubygems 3.2.0.rc.2
   unless Gem::Requirement.new("> 1", "< 2") == Gem::Requirement.new("< 2", "> 1")
     class Requirement
       module OrderIndependentComparison
--- a/bundler/spec/install/gemfile/platform_spec.rb
+++ b/bundler/spec/install/gemfile/platform_spec.rb
@@ -256,6 +256,35 @@ RSpec.describe "bundle install across platforms" do
     expect(the_bundle).not_to include_gem "CFPropertyList"
   end
 
+  it "works with gems with platform-specific dependency having different requirements order", :rubygems => ">= 3.2.0.rc.2" do
+    simulate_platform x64_mac
+
+    update_repo2 do
+      build_gem "fspath", "3"
+      build_gem "image_optim_pack", "1.2.3" do |s|
+        s.add_runtime_dependency "fspath", ">= 2.1", "< 4"
+      end
+      build_gem "image_optim_pack", "1.2.3" do |s|
+        s.platform = "universal-darwin"
+        s.add_runtime_dependency "fspath", "< 4", ">= 2.1"
+      end
+    end
+
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo2)}"
+    G
+
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo2)}"
+
+      gem "image_optim_pack"
+    G
+
+    expect(err).not_to include "Unable to use the platform-specific"
+
+    expect(the_bundle).to include_gem "image_optim_pack 1.2.3 universal-darwin"
+  end
+
   it "fetches gems again after changing the version of Ruby" do
     gemfile <<-G
       source "#{file_uri_for(gem_repo1)}"
--- a/bundler/spec/install/gemfile/platform_spec.rb
+++ b/bundler/spec/install/gemfile/platform_spec.rb
@@ -256,7 +256,7 @@ RSpec.describe "bundle install across platforms" do
     expect(the_bundle).not_to include_gem "CFPropertyList"
   end
 
-  it "works with gems with platform-specific dependency having different requirements order", :rubygems => ">= 3.2.0.rc.2" do
+  it "works with gems with platform-specific dependency having different requirements order" do
     simulate_platform x64_mac
 
     update_repo2 do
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
@toy I couldn't repro the problem on my linux box with recent versions of bundler & rubygems. Maybe this is fixed in recent versions? We started sorting requirements [here](https://github.com/rubygems/rubygems/pull/2230), but then reverted it [here](https://github.com/rubygems/rubygems/pull/2652). Reason for the revert was that it wasn't necessary for any of the problems we wanted to fix, and that it was pointed out that requirement ordering inside gemspecs might be there for a reason.

toy:
@deivid-rodriguez
Reproducing using docker:
```shell
docker run --rm -i ruby:2.7 bash <<BASH
gem install image_optim_pack --platform ruby

cat <<GEMFILE > Gemfile
source 'https://rubygems.org'
gem 'image_optim'
GEMFILE
bundle

cat <<GEMFILE > Gemfile
source 'https://rubygems.org'
gem 'image_optim'
gem 'image_optim_pack'
GEMFILE
bundle
BASH
```
For me output is (note the `Unable to use the platform-specific` …):
```
Successfully installed fspath-3.1.2
Successfully installed image_size-2.1.0
Successfully installed exifr-1.3.6
Successfully installed progress-3.5.2
Successfully installed in_threads-1.5.4
Rails image assets optimization is extracted into image_optim_rails gem
You can safely remove `config.assets.image_optim = false` if you are not going to use that gem
Successfully installed image_optim-0.27.0
Successfully installed image_optim_pack-0.6.0.20200215
7 gems installed
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Using bundler 2.1.4
Using exifr 1.3.6
Using fspath 3.1.2
Using image_size 2.1.0
Using in_threads 1.5.4
Using progress 3.5.2
Using image_optim 0.27.0
Bundle complete! 1 Gemfile dependency, 7 gems now installed.
Fetching gem metadata from https://rubygems.org/.....
Resolving dependencies...
Unable to use the platform-specific (x86_64-linux) version of image_optim_pack (0.6.0.20200215) because it has different dependencies from the ruby version. To use the platform-specific version of the gem, run `bundle config set specific_platform true` and install again.
Using bundler 2.1.4
Using exifr 1.3.6
Using fspath 3.1.2
Using image_size 2.1.0
Using in_threads 1.5.4
Using progress 3.5.2
Using image_optim 0.27.0
Using image_optim_pack 0.6.0.20200215
Bundle complete! 2 Gemfile dependencies, 8 gems now installed.
```
I understand #2627, the order of requirements should not randomly change in the output, but the issue is still there and as far as I know the order of requirements is not important for comparison. I'll try to change my PR to fix the issue while persisting the order of requirements in the output.

deivid-rodriguez:
@toy Thanks for the docker repro, issue confirmed on my side :+1:. Your suggestion also looks perfectly fine, as long as the original order is preserved for other things, it sounds perfectly reasonable to ignore it for requirement comparison :+1:.

toy:
@deivid-rodriguez Restored changes in `test/rubygems/test_gem_specification.rb` prove that order is preserved.
Not certain if it is ok like this or it is better to find a way to not reevaluate `_sorted_requirements` and `_tilde_requirements` multiple time during same comparison.

deivid-rodriguez:
I'm not sure either. Does memoizing do the trick?

Something else that would be nice is to have a test in `bundler` with the steps provided in the docker image in addition to the unit test. No need to backport the fix itself to `bundler,` restricting the spec to rubygems >= 3.2.0.rc.2 only should be enough and would also document that the scenario needs the latest rubygems to work.

toy:
> I'm not sure either. Does memoizing do the trick?

It should, I had an impression that class is mutable, but reading through it I didn't find mutating methods.
So I'll add memoization.

> Something else that would be nice is to have a test in `bundler` with the steps provided in the docker image in addition to the unit test. No need to backport the fix itself to `bundler,` restricting the spec to rubygems >= 3.2.0.rc.2 only should be enough and would also document that the scenario needs the latest rubygems to work.

I'll try to create a test, `bundler/spec/install/gemfile/platform_spec.rb` looks like the right place for it.

deivid-rodriguez:
Nice @toy, thanks!

I said there was no need to backport the fix itself to `bundler`, but after a second thought it'd be nice to get the fixed in `bundler` independently of the underlying `rubygems`. Could you try something like this in `lib/bundler/rubygems_ext.rb` to see if it works fine against all supported rubies & rubygems?

```diff
diff --git a/bundler/lib/bundler/rubygems_ext.rb b/bundler/lib/bundler/rubygems_ext.rb
index 1d3bdc5565..76b779b5bb 100644
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -129,6 +129,27 @@ def to_lock
     end
   end
 
+  unless Requirement.instance_methods(false).include?(:_sorted_requirements)
+    module OrderIndependentComparison
+      def ==(other)
+        rubygems_comparison = super
+        return rubygems_comparison unless rubygems_comparison == false
+
+        _sorted_requirements == other._sorted_requirements
+      end
+
+      protected
+
+      def _sorted_requirements
+        @_sorted_requirements ||= requirements.sort_by(&:to_s)
+      end
+    end
+
+    class Requirement
+      prepend OrderIndependentComparison
+    end
+  end
+
   class Platform
     JAVA  = Gem::Platform.new("java") unless defined?(JAVA)
     MSWIN = Gem::Platform.new("mswin32") unless defined?(MSWIN)
```

toy:
@deivid-rodriguez I was not sure about calling `super` before and tilde requirements were not handled specially (`Gem::Requirement.new('~> 1.3') == Gem::Requirement.new('~> 1.3.0')` would be `true`).
So I'm patching the class only if it does not handle comparison already and then compare new instances with sorted requirements if requirements are not sorted.

toy:
Force-pushed to keep history clean: found a bug and memoized

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -129,6 +129,34 @@ def to_lock
     end
   end
 
+  unless Gem::Requirement.new("> 1", "< 2") == Gem::Requirement.new("< 2", "> 1")
```
deivid-rodriguez, line 132: Nice idea :+1:!

deivid-rodriguez:
Smart! Thanks for fixing my buggy code :sweat_smile:!

deivid-rodriguez (approved):
:+1:

bundler/lib/bundler/rubygems_ext.rb
deivid-rodriguez: A small little comment here. Feature detection is great, but I also somewhat like hardcoded version checks, because it makes it harder to miss future cleanup opportunities when dropping support for old versions.

So perhaps we can find a middle ground by adding a `# available since 3.2.0.rc.2` comment so that when a I grep for "3.2.0" in the future when dropping rubygems 3.2.0 support, I don't miss this.

toy: Extended to `# comparison is done order independently since rubygems 3.2.0.rc.2`, do you think it is better or not?

deivid-rodriguez: Sure, yes!

deivid-rodriguez:
Thanks @toy!! :purple_heart:

PR #3893, opened 2020-08-22
fix: MiniTest file and class should starts with 'test'

zeero:
# Description:

`bundle gem ${newgem} --test=minitest` command generates `${Newgem}Test` test class in `${newgem}_test.rb` file.
[In the official documentation example](https://docs.ruby-lang.org/en/2.0.0/MiniTest.html#module-MiniTest-label-Unit+tests), Minitest's test class starts with `Test`, instead of ending with `Test`.
Neither will cause any real harm, but I think it's better to match the official one.

`bundle gem ${newgem} --test=minitest` command generates `Test${Newgem}` test class in `$test_{newgem}.rb` file.
Then I modify the Rakefile template to work with `$test_{newgem}.rb` file.

______________

# Tasks:

- [x] Write tests

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -157,8 +157,8 @@ bundler/lib/bundler/templates/newgem/rspec.tt
 bundler/lib/bundler/templates/newgem/rubocop.yml.tt
 bundler/lib/bundler/templates/newgem/spec/newgem_spec.rb.tt
 bundler/lib/bundler/templates/newgem/spec/spec_helper.rb.tt
-bundler/lib/bundler/templates/newgem/test/minitest/newgem_test.rb.tt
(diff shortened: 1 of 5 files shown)


hsbt:
@zeero Can you rebase from the current master branch? I fixed the fragile tests on Windows.

zeero:
@hsbt 
Thanks for your comments and information!
I rebased this branch with the current master.

deivid-rodriguez:
Late to the party but I don't have a strong opinion on this since I try to avoid `minitest` as much as possible. However, it's a deviation from the current behaviour, so people happy with the current behaviour will not be happy. Also, I feel the `_test.rb` is more common out there. So, I would personally not have accepted this, since I don't think there was a strong reason to change it.

Anyways, no big deal.

PR #3897, opened 2020-08-24
Skip the example with the drive letter of Windows

hsbt:
# Description:

In GitHub Actions, The drive letter returns a downcase value.

```
  You have changed in the Gemfile:\n* rack from `no specified source` to `D:/a/rubygems/rubygems/bundler/tmp/2/libs/rack (at master@d8e48a3)`
```
  vs
```
  * rack from `no specified source` to `d:/a/rubygems/rubygems/bundler/tmp/2/libs/rack (at master@d8e48a3)`
```

The examples of Windows environment is always failed in the current master branch.

Skip the failing exapmles in Windows environment.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/bundler/spec/install/deploy_spec.rb
+++ b/bundler/spec/install/deploy_spec.rb
@@ -361,7 +361,10 @@ RSpec.describe "install in deployment or frozen mode" do
       bundle "config --local deployment true"
       bundle :install, :raise_on_error => false
       expect(err).to include("deployment mode")
-      expect(err).to include("You have deleted from the Gemfile:\n* source: #{lib_path("rack-1.0")} (at master@#{revision_for(lib_path("rack-1.0"))[0..6]}")
+      # The drive letter of the Windows environment is fragile value in GitHub Actions
+      unless Gem.win_platform?
+        expect(err).to include("You have deleted from the Gemfile:\n* source: #{lib_path("rack-1.0")} (at master@#{revision_for(lib_path("rack-1.0"))[0..6]}")
+      end
       expect(err).not_to include("You have added to the Gemfile")
       expect(err).not_to include("You have changed in the Gemfile")
     end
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I'm curious about how this started failing. Was this a change in the github actions environment or a change we made?

MSP-Greg:
I believe that Actions Windows images have always had the HDD as `C:` (uppercase) and the SSD drive as `d:` (lowercase).  The repo being tested is always on the SSD (`d:`). Since Actions flushes/removes older logs, it's impossible to verify.

Hence, I believe this is a RG/Bundler change.

deivid-rodriguez:
Thanks, that's useful. It should be a matter of bisecting the latest commits.

MSP-Greg:
JFYI,

Currently, Actions Windows does not have an `ENV['TMPDIR']` setting.  Ruby then uses another ENV setting (`TEMP` or `TMP`) for the path of the tempdir.  On Actions, and most Windows installs, those point a temp folder on the same drive as the OS, hence, `C:`.

For a long time, it has been set in [MSP-Greg/setup-ruby-pkgs](https://github.com/MSP-Greg/setup-ruby-pkgs) to the value of `ENV['RUNNER_TEMP']`, which is on the SSD (`d:`).  Lot's of extension gems are using the Action for updating gcc and installing packages (cross-platform).

I believe the systems currently used for Actions CI are eight core systems, with VM's providing two cores to each CI job.  Depending on lots of factors, the HDD may slow down IO, so that was why `setup-ruby-pkgs`  changed the temp folder to `d:`.

Long story short, `ruby/setup-ruby` may also change this setting in the near future.  I wanted to mention this, but it shouldn't affect this issue...

MSP-Greg:
@deivid-rodriguez 

Well, I just added some checks to my 'Actions runner setup/configuration' repo.

PowerShell has a variable `$PSScriptRoot`, which is equivalent to `__dir__`.  I ran both types of scripts, and both show the SSD as `D:`.

Also, the ENV variables pointed to the SSD have a been changed to `D:`.  The last time I ran it (2020-07-13), all the ENV variables used a lowercase `d:`.  Sorry about the previous info.

deivid-rodriguez:
Ok, so change in Github Actions after all. In order to make our specs stronger, should we normalize casing of the affected environment variables?

PR #3900, opened 2020-08-24
Fix regression related to locked ruby

deivid-rodriguez:

A few months ago we introduced [a change in how we handle the locked ruby version](https://github.com/rubygems/bundler/pull/7559) that has caused a (fortunately still unreleased) regression.

The regression will be hit be applications:

* Using the `ruby` DSL.
* Having a locked ruby version different from the running ruby version.
* Running `bundle update`.

In that case, whereas previously things would just work, we now get an error like the following:

```
Could not find gem 'Ruby (>= 2.2.2, = 2.7.1.83, = 2.6.5.114)' in the local ruby installation. The source contains 'Ruby' at: 2.7.1.83
```

This PR reverts the offending PR, and adds a regression spec.

The offending PR (that tried to reactivate this unsued code, but caused this regression), also linked to an alternative PR removing the unused code instead. As a bonus, this PR backports that alternative solution to the new repository.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -868,8 +868,8 @@ module Bundler
       @metadata_dependencies ||= begin
         ruby_versions = concat_ruby_version_requirements(@ruby_version)
         if ruby_versions.empty? || !@ruby_version.exact?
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Thanks @indirect!

PR #3902, opened 2020-08-25
Fix incorrect re-resolution when path gem excluded and not available

### LINKED ISSUE #3177: Bundler requires all `:path` gems to be in default group
If you have a Gemfile with a local _path_ Gem such as:

```
group :development, :test do
   gem 'demo', :path => "./ruby-gems/demo"
end
```
Even though the `demo` gem is **not used** in any non-test files; bundler will complain:
```
The path `uri:classloader://warbler-example/ruby-gems/demo` does not exist.
```

This is an issue from running within a JAR using JRuby+ Warbler

I have created a minimal example at: https://github.com/fzakaria/warbler-bundler-jruby-failure/tree/demo-development
(Specifically the _demo-development_ branch)

## Environment
```
jruby 9.1.17.0 (2.3.3) 2018-04-20 d8b1ff9 OpenJDK 64-Bit Server VM 25.232-b09 on 1.8.0_232-b09 +jit [darwin-x86_64]

Darwin MacBook-Pro.localdomain 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64
```

deivid-rodriguez:
# Description:

According to the debugging messages, if specs for a path gem are not available at `bundler/setup` time, but the group the gem belongs to has been excluded, it should be assumed that the path source has no changes and continue.

However, there's another check for re-resolution that checks completeness for the current platform that was not filtering for requested groups only, and making this case to incorrectly re-resolve, and as a result fail due to the path gem not being available (even if not requested).

This commit should fix the problem.

Fixes #3177.

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -828,7 +828,7 @@ module Bundler
       end
 
       resolve = SpecSet.new(converged)
-      @locked_specs_incomplete_for_platform = !resolve.for(expand_dependencies(deps), @unlock[:gems], true, true)
+      @locked_specs_incomplete_for_platform = !resolve.for(expand_dependencies(requested_dependencies & deps), @unlock[:gems], true, true)
       resolve = resolve.for(expand_dependencies(deps, true), @unlock[:gems], false, false, false)
       diff    = nil
 
--- a/bundler/spec/runtime/setup_spec.rb
+++ b/bundler/spec/runtime/setup_spec.rb
@@ -635,6 +635,26 @@ RSpec.describe "Bundler.setup" do
       expect(the_bundle).to include_gems "activesupport 2.3.2"
     end
 
+    it "remembers --without and does not bail on bare Bundler.setup, even in the case of path gems no longer available" do
+      bundle "config --local without development"
+
+      path = bundled_app(File.join("vendor", "foo"))
+      build_lib "foo", :path => path
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem "activesupport", "2.3.2"
+        gem 'foo', :path => 'vendor/foo', :group => :development
+      G
+
+      FileUtils.rm_rf(path)
+
+      ruby "require 'bundler'; Bundler.setup", :env => { "DEBUG" => "1" }
+      expect(out).to include("Assuming that source at `vendor/foo` has not changed since fetching its specs errored")
+      expect(out).to include("Found no changes, using resolution from the lockfile")
+      expect(err).to be_empty
+    end
+
     it "remembers --without and does not include groups passed to Bundler.setup" do
       bundle "config --local without rails"
       install_gemfile <<-G


fzakaria:
Wow that was the fix !  So small :+1: 
I tried to introspect bundler myself but to be honest the code was a bit confusing for me to navigate.

I'm on paternity now; but I will try to test out the fix by EOW :)

deivid-rodriguez:
I know, bundler's code gets superconfusing. Your investigation in the original ticket was very helpful and on point, and made it much easier to figure this out. Still let's wait for the CI and your tests, hopefully this is correct.

deivid-rodriguez:
> I'm on paternity now;

Congratulations, by the way :smile:

fzakaria:
thanks! 
Yes I try when possible at least to give as much information.
Glad it helped; 

Our codebase is pretty old & large so we do quite a lot of "interesting" Ruby/JRuby/Bundler/Warbler patterns.

deivid-rodriguez:
@fzakaria Did you find time to try this? No rush at all, I was just wondering, since I looked for the meaning of the EOW acronym and it means End of Week :smile:

fzakaria:
Yea ; sorry for the late reply! the baby is making things tough to follow-through on :+1: 

I tested it in fact with my minimal reproduction project. 
https://github.com/fzakaria/warbler-bundler-jruby-failure

Here are the steps:
1. Confirm failure; using **bundler 2.0.2** (separate issue affecting 2.1.4)

```ruby
ruby '2.5.7', engine: 'jruby', engine_version: '9.2.13.0'
source 'https://rubygems.org'

group :development, :test do
   gem 'demo', :path => "./ruby-gems/demo"
end
gem 'warbler', '2.0.5'
gem 'jruby-jars', '9.2.13.0'
```

```
$ bundle exec warble
$ java -jar warbler-bundler-jruby-failure.jar

The path `uri:classloader://warbler-bundler-jruby-failure/ruby-gems/demo` does not exist.
```

2. Checkout your pull-request and build the gem
```
# remove all previous bundlers
gem uninstall bundler
gem build bundler.gemspec`
gem install bundler-2.2.0.rc.2.gem 
```

3. Profit

```
$ bundle exec warble
$ java -jar warbler-bundler-jruby-failure.jar

Starting the JAR!
```

PS: is there a more fancier way to test developing gems ? Python has the ability to perform a "dev install" which symlinks to the directory rather than building a full gem.

deivid-rodriguez:
Great, thanks so much for double checking even if it's hard now with the baby! ;)

> PS: is there a more fancier way to test developing gems ? Python has the ability to perform a "dev install" which symlinks to the directory rather than building a full gem.

With other gems, you would normally use the `path:` option in the `Gemfile` and don't need to do anything else, but for bundler itself, we need to do extra magic like what you did. I normally run `bin/rake install`, which builds and installs in one step.

deivid-rodriguez:
Merging since @fzakaria has confirmed this works and the PR received no other feedback.

Issue #3757, opened 2020-06-24
`bundle clean --force` removes default gem executables


deivid-rodriguez:
Thank you for contributing to the [rubygems](https://github.com/rubygems/rubygems) repository, and specifically to the [Bundler](https://bundler.io/) gem.

### Please describe the problem as clearly as you can

We have a `bundle clean --force` command that removes every gem not in the current bundle from your system. This command goes too far and removes executables from default gems.

### Please post system independent reproduction steps

On any application, and on ruby 2.7, run

```
$ irb --version
irb 1.2.3 (2020-02-15)
$ bundle clean --force
Removing irb (1.2.3)
$ irb --version
Traceback (most recent call last):
	1: from /home/deivid/.rbenv/versions/2.7.1/bin/irb:23:in `<main>'
/home/deivid/.rbenv/versions/2.7.1/bin/irb:23:in `load': cannot load such file -- /home/deivid/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/irb-1.2.3/exe/irb (LoadError)
```

### What command did you run?

    bundle clean --force

### What were you expecting to happen?

All gems are removed, but not executables from default gems, because default gems should be always available.

### What actually happened?

IRB is gone.

NOTE: I experimented with a better issue report template in this issue, will open a PR for it :)

rubyFeedback:
Hehe the template is quite funny since it looks as if you are almost talking to yourself there,
in particular with the "Thank you for contributing to the rubygems repository" intro. I mean
you are not exactly a newcomer to rubygems ...  ;)

deivid-rodriguez:
Hehehe, yeah.

In https://github.com/rubygems/rubygems/pull/3784 I put that appreciation message inside `<!-- ... -->`, so that the OP gets the appreciation, but it doesn't show up by default in the reported issue.

PR #3906, opened 2020-08-26
Make `--default` and `--install-dir` options to `gem install` play nice together

deivid-rodriguez:
# Description:

These options are probably not really used in real life together, but I happened to need them while writing a spec for a bundler issue fix. In particular, for https://github.com/rubygems/rubygems/issues/3757.

## What was the end-user or developer problem that led to this PR?d

The problem is that when `--default` and `--install-dir` options are given to `gem install`, the default specification file should be installed to the `specifications/default` folder in `install-dir`, not to the one in the default gem directory.

The fix seemed really simple, but due to the cumbersome setup of rubygems tests, it broke a lot of tests, so I had to refactor rubygems test setup to play nice with the change. More information is given in the individual commits.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -433,7 +433,7 @@ class Gem::Installer
   #
 
   def default_spec_file
-    File.join Gem.default_specifications_dir, "#{spec.full_name}.gemspec"
+    File.join gem_home, "specifications", "default", "#{spec.full_name}.gemspec"
   end
 
   ##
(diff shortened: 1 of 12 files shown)


hsbt:
Basically it seems good. I will check the behavior changes of the default gems installation.

deivid-rodriguez:
@hsbt Did you have time to look at this? I don't think it should affect any realworld behavior since it's a very niche combination of flags. In any case, if it would affect I believe it's too fix a bug so it's fine.

But great to double check of course :+1:.

hsbt:
Sorry, I'm working on RubyKaigi takeout 2020 as stuff in this week. I'll review this after that 🙇

deivid-rodriguez:
Sure, thanks! And keep up the good work :muscle:.

deivid-rodriguez:
@hsbt friendly ping :)

hsbt (approved):

deivid-rodriguez:
Thanks so much @hsbt!

PR #3908, opened 2020-08-26
Add CHANGELOG.md to newgem generator

kieranklaassen:
This feature adds a `CHANGELOG.md` file when running `bundle gem mygem` in addition to all other gem skeleton files.

I keep adding it myself and it is best practice to keep a changelog. Let me know if there is any reason this was not there yet or if there is anything I can improve. Thanks!

### DIFF
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -4,6 +4,7 @@ RSpec.describe "bundle gem" do
   def gem_skeleton_assertions
     expect(bundled_app("#{gem_name}/#{gem_name}.gemspec")).to exist
     expect(bundled_app("#{gem_name}/README.md")).to exist
+    expect(bundled_app("#{gem_name}/CHANGELOG.md")).to exist
     expect(bundled_app("#{gem_name}/Gemfile")).to exist
     expect(bundled_app("#{gem_name}/Rakefile")).to exist
     expect(bundled_app("#{gem_name}/lib/#{require_path}.rb")).to exist
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -4,7 +4,6 @@ RSpec.describe "bundle gem" do
   def gem_skeleton_assertions
     expect(bundled_app("#{gem_name}/#{gem_name}.gemspec")).to exist
     expect(bundled_app("#{gem_name}/README.md")).to exist
-    expect(bundled_app("#{gem_name}/CHANGELOG.md")).to exist
     expect(bundled_app("#{gem_name}/Gemfile")).to exist
     expect(bundled_app("#{gem_name}/Rakefile")).to exist
     expect(bundled_app("#{gem_name}/lib/#{require_path}.rb")).to exist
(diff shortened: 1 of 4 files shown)


simi:
This is really good addition. The only question I do have here is do we need also corresponding CLI flags `--changelog` and `--no-changelog`?

kieranklaassen:
@simi Yeah, I was thinking the same. My personal decision was that everyone should have a CHANGELOG. It is like the README. But I'm happy to add the flags if others want to weigh in on this decision. Cheers!

indirect:
Personally I would vote for no option, and if you reallllly don't want a changelog you can delete the file after generating. 😄

deivid-rodriguez (approved):
:clap: :clap:

hsbt:
Can we change it with flag option?

I understood this is useful then nothing. But I have some of project to use GitHub releases instead of `CHANGELOG`.  When it's case, the default behavior is a little noisy for me.

kieranklaassen:
@hsbt Would it be sufficient to delete the CHANGELOG.md after generating the new Gem like @indirect is suggesting? Or would you prefer to have only the `--no-changelog` available if you are going to use any other way to keep track of your changelog? 

If you think we should add the flag I'm happy to do add it. I would just make sure all people are nudged towards keeping a `CHANGELOG.md` since that is the most universal way to keep track.

hsbt:
It's good to make it optional like https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/cli/gem.rb#L131

kieranklaassen:
@hsbt I committed your requested changes.

simi (approved):

hsbt:
@kieranklaassen I'm not sure why Windows CI was failed with this PR. I re-run them twice. But they are failed by same failure.

deivid-rodriguez:
You need to add `"BUNDLE_GEM__CHANGELOG" => false` at the main before hook in the file to ensure that readline doesn't come into play in those tests, since it still seems to give some issues on Windows.

deivid-rodriguez:
Also, can you add a "on first run asks about CHANGELOG" spec, similar to the ones already present for CoC, CI service, and so on?

bundler/spec/commands/newgem_spec.rb
```diff
@@ -156,6 +156,26 @@ def bundle_exec_rubocop
     end
   end
 
+  shared_examples_for "--changelog flag" do
+    before do
+      bundle! "gem #{gem_name} --changelog"
+    end
+    it "generates a gem skeleton with CHANGELOG license" do
```
colby-swandale, line 163:
```suggestion
    it "generates a gem skeleton with a CHANGELOG" do
```

bundler/spec/commands/newgem_spec.rb
```diff
@@ -156,6 +156,26 @@ def bundle_exec_rubocop
...
+      gem_skeleton_assertions
+      expect(bundled_app("#{gem_name}/CHANGELOG.md")).to exist
+    end
+  end
+
+  shared_examples_for "--no-changelog flag" do
+    before do
+      bundle! "gem #{gem_name} --no-changelog"
+    end
+    it "generates a gem skeleton without CHANGELOG license" do
```
colby-swandale, line 173:
```suggestion
    it "generates a gem skeleton without a CHANGELOG" do
```

colby-swandale (commented):
Looks good, just a couple of small things from me.

kieranklaassen:
Got a bit busy last week. I'll get the missing tests in like requested and fi them for the Windows builds.

deivid-rodriguez:
@kieranklaassen Are you still planning to update this?

kieranklaassen:
@deivid-rodriguez Yes! I'm swamped at work but it's on my radar! Sorry, for the delay.

deivid-rodriguez:
No problem, thanks for the update!

deivid-rodriguez:
This seems like a good opportunity for a newcomer to be able to help out, so I'll tag this as "good first issue" in case somebody is interested in wrapping it up :)

deivid-rodriguez:
Hi @kieranklaassen, @jlgoedert has kindly volunteered to get this over the line, so I'll close this PR in favor of the new one. Of course, your attribution will be kept. Thanks a lot for your contribution!

PR #3909, opened 2020-08-27
Fix platform issues when running under a frozen bundle

### LINKED ISSUE #3874: Bundler local filters out all "ruby" platform gems on JRuby
It appears that Bundler is filtering out all gems that don't have a platform of "java" when trying to activate a local/vendored bundle.

The original issue reported by @eregon is here: https://github.com/jruby/jruby/issues/6349

I dug into this and discovered that `expand_dependencies` is filtering out all dependencies in the "ruby" platform if the `generic_local_platform` is not also "ruby". This means that only "java" platform gems get included, which clearly won't work.

https://github.com/jruby/jruby/issues/6349#issuecomment-668149352

I do not know why it breaks in this situation and not when running a bundled application normally. I suspect it has something to do with how local bundling limits gem visibility to only the gems vendored into the application.

The "ruby" platform is both a marker for C extensions and a default or fallback platform. Removing all "ruby" gems if the platform differs is not correct logic; as with normal gem resolution, a platform-specific gem should be used if available, and the default "ruby" platform gem should be used otherwise.

This should affect any implementation that ends up with a `generic_local_platform` that is not simply "ruby" but I do not know of any implementation that both supports Bundler and also has this different platform.

deivid-rodriguez:

In the case of running under jruby, on a frozen bundle, and against a lockfile only including the ruby platform, `expand_dependencies` would end up filtering out all dependencies, causing errors.

The reason is that on a frozen bundle, the local platform is not added to the array of definition platforms, since only platform information from the lockfile should be used. That means that under the previous logic, `expand_dependencies` would filter out everything.

The solution is to make sure to always _at least_ expand all dependencies with the local platform.

Note that it should still be preferred to "unfreeze" the application, and rebundle using jruby, so that platform specific dependencies are picked up. But if the desired to use only pure ruby versions of dependencies, this changes gets that case working too without weird errors.

Fixes #3874.

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -204,10 +204,6 @@ module Bundler
       @locked_specs - specs
     end
 
-    def new_platform?
-      @new_platform
-    end
-
     def missing_specs
       missing = []
       resolve.materialize(requested_dependencies, missing)
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -553,10 +553,9 @@ module Bundler
     end
 
     def current_platforms
-      current_platform = Bundler.local_platform
       [].tap do |platforms|
-        platforms << current_platform if Bundler.feature_flag.specific_platform?
-        platforms << generic(current_platform)
+        platforms << local_platform if Bundler.feature_flag.specific_platform?
+        platforms << generic_local_platform
       end
     end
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I'll merge this. @headius if you have time to verify the fix, try out the default branch and we can address any issues post-merge.

PR #3910, opened 2020-08-28
Actions daily workflows - only run in rubygems account

MSP-Greg:
# Description:

Support contributors by shutting off daily CI in forks.

- [ ] Write tests

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   daily_bundler:
     runs-on: ubuntu-18.04
+    if: github.repository == "rubygems/rubygems"
     strategy:
       matrix:
         ruby: [ ruby-head ]
--- a/.github/workflows/daily-rubygems.yml
+++ b/.github/workflows/daily-rubygems.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   daily_rubygems:
     runs-on: ubuntu-18.04
+    if: github.repository == "rubygems/rubygems"
     strategy:
       matrix:
         ruby: [ ruby-head, truffleruby-head ]
--- a/.github/workflows/monthly-bundler.yml
+++ b/.github/workflows/monthly-bundler.yml
@@ -8,6 +8,7 @@ jobs:
   monthly_bundler:
     name: Refresh month of man pages
     runs-on: ubuntu-18.04
+    if: github.repository == "rubygems/rubygems"
     steps:
       - uses: actions/checkout@v2
 


simi:
Same daily fail is in my fork as well. Is that problem with CI job? I see it pass in here.

deivid-rodriguez:
Aha! I now understand the why of this request, I thought it was a contribution to help fight climate change :laughing:. It'd be interesting to understand why the daily bundler job fails on forks.

deivid-rodriguez:
If I understand correctly, this only started happening very recently since nobody noticed this until now. Right?

deivid-rodriguez:
Actually, this is not related to forks, but a real problem with HEAD versions in the canonical repo, since our daily workflow seems broken in the same way (and also with our daily notifications, since we didn't get notified about it).

So thanks for reporting this!

simi:
22. 8. 2:20 was the first fail in my fork

deivid-rodriguez:
Culprit is most likely https://github.com/ruby/ruby/commit/8a40e9b5a20314a509c8caaadce19a0b52680419.

deivid-rodriguez:
The `set` library seems pure ruby so the easy fix is to vendor it. However, there's no rubygems.org release nor github repository. 

@hsbt Could you clarify what the situation regarding `set` as a default gem is? Also, could talk a bit about your criteria for promoting standard library code to default gems? When default gem promotion affects code heavily used by `bundler`, we need to do extra work to support it. Otherwise there's no benefit from the promotion, since `bundler` will load the version that comes with ruby and applications won't be able to upgrade.

deivid-rodriguez:
Reported the issue at https://github.com/rubygems/rubygems/issues/3912 to reduce the noise on this PR.

hsbt:
> @hsbt Could you clarify what the situation regarding `set` as a default gem is?

Because It's still preparation status.

.github/workflows/daily-bundler.yml
```diff
@@ -7,6 +7,7 @@ on:
 jobs:
   daily_bundler:
     runs-on: ubuntu-18.04
+    if: github.repository == "rubygems/rubygems"
```
eregon, line 10: Can you also do this for `monthly-bundler.yml` ?

eregon (approved):
:+1: from me

.github/workflows/daily-bundler.yml
MSP-Greg: Fixed.  Sorry, I missed the monthly...

MSP-Greg:
@deivid-rodriguez

> a contribution to help fight climate change

I didn't mention that.  Have you got any extra lights on?  How about that thermostat setting?

> since nobody noticed this until now

I normally turn actions off for repos with cron jobs.  Kind if a PITA.

deivid-rodriguez (approved):

hsbt:
👍

Issue #3912, opened 2020-08-28
Handle promotion of `set` to a default gem


deivid-rodriguez:
### Describe the problem as clearly as you can

Set was promoted to a default gem at https://github.com/ruby/ruby/commit/8a40e9b5a20314a509c8caaadce19a0b52680419. Since `bundler` uses `set` quite a lot, it will now load the default version of `set` very early and applications won't be able to choose any version of `set` via the Gemfile. We should fix this issue because otherwise the promotion is quite pointless. Potential fixes:

* Vendor set. It should be ok since it seems pure ruby, but there's no github repo nor rubygems release at the moment.

* Revert the promotion.

### Post system independent steps to reproduce the problem

See https://github.com/rubygems/rubygems/actions?query=workflow%3Adaily-bundler.

deivid-rodriguez:
As per https://github.com/rubygems/rubygems/pull/3910#issuecomment-682433984, should we go with vendoring `set` once the github repository is available, then?

hsbt:
I will make the repository to public before the final release of Ruby 3.0.

deivid-rodriguez:
@hsbt I created #4297 to fix this by removing our dependency on `set`.

PR #3913, opened 2020-08-28
Add `set` as a non supported default gem

deivid-rodriguez:
# Description:

As per #3912, I'm adding `set` as an exemption to the specs as a non supported default gem, until we can handle its promotion.

This is to get the daily bundler workflow against ruby-head back to green.

# Tasks:

- [ ] Write tests


hsbt (approved):

colby-swandale (approved):

deivid-rodriguez:
Thanks both!

Issue #3598, opened 2020-05-06
bundle package --all fails if BUNDLE_CACHE_ALL is set to true in .bundle/config


jesseadams:
1. Do you have a reproduction script? It would be really helpful for us if you provide one.

With the following barebones Gemfile, or any Gemfile:
```
source 'https://rubygems.org'
```

Run:

```
$ bundle package --all
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
Updating files in vendor/cache
$ cat .bundle/config
---
BUNDLE_CACHE_ALL: "true"
$ bundle package --all
Unknown switches "--all"
```

2. What are you trying to accomplish?

    Your answer: Trying to run `bundle package --all` consistently in an automated deployment pipeline.

3. What command did you run?

    Your answer: `bundle package --all` multiple times during various stages of building a docker image.

4. What were you expecting to happen?

    Your answer: Bundler to correctly and consistently package the gems

5. What actually happened?

    Your answer: Bundler correctly packaged the gems the first time and them automatically put `BUNDLE_CACHE_ALL: "true"` into .bundle/config. The second time running the same command generates the following error: `Unknown switches "--all"`

6. Is there an exception backtrace? If so, please copy it below.

    Your answer: No backtrace available

7. Please run `bundle env` and paste the output below:

On my local machine:

## Environment

```
Bundler       2.1.4
  Platforms   ruby, x86_64-darwin-18
Ruby          2.7.1p83 (2020-03-31 revision a0c7c23c9cec0d0ffcba012279cd652d28ad5bf3) [x86_64-darwin18]
  Full Path   /usr/local/opt/ruby/bin/ruby
  Config Dir  /usr/local/Cellar/ruby/2.7.1/etc
RubyGems      3.1.2
  Gem Home    /usr/local/lib/ruby/gems/2.7.0
  Gem Path    /Users/jradams/.gem/ruby/2.7.0:/usr/local/lib/ruby/gems/2.7.0:/usr/local/Cellar/ruby/2.7.1/lib/ruby/gems/2.7.0
  User Home   /Users/jradams
  User Path   /Users/jradams/.gem/ruby/2.7.0
  Bin Dir     /usr/local/lib/ruby/gems/2.7.0/bin
Tools
  Git         2.25.1
  RVM         not installed
  rbenv       rbenv 1.1.2
[...]

PR #3914, opened 2020-08-28
Restore `bundle cache --all` in all cases

deivid-rodriguez:
# Description:

The `bundle cache --all` flag has been going back and forth recently, since I'm not too clear what our plan on it is.

The current situation, where `bundle cache --all` only works the first time it's run is definitely not what we want.

My proposal is to restore the flag, but deprecate it. The rationale is the same for all the other flags that are silently rememebered. Instead, recommend users to use configuration, and set `bundle config cache_all true` if they want git and path gems to also be cached.

The change in the default value of `cache_all` set for bundler 3 (where path and git gems will be cached by default) is kept.

Fixes https://github.com/rubygems/rubygems/issues/3598.
______________

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -440,10 +440,9 @@ module Bundler
     end
 
     desc "cache [OPTIONS]", "Locks and then caches all of the gems into vendor/cache"
-    unless Bundler.feature_flag.cache_all?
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I'll split the deprecation to a separate PR, since the bug fix and the deprecation are different tasks, and should appear in different changelog sections.

Issue #3916, opened 2020-08-30
bundle config key value is --global by default


eregon:
### Describe the problem as clearly as you can

`bundle config key value` is `--global` by default (stored in `~/.bundle/config`), while I think most usages would benefit from being local.

For instance, typical mistakes are `bundle config path vendor/bundle`, `bundle config without db`, etc.

Which unfortunately is also what is shown by deprecation messages, and I believe that part is an outright bug (Bundler 2.1.4) because the old flags did write to `.bundle/config` (= `--local`):
```
$ bundle install --path vendor/bundle
[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path 'vendor/bundle'`, and stop using this flag

$ bundle install --without db
[DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'db'`, and stop using this flag
```

Note also that the very similar command `git config key value` defaults to `--local`.

I think it would be much safer and intuitive to change `bundle config key value` to also default to `--local`.
Most of Bundler configuration is project-specific and doing it globally makes often little sense, especially when not stating explicitly one wants it to be global.

deivid-rodriguez:
I agree, we should fix the deprecation messages and consider changing `bundle config` default to be local :+1:.

eregon:
PR to fix the deprecation messages and always show `--local`/`--global` explicitly for `bundle config set` commands: https://github.com/rubygems/rubygems/pull/3917

eregon:
cc @chrisseaton since you ran into this bug (but I'm sure you're not the only one, this exists since Bundler 2 IIRC).

eregon:
Could we reopen to track changing the default? Or should I file a new issue?

deivid-rodriguez:
I prefer a new ticket referencing this one.

eregon:
OK, https://github.com/rubygems/rubygems/issues/3926

deivid-rodriguez:
Thanks, I also edited the title of this ticket to reflect what we fixed in the PR that closed this.

PR #3917, opened 2020-08-30
Fix deprecation messages for `bundle install` flags, the config should be --local as before

eregon:
* Closes https://github.com/rubygems/rubygems/issues/3916.
* Always show --local or --global for `bundle config set` commands as
  the default is global which can be not intuitive (#3916).

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/fetcher.rb
+++ b/bundler/lib/bundler/fetcher.rb
@@ -47,7 +47,7 @@ module Bundler
         remote_uri = filter_uri(remote_uri)
         super "Authentication is required for #{remote_uri}.\n" \
           "Please supply credentials for this source. You can do this by running:\n" \
-          " bundle config set #{remote_uri} username:password"
+          " bundle config set --global #{remote_uri} username:password"
       end
     end
     # This error is raised if HTTP authentication is provided, but incorrect.
--- a/bundler/man/bundle-config.1.txt
+++ b/bundler/man/bundle-config.1.txt
@@ -58,8 +58,8 @@ DESCRIPTION
        Executing bundle with the BUNDLE_IGNORE_CONFIG environment variable set
        will cause it to ignore all configuration.
 
-       Executing bundle  config  set  disable_multisource  true  upgrades  the
-       warning  about  the  Gemfile  containing multiple primary sources to an
+       Executing bundle config set --local disable_multisource  true  upgrades
+       the warning about the Gemfile containing multiple primary sources to an
        error. Executing bundle  config  unset  disable_multisource  downgrades
        this error to a warning.
 
@@ -70,7 +70,8 @@ REMEMBERING OPTIONS
 
        However,  this will be changed in bundler 3, so it's better not to rely
        on this behavior. If these options must be remembered, it's  better  to
-       set them using bundle config (e.g., bundle config set path foo).
+       set  them  using  bundle  config  (e.g., bundle config set --local path
+       foo).
 
        The options that can be configured are:
 
@@ -120,7 +121,7 @@ BUILD OPTIONS
 
 
 
-           bundle config set build.mysql --with-mysql-config=/usr/local/mysql/bin/mysql_config
+           bundle config set --global build.mysql --with-mysql-config=/usr/local/mysql/bin/mysql_config
 
 
 
@@ -376,7 +377,7 @@ LOCAL GIT REPOS
 
 
 
-           bundle config set local.GEM_NAME /path/to/local/git/repository
+           bundle config set --local local.GEM_NAME /path/to/local/git/repository
 
 
 
@@ -385,7 +386,7 @@ LOCAL GIT REPOS
 
 
 
-           bundle config set local.rack ~/Work/git/rack
+           bundle config set --local local.rack ~/Work/git/rack
 
 
 
@@ -418,7 +419,7 @@ MIRRORS OF GEM SOURCES
 
 
 
-           bundle config set mirror.SOURCE_URL MIRROR_URL
+           bundle config set --global mirror.SOURCE_URL MIRROR_URL
 
 
 
@@ -427,7 +428,7 @@ MIRRORS OF GEM SOURCES
 
 
 
-           bundle config set mirror.http://rubygems.org http://rubygems-mirror.org
+           bundle config set --global mirror.http://rubygems.org http://rubygems-mirror.org
 
 
 
@@ -437,7 +438,7 @@ MIRRORS OF GEM SOURCES
 
 
 
-           bundle config set mirror.SOURCE_URL.fallback_timeout TIMEOUT
+           bundle config set --global mirror.SOURCE_URL.fallback_timeout TIMEOUT
 
 
 
@@ -445,7 +446,7 @@ MIRRORS OF GEM SOURCES
 
 
 
-           bundle config set mirror.https://rubygems.org.fallback_timeout 3
+           bundle config set --global mirror.https://rubygems.org.fallback_timeout 3
 
 
 
@@ -458,7 +459,7 @@ CREDENTIALS FOR GEM SOURCES
 
 
 
-           bundle config set SOURCE_HOSTNAME USERNAME:PASSWORD
+           bundle config set --global SOURCE_HOSTNAME USERNAME:PASSWORD
 
 
 
@@ -467,7 +468,7 @@ CREDENTIALS FOR GEM SOURCES
 
 
 
-           bundle config set gems.longerous.com claudette:s00pers3krit
+           bundle config set --global gems.longerous.com claudette:s00pers3krit
 
 
 
@@ -484,7 +485,7 @@ CREDENTIALS FOR GEM SOURCES
 
 
 
-           bundle config set https://github.com/bundler/bundler.git username:password
+           bundle config set --global https://github.com/bundler/bundler.git username:password
 
 
 
--- a/bundler/man/bundle-config.ronn
+++ b/bundler/man/bundle-config.ronn
@@ -47,7 +47,7 @@ configuration only from the local application.
 Executing bundle with the `BUNDLE_IGNORE_CONFIG` environment variable set will
 cause it to ignore all configuration.
 
-Executing `bundle config set disable_multisource true` upgrades the warning about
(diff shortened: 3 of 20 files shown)


bundler/spec/commands/outdated_spec.rb
```diff
@@ -555,15 +555,15 @@ def test_group_option(group)
...
+  context "after bundle config set --local deployment true" do
     before do
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo2)}"
 
         gem "rack"
         gem "foo"
       G
-      bundle "config set deployment true"
+      bundle "config --local deployment true"
```
deivid-rodriguez, line 566: Could you keep `set`?

bundler/lib/bundler/fetcher.rb
```diff
@@ -47,7 +47,7 @@ def initialize(remote_uri)
         remote_uri = filter_uri(remote_uri)
         super "Authentication is required for #{remote_uri}.\n" \
           "Please supply credentials for this source. You can do this by running:\n" \
-          " bundle config set #{remote_uri} username:password"
+          " bundle config set --global #{remote_uri} username:password"
```
deivid-rodriguez, line 50: In the other places, you chose `--local` for configuring credentials for a source. Should we use that here too?

bundler/man/bundle-config.ronn
```diff
@@ -333,21 +333,21 @@ Bundler supports overriding gem sources with mirrors. This allows you to
 configure rubygems.org as the gem source in your Gemfile while still using your
 mirror to fetch gems.
 
-    bundle config set mirror.SOURCE_URL MIRROR_URL
+    bundle config set --local mirror.SOURCE_URL MIRROR_URL
```
eregon, line 336: Not sure if better to show `--local` or `--global` for mirror and authentication (just below).
But at least it seems better to show one of them, so users can consider whether they want it local or global and not rely on the default which might not do the right thing.

deivid-rodriguez (commented):
Looks good! Being explicit is a good step forward for now, and makes it easier to change the default in the future.

bundler/man/bundle-config.1.txt
```diff
@@ -70,43 +70,44 @@ REMEMBERING OPTIONS
...
 
-       bin    Creates   a   directory   (defaults  to  ~/bin)  and  place  any
+       bin    Creates  a  directory  (defaults  to  ~/bin)   and   place   any
               executables  from  the  gem  there.  These  executables  run  in
               Bundler's context. If used, you might add this directory to your
-              environment's PATH variable. For  instance,  if  the  rails  gem
+              environment's  PATH  variable.  For  instance,  if the rails gem
               comes with a rails executable, this flag will create a bin/rails
-              executable that ensures that all referred dependencies  will  be
+              executable  that  ensures that all referred dependencies will be
```
eregon, line 83: My `groff` seems to generate unnecessary changes like here, could someone run `rake man:build` with a version that doesn't cause so many changes? Then I'll pick it up in this PR.

bundler/man/bundle-config.ronn
deivid-rodriguez: Agreed. Since we are unsure, let's go with `--global` I guess to match current "flagless" behaviour?

bundler/lib/bundler/fetcher.rb
eregon: Done, but might change again for https://github.com/rubygems/rubygems/pull/3917#discussion_r479780099

bundler/man/bundle-config.1.txt
deivid-rodriguez: Yes, I hate it :angry:. My `groff` version does that too. I think we might want to pass a ruby script over the result of `groff` (instead of just `col -xb`) to eliminate these whitespace differences.

bundler/lib/bundler/fetcher.rb
deivid-rodriguez: Yeah, I saw your other comment after I posted this.

bundler/man/bundle-config.ronn
eregon: OK, switched to `--global` for mirror & auth config commands.

bundler/spec/commands/outdated_spec.rb
eregon: Done

eregon:
@deivid-rodriguez I think I addressed everything now, and I added a CHANGELOG entry, should be good to go, except that regeneration of `bundle-config.1.txt` (unless we're fine with it?).

bundler/man/bundle-config.1.txt
eregon: I made https://github.com/eregon/rubygems/commit/ad68c24e09e812b906dbe4a1542ac33f6d7482e8 that fixes the extra `1m` and `22m` in the diff:
```diff
-NAME
-       bundle-add - Add gem to the Gemfile and run bundle install
+1mNAME0m
+       1mbundle-add 22m- Add gem to the Gemfile and run bundle install
-SYNOPSIS
-       bundle     add     GEM_NAME     [--group=GROUP]     [--version=VERSION]
+1mSYNOPSIS0m
+       1mbundle     add     4m22mGEM_NAME24m     [--group=GROUP]     [--version=VERSION]
        [--source=SOURCE]   [--git=GIT]   [--branch=BRANCH]    [--skip-install]
```
but it does not fix those whitespace changes. So I wonder if `groff` somehow forgot to strip those escape sequences and that causes the extra spacing changes.

eregon:
FWIW I like `bundle config --local path vendor/bundle` better than `bundle config set --local path vendor/bundle` but given that seems deprecated for Bundler 3 I went with the Bundler 3 syntax and explicit `set` (which is another difference to `git config`).
(Specifically if we change the default to `--local`, `bundle config path vendor/bundle` sounds nicer than `bundle config set path vendor/bundle` to me)

bundler/man/bundle-config.1.txt
eregon: BTW, what are the `.txt` for? Isn't the `.ronn` enough for "readable without `man`"?

deivid-rodriguez: > but it does not fix those whitespace changes. So I wonder if groff somehow forgot to strip those escape sequences and that causes the extra spacing changes.

I think the extra spacing is generated due to `ronn` generating TABs at random places when you make changes to a ronn source. But not really sure.

> BTW, what are the .txt for? Isn't the .ronn enough for "readable without man"?

That's a great point. We can probably print the `ronn` sources when `man` is not available and stop generating these txt files.

deivid-rodriguez:
> @deivid-rodriguez I think I addressed everything now, and I added a CHANGELOG entry, should be good to go, except that regeneration of bundle-config.1.txt (unless we're fine with it?).

You can skip the changelog message since I plan to start generating the changelog automatically from merge PR titles and labels in the next version.

> FWIW I like `bundle config --local path vendor/bundle` better than `bundle config set --local path vendor/bundle` but given that seems deprecated for Bundler 3 I went with the Bundler 3 syntax and explicit set (which is another difference to `git config`).
(Specifically if we change the default to `--local`, `bundle config path vendor/bundle` sounds nicer than `bundle config set path vendor/bundle` to me)

I'm also unsure about deprecating the old `bundle config` interface, that's why I delayed it until it can be better discussed, and the original reasons properly reviewed: https://github.com/rubygems/bundler/pull/7475. But we're using the new interface internally in a consistent way at the moment, so let's keep doing it for the moment :+1:.

deivid-rodriguez:
Also, I think we can ignore the annoying txt diff for this PR, and try to get rid of txt files at all as you suggested.

eregon:
> You can skip the changelog message since I plan to start generating the changelog automatically from merge PR titles and labels in the next version.

Should I remove that change then?

> I'm also unsure about deprecating the old `bundle config` interface, that's why I delayed it until it can be better discussed, and the original reasons properly reviewed: [rubygems/bundler#7475](https://github.com/rubygems/bundler/pull/7475). But we're using the new interface internally in a consistent way at the moment, so let's keep doing it for the moment .

Is there some good place to discuss that? I feel this deprecation is going to be rather annoying (and require verbose commands) for very little gains. The only conflict would be if one needs to set a key named `list/get/set/unset` but that seems extremely unlikely.

> Also, I think we can ignore the annoying txt diff for this PR, and try to get rid of txt files at all as you suggested.

Will you make a PR or is there a plan for that then? :smiley: 
I guess we should wait to merge this PR after that?

I think this PR should be merged for the next Bundler 2 release, it's quite bad if people configure globally when they do not intend to (can cause confusion, or an impression that the deprecation messages are just broken and due to that might be ignored more).

deivid-rodriguez:
> Should I remove that change then?

It will be overwritten unless the PR title matches the current wording exactly, so yeah I'd rather remove it!

> Is there some good place to discuss that? I feel this deprecation is going to be rather annoying (and require verbose commands) for very little gains. The only conflict would be if one needs to set a key named list/get/set/unset but that seems extremely unlikely.

Feel free to open a new ticket, investigate why the new interface was added, and explain why you think the reasons for making it compulsory are wrong.

> Will you make a PR or is there a plan for that then? :smiley:
> I guess we should wait to merge this PR after that?

I like your idea, but it came up literally 1 hour ago, so I'm not sure when I (or anybody) will get to it. Better not to block this PR on that.

> I think this PR should be merged for the next Bundler 2 release, it's quite bad if people configure globally when they do not intend to (can cause confusion, or an impression that the deprecation messages are just broken and due to that might be ignored more).

Sure, just be patient! :)
Sa little
This is straightforward and will make it to the next release, just nbe

bundler/man/bundle-config.1.txt
eregon: I ended up using `git add -p`, and looking at which lines were relevant with `git diff -w`. Quite bothersome but it worked.

eregon:
OK, CHANGELOG entry removed. I worked around and now the diff for `bundle-config.1.txt` should be minimal.
Ready to be merged :)

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks!

PR #3923, opened 2020-09-01
Fix issues when syncronizing man pages

deivid-rodriguez:
We had an issue where every month our CI would start failing because the check that makes sure that our man pages are in sync with their sources would start failing.

We fixed it by automatically updating the month at the begining of each month. However, we changed permissions in the default branch, and the task started failing.

Here I propose an alternative solution, which is ignoring the month for the man pages syncronization check.

The month is now no longer updated every month, but instead only when man pages are explicitly rebuilt.

# Tasks:

- [ ] Write tests


deivid-rodriguez:
This PR implements the second proposal I made in https://github.com/rubygems/rubygems/issues/3582, since the 3rd turned out to be error prone and problematic.

PR #3924, opened 2020-09-02
The bug report template should be printed to standard error

deivid-rodriguez:

While writing a spec for another bug, I noticed that the bug report template is being printed to STDOUT. I believe that's unexpected.

So I changed it to be printed to STDERR.

I also added some optional color parameters to the UI helpers, so that I could keep the current output colors.

# Tasks:

- [ ] Write tests


colby-swandale (approved):

deivid-rodriguez:
Thanks @colby-swandale!

PR #3927, opened 2020-09-03
Move shebang to the top of `bin/console` template

speckins:
In an executable script, the shebang line should be the first line (the
file needs to start with the bytes 0x23 0x21).  Putting a comment above
it will break the script.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/bin/console.tt
+++ b/bundler/lib/bundler/templates/newgem/bin/console.tt
@@ -1,6 +1,5 @@
-# frozen_string_literal: true
-
 #!/usr/bin/env ruby
+# frozen_string_literal: true
 
 require "bundler/setup"
 require "<%= config[:namespaced_path] %>"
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (changes requested):
Thank you, nice catch!

As illustrated by the failing specs, you need to restore the `# frozen_string_literal: true` line as the second line in the file.

speckins:
I've made the requested changes.

deivid-rodriguez:
Thanks! Just one final request, could you add a couple of lines below this code here:

https://github.com/rubygems/rubygems/blob/2ceb821579d491219ce352b858120f7b5c0a467b/bundler/spec/commands/newgem_spec.rb#L377-L378

to assert that the first two characters in the generated binstub are "!#"?

That way we add some test coverage for this bug fix.

speckins:
I appended those expectations to that same test rather than creating a separate test; I hope that's okay.

deivid-rodriguez (approved):
Yes, it looks really good, thanks!

hsbt:
👍

Issue #3931, opened 2020-09-05
How can I override `path` to the default path (Gem.dir)?


ikngtty:
### Describe the problem as clearly as you can

My global config, `~/.bundle/config`, is:

```yaml
---
BUNDLE_PATH: "vendor/bundle"
```

but locally I want to cancel it and install it to the default path, "Gem.dir".

I set a local config, `./.bundle/config`, as:

```yaml
---
BUNDLE_PATH__SYSTEM: "true"
```

but it doesn't work. I try `bundle install` or `bundle update` and gems are installed to `./vendor/bundle`.

## Environment

```
Bundler       2.1.4
  Platforms   ruby, x86_64-darwin-19
Ruby          2.7.1p83 (2020-03-31 revision a0c7c23c9cec0d0ffcba012279cd652d28ad5bf3) [x86_64-darwin19]
  Full Path   /Users/ikngtty/.rbenv/versions/2.7.1/bin/ruby
  Config Dir  /Users/ikngtty/.rbenv/versions/2.7.1/etc
RubyGems      3.1.2
  Gem Home    /Users/ikngtty/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0
  Gem Path    /Users/ikngtty/.gem/ruby/2.7.0:/Users/ikngtty/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0
  User Home   /Users/ikngtty
  User Path   /Users/ikngtty/.gem/ruby/2.7.0
  Bin Dir     /Users/ikngtty/.rbenv/versions/2.7.1/bin
Tools         
  Git         2.28.0
  RVM         not installed
  rbenv       rbenv 1.1.2
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2020-09-05
Git SHA           unknown
Released Version  false
```

## Bundler settings

```
path
  Set for the current user (/Users/ikngtty/.bundle/config): "vendor/bundle"
path.system
  Set for your local app (/Users/ikngtty/Configurations/gem/.bundle/config): true
```

### Gemfile

```ruby
# frozen_string_literal: true

source 'https://rubygems.org'

git_source(:github) {|repo_name| "https://github.com/#{repo_name}" }

gem 'rails'
gem 'tw'
gem 'solargraph'
gem 'reek'
gem 'travis'
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (5.2.4.3)
[...]

PR #3933, opened 2020-09-07
Prioritize `path.system` over `path` when it makes sense

deivid-rodriguez:

If a user configures `path.system true` locally, and `path foo` globally, gems are still installed to `foo`, even if they should be installed to `GEM_PATH`, give that local configuration should have more priority.

My fix is to remove some special logic that was preventing this case to work, and also to relax the `path` and `path.system` validation that makes sure there are no conflicts between them to allow this case.

Fixes https://github.com/rubygems/rubygems/issues/3931.

# Tasks:

- [x] Write tests

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -195,14 +195,16 @@ module Bundler
 
     # for legacy reasons, in Bundler 2, we do not respect :disable_shared_gems
     def path
-      key  = key_for(:path)
-      path = ENV[key] || @global_config[key]
-      if path && !@temporary.key?(key) && !@local_config.key?(key)
-        return Path.new(path, false)
+      configs.each do |_level, settings|
+        path = value_for("path", settings)
+        path_system = value_for("path.system", settings)
+        disabled_shared_gems = value_for("disable_shared_gems", settings)
+        next if path.nil? && path_system.nil? && disabled_shared_gems.nil?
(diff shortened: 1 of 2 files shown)


colby-swandale:
Is this also a reactor? It's hard to spot what exactly the "fix" for the linked issue is.

deivid-rodriguez:
Yes, I had to introduce quite a bit of refactoring to figure this out. The fix is the last commit: https://github.com/rubygems/rubygems/pull/3933/commits/418998cab819b230352e1595d0fcfb318375f7fd (the one including a regression spec, the rest of commits should keep specs green, but they don't fix anything).

The fix is essentially, instead of resolving `path.system` and `path` independently, we should resolve them at the same time. By "resolving" I mean going through each level of configuration (local, env, global) and choosing the first "non empty" one. It

deivid-rodriguez:
@colby-swandale Do you plan on reviewing this? I was thinking of going ahead and merging if not.

I could split the refactoring to a separate changeset but since the refactoring is aimed at making the fix easier, and backporting the fix will require backporting the refactoring, I figured it was best to keep them together. I think a "commit by commit" review should work.

colby-swandale:
Sorry but I don't have the mental bandwidth to review at the moment.

deivid-rodriguez:
No problem. I'll merge for now, and we can iterate if any problems are found.

PR #3940, opened 2020-09-11
Avoid duplicated generation of APISpecification objects

mame:
# Description:

`gem install aws-sdk` is painfully slow

As far as I could see, `Gem::Resolver::APISpecification` objects are
supposed to be immutable.  If my guessing is correct, then we can cache
and reuse its instances for performance.
At least, `rake` passes on my machine.

Before this change:

```
$ time ruby -I lib bin/gem install --no-doc aws-sdk
Successfully installed aws-sdk-3.0.1
1 gem installed
real    0m37.104s
user    0m36.952s
sys     0m0.333s
```

After this change:

```
$ time ruby -I lib bin/gem install --no-doc aws-sdk
Successfully installed aws-sdk-3.0.1
1 gem installed
real    0m23.905s
user    0m23.740s
sys     0m0.365s
```

# Tasks:

- [ ] Write tests

### DIFF
--- a/lib/rubygems/resolver/api_specification.rb
+++ b/lib/rubygems/resolver/api_specification.rb
@@ -7,6 +7,17 @@
 
 class Gem::Resolver::APISpecification < Gem::Resolver::Specification
   ##
+  # We assume that all instances of this class are immutable;
+  # so avoid duplicated generation for performance.
+  @@cache = {}
+  def self.new(set, api_data)
+    cache_key = [set, api_data]
+    cache = @@cache[cache_key]
+    return cache if cache
+    @@cache[cache_key] = super
+  end
+
+  ##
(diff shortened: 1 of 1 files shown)


hsbt:
Thanks!

lib/rubygems/resolver/api_specification.rb
```diff
@@ -6,6 +6,17 @@
 # is the name, version, and dependencies.
 
 class Gem::Resolver::APISpecification < Gem::Resolver::Specification
+  ##
+  # We assume that all instances of this class are immutable;
+  # so avoid duplicated generation for performance.
+  @@cache = {}
```
eregon, line 12: Class variables are pretty inefficient, I think on all Ruby implementations.
Should we use a constant or `@ivar` on the class instead here?

PR #3942, opened 2020-09-11
Revert "Avoid duplicated generation of APISpecification objects"

kadirselcuk:
Reverts rubygems/rubygems#3940


kadirselcuk (commented):
APIS kodunu geri almak

PR #3951, opened 2020-09-17
Fix `typo_squatting?` false positive for `rubygems.org` itself

andy-smith-msm:
# Description:

Correct typo_squatting behaviour to allow adding https://rubygems.org without triggering input

AMI prebuilt using internal gem server + another scripted scenario for some of our builds which re-add https://rubygems.org

This triggers the prompt saying that https://rubygems.org is similar to https://rubygems.org which totally breaks the script due to a) No option to default -y b) Strict tty checks with no default which cause a failure and mean for horrible workarounds c) This specific issue which is independent but contributes to the overall problem.

Ensure that correctly entered https://rubygems.org does _not_ raise the result that it is similar, because it is _not_ similar but identical and poses no issue with the desired effect (that stops folks accidentally configuring camped domains and getting supply chain attacks).

We do this by instead of using <= distrance_threshold (which also covers zero) by instead using a range check from 1..distance_threshold
______________

# Tasks:

- [X] Write tests
- [X] Get code review from coworkers / friends

### DIFF
--- a/lib/rubygems/source.rb
+++ b/lib/rubygems/source.rb
@@ -225,7 +225,7 @@ class Gem::Source
 
   def typo_squatting?(host, distance_threshold=4)
     return if @uri.host.nil?
-    levenshtein_distance(@uri.host, host) <= distance_threshold
+    levenshtein_distance(@uri.host, host).between? 1, distance_threshold
   end
 end
 
--- a/test/rubygems/test_gem_source.rb
+++ b/test/rubygems/test_gem_source.rb
@@ -240,6 +240,11 @@ class TestGemSource < Gem::TestCase
     refute rubygems_source.typo_squatting?("rubysertgems.org")
   end
 
+  def test_typo_squatting_false_positive
+    rubygems_source = Gem::Source.new("https://rubygems.org")
+    refute rubygems_source.typo_squatting?("rubygems.org")
+  end
+
   def test_typo_squatting_custom_distance_threshold
     rubygems_source = Gem::Source.new("https://rubgems.org")
     distance_threshold = 5


sean-msmg (approved):

sean-msmg (approved):

androidwiltron (approved):

paulsmyth-msg (approved):

simi:
I do agree on handling this better than current state, but wouldn't be better to check if that source is already present and just warn nothing is happening since that source is already present (ant thus there's nothing to do)?

sean-msmg:
The typo_squatting? check is hard-coded to rubygems.org, not based on existing sources. The bug means you can't clear out all sources then re-add rubygems.org

https://github.com/rubygems/rubygems/blob/6914b4ec439ae1e7079b3c08576cb3fbce68aa69/lib/rubygems/commands/sources_command.rb#L67

andy-smith-msm:
@simi this would be a workaround, I've tried to stay faithfully to ensuring the typo_squatting? method more accurately reflects what it was supposed to do as per the original case, which is to spot nearly identical but not identical matches and avoid folks becoming vulnerable to MITM/supply chain attacks.

There are a number of issues in working around it for us and this is only one of the defects really that should be raised against the current implementation. I chose the one that seemed to provide correctness in the single method implementation against the original driver.

andy-smith-msm:
> The typo_squatting? check is hard-coded to rubygems.org, not based on existing sources. The bug means you can't clear out all sources then re-add rubygems.org
> 
> https://github.com/rubygems/rubygems/blob/6914b4ec439ae1e7079b3c08576cb3fbce68aa69/lib/rubygems/commands/sources_command.rb#L67

For clarity @simi this is what we have.

We have a cloud pipeline with servers built with configurations that already exclude https://rubygems.org in favour of our servers. We have some mechanisms for individual hosts that need to reset this on system build and this mechanism works find with earlier versions of gem but we are trying to migrate forward to gem 3.x

simon71 (approved):

simi (approved):
Thanks for additional info, that makes it clear.

PS: We should also support for `-y` or `--force` flag later as well to not block this quick fix.

andy-smith-msm:
> Thanks for additional info, that makes it clear.
> 
> PS: We should also support for `-y` or `--force` flag later as well to not block this quick fix.


Thanks @simi , yes for the related impediments I was considering doing this also in a seperate PR.

simi:
@andy-smith-msm I have created placeholder for followup - https://github.com/rubygems/rubygems/issues/3955.

andy-smith-msm:
> @andy-smith-msm I have created placeholder for followup - #3955.

good stuff, I'll try to spend some time on this and quote #3955 in the commit ID

deivid-rodriguez (approved):
Looks good to me too!

deivid-rodriguez:
Thanks for this @andy-smith-msm!

PR #1454, opened 2016-01-14
Allow the user to omit .gemspec when doing gem build

sblackstone:
When in the directory of a gem and using bash autocomplete, you often have directories which match the name of the gem as well as your gemspec file - so "gem build *someprefix* *tab*" then makes you stop and type .g *tab* in order to autocomplete the name.

This patch changes this so if you omit .gemspec and the file isn't found, it will try attaching .gemspec before giving up.

PR #3954, opened 2020-09-17
Fix `gem build -C`

### LINKED ISSUE #3953: The `gem build -C` does not work as expected
This is followup of #2587 and I am not really sure what is the purpose of this code:

https://github.com/rubygems/rubygems/blob/8fd5dee8fcfdda7db74c579616fb5a718c287e6d/lib/rubygems/commands/build_command.rb#L87-L89

If I specify the `-C` option, why the `Dir.chdir` does not use the specified path? I think the current code is clumsy. It tries to do guess everything, but nothing is done properly. For example, I don't understand why it tries to used `gem_name` and then tries also the `"#{gem_name}.gemspec"` when this behavior is not documented. Is the .gemspec file expected to be located in the current directory or is the path expected?

This is specifically behavior I complain about:

~~~
$ ls | grep gemspec
abrt.gemspec

$ mkdir ci

$ cd ci

$ gem build *.gemspec -C ..
ERROR:  Gemspec file not found: *.gemspec.gemspec

$ # Hmmm, what is this ^^ error. The *.gemspec.gemspec is confusing. I have not asked for anything like this.

$ cp ../abrt.gemspec .

$ gem build abrt.gemspec -C ..
abrt.gemspec
    ["LICENSE.txt", "README.md", "Rakefile"] are not files

$ # But why this ^^ does not work?

$ gem build ../*.gemspec -C ..
  Successfully built RubyGem
  Name: abrt
  Version: 0.4.0
  File: abrt-0.4.0.gem

$ ll
total 4
-rw-rw-r--. 1 vondruch vondruch 817 Sep 17 14:36 abrt.gemspec

$ # But where is my package? :(
~~~

voxik:
The directory specified by `-C` option must be reflected, while the already loaded spec file must be used, wherever it came from. This also fixes confusing messages when .gemspec file was not found at all.

Fixes #3953

I have not tried to run the test suite, so let me see what is going to fail ...

### DIFF
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -23,7 +23,9 @@ class Gem::Commands::BuildCommand < Gem::Command
       options[:output] = value
     end
 
-    add_option '-C PATH', '', 'Run as if gem build was started in <PATH> instead of the current working directory.' do |value, options|
+    add_option '-C PATH', '', 'Look for the code of the gem to be built',
+                              'in <PATH>. The GEMSPEC_FILE given as an',
+                              'argument is not influenced by this option' do |value, options|
       options[:build_path] = value
     end
   end
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -85,8 +85,7 @@ Gems can be saved to a specified filename with the output option:
       spec = Gem::Specification.load(gemspec)
 
       if options[:build_path]
-        Dir.chdir(File.dirname(gemspec)) do
-          spec = Gem::Specification.load(File.basename(gemspec))
+        Dir.chdir(options[:build_path]) do
           build_package(spec)
         end
       else
--- a/test/rubygems/test_gem_commands_build_command.rb
+++ b/test/rubygems/test_gem_commands_build_command.rb
@@ -272,6 +272,47 @@ class TestGemCommandsBuildCommand < Gem::TestCase
     assert_equal "this is a summary", spec.summary
   end
 
+  def test_execute_outside_dir_with_external_gemspec
+    gemspec_dir = File.join @tempdir, 'gemspec_dir'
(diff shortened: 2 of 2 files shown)


voxik:
I have pushed update which should fix the failing `TestGemCommandsBuildCommand#test_can_find_gemspecs_without_dot_gemspec` test case (which was broken from beginning #1454).

deivid-rodriguez:
Thanks @voxik! This looks good to me. Only requests:

* Could you add a test to cover https://github.com/rubygems/rubygems/pull/3954/commits/15e431b535d6ae550d1c2d865c06a5e0c90a50f1?
* Could you split the second bug fix to a separate PR?

deivid-rodriguez:
Hi @voxik.

I'm now re-reading this patch and I think this does not solve the problem as expected.

The help for the `-C` flag says

> Run as if gem build was started in <PATH> instead of the current working directory.

However, in this patch, the gemspec is resolved before switching directories, so if it exists in the current directory, it will be taken from there. I would expect the given `gemspec` to be picked up _as is_ if an absolute path is given, but otherwise be resolved from the path specified in the `-C` CLI flag.

#3983 has been proposed fixing this issue as I would have expected, but maybe you had other expectations for the `-C` flag, so if that's the case, let's discuss it.

voxik:
>     * Could you split the second bug fix to a separate PR?

Done: #3988

voxik:
>     * Could you add a test to cover [15e431b](https://github.com/rubygems/rubygems/commit/15e431b535d6ae550d1c2d865c06a5e0c90a50f1)?

Done

deivid-rodriguez:
Just to avoid having two separate PRs fixing the same thing, I'll close this in favor of https://github.com/rubygems/rubygems/pull/4022. If you prefer to change this PR yourself to follow the approach discussed in #4022, I can also close that one and follow up in here, I don't mind either way.

PR #3956, opened 2020-09-18
Add `--force` option to `gem sources` command

### LINKED ISSUE #3955: Support -y or --force option in ask_yes_no.
As raised in https://github.com/rubygems/rubygems/pull/3951, there's no way to provide CLI argument to force yes answer to all UI yes/no interactions breaking automated scripts.

To solve this, it should be possible to add global `--force` or `-y` argument setting some global `Gem` state and use it in `ask_yes_no` method to provide explicit value when provided.

https://github.com/rubygems/rubygems/blob/6914b4ec439ae1e7079b3c08576cb3fbce68aa69/lib/rubygems/user_interaction.rb#L250-L281

andy-smith-msm:
Allows --force to override use of http URLs and typo checks

# Description:

As discussed in #3955 implements a --force option to the sources command

Discussed in #3955 as a complete solution to scripted use of gem which cannot handle no tty input within ask_yes_no

Localised fix within commands/sources_command.rb since implementation within user_interaction.rb suggests the creation of user_interaction_options.rb to follow the framework but this would require a larger scale change to centralise the several already implemented --force options within other modules. --force chosen to be consistent with other module usage vs -y.

Also note that --force potentially does different things in other modules so the refactor mentioned is potentially extensive which is why the override is confined to the sources command implementation.

Closes #3955.
______________

# Tasks:

- [X] Write tests

### DIFF
--- a/lib/rubygems/commands/sources_command.rb
+++ b/lib/rubygems/commands/sources_command.rb
@@ -34,6 +34,10 @@ class Gem::Commands::SourcesCommand < Gem::Command
       options[:update] = value
     end
 
+    add_option '-f', '--[no-]force', "Do not show any confirmation prompts and behave as if 'yes' was always answered" do |value, options|
+      options[:force] = value
+    end
+
     add_proxy_option
   end
 
@@ -71,7 +75,7 @@ class Gem::Commands::SourcesCommand < Gem::Command
 Do you want to add this source?
       QUESTION
 
-      terminate_interaction 1 unless ask_yes_no question
+      terminate_interaction 1 unless options[:force] || ask_yes_no(question)
     end
   end
 
@@ -86,7 +90,7 @@ https://rubygems.org is recommended for security over #{uri}
 Do you want to add this insecure source?
       QUESTION
 
-      terminate_interaction 1 unless ask_yes_no question
+      terminate_interaction 1 unless options[:force] || ask_yes_no(question)
     end
   end
 
--- a/test/rubygems/test_gem_commands_sources_command.rb
+++ b/test/rubygems/test_gem_commands_sources_command.rb
@@ -107,6 +107,36 @@ class TestGemCommandsSourcesCommand < Gem::TestCase
     assert_empty ui.error
   end
 
+  def test_execute_add_allow_typo_squatting_source_forced
+    rubygems_org = "https://rubyems.org"
+
+    spec_fetcher do |fetcher|
+      fetcher.spec("a", 1)
+    end
+
+    specs = Gem::Specification.map do |spec|
+      [spec.name, spec.version, spec.original_platform]
+    end
+
+    specs_dump_gz = StringIO.new
+    Zlib::GzipWriter.wrap(specs_dump_gz) do |io|
+      Marshal.dump(specs, io)
+    end
+
+    @fetcher.data["#{rubygems_org}/specs.#{@marshal_version}.gz"] = specs_dump_gz.string
+    @cmd.handle_options %W[--force --add #{rubygems_org}]
+
+    @cmd.execute
+
+    expected = "https://rubyems.org added to sources\n"
+    assert_equal expected, ui.output
+
+    source = Gem::Source.new(rubygems_org)
+    assert Gem.sources.include?(source)
+
+    assert_empty ui.error
+  end
+
   def test_execute_add_deny_typo_squatting_source
     rubygems_org = "https://rubyems.org"
 
@@ -281,6 +311,36 @@ source http://gems.example.com/ already present in the cache
 
     assert_equal expected, @ui.output
     assert_empty @ui.error
+  end
+
+  def test_execute_add_http_rubygems_org_forced
+    rubygems_org = "http://rubygems.org"
+
+    spec_fetcher do |fetcher|
+      fetcher.spec("a", 1)
+    end
+
+    specs = Gem::Specification.map do |spec|
+      [spec.name, spec.version, spec.original_platform]
+    end
+
+    specs_dump_gz = StringIO.new
+    Zlib::GzipWriter.wrap(specs_dump_gz) do |io|
+      Marshal.dump(specs, io)
+    end
+
+    @fetcher.data["#{rubygems_org}/specs.#{@marshal_version}.gz"] = specs_dump_gz.string
+    @cmd.handle_options %W[--force --add #{rubygems_org}]
+
+    @cmd.execute
+
+    expected = "http://rubygems.org added to sources\n"
+    assert_equal expected, ui.output
+
+    source = Gem::Source.new(rubygems_org)
+    assert Gem.sources.include?(source)
+
+    assert_empty ui.error
   end
 
   def test_execute_add_https_rubygems_org


lib/rubygems/commands/sources_command.rb
```diff
@@ -71,7 +75,7 @@ def check_typo_squatting(source)
 Do you want to add this source?
       QUESTION
 
-      terminate_interaction 1 unless ask_yes_no question
+      terminate_interaction 1 unless options[:force] || ask_yes_no(question)
```
deivid-rodriguez, line 78: `ask_yes_no` already supports a default parameter, so it should work if you pass `options[:force]` as the second argument?

andy-smith-msm: @deivid-rodriguez yes, if you take a look at the implementation this default parameter does something different. It is only used when tty? is false and to change that contract is beyond this. I had considered an additional force option on that method to keep the existing behaviours but settled on this for safety.

deivid-rodriguez:
A tiny note, don't reference issues in the PR title or commit message, only in the PR body.

Rationale is:

* For PR titles, we use those to generate the changelog, and they won't get properly autolinked in that context.
* For commit messages, I prefer git history to be independent of the source tracker, and to force the commit message author to explain what's being fixed independently without resorting to linking to github. Also, when referencing issues in commits and force-pushing, and reference is generated in the reference ticket for every force-push wich I find a bit annoying :sweat_smile:.

lib/rubygems/commands/sources_command.rb
deivid-rodriguez: You're right :+1:.

andy-smith-msm:
> A tiny note, don't reference issues in the PR title or commit message, only in the PR body.
> 
> Rationale is:
> 
> * For PR titles, we use those to generate the changelog, and they won't get properly autolinked in that context.
> * For commit messages, I prefer git history to be independent of the source tracker, and to force the commit message author to explain what's being fixed independently without resorting to linking to github. Also, when referencing issues in commits and force-pushing, and reference is generated in the reference ticket for every force-push wich I find a bit annoying 😅.

Updated commit message as requested :)

lib/rubygems/commands/sources_command.rb
```diff
@@ -34,6 +34,10 @@ def initialize
       options[:update] = value
     end
 
+    add_option '-f', '--[no-]force', 'Force prompted confirmations' do |value, options|
```
deivid-rodriguez, line 37: Maybe a better help text would be "Automatically answer 'yes' to all confirmation prompts" or something like that?

andy-smith-msm: If you like, I will update.

I would suggest we stick with --force as an option though for consistency with other commands.

andy-smith-msm: Updated as suggested :+1:

deivid-rodriguez: Yup, I only meant to update the help text, the name seems fine :+1:.

sean-msmg (approved):

adamarnold-msm (approved):

simi:
:information_source: This adds `--force` just to this one command only. There's chance to make it global flag (like --verbose) and handle this in `ask_yes_no` globally.

andy-smith-msm:
> ℹ️ This adds `--force` just to this one command only. There's chance to make it global flag (like --verbose) and handle this in `ask_yes_no` globally.

I had a chat with @deivid-rodriguez about the approach here so here is the full summary.

1. ask_yes_no default only works when tty? is false. Changing this means changing behaviours that existing users may have workarounds in. Could be a breaking change.
2. ask_yes_no option should probably be consistent with --force used in other commands so extending it's usage would require a refactor of all existing commands and a clear understanding of how --force differs for those other commands. I aggree that it is cleanest to implement a user_interface_options.rb, link ask_yes_no to that etc. but this would be a much larger PR and risker a refactor and I also didn't want to mix a large refactor with a small code change.
3. --force options are consistently currently implemented in the command scripts so it seems in keeping with the project for this commit to do the same pending a larger refactor effort.

simi (approved):

deivid-rodriguez:
Yes @simi, adding a global `--force` option or similar makes total sense, but it probably requires some refactoring and a review of all what all current `--force` options currently do, in order to not create any naming conflicts. I would open a new ticket to track that improvement.

andy-smith-msm:
9 failed checks but all it says on each one is 'This check failed' on the Details. Could this be a build server restart or something?

simi:
https://github.com/rubygems/rubygems/actions/runs/261103746

> This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.

Yes, it seems like CI problem.

andy-smith-msm:
@simi is there anything that will restart these after it fails in this state or do I have to squash on an empty commit to trigger it to run everything again based on a new commit hash?

simi:
I can't restart, but you can do `git commit --amend --no-edit` and force push.

andy-smith-msm:
> I can't restart, but you can do `git commit --amend --no-edit` and force push.

Thanks @simi, done :)

test/rubygems/test_gem_commands_sources_command.rb
```diff
@@ -107,6 +107,39 @@ def test_execute_add_allow_typo_squatting_source
...
+    end
+
+    specs_dump_gz = StringIO.new
+    Zlib::GzipWriter.wrap(specs_dump_gz) do |io|
+      Marshal.dump(specs, io)
+    end
+
+    @fetcher.data["#{rubygems_org}/specs.#{@marshal_version}.gz"] = specs_dump_gz.string
+    @cmd.handle_options %W[--force --add #{rubygems_org}]
+    ui = Gem::MockGemUi.new("y")
```
deivid-rodriguez, line 128: If I understand this test correctly, you don't need a special UI here mocking input, since no questions will be asked.

test/rubygems/test_gem_commands_sources_command.rb
```diff
@@ -283,6 +316,39 @@ def test_execute_add_http_rubygems_org
...
```
deivid-rodriguez, line 337: Same here about the UI, but I wonder if we need this test at all since this case will be fixed by https://github.com/rubygems/rubygems/pull/3951, rright?

deivid-rodriguez (commented):
Looks good to me, just noticed a couple of small things.

test/rubygems/test_gem_commands_sources_command.rb
andy-smith-msm: ah good point, yes I can remove that.

test/rubygems/test_gem_commands_sources_command.rb
deivid-rodriguez: Oh, I see, thanks!

test/rubygems/test_gem_commands_sources_command.rb
andy-smith-msm: Updated tests with the mock wrapper remove have been rebased into a new commit and pushed

lib/rubygems/commands/sources_command.rb
```diff
@@ -34,6 +34,10 @@ def initialize
       options[:update] = value
     end
 
+    add_option '-f', '--[no-]force', "Automatically answer 'yes' to all confirmation prompts" do |value, options|
```
deivid-rodriguez, line 37: Since the actually behaviour is that we skip showing the prompts at all, as opposed to showing them and answering `yest` automatically, we could maybe make the help text slightly more accurate by saying:

```suggestion
    add_option '-f', '--[no-]force', "Do show any confirmation prompts and behave as if 'yes' was always answered" do |value, options|
```

andy-smith-msm: Using that phrase slightly changed to say 'Do not show'

andy-smith-msm: Update pushed

deivid-rodriguez (approved):
Thanks!

PR #3960, opened 2020-09-23
rubygems is needed to run solo-file test

hsbt:
# Description:

When We invoked with only `test/rubygems/test_gem_source_list.rb`, It failed the missing file.
______________

# Tasks:

- [ ] Write tests

### DIFF
--- a/test/rubygems/test_gem_source_list.rb
+++ b/test/rubygems/test_gem_source_list.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+require 'rubygems'
 require 'rubygems/source_list'
 require 'rubygems/test_case'
 


deivid-rodriguez:
I assume you're running this file with `RUBYOPT=--disable-gems`? Otherwise `rubygems` should be required automatically. If we want to support this, we should probably add this require on top of the `lib/rubygems/test_case.rb` instead?

hsbt:
> I assume you're running this file with `RUBYOPT=--disable-gems`?

Maybe yes. It was reported by @ko1 . The ruby core test suite uses `--disable-gems`.

deivid-rodriguez (approved):
We can unify this more as a follow up to this PR.

Issue #3363, opened 2018-06-18
`settings[:gemfile]` from `.bundle/config` file is not respected when doing `require 'bundle/setup'`


faucct:
Gemfile configuration from `.bundle/config` is only respected when you use `Bundler::CLI`. Can someone tell me why, please?

hsbt:
This was fixed in #4892.

`Bundler.setup`, and therefore require "bundler/setup", now reads the gemfile setting from `.bundle/config` the same way `Bundler::CLI` does.

Issue #3294, opened 2019-11-19
Bundler 2 does not auto-switch versions when BUNDLE_GEMFILE is set via .bundle/config


jtokoph:
Bundler 2 is supposed to detect if a project was bundled with v1 based on the `BUNDLED WITH` lines in the lockfile.

This is not working correctly if:
1. The Gemfile is in a subdirectory or with an alternate name
2. The alternate location is specified in `.bundle/config` with `BUNDLE_GEMFILE`
3. `bundle update` is run from the project root directory

Note: This is working correctly if one of the following is true:
- `bundle update` is run from the subdirectory that includes the Gemfile
- _or_ `bundle update` is run with `BUNDLE_GEMFILE` env var set to the alternate location

## Expected result

Running bundler 2.x on a project that was bundled with 1.x would be that  `BUNDLED_WITH` would not be updated.

## Observed result

`BUNDLED WITH` is updated to 2.x

## Example repo to reproduce

A repository that demonstrates this behavior is located here: https://github.com/jtokoph/bundler2bug

The readme has steps to reproduce.

RomainMorlevat:
Same issue on one project here, with Gemfile in the (Rails) root folder.

hsbt:
This works on current master. The `gemfile` setting from `.bundle/config` is promoted to `BUNDLE_GEMFILE` before the version check runs, so `SelfManager` reads `BUNDLED WITH` from that lockfile and switches to the version it names.

PR #3963, opened 2020-09-25
Make config location relative to the current working directory

### LINKED ISSUE #2048: Installation fails on Nano Server 1709
Installation on the latest Nano Server docker image (microsoft/nanoserver:1709) fails with the error

gem install --local C:\tmp\rubygems-update-2.6.13.gem --no-ri --no-rdoc
C:/Ruby_2.2.4_x64/lib/ruby/2.2.0/rubygems/config_file.rb:89:in `join': no implicit conversion of nil into String (TypeError)

This looks to be due to the usage of SHGetFolderPath in SYSTEM_CONFIG_PATH, which is returning blank values on Nano Server ([https://github.com/dotnet/corefx/issues/19110](url)) - instead it would be better to use the ALLUSERSPROFILE environment variable.

deivid-rodriguez:
This PR ports https://github.com/rubygems/bundler/pull/6713 from the old repository.

### What was the end-user problem that led to this PR?

There are two related end-user problems.

* When a custom `Gemfile` is configured via `BUNDLE_GEMFILE` and we run `bundle config`, the local config is not generated in the current working directory. The reason is in the diagnosis section, but the end user problem is that it is very counterintuitive. On bundler 2 this is aggravated because of "sticky options". See https://github.com/bundler/bundler/issues/2048#issuecomment-7615264 for an example.

* When configuring a custom gemfile via settings, it is not respected when requiring `bundler/setup`. See #6589.

### What was your diagnosis of the problem?

My diagnosis was that the config location is currently resolved from the current `Gemfile` location, and that leads to configuration being created on counterintuitive locations. For example, you have `BUNDLE_GEMFILE` set to a particular non-root location, and run `bundle config` on the root folder, but no local config is apparently generated... until you figure out it's being generated in the same folder `BUNDLE_GEMFILE` is pointing to.

Also, since the config location depends on the location of the current `Gemfile`, we can't consistently resolve the `Gemfile` from configuration, because of a chicken-and-egg problem. We need the `Gemfile` to figure out where the config is, but we need the config to figure out where the `Gemfile` is. This is why #3363 can't currently work.

### What is your fix for the problem, implemented in this PR?

My fix was to change how we resolve the configuration root, and instead look up in the directory hierarchy for an existing local config, and otherwise default to the current working directory. With that in place, fixing #6589 is easy. My changes led to a single spec failure. Even if it's just one, it is backwards incompatible if backwards compatibility is defined by our current set of specs. Since I'd like to be able to opt-in to the new behavior right now on bundler 1, I made a feature flag for this change. This feature flag needed to be different from the other feature flags we currently have, because it can't be config-based (chicken-and-egg problem again). So I introduced "env-based feature flags".

Once we have the new configuration root resolution in place, the fix for #6589 is to add knowledge to the `find_gemfile` method to read from settings (and not just from env).

### Why did you choose this fix out of the possible options?

I chose this fix because changing the way we resolve configuration seemed like the only way to resolve the problems that I wanted to fix here.

Fixes https://github.com/rubygems/rubygems/issues/3363.
Fixes https://github.com/rubygems/bundler/issues/2048 (the remaining part of it, https://github.com/rubygems/bundler/issues/2048#issuecomment-7615264 and https://github.com/rubygems/bundler/issues/2048#issuecomment-47390038).
Fixes https://github.com/rubygems/rubygems/issues/3294.

# Tasks:

- [x] Write tests

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,2 +1,0 @@
----
-BUNDLE_PATH__SYSTEM: "true"
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -29,7 +29,8 @@ jobs:
         run: echo "REF=$(ruby -v | cut -d')' -f1 | cut -d' ' -f5)" >> $GITHUB_ENV
       - uses: actions/checkout@v2
         with:
-          repository: ruby/ruby
+          repository: deivid-rodriguez/ruby
+          ref: workaround-bundler-issue
           path: ruby/ruby
           fetch-depth: 10
       - name: Checkout the latest buildable revision
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -29,8 +29,7 @@ jobs:
         run: echo "REF=$(ruby -v | cut -d')' -f1 | cut -d' ' -f5)" >> $GITHUB_ENV
       - uses: actions/checkout@v2
         with:
-          repository: deivid-rodriguez/ruby
-          ref: workaround-bundler-issue
+          repository: ruby/ruby
           path: ruby/ruby
           fetch-depth: 10
       - name: Checkout the latest buildable revision
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,7 @@ require "rake/testtask"
 
 desc "Setup Rubygems dev environment"
 task :setup do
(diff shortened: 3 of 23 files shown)


brodock:
what is missing here to get this merged?

deivid-rodriguez:
This needs a rebase and a second look to evaluate compatibility. I'll look into it next year.

PR #3974, opened 2020-09-28
Disable `fail-fast` mode in CI

deivid-rodriguez:
Github Actions "fails fast" by default. However, most of the times when I get an unexpected build failure, I'll want to know if it's specific to a ruby-version or other variables our matrix use, so a `fail-fast: false` by default is more useful to me.

This is obviously heavier, but not that much heavier I believe because by the time a job fails, other jobs might already be closed to completion, and this does not affect the majority of cases where everything passes.  

# Tasks:

- [ ] Write tests

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -8,9 +8,6 @@ jobs:
   daily_bundler:
     runs-on: ubuntu-18.04
     if: github.repository == 'rubygems/rubygems'
-    strategy:
-      matrix:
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
This is specially important for environments like Windows which are currently hard to simulate since none of the current maintainers use Windows.

hsbt:
I'm +1 to this.

hsbt (approved):

deivid-rodriguez:
Awesome, thanks @hsbt!

PR #3985, opened 2020-10-01
Fix SystemStackError when overriding `Warning.warn`

### LINKED ISSUE #2588: Stack level too deep when overriding `Warning.warn`
Hello everyone 👋 ,

Thanks a lot for maintaining rubygems ❤️. I encountered an issue when upgrading rubygems to 3.0 yesterday, the exception is: `SystemStackError (stack level too deep)`

The issue can be reproduced when your application uses the new `Warning.warn` method from ruby 2.5 (https://ruby-doc.org/core-2.5.0/Warning.html)

```ruby
module Warning
  def warn(str)
    # some custom pre-processing of the warning message
    super
  end
end

warn 'Foo Bar'
```

<details>
  <summary> StackTrace </summary>

   ```
       16: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `block in <module:Kernel>'
       15: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `call'
       14: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `warn'
       13: from (irb):4:in `warn'
       12: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `block in <module:Kernel>'
       11: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `call'
       10: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `warn'
        9: from (irb):4:in `warn'
        8: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `block in <module:Kernel>'
        7: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `call'
        6: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `warn'
        5: from (irb):4:in `warn'
        4: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `block in <module:Kernel>'
        3: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `call'
        2: from /opt/rubies/2.5.3/lib/ruby/site_ruby/2.5.0/rubygems/core_ext/kernel_warn.rb:17:in `warn'
        1: from (irb):4:in `warn'
SystemStackError (stack level too deep)
```
</details>

-------------

This commit seems to have caused this regression https://github.com/rubygems/rubygems/commit/7153a317c3e545658459d84aeafdad891d3a8a36#diff-5f3c1066128b69ac52cca10be3eee4f0

```
$ gem env version
3.0.2
$ ruby -v
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin17]
```

Thanks !

### PR DESCRIPTION
**Pull request by deivid-rodriguez**

When overriding `Warning.warn` like this:

```ruby
module Warning
  def warn(str)
    # some custom pre-processing of the warning message
    super
  end
end

warn 'Foo Bar'
```

Rubygems would throw a `SystemStackError`.

While experimenting with a solution, I changed our monkeypatch to use `prepend` instead, and that revealed an off-by-one error in the implementation. The previous implementation was working because (I have no idea why), two duplicated entries were present in the callstack for `kernel_require.rb`, automatically "correcting" the issue.

This implementation feels better and works as expected with `Warning.warn`.

Fixes #2588.

# Tasks:

- [x] Write tests

### DIFF
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -3,23 +3,15 @@
 # `uplevel` keyword argument of Kernel#warn is available since ruby 2.5.
 if RUBY_VERSION >= "2.5"
 
-  module Kernel
-    rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
+  module Gem::KernelExt
+    def warn(*messages, **kw)
+      rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
 
-    original_warn = method(:warn)
-
-    remove_method :warn
-
-    class << self
-      remove_method :warn
-    end
-
-    module_function define_method(:warn) {|*messages, **kw|
       unless uplevel = kw[:uplevel]
         if Gem.java_platform?
-          return original_warn.call(*messages)
+          return super(*messages)
         else
-          return original_warn.call(*messages, **kw)
+          return super(*messages, **kw)
         end
       end
 
@@ -36,19 +28,23 @@ if RUBY_VERSION >= "2.5"
             break
           end
 
-          start += 1
-
           if path = loc.path
             unless path.start_with?(rubygems_path) or path.start_with?('<internal:')
               # Non-rubygems frames
               uplevel -= 1
+              break if uplevel < 0
             end
           end
+
+          start += 1
         end
         kw[:uplevel] = start
       end
 
-      original_warn.call(*messages, **kw)
-    }
+      super(*messages, **kw)
+    end
   end
+
+  extend Gem::KernelExt
+  Kernel.singleton_class.prepend Gem::KernelExt
 end
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -45,10 +45,9 @@ if RUBY_VERSION >= "2.5"
             end
           end
         end
-        uplevel = start
+        kw[:uplevel] = start
       end
 
-      kw[:uplevel] = uplevel
       original_warn.call(*messages, **kw)
     }
   end
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -677,6 +677,22 @@ class TestGemRequire < Gem::TestCase
         end
       end
     end
+
+    def test_no_crash_when_overriding_warn_with_warning_module
+      skip "https://github.com/oracle/truffleruby/issues/2109" if RUBY_ENGINE == "truffleruby"
+
+      Dir.mktmpdir("warn_test") do |dir|
+        File.write(dir + "/main.rb", "module Warning; def warn(str); super; end; end; warn 'Foo Bar'")
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--disable=gems", "-C", dir, "main.rb")
+        end
+        assert_match(/Foo Bar\n$/, err)
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--enable=gems", "-C", dir, "main.rb")
+        end
+        assert_match(/Foo Bar\n$/, err)
+      end
+    end
   end
 
   private


lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -3,23 +3,15 @@
 # `uplevel` keyword argument of Kernel#warn is available since ruby 2.5.
 if RUBY_VERSION >= "2.5"
 
-  module Kernel
-    rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
+  module KernelExt
```
eregon, line 6: This declares a global `KernelExt`, should probably be under `Gem::` to avoid conflicts.

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -36,20 +28,25 @@ class << self
...
 
-      kw[:uplevel] = uplevel
-      original_warn.call(*messages, **kw)
-    }
+      super(*messages, **kw)
+    end
   end
+
+  Kernel.prepend(KernelExt)
+  Object.include(Kernel)
```
eregon, line 49: This should be unnecessary.

deivid-rodriguez: Did you try it? I think it's necessary because if `Kernel` has already been included into `Object`, the enhanced `Kernel` won't take effect on `warn` calls without a receiver.

eregon: I'll try it tomorrow.
Changing `Kernel#warn` is quite tricky, much harder than `Warning.warn`, I wish there would be a proper API for that.

One downside of this approach is it adds an extra entry in ancestors for every `Object`. That slows down every method lookup that would read a Kernel method or above.
And I guess it won't work for things like `Delegator < BasicObject` which `include Kernel` (although Delegator makes a copy of Kernel so that case is more complex than that).

One idea would be to still monkey patch, but to call the original Kernel#warn with `Warning` as self when self is `Warning` to avoid the recursion. In other words, to preserve the value of `self` through this override. It could also be done with an UnboundMethod and then `bind(self).call()`.

deivid-rodriguez: I see. Would you be able to propose a follow up patch fixing this?

deivid-rodriguez: I changed it to extend `Object` instead of prepending. Should it better since it does not modify `Object`'s ancestors chain?

eregon: I just made https://github.com/rubygems/rubygems/pull/3987

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -36,20 +28,23 @@ class << self
...
 
-      kw[:uplevel] = uplevel
-      original_warn.call(*messages, **kw)
-    }
+      super(*messages, **kw)
+    end
   end
+
+  extend Gem::KernelExt
+  Kernel.singleton_class.prepend Gem::KernelExt
```
eregon, line 49: I think this will not work for a non-main-object non-`Kernel.warn` receiver, like if you call `warn` in some method of some class.

deivid-rodriguez: Could you add a test for that to your PR, to prove that your approach works better? Thanks!

lib/rubygems/core_ext/kernel_warn.rb
eregon: Could you review that one? I think it's the safest approach with the least side effects.
`prepend` / `extend` look nicer, but they don't achieve the same effect and they have various corner cases where they don't work (for modules like Kernel, for classes it would work better).

deivid-rodriguez: Thanks!

deivid-rodriguez:
Closing in favor of #3987.

lib/rubygems/core_ext/kernel_warn.rb
eregon: Confirmed it doesn't work in this case, in that that would call the original Kernel#warn and skip RubyGems' variant.
Trying to write a test for it, it's not easy.

PR #3987, opened 2020-10-02
Preserve `self` when calling the original Kernel#warn method

eregon:
Alternative to https://github.com/rubygems/rubygems/pull/3985 which does not require adding a new module next to Kernel, and also works if some class inherits from `BasicObject` and `include Kernel`.

* This fixes infinite recursion when monkey-patching Warning#warn, which is [not recommended](https://github.com/rubygems/rubygems/issues/2588#issuecomment-702379220), but still should behave the same whether RubyGems is loaded or not.
* Fixes https://github.com/rubygems/rubygems/issues/2588.

cc @deivid-rodriguez

### DIFF
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -6,7 +6,7 @@ if RUBY_VERSION >= "2.5"
   module Kernel
     rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
 
-    original_warn = method(:warn)
+    original_warn = instance_method(:warn)
 
     remove_method :warn
 
@@ -17,9 +17,9 @@ if RUBY_VERSION >= "2.5"
     module_function define_method(:warn) {|*messages, **kw|
       unless uplevel = kw[:uplevel]
         if Gem.java_platform?
-          return original_warn.call(*messages)
+          return original_warn.bind(self).call(*messages)
         else
-          return original_warn.call(*messages, **kw)
+          return original_warn.bind(self).call(*messages, **kw)
         end
       end
 
@@ -48,7 +48,7 @@ if RUBY_VERSION >= "2.5"
         kw[:uplevel] = start
       end
 
-      original_warn.call(*messages, **kw)
+      original_warn.bind(self).call(*messages, **kw)
     }
   end
 end
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -45,10 +45,9 @@ if RUBY_VERSION >= "2.5"
             end
           end
         end
-        uplevel = start
+        kw[:uplevel] = start
       end
 
-      kw[:uplevel] = uplevel
       original_warn.call(*messages, **kw)
     }
   end
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -677,6 +677,47 @@ class TestGemRequire < Gem::TestCase
         end
       end
     end
+
+    def test_no_crash_when_overriding_warn_with_warning_module
+      skip "https://github.com/oracle/truffleruby/issues/2109" if RUBY_ENGINE == "truffleruby"
+
+      Dir.mktmpdir("warn_test") do |dir|
+        File.write(dir + "/main.rb", "module Warning; def warn(str); super; end; end; warn 'Foo Bar'")
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--disable=gems", "-C", dir, "main.rb")
+        end
+        assert_match(/Foo Bar\n$/, err)
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--enable=gems", "-C", dir, "main.rb")
+        end
+        assert_match(/Foo Bar\n$/, err)
+      end
+    end
+
+    def test_expected_backtrace_location_when_inheriting_from_basic_object_and_including_kernel
+      Dir.mktmpdir("warn_test") do |dir|
+        File.write(dir + "/main.rb", "\nrequire 'sub'\n")
+        File.write(dir + "/sub.rb", <<-'RUBY')
+          require 'rubygems'
+          class C < BasicObject
+            include ::Kernel
+            def deprecated
+              warn "This is a deprecated method", uplevel: 2
+            end
+          end
+          C.new.deprecated
+        RUBY
+
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--enable=gems", "-C", dir, "-I", dir, "main.rb")
+        end
+        assert_match(/main\.rb:2: warning: This is a deprecated method$/, err)
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--enable=gems", "-C", dir, "-I", dir, "main.rb")
+        end
+        assert_match(/main\.rb:2: warning: This is a deprecated method$/, err)
+      end
+    end
   end
 
   private


lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -5,21 +5,19 @@
 
   module Kernel
     rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
-
```
deivid-rodriguez, line 8: Please avoid unnecessary whitespace changes. Thanks!

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -5,21 +5,19 @@
...
-    original_warn = method(:warn)
+    original_warn = instance_method(:warn)
 
     remove_method :warn
-
```
deivid-rodriguez, line 12: Same here about whitespace.

deivid-rodriguez:
Looks good to me @eregon.

> I think it's the safest approach with the least side effects.
prepend / extend look nicer, but they don't achieve the same effect and they have various corner cases where they don't work (for modules like Kernel, for classes it would work better).

Regarding this, sure, definitely better to have safer code than to have nicer looking code. Let's add an extra test that fails with the prepend / extend approach to justify it.

lib/rubygems/core_ext/kernel_warn.rb
eregon: I think those were due to RuboCop rules, but we change them, it seems like a good opportunity for me to clean them, as the original author (7153a317c3e545658459d84aeafdad891d3a8a36) did not intend those extra lines.
Is it OK to keep that change?
I can revert that if needed, but IMHO it's more readable this way.

deivid-rodriguez: Let's focus on the fix itself and leave style appeasing tweaks aside.

eregon:
Test added.
On #3985 that test fails with:
```
  1) Failure:
TestGemRequire#test_no_crash_when_overriding_warn_with_warning_module [/home/eregon/code/rubygems/test/rubygems/test_require.rb:710]:
Expected /^warn comes from .+rubygems\/core_ext\/kernel_warn.rb/ to match "main.rb:2: warning: method redefined; discarding old warn\n" +
"Foo Bar\n" +
"warn comes from core\n".
```
That is, the original Kernel#warn is used and not RubyGems version, which is unintended.

lib/rubygems/core_ext/kernel_warn.rb
eregon: Done

test/rubygems/test_require.rb
```diff
@@ -677,6 +677,39 @@ def test_require_bundler_with_bundler_version
...
+            end
+          end
+          C.new.some_method
+        RUBY
+        _, err = capture_subprocess_io do
+          system(*ruby_with_rubygems_in_load_path, "-w", "--disable=gems", "-C", dir, "main.rb")
+        end
+        assert_match(/^Foo Bar\n/, err)
+        assert_match(/^warn comes from/, err)
+        refute_match(%r{^warn comes from .+rubygems/core_ext/kernel_warn.rb}, err)
```
eregon, line 705: The reason for this `refute` vs asserting it comes from core is that's it's not easy to assert it comes from core, because methods defined in Ruby will have a source_location. In Ruby 2.7:
```
$ ruby --disable-gems -e 'p method(:warn).source_location'
["<internal:warning>", 42]
```

eregon:
Should be ready for final review & merge when CI passed now.

deivid-rodriguez:
I'm not sure I understand this very well, but I'm not clear what the end user problem that this fixes is.

* In https://github.com/rubygems/rubygems/pull/2442, the end user problem is clear to me: a `warn` call during a `require` using the ruby 2.5+ `uplevel` parameter would show an unintended location for the warning, differing from the one when using standard `require`. The reason for customizing `warn` was that PR.

* In #2588  the problem is also clear, a SystemStackError.

But this one I'm not totally sure. If not necessary for any end user behaviour, it seems nice to me that the original `Kernel#warn` is used? :thinking:

Could you split the new checks to a separate test, since they are unrelated to `test_no_crash_when_overriding_warn_with_warning_module`, and try to explain this better?

eregon:
In short, it fixes #2588, without changing any other behavior or changing the ancestors of any class.

I wish without RubyGems it would just be a SystemStackError because redefining Warning#warn is IMHO the mistake in the first place (documented in https://github.com/ruby/ruby/pull/3621), but since there is a workaround in Ruby it sounds indeed nice that the RubyGems version of Kernel#warn behaves the same.
The previous version would unintentionally change the receiver for `warn`, which this PR addresses.

I guess your question is mostly is there a case where we'd want RubyGems' `warn` and the receiver is not simple the main object or Kernel.warn? Someone could do `some_object.send(:warn, message)` since `send` is skipped on CRuby at the top-level.
Someone might also pass a value >1 to `uplevel:` in which case it might end up on the `require` frame, when we always want to skip it.

eregon:
@deivid-rodriguez I've updated the test with something more realistic, it should be clearer now.
Essentially, `warn` always wants to skip core library methods (such as the original `require`), and also RubyGems `require` since we believe (in the vast majority of cases) it's not itself the source of the warning (and that showing the location of RubyGems' require would be useless).

BTW on Ruby 3, `warn` already skips `<internal:` core library methods defined in Ruby:
```
$ ruby -ve 'tap { warn "foo", uplevel: 1; p caller(0) }'
ruby 3.0.0dev (2020-09-30T06:55:07Z master ce986b41ca) [x86_64-linux]
-e:1: warning: foo
["-e:1:in `block in <main>'", "<internal:kernel>:90:in `tap'", "-e:1:in `<main>'"]
```
We see warn uses `-e:1` and not `<internal:kernel>:90` as the caller.

It's kind of a shame that we need to redefine `Kernel#warn` when there is existing filtering for `<internal:`.
I think it would be worth proposing a new feature on https://bugs.ruby-lang.org/ like e.g. `Kernel.ignore_for_warnings %r{/rubygems/core_ext/kernel_require\.rb}`. Thoughts?
It would only apply to new Ruby versions so we'd still need to override on older versions though.

eregon:
I didn't notice the lint failed, fixed now.
@deivid-rodriguez OK to merge this when it's green?

eregon:
I got a new idea, and it seems to work on Ruby 3 and on TruffleRuby, so there we wouldn't need to override `Kernel#warn` at all!

The idea is we could set the "file" value for RubyGems' require as starting with `<internal:` through `eval`, and that will result in it being ignored by `Kernel#warn`, like [core library methods](https://gist.github.com/eregon/912e6359e83781c5fa1c638d3768c526).

```ruby
eval <<RUBY, nil, '<internal:skip-me-for-warnings>', __LINE__+1
def mycaller
  foo
end
RUBY

def foo
  warn "WARNING", uplevel: 1
end

mycaller
```

On Ruby 3 and on TruffleRuby it gives:
```
warn.rb:11: warning: WARNING
```
On Ruby 2 (I tried 2.7 and 2.6) it gives:
```
<internal:skip-me-for-warnings>:3: warning: WARNING
```

Anyway, we still need the fallback for Ruby 2, so I think we should first merge this PR.
But this sounds like a better way forward, isn't it?
cc @nobu

eregon:
@deivid-rodriguez The CI passes now, could you merge this?

eregon:
Regarding https://github.com/rubygems/rubygems/pull/3987#issuecomment-704334889 I got confused, `<internal:` is removed by RubyGems' Kernel#warn. With `--disable=gems` it shows `<internal:skip-me-for-warnings>:3` on Ruby 3. I'll probably file a pull request for that.
At least Kernel#warn seems to skip core library methods in Ruby like in this case:
```
$ ruby -v --disable=gems -e 'def deprecated; warn "use X instead", uplevel: 1; end; tap { deprecated; raise "error" };'
ruby 3.0.0preview1 (2020-09-25 master 0096d2b895) [x86_64-linux]
-e:1: warning: use X instead
-e:1:in `block in <main>': error (RuntimeError)
	from <internal:kernel>:90:in `tap'
	from -e:1:in `<main>'
```
(shows `-e:1: warning: use X instead` and not `<internal:kernel>:90: warning: use X instead`)

eregon:
Issue on the CRuby tracker to automatically skip `<internal:` entries, and this could be the way forward to not need to override Kernel#warn at all in RubyGems: https://bugs.ruby-lang.org/issues/17259

deivid-rodriguez:
I rebased this PR, squashed some commits and slightly modified the test, and your commit attribution was accidentally lost in that process. I'm really sorry about that, it was completely unintentional.

eregon:
A bit unfortunate, but not a big deal, thanks for merging.

PR #3988, opened 2020-10-02
Append '.gemspec' extension only when it is not present.

voxik:
This avoids confusing messages such as:

~~~
$ gem build *.gemspec
ERROR:  Gemspec file not found: *.gemspec.gemspec
~~~

Please note that the `TestGemCommandsBuildCommand#test_can_find_gemspecs_without_dot_gemspec`
testcase was probably wrong from the beginning, trying to apply `.gemspec` extension to `@gem.spec_name`, which already contains the extension.

Fixes minor issue mentioned in #3953.

### DIFF
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -79,7 +79,7 @@ Gems can be saved to a specified filename with the output option:
   end
 
   def build_gem(gem_name)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
:+1:

deivid-rodriguez:
Since this fix is independent from the whole #3954 vs #3983 discussion, I'm merging it!

Issue #4012, opened 2020-10-09
`bundler install` should not fail when a native platform gem's required_ruby_version is not met, and a suitable `ruby` platform gem is available


flavorjones:
### Describe the problem as clearly as you can

Given:
- I am running a specific version of Ruby
- a Gem is available as both a `ruby` platform gem and a native (e.g., `x86_64-linux`) platform gem
  - where my specific version of Ruby **does not** meet the requirements set by the **native** platform gem via `spec.required_ruby_version`
  - where my specific version of Ruby **does** meet the requirements set by the **`ruby`** platform gem via `spec.required_ruby_version`

I expect `bundler install` (and/or `gem install`?) to install the `ruby` platform gem.

More specifically, I expect `bundler install` to iterate through the list of Rubygems Platforms for my system, from more-specific to more-general, until it finds a gem that meets all the requirements. It should not give up and raise an error until the list of platforms has been exhausted.

Instead, I see `bundler install` select the native gem, and then fail while resolving dependencies with an error message like:

```
nokogiri-1.11.0.rc3-x86_64-linux requires ruby version < 2.8.dev, >= 2.4, which is incompatible with the current version, ruby 3.0.0p-1
```

### Additional context

Given the work done at https://github.com/rubygems/bundler/pull/7522 to improve platform-specific gem resolution, I expected this to eventually install the matching `ruby` gem rather than give up after failing to install the non-matching native gem. Perhaps this is my misunderstanding of that issue.

As a maintainer of Nokogiri, about to release native gems (see https://github.com/sparklemotion/nokogiri/issues/2075), I'd like very much for this to work as described. I believe the current behavior will cause confusion for Ruby users whenever a new version of Ruby is made available, and that will create additional support work for native gem maintainers.

I can imagine this unfortunate scenario:

1. ruby 3.0.0 final is released in December 2020
2. nokogiri v1.11.0 is simultaneously released with native gems that support ruby `~> 3.0.0`
3. time passes
4. ruby 3.1 final is released in December 2021 (or, more realistically, ruby 3.1.0preview1 is released in September 2021 and people start trying it out)
5. developers everywhere try to upgrade and see the above error message
6. developers open multiple issues on bundler and nokogiri, and post on stackoverflow, and complain on twitter ...
7. adoption of ruby 3.1.0 is slowed or inhibited

[...]

Issue #3200, opened 2017-10-29
Gem in gemspec without ruby platform version is resolved to a wrong platform


toy:
I've stumbled upon an issue with using bundler for developing a gem that has a dependency on `sys-proctable` gem. The issue seems to be caused by two characteristics of `sys-proctable` gem: it has only platform specific versions and two of them (darwin and freebsd) have additional dependency on ffi gem. I was able to reproduce the behaviour on freebsd (10.3) and osx (darwin 16). 

Given `Gemfile`:
```ruby
source 'https://rubygems.org'
gemspec
```
and `test.gemspec`:
```ruby
Gem::Specification.new do |s|
  s.name        = 'test'
  s.version     = '1.0.0'
  s.summary     = 'test summary'
  s.authors     = ['Max Mustermann']
  s.add_dependency 'sys-proctable', '~> 1.0'
end
```
and running `bundle` on freebsd, produces:
```
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/.
Resolving dependencies...
Unable to use the platform-specific (universal-freebsd) version of sys-proctable (1.1.5) because it has different dependencies from the universal-aix5 version. To use the platform-specific version of the gem, run `bundle config specific_platform true` and install again.
Using bundler 1.15.4
Fetching sys-proctable 1.1.5 (universal-aix5)
Installing sys-proctable 1.1.5 (universal-aix5)
Using test 1.0.0 from source at `.`
Bundle complete! 1 Gemfile dependency, 3 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
```

Running `bundle config specific_platform true`, removing the `sys-proctable` gem from the system and repeating `bundle` will remove the warning, but will still fetch the `universal-aix5` platform specific gem.

deivid-rodriguez:
Hello! :wave: :wave: 

I run Samuel's specs and it passes now! So it sounds that we might've fixed this with the recent multiplatform improvements! :tada: 

@toy Could you verify that this is now working fine?

toy:
@deivid-rodriguez Using bundler 2.2.0.rc.2 if I change the dependency to `s.add_dependency 'sys-proctable', '~> 1.1.0'` (as sys-proctable `1.2.x` doesn't have platform specific versions) I still see the warning and universal-aix5 version is installed.

deivid-rodriguez:
[...]

PR #4015, opened 2020-10-13
Enable `specific_platform` by default

deivid-rodriguez:
To properly support gems providing native platform versions with different dependencies than their pure ruby counterparts.

Nokogiri will start doing this soon. In particular, for each version, the pure ruby variant will not have a `required_ruby_version` upper bound, but the native variant will. Bundler should always install a valid bundle in this case.

Closes https://github.com/rubygems/rubygems/issues/3200.
Closes https://github.com/rubygems/rubygems/issues/4012.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -559,10 +559,7 @@ module Bundler
     end
 
     def current_platforms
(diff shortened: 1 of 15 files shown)


deivid-rodriguez:
Let's do this!

PR #4017, opened 2020-10-15
Clean built extensions after install

### LINKED ISSUE #3958: Cleanup object files on gems with extensions
I've just noticed that gems with compiled extensions use a lot of disk space, most of it are object files that, as far as I understand, are not needed on runtime.

For example, on sassc-2.4.0 does use (on my linux computer) 114.3 MiB, where extension code is 110.9 MiB but the important files (as far as I understand) are just `libsass`, with 1,7 MB and `libsass.so` 3,2MB. That is about 108 MB of extra space that could also make slower deployments, when generating bundles and testing when caching gems on CI, or simply on developers machines that duplicate gems on different environments (per project GEM_HOME).

```
$ gem env version 3.0.3
```

What are your thoughts on this?

eloyesp:
This should fix #3958, it is based on a comment of @simi there.

It just adds a `clean` step after `install` to remove object files.

### DIFF
--- a/bundler/lib/bundler/stub_specification.rb
+++ b/bundler/lib/bundler/stub_specification.rb
@@ -16,8 +16,17 @@ module Bundler
       # Stub has no concept of source, which means that extension_dir may be wrong
       # This is the case for git-based gems. So, instead manually assign the extension dir
       return unless source.respond_to?(:extension_dir_name)
-      unique_extension_dir = [source.extension_dir_name, File.basename(full_gem_path)].uniq.join("-")
-      extension_dir = File.join(extensions_dir, unique_extension_dir)
+      extension_dir_name = source.extension_dir_name
+      unique_extension_dir = [extension_dir_name, File.basename(full_gem_path)].uniq.join("-")
+      extension_dir = File.join(
+        stub.extensions_dir,
+        '..', '..', '..', # Fix stub.extension_dir
+        extension_dir_name,
+        "extensions",
+        Gem::Platform.local.to_s,
+        Gem.extension_api_version,
+        unique_extension_dir
+      )
       stub.extension_dir = File.expand_path(extension_dir)
     end
 
--- a/bundler/lib/bundler/stub_specification.rb
+++ b/bundler/lib/bundler/stub_specification.rb
@@ -16,8 +16,9 @@ module Bundler
       # Stub has no concept of source, which means that extension_dir may be wrong
       # This is the case for git-based gems. So, instead manually assign the extension dir
       return unless source.respond_to?(:extension_dir_name)
-      path = File.join(stub.extensions_dir, source.extension_dir_name)
-      stub.extension_dir = File.expand_path(path)
+      unique_extension_dir = [source.extension_dir_name, File.basename(full_gem_path)].uniq.join("-")
+      extension_dir = File.join(extensions_dir, unique_extension_dir)
+      stub.extension_dir = File.expand_path(extension_dir)
     end
 
     def to_yaml
--- a/bundler/spec/install/gems/native_extensions_spec.rb
+++ b/bundler/spec/install/gems/native_extensions_spec.rb
@@ -141,6 +141,8 @@ RSpec.describe "installing a gem with native extensions" do
       gem "c_extension_two", :git => #{lib_path("gems").to_s.dump}
     G
 
+    run 'puts $LOAD_PATH.grep(/extensions/).each { |path| raise "Invalid LOAD_PATH: #{ path }" unless File.directory?(path) }'
+
     run "Bundler.require; puts CExtension_one.new.value; puts CExtension_two.new.value"
     expect(out).to eq("one\ntwo")
   end
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -222,4 +222,8 @@ EOF
 
     destination
   end
+
+  def clean_intermediate_files
+    FileUtils.rm Dir["#{@spec.gem_dir}/**/*.{so,o,bundle}"]
+  end
 end
--- a/lib/rubygems/ext/cargo_builder.rb
+++ b/lib/rubygems/ext/cargo_builder.rb
@@ -139,9 +139,7 @@ class Gem::Ext::CargoBuilder < Gem::Ext::Builder
   end
 
   def final_extension_path(dest_path)
-    dylib_path = cargo_dylib_path(dest_path)
-    dlext_name = "#{spec.name}.#{makefile_config("DLEXT")}"
-    dylib_path.gsub(File.basename(dylib_path), dlext_name)
+    File.join dest_path, "#{spec.name}.#{makefile_config("DLEXT")}"
   end
 
   def cargo_dylib_path(dest_path)
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -178,6 +178,11 @@ module Gem::InstallUpdateOptions
                "Suggest alternates when gems are not found") do |v,o|
       options[:suggest_alternate] = v
     end
+
+    add_option(:"Install/Update", "--dirty",
+               "Leave build artifacts on the gem directory") do |v,o|
+      options[:dirty] = v
+    end
   end
 
   ##
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -841,6 +841,7 @@ TEXT
     builder = Gem::Ext::Builder.new spec, build_args
 
     builder.build_extensions
+    builder.clean_intermediate_files
   end
 
   ##
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -841,7 +841,7 @@ TEXT
     builder = Gem::Ext::Builder.new spec, build_args
 
     builder.build_extensions
-    builder.clean_intermediate_files
+    builder.clean_intermediate_files unless options[:dirty]
   end
 
   ##
--- a/test/rubygems/test_gem_ext_cargo_builder.rb
+++ b/test/rubygems/test_gem_ext_cargo_builder.rb
@@ -55,7 +55,7 @@ class TestGemExtCargoBuilder < Gem::TestCase
     end
 
     output = output.join "\n"
-    bundle = File.join(@dest_path, "release/rust_ruby_example.#{RbConfig::CONFIG['DLEXT']}")
+    bundle = File.join(@dest_path, "rust_ruby_example.#{RbConfig::CONFIG['DLEXT']}")
 
     assert_match(/Finished/, output)
     assert_match(/release/, output)
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -26,6 +26,7 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
       -i /install_to
       -w
       --post-install-message
+      --dirty
     ]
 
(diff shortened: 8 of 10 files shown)


ma-6cji (approved):

simi:
1. Some tests do rely on this.
2. I think it would be great to make this configurable somehow and start with this disabled by default.

deivid-rodriguez:
It's probably worth investigating the failing tests first before making a call on how to handle this. The only failing tests are some tests I added very recently to test rubygems custom require behaviour regarding `.rb` vs `.so` selection, and the issue might be perfectly with the tests themselves :shrug:

simi:
Any news in here?

deivid-rodriguez:
I just rebased this PR, and fix the test failure, which was indeed expected and doesn't signal any problem with doing this. I think doing this is safe.

deivid-rodriguez:
Oh wow, let me take back my previous message, this seems to break a lot of Bundler specs, actually. I will have a look.

eloyesp:
As far as I could research, there is one remaining issue in [`Gem::Ext::ExtConfBuilder#build`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L13) that install the gem extensions within the gem itself by seting the `sitedir` argument for make, and making the `clean` task remove the actual installed files. The failure raises when the `full_tmp_dir` is being accessed after being cleaned up.

But I'm not sure why is ExtConfBuilder being used and why the sitedir is set to a tmpdir on the gem folder in that case.

eloyesp:
Ok, I now:

- fixed the failing tests properly (the ones checking the intermediate files instead of the installed ones).
- Added clean step on builder (that is called on install) that is quite simple to understand and have proper tests.
- Added an option `--dirty` that skip that step, just in case there is some gem broken by this change and added tests for it.
- Rebased everything on top of the `master` branch.

As far as I could run the tests everything just pass, but it seems that CI won't run without approval.

deivid-rodriguez:
Thanks for working on this @eloyesp, Since this would be your first contribution to this repo, CI needs manual approval. Just did that!

I'm not sure about `--dirty` option? I can't think of a case why someone would need artifacts to be left around, so I rather not commit to a new public CLI option that may not be used by anyone.

eloyesp:
The  `dirty` option is added because of a possible gem that may depend on the build artifacts, it could be replaced by an environment variable, the idea is simply to give an easy workaround for a broken gem.

Also, I were reviewing the failures and it seems that the cargo builder depends on the build artifacts (or at least the tests do), I cannot reproduce it here because I could not make cargo run here.

There is a single failure for bundler, but could not reproduce the issue (the test pass locally).

And there were two warnings because of an unused variable, I've fixed those.

deivid-rodriguez:
> As far as I could research, there is one remaining issue in [Gem::Ext::ExtConfBuilder#build](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L13) that install the gem extensions within the gem itself by seting the sitedir argument for make, and making the clean task remove the actual installed files. The failure raises when the full_tmp_dir is being accessed after being cleaned up.

> But I'm not sure why is ExtConfBuilder being used and why the sitedir is set to a tmpdir on the gem folder in that case.

I don't understand this case well, but I'd like to because using `make clean` feels... cleaner 😄. So, with the previous `make clean` approach, there was a single test failure, correct? Can you link push a separate branch with that approach and the failures it creates, I'd like to have a look.

eloyesp:
The problem with that approach is that the make method is a class method with many parameters, it is already doing too much and having lots of smells (in the [reek](https://github.com/troessner/reek) sense), adding the clean step there makes it much harder to debug and understand what happens.

While adding the single "clean" step in the array is clean enough adding that as an optional step adds lots of noise and making that cleaner would require a real refactor (or adding the seventh argument to that method).

simi:
I don't see problem with 7th optional argument. We need to maximize backward compatibility, this is the price we need to pay sometimes. Refactoring would be welcomed as well (not changing behaviour for now, just restructuring the code). All `ext` part of rubygems is a little confusing IMHO.

:thinking: I don't think `FileUtils.rm Dir["#{@spec.gem_dir}/**/*.{so,o,bundle}"]` is good approach, since extension can do actually a lot of different stuff (create various files) on different platforms (when extension can differ). What about linux static builds? `*.a` What about rules files `*.d`? What about Mac OS dynlibs aka `*.dylib`? We should go with `make clean` as standard task to cleanup environment. That is supported by `mkmf`, it seems to be supported by next release of rust `mkmf`, ...

eloyesp:
I understand that doing an actual removal with `FileUtils.rm` is not beautiful, but it have a clear advantage it is easier to understand and extend, that is this should get rid of all that you listed:

    FileUtils.rm Dir["#{@spec.gem_dir}/**/*.{so,o,bundle,a,d,dylib}"]

Refactoring everything on `ext` is actually an important task, but it is impossible for me (because I lack the time and the confidence with the code). The main issue with such a refactor is that there is a lot of coupling between the tests and the code, making any refactor require rewrite most of the test suite.

I were exploring the option of simply adding a "clean" step, and the least ugly way to get that working was using a global variable or something like that, that would make the needed refactor even harder to implement (adding more technical debt).

Once the refactor is actually done, I think that the best approach would be to use "make clean" as you all recomend, it is obviously cleaner that way.

simi:
I understand, but I still think we just can't remove random files we don't own in RubyGems. Those files are owned (created) by make tasks and they should be cleaned using the same owner.

For now, to keep it simple, what about to extend this line with post optional task?

https://github.com/rubygems/rubygems/blob/a09fdd10145d065d7f45917cd72a5a2f16720511/lib/rubygems/ext/builder.rb#L43

```ruby
["clean", "", "install", ENV['RUBYGEMS_POST_MAKE_TASK']].compact.each do |target|
```

Would that be enough for now for opt-in behaviour of running `make clean` doing something like `RUBYGEMS_POST_MAKE_TASK=clean gem install xyzl`? Would you mind to give it a try?

eloyesp:
Ok, I've created another pull-request with that implementation, it uses an ENV variable (that is another way to use global state) making testing and refactoring even harder.

deivid-rodriguez:
From what I understand, there's two discussions going on:

* Whether to `make clean` or remove specific files. I agree with @simi here.
* Whether to make this opt in, or not. I'm not necessarily against making this opt-in, but I don't know of any gem relying on intermediate build artifacts present in the final installation directory. Maybe one such gem exists, but I tend towards potentially breaking some users of this potential gem in favor of benefiting the rest of the community by default. Apparently the cargo builder relies on intermediate build artifacts being present in the installation folder? Can we look a bit more into that? Maybe it's just a problem with the tests?

eloyesp:
Yes and no, I do think as everyone here that using "make clean" is better than removing files using a glob, but I prefer the last given the current implementation of the extension building process complexity. That is, adding more complexity there adds more technical debt and makes it even harder to refactor the code.

So, I'm not against using "make clean", I'm against the environment variable. That is the trade-off I'm talking about.

Regarding the opt-in or opt-out, I also think that opt-out is the way to go, but it is quite important to make that opt-out very clear, that is why I've added the `dirty` flag there, it sounds bad.

simi:
> Ok, I've created another pull-request with that implementation, it uses an ENV variable (that is another way to use global state) making testing and refactoring even harder.

That was just a quick idea to roll it out quickly for testing. Later we can integrate this in better way and turn it opt-out (enabled) by default. If you don't like this idea, we can continue exploring.

simi:
@deivid-rodriguez cargo/ruby connection did this recently -https://github.com/oxidize-rb/rb-sys/pull/107.

eloyesp:
Ok, I'm almost sure I found the bug with cargo builder... It seems that when installing it adds a "release" subfolder that makes require fail.

```
# doing this manually in the middle of the test make it pass
cp tmp/test_rubygems_20221201-5322-h1rss4/gemhome/extensions/x86_64-linux/3.1.0-static/rust_ruby_example-0.1.0/release/rust_ruby_example.so tmp/test_rubygems_20221201-5322-h1rss4/gemhome/extensions/x86_64-linux/3.1.0-static/rust_ruby_example-0.1.0/
```

I'm working on a proper fix.

eloyesp:
That last commit should fix most of the failing tests (I could not reproduce the bundler related failure locally so I hope this fix that issue).

Now, we can:

- leave that clean task as is, and improve it in a later iteration with more information.
- add a little bit of love to that glob with the information we already have.
- try different strategies like running "make clean", something like a partial pristine (remove anything not listed in the gem itself). (with the current implementation this would require adding ~50 LOC)

My recommendation would be the first or the second one, leaving the improvements to a later iteration. I think that after a proper refactor the `make clean` option should be a one-liner, but adding it now will just make that refactor even harder (as it will need to duplicate most of the `build_extensions` related code.)

eloyesp:
This extra line makes the test fail in a much clear way:

``` ruby
run 'puts $LOAD_PATH.grep(/extensions/).each { |path| raise "Invalid LOAD_PATH: #{ path }
  " unless File.directory?(path) }'
```

eloyesp:
Found the issue, but I will not be able to fix it in the near future. The problem is that bundler when looking for performance is breaking the DRY rule duplicating the Specification using different implementations on different stages.

While the gem is being installed the extensions_path is defined in [rubygems_ext.rb](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/rubygems_ext.rb#L67) monkeypatching Gem::Specification.

But afterwards on [runtime](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/runtime.rb#L24) those specs are [a special collection](https://github.com/rubygems/rubygems/blob/771c94e14d6c86c8df3d1df0d2679ed59316252a/bundler/lib/bundler/spec_set.rb) of [stub specifications](https://github.com/rubygems/rubygems/blob/771c94e14d6c86c8df3d1df0d2679ed59316252a/bundler/lib/bundler/stub_specification.rb) that should not be confused with [stub specifications](https://github.com/rubygems/rubygems/blob/771c94e14d6c86c8df3d1df0d2679ed59316252a/lib/rubygems/stub_specification.rb).

The problem is that there is a mismatch regarding what the `extension_dir` actually is.

I've been digging several hours in this, but I cannot fix that as it requires much more knowledge about how to debug this.

eloyesp:
I mostly sure I fixed the bundler bug, but it would need a proper refactor to reuse instead of duplicate. I've just changed the stub implementation to match.

eloyesp:
It seems that _some_ extensions are being installed in different paths, so now that I modify the "extension_path" to make it work for one gem, it is broken for other gem, may be there is a better way to modify the "extension_path" to make everyone happy.

eloyesp:
I'm still working on this bug, it seems that bundler installs git sourced gems extensions on two different paths depending something (not sure what).

The two cases are on the specs, but they are not specified, so the specs pass because ruby finds the build artifacts (`$LOAD_PATH` is set to a non-existing directory but that is ok for the specs).

The test on `spec/install/gems/native_extensions_spec.rb:93` install extensions on something like:

    #{ stub.base_dir }/#{ source.extension_dir_name }/extensions/etc

While the test `./spec/install/gemfile/git_spec.rb:1203` installs the extensions on:

    #{ stub.base_dir }/extensions/etc

I could not find where is the difference, what makes this case use different folder structure. But without fixing this bug, adding the `clean artifacts` feature on rubygems might break some bundler setups with git gems.

I would greatly appreciate some help here. Regards.

deivid-rodriguez:
@eloyesp I'll try to have a look and see if I can figure this out. Thanks so much for your persistence on this issue, it's of great help!

deivid-rodriguez:
Closing in favor of #6133, let's continue the discussion there.

Issue #3984, opened 2020-10-01
query command deprecation clarification


hurricup:
I'm having a problem or would like to suggest feature. (more like request for clarification)

My current problem is deprecation of `query` command without describing alternative.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
3.2.0.rc.1
```

We have a deprecation warning for the query command, but [documentation](https://guides.rubygems.org/command-reference/#gem-query) does not say anything about deprecation or discouraging using it. Also, warning provides no information about what to use instead.

deivid-rodriguez:
You're right!

We deprecated the `query` command in favour of either `gem search` or `gem list`, dependending on whether you're looking for remote gems, or gems installed on your system. There was an old note in `gem query` that said:

> You should really use the list and search commands instead.  This command is too hard to use.

And we decided the note made sense and deprecated it.

We should:

* Add a note to `guides.rubygems.org` mentioning that the command was deprecated (and in which version).
* Improve the deprecation message to point to the alternatives.

PR #4021, opened 2020-10-16
Suggest alternatives in `gem query` deprecation

landongrindheim:
##  What was the end-user or developer problem that led to this PR?
`gem query` was deprecated in rubygems 3.2. Prior to the command being deprecated, there was an injunction to use `list` or `search`. While the `query` command is going away, the usefulness of this suggestion is still present, as pointed out in #3984.

[@deivid-rodriguez's comment](https://github.com/rubygems/rubygems/issues/3984#issuecomment-702048149) on the [linked issue](#3984) suggested adding suggestions for the now-deprecated `query` command, along with adding information about the deprecation to rubygems.org.

Note: since the described problem requires changes in two repositories, this is the first of two pull requests to address this issue.

### DIFF
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -9,6 +9,14 @@ class Gem::Commands::QueryCommand < Gem::Command
 
   include Gem::QueryUtils
 
+  alias warning_without_suggested_alternatives deprecation_warning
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
This looks good to me, thanks!

PR #4023, opened 2020-10-20
Omit deprecated commands from command help output

### LINKED ISSUE #3984: query command deprecation clarification

landongrindheim:
`gem query` was deprecated in rubygems 3.2. It was noted in #3984 that there should be a deprecation warning in the CLI (done in #4021) and a message in [the guides](https://guides.rubygems.org/command-reference/#gem-query).

A PR was opened to address the message (rubygems/guides#270) in Guides, but it [was pointed out](https://github.com/rubygems/guides/pull/270#issuecomment-712832643) that the file being changed is auto-generated. This change leans into that auto-generation and should make rubygems/guides#270 redundant.

Closes #3984 

Add the deprecation note to the command's summary so that it can be relied upon for auto-generating the documentation. This note will display near the heading in the guides.

### DIFF
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -32,7 +32,7 @@ class Gem::Commands::QueryCommand < Gem::Command
     add_query_options
   end
 
-  def description
+  def description # :nodoc:
     <<-EOF
 The query command is the basis for the list and search commands.
 
--- a/lib/rubygems/commands/query_command.rb
+++ b/lib/rubygems/commands/query_command.rb
@@ -31,4 +31,13 @@ class Gem::Commands::QueryCommand < Gem::Command
 
     add_query_options
   end
+
+  def description
+    <<-EOF
+The query command is the basis for the list and search commands.
+
+You should really use the list and search commands instead.  This command
+is too hard to use.
+    EOF
+  end
 end
--- a/lib/rubygems/commands/help_command.rb
+++ b/lib/rubygems/commands/help_command.rb
@@ -332,6 +332,8 @@ platform.
     @command_manager.command_names.each do |cmd_name|
       command = @command_manager[cmd_name]
 
+      next if command.deprecated?
+
       summary =
         if command
           command.summary
--- a/lib/rubygems/query_utils.rb
+++ b/lib/rubygems/query_utils.rb
@@ -57,15 +57,6 @@ module Gem::QueryUtils
     "--local --name-matches // --no-details --versions --no-installed"
   end
 
-  def description # :nodoc:
-    <<-EOF
-The query command is the basis for the list and search commands.
-
-You should really use the list and search commands instead.  This command
-is too hard to use.
-    EOF
-  end
-
   def execute
     gem_names = Array(options[:name])
 
--- a/test/rubygems/test_gem_commands_help_command.rb
+++ b/test/rubygems/test_gem_commands_help_command.rb
@@ -40,13 +40,26 @@ class TestGemCommandsHelpCommand < Gem::TestCase
 
     util_gem 'commands' do |out, err|
       mgr.command_names.each do |cmd|
-        assert_match(/\s+#{cmd}\s+\S+/, out)
+        unless mgr[cmd].deprecated?
+          assert_match(/\s+#{cmd}\s+\S+/, out)
+        end
       end
 
       if Gem::HAVE_OPENSSL
         assert_empty err
 
         refute_match 'No command found for ', out
+      end
+    end
+  end
+
+  def test_gem_help_commands_omits_deprecated_commands
+    mgr = Gem::CommandManager.new
+
+    util_gem 'commands' do |out, err|
+      deprecated_commands = mgr.command_names.select {|cmd| mgr[cmd].deprecated? }
+      deprecated_commands.each do |cmd|
+        refute_match(/\s+#{cmd}\s+\S+/, out)
       end
     end
   end


lib/rubygems/commands/query_command.rb
```diff
@@ -31,4 +37,13 @@ def initialize(name = 'query',
 
     add_query_options
   end
+
+  def description
```
deivid-rodriguez, line 41: We should keep the `# :nodoc:` for consistency with the other command descriptions.

```suggestion
  def description # :nodoc:
```

lib/rubygems/commands/query_command.rb
```diff
@@ -9,6 +9,13 @@ class Gem::Commands::QueryCommand < Gem::Command
 
   include Gem::QueryUtils
 
+  NAME    = 'query'
+  SUMMARY = <<~SUMMARY
+    Query gem information in local or remote repositories
+
+    **Note:** `query` is deprecated as of RubyGems 3.2
+  SUMMARY
```
deivid-rodriguez, line 17: Summaries should be one liners, otherwise `gem help commands` output is a bit messed up:

```
$ gem help commands
GEM commands are:

    build                  Build a gem from a gemspec
    cert                   Manage RubyGems certificates and signing settings
    check                  Check a gem repository for added or missing files
    cleanup                Clean up old versions of installed gems
    contents               Display the contents of the installed gems
    dependency             Show the dependencies of an installed gem
    environment            Display information about the RubyGems environment
    fetch                  Download a gem and place it in the current directory
    generate_index         Generates the index files for a gem server directory
    help                   Provide help on the 'gem' command
    info                   Show information for the given gem
    install                Install a gem into the local repository
    list                   Display local gems whose name matches REGEXP
    lock                   Generate a lockdown list of gems
    mirror                 Mirror all gem files (requires rubygems-mirror)
    open                   Open gem sources in editor
    outdated               Display all gems that need updates
    owner                  Manage gem owners of a gem on the push server
    pristine               Restores installed gems to pristine condition from
                           files located in the gem cache
    push                   Push a gem up to the gem server
    query                  Query gem information in local or remote repositories
                           
                           **Note:** `query` is deprecated as of RubyGems 3.2
    rdoc                   Generates RDoc for pre-installed gems
    regenerate_binstubs    Re run generation of executable wrappers for gems.
    search                 Display remote gems whose name matches REGEXP
    server                 Documentation and gem repository HTTP server
    signin                 Sign in to any gemcutter-compatible host. It defaults
                           to https://rubygems.org
    signout                Sign out from all the current sessions.
    sources                Manage the sources and cache file RubyGems uses to
                           search for gems
    specification          Display gem specification (in yaml)
    stale                  List gems along with access times
    uninstall              Uninstall gems from the local repository
    unpack                 Unpack an installed gem to the current directory
    update                 Update installed gems to the latest version
    which                  Find the location of a library file you can require
    yank                   Remove a pushed gem from the index

For help on a particular command, use 'gem help COMMAND'.

Commands may be abbreviated, so long as they are unambiguous.
e.g. 'gem i rake' is short for 'gem install rake'.
```

deivid-rodriguez (commented):
I was reviewing this, but then I had a maybe better idea. What if, instead, we change the script in https://github.com/rubygems/guides/ to ignore deprecated commands and don't list them in the command reference section at all. I think, once we deprecate a command, it's no worth documenting it at all?

Thoughts?

landongrindheim:
> What if, instead, we change the script in https://github.com/rubygems/guides/ to ignore deprecated commands and don't list them in the command reference section at all. I think, once we deprecate a command, it's no worth documenting it at all?

@deivid-rodriguez This makes sense to me -- why hold the user's hand when the intent is to get them to stop using the functionality? I've poked around online for approaches to documenting deprecated functionality. This is what I've seen so far:

* Ruby stopped documenting `Fixnum` in [v2.4](https://ruby-doc.org/core-2.4.0/Fixnum.html) (It was present in [v2.3](https://ruby-doc.org/core-2.3.0/Fixnum.html))
* Ruby still [documents `safe_level`](https://docs.ruby-lang.org/en/2.6.0/ERB.html) as an argument to `ERB.new`, even though it was [deprecated in v2.6](https://www.ruby-lang.org/en/news/2018/12/25/ruby-2-6-0-released/)
* ActiveRecord still [documents `update_attributes`](https://api.rubyonrails.org/classes/ActiveRecord/Persistence.html#method-i-update_attributes) (with aliasing) even though it has been deprecated.
* Elixir documents [the Behaviour module](https://hexdocs.pm/elixir/Behaviour.html), noting that it has been deprecated.

You're doing the heavy lifting here, so I think you should feel free to choose the direction you'd like this to go 🙂

If this PR can be helpful, (say, adding `**deprecated**` to the summary _instead of_ the newline which breaks functionality with `gem help`), please let me know and I'll adjust accordingly.

lib/rubygems/commands/query_command.rb
landongrindheim: Good call! Sorry for the miss on this 😬 

I'm happy to make adjustments here. I think they'll depend on the direction you want to take this. A couple alternatives that make sense to me:
* `gem query`'s summary becomes `Query gem information in local or remote repositories **deprecated**`
* The deprecation message is moved to `#description`
* The deprecation is removed altogether in favor of not rendering documentation for deprecated commands
  * Should `gem help commands ` be consistent with the guides and skip deprecated commands?
  * Should there be mention of the deprecation anywhere other than in CLI output upon use?

Please let me know the direction you'd like this to take and I'll follow suit 😄

deivid-rodriguez:
Thanks for the research! Yeah, that's exactly what I was thinking. If we intend users to stop using a command, not documenting it sounds like a good step in that direction.

> The deprecation is removed altogether in favour of not rendering documentation for deprecated commands

I'd go with this one.

>  Should gem help commands be consistent with the guides and skip deprecated commands?

Yes, that's sounds like the right way forward to me.

> Should there be mention of the deprecation anywhere other than in CLI output upon use?

I don't think so, just warning on usage and recommending alternatives seems fine to me.

landongrindheim:
@deivid-rodriguez What do yo think of this approach?

lib/rubygems/commands/query_command.rb
```diff
@@ -17,9 +22,8 @@ def deprecation_warning
     alert_warning message unless Gem::Deprecate.skip
   end
 
-  def initialize(name = 'query',
-                 summary = 'Query gem information in local or remote repositories')
-    super name, summary,
+  def initialize
+    super NAME, SUMMARY,
```
deivid-rodriguez, line 26: These changes are not actually used in the end, right? Let's revert them to keep the structure of command files consistent.

lib/rubygems/commands/query_command.rb
```diff
@@ -31,4 +35,13 @@ def initialize(name = 'query',
...
   end
+
+  def description # :nodoc:
+    <<~DESCRIPTION
+      The query command is the basis for the list and search commands.
+
+      You should really use the list and search commands instead.  This command
+      is too hard to use.
+    DESCRIPTION
+  end
```
deivid-rodriguez, line 46: Moving the description to the command itself seems like a good idea even if no longer necessary for the fix. Let's keep the previous `EOF` separator though. `DESCRIPTION` looks better but I prefer to update that consistently for all commands and to enforce good names when doing it with `rubocop`.

deivid-rodriguez (commented):
The approach looks good! Only added a couple of minor comments.

Could you squash everything into two commits, one that moves the description to the query command, and the other one that implements "removing" deprecated commands from help output?

lib/rubygems/commands/query_command.rb
landongrindheim: It's kind of a mix, TBH. `QueryCommand` is the only command that has arguments given to `#initialize`.  The rest of the commands have a structure similar to:
```ruby
  def initialize
    super 'open', 'Open gem sources in editor'
```

Unless I hear otherwise from you, I'll conform to the style of the other commands. This would result in this chunk of code being changed, but it would increase consistency, which seems to be a priority here 😄 If you prefer that they be fully reverted, I'll do that 👍

deivid-rodriguez: Normally my preference is to introduce the minimal set of changes to fix the issue, and do any refactorings or consistency improvements separately. Separated commits are fine though if the changes are related to the fix at hand.

deivid-rodriguez: I'd go with reverting the changes altogether to make the actual fix/improvement of this PR more apparent, and maybe making the command structure consistent on a separate PR.

deivid-rodriguez (approved):
Thanks so much for this fix and for your patience with my picky reviews :smiley:

hsbt (approved):
It's nice feature.

PR #4026, opened 2020-10-21
Make Gem::SystemExitException properly exit with a given code

knu:

The gem command does not properly exit with a non-zero exit code in some cases.

I found gem update --system did not properly return a non-zero exit code when it failed with a permission error.

```
$ gem update --system; echo $?
Updating rubygems-update
Successfully installed rubygems-update-3.1.4
Installing RubyGems 3.1.4
ERROR:  While executing gem ... (Errno::EACCES)
   Permission denied @ rb_sysopen - /usr/local/lib/ruby/site_ruby/2.6.0/ubygems.rb
0
$ gem --version
3.0.3
```

The cause was in how Gem::SystemExitException initializes itself.  It didn't pass an exit code to the super method.  See the document of SystemExit.new() for details.

### DIFF
--- a/bin/gem
+++ b/bin/gem
@@ -7,12 +7,7 @@
 
 require 'rubygems'
 require 'rubygems/gem_runner'
-require 'rubygems/exceptions'
 
 args = ARGV.clone
 
-begin
-  Gem::GemRunner.new.run args
-rescue Gem::SystemExitException => e
-  exit e.exit_code
-end
+Gem::GemRunner.new.run args
--- a/setup.rb
+++ b/setup.rb
@@ -21,7 +21,6 @@ Dir.chdir File.dirname(__FILE__)
 $:.unshift File.expand_path('lib')
 require 'rubygems'
 require 'rubygems/gem_runner'
-require 'rubygems/exceptions'
 
 Gem::CommandManager.instance.register_command :setup
 
@@ -31,8 +30,4 @@ if ENV["GEM_PREV_VER"]
 end
 args.unshift 'setup'
 
-begin
-  Gem::GemRunner.new.run args
-rescue Gem::SystemExitException => e
-  exit e.exit_code
-end
+Gem::GemRunner.new.run args
--- a/test/rubygems/test_exit.rb
+++ b/test/rubygems/test_exit.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+require 'rubygems'
+
+class TestExit < Gem::TestCase
+  def test_exit
+    system(*ruby_with_rubygems_in_load_path, "-e", "raise Gem::SystemExitException.new(2)")
+    assert_equal 2, $?.exitstatus
+  end
+end
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -561,6 +561,7 @@ test/rubygems/ssl_key.pem
 test/rubygems/test_bundled_ca.rb
 test/rubygems/test_config.rb
 test/rubygems/test_deprecate.rb
+test/rubygems/test_exit.rb
 test/rubygems/test_gem.rb
 test/rubygems/test_gem_available_set.rb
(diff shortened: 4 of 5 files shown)


knu:
The build failure with TruffleRuby seems to be due to an incompatibility with MRI.
I've filed an issue: https://github.com/oracle/truffleruby/issues/2128

deivid-rodriguez (commented):
Hi ~@kou~@knu (sorry for the unintentional ping @kou)! Thanks for the fix!

I think the actual cause of your issue was actually fixed by https://github.com/rubygems/rubygems/commit/be4c79030d78e48b25cdb17df8fd9145c5cb93e6. Which should probably be changed to `exit($?.exitstatus)` instead of `abort`, by the way.

However, your change makes sense to me and I think it allows us to remove the following `rescue`'s?

https://github.com/rubygems/rubygems/blob/fac8d592f98ce94af2ea258a03515b28ba44ab87/bin/gem#L14-L18

https://github.com/rubygems/rubygems/blob/fac8d592f98ce94af2ea258a03515b28ba44ab87/setup.rb#L35-L39

knu:
You may be right.  I didn’t actually reproduce the EACCES case with RubyGems > 3.0.3, so the problem may not be there in the latest version.  Thanks for the info.

deivid-rodriguez:
Could you add the suggested cleanup to this PR if it makes sense to you? And also squash everything to one commit since you're at it?

knu:
@deivid-rodriguez Sure, done.  This can cause a small regression with TruffleRuby that the exit code becomes 1 where it should be >1, but it will hopefully be fixed on their side.

deivid-rodriguez:
You're right!

Do you think it's worth keeping the behaviour on truffleruby until the issue is fixed there? I think it should be a small enough issue so that nobody will notice, but I feel guilty knowingly introducing a bug on a supported platform, even if small :sweat_smile:.

knu:
At least an issue is filed and it is linked to this issue, and a command failure is still a failure, so I don't think you should feel so guilty. 😇

deivid-rodriguez:
Sure, it just bothers me a bit that in the end I don't think this PR is fixing an actual issue, only improving the implementation, but it is potentially introducing an issue for a truffleruby user relying on specific exit codes of rubygems errors. I'm aware that such a user might not exist, but still :man_shrugging:.

deivid-rodriguez:
I think at this point it's perfectly fine to drop support for truffleruby 20.3, we already stopped testing with it. This issue is fixed since truffleruby 21.0, so it should be all good.

@knu can you rebase this PR?

test/rubygems/test_exit.rb
```diff
@@ -0,0 +1,11 @@
+require 'rubygems/test_case'
+require 'rubygems'
+
+class TestExit < Gem::TestCase
+  def test_exit
+    skip "https://github.com/oracle/truffleruby/issues/2128" if RUBY_ENGINE == "truffleruby"
```
deivid-rodriguez, line 6: This skip can be removed now.

bin/gem
```diff
@@ -11,8 +11,4 @@ require 'rubygems/exceptions'
 
 args = ARGV.clone
 
-begin
-  Gem::GemRunner.new.run args
-rescue Gem::SystemExitException => e
```
deivid-rodriguez, line 16: Since we're not directly using this exception constant, I think the require above for `rubygems/exceptions` can also be removed.

setup.rb
```diff
@@ -32,8 +32,4 @@
 end
 args.unshift 'setup'
 
-begin
-  Gem::GemRunner.new.run args
-rescue Gem::SystemExitException => e
```
deivid-rodriguez, line 37: Same comment about no longer needing to require `rubygems/exceptions`.

simi:
:+1:

deivid-rodriguez:
Rebased and comments addressed!

Issue #4027, opened 2020-10-22
Bundler and RubyGems install directory mismatch


felipec:
Many people don't install to the default directory, but to the user directory (--user-install), in fact, it's the default on many distributions.

Unfortunately this causes a mismatch between bundler and gem, since bundler doesn't read `gemrc` configuration files.

This problem has been explored in many issues: rubygems/bundler#710, rubygems/bundler#2565, rubygems/bundler#2667, chef/chef-dk#148. No satisfactory resolution from the development team.

I took the time to develop a simply and easy solution myself that should absolutely work for everyone, an environment variable: GEM_USER_INSTALL, which just works ™. And thus I am creating yet another issue ticket.

hsbt:
> Many people don't install to the default directory, but to the user directory (--user-install), in fact, it's the default on many distributions.

I don't think so. The users who are using the default package provided many of Linux distributions are minority. also see https://rails-hosting.com/2020/#ruby-rails-version-updates

I'm Ruby programmer over the 20+ years. I and my company also didn't use it. We use the custom runtime build by like ruby-build.

> No satisfactory resolution from the development team.

This comment is a little offensive.

felipec:
@hsbt 

> I don't think so. The users who are using the default package provided many of Linux distributions are minority.

Really? Is that why 64% of Arch Linux users have [installed ruby gems](https://pkgstats.archlinux.de/packages#query=rubygems)? Is that also why Debian packages rubygems directly into the `ruby` package?

> https://rails-hosting.com/2020/#ruby-rails-version-updates

Rails developers != Ruby users.

> This comment is a little offensive.

You think reality is offensive?

deivid-rodriguez:
Whatever the popularity of Linux distributions is, I agree we could do better here. What about honouring `--user-install` if bundler is installing to system gems (i.e., no custom `BUNDLE_PATH` configuration is given). I think it's reasonable to respect whatever rubygems considers as the preferred location to install system gems.

felipec:
@deivid-rodriguez I also created a patch for that, but it's so ugly it seems like a hack. Basically we would need to parse `Gem.configuration` and find the `--user-install` string. I did it only for `gem`, but the same has to be checked for `install`.

```
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -163,6 +163,8 @@ def sources
     end
 
     def gem_dir
[...]

PR #4028, opened 2020-10-22
Add GEM_USER_INSTALL environment variable

felipec:
Many (most) people don't install gems in the default directory, but in
the user's directory (--user-install), this can be configured by
distributions using /etc/gemrc.

However, that doesn't work for bundler, which has no global
configuration.

Many bugs have been opened about this mismatch, but there's no easy
clean solution, except this.

The environment variable GEM_USER_INSTALL overrides the installation
directory (home) so it is the same as the user's home directory, so
there's effectively only one directory (and --[no-]user-install has no
effect).

Distributions can set this variable on, and then finally 'gem install'
and 'bundle install' would use the same location without user
intervention.

Fixes issue #4027.

Signed-off-by: Felipe Contreras <felipe.contreras@gmail.com>

### DIFF
--- a/lib/rubygems/commands/environment_command.rb
+++ b/lib/rubygems/commands/environment_command.rb
@@ -52,6 +52,9 @@ GEM_HOME environment variables. GEM_HOME sets the default repository to
 install into. GEM_PATH allows multiple local repositories to be searched for
 gems.
 
+GEM_USER_INSTALL forces the default repository to be in the user's home
+directory when set (so --[no-]user-install becomes moot).
+
 If you are behind a proxy server, RubyGems uses the HTTP_PROXY,
 HTTP_PROXY_USER and HTTP_PROXY_PASS environment variables to discover the
 proxy server.
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -23,7 +23,8 @@ class Gem::PathSupport
   # hashtable, or defaults to ENV, the system environment.
   #
   def initialize(env)
-    @home = env["GEM_HOME"] || Gem.default_dir
+    default_dir = env["GEM_USER_INSTALL"] ? Gem.user_dir : Gem.default_dir
+    @home = env["GEM_HOME"] || default_dir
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Thanks for your contribution @felipec. However, I don't like this solution. If you're willing to set an environment variable, why not setting `GEM_HOME` or `GEM_PATH` directly?

felipec:
@deivid-rodriguez If I set `GEM_USER_INSTALL`, the gems will be installed to `~/.gem/ruby/2.7.0`, if I set `GEM_HOME` to that, that will work **until** there's an update to the ruby version, or I change the engine, then my gems will be all mixed up.

There's a reason gems are installed to `~/.gem/$engine/$version` by default.

deivid-rodriguez:
Right, sorry, `BUNDLE_PATH` would be the right way. If you set `BUNDLE_PATH=~/.gem`, I think it should be the same as `--user-install` in the rubygems side.

felipec:
@deivid-rodriguez Yes, but I don't want to set two configurations. `bundler` should by default install to the same location as `gem`, with **one** configuration.

deivid-rodriguez:
Sure, you want a clean solution and I give you a workaround. I'd like to think a bit more about what the cleanest solution is.

deivid-rodriguez:
The environment variable approach has the advantage of being opt-in for the majority of users (those developing using version managers, like rbenv, rvm, chruby and so on), but with the possibility of being provided by default to OS ruby packages users (if OS packagers set it by default on their integrations).

This is something I didn't consider initially and I admit feels nice.

Pinging OS packagers @terceiro and @voxik since I'd also like to hear their opinions about this. I believe many OS package users don't use `bundler` at all, but still.

felipec:
> I believe many OS package users don't use `bundler` at all, but still.

They may not **normally** use bundler, but occasionally the may download the source code of a project that uses bundler.

deivid-rodriguez:
Let's have a fresh discussion about this at https://github.com/rubygems/rubygems/issues/4031.

Issue #4031, opened 2020-10-23
Add a way to disable "auto-sudo"


deivid-rodriguez:
### Describe the problem as clearly as you can

Currently bundler detects if it's lacking write permissions on the default `GEM_HOME` and automatically "upgrades privileges" if needed.

It turns out the main users running into these issues (OS ruby package users) don't actually want this behaviour. Instead, they usually prefer "downgrading" to a folder with less permissions than upgrading the running user's privileges.

Also, although this shouldn't influence what we do here and it should just be fixed, there's quite a few issues with the current "auto-sudo" implementation, as can be seeing by searching existing issues.

So, we should consider providing some enhancements to this. Some ideas:

* Provide a way to opting in to the alternative behaviour of "downgrading the folder rather than upgrading the user".
* Ask user when this problem first appear about the preferred behaviour and save it.

Note that rubygems currently does not try to "upgrade to sudo" and provides a way out by configuring `gem: --user-install` in your `~/.gemrc` file. Also, note that we're considering making this behaviour the default if the default folder is not writable in #2847.

I'm particularly interested in @voxik, @terceiro, and @duckinator opionions.

eregon:
Agreed, I think using `sudo` for installing gems is almost always a mistake. I think most users want to just install wherever it's possible with current permissions, or they actually want system packages provided by the package manager.

I actually thought Bundler would install under $HOME if it cannot write to the GEM_DIR, but seems not.

A good alternative for Bundler is to `bundle config set --local path vendor/bundle` (which is unfortunately quite verbose, longer than the deprecated `bundle install --path vendor/bundle`). That won't share gems between applications though, which might or not be wanted.

terceiro:
IMO  installing to $HOME should be the default if the user has no write permissions to the global GEM_DIR. i.e. `gem install` or `bundle install` should by default install to $HOME if the user is not root. There are use cases for installing gems globally, and if you want that, just run `gem install` or `bundle install` as root.

There is one problem with user install, though: programs provided by the installed gems do not end up in `$PATH` without explicit user intervention (.bashrc).

terceiro:
I realized that my opinion above is perhaps too strong, and that may be off-topic in the context of the current discussion (providing a way of opting out of "auto sudo"). I think making bundler respect the rubygems --user-install configuration is a good way of doing it.

duckinator:
I almost exclusively use OS-provided ruby packages, and have definitely had problems with this.

[...]

PR #4032, opened 2020-10-23
add BUNDLE_NO_SUDO env var to disallow sudo

ryantm:

https://github.com/NixOS/nixpkgs/issues/101479

In nixpkgs we create a declarative readonly bundler filesystem environment, so we do not want bundler to try to edit these files, and we do not need it to warn us that it cannot edit them.

Executing bundle with the BUNDLE_NO_SUDO environment variable set will disallow using sudo.

### DIFF
--- a/bundler/lib/bundler/man/bundle-config.1.ronn
+++ b/bundler/lib/bundler/man/bundle-config.1.ronn
@@ -47,6 +47,9 @@ configuration only from the local application.
 Executing bundle with the `BUNDLE_IGNORE_CONFIG` environment variable set will
 cause it to ignore all configuration.
 
+Executing bundle with the `BUNDLE_NO_SUDO` environment variable set will
+disallow using sudo.
+
 Executing `bundle config set --local disable_multisource true` upgrades the warning about
 the Gemfile containing multiple primary sources to an error. Executing `bundle
 config unset disable_multisource` downgrades this error to a warning.
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -257,7 +257,7 @@ module Bundler
     def allow_sudo?
       key = key_for(:path)
       path_configured = @temporary.key?(key) || @local_config.key?(key)
-      !path_configured
+      !path_configured && ENV["BUNDLE_NO_SUDO"].nil?
     end
 
     def ignore_config?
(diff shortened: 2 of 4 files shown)


duckinator:
@deivid-rodriguez what're your thoughts on this PR? I'm in favor of it, but I'm unsure how it would interact with #4031.

ryantm:
@duckinator thanks for the review. All checks are passing now.

deivid-rodriguez (commented):
Hello @ryantm! Thanks for this patch!

I quickly had a look at the screenshot you posted at the related nixpkgs issue, and that would sound like a bug to me? I mean, why would bundler warn about write permissions on a `rails console` invocation? :thinking:

This is partly why I was expecting to have a look at the longstanding "sudo issues" we have opened before making a decision here or in #4031.

Anyways, regardless of that, this PR and the approach look good. I would add some tests (other than the unit test you added) to document the expectations here (what happens when this is enabled and bundler needs to write on a folder where it has no permissions? A permissions error?). I'm also unsure whether we should promote this to a first class setting or whether an environment variable is fine.

ryantm:
@deivid-rodriguez Thanks for the review!

Yeah, it does seem like a bug that commands that are not supposed to effect the disk are calling  `requires_sudo?` https://github.com/rubygems/rubygems/blob/dc133dceca0ea12b1fb0ad492ad3c9a607d1a430/bundler/lib/bundler.rb#L471 and that would be a better direct fix to my problem if we can easily fix it.

I don't know what "first class setting" means. Do you have some documentation about bundler settings I could read?

deivid-rodriguez:
Great, I'll try to fix that issue then since I've seen it reported several times.

Then we can decide whether we still want to move forward with this PR. I think it will still makes sense, but I prefer to have a look at the issue first because I presume I will learn more about the whole sudo thing, so we'll make a more informed decision :)

Regarding bundler settings, yes, have a look at: https://bundler.io/man/bundle-config.1.html. Normally when we make things configurable in bundler, we allow different ways to configure them (environment variables is just one of them).

jsierles:
Hey, was there any progress on figuring this out? It appears to still be happening.

duckinator:
@ryantm @jsierles this is being resolved by removing the "auto-sudo" functionality entirely (see #5888).

deivid-rodriguez:
Yes, thanks so much for starting this anyways @ryantm!

ryantm:
Great, that should fix my issue!

PR #4033, opened 2020-10-24
Exclude CI files and issue templates from gems

nobu:
These files are never used to install or run the gems, but just waste.

Generated gem size will be reduced.

This would not be so frequent, but it actually reduced a half in [this case](https://github.com/newrelic/newrelic-ruby-agent/pull/458).

Exclude such files/directories from generated gemspec files.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{\A(?:test|spec|features)/}) }
+    `git ls-files -z`.split("\x0").reject do |f|
+      f.match(%r{\A(?:(?:test|spec|features)/|\.(?:git|travis|circleci)|appveyor)})
+    end
   end
   spec.bindir        = "exe"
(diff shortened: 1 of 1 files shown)


bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -25,7 +25,9 @@ Gem::Specification.new do |spec|
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{\A(?:test|spec|features)/}) }
+    `git ls-files -z`.split("\x0").reject do |f|
+      f.match(%r{\A(?:(?:test|spec|features)/|\.(?:git|travis|circleci)|appveyor)})
```
hsbt, line 29:
```suggestion
      f.match(%r{\A(?:(?:test|spec|features)/|\.(?:github|travis|circleci)|appveyor)})
```

It seems `.github` is correctly. But If you hope to remove like `.gitignore`, It's okay to the current code.

nobu: Yes, I intended to exclude `.gitlab-ci` also not only `.gitignore`, since 22cb599bcc considers it.

nobu: I believe that all `.git*` files are git-related, and assume that gems don't need them to work.

hsbt: 👍 I see. I agreed with your opinion.

PR #4038, opened 2020-10-26
Make printed drive letters consistent on Windows

deivid-rodriguez:

Sometimes driver letters are printed with inconsistent casing on Windows. This is unexpected and sometimes can cause surprising test failures, like the ones I'm unskipping here and like https://github.com/rubygems/rubygems/runs/1310494829?check_suite_focus=true.

The problem is that the helper that removes credentials from URLs for display, does the following when passed an actual file with a Windows driver letter.

```
URI("A:/foo").to_s
=> "a:/foo"
```

Namely, the `uri` library interprets the drive letter as the scheme and downcases it, causing the inconsistency.

My fix is to make the helper return the original argument when the string given is an actual file, since I believe in that case we can safely assume it includes no credentials.

An alternative would be to wrap the passed string with `file://` to give it a proper scheme, and then remove it after cleanup and before returning from the method. However, the initial approach is simpler and I think it's also safe.

### DIFF
--- a/bundler/lib/bundler/uri_credentials_filter.rb
+++ b/bundler/lib/bundler/uri_credentials_filter.rb
@@ -8,6 +8,8 @@ module Bundler
       return uri_to_anonymize if uri_to_anonymize.nil?
       uri = uri_to_anonymize.dup
       if uri.is_a?(String)
+        return uri if File.exist?(uri)
+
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Ready now, added a description!

PR #4043, opened 2020-10-29
Fix `--platform` option to `gem specification` being ignored

deivid-rodriguez:

While debugging a different issue, I wanted to know how the specification of the precompiled version of the `google-protobuf` gem for linux looked like.

So I was running

```
gem specification --remote google-protobuf --platform x86_64-linux
```

But I would keep getting the pure ruby version of the specification.

I think, if `--platform` option is given, it should print only specifications for the given platform.


bronzdoc (approved):
LGTM

PR #4045, opened 2020-11-01
Prevent remove command from deleting gemfile lines that are comments

jlgoedert:

Running `bundle remove foo`, removes the respective gem, "foo", from the Gemfile, but also removes any comments containing `gem "foo"`, where "foo" is the name of the removed gem (fixes https://github.com/rubygems/rubygems/issues/3276).

In the code that removes text from gemfiles, there is a regular expression that identifies lines where the gems targeted for removal are mentioned. To complement this, I use an additional regular expression to make sure the mention of the gem is not a line comment.

I included two tests. One test verifies such a line comment is not deleted. The other test makes sure a typical inline comment is still removed in a normal `bundle remove` scenario.

Please let me know if you think there are important cases that I missed. It is my understanding that the
```
=begin
=end 
```
comment syntax is very rarely used.

Any and all feedback is greatly appreciated!

### DIFF
--- a/bundler/spec/commands/remove_spec.rb
+++ b/bundler/spec/commands/remove_spec.rb
@@ -83,7 +83,7 @@ RSpec.describe "bundle remove" do
     end
   end
 
-  describe "remove mutiple gems from gemfile" do
+  describe "remove multiple gems from gemfile" do
     context "when all gems are present in gemfile" do
       it "shows success fir all removed gems" do
         gemfile <<-G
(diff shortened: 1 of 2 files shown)


bundler/spec/commands/remove_spec.rb
```diff
@@ -616,4 +616,43 @@
...
+        expect(out).to include("rack was removed.")
+        gemfile_should_be <<-G
+          source "#{file_uri_for(gem_repo1)}"
+
+          # gem "rack" might be used in the future
+        G
+      end
+    end
+
+    context "when gem specified for removal has an inline comment" do
```
deivid-rodriguez, line 641: How about a situation where `gem "rack"` is an inline comment to a different gem, like

```
gem "puma" # implements interface provided by gem "rack"
```

Does it work?

jlgoedert: Thank you for the feedback. I included two new tests to show this situation is now covered. I hope this is what you had in mind. Please let me know if I can make any improvements.

deivid-rodriguez (approved):

Issue #3347, opened 2020-03-14
[Windows] bundle install with platform - platform ignored


magynhard:
Hi guys,

I'm using the eventmachine gem, which delivers with a native C extension that needs to be build when installing. This works only when i manually install it by gem command, not with bundler.

When i install it manually, it is only working in this way:
```ruby
# commandline
gem install eventmachine --platform ruby
```

> Fetching: eventmachine-1.2.7.gem (100%)
Temporarily enhancing PATH for MSYS/MINGW...
Building native extensions.  This could take a while...
Successfully installed eventmachine-1.2.7
Parsing documentation for eventmachine-1.2.7
Installing ri documentation for eventmachine-1.2.7
Done installing documentation for eventmachine after 9 seconds
1 gem installed

If i don't specify the platform, it installs without compiling, and i get an error, when running my ruby app:
```ruby
# commandline
gem install eventmachine
```

> Fetching: eventmachine-1.2.7-x64-mingw32.gem (100%)
Successfully installed eventmachine-1.2.7-x64-mingw32
Parsing documentation for eventmachine-1.2.7-x64-mingw32
Installing ri documentation for eventmachine-1.2.7-x64-mingw32
Done installing documentation for eventmachine after 8 seconds
1 gem installed

> Unable to load the EventMachine C extension; To use the pure-ruby reactor, require 'em/pure_ruby'
cannot load such file -- 2.4/rubyeventmachine (LoadError)

But of course i want to put the dependency in my Gemfile, but the platforms option does not behave like it does with the gem command:

```ruby
# Gemfile
gem 'eventmachine', :platforms => :ruby
```

When i run `bundler install` it behaves just only like `gem install eventmachine`, without `--platform` parameter.

Since it is important to me that all developers can install the dependencies with `bundle install`, getting a running environment, I have a problem here.

> Bundler 1.16.2
> ruby 2.4.4p296 (2018-03-28 revision 63013) [x64-mingw32]
> MS Windows 7 Enterprise

dsisnero:
I have same problem.  ruby on windows- nokogiri
bundle config build.nokogiri --use-system-libraries 
I need option
bundle  config build.nokogiri --use-system-libraries --platform=ruby

deivid-rodriguez:
Is this still reproducible?

asm256:
Yes.
Latest [EventMachine](https://rubygems.org/gems/eventmachine) is still need option `--platform ruby` on gem install.

[...]

PR #4049, opened 2020-11-05
Extend `gem` DSL with a `force_ruby_platform` option

deivid-rodriguez:

Sometimes it would be useful, specially when we enable the `specific_platform` setting by default to be able to opt-out of the standard platform priority and always force the ruby variant for specific gems.

Add a `:force_ruby_platform` option to the `gem` DSL.

This is the "per gem" equivalent of the `force_ruby_platform` setting.

This can be handy (assuming the pure ruby variant works just fine) when:

* You're having issues with the platform specific variant.

* The platform specific variant does not yet support a newer ruby (and thus has a `required_ruby_version` upper bound), but you still want your Gemfile{.lock} files to resolve under that ruby.

This PR closes https://github.com/rubygems/rubygems/issues/3347.

### DIFF
--- a/bundler/lib/bundler/dependency.rb
+++ b/bundler/lib/bundler/dependency.rb
@@ -109,7 +109,7 @@ module Bundler
       @env            = options["env"]
       @should_include = options.fetch("should_include", true)
       @gemfile        = options["gemfile"]
-      @force_ruby_platform = options["force_ruby_platform"]
+      @force_ruby_platform = options.fetch("force_ruby_platform", default_force_ruby_platform)
 
       @autorequire = Array(options["require"] || []) if options.key?("require")
     end
@@ -158,6 +158,18 @@ module Bundler
       super
     rescue NoMethodError
       requirement != ">= 0"
+    end
+
+    private
+
+    # The `:force_ruby_platform` attribute is `false` by default, except for
+    # TruffleRuby. TruffleRuby generally needs to force the RUBY platform
+    # variant unless the name is explicitly allowlisted.
+
+    def default_force_ruby_platform
+      return false unless Bundler.current_ruby.truffleruby?
+
+      !Gem::Platform::REUSE_AS_BINARY_ON_TRUFFLERUBY.include?(name)
     end
   end
 end
--- a/bundler/lib/bundler/dependency.rb
+++ b/bundler/lib/bundler/dependency.rb
@@ -7,7 +7,7 @@ require_relative "rubygems_ext"
 module Bundler
   class Dependency < Gem::Dependency
     attr_reader :autorequire
-    attr_reader :groups, :platforms, :gemfile, :git, :github, :branch, :ref
+    attr_reader :groups, :platforms, :gemfile, :git, :github, :branch, :ref, :force_ruby_platform
 
     # rubocop:disable Naming/VariableNumber
     PLATFORM_MAP = {
@@ -109,6 +109,7 @@ module Bundler
       @env            = options["env"]
       @should_include = options.fetch("should_include", true)
       @gemfile        = options["gemfile"]
+      @force_ruby_platform = options["force_ruby_platform"]
 
       @autorequire = Array(options["require"] || []) if options.key?("require")
     end
--- a/bundler/lib/bundler/man/gemfile.5.ronn
+++ b/bundler/lib/bundler/man/gemfile.5.ronn
@@ -231,6 +231,20 @@ All operations involving groups ([`bundle install`](bundle-install.1.html), `Bun
 `Bundler.require`) behave exactly the same as if any groups not
 matching the current platform were explicitly excluded.
 
+### FORCE_RUBY_PLATFORM
+
+If you always want the pure ruby variant of a gem to be chosen over platform
+specific variants, you can use the `force_ruby_platform` option:
+
+    gem "ffi", force_ruby_platform: true
+
+This can be handy (assuming the pure ruby variant works fine) when:
+
+* You're having issues with the platform specific variant.
+* The platform specific variant does not yet support a newer ruby (and thus has
+  a `required_ruby_version` upper bound), but you still want your Gemfile{.lock}
+  files to resolve under that ruby.
+
 ### SOURCE
 
 You can select an alternate Rubygems repository for a gem using the ':source'
--- a/bundler/lib/bundler/man/gemfile.5.ronn
+++ b/bundler/lib/bundler/man/gemfile.5.ronn
@@ -495,8 +495,8 @@ the `.gemspec` file.
 
 The `gemspec` method adds any runtime dependencies as gem requirements in the
 default group. It also adds development dependencies as gem requirements in the
-`development` group. Finally, it adds a gem requirement on your project (`:path
-=> '.'`). In conjunction with `Bundler.setup`, this allows you to require project
+`development` group. Finally, it adds a gem requirement on your project (`path:
+'.'`). In conjunction with `Bundler.setup`, this allows you to require project
 files in your test code as you would if the project were installed as a gem; you
 need not manipulate the load path manually or require project files via relative
 paths.
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -222,27 +222,9 @@ module Gem
     MINGW = Gem::Platform.new("x86-mingw32")
     X64_MINGW = [Gem::Platform.new("x64-mingw32"),
                  Gem::Platform.new("x64-mingw-ucrt")].freeze
-  end
 
-  Platform.singleton_class.module_eval do
-    unless Platform.singleton_methods.include?(:match_spec?)
-      def match_spec?(spec)
-        match_gem?(spec.platform, spec.name)
-      end
-
-      def match_gem?(platform, gem_name)
-        match_platforms?(platform, Gem.platforms)
-      end
-
-      private
-
-      def match_platforms?(platform, platforms)
-        platforms.any? do |local_platform|
-          platform.nil? ||
-            local_platform == platform ||
-            (local_platform != Gem::Platform::RUBY && local_platform =~ platform)
-        end
-      end
+    if RUBY_ENGINE == "truffleruby" && !defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)
+      REUSE_AS_BINARY_ON_TRUFFLERUBY = %w[libv8 sorbet-static].freeze
     end
   end
 
--- a/bundler/spec/install/gemfile/force_ruby_platform_spec.rb
+++ b/bundler/spec/install/gemfile/force_ruby_platform_spec.rb
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+RSpec.describe "bundle install with force_ruby_platform DSL option", :jruby do
+  context "when no transitive deps" do
+    before do
+      build_repo4 do
+        # Build a gem with platform specific versions
+        build_gem("platform_specific") do |s|
+          s.write "lib/platform_specific.rb", "PLATFORM_SPECIFIC = '1.0.0 RUBY'"
+        end
+
+        build_gem("platform_specific") do |s|
+          s.platform = Bundler.local_platform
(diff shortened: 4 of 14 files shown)


bundler/spec/install/gemfile/force_ruby_platform_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it "still pulls the ruby variant" do
+      simulate_platform "ruby" do
+        install_gemfile <<-G
+          source "#{file_uri_for(gem_repo4)}"
+
+          gem "depends_on_platform_specific"
+          gem "platform_specific", :force_ruby_platform => true
+        G
+
+        expect(the_bundle).to include_gems "platform_specific 1.0.0 RUBY"
```
simi, line 60: :thinking: It would make sense to me to add also specs for not setting `force_ruby_platform` here to see (and actually test) the difference.

```ruby
gem "platform_specific-forced", :force_ruby_platform => true
gem "platform_specific"
# ...
expect(the_bundle).to include_gems "platform_specific-forced 1.0.0 RUBY"
expect(the_bundle).to include_gems "platform_specific 1.0.0 JAVA"
```

deivid-rodriguez: Sure.

simi (approved):

deivid-rodriguez:
I'd like to introduce this, we probably need a new minor but it seems ok? I was kind of waiting to introduce #4076 before this, because it's a really nice feature to have before introducing DSL changes.

With that in place, you could freely use new `DSL` features in your Gemfile like `gem "google-protobuf", force_ruby_platform: true` and happily distribute them, knowing that all users of your Gemfile will transparently get a bundler version that understands those features. Of course, people would need to run a bundler version that has #4076. Maybe not a life changer, but it sounds really nice to me.

If we proceeed with this PR _without_ automatic installation of the `BUNDLED WITH` version, then users not running a proper version will get an error like:

> You passed `:force_ruby_platform` as an option for gem "google-protobuf", but it is invalid. Valid options are: <list_of_valid_options>. You may be able to resolve this by upgrading Bundler to the newest version.

Not as nice as "just works", but not that bad either.

ncri:
When is this going to be merged?

ncri:
Anyone reading here? Looks this has already been approved?

ngan:
@deivid-rodriguez will this make it into next release? 🙏

deivid-rodriguez:
It depends on whether I manage to make it actually work or not 😆,  but I hope to be able to release it soon!

ncri:
That would be excellent! :)

eregon:
This functionality would be great, recently @bjfish and I were wondering if `gem 'foo', platforms: :ruby` was related but actually not, that Bundler `gem name, platforms:` is completely unrelated to the the `Gem.platforms`.

bundler/lib/bundler/dependency.rb
```diff
@@ -159,5 +159,13 @@ def specific?
     rescue NoMethodError
       requirement != ">= 0"
     end
+
+    private
+
+    def default_force_ruby_platform
+      return false unless Bundler.current_ruby.truffleruby?
+
+      !%w[libv8 sorbet-static].include?(name)
```
eregon, line 168: (reposting it here so we can discuss this on the PR more easily, vs a comment on a commit)
Could we use `match_gem?` here so we avoid duplicating the logic of https://github.com/oracle/truffleruby/blob/b131d3c146afe6bed0f119d0891b34ba68ac5deb/lib/mri/rubygems/platform.rb#L38-L47 ?
That list is subject to change and needs to remain in the truffleruby repo.

deivid-rodriguez: This was just a WIP but I was planning to use the constant.

deivid-rodriguez: Truffleruby wants to force the ruby platform for some specific gems, using this feature seems like a perfect use case for me.

eregon: > Truffleruby wants to force the ruby platform for some specific gems, using this feature seems like a perfect use case for me.

Ah right, this would be a cleaner way to integrate than to override `match_gem?` in TruffleRuby, is that what you mean?
That sounds good in general.
I'd be more comfortable if this `default_force_ruby_platform` was a method on e.g. `Gem` (and it would be passed the gem name and the version, as the answer might depend on the version) and TruffleRuby could override that method in `rubygems/defaults/truffleruby.rb`

deivid-rodriguez: > Is there are reason we can't use match_gem? here?

We can, but using this feature feels way better to me. As you can see in d3317b8192cc912f9219c84f2b6302548be1fe34, it allows us to remove all the hacks we need to keep just for truffleruby, and that we introduced because of causing regressions. I think this is much cleaner and less likely to cause issues.

deivid-rodriguez: > I'd be more comfortable if this default_force_ruby_platform was a method on e.g. Gem (and it would be passed the gem name and ideally the version too, as the answer might depend on the version) and TruffleRuby could override that method in rubygems/defaults/truffleruby.rb

I think passing the version is overkill to me. We haven't run resolution at this point yet, so we only have Gemfile information here. We could use the version if it's explicit in the Gemfile, but I think it's an unnecessary complication. Let's stick to this. Regarding moving the method, I don't think this is a good fit as a global `Gem` method, and I don't want to pollute `Gem` more unnecessarily. Why will you want to override this method?

deivid-rodriguez: I added a commit to use `Gem::Platform::REUSE_AS_BINARY_ON_TRUFFLERUBY`. I'm happy to backfill it for older versions of truffleruby that didn't have it.

eregon: Yes, agreed having the version is not worth the trouble at this point, just the gem name is enough currently.

> Regarding moving the method, I don't think this is a good fit as a global Gem method, and I don't want to pollute Gem more unnecessarily. 

It could be somewhere else, e.g., on Gem::Platform or any other place.
An instance method on Bundler::Dependency feels weird to me if we ever need that logic anywhere else, but maybe we don't, WDYT?
I guess `Bundler::Dependency#force_ruby_platform` is already a good way to get the result of this check.
An advantage of having it on Bundler::Dependency is we have more information should we ever need it. Would be the same if it's a method somewhere else and it's passed the Bundler::Dependency instance, but then indeed there is no point to move the method.

> Why will you want to override this method?

Because I want some flexibility, for instance I'd think the list might depend on some TruffleRuby option, or depend on the platform, or might want to match gem names with a regexp (e.g., to handle potential forks of those gems, or some related gems which should all be installed as binary).
In other words, a constant like that does not feel like a proper API, a method seems much more flexible.
But I'm OK to override `Bundler::Dependency#default_force_ruby_platform` in TruffleRuby, so that could be the "extension point".

It also feels cleaner to me if TruffleRuby overrides the method and the default is just `false`, because then there are no truffleruby specifics here, except for a comment on this new method to mention it is overwritten by truffleruby (but possibly also by other Ruby implementations which want to override it in the future).

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -221,6 +221,10 @@ class Platform
     MSWIN64 = Gem::Platform.new("mswin64") unless defined?(MSWIN64)
     MINGW = Gem::Platform.new("x86-mingw32") unless defined?(MINGW)
     X64_MINGW = Gem::Platform.new("x64-mingw32") unless defined?(X64_MINGW)
+
+    if RUBY_ENGINE == "truffleruby" && !defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)
+      REUSE_BINARY_ON_TRUFFLERUY = %w[libv8 sorbet-static]
```
eregon, line 226: It's `REUSE_AS_BINARY_ON_TRUFFLERUBY` not `REUSE_BINARY_ON_TRUFFLERUY`.
I'd prefer if we don't depend on that constant name going forward though.

bundler/lib/bundler/dependency.rb
eregon: > It also feels cleaner to me if TruffleRuby overrides the method and the default is just `false`, because then there are no truffleruby specifics here, except for a comment on this new method to mention it is overwritten by truffleruby (but possibly also by other Ruby implementations which want to override it in the future).

Ah, but that can't be because then existing TruffleRuby releases would just break with latest Bundler, I didn't think about that ealier.
OK, then using the constant on those older TruffleRuby versions seems unavoidable.
Are you OK if newer TruffleRuby releases override this method? Could you add a comment that the method might be overridden by Ruby implementations and e.g. truffleruby?

bundler/lib/bundler/rubygems_ext.rb
eregon: (still a typo here, missing a B before the last Y)

bundler/lib/bundler/dependency.rb
deivid-rodriguez: I think you can send us a patch with whatever comment is needed when/if you need to override the method in the future

bundler/lib/bundler/dependency.rb
```diff
@@ -158,5 +159,13 @@ def specific?
     rescue NoMethodError
       requirement != ">= 0"
     end
+
+    private
+
+    def default_force_ruby_platform
```
eregon, line 165: I'd like to suggest this:
```suggestion
    # Note that TruffleRuby (and possibly other Ruby implementations) might override this method for more flexibility.
    # This default implementation is provided for compatibility with older TruffleRuby releases.
    def default_force_ruby_platform
```

deivid-rodriguez: As I said above, let's discuss this when/if you decide to override the method in the future. It seems clear to me now what the method is doing.

bundler/lib/bundler/man/gemfile.5.ronn
```diff
@@ -231,6 +231,20 @@ All operations involving groups ([`bundle install`](bundle-install.1.html), `Bun
 `Bundler.require`) behave exactly the same as if any groups not
 matching the current platform were explicitly excluded.
 
+### FORCE_RUBY_PLATFORM
+
+If you always want the pure ruby variant of a gem to be chosen over platform
+specific variants, you can use the `force_ruby_platform` option:
+
+    gem "ffi", :force_ruby_platform => true
```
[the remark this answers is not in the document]
deivid-rodriguez: Fair enough, I used our internal style for the documentation, but we should use the most common style.

eregon:
@deivid-rodriguez Thank you for your work on this!
It seems like a great way to implement the behavior for truffleruby, much better than having to change/rely on `Gem.platforms` like I did initially (I didn't realize it'd be enough to change the `force_ruby_platform` on a dependency).

I'll have to check if anything else in RubyGems uses `Gem.platforms` as a way to select gems, I'd guess yes for e.g. `gem install ...`, so probably we still need `match_gem?`, etc for RubyGems, or maybe some newer way piggy-backing on RubyGems's equivalent of the `force_ruby_platform` flag.

bundler/lib/bundler/dependency.rb
eregon: OK. It wasn't clear to me initially on this PR, so I think such a comment would be helpful for anyone else reading this code.

deivid-rodriguez:
> It seems like a great way to implement the behavior for truffleruby, much better than having to change/rely on Gem.platforms like I did initially (I didn't realize it'd be enough to change the force_ruby_platform on a dependency).

I didn't occur to me either until I was working on this feature :)

> I'll have to check if anything else in RubyGems uses Gem.platforms as a way to select gems, I'd guess yes for e.g. gem install ..., so probably we still need match_gem?, etc for RubyGems, or maybe some newer way piggy-backing on RubyGems's equivalent of the force_ruby_platform flag.

Yes, I think `match_gem?`, etc are still needed for RubyGems, but for now it feels nice to be able to simplify the Bundler side.

eregon:
Accidentally closed by https://github.com/rubygems/rubygems/pull/5674

deivid-rodriguez:
Ooops.

deivid-rodriguez:
Alright let's do this!

Issue #4018, opened 2020-10-15
Consider delaying the change to resolve against all `Gemfile` platforms by default


deivid-rodriguez:
### Describe the problem as clearly as you can

In bundler 2.2.0, as per https://github.com/rubygems/bundler/pull/7580, bundler will start resolving against all platforms directly specified in the `Gemfile`. See https://github.com/rubygems/bundler/pull/7580 for the rationale behind it.

This change will have some consequences:

* It will make `bundle install` fail in some cases that were previously not failing.

* It will make resolution less slower for gemfiles including platforms different than that currently running platform, because it will resolve all platforms by default and not just the current platform.

An example of the former is if you're developing a gem with both jruby & C-extensions, using the same `gemspec` for both, running `bundle install` under `jruby` and having a `Gemfile` with a gem specific to a platform different than jruby.

This happens in https://github.com/mongodb/bson-ruby (that has both java and C-extensions, and includes `gem "byebug", platform: :mri` in its `Gemfile`). In such a case, `bundler` will try to resolve against all platforms, and for each of them it will try to use the same gemspec (because of `gemspec` DSL in the `Gemfile`). It will eventually fail because the gemspec picks the currently running platform but will fail to be found for other platforms.

The solution would involve maintaining several gemspecs for each supported platform, but I think most gems like this don't do that, and the error given doesn't make it clear that this is a solution (or what the culprit is).
  
### Post steps to reproduce the problem

```shell
git clone https://github.com/mongodb/bson-ruby
cd bson-ruby
rbenv shell jruby
gem install bundler:2.2.0.rc.2
bundle install
```

### What were you expecting to happen?

No failure, lockfile generated.

### What actually happened?

```
$ bundle install
[bson] Warning: No private key present, creating unsigned gem.
Fetching gem metadata from https://rubygems.org/........
Could not find gem 'bson' in source at `.`.
The source contains 'bson' at: 4.10.0 java
```

### Proposed "solution"

After looking back, the initial reason to start doing this was to "fix" the annoying warning `bundler` prints in this situation:

> The dependency tzinfo-data (>= 0) will be unused by any of the platforms Bundler is installing for. Bundler is installing for ruby but the dependency is only for x86-mingw32, x86-mswin32, x64-mingw32, java. To add those platforms to the bundle, run `bundle lock --add-platform mingw, mswin, x64_mingw, jruby`.

[...]

PR #4052, opened 2020-11-06
Revert resolving all Gemfile platforms automatically

deivid-rodriguez:

This is the second time I revert this change, since I have found some cases where this might cause pain, and I want to battle test multiplatform fixes before attempting this.

See https://github.com/rubygems/rubygems/issues/4018 for more on this.

My fix is to revert resolving only for the current platform like it's been done historically.

I think ideally, when we introduce something like this, we should consider making it opt-in by recording this intention in the `Gemfile`, maybe with an unscoped `platforms` DSL, or by an explicit CLI flag, like `bundle install --all-platforms`, or `bundle lock --all-platforms`.

The part I'm not reverting is removing the warning about dependencies being unused under the current platform, since it is pretty much expected on a multiplatform context.

Closes https://github.com/rubygems/rubygems/issues/4018.

### DIFF
--- a/bundler/lib/bundler/cli/outdated.rb
+++ b/bundler/lib/bundler/cli/outdated.rb
@@ -76,8 +76,6 @@ module Bundler
         next unless gems.empty? || gems.include?(current_spec.name)
 
         active_spec = retrieve_active_spec(definition, current_spec)
-        next unless active_spec
-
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I will revisit this for the next version.

PR #4053, opened 2020-11-07
Remove duplicate cases

knqyf263:
## What is your fix for the problem, implemented in this PR?
Remove duplicate cases

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4054, opened 2020-11-09
Lock parallel ~> 1.19.2

hsbt:

The latest version of `parallel` gem show https://github.com/rubygems/rubygems/pull/4053/checks?check_run_id=1371824322#step:7:460 

Lock the old version of `parallel` that is supporeted Ruby 2.3 and 2.4.


deivid-rodriguez:
Thanks for fixing this! Just changing the label because this is a "test only fix" that doesn't affect end users, so we don't want it to appear in the changelog.

PR #4066, opened 2020-11-17
New gem template: prefer `require_relative` to `require`

marcandre:
New gem template uses `require_relative` in the gemspec but `require` in the `lib/gem.rb`. This PR changes that second occurrence.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/lib/newgem.rb.tt
+++ b/bundler/lib/bundler/templates/newgem/lib/newgem.rb.tt
@@ -1,8 +1,8 @@
 # frozen_string_literal: true
 
-require "<%= config[:namespaced_path] %>/version"
+require_relative "<%= File.basename(config[:namespaced_path]) %>/version"
 <%- if config[:ext] -%>
-require "<%= config[:namespaced_path] %>/<%= config[:underscored_name] %>"
+require_relative "<%= File.basename(config[:namespaced_path]) %>/<%= config[:underscored_name] %>"
 <%- end -%>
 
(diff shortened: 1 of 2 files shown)


marcandre:
Mmm, that's incorrect in case of `bundle gem with-dash` though.

marcandre:
Should be fixed, tests too 😅

deivid-rodriguez (approved):
Thanks @marcandre!

bobf:
@deivid-rodriguez Can you give some info about the benefits of using `require_relative` over `require` here ?

The PR description mentions that it is to bring things in line with the `.gemspec` definition but the relevant commit [here](https://github.com/rubygems/rubygems/commit/9775fb80bc6de3fbf56abd2d08788d87b84f4968) doesn't seem to provide any extra context and isn't linked to a PR.

Should `require_relative` be used for other files in the gem and are you able to explain why this is better ? Does it provide some benefits for disambiguation ?

I would appreciate any feedback so I can adapt to newer/better practices if needed.

deivid-rodriguez:
Hello!

My personal recommendation is to use `require_relative` for files relative to your project. That gives information to the person that reads the code that you're loading an internal dependency, not a third party one. It also has the performance benefit that it doesn't need to search the `$LOAD_PATH` since `require_relative` knows exactly where the file it needs to load is located.

bobf:
@deivid-rodriguez That makes sense - thanks a lot for the quick response. I will use this strategy in future.

Have a nice Christmas !

Issue #88, opened 2011-06-09
rubygems creates ruby wrappers around non-ruby executables, causing them to fail


olbrich:
when providing executables with a gem, they get wrapped in a ruby wrapper that provides some support for using previous
versions of a gem. If the executable is not a ruby file (e.g., a bash script, php, etc..) then the wrapper will fail.

Suggestions:
- allow the gemspec to force the gem to be installed with '--no-wrapper' and leave it to the developer to properly
  wrap executables.
- check the shebang line of the executable and skip wrapping for anything that explicitly uses something other than
  ruby
- provide a warning at the time the gem is built that files will be wrapped (check for .bat files, or shebangs that don't include ruby... perhaps not perfect, but it will probably catch most cases).
- document the existing behavior better and provide instructions for work arounds for non-ruby binaries.

kigster:
@steveklabnik @zenspider @olbrich @HugoGiraudel I just bumped into another valid reason in favor of allowing gem developers to install non-ruby executables automatically.  

I am unaware of any way that a child process can affect the parent's process environment on UNIX, in particular, user's current directory. This is no different from Ruby scripts.  However, were it a shell script, you can simply `eval` a string containing commands, and affect the parent shell process, including changing your current directory.

[I am building a ruby version](https://github.com/kigster/warp-dir) of the `wd` (warp directory) tool that's been [available for `zsh`](https://github.com/mfaerevaag/wd) for quite some time now. I love the concept, and want to expand on it, but the core functionality of this tool is to be able to change the directory of the user that runs `wd` in the first place. It's easy to do in a shell, but impossible to do in Ruby.

I am currently shipping two executables with the gem: `warp-dir` – is a ruby file that runs the gem's code. To work around the limitation of not being able to change the directory of the parent process, I am having the ruby code generate shell compatible executable code (even for printing help). Talk about mete-meta programming :) So `warp-dir` ruby script is not meant to be run directly by humans.  You are supposed to do something like this: `eval $(warp-dir --help)`.  So the `wd` executable is a tiny script that provides this functionality.  And it would be lovely to have it be installed without ruby wrappers automatically. Gem user should really not worry about this type of thing.

I am glad that it looks like I can choose `--no-wrapper` flag while installing the gem, but it would be much more preferable if the gem author could explicitly tell rubygems not to create wrappers regardless.

One possible way to do this in a backwards compatible way, is by introducing a new config option for such scripts:

```ruby
[...]

PR #4067, opened 2020-11-17
Automatically don't wrap non-ruby executables

iboB:
This is still a draft. The functionality is in, but docs are not updated. 

This is the result of the discussion in #88 

In short: automatically do not create wrappers of non-ruby executables, where a ruby executable is defined as having a shebang which contains "ruby"

There was a test which basically checked that what is defined as a "non-ruby executable" was wrapped. I removed this test, and added a new one which checks that it's instead symlinked.

Some other tests which relied on the old behavior needed tweaks:  their mock executables didn't contain a ruby shebang, so now they do

@deivid-rodriguez if you're ok with this so far, I'll make changes to the docs and submit the full PR

# Make sure he following tasks are checked


iboB:
Uh-oh... some tests fail on Windows. Executables missing ruby shebang again, and thus not getting wrapped. If we're ok with the current direction of this PR, I'll fix them

PR #4069, opened 2020-11-17
Don't wrap non-ruby executables and explicit no-wrap requests. 

iboB:
This is still a draft. The functionality is in, but docs are not updated.

This is the result of the discussion in #88

In short: apart from checking `@wrappers` the installer performs an additional check before deciding whether to wrap executables:

* Check if the first line is a shebang: 
    * if it is, check if it contains the string 'ruby', 
    * if it doesn't, assume this is not a ruby executable and don't wrap
* If it's not a shebang, check if it contains the magic string 'rubygems: no-wrap'
    * if it does, don't wrap as well. This will be the way to deploy windows batch files
    * if it doesn't contain the magic string, assume it's a shebang-less ruby executable and wrap away

This doesn't break any of the existing tests. I find it highly unlikely that any existing gem will be affected by this

Future gems will be able to provide shell scripts as non-wrapped executables, without requesting the users to explicitly use `--no-wrappers` when installing

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -474,6 +474,26 @@ class Gem::Installer
     end
   end
 
+  ##
+  # Checks if +bin_path+ should be wrapped if we're generating wrappers
+  #
+  # If the first line of the file is a shebang check if it mentions ruby and
+  # return false if it doesn't (don't wrap non-ruby executable)
+  #
+  # If the first line is not a shebang, check if it contains the magic string:
+  # "rubygems: no-wrap" and return false if it does (don't wrap if no-wrap was
+  # requested)
+  #
+  # Otherwise wrap
+
+  def wrappable_executable?(bin_path)
+    File.open bin_path, 'r' do |f|
+      line1 = f.gets
+      break line1['ruby'] if line1.start_with?('#!') # if the line is a shebang check for ruby
+      !line1['rubygems: no-wrap'] # check for magic string
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


iboB:
Er... I saw the failing tests on windows, but to be honest I can't see how my changes could've caused the fails

lib/rubygems/installer.rb
```diff
@@ -474,6 +474,26 @@ def generate_windows_script(filename, bindir)
...
+  # requested)
+  #
+  # Otherwise wrap
+
+  def wrappable_executable?(bin_path)
+    File.open bin_path, 'r' do |f|
+      line1 = f.gets
+      break line1['ruby'] if line1.start_with?('#!') # if the line is a shebang check for ruby
+      !line1['rubygems: no-wrap'] # check for magic string
+    end
```
knu, line 494: Using String#[] is not the best way to check if a string contains a substring, as it's not a boolean operation.
Also, the method should be aware that IO#gets may return nil if the file is empty.

You could use a regexp here.

```ruby
# If the first line is a non-ruby shebang or contains a magic string, return false.
/\A#!(?!.*ruby)|rubygems: no-wrap/ !~ File.open(bin_path, &:gets)
```

knu (commented):
Is this PR still a draft?

iboB:
@knu it is still a draft. Waiting on @deivid-rodriguez on whether we should go on with this functionality. If yes, then I'll add some docs changes in the same PR. 

Thanks for the review!

Issue #3053, opened 2019-12-27
avoiding Kernel.warn patch


kares:
Seems like RGs 3.0 introduced a `Kernel.warn` replacement.
https://github.com/rubygems/rubygems/commit/7153a317c3e545658459d84aeafdad891d3a8a36#diff-5f3c1066128b69ac52cca10be3eee4f0
This is getting installed "globally" - but does not seem like filtering `:uplevel` is always necessary.

I understand why its being introduced but is it really needed globally?
My concern is that we're potentially walking the (uplevel) stack more than once.
If someone would able to confirm we could refactor to use a proper `Gem.warn` internally, instead. 

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

deivid-rodriguez:
Hi @kares!

Sorry for the delaying in getting back to you here.

The `Kernel.warn` patch was added at https://github.com/rubygems/rubygems/pull/2442. The idea was to fix a small issue where rubygems require would behave differently from ruby's require.

In https://github.com/rubygems/rubygems/pull/3985 I'm changing the patch to a version that seems friendlier and behaves better, so if you were having issues with the current version, maybe this one will behave better for you too.

Hope this helps!

deivid-rodriguez:
It sounds like we'll probably be able to remove the `Kernel.warn` patch in the future, have a look at #3987. Closing this!

PR #4075, opened 2020-11-20
Do not override Kernel#warn when there is no need

eregon:
* On Ruby implementations where https://bugs.ruby-lang.org/issues/17259
  is available, backtrace entries starting with `<internal:` are ignored
  for Kernel#warn. We can then define RubyGems's Kernel#require with
  such a filename and it will automatically be skipped for Kernel#warn.
* This is much less fragile (overriding Kernel#warn causes multiple
  issues and is very difficult to do properly) and it is also more
  efficient (the override would walk the stack multiple times).

Based on the idea in https://github.com/rubygems/rubygems/pull/3987#issuecomment-704334889

cc @deivid-rodriguez 

There were various issues due to monkey patching Kernel#warn, a quick search reveals at least:
https://github.com/rubygems/rubygems/issues/3053
https://github.com/rubygems/rubygems/issues/2588
https://github.com/rubygems/rubygems/pull/2911
and ~12 PRs seem to match `Kernel#require`: https://github.com/rubygems/rubygems/pulls?q=is%3Apr+Kernel%23warn
(original PR: https://github.com/rubygems/rubygems/pull/2442)

Also as a Ruby implementor, I can say overriding Kernel methods when it's not strictly needed leads to all kind of troubles.
Notably I got some headaches to handle the filtering in TruffleRuby and still let the filtering on RubyGems understand what's going on (it was not enough to just filter twice to give an idea of the complexity).

And of course having subtle behavior differences when passing `--disable-gems` for Kernel#warn is very confusing (I got tricked by this multiple times).

This PR is a nice way to not need to monkey-patch at all, and still always skip RubyGems' `require` when using `warn('message', uplevel: 1)`, for instance at the top of a file to emit a deprecation warning. So it fixes the original problem https://github.com/rubygems/rubygems/pull/2442 without monkey-patching anything.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -119,6 +119,10 @@ module Gem
   # to avoid deprecation warnings in Ruby 2.7.
   UNTAINT = RUBY_VERSION < '2.7' ? :untaint.to_sym : proc{}
 
+  # When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
+  KERNEL_WARN_IGNORES_INTERNAL_ENTRIES = RUBY_ENGINE == "truffleruby" ||
+      (RUBY_ENGINE == "ruby" && RUBY_VERSION >= '3.0')
+
   ##
   # An Array of Regexps that match windows Ruby platforms.
 
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -17,6 +17,8 @@ module Kernel
     private :gem_original_require
   end
 
+  file = Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES ? "<internal:#{__FILE__}>" : __FILE__
+  module_eval <<'RUBY', file, __LINE__
   ##
   # When RubyGems is required, Kernel#require is replaced with our own which
   # is capable of loading gems on demand.
@@ -166,6 +168,7 @@ module Kernel
       end
     end
   end
+RUBY
 
   private :require
 
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -18,7 +18,7 @@ module Kernel
   end
 
   file = Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES ? "<internal:#{__FILE__}>" : __FILE__
-  module_eval <<'RUBY', file, __LINE__
+  module_eval <<'RUBY', file, __LINE__ + 1
   ##
   # When RubyGems is required, Kernel#require is replaced with our own which
   # is capable of loading gems on demand.
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # `uplevel` keyword argument of Kernel#warn is available since ruby 2.5.
-if RUBY_VERSION >= "2.5"
+if RUBY_VERSION >= "2.5" && !Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES
 
   module Kernel
     rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -709,7 +709,7 @@ class TestGemRequire < Gem::TestCase
         RUBY
 
         _, err = capture_subprocess_io do
-          system(*ruby_with_rubygems_in_load_path, "-w", "--enable=gems", "-C", dir, "-I", dir, "main.rb")
+          system(*ruby_with_rubygems_in_load_path, "-w", "--disable=gems", "-C", dir, "-I", dir, "main.rb")
         end
         assert_match(/main\.rb:2: warning: This is a deprecated method$/, err)
         _, err = capture_subprocess_io do


eregon:
@deivid-rodriguez Could you or some other maintainer (who should I ping?) review this?
I think it's important to merge this so it can be in Ruby 3.

lib/rubygems/core_ext/kernel_warn.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 # `uplevel` keyword argument of Kernel#warn is available since ruby 2.5.
-if RUBY_VERSION >= "2.5"
+if RUBY_VERSION >= "2.5" && !Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES
```
deivid-rodriguez, line 4: Since the require of this file is already "protected", I think we don't need to check it again? If someone is loading this file directly for some reason, I guess we can let run the code?

deivid-rodriguez (commented):
Looks good to me, just a minor comment but looks fine either way!

lib/rubygems.rb
```diff
@@ -119,6 +119,10 @@ module Gem
   # to avoid deprecation warnings in Ruby 2.7.
   UNTAINT = RUBY_VERSION < '2.7' ? :untaint.to_sym : proc{}
 
+  # When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
+  KERNEL_WARN_IGNORES_INTERNAL_ENTRIES = RUBY_ENGINE == "truffleruby" ||
```
simi, line 123: Isn't there any different way (without hardcoding engine and version) to check this functionality is present? If not, can't we do some sanity check (not printed to stdout) to check how warn behaves to detect?

lib/rubygems/core_ext/kernel_warn.rb
eregon: I prefer to be sure, i.e. that Kernel#warn is never overridden when `Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES` is a guarantee.
I see this as the main check and then the other check (`require 'rubygems/core_ext/kernel_warn' unless Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES`) is just an optimization.

eregon: This way, even if some other part of RubyGems at some point in the future does `require 'rubygems/core_ext/kernel_warn'`, it will still only override `Kernel#warn` if needed, which seems more future-proof.

lib/rubygems.rb
eregon: We would have to temporarily switch `$stderr` to test that (and to what? we don't want to load StringIO just for that).
RubyGems could be required potentially concurrently in some thread, so that change could end up breaking something else.
I think it's difficult to feature-detect this one.

If we want something more opt-in/flexible we could let `require "rubygems/defaults/#{RUBY_ENGINE}"` opt in for this behavior.
Then we'd need e.g. some `@ivar` on Gem and a method to set it like `Gem.warn_ignores_internal_entries`.

deivid-rodriguez: I thought the same thing but this thing felt hard to feature-detect.

eregon: OK as-is, or do you want I make it opt-in from `require "rubygems/defaults/#{RUBY_ENGINE}"`?
If the latter, could you decide the method names? (to set and to query)

deivid-rodriguez: As a small note, I wrote https://github.com/rubygems/rubygems/pull/4075/files#r536002463 before reading https://github.com/rubygems/rubygems/pull/4075/files#r536001783. It was a reply to @simi's comment, not to @eregon's explanation. I'm :+1: with the hardcoded version/implementation check.

lib/rubygems/core_ext/kernel_warn.rb
deivid-rodriguez: I can't imagine how we could end up requiring the file from some other part of rubygems, honestly. But assuming that could happen, then I'd say let's remove the "optimization" on the require itself so we don't have carry it around everywhere. Checking the same thing right before & after a require feels weird, specially if there's an extra condition that you only check in one of the two places.

eregon: OK, I removed the optimization condition, so the file is always loaded and all conditions are inside.

lib/rubygems/core_ext/kernel_require.rb
```diff
@@ -18,7 +18,7 @@ module Kernel
   end
 
   file = Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES ? "<internal:#{__FILE__}>" : __FILE__
-  module_eval <<'RUBY', file, __LINE__
+  module_eval <<'RUBY', file, __LINE__ + 1
```
deivid-rodriguez, line 21: Oh, nice catch! Since we almost missed this regression, should we add a test where an exception is raised during a require and check that the backtrace is correct?

eregon: It seems difficult, because we would have to hardcode some line numbers (e.g. the line of `raise` or the line of `def`), no?

I also noticed many other places use `__LINE__ + 1` in the codebase, so it's just the correct pattern that I unfortunately missed initially.

deivid-rodriguez: Yeah, well I was thinking of a "clever" solution of reading and greping the file and inferring line numbers from that, but it's probably not worth it?

deivid-rodriguez: Maybe a rubocop cop is best!

deivid-rodriguez: https://github.com/rubygems/rubygems/pull/4097

lib/rubygems.rb
deivid-rodriguez: Let's leave it as it then since there was no further objections.

deivid-rodriguez:
Thanks @eregon!

PR #4076, opened 2020-11-20
Change `bundle install` with a lockfile to respect the `BUNDLED WITH` bundler version

duckinator:
This implements part of the [Bundler Version Locking](https://github.com/rubygems/rfcs/pull/29) RFC.

It's hard to ensure that all developers and environments of an application use the exact same bundler version. There's many reasons why that's a good thing, for example, wanting to enforce that a security release is used, so that nobody is vulnerable to a security issue.

My fix is to respect the `BUNDLED WITH` version if there is a lockfile. This section was originally introduced for this exact purpose but was never actually enforced.

The strategy is that when `bundle install` is run, if the running version does not match the `BUNDLED WITH` version, `bundler` will first install the `BUNDLED WITH` version, and then re-exec using that version.

This strategy is implemented as conservatively as we could:

* Other versions can still be specific with the `BUNDLER_VERSION` environment variable or with `bundle _<version>_`.
* If anything goes wrong when installing the `BUNDLED WITH` version, bundler will go on using the running version as a fallback.
* This feature is only applied when `bundler` is run in combination with the latest rubygems.
 

duckinator:
Sorry for all the noise. Apparently GitHub removed the original way to create a draft PR, and now you have to create a normal PR and click a link that's styled identically to the text around it. :upside_down_face:

deivid-rodriguez:
The way I normally do it is by clicking on the "arrow down" in the "Create Pull Request" green button. There you can select to create a draft PR instead.

duckinator:
@deivid-rodriguez that's what I usually do, but it's not there for me anymore lmao

deivid-rodriguez:
Weird, I just [created a draft PR](https://github.com/rubygems/rubygems/pull/4077) like that.

duckinator:
Either way, here's a screenshot showing this PR working! :3 

Screenshot_20201120_154127

duckinator:
Fixed some linting failures. (I'm having trouble getting rubocop to run locally, so had to rely on CI for that, heh.)

duckinator:
I fixed the linter errors but introduced new ones, the first time. I think it's right, now. :joy:

duckinator:
I have no idea what's causing those CI errors; will have to look into it another day.

duckinator:
Getting back into this again. Rebased off master.

duckinator:
Rebased off master. Also realized I finished an item from the TODO checklist at some point already, so checked that off.

duckinator:
Huh, okay, so there's no actual way to determine whether it was started using that mechanism or not. That makes me wonder if this all needs to be dragged up to the Bundler binstub.

Thanks for the info, @simi. I'll have to think a bit on how to approach this.

deivid-rodriguez:
@duckinator I'm taking over this PR now if you don't mind.

duckinator:
@deivid-rodriguez sorry for the delay in replying, but go for it. 👍

deivid-rodriguez:
Alright, just to document progress here. Things are green and a pretty satisfied with this as an initial version. The current approach is the following. Parse the `BUNDLED WITH` version and install that and `Kernel.exec` to it unless:

* There's no lockfile or no `BUNDLED WITH` section inside the lockfile.
* The running version matches the `BUNDLED WITH` version exactly.
* The lockfile is locked to a development version of bundler.
* An explicit `BUNDLER_VERSION` is set on the environment.

Another exception is that `bundle _<version>_` has been passed explicitly in the command line. However, binstubs completely remove that argument when they process, so we have no way to know whether `_<version>_` has been passed originally. The solution I came up with for that is to change the bundler binstub to explicitly set `BUNDLER_VERSION` when `_<version>_`. That way this case falls under the fourth bullet point above. It also prompted me to also require the latest version of rubygems (the one that generates this modified bundler binstubs) for this feature to be enabled.

In the next couple of days I want to wrap this up with:

* Some more tests.
* Rescue any error that I might've missed when installing bundler, and go on with the running bundler in that case (with some logging of the error I guess).
* Detect the situation where the running bundler location is not writable, and skip the feature there too, because we will try to install the new version alongside with the running bundler version and that will fail. I hope this will seamlessly skip this feature when using OS packaged rubies.
* Investigate why I needed https://github.com/rubygems/rubygems/pull/4076/commits/b398f8096138e3a7d93988f7f14372160b677ebb.

deivid-rodriguez:
Funny how normally tests are supposed to catch incorrect implementations. In this case, a correct implementation caught two bugs in the tests 😄.

deivid-rodriguez:
Alright, I'm setting this as ready. I will split a couple of changes to separate PRs but other than that this is ready.

deivid-rodriguez:
I will rebase this after merging #5179 and #5180, and then merge it, unless I receive some feedback before.

ngan:
Thanks is awesome. Thanks @duckinator and @deivid-rodriguez

duckinator:
@deivid-rodriguez thank you for making sure this PR got wrapped up and merged. I appreciate it a lot!

ysbaddaden:
Is it possible to DISABLE or OPT-OUT from that feature?

Bundler keeps on downgrading itself in my Docker images. I just want to use the bundler that is already available, not downgrade to the version from an older image :sob:

simi:
> Is it possible to DISABLE or OPT-OUT from that feature?
> 
> Bundler keeps on downgrading itself in my Docker images. I just want to use the bundler that is already available, not downgrade to the version from an older image sob

The proper way to handle is to update the bundler in `Gemfile.lock` to the one you use, in your case to the one you have in docker image.

PR #4077, opened 2020-11-20
Check whether some specs are still failing on Windows

deivid-rodriguez:
These specs mention a ticket that's closed. So either the issue is not really fixed, and we should reopen it, or these specs pass. I'm double checking that here. 


deivid-rodriguez:
It's not actually fixed, so I reopened the mentioned issue.

PR #4081, opened 2020-11-23
Let Net::HTTP decompress the index instead of doing it manually

### LINKED ISSUE #3860: Fails with stringio gem
Recently I updated `csv` gem in my project, that contains adding `stringio` gem https://github.com/ruby/csv/commit/09dd9f27717f2159a5c4a4949a50f91d05fb640f

The bundler fails with:
```
/home/runner/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0/gems/bundler-2.1.4/lib/bundler/runtime.rb:312:in `check_for_activated_spec!': You have already activated stringio 0.0.2, but your Gemfile requires stringio 0.1.3. Since stringio is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports stringio as a default gem. (Gem::LoadError)
```

This error appears on Github Actions pipeline, but I cannot reproduce it locally.

I'm using Ruby 2.6.6 with newest bundler `2.1.4`

eregon:
* Setting the Accept-Encoding header actually disables the automatic
  decompression as documented in Net::HTTP.
  https://github.com/ruby/ruby/blob/79a8ed07650dcbb36ec4b49a22596275e6c0fe23/lib/net/http/response.rb#L105-L107
  https://github.com/ruby/ruby/blob/c5eb24349a4535948514fe765c3ddb0628d81004/lib/net/http/request.rb#L11-L12
* This is also more memory efficient as Net::HTTP decompresses chunk
  by chunk instead of having the entire compressed String in memory
  and also wrapped in a StringIO which is particularly inefficient.
* See https://github.com/oracle/truffleruby/issues/2127#issuecomment-730492319
  for more details.

As a result, `bundler` no longer uses `stringio`, so the require for it can be removed avoid potential issues with the `stringio` default gem. So this PR also closes #3860.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -2,7 +2,6 @@
 
 require_relative "../vendored_fileutils"
 require "stringio"
-require "zlib"
 
 module Bundler
   class CompactIndexClient
@@ -45,18 +44,12 @@ module Bundler
               else
                 "bytes=#{local_temp_path.size}-"
               end
-          else
-            # Fastly ignores Range when Accept-Encoding: gzip is set
-            headers["Accept-Encoding"] = "gzip"
           end
 
           response = @fetcher.call(remote_path, headers)
           return nil if response.is_a?(Net::HTTPNotModified)
 
           content = response.body
-          if response["Content-Encoding"] == "gzip"
-            content = Zlib::GzipReader.new(StringIO.new(content)).read
-          end
 
           SharedHelpers.filesystem_access(local_temp_path) do
             if response.is_a?(Net::HTTPPartialContent) && local_temp_path.size.nonzero?
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative "../vendored_fileutils"
-require "stringio"
 
 module Bundler
   class CompactIndexClient
--- a/bundler/spec/bundler/compact_index_client/updater_spec.rb
+++ b/bundler/spec/bundler/compact_index_client/updater_spec.rb
@@ -17,7 +17,6 @@ RSpec.describe Bundler::CompactIndexClient::Updater do
     let(:response) { double(:response, :body => "abc123") }
 
     it "treats the response as an update" do
-      expect(response).to receive(:[]).with("Content-Encoding") { "" }
       expect(response).to receive(:[]).with("ETag") { nil }
       expect(fetcher).to receive(:call) { response }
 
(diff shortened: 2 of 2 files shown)


eregon:
Looks like there is a strange bug in RuboCop, I didn't touch those lines: https://github.com/rubygems/rubygems/pull/4081/checks?check_run_id=1442980333

eregon:
Could someone review this PR?

deivid-rodriguez (commented):
Thanks @eregon. I checked the `net/http` code and I think I get it.

If "Accept-Encoding" is set, like we're doing,  `net/http` sets a flag to avoid handling decoding of the response (`@decode_content = false`):

https://github.com/ruby/net-http/blob/71ba18910eab3876b153cf8e4f1cf1045ed69501/lib/net/http/generic_request.rb#L73

If it's not set, `net/http` sets it to `gzip` and sets the flag to handle it internally (`decode_content = true`):

https://github.com/ruby/net-http/blob/71ba18910eab3876b153cf8e4f1cf1045ed69501/lib/net/http/generic_request.rb#L34-L44

Then when it comes to handling the response, it skips decompression if the `@decode_content` flag is set, and also removes the `Content-Encoding` header from the response:

https://github.com/ruby/net-http/blob/71ba18910eab3876b153cf8e4f1cf1045ed69501/lib/net/http/response.rb#L257-L287

So in that case, the block below where we handle it ourselves will be skipped since `Content-Encoding` won't be set.

What I'm think is, since `net/http` also handles availability of `zlib` automatically, could we skip setting the flag altogether and also our handling of the response at 

https://github.com/rubygems/rubygems/blob/184c84d88e6deb9145b811d51b1e26462b6b5dca/bundler/lib/bundler/compact_index_client/updater.rb#L60-L62

?

eregon:
I think we could do that, except maybe for `Net::HTTP::HAVE_ZLIB == false` cases.
I'm not sure if that happens in practice, but it seems possible if e.g. the zlib C extension was not compiled.
OTOH if zlib was not compiled, it would already not work because there is `require "zlib"` at the top of this file, so maybe we don't care about that case and raise an exception if it's false?

And anyway, if zlib is not available, Net::HTTP would not set `Accept-Encoding: gzip`, so it seems we can safely remove anything related to zlib in this file, and let Net::HTTP do the right thing.
The only concern seems then to be if some older `net/http` doesn't properly handle that gzip stuff, I'll check it for Ruby 2.3.

deivid-rodriguez:
I think we just run into this issue after upgrading to truffleruby 20.3: https://github.com/rubygems/rubygems/runs/1658684625?

I rebased this and removed `zlib` usages.

I could probably have gone all the way and also remove the require and the rescue `Zlib::GzipFile`. I think the require can go too, and regarding the rescue, it seems useful to avoid some errors: https://github.com/rubygems/bundler/issues/6261. But if it's having an effect, it's probably hiding server side issues, so we might want to add a warning a leave it there for a while to make sure nobody complains before we removing it :man_shrugging:. Anyways, for another time. I'll check CI here with the current changs for now.

eregon:
Looks good to me. Yes, that should help with that error.

We're also working on fixing this in general in TruffleRuby, but that will need to wait the 21.2 release probably.
So it'd be nice to have this fix as a Bundler release, so that it's also applied when users install the latest Bundler on TruffleRuby.

eregon:
And thanks for rebasing and improving this PR! (I had it as a TODO but I have too many TODOs right now.)

deivid-rodriguez:
I think `stringio` is no longer used, so we can safely remove that require too, and this PR would also close #3860.

eregon:
Even better :+1:

deivid-rodriguez:
Thanks @eregon!

PR #4082, opened 2020-11-24
Support non-gnu libc linux platforms

### LINKED ISSUE #3174: Platform comparison mixing libc families
Hi! After #2918 and #2922, I've extended my own gem repository to contain glibc-linked gems also, and found out that in a machine with glibc, where the platform is `x86_64-linux`, bundler installs `x86_64-linux-musl` gems as if they were compatible.  I checked again and found out this on `rubygems/platform.rb`:

```ruby
  ##
  # Does +other+ match this platform?  Two platforms match if they have the
  # same CPU, or either has a CPU of 'universal', they have the same OS, and
  # they have the same version, or either has no version.
  #
  # Additionally, the platform will match if the local CPU is 'arm' and the
  # other CPU starts with "arm" (for generic ARM family support).

  def ===(other)
```

The "or either has no version" bit was the one that ends up comparing `x86_64-linux` to `x86_64-linux-musl` and finding them equal.  I patched the version check to `@version == other.version` and bundler installed the gem from the correct platform, but I'm wondering if it's ok to submit this patch as a PR since the documentation was so specific about it.  Could you clarify?  Thanks!

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

```
$ gem env version
3.0.3
```

lloeki:

Attempting to install a gem published as both `*-linux` and `*-linux-musl` results in the incorrect gem being picked up, causing build failures due to binary incompatibility. This is caused by the `nil` wildcard swallowing the libc information upon version comparison.

Followup to #2922
Closes #3174
    
Handle the `linux` case by performing only non-wildcard equality on the version and asserting `'gnu'` and `nil` equivalence, while preserving the current behaviour for other OSes.

Impact of this change is expected to be limited to none, as `*-linux` gems working on non-`*-linux` platforms such as *`-linux-musl` are vanishingly rare due to the stringent requirements and moves required to produce them. Therefore, as codified in the codebase preceding this pull request, it can be assumed that the set of gems published as `*-linux` is strictly equal to the set of gems that would be marked `-linux-gnu` only (if that were a thing).

Alternative considered: implement an additional, optional `@libc` attribute, thus setting the `@version` one to be always `nil` for an `@os` being `'linux'`. This would do away with some of the hacks in there, but would significantly alter the existing code for very little immediate value, and would break code currently depending on checking for `@version` being `musl` on recent rubygems.

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -66,7 +66,7 @@ class Gem::Platform
     when String then
       arch = arch.split '-'
 
-      if arch.length > 2 and arch.last !~ /\d/ # reassemble x86-linux-gnu
+      if arch.length > 2 and arch.last !~ /\d+(\.\d+)?$/ # reassemble x86-linux-{libc}
         extra = arch.pop
         arch.last << "-#{extra}"
       end
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -146,7 +146,8 @@ class Gem::Platform
   ##
   # Does +other+ match this platform?  Two platforms match if they have the
   # same CPU, or either has a CPU of 'universal', they have the same OS, and
-  # they have the same version, or either has no version.
+  # they have the same version, or either has no version (except for 'linux'
+  # where the version is the libc name, with no version standing for 'gnu')
   #
   # Additionally, the platform will match if the local CPU is 'arm' and the
   # other CPU starts with "arm" (for generic ARM family support).
@@ -162,7 +163,11 @@ class Gem::Platform
     @os == other.os and
 
     # version
-    (@version.nil? or other.version.nil? or @version == other.version)
+    (
+      (@os != 'linux' and (@version.nil? or other.version.nil?)) or
+      (@os == 'linux' and ((@version.nil? and other.version == 'gnu') or (@version == 'gnu' and other.version == nil))) or
+      @version == other.version
+    )
   end
 
   ##
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -165,7 +165,6 @@ class Gem::Platform
     # version
     (
       (@os != 'linux' and (@version.nil? or other.version.nil?)) or
-      (@os == 'linux' and ((@version.nil? and other.version == 'gnu') or (@version == 'gnu' and other.version == nil))) or
       @version == other.version
     )
   end
--- a/test/rubygems/test_gem_platform.rb
+++ b/test/rubygems/test_gem_platform.rb
@@ -145,6 +145,7 @@ class TestGemPlatform < Gem::TestCase
     test_cases.each do |arch, expected|
       platform = Gem::Platform.new arch
       assert_equal expected, platform.to_a, arch.inspect
+      assert_equal expected, Gem::Platform.new(platform.to_s).to_a, arch.inspect
     end
   end
 
--- a/test/rubygems/test_gem_platform.rb
+++ b/test/rubygems/test_gem_platform.rb
@@ -134,7 +134,9 @@ class TestGemPlatform < Gem::TestCase
       'i386-solaris2.8'        => ['x86',       'solaris',   '2.8'],
       'mswin32'                => ['x86',       'mswin32',   nil],
       'x86_64-linux'           => ['x86_64',    'linux',     nil],
+      'x86_64-linux-gnu'       => ['x86_64',    'linux',     nil],
       'x86_64-linux-musl'      => ['x86_64',    'linux',     'musl'],
+      'x86_64-linux-uclibc'    => ['x86_64',    'linux',     'uclibc'],
       'x86_64-openbsd3.9'      => ['x86_64',    'openbsd',   '3.9'],
       'x86_64-openbsd4.0'      => ['x86_64',    'openbsd',   '4.0'],
       'x86_64-openbsd'         => ['x86_64',    'openbsd',   nil],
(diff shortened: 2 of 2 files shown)


lib/rubygems/platform.rb
```diff
@@ -171,7 +172,11 @@ def ===(other)
     @os == other.os and
 
     # version
-    (@version.nil? or other.version.nil? or @version == other.version)
+    (
+      (@os != 'linux' and (@version.nil? or other.version.nil?)) or
+      (@os == 'linux' and ((@version.nil? and other.version == 'gnu') or (@version == 'gnu' and other.version == nil))) or
```
lloeki, line 177: This code should never happen as of today's platform parser. I included it as a defensive coding practice should the platform parser code evolve into not turning `*-linux-gnu` into `*-linux`.

lib/rubygems/platform.rb
```diff
@@ -71,7 +71,7 @@ def initialize(arch)
     when String then
       arch = arch.split '-'
 
-      if arch.length > 2 and arch.last !~ /\d/ # reassemble x86-linux-gnu
+      if arch.length > 2 and arch.last !~ /\d+(\.\d+)?$/ # reassemble x86-linux-{libc}
```
lloeki, line 74: This makes the code consistent with the line below in terms of what a "version" is:

```
if arch.length == 2 and arch.last =~ /^\d+(\.\d+)?$/ # for command-line
```

I did not pursue further change to limit the scope, but it shows possible room for future enhancements.

test/rubygems/test_gem_platform.rb
```diff
@@ -143,6 +145,7 @@ def test_initialize
     test_cases.each do |arch, expected|
       platform = Gem::Platform.new arch
       assert_equal expected, platform.to_a, arch.inspect
+      assert_equal expected, Gem::Platform.new(platform.to_s).to_a, arch.inspect
```
lloeki, line 148: An alternative test might be `assert_equal platform, Gem::Platform.new(platform.to_s), arch.inspect`

lib/rubygems/platform.rb
deivid-rodriguez: If this is not needed, I'd rather not include it. In any case, aren't you missing the "one of them is nil" case if os is linux? Also you're introducing a small coding style inconsistency be checking for `nil` with `== nil` instead of `nil?`.

deivid-rodriguez (commented):
Hi @lloeki, thanks so much for this!

I'm sorry it took me so long to look at it since it seems like a very interesting PR. As far as I understand, it unblocks `sassc-ruby` from providing a precompiled version. And `sassc-ruby` is the gem I'm aware of with the extension that takes the most to compile.

lib/rubygems/platform.rb
lloeki: > In any case, aren't you missing the "one of them is nil" case if os is linux?

Through these cases I did infer that in general `nil` is some sort of wildcard meaning "match anything" (e.g for `darwin` that would be "any darwin-XY version") but in the case of `linux` there's this historically introduced special case that `linux` == `linux-gnu` and nothing else.

That's the source of the overall issue:

- on a musl system such as Alpine (`linux-musl`) it means that current rubygems finds a match in gems marked as `linux` and can attempt to fetch+install them (even if there are `linux-musl` gems published), which is incorrect.
- on a glibc system (`linux` a.k.a `linux-gnu`) it means that current rubygems finds a match in gems marked as `linux-musl` and can attempt to fetch+install them (even if there are `linux` or theoretical `linux-gnu` gems published), which is incorrect.

> Also you're introducing a small coding style inconsistency be checking for nil with == nil instead of nil?

Good catch! Will fix...

lloeki:
> I'm sorry it took me so long to look at it since it seems like a very interesting PR

No worries :) I'm glad it's being considered!

This is a big conundrum, and the patch here presents an "ideal" version of a fix: I feel compelled to tell you that I am wondering how to handle the merge of this patch once merged.

Indeed people will eventually start pushing linux-musl gems (they don't currently because it's not working on musl, but once it does...), and non-up-to-date rubygems running on glibc (hence `linux`) may try to fetch `linux-musl` binaries, possibly wreaking havoc in currently working setups.

It all depends on what the gem server returns as a resultset and the order in which gem matches and dependencies are processed, but it does happen. You can try it with either `gem` or `bundle` against the `libv8-node` gem (which has `linux` and `linux-musl` versions published) to wrap your head around it.

> As far as I understand, it unblocks sassc-ruby from providing a precompiled version

That's one of the blockers, yes.

> And sassc-ruby is the gem I'm aware of with the extension that takes the most to compile.

I assume you have never tried to build `libv8` from source do you? ;) j/k, sassc is indeed a painful one. Nokogiri is a nice one too. In fact, anything involving C++ is generally a pain.

lloeki:
> Indeed people will eventually start pushing linux-musl gems (they don't currently because it's not working on musl, but once it does...), and non-up-to-date rubygems running on glibc (hence linux) may try to fetch linux-musl binaries, possibly wreaking havoc in currently working setups.

@deivid-rodriguez seems like my concerns are unwarranted, libv8-node has both `linux` and `linux-musl` gems published and everything works as expected:

```
# recent ruby
docker run --rm -it ruby:2.7 /bin/sh -c 'gem install libv8-node -v 15.5.1.0.beta1'
docker run --rm -it ruby:2.7 /bin/sh -c 'git clone https://github.com/lloeki/rubygems.git --branch support-non-gnu-libc-linux && cd rubygems && rake setup && ruby -Ilib bin/gem install libv8-node -v 15.5.1.0.beta1'
# old ruby (oldest supported by current rubygems)
docker run --rm -it ruby:2.3 /bin/sh -c 'gem install libv8-node -v 15.5.1.0.beta1'
docker run --rm -it ruby:2.3 /bin/sh -c 'git clone https://github.com/lloeki/rubygems.git --branch support-non-gnu-libc-linux && cd rubygems && rake setup && ruby -Ilib bin/gem install libv8-node -v 15.5.1.0.beta1'
# older ruby (no clone of branch because no support from current rubygems)
docker run --rm -it ruby:2.0 /bin/sh -c 'gem install libv8-node -v 15.5.1.0.beta1'
# on musl, without fix, picks the wrong one (expected)
docker run --rm -it ruby:2.3-alpine /bin/sh -c 'gem install libv8-node -v 15.5.1.0.beta1'
docker run --rm -it ruby:2.7-alpine /bin/sh -c 'gem install libv8-node -v 15.5.1.0.beta1'
# on musl, with fix, picks the correct one
docker run --rm -it ruby:2.3-alpine /bin/sh -c 'apk add build-base git && git clone https://github.com/lloeki/rubygems.git --branch support-non-gnu-libc-linux && cd rubygems && rake setup && ruby -Ilib bin/gem install libv8-node -v 15.5.1.0.beta1'
docker run --rm -it ruby:2.7-alpine /bin/sh -c 'apk add build-base git && git clone https://github.com/lloeki/rubygems.git --branch support-non-gnu-libc-linux && cd rubygems && rake setup && ruby -Ilib bin/gem install libv8-node -v 15.5.1.0.beta1'
```

So the fix and its ultimate impact presented here appears to work in all cases.

lib/rubygems/platform.rb
lloeki: Resolved by removing the unreachable condition.

lloeki:
@deivid-rodriguez resolved review and rebased onto master.

deivid-rodriguez:
@lloeki This is awesome, thanks for your work. I reviewed it and it looks just fine to me! :tada:.

Only thing I noticed is that it doesn't seem actually rebased? First commit before your changes is from last November, it seems? https://github.com/rubygems/rubygems/commits/26f3bd99973d97ecadf247f61a2cf0973dbb58dd.

Can you rebase it just to make sure it's all still :heavy_check_mark:? And while at it, I'd squash the last commit into the first one.

lloeki:
Huh. Not sure what I did there, maybe I forgot to rebase on the correct remote. Go figure, it should be correctly rebased now.

deivid-rodriguez (approved):
The squashing I suggested was not a big deal, so I'm merging this. Thanks so much for this fix @lloeki, I expected to release it very soon.

lloeki:
Ah sorry about the squash, I was scratching my head a bit to much about why I messed up the rebase and forgot about it ^^’. Thanks!

deivid-rodriguez:
No worries!

PR #4083, opened 2020-11-24
Update `set_rubyopt` in `Bundler::SharedHelpers` to support paths containing spaces

scotje:
In situations where Ruby is installed into a base path that contains one
or more spaces, the previous behavior could lead to Ruby attempting to
interpret the various components of the require path set via RUBYOPT as
distinct options, leading to errors like:

```
C:/Program Files/Puppet
Labs/DevelopmentKit/private/ruby/2.5.8/bin/ruby.exe: invalid switch in
RUBYOPT: -F (RuntimeError)
```

This change wraps the entire `-r` option in quotes to ensure that Ruby
interprets it all as a single value.

Since this is an implementation detail of a private method, I wasn't sure 
where the best place to add a test would be, but I'm happy to add tests if
someone can point me in the right direction.

### DIFF
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -303,7 +303,7 @@ module Bundler
 
     def set_rubyopt
       rubyopt = [ENV["RUBYOPT"]].compact
-      setup_require = "-r#{File.expand_path("setup", __dir__)}"
+      setup_require = "\"-r#{File.expand_path("setup", __dir__)}\""
       return if !rubyopt.empty? && rubyopt.first =~ /#{setup_require}/
       rubyopt.unshift setup_require
       Bundler::SharedHelpers.set_env "RUBYOPT", rubyopt.join(" ")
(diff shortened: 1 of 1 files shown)


scotje:
Looking into the failing specs....

scotje:
This seems to work if there is only one option passed via `RUBYOPT` but breaks down if multiple flags are being passed. :(

deivid-rodriguez:
Perhaps you want to `shelljoin` the different `RUBYOPT` components, instead of `.join(" ")`? I guess `bundler` specs won't pass on your system due to this issue, could you show an example of one of the failures, or alternatively, a way to reproduce on a similar windows system (I do have a Windows VM with spaces in ruby's path I believe).

deivid-rodriguez:
It sounds like `shellwords` here is going to be a problem because it activates the `shellwords` gem, which we don't want. Any updates on how to reproduce this issue?

deivid-rodriguez:
Since we got no further feedback here, we can't reproduce the issue, and the current changes break a lot of tests, I'll close this for now.

PR #4084, opened 2020-11-25
Fix `Bundler::Plugin::API::Source#to_s` having empty source type

ayanko:

I found this small bug during developing bundler plugin.

The fix is self explanatory and covered with spec.

- [x] ~Write code to solve the problem~

### DIFF
--- a/bundler/lib/bundler/plugin/api/source.rb
+++ b/bundler/lib/bundler/plugin/api/source.rb
@@ -260,7 +260,7 @@ module Bundler
         end
 
         def to_s
-          "plugin source for #{options[:type]} with uri #{uri}"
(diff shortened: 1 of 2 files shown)


bundler/lib/bundler/plugin/api/source.rb
```diff
@@ -260,7 +260,7 @@ def to_lock
         end
 
         def to_s
-          "plugin source for #{options[:type]} with uri #{uri}"
+          "plugin source for #{@type} with uri #{@uri}"
```
ayanko, line 263: Use instance variables to consistent with `#to_lock` implementation.

deivid-rodriguez (approved):
Thanks for this fix!

PR #4089, opened 2020-11-26
Remove Syck traces

voxik:
Syck can't be used as YAML engine since Ruby 2.2 and most of the support was removed already. This is attempt to remove the remaining traces. Let's see what happens.


voxik:
Hm, right. I have not executed test suite and there are a few test cases testing specifically the syck behavior.

deivid-rodriguez:
I think this is essentially a dup of https://github.com/rubygems/rubygems/pull/3797?

voxik:
You are right :see_no_evil:

voxik:
I think we differ only here:

https://github.com/rubygems/rubygems/pull/3797/files#diff-dd90140a6483cfb50fbbc3268f0d77984cd96e17792201e53bb3c0dc7ec435f9L627

Closing this now. Sorry for the nose.

deivid-rodriguez:
No problem! :+1:

PR #4093, opened 2020-12-03
Complete specs for newgem generator CHANGELOG.md

jlgoedert:

The discussion in [the original PR](https://github.com/rubygems/rubygems/pull/3908) welcomes a newcomer to resolve the leftover suggestions to get the original PR over the goal line. The original PR adds the option to include a CHANGELOG.md when running `bundle gem <GEM>`. Thank you @kieranklaassen for the excellent improvement and thank you for opening up another opportunity to contribute. 

I simply fixed a typo in the spec and followed the instructions to set the CHANGELOG global config to false in the main before hook. I also wrote the spec to verify there is a prompt (on the first run) that asks if the user would like to include a CHANGELOG.md.

### DIFF
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -158,6 +158,26 @@ RSpec.describe "bundle gem" do
     end
   end
 
+  shared_examples_for "--changelog flag" do
+    before do
+      bundle "gem #{gem_name} --changelog"
+    end
+    it "generates a gem skeleton with a CHANGELOG", :readline do
+      gem_skeleton_assertions
+      expect(bundled_app("#{gem_name}/CHANGELOG.md")).to exist
+    end
+  end
+
+  shared_examples_for "--no-changelog flag" do
+    before do
+      bundle "gem #{gem_name} --no-changelog"
+    end
+    it "generates a gem skeleton without a CHANGELOG", :readline do
+      gem_skeleton_assertions
+      expect(bundled_app("#{gem_name}/CHANGELOG.md")).to_not exist
+    end
+  end
+
   shared_examples_for "--rubocop flag" do
     before do
       bundle "gem #{gem_name} --rubocop"
@@ -904,6 +924,22 @@ RSpec.describe "bundle gem" do
       it_behaves_like "--rubocop flag"
       it_behaves_like "--no-rubocop flag"
     end
+
+    context "with changelog option in bundle config settings set to true" do
+      before do
+        global_config "BUNDLE_GEM__CHANGELOG" => "true"
+      end
+      it_behaves_like "--changelog flag"
+      it_behaves_like "--no-changelog flag"
+    end
+
+    context "with changelog option in bundle config settings set to false" do
+      before do
+        global_config "BUNDLE_GEM__CHANGELOG" => "false"
+      end
+      it_behaves_like "--changelog flag"
+      it_behaves_like "--no-changelog flag"
+    end
   end
 
   context "gem naming with underscore", :readline do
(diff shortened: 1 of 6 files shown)


bundler/spec/commands/newgem_spec.rb
```diff
@@ -157,6 +158,25 @@ def bundle_exec_rubocop
       end
     end
   end
+  shared_examples_for "--changelog flag" do
```
colby-swandale, line 161: Missing line between `end` & `shared_examples_for`

jlgoedert:
I'll try to get CI to pass over the weekend. Any advice is appreciated. I am guessing I misunderstood @deivid-rodriguez when he said:
> You need to add `"BUNDLE_GEM__CHANGELOG" => false` at the main before hook in the file to ensure that readline doesn't come into play in those tests, since it still seems to give some issues on Windows.

deivid-rodriguez:
@jlgoedert Maybe it's a different issue then. Another hint is that some new gem specs are tagged with `:readline` and I believe that's precisely to skip "Windows readline weirdness", so maybe you need to do the same for the specs that are failing.

jlgoedert:
Thank you! I'll give this a try in addition to some other investigation.

bundler/spec/commands/newgem_spec.rb
```diff
@@ -158,6 +159,26 @@ def bundle_exec_rubocop
...
+
+  shared_examples_for "--no-changelog flag" do
+    before do
+      bundle! "gem #{gem_name} --no-changelog"
+    end
+    it "generates a gem skeleton without a CHANGELOG", :readline do
+      gem_skeleton_assertions
+      expect(bundled_app("#{gem_name}/CHANGELOG.md")).to_not exist
+    end
+  end
```
deivid-rodriguez, line 180: @jlgoedert You're defining shared examples, but not actually using them! You need to `it_behaves_like "--changelog flag"` and `it_behaves_like "--no-changelog flag"`somewhere. Have a look at how it's done for `--rubocop` for example.

deivid-rodriguez (changes requested):
Noticed a small issue in the specs. Thanks for moving this forward!

bundler/spec/commands/newgem_spec.rb
jlgoedert: @deivid-rodriguez Nice catch! Actually testing this revealed a couple small issues that I corrected. As always, thanks for the feedback.

bundler/spec/commands/newgem_spec.rb
```diff
@@ -28,6 +28,7 @@ def bundle_exec_rubocop
   let(:require_path) { "mygem" }
 
   before do
+    global_config "BUNDLE_GEM__CHANGELOG" => "false"
```
deivid-rodriguez, line 31: Could you confirm whether this line is actually necessary? Since it's not needed for other analogous flags, I have the feelling that it's not needed here either.

deivid-rodriguez (commented):
Just one last little comment, but it's looking almost ready! :tada:

deivid-rodriguez:
Also, if you can, could you squash everything to a single commit?

bundler/spec/commands/newgem_spec.rb
jlgoedert: You are correct! Removed it 👍

jlgoedert:
@deivid-rodriguez I think everything is appropriately squashed. Thanks for the guidance!

deivid-rodriguez (approved):
This looks good to me!

simi (approved):
Good job!

deivid-rodriguez:
Thanks @jlgoedert for the work!  :tada: 

And thanks @simi, I had completely forgotten to merge this one! :sweat_smile:

Issue #4099, opened 2020-12-05
RubyGems ssl broken via outdated cert when system ca file missing.


MSP-Greg:
When Ruby master build is run, it may not have system certs available.  As such, it relies on the certs provided by RubyGems.  An outdated cert causes this to fail.

See https://github.com/ruby/ruby/pull/3854 for a fix.  Replaced expired AddTrustExternalCARoot.pem with GlobalSignRootCA_R3.pem.

Needs **`rake update_manifest`**

This issue is related to:

  - [x] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:
NA

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

PR #4100, opened 2020-12-06
Add GlobalSign Root CA - R3 cert and remove outdated certs

sonalkr132:
root CA of rubygems.org (and all subdomains) was updated from GlobalSign Organization Validation CA - SHA256 - G2 to GlobalSign Root CA - R3.
GlobalSignRootCA.pem was previously used to verify server cert if system certs could not verify rubygems.org cert.

Fixes when rubygem.org cert could not be verified by using system certs:
```
SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)                                          
```
closes: #4099 

add GlobalSign R3 CA cert. used [here](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/remote_fetcher.rb#L88) to configure remove fetcher.


hsbt (approved):

hsbt:
@sonalkr132 Thanks! I will backport this to RG 2.7, 3.0 and 3.1.

sonalkr132:
Thanks. Perhaps we should wait until @dwradcliffe confirms the removal of the old cert.
AddTrustExternalCARoot.pem was expired so it is fine to remove that. DigiCertHighAssuranceEVRootCA.pem is for cloudfront and I am positively sure that we don't have any CloudFront endpoints, however, a confirmation would be nice. It was added [more than seven years ago](https://github.com/rubygems/rubygems/blob/5ee6a59784b1736553e16fda374c18491bb66abe/lib/rubygems/ssl_certs/DigiCertHighAssuranceEVRootCA.pem).

hsbt:
Ah, OK. I'm waiting to release the new versions of rubygems until approval from @dwradcliffe .

indirect:
I can confirm that we previously used CloudFront as the S3 CDN, and we now use Fastly instead. It is ok to remove `DigiCertHighAssuranceEVRootCA.pem`. 👍🏻

dwradcliffe:
I think there was still a cloudfront domain setup for legacy clients buts it’s probably time to stop supporting that.
👍🏻

Issue #2900, opened 2019-08-30
gem outdated should list all newer versions of each outdated gem (instead of just the latest)


rgoldberg:
`gem outdated` should be able to list all newer versions of each outdated gem (instead of just the latest).

Maybe it should:

- only list the newest by default, but be able to list all newer via a switch
- list all newer by default, but be able to only list the newest via a switch
- always list all newer

This would be useful when the latest version does not support the running version of ruby.

The existing feature request #2575 would be nicer (yet more time consuming to implement) than this one.  This, however, might be a useful, quick to implement, stopgap feature.

This issue is related to:

  - [X] The command line `gem`

PR #4110, opened 2020-12-08
Respect `required_ruby_version` and `required_rubygems_version` constraints when looking for `gem install` candidates

deivid-rodriguez:

The end user problem is a long standing issue in the rubygems client. If your current ruby (similar with rubygems) version does not support the latest version of your target gem for installation, `gem install` will fail.

Instead, it should install the newest version that's compatible with your current ruby (and rubygems).

This has caused pain in the past in CI when we released a new version of bundler dropping support for old rubies, and suddenly CIs running `gem install bundler` would start failing on those rubies.

My fix is to start using the CompactIndex API when looking for dependencies, which provides this information.

This will be another step towards phasing out the old dependency API in the future, since rubygems will no longer use it.

Before this PR:

```
$ ruby -v
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-linux]

$ gem install activesupport 
ERROR:  Error installing activesupport:
	The last version of activesupport (>= 0) to support your Ruby & RubyGems was 5.2.4.4. Try installing it with `gem install activesupport -v 5.2.4.4`
	activesupport requires Ruby version >= 2.5.0. The current ruby version is 2.4.10.364.
```

After this PR:

```
$ ruby -v
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-linux]

$ gem install activesupport 
Successfully installed activesupport-5.2.4.4
1 gem installed
```

Currently I'm marking this PR as WIP, because I don't like this behaviour:

```
$ ruby -v
ruby 2.4.10p364 (2020-03-31 revision 67879) [x86_64-linux]

$ gem install activesupport:6.0.3
ERROR:  Could not find a valid gem 'activesupport' (= 6.0.3) (required by 'activesupport' (6.0.3)) in any repository
ERROR:  Possible alternatives: active_support-custom_logger, active_support-dependencies_patch, active_support-encrypted_redis_cache_store, active_support-lazy_load_patch, active_support_alias_class_method, activesupport, activesupport-cache-redis_multiplexer, activesupport-cascadestore, activesupport-core-ext, activesupport-current_attributes
```

Instead, I should get an error like the one you currently get when running `gem install activesupport` on ruby 2.4, telling you about your current ruby not being supported by the required activesupport version.

Fixes https://github.com/rubygems/rubygems/issues/2575.
Fixes https://github.com/rubygems/rubygems/issues/3089.

Closes https://github.com/rubygems/rubygems/issues/2900.
Closes https://github.com/rubygems/rubygems/pull/2655.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -62,6 +62,8 @@ jobs:
       - name: Check bundler install didn't hit the network
         run: if grep -q 'GET http' output.txt; then false; else true; fi
         working-directory: ./bundler
+      - name: Check rails can be installed
+        run: gem install rails
     timeout-minutes: 10
 
   install_rubygems_windows:
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -412,6 +412,7 @@ lib/rubygems/requirement.rb
 lib/rubygems/resolver.rb
 lib/rubygems/resolver/activation_request.rb
 lib/rubygems/resolver/api_set.rb
+lib/rubygems/resolver/api_set/gem_parser.rb
 lib/rubygems/resolver/api_specification.rb
 lib/rubygems/resolver/best_set.rb
 lib/rubygems/resolver/composed_set.rb
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -52,6 +52,7 @@ bundler/lib/bundler/cli/update.rb
 bundler/lib/bundler/cli/viz.rb
 bundler/lib/bundler/compact_index_client.rb
 bundler/lib/bundler/compact_index_client/cache.rb
+bundler/lib/bundler/compact_index_client/gem_parser.rb
 bundler/lib/bundler/compact_index_client/updater.rb
 bundler/lib/bundler/constants.rb
 bundler/lib/bundler/current_ruby.rb
--- a/bundler/lib/bundler/compact_index_client/cache.rb
+++ b/bundler/lib/bundler/compact_index_client/cache.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative "gem_parser"
+
 module Bundler
   class CompactIndexClient
     class Cache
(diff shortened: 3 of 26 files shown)


indirect:
Hooray! Great work getting this set up. 😄

deivid-rodriguez:
Thanks @indirect! :heart:

I improved the error message when we explicitly try to install a gem that doesn't satisfy the current ruby version. For example, `gem install activesupport:6.0.3` on ruby 2.4. Now we get:

```
$ gem install rails:6.0.3
ERROR:  Error installing rails:
	rails-6.0.3 requires Ruby version >= 2.5.0. The current ruby version is 2.4.10.364.
```

I also cherry-picked several tests from #2655.
 
Finally, I added several changes to restore some unintended behavioural changes that I had made.

I'll do another pass but it should be mostly ready now!

deivid-rodriguez:
I updated the PR to make more explicit the code from `bundler` being duplicated in rubygems. I believe it should be ready now, and I'd like to ship this fix with rubygems 3.2.3 before the final ruby release.

simi:
Shouldn't 2.3.0 version fit better here? IMHO it is not minor change valid for patch release.

deivid-rodriguez:
It is a bug fix :man_shrugging:.

What I'd like to accomplish to releasing it with rubygems 3.2.3 is that we can announce the fix and recommend people using `gem install bundler` in the workflows running old rubies to upgrade rubygems. That way when we drop support for old rubies in master,  and release a new version with ruby 2.3 and ruby 2.4 support. The workflows of these people won't break.

But if you consider this very risky, we can delay it.

deivid-rodriguez:
Basically, have a version out there that supports ruby 2.3 and ruby 2.4 and has this fix.

deivid-rodriguez:
@simi What should we do, then? Alternatively to shipping it now, we can also hold it until rubygems 3.3.0, and hold support for 2.3 and 2.4 as well until after 3.3.0. That way we also achieve my preference of shipping a version that supports ruby 2.3 and ruby 2.4 and has this fix.

But I still think this is a bug fix that we could ship now. I have the ability to underestimate how changes can be breaking for people, but I would really expect this one to be smooth.

simi:
:thinking: @deivid-rodriguez this would be really nice to get released with Ruby 3.0 if that's still possible.

deivid-rodriguez:
@simi So what are you suggesting that we do? Ship this now, but call the new version 3.3.0? Or are you ok with shipping this as a bug fix?

simi:
IMHO 3.3.0 would be better if that is not breaking any of the following plans (including it into 3.0 Ruby for example).

deivid-rodriguez:
But then we can't satisfy @hsbt's preference of releasing bundler & rubygems "in sync" unless we push an artificial bundler 2.3.0 release without new features.

deivid-rodriguez:
If this PR introduces some regression, it won't matter whether we call this a bug fix, or a feature, or whether we include it in 3.2.3 or 3.3.0, right? I will be a bug that we will have to fix.

simi:
Sure, I'm not going to block this actually just because of versioning. Let's ship it somehow :pray:

deivid-rodriguez:
Great, let's do this then!

PR #4113, opened 2020-12-09
Fix Resolver::APISet to always include prereleases when necessary

### LINKED ISSUE #4112: gem command neglects --pre option

My current problem is "gem command neglects --pre option".

---
I've run `gem i rails --pre` command on Ruby 3.0.0preview2 and gem v.3.2.0, gem command installed rails-6.0.3.4 (not rails 6.1.0.rc2).
It looks that gem command neglect --pre option.

```
$ gem i rails --pre
Fetching rails-6.0.3.4.gem
Successfully installed rails-6.0.3.4
1 gem installed

$ ruby -v
ruby 3.0.0preview2 (2020-12-08 master d7a16670c3) [x86_64-darwin19]

$ gem -v
3.2.0

$ gem update --system
Latest version already installed. Done.
```

- Ruby: 3.0.0preview2
  - rbenv, ruby-build
- gem: 3.2.0
- OS: MacOS 10.15.7

In Ruby 2.7.2 and gem v.3.1.4, gem command installs rails 6.1.0.rc2 as intended.

---

This issue is related to:

  - [ ] Network problems
  - [x] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
$ gem env version
3.2.0
```

deivid-rodriguez:

```
gem install rails --pre
```

fails to install the latest rails prerelease.

Previous versions of molinillo, specifically before https://github.com/CocoaPods/Molinillo/commit/d29a053fe9ee6f07b1d7d6f501f882fa307c9390, would include special logic for dealing with prereleases. Now this logic needs to be implemented by the specification provider, in this case, rubygems.

That change in molinillo made an issue in our `Resolver::APISet` implementation manifest, where even if the set has been created with the `prerelease` flag, it would fail to return prereleases unless the dependency explicitly allowed them. For example, it would return prereleases given the dependency "activesupport, = 6.1.rc2", but it will fail to return prereleases given the dependency "activesupport, >= 4.2.0". In the latter case, we also want prereleases returned when the `--pre` flag has been passed to `gem install`, i.e., when the API set has been created with the prerelease flag.

Fixes #4112.

### DIFF
--- a/lib/rubygems/resolver/api_set.rb
+++ b/lib/rubygems/resolver/api_set.rb
@@ -52,7 +52,7 @@ class Gem::Resolver::APISet < Gem::Resolver::Set
     end
 
     versions(req.name).each do |ver|
-      if req.dependency.match? req.name, ver[:number]
+      if req.dependency.match? req.name, ver[:number], @prerelease
         res << Gem::Resolver::APISpecification.new(self, ver)
       end
     end
--- a/test/rubygems/test_gem_resolver_api_set.rb
+++ b/test/rubygems/test_gem_resolver_api_set.rb
@@ -55,6 +55,35 @@ class TestGemResolverAPISet < Gem::TestCase
     assert_equal expected, set.find_all(a_dep)
   end
 
+  def test_find_all_prereleases
+    spec_fetcher
+
+    data = [
+      { :name         => 'a',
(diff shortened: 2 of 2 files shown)


hsbt (approved):
👍

PR #4121, opened 2020-12-10
Better CI job names

deivid-rodriguez:

Our CI configuration uses exact versions. This helps us with the determinism of our CI, making it less dependent on external changes, following the principle that CI should fail if and only if the code inside a PR breaks something.

However, since exact versions are used to build job names, that means every time we bump the tested versions, the job names change and we have to change our protected branch configuration to enforce the checks with the new names.

With these changes I expect to keep the best of both worlds: CI determinism, and stable and descriptive job names.

### DIFF
--- a/.github/workflows/daily-bundler.yml
+++ b/.github/workflows/daily-bundler.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   daily_bundler:
+    name: Bundler (ruby-head)
     runs-on: ubuntu-18.04
     if: github.repository == 'rubygems/rubygems'
     env:
--- a/.github/workflows/daily-rubygems.yml
+++ b/.github/workflows/daily-rubygems.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   daily_rubygems:
+    name: Rubygems (ruby-head)
     runs-on: ubuntu-18.04
     if: github.repository == 'rubygems/rubygems'
     strategy:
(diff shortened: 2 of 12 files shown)


simi (approved):
> However, since exact versions are used to build job names, that means every time we bump the tested versions, the job names change and we have to change our protected branch configuration to enforce the checks with the new names.

If I understand it well, this is the reason why this PR hangs for jobs to wait, which will never happen since they are renamed now. Let's :pray: after merge last settings tweak will be needed because of this.

deivid-rodriguez:
Yes, in order to make this PR get a green state, I'll have to go to the protected branches configuration and change all of the required checks to be the new ones.

But this is the last time we'll have to do this! :tada:

deivid-rodriguez:
imagen

:tada: :tada: :tada: :tada: :tada: :tada:

deivid-rodriguez:
I'll go ahead now, uncheck a bunch of checkboxes, then check a bunch of checkboxes and then merge :)

deivid-rodriguez:
Oh wow, I changed the windows bundler workflow to use exact versions, since it was the only workflow not using them and... I got a segfault :open_mouth:.

deivid-rodriguez:
Not sure what happened there, I suspect a bad tools cache in the github actions environment. I'm going to ignore it since I don't think it will come up again.

Issue #4130, opened 2020-12-12
Got "NoMethodError: undefined method `version' for nil:NilClass" when running "bundle outdated --strict"


jasnow:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/Users/USER/.rvm/rubies/ruby-2.7.2/bin/bundle outdated --strict`
  This command has been is a script I have been using for a LONG time.

- **What did you expect to happen?**

  I expected Bundler to "Only list newer versions allowed by your Gemfile requirements"

- **What happened instead?**

  Instead, what happened was got error below and backtrace.

- **Have you tried any solutions posted on similar issues in 
  our issue tracker, stack overflow, or google?**

  Had problem searching but I tried. Did not see my problem.

- **Have you read our issues document, 
  https://github.com/rubygems/rubygems/blob/master/bundler/doc/contributing/ISSUES.md?**

  Scanned it.

## Backtrace

```
NoMethodError: undefined method `version' for nil:NilClass
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/cli/outdated.rb:81:in `block in run'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/cli/outdated.rb:75:in `each'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/cli/outdated.rb:75:in `run'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/cli.rb:442:in `outdated'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
  /Users/USER/.rvm/rubies/ruby-2.7.2/lib/ruby/site_ruby/2.7.0/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
[...]

PR #4133, opened 2020-12-13
Fix `bundle outdated --strict` crash

deivid-rodriguez:

This command would crash when there are some dependencies not used under
the current platform.

Check for `nil` at the proper place.

Fixes https://github.com/rubygems/rubygems/issues/4130.

### DIFF
--- a/bundler/lib/bundler/cli/outdated.rb
+++ b/bundler/lib/bundler/cli/outdated.rb
@@ -76,6 +76,8 @@ module Bundler
         next unless gems.empty? || gems.include?(current_spec.name)
 
         active_spec = retrieve_active_spec(definition, current_spec)
+        next unless active_spec
+
         next unless filter_options_patch.empty? || update_present_via_semver_portions(current_spec, active_spec, options)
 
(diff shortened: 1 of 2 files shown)


simi:
Nice, I wasn't sure this is intended. Since if I understand it now, outdated gems with currently unsupported platform would not be reported, but they were before 2.2.0, which (I'm not sure) could be breaking change.

deivid-rodriguez:
Ha, good catch! I assumed they were ignored in the past too. Definitely a less useful behaviour, I'll have a look and see if I can fix it.

deivid-rodriguez:
@simi I double checked this and bundler 2.1.4 has the same behaviour. That behaviour could even be considered a bug, because `bundle update --lock` updates the lockfile for all platforms. But since the behaviour is unchanged from 2.1.4, I'd ship this to fix the crash, and open a separate ticket to fix the behaviour.

simi (approved):

deivid-rodriguez:
Thanks @simi!! :tada:

Issue #3926, opened 2020-09-02
Changing the default of `bundle config set` to be --local instead of --global


eregon:
### Describe the problem as clearly as you can

`bundle config set key value` is currently `--global` by default (stored in `~/.bundle/config`), while I think most usages would benefit from being local.

For instance, typical mistakes are `bundle config path vendor/bundle`, `bundle config without db`, etc.

Note also that the very similar command `git config key value` defaults to `--local`.

I think it would be much safer and intuitive to change `bundle config set key value` to also default to `--local`.
Most of Bundler configuration is project-specific and doing it globally makes often little sense, especially when not stating explicitly one wants it to be global.

Extracted from https://github.com/rubygems/rubygems/issues/3916

eregon:
I think the user's expectation is that running pretty much any `bundle` command in a project with a Gemfile only affects that project, and nothing else.
I think system-wide changes such as global config should always be explicit, as otherwise they can be very confusing.

eregon:
Thanks for fixing this!

Issue #3708, opened 2020-06-11
Calling bundle install with BUNDLE_WITH runtime env var persists it in .bundle/config


mokagio:
I checked all the following documents and couldn't find a solution for my issue:

- [x] [Troubleshooting common issues](https://github.com/rubygems/rubygems/blob/master/bundler/doc/TROUBLESHOOTING.md)
- [x] [Bundler documentation site](https://bundler.io/)
- [x] [Bundler man pages](https://bundler.io/man/bundle.1.html)
- [x] [Bundler command line reference](https://bundler.io/v2.0/commands.html)

### 1. Do you have a reproduction script?

https://gist.github.com/mokagio/7dfcd75ffc2fac278c2bde4862c3164b

### 2. What are you trying to accomplish?

We use some Ruby tooling across the Automattic iOS apps. We want to achieve two things:

- Vendor our gems in a folder within the repository, `vendor/bundle`
- Have an optional group in the `Gemfile`, which should be used only by platforms devs and CI.

This setup results in an unexpected change to the `.bundle/config` when installing the optional group, more on it below.

The need for the optional group is because we use `rmagick ~> 3.2.0`, which requires an older version of ImageMagick. We don't want the product devs to have to worry about setting up their machine correctly to support it.

### 3. What command did you run?

In order to make sure gems are vendored, I tried to track `.bundle/config` in the repo. 

```
---
BUNDLE_PATH: "vendor/bundle"
```

In order to install the optional group, I run

```
BUNDLE_WITH=optional_group bundle install
```

### 4. What were you expecting to happen?

I was expecting the installation to be successful, including the gem in the optional group, and no change to occur in `.bundle/config`.

5. What actually happened?

The installation was successful, including the gem in the optional group, but `.bundle/config` changed, like this:

```diff
 ---
 BUNDLE_PATH: "vendor/bundle"
+BUNDLE_WITH: "optional_group"
```

I wasn't expecting that. My understanding of environment variables is that they can be used to override configurations at runtime. I don't understand why `bundle  install` updated its config file with the environment value I passed it.

Using `bundle install --with optional_group`, which is a deprecated approach, has the same result.

The [docs](https://bundler.io/v2.0/bundle_config.html#REMEMBERING-OPTIONS) say:

> Flags passed to `bundle install` or the Bundler runtime, such as `--path` foo or `--without` production, are **not remembered between commands**. If these options must be remembered, they must be set using bundle config (e.g., `bundle config path foo`).
[...]

PR #4152, opened 2020-12-15
Fix `bundle config` inside an application saving configuration globally

deivid-rodriguez:

Users expect `bundle config` to save configuration to the local application when `bundle config` is run on an application context. This is because other tools work like that (`git`), and because usual app specific configurations (`without`, `path`) are no generally desired to be applied to all the apps of a user.

Also, there's no documentation anywhere about it being global by default.

Someone could be relaying on this, but I'm going to call this a bug, and just change the behavior, because we keep getting issues about it.

Change `bundle config` to be local by default like everyone expects.

Fixes https://github.com/rubygems/rubygems/issues/3926.
Fixes https://github.com/rubygems/rubygems/issues/4142.
Fixes https://github.com/rubygems/rubygems/issues/4813.

### DIFF
--- a/bundler/lib/bundler/cli/config.rb
+++ b/bundler/lib/bundler/cli/config.rb
@@ -180,7 +180,7 @@ module Bundler
         scopes = %w[global local].select {|s| options[s] }
         case scopes.size
         when 0
-          @scope = "global"
+          @scope = inside_app? ? "local" : "global"
           @explicit_scope = false
         when 1
           @scope = scopes.first
@@ -188,6 +188,15 @@ module Bundler
           raise InvalidOption,
             "The options #{scopes.join " and "} were specified. Please only use one of the switches at a time."
         end
+      end
+
+      private
+
+      def inside_app?
+        Bundler.root
+        true
+      rescue GemfileNotFound
+        false
       end
     end
   end
--- a/bundler/spec/commands/config_spec.rb
+++ b/bundler/spec/commands/config_spec.rb
@@ -43,6 +43,12 @@ RSpec.describe ".bundle/config" do
       G
     end
 
+    it "is local by default" do
+      bundle "config set foo bar"
+      expect(bundled_app(".bundle/config")).to exist
+      expect(home(".bundle/config")).not_to exist
+    end
+
     it "can be moved with an environment variable" do
       ENV["BUNDLE_APP_CONFIG"] = tmp("foo/bar").to_s
       bundle "config set --local path vendor/bundle"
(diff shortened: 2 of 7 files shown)


deivid-rodriguez:
Ok, some some specs _do break_ when changing this, although none of them is explicitly testing this behaviour so it's just and don't show any reasonable usages of `bundle config` being global by default.

The last spec I fixed at https://github.com/rubygems/rubygems/pull/4152/commits/1c509c444289674d6898a11002086affb9216343 _does_ show a "reasonable" use. Say, you have two `Gemfile` files, one at the root of your project and one at `./gemfiles/rails_52/Gemfile`. For each of these folders, `bundle config --local` will be saved to `.bundle/config` and `./gemfiles/rails_52/.bundle/config` respectively. So `bundle config` being global by default allows this case to automatically "share" configurations for all `Gemfiles`, which is normally what you want.

So, I see two ways forward:

* Change this PR to apply only to bundler 3, and create a separate PR that restores the behaviour of copying global `without` configurations to the local configuration when `bundle install` is run (with a deprecation warning).

* Keep going down the rabbit whole and change the criteria to choose the base folder for the configuration. Right now it's: folder of the first `Gemfile` up in the hierarchy, and if not found, folder of the first `.bundle` folder up in the hierarchy. If we reversed this and considered first the presence of `.bundle` folders, the behaviour of the multi-Gemfile setup explained above would be preserved I think.

I'm a bit hesitant here, but I currently lean towards the first option.

deivid-rodriguez:
Actually for the second option, the idea I was developing in https://github.com/rubygems/rubygems/pull/3963 would work too.

deivid-rodriguez:
I want to get the fix for #4142 out soon, so I went with the first option. I created #4154 to revert the change that caused the regression, and then I'll update this PR to change `bundle config` default.

deivid-rodriguez:
Alright, I rebased this and added specs.

I was thinking of just moving this forward and assume any potential incompatibility, but I think we can detect the situations when this would be backwards incompatible and keep the global default in that case with a warning. I will work on that tomorrow.

deivid-rodriguez:
The potential incompatibilities are very low here, since this is the behavior everybody expects, so I will move this forward. After merging this, I will reapply the fix for https://github.com/rubygems/rubygems/issues/3708 and https://github.com/rubygems/rubygems/issues/4835. The only reason that fix was reverted was that, due to the current weird priority, it generated some regressions, but as soon as this priority gets fixed, it should cause no problems.

PR #4162, opened 2020-12-16
Reuse otp passed by user in api request retry

sonalkr132:

Fixes failed commands with access update or fist invocation (no local API key):
```
$ gem yank somegem -v 0.0.1
Don't have an account yet? Create one at http://localhost:3000/sign_up
   Email:   some@rubygems.org
  Password:   
Code:   320292
Added yank_rubygem scope to the existing API key
You have enabled multifactor authentication but no OTP code provided.
Please fill it and retry.
```

reuse otp passed by the user in api request retry

### DIFF
--- a/lib/rubygems/test_utilities.rb
+++ b/lib/rubygems/test_utilities.rb
@@ -38,7 +38,7 @@ class Gem::FakeFetcher
     @paths = []
   end
 
-  def find_data(path, nargs = 3)
+  def find_data(path)
     return Gem.read_binary path.path if URI === path and 'file' == path.scheme
 
     if URI === path and "URI::#{path.scheme.upcase}" != path.class.name
@@ -54,10 +54,11 @@ class Gem::FakeFetcher
       raise Gem::RemoteFetcher::FetchError.new("no data for #{path}", path)
     end
 
-    data = @data[path]
-
-    data.flatten! and return data.shift(nargs) if data.respond_to?(:flatten!)
-    data
+    if @data[path].kind_of?(Array) && @data[path].first.kind_of?(Array)
+      @data[path].shift
+    else
+      @data[path]
+    end
   end
 
   def fetch_path(path, mtime = nil, head = false)
--- a/test/rubygems/test_gem_commands_push_command.rb
+++ b/test/rubygems/test_gem_commands_push_command.rb
@@ -404,11 +404,13 @@ class TestGemCommandsPushCommand < Gem::TestCase
     assert_equal '111111', @fetcher.last_request['OTP']
   end
 
-  def test_sending_gem_unathorized_api_key
+  def test_sending_gem_unathorized_api_key_with_mfa_enabled
+    response_mfa_enabled = "You have enabled multifactor authentication but your request doesn't have the correct OTP code. Please check it and retry."
     response_forbidden = "The API key doesn't have access"
     response_success   = 'Successfully registered gem: freewill (1.0.0)'
 
     @fetcher.data["#{@host}/api/v1/gems"] = [
+      [response_mfa_enabled, 401, 'Unauthorized'],
       [response_forbidden, 403, 'Forbidden'],
       [response_success, 200, "OK"],
     ]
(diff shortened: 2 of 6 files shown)


sonalkr132:
The first invocation has one more quirk, which exists in previous versions as well:
```
Don't have an account yet? Create one at http://localhost:3000/sign_up
   Email:   some@rubygems.org
Password:   

Code:   569043
Signed in.
Pushing gem to http://localhost:3000...
Code:   569043
Successfully registered gem: sometsomeee (3.10.50.15)
```
It asks for OTP twice because `sign_in` and real execution are separate invocations([1](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/commands/push_command.rb#L64-L66)). Is it okay to use `options[:otp]` as an instance variable, which will be set in `sign_in`?

lib/rubygems/test_utilities.rb
```diff
@@ -54,10 +54,11 @@ def find_data(path, nargs = 3)
...
 
-    data = @data[path]
-
-    data.flatten! and return data.shift(nargs) if data.respond_to?(:flatten!)
-    data
+    if @data[path].kind_of?(Array) && @data[path].first.kind_of?(Array)
+      @data[path].shift
+    else
+      @data[path]
+    end
```
sonalkr132, line 61: this was needed so that `@fetcher.data["#{@host}/api/v1/gems"]` can return more than two responses. `find_data` seems a bit overloaded. `data.flatten! and return data.shift(nargs) if data.respond_to?(:flatten!)` was my previous last-ditch effort of supporting more than one response from faker. does this need something more elegant?

deivid-rodriguez: I think we could improve it in the future but what you did here is more flexible and no less elegant than the previous solution, so I'm +1 on what you did.

deivid-rodriguez:
> It asks for OTP twice because sign_in and real execution are separate invocations(1). Is it okay to use options[:otp] as an instance variable, which will be set in sign_in?

It seems ok to me!

Something I don't understand is how the scope is getting properly updated now if it needs an otp code too?

sonalkr132:
update scope was promoting for otp, through `rubygems_api_request`. This version has one fewer prompt (and request) for update scope case.

deivid-rodriguez:
Aahhhh right, sorry, I get it now.

deivid-rodriguez (approved):
This looks good to me!

Another quirk I see (I haven't tried it but I think it happes from reading the code) is that if `--opt` is given on the command line, the given code will be used for sign_in, but then other requests will still ask for an otp code, right?

If we implement your instance variable suggestion, we're it's filled up lazily from either `options[:otp]` or asking the user, I think both cases would be fixed?

In any case, this PR keeps existing behaviour and fixes an issue,  so if you want to merge this now and fix the other cases later, I'm good with it!

deivid-rodriguez:
@sonalkr132 Not sure if you're still there, but if you give me the OK now, I'll sneak this into the 3.2.2 release.

sonalkr132:
> the given code will be used for sign_in, but then other requests will still ask for an otp code, right?

not really. for them the otp is passed from the command file itself ([1](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/commands/push_command.rb#L94)). `if mfa_unauthorized?(response)` is basically fallback.

```
RUBYGEMS_HOST="https://staging.rubygems.org" gem push sometsomeee-3.10.50.15.gem  --otp 487877
Enter your https://staging.rubygems.org credentials.
Don't have an account yet? Create one at https://staging.rubygems.org/sign_up
   Email:   aditya.prakash132@gmail.com
Password:   

Signed in.
Pushing gem to https://staging.rubygems.org...
Successfully registered gem: sometsomeee (3.10.50.15)
```

test/rubygems/test_gem_commands_push_command.rb
```diff
@@ -417,17 +419,54 @@ def test_sending_gem_unathorized_api_key
...
+    end
+
+    mfa_notice = "You have enabled multi-factor authentication. Please enter OTP code."
+    assert_match mfa_notice, @ui.output
+    assert_match "Enter your https://rubygems.example credentials.", @ui.output
+    assert_match "Email:", @ui.output
+    assert_match "Password:", @ui.output
+    assert_match "Signed in with API key:", @ui.output
+    assert_match response_success, @ui.output
+    assert_equal '11111', @fetcher.last_request['OTP']
```
sonalkr132, line 469: this assert fails with `nil` without b09e93208f039e2021bd7d0e26cac2ffb808b3de

deivid-rodriguez (approved):
Looks good to me @sonalkr132!

deivid-rodriguez:
I'd like to rename the PR to something more close to the use problem being fixed, maybe: "Fix issue with where CLI commands making more than one request to rubygems.org needing an OTP code would crash or ask for the code twice"?

deivid-rodriguez:
Anyways, merging.

Thanks for fixing this and improving the code all around!

deivid-rodriguez:
To clarify why I'd like to rename the PR, our scripts use the PR title to generate the changelog, so the title of the PR is what the users will see when checking the changelog. So the idea would be that PR titles are end user focused. No big deal though!

PR #4168, opened 2020-12-17
Use a CHANGELOG in markdown for rubygems

deivid-rodriguez:

The problem is very minor, but github releases doens't look nicely formatted since github doesn't understand our current format (I think it's rdoc).

My fix is to use markdown since it's more standard these days and it's what bundler uses so helps unifying things.


hsbt (commented):
👍

simi (approved):

Issue #4166, opened 2020-12-17
Bundler 2.2.1, multiple platforms and bundle install --frozen


knu:
### Describe the problem as clearly as you can

Bundler 2.2+ started to output platform-specific gem entries to Gemfile.lock, and it seems `bundle install --frozen` installs gems for the specified platform regardless of the running platform, if the lock file does not include entries for their variants appropriate for the running platform.

We have a project targeted to x86_64 Linux where the majority of developers are on macOS.  We have been sharing a single Gemfile.lock in the repository for both development and deployment.  Here's our development workflow that may suffer from the new behavior:

1. A developer adds a new gem or upgrades an existing gem on macOS and creates a pull request.  The gem has a binary variant for macOS, thus Gemfile.lock includes an entry that goes `some-gem (x.y.z-universal-darwin)` and none for `x86_86-linux`.
2. CI and deployment on Linux fail because Gems are installed with `bundle install --frozen` for strict integrity there.  The command runs and exits successfully, but the `some-gem` installed by Bundler is for macOS (`universal-darwin`), so the test suite and application fail to load the gem and do not work.

Disabling the frozen option may be an option (in exchange for a strict integrity check), but still, we cannot easily maintain a universal Gemfile.lock across platforms.  This may be another story, though.

### Post steps to reproduce the problem

1. Create a Gemfile.lock containing a binary gem entry targeted for a specific platform, run `bundle` there, and check in Gemfile.lock.
2. Check out Gemfile and Gemfile.lock on another platform and run `bundle install --frozen`.
3. See from the output that it has only installed gems for the first platform.  Or run `bundle exec gem list`.

### Which command did you run?

`bundle install --frozen` (`bundle config set --local frozen true && bundle`)

### What were you expecting to happen?

It should refuse to install gems for other (incompatible) platforms, and exit with a non-zero status code.

### What actually happened?

It installed alien binary gems that wouldn't load or work, without an error.

### If not included with the output of your command, run `bundle env` and paste the output below

```
Bundler       2.2.1
  Platforms   ruby, x86_64-darwin
Ruby          2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin]
  Full Path   /Users/knu/arch/darwin/mach/x86_64/rbenv/versions/2.6.6/bin/ruby
  Config Dir  /Users/knu/arch/darwin/mach/x86_64/rbenv/versions/2.6.6/etc
RubyGems      3.1.2
[...]

PR #4172, opened 2020-12-17
Fix bundler installing gems for a different platform when running in frozen mode and current platform not in the lockfile

deivid-rodriguez:

Since bundler 2.2.1, bundler could end up installing dependencies not valid for the current platform.

RUBY is no longer included as a platform in the lockfile, however it's still included in the list of rubygems platforms, so the check was not doing the right thing.

Instead, use our own logic.

Fixes https://github.com/rubygems/rubygems/issues/4166.

### DIFF
--- a/bundler/lib/bundler/cli/update.rb
+++ b/bundler/lib/bundler/cli/update.rb
@@ -82,7 +82,7 @@ module Bundler
           locked_spec = locked_info[:spec]
           new_spec = Bundler.definition.specs[name].first
           unless new_spec
-            if Bundler.rubygems.platforms.none? {|p| locked_spec.match_platform(p) }
+            unless locked_spec.match_platform(Bundler.local_platform)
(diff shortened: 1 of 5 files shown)


indirect (approved):

deivid-rodriguez:
I noticed that this error can only happen in frozen mode, so I think it'd be worth mentioning that as the cause of the error. I'll push an update.

However, I'm not sure about remediation steps. It's somewhat tricky because `bundle lock --add-platform x86_64-linux` is not going to work in frozen mode (I wonder if a command explicitly meant to update the lockfile should override frozen mode, but that's another story). So we'll need to mention that the command needs to be run in a non frozen environment, or maybe mention `bundle config unset frozen && bundle lock --add-platform x86_84-linux && bundle config set --local frozen true` directly?

simi (approved):

deivid-rodriguez:
I'll merge this now, and we can further tweak it later.

Issue #4153, opened 2020-12-15
Bundler 2.2.1 - bundle package fails


sterankin:
Our CI build broke today with the below error.  Its seems the Docker image is install bundler 2.2.1 and when we run docker package, it generates the following error report:

I was able to fix by installing bundler 2.1.4 and using that in my CI build, but I am hoping someone can help with the below.

Ruby 2.5.8

**What did you do?**

I ran the command `/usr/local/bundle/bin/bundle package`

**What did you expect to happen?**

I expected Bundler to bundle package as normal

 **What happened instead?**

Instead, I got the error below.

**Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**
No, I downgraded bundler to 2.1.4 to get it to work

## Backtrace

```
RuntimeError: Specs already loaded
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/definition.rb:160:in `resolve_remotely!'
usr/local/bundle/gems/bundler-2.2.1/lib/bundler/installer.rb:306:in `resolve_if_needed'
 /usr/local/bundle/gems/bundler-2.2.1/lib/bundler/installer.rb:83:in `block in run'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/process_lock.rb:12:in `block in lock'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/process_lock.rb:9:in `open'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/process_lock.rb:9:in `lock'
 /usr/local/bundle/gems/bundler-2.2.1/lib/bundler/installer.rb:72:in `run'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/installer.rb:24:in `install'
 /usr/local/bundle/gems/bundler-2.2.1/lib/bundler/cli/install.rb:64:in `run'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/cli/cache.rb:33:in `install'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/cli/cache.rb:17:in `run'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/cli.rb:482:in `cache'
/usr/local/bundle/gems/bundler-2.2.1/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
[...]

PR #4175, opened 2020-12-18
Fix crash when `bundle exec`'ing to bundler

deivid-rodriguez:

The problem was that `bundle exec`'ing to `bundler` itself would crash if the lockfile was generated with bundler 2.1.4 and it's missing the specific platform.

I can't think of any reason to `bundle exec` to `bundler`, but it shouldn't be a way to make bundler crash.

In this case, the crash was due to `bundle exec`'ing on a definition missing the current specific platform (due to lockfile generated with older bundler). The definition would resolve gems in the main process, and then would tried to be re-resolved again inside the "bundle exec" context because of the missing platform. This double resolution would cause bundler to crash.

I don't see why we should raise an exception in this situation, specially such a cryptic one, so don't raise it.

Fixes https://github.com/rubygems/rubygems/issues/4153.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -162,7 +162,7 @@ module Bundler
     end
 
     def resolve_remotely!
-      raise "Specs already loaded" if @specs
+      return if @specs
       @remote = true
       sources.remote!
       specs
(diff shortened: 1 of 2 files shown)


simi (approved):

deivid-rodriguez:
Thanks for having a look @simi!

PR #4177, opened 2020-12-21
Fix misspellings in default API key name

hsbt:
Forward-port from https://github.com/ruby/ruby/commit/63849a1cd98c7ae3a7d6af3f69814f01dfae25ea


deivid-rodriguez:
Thank you!

Issue #4176, opened 2020-12-19
Deployment to AWS Lambda - clarifications and recommendations


brandonhilkert:
First off, thank you for `bundler` and thank you for aiding the terrific ecosystem it helps make a Ruby a joy to work with.

We have a handful of functions deployed to AWS Lambda. We deploy to them using the serverless framework (not super important, but it's the thing doing the zipping of all the packages and uploading to S3). Because Lambda doesn't have a "build" step, everything has to be ready to go when the function runs. 

We use CircleCI for CI/CD and our deployment step does the following:

```
bundle install --path vendor/bundle
```

In our function, we then require dependencies at the top and everything works great:

```ruby
require 'json'
require 'base64'
require 'singleton'

require 'aws-sdk-dynamodb'
```

One of our functions now depends on a private gem in our Github org named `muzzle` that's referenced via the following line in `Gemfile:
```
gem "muzzle", git: "git@github.com:[GH ORG]/muzzle.git"
```

The CI user can bundle it, but when I look at the `vendor/bundle`, I see the following tree (I've removed the unexciting bits):

```
vendor/
└── bundle
    └── ruby
        └── 2.7.0
            ├── bin
            │   ├── dotenv
            │   ├── httparty
            │   ├── rackup
            │   ├── rake
            │   ├── sidekiq
            │   └── sidekiqmon
            ├── build_info
            ├── bundler
            │   └── gems
            │       └── muzzle-325f470ea758
            │           ├── [gem contents]
            ├── cache
            │   ├── activesupport-6.0.3.4.gem
            │   ├── addressable-2.7.0.gem
            │   ├── aws-eventstream-1.0.3.gem
            │   ├── aws-partitions-1.278.0.gem
            │   ├── aws-sdk-core-3.90.1.gem
            │   ├── aws-sdk-dynamodb-1.43.0.gem
            │   ├── aws-sdk-firehose-1.25.0.gem
            │   ├── aws-sdk-kinesis-1.20.0.gem
            │   ├── aws-sdk-kms-1.30.0.gem
            │   ├── aws-sdk-s3-1.61.1.gem
            │   ├── aws-sdk-sqs-1.26.0.gem
            │   ├── aws-sigv4-1.1.1.gem
            │   ├── bundler
            │   │   └── git
            │   │       └── muzzle-428facdd2dddb1bb68f1b5e36e09232e3d124f96
            │   │           ├── HEAD
            │   │           ├── config
            │   │           ├── description
[...]

Issue #4155, opened 2020-12-15
2.2.x forces writable lockfile when running app


btalbot:
### Describe the problem as clearly as you can

`bundle check` or `bundle exec` should not attempt to update an existing Gemfile.lock and die with an error when that fails.

### Post steps to reproduce the problem

Start with Gemfile.lock created using pre-2.2 bundler; in this case 2.1.4

```
cat Gemfile
source "https://rubygems.org"

cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  ruby

DEPENDENCIES

BUNDLED WITH
   2.1.4
```

### Which command did you run?

Ensure lockfile is not-writable which is expected when docker app is run (as non file owner) without permissions to alter its own app files.

```
chmod -w Gemfile.lock

bundle check
Resolving dependencies...
There was an error while trying to write to `/tmp/bundler-2.2.0-tester/Gemfile.lock`. It is likely that you need to grant write permissions for that path.

bundle exec ruby -e 'puts RUBY_VERSION'
There was an error while trying to write to `/tmp/bundler-2.2.0-tester/Gemfile.lock`. It is likely that you need to grant write permissions for that path.

```

### What were you expecting to happen?

Bundler 2.2 to be usable to run app assembled using bundler 2.1 without altering source files.

### What actually happened?

Bundler 2.2 exists with error that it cannot update the lockfile.

### If not included with the output of your command, run `bundle env` and paste the output below

this issue is demonstrated using macos, but occurs on the linux distro we're using for docker containers as well.

## Environment

```
Bundler       2.2.1
  Platforms   ruby, x86_64-darwin-19
Ruby          2.6.6p146 (2020-03-31 revision 67876) [x86_64-darwin19]
  Full Path   /Users/btalbot/.rbenv/versions/2.6.6/bin/ruby
  Config Dir  /Users/btalbot/.rbenv/versions/2.6.6/etc
RubyGems      3.0.3
  Gem Home    /Users/btalbot/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0
  Gem Path    /Users/btalbot/.gem/ruby/2.6.0:/Users/btalbot/.rbenv/versions/2.6.6/lib/ruby/gems/2.6.0
  User Home   /Users/btalbot
  User Path   /Users/btalbot/.gem/ruby/2.6.0
  Bin Dir     /Users/btalbot/.rbenv/versions/2.6.6/bin
Tools
[...]

PR #4179, opened 2020-12-21
Restore full compatibility with previous lockfiles

deivid-rodriguez:

What I thought was a harmless change in the lockfile format (adding the specific platform to the list of platforms in the lockfile) has caused several issues for users, just because of the fact of generating changes in the lockfile during `bundle exec`, even if the bundle can be satisfied. See #4176 or #4155 for examples.

My fix is to restore the previous behaviour if dealing with an old lockfile. An old lockfile is a lockfile where the most specific ruby platform is RUBY itself, since previous lockfiles wouldn't lock the specific platform.

The rationale for doing this is that the bug locking the specific platform was meant to fix was a very rare edge case where different variants of the same gem would have either different dependencies or different ruby/rubygems version requirements. However, if people have a working lockfile, we can assume they are not running into this issue, so the most user friendly approach is to keep the same behaviour for this people, and only apply the bug fix to freshly generated lockfiles, and to people explicitly requesting specific platforms on existing lockfiles through `bundle lock --add-platform`.

Fixes https://github.com/rubygems/rubygems/issues/4176 and all regressions due to bundler 2.2.x creating lockfile changes over 2.1.x Gemfiles.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -212,13 +212,10 @@ module Bundler
         end
     end
 
-    def locked_bundler_version
-      return nil unless defined?(@definition) && @definition
+    def most_specific_locked_platform?(platform)
+      return false unless defined?(@definition) && @definition
 
-      locked_gems = definition.locked_gems
-      return nil unless locked_gems
-
-      locked_gems.bundler_version
+      definition.most_specific_locked_platform == platform
     end
 
     def ruby_scope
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
In this PR I'm actually fixing two issues, but they are quite related. Should I split it? I'd like to get this out tomorrow, so I won't to rush a bit :sweat_smile:.

deivid-rodriguez:
I splitted the other fix to a separate PR and merged it, and this one should be ready as well now.

PR #4182, opened 2020-12-22
Move aws-sdk-s3 gem to Gemfile

deivid-rodriguez:

While releasing I got an error because I must've cleaned up this gem from my system and I had to rerun the remaining release tasks manually.

I think we can make this kind of issue less likely to happen by moving the gem together with all the other development gems. Since a working bundle is needed for most release tasks, it will make this error much less likely to happen.


hsbt:
👍

Issue #4187, opened 2020-12-23
No such file or directory error while installing dependencies in Docker


glebsonik:
I tried to install gems with Bundler in Docker container using a bash script inside of CircleCI configuration file. Usually it works perfectly fine, but since yesterday in about 5% of containers installation fails (average number of concurrent containers is 60). I tried to use previous release (2.1.4), and it worked just as expected.
## Backtrace

```
Errno::ENOENT: No such file or directory @ realpath_rec - /usr/local/bundle/gems/debase-ruby_core_source-0.10.11/lib
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:485:in `realpath'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:485:in `block in mkdir_p_safe'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:484:in `each'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:484:in `reduce'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:484:in `mkdir_p_safe'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:406:in `block (2 levels) in extract_tar_gz'
  /usr/local/lib/ruby/2.6.0/rubygems/package/tar_reader.rb:65:in `each'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:389:in `block in extract_tar_gz'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:522:in `block in open_tar_gz'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:519:in `wrap'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:519:in `open_tar_gz'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:388:in `extract_tar_gz'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:368:in `block (2 levels) in extract_files'
  /usr/local/lib/ruby/2.6.0/rubygems/package/tar_reader.rb:65:in `each'
  /usr/local/lib/ruby/2.6.0/rubygems/package.rb:365:in `block in extract_files'
  /usr/local/lib/ruby/2.6.0/rubygems/package/file_source.rb:30:in `open'
  /usr/local/lib/ruby/2.6.0/rubygems/package/file_source.rb:30:in `with_read_io'
[...]

PR #4188, opened 2020-12-23
Update man pages to reflect to new default for bundle install jobs

deivid-rodriguez:

We updated the default in bundler 2.2, but didn't update docs.

Update docs.


deivid-rodriguez:
If the problem in https://github.com/rubygems/rubygems/issues/4187 is confirmed, we might want to switch back to the old default of 1 when running old rubygems.

PR #4190, opened 2020-12-23
Never spawn subshells when building extensions

deivid-rodriguez:

In https://github.com/rubygems/rubygems/pull/4189 I was expecting CI to pass, however it failed because apparently on Windows, `Open3.capture2e` does not seem to start a new shell even when given a string argument, and that made my changes break some tests.

I think not spawning a new shell should be our default behaviour.

My fix is to change our `Open3.capture2e` calls to always receive an array, so that it always try to run the command directly instead of starting a new shell.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -29,7 +29,7 @@ class Gem::Ext::Builder
       make_program = (/mswin/ =~ RUBY_PLATFORM) ? 'nmake' : 'make'
     end
 
-    destdir = '"DESTDIR=%s"' % ENV['DESTDIR']
+    destdir = 'DESTDIR=%s' % ENV['DESTDIR']
 
     ['clean', '', 'install'].each do |target|
       # Pass DESTDIR via command line to override what's in MAKEFLAGS
@@ -37,7 +37,7 @@ class Gem::Ext::Builder
         make_program,
         destdir,
         target,
-      ].join(' ').rstrip
+      ].reject(&:empty?)
       begin
         run(cmd, results, "make #{target}".rstrip, make_dir)
       rescue Gem::InstallError
@@ -56,7 +56,7 @@ class Gem::Ext::Builder
         p(command)
       end
       results << "current directory: #{dir}"
-      results << (command.respond_to?(:shelljoin) ? command.shelljoin : command)
+      results << command.shelljoin
 
       require "open3"
       # Set $SOURCE_DATE_EPOCH for the subprocess.
--- a/lib/rubygems/ext/cmake_builder.rb
+++ b/lib/rubygems/ext/cmake_builder.rb
@@ -4,8 +4,7 @@ require_relative '../command'
 class Gem::Ext::CmakeBuilder < Gem::Ext::Builder
   def self.build(extension, dest_path, results, args=[], lib_dir=nil, cmake_dir=Dir.pwd)
     unless File.exist?(File.join(cmake_dir, 'Makefile'))
-      cmd = "cmake . -DCMAKE_INSTALL_PREFIX=#{dest_path}"
-      cmd << " #{Gem::Command.build_args.join ' '}" unless Gem::Command.build_args.empty?
+      cmd = ["cmake", ".", "-DCMAKE_INSTALL_PREFIX=#{dest_path}", *Gem::Command.build_args]
 
       run cmd, results, class_name, cmake_dir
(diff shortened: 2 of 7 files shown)


lib/rubygems/ext/rake_builder.rb
```diff
@@ -25,7 +25,7 @@ def self.build(extension, dest_path, results, args=[], lib_dir=nil, extension_di
       end
     end
 
-    rake_args = ["RUBYARCHDIR=#{dest_path}", "RUBYLIBDIR=#{dest_path}", *args]
+    rake_args = ["RUBYARCHDIR=#{dest_path}".shellescape, "RUBYLIBDIR=#{dest_path}".shellescape, *args]
```
simi, line 28: Are `args` safe enough to skip `shellescape` here?

deivid-rodriguez: I'm not sure but note that previously the `run` method would `.join(" ")` its arguments (not `.shelljoin`) and pass them to `Open3.capture2e`. So if anything is broken after these changes, it was already broken before. The only reason I added `shellescape` here was to keep some tests that asserted for `RUBYARCHDIR\\=...` in error messages passing.

deivid-rodriguez: Oooops. Since we don't go through the shell when spawning subcommands, it turns out we MUST NOT shellescape arguments! I'll fix this.

test/rubygems/test_gem_ext_rake_builder.rb
```diff
@@ -24,7 +24,7 @@ def test_class_build
 
       refute_match %r{^rake failed:}, output
       assert_match %r{^#{Regexp.escape Gem.ruby} mkrf_conf\.rb}, output
-      assert_match %r{^#{Regexp.escape rake} RUBYARCHDIR\\=#{Regexp.escape @dest_path} RUBYLIBDIR\\=#{Regexp.escape @dest_path}}, output
+      assert_match %r{^#{Regexp.escape rake} RUBYARCHDIR=#{Regexp.escape @dest_path} RUBYLIBDIR=#{Regexp.escape @dest_path}}, output
```
deivid-rodriguez, line 27: This can be actually considered a bug fix.

The previous shellscaping of arguments was meant for arguments to "survive" the extra shell and arrive with the correct value to the final program (`rake` in this case). But for error messages, in addition to making the command less clear, copy-pasting it into your shell won't work.

For example, in this case, `rake`'s behavior when given `=` separated args is to set environment variables. See https://github.com/ruby/rake/blob/c2eeae2fe2b67170472a1441ebf84d3a238c3361/lib/rake/application.rb#L751-L759.

Using a "shellescaped" command breaks that:

```
$ echo "puts ENV['RUBYLIBDIR']" > Rakefile
$ rake RUBYLIBDIR=. -T
.
$ rake RUBYLIBDIR\\=. -T
```

lib/rubygems/ext/rake_builder.rb
deivid-rodriguez: I removed the shellescaping here and that unbroke things.

test/rubygems/test_gem_ext_rake_builder.rb
deivid-rodriguez: I decided to be a good boy and splitted what I think is a bug fix to a separate PR: https://github.com/rubygems/rubygems/pull/4194.

nobu: Quoting whitespaces may be needed instead?

deivid-rodriguez: As far as I understand, quoting wouldn't be needed since there's no longer a shell involved? In any case, we're not currently quoting anything, so I think it'd be part of a separate bug fix if needed?

nobu: > But for error messages, in addition to making the command less clear, copy-pasting it into your shell won't work.

When copy-pasting, spaces in a "word" need to be quoted or escape, no?

deivid-rodriguez: Oh, you're totally correct @nobu. I was thinking of the command that gets run, not the command that we print. So... I'll have yet another look.

deivid-rodriguez: It should be fixed now!

PR #4194, opened 2020-12-23
Fix commands printed on rake extension build failures

deivid-rodriguez:

In the case of a rake extension build failure, the commands that are printed as the ones that failed are incorrect.

The commands we print with the output on failures shouldn't be shellscaped, because if the `run` method is passed an array the command is invoked directly without going through an extra shell.

My fix is to not shellescape them.

Fixes the issue detected at https://github.com/rubygems/rubygems/pull/4190/files#r548235855.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -56,7 +56,7 @@ class Gem::Ext::Builder
         p(command)
       end
       results << "current directory: #{dir}"
-      results << (command.respond_to?(:shelljoin) ? command.shelljoin : command)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
As pointed out by @nobu at https://github.com/rubygems/rubygems/pull/4190#discussion_r548345537, this is incorrect. I'm confusing how we represent the string in the code with how the string gets printed. The string "rake RUBYLIBDIR\\=foo" gets printed as "rake RUBYLIBDIR\=" (which works fine when copy-pasted in the shell), because we're making it explicit that the slash is a real slash part of the string, not the beginning of a escape sequence.

PR #4196, opened 2020-12-24
Fix Windows + JRuby no longer being able to install git sources

### LINKED ISSUE #4195: [Regression] bundle 2.2.x fails to install dependencies from Git on JRuby + Windows

If user tries to `bundle install` on a Gemfile that has Git-repository dependency, bundler fails with `The source does not contain any versions of '<dependency>'`:

### Steps to reproduce the problem

Assuming that you use Windows, have JRuby as default Ruby and have Bundler 2.2.x (2.2.3 still has this bug):

1. Create an empty project with the following `Gemfile`:

```
source "https://rubygems.org"

gem 'prawn-table', git: 'https://github.com/prawnpdf/prawn-table', ref: '515f2db294866a343b05d15f94e5fb417a32f6ff'
```

2. Run `bundle install`

### Expected

Bundler happily installs dependencies

### Actual

```
Fetching https://github.com/prawnpdf/prawn-table
Could not find gem 'prawn-table' in https://github.com/prawnpdf/prawn-table (at
Fetching gem metadata from https://rubygems.org/
515f2db@515f2db).
The source does not contain any versions of 'prawn-table'
Error: Process completed with exit code 1.
```

Bug is **NOT** present in Bundler 2.1.4.

### Additional notes

For your convenience, I've created an minimal testcase repository: https://github.com/slonopotamus/bundler-git-dep

Also, here's GitHub Actions workflow run that reproduces described scenario: https://github.com/slonopotamus/bundler-git-dep/runs/1605766634?check_suite_focus=true

You can clearly see that only Windows+JRuby is affected, other platforms do not exhibit this bug.

deivid-rodriguez:

The problem is that bundler 2.2 can no longer install git sources under Windows + JRuby.

First I considered it could be a Windows escaping issue, so I made some changes so that the `git` commands `bundler` runs under the hood no longer use a shell but call `git` directly. This is similar to #4190, but for `bundler`. I'm unsure whether that fixed anything regarding this issue, but I'd say no because Ruby on Windows CI was previously passing, so it must be something specific to jruby. Anyways, this change seems like a good move anyways, so I left it there.

After that, I remembered being hit by some issues with jruby not properly handling the `:chdir` parameter to `Process.spawn` and friends. See https://github.com/jruby/jruby/issues/6153.

So to workaround this issue I tried using the `-C` CLI flag to `git` instead of the `:chdir` parameter to `Process.spawn`, and that worked just fine!

I also added a bare integration test to specifically check this simple case on JRuby + Windows.

Fixes #4195.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -71,14 +71,20 @@ jobs:
     timeout-minutes: 10
 
   install_rubygems_windows:
-    name: Install Rubygems on Windows
+    name: Install Rubygems on Windows (${{ matrix.ruby.name }})
     runs-on: windows-2019
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby:
+          - { name: 2.7, value: 2.7.2 }
+          - { name: jruby-9.2, value: jruby-9.2.14.0 }
     steps:
       - uses: actions/checkout@v2
       - name: Setup ruby
         uses: ruby/setup-ruby@v1
         with:
-          ruby-version: 2.7.2
+          ruby-version: ${{ matrix.ruby.value }}
           bundler: none
       - name: Install rubygems
         run: ruby setup.rb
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
The [last commit in this branch](c2bd033513d34962932c8aa9a426725cbbb4cfd1) seems to fix #4195!

Issue #3799, opened 2020-07-08
Could boot performance be improved? Caching default gems perhaps?


casperisfine:
Sorry if this was discussed elsewhere, I tried to search for previous discussions but given how generic the subject is I couldn't find anything.

### Context

As many people know, Rubygems initialization is the biggest part of MRI's boot:

```
$ time ruby -e '1'

real	0m0.076s
user	0m0.051s
sys	0m0.021s
$ time ruby --disable-gems -e '1'

real	0m0.029s
user	0m0.011s
sys	0m0.015s
```

So (on my machine) "raw" MRI initialization is `29ms`, and rubygems adds `47ms` to it.

When profiling, we can see that a large part of that comes from loading default gems:

```
==================================
  Mode: cpu(10)
  Samples: 545 (52.98% miss rate)
  GC: 25 (4.59%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       237  (43.5%)         190  (34.9%)     Gem::Specification.load
       419  (76.9%)          40   (7.3%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/rubygems.rb>
        66  (12.1%)          34   (6.2%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/did_you_mean.rb>
        70  (12.8%)          33   (6.1%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/rubygems/specification.rb>
        30   (5.5%)          30   (5.5%)     Gem::BasicSpecification#internal_init
        22   (4.0%)          22   (4.0%)     (sweeping)
        93  (17.1%)          21   (3.9%)     <module:Gem>
       106  (19.4%)          16   (2.9%)     Kernel#require
        15   (2.8%)          14   (2.6%)     Gem::Specification#files
        10   (1.8%)          10   (1.8%)     Gem::PathSupport#expand
        11   (2.0%)           9   (1.7%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/did_you_mean/spell_checker.rb>
        10   (1.8%)           9   (1.7%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/rubygems/user_interaction.rb>
        11   (2.0%)           9   (1.7%)     <require:/opt/rubies/2.8.0-dev/lib/ruby/2.8.0/did_you_mean/spell_checkers/name_error_checkers.rb>
[...]

PR #4199, opened 2020-12-26
Speed up boot time by adding default spec files to stub lines

deivid-rodriguez:

Rubygems boot time is slow. One of the causes if that all default gems are evaluated, in order to figure out which files they include and be able to "upgrade requires" in case a higher version is installed. So for example, when `require "psych"` is run, rubygems checks whether there's any file named "psych.rb` inside a default gem, and in that case activates the highest version of that default gem in the system.

This works, but evaluating all default specs in order to do this is expensive.

My fix is, similarly to what we do with other information inside gemspecs to avoid having to actually re-evaluate once they are insalled, is to create a new stub line including the list of files inside the gem. So, before this patch, the `psych` gemspec installed would like like this:

```ruby
# -*- encoding: utf-8 -*-
# stub: psych 3.3.0 ruby lib
# stub: ext/psych/extconf.rb

Gem::Specification.new do |s|
  s.name = "psych".freeze
  s.version = "3.3.0"

  s.required_rubygems_version = Gem::Requirement.new(">= 0".freeze) if s.respond_to? :required_rubygems_version=
  s.require_paths = ["lib".freeze]
  s.authors = ["Aaron Patterson".freeze, "SHIBATA Hiroshi".freeze, "Charles Oliver Nutter".freeze]
  s.date = "2021-07-08"
  s.description = "Psych is a YAML parser and emitter. Psych leverages libyaml[https://pyyaml.org/wiki/LibYAML]\nfor its YAML parsing and emitting capabilities. In addition to wrapping libyaml,\nPsych also knows how to serialize and de-serialize most Ruby objects to and from the YAML format.\n".freeze
  s.email = ["aaron@tenderlovemaking.com".freeze, "hsbt@ruby-lang.org".freeze, "headius@headius.com".freeze]
  s.extensions = ["ext/psych/extconf.rb".freeze]
  s.extra_rdoc_files = ["README.md".freeze]
  s.files = ["README.md".freeze, "ext/psych/extconf.rb".freeze, "psych.rb".freeze, "psych.so".freeze, "psych/class_loader.rb".freeze, "psych/coder.rb".freeze, "psych/core_ext.rb".freeze, "psych/exception.rb".freeze, "psych/handler.rb".freeze, "psych/handlers/document_stream.rb".freeze, "psych/handlers/recorder.rb".freeze, "psych/json/ruby_events.rb".freeze, "psych/json/stream.rb".freeze, "psych/json/tree_builder.rb".freeze, "psych/json/yaml_events.rb".freeze, "psych/nodes.rb".freeze, "psych/nodes/alias.rb".freeze, "psych/nodes/document.rb".freeze, "psych/nodes/mapping.rb".freeze, "psych/nodes/node.rb".freeze, "psych/nodes/scalar.rb".freeze, "psych/nodes/sequence.rb".freeze, "psych/nodes/stream.rb".freeze, "psych/omap.rb".freeze, "psych/parser.rb".freeze, "psych/scalar_scanner.rb".freeze, "psych/set.rb".freeze, "psych/stream.rb".freeze, "psych/streaming.rb".freeze, "psych/syntax_error.rb".freeze, "psych/tree_builder.rb".freeze, "psych/versions.rb".freeze, "psych/visitors.rb".freeze, "psych/visitors/depth_first.rb".freeze, "psych/visitors/emitter.rb".freeze, "psych/visitors/json_tree.rb".freeze, "psych/visitors/to_ruby.rb".freeze, "psych/visitors/visitor.rb".freeze, "psych/visitors/yaml_tree.rb".freeze, "psych/y.rb".freeze]
  s.homepage = "https://github.com/ruby/psych".freeze
  s.licenses = ["MIT".freeze]
  s.rdoc_options = ["--main".freeze, "README.md".freeze]
  s.required_ruby_version = Gem::Requirement.new(">= 2.4.0".freeze)
  s.rubygems_version = "3.3.0.dev".freeze
  s.summary = "Psych is a YAML parser and emitter".freeze
end
```

but now it looks like this (note the 4th comment on top of the file):

```ruby
# -*- encoding: utf-8 -*-
# stub: psych 3.3.0 ruby lib
# stub: ext/psych/extconf.rb
# files-stub: README.mdext/psych/extconf.rbpsych.rbpsych.sopsych/class_loader.rbpsych/coder.rbpsych/core_ext.rbpsych/exception.rbpsych/handler.rbpsych/handlers/document_stream.rbpsych/handlers/recorder.rbpsych/json/ruby_events.rbpsych/json/stream.rbpsych/json/tree_builder.rbpsych/json/yaml_events.rbpsych/nodes.rbpsych/nodes/alias.rbpsych/nodes/document.rbpsych/nodes/mapping.rbpsych/nodes/node.rbpsych/nodes/scalar.rbpsych/nodes/sequence.rbpsych/nodes/stream.rbpsych/omap.rbpsych/parser.rbpsych/scalar_scanner.rbpsych/set.rbpsych/stream.rbpsych/streaming.rbpsych/syntax_error.rbpsych/tree_builder.rbpsych/versions.rbpsych/visitors.rbpsych/visitors/depth_first.rbpsych/visitors/emitter.rbpsych/visitors/json_tree.rbpsych/visitors/to_ruby.rbpsych/visitors/visitor.rbpsych/visitors/yaml_tree.rbpsych/y.rb

Gem::Specification.new do |s|
  s.name = "psych".freeze
  s.version = "3.3.0"

  s.required_rubygems_version = Gem::Requirement.new(">= 0".freeze) if s.respond_to? :required_rubygems_version=
  s.require_paths = ["lib".freeze]
  s.authors = ["Aaron Patterson".freeze, "SHIBATA Hiroshi".freeze, "Charles Oliver Nutter".freeze]
  s.date = "2021-07-08"
  s.description = "Psych is a YAML parser and emitter. Psych leverages libyaml[https://pyyaml.org/wiki/LibYAML]\nfor its YAML parsing and emitting capabilities. In addition to wrapping libyaml,\nPsych also knows how to serialize and de-serialize most Ruby objects to and from the YAML format.\n".freeze
  s.email = ["aaron@tenderlovemaking.com".freeze, "hsbt@ruby-lang.org".freeze, "headius@headius.com".freeze]
  s.extensions = ["ext/psych/extconf.rb".freeze]
  s.extra_rdoc_files = ["README.md".freeze]
  s.files = ["README.md".freeze, "ext/psych/extconf.rb".freeze, "psych.rb".freeze, "psych.so".freeze, "psych/class_loader.rb".freeze, "psych/coder.rb".freeze, "psych/core_ext.rb".freeze, "psych/exception.rb".freeze, "psych/handler.rb".freeze, "psych/handlers/document_stream.rb".freeze, "psych/handlers/recorder.rb".freeze, "psych/json/ruby_events.rb".freeze, "psych/json/stream.rb".freeze, "psych/json/tree_builder.rb".freeze, "psych/json/yaml_events.rb".freeze, "psych/nodes.rb".freeze, "psych/nodes/alias.rb".freeze, "psych/nodes/document.rb".freeze, "psych/nodes/mapping.rb".freeze, "psych/nodes/node.rb".freeze, "psych/nodes/scalar.rb".freeze, "psych/nodes/sequence.rb".freeze, "psych/nodes/stream.rb".freeze, "psych/omap.rb".freeze, "psych/parser.rb".freeze, "psych/scalar_scanner.rb".freeze, "psych/set.rb".freeze, "psych/stream.rb".freeze, "psych/streaming.rb".freeze, "psych/syntax_error.rb".freeze, "psych/tree_builder.rb".freeze, "psych/versions.rb".freeze, "psych/visitors.rb".freeze, "psych/visitors/depth_first.rb".freeze, "psych/visitors/emitter.rb".freeze, "psych/visitors/json_tree.rb".freeze, "psych/visitors/to_ruby.rb".freeze, "psych/visitors/visitor.rb".freeze, "psych/visitors/yaml_tree.rb".freeze, "psych/y.rb".freeze]
  s.homepage = "https://github.com/ruby/psych".freeze
  s.licenses = ["MIT".freeze]
  s.rdoc_options = ["--main".freeze, "README.md".freeze]
  s.required_ruby_version = Gem::Requirement.new(">= 2.4.0".freeze)
  s.rubygems_version = "3.3.0.dev".freeze
  s.summary = "Psych is a YAML parser and emitter".freeze
end
```
That way we can use the existing mechanism to avoid evaulating gemspecs if stub lines already include the information we need.

When rubygems is updated, all default gemspecs will be regenerated to include this line. When rubygems is downgraded, gemspecs are left there without that line, but I don't think that will cause any issues since the comment will be ignored by older rubygems.

The results I'm getting with this optimization are pretty unstable, but they show a consistent speedup. Compared to current master branch, rubygems boots ~20% faster, and the overhead of rubygems over ruby boot time goes from ~75% slower to ~50% slower.

```
$ hyperfine --prepare 'git checkout speedup && rake install' --command-name optimization 'ruby -e1' --prepare 'git checkout master && rake install' --command-name current 'ruby -e1'
Benchmark #1: optimization
  Time (mean ± σ):      71.8 ms ±   5.7 ms    [User: 59.0 ms, System: 15.0 ms]
  Range (min … max):    66.6 ms …  85.3 ms    10 runs
 
Benchmark #2: current
  Time (mean ± σ):      84.4 ms ±   4.2 ms    [User: 73.5 ms, System: 13.2 ms]
  Range (min … max):    79.2 ms …  90.7 ms    10 runs
 
Summary
  'optimization' ran
    1.18 ± 0.11 times faster than 'current'
```

```
$ hyperfine --prepare 'git checkout speedup && rake install' --command-name optimization 'ruby -e1' --prepare 'git checkout master && rake install' --command-name current 'ruby -e1' --prepare ''  --command-name without-rubygems 'ruby --disable-gems -e1'
Benchmark #1: optimization
  Time (mean ± σ):      67.8 ms ±   1.1 ms    [User: 59.9 ms, System: 10.0 ms]
  Range (min … max):    66.9 ms …  70.5 ms    10 runs
 
Benchmark #2: current
  Time (mean ± σ):      80.5 ms ±   0.7 ms    [User: 71.8 ms, System: 10.8 ms]
  Range (min … max):    79.7 ms …  81.7 ms    10 runs
 
Benchmark #3: without-rubygems
  Time (mean ± σ):      46.6 ms ±  10.4 ms    [User: 32.1 ms, System: 17.1 ms]
  Range (min … max):    33.0 ms …  78.1 ms    77 runs
 
Summary
  'without-rubygems' ran
    1.46 ± 0.33 times faster than 'optimization'
    1.73 ± 0.39 times faster than 'current'
```

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -182,7 +182,9 @@ command to remove old versions.
       say "Installing RubyGems #{version}" unless options[:silent]
 
       installed = preparing_gem_layout_for(version) do
-        system Gem.ruby, "--disable-gems", "setup.rb", *args
+        preparing_default_gemspecs_for(version) do
+          system Gem.ruby, "--disable-gems", "setup.rb", *args
+        end
       end
 
       say "RubyGems system software updated" if installed unless options[:silent]
@@ -206,6 +208,16 @@ command to remove old versions.
       end
 
       status
+    end
+  end
+
+  def preparing_default_gemspecs_for(version)
+    if Gem::Version.new(version) >= Gem::Version.new("3.3.a")
+      yield
+    else
+      Gem.regenerate_default_specs(include_stub_files: false)
+
+      yield
     end
   end
 
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -182,9 +182,7 @@ command to remove old versions.
       say "Installing RubyGems #{version}" unless options[:silent]
 
       installed = preparing_gem_layout_for(version) do
-        preparing_default_gemspecs_for(version) do
-          system Gem.ruby, "--disable-gems", "setup.rb", *args
-        end
+        system Gem.ruby, "--disable-gems", "setup.rb", *args
       end
 
       say "RubyGems system software updated" if installed unless options[:silent]
@@ -208,16 +206,6 @@ command to remove old versions.
       end
 
       status
-    end
-  end
-
-  def preparing_default_gemspecs_for(version)
-    if Gem::Version.new(version) >= Gem::Version.new("3.3.a")
-      yield
-    else
-      Gem.regenerate_default_specs(include_stub_files: false)
-
-      yield
     end
   end
 
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -88,6 +88,14 @@ jobs:
         working-directory: ./bundler
       - name: Check rails can be installed
         run: gem install rails --verbose --backtrace
+      - name: Check default gem gets activated
+        run: ruby -rpsych -e 'exit(Gem.loaded_specs["psych"].version.to_s != "4.0.4")'
+      - name: Check upgraded default gem gets activated
+        run: |
+          gem install psych:4.0.4
+          ruby -rpsych -e 'puts "Checking psych 4.0.4 gets activated...."; exit(Gem.loaded_specs["psych"].version.to_s == "4.0.4")'
+          ruby -rpsych -e 'puts "Checking only psych 4.0.4 features are loaded between #{$LOADED_FEATURES}"; exit($LOADED_FEATURES.grep(/psych/).all?{|lf| lf =~ /psych-4.0.4/})'
+        if: matrix.ruby.name != '2.3'
     timeout-minutes: 10
 
   install_rubygems_windows:
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1203,10 +1203,18 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
           next unless $~
         end
 
-        spec.activate if already_loaded?(file)
+        spec.to_spec.activate if already_loaded?(file)
 
         @path_to_default_spec_map[file] = spec
         @path_to_default_spec_map[file.sub(suffix_regexp, "")] = spec
+      end
+    end
+
+    def regenerate_default_specs(include_stub_files:)
+      Gem::Util.glob_files_in_dir("*.gemspec", Gem.default_specifications_dir).map do |path|
+        spec = Gem::Specification.load path
+
+        write_binary(spec.loaded_from, spec.to_ruby(include_stub_files: include_stub_files))
       end
     end
 
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1210,11 +1210,11 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
       end
     end
 
-    def regenerate_default_specs(include_stub_files:)
+    def regenerate_default_specs
       Gem::Util.glob_files_in_dir("*.gemspec", Gem.default_specifications_dir).map do |path|
         spec = Gem::Specification.load path
 
-        write_binary(spec.loaded_from, spec.to_ruby(include_stub_files: include_stub_files))
+        write_binary(spec.loaded_from, spec.to_ruby)
       end
     end
 
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -615,7 +615,7 @@ abort "#{deprecation_message}"
   end
 
   def regenerate_default_specs
-    Gem.regenerate_default_specs(include_stub_files: true)
+    Gem.regenerate_default_specs
   end
 
   private
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -18,7 +18,8 @@ class Gem::Commands::SetupCommand < Gem::Command
           :site_or_vendor => "sitelibdir",
           :destdir => "", :prefix => "", :previous_version => "",
           :regenerate_binstubs => true,
-          :regenerate_plugins => true
+          :regenerate_plugins => true,
+          :regenerate_default_specs => true
 
     add_option "--previous-version=VERSION",
                "Previous version of RubyGems",
@@ -90,6 +91,11 @@ class Gem::Commands::SetupCommand < Gem::Command
     add_option "--[no-]regenerate-plugins",
                "Regenerate gem plugins" do |value, options|
       options[:regenerate_plugins] = value
+    end
+
+    add_option "--[no-]regenerate-default-specifications",
+               "Regenerate default gem specifications" do |value, options|
+      options[:regenerate_default_specifications] = value
     end
 
     add_option "-f", "--[no-]force",
@@ -182,6 +188,7 @@ By default, this RubyGems will install gem as:
 
     regenerate_binstubs(bin_dir) if options[:regenerate_binstubs]
     regenerate_plugins(bin_dir) if options[:regenerate_plugins]
+    regenerate_default_specs if options[:regenerate_default_specs]
 
     uninstall_old_gemcutter
 
@@ -605,6 +612,10 @@ abort "#{deprecation_message}"
 
     command = Gem::Commands::PristineCommand.new
     command.invoke(*args)
+  end
+
+  def regenerate_default_specs
+    Gem.regenerate_default_specs(include_stub_files: true)
   end
 
   private
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -875,7 +875,7 @@ class Gem::Specification < Gem::BasicSpecification
   # Loads the default specifications. It should be called only once.
 
   def self.load_defaults
-    each_spec([Gem.default_specifications_dir]) do |spec|
+    default_stubs.each do |spec|
       # #load returns nil if the spec is bad, so we just ignore
       # it at this stage
       Gem.register_default_spec(spec)
@@ -2462,13 +2462,14 @@ class Gem::Specification < Gem::BasicSpecification
   # be eval'ed and reconstruct the same specification later.  Attributes that
   # still have their default values are omitted.
 
-  def to_ruby
+  def to_ruby(include_stub_files: true)
     mark_version
     result = []
     result << "# -*- encoding: utf-8 -*-"
     result << "#{Gem::StubSpecification::PREFIX}#{name} #{version} #{platform} #{raw_require_paths.join("\0")}"
     result << "#{Gem::StubSpecification::PREFIX}#{extensions.join "\0"}" unless
       extensions.empty?
+    result << "#{Gem::StubSpecification::FILES_PREFIX}#{files.join("\0")}" if include_stub_files && default_gem?
     result << nil
     result << "Gem::Specification.new do |s|"
 
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2462,14 +2462,14 @@ class Gem::Specification < Gem::BasicSpecification
   # be eval'ed and reconstruct the same specification later.  Attributes that
   # still have their default values are omitted.
 
-  def to_ruby(include_stub_files: true)
+  def to_ruby
     mark_version
     result = []
     result << "# -*- encoding: utf-8 -*-"
     result << "#{Gem::StubSpecification::PREFIX}#{name} #{version} #{platform} #{raw_require_paths.join("\0")}"
     result << "#{Gem::StubSpecification::PREFIX}#{extensions.join "\0"}" unless
       extensions.empty?
-    result << "#{Gem::StubSpecification::FILES_PREFIX}#{files.join("\0")}" if include_stub_files && default_gem?
+    result << "#{Gem::StubSpecification::FILES_PREFIX}#{files.join("\0")}" if default_gem?
     result << nil
     result << "Gem::Specification.new do |s|"
 
--- a/lib/rubygems/stub_specification.rb
+++ b/lib/rubygems/stub_specification.rb
@@ -7,15 +7,17 @@
 class Gem::StubSpecification < Gem::BasicSpecification
   # :nodoc:
   PREFIX = "# stub: ".freeze
+  FILES_PREFIX = "# files-stub: ".freeze
 
   # :nodoc:
   OPEN_MODE = "r:UTF-8:-".freeze
 
   class StubLine # :nodoc: all
-    attr_reader :name, :version, :platform, :require_paths, :extensions,
+    attr_reader :name, :version, :platform, :require_paths, :files, :extensions,
                 :full_name
 
     NO_EXTENSIONS = [].freeze
+    NO_FILES = [].freeze
 
     # These are common require paths.
     REQUIRE_PATHS = { # :nodoc:
@@ -32,13 +34,14 @@ class Gem::StubSpecification < Gem::BasicSpecification
       "lib" => ["lib"].freeze,
     }.freeze
 
-    def initialize(data, extensions)
+    def initialize(data, files, extensions)
       parts          = data[PREFIX.length..-1].split(" ".freeze, 4)
       @name          = parts[0].freeze
       parts.insert(1, 0) unless Gem::Version.correct?(parts[1])
       @version       = Gem::Version.new(parts[1])
 
       @platform      = Gem::Platform.new parts[2]
+      @files         = files
       @extensions    = extensions
       @full_name     = if platform == Gem::Platform::RUBY
         "#{name}-#{version}"
@@ -112,13 +115,21 @@ class Gem::StubSpecification < Gem::BasicSpecification
             file.readline # discard encoding line
             stubline = file.readline.chomp
             if stubline.start_with?(PREFIX)
-              extensions = if /\A#{PREFIX}/ =~ file.readline.chomp
+              second_stub_line = file.readline.chomp
+
+              extensions = if /\A#{PREFIX}/ =~ second_stub_line
                 $'.split "\0"
               else
                 StubLine::NO_EXTENSIONS
               end
 
-              @data = StubLine.new stubline, extensions
+              files = if default_gem? && (/\A#{FILES_PREFIX}/ =~ second_stub_line || /\A#{FILES_PREFIX}/ =~ file.readline.chomp)
+                $'.split "\0"
+              else
+                StubLine::NO_FILES
+              end
+
+              @data = StubLine.new stubline, files, extensions
             end
           rescue EOFError
           end
@@ -171,6 +182,13 @@ class Gem::StubSpecification < Gem::BasicSpecification
 
   def version
     data.version
+  end
+
+  ##
+  # List of files in the gem
+
+  def files
+    data.files
   end
 
   def full_name
--- a/lib/rubygems/stub_specification.rb
+++ b/lib/rubygems/stub_specification.rb
@@ -35,11 +35,8 @@ class Gem::StubSpecification < Gem::BasicSpecification
     def initialize(data, extensions)
       parts          = data[PREFIX.length..-1].split(" ".freeze, 4)
       @name          = parts[0].freeze
-      @version       = if Gem::Version.correct?(parts[1])
-        Gem::Version.new(parts[1])
-      else
-        Gem::Version.new(0)
-      end
+      parts.insert(1, 0) unless Gem::Version.correct?(parts[1])
+      @version       = Gem::Version.new(parts[1])
 
       @platform      = Gem::Platform.new parts[2]
       @extensions    = extensions
--- a/lib/rubygems/stub_specification.rb
+++ b/lib/rubygems/stub_specification.rb
@@ -188,7 +188,11 @@ class Gem::StubSpecification < Gem::BasicSpecification
   # List of files in the gem
 
   def files
-    data.files
+    if default_gem?
+      data.files.any? ? data.files : to_spec.files
+    else
+      to_spec.files
+    end
   end
 
   def full_name
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -443,42 +443,6 @@ class TestGemRequire < Gem::TestCase
     end
   end
 
-  def test_realworld_default_gem
-    omit "this test can't work under ruby-core setup" if ruby_repo?
-
-    cmd = <<-RUBY
-      $stderr = $stdout
-      require "json"
-      puts Gem.loaded_specs["json"]
-    RUBY
-    output = Gem::Util.popen(*ruby_with_rubygems_in_load_path, "-e", cmd).strip
-    assert $?.success?
-    refute_empty output
-  end
-
-  def test_realworld_upgraded_default_gem
-    omit "this test can't work under ruby-core setup" if ruby_repo?
-
-    newer_json = util_spec("json", "999.99.9", nil, ["lib/json.rb"])
-    install_gem newer_json
-
-    path = "#{@tempdir}/test_realworld_upgraded_default_gem.rb"
-    code = <<-RUBY
-      $stderr = $stdout
-      require "json"
-      puts Gem.loaded_specs["json"].version
-      puts $LOADED_FEATURES
-    RUBY
-    File.write(path, code)
-
-    output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
-    assert $?.success?
-    refute_empty output
-    assert_equal "999.99.9", output.lines[0].chomp
-    # Make sure only files from the newer json gem are loaded, and no files from the default json gem
-    assert_equal ["#{@gemhome}/gems/json-999.99.9/lib/json.rb"], output.lines.grep(%r{/gems/json-}).map(&:chomp)
-  end
-
   def test_default_gem_and_normal_gem
     default_gem_spec = new_default_spec("default", "2.0.0.0",
                                         nil, "default/gem.rb")


deivid-rodriguez:
I'm working on more boot performance improvements and now I have the following performance locally:

```
Benchmark #1: ruby -e1
  Time (mean ± σ):      56.2 ms ±   1.9 ms    [User: 46.3 ms, System: 12.2 ms]
  Range (min … max):    53.9 ms …  62.4 ms    52 runs
 
Benchmark #2: ruby --disable-gems -e1
  Time (mean ± σ):      43.8 ms ±   8.5 ms    [User: 31.4 ms, System: 15.2 ms]
  Range (min … max):    34.3 ms …  55.3 ms    84 runs
 
Summary
  'ruby --disable-gems -e1' ran
    1.28 ± 0.25 times faster than 'ruby -e1'
```

Will try to get this ready soon.

deivid-rodriguez:
I updated this PR. The results I posted in the previous message were with further optimizations, but I think I'll propose this first and then the other optimizations separately.

deivid-rodriguez:
I'd like to do some more testing around backwards compatibility but other than that this should be ready!

lib/rubygems/commands/update_command.rb
```diff
@@ -209,6 +211,16 @@ def preparing_gem_layout_for(version)
     end
   end
 
+  def preparing_default_gemspecs_for(version)
+    if Gem::Version.new(version) >= Gem::Version.new("3.3.a")
```
deivid-rodriguez, line 215: Actually I can see now why the gemspecs were not being regenerated back when downgrading in my testing: master is also `>= 3.3.a`. But I think this might not be actually needed because the extra comment doesn't seem to cause any issues when running old versions.

deivid-rodriguez:
Also added some code locally to compare this approach with the one proposed at https://github.com/rubygems/rubygems/issues/3799 of caching default marshalled specs, and it seems that this approach is consistently faster:

```
$ hyperfine --warmup 1 --command-name current 'ruby -e1' --command-name stub_default_spec_files 'STUB_DEFAULT_SPEC_FILES=true ruby -e1' --command-name 'cache_default_specs' 'CACHE_DEFAULT_SPECS=true ruby -e1'
Benchmark #1: current
  Time (mean ± σ):      79.0 ms ±   1.5 ms    [User: 68.1 ms, System: 13.3 ms]
  Range (min … max):    77.5 ms …  85.1 ms    37 runs
 
Benchmark #2: stub_default_spec_files
  Time (mean ± σ):      66.3 ms ±   1.5 ms    [User: 56.5 ms, System: 12.1 ms]
  Range (min … max):    64.5 ms …  71.6 ms    43 runs
 
Benchmark #3: cache_default_specs
  Time (mean ± σ):      75.6 ms ±   1.7 ms    [User: 64.8 ms, System: 13.1 ms]
  Range (min … max):    73.2 ms …  79.2 ms    38 runs
 
Summary
  'stub_default_spec_files' ran
    1.14 ± 0.04 times faster than 'cache_default_specs'
    1.19 ± 0.03 times faster than 'current'
```

eregon:
The numbers for `ruby --disable-gems` look very high to me, maybe you're using rbenv or RVM?
For startup measurement it's best to ensure the real ruby binary is already in PATH, without any wrapper before.
Or potentially just pass a full path to the real ruby binary.

deivid-rodriguez:
I rebased this and re-measured with my current Ruby installation which does not use shims:

```
$ hyperfine -N --command-name cache-default-marshaled-specs --prepare "./tmp/prepare.sh speedup-alt" "ruby -e1"  --command-name this-pr --prepare "./tmp/prepare.sh speedup" "ruby -e1" --command-name master --prepare "./tmp/prepare.sh master" "ruby -e1" --warmup 5
Benchmark 1: cache-default-marshaled-specs
  Time (mean ± σ):      29.5 ms ±   0.1 ms    [User: 23.2 ms, System: 5.5 ms]
  Range (min … max):    29.3 ms …  29.7 ms    10 runs
 
Benchmark 2: this-pr
  Time (mean ± σ):      27.4 ms ±   0.4 ms    [User: 20.8 ms, System: 5.8 ms]
  Range (min … max):    27.1 ms …  28.4 ms    10 runs
 
Benchmark 3: master
  Time (mean ± σ):      32.1 ms ±   0.2 ms    [User: 25.3 ms, System: 5.9 ms]
  Range (min … max):    31.9 ms …  32.5 ms    10 runs
 
Summary
  'this-pr' ran
    1.08 ± 0.01 times faster than 'cache-default-marshaled-specs'
    1.17 ± 0.02 times faster than 'master'
```

eregon:
I'm a little bit worried to store this information in the `.gemspec` for two reasons:
* It can make the gemspec a lot bigger (a gem can have many files), and so if we actually need to load it fully we'll take longer to read it
* It makes the gemspec unreadable in most editors/CLI tools due to the `\0` separators, which is unfortunate when debugging/investigating (already an issue with `stub` lines).

Would it make sense to have a separate file for this information? That would also enable storing the information in a custom format, and not need to buffer a large part of the file just to find the line endings, which would likely mean additional speedups.

deivid-rodriguez:
> It can make the gemspec a lot bigger (a gem can have many files), and so if we actually need to load it fully we'll take longer to read it

Default gems don't usually have many files, and we only do this for default gems. Do you really think this can be a noticeable problem?

> It makes the gemspec unreadable in most editors/CLI tools due to the \0 separators, which is unfortunate when debugging/investigating (already an issue with stub lines).

As you said, this is an existing issue, and it seems acceptable to me. We can consider changing it, there's an open issue for that, but only a single person reported this as an issue in like 15 years.

> Would it make sense to have a separate file for this information? That would also enable storing the information in a binary format, and not need to buffer a large part of the file just to find the line endings, which would likely mean additional speedups.

I guess, but I think this is an improvement already and the downsides seem quite weak.

eregon:
> Default gems don't usually have many files, and we only do this for default gems. Do you really think this can be a noticeable problem?

Yes, otherwise I wouldn't mention it. It would basically double the size of every default gem `.gemspec` file.
(talking about file size, the `.freeze` everywhere look not nice and are a waste of file size, `# frozen_string_literal: true` seems much better)

Since `s.files =` seems by far the largest field of a gemspec, it is a bit surprising that this approach gains something. Maybe there is some non-trivial costs for other fields being assigned? Otherwise we're just measuring Ruby parser speed vs read with buffer, split by lines, split by \0, at the cost of storing the information twice.

```
> Gem::Specification.each.select(&:default_gem?).map { [_1.files.size, _1.name] }.sort
...
 [8, "optparse"],
 [11, "net-http"],
 [12, "fiddle"],
 [14, "csv"],
 [14, "digest"],
 [14, "drb"],
 [15, "uri"],
 [17, "reline"],
 [18, "did_you_mean"],
 [19, "openssl"],
 [26, "json"],
 [26, "racc"],
 [40, "psych"],
 [46, "irb"],
 [124, "rdoc"],
 [211, "bundler"]]
```

> but only a single person reported this as an issue in like 15 years.

Yet multiple people seem annoyed by it, IMHO worth fixing. I didn't report it until now but I've been annoyed by that several times (when debugging stuff related to default gems).
I'm probably more exposed to it than the average Ruby user though.

deivid-rodriguez:
> Yes, otherwise I wouldn't mention it. It would basically double the size of every default gem .gemspec file.

Mmmm. By size you mean the size of the file? That statement seems like an exaggeration if so, it only affect two gemspecs noticiably, `rdoc` and `bundler`, out of 70 in Ruby 3.1.2.

#### Before

```
$ du -hs /Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/*|sort -nr
 28K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/bundler-2.4.0.dev.gemspec
 12K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/rdoc-6.4.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/zlib-2.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/yaml-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/weakref-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/uri-0.11.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/un-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tsort-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tmpdir-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/timeout-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/time-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tempfile-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/syslog-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/strscan-3.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/stringio-3.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/singleton-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/shellwords-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/set-1.0.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/securerandom-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ruby2_keywords-0.0.5.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/rinda-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/resolv-replace-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/resolv-0.2.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/reline-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/readline-ext-0.1.4.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/readline-0.0.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/racc-1.6.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/psych-4.0.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pstore-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/prettyprint-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pp-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pathname-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ostruct-0.5.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/optparse-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/openssl-3.0.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/open3-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/open-uri-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/observer-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/nkf-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/net-protocol-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/net-http-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/mutex_m-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/logger-1.5.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/json-2.6.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/irb-1.4.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ipaddr-1.2.4.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-wait-0.2.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-nonblock-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-console-0.5.11.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/getoptlong-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/forwardable-1.3.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/find-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fileutils-1.6.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fiddle-1.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fcntl-1.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/etc-1.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/error_highlight-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/erb-2.2.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/english-0.7.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/drb-2.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/digest-3.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/did_you_mean-1.6.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/delegate-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/date-3.2.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/csv-3.2.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/cgi-0.3.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/bigdecimal-3.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/benchmark-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/base64-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/abbrev-0.1.0.gemspec
```

#### After

```
$ du -hs /Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/*|sort -nr
 28K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/bundler-2.4.0.dev.gemspec
 12K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/rdoc-6.4.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/zlib-2.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/yaml-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/weakref-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/uri-0.11.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/un-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tsort-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tmpdir-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/timeout-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/time-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/tempfile-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/syslog-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/strscan-3.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/stringio-3.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/singleton-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/shellwords-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/set-1.0.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/securerandom-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ruby2_keywords-0.0.5.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/rinda-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/resolv-replace-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/resolv-0.2.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/reline-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/readline-ext-0.1.4.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/readline-0.0.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/racc-1.6.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/psych-4.0.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pstore-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/prettyprint-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pp-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/pathname-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ostruct-0.5.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/optparse-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/openssl-3.0.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/open3-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/open-uri-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/observer-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/nkf-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/net-protocol-0.1.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/net-http-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/mutex_m-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/logger-1.5.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/json-2.6.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/irb-1.4.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/ipaddr-1.2.4.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-wait-0.2.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-nonblock-0.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/io-console-0.5.11.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/getoptlong-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/forwardable-1.3.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/find-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fileutils-1.6.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fiddle-1.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/fcntl-1.0.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/etc-1.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/error_highlight-0.3.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/erb-2.2.3.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/english-0.7.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/drb-2.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/digest-3.1.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/did_you_mean-1.6.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/delegate-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/date-3.2.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/csv-3.2.2.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/cgi-0.3.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/bigdecimal-3.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/benchmark-0.2.0.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/base64-0.1.1.gemspec
4,0K	/Users/deivid/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0/specifications/default/abbrev-0.1.0.gemspec
```

This could be optimized further by limiting the list of files to requirable files, although that would require some work.


> Since `s.files =` seems by far the largest field of a gemspec, it is a bit surprising that this approach gains something. Maybe there is some non-trivial costs for other fields being assigned? Otherwise we're just measuring Ruby parser speed vs read with buffer, split by lines, split by \0, at the cost of storing the information twice.

The cost we saving is the cost of loading all default marshaled specs and evaluating all of them, every time we boot. With this approach we only need to read the first few lines of the stub specification files. This approach just follows what's done for other fields, for the same reason.

> Yet multiple people seem annoyed by it, IMHO worth fixing. I didn't report it until now but I've been annoyed by that several times (when debugging stuff related to default gems).

I only remember one person annoyed in the issue I linked to (you are the second), but as I said in that issue, I don't mind fixing it. Do you consider fixing it a blocker for this?

deivid-rodriguez:
Well I'm now reviewing the list of files generated for some gems, and I think this patch is still not fully correct... Will review it more closely.

eregon:
> By size you mean the size of the file?

Yes. It looks like in the `du/ls` output above it's twice the same, I guess a copy/paste mistake.
(`ls -l` would also show more precise sizes)

I think even if it only affects a few gems now, it's not unlikely it will affect more in the future and there are already many default gems. Also this approach could probably generalize to non-default gems, and then the issue would be more prominent.
The reason many default gems don't have many Ruby files is many are C extensions, but that will change once JRuby/TruffleRuby integrate their handling of those default gems upstream.

> The cost we saving is the cost of loading all default marshaled specs

What do you mean by marshaled? That it's generated Ruby code?

> and evaluating all of them, every time we boot. With this approach we only need to read the first few lines of the stub specification files. This approach just follows what's done for other fields, for the same reason.

Yes. I think there shouldn't be much of a difference between them (it's more or less the same work from a conceptual point of view), except of course going through the Ruby parser and evaluating vs manually parsing those first lines.
I wonder how much we'd gain in time if we'd use `# frozen_string_literal: true ` + `%w[file1 file2]`.

> I only remember one person annoyed in the issue I linked to (you are the second), but as I said in that issue, I don't mind fixing it. Do you consider fixing it a blocker for this?

I think currently the `\0` is only there when there are multiple native extensions, so basically just json and a few others maybe. But this would make all default gemspecs unreadable in most editors and CLI tools, so it makes that issue significantly worse.

deivid-rodriguez:
> Yes. It looks like in the du/ls output above it's twice the same, I guess a copy/paste mistake.
(ls -l would also show more precise sizes)

Sorry, edited. As you can see, only two gemspecs are noticiably affected.

> I think even if it only affects a few gems now, it's not unlikely it will affect more in the future and there are already many default gems. Also this approach could probably generalize to non-default gems, and then the issue would be more prominent.

Again, this seems a very weak concern, since you have not even proved that this is actually an issue at all. And no, non-default gems don't use the `files` field at all, so won't ever need this.

> What do you mean by marshaled? That it's generated Ruby code?

Sorry, forget about this. I mixed up the approaches to this issue. The approach proposed at #3799 was to `Marshal.dump` loaded default specifications save that to a file, and then `Marshal.load` from the file on subsequent boots. This approach is completely unrelated to that.

> Yes. I think there shouldn't be much of a difference between them (it's more or less the same work from a conceptual point of view), except of course going through the Ruby parser and evaluating vs manually parsing those first lines.
I wonder how much we'd gain in time if we'd use # frozen_string_literal: true + %w[file1 file2].

I took the idea from what was already done to avoid evaluating full gemspecs just for using common fields like the `version` or `platform`. Not sure why it's so slow, but the stack profiling showed at #3799 shows that a big amount of boot time is spent loading specification files. I guess because `eval` is slow? In this case, the culprit for loading the full gemspec instead of just reading the `stub` lines was the `files` field, hence the approach taken here.

> I think currently the \0 is only there when there are multiple native extensions, so basically just json and a few others maybe. But this would make all default gemspecs unreadable in most editors and CLI tools, so it makes that issue significantly worse.

Yes, that why I asked you previously whether you consider this issue a blocker for intruducing this. If it affected the same set of gemspecs then it would be obviously not a blocker. By the way, which editors did you actually check? Just saying because I tried two editors and they handled it just fine.

eregon:
> Sorry, edited. As you can see, only two gemspecs are noticiably affected.

But that's only looking by 4KB rounded sizes, which is very imprecise.
Also we have to consider that the current encoding of `.files =` is very inefficient, but if it was `%w[file1 file2]` it would be as efficient as the stub line, so the characters for representing the list of files would be exactly the double with this approach.

> By the way, which editors did you actually check?

IIRC IntelliJ and Atom, also `cat` and `most` on the command line.
In some of them the file is recognized as binary and not text, and so not shown properly (because it contains \0).

eregon:
Regarding the general approach, having a list of all files for default gems feels a bit overkill.
The way lazy rubygems deals with this is just assume the gem name is the prefix for all its required files, so whether it is a default gem file can be checked with much less data: https://github.com/oracle/truffleruby/blob/master/src/main/ruby/truffleruby/core/truffle/gem_util.rb

deivid-rodriguez:
> But that's only looking by 4KB rounded sizes, which is very imprecise.

I really don't understand how this minimal difference in file size could affect performance noticeably. It seems useless to me to measure the file size difference more precisely. It would be a similar concern to oppose to someone extend the gemspec description of a default gem because it will make things slower, no?

> IIRC IntelliJ and Atom, also cat and most on the command line.

I don't see the issue with `cat` and `more`, they work perfectly on my side. Either the ignore the "\0" altogether (cat), or they show a little marker (`more`). Anyways, I'll just fix that issue since it will probably take less that keep discussing this 😆.

> Regarding the general approach, having a list of all files for default gems feels a bit overkill.

Note that this is not the approach of this PR, this is how things already work. Let's focus on this PR.

All in all, I'm not convinced at all by your reservations, but I'll implement changing the marker before moving this forward. I'll refrain from commenting here anymore until I'm done with that.

eregon:
My main reservation is when looking at the file, and seeing two huge list of filenames it looks quite messy and obvious duplication.
I think it'd be nicer to have a separate file with that information, and potentially use a binary format there instead of a line-based format for the best speed.

FWIW, loading default gemspecs is quite slow, here is a flamegraph on truffleruby: [flamegraph.zip](https://github.com/rubygems/rubygems/files/9396697/flamegraph.zip) (as a zip so it remains dynamic).
Notably the `sort` from `Gem::Specification#files` seems expensive and is probably unnecessary.

It'd be nice to not need to do so much work just to find out if the require is requiring a default gem.
In this case, it actually takes longer to figure `load_defaults` than actually loading `chunky_png` and `zlib`.

voxik:
I also have reservation against this PR. Not because I think that RubyGems should not be faster, but because the default gems would deserve rework such as #2909. If that landed, StdLib could truly be split into gems. How would be preserved the current StdLib functionality (with `--disable-gems`?), that is a question. But Ruby could somehow populate the `$LOAD_PATH`. In that case, Ruby's own `require` would be in charge of loading the libraries and RubyGems would do less.

voxik:
> but don't understand why those big projects should block quick wins like this, unless there are any real downsides of doing this, which I have not yet heard of.

Because quick wins like this makes stuff more complicated and also more fragile. How many lines of code this adds? Prior, there were already 3 places where the file list could be taken from (directory, original .gemspec, installed gemspec) and this adds 4th location.

deivid-rodriguez:
I think this is a good compromise and does not add that much complexity. We can always switch to simpler solutions in the future like the ones that have been suggested here, but reusing this technique also for default specification files seems like an easy win for now to me.

deivid-rodriguez:
I'll try to find some time to rebase this and remeasure. The speed up now may be more than 20% given that there are more default gems now.

PR #4205, opened 2020-12-27
Update rubocop to supported ruby

lukaso:

Rubocop was struggling to run as the newer versions of rubocop as it [no longer support ruby version 2.3](https://docs.rubocop.org/rubocop/compatibility.html).

Update ruby version in `.rubocop.yml`.

- [n/a ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,7 @@ AllCops:
     - 'lib/rubygems/resolver/molinillo/**/*'
     - 'pkg/**/*'
     - 'tmp/**/*'
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.7
 
 Layout/AccessModifierIndentation:
   Enabled: true


deivid-rodriguez (commented):
Hi @lukaso.

This means potentially using cops specific to 2.7 syntax, which we don't want since we still support ruby 2.3. Technically we could do this and hope that if someone makes a mistake and enables a cop that it's not compatible with our range of supported rubies, our tests will catch it. But it sounds like there's little benefit on doing it. We can upgrade this to our new minimum once we drop ruby 2.3 support.

What you need to do is run `rake setup` from the root of the repository, so that you get the proper `rubocop` version installed that we use, and that it stills support ruby 2.3.

deivid-rodriguez:
I'm closing this due to the above but if something is not clear or you disagree, please let me know!

PR #4209, opened 2020-12-28
Fix bundle man pages display on truffleruby

deivid-rodriguez:

The problem is that when `bundler` is installed as a default gem on truffleruby, our pretty man pages are not displayed.

By default, you get the following on truffleruby:

<details>
<summary>bundle install --help (thor's fallback)</summary>

```
Usage:
  bundle install [OPTIONS]

Options:
      [--binstubs=Generate bin stubs for bundled gems to ./bin]                                                                                                             
      [--clean=Run bundle clean automatically after install], [--no-clean]                                                                                                  
      [--deployment=Install using defaults tuned for deployment environments], [--no-deployment]                                                                            
      [--frozen=Do not allow the Gemfile.lock to be updated after this install], [--no-frozen]                                                                              
      [--full-index=Fall back to using the single-file index of all gems], [--no-full-index]                                                                                
      [--gemfile=Use the specified gemfile instead of Gemfile]                                                                                                              
  -j, [--jobs=Specify the number of jobs to run in parallel]                                                                                                                
      [--local=Do not attempt to fetch gems remotely and use the gem cache instead], [--no-local]                                                                           
      [--no-cache=Don't update the existing gem cache.]                                                                                                                     
  --force, [--redownload=Force downloading every gem.], [--no-redownload]                                                                                                   
      [--no-prune=Don't remove stale gems from the cache.]                                                                                                                  
      [--path=Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME). Bundler will remember this value for future installs on this machine]           
      [--quiet=Only output warnings and errors.], [--no-quiet]                                                                                                              
      [--shebang=Specify a different shebang executable name than the default (usually 'ruby')]                                                                             
      [--standalone=Make a bundle that can work without the Bundler runtime]                                                                                                
      [--system=Install to the system location ($BUNDLE_PATH or $GEM_HOME) even if the bundle was previously installed somewhere else for this application], [--no-system]  
      [--trust-policy=Gem trust policy (like gem install -P). Must be one of HighSecurity|MediumSecurity|LowSecurity|AlmostNoSecurity|NoSecurity]                           
      [--without=Exclude gems that are part of the specified named group.]                                                                                                  
      [--with=Include gems that are part of the specified named group.]                                                                                                     
      [--no-color]                                                                                                                                                          # Disable colorization in output
  -r, [--retry=NUM]                                                                                                                                                         # Specify the number of times you wish to attempt network commands
  -V, [--verbose], [--no-verbose]                                                                                                                                           # Enable verbose output mode

Description:
  Install will install all of the gems in the current bundle, making them available for use. In a freshly checked out repository, this command will give you the same gem versions as the last person who updated the Gemfile and ran `bundle 
  update`.

  Passing [DIR] to install (e.g. vendor) will cause the unpacked gems to be installed into the [DIR] directory rather than into system gems.

  If the bundle has already been installed, bundler will tell you so and then exit.
```

</details>

That's thor's automatically generated help, much less curated than our man page:

<details>
<summary>bundle install --help (pretty)</summary>

```
BUNDLE-INSTALL(1)                                                                                                                                                                                                      BUNDLE-INSTALL(1)

NAME
       bundle-install - Install the dependencies specified in your Gemfile

SYNOPSIS
       bundle  install  [--binstubs[=DIRECTORY]]  [--clean]  [--deployment]  [--frozen]  [--full-index]  [--gemfile=GEMFILE] [--jobs=NUMBER] [--local] [--no-cache] [--no-prune] [--path PATH] [--quiet] [--redownload] [--retry=NUMBER]
       [--shebang] [--standalone[=GROUP[ GROUP...]]] [--system] [--trust-policy=POLICY] [--with=GROUP[ GROUP...]] [--without=GROUP[ GROUP...]]

DESCRIPTION
       Install the gems specified in your Gemfile(5). If this is the first time you run bundle install (and a Gemfile.lock does not exist), Bundler will fetch all remote sources, resolve dependencies and install all needed gems.

       If a Gemfile.lock does exist, and you have not updated your Gemfile(5), Bundler will fetch all remote sources, but use the dependencies specified in the Gemfile.lock instead of resolving dependencies.

       If a Gemfile.lock does exist, and you have updated your Gemfile(5), Bundler will use the dependencies in the Gemfile.lock for all gems that you did not update, but will re-resolve the dependencies of gems that you did update.
       You can find more information about this update process below under CONSERVATIVE UPDATING.

OPTIONS
       The  --clean,  --deployment, --frozen, --no-prune, --path, --shebang, --system, --without and --with options are deprecated because they only make sense if they are applied to every subsequent bundle install run automatically
       and that requires bundler to silently remember them. Since bundler will no longer remember CLI flags in future versions, bundle config (see bundle-config(1)) should be used to apply them permanently.

       --binstubs[=<directory>]
              Binstubs are scripts that wrap around executables. Bundler creates a small Ruby file (a binstub) that loads Bundler, runs the command, and puts it in bin/. This lets you link the binstub inside of an application to the
              exact gem version the application needs.

              Creates  a  directory  (defaults  to  ~/bin) and places any executables from the gem there. These executables run in Bundler´s context. If used, you might add this directory to your environment´s PATH variable. For in‐
              stance, if the rails gem comes with a rails executable, this flag will create a bin/rails executable that ensures that all referred dependencies will be resolved using the bundled gems.

       --clean
              On finishing the installation Bundler is going to remove any gems not present in the current Gemfile(5). Don´t worry, gems currently in use will not be removed.

              This option is deprecated in favor of the clean setting.

(...)
```

</details>

In addition to that, if you install bundler as a default gem from source, you'll get a crash when running a help command, like this:

```
$ bundle install --help
(... BUG REPORT TEMPLATE HERE ...)
TypeError: no implicit conversion of nil into String
  <internal:core> core/type.rb:277:in `convert_type'
  <internal:core> core/type.rb:257:in `rb_convert_type'
  <internal:core> core/kernel.rb:171:in `StringValue'
  <internal:core> core/type.rb:438:in `coerce_to_path'
  <internal:core> core/file.rb:744:in `join'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/cli.rb:128:in `help'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/cli.rb:30:in `dispatch'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/cli.rb:24:in `start'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/gems/gems/bundler-2.3.0.dev/exe/bundle:49:in `block in <top (required)>'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/ruby/site_ruby/20.3.0/bundler/friendly_errors.rb:130:in `with_friendly_errors'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/lib/gems/gems/bundler-2.3.0.dev/exe/bundle:37:in `<top (required)>'
  <internal:core> core/kernel.rb:401:in `load'
  <internal:core> core/kernel.rb:401:in `load'
  /home/deivid/.rbenv/versions/truffleruby-20.3.0/bin/bundle:23:in `<main>'
```

My fix is to move generated man pages to `lib/bundler/man` together with their sources and display them from there, so that we no longer depend on `RbConfig::CONFIG["mandir"]`, which is not set on truffleruby.


deivid-rodriguez:
I pushed https://github.com/rubygems/rubygems/compare/restore_man_bundle_for_now with an alternative. It has the downside that it shows the ronn sources as opposed to the actual man pages on truffleruby, and that it restores all the code that I was able to delete. But it restores `bundle install --help` to not crashing on truffleruby, and doesn't change any future behaviour of `man bundle`.

If I get no feedback I'll still go with what we have now, but let me know if you prefer the alternative and I will proceed with it.

eregon:
Thanks for this :) I just noticed it through the release notes.
We could define `RbConfig::CONFIG["mandir"]` on TruffleRuby, but I agree that it's probably best people use `bundle command --help` rather than `man bundle` as that only works for the default Bundler of the system Ruby (out of the box => for most users).

PR #4212, opened 2020-12-29
Prepare bundler 2.2.4 & rubygems 3.2.4 releases

deivid-rodriguez:
Getting ready for a new round of releases.


deivid-rodriguez:
@nobu has raised some issues about #4209, so I'll hold the release until we figure it out.

PR #4216, opened 2020-12-30
Speed up complex dependency resolves by creating DepProxy factory and cache

lukaso:

Running some (relatively complex) resolution cases can lead to extremely poor performance of `Resolving dependencies`. Here is a trivial case to reproduce this which runs in 22 minutes on my laptop.

Gemfile
```ruby
source 'https://rubygems.org'

gem "actionmailer"
gem "mongoid", ">= 0.10.2"
```

The same problem has been seen in the real world at [Simply Business](https://www.simplybusiness.co.uk) where we have repeatedly run into resolution taking excessively long or even hanging.

### Solution

The solution in this PR takes the above case from 22 minutes to 1 minute 32 seconds (on my laptop).

On other runs with different gem files, I've seen improvements including a 30% improvement and a 5 x improvement.

For the general case, there is no performance implications.

The basic solution (I will describe the diagnostics below) that led me to this requires two things:

1. Dedup `DepProxy` by creating a factory method and comparing requests for new `DepProxy`s to existing proxies (which creates a high cache hit rate). The using the hash of the objects to cache. This also allows comparison of two exactly the same `DepProxy`s to use the inbuilt `Object#==` which is very fast. This does mean that the objects are now shared. If they were immutable before, this would be totally fine, but even if they aren't, it appears not to cause problems.
2. Changing the Molinillo `build_details_for_unwind` in `resolution.rb` with an implementation for `relevant_unused_unwinds` that uses `Array#-` which takes advantage of object comparison rather than `Array#&` which uses `Array#hash` and in this case is substantially slower. Without the Molinillo change, the `DepProxy` factory is performance neutral.

`build_details_for_unwind` and specifically `Array#&` were where most of the time in the cases above were spent, as the analysis will demonstrate.

#### Note on the PR

I present the full PR with vendored changes, assuming if you would like to move forward I would submit a PR to `Molinillo`.

### Analysis

The added `slow_perf_spec.rb` can run against this branch or `master`. On this branch it runs in:

```
Finished in 4 minutes 37 seconds (files took 0.25304 seconds to load)
```

On `master` it runs:

```
Finished in 6 minutes 29 seconds (files took 0.27068 seconds to load)
```

This gives a savings of two minutes or ~ 30%.

Here is the output of the old and new algorithms:

#### DepProxy branch

```sh
$ rm Gemfile.lock; time ~/projects/rubygems/bundler/bin/bundle --gemfile=Gemfile
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...............
Using concurrent-ruby 1.1.7
Using minitest 5.14.2
Using zeitwerk 2.4.2
Using builder 3.2.4
Using erubi 1.10.0
Using mini_portile2 2.4.0
Using crass 1.0.6
Using rack 2.2.3
Using mini_mime 1.0.2
Using bson 4.11.1
Using bundler 2.3.0.dev
Using leshill-will_paginate 2.3.11
Using durran-validatable 2.0.1
Using i18n 1.8.5
Using nokogiri 1.10.10
Using tzinfo 2.0.4
Using loofah 2.8.0
Using activesupport 6.1.0
Using mongo 2.14.0
Using rails-html-sanitizer 1.3.0
Using rack-test 1.1.0
Using mail 2.7.1
Using globalid 0.4.2
Using rails-dom-testing 2.0.3
Using activejob 6.1.0
Using actionview 6.1.0
Using actionpack 6.1.0
Using mongoid 1.0.6
Using actionmailer 6.1.0
Bundle complete! 2 Gemfile dependencies, 29 gems now installed.

real	1m31.824s
user	1m30.883s
sys	0m0.497s
```

#### Master branch

```sh
$ rm Gemfile.lock; time ~/projects/rubygems/bundler/bin/bundle --gemfile=Gemfile
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies........................
Using concurrent-ruby 1.1.7
Using minitest 5.14.2
Using zeitwerk 2.4.2
Using builder 3.2.4
Using erubi 1.10.0
Using mini_portile2 2.4.0
Using crass 1.0.6
Using rack 2.2.3
Using mini_mime 1.0.2
Using bson 4.11.1
Using bundler 2.3.0.dev
Using leshill-will_paginate 2.3.11
Using durran-validatable 2.0.1
Using i18n 1.8.5
Using tzinfo 2.0.4
Using nokogiri 1.10.10
Using activesupport 6.1.0
Using mail 2.7.1
Using rails-dom-testing 2.0.3
Using rack-test 1.1.0
Using globalid 0.4.2
Using mongo 2.14.0
Using loofah 2.8.0
Using activejob 6.1.0
Using rails-html-sanitizer 1.3.0
Using mongoid 1.0.6
Using actionview 6.1.0
Using actionpack 6.1.0
Using actionmailer 6.1.0
Bundle complete! 2 Gemfile dependencies, 29 gems now installed.

real	22m27.811s
user	22m15.555s
sys	0m4.723s
```

##### Original implementation

```
Measure Mode: wall_time
Thread ID: 70207129452500
Fiber ID: 70207121129620
Total: 514.729486
Sort by: self_time

 %self      total      self      wait     child     calls  name                           location
 54.20    481.506   278.970     0.000   202.536  2230632   Array#&
 38.29    197.141   197.101     0.000     0.039 870039195   Bundler::DepProxy#hash         /Users/lukas.oberhuber/projects/rubygems/bundler/lib/bundler/dep_proxy.rb:12
  0.95     12.820     4.877     0.000     7.943  7039049   Bundler::DepProxy#eql?         /Users/lukas.oberhuber/projects/rubygems/bundler/lib/bundler/dep_proxy.rb:16
  0.72    493.137     3.682     0.000   489.455   233778  *Array#select
  0.65      7.320     3.367     0.000     3.953  8345705   Gem::Dependency#==             /Users/lukas.oberhuber/.rbenv/versions/2.6.5/lib/ruby/2.6.0/rubygems/dependency.rb:173
  0.37      2.953     1.899     0.000     1.054  2239519   Bundler::Molinillo::Resolver::Resolution::UnwindDetails#<=> /Users/lukas.oberhuber/projects/rubygems/bundler/lib/bundler/vendor/molinillo/lib/molinillo/resolution.rb:83
  0.37     10.352     1.884     0.000     8.469   138522  *Array#each
  0.36      1.876     1.876     0.000     0.000 15642359   Bundler::Molinillo::Resolver::Resolution::UnwindDetails#state_index
  ...
```

##### With adding a DepProxy factory

```
Measure Mode: wall_time
Thread ID: 70284950585360
Fiber ID: 70284980000360
Total: 97.786107
Sort by: self_time

 %self      total      self      wait     child     calls  name                           location
 76.78     75.077    75.077     0.000     0.000  2327396   Array#-
  3.54     85.954     3.465     0.000    82.489    95397   Array#select
  1.87      2.805     1.831     0.000     0.974  2239519   Bundler::Molinillo::Resolver::Resolution::UnwindDetails#<=> /Users/lukas.oberhuber/projects/rubygems/bundler/lib/bundler/vendor/molinillo/lib/molinillo/resolution.rb:83
  1.83      1.789     1.789     0.000     0.000 15642359   Bundler::Molinillo::Resolver::Resolution::UnwindDetails#state_index
  1.81      9.044     1.774     0.000     7.270   138532  *Array#each
  0.95      1.570     0.928     0.000     0.642   260374   Gem::Version#<=>               /Users/lukas.oberhuber/.rbenv/versions/2.6.5/lib/ruby/2.6.0/rubygems/version.rb:342
  0.87      0.852     0.852     0.000     0.000  4454118   Bundler::Molinillo::Resolver::Resolution::UnwindDetails#requirements_unwound_to_instead
  0.74      3.494     0.720     0.000     2.774  2218106   Comparable#>
  0.66      1.028     0.650     0.000     0.378   350426   Array#reject!
  ...
```

### Other interesting findings

Interestingly, by switching the `mongoid` version from `>=0.10.2` to `>=0.10.3`, steps to resolve go from 6264 down to 827 due to picking a different starting requirment (although I haven't analysed the algorithm to understand why the change is triggered). Red or `<` version is `>=0.10.3` and Green or `>` is `>=0.10.2`. There is also a dramatic slowdown in dependency resolution between the two options, due to the very slow algorithm (even with the fixes proposed here) related to looking for previous unwind states.

```diff
< BUNDLER: Starting resolution (2020-12-30 00:44:11 +0000)
< BUNDLER: User-requested dependencies: [#<Bundler::DepProxy:0x00007fef5f0d8058 @dep=<Bundler::Dependency type=:runtime name="actionmailer" requirements=">= 0">, @__platform=#<Gem::Platform:0x00007fef5f016098 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=325717142915182719>, #<Bundler::DepProxy:0x00007fef66073408 @dep=<Bundler::Dependency type=:runtime name="mongoid" requirements=">= 0.10.3">, @__platform=#<Gem::Platform:0x00007fef5f016098 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=1448836936842474097>, #<Bundler::DepProxy:0x00007fef66073070 @dep=<Bundler::Dependency type=:runtime name="Ruby\u0000" requirements=">= 0">, @__platform=#<Gem::Platform:0x00007fef5f016098 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-4097237829671024793>, #<Bundler::DepProxy:0x00007fef66072800 @dep=<Bundler::Dependency type=:runtime name="RubyGems\u0000" requirements="= 3.1.4">, @__platform=#<Gem::Platform:0x00007fef5f016098 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-3847635433202042685>]
---
> BUNDLER: Starting resolution (2020-12-30 00:40:45 +0000)
> BUNDLER: User-requested dependencies: [#<Bundler::DepProxy:0x00007fc6070f4518 @dep=<Bundler::Dependency type=:runtime name="actionmailer" requirements=">= 0">, @__platform=#<Gem::Platform:0x00007fc6078f1ec8 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-3111283843832305287>, #<Bundler::DepProxy:0x00007fc60c147dd0 @dep=<Bundler::Dependency type=:runtime name="mongoid" requirements=">= 0.10.2">, @__platform=#<Gem::Platform:0x00007fc6078f1ec8 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-4225324059996933406>, #<Bundler::DepProxy:0x00007fc60c147358 @dep=<Bundler::Dependency type=:runtime name="Ruby\u0000" requirements=">= 0">, @__platform=#<Gem::Platform:0x00007fc6078f1ec8 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-329411114533298518>, #<Bundler::DepProxy:0x00007fc60c146ed0 @dep=<Bundler::Dependency type=:runtime name="RubyGems\u0000" requirements="= 3.1.4">, @__platform=#<Gem::Platform:0x00007fc6078f1ec8 @cpu="x86_64", @os="darwin", @version="19" @cpu="x86_64", @os="darwin", @version="19">, @hash=-3561920153128883523>]
13,20c13,20
< BUNDLER:     Creating possibility state for mongoid (>= 0.10.3) x86_64-darwin-19 (428 remaining)
< BUNDLER:       Attempting to activate [mongoid (7.2.0) (x86_64-darwin-19)]
< BUNDLER:       Activated mongoid at [mongoid (7.2.0) (x86_64-darwin-19)]
< BUNDLER:       Requiring nested dependencies (activemodel (>= 5.1, < 6.1) x86_64-darwin-19, mongo (>= 2.10.5, < 3.0.0) x86_64-darwin-19)
< BUNDLER:       Creating possibility state for actionmailer x86_64-darwin-19 (744 remaining)
< BUNDLER:         Attempting to activate [actionmailer (6.1.0) (x86_64-darwin-19)]
< BUNDLER:         Activated actionmailer at [actionmailer (6.1.0) (x86_64-darwin-19)]
< BUNDLER:         Requiring nested dependencies (activesupport (= 6.1.0) x86_64-darwin-19, actionpack (= 6.1.0) x86_64-darwin-19, actionview (= 6.1.0) x86_64-darwin-19, activejob (= 6.1.0) x86_64-darwin-19, mail (~> 2.5, >= 2.5.4) x86_64-darwin-19, rails-dom-testing (~> 2.0) x86_64-darwin-19)
---
> BUNDLER:     Creating possibility state for actionmailer x86_64-darwin-19 (744 remaining)
> BUNDLER:       Attempting to activate [actionmailer (6.1.0) (x86_64-darwin-19)]
> BUNDLER:       Activated actionmailer at [actionmailer (6.1.0) (x86_64-darwin-19)]
> BUNDLER:       Requiring nested dependencies (activesupport (= 6.1.0) x86_64-darwin-19, actionpack (= 6.1.0) x86_64-darwin-19, actionview (= 6.1.0) x86_64-darwin-19, activejob (= 6.1.0) x86_64-darwin-19, mail (~> 2.5, >= 2.5.4) x86_64-darwin-19, rails-dom-testing (~> 2.0) x86_64-darwin-19)
> BUNDLER:       Creating possibility state for mongoid (>= 0.10.2) x86_64-darwin-19 (430 remaining)
> BUNDLER:         Attempting to activate [mongoid (7.2.0) (x86_64-darwin-19)]
> BUNDLER:         Activated mongoid at [mongoid (7.2.0) (x86_64-darwin-19)]
> BUNDLER:         Requiring nested dependencies (activemodel (>= 5.1, < 6.1) x86_64-darwin-19, mongo (>= 2.10.5, < 3.0.0) x86_64-darwin-19)
```


deivid-rodriguez:
Oh wow, we definitely want this  it's great :tada:! I'll have a detailed look tomorrow :+1:.

indirect:
Great investigation work here, thanks for the PR!

lukaso:
My pleasure. Thanks.

lukaso:
This additional change looks like it took this case from 1m32s down to 10s on my laptop. The `requirements_unwound_to_instead` array was not being deduplicated on insert.

bundler/lib/bundler/dep_proxy.rb
```diff
@@ -4,18 +4,27 @@ module Bundler
...
+
     def initialize(dep, platform)
       @dep = dep
       @__platform = platform
     end
 
-    def hash
-      @hash ||= [dep, __platform].hash
+    private_class_method :new
+    private :dup
```
deivid-rodriguez, line 19: This is to avoid potential future programming mistakes where bundler code tries to instantiate or dup `DepProxy` objects directly, right? Should we add a couple of unit tests to cover those code paths and illustrate this?

bundler/lib/bundler/dep_proxy.rb
```diff
@@ -4,18 +4,27 @@ module Bundler
...
+
+    # reset before calling
+    def self.setup!
+      @proxies = {}
```
deivid-rodriguez, line 23: Is this in case bundler tries to resolve two different definitions in the same invokation? I'm not sure `bundler` whether ever does that... If it's for tests only, should we move this to test code?

bundler/lib/bundler/dep_proxy.rb
```diff
@@ -4,18 +4,27 @@ module Bundler
...
     end
 
-    def ==(other)
-      return false if other.class != self.class
-      dep == other.dep && __platform == other.__platform
+    def hash
+      @hash ||= [dep, __platform].hash
```
deivid-rodriguez, line 27: If I understood your description correctly, molinillo now always does "object based comparison", so the specific hash ruby (or us) assigns to the object should be irrelevant since it's no longer used (and even if molinillo still used it somewhere for comparisons, it will still work since same objects should have the same default `#hash`, correct? In order words, do we need this method?

bundler/spec/realworld/edgecases_spec.rb
```diff
@@ -450,10 +450,13 @@ def rubygems_version(name, requirement)
...
 
+    expect(out).to include("BUNDLER: Unactivated: \n")
+    expect(out).to include("BUNDLER: Finished resolution (")
+    steps = out.match(/BUNDLER: Finished resolution \((\d+) steps\)/)&.[](1)&.to_i
     if Bundler.feature_flag.bundler_3_mode?
-      expect(out).to include("BUNDLER: Finished resolution (2492 steps)")
+      expect(steps).to be_between(2420, 2520)
     else
-      expect(out).to include("BUNDLER: Finished resolution (2722 steps)")
+      expect(steps).to be_between(2650, 2750)
```
deivid-rodriguez, line 459: Why is this change? Is it because the number of resolution steps in the spec is now non-deterministic somehow, or because you want to make the spec a bit more resilient to future changes in the resolver? If it's the latter, resolver's code hardly ever changes, so I don't mind having to update this spec when changes we make affect the number of steps needed. Actually, the performance spec you added is essentially testing the same thing as here: that the resolver doesn't take too long, so we can probably keep both tests but unify the approach.

deivid-rodriguez (commented):
Hei, this is really awesome, thanks soooooo much! :heart: 

I made some small comments, but this is going in, soon! I think you can start proposing the molinillo changes upstream :smiley:.

Also, rubygems has a separate copy of molinillo, maybe it's worth updating that copy too.

bundler/lib/bundler/dep_proxy.rb
lukaso: Done in e2573800b5357949ff7be1e6c8afd8cd9e05deaf

bundler/lib/bundler/dep_proxy.rb
lukaso: Agreed this is only for test code. There are other components that are setup in the resolver, so it seemed reasonable to do it there anyway.
I don't think we can remove this method, so it is about putting the call to `DepProxy.setup!` in a `before(:each)` block in `spec_helper.rb` if you think that would be better.

bundler/spec/realworld/edgecases_spec.rb
lukaso: On my mac on the `master` using ruby 2.7.2 this test failed because the number of steps were different. So it appears to be deterministic on github actions in exactly the case it is run in, but only then. Given that, it felt like a fragile test. With a range, this test could also be opened up beyond just ruby 2.7.2. 

Here is the beginning of the very long output:

```sh
$ bin/rspec spec/realworld/edgecases_spec.rb:349 --tag realworld                                                                            (master %=)
Run options:
  include {:realworld=>true, :locations=>{"./spec/realworld/edgecases_spec.rb"=>[349]}}
  exclude {:jruby=>true, :readline=>false, :permissions=>false, :no_color_tty=>false, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.30.0", :sudo=>true}

Randomized with seed 16292
F

Retried examples: 1
  real world edgecases doesn't hang on big gemfile
  [5/5] ./spec/realworld/edgecases_spec.rb:349


Failures:

  1) real world edgecases doesn't hang on big gemfile
     Failure/Error: expect(out).to include("BUNDLER: Finished resolution (2722 steps)")

       expected "Starting resolution (2021-01-04 17:54:33 +0000)\nUser-requested dependencies: [#<Gem::Resolver::Depe...Writing lockfile to /Users/lukas.oberhuber/projects/rubygems/bundler/tmp/1/bundled_app/Gemfile.lock" to include "BUNDLER: Finished resolution (2722 steps)"
       Diff:
       @@ -1,10033 +1,20065 @@
       -BUNDLER: Finished resolution (2722 steps)
       +Starting resolution (2021-01-04 17:54:33 +0000)
```

Way further down:

```sh
BUNDLER: Finished resolution (2738 steps) (Took 10.813019 seconds) (2021-01-04 17:54:47 +0000)
```

bundler/lib/bundler/dep_proxy.rb
deivid-rodriguez: I was thinking of something like `DepProxy.instance_variable_set(:@proxies, {})` inside tests, to make it clear that it's not a public method, but just some hack for testing several different resolutions in the same test run.

bundler/spec/realworld/edgecases_spec.rb
deivid-rodriguez: I see. How about changing to the same strategy you used for the test you added, just making sure it doesn't take too long?

deivid-rodriguez: Actually, it seems fine as it is, maybe leave a note about this for future reference.

bundler/lib/bundler/dep_proxy.rb
lukaso: Because all the tests run their own process, right? Let me look into it.

Separate question: I was wondering if there's a way to use pry inside a test run, or to do the test set up manually so that you can use the fake repos with `pry`. Is there any way to do that? I tried and failed miserably.

bundler/lib/bundler/dep_proxy.rb
lukaso: You are right, especially according to this definition: https://stackoverflow.com/questions/1791171/ruby-object-hash. I'll remove it.

bundler/spec/realworld/edgecases_spec.rb
lukaso: Will do.

bundler/lib/bundler/dep_proxy.rb
deivid-rodriguez: Normally tests run on their own process, yeah, but I think resolver tests don't and that's  why I think we need to reset the dep_proxy cache between them so that they don't pollute each other.

Regarding `pry`, I normally use puts debugging when I need to debug tests that run subproceses (most), since we I tried using a REPL I failed miserably too.

lukaso: Need for `setup!` has been removed in 5aeaf9c8ee8dda2fa9fc812059e77d2d6d98502a.

bundler/lib/bundler/dep_proxy.rb
lukaso: Removed in 37c6a7e4c273f9c8cfaeaef7b56a011521497f0d

bundler/spec/realworld/edgecases_spec.rb
lukaso: Done in 8574da2823d1c9b11f0e5bab69a5dd597f15beb8

bundler/lib/bundler/dep_proxy.rb
deivid-rodriguez: Awesome, nice catch! I'm thinking though that we should probably push this fix all the way down to `Gem::Requirement`, which is where the actual issue is, right?

bundler/lib/bundler/dep_proxy.rb
deivid-rodriguez: Thanks :heart:!

bundler/spec/realworld/edgecases_spec.rb
deivid-rodriguez: :+1:

bundler/lib/bundler/dep_proxy.rb
lukaso: Maybe. It's intentional so the question is why is it that way? Canonical segments removes any trailing 0s on versions. So '2.0.0' becomes [2] and 2.1.0 becomes [2,1].

```ruby
  def hash # :nodoc:
    canonical_segments.hash
  end

  def canonical_segments
    @canonical_segments ||=
      _split_segments.map! do |segments|
        segments.reverse_each.drop_while {|s| s == 0 }.reverse
      end.reduce(&:concat)
  end
```

deivid-rodriguez: But that's `Gem::Version`. I think for `Gem::Version` this is indeed intentional. I not 100% sure of the reason, but I think "2.0.0" and "2.0" are the same for most purposes, but it's useful to keep them different for others (for example, for display purposes?).

But for `Gem::Requirement` I don't think there's any reason why `~> 2.0.0` should ever be considered the same as `~> 2.0`.

lukaso: OK, I'm with you. let me have a go at it. However, that doesn't change the fact that we have to do it here, so I would do as a separate pull request. The reason is that this has to run against any version of rubygems, right?

deivid-rodriguez: If we need a change in `rubygems` inside `bundler`, where normally backport the change into bundler when using older rubygems through a monkeypatch in `bundler/lib/bundler/rubygems_ext.rb`. That said, I wouldn't mind leaving it as it is for now and change it later.

bundler/spec/realworld/edgecases_spec.rb
deivid-rodriguez: Do you have any idea why the number of steps now changed? I'm refering to https://github.com/rubygems/rubygems/pull/4216/commits/acc478ddbb031236740ae11506fb7cb040c9cdb1.

Anyways, I'm think that maybe it's better to keep the exact number of steps here. It makes us more aware of subtle changes in the resolver. I think we don't need to open these specs to other rubies, they represent specific realworld cases so it's fine to run the exact scenario of the original report.  And regarding running specs locally on MacOS (which is what prompted you to change this),I'm planning to check it in CI (see #4245), so it will be fixed by that :+1:.

Anyways, I realize I'm probably just being annoying at this point, and I'm planning an imminent release, so I can take this over and make  the last tweaks if that's ok with you.

lukaso: > Anyways, I realize I'm probably just being annoying at this point, and I'm planning an imminent release, so I can take this over and make the last tweaks if that's ok with you.

Not annoying. All good comments. In any case, go for it. Happy for you to drive it to live :)

lukaso: Don't forget one of the changes is on the vendored `Molinillo`.

deivid-rodriguez: Oh, of course, I forgot! Are you planning to propose them upstream? In any case, the changes we upstream to molinillo are usually very slow to get released, so what I normally do is include them and add a note to our "vendoring rake tasks" to keep track of the unreleased patches we are shipping.

deivid-rodriguez:
Ok, I've been working on finishing this up and it's almost ready. I'll force-push to this branch soon.

Current performance on my machine with the problematic Gemfile is ~3 seconds without the molinillo patch, and ~10 seconds without it :tada:.

simi:
I'll take a look once ready.

deivid-rodriguez:
Great! I created a PR to Molinillo: https://github.com/CocoaPods/Molinillo/pull/146. If it passes Molinillo's specs I think we can ship it.

lukaso:
> Ok, I've been working on finishing this up and it's almost ready. I'll force-push to this branch soon.
> 
> Current performance on my machine with the problematic Gemfile is ~3 seconds with the molinillo patch, and ~10 seconds without it 🎉.

Fantastic news.

deivid-rodriguez:
Yay, it's great! I added some tweaks to get this ready. Suming them up:

* Removed the changes to the realworld specs, since I wasn't sure. I'll still get specs working on MacOS on a separate PR.
* Given that the spec now takes 3 seconds, I cut the limit to fail the realworld spec to 120 seconds.
* I tweaked the fix in #4250 here, added a "polyfill" to `bundler` to support older rubygems not including the fix, and removed the tweak in hashes from `DepProxy`.
* I added a note in the Rakefile to mention that we're cherry-picking some unreleased changes to Molinillo.
* I regenerated cassettes, given recent changes from other PRs.

I think that's it.

@simi You can have a look now!

bundler/lib/bundler/dep_proxy.rb
```diff
@@ -4,19 +4,18 @@ module Bundler
   class DepProxy
     attr_reader :__platform, :dep
 
+    @proxies = {}
+
+    def self.get_proxy(dep, platform)
+      @proxies[[dep, platform].hash] ||= new(dep, platform).freeze
```
segiddins, line 10: I think the key here should be `[dep, platform]`, instead of that array's hash, otherwise hash collisions would lead to the wrong instance being returned

deivid-rodriguez: I mean, that's simpler and I can update it, but I don't understand why the current version could cause issues.

`Object` docs say:

> The hash value for an object may not be identical across invocations or implementations of Ruby. If you need a stable identifier across Ruby invocations and implementations you will need to generate one with a custom method.

That means it _is_ stable across the same invocation, so `[dep, platform].hash` will always return the same thing, which is what we want here, right?

indirect: I think the danger is not that the value might change, but that _different_ objects could have the same hash. Hash collisions have definitely been the cause of horrible Bundler bugs in the past, on 32-bit machines where completely different values hashed to the same hash. 😬

deivid-rodriguez: Oh I get it now! Sure, better stay on the safe side, a bug there would be terrible :grimacing:.

deivid-rodriguez: Nice catch @segiddins!

deivid-rodriguez: However, are we sure that using an array as the hash key wouldn't be problematic either? It seems to me that ruby uses `#hash` under the hood anyways so maybe it has the same issue.

Maybe

```diff
diff --git a/bundler/lib/bundler/dep_proxy.rb b/bundler/lib/bundler/dep_proxy.rb
index c7e2b40634..0c7eb8fdb6 100644
--- a/bundler/lib/bundler/dep_proxy.rb
+++ b/bundler/lib/bundler/dep_proxy.rb
@@ -4,10 +4,10 @@ module Bundler
   class DepProxy
     attr_reader :__platform, :dep
 
-    @proxies = {}
+    @proxies = Hash.new{|hash, key| hash[key] = {} }
 
     def self.get_proxy(dep, platform)
-      @proxies[[dep, platform].hash] ||= new(dep, platform).freeze
+      @proxies[dep][platform] ||= new(dep, platform).freeze
     end
 
     def initialize(dep, platform)
```

is better?

indirect: I think, although I am not 100% sure, that the implementation of `Hash` uses .hash, but does not treat two objects with the same hash as identical. Like most hash tables or hash maps, I believe it uses the hash value to determine a bucket, and then uses object equality checks within that bucket. So if we use `.hash`, I think we are removing the Hash implementation's ability to tell the difference between two objects that have the same hash but are not identical objects. That said, I don't have time to check the MRI implementation right now, so I could be wrong!

deivid-rodriguez: Still, are we sure that different arrays can have the same `#hash`? It really seems super weird to me, why would ruby do that? Shouldn't it generate a unique value for each array?

deivid-rodriguez: Ok, I think the docs are pretty clear that this can't happen: https://ruby-doc.org/core-3.0.0/Array.html#method-i-hash. I think there's no issue here.

indirect: Here's a somewhat reasonable SO discussion on the subject: https://stackoverflow.com/questions/36952150/ruby-internals-and-how-to-guarantee-unique-hash-values. Calculated hash values can always have collisions, because the hash result has less possible values than there are possible objects. You cannot guarantee every possible object will have a unique hash, so you have to use it as an optimization step rather than the final check.

deivid-rodriguez: (@indirect I missed to read your answer before posting my comments, checking it now)

indirect: (IIRC, hash collisions caused occasional confusing bugs, but we didn't realize what was happening until someone showed up with a Ruby compiled to run on a 16-bit CPU, where the possible hash values were so few that hash collisions were happening every few seconds that Bundler was running.)

deivid-rodriguez: I _think_ we are confusing the `#hash` (_method_) ruby uses to determine the key used under the hood when using objects as keys with the hash (function) used by the hash (data structure) implementation to map keys to buckets. The hash (function) can certainly have collisions that should be handled by the implementation, but I don't see why the `#hash` method on an Array should have collisions and not be unique to the array contents.

segiddins:
Because it’s mapping from an infinite dimensional space (any possible array) to a fixed-width integer. 

On Jan 7, 2021, at 12:58 PM, David Rodríguez <notifications@github.com> wrote:

﻿
@deivid-rodriguez commented on this pull request.

In bundler/lib/bundler/dep_proxy.rb:

> @@ -4,19 +4,18 @@ module Bundler
   class DepProxy
     attr_reader :__platform, :dep
 
+    @proxies = {}
+
+    def self.get_proxy(dep, platform)
+      @proxies[[dep, platform].hash] ||= new(dep, platform).freeze
I think we are confusing the #hash (method) ruby uses to determine the key used under the hood when using objects as keys with the hash (function) used by the hash (data structure) implementation to map keys to buckets. The hash (function) can certainly have collisions that should be handled by the implementation, but I don't see why the #hash method on an Array should have collisions and not be unique to the array contents.

deivid-rodriguez:
That makes sense @segiddins, but won't ruby take care of collisions when generating hashes? If you implement your own `#hash` you need to make sure that you can't have collisions, but I would expect ruby's `Array#hash` to not generate collisions? At least the docs seem to suggest so.

In any case, even if we were to really fix this I'm not sure how to go about it, since using an array as the hash key I would imagine `ruby` would use `Array#hash` under the hood for the key, and with my suggestion we still end up using `Gem::Platform` and `Gem::Requirement` as keys, which use `Array#hash` to implement their own `#hash` methods.

So, to sum up, I don't think this is a problem in practice so I'm fine with whatever alternative you or @indirect think is safest. I'll just update this to use the `Array` as the key as you suggested.

bundler/lib/bundler/dep_proxy.rb
lukaso: Before going with these alternate implementations, make sure that two *different* Gem::Dependency objects with identical contents really do create the same lookup key. Otherwise something like `to_s` could be used. So we could go with 

```ruby
@proxies["#{dep.to_s}::#{platform.to_s"] ||= new(dep, platform).freeze
```

Assuming both `to_s` functions have all the contents that make deps and platforms unique.

Without the cache hits on creation, this pr will not improve anything.

deivid-rodriguez: I think `to_s` wouldn't work because we want things like "foo >= 2.0" and  "foo >= 2.0.0" to get the same key. But the other alternatives suggested I think would work equally fine.

deivid-rodriguez:
@simi In case it helps with your review (assuming you're interested in reviewing the part about hash collisions specifically), I made some research on ruby sources:

This defines `Hash#[]`: https://github.com/ruby/ruby/blob/55e52c19e74d5df90560ea1cc4f2a2b9f5d7a5c4/hash.c#L6874, which is this method: https://github.com/ruby/ruby/blob/55e52c19e74d5df90560ea1cc4f2a2b9f5d7a5c4/hash.c#L2065, which calls the function `hash_st_like_lookup` and ends up calling `st_lookup` which implements the juicy part: https://github.com/ruby/ruby/blob/55e52c19e74d5df90560ea1cc4f2a2b9f5d7a5c4/st.c#L1007-L1031. I think the `do_hash` call in the second line calls `#hash` on the given key, in particular  this line: https://github.com/ruby/ruby/blob/55e52c19e74d5df90560ea1cc4f2a2b9f5d7a5c4/st.c#L320.

If my interpretation of the C code is correct (which is assuming too much already :sweat_smile:), that would mean that all alternatives proposed here are equally safe (or unsafe, although I would say they are all safe).

lukaso:
> If my interpretation of the C code is correct (which is assuming too much already 😅), that would mean that all alternatives proposed here are equally safe (or unsafe, although I would say they are all safe).

I think the `[dep, platform]` without the `.hash` is best. It uses ruby's hash key functionality which does not allow hash collisions. I ran a test and there are many cache hits with this variant, meaning two instances of `dep` that are equal evaluate to the same key. Given this, the current implementation on the branch is the best from my reading and testing.

deivid-rodriguez:
Still need to investigate it more, but as of today, the `Gemfile` introduced as part of #4134 is again unbearably slow (I aborted it after 18 minutes).

It seems that some recent gem releases made the number of steps in that resolution go from ~1000 to ~6000. With this PR, resolution is at least workable taking ~ 1minute.

All that to say, we should merge this. I think it has gone through enough scrutiny already. If nobody beats me to it, I'll be merging this later today, we can address any late maintainer feedback post-merge.

simi (approved):
Good job! I got some minor comments regarding naming and potential places for additional comments. But nothing to block this for now. I'll introduce them in separate PR later.

Issue #4218, opened 2020-12-30
`rake release` not working with Bundler 2.2.3


schneems:
### Post steps to reproduce the problem

Trying to release a gem fails with Bundler 2.2.3 with Ruby 3.0.0

```
$ bundle exec rake release
dead_end 1.1.1 built to pkg/dead_end-1.1.1.gem.
Tagged v1.1.1.
Untagging v1.1.1 due to error.
rake aborted!
Running `git config --get branch.heads/v1.1.1.remote` failed. Run this command directly for more detailed output.
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:204:in `sh'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:128:in `default_remote'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:118:in `git_push'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:71:in `block (2 levels) in install'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:174:in `tag_version'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/gem_helper.rb:71:in `block in install'
/Users/rschneeman/.gem/ruby/3.0.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/cli/exec.rb:63:in `load'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/cli/exec.rb:63:in `kernel_load'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/cli/exec.rb:28:in `run'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/cli.rb:497:in `exec'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.3/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
[...]

PR #4219, opened 2020-12-30
Fix running `rake release` from an ambiguous ref

deivid-rodriguez:

In bundler 2.2.0, we started pushing only the explicit tag that we are
releasing, as opposed to all local tags, because pushing all tags is unnecessary
and can lead to surprising results depending on your local remotes.

However, with this change we need to manually extract the remote to push to, and
for that we hadn't considered the edge case of releasing from an ambigous
reference, for example, releasing from a branch with the same name as the tag.

My fix is to consider that `git rev-parse --abbrev-ref HEAD` will return a
qualified reference name if the current HEAD ref is ambiguous, and also to pass
a qualified ref to `git push` when pushing the tag because otherwise it will
fail for the same reason.

Fixes https://github.com/rubygems/rubygems/issues/4218.

### DIFF
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -117,12 +117,13 @@ module Bundler
     def git_push(remote = nil)
       remote ||= default_remote
       perform_git_push remote
-      perform_git_push "#{remote} #{version_tag}"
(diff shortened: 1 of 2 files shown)


schneems:
Awesome sauce. Thanks!

PR #4220, opened 2020-12-30
Improve rubocop setup in the new gem template

### LINKED ISSUE #4231: `bundle gem` and Gemfile template
It's not clear to me why provide versions in the Gemfile, pessimistic at that. I always remove them entirely.

If the maintainers won't consider removing them, please use a recent version for `RuboCop`.  Bundler 2.2.4 uses `~0.80`. Current major version is `1` and should be used. FYI, version 0.80.0 was released almost a year ago and is 35 versions behind current stable...

deivid-rodriguez:

By default, until we drop support for ruby 2.3 and ruby 2.4, we generate gems that still specific ruby 2.3 as the minimum version.

However, in the generated Gemfile we specify `~> 0.80` as the rubocop constraint. This will end up install a rubocop version that doesn't actually support ruby 2.3.

The gem author may not notice that and might end up pushing a gem that does not support ruby 2.3 while claiming it does.

My fix is to specify a target rubocop version in the generated configuration and make sure that a version supporting it is specified in the Gemfile.

Closes #4231.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -147,6 +147,7 @@ module Bundler
         "For more information, see the RuboCop docs (https://docs.rubocop.org/en/stable/) " \
         "and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide).")
         config[:rubocop] = true
+        config[:rubocop_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.80.1" : "0.90.0"
         Bundler.ui.info "RuboCop enabled in config"
         templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
       end
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -59,6 +59,7 @@ module Bundler
         :exe              => options[:exe],
         :bundler_version  => bundler_dependency_version,
         :github_username  => github_username.empty? ? "[USERNAME]" : github_username,
+        :required_ruby_version => Gem.ruby_version < Gem::Version.new("2.4.a") ? "2.3.0" : "2.4.0",
       }
       ensure_safe_gem_name(name, constant_array)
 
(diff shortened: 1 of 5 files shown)


hsbt:
Should we drop to support < Ruby 2.6 soon?

Ruby 2.3 is officially unsupported by Ruby core team. So, We will deprecate Ruby 2.5 in 2021. I'm bit of against to add old Ruby support. We should encourage to use supported version of Ruby for gem authors.

deivid-rodriguez:
Sure, but note that I'm not adding support for anything, we already have ">= 2.3" in the default generated gemspec. I'm just preventing a potentially problematic situation.

How about I ship this as a bug fix, and then we drop support for 2.3 and 2.4 in master?

deivid-rodriguez:
I rebased #3398 for now. It had many conflicts and I did the rebase carelessly so I probably made mistakes. But it's a start :)

deivid-rodriguez:
Actually, I agree with you that new gems should focus only on supported rubies. Bundler & rubygems support is another story, I think we should wait a bit more.

How about I change this PR to generate new gems with `required_ruby_version >= 2.6` and adapt the rubocop version accordingly?

hsbt:
I'm 👍  to add this to only bundler-2.2.x because it still suggests using Ruby 2.3+.

>How about I ship this as a bug fix, and then we drop support for 2.3 and 2.4 in master?
>How about I change this PR to generate new gems with required_ruby_version >= 2.6 and adapt the rubocop version accordingly?

I also agreed above.

deivid-rodriguez:
Ok, so bumping the minimum `required_ruby_version` was not going to work for users of old rubies since they would be creating gems that they can't run themselves :rofl:.

So I went with this compromise solution:

* Bump the `required_ruby_version` in the default template 2.4, unless you're running 2.3.
* Add the latest rubocop version to the Gemfile (which supports ruby 2.4), unless you're running 2.3.
* Setup a proper `TargetRubyVersion` in the generated rubocop configuration matching the `required_ruby_version` specification.

PR #4224, opened 2020-12-31
Support repositories with default branch not named master

### LINKED ISSUE #4009: Using main branch as default I get that Revision master does not exist
Hi 👋

I have a personal gem at my_repo that a default branch is a main.
And I need specify a branch to works.
Using like that does't works

```ruby
gem 'my_gem', git: 'my_repo'
```

Works fine if I specify a branch

```ruby
gem 'my_gem', git: 'my_repo', branch: :main
```

Branch main not should be checked too?

`Bundler version 2.1.4`

deivid-rodriguez:

Bundler doesn't support non-master default branches, so you can't currently point to, for example, [simplecov's repo](https://github.com/simplecov-ruby/simplecov), unless you add an explicit `branch: => "main"` to your git sources configuration.

Bundler should just work in this case without having to make the default branch explicit.

My fix is to make sure the code doesn't assume a specific branch until we have actually cloned the repository. This means that when a git sources is printed, it no longer includes "master" if  the source is only present in the `Gemfile` and has not yet been installed.

Fixes #4009.

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -22,7 +22,7 @@ module Bundler
         @uri        = options["uri"] || ""
         @safe_uri   = URICredentialsFilter.credential_filtered_uri(@uri)
         @branch     = options["branch"]
-        @ref        = options["ref"] || options["branch"] || options["tag"] || "master"
+        @ref        = options["ref"] || options["branch"] || options["tag"]
         @submodules = options["submodules"]
         @name       = options["name"]
         @version    = options["version"].to_s.strip.gsub("-", ".pre.")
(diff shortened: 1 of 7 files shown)


simi:
If I understand this well, this also fixes "bug" when `master` branch is present, but it is not default and is still picked when no branch is specified. It will pick correct default branch even non-default `master` one is present. :muscle:

deivid-rodriguez:
Now that you mention it, yeah, it sounds like that's what the old code was doing!

deivid-rodriguez:
It should be ready now unless CI proves me wrong!

simi (approved):

deivid-rodriguez:
Thanks @simi! Since #4233 changes the same area of the code, I gave this a final rebase just to make sure the PRs don't semantically conflict somehow. I'll merge after CI :+1:.

deivid-rodriguez:
CI got a flaky issue on jruby which I'm trying to address at #4235. I'll rebasing to see how this plays with the latest CI additions, and then merge.

PR #4236, opened 2021-01-05
Reduce the number of CI jobs

deivid-rodriguez:

The problem is that we have a very big CI matrix and we still need to add entries for the recent ruby 3.0.0 and rubygems 3.2.x releases.

My fix is to reduce the number of CI jobs. The rationale for keeping these jobs is that normally bundler is run against either the rubygems version that ships with ruby, or the latest rubygems version if it has been upgraded. Upgrades to intermediate versions of rubygems are quite rare, so I think we can skip all jobs in the middle.

### DIFF
--- a/.github/workflows/older-rubygems-bundler.yml
+++ b/.github/workflows/older-rubygems-bundler.yml
@@ -19,42 +19,18 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I think this is a good idea and an easy way to cut some jobs off. I don't think this intermediate CI entries have ever detected an issue not detected by other jobs.

simi (approved):

PR #4237, opened 2021-01-05
Introduce disable_local_revision_check config

rafaelfranca:
This config, in combination with disable_local_branch_check make it
so `local.library` configuration work in the same way as changing
them Gemfile to point to a path.

This is useful specially in bisect scripts, which will change the HEAD
of the repository to point to a commit that will not include the commit
specified in the Gemfile.lock.

### DIFF
--- a/bundler/spec/install/gemfile/git_spec.rb
+++ b/bundler/spec/install/gemfile/git_spec.rb
@@ -598,6 +598,24 @@ RSpec.describe "bundle install with git sources" do
       bundle :install, :raise_on_error => false
       expect(err).to match(/The Gemfile lock is pointing to revision \w+/)
     end
+
+    it "does not explode on invalid revision on install" do
+      build_git "rack", "0.8"
+
+      build_git "rack", "0.8", :path => lib_path("local-rack") do |s|
+        s.write "lib/rack.rb", "puts :LOCAL"
+      end
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack", :git => "#{lib_path("rack-0.8")}", :branch => "master"
+      G
+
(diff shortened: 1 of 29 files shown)


bundler/spec/install/gemfile/git_spec.rb
```diff
@@ -598,6 +598,24 @@
       bundle :install, :raise_on_error => false
       expect(err).to match(/The Gemfile lock is pointing to revision \w+/)
     end
+
+    it "explodes on invalid revision on install" do
```
dorianmariefr, line 602:
```suggestion
    it "doesn't explode on invalid revision on install" do
```

?

rafaelfranca: Thanks!

simi (approved):
Looks good to me.

:thinking: To simplify this a little, we can unify both settings to disable checks to one setting key -> `disable_local_git_check`. But that can be done later if needed.

deivid-rodriguez (commented):
Hi @rafaelfranca, thanks for this!

I'm not sure I understand the reason for this PR very well. I sometimes bisect using `bundle config local.` and things work just fine for me.

Normally, your configured local path is a clone of the repository you are bisecting, so whatever revision is locked in your lockfile should be present in your local copy and this should pose no issues, right?

I imagine you could run into this issue if your local clone is out of date and the revision in the lockfile pointing the repository clone is not yet present there. In that case, running `git fetch` on it should fix it, and that's probably better than ignoring the error since otherwise you might not be bisecting against up to date code anyways.

Let me know your thoughts.

deivid-rodriguez (approved):
Well shame on me for not properly reading the description of the PR. You explained the exact kind of situation where this setting could be useful (bisecting scripts that change the HEAD of the repo).

Issue #4234, opened 2021-01-05
MacOS system ruby (universal.x86_64-darwin19) chooses to install arm64-darwin native gem


flavorjones:
:wave: I'm a maintainer of Nokogiri, and we recently started shipping native gems for darwin:

- https://rubygems.org/gems/nokogiri/versions/1.11.0-x86_64-darwin
- https://rubygems.org/gems/nokogiri/versions/1.11.0-arm64-darwin

We received a bug report yesterday (https://github.com/sparklemotion/nokogiri/issues/2165) from @rgoldberg that I'm not sure how to address, and I'm hoping someone from the RubyGems team can help me understand what's going on.

**Please describe what you're seeing**

On an Intel-based mac running Catalina (10.15.7), using the system-installed ruby (`/usr/bin/ruby`) and gem (`/usr/bin/gem`) which reports:

```
RubyGems Environment:
  - RUBYGEMS VERSION: 3.0.3
  - RUBY VERSION: 2.6.3 (2019-04-16 patchlevel 62) [universal.x86_64-darwin19]
  - INSTALLATION DIRECTORY: /Library/Ruby/Gems/2.6.0
  - USER INSTALLATION DIRECTORY: /Users/flavorjones/.gem/ruby/2.6.0
  - RUBY EXECUTABLE: /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/ruby
  - GIT EXECUTABLE: /Users/flavorjones/.nix-profile/bin/git
  - EXECUTABLE DIRECTORY: /usr/local/bin
  - SPEC CACHE DIRECTORY: /Users/flavorjones/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /Library/Ruby/Site
  - RUBYGEMS PLATFORMS:
    - ruby
    - universal-darwin-19
  - GEM PATHS:
     - /Library/Ruby/Gems/2.6.0
     - /Users/flavorjones/.gem/ruby/2.6.0
     - /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/gems/2.6.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
     - "gem" => "--no-ri --no-rdoc --no-document"
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - <redacted>
[...]

PR #4238, opened 2021-01-05
Prefer `RUBY_PLATFORM`  to `RbConfig` to extract the local platform when it has the CPU embedded

deivid-rodriguez:

On some MacOS platforms, it can happen that rubygems installs a arm targeted gem on an intel based platform.

The problem is rubygems infers the local platform to be "universal", an in that case, rubygems considers any darwin variant a good candidate for installation.  This could also be a problem in ruby, I'll double check with ruby-core developers, but for now I'm proposing a workaround/fix in rubygems.

The fix is us to make sure we infer the local platform from `RUBY_PLATFORM` since it includes * more accurate information in this case.

I added an extra commit to make it so that the new platform inferred in this case is considered x86_64, so that variant gets installed.

However, unless we understand what the correct thing to do is here, I don't think we should introduce that second change and let rubygems choose the non-platform specific version.

Fixes https://github.com/rubygems/rubygems/issues/4234.

NOTE: The PR also includes a bunch of test changes to make sure that the platform is only mocked when needed and side effects are restored afterwards. I don't think it's strictly necessary for this, but it makes our tests better and potentially catch more platform specific issues.

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -76,6 +76,7 @@ class Gem::Platform
 
       @cpu = case cpu
              when /i\d86/ then 'x86'
+             when /universal\.(.*)/ then $1
              else cpu
              end
 
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -13,6 +13,7 @@ class Gem::Platform
 
   def self.local
     arch = RbConfig::CONFIG['arch']
+    arch = RUBY_PLATFORM if RUBY_PLATFORM =~ /universal\..*darwin/
     arch = "#{arch}_60" if arch =~ /mswin(?:32|64)$/
     @local ||= new(arch)
   end
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -981,14 +981,11 @@ Also, a list:
 
   def util_set_arch(arch)
     RbConfig::CONFIG['arch'] = arch
-    platform = Gem::Platform.new arch
 
     Gem.instance_variable_set :@platforms, nil
     Gem::Platform.instance_variable_set :@local, nil
 
     yield if block_given?
-
-    platform
   end
 
   ##
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -296,8 +296,7 @@ class Gem::TestCase < Minitest::Test
   # directed to a temporary directory.  All install plugins are removed.
   #
   # If the +RUBY+ environment variable is set the given path is used for
-  # Gem::ruby.  The local platform is set to <tt>i386-mswin32</tt> for Windows
-  # or <tt>i686-darwin8.10.1</tt> otherwise.
+  # Gem::ruby.
 
   def setup
     @orig_env = ENV.to_hash
@@ -409,14 +408,6 @@ class Gem::TestCase < Minitest::Test
     Gem.searcher = nil
     Gem::SpecFetcher.fetcher = nil
 
-    @orig_arch = RbConfig::CONFIG['arch']
-
-    if win_platform?
-      util_set_arch 'i386-mswin32'
-    else
-      util_set_arch 'i686-darwin8.10.1'
-    end
-
     @orig_hooks = {}
     %w[post_install_hooks done_installing_hooks post_uninstall_hooks pre_uninstall_hooks pre_install_hooks pre_reset_hooks post_reset_hooks post_build_hooks].each do |name|
       @orig_hooks[name] = Gem.send(name).dup
@@ -441,8 +432,6 @@ class Gem::TestCase < Minitest::Test
         $LOADED_FEATURES.replace @orig_LOADED_FEATURES
       end
     end
-
-    RbConfig::CONFIG['arch'] = @orig_arch
 
     if defined? Gem::RemoteFetcher
       Gem::RemoteFetcher.fetcher = nil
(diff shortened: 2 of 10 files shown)


deivid-rodriguez:
Something seems off for jruby, I'll double check.

lib/rubygems/platform.rb
```diff
@@ -75,6 +75,7 @@ def initialize(arch)
 
       @cpu = case cpu
              when /i\d86/ then 'x86'
+             when /universal\.x86_64/ then 'x86_64'
```
larskanis, line 78: It might be useful to capture the architecture out of `cpu`. For an arm based Mac the RUBY_PLATFORM is `universal.arm64e-darwin20`. See [here](https://github.com/ffi/ffi/issues/860#issuecomment-748541442).

djberg96:
> Something seems off for jruby, I'll double check.

In JRuby, the RUBY_PLATFORM constant returns "java", so I'm pretty sure we can't do this.

djberg96:
> The problem is rubygems infers the local platform to be "universal", an in that case, rubygems considers any darwin variant a good candidate for installation.

I'd argue that it's up to the gem publisher to set the cpu arch properly since it comes up so rarely.

lib/rubygems/platform.rb
deivid-rodriguez: Nice, that makes sense, thanks!

deivid-rodriguez:
> In JRuby, the RUBY_PLATFORM constant returns "java", so I'm pretty sure we can't do this.

You're correct, for JRuby we should probably still use the old way, and I think we should report this divergence to ruby-core?

djberg96:
> > In JRuby, the RUBY_PLATFORM constant returns "java", so I'm pretty sure we can't do this.
> 
> You're correct, for JRuby we should probably still use the old way, and I think we should report this divergence to ruby-core?

What about other implementations like Rubinius or Truffleruby, though?

deivid-rodriguez:
>  I'd argue that it's up to the gem publisher to set the cpu arch properly since it comes up so rarely.

Of course, but rubygems needs to not only consider the platform declared by the gem package's author, but also the current platform where it's running in order to figure out if a given platform specific gem is a valid candidate for installation.

deivid-rodriguez:
> What about other implementations like Rubinius or Truffleruby, though?

Sure, we should make sure that whatever we do is correct for every implementation that we support.

deivid-rodriguez:
Doing some git archaeology, I think this divergence was introduced by https://github.com/ruby/ruby/commit/f18d0400616b1f456214e05738badfc0f9369449.

However I think it might contain a bug since from the commit message I think it's supposed to fill in `RbConfig::CONFIG["arch"]` with the value of `RUBY_PLATFORM`, but replacing the embedded CPU (`universal.<cpu>-darwin20`) with some compiler environment variables. However it seems to be completely wiping out the CPU part instead.

deivid-rodriguez:
@nobu fixed this in https://github.com/ruby/ruby/commit/96ce1d9a0ff64494753ad4730f36a0cd7e7a89e7, so we can special case this in rubygems but keep using `RbConfig` other than that until the change is propagated to our supported rubies.

lib/rubygems/platform.rb
deivid-rodriguez: Updated!

deivid-rodriguez:
I think this should be mostly ready now. I could extract the tests refactoring to a separate PR to make the fix more apparent.

eregon:
> What about other implementations like Rubinius or Truffleruby, though?

No need to worry, only JRuby does not reflect the native platform in `RUBY_PLATFORM`, which seems mostly for legacy reasons (https://github.com/jruby/jruby/issues/6152).

---

FWIW, in https://bugs.ruby-lang.org/issues/17021 it seems MRI developers recommend to use target_* over host_*.

Changing only if `RUBY_PLATFORM =~ /universal\..*darwin/` as done in the PR is probably safer though, so :+1: .

deivid-rodriguez:
Apparently the only person running reliably into this problem tried this PR and it didn't seem to work properly: https://github.com/sparklemotion/nokogiri/issues/2165#issuecomment-759909457. At least I wasn't expecting all those warnings to appear after this PR for anyone :man_shrugging:.

lib/rubygems/platform.rb
```diff
@@ -75,6 +76,7 @@ def initialize(arch)
 
       @cpu = case cpu
              when /i\d86/ then 'x86'
+             when /universal\.(.*)/ then $1
```
drbrain, line 79: I think `universal` included in `Gem::Platform.local` is incorrect, so stripping `universal.` leaving only `x86_64` or `arm64e` as the CPU value is correct as implemented here.

Perhaps this regexp should also be anchored, `/\A universal\.(.*)/`?

deivid-rodriguez:
In case it helps, thanks to a @lloeki's comment somewhere on github which I can no longer find, I found MacOS patches on top of the official ruby: https://opensource.apple.com/source/ruby/ruby-145.40.1/patches/.

Most of them seem totally related to this issue, like https://opensource.apple.com/source/ruby/ruby-145.40.1/patches/0001-Revert-Support-Universal-Binary-for-macOS.patch.auto.html reverting https://github.com/ruby/ruby/commit/529adb68727c591204d9d41aab513ba06f12f776, or https://opensource.apple.com/source/ruby/ruby-145.40.1/patches/tool_mkconfig.rb.diff.auto.html changing the same area of the code @nobu changed in https://github.com/ruby/ruby/commit/96ce1d9a0ff64494753ad4730f36a0cd7e7a89e7.

My impression is that MacOS developers don't want to collaborate with ruby too much :(

PR #4241, opened 2021-01-06
Bump minitest development dependency to 5.14.3

deivid-rodriguez:

Our daily CI is failing.

Make sure we use a minitest version compatible with ruby's HEAD.

### DIFF
--- a/dev_gems.rb.lock
+++ b/dev_gems.rb.lock
@@ -30,7 +30,7 @@ GEM
     jmespath (1.4.0)
     json (2.3.0)
     json (2.3.0-java)
-    minitest (5.14.2)
+    minitest (5.14.3)
     minitest-bisect (1.5.1)
       minitest-server (~> 1.0)
       path_expander (~> 1.1)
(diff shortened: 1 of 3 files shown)


dev_gems.rb.lock
```diff
@@ -112,4 +112,4 @@ DEPENDENCIES
   webrick (~> 1.6)
 
 BUNDLED WITH
-   2.2.3
+   2.3.0.dev
```
simi, line 115: This means dev bundler should be bundled with itself, is that ok?

simi (approved):

dev_gems.rb.lock
deivid-rodriguez: Yes, I think it's good to test development bundler as much as we can, so if we can use it to bundle our own development dependencies, it sounds nice to me.

The reason I added this was that I normally have dev bundler installed glocally, so everytime I run `rake setup` the lockfile changes to the dev version. By always using the dev version that no longer happens.

deivid-rodriguez: Good to go then, @simi?

PR #4242, opened 2021-01-06
Add ruby 3.0.0 and rubygems 3.2.4 to CI

deivid-rodriguez:

We're still not testing against ruby 3.0.0, and bundler against the latest 3.2 release of rubygems.

My fix is to add the missing entries to the CI matrix.


deivid-rodriguez:
It seems like it's all good, and the bottleneck is still Windows, so I guess we can do this without #4236.

PR #4245, opened 2021-01-06
Add MacOS to CI

deivid-rodriguez:

We're not testing bundler against MacOS in CI.

By adding it, we achieve several things:

* It can help us catch MacOS specific issues.
* It aligns bundler specs with rubygems which already run on MacOS.
* It allows bundler developers using MacOS to have a green suite.
* It can reduce friction with ruby-core by catching CI issues that will be detected later by them otherwise.

My fix is to add MacOS to bundler's CI.

Note that on initial runs some specs failed that test compatibility features that should've been removed a long time ago failed. Just to get to an initial green point I tentatively removed those, but we can discuss more about that before merging this.


djberg96:
We use Travis, yes? I think their matrix says it's supported, but it always bombed when I actually tried to use the thing. Not saying this is a bad idea, just something to keep in mind.

deivid-rodriguez:
No, we use Github Actions. I've heard and seen MacOS work inconsistently on Github Actions (see https://github.com/actions/virtual-environments/issues/841), but we already use it for rubygems and it works mostly fine, so I guess we can try.

simi:
@deivid-rodriguez is the only problem now in different output of an error?

```diff
-Retrying fetcher due to error (2/4): Bundler::HTTPError Could not fetch specs from http://127.0.0.1:21454/ due to underlying error <Errno::ECONNREFUSED: Failed to open TCP connection to 127.0.0.1:21454 (Connection refused - connect(2) for "127.0.0.1" port 21454) (http://127.0.0.1:21454/specs.4.8.gz)>
+Retrying fetcher due to error (2/4): Bundler::HTTPError Could not fetch specs from http://127.0.0.1:21454/ due to underlying error <timed out (http://127.0.0.1:21454/specs.4.8.gz)>
```

deivid-rodriguez:
Weird, I rebased it to see how it's going now.

deivid-rodriguez:
@simi Finally resolved the remaining issues and this is now green!

PR #4247, opened 2021-01-06
Ensure `Version#canonical_segments` cannot be modified

agrobbin:

Since `Gem::Version` instances are cached based on the version string, `canonical_segments` should not be modifiable, otherwise things like `Gem::Version#<=>` (and any other uses of `canonical_versions`) will have different results.

Consider this (on Rails 6.0.3.4):

```ruby
Rails.gem_version > Gem::Version.new('6.0')
```

This should be `true`, but we actually had it return `false` because another gem we depended on did something like this:

```ruby
segments = Rails.gem_version.canonical_segments

segments.pop until segments.length <= 2
```

This modified the `canonical_segments` array in place, resulting in the `false` return from `>`.

The gem in question should definitely be calling `dup` on `canonical_segments`, but there's nothing that stops it from *not* doing so, and manipulating the array directly.

Now, if someone tries to do that, they'll get a `FrozenError`, which seems preferable!

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -372,7 +372,7 @@ class Gem::Version
     @canonical_segments ||=
       _split_segments.map! do |segments|
         segments.reverse_each.drop_while {|s| s == 0 }.reverse
-      end.reduce(&:concat)
+      end.reduce(&:concat).freeze
   end
 
   def freeze
--- a/test/rubygems/test_gem_version.rb
+++ b/test/rubygems/test_gem_version.rb
@@ -214,6 +214,7 @@ class TestGemVersion < Gem::TestCase
     assert_equal [1], v("1.0.0").canonical_segments
     assert_equal [1, "a", 1], v("1.0.0.a.1.0").canonical_segments
     assert_equal [1, 2, 3, "pre", 1], v("1.2.3-1").canonical_segments
+    assert v("1.0.0").canonical_segments.frozen?
   end
 
   def test_frozen_version


deivid-rodriguez (commented):
This makes total sense.

I'm just worried about abruptly breaking any gems that might be unintentionally doing this. You might argue that they are already broken but this essentially means that activerecord-pg_enum will raise an error whenever is used with whatever rubygems version ships this fix, right?

I guess we should at least wait for the fix to be released upstream, and consider making this change on our next major bump.

agrobbin:
That's a totally fair concern, @deivid-rodriguez. I would also love to know how you knew the gem in question was activerecord-pg_enum!

I've issued alassek/activerecord-pg_enum#17 to fix the issue in that gem, though I'm not sure if there are other gems of popularity that do similar things.

In the meantime, let me know what I should do here. Not sure if holding this until the next major version means the PR hangs out for awhile, if it gets merged to a different branch, or just gets merged into `master` regardless.

alassek:
I don't have any data to back this up, but in 12 years of professional Ruby I've never seen anyone else using `Gem::Version` as a value object in the way that I am. I certainly didn't intend my usage to have side-effects.

But, this is `rubygems` after all so I understand extreme caution.

deivid-rodriguez:
Yeah, I figured the same thing. I'm probably overly conservative here, but I always turn out to have been overly optimistic in the end. So I'll try the other way around this time :sweat_smile:.

> That's a totally fair concern, @deivid-rodriguez. I would also love to know how you knew the gem in question was activerecord-pg_enum!

It's because I have extrasensory perception. Nah, I just poked around your profile hoping to find it and I did :)

> In the meantime, let me know what I should do here. Not sure if holding this until the next major version means the PR hangs out for awhile, if it gets merged to a different branch, or just gets merged into master regardless.

I think I'll tag the PR with "major bump" for now  and we can leave it here until the right time comes.

deivid-rodriguez:
Hei @agrobbin!

In the end, we ended up taking the same aggressive approach proposed by this PR at #6970.

I'm sorry your contribution wasn't credited in the end, we greatly appreciate it though! ❤️

agrobbin:
No problem, @deivid-rodriguez, I'm glad some version was merged!

PR #4248, opened 2021-01-06
Fix `gem outdated` incorrectly handling platform specific gems

deivid-rodriguez:

If you have an up to date platform specific gem installed, and a lower ruby specific version, it would report the gem as out of date.

See https://github.com/sparklemotion/nokogiri/issues/2165#issuecomment-754397985.

My fix is to  change the outcome of `Gem.latest_specs` to return the latest spec for each gem, not the latest spec for each gem and platform.

I don't see how the previous behavior could be useful, and it was causing this bug.

I also had to add a small tweak to a sorting method because one test was no longer relying on the previous behavior of `Gem.latest_specs` and was no returning the RUBY variant first in the edge case of two gems with the same version but different platform installed at the same time.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -844,7 +844,9 @@ class Gem::Specification < Gem::BasicSpecification
     specs.sort! do |a, b|
       names = a.name <=> b.name
       next names if names.nonzero?
-      b.version <=> a.version
+      versions = b.version <=> a.version
+      next versions if versions.nonzero?
+      b.platform == Gem::Platform::RUBY ? -1 : 1
     end
   end
 
@@ -1080,20 +1082,15 @@ class Gem::Specification < Gem::BasicSpecification
   end
 
   def self._latest_specs(specs, prerelease = false) # :nodoc:
-    result = Hash.new {|h,k| h[k] = {} }
-    native = {}
+    result = {}
 
     specs.reverse_each do |spec|
       next if spec.version.prerelease? unless prerelease
 
-      native[spec.name] = spec.version if spec.platform == Gem::Platform::RUBY
-      result[spec.name][spec.platform] = spec
+      result[spec.name] = spec
     end
 
-    result.map(&:last).map(&:values).flatten.reject do |spec|
-      minimum = native[spec.name]
-      minimum && spec.version < minimum
-    end.sort_by{|tup| tup.name }
+    result.map(&:last).flatten.sort_by{|tup| tup.name }
   end
 
   ##
--- a/test/rubygems/test_gem_commands_outdated_command.rb
+++ b/test/rubygems/test_gem_commands_outdated_command.rb
@@ -28,4 +28,22 @@ class TestGemCommandsOutdatedCommand < Gem::TestCase
     assert_equal "foo (0.2 < 2.0)\n", @ui.output
     assert_equal "", @ui.error
   end
+
+  def test_execute_with_up_to_date_platform_specific_gem
+    spec_fetcher do |fetcher|
+      fetcher.download 'foo', '2.0'
+
+      fetcher.gem 'foo', '1.0'
+      fetcher.gem 'foo', '2.0' do |s|
+        s.platform = Gem::Platform.local
+      end
+    end
+
+    use_ui @ui do
+      @cmd.execute
+    end
+
+    assert_equal "", @ui.output
+    assert_equal "", @ui.error
+  end
 end
--- a/test/rubygems/test_gem_dependency_installer.rb
+++ b/test/rubygems/test_gem_dependency_installer.rb
@@ -44,7 +44,7 @@ class TestGemDependencyInstaller < Gem::TestCase
       s.add_development_dependency 'c'
     end
 
-    util_reset_gems
+    util_setup_spec_fetcher(@a1, @a1_pre, @b1, @d1)
   end
 
   def test_install
@@ -1125,16 +1125,6 @@ class TestGemDependencyInstaller < Gem::TestCase
     @d1, @d1_gem = util_gem 'd', '1'
     @d2, @d2_gem = util_gem 'd', '2'
 
-    util_reset_gems
-  end
-
-  def util_reset_gems
-    @a1     ||= nil
-    @b1     ||= nil
-    @a1_pre ||= nil
-    @d1     ||= nil
-    @d2     ||= nil
-
-    util_setup_spec_fetcher(*[@a1, @a1_pre, @b1, @d1, @d2].compact)
+    util_setup_spec_fetcher(@d1, @d2)
   end
 end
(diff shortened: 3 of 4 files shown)


deivid-rodriguez:
This changes what `Gem.latest_specs` returns, but honestly I can see no usage for the previous behaviour. The usual bug fix vs breaking change debate. If someone is relying on the broken behaviour, I'd rather break things and get a bug report to understand what they are doing.

rgoldberg:
Wouldn't it make more sense for the function in question to return a map from platform to newest version for that platform?

If there is no version specific to your platform, then you can lookup the "RUBY" version for your fallback using an empty string as the key.

A helper function could be made that does this for you.

Instead of replacing the existing function, these could both be new functions, so as not to disrupt usage of the existing function, which could be deprecated.

deivid-rodriguez:
I'm not sure this method was expected to be generally used. I've never seen it used in the wild, but people are probably indeed using it, since people use anything defined in rubygems :rofl:. And it's not private nor `:nodoc:`, so yeah, I guess you're right :man_shrugging:.

What bothers me is that even if people are using it, I think they are probably getting results they don't want since what the method  currently does is just weird to me. The only usage I know of (`gem outdated`) was buggy because of it.

The problem with deprecating things in rubygems is that the deprecations are seen by final users, not by gem authors, and final users can't immediately do anything to fix them, so they are just annoying.

That's why I prefer to just call this a bug fix and change the behaviour if the number of affected libraries is considered small. Also, this change could "break things" if the library relies on the previous behaviour, or fix things if it relies on the new behavior.

@hsbt Could you check usage of `Gem.latest_specs` in the wild using your magic codesearch thing?

deivid-rodriguez:
Also, this method is used by `latest_spec_for` too, a method I added recently for the new rubygems plugin layout, so the regeneration of rubygems plugins is probably also buggy because of this. Although I don't know of any precompiled gem using rubygems extensions, but again, it probably exists too :sweat_smile:.

rgoldberg:
What do Ruby Gems maintainers think about this?

I'd advise adding the 2 new functions I mentioned, deprecating the existing one, and using the helper function in `gem outdated`.

If you're concerned about creating an unnecessary map in the helper function that isn't returned to the caller, it can obviously be optimized to just find the correct value without creating any map.

deivid-rodriguez:
I don't think we should create a new method that maps platforms to the latest specs per platform, because nobody has requested it, and I don't see how it can be useful for any one.

I do care about backwards compatibility, so if we deem fixing the current behaviour of `Gem.latest_specs` risky for a bugfix release, then I would add a separate helper that does the correct thing and leave `Gem.latest_specs` untouched for now.

rgoldberg:
The function should accept a platform name.

It should return the newest version for the specified platform, as outlined later.

You could create a no-arg helper function that calls the main function with the name of the current running platform

If versions exist for the current platform, and if the newest platform-specific version is at least as new as the newest non-platform-specific version or if there are no non-platform-specific versions, then return the newest platform-specific version.

If versions exist for the current platform, and if the newest platform-specific version is older than the newest non-platform-specific version, then return the newest non-platform-specific version.

If no versions exist for the current platform, but non-platform-specific versions exist, return the newest non-platform-specific version.

If no versions exist for the current platform, and if there are no non-platform-specific versions, then return null, or otherwise indicate the absence of a version for this platform.

What is the "platform name" of the non-platform-specific version?  Would it be an empty string?  Would it be "RUBY"?

deivid-rodriguez:
@rgoldberg What you describe is exactly what this PR implements I believe :rofl:.

deivid-rodriguez:
> What is the "platform name" of the non-platform-specific version? Would it be an empty string? Would it be "RUBY"?

For non-platform specific versions, we normally omit the platform, so empty string yeah.

rgoldberg:
It seems odd to create a hash to just return a single string.

How does the function know the platform for which to return the latest version?

When `_latest_specs` is called, what is the value of `Gem::Specification._all` passed in as the value of `specs`?  i.e. what type of object is it, and what would it contain for a simple test with the following platform to version data?

arm: 3
x86: 5
not-platform-specific: 4

I don't code in Ruby, so I'm not about to setup a whole Ruby environment & learn Ruby to figure out all this.

deivid-rodriguez:
> It seems odd to create a hash to just return a single string.

It returns an array where each element is a gem version installed on your system, the latest of each gem.

> How does the function know the platform for which to return the latest version?

Whatever the platform of the latest version in each case is. In case there's a specific and non-specific variants of the same version number installed, the platform specific one is returned.

In the case you present, it would return "x86: 5" because it's the highest version in the system. That would normally match your current platform in the great majority of the cases unless you've hit a bug that installs an incorrect variant for your platform (like https://github.com/sparklemotion/nokogiri/issues/2165). Still seems correct and useful behaviour in that case to me.

rgoldberg:
> Whatever the platform of the latest version in each case is. In case there's a specific and non-specific variants of the same version number installed, the platform specific one is returned.

What you're saying isn't the algorithm that I wrote.

e.g., for the following data:

arm: 3
x86: 5
not-platform-specific: 4

My algorithm would return return 4 for arm, whereas an algorithm following your statement would return 3 for arm.

If a newer version that isn't platform-specific has been released, I want that newer version.  e.g., maybe the gem switched from platform-specific code to code that works on all platforms.

Can you give me an examples of a value for `Gem::Specification._all`, and the return value for that argument?  In something like JSON notation for arrays & hashes / objects?

deivid-rodriguez:
I would want `x86: 5` returned, which would be the correct answer for "what's the latest version installed on my system" and would also inform me of a problem with my installation, which includes an "alien gem" from a platform not maching mine.

deivid-rodriguez:
Just to clarify, this method does nothing "remote", it only checks what's installed on your system and gives you the highest version of each gem. Well, that's what it does _with my changes_, before it would do something different, in the case you posted it would return all 3.

rgoldberg:
@deivid-rodriguez I misread your comment before my previous comment, so I rewrote my previous comment (a few times).  Sorry for any confusion.  Can you look at my rewritten previous comment, and answer its current questions?

deivid-rodriguez:
I'm sorry @rgoldberg but I'm not sure where you're trying to get at, and what's the problem that you're trying to solve. I don't think we need to return anything extra, this is enough to resolve the issue you have raised: `gem outdated` incorrectly telling you that a gem is out of date.

Also, and I don't mean to be rude, I totally respect that you don't want to learn ruby or setup a development environment. In that case, please leave only "final user feedback" about what your expectations are _as a user_ when using rubygems, but don't use me as a REPL to replace your missing development environment.

deivid-rodriguez:
We haven't checked usage in the wild of `Gem.latest_specs` (I haven't done it because I don't currently have 500Gb available to setup https://github.com/akr/gem-codesearch). But honestly, this seems like a clear bug fix to me, so I would move it forward as it is.

simi (approved):
I think I have seen using `Gem.latest_specs` somewhere in the wild before. But I can't remember where and I can't find it :/

deivid-rodriguez:
My point is that even if someone is using it, the previous result doesn't make sense to me, so the way I see it we would be just fixing a bug for those people using it.

PR #4250, opened 2021-01-06
Adds additional uniqueness to `Requirement` hashes

lukaso:

`Gem::Requirement#hash` currently returns the same hash for "\~> 2.0" and "\~> 2.0.0" even though these are not equal according to `Gem::Requirement#==`.

This fix returns a unique hash that is similar equality guarantees as `Gem::Requirement#==`,
though it also provides different hashes when an exact match is
requested, whereas `==` only differentiates them when there is a tilde
requirement.

`Gem::Requirement#==` gives:

```
  assert_requirement_equal "= 1.0", "= 1.0.0"
```

Whereas `Gem::Requirement#hash` gives:

```
  refute_requirement_hash_equal "= 1.0", "= 1.0.0"
```

I am not at all sure if the functioning of `Gem::Requirement#==` should not be changed as gem bumping depends on the full number of points even when they are 0, much like in the case of a tilde requirement.

### DIFF
--- a/lib/rubygems/requirement.rb
+++ b/lib/rubygems/requirement.rb
@@ -190,7 +190,7 @@ class Gem::Requirement
   end
 
   def hash # :nodoc:
-    requirements.sort.hash
+    requirements.map {|r| r.first == "~>" || r.first == "=" ? [r[0], r[1].to_s] : r }.sort.hash
   end
 
   def marshal_dump # :nodoc:
--- a/test/rubygems/test_gem_requirement.rb
+++ b/test/rubygems/test_gem_requirement.rb
@@ -401,6 +401,27 @@ class TestGemRequirement < Gem::TestCase
     assert_equal r1.hash, r2.hash
   end
 
+  def test_hash_equality_same_as_equals
+    refute_requirement_hash_equal "= 1.2", "= 1.3"
+    refute_requirement_hash_equal "= 1.3", "= 1.2"
+
+    refute_requirement_hash_equal "~> 1.3", "~> 1.3.0"
+    refute_requirement_hash_equal "~> 1.3.0", "~> 1.3"
+
+    assert_requirement_hash_equal ["> 2", "~> 1.3", "~> 1.3.1"], ["~> 1.3.1", "~> 1.3", "> 2"]
+
+    assert_requirement_hash_equal ["> 2", "~> 1.3"], ["> 2.0", "~> 1.3"]
+    assert_requirement_hash_equal ["> 2.0", "~> 1.3"], ["> 2", "~> 1.3"]
+
+    refute_requirement_hash_equal "= 1.0", "= 1.0.0"
+    refute_requirement_hash_equal "= 1.1", "= 1.1.0"
+    refute_requirement_hash_equal "= 1", "= 1.0.0"
+
+    refute_requirement_hash_equal "1.0", "1.0.0"
+    refute_requirement_hash_equal "1.1", "1.1.0"
+    refute_requirement_hash_equal "1", "1.0.0"
+  end
+
   # Assert that two requirements are equal. Handles Gem::Requirements,
   # strings, arrays, numbers, and versions.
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez (commented):
Thanks @lukaso!

I couldn't understand why special casing `=` here, I believe the only special case is `~>`?

lukaso:
> Thanks @lukaso!
> 
> I couldn't understand why special casing `=` here, I believe the only special case is `~>`?

It depends very much on the purpose of the PR. You suggested I raise it based on being able to distinguish "2.0.0" from "2.0". If the only case where we want to do this is the same as `==` then yes, that special case should be removed. If the purpose is to provide unique hashes for dependency requirements as discussed from [here](https://github.com/rubygems/rubygems/pull/4216#discussion_r552235295) then we need the `=` case to be special cased.

Either way it doesn't solve the problem discussed above due to bundler using old rubygems versions as well as new ones.

deivid-rodriguez:
I think I might be misunderstanding something but my expectation is that for two requirements like `= 2.0` and `= 2.0.0`, you get the same `DepProxy` instance, but for `~> 2.0` and `~> 2.0.0` you get two different instances.

So what you are saying is that unless we special case `=` here, we can't get rid of `DepProxy.setup!` in #4216, correct? In that case, I think the problem might be that we _do_ need to reset the dep proxy cache after each resolver spec, since each spec might setup different (and potentially conflicting) requirements that can pollute the `DepProxy` of other specs.

deivid-rodriguez:
Hei @lukaso! So I had a closer look, and yeah, the problem is specs polluting the `DepProxy` of other specs.

For example, I'm running the following without 7c2e86c124803324419f78965e713acca0680bbc:

```
Run options:
  include {:ids=>{"./spec/resolver/basic_spec.rb"=>["1:21:6", "1:21:12:1"]}}
  exclude {:jruby=>true, :readline=>false, :permissions=>false, :no_color_tty=>false, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.30.0", :realworld=>true, :sudo=>true}

Randomized with seed 58440

Resolving
  conservative
    resolves foo only to latest patch strict
    revert to previous
      could revert to a previous version level patch (FAILED - 1)


Retried examples: 0

Failures:

  1) Resolving conservative revert to previous could revert to a previous version level patch
     Failure/Error: expect(got).to include(s)
       expected ["bar-2.2.3", "foo-1.4.4"] to include "bar-2.1.1"
     # ./spec/support/indexes.rb:46:in `block in should_resolve_and_include'
     # ./spec/support/indexes.rb:45:in `each'
     # ./spec/support/indexes.rb:45:in `should_resolve_and_include'
     # ./spec/support/indexes.rb:75:in `should_conservative_resolve_and_include'
     # ./spec/resolver/basic_spec.rb:288:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:104:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:104:in `block (3 levels) in <top (required)>'
     # ./spec/support/helpers.rb:352:in `block in with_gem_path_as'
     # ./spec/support/helpers.rb:366:in `without_env_side_effects'
     # ./spec/support/helpers.rb:348:in `with_gem_path_as'
     # ./spec/spec_helper.rb:103:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:75:in `block (2 levels) in <top (required)>'
     # ./spec/support/rubygems_ext.rb:110:in `load'
     # ./spec/support/rubygems_ext.rb:110:in `gem_load_and_activate'

Finished in 0.79698 seconds (files took 0.13155 seconds to load)
2 examples, 1 failure

Failed examples:

rspec ./spec/resolver/basic_spec.rb:287 # Resolving conservative revert to previous could revert to a previous version level patch

Randomized with seed 58440
```

The problem is that the first spec sets up the dependency `bar ~>  2.1` for its own resolution. Then the second spec tries to setup the `bar ~> 2.1.0` dependency on foo, however due to this bug the previously cached dependency is setup instead.

If I apply this patch but changing it to special case "~>" only, the the spec passes.

deivid-rodriguez:
We'll introduce this fix as part of #4216 in the end, so closing here. Thanks so much for all the work @lukaso!

PR #4252, opened 2021-01-07
Workaround for another jruby crash when autoloading a constant

deivid-rodriguez:

See https://github.com/rubygems/rubygems/runs/1662234887.

JRuby has issue with autoload that come up once in a while in our test suite. Normally they happen when a file is required explicitly and the constant it provides is also autoloaded. We normally workaround those issues by removing either the require or the autoload, and they never reappear.

However in this case I can't see the duplicated require/autoload, so since the constant is only used in two places I'm replacing the `autoload` with two inline requires.


simi (approved):

deivid-rodriguez:
Thanks @simi!

PR #4255, opened 2021-01-07
Rerecord VCR cassettes

deivid-rodriguez:
I like to do this once in a while to make sure the specs test a more up to date realworld.


simi:
In theory we should do this after each change on API in rubygems.org and index.rubygems.org. Are there any other endpoints used?

deivid-rodriguez:
Just the dependency endpoints are used. Still, even if no API changes, I think it's good to do it once in a while. As the space of released gems grows, the resolution problem usually becomes harder, so it doesn't hurt to check how problematic Gemfiles behave in the current realworld.

simi:
@deivid-rodriguez sure, thanks for the info.

simi (approved):

deivid-rodriguez:
Something quite interesting happened after merging #4081.

Now the response files generated by VCR are plain text, I think because now `net/http` takes care of decompressing the content, correct @eregon? Seems nice to me.

I'll make a followup PR to catch up with this change.

eregon:
That makes sense as I think VCR monkey-patches Net::HTTP outer methods and so the whole decompression is all internal to Net::HTTP.

deivid-rodriguez:
Yeah, I was thinking the same. We don't use the real VCR but a custom implementation of the same thing, but we monkeypatch `net/http` in a similar way I believe.

PR #4271, opened 2021-01-11
Restore accepting custom make command with extra options as the `make` env variable

terceiro:

Debian's gem2deb passes a set of build variables to Rubygems extension builders using ENV['make']. These options are used to produce hardened binaries, as well as to help with reproducible builds.

After ad632bb0800a07433b8d9a9dd8269dd3bcb2f068, it no longer works.

Consider this case and shellsplit it before passing it to process spawning helpers.
 
### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -28,13 +28,14 @@ class Gem::Ext::Builder
     unless make_program
       make_program = (/mswin/ =~ RUBY_PLATFORM) ? 'nmake' : 'make'
     end
+    make_program = Shellwords.split(make_program)
 
     destdir = 'DESTDIR=%s' % ENV['DESTDIR']
 
     ['clean', '', 'install'].each do |target|
       # Pass DESTDIR via command line to override what's in MAKEFLAGS
       cmd = [
-        make_program,
+        *make_program,
         destdir,
         target,
       ].reject(&:empty?)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Thanks @terceiro, looks great! Sorry for the regression.

deivid-rodriguez:
@terceiro I planning to get this fix out in about a week.

deivid-rodriguez:
@terceiro I was thinking about this, and while I'm happy to fix this regression and get you back on track,I wonder if this is the right thing to do. I would've never expected people to pass command line flags as part of `ENV["make"]`, hence the regression :sweat_smile:. I was only expecting people to configure the specific path or make variant in there.

Is this something you do often, and that other libraries usually support? Would it be better that we respected a separate environment variable for this, such as,  `MAKEFLAGS` or any other variable commonly used/documented for this?

terceiro:
@deivid-rodriguez in my experience most projects that support overriding commands with environment variables like this will usually handle a command with extra arguments, but I don't really have data to prove it.

Maybe we could use MAKEFLAGS as well, in this case I think rubygems would not need to do anything.

deivid-rodriguez:
Yeah, I think you're right. I just remembered how `ruby-core` customizes the `gem` CLI command for its environment: https://github.com/ruby/ruby/blob/0036648a420f945624898568bb82bc5f83195d12/tool/runruby.rb#L115.

Thanks for the feedback, your fix is good :+1:.

PR #4287, opened 2021-01-15
Improve realworld specs and regenerate cassettes

deivid-rodriguez:

Our realworld specs are hard to debug.

This PR includes several improvements:

* When specs testing gemfile performance fail, they now show a better message.
* On failures, command output is displayed _before_ failure message. Failure message is usually more helpful so we don't want long commands output to hide it.
* Regenerate cassettes. One Gemfile now performs worse with updated specs. I will fix that in a separate PR.
* Remove `:sometimes` filter. I don't think these specs fail intermittently, but if they do, we should fix that instead of hiding errors by retrying.


bundler/spec/realworld/edgecases_spec.rb
```diff
@@ -451,9 +451,9 @@ def rubygems_version(name, requirement)
     bundle :lock, :env => { "DEBUG_RESOLVER" => "1" }
 
     if Bundler.feature_flag.bundler_3_mode?
-      expect(out).to include("BUNDLER: Finished resolution (1336 steps)")
+      expect(out).to display_total_steps_of(6488)
     else
-      expect(out).to include("BUNDLER: Finished resolution (1395 steps)")
+      expect(out).to display_total_steps_of(6827)
```
deivid-rodriguez, line 456: This spec is now much slower, but #4289 fixes that!

Issue #4275, opened 2021-01-12
Bundler 2.2.3 introduced a significant performance regression for our bundle, not fixed in 2.2.5


haines:
### Describe the problem as clearly as you can

Our `bundle install` (113 Gemfile dependencies, 253 gems) is very slow on recent versions of Bundler.

| Bundler version | Install time |
|---|---|
| 2.1.4 | 0.69s |
| 2.2.0 | 1.08s |
| 2.2.1 | 1.15s |
| 2.2.2 | 1.05s |
| 2.2.3 | 33.61s |
| 2.2.4 | 32.88s |
| 2.2.5 | 32.07s |

### Post steps to reproduce the problem

I don't want to disclose our full dependencies too widely but can add maintainers to a private repo on an individual basis.

### Which command did you run?

`bundle install`

### What were you expecting to happen?

Gems to install in a reasonable time.

### What actually happened?

Gems took a very long time to install.

#### Environment

```
Bundler       2.2.5
  Platforms   ruby, x86_64-darwin-19
Ruby          2.7.2p137 (2020-10-01 revision 5445e0435260b449decf2ac16f9d09bae3cafe72) [x86_64-darwin19]
  Full Path   /Users/andrew/.rubies/ruby-2.7.2/bin/ruby
  Config Dir  /Users/andrew/.rubies/ruby-2.7.2/etc
RubyGems      3.2.5
  Gem Home    /Users/andrew/.gem/ruby/2.7.2
  Gem Path    /Users/andrew/.gem/ruby/2.7.2:/Users/andrew/.rubies/ruby-2.7.2/lib/ruby/gems/2.7.0
  User Home   /Users/andrew
  User Path   /Users/andrew/.gem/ruby/2.7.0
  Bin Dir     /Users/andrew/.gem/ruby/2.7.2/bin
Tools         
  Git         2.30.0
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

#### Bundler Build Metadata

```
Built At          2021-01-11
Git SHA           ba867aed7f
Released Version  true
```

deivid-rodriguez:
Thanks for the report @haines, could you add me to your private repo I guess?

haines:
Done! Thanks for looking into it @deivid-rodriguez ❤️

deivid-rodriguez:
[...]

PR #4277, opened 2021-01-12
Fix resolver edge cases and speed up bundler

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

It all started with https://github.com/rubygems/rubygems/issues/4275. Recent versions of bundler have caused a heavy `bundle install` performance regression, even in cases when there's a lockfile and all gems are already installed :grimacing:.

## What is your fix for the problem, implemented in this PR?

After an initial investigation of the problem, I noticed that reverting https://github.com/rubygems/bundler/pull/6495 fixed the issue. That gave me the idea that this PR fixed a resolution problem, but at an incorrect place (note how the platform filtering happens on top of the resolve (independently of whether resolution actually happened, or it was read from the lockfile).

So, the challenge was to fix that same problem inside the resolver, so that no overhead is needed in any case. The result of doing this (other than fixing this regression) is speeding up all `bundler/setup` requires. In the case of a generated default Rails application, the speed up is ~5-10% of boot time.

My first solution worked, but created an overhead on most resolutions, that would be a bit slower. The reason is that spec groups need to be modified dynamically depending on the currently activated platforms, so that whole `search_for` result for each dependency can no longer be fully memoized.

So I did a series on refactorings on the resolution process to not hurt performance. To measure results, I compiled a series of Gemfiles that have proven to be problematic in the past, and run the resolver before an after this PR. The results on my machine are as follows:

#### Before this PR

```
+-------------------------------+---------+---------------------+
|  Gemfile                      |  Steps  |  Seconds            |
+-------------------------------+---------+---------------------+
|  depfu                        |  242    |  0.985466332        |
|  solidus_shipwire             |  148    |  1.175150116        |
|  active_model_as_json_filter  |  948    |  1.262431201        |
|  simply_bussines              |  2061   |  1.623000041        |
|  tricky_groups                |  889    |  3.18565573         |
|  network_executive            |  2179   |  3.354704401        |
|  nix                          |  1087   |  4.492709607        |
|  fury-bug                     |  2471   |  4.693998945        |
|  doorkeeper-couchbase         |  8880   |  4.740346119        |
|  big                          |  1005   |  5.792436901        |
|  spree_pag_seguro             |  2192   |  8.232739329        |
|  refinerycms-photo-gallery    |  4814   |  9.35598295         |
|  activeadmin-redactor         |  4191   |  19.830266492       |
|  staple                       |  7330   |  22.273399876       |
|  mongoid_session_store_json   |  14497  |  29.598622289       |
[...]

PR #4289, opened 2021-01-15
Fix resolver dependency comparison

deivid-rodriguez:
This PR fixes the last "infinite resolution Gemfile" we know about :tada:

The Gemfile in https://github.com/rubygems/rubygems/issues/3788 doesn't resolve in reasonable time. I tried it against current master and aborted it after ~30minutes & 500K steps.

Our underlying resolver, Molinillo, is prepared to group spec groups that all have the same dependencies to simplify resolution. However, our dependency proxies were not properly comparable between each other and these groups were not being properly built, causing candidates with the same dependencies to be tried independently.

This was actually the cause of a very slow Gemfile, and I had added a fix that prevented fallback groups from being added if the dependencies of the ruby variants were the same as the platform specific variants.  Although that fix worked for that case, it's not a proper fix in general as proved by the "infinite resolution" that's being fixed here.

The proper fix is to ignore platforms when comparing dependency proxy objects. Platforms are only useful _outside_ of Molinillo for searching valid candidate specifications for a dependency according to its platform, but inside Molinillo, two `DepProxy` objects with the same dependencies should be considered the same and grouped together even if they have different `__platform`.

Fixes https://github.com/rubygems/rubygems/issues/3788.


deivid-rodriguez:
We found a simple Gemfile where these changes make performance much worse, so I'm blocking this until I can investigate why.

deivid-rodriguez:
Ok, I found a solution that fixes the `Gemfile` for which the initial approach had regressed. The current set of changes consistently improve resolution times accross the board except for one unlucky case where the greater number of spec_groups changes the sorting priority and can result in a higher number of steps. This type of cases still resolve fine and quick though.

In addition to fixing #3788 and other previously "irresolvable" Gemfiles, this PR also closes #4275, since we no longer need to do  an extra platform filtering pass after resolving in `Definition#resolve`, resulting in a 8x improvement in that case.

As per @lukaso's suggestion, tomorrow I'll try to restore the previous `DepProxy#==` semantics by overriding the logic in `Molinillo` where the comparison needs to ignore platforms. I think the result might also be simpler and might not need mutation of `DepProxy` objects.

deivid-rodriguez:
@lukaso I restored the previous more sane `DepProxy` equal, let me know what you think!

deivid-rodriguez:
Also note that this PR not only fixes and speeds up the resolver, but also provides an improvement in `bundler` boot time in general. :tada:.

For example, on the default `rails new` application on my system, `ruby -rbundler/setup -e1` takes ~20ms less than on 2.2.6. Not much, but hey, it's something.

deivid-rodriguez:
Finding out more stuff here. It turns out that after restoring the previous comparison and only changing the equality logic inside Molinillo where specs with the same dependencies are grouped together, resolving works just as well as with the rest of the fixes, but faster. So I'm planning to introduce the dependency comparison fix first, and move the rest of the changes to #4277, focusing on optimizing them to keep the performance this PR will achieve.

deivid-rodriguez:
I proposed the Molinillo change upstream :crossed_fingers:.

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -224,7 +224,7 @@ def parse_spec(line)
 
         # Avoid introducing multiple copies of the same spec (caused by
         # duplicate GIT sections)
-        @specs[@current_spec.identifier] ||= @current_spec
+        @specs[@current_spec.hash] ||= @current_spec
```
lukaso, line 227: Wouldn't this hash cause the same issues we saw with the `DepProxy` hash? Should this maybe use the constituents of the hash? (don't know how much this code is called, and I'm guessing the previous implementation has worked relatively well to date, but the risk remains).

deivid-rodriguez: Hei @lukaso, thanks for your comment.

I'm really not sure. Thinking about this issue more, my understanding is that as long as you rely on `#hash` values being generated by ruby, then it is irrelevant whether you use `#hash` or the raw object instance as the key to a hash, since ruby will always use its internally generated hash vale that it's guaranteed to be unique. If you implement your own `#hash` method though, you need to make sure that the result cannot have collissions. However, [the docs on user-defined #hash values](https://ruby-doc.org/core-3.0.0/Hash.html#class-Hash-label-User-Defined+Hash+Keys) make it seem like the implementation I'm using here (XOR'ing the `#hash` value of the different data components of the object) is equally safe, so I tend to say this is fine.

That said, I'm fine with reverting this change and restoring the `Identifier` class too.

In any case, my plan is to move these changes to #4277 and leave on the dep proxy comparison fix in this PR. I'll do this on Monday.

lukaso: From what I can tell, your implementation is exactly as good  as the `Identifier` class one you replaced in terms of `hash`. That was also using hash. So no need to revert. Both implementations have the same potential flaw. I think you'd have to use an array of the actual values as the key to `@specs` in order to avoid the issue, exactly as we are doing on `DepProxy`. And then there's the question of how often this code is called (could it cause a performance issue?).

deivid-rodriguez: > From what I can tell, your implementation is exactly as good as the Identifier class one you replaced in terms of hash. That was also using hash. 

Was it? I don't see a custom `#hash` in the class.

Anyways, now I understand that your comment is the exact same concern raised in https://github.com/rubygems/rubygems/pull/4216#discussion_r553545164. Just like in that discussion, I don't believe there's an issue here, but I will change the code as per your request. It's going to look a bit weird though. It'd be

```suggestion
        @specs[[@current_spec.name, @current_spec.version, @current_spec.platform_string] ||= @current_spec
```

Correct?

Regarding the performance, I don't understand how this could cause a performance issue, even if it was a hot path. I believe it's not since this is the lockfile parser, and lockfile parsing should happen only once, but still.

lukaso: Yes, correct, that was the concern (not raised by me, but noticed same behaviour here).

deivid-rodriguez: Great, I will update this as we did there. For now I removed all these changes from this PR to focus on the dependency requirement fix. I'll move the rest of the improvements to #4277 after merging this one (I'll merge it after CI).

deivid-rodriguez: Just to complete my further thoughts on this issue, it seems that ruby properly resolves collissions as long as you provide a proper `eql?` definition. So, in the example given at https://ruby-doc.org/core-3.0.0/Hash.html#class-Hash-label-User-Defined+Hash+Keys:

```ruby
class Book
  attr_reader :author, :title

  def initialize(author, title)
    @author = author
    @title = title
  end

  def ==(other)
    self.class === other &&
      other.author == @author &&
      other.title == @title
  end

  alias eql? ==

  def hash
    @author.hash ^ @title.hash # XOR
  end
end

book1 = Book.new 'matz', 'Ruby in a Nutshell'
book2 = Book.new 'matz', 'Ruby in a Nutshell'

reviews = {}

reviews[book1] = 'Great reference!'
reviews[book2] = 'Nice and compact!'

reviews.length #=> 1
```

You can check forcing a `#hash` collision and still getting the expected behaviour:

```ruby
book3 =  Book.new 'Ruby in a Nutshell', 'matz'

book3.hash == book2.hash # => true

reviews[book3] = "Different"

reviews.lenght # => 2
```

So, again, no issue here.

deivid-rodriguez: @lukaso I moved all these changes to #4277. This should be fixed there!

PR #4286, opened 2021-01-15
Swallow any system call error in `ensure_gem_subdirs` to support jruby embed paths

kares:
JRuby expects the method to not raise with embed paths (and is blocking a 3.2 update).
This has changed in RGs 3.2 to only handle `EACCESS`: https://github.com/rubygems/rubygems/commit/0c85553cd9be78ae28201a1464b226589e009539

NOTE: wasn't sure whether to target master or 3.2 branch.

PR #4291, opened 2021-01-15
Swallow any system call error in `ensure_gem_subdirs`

kares:
https://github.com/rubygems/rubygems/pull/4286 (with added asserts)

JRuby expects the method to not raise with embed paths (and is blocking a 3.2 update).
This has changed in RGs 3.2 to only handle EACCESS: 0c85553

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -469,7 +469,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
       next if File.exist? subdir
       begin
         FileUtils.mkdir_p subdir, **options
-      rescue Errno::EACCES, Errno::ENOTDIR
+      rescue SystemCallError
       end
     end
   ensure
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@kares Thanks!

I know it's more work but can you cherry-pick the untouched commit from the stable branch for now, and then propose other enhancements to master on a separate PR? Just so that this plays nice with our release scripts and I don't have to fix conflicts manually when releasing.

simi (approved):
:information_source: @deivid-rodriguez I'm not sure how to merge this to not mess with release scripts (I assume "rebase and merge" would be the one).

deivid-rodriguez:
I think standard merge should be ok as long as we don't tag the PR for backport (we have used only that one so far since we started merging PRs manually instead of using a bot). Technically even if we tag this PR for backport, release scripts should detect that it already backported and ignore, so no big deal either, although I'm not sure the detection would work in this case.

My concern with merging a different PR here and on the stable branch was to miss backporting things if this PR had extra stuff that could later cause conflicts.

So to sump up, this should be ready!

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks @kares!

PR #4297, opened 2021-01-18
Support gemified `set`

deivid-rodriguez:

The problem is that users can't currently specify `set` in their Gemfiles with the version they want to use, because `bundler` uses `set` and activates the default version of set unconditionally.

To be honest though, the reason that made me work on this was not that one, but that I found that some usages of `Set` are actually much slower than using plan arrays. In particular, I was making some improvements to the resolver, and found that due to the nature of my changes (making inclusion checks on the `@platforms` array quite frequently), they were slowing down the resolver quite a lot (from 9 seconds to 12 seconds on a Gemfile proved problematic in the past). When switching to plain arrays, there was no performance penalty.

My fix is to replace all usages of `Set` preventing it to be supported as a default gem, fixing both of the above issues.

This requires changes in both `thor` and `Molinillo` which I will propose upstream.

Fixes https://github.com/rubygems/rubygems/issues/3912.


deivid-rodriguez:
I submitted upstream PRs:

* https://github.com/erikhuda/thor/pull/746.
* https://github.com/CocoaPods/Molinillo/pull/147.

deivid-rodriguez:
Thor PR was merged, Molinillo's one is waiting on feedback from Samuel. I'll update the comments about the vendored libraries and their differences with their latest released versions later.

deivid-rodriguez:
Thor 1.1.0 has been released :tada:, I'll update this PR to use that.

deivid-rodriguez:
I updated the PR to use thor 1.1.0. I wouldn't rely on the Molinillo PR to be attended soon, so this should be ready.

deivid-rodriguez:
Got some feedback in the Molinillo PR so I updated the PR in Molinillo and this PR to include the modified version.

deivid-rodriguez:
Pull request to Molinillo was merged, and now this PR reflects that.

bundler/lib/bundler/index.rb
```diff
@@ -65,11 +63,14 @@ def search(query, base = nil)
     def unsorted_search(query, base)
       results = local_search(query, base)
 
-      seen = results.map(&:full_name).to_set unless @sources.empty?
+      seen = results.map(&:full_name).uniq unless @sources.empty?
 
       @sources.each do |source|
         source.unsorted_search(query, base).each do |spec|
-          results << spec if seen.add?(spec.full_name)
+          next if seen.include?(spec.full_name)
```
segiddins, line 70: is this a potential performance regression? would we be better off using a dictionary here to keep amortized constant-time inclusion checks?

deivid-rodriguez: I'm not sure, I'll have a closer look.

The main motivation of this PR was that I found that in some cases using `Set` performs much worse than using plain Arrays for some reason (see the PR description), so I took the chance to remove `set` usages to fix that performance regression and at the same time support `set` as a gem.

hsbt (approved):
I'm +1 for the changes under the `lib/*` and `bundler/lib/*`.

deivid-rodriguez:
Thanks @hsbt! I will review @segiddins comment, and try to get this ready soon.

bundler/lib/bundler/index.rb
deivid-rodriguez: @segiddins I run a few tests to check performance and I didn't see any noticiable difference, I think it's because this is tipically a very small array.

I'd say let's go with what we have to allow `set` inside `Gemfile`. If this causes issues, they can be addressed later :+1:.

deivid-rodriguez:
Let's do this a see how it goes!

KaycoinsInc (approved):

KaycoinsInc (approved):

PR #4303, opened 2021-01-18
Respect user configured default branch in README links in new generated gems

mostafahussein:
This PR will replace `master` with `main` as a default branch name at the README.md file which will be generated when creating a new gem.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -176,6 +176,11 @@ module Bundler
         )
       end
 
+      if use_git
+        Bundler.ui.info "Initializing git repo in #{target}"
+        `git init #{target}`
+      end
+
       templates.each do |src, dst|
         destination = target.join(dst)
         SharedHelpers.filesystem_access(destination) do
@@ -191,8 +196,6 @@ module Bundler
       end
 
       if use_git
-        Bundler.ui.info "Initializing git repo in #{target}"
-        `git init #{target}`
         Dir.chdir(target) do
           `git add .`
         end
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -184,6 +184,8 @@ module Bundler
       if use_git
         Bundler.ui.info "Initializing git repo in #{target}"
         `git init #{target}`
+
+        config[:git_default_branch] = File.read("#{target}/.git/HEAD").split("/").last.chomp
       end
 
       templates.each do |src, dst|
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -39,11 +39,11 @@ module Bundler
       constant_name = name.gsub(/-[_-]*(?![_-]|$)/) { "::" }.gsub(/([_-]+|(::)|^)(.|$)/) { $2.to_s + $3.upcase }
       constant_array = constant_name.split("::")
 
-      git_installed = Bundler.git_present?
+      use_git = Bundler.git_present? && options[:git]
 
-      git_author_name = git_installed ? `git config user.name`.chomp : ""
-      github_username = git_installed ? `git config github.user`.chomp : ""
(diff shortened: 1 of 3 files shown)


simi:
Hello @mostafahussein.

We will probably need to change a default branch generated as well to get those information in sync. I did quick test and GitHub (IMHO still home of most Ruby opensource code) creates "main" default repository for newly created repos. I have found out there is setting at `https://github.com/settings/repositories` which defaults to "main" (if I remember well, on my account default branch was called "master" and I have not updated this manually).

I also tried to use `bundle gem` command to bootstrap new gem, but it is created with "master" branch.

I'm afraid we would probably need new CLI switch to specify default branch and probably get in sync with GitHub to suggest "main" as default to properly handle all cases. During git init, we should respect that by using `git init --initial-branch=branch-name`.

On the other side, we can also try to get user default for git default branch by reading `init.defaultBranch` git global config and suggest that one if user has already this set.

Also in README.md this is related only to CoC link. Maybe it would be enough to remove the link part to not need to handle the url construction.

mostafahussein:
Hello @simi ,
Thank you for your feedback, I have pushed a new code based on it, can you please review it and let me what should be modified or updated?

deivid-rodriguez:
I feel this is something people either adopt for all projects, or for none. So maybe respecting `init.defaultBranch` if it exists and otherwise using `main` would be enough?

mostafahussein:
I have updated the code to either use `initialBranch` or `main` as default but I have a concern regarding the usage of `--initial-branch` as this was [introduced](https://github.blog/2020-07-27-highlights-from-git-2-28/#introducing-init-defaultbranch) since `Git 2.28`. So it might be a breaking change for other users, Is there anything that should be done regarding this?

deivid-rodriguez:
I didn't mean to overrule @simi's suggestion, just thought it was worth discussing it. What do you think @simi?

Regarding the backwards compatibility concern, I think maybe the easiest way is to not pass `--initial-branch` at all. Instead, let `git init` do what it is configured to do, and then _read_ the branch that was generated and use that in the README?

mostafahussein:
I also thought that the use of the default branch config or `main` only would be better  because if we let the user type the value then we need to add some sort of regex validation. If such complexity isn't necessary then the current implementation would be enough otherwise we can add a question and let the users decide for themselves and validate the answer with regex

hsbt:
The current implementation of rubygems didn't enforce to use `master` because we relied on git default rule. It's out of scope of rubygems.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -180,7 +182,7 @@ def run
       if Bundler.git_present? && options[:git]
         Bundler.ui.info "Initializing git repo in #{target}"
         Dir.chdir(target) do
-          `git init`
+          `git init --initial-branch=#{git_default_branch}`
```
deivid-rodriguez, line 185: My suggestion is that you don't pass `--initial-branch` at all. Instead, add `git rev-parse --abbrev-ref HEAD` after `git init` has run to check the name of the created branch, and fill in `git_default_branch` with that.

That we we don't have any backwards compatibility issues and we still respect user's choice.

deivid-rodriguez:
> If we didn't pass --initial-branch then git will follow the default strategy before 2.80 which is to create a master branch as I guess not all users consider using the init.defaultBranch option

I'm sorry I missed this edit. Newest `git` versions show a pretty annoying warning if you don't explicitly configure `init.defaultBranch`, so I think not passing `--initial-branch` is good enough for now. Let's remove the last hardcoded "master" from our code base, and we can reevaluate further changes in the future.

deivid-rodriguez:
I went ahead and implemented the suggestions made, since we got no further feedback but I wanted to stop hardcoding "master" in generated gems.

Turned out to be much trickier than I expected, because the templates were being rendered _before_ the git repository was initialized, so I had to refactor things around a bit, but it should be working now :+1:.

deivid-rodriguez:
Let's do it!

Issue #4307, opened 2021-01-21
Sometimes `rake release` misdetects the current branch


deivid-rodriguez:
Sorry for the radio silence here. I'm still seeing the error with bundler 2.2.6:

```
$ bundle --version
Bundler version 2.2.6
$ bundle exec rake release
dead_end 1.1.3 built to pkg/dead_end-1.1.3.gem.
Tagged v1.1.3.
Untagging v1.1.3 due to error.
rake aborted!
Running `git config --get branch.module_function.remote` failed. Run this command directly for more detailed output.
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:204:in `sh'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:128:in `default_remote'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:118:in `git_push'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:71:in `block (2 levels) in install'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:174:in `tag_version'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/gem_helper.rb:71:in `block in install'
/Users/rschneeman/.gem/ruby/3.0.0/gems/rake-12.3.3/exe/rake:27:in `<top (required)>'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/cli/exec.rb:63:in `load'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/cli/exec.rb:63:in `kernel_load'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/cli/exec.rb:28:in `run'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/cli.rb:494:in `exec'
/Users/rschneeman/.gem/ruby/3.0.0/gems/bundler-2.2.6/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
[...]

PR #4309, opened 2021-01-22
Fix releasing from a not yet pushed branch

deivid-rodriguez:

The problem is that since recent versions, releasing from an unpushed branch no longer works.

My fix is to always give an explicit branch and remote to git push.

There's a bit of yet missing logic to infer the remote. In particular, we should check `branch.<name>.pushRemote` with the most priority, and also `push.defaultRemote` with more priority than "origin".

But I focused on fixing the issue on this PR and will reticket the other stuff.

Fixes https://github.com/rubygems/rubygems/issues/4307.

### DIFF
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -116,19 +116,21 @@ module Bundler
 
     def git_push(remote = nil)
       remote ||= default_remote
-      perform_git_push remote
+      perform_git_push "#{remote} refs/heads/#{current_branch}"
       perform_git_push "#{remote} refs/tags/#{version_tag}"
       Bundler.ui.confirm "Pushed git commits and release tag."
     end
 
     def default_remote
+      remote_for_branch, status = sh_with_status(%W[git config --get branch.#{current_branch}.remote])
+      return "origin" unless status.success?
+
+      remote_for_branch.strip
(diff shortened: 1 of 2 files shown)


schneems:
Works for me

```
$ ruby -I /Users/rschneeman/Documents/projects/rubygems/bundler/lib /Users/rschneeman/Documents/projects/rubygems/bundler/exe/bundle -v
Bundler version 2.3.0.dev
$ ruby -I /Users/rschneeman/Documents/projects/rubygems/bundler/lib /Users/rschneeman/Documents/projects/rubygems/bundler/exe/bundle exec rake release
c9fb36f7a709d9ddba4bf7da866e2a1 0.1.0 built to pkg/c9fb36f7a709d9ddba4bf7da866e2a1-0.1.0.gem.
Tagged v0.1.0.
Pushed git commits and release tag.
Code:   ^CERROR:  Interrupted
rake aborted!
```

(Note I bailed out of actually pushing, the tagging worked). Thanks!

schneems:
Note  to self bundler silently locks to old versions so running this is not enough:

```
gem install bundler
bundle exec rake release
```

You need to 

```
gem install bundler
bundle update --bundler
git add Gemfile.lock
git commit -m "Update bundler version"
git push
bundle exec rake release
```

PR #4310, opened 2021-01-22
Avoid showing platforms with requirements in error messages

deivid-rodriguez:

When conflicts in dependencies happens, bundler shows the platform next to the
requirement. This is not helpful in general and it has confused me a few times,
so I fixed it.

The platform was being printed because Molinillo was ignoring the proc we passed
it to customize how requirements are printed. I fixed Molinillo to respect it.

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -151,7 +151,8 @@ end
 load "task/automatiek.rake"
 
 # We currently ship Molinillo master branch as of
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This was included with https://github.com/rubygems/rubygems/pull/4308, but it's a different enhancement not specific to the bundler dependency, so I extracted to a separate PR.

Molinillo fix has been proposed at https://github.com/CocoaPods/Molinillo/pull/149.

PR #4311, opened 2021-01-22
Replace "iff" with "whether" and "if and only if"

gmcgibbon:
Rubygems part of https://github.com/ruby/ruby/pull/4035, except the vendored code.

iff means if and only if, but readers without that knowledge might
assume this to be a spelling mistake. To me, this seems like
exclusionary language that is unnecessary.

Simply using "if and only if" or "whether" should suffice.


deivid-rodriguez:
This is actually vendored code for us :sweat_smile:. Could you sync with the upstream, please? It's https://github.com/CocoaPods/Molinillo.

hsbt:
Ah, right. I forgot this.

@gmcgibbon Can you also submit this fixes to Molinillo? Thanks.

gmcgibbon:
Yes, will do. Thank you!

Issue #4261, opened 2021-01-08
new plugins support broken when using buildroot


darix:
using the following install commandline

```
install cmdline: ["/usr/bin/gem.ruby3.0", "install", "--verbose", "--local", "--build-root", "/home/abuild/rpmbuild/BUILDROOT/rubygem-yard-0.9.26-0.x86_64", "-f", "--no-document", "/home/abuild/rpmbuild/SOURCES/yard-0.9.26.gem"]
```

leads to

```
abuild@09c1149cc7cf:~/rpmbuild/BUILDROOT/rubygem-yard-0.9.26-0.x86_64> cat ./usr/lib64/ruby/gems/3.0.0/plugins/yard_plugin.rb
require '/home/abuild/rpmbuild/BUILDROOT/rubygem-yard-0.9.26-0.x86_64/usr/lib64/ruby/gems/3.0.0/gems/yard-0.9.26/lib/rubygems_plugin.rb'
```

The generated file should not contain the buildroot.

```
$ gem env version
gem.ruby3.0 env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.2.3
  - RUBY VERSION: 3.0.0 (2020-12-25 patchlevel 0) [x86_64-linux-gnu]
  - INSTALLATION DIRECTORY: /usr/lib64/ruby/gems/3.0.0
  - USER INSTALLATION DIRECTORY: /home/abuild/.local/share/gem/ruby/3.0.0
  - RUBY EXECUTABLE: /usr/bin/ruby.ruby3.0
  - GIT EXECUTABLE: 
  - EXECUTABLE DIRECTORY: /usr/bin
  - SPEC CACHE DIRECTORY: /home/abuild/.local/share/gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - x86_64-linux
  - GEM PATHS:
     - /usr/lib64/ruby/gems/3.0.0
     - /home/abuild/.local/share/gem/ruby/3.0.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => true
     - :bulk_threshold => 1000
     - :benchmark => false
     - :install => "--format-executable --no-user-install"
     - "install" => "--format-executable --no-user-install"
     - :format_executable => true
     - :update => "--format-executable --no-user-install"
     - "update" => "--format-executable --no-user-install"
[...]

PR #4317, opened 2021-01-25
Generate plugin wrappers with relative requires

deivid-rodriguez:

When using the `--build-root` option to `gem install`, generated plugin wrappers include the `--build-root` parameter which is unexpected.

My fix is to use relative links, which shouldn't change behaviour and it fixes the issue.

Fixes https://github.com/rubygems/rubygems/issues/4261.


darix:
This fixes my issue.

simi (approved):

PR #4328, opened 2021-01-29
Porting ruby-core test suite changes

deivid-rodriguez:

Ruby-core team commit changes in our code directly to their repository.

Port those changes.

### DIFF
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -49,7 +49,7 @@ jobs:
       - name: Test RubyGems
         run: |
           ruby tool/sync_default_gems.rb rubygems
-          make test-all TESTS="rubygems -j2"
+          make -j2 -s test-all TESTS="rubygems --no-retry"
         working-directory: ruby/ruby
         if: matrix.target == 'Rubygems'
       - name: Test Bundler
--- a/bundler/spec/bundler/gem_helper_spec.rb
+++ b/bundler/spec/bundler/gem_helper_spec.rb
@@ -98,6 +98,7 @@ RSpec.describe Bundler::GemHelper do
         context "before installation" do
           it "raises an error with appropriate message" do
             task_names.each do |name|
+              skip "Rake::FileTask '#{name}' exists" if File.exist?(name)
               expect { Rake.application[name] }.
                 to raise_error(/^Don't know how to build task '#{name}'/)
             end
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -312,7 +312,6 @@ class Gem::TestCase < Minitest::Test
     ENV['XDG_DATA_HOME'] = nil
     ENV['SOURCE_DATE_EPOCH'] = nil
     ENV['BUNDLER_VERSION'] = nil
-    ENV["TMPDIR"] = @tmp
 
     @current_dir = Dir.pwd
     @fetcher     = nil
@@ -323,13 +322,10 @@ class Gem::TestCase < Minitest::Test
     # capture output
     Gem::DefaultUserInteraction.ui = Gem::MockGemUi.new
 
-    tmpdir = File.realpath Dir.tmpdir
-    tmpdir.tap(&Gem::UNTAINT)
-
-    @tempdir = File.join(tmpdir, "test_rubygems_#{$$}")
+    @tempdir = Dir.mktmpdir("test_rubygems_", @tmp)
     @tempdir.tap(&Gem::UNTAINT)
 
-    FileUtils.mkdir_p @tempdir
+    ENV["TMPDIR"] = @tempdir
 
     @orig_SYSTEM_WIDE_CONFIG_FILE = Gem::ConfigFile::SYSTEM_WIDE_CONFIG_FILE
     Gem::ConfigFile.send :remove_const, :SYSTEM_WIDE_CONFIG_FILE
--- a/test/rubygems/test_gem_util.rb
+++ b/test/rubygems/test_gem_util.rb
@@ -72,10 +72,10 @@ class TestGemUtil < Gem::TestCase
     ]
 
     files_with_absolute_base = Gem::Util.glob_files_in_dir('*.rb', File.join(@tempdir, 'g'))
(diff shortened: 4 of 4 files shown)


hsbt:
@deivid-rodriguez Thanks for this ❤️ 

Did you track the recent [changes](https://github.com/ruby/ruby/commits/master/spec/bundler
) by @znz ? 

It fixed the out-of-place build of ruby-core repository. I'm not sure it fixed by https://github.com/rubygems/rubygems/pull/4327 yet.

deivid-rodriguez:
Yes, I noticed all those changes and decided to remove all the hacks since they are hard to maintain and they don't buy us much. Next time we sync those changes by @znz will be overwritten, but the code will no longer cause any issues for ruby-core since it's gone.

I have  a question for @nobu regarding 79490fa79854770e5bc325145f385b54fc426e7c. I just noticed a leftover temporary directory left at the root of the repo after running tests with this change, is it expected?

nobu:
No, I wanted all temporary directories to be removed, but just one directory is created to write a `gemrc` file  **after** `teardown`, and I don't know how it runs.
Seems no finalizer, `at_exit`, `END {}`.

deivid-rodriguez:
Until we figure it out, can we create this temporary folder inside `tmp/` so that, since `tmp/` is gitignored, the leftovers don't show up as untracked changes?

nobu:
Of course it doesn’t matter for me, as far as these directories get cleared eventually.

nobu:
JFYI, this is the branch I tried once.
https://github.com/nobu/rubygems/tree/temporary-index-directories

deivid-rodriguez:
Thanks @nobu, I'll have a look :+1:.

deivid-rodriguez:
@nobu I pushed a new version that shouldn't leave any test folders around and I think it should work fine for ruby-core too. I'll merge it after CI.

deivid-rodriguez:
I created a PR in the ruby repo to check CI against this branch and it failed, but it passes on my machine locally so I'm not sure what's going on. Any ideas?

simi (approved):

deivid-rodriguez:
I confirmed that this PR is breaking the ruby-core CI. I'll investigate why it broke and why our "ruby-core simulation" CI didn't catch it.

deivid-rodriguez:
I think I get it now. Ruby-core runs rubygems tests in parallel (we should do that too, by the way). The way this PR is setup (where `tmp/` is cleaned up after each test), tests in one thread will mess up the directory layout for other tests in other threads. Will fix.

deivid-rodriguez:
@nobu I think there's something work the the processing of the parallel test workers. With the broken code, `make -j2 -s test-all TESTS=rubygems` still reports a zero exit code.

nobu:
There are so many "No such file or directory" on temporary directories, which seem caused by some races.

deivid-rodriguez:
> There are so many "No such file or directory" on temporary directories, which seem caused by some races.

Yes, I already have a fix for this, I'm just trying to make sure our catches it next time something like this happens, and I found that `make -j2 -s test-all TESTS=rubygems` still returns a zero exit code even when all those errors happen.

deivid-rodriguez:
Oh, I think I understand now. If workers fail, ruby-core retries to run sequentially and that succeeds, so the issue is not detected.

`make -j2 -s test-all TESTS="rubygems --no-retry"` does what I want :+1:.

deivid-rodriguez:
Yes, that did the trick. Our CI should be able to catch a similar issue in the future.

nobu:
A question: in 3e522bac65d8c12b90949c09b5f61466ae682153, 
> since the current directory is changed by our tests, tests can fail.

I can't get this point.
What does "changed" mean?
First I thought the case `Dir.tmpdir` returns a relative path and not accessible after `Dir.chdir`, but the result should have been absolute always.

deivid-rodriguez:
As far as I understand, `Dir.chdir` is not thread safe, and rubygems tests do that at the moment. When I introduced that commit, I was having issues with `Dir.tmpdir` silently not respecting `ENV["TMPDIR"]` (https://bugs.ruby-lang.org/issues/16975) and falling back to the pwd. When that happened the path of generated files would be unpredictable and some of our tests that make assertions  on the number of parts in those paths (for example the test I changed in this PR) would fail.

deivid-rodriguez:
It should be ready now and I also checked that it passes ruby-core CI at https://github.com/ruby/ruby/pull/4143.

simi (approved):

deivid-rodriguez:
Thanks @simi!!

nobu:
Still a garbage is left.

```sh
$ ruby tool/sync_default_gems.rb rubygems
Sync rubygems/rubygems
$ rm -fr tmp
$ make test-all TESTS='rubygems --no-retry' > /dev/null
Skipping Gem::PackageTask tests.  rake not found.
$ ls tmp/
test_rubygems_20210204-76044-xpvcqw/
$ find tmp -type f -print -exec cat {} \;
tmp/test_rubygems_20210204-76044-xpvcqw/userhome/.config/gem/gemrc
---
:backtrace: false
:bulk_threshold: 1000
:sources: []
:update_sources: true
:verbose: true
:concurrent_downloads: 8
http_proxy:
```

nobu:
Seems writing to different (the previous?) `user_home` path.

deivid-rodriguez:
#4344 should fix it! Rubygems abuses memoization :man_shrugging:

nobu:
Thanks, it work perfectly.

nobu:
> When that happened the path of generated files would be unpredictable and some of our tests that make assertions on the number of parts in those paths (for example the test I changed in this PR) would fail.

Thank you for the clarification.
So it is ok without `@tmp` now?

deivid-rodriguez:
I'm not sure but is it a problem?

nobu:
Just I prefer temporary files/directories under `TMPDIR`, not to pollute the source directory.
I'll make another PR about this later, perhaps.

deivid-rodriguez:
I actually prefer the current solution with a local `tmp/` folder. It makes things more self-contained and easier to debug, in my opinion.

PR #4334, opened 2021-02-01
Fix `gem install` crashing on gemspec with nil required_ruby_version

pbernays:

Installing a gem from GemFury in ruby 3.0.0 fails with ```undefined method `satisfied_by?' for nil:NilClass```

An inspection of the gemspec showed both `required_ruby_version` and `required_rubygems_version` were set to nil.

Took the fix for `required_rubygems_version` a step further and applied it to `required_ruby_version`


deivid-rodriguez (commented):
Interesting. When I looked at this, I understood that `required_ruby_version` was present since the initial version of the spec, so this couldn't happen. After I second look, I still believe that's true, but even if that's the case, it seems like gems with this field set to `nil` still crept in, also confirmed by commits like this: https://github.com/rubygems/rubygems/commit/1b700ea30490e1376e2fb640b16bc7c216c6f6c3. So thanks for fixing this!

Could you make the following changes:

* Make a separate test analogous to the other one, instead of extending it.
* Add an already decompressed test file to keep it consistent with the previous test.
* Change the comment to mention that some marshalled specs where generated with this field set to `nil` instead of mentioning the `Gem::Specification` API version.

deivid-rodriguez:
Oh, and you also need to run `rake update_manifest` to get the new test file added to the package.

pbernays:
Not a problem. Since the test file is uncompressed, should I drop the `.rz` extension or would you prefer I leave as is?

pbernays:
Ah, nvm. I see rubocop failing because it picks up the marshalled `.gemspec`. Will adjust the filenames.

deivid-rodriguez (approved):
Thank you!

PR #4356, opened 2021-02-07
Fix `Gem::Specification.stubs_for` returning wrong named specs

tompng:

Gem::Specification.find_all_by_name returns wrong specs
```ruby
Gem::Specification.reset
p Gem::Specification.stubs_for('pry').map(&:name) #=> ['pry', 'pry-doc']
p Gem::Specification.find_all_by_name('pry').map(&:name) #=> ['pry', 'pry-doc']
Gem::Specification.stubs
p Gem::Specification.stubs_for('pry').map(&:name) #=> ['pry']
p Gem::Specification.find_all_by_name('pry').map(&:name) #=> ['pry']
```

```ruby
# in Gem::Specification.stubs_for
@@stubs_by_name[name] ||= stubs_for_pattern("#{name}-*.gemspec")
```
`stubs_for_pattern('pry')` returns `pry`, `pry-doc`, `pry-foo`, `pry-bar`.
this PR rejects `pry-doc`, `pry-foo`, `pry-bar`, which name does not match.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -830,7 +830,9 @@ class Gem::Specification < Gem::BasicSpecification
     if @@stubs
       @@stubs_by_name[name] || []
     else
-      @@stubs_by_name[name] ||= stubs_for_pattern("#{name}-*.gemspec")
+      @@stubs_by_name[name] ||= stubs_for_pattern("#{name}-*.gemspec").select do |s|
+        s.name == name
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
I don't know how to do this directly with a glob, so this looks good to me! :+1:

simi (approved):

simi:
@deivid-rodriguez I think glob can't handle this and even it could, it is probably possible to make `-` in version as well (even seems not supported by default tools).

PR #4384, opened 2021-02-17
Restore the ability to manually install extension gems

### LINKED ISSUE #4375: Incorrect Code Figuring Out Missing Extensions and Deleting Gems

Bundler is incorrectly deciding some gems have missing extensions when in fact they do not. This breaks the ability to manually install gems and then use bundler. This is a new breakage.

Let's go through an example. On mingw64, without ruby devkit installed, install the mysql2 gem:

```gem install --platform ruby mysql2```

It will fail because the mysql2 extconf.rb file has some incorrect logic.  Specifically line 184 should be changed from:

```if RUBY_PLATFORM =~ /mswin|mingw/ && !defined?(RubyInstaller)```

To:

```if RUBY_PLATFORM =~ /mswin/ && !defined?(RubyInstaller)```

So locally fix that line in extconf.rb, ruby, then rerun `ruby extconf.rb --with-mysql2-dir=/mingw64`, then run `make` to build mysql2.so successfully.

Next, follow the instructions from `gem help install` for this exact situation:

```
[build fails]
      Gem files will remain installed in \
      /path/to/gems/some_extension_gem-1.0 for inspection.
      Results logged to /path/to/gems/some_extension_gem-1.0/gem_make.out
      $ [cd /path/to/gems/some_extension_gem-1.0]
      $ [edit files or what-have-you and run make]
      $ gem spec ../../cache/some_extension_gem-1.0.gem --ruby > \
                 ../../specifications/some_extension_gem-1.0.gemspec
      $ gem list some_extension_gem
```
This works great, mysql2 gem is now installed. 

However, with the latest version of bundler this fails as soon as you run ```bundler update``` or ```bundler install``` because bundler does not think the extension exists and it deletes the gem from the list of gems!

Try it:

1. Have a Gemfile that requires mysql2 gem
2. ```bundle install```
3.  ```gem list mysql2``` now shows nothing!

I'm not sure where bundler is deleting the gem (it really should not be), but I do know where the faulty extension code is. See stub_specification.rb:

```
  def missing_extensions?
      return false if default_gem?
      return false if extensions.empty?
      return false if File.exist? gem_build_complete_path

      true
    end
```
This is new code - previously it was:

```
    def missing_extensions?
      stub.missing_extensions?
    end
```

The incorrect line is:

```return false if File.exist? gem_build_complete_path```

The file its looking for (gem_build_complete_path) is:

```
ruby-2.7.2\lib\ruby\gems\2.7.0\extensions\x64-mingw32\2.7.0\mysql2-0.5.3\gem_build.complete
```

Of course that file doesn't exist because we manually built the gem and its shared library.

Like I said, this is a new issue - this type of installation has worked fine in the past.

So a few thoughts:

* It seems strange bundler is even doing this check? Shouldn't RubyGems verify a gem was correctly installed at the time of installation? Not at some later time like bundler is doing here?
* Checking for gem_make.out seems like a poor choice. Shouldn't the .so file be checked instead?
* Can the "real" location of the so file be checked - ie, the gem path (gem/mysql2/lib/2.7/mysql2.so) or in its extension directory?
* If none of the above, then can ```gem help install`` be updated to note you have to create this dummy gem_build.complete file? And a command on how to do it?

Thanks

## Environment

```
Bundler       2.2.9
  Platforms   ruby, x64-mingw32
Ruby          2.7.2p137 (2020-10-01 revision 5445e0435260b449decf2ac16f9d09bae3cafe72) [x64-mingw32]
  Full Path   C:/msys64/usr/local/ruby-2.7.2/bin/ruby.exe
  Config Dir  C:/ProgramData
RubyGems      3.2.9
  Gem Home    C:/msys64/usr/local/ruby-2.7.2/lib/ruby/gems/2.7.0
  Gem Path    D:/cfis/.gem/ruby/2.7.0;C:/msys64/usr/local/ruby-2.7.2/lib/ruby/gems/2.7.0
  User Home   D:/cfis
  User Path   D:/cfis/.gem/ruby/2.7.0
  Bin Dir     C:/msys64/usr/local/ruby-2.7.2/bin
Tools
  Git         2.30.0.windows.2
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2021-02-08
Git SHA           30e740c073
Released Version  true
```

cfis:
Update bundler to include manually_installed gems that have extensions. Manually installed gems are gems that fail to build on installation but are then fixed by hand in-place and registered with RubyGems following the instructions provided by "gem help install". This functionality was unintentionally broken by af7e4b4a.

Fixes #4375.

### DIFF
--- a/bundler/lib/bundler/stub_specification.rb
+++ b/bundler/lib/bundler/stub_specification.rb
@@ -26,11 +26,19 @@ module Bundler
 
     # @!group Stub Delegates
 
+    def manually_installed?
+      # This is for manually installed gems which are gems that were fixed in place after a
+      # failed installation. Once the issue was resolved, the user then manually created
+      # the gem specification using the instructions provided by `gem help install`
+      installed_by_version == Gem::Version.new(0)
+    end
+
     # This is defined directly to avoid having to loading the full spec
     def missing_extensions?
       return false if default_gem?
       return false if extensions.empty?
       return false if File.exist? gem_build_complete_path
+      return false if manually_installed?
 
       true
     end
--- a/bundler/spec/bundler/stub_specification_spec.rb
+++ b/bundler/spec/bundler/stub_specification_spec.rb
@@ -6,6 +6,7 @@ RSpec.describe Bundler::StubSpecification do
       s.name = "gemname"
       s.version = "1.0.0"
       s.loaded_from = __FILE__
+      s.extensions = "ext/gemname"
     end
 
     described_class.from_stub(gemspec)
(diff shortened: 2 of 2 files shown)


deivid-rodriguez (approved):
Honestly I don't like this too much and would like to remove this documented instructions in favor of better alternatives for the next major.

But this was working before, and was documented, so fixing it for now.

Thanks @cfis!

cfis:
Thanks! Happy to test whatever changes you have in mind for the future when the time comes.

deivid-rodriguez:
My idea to properly fix this would be to add the ability to install gems from git sources, like bundler can do. With that in place, you could do the right thing here: fork the gem, fix the bug in the extension at the source, contribute the fix upstream, and then install your fixed version from your fork while the fix is released.

cfis:
Sounds like a good plan to me.

deivid-rodriguez:
Thanks!

PR #4390, opened 2021-02-18
Improve `include_gems` matcher

deivid-rodriguez:

The `include_gems` matcher relies on the output of a bundler subshell to work. This can get broken with weird errors if you introduce debugging output while developing bundler.

Rely on process status instead, and also do all checks in a single subprocess instead of many for efficiency.

This is a follow up to #4388, but for the `include_gems` "positive" matcher.


deivid-rodriguez:
Changes here break specs on jruby. I'll wait till jruby 9.2.16.0 is shipped, since it should bring some bug fixes when spawing subprocess and might fix this.

deivid-rodriguez:
Upgrading to JRuby 9.2.16.0 indeed fixes this and allows us to remove a JRuby specific workaround from our specs.

PR #4381, opened 2021-02-15
Lock GEM sources separately and fix locally installed specs confusing bundler

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Our latest release broke some simple cases of using multiple rubygems sources. This PR tries to reintroduce some of the fixes without breaking anything. In particular, this PR introduces separate GEM sources in the lockfile, which avoids "Dependency Confusion" for all usages with an existing lockfile. Making this change in 2.2.10 made an issue surface where already installed gems would confuse bundler and make it use gems from the wrong source (#4383). Actually, a similar issue had been [reported before](https://github.com/rubygems/rubygems/issues/4016) and it was not tied to the change of locking GEM sources separately but to the way sources work in bundler in general. This PR also fixes that issue.

## What is your fix for the problem, implemented in this PR?

Lock GEM sources separately in the lockfile making it predictable and deterministic where each gem is pulled from once you have a lockfile. To avoid the problem of unintentionally unlocking all GEM sources in existing lockfiles and thus upgrading everything (even on `bundle install`, not only on `bundle update`), we keep the current behaviour for existing lockfiles, but show a warning for lockfiles with "merged GEM sections" recommending users to run `bundle update` or  regenerate the lockfile from scratch, so that they can opt-in to lockfile GEM sources separately.

To fix the issue of locally installed gems confusing bundler, we change the existing behaviour of always falling back to locally installed specs, and restrict that fallback to the default source only.
   
Fixes #3643.
Fixes #4016.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

Issue #750, opened 2013-12-17
Installing gem 2nd time results in infinity loop


voxik:
Doing `$ gem install bcrypt-ruby && gem install bcrypt-ruby`, i.e. installing bcrypt-ruby twice results in infinite loop (:warning: be careful when trying to reproduce).

It seems that running extconf.rb spawns new instance of Ruby, which thinks that the extension is not build and tries it to build again, executing extonf.rb once more etc. After abort, Ruby is unusable until you delete the gem leftovers.

I am afraid, that similar issue might happen when extension build fails for some reason.

Not sure if the automatic build of extensions is good idea ...

PR #4408, opened 2021-02-25
Fix race conditions when reading & writing gemspecs concurrently

deivid-rodriguez:

While working on #4381, I added a spec that failed sporadically because of several warnings being printed about read gemspecs being `nil`.

Since I had seen this issue many times in the past, having a spec reproducing seemed like a good opportunity to look at it.

The problem was that gemspec writes were not protected by file locks, and even when protected file locks were added after the file was created, not during creation, so race conditions still appeared.

Another issue with concurrent gemspec reads and installations, is that rubygems installer would remove the gemspec if present at the beginning on the installation, and then install the new gemspec in the end. This was introduced at https://github.com/rubygems/rubygems/commit/af604436d8141c34cb2e1e645b9b0d47bfd55a55 to fix some [hangs when running two `gem install` commands in a row](https://github.com/rubygems/rubygems/issues/750). But that issue was never reproducible, and it still isn't, so I'm reverting it so that I can get my regression test green.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -800,11 +800,11 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   ##
   # Safely write a file in binary mode on all platforms.
   def self.write_binary(path, data)
+    File.open(path, File::RDWR | File::CREAT | File::BINARY | File::LOCK_EX) do |io|
+      io.write data
+    end
+  rescue *WRITE_BINARY_ERRORS
     File.open(path, 'wb') do |io|
-      begin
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Since this is not fixing a real problem in rubygems, and I ended up taking a different approach for fixing the realworld issue inside bundler. I think the changes here still have value, but I won't be introducing them for the moment.

simi:
:muscle:

deivid-rodriguez:
We got a hang in CI after merging this PR 😬, I will monitor it and revert it if it keeps coming up.

PR #4409, opened 2021-02-26
Fix sporadic warnings about `nil` gemspec on install/update and make those faster

deivid-rodriguez:

The problem is that sometimes running `bundle update` ends up printing some annoying warnings about `nil` gemspecs like

```
[/home/runner/work/rubygems/rubygems/bundler/tmp/2/gems/system/specifications/zeitwerk-2.4.2.gemspec] isn't a Gem::Specification (NilClass instead).
```

The fix is to completely override rubygems installer to get rid of the places where it accessed and modified shared gem specification cache. In particular

* Stop calling `Gem::Specification.reset` after each gem installation and instead move it to the main thread, after installation of all gems has finished.
* Stop using rubygems implementation of `Gem::Specification.latest_spec_for(name)` in favour of `Gem::Specification.stubs_for(name).first`, which does the same thing but without traversing and potentially loading all specifications.
* Remove all the other code that's not actually necessary for bundler.

Technically only the first point would be necessary to fix the race condition but I figured since I needed to overwrite the method I would only include what's needed, and make it faster.

As a result, `bundle install` is noticiably faster now.

### DIFF
--- a/bundler/lib/bundler/installer.rb
+++ b/bundler/lib/bundler/installer.rb
@@ -89,6 +89,8 @@ module Bundler
         end
         install(options)
 
+        Gem::Specification.reset # invalidate gem specification cache so that installed gems are immediately available
+
         lock unless Bundler.frozen_bundle?
         Standalone.new(options[:standalone], @definition).generate if options[:standalone]
       end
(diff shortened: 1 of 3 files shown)


indirect (approved):
🎉

PR #4410, opened 2021-02-26
Fix deployment install with duplicate path gems added to Gemfile

internets:

When adding the same path gem twice to a Gemfile, bundle install will 
succeed with deployment set to "false" but fail with deployment set to 
"true". The error message given is "The gemspecs for path gems changed". 
To fix, duplicate dependencies must be removed before checking against 
the locked dependencies.

The implementation of `Bundler::Definition#dependencies_for_source_changed?` prior to #4297 relied on using `Set.new(a) != Set.new(b)` to compare the source dependencies and lockfile dependencies.
This restores the duplicate pruning behavior of `Set` without relying on `Set` itself.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -601,7 +601,7 @@ module Bundler
       deps_for_source = @dependencies.select {|s| s.source == source }
       locked_deps_for_source = @locked_deps.values.select {|dep| dep.source == locked_source }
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):

deivid-rodriguez:
Thanks!

internets:
My pleasure! Thanks for all your work on bundler (and others)!

Issue #3662, opened 2020-05-27
Improve `SharedHelpers.major_deprecation`


deivid-rodriguez:
We currently use this helper for deprecation messages. You pass a major version and a deprecation message to it and it will display the deprecation warning if running that major version. That works well.

For example.

```
Bundler::SharedHelpers.major_deprecation 2, "The --binstubs option will be removed in favor of `bundle binstubs`"
```

will print the warning

> [DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs`.

when running bundler 2.

However, the helper will also _fail_ with the same message when running higher versions of bundler. For example, if running bundler 3 it will fail hard with the error

> [REMOVED] The --binstubs option will be removed in favor of `bundle binstubs`.

This last "mode" reads not great in my opinion, because deprecation messages don't work for "removals", since they talk in future tense instead of present tense, and sometimes hard errors need to be worded differently from deprecations.

I believe we should improve this.

nevinera:
@deivid-rodriguez - my instinct here is that we should add an optional `removed` keyword argument that is used for the third case, and then update most of the call-sites to supply it.

If it seems safe enough, I could drop the positional arguments entirely and go with three _required_ kwargs, `version`, `deprecated`, and `removed`, and update _all_ the call-sites, but I'm not sure how much merge-churn to expect - I'd prefer you made that call? I'll draft up the first version to start.

deivid-rodriguez:
I agree with the approach you chose of adding a removed `:keyword` parameter. Thanks for working on this, I'll try to review the PR later today.

PR #4411, opened 2021-02-27
Add standard option alongside rubocop to `bundle gem`

searls:
This commit converts the --rubocop and --no-rubocop flags to a new --linter option. The option supports either the standard or rubocop gems, and (similar to the --ci option), it can also be set either interactively or via a new `gem.linter` global config.

I do have one outstanding question for the rubygems maintainers that should be discussed before merge, however. Currently, there are [a number of specs](https://github.com/rubygems/rubygems/blob/master/bundler/spec/commands/newgem_spec.rb#L237-L263) that ensure projects that are generated with RuboCop configurations will initially pass RuboCop. However, the default project template will _not_ pass by default:

```
standard: Use Ruby Standard Style (https://github.com/testdouble/standard)
       standard: Run `standardrb --fix` to automatically fix some problems.
         mygem.gemspec:6:12: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:7:15: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:8:15: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:9:13: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:11:15: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:12:19: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:13:16: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:29:14: Layout/ExtraSpacing: Unnecessary spacing detected.
         mygem.gemspec:30:19: Layout/ExtraSpacing: Unnecessary spacing detected.
```

Should we leave this as-is, or should we execute `rake standard:fix` after the project is initialized to ensure the project successfully builds without requiring users to fix it themselves?

When creating a new gem, I like to use [standard](https://github.com/testdouble/standard) for automated linting & formatting (full disclosure: I maintain standard). If you're not familiar with standard, it's built on top of rubocop and provides a fixed ruleset that is intentionally not customizable by users, for the purpose of promoting consistency across Ruby projects.

When I saw that the `bundle gem` CLI included an option for adding RuboCop, I figured it'd be neat to include Standard alongside it, similar to how the `--ci` option includes four CI service options.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -162,7 +162,7 @@ module Bundler
         Bundler.ui.info "RuboCop enabled in config"
         templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
       when "standard"
-        config[:linter_version] = "0.2.5" if Gem.ruby_version < Gem::Version.new("2.4.a")
+        config[:linter_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.2.5" : "1.0"
         Bundler.ui.info "Standard enabled in config"
         templates.merge!("standard.yml.tt" => ".standard.yml")
       end
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -155,15 +155,16 @@ module Bundler
         templates.merge!("CHANGELOG.md.tt" => "CHANGELOG.md")
       end
 
-      if ask_and_set(:rubocop, "Do you want to add rubocop as a dependency for gems you generate?",
-        "RuboCop is a static code analyzer that has out-of-the-box rules for many " \
-        "of the guidelines in the community style guide. " \
-        "For more information, see the RuboCop docs (https://docs.rubocop.org/en/stable/) " \
-        "and the Ruby Style Guides (https://github.com/rubocop-hq/ruby-style-guide).")
-        config[:rubocop] = true
-        config[:rubocop_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.81.0" : "1.7"
+      config[:linter] = ask_and_set_linter
+      case config[:linter]
+      when "rubocop"
+        config[:linter_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.81.0" : "1.7"
         Bundler.ui.info "RuboCop enabled in config"
         templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
+      when "standard"
+        config[:linter_version] = "0.2.5" if Gem.ruby_version < Gem::Version.new("2.4.a")
+        Bundler.ui.info "Standard enabled in config"
+        templates.merge!("standard.yml.tt" => ".standard.yml")
       end
 
       templates.merge!("exe/newgem.tt" => "exe/#{name}") if config[:exe]
@@ -306,6 +307,58 @@ module Bundler
       end
 
       ci_template
+    end
+
+    def ask_and_set_linter
+      linter_template = options[:linter] || Bundler.settings["gem.linter"]
+      linter_template = deprecated_rubocop_option if linter_template.nil?
+
+      if linter_template.to_s.empty?
+        Bundler.ui.confirm "Do you want to add a code linter and formatter to your gem? " \
+          "Supported Linters:\n" \
+          "* RuboCop:       https://rubocop.org\n" \
+          "* Standard:      https://github.com/testdouble/standard\n" \
+          "\n"
+        Bundler.ui.info hint_text("linter")
+
+        result = Bundler.ui.ask "Enter a linter. rubocop/standard/(none):"
+        if result =~ /rubocop|standard/
+          linter_template = result
+        else
+          linter_template = false
+        end
+      end
+
+      if Bundler.settings["gem.linter"].nil?
+        Bundler.settings.set_global("gem.linter", linter_template)
+      end
+
+      # Once gem.linter safely set, unset the deprecated gem.rubocop
+      unless Bundler.settings["gem.rubocop"].nil?
+        Bundler.settings.set_global("gem.rubocop", nil)
+      end
+
+      if options[:linter] == Bundler.settings["gem.linter"]
+        Bundler.ui.info "#{options[:linter]} is already configured, ignoring --linter flag."
+      end
+
+      linter_template
+    end
+
+    def deprecated_rubocop_option
+      if !options[:rubocop].nil?
+        if options[:rubocop]
+          Bundler::SharedHelpers.major_deprecation 2, "--rubocop is deprecated, use --linter=rubocop"
+          "rubocop"
+        else
+          Bundler::SharedHelpers.major_deprecation 2, "--no-rubocop is deprecated, use --linter"
+          false
+        end
+      elsif !Bundler.settings["gem.rubocop"].nil?
+        Bundler::SharedHelpers.major_deprecation 2,
+          "config gem.rubocop is deprecated; we've updated your config to use gem.linter instead"
+        Bundler.settings["gem.rubocop"] ? "rubocop" : false
+      end
     end
 
     def bundler_dependency_version
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -21,6 +21,16 @@ RSpec.describe "bundle gem" do
     bundle "exec rubocop --debug --config .rubocop.yml", :dir => bundled_app(gem_name)
   end
 
+  def bundle_exec_standardrb
+    prepare_gemspec(bundled_app(gem_name, "#{gem_name}.gemspec"))
+    standard_version = RUBY_VERSION > "2.4" ? "" : " -v 0.2.5"
+    gems = ["minitest", "rake", "rake-compiler", "rspec", "standard#{standard_version}", "test-unit"]
+    gems.unshift "parallel -v 1.19.2" if RUBY_VERSION < "2.5"
+    path = Bundler.feature_flag.default_install_uses_path? ? local_gem_path(:base => bundled_app(gem_name)) : system_gem_path
+    realworld_system_gems gems, :path => path
+    bundle "exec standardrb --debug", :dir => bundled_app(gem_name)
+  end
+
   let(:generated_gemspec) { Bundler.load_gemspec_uncached(bundled_app(gem_name).join("#{gem_name}.gemspec")) }
 
   let(:gem_name) { "mygem" }
@@ -147,8 +157,68 @@ RSpec.describe "bundle gem" do
   end
 
   shared_examples_for "--rubocop flag" do
+    context "is deprecated", :bundler => "< 3" do
+      before do
+        bundle "gem #{gem_name} --rubocop"
+      end
+
+      it "generates a gem skeleton with rubocop" do
+        gem_skeleton_assertions
+        expect(bundled_app("test-gem/Rakefile")).to read_as(
+          include("# frozen_string_literal: true").
+          and(include('require "rubocop/rake_task"').
+          and(include("RuboCop::RakeTask.new").
+          and(match(/default:.+:rubocop/))))
+        )
+      end
+
+      it "includes rubocop in generated Gemfile" do
+        allow(Bundler::SharedHelpers).to receive(:find_gemfile).and_return(bundled_app_gemfile)
+        builder = Bundler::Dsl.new
+        builder.eval_gemfile(bundled_app("#{gem_name}/Gemfile"))
+        builder.dependencies
+        rubocop_dep = builder.dependencies.find {|d| d.name == "rubocop" }
+        expect(rubocop_dep).not_to be_nil
+      end
+
+      it "generates a default .rubocop.yml" do
+        expect(bundled_app("#{gem_name}/.rubocop.yml")).to exist
+      end
+    end
+  end
+
+  shared_examples_for "--no-rubocop flag" do
+    context "is deprecated", :bundler => "< 3" do
+      define_negated_matcher :exclude, :include
+
+      before do
+        bundle "gem #{gem_name} --no-rubocop"
+      end
+
+      it "generates a gem skeleton without rubocop" do
+        gem_skeleton_assertions
+        expect(bundled_app("test-gem/Rakefile")).to read_as(exclude("rubocop"))
+        expect(bundled_app("test-gem/#{gem_name}.gemspec")).to read_as(exclude("rubocop"))
+      end
+
+      it "does not include rubocop in generated Gemfile" do
+        allow(Bundler::SharedHelpers).to receive(:find_gemfile).and_return(bundled_app_gemfile)
+        builder = Bundler::Dsl.new
+        builder.eval_gemfile(bundled_app("#{gem_name}/Gemfile"))
+        builder.dependencies
+        rubocop_dep = builder.dependencies.find {|d| d.name == "rubocop" }
+        expect(rubocop_dep).to be_nil
+      end
+
+      it "doesn't generate a default .rubocop.yml" do
+        expect(bundled_app("#{gem_name}/.rubocop.yml")).to_not exist
+      end
+    end
+  end
+
+  shared_examples_for "--linter=rubocop flag" do
     before do
-      bundle "gem #{gem_name} --rubocop"
+      bundle "gem #{gem_name} --linter=rubocop"
     end
 
     it "generates a gem skeleton with rubocop" do
@@ -175,11 +245,38 @@ RSpec.describe "bundle gem" do
     end
   end
 
-  shared_examples_for "--no-rubocop flag" do
+  shared_examples_for "--linter=standard flag" do
+    before do
+      bundle "gem #{gem_name} --linter=standard"
+    end
+
+    it "generates a gem skeleton with standard" do
+      gem_skeleton_assertions
+      expect(bundled_app("test-gem/Rakefile")).to read_as(
+        include('require "standard/rake"').
+        and(match(/default:.+:standard/))
+      )
+    end
+
+    it "includes standard in generated Gemfile" do
+      allow(Bundler::SharedHelpers).to receive(:find_gemfile).and_return(bundled_app_gemfile)
+      builder = Bundler::Dsl.new
+      builder.eval_gemfile(bundled_app("#{gem_name}/Gemfile"))
+      builder.dependencies
+      standard_dep = builder.dependencies.find {|d| d.name == "standard" }
+      expect(standard_dep).not_to be_nil
+    end
+
+    it "generates a default .standard.yml" do
+      expect(bundled_app("#{gem_name}/.standard.yml")).to exist
+    end
+  end
+
+  shared_examples_for "--linter=none flag" do
     define_negated_matcher :exclude, :include
 
     before do
-      bundle "gem #{gem_name} --no-rubocop"
+      bundle "gem #{gem_name} --linter=none"
     end
 
     it "generates a gem skeleton without rubocop" do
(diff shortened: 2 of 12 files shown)


deivid-rodriguez:
@searls Thanks for this.

I'm happy with adding a `--linter` option that allows you to setup "standard". But I think we should maintain CLI compatibility for now for people using the existing `--rubocop` option and configuration.

Regarding having a new gem that passes standard checks, I think that'd be a good thing and in my opinion, given we're only getting one offense at the moment, I would fix the issue directly inside our templates.

searls:
Thanks @deivid-rodriguez! That sounds like a great plan. I'll update the PR when I get a chance (I'm restoring my Mac right now so it might be a little while)

Regarding the `--rubocop` flag: so that it's not left in Bundler forever, do you all have an approach for flagging things like CLI options for deprecation so that the `--rubocop` flag can be removed at the next appropriate release?

deivid-rodriguez:
`bundle gem` configuration & CLI flags are currently quite flexible, but as a result a deprecation like this, and making these two options be available simultaneously, sounds quite tricky to get right :grimacing:.  We currently have some deprecated CLI flags (you can grep for `SharedHelpers.major_deprecation` and will find some), but I don't think we have deprecated configurations at the moment, so that might be more work.

I guess the deprecation should work like this:

* If `--rubocop` flag is used, warn and recommend `--linter=rubocop`, and run the command as if `--linter=rubocop` had been specified.

* If `--rubocop=false` flag is used, warn and recommend `--linter=false`, and run the command as if `--linter=false` had been specified.

* If conflicting `--rubocop`and `--linter` options are used at the same time... I guess we can ignore that case, honestly? People doing that should get whatever they deserve :laughing:.

* If `gem.rubocop false` is configured and applied (because no CLI flags given), we should act as if `gem.linter false` was configured, and tell people to replace the `gem.rubocop false` configuration with `gem.linter <rubocop|standard|false>`.

* If `gem.rubocop true` is configured and applied (because no CLI flags given), we should act as if `gem.linter rubocop` was configured, and tell people to replace the `gem.rubocop true` configuration in favour of `gem.linter <rubocop|standard|false>`.

* If you configure `gem.linter` directly (`bundle config linter <value>`) and there's an existing `gem.rubocop` configuration, I would remove the existing `gem.rubocop` configuration first without warning.

* When nothing is yet configured, the current implementation sounds just right: change the initial question to ask for a generic linter.

searls:
Hi @deivid-rodriguez, thanks for the detailed list of edge cases & requirements, it was helpful to work through. I've pushed up changes that restores `--rubocop`/`--no-rubocop` and `gem.rubocop` (they're resorted to after `--linter` and `gem.linter`), added deprecation warnings when they're used, and when they're used, `gem.linter` is written based on the. I also added tests for each of these deprecation warnings.

Finally, I changed the template's gemspec so that its style will pass both Rubocop & Standard by default

searls:
The changes I made to the `major_deprecation_spec` are failing under Windows and I can't really figure out why (I see a lot of `ZeroDivisionError`, which looks like it's masking whatever's happening)

deivid-rodriguez:
Oh yeah, that's a long standing issue I should have a look at (one of the many :sweat_smile:). For now what we're doing in other  similar specs is tag them with `:readline` so that they are skipped on Windows.

deivid-rodriguez:
Same backtrace was reported to irb here: https://github.com/ruby/irb/issues/50, and the issue should've been fixed/workarounded inside `reline` since. But we can't yet upgrade `reline` in our CI since new versions don't support rubies that we do support. So skipping on Windows seems fine for now.

searls:
Ok, @deivid-rodriguez I made that change but there are still some failures that I don't really understand and which had been green previously https://github.com/rubygems/rubygems/pull/4411/checks?check_run_id=2004639922

deivid-rodriguez:
Those failure are for "bundler 3" entries, which test future behavior of bundler 3. The results seem expected since all those commands will start failing on bundler 3 since the deprecated flag will be removed. You can use the `:bundler => "< 3"` rspec filter for those specs.

searls:
@deivid-rodriguez oh neat; it's cool how sophisticated the deprecation behavior in bundler is

deivid-rodriguez:
Hopefully other users will agree with you, we have never yet completed a full deprecation cycle... :sweat_smile:. We've been unintentionally breaking things sometimes, but not yet intentionally doing that :laughing:. I think we can improve the deprecation helpers though, for example, I reported https://github.com/rubygems/rubygems/issues/3662.

searls:
Hi @deivid-rodriguez, AFAICT I've gotten everything else working so far, but the two remaining failed checks don't make sense to me: https://github.com/rubygems/rubygems/pull/4411/checks?check_run_id=2005561808

deivid-rodriguez:
That seems like a real issue with the implementation.

Previously those specs were pre-configuring `gem.rubocop` explicitly to be `false`, to avoid `bundle gem` asking for input. However, now that you changed it to set `gem.linter` to `false` instead, it sounds like `reline` is coming into play which I think means `bundle gem` is asking offering the user to choose a linter even if it shouldn't since `gem.linter` is explicitly configured as `false`.

searls:
@deivid-rodriguez thanks for explaining this to me! I'll take time this evening to look at it

deivid-rodriguez:
No problem!

searls:
@deivid-rodriguez I think I understood your feedback, but I'm struggling to connect it to the currently failing checks (all Windows, ZeroDivisionError, like the readline comment you had above). Can you take another look or help me connect the dots?

deivid-rodriguez:
I think `readline` is coming into play where it shouldn't, which would mean `bundle gem` is prompting the user to choose a linter where it shouldn't. Actually my advice above of using the `:readline` to skip this issue was probably wrong and the problem was due to the same issue we're facing now.

This is the issue:

```
$ bundle config gem.linter false
$ bundle gem foo
Creating gem 'foo'...
MIT License enabled in config
Code of conduct enabled in config
Changelog enabled in config
Do you want to add a code linter and formatter to your gem? Supported Linters:
* RuboCop:       https://rubocop.org
* Standard:      https://github.com/testdouble/standard

Your choice will only be applied to this gem.
Enter a linter. rubocop/standard/(none): ^C
```

In the above case, bundler shouldn't prompt because the user explicitly configured `gem.linter` as false.

This is how other similar flags already work. For example,

```
$ bundle config gem.ci false
$ bundle gem foo
Creating gem 'foo'...
MIT License enabled in config
Code of conduct enabled in config
Changelog enabled in config
      create  foo/Gemfile
      create  foo/lib/foo.rb
      create  foo/lib/foo/version.rb
      create  foo/foo.gemspec
      create  foo/Rakefile
      create  foo/README.md
      create  foo/bin/console
      create  foo/bin/setup
      create  foo/.gitignore
      create  foo/.rspec
      create  foo/spec/spec_helper.rb
      create  foo/spec/foo_spec.rb
      create  foo/LICENSE.txt
      create  foo/CODE_OF_CONDUCT.md
      create  foo/CHANGELOG.md
Initializing git repo in /home/deivid/Code/rubygems/rubygems/foo
Gem 'foo' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html
```

searls:
Thanks, @deivid-rodriguez! I'm new to thinking about the `~/.bundle/config` file, so in experimenting with your example, I think I found that the ci option is also behaving the way you describe above.

Here is a longish shell transcript:

```
$ bundle config ci false
$ bundle config linter false
$ cat ~/.bundle/config 
---
BUNDLE_CI: "false"
BUNDLE_LINTER: "false"
```

And then when creating a gem, it interactively asks for both CI and for linter:

```
$ bundle gem foo
Creating gem 'foo'...
Do you want to generate tests with your gem?
Future `bundle gem` calls will use your choice. This setting can be changed anytime with `bundle config gem.test`.
Enter a test framework. rspec/minitest/test-unit/(none): 
HEYO 
AND true
Do you want to set up continuous integration for your gem? Supported services:
* CircleCI:       https://circleci.com/
* GitHub Actions: https://github.com/features/actions
* GitLab CI:      https://docs.gitlab.com/ee/ci/
* Travis CI:      https://travis-ci.org/

Future `bundle gem` calls will use your choice. This setting can be changed anytime with `bundle config gem.ci`.
Enter a CI service. github/travis/gitlab/circle/(none): 
Do you want to license your code permissively under the MIT license?
This means that any other developer or company will be legally allowed to use your code for free as long as they admit you created it. You can read more about the MIT license at https://choosealicense.com/licenses/mit. y/(n): 
Do you want to include a code of conduct in gems you generate?
Codes of conduct can increase contributions to your project by contributors who prefer collaborative, safe spaces. You can read more about the code of conduct at contributor-covenant.org. Having a code of conduct means agreeing to the responsibility of enforcing it, so be sure that you are prepared to do that. Be sure that your email address is specified as a contact in the generated code of conduct so that people know who to contact in case of a violation. For suggestions about how to enforce codes of conduct, see https://bit.ly/coc-enforcement. y/(n): 
Do you want to include a changelog?
A changelog is a file which contains a curated, chronologically ordered list of notable changes for each version of a project. To make it easier for users and contributors to see precisely what notable changes have been made between each release (or version) of the project. Whether consumers or developers, the end users of software are human beings who care about what's in the software. When the software changes, people want to know why and how. see https://keepachangelog.com y/(n): 
HEYO 
AND true
Do you want to add a code linter and formatter to your gem? Supported Linters:
* RuboCop:       https://rubocop.org
* Standard:      https://github.com/testdouble/standard

Future `bundle gem` calls will use your choice. This setting can be changed anytime with `bundle config gem.linter`.
Enter a linter. rubocop/standard/(none): 
Initializing git repo in /Users/justin/tmp/foo
      create  foo/Gemfile
      create  foo/lib/foo.rb
      create  foo/lib/foo/version.rb
      create  foo/foo.gemspec
      create  foo/Rakefile
      create  foo/README.md
      create  foo/bin/console
      create  foo/bin/setup
      create  foo/.gitignore
Gem 'foo' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html
```

searls:
Ok, things are passing!

bundler/lib/bundler/cli/gem.rb
```diff
@@ -155,15 +155,16 @@ def run
...
-        config[:rubocop] = true
-        config[:rubocop_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.81.0" : "1.7"
+      config[:linter] = ask_and_set_linter
+      case config[:linter]
+      when "rubocop"
+        config[:linter_version] = Gem.ruby_version < Gem::Version.new("2.4.a") ? "0.81.0" : "1.7"
         Bundler.ui.info "RuboCop enabled in config"
         templates.merge!("rubocop.yml.tt" => ".rubocop.yml")
+      when "standard"
+        config[:linter_version] = "0.2.5" if Gem.ruby_version < Gem::Version.new("2.4.a")
```
deivid-rodriguez, line 165: Would it be too bad if we locked the version in general, at least until you reach 1.0? Users might get an outdated standard version by default, but we'll probably get updates proposed and we'll try to keep it up to date.

Our CI is hard enough to keep green, so we don't want it to also break to due to changes in third party libraries.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -308,6 +309,52 @@ def ask_and_set_ci
...
+        Bundler.ui.info "#{options[:linter]} is already configured, ignoring --linter flag."
+      end
+
+      linter_template
+    end
+
+    def deprecated_rubocop_option
+      if !options[:rubocop].nil?
+        if options[:rubocop]
+          Bundler::SharedHelpers.major_deprecation 2, "--rubocop is deprecated, use --linter"
```
simi, line 346: Would it be beneficial to provide exact command `--linter=rubocop` to make it super simple to migrate?

bundler/spec/commands/newgem_spec.rb
```diff
@@ -175,11 +245,38 @@ def bundle_exec_rubocop
...
+      standard_dep = builder.dependencies.find {|d| d.name == "standard" }
+      expect(standard_dep).not_to be_nil
+    end
+
+    it "generates a default .standard.yml" do
+      expect(bundled_app("#{gem_name}/.standard.yml")).to exist
+    end
+  end
+
+  shared_examples_for "--linter=none flag" do
```
deivid-rodriguez, line 275: This is interesting. I think you probably used "none" here because the prompt says "rubocop/standard/(none)", right? I think "none" in the prompt is meant as "leave empty, just hit enter". I think `--linter=none` and in general `--linter=any_other_undefined_value` works fine accidentally because there's no behaviour associated to it.

Actually "none" reads slightly better to me, and we will never add a linter named "none" that would mess things up, but since `false` is already used for configuring through ENV (`BUNDLE_GEM__LINTER=false`), or through bundler config (`bundle config linter false`), I'd stick to `--linter=false` here?

By the way, thanks for testing this case, I don't think it's tested for other similar flags.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -308,6 +309,52 @@ def ask_and_set_ci
...
+          "rubocop"
+        else
+          Bundler::SharedHelpers.major_deprecation 2, "--no-rubocop is deprecated, use --linter"
+          false
+        end
+      elsif !Bundler.settings["gem.rubocop"].nil?
+        Bundler::SharedHelpers.major_deprecation 2, "gem.rubocop is deprecated, use gem.linter"
```
deivid-rodriguez, line 353: Here I'd say something like

> Bundler is using your existing `gem.rubocop` configuration, but this configuration is deprecated. Make sure you migrate this configuration by running `bundle config set linter <value>`.

But then you'd need to implement the 6th bullet point I suggesed in https://github.com/rubygems/rubygems/pull/4411#issuecomment-787450937, and maybe I'm asking too much from this PR... :thinking:

deivid-rodriguez (commented):
I added some comments, it's looking great!

bundler/lib/bundler/cli/gem.rb
searls: ok, done

bundler/spec/commands/newgem_spec.rb
searls: Yeah, I used "none" just because it reads nicer. Both this and the ci flag will take non-"false" flags and parse them as strings, and then when they aren't present in the `case` statement, they'll ultimately resolve to false. Seems fine?

deivid-rodriguez: Sure, let's leave it as it is.

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez, line 165: (I  made this comment in my initial review but somehow I deleted it, reposting it now)

Would it be too bad if we locked the version in general, at least until you reach 1.0? Users might get an outdated standard version by default, but we'll probably get updates proposed and we'll try to keep it up to date.

Our CI is hard enough to keep green, so we don't want it to also break to due to changes in third party libraries.

bundler/lib/bundler/cli/gem.rb
searls: Ok, I've implemented and tested that case now. If `gem.rubocop` is set, gem.linter will be written and gem.rubocop will be deleted (with a message to that effect). [test here](https://github.com/searls/rubygems/blob/master/bundler/spec/commands/newgem_spec.rb#L1015-L1034)

bundler/lib/bundler/cli/gem.rb
searls: I take your point. Here are some of my thoughts:

As a consultant, I visit lots and lots of Ruby teams, and something that's very common (but thanks to Bundler's sophistication, usually unnecessary) is for teams to proactively lock down versions of everything in their Gemfile and only make updates harder for themselves. When a version is pinned in a Gemfile, it (should) communicate to the reader "there is something dangerous about upgrading this and trusting the lockfile", but if most or every gem is locked by default (as it is in the Gemfile generated by Rails, too), then a pinned version doesn't tell the reader anything useful; it just makes staying up-to-date harder. I think tools that generate Gemfiles (chief of them Bundler, of course) should lead by example and default to only pinning versions where there's a known risk of upgrading, to help people get and stay in the habit of staying better on-top-of their dependencies by trusting bundle update and version control to role back any future incompatibilities

As for the risk to Bundler's CI (because I can empathize having worked to get all 72 checks green), the integrations I've added with standard that could break things are (1) a name change, (2) a change to the require statement in the Rakefile, (3) the name of the standard rake command, and (4) they default style ruleset as it applies to the gemspec (which is a pretty minimal file). At least for items 1-3, I'm so confident that these won't change that I'm willing to commit to proactively fix them in Bundler before they do.

To your third point, we haven't come to decide what 1.0 will "mean" from Standard's perspective (e.g. should it be a commitment that code fixes shouldn't churn git without a major bump? Obviously Rubocop isn't adhering to that, so that'd be tough). But one thing I will say is that the road from here to wherever 1.0 is to increase its stability and to relax its rules that cause false negatives, so in practice I wouldn't recommend anybody pin their Gemfile to a version now (and pinning it in this tool is essentially a context-free recommendation, re: the above). If this ends up causing you all any grief, I'm willing to come back and offer some kind of solution to stabilize your build (including pinning the version in tests somehow without leaking that through to the end user)

Are you comfortable moving forward on this as-is, or do you want it pinned before merge?

deivid-rodriguez: I agree with your points.

The key is each one interpretation of "should lead by example and default to only pinning versions where there's a known risk of upgrading". I think it's a common (and reasonable) conception that gems in their 0.x series are inherently unstable, so pinning those in the default template seems totally reasonable to me. And given we don't pin other gems in there, I think it's easy to understand why the gem is being pinned.

Regarding the risk to bundler's CI, (4) is exactly the issue I had in mind, reinforced by the fact that standard warns about this on its own README.

I'd say let's pin it.

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Thanks!

Well my suggestion was recommending running `bundle config gem.linter <value>` in the deprecation message and changing the behaviour of `bundle config` to do the update. I was thinking of suggesting on the fly updating of the old setting like you did here, but I thought it was too magical. But after a second thought your approach is nicer and simpler :+1:.

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: >  (4) they default style ruleset as it applies to the gemspec (which is a pretty minimal file)

Just to clarify, it's not only the gemspec but all files in the default gem skeleton, right? Definitely not a lot, but more than a single small file, and they might grow.

bundler/lib/bundler/cli/gem.rb
simi: Just to mention, I do remember some reports of running rubygems on read-only file systems complaining about unexpected file writes. But since this config is not meant for production runtime, but for devs, I think it is ok to migrate on the fly. :muscle:

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: > As a consultant, I visit lots and lots of Ruby teams, and something that's very common (but thanks to Bundler's sophistication, usually unnecessary) is for teams to proactively lock down versions of everything in their Gemfile and only make updates harder for themselves.

I can confirm I've seen this too, and I'm willing to improve our documentation with better Gemfile best practices that most people can agree with. I bet if you properly explain this to teams currently doing this, they would change their practice.

searls:
@deivid-rodriguez okay, two updates:

* Got the build passing on all checks
* Based on the discussion of pinning version, because we've been comfortable with its general stability for a while now, I switched gears and implemented the last several patches blocking StandardRB from hitting 1.0.0 and released 1.0.0

deivid-rodriguez:
That's great @searls!

Can you then follow the same style the other gems in the generated `Gemfile` use, and add a `~> 1.0` requirement in there? I guess it's a separate discussion whether that's a good practice or not, but let's respect the current style for now.

searls:
> Can you then follow the same style the other gems in the generated `Gemfile` use, and add a `~> 1.0` requirement in there? I guess it's a separate discussion whether that's a good practice or not, but let's respect the current style for now.

Ok, done and build passing. Good to merge?

simi (approved):

deivid-rodriguez:
Yay!

hsbt:
@searls 🎉

bbatsov:
Funny enough I never noticed there was a `--rubocop` option to begin with. 😆

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Yeah, I had the same thought, but since `bundle gem` already writes configuration and it's never been an issue, this seems fine too indeed!

deivid-rodriguez:
It's not been out for long, it was released last July with 2.2.0.rc1, but we only released the final version in December. So no surprise people are still noticing it :sweat_smile:. Probably many people will first notice it when it's already deprecated :rofl:.

bbatsov:
Oh, well... 🤣

searls:
@deivid-rodriguez hi Deivid, I was hoping to demo this feature at RubyConf next week -- when do you think it might be next included in a release?

deivid-rodriguez:
Initial plan was to ship with bundler 2.3 sometime next month, but we've been shipping this kind of small enhancements in tiny releases, so if this is urgent for you I can ship it with next week's patch level release. Would next Monday work for you?

searls:
That would be awesome, thank you! And if it all goes sideways and breaks I'll be extra motivated to help fix it before my talk Tuesday lol

PR #4418, opened 2021-03-01
Fix gems sometimes being pulled from irrelevant sources

deivid-rodriguez:

The end user problem is that we have several issues with the sources where each dependency is pulled for resolution. In 2.2.10 I rushed too much and tried to fix all those issues at once, but that caused problems and had to be reverted. This time I want to make small improvements, one at a time.

In this PR, I improve the following:

Previously with the following Gemfile:

```ruby
source "https://source1" # doesn't include lib1 nor lib2

gem "lib1"

source "https://source2" do # includes lib1 and lib2
  gem "lib2"
end
```

bundler would resolve "fine" by pulling "lib1" from `https://source2`. I don't see how that would be expected. Bundler should look for "lib1" exclusively in the default source (`https://source1`) and print an error because the gem could not be found there. This PR fixes that issue.

The fix is to not use a "global merged source" that mixes all sources in the Gemfile for top-level dependencies. Instead, set the default source exclusively for them as one would expect. 

In addition to a spec for the fixed issue, I also added several specs to make sure that none of the refactorings I made here would make any of the previous issues in the bundler 2.2.10 release regress.

These changes were extracted from #4381, just to minimize the number of changes and try to introduce improvements one at a time.

### DIFF
--- a/.github/workflows/ruby-core.yml
+++ b/.github/workflows/ruby-core.yml
@@ -46,15 +46,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           path: rubygems/rubygems
-      - name: Test RubyGems
+      - name: Sync tools
         run: |
           ruby tool/sync_default_gems.rb rubygems
-          make -j2 -s test-all TESTS="rubygems --no-retry"
+          ruby tool/sync_default_gems.rb bundler
+        working-directory: ruby/ruby
+      - name: Test RubyGems
+        run: make -j2 -s test-all TESTS="rubygems --no-retry"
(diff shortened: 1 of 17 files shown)


simi:
Looks ok to me. I did some local testing (including production Gemfile).

deivid-rodriguez:
I removed the improvement in the error messages since it required some conditional logic depending on the rubygems version being run and I want to keep this small. I can improve that later.

deivid-rodriguez:
A few people tried this and it seems all good, so moving it forward!

Issue #4429, opened 2021-03-05
Three tests fail in Fedora packaging environment


voxik:
Trying to update rubygems package on Fedora, I observe following error:

~~~
  1) Failure:
TestKernel#test_gem_failing_inside_require_doesnt_cause_double_exceptions [/builddir/build/BUILD/rubygems-3.2.13/test/rubygems/test_kernel.rb:107]:
Expected: 1
  Actual: 0
~~~

Digging deeper, this is the actual issue:

~~~
$ ruby -e 'Dir.glob "./test/**/test_kernel.rb", &method(:require)' -- -n /test_gem_failing_inside_require_doesnt_cause_double_exceptions/
Ignoring json-2.5.1 because its extensions are not built. Try: gem pristine json --version 2.5.1
Run options: -n /test_gem_failing_inside_require_doesnt_cause_double_exceptions/ --seed 11053

# Running:


From: /builddir/build/BUILD/rubygems-3.2.13/test/rubygems/test_kernel.rb @ line 104 :

     99:       { "GEM_HOME" => Gem.paths.home },
    100:       *ruby_with_rubygems_in_load_path,
    101:       "-r",
    102:       "./activate.rb"
    103:     )
 => 104: binding.irb
    105:     load_errors = output.split("\n").select {|line| line.include?("Could not find") }
    106: 
    107:     assert_equal 1, load_errors.size
    108:   end
    109: 

irb(#<TestKernel:0x000055873bfe3180>):001:0> p output
"/builddir/build/BUILD/rubygems-3.2.13/tmp/test_rubygems_20210305-31-sxdh6q/activate.rb:1:in `<top (required)>': undefined method `gem' for main:Object (NoMethodError)\n\tfrom -:in `require'\n"
=> "/builddir/build/BUILD/rubygems-3.2.13/tmp/test_rubygems_20210305-31-sxdh6q/activate.rb:1:in `<top (required)>': undefined method `gem' for main:Object (NoMethodError)\n\tfrom -:in `require'\n"
irb(#<TestKernel:0x000055873bfe3180>):002:0> 
F

Finished in 2499.142949s, 0.0004 runs/s, 0.0004 assertions/s.

  1) Failure:
TestKernel#test_gem_failing_inside_require_doesnt_cause_double_exceptions [/builddir/build/BUILD/rubygems-3.2.13/test/rubygems/test_kernel.rb:107]:
Expected: 1
  Actual: 0

[...]

PR #4440, opened 2021-03-08
[WIP] Run tests with `--disable-gems`

voxik:
This PR is related to discussion in #4429. It makes the test suite execute with `--disable-gems` option. This should help to reduce the initial footprint of loaded libraries and reduce the risk of inadvertently testing system RubyGems instead of upstream version.

### DIFF
--- a/.github/workflows/ubuntu-rubygems.yml
+++ b/.github/workflows/ubuntu-rubygems.yml
@@ -36,6 +36,11 @@ jobs:
       - name: Run Test
         run: rake test
         if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
+      - name: Run Test (--disable-gems)
+        env:
+          RUBYOPT: --disable-gems
+        run: rake test
+        if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
       - name: Run Test (JRuby)
         run: JRUBY_OPTS=--debug rake test
         if: startsWith(matrix.ruby.name, 'jruby')


voxik:
This is one way to change the test suite. The other way would be to execute change the content of the default Rake task. That way, the test would be executed similarly for all platforms. Will try that ....

deivid-rodriguez:
I'd rather not change the way rubygems runs tests in general. Most users of rubygems don't have `RUBYOPT=--disable-gems`, so I'd like to keeping testing that case.

Instead, I would add an extra matrix entry with `RUBYOPT=--disable-gems` and make sure it also passes.

voxik:
This is the option I was talking about:

~~~diff
diff --git a/Rakefile b/Rakefile
index 8b1164559..9ca37b98f 100644
--- a/Rakefile
+++ b/Rakefile
@@ -27,11 +27,16 @@ task :git_hooks do
 end
 
 Rake::TestTask.new do |t|
+  ENV['RUBYOPT'] = '--disable-gems'
+
+  t.loader = :direct
+
   t.ruby_opts = %w[-w]
   t.ruby_opts << '-rdevkit' if RbConfig::CONFIG['host_os'].include?('mingw')
 
   t.libs << "test"
   t.libs << "bundler/lib"
+  t.libs << "."
 
 end
~~~

And I'll explore how to extend the matrix as you suggested

deivid-rodriguez:
Don't mind moving to the "direct" loader, but I'd rather massage the test file list so that it plays nice with adding "test" to the `$LOAD_PATH` rather than adding "." to the `$LOAD_PATH` as well.

So something like

```diff
 
+require "pathname"
+
 Rake::TestTask.new do |t|
+  t.loader = :direct
+
   t.ruby_opts = %w[-w]
   t.ruby_opts << '-rdevkit' if RbConfig::CONFIG['host_os'].include?('mingw')
 
   t.libs << "test"
   t.libs << "bundler/lib"
 
-  t.test_files = FileList['test/**/test_*.rb']
+  t.test_files = Dir.glob('test/**/test_*.rb').map {|p| Pathname.new(p).relative_path_from("test") }
 end
 
```
or maybe passing absolute paths directly?

hsbt:
Objection. `--disable-gems` is only debugging feature for ruby-core, not distribution vendor and ruby users. so It should not be use on primary test of CI.

deivid-rodriguez:
The idea is to not use it for primary testing of CI as per https://github.com/rubygems/rubygems/pull/4440#issuecomment-793023051. But I don't mind having green tests with `--disable-gems` and adding an extra CI matrix entry to test it, if that helps packagers with upgrading.

deivid-rodriguez:
Either 1 or 5 work for me, I don't have a strong opinion on either. I think @hsbt and @simi feel more strongly against this that I feel in favour. If this is the case, please go ahead and close this PR.

deivid-rodriguez:
I guess issues on ruby 3 should be fixed by https://github.com/ruby/rake/pull/379.

deivid-rodriguez:
By the way, even if we decide to move this forward, I don't see how this part of the commit message is relevant here:

> This should help to reduce the initial footprint of loaded libraries and reduce the risk of inadvertently testing system RubyGems instead of upstream version.

Our tests always use `-Ilib` as a ruby option, so both statements seem false to me. Local rubygems will always be loaded and the footprint of our tests will be the same regardless of `--disable-gems`.

voxik:
> By the way, even if we decide to move this forward, I don't see how this part of the commit message is relevant here:

Have you ever tried to do something like `$ mv lib/rubygems.rb{,.bak}`? I am quite sure the test suite will mostly work, because it is going to use the system version of `rubygems.rb` file. Similarly, if you have `RUBYOPT="-Isome/path -rrubygems"` I don't think the test suite is doing much to at least try to detect is it going to use some unexpected version of library.

I am not saying that `--disable-gems` is going to magically solve the issues, but I think that the test suite should be much more precise about the load paths.

voxik:
> > I completely disagree with @hsbt on `--disable-gems` being just debugging feature
> 
> It's not only my opinion, it's opinion of ruby-core team.

Thank you for opening the https://bugs.ruby-lang.org/issues/17684

voxik:
1. I have already removed the use of `--disable-gems` in the Fedora package. I'm here mostly because I have promised to deliver test in #4429 and I think it is good for the purpose I have stated in the commit message. If the result of this exercise will be that `--disable-gems` is deprecated, I might disagree but at least it will be documented.

2. There at least one more interesting failure in Fedora, which is why I said earlier that "I think that the test suite should be much more precise about the load paths." I cannot just drop everything into one issue, because it would be even bigger mess then it is now :/

deivid-rodriguez:
@hsbt If you're still against adding a secondary matrix entry to test `--disable-gems`, please go ahead and close. Otherwise @voxik can push the actual fix and get this green.

deivid-rodriguez:
@voxik Since there was no feedback from @hsbt feel free to move this forward.

hsbt:
I will handle this tomorrow.

hsbt:
I decided to close this. We should reduce the initial footprint or load time of rubygems at first. The `--disable-gems` way is the final solution for us.

PR #4456, opened 2021-03-14
Don't generate duplicate redundant sources in the lockfile

### LINKED ISSUE #4448: Duplicate Sections in Gemfile.lock
Bundler version `2.2.14` duplicates the "GEM remote: https://rubygems.org/" section in the Gemfile.lock, if it is specified twice in the `Gemfile`. This doesn't happen for Bundler version `2.2.13`.

I create a [this repository](https://github.com/ThomasKoppensteiner/bundler_duplicate_sections) containing some minimum working examples.

A similar behavior was discussed [here](https://github.com/rubygems/rubygems/issues/4380#issuecomment-779698508). 

I'm not sure if this causes any problems, but the `Gemfile.lock` becomes unnecessarily large and harder to review.

deivid-rodriguez:

Since we lock separate rubygems sources in the lockfile, it can happen that we generate duplicate redundant sources. It doesn't seem to cause specific issues, but it makes the lockfile harder to review.

Avoid this by adding a `uniq` call when generating the rubyems source list to use for writing the lockfile.

Fixes #4448.

### DIFF
--- a/bundler/lib/bundler/source_list.rb
+++ b/bundler/lib/bundler/source_list.rb
@@ -88,7 +88,7 @@ module Bundler
     def lock_sources
       lock_sources = (path_sources + git_sources + plugin_sources).sort_by(&:to_s)
       if disable_multisource?
-        lock_sources + rubygems_sources.sort_by(&:to_s)
+        lock_sources + rubygems_sources.sort_by(&:to_s).uniq
       else
         lock_sources << combine_rubygems_sources
       end
--- a/bundler/spec/lock/lockfile_spec.rb
+++ b/bundler/spec/lock/lockfile_spec.rb
@@ -641,6 +641,30 @@ RSpec.describe "the lockfile format" do
     G
   end
 
+  it "removes redundant sources" do
+    install_gemfile <<-G
+      source "#{file_uri_for(gem_repo2)}/"
+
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/source_list.rb
```diff
@@ -88,7 +88,7 @@ def get(source)
     def lock_sources
       lock_sources = (path_sources + git_sources + plugin_sources).sort_by(&:to_s)
       if disable_multisource?
-        lock_sources + rubygems_sources.sort_by(&:to_s)
+        lock_sources + rubygems_sources.sort_by(&:to_s).uniq
```
simi, line 91: I'm just wondering if duplication in `rubygems_sources` would not cause similar problem in any other place (like in `all_sources`). :thinking:

simi (approved):

bundler/lib/bundler/source_list.rb
deivid-rodriguez: I wondered the same thing and tried changing `rubygems_sources` too, but changing that method broke some other specs, so I cancelled that.

deivid-rodriguez:
I'll merge since @simi's comment seemed not a blocker given the approval!

simi:
yup, definitely not a blocker :balloon:

PR #4467, opened 2021-03-16
Filter outdated gems to the current version of Ruby

hosamaly:

Previously, `bundle outdated` would show the latest version of a gem, but it didn't check whether this gem could be used in the current version of Ruby. This resulted in some false-positives.

This PR filters out those gems that cannot be used on the current version of Ruby.

It's not foolproof, however, because we're only checking the requirements of the gem and not its dependencies. This means that we may still show versions that couldn't be installed on this version of Ruby. For example, the `factory_bot_rails` gem doesn't specify a required Ruby version, even though its dependency `factory_bot` specifies one.

## Is there more work to be done?

I didn't know how to write a test for this feature. Specifically, I didn't know how to adapt the code in `outdated_spec` to add a dependency that requires a specific version of Ruby. I'd appreciate your help with this.

### DIFF
--- a/bundler/lib/bundler/cli/outdated.rb
+++ b/bundler/lib/bundler/cli/outdated.rb
@@ -153,6 +153,12 @@ module Bundler
         if !current_spec.version.prerelease? && !options[:pre] && active_specs.size > 1
           active_specs.delete_if {|b| b.respond_to?(:version) && b.version.prerelease? }
         end
+
+        active_specs.delete_if do |spec|
+          spec.respond_to?(:required_ruby_version) &&
+            !spec.required_ruby_version.satisfied_by?(Gem.ruby_version)
+        end
+
         active_spec = active_specs.last
       end
 
--- a/bundler/spec/commands/outdated_spec.rb
+++ b/bundler/spec/commands/outdated_spec.rb
@@ -652,6 +652,41 @@ RSpec.describe "bundle outdated" do
     it_behaves_like "version update is detected"
   end
 
+  context "update available for a gem that requires a specific version of Ruby" do
+    before do
+      build_repo4 do
+        {
+          "1.0.0" => nil,
+          "2.0.0" => ">= #{Gem.ruby_version}",
+          "3.0.0" => "> #{Gem.ruby_version}",
+        }.each do |gem_version, required_ruby_version|
+          build_gem "ruby_support", gem_version do |s|
+            s.required_ruby_version = required_ruby_version
+          end
+        end
+      end
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo4)}"
+        gem "ruby_support", '1.0.0'
+      G
+    end
+
+    subject { bundle "outdated", :raise_on_error => false }
+
+    it_behaves_like "version update is detected"
+
+    it "detects the gem version supported on this version of Ruby" do
+      subject
+      expect(out).to include("2.0.0")
+    end
+
+    it "excludes the gem version that requires a more recent version of Ruby" do
+      subject
+      expect(out).not_to include("3.0.0")
+    end
+  end
+
   context "when on a new machine" do
     before do
       simulate_new_machine


deivid-rodriguez:
Thanks @hosamaly. For the spec, you can grep the specs for `required_ruby_version` and see how gems with this property are built for testing. If I recall correcty, it was something like:

```ruby
build_gem "foo" do |s|
  s.required_ruby_version = "< 3.1.dev"
end
```

hosamaly:
Thanks for the pointer, @deivid-rodriguez. I've now added some tests.

deivid-rodriguez:
@hosamaly I was thinking about this PR, and I'm not so sure about it.

As far as I understand, currently `bundle outdated` shows the latest version of every dependency unconditionally, no matter that `bundle update` can update directly to it or not. For example, say I have the following `Gemfile` and `Gemfile.lock`:

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

gem "rails", "~> 6.0.0"
```

```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (6.0.3.5)
      actionpack (= 6.0.3.5)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
    actionmailbox (6.0.3.5)
      actionpack (= 6.0.3.5)
      activejob (= 6.0.3.5)
      activerecord (= 6.0.3.5)
      activestorage (= 6.0.3.5)
      activesupport (= 6.0.3.5)
      mail (>= 2.7.1)
    actionmailer (6.0.3.5)
      actionpack (= 6.0.3.5)
      actionview (= 6.0.3.5)
      activejob (= 6.0.3.5)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 2.0)
    actionpack (6.0.3.5)
      actionview (= 6.0.3.5)
      activesupport (= 6.0.3.5)
      rack (~> 2.0, >= 2.0.8)
      rack-test (>= 0.6.3)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.2.0)
    actiontext (6.0.3.5)
      actionpack (= 6.0.3.5)
      activerecord (= 6.0.3.5)
      activestorage (= 6.0.3.5)
      activesupport (= 6.0.3.5)
      nokogiri (>= 1.8.5)
    actionview (6.0.3.5)
      activesupport (= 6.0.3.5)
      builder (~> 3.1)
      erubi (~> 1.4)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.1, >= 1.2.0)
    activejob (6.0.3.5)
      activesupport (= 6.0.3.5)
      globalid (>= 0.3.6)
    activemodel (6.0.3.5)
      activesupport (= 6.0.3.5)
    activerecord (6.0.3.5)
      activemodel (= 6.0.3.5)
      activesupport (= 6.0.3.5)
    activestorage (6.0.3.5)
      actionpack (= 6.0.3.5)
      activejob (= 6.0.3.5)
      activerecord (= 6.0.3.5)
      marcel (~> 0.3.1)
    activesupport (6.0.3.5)
      concurrent-ruby (~> 1.0, >= 1.0.2)
      i18n (>= 0.7, < 2)
      minitest (~> 5.1)
      tzinfo (~> 1.1)
      zeitwerk (~> 2.2, >= 2.2.2)
    builder (3.2.4)
    concurrent-ruby (1.1.8)
    crass (1.0.6)
    erubi (1.10.0)
    globalid (0.4.2)
      activesupport (>= 4.2.0)
    i18n (1.8.9)
      concurrent-ruby (~> 1.0)
    loofah (2.9.0)
      crass (~> 1.0.2)
      nokogiri (>= 1.5.9)
    mail (2.7.1)
      mini_mime (>= 0.1.1)
    marcel (0.3.3)
      mimemagic (~> 0.3.2)
    method_source (1.0.0)
    mimemagic (0.3.5)
    mini_mime (1.0.2)
    minitest (5.14.4)
    nio4r (2.5.7)
    nokogiri (1.11.2-x86_64-linux)
      racc (~> 1.4)
    racc (1.5.2)
    rack (2.2.3)
    rack-test (1.1.0)
      rack (>= 1.0, < 3)
    rails (6.0.3.5)
      actioncable (= 6.0.3.5)
      actionmailbox (= 6.0.3.5)
      actionmailer (= 6.0.3.5)
      actionpack (= 6.0.3.5)
      actiontext (= 6.0.3.5)
      actionview (= 6.0.3.5)
      activejob (= 6.0.3.5)
      activemodel (= 6.0.3.5)
      activerecord (= 6.0.3.5)
      activestorage (= 6.0.3.5)
      activesupport (= 6.0.3.5)
      bundler (>= 1.3.0)
      railties (= 6.0.3.5)
      sprockets-rails (>= 2.0.0)
    rails-dom-testing (2.0.3)
      activesupport (>= 4.2.0)
      nokogiri (>= 1.6)
    rails-html-sanitizer (1.3.0)
      loofah (~> 2.3)
    railties (6.0.3.5)
      actionpack (= 6.0.3.5)
      activesupport (= 6.0.3.5)
      method_source
      rake (>= 0.8.7)
      thor (>= 0.20.3, < 2.0)
    rake (13.0.3)
    sprockets (4.0.2)
      concurrent-ruby (~> 1.0)
      rack (> 1, < 3)
    sprockets-rails (3.2.2)
      actionpack (>= 4.0)
      activesupport (>= 4.0)
      sprockets (>= 3.0.0)
    thor (1.1.0)
    thread_safe (0.3.6)
    tzinfo (1.2.9)
      thread_safe (~> 0.1)
    websocket-driver (0.7.3)
      websocket-extensions (>= 0.1.0)
    websocket-extensions (0.1.5)
    zeitwerk (2.4.2)

PLATFORMS
  x86_64-linux

DEPENDENCIES
  rails (~> 6.0.0)

BUNDLED WITH
   2.3.0.dev
```

Then `bundle outdated` gives

```
Resolving dependencies....

Gem            Current  Latest  Requested  Groups
actioncable    6.0.3.5  6.1.3
actionmailbox  6.0.3.5  6.1.3
actionmailer   6.0.3.5  6.1.3
actionpack     6.0.3.5  6.1.3
actiontext     6.0.3.5  6.1.3
actionview     6.0.3.5  6.1.3
activejob      6.0.3.5  6.1.3
activemodel    6.0.3.5  6.1.3
activerecord   6.0.3.5  6.1.3
activestorage  6.0.3.5  6.1.3
activesupport  6.0.3.5  6.1.3
rails          6.0.3.5  6.1.3   ~> 6.0.0   default
railties       6.0.3.5  6.1.3
tzinfo         1.2.9    2.0.4
```

So running `bundle update tzinfo` or any other dependency will not update anything due to the constraint on Rails. In this case, it's easy to see what's holding things but it's not easy in general for bigger Gemfiles. In your case, it's the same case, but it's your current running ruby version holding things.

In my opinion, an improvement here would be to add a new `bundle outdated` mode that actually resolves like `bundle update` would do, and gives you the highest version of each dependency you can upgrade to.

But I'm not sure fixing this particular case in isolation would be an improvement or just introduce an inconsistency for no good reason. What's your use case here? Would maybe adding a new column with the ruby requirement of each top version also fix your use case?

hosamaly:
Thanks for your response, @deivid-rodriguez.

I'm currently working on upgrading an old application that has many direct and transitive dependencies. I'm trying to update the dependencies as much as possible while keeping the same version of Ruby. Once that's done, I intend to update the version of Ruby as well.

At the moment, running `bundle outdated` produces a long list of dependency updates, some of which have already been maxed out within this version of Ruby. Additionally, the suggested updated versions are, in many cases, incompatible with this version of Ruby.

I, too, would like to see a mode that resolves like `bundle update`, but that seemed too big of a task for me at the moment.

If adding columns to the output is the right way, then I'd suggest that we need two columns: one for the most recent version within the current version of Ruby and one for the most recent version ever.

I'm inclined to think that it may be easier to add a filter option that limits the output to those versions supported in the current version of Ruby. Perhaps something like `bundle outdated --filter-required-ruby`. What do you think about this?

deivid-rodriguez:
> At the moment, running bundle outdated produces a long list of dependency updates, some of which have already been maxed out within this version of Ruby. Additionally, the suggested updated versions are, in many cases, incompatible with this version of Ruby.

I can see how getting the highest possible versions you can upgrade to within the current ruby would be helpful. A new column with the required ruby version for the max version would give you a hint that you might've already maxed out the dependency, but doesn't give you a definitive answer to that question, so I see how it won't really help. 

I think I'd rather fix the issue in general, given that the currently proposed solution only works for required ruby version, not for other "conflicts", and even in that case, it only works in some cases as you pointed out in the PR body.

A `bundle outdated` mode that resolves just like `bundle update` might not be as hard as it seems. I think we would need to use the resolver exactly in the same way as `bundle update` uses it, and then loop through the result.

What I don't currently have is a good name for this mode, actually :sweat_smile:. Maybe we can add two columns as you suggested, one with the "absolute max version" and one with the "relative max version", and print the information by default.

Thoughts?

hosamaly:
Adding two columns sounds useful, and adding another mode that acts like a dry-run of `bundle update` is useful as well. However, I'm afraid I may not be able to work on either of them at the moment.

Nonetheless, adding a filter option doesn't sound too hard. What do you think about that?

deivid-rodriguez:
@hosamaly I had a closer look at this and it seems that there's already a `--strict` flag that does exactly what was suggested, so I think this PR is not necessary. Can you confirm?

hosamaly:
Indeed, @deivid-rodriguez, it looks like the `--strict` flag does this.

I had tried the flag but was misled when I saw some gems being suggested even though I had known that they don't support recent versions of Ruby. After double-checking, I can now see that the issue is with the gems themselves: their gemspecs still require a smaller version of Ruby than they support. Clearly, this isn't something that Bundler can help with.

Thanks a lot for your support. I'll now close this PR.

deivid-rodriguez:
No problem, thanks for taking the time to figure this out with me :+1:.

Issue #4462, opened 2021-03-16
only_update_to_newer_versions regression in v2.2 - bug or feature?


theflow:
### Describe the problem as clearly as you can

With https://github.com/rubygems/rubygems/pull/3692 included in Bundler 2.2 I'm seeing a change on how `only_update_to_newer_versions` works with indirect dependencies. It now does downgrade dependencies, even with that option set where before it wouldn't.

### Post steps to reproduce the problem

I've put an example Gemfile/Gemfile.lock here to reproduce: https://gist.github.com/theflow/e8690bd29ede796fad803b46679650cc

When it's only updating indirect dependencies, the `only_update_to_newer_versions` flag doesn't work anymore as it downgrades `unicode-display_width` here. With Bundler before v2.2 it wouldn't update anything at all.

```
$ BUNDLE_ONLY_UPDATE_TO_NEWER_VERSIONS=true bundle update

Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Using ansi 1.5.0
Using bundler 2.2.13
Using docile 1.3.5
Using simplecov-html 0.12.3
Using simplecov_json_formatter 0.1.2
Using unicode-display_width 1.7.0 (was 2.0.0)
Using terminal-table 3.0.0 (was 1.6.0)
Using tabularize 0.2.10
Using simplecov 0.21.2
Using simplecov-console 0.9.1
Bundle updated!
```

It's clear from that PR that this is somewhat intentional and I'm not sure this is fixable as we have two competing behaviors we want here. So not really sure if that's a bug or a feature 🙃

deivid-rodriguez:
Hi @theflow, that's interesting.

I see how you'd like `bundle update` not to change anything in that particular case: updating one transitive dependency and downgrading another one doesn't seem like an improvement so better leave things as they are. On the other hand, my PR looks correct too since bundler should do what's necessary with your transitive dependencies to be able to find a resolution for your direct dependencies, even if that requires downgrading a transitive dependency.

I guess the best solution would be to try the previous implementation first (do not allow downgrading anything), and only fallback to the current implementation if no valid resolution was found.

theflow:
This is a case from Depfu, where we never want to have anything downgraded, but where we are also sure we don't have the use case you were fixing in #3692.

So I don't know if this is really a problem for normal users, you're the better judge of that.

I'm also ok with leaving the current behavior as solving this might not be worth the complexity. I can easily monkey-patch it for our case.

[...]

PR #4473, opened 2021-03-19
Never downgrade top level gems when running `bundle update`

deivid-rodriguez:

Today I run `bundle update` in my library and it downgraded a top-level dependency. In my opinion, nobody expects this, so this is a bug.

My fix is to enable to `only_update_to_newer_versions` feature flag by default. The behaviour can still be further refined (see #4462), but enabling the behaviour by default is still better than current behaviour.
 

hsbt:
👍  I agreed to enable this always.

Issue #4425, opened 2021-03-03
rubygems-update 3.2.13 breaks sorbet installation on alpine


garybethea:
My current problem is that using `rubygems-update 3.2.13`, running `bundle install` on Alpine fails with the error, `Could not find sorbet-static-0.5.6317-universal-darwin-14 in any of the sources` even though it is present in `Gemfile.lock`.

This issue is related to:

  - [ ] Network problems
  - [ X] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
Platform: ruby:2.6.5-alpine
rubygems-update: 3.2.13
bundler: 2.1.4
```

### Output
```
#!/bin/bash -eo pipefail
aws_region="HIDDEN"
image_prefix="HIDDEN"

.circleci/build_container \
  --image-prefix "${image_prefix}" \
  --git-sha "${CIRCLE_SHA1}" \

Login Succeeded
Sending build context to Docker daemon  8.201MB
Step 1/22 : FROM ruby:2.6.5-alpine as base
 ---> 59d4ec426ed6
Step 2/22 : LABEL maintainer "HIDDEN@HIDDEN.com"
 ---> Using cache
 ---> ae9448a3f518
Step 3/22 : ARG BUNDLE_GITHUB__COM
 ---> Using cache
 ---> b3cb033adb4e
Step 4/22 : ARG BUNDLER_VERSION=2.1.4
 ---> Using cache
 ---> 061715bc2d57
Step 5/22 : ARG APP_PATH=/app
 ---> Using cache
 ---> b7f859aa0ffa
Step 6/22 : ENV BUNDLER_VERSION ${BUNDLER_VERSION}
 ---> Using cache
 ---> c268d0e3dfa6
Step 7/22 : RUN mkdir -p ${APP_PATH}
 ---> Using cache
 ---> 8e723f4ca6d9
Step 8/22 : WORKDIR ${APP_PATH}
 ---> Using cache
 ---> ca1144a4e2d2
Step 9/22 : RUN apk upgrade && apk add --no-cache   build-base   postgresql-dev   git   less   nodejs-current   tzdata   bash   snappy-dev   libexecinfo   automake   autoconf   libtool
 ---> Using cache
 ---> c2662b416907
Step 10/22 : COPY Gemfile* ${APP_PATH}/
 ---> Using cache
 ---> 7282b1f406f0
[...]

PR #4488, opened 2021-03-24
Backport non gnu libc linux support from RubyGems

lloeki:

See #3174, #4082, #4425. 

Implement the non-ideal but pragmatic behaviour we agreed upon:

- "the previous state of things" i.e `nil` version in gems is a wildcard (i.e a `linux` gem it can be picked by Rubies on `linux` and `linux-musl`)
- with the exception that if there's a `linux-musl` variant, it should always be preferred on Alpine (i.e a `linux-musl` Ruby)
- and a mirror constraint that on a `linux` Ruby it should never pick `linux-musl` gems

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -237,6 +237,32 @@ module Gem
     MINGW = Gem::Platform.new("x86-mingw32")
     X64_MINGW = [Gem::Platform.new("x64-mingw32"),
                  Gem::Platform.new("x64-mingw-ucrt")].freeze
+
+    if Gem::Platform.new("x86_64-linux-musl") === Gem::Platform.new("x86_64-linux")
+      remove_method :===
+
+      def ===(other)
+        return nil unless Gem::Platform === other
+
+        # universal-mingw32 matches x64-mingw-ucrt
+        return true if (@cpu == "universal" || other.cpu == "universal") &&
+                       @os.start_with?("mingw") && other.os.start_with?("mingw")
+
+        # cpu
+        ([nil,"universal"].include?(@cpu) || [nil, "universal"].include?(other.cpu) || @cpu == other.cpu ||
+        (@cpu == "arm" && other.cpu.start_with?("arm"))) &&
+
+          # os
+          @os == other.os &&
+
+          # version
+          (
+            (@os != "linux" && (@version.nil? || other.version.nil?)) ||
+            (@os == "linux" && ((@version.nil? && ["gnu", "musl"].include?(other.version)) || (@version == "gnu" && other.version.nil?))) ||
+            @version == other.version
+          )
+      end
+    end
   end
 
   Platform.singleton_class.module_eval do
--- a/bundler/spec/resolver/platform_spec.rb
+++ b/bundler/spec/resolver/platform_spec.rb
@@ -82,6 +82,79 @@ RSpec.describe "Resolving platform craziness" do
     should_resolve_as %w[foo-1.0.0-x64-mingw32]
   end
 
+  describe "on a linux platform", :rubygems => ">= 3.1.0.pre.1" do
+    # Ruby's platform is *-linux => platform's libc is glibc, so not musl
+    # Ruby's platform is *-linux-musl => platform's libc is musl, so not glibc
+    # Gem's platform is *-linux => gem is glibc + maybe musl compatible
+    # Gem's platform is *-linux-musl => gem is musl compatible but not glibc
+
+    it "favors the platform version-specific gem on a version-specifying linux platform" do
+      @index = build_index do
+        gem "foo", "1.0.0"
+        gem "foo", "1.0.0", "x86_64-linux"
+        gem "foo", "1.0.0", "x86_64-linux-musl"
+      end
+      dep "foo"
+      platforms "x86_64-linux-musl"
+
+      should_resolve_as %w[foo-1.0.0-x86_64-linux-musl]
+    end
+
+    it "favors the version-less gem over the version-specific gem on a gnu linux platform" do
+      @index = build_index do
+        gem "foo", "1.0.0"
+        gem "foo", "1.0.0", "x86_64-linux"
+        gem "foo", "1.0.0", "x86_64-linux-musl"
+      end
+      dep "foo"
+      platforms "x86_64-linux"
+
+      should_resolve_as %w[foo-1.0.0-x86_64-linux]
+    end
+
+    it "ignores the platform version-specific gem on a gnu linux platform" do
+      @index = build_index do
+        gem "foo", "1.0.0", "x86_64-linux-musl"
+      end
+      dep "foo"
+      platforms "x86_64-linux"
+
+      should_not_resolve
+    end
+
+    it "falls back to the platform version-less gem on a linux platform with a version" do
+      @index = build_index do
+        gem "foo", "1.0.0"
+        gem "foo", "1.0.0", "x86_64-linux"
+      end
+      dep "foo"
+      platforms "x86_64-linux-musl"
+
+      should_resolve_as %w[foo-1.0.0-x86_64-linux]
+    end
+
+    it "falls back to the ruby platform gem on a gnu linux platform when only a version-specifying gem is available" do
+      @index = build_index do
+        gem "foo", "1.0.0"
+        gem "foo", "1.0.0", "x86_64-linux-musl"
+      end
+      dep "foo"
+      platforms "x86_64-linux"
+
+      should_resolve_as %w[foo-1.0.0]
+    end
+
+    it "falls back to the platform version-less gem on a version-specifying linux platform and no ruby platform gem is available" do
+      @index = build_index do
+        gem "foo", "1.0.0", "x86_64-linux"
+      end
+      dep "foo"
+      platforms "x86_64-linux-musl"
+
+      should_resolve_as %w[foo-1.0.0-x86_64-linux]
+    end
+  end
+
   it "takes the latest ruby gem if the platform specific gem doesn't match the required_ruby_version" do
     @index = build_index do
       gem "foo", "1.0.0"
--- a/bundler/spec/support/indexes.rb
+++ b/bundler/spec/support/indexes.rb
@@ -33,6 +33,10 @@ module Spec
       Bundler::Resolver.resolve(deps, source_requirements, *args)
     end
 
+    def should_not_resolve
+      expect { resolve }.to raise_error(Bundler::GemNotFound)
+    end
+
     def should_resolve_as(specs)
       got = resolve
       got = got.map(&:full_name).sort


lloeki:
So, this is WIP: previous work + messy `SAVEPOINT` commit that I'll clean up later, but I needed to push some base to discuss upon.

I did it TDD/BDD by adding rubygems and bundler tests that make sense WRT the behaviour, and made them pass.

There's still some strange behaviour that I hacked in/out, so I'm not entirely sure I did not create breakage at this stage on other platforms. I also saw some strange things I can't seem to wrap my head around: most likely I'm missing something but they could be legacy code, or bugs.

I'll continue working on this as I figure things out.

CONTRIBUTING.md
```diff
@@ -42,8 +42,8 @@ To run commands like `gem install` from the repo:
 To run bundler test:
 
     $ cd bundler
-    $ bin/rake spec:deps
-    $ bin/rspec spec
+    $ RUBYOPT='-I../lib' bin/rake spec:deps
+    $ RUBYOPT='-I../lib' bin/rspec spec
```
lloeki, line 46: Seems like by default the repo bundler runs against Ruby's rubygems instead of the repo one, so I had to do this to pick my changes.

Just a note for now but feels worthy of adding to the doc.

bundler/doc/development/SETUP.md
```diff
@@ -32,6 +32,10 @@ To work on Bundler, you'll probably want to do a couple of things:
 
         $ alias dbundle='/path/to/bundler/repo/bin/bundle'
 
+## Jointly developing on bundler and rubygems
+
+RUBYOPT='-I../lib' bin/rake spec
```
lloeki, line 37: Same as above, just a note for now but feels worthy of adding to the doc.

bundler/lib/bundler/gem_helpers.rb
```diff
@@ -55,7 +56,8 @@ def select_best_platform_match(specs, platform)
     class PlatformMatch
       def self.specificity_score(spec_platform, user_platform)
         return -1 if spec_platform == user_platform
-        return 1_000_000 if spec_platform.nil? || spec_platform == Gem::Platform::RUBY || user_platform == Gem::Platform::RUBY
+        return 1_000_000 if spec_platform.nil? || spec_platform == Gem::Platform::RUBY || user_platform == Gem::Platform::RUBY # <= last one is weird, not sure it should happen
```
lloeki, line 59: Here I'm not seeing when `user_platform == Gem::Platform::RUBY` is a legit case:

- for `ruby` gem and runtime platform, it would be covered by `spec_platform == user_platform` 
- it would just let through any non-`ruby` gem on `ruby` runtime platform†
- it would just let through any non-`ruby` gem on any runtime platform, given `spec_group_ruby` adds a `ruby` runtime platform processing step

So I thought, `1_000_000` is ultra low priority (IIUC), but maybe `return nil` when it's a definitive not-a-match-at-all, then `.compact!` the `sort_by` result would give only those `sorted_matching` that truly have a chance to match, without the ones we are sure we do not want.

† assuming there is any before `specificity_score` is called, which seems like it should often be the case, except a second question seems to ram some into this.

bundler/lib/bundler/match_platform.rb
```diff
@@ -11,11 +11,16 @@ def match_platform(p)
...
     def self.platforms_match?(gemspec_platform, local_platform)
+      # p({ gem: gemspec_platform, local: local_platform })
       return true if gemspec_platform.nil?
-      return true if Gem::Platform::RUBY == gemspec_platform
-      return true if local_platform == gemspec_platform
+      return true if gemspec_platform == Gem::Platform::RUBY
+      return true if gemspec_platform == local_platform
       gemspec_platform = Gem::Platform.new(gemspec_platform)
-      return true if GemHelpers.generic(gemspec_platform) === local_platform
+      #return true if GemHelpers.generic(gemspec_platform) === local_platform
```
lloeki, line 19: That's the second question I had. I added a test to make sure that without a `ruby` nor a `linux` gem, a `linux-musl` gem by itself would never get picked by a `linux` runtime, but the resolver still returned the `linux-musl` gem!

It was really unclear why, so I went through the whole resolver thing to find that it's not the `linux` platform `spec_group` but the additional `spec_group_ruby` that causes this, because:

```
pry> Bundler::GemHelpers.generic(Gem::Platform.new('x86_64-linux-musl'))
=> "ruby"
```

Which equals `local_platform from the resolver's `spec_group_ruby` in `search_for`.

This means that in the results, `linux-musl` still appears for `linux` through the added `ruby` candidate spec group, and since `specificity_score` doesn't remove it, it bubbles up to the final result even though it's absent from the platform-specific `spec_group`...

The possible leakage also applies in other cases (such as `linux` appearing for `linux-musl` through `ruby`) but the end result is so that this is equivalent to what we want (and what appears in the `spec_group`). If I got this right, I suspect in most cases the problem is hidden this way but it could cause subtle issues here and there.

lib/rubygems/platform.rb
```diff
@@ -146,10 +146,17 @@ def hash # :nodoc:
...
   #
   # Additionally, the platform will match if the local CPU is 'arm' and the
   # other CPU starts with "arm" (for generic ARM family support).
+  #
+  # Of note, this method is not commutative. Indeed the OS 'linux' has a
+  # special case: the version is the libc name, yet while "no version" stands
+  # as a wildcard for a binary gem platform (as for other OSes), for the
+  # runtime platform "no version" stands for 'gnu'. To be able to disinguish
+  # these, the method receiver is the gem platform, while the argument is
+  # the runtime platform.
```
lloeki, line 159: Given the goal we had, and the fact that `===` is not commutative already in many places (e.g class membership) this is how I implemented the constraint.

Therefore calls to `Gem::Platform#===` should always be in that order: `gem_platform === runtime_platform`, turns out this was already the case almost everywhere.

test/rubygems/test_gem_platform.rb
```diff
@@ -261,6 +264,42 @@ def test_nil_cpu_arch_is_treated_as_universal
...
+    x86_linux_musl = Gem::Platform.new 'i686-linux-musl'
+    x86_linux_uclibc = Gem::Platform.new 'i686-linux-uclibc'
+
+    # a naked linux runtime is implicit gnu, as it represents the common glibc-linked runtime
+    assert(x86_linux === x86_linux_gnu, 'linux =~ linux-gnu')
+    assert(x86_linux_gnu === x86_linux, 'linux-gnu =~ linux')
+
+    # musl and explicit gnu should differ
+    # refute(x86_linux_gnu === x86_linux_musl, 'linux-gnu =~ linux-musl')
+    # refute(x86_linux_musl === x86_linux_gnu, 'linux-musl =~ linux-gnu')
```
lloeki, line 287: It might be worth thinking about introducing `linux-gnu` someday, for those good willed maintainers that know their binary is definitely not musl compatible but with no `linux-musl` published (yet, or ever).

test/rubygems/test_gem_platform.rb
```diff
@@ -261,6 +264,42 @@ def test_nil_cpu_arch_is_treated_as_universal
...
+    refute(x86_linux_uclibc === x86_linux_musl, 'linux-uclibc =~ linux-musl')
+    refute(x86_linux_musl === x86_linux_uclibc, 'linux-musl =~ linux-uclibc')
+
+    # musl host runtime accepts libc-generic or statically linked gems...
+    assert(x86_linux === x86_linux_musl, 'linux =~ linux-musl')
+    # ...but implicit gnu runtime generally does not accept musl-specific gems
+    refute(x86_linux_musl === x86_linux, 'linux-musl =~ linux')
+
+    # other libc are not glibc compatible
+    # refute(x86_linux === x86_linux_uclibc, 'linux =~ linux-uclibc')
```
lloeki, line 299: The case of these non-musl, non-gnu libc that are known not to be compatible at all with glibc binaries (so zero chance of `linux` working) remains unsolved here. Anecdotal I know, but one day musl was anecdotal as well.

CONTRIBUTING.md
deivid-rodriguez: Yes, thanks @lloeki, this should be documented, definitely. What I usually do is `export RGV=..`, which is what tests do, and make sure all development binstubs automatically run pointing to the correct copy.

bundler/lib/bundler/gem_helpers.rb
deivid-rodriguez: I _think_ it can be assumed that it's only for sorting but shouldn't remove any specs. Filtering out specs that should never match should be done previously to using this method.

Just to try to give you some context from what I recall from fighting against these issues, this method is used in a few different situations, where "user_platform" doesn't always mean "runtime platform" but a platform we are resolving dependencies or materializing a resolution against. Your reasoning might still make sense, but just wanted to clarify that. For example:

* We might be on linux, but resolve against java. This is what happens when you do `bundle lock --add-platform java`, or when you have a lockfile including JAVA in the list of platforms. In those cases, a resolution is found that not only matches the running platform, but also JAVA.

* When I say materializing I mean picking all the specs in a resolution (if recorded, in the lockfile) and selecting only those which are relevant for a specific platform. For example, the lockfile can include specs for different platforms at the same time, like `foo-1.0` and `foo-1.0-x86_64-linux` if the lockfile includes both say `x86_64-linux` and `ruby` in the `PLATFORMS` sections. It can even include specs that are only relevant for some platforms, for example, if `foo-1.0-x86_64-linux` depends on `bar`, but `foo-1.0` does not depend on `bar`.

Also, and still not sure whether it's relevant to these issues, but just to give some context on this "ruby spec_group". Since different platform specific variants can have different dependencies, the resolver always adds this fallback spec group, so that it can consider all possibilities. So in the example I gave above, when resolving for `x86_64-linux`, one resolution possibility would be the `<foo-1.0, x86_64_linux>` spec group, an another one would be the `<foo-1.0, ruby>` fallback spec group, and they need to be considered as different possibilities because one of them brings in the `bar` dependency requirement while the other one does not.

Not sure if all of the above answers the question but hopefully brings some light.

bundler/lib/bundler/match_platform.rb
deivid-rodriguez: Some platforms are "forced" to not have a generic platform (i.e., only match their own platform).

https://github.com/rubygems/rubygems/blob/5cfad4bd6661f2370bff0aaef1e81802420d2cef/bundler/lib/bundler/gem_helpers.rb#L6-L14

Perhaps our constraints for `x86_64-linux-musl` match this case.

deivid-rodriguez:
@lloeki I left a few comments, hopefully they can help :pray:.

nightpool:
After trying to upgrade to mini_racer 0.4.0 and libv8-node this weekend, I found that this issue has broken my Heroku deployment:

```
remote:        Fetching libv8-node 15.14.0.1 (x86_64-linux-musl)
remote:        Installing libv8-node 15.14.0.1 (x86_64-linux-musl)
....
remote:        Bundle completed (183.45s)
remote:        Cleaning up the bundler cache.
remote:        Could not find libv8-node-15.14.0.1 in any of the sources
remote: -----> Writing config/database.yml to read from DATABASE_URL
remote: -----> Detecting rake tasks
remote:        Could not detect rake tasks
remote:        ensure you can run `$ bundle exec rake -P` against your app
remote:        and using the production group of your Gemfile.
remote:        /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:91:in `block in materialize': Could not find libv8-node-15.14.0.1 in any of the sources (Bundler::GemNotFound)
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `map!'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `materialize'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:170:in `specs'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:237:in `specs_for'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:226:in `requested_specs'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:108:in `block in definition_method'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:20:in `setup'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler.rb:107:in `setup'
remote:        	from /tmp/build_3933b252/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/setup.rb:20:in `<top (required)>'
remote:        	from /tmp/build_3933b252/vendor/ruby-2.5.8/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
remote:        	from /tmp/build_3933b252/vendor/ruby-2.5.8/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
remote:        	from /tmp/build_3933b252/vendor/bundle/bin/rake:27:in `<main>'
```

At least, I'm assuming that this is the same issue? Is there any known workaround?

lloeki:
Might be, it fetches linux-musl when (I assume) Heroku is glibc-based?

I really have to find some time to get back on this and finish the PR.

nightpool:
Thanks @lloeki. I'm not sure why it would be giving that specific error though, considering that there aren't any issues at build time. I also tried pushing the precompiled x86_64 gem to my github rubygems server ([here](https://github.com/nightpool/libv8-node/packages/811334) and got a very similar error:

```
remote:        Fetching libv8-node 15.14.0.1
remote:        Installing libv8-node 15.14.0.1
[...]
remote:        Bundle completed (182.41s)
remote:        Cleaning up the bundler cache.
remote:        Could not find libv8-node-15.14.0.1 in any of the sources
remote: -----> Writing config/database.yml to read from DATABASE_URL
remote: -----> Detecting rake tasks
remote:        Could not detect rake tasks
remote:        ensure you can run `$ bundle exec rake -P` against your app
remote:        and using the production group of your Gemfile.
remote:        /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:91:in `block in materialize': Could not find libv8-node-15.14.0.1 in any of the sources (Bundler::GemNotFound)
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `map!'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/spec_set.rb:85:in `materialize'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:170:in `specs'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:237:in `specs_for'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/definition.rb:226:in `requested_specs'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:108:in `block in definition_method'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/runtime.rb:20:in `setup'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler.rb:107:in `setup'
remote:        	from /tmp/build_4900502f/vendor/bundle/ruby/2.5.0/gems/bundler-1.17.3/lib/bundler/setup.rb:20:in `<top (required)>'
remote:        	from /tmp/build_4900502f/vendor/ruby-2.5.8/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
remote:        	from /tmp/build_4900502f/vendor/ruby-2.5.8/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
remote:        	from /tmp/build_4900502f/vendor/bundle/bin/rake:27:in `<main>'
```

bundler/lib/bundler/gem_helpers.rb
lloeki: Yes it does! It was especially helpful as I had to dig in deep to [produce a patch](https://github.com/rubygems/rubygems/issues/3174#issuecomment-963166985) specifically for bundler 1.17.

lib/rubygems/platform.rb
lloeki: [While doing this](https://github.com/rubygems/rubygems/issues/3174#issuecomment-963166985) it turns out I found other places where this test is called in a certain order, so I [reversed the order](https://gist.github.com/lloeki/12dcf61324f64a2fa1e8a8b2109c1f00#file-musl-patch-L54).

lloeki:
I'm going to try to redo this based on [the rubygems 3.0 + bundler 1.17 patch I produced](https://gist.github.com/lloeki/12dcf61324f64a2fa1e8a8b2109c1f00#file-musl-patch), which seems to produce good results. Of course only the specs will tell if it fully works.

wwahammy:
@lloeki thanks for all the work so far on this! Is there anything I can do to help get this across the finish line? I'm certainly not familiar with the bundler source but I'm happy to help in any way I can.

virusman:
@lloeki Any hope for this to get merged?

lloeki:
@virusman not as is.

FYI I'm rethinking how it should work and what the expectations of the `platform.version` component should be so that it doesn't break anything.

 Do you have a need for a specific gem outside of `mini_racer`? The more information I have about how people use multiple platforms, the better.

vbabenkoru:
Yeah I need this for mini_racer… The recent changes in rubygems and bundler broke it, and I can’t even deploy the current version of my app without downgrading rubygems and bundler because of that. No workarounds I’ve seen have worked for bundler.

ronaldtse:
We're also encountering this issue with 4 other gems:
* https://github.com/lutaml/expressir/issues/106
* https://github.com/metanorma/emf2svg-ruby/issues/17
* https://github.com/fontist/extract_ttc/issues/7
* https://github.com/fontist/ffi-libarchive-binary/issues/11

On Alpine with Ruby 3.1 (3.1-alpine at https://hub.docker.com/_/ruby) we are seeing the installation of a `libc` gem (`x86_64-linux`) but it fails on inclusion due to `musl` being unable to load a `libc` compiled library.

The only two options we have as gem maintainers are:
1. The `x86_64-linux` platform gem works natively on `musl`. As @lloeki says it can be impractical.
2. We publish platform gems for `x86_64-linux-musl` so that on Alpine the `x86_64-linux-musl` is always used.

cc: @maxirmx @alexeymorozov @camobap

hsbt:
At first, the current ruby language is not supported alpine linux and musl officially. The ruby core team does not testing them with https://rubyci.org/

ronaldtse:
Thanks @hsbt for the clarification! It was confusing whether it is supported or not given that the Ruby docker images include one built on Alpine... I guess it's a good time to investigate.

hsbt:
I understood alpine image was published by docker team. But the ruby core team didn't work together it. we have our official docker image only https://github.com/ruby/ruby-docker-images/pkgs/container/ruby .

This means the native extension built by musl may not works well with the ruby language. I highly recommend to use debian based image or official docker image published by the ruby core team.

ronaldtse:
I wasn't even aware that the official Ruby docker images are different from those from the Docker team. Thank you for bringing this awareness here @hsbt !

bundler/lib/bundler/match_platform.rb
deivid-rodriguez: @lloeki This line has now been removed, I think it was indeed incorrect. So this should no longer be an issue!

deivid-rodriguez:
Good news, it seems that #5495 indeed unblocked this PR and it's now all green :)

I'll be splitting it into two PRs, one for RubyGems, and one for Bundler, since they are separate improvements after all. But other than that this will be ready soon!

deivid-rodriguez:
I fixed one `gem install` bug still installing must on non-musl Linux at https://github.com/rubygems/rubygems/pull/4488/commits/bce31848e64fa1ee1836e222690b216c89f3c8be, and also implemented the extra matching rules mentioned by @lloeki and left out as commented out tests, although I'm not sure if it's getting ahead of ourselves too much.

lloeki:
So nice, thanks @deivid-rodriguez.

Since comparing a gem platform and the actual ruby platform is asymmetric, I was wondering whether we should introduce a `Ruby::Platform` or `Gem::RubyPlatform` to distinguish from `Gem::Platform`.

deivid-rodriguez:
> Since comparing a gem platform and the actual ruby platform is asymmetric, I was wondering whether we should introduce a Ruby::Platform or Gem::RubyPlatform to distinguish from Gem::Platform.

Yes, something feels a bit off now that you have to be very careful when using `Gem::Platform#===` since it's no longer commutative. We should improve it, and your idea seems interesting.

deivid-rodriguez:
> Yes, something feels a bit off now that you have to be very careful when using Gem::Platform#=== since it's no longer commutative. We should improve it, and your idea seems interesting.

In any case, I was just noticing that it's already not commutative, so I think we can move this forward and reschedule extracting `Gem::RubyPlatform` (maybe could be named `Gem::Platform::Local` or `Gem::Platform::Runtime`).

lib/rubygems/platform.rb
deivid-rodriguez: As already observed, `Gem::Platform#===` is already not commutative so your implementation seems fine for the time being!

test/rubygems/test_gem_platform.rb
deivid-rodriguez: I did introduce checks for this but I think we may be getting ahead of ourselves, I'm not fully sure whether we should include it with this patch or not.

test/rubygems/test_gem_platform.rb
deivid-rodriguez: I fixed this one too! Never seen these in the wild, but if they're not compatible it seems good to make sure we never match "standard" linux.

deivid-rodriguez:
I merged the RubyGems part of this separately at #5852, and this PR now only backports the feature to Bundler so that this works consistently across RubyGems version. I only backported all the way back to RubyGems 3.1.0.pre.1 because trying to backport further required further changes and was a mess.

ronaldtse:
Thank you @deivid-rodriguez for #5852 ! We look forward to seeing this merged too and getting to work with non-glibc platforms!

deivid-rodriguez:
Thank @lloeki, he did all the heavy lifting! :)

lloeki:
You're welcome, there were so many users impacted by this I could don't let that one slide :)

ronaldtse:
Of course a huge thank you to @lloeki ! Looking forward to publishing platform gems for musl...

ronaldtse:
Thank you for merging!

deivid-rodriguez:
Unfortunately, this PR created again a regression in musl platforms 😬. I'll try hard to fix it to hopefully avoid having to revert the feature.

Issue #4370, opened 2021-02-11
Adding platform without version does not add current-platform version of libv8 to specs on macOS


owst:
We're in the process of upgrading from Bundler 1.17.2 to 2.2.9, and as part of that process, I'm attempting to use the improved platform support, but have hit a problem with macOS in that the current-platform specific gem doesn't appear to be added when adding a version-less platform to the lock file. 

Starting with the following Gemfile/Gemfile.lock on my mac, with:
```
$ gem env platform
ruby:x86_64-darwin-18
```

```ruby
source 'https://rubygems.org'

gem 'therubyracer'
```

```
GEM
  remote: https://rubygems.org/
  specs:
    libv8 (3.16.14.19)
    ref (2.0.0)
    therubyracer (0.12.3)
      libv8 (~> 3.16.14.15)
      ref

PLATFORMS
  ruby

DEPENDENCIES
  therubyracer

BUNDLED WITH
   1.17.2
```

I first updated to use Bundler 2 (2.2.9) with `gem install bundler` and then `bundle update --bundler` - the only change was to `BUNDLED WITH` which became 2.2.9 as expected. Then I tried to add a versionless mac platform to the lockfile with `bundle lock --add-platform x86_64-darwin`, which succeeded, but the diff doesn't look quite right:
```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index d3e1ab0..2e6acdc 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,9 @@ GEM
   remote: https://rubygems.org/
   specs:
     libv8 (3.16.14.19)
+    libv8 (3.16.14.19-x86_64-darwin-15)
+    libv8 (3.16.14.19-x86_64-darwin-16)
+    libv8 (3.16.14.19-x86_64-darwin-17)
     ref (2.0.0)
     therubyracer (0.12.3)
       libv8 (~> 3.16.14.15)
@@ -9,6 +12,7 @@ GEM
 
 PLATFORMS
   ruby
+  x86_64-darwin
 
 DEPENDENCIES
   therubyracer
```

specifically, there is no `libv8 (3.16.14.19-x86_64-darwin-18)` (my current platform version). Therefore, a subsequent `bundle install` falls back to `Installing libv8 3.16.14.19 with native extensions`, which I noticed because compiling v8 fails, due to an unrelated problem.

If I manually insert the missing version into the lock file, it _is_ installed without issue. 

[...]

PR #4497, opened 2021-03-26
Fix incorrect logic for filtering metadata matching candidates

deivid-rodriguez:

The problem is that when bundling a bundler 2.1 lockfile (without platforms locked), we want to resolve platforms at materialization time for backwards compatibility. However, since https://github.com/rubygems/rubygems/pull/4449, the logic to do this is incorrect and can lead to the standard variant being preferred over a platform specific variant as explained at https://github.com/rubygems/rubygems/issues/4370#issuecomment-807952686.

The previous logic was discarding both specs coming from the old dependency API (`RemoteSpecification`'s), and installed specifications (`StubSpecification`, which inherit from `RemoteSpecification`).

We don't actually want to discard any of those. In particular, discarding already installed specifications can cause another not platform specific variant to be chosen in the end, which is unexpected.

### DIFF
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -74,9 +74,9 @@ module Bundler
           MatchPlatform.platforms_match?(spec.platform, platform_object)
         end
         installable_candidates = same_platform_candidates.select do |spec|
-          !spec.is_a?(RemoteSpecification) &&
-            spec.required_ruby_version.satisfied_by?(Gem.ruby_version) &&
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
:information_source: In my opinion, and that's the reason for this bug since I didn't know it until now, it's very unexpected that `StubSpecification`, which represents an installed specification, inherits from `RemoteSpecification` which represents a specification coming from the old dependency API. They do have similar logic regarding lazily loading the underlying full specification on demand, but other than that they are completely different classes, so I think they should not inherit from each other. But that's a bigger refactoring and I wanted to focus on fixing the bug.

PR #4498, opened 2021-03-28
Enable bundler-cache for GitHub Actions template

andyw8:
This seems like a useful default.

Details: https://github.com/actions/cache/blob/main/examples.md#ruby---bundler

### DIFF
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -11,6 +11,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: <%= RUBY_VERSION %>
+        bundler-cache: true
     - name: Run the default task
       run: |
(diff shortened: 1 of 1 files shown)


bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -11,6 +11,7 @@ jobs:
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: <%= RUBY_VERSION %>
+        bundler-cache: true
```
deivid-rodriguez, line 14: Yes! But then the `gem install` && `bundle install` lines below can be removed :)

andyw8: Oh yes, updated.

deivid-rodriguez (approved):
Thanks @andyw8!

andyw8:
Thanks to @swanson for the idea.

PR #4511, opened 2021-04-07
Bump patch level versions of ruby in CI

deivid-rodriguez:

We should run tests against the latest releases of ruby.

Do just that.

### DIFF
--- a/.github/workflows/ubuntu-bundler.yml
+++ b/.github/workflows/ubuntu-bundler.yml
@@ -18,10 +18,10 @@ jobs:
         ruby:
           - { name: ruby-2.3, value: 2.3.8 }
           - { name: ruby-2.4, value: 2.4.10 }
-          - { name: ruby-2.5, value: 2.5.8 }
-          - { name: ruby-2.6, value: 2.6.6 }
-          - { name: ruby-2.7, value: 2.7.2 }
-          - { name: ruby-3.0, value: 3.0.0 }
+          - { name: ruby-2.5, value: 2.5.9 }
+          - { name: ruby-2.6, value: 2.6.7 }
+          - { name: ruby-2.7, value: 2.7.3 }
+          - { name: ruby-3.0, value: 3.0.1 }
 
         bundler:
           - { name: 2, value: '' }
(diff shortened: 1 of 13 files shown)


simi:
For example `2.5.9` ruby version is not found in ruby-setup GH action (https://github.com/rubygems/rubygems/pull/4511/checks?check_run_id=2332121751), but it was added already at ruby-setup (https://github.com/ruby/setup-ruby/commit/d2f535c1eda3aae4274db5cc4626d384b5b000a8). @eregon any idea?

deivid-rodriguez:
It's because https://github.com/oneclick/rubyinstaller2 has not yet released the new patch levels.

.github/workflows/ubuntu-bundler.yml
```diff
@@ -18,9 +18,9 @@ jobs:
         ruby:
           - { name: ruby-2.3, value: 2.3.8 }
           - { name: ruby-2.4, value: 2.4.10 }
-          - { name: ruby-2.5, value: 2.5.8 }
-          - { name: ruby-2.6, value: 2.6.6 }
-          - { name: ruby-2.7, value: 2.7.2 }
+          - { name: ruby-2.5, value: 2.5.9 }
+          - { name: ruby-2.6, value: 2.6.7 }
+          - { name: ruby-2.7, value: 2.7.3 }
           - { name: ruby-3.0, value: 3.0.0 }
```
deivid-rodriguez, line 24: Forgot this update for some reason!

.github/workflows/windows-bundler.yml
```diff
@@ -23,10 +23,10 @@ jobs:
...
           - { name: jruby-9.2, value: jruby-9.2.16.0 }
           - { name: ruby-2.4, value: 2.4.10 }
-          - { name: ruby-2.5, value: 2.5.8 }
-          - { name: ruby-2.6, value: 2.6.6 }
-          - { name: ruby-2.7, value: 2.7.2 }
-          - { name: ruby-3.0, value: 3.0.0 }
+          - { name: ruby-2.5, value: 2.5.9 }
+          - { name: ruby-2.6, value: 2.6.7 }
+          - { name: ruby-2.7, value: 2.7.3 }
+          - { name: ruby-3.0, value: 3.0.1 }
```
eregon, line 29: You could potentially use `'2.5'`/`'2.6'`/etc and then it would automatically update when RubyInstaller has a build for it (and I run a script in setup-ruby). Depends if you want Ruby patch updates to be explicit or not.

deivid-rodriguez: Yeah, we're aware, thanks! I do prefer explicit patch levels. I understand auto-selection can be handy for many users, but for us I prefer explicit, given how hard our CI is to keep green, and how deeply connected with core ruby rubygems & bundler are.

PR #4512, opened 2021-04-07
Lock all test gems

deivid-rodriguez:

Our check that `rubocop` or `standard` can be run without offenses on a freshly generated gem is very brittle and usually fails due various reasons when new release of those gems or their dependencies happen. This happened recently, causing our CI to be currently broken.

We could add workarounds for those issues, but I prefer to not be subject to third party releases breaking our tests at all. So this PR tries to achieve that.

My fix is to keep the `standard` and `rubocop` depenencies of our specs in a Gemfile and lockflie duples, that are installed to a separate path when tests are setup. Then the specs that check whether a `standard` or `rubocop` run is clean use that location as their `BUNDLE_PATH`. 


deivid-rodriguez:
I set this to be automerged to get CI back to green.

simi:
💪

PR #4522, opened 2021-04-11
Cancel `deployment_means_frozen` change

deivid-rodriguez:

We have the following change scheduled for bundler 3:

* Remove the `frozen` setting.
* Change the current `deployment` setting to mean what the `frozen` setting means today.

I think this is not a good change to do now, because:

* It does not have a clean deprecation/migration path: we can't deprecate frozen and give an alternative, because `deployment` does not an work today as it will work in bundler 3.

* When removing `frozen` was proposed, one of the motivations was that there was no `deployment` setting, only a `--deployment` flag. Today we do have a `deployment` setting so that makes the migration path harder.

I do understand that a `deployment` setting that means `frozen` + `path=vendor/bundle` might be kind of obscure/unnecessary, specially on bundler 3 where bundler will install locally, not globally, by default. But I think we have sufficient breaking changes for bundler 3, like the change to install locally instead of globally which will be big, so I prefer to cancel this one, and maybe consider it for later (once bundler 3 is settled and the `deployment` setting _is_ actually unnecessary most of the times).

My fix is to completely cancel the breaking change for the moment.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -197,7 +197,7 @@ module Bundler
 
     def frozen_bundle?
       frozen = settings[:deployment]
-      frozen ||= settings[:frozen] unless feature_flag.deployment_means_frozen?
+      frozen ||= settings[:frozen]
       frozen
     end
 
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -33,12 +33,8 @@ module Bundler
 
         options[:local] = true if Bundler.app_cache.exist?
 
-        if Bundler.feature_flag.deployment_means_frozen?
-          Bundler.settings.set_command_option :deployment, true
-        else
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
@indirect I just wanted to double check this with you. Do you think my plan/explanation makes sense? Original discussion is at https://github.com/rubygems/bundler/issues/4619 and the linked PR.

indirect:
Yes, I think it makes sense! The underlying goal is to clear up the confusion caused by `deployment` non-obviously including both `frozen` mode and also `path: vendor/bundle`. Some future migration path that resolves that confusing by pulling out the path setting and making it explicit seems like the way to go to me, but it also doesn't seem terribly urgent.

deivid-rodriguez:
I agree. In my opinion, the easiest way to resolve the confusion would be to deprecate the `deployment` configuration in favor of separate `frozen` and `path` configurations once we release bundler 3. I'll reticket that improvement so it doesn't get lost :+1:.

PR #4529, opened 2021-04-13
Prefer File.open instead of Kernel#open

mame:
... because Kernel#open may allow unintentional command injection.
At present, I don't know how to exploit these usages of Kernel#open, but
I'm afraid if they may be used with untrusted user input accidentally in
future. As far as I see, there is no reason to use Kernel#open.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -813,7 +813,7 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   ##
   # Safely write a file in binary mode on all platforms.
   def self.write_binary(path, data)
-    open(path, 'wb') do |io|
+    File.open(path, 'wb') do |io|
       begin
         io.flock(File::LOCK_EX)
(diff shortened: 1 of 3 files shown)


hsbt:
👍

deivid-rodriguez:
@hsbt I don't think this is a security fix, since I don't think it can't really be exploited at the moment. I think it's better to tag it as `rubgygems: enhancements` since it's just a preventive measure in case it somehow became exploitable in the future. I already used that same criteria to tag https://github.com/rubygems/rubygems/pull/4506 as an enhancement. Same thing would also apply to #4530.

hsbt:
I have no strong opinion to remain that. I replaced your suggestion.

deivid-rodriguez:
Alright @hsbt! It just felt to me that tagging this as a security fix would essentially imply that everyone using `Kernel#open` has a security vulnerability, which didn't seem like a good thing to imply.

PR #4530, opened 2021-04-13
Prefer File.read instead of IO.read

mame:
Ditto to IO.readlines. IO.read and IO.readlines may run a command.

Related: https://github.com/rubygems/rubygems/pull/4529


deivid-rodriguez (approved):
Seems worth another rubocop like https://github.com/rubocop/rubocop/pull/5319!

PR #4533, opened 2021-04-13
Add a git version guard for init.defaultBranch

mame:
Some specs use `git config --global init.defaultBranch main`, but the
configuration has been implemented since git 2.28.0.
I'm using git 2.25.1 (which is bundled in Ubuntu 20.04), so these specs
fail as follows.

```
Failures:

  1) bundle install git sources displays the correct default branch
...
       Diff:
       @@ -1,10 +1,19 @@
       -Using foo 1.0 from file:///home/mame/work/rubygems/bundler/tmp/1/libs/foo (at main@cdca7ca)
       ...
       +Using foo 1.0 from file:///home/mame/work/rubygems/bundler/tmp/1/libs/foo (at master@cdca7ca)
       ...
```

This changeset adds `:git => ">= 2.28.0"` to suppress the failures.

### DIFF
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -1149,7 +1149,7 @@ RSpec.describe "bundle gem" do
       it_behaves_like "--no-mit flag"
     end
 
-    context "with coc option in bundle config settings set to true" do
+    context "with coc option in bundle config settings set to true", :git => ">= 2.28.0" do
       before do
         global_config "BUNDLE_GEM__COC" => "true"
       end
@@ -1157,7 +1157,7 @@ RSpec.describe "bundle gem" do
       it_behaves_like "--no-coc flag"
     end
 
-    context "with coc option in bundle config settings set to false" do
+    context "with coc option in bundle config settings set to false", :git => ">= 2.28.0" do
       before do
         global_config "BUNDLE_GEM__COC" => "false"
       end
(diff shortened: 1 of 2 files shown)


simi (approved):

bundler/spec/commands/newgem_spec.rb
```diff
@@ -1149,15 +1149,15 @@ def create_temporary_dir(dir)
       it_behaves_like "--no-mit flag"
     end
 
-    context "with coc option in bundle config settings set to true" do
+    context "with coc option in bundle config settings set to true", :git => ">= 2.28.0" do
```
deivid-rodriguez, line 1152: Can you instead tag the specific spec that runs `git config --global init.defaultBranch main` and then checks that links to `main`are present in the generated README? It t would make it more apparent why the spec only works on some git versions, and allows you to run all the other specs that do pass.

mame: Is this right? https://github.com/rubygems/rubygems/pull/4533/commits/691b61751e7324564394cdab4ba6398170d50080

mame: Ah, I missed "then checks that links to mainare present in the generated README".

deivid-rodriguez: which I believe is the only spec that should be failing for you.

mame: Ahh, I see! https://github.com/rubygems/rubygems/pull/4533/commits/dab3fc5de156b47bd23dea2adfa5178c133a4c23 It works great on my machine.

deivid-rodriguez: Yes, I meant exactly that, except that you missed the ">=" in the filter! Thanks @mame!

mame: Fixed 😓

deivid-rodriguez (approved):
Thanks for your patience @mame, and sorry for the bad explanation :sweat_smile:.

PR #4551, opened 2021-04-21
Fix usage messages to reflect the current POSIX-compatible behaviour

### LINKED ISSUE #4538: Gem respects POSIXLY_CORRECT but that is not mentioned anywhere

My current problem is when POSIXLY_CORRECT env variable is set, gem install
does not follow its own help message.

```
+root@13d23e3307b4:/# gem help install | head -n1
Usage: gem install GEMNAME [GEMNAME ...] [options] -- --build-flags [options]
```

Usage suggests that correct command is:

```
gem install bundler --conservative --user -N --version '~> 1.17'
```

However, when executed with POSIXLY_CORRECT:

```
+root@13d23e3307b4:/# POSIXLY_CORRECT=1 gem install bundler --conservative --user -N --version '~> 1.17'
Fetching bundler-2.2.16.gem
Successfully installed bundler-2.2.16
Parsing documentation for bundler-2.2.16
Installing ri documentation for bundler-2.2.16
Done installing documentation for bundler after 3 seconds
ERROR:  Could not find a valid gem '~> 1.17' (>= 0) in any repository
1 gem installed
+root@13d23e3307b4:/# echo $?
2
```

Not exactly what I would expect. This is not mentioned in the help message as far as I can tell:

```
+root@13d23e3307b4:/# gem help install | grep POSIXLY || echo NOTHING
NOTHING
```

Putting the gem name to the end makes it work:

```
+root@13d23e3307b4:/# POSIXLY_CORRECT=1 gem install --conservative --user -N --version '~> 1.17' bundler
Fetching bundler-1.17.3.gem
WARNING:  You don't have /root/.gem/ruby/2.7.0/bin in your PATH,
          gem executables will not run.
Successfully installed bundler-1.17.3
1 gem installed
```

I'm not sure what the correct solution is, but the current state seems broken.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
+root@13d23e3307b4:/# gem env version
3.1.2
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md) (did not read).

graywolf-at-work:
When POSIXLY_CORRECT environment variable is set, the argument parsing
tries to behave in POSIX-compatible manner. That is fine, but leads to
usage messages no longer being accurate:

    $ POSIXLY_CORRECT= gem open foo -e nano
        Too many gem names (foo, -e, nano); please specify only one

Correct invocation would be:

    $ POSIXLY_CORRECT= gem open -e nano foo
    Unable to find gem 'foo'

However, usage help says nothing about that. This commit therefore
shuffles the usage order to make it always correct.

Closes #4538.

Described in the commit message.

- [?] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
  -> I'm not sure what to do about this. There are no existing tests that break with this and introducing new just for checking if correct message is printed seems weird.

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -130,7 +130,7 @@ You can use `i` command instead of `install`.
   end
 
   def usage # :nodoc:
-    "#{program_name} GEMNAME [GEMNAME ...] [options] -- --build-flags"
+    "#{program_name} [options] GEMNAME [GEMNAME ...] -- --build-flags"
   end
 
   def check_install_dir # :nodoc:
--- a/lib/rubygems/commands/open_command.rb
+++ b/lib/rubygems/commands/open_command.rb
@@ -36,7 +36,7 @@ class Gem::Commands::OpenCommand < Gem::Command
   end
 
   def usage # :nodoc:
-    "#{program_name} GEMNAME [-e COMMAND]"
+    "#{program_name} [-e COMMAND] GEMNAME"
   end
 
   def get_env_editor
--- a/lib/rubygems/commands/yank_command.rb
+++ b/lib/rubygems/commands/yank_command.rb
@@ -24,7 +24,7 @@ data you will need to change them immediately and yank your gem.
   end
 
   def usage # :nodoc:
-    "#{program_name} GEM -v VERSION [-p PLATFORM] [--key KEY_NAME] [--host HOST]"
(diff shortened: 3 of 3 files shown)


deivid-rodriguez (approved):
Looks great @graywolf-at-work, thanks so much!

PR #4362, opened 2021-02-08
Fix encoding issue on compact index updater

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Sometimes bundler crashes with an encoding error: https://github.com/rubygems/rubygems/issues/4360.

## What is your fix for the problem, implemented in this PR?

We should write versions file binarily so that no transcoding happens when the encoding of the file being created and the encoding of the data don't match.

I'm not sure whether this problem was introduced with https://github.com/rubygems/rubygems/pull/4081 or not, though. It seems to me that it should reproduce either way.

Fixes #4360.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4563, opened 2021-04-26
Discard partial range responses without etag

### LINKED ISSUE #4514: Can't bundle new version of Sidekiq Pro
Hi, I released Sidekiq Pro 5.2.2 yesterday and people are telling me that they can't get it with Bundler 2.2.15. I noticed that `~/.bundle/cache/compact_index/*/versions` is out of date on my machine. The cached version is from Feb.

```
~/.bundle/cache/compact_index/enterprise.contribsys.com.xxxxxxx.443.cdac25812e55451c9ec774c09cafb44f$ more versions 
created_at: 2021-02-04T22:11:58+00:00
```

My server version is from yesterday.

```
/var/www/enterprise$ more versions 
created_at: 2021-04-07T20:37:17+00:00
```

Even if I put `gem 'sidekiq-pro', '5.2.2'`, Bundler tells me that version is not available.

Is there any debug logging I can enable or other way to diagnose the problem? Can you reproduce the issue with your Sidekiq Pro credentials?

### LINKED ISSUE #4543: Bundler should handle invalid responses to partial content requests
Bundler should discard any partial range response and the local file if the response comes back without an etag and download the entire thing. See #4514.

deivid-rodriguez:

Some custom gem servers (at least sidekiq's pro one) don't support "incremental API content", yet they respond to partial range responses, providing no Etag.

Previously bundler would handle this just fine, because it would discard the response given and retry a "full range request" in this case.

However, in #3865 we changed bundler to accept the response in this case, which would be appended to the existing response. However, since new versions are not appended to the remote versions file (no "incremental API content"), this would result in a bad response being written to the local client compact index cache. This should be the reason for the problems reported in #4514.

My fix, for now, is to revert #3865. But as a further improvement, we can keep the "accept the response if no etag" idea, but only if we get a 200 response, not a partial content (206) response.

If we did the latter, sidekiq-pro can fix their server code to always serve full responses, even when requested partial ranges, and things will work in a more efficient way, since it wouldn't require double fallback requests.

EDIT: Extra improvement was also implemented :+1:.

Closes #4514.
Closes #4543.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -50,16 +50,20 @@ module Bundler
 
           content = response.body
 
-          SharedHelpers.filesystem_access(local_temp_path) do
+          etag = (response["ETag"] || "").gsub(%r{\AW/}, "")
+          correct_response = SharedHelpers.filesystem_access(local_temp_path) do
             if response.is_a?(Net::HTTPPartialContent) && local_temp_path.size.nonzero?
               local_temp_path.open("a") {|f| f << slice_body(content, 1..-1) }
+
+              etag_for(local_temp_path) == etag
             else
               local_temp_path.open("wb") {|f| f << content }
+
+              etag.length.zero? || etag_for(local_temp_path) == etag
             end
           end
 
-          etag = (response["ETag"] || "").gsub(%r{\AW/}, "")
-          if etag.length.zero? || etag_for(local_temp_path) == etag
+          if correct_response
             SharedHelpers.filesystem_access(local_path) do
               FileUtils.mv(local_temp_path, local_path)
             end
--- a/bundler/spec/bundler/compact_index_client/updater_spec.rb
+++ b/bundler/spec/bundler/compact_index_client/updater_spec.rb
@@ -38,7 +38,6 @@ RSpec.describe Bundler::CompactIndexClient::Updater do
 
   context "when bundler doesn't have permissions on Dir.tmpdir" do
     it "Errno::EACCES is raised" do
-      local_path # create local path before stubbing mktmpdir
       allow(Bundler::Dir).to receive(:mktmpdir) { raise Errno::EACCES }
 
       expect do
--- a/bundler/spec/install/gems/compact_index_spec.rb
+++ b/bundler/spec/install/gems/compact_index_spec.rb
@@ -826,6 +826,28 @@ The checksum of /versions does not match the checksum provided by the server! So
     expect(the_bundle).to include_gems "rack 1.0.0"
   end
 
+  it "performs full update if server endpoints serve partial content responses but don't have incremental content and provide no Etag" do
+    build_repo4 do
+      build_gem "rack", "0.9.1"
+    end
+
+    install_gemfile <<-G, :artifice => "compact_index", :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }
+      source "#{source_uri}"
+      gem 'rack', '0.9.1'
+    G
+
+    update_repo4 do
+      build_gem "rack", "1.0.0"
+    end
+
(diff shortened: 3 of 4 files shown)


indirect:
This seems reasonable! My one caution is to make sure that Bundler actually accepts responses from servers that don't include an etag at all. At the time of my original PR, Bundler would refuse responses with no etag, and act like the compact index didn't exist. If that's no longer the case, sounds like this takes care of everything. 😄

indirect (approved):

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -92,11 +92,8 @@ def slice_body(body, range)
 
       def checksum_for_file(path)
         return nil unless path.file?
-        # This must use File.read instead of Digest.file().hexdigest
-        # because we need to preserve \n line endings on windows when calculating
-        # the checksum
         SharedHelpers.filesystem_access(path, :read) do
-          SharedHelpers.digest(:MD5).hexdigest(File.read(path))
+          SharedHelpers.digest(:MD5).file(path).hexdigest
```
deivid-rodriguez, line 96: I changed this method because one spec was failing with an invalid encoding error. I think this code was now incorrect since we started writing the versions file with "binary" encoding at https://github.com/rubygems/rubygems/pull/4362.

deivid-rodriguez:
> This seems reasonable! My one caution is to make sure that Bundler actually accepts responses from servers that don't include an etag at all. At the time of my original PR, Bundler would refuse responses with no etag, and act like the compact index didn't exist. If that's no longer the case, sounds like this takes care of everything. smile

Yes. What it does now is falling back to "full update" mode by retrying the request without `If-None-Match` and `Range` headers, which seems reasonable.

indirect:
Perfect 👍🏻

bundler/lib/bundler/compact_index_client/updater.rb
deivid-rodriguez: Well, now we have some mostl likely related specs failing on Windows. I'll dig deeper into this and possibly move this fix/changes to a separate PR.

deivid-rodriguez:
Alright, I keep going a bit in circles here, but at the same time knowing more about the issue.

After all you're 100% right when you said:

> This seems reasonable! My one caution is to make sure that Bundler actually accepts responses from servers that don't include an etag at all. At the time of my original PR, Bundler would refuse responses with no etag, and act like the compact index didn't exist. If that's no longer the case, sounds like this takes care of everything. :smile:

This is actually how bundler behaved before #3865. Both partial range requests and full requests required an Etag, so we would indeed end up raising a `MisMatchedChecksumError` in the missing Etag case. The question is whether this is a good behaviour or not. The result of the error is that it would be rescued here to decide that the compact index API is not an available fetching method:

https://github.com/rubygems/rubygems/blob/563933a08d1d2823cf6d81fb08b96ebf3a2101f5/bundler/lib/bundler/fetcher/compact_index.rb#L70-L84

So the way bundler behaved before your PR (and after this one if we merge it) is that it falls back to other fetching methods. In the case of sidekiq-pro, to the full index, which works just fine.

A theory of why we fixed this in the first place is that some gem servers would provide no Etag _and_ implement no other fetching methods other than the compact index API. If this is the case, yeah, my PR will break them again :grimacing:. So it sounds like I need to go all the way and keep accepting the response without an ETag in case we're requesting a full (not partial) update.

bundler/lib/bundler/compact_index_client/updater.rb
deivid-rodriguez: Final version doesn't need these changes.

deivid-rodriguez:
Alright, I implemented only discarding the no etag response when doing a partial update, but accepting it for full updates. This should be finally ready now assuming CI passes.

indirect (approved):
Yep, this looks like exactly what I was thinking. Thanks!

PR #4565, opened 2021-04-27
Improve authentication required error message to include an alternative using `ENV`

deivid-rodriguez:

The convention for environment variables and credentials is tricky, and it can be hard to figure out the name of the environment variable that needs to be set.

Include the specific environment variable that needs to be set with the error message.

Closes https://github.com/rubygems/rubygems/issues/3255.

### DIFF
--- a/bundler/spec/bundler/fetcher/downloader_spec.rb
+++ b/bundler/spec/bundler/fetcher/downloader_spec.rb
@@ -83,6 +83,11 @@ RSpec.describe Bundler::Fetcher::Downloader do
           /Authentication is required for www.uri-to-fetch.com/)
       end
 
+      it "should raise a Bundler::Fetcher::AuthenticationRequiredError with advices" do
+        expect { subject.fetch(uri, options, counter) }.to raise_error(Bundler::Fetcher::AuthenticationRequiredError,
+          /`bundle config set --global www\.uri-to-fetch\.com username:password`.*`BUNDLE_WWW__URI___TO___FETCH__COM`/m)
+      end
+
       context "when the there are credentials provided in the request" do
         let(:uri) { Bundler::URI("http://user:password@www.uri-to-fetch.com") }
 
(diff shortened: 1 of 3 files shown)


bundler/spec/bundler/fetcher/downloader_spec.rb
```diff
@@ -83,6 +83,11 @@
           /Authentication is required for www.uri-to-fetch.com/)
       end
 
+      it "should raise a Bundler::Fetcher::AuthenticationRequiredError with advices" do
+        expect { subject.fetch(uri, options, counter) }.to raise_error(Bundler::Fetcher::AuthenticationRequiredError,
+          /`bundle config set --global www\.uri-to-fetch\.com username:password`.*`BUNDLE_WWW__URI-TO-FETCH__COM`/m)
```
ianfixes, line 88: The `-` characters in `BUNDLE_WWW__URI-TO-FETCH__COM` aren't [valid in variable names](https://www.tutorialspoint.com/unix/unix-using-variables.htm#:~:text=Variable%20Names,have%20their%20names%20in%20UPPERCASE.), although I'm not sure whether this matters for the test

ianfixes:
This looks great, thank you! Is there anything I can contribute to help with the failing tests?

bundler/spec/bundler/fetcher/downloader_spec.rb
deivid-rodriguez: Spec failure was not related, but that's a good point though!

We should either fix that issue right now, or open a separate ticket about it and avoid the ENV variable recommendation in that particular case.

deivid-rodriguez: According to https://stackoverflow.com/questions/3523028/valid-characters-of-a-hostname, only letters, numbers, and hyphens are valid within a hostname component, so I guess we can add the `___` (triple underscore) <-> `-` (hyphen) to this translation and fix the issue completely.

deivid-rodriguez: I proposed a fix at https://github.com/rubygems/rubygems/pull/4571.

ianfixes:
I think this might need a slight tweak... see #4601

deivid-rodriguez:
Nice catch, thanks!

Issue #4414, opened 2021-02-28
Redact private data from `bundle env` output


deivid-rodriguez:
### Describe the problem as clearly as you can

As per https://github.com/rubygems/rubygems/pull/4359, Our issue templates tells users to copy the output of `bundle env` without remembering them to anonymize any private data before doing so.

Ideally we should anonymize any private data included in `bundle env` output by default, but since that might be tricky, we should at least add a note about it to the template.

aaronjensen:
Just a note: even if you succeed in redacting data (which, I would think would just be a matter of only emitting bundle configuration for named/hard-coded configuration options, which should not include credentials), you should still encourage anonymizing the data because you cannot force people to upgrade and there are already old versions in the wild that emit private data.

deivid-rodriguez:
Sure, I just added a note about this at #4416.

deivid-rodriguez:
#4566 should fix this!

PR #4566, opened 2021-04-27
Redact credentials from `bundle env` and `bundle config`

deivid-rodriguez:

Currently `bundle env`  and `bundle config` prints any credentials configured. Users may blindly copy-paste this output, for example to a bundler issue, unintentionally leaking their credentials.

Redact any credentials from `bundle env` and `bundle config` output.

Closes https://github.com/rubygems/rubygems/issues/4414.

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -13,6 +13,7 @@ module Bundler
       auto_install
       cache_all
       cache_all_platforms
+      clean
       default_install_uses_path
       deployment
       disable_checksum_validation
@@ -25,11 +26,14 @@ module Bundler
       force_ruby_platform
       forget_cli_options
       frozen
+      gem.changelog
       gem.coc
       gem.mit
+      git.allow_insecure
       global_gem_cache
       ignore_messages
       init_gems_rb
+      inline
       no_install
       no_prune
       path_relative_to_cwd
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -129,7 +129,7 @@ module Bundler
 
       keys.map do |key|
         key.sub(/^BUNDLE_/, "").gsub(/__/, ".").downcase
-      end
+      end.sort
     end
 
     def local_overrides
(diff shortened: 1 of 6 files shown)


bundler/lib/bundler/settings.rb
```diff
@@ -340,6 +372,10 @@ def array_to_s(array)
       array.join(":").tr(" ", ":")
     end
 
+    def redact(password)
+      "*" * password.length
```
deivid-rodriguez, line 376: I can make this a fixed number of `*` to avoid potentially leaking a password length, not sure whether we should care about that.

deivid-rodriguez: Went ahead and changed it to the fixed string `[REDACTED]`.

bundler/spec/commands/config_spec.rb
```diff
@@ -408,6 +408,14 @@
       expect(out).to eq "spec_run=true"
     end
 
+    it "list with credentials" do
+      bundle "config list", :env => { "BUNDLE_GEMS__MYSERVER__COM" => "user:password" }
+      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\nspec_run\nSet via BUNDLE_SPEC_RUN: \"true\"\n\ngems.myserver.com\nSet via BUNDLE_GEMS__MYSERVER__COM: \"user:[REDACTED]\""
+
+      bundle "config list", :parseable => true, :env => { "BUNDLE_GEMS__MYSERVER__COM" => "user:password" }
+      expect(out).to eq "spec_run=true\ngems.myserver.com=user:password"
```
deivid-rodriguez, line 416: As this spec proves, credentials are redacted from regular `bundle config` out, but not when using the `--parseable` flag, which seems like the right thing to do.

simi: Seems reasonable, but would be good to document this behaviour to prevent unexpected output.

simi (approved):

indirect (approved):

deivid-rodriguez:
Thank for your your reviews! This run into some weird failure on truffleruby. I rebased it and will set auto-merge now :+1:.

PR #4586, opened 2021-05-08
Undeprecate bundle show

### LINKED ISSUE #3243: add `bundle which <GEM>` command
We already have `bundle open <GEM>`, but sometimes I just want the full path to an installed gem, instead of opening a new `$EDITOR`.

I think we should add a `which` command, that just prints the path of the installed gem.

Example:

```
$ bundle which racc
/home/user/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/racc-1.4.16
```

NeimadTL:
The timeline of how show got deprecated is on this issue's comment: https://github.com/rubygems/rubygems/issues/3243#issuecomment-780379254. We eventually decided to undeprecate show as no apparent reason was give
to deprecate it.

`bundle show` was deprecated for no apparent reason.
timeline (see [here](https://github.com/rubygems/rubygems/issues/3243#issuecomment-780379254))

`bundle show` has been undeprecated.

Closes #3243.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -308,39 +308,19 @@ module Bundler
       end
     end
 
-    unless Bundler.feature_flag.bundler_3_mode?
-      desc "show GEM [OPTIONS]", "Shows all gems that are part of the bundle, or the path to a given gem"
-      long_desc <<-D
-        Show lists the names and versions of all gems that are required by your Gemfile.
-        Calling show with [GEM] will list the exact location of that gem on your machine.
-      D
-      method_option "paths", :type => :boolean,
-                             :banner => "List the paths of all gems that are required by your Gemfile."
-      method_option "outdated", :type => :boolean,
-                                :banner => "Show verbose output including whether gems are outdated."
-      def show(gem_name = nil)
-        if ARGV[0] == "show"
-          rest = ARGV[1..-1]
-
-          if flag = rest.find{|arg| ["--verbose", "--outdated"].include?(arg) }
-            Bundler::SharedHelpers.major_deprecation(2, "the `#{flag}` flag to `bundle show` was undocumented and will be removed without replacement")
-          else
-            new_command = rest.find {|arg| !arg.start_with?("--") } ? "info" : "list"
-
-            new_arguments = rest.map do |arg|
-              next arg if arg != "--paths"
-              next "--path" if new_command == "info"
-            end
-
-            old_argv = ARGV.join(" ")
-            new_argv = [new_command, *new_arguments.compact].join(" ")
-
-            Bundler::SharedHelpers.major_deprecation(2, "use `bundle #{new_argv}` instead of `bundle #{old_argv}`")
-          end
-        end
-        require_relative "cli/show"
-        Show.new(options, gem_name).run
-      end
+    desc "show GEM [OPTIONS]", "Shows all gems that are part of the bundle, or the path to a given gem"
+    long_desc <<-D
+      Show lists the names and versions of all gems that are required by your Gemfile.
+      Calling show with [GEM] will list the exact location of that gem on your machine.
+    D
+    method_option "paths", :type => :boolean,
+                           :banner => "List the paths of all gems that are required by your Gemfile."
+    method_option "outdated", :type => :boolean,
+                              :banner => "Show verbose output including whether gems are outdated."
+    def show(gem_name = nil)
+      SharedHelpers.major_deprecation(2, "the `--outdated` flag to `bundle show` was undocumented and will be removed without replacement") if ARGV.include?("--outdated")
+      require_relative "cli/show"
+      Show.new(options, gem_name).run
     end
 
     desc "list", "List all gems in the bundle"
--- a/bundler/spec/other/major_deprecation_spec.rb
+++ b/bundler/spec/other/major_deprecation_spec.rb
@@ -569,18 +569,6 @@ The :gist git source is deprecated, and will be removed in the future. Add this 
       G
     end
 
-    context "without flags" do
-      before do
-        bundle :show
-      end
-
-      it "prints a deprecation warning recommending `bundle list`", :bundler => "< 3" do
-        expect(deprecations).to include("use `bundle list` instead of `bundle show`")
-      end
-
-      pending "fails with a helpful message", :bundler => "3"
-    end
-
     context "with --outdated flag" do
       before do
         bundle "show --outdated"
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/cli.rb
```diff
@@ -316,28 +316,11 @@ def update(*gems)
...
-              next arg if arg != "--paths"
-              next "--path" if new_command == "info"
-            end
-
-            old_argv = ARGV.join(" ")
-            new_argv = [new_command, *new_arguments.compact].join(" ")
-
-            Bundler::SharedHelpers.major_deprecation(2, "use `bundle #{new_argv}` instead of `bundle #{old_argv}`")
-          end
-        end
```
deivid-rodriguez, line 340: This whole block of code is wrapped under a `unless Bundler.feature_flag.bundler_3_mode?` condition. We need to remove that condition so that the command doesn't get removed when we release bundler 3.

deivid-rodriguez (changes requested):
@NeimadTL Thanks for not forgetting about this :heart:

Made one comment, but other than that it seems all good!

bundler/lib/bundler/cli.rb
```diff
@@ -316,28 +316,11 @@ def update(*gems)
       D
       method_option "paths", :type => :boolean,
                              :banner => "List the paths of all gems that are required by your Gemfile."
+      method_option "verbose", :type => :boolean,
+                               :banner => "Show verbose output."
```
deivid-rodriguez, line 320: The verbose flag is global and not specific to any command, so there's no need to specify it here. It was a mistake that the deprecation message also applied to the `--verbose` flag.

As an extra improvement, could you change [the docs](https://github.com/rubygems/rubygems/blob/8e204557a657d3a65f40f4b1598d92ab2adcbca8/bundler/lib/bundler/man/bundle-show.1.ronn) to document the missing `--outdated` flag? (You'll need to run `bin/rake man:build` after that to regenerate the other man pages after that). Alternative, since `bundle show --outdated` is not really useful IMO, we could keep the deprecation for that. What do you think?

bundler/lib/bundler/cli.rb
NeimadTL: ok, will do that, thanks for the heads up ;)

bundler/lib/bundler/cli.rb
NeimadTL: Thanks for the information about `--verbose`. Will add the deprecation for `--outdated` since there's `bundle outdated` already as you told me.

NeimadTL:
@deivid-rodriguez , just made the recommended changes. Thanks again for all the information ;)

bundler/lib/bundler/cli.rb
deivid-rodriguez: Thanks!

bundler/lib/bundler/cli.rb
deivid-rodriguez: I agree restoring the deprecation of `bundle show --outdated` is best :+1:.

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -574,10 +574,6 @@
         bundle :show
       end
 
-      it "prints a deprecation warning recommending `bundle list`", :bundler => "< 3" do
-        expect(deprecations).to include("use `bundle list` instead of `bundle show`")
-      end
-
```
deivid-rodriguez, line 580: For the specs being removed, they're usually following by a pending "fails with a helpful" message spec, that should've been implemented before we release bundler 3. Since these specs won't be implemented anymore, they should be removed too.

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -598,10 +594,6 @@
         bundle "show --verbose"
       end
 
-      it "prints a deprecation warning informing about its removal", :bundler => "< 3" do
-        expect(deprecations).to include("the `--verbose` flag to `bundle show` was undocumented and will be removed without replacement")
-      end
-
```
deivid-rodriguez, line 604: Same comment about removing the pending spec below.

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -622,10 +610,6 @@
         bundle "show --paths"
       end
 
-      it "prints a deprecation warning recommending `bundle list`", :bundler => "< 3" do
-        expect(deprecations).to include("use `bundle list` instead of `bundle show --paths`")
-      end
-
```
deivid-rodriguez, line 628: Same comment about removing the pending spec below.

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -634,10 +618,6 @@
         bundle "show rack --paths"
       end
 
-      it "prints deprecation warning recommending `bundle info`", :bundler => "< 3" do
-        expect(deprecations).to include("use `bundle info rack --path` instead of `bundle show rack --paths`")
-      end
-
```
deivid-rodriguez, line 640: Same comment about removing the pending spec below.

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -587,8 +583,8 @@
       end
 
       it "prints a deprecation warning informing about its removal", :bundler => "< 3" do
-        expect(deprecations).to include("the `--outdated` flag to `bundle show` was undocumented and will be removed without replacement")
-      end
+       expect(deprecations).to include("The `--outdated` flag to `bundle show` was undocumented and will be removed without replacement")
```
deivid-rodriguez, line 586: Fixing the case seems fine, but this is also removing one whitespace from the beginning of the line, making the indentation inconsistent. Can you fix it?

deivid-rodriguez (changes requested):
One final round of comments, but seems mostly ready!

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -573,8 +573,6 @@
       before do
         bundle :show
       end
-
```
deivid-rodriguez, line 576: @NeimadTL I just noticed that actually, all specs inside this "`context "bundle show" do` block, except for the one about the `--outdated` flag can now be removed. Can you do it?

bundler/spec/other/major_deprecation_spec.rb
```diff
@@ -587,8 +581,8 @@
       end
 
       it "prints a deprecation warning informing about its removal", :bundler => "< 3" do
-        expect(deprecations).to include("the `--outdated` flag to `bundle show` was undocumented and will be removed without replacement")
-      end
+       expect(deprecations).to include("the `--outdated` flag to `bundle show` was undocumented and will be removed without replacement")
+     end
```
deivid-rodriguez, line 585: Actually, it's the indentation of the line that was causing the extra error. Reverting the message casing is fine though, since it's at least a different formatting issue than what we're fixing here. This is how this should look I believe.

```suggestion
        expect(deprecations).to include("the `--outdated` flag to `bundle show` was undocumented and will be removed without replacement")
      end
```

You can run `bin/rubocop` locally from the bundler folder to double check that the indentation is ok.

deivid-rodriguez (changes requested):
Sorry I keep noticing more things we should delete :sweat_smile:.

bundler/spec/other/major_deprecation_spec.rb
NeimadTL: done. thanks for the rubocop hint

bundler/spec/other/major_deprecation_spec.rb
deivid-rodriguez: Yes, it's correct, thanks!

bundler/spec/other/major_deprecation_spec.rb
deivid-rodriguez: Thanks for fixing it!

deivid-rodriguez:
Everything was addressed but there was one leftover issue where you had unintentionally committed a `bundler/.rake_tasks~` file to the repository. I fixed that for you and rebased and squashed the PR since I was at it. Will merge after CI.

Thanks so much @NeimadTL for your contribution! :heart:

PR #4587, opened 2021-05-10
Reset `XDG_CONFIG_HOME` for bundler specs

deivid-rodriguez:

Some spec failures under ruby-core environment.

My fix is an alternative to https://github.com/ruby/ruby/commit/50f17241a32d837403fae68dc1ed0f046506d3e8 and prior commits, which have not yet been ported to our repo.

It seems simpler and more consistent with the current setup to me. @znz what do you think?


hsbt (approved):
👍

simi (approved):

PR #4598, opened 2021-05-12
Use simpler notation for generated `required_ruby_version`

deivid-rodriguez:

I noticed this issue when I stumped across https://github.com/dependabot/dependabot-core/issues/2963. We can't really fix that dependabot issue ourselves, but we can generate new gemspecs without any dependencies on this class.

There's no need to reference an internal rubygems class, even though this is a core class that will probably never be renamed/removed, but still it's unnecessary to depend on it here. Use the simpler notation for `required_ruby_version`.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -14,7 +14,7 @@ Gem::Specification.new do |spec|
 <%- if config[:mit] -%>
   spec.license = "MIT"
 <%- end -%>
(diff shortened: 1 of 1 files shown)


simi:
:muscle:

ybiquitous:
@deivid-rodriguez Thank you for this change! 👍🏼 

I've left a comment in dependabot/dependabot-core#2963, so maybe that issue will be closed soon.

deivid-rodriguez:
No problem! If they won't to play nice with existing notation they will still need to fix that issue, but hopefully this change will alleviate the issue going forward.

PR #4599, opened 2021-05-12
Move test specific classes to `test/`

deivid-rodriguez:

Apparently `Gem::TestCase` is a class supposed to be exposed to rubygems end users. Not sure why, maybe for testing rubygems plugins. I don't think people is using it, so I'd rather move these test clases to `test/`.

In this approach, I deprecated requiring these classes. However, in #3141 we have probably already broken this, so I'm not sure whether is worth doing this and we can just move forward and remove these files.


hsbt:
I'll support this. I agreed to hide test case class in this library. 

FYI: rdoc has been removed them at https://github.com/ruby/rdoc/pull/737

deivid-rodriguez:
I rebased this and completely remove all the deprecation dance since now that we shipped the migration of the base test classes to `test-unit`, we have already broken anyone actually depending on this (hopefully nobody), so going through a deprecation is pointless.

hsbt (approved):
Great work!

PR #4600, opened 2021-05-12
Use `Time.stub :now` to avoid a random failure

mame:
Essentially this reverts 45464bfcbdf9f9cfb440950bc57a27d237627a17.
The commit removed a mock of Time.now, which caused a random failure.

http://rubyci.s3.amazonaws.com/ubuntu1804/ruby-master/log/20210512T123004Z.fail.html.gz
```
  1) Failure:
TestGemPackageTarWriter#test_add_file_signer [/home/chkbuild/chkbuild/tmp/build/20210512T123004Z/ruby/test/rubygems/test_gem_package_tar_writer.rb:117]:
Field mtime of the tar header differs..
<"14046746312\u0000"> expected but was
<"14046746311\x00">.
```

Object#stub is defined at f1af59fe02ef2cc58f13e2742e4cc6cf8c2a1a20, so
now `Time.stub :now` works.

### DIFF
--- a/test/rubygems/test_gem_package_tar_writer.rb
+++ b/test/rubygems/test_gem_package_tar_writer.rb
@@ -27,58 +27,67 @@ class TestGemPackageTarWriter < Gem::Package::TarTestCase
   end
 
   def test_add_file
-    @tar_writer.add_file 'x', 0644 do |f|
-      f.write 'a' * 10
-    end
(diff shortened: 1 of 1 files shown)


simi (approved):

deivid-rodriguez (approved):
Awesome @mame, thanks!

PR #4609, opened 2021-05-19
Fix dependency confusion issues with implicit dependencies

deivid-rodriguez:

The end user problem is that the current source priority of bundler is not secure, potentially causing private gem names to be name-squatted on public gem sources. In particular, indirect dependencies are currently subject to this problem.
 
My fix is to make sure each dependency (including indirect dependencies) gets an explicit source requirement before resolution starts. If a dependency is present in multiple sources, scoped sources should always be preferred to global sources, and if there's an ambiguity between different block sources, and error (bundler 3) or warning (bundler 2) should be raised.

Fixes https://github.com/rubygems/rubygems/issues/3982.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -866,7 +866,7 @@ module Bundler
       source_requirements = if precompute_source_requirements_for_indirect_dependencies?
         { :default => sources.default_source }.merge(source_map.all_requirements)
       else
-        { :global => Source::RubygemsAggregate.new(sources, source_map) }.merge(source_map.direct_requirements)
+        { :default => Source::RubygemsAggregate.new(sources, source_map) }.merge(source_map.direct_requirements)
       end
       metadata_dependencies.each do |dep|
         source_requirements[dep.name] = sources.metadata_source
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -286,45 +286,6 @@ module Bundler
       end
     end
 
-    def build_index
-      Index.build do |idx|
-        dependency_names = @dependencies.map(&:name)
-
-        sources.all_sources.each do |source|
-          source.dependency_names = dependency_names - source_map.pinned_spec_names(source)
-          idx.add_source source.specs
-          dependency_names.concat(source.unmet_deps).uniq!
-        end
-
-        double_check_for_index(idx, dependency_names)
-      end
-    end
-
-    # Suppose the gem Foo depends on the gem Bar.  Foo exists in Source A.  Bar has some versions that exist in both
-    # sources A and B.  At this point, the API request will have found all the versions of Bar in source A,
-    # but will not have found any versions of Bar from source B, which is a problem if the requested version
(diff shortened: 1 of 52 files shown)


zofrex:
I tested this with a project with an indirect private dependency and it works! With this patch applied, Bundler only fetches the indirect dependency from the private server, and ignores the higher-versioned gem on rubygems.org. Awesome!

simi (approved):

Fryguy:
Tried with ManageIQ, and some other test cases and this PR works for me!

voxik:
Is there chance to have this backported into Ruby 2.5+? I am asking, because Red Hat security folks rates this as high/important issue [1]. But this does not look as something which can be easily backported unfortunately.


[1]: https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2020-36327

voxik:
> Is there chance to have this backported into Ruby 2.5+?

Actually 2.5 is out of support now, right?

deivid-rodriguez:
Yes: https://www.ruby-lang.org/en/downloads/.

This would be pretty hard to backport indeed. It's a big patch and the result of work & refactoring across many versions, for example, it's built on top of https://github.com/rubygems/rubygems/pull/4381 which is also another related security improvement. I guess the only chance would be to completely upgrade bundler on older but still supported rubies like 2.7.

bundler/lib/bundler/definition.rb
```diff
@@ -287,50 +286,6 @@ def resolve
       end
     end
 
-    def index
```
rmacklin, line 290: @deivid-rodriguez, it seems that the removal of this method broke https://rubygems.org/gems/gemsurance as that method was used [here](https://github.com/appfolio/gemsurance/blob/2680f874c76247cf41fface3656b7764e817457c/lib/gemsurance/gem_info_retriever.rb#L122), resulting in a `NoMethodError` on bundler v2.2.18:
```
undefined method `index' for #<Bundler::Definition:0x00007fd7dd23a958> (NoMethodError)
```

Given that this was removed in a patch version of bundler (the method existed in bundler v2.2.17), I'm guessing it wasn't considered public API in the first place. But I figured I'd mention it in case it was okay to use the `index` method externally and this breakage was actually unintended.

deivid-rodriguez: Hei @rmacklin, thanks for letting me know.

Yeah, this is not public API. The method was using a merged index (gems coming from different sources being used indifferently), which was the underlying problem causing "dependency confusion". So I decided to remove it since I think it was making `bundle outdated` suggest unexpected (potentially insecure) gems.

I guess `gemasurance` can adapt https://github.com/rubygems/rubygems/pull/4609/commits/1d1d8a8f0816c9734cd0500da49df68cfa50bdc5 to its needs: use the same source as the currently resolved spec instead of using a merged source (note that commit caused a regression fixed here: https://github.com/rubygems/rubygems/commit/64168ebceb6d99b27c189ad7ec4819c9da69f502.

rmacklin: Thanks for the confirmation!

I think we may be able to replace `gemsurance` with a combination of `bundler-audit` and `bundle outdated`.

PR #4611, opened 2021-05-20
Get ruby-core CI back to green

deivid-rodriguez:

Our CI job testing with ruby-core environment is broken.

* Port https://github.com/ruby/ruby/commit/8e91b969df08b7a2eb27a5d6d38733eea42dc7ad from ruby-core, and make it compatible with psych 3 (our environment) and psych 4 (ruby-core environment).

### DIFF
--- a/lib/rubygems/test_case.rb
+++ b/lib/rubygems/test_case.rb
@@ -692,6 +692,28 @@ class Gem::TestCase < Test::Unit::TestCase
     path
   end
 
+  ##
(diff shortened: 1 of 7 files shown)


hsbt:
👍  This is good to start with Psych 4.0.0.

Maybe, We should use `Psych.load( ... permitted_class: [Gem::Specification, Symbol])` instead of `Pysch.unsafe_load`. I will update them after freezing specification of Psych 4.0.0.

PR #4616, opened 2021-05-24
Make `ruby -r` work with `bundle exec`

### LINKED ISSUE #4025: `bundle exec ruby` is not always equivalent to `ruby -rbundler/setup`.

If the `ruby` command includes other require flags, they might get applied to early before the `-rbundler/setup` flag has been applied.

We should probably special case `bundle exec ruby` and change it so that `-r` flags are applied in the right order. Possibly also take the chance to print a deprecation warning and recommend `ruby -rbundler/setup` instead.

```shell
# Make sure you don't have `pry` installed globally
mkdir /tmp/test
cd /tmp/test
echo -e 'source "https://rubygems.org"\ngem "pry"' > Gemfile
bundle config set --local path vendor/bundle
bundle install
bundle exec ruby -rpry -e1
```

```shell
bundle exec ruby -rpry -e1
```

Empty successful output.

### What actually happened?

An error like

> Traceback (most recent call last):
	1: from /home/deivid/.rbenv/versions/2.7.2/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:83:in `require'
/home/deivid/.rbenv/versions/2.7.2/lib/ruby/site_ruby/2.7.0/rubygems/core_ext/kernel_require.rb:83:in `require': cannot load such file -- pry (LoadError)

_Originally reported by @pocke in https://github.com/rubygems/rubygems/issues/3298#issuecomment-501358633_

deivid-rodriguez:

When `bundle exec''ing to `ruby`, and using `-r` flags, requiring gems in the `Gemfile` won't work because `bundler/setup` is required through `-r` options inside `RUBYOPT`, which are processed after `-r` CLI flags.

My fix is to special case `bundle exec ruby` and prepend `-rbundler/setup` to ruby CLI flags, so that gems in the `Gemfile` are made available _before_ other `-r` options are processed.

Note: The test I wrote for this revealed an issue with `Bundler.which` under Windows, which I also fixed.

Fixes #4025.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -512,14 +512,18 @@ EOF
     end
 
     def which(executable)
+      exts = (RbConfig::CONFIG["EXECUTABLE_EXTS"]&.split || []) | [RbConfig::CONFIG["EXEEXT"]]
+
       if File.file?(executable) && File.executable?(executable)
         executable
       elsif paths = ENV["PATH"]
         quote = '"'.freeze
         paths.split(File::PATH_SEPARATOR).find do |path|
           path = path[1..-2] if path.start_with?(quote) && path.end_with?(quote)
-          executable_path = File.expand_path(executable, path)
-          return executable_path if File.file?(executable_path) && File.executable?(executable_path)
+          exts.find do |ext|
+            executable_path = File.expand_path(executable + ext, path)
+            return executable_path if File.file?(executable_path) && File.executable?(executable_path)
+          end
         end
       end
     end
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I'm closing this since it doesn't work for shell scripts invoking ruby. I will extract some changes from here that still make sense (fixes to `Bundler.which` on Windows) to a separate PR. A better approach that works in general is being discussed at https://bugs.ruby-lang.org/issues/17883.

PR #4537, opened 2021-04-15
Easier gemspec syncing with ruby-core

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Syncying bundler with ruby-core is tricky. After https://github.com/ruby/ruby/commit/0d9496f924d36534bd524791554d49dc0026b0e0, it can be done through `ruby tool/sync_defaullt_gems.rb bundler` without manual modifications, but if we make changes to the gemspec, it won't be ported properly. For example, bumping the minimum required ruby version or other similar changes.

## What is your fix for the problem, implemented in this PR?

With these tweaks to our gemspec, and a patch to the ruby sync tool (https://github.com/ruby/ruby/blob/26888d5e032202328e10881550477fd036c8e805/tool/sync_default_gems.rb#L97-L108) like this:

```diff
diff --git a/tool/sync_default_gems.rb b/tool/sync_default_gems.rb
index 9c8afecf10..3a250cc075 100644
--- a/tool/sync_default_gems.rb
+++ b/tool/sync_default_gems.rb
@@ -98,12 +98,18 @@ def sync_default_gems(gem)
     rm_rf(%w[lib/bundler lib/bundler.rb libexec/bundler libexec/bundle spec/bundler tool/bundler/*])
     cp_r(Dir.glob("#{upstream}/bundler/lib/bundler*"), "lib")
     cp_r(Dir.glob("#{upstream}/bundler/exe/bundle*"), "libexec")
-    cp_r("#{upstream}/bundler/bundler.gemspec", "lib/bundler")
+
+    gemspec_content = File.readlines("#{upstream}/bundler/bundler.gemspec").map do |line|
+      next if line =~ /extra_rdoc_files/
+
+      line.gsub("bundler.gemspec", "lib/bundler/bundler.gemspec").gsub('"exe"', '"libexec"')
+    end.compact.join
+    File.write("lib/bundler/bundler.gemspec", gemspec_content)
+
     cp_r("#{upstream}/bundler/spec", "spec/bundler")
     cp_r(Dir.glob("#{upstream}/bundler/tool/bundler/test_gems*"), "tool/bundler")
     cp_r(Dir.glob("#{upstream}/bundler/tool/bundler/rubocop_gems*"), "tool/bundler")
     cp_r(Dir.glob("#{upstream}/bundler/tool/bundler/standard_gems*"), "tool/bundler")
-    `git checkout lib/bundler/bundler.gemspec`
[...]

PR #4618, opened 2021-05-25
Prepare rubygems 3.2.18 and bundler 2.2.18

deivid-rodriguez:
Let's get another round of releases out there.


deivid-rodriguez:
Running `ruby-core` CI in the stable branch forces us to backport PRs like https://github.com/rubygems/rubygems/pull/4537 or https://github.com/rubygems/rubygems/pull/3141 in order to keep CI geen here. Even though these PRs should not include any user facing changes, it seems unnecessary to do this, since we might want to skip this job in the stable branch.

simi (approved):

PR #4619, opened 2021-05-25
Better defaults for GitHub Actions template generated by `bundle gem`

### LINKED ISSUE #4230: Template for github actions recommendations
1)
```
with:
        ruby-version: 3.0.0
```

Should be using a matrix, even if you only want a single version, as it makes it easier to add versions.
Morevover, I feel that just the current stable ruby version is not the best default. My recommendation would be current and two preceding, but I leave that up to the maintainers.
More importantly, I believe that `ruby-head` should be there by default. This encourages people to report issues they might encounter with Ruby as it gets developed, provide valuable feedback on an upcoming incompatibilities, or simply prepare their code for the next version.

2)
```
# replace
on: [push,pull_request]
# with
on:
  push:
    branches:
    - master
  pull_request:
```

Otherwise, CI runs twice, once when the branch is pushed, and once when the PR is created. `master` should be replaced by whatever the default main branch in the git config.

See https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662

NeimadTL:

Some room for improvement on the default github actions template.

As the CI was running two times in a row, once for push and once
again for pull, we decided to make different section for each event.
We're using matrix as it makes it easier to add versions and also using
the default main branch in the git config from now by the use of the
:git_default_branch key...

Fixes #4230.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -9,12 +9,18 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        ruby:
+          - <%= RUBY_VERSION %>
+
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: <%= RUBY_VERSION %>
+        ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
     - name: Run the default task
       run: bundle exec rake
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -1,6 +1,10 @@
 name: Ruby
 
-on: [push,pull_request]
+on:
+  push:
+    - <%= config[:git_default_branch] %>
+
+  pull_request:
 
 jobs:
   build:


bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -10,7 +18,7 @@ jobs:
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: <%= RUBY_VERSION %>
+        ruby-version: ${{ matrix.ruby }}
```
simi, line 21: I would prefer to keep `RUBY_VERSION` in here, since I think there is no need for any matrix to support 1 particular version only. Alternatively (if we would like to keep matrix) it would be enough to don't hardcode `3.0.0` in the template and use `RUBY_VERSION` in the matrix instead.

NeimadTL: @simi , I'll probably add other versions. I asked @deivid-rodriguez, if we'll take this [comment](https://github.com/rubygems/rubygems/issues/4230#issue-778568037) into account or not. Would appreciate if you can give your thoughts on that as well. You're totally right for the hardcoded version, will update it.

MSP-Greg: You're thinking something like the following?
```yaml
strategy:
  matrix:
    ruby: [ <%= RUBY_VERSION %> ]
```

At least this leaves the template such that the matrix could be expanded.

Since most gems support a range of rubies,  one could...

NeimadTL: @MSP-Greg , yep something like that with other versions in it: current, the two preceding ones and ruby-head

deivid-rodriguez: yes, the rationale for moving to a matrix is that it makes it easier to add new version, but on the related issue @marcandre also suggests adding `ruby-head`, and also a couple more stable rubies. In my opinion it's a good idea. @simi What do you think?

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -1,6 +1,14 @@
 name: Ruby
 
-on: [push,pull_request]
+strategy:
+  matrix:
+    ruby: 3.0.0
+
+on:
+  push:
+    config[:git_default_branch]
```
deivid-rodriguez, line 9: This should be an array including only the default branch, which you need to wrap with `<%= %>`. So

```suggestion
    - <%= config[:git_default_branch] %>
```

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
deivid-rodriguez: Ok, let's start with a matrix including the current ruby and `ruby-head` I guess.

deivid-rodriguez: Actually `ruby-head` should go into `allowed_failures` by default, and the configuration might get tricky... Maybe let's leave just `RUBY_VERSION` for now and discuss how to add `ruby-head` separately?

PR #4624, opened 2021-05-26
Simplify `setup.rb` code

deivid-rodriguez:

The previous way was hitting a bug in truffleruby, make `ruby setup.rb` crash under than implementation.

Do things in a simpler way.

### DIFF
--- a/setup.rb
+++ b/setup.rb
@@ -11,8 +11,7 @@ if ENV['RUBYOPT'] or defined? Gem
   ENV.delete 'RUBYOPT'
 
   require 'rbconfig'
-  cmd = [RbConfig.ruby, 'setup.rb', *ARGV].compact
-  cmd[1,0] = "--disable-gems"
+  cmd = [RbConfig.ruby, '--disable-gems', 'setup.rb', *ARGV].compact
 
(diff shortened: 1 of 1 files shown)


setup.rb
```diff
@@ -11,8 +11,7 @@
   ENV.delete 'RUBYOPT'
 
   require 'rbconfig'
-  cmd = [RbConfig.ruby, 'setup.rb', *ARGV].compact
-  cmd[1,0] = "--disable-gems"
+  cmd = [RbConfig.ruby, '--disable-gems', 'setup.rb', *ARGV].compact
```
simi, line 14: :thinking: Do you have any idea why `compact` is needed at all? `RbConfig.ruby` will never be `nil` and even if it would be `nil`, it will fail later. `ARGV` is an array (but even it would be `nil`, deconstruction will handle it properly).

simi (approved):

setup.rb
deivid-rodriguez: It seems completely unnecessary indeed, I can remove that too since we're at it.

deivid-rodriguez: Done!

Issue #4588, opened 2021-05-10
Bundler 2.2.17 fails with "Could not find sorbet-static-0.5.6403-universal-darwin-20 in any of the sources" on Linux on TruffleRuby


eregon:
### Describe the problem as clearly as you can

When running on truffleruby-head/dev, which you can install with ` ruby-build truffleruby-dev ~/.rubies/truffleruby-dev`.
There is a change to mark the binary-only `sorbet-static` gem as fine to install on TruffleRuby, this is needed to reproduce, it should merge soon, I'll ping here when it's done and there is a build with it.

Using this Gemfile:
```ruby
source 'https://rubygems.org'

gem 'sorbet', '0.5.6403'
gem 'libv8', '8.4.255.0'
```

The resolution seems to go wrong, because the generated Gemfile.lock on TruffleRuby is:
```
GEM
  remote: https://rubygems.org/
  specs:
    libv8 (8.4.255.0)
    sorbet (0.5.6403)
      sorbet-static (= 0.5.6403)
    sorbet-static (0.5.6403-universal-darwin-14)
    sorbet-static (0.5.6403-universal-darwin-15)
    sorbet-static (0.5.6403-universal-darwin-16)
    sorbet-static (0.5.6403-universal-darwin-17)
    sorbet-static (0.5.6403-universal-darwin-18)
    sorbet-static (0.5.6403-universal-darwin-19)
    sorbet-static (0.5.6403-universal-darwin-20)
    sorbet-static (0.5.6403-x86_64-linux)

PLATFORMS
  ruby

DEPENDENCIES
  libv8 (= 8.4.255.0)
  sorbet (= 0.5.6403)

BUNDLED WITH
   2.2.17
```

Compare that to CRuby where it is:
```
GEM
  remote: https://rubygems.org/
  specs:
    libv8 (8.4.255.0-x86_64-linux)
    sorbet (0.5.6403)
      sorbet-static (= 0.5.6403)
    sorbet-static (0.5.6403-x86_64-linux)

PLATFORMS
  x86_64-linux

DEPENDENCIES
  libv8 (= 8.4.255.0)
  sorbet (= 0.5.6403)

BUNDLED WITH
   2.2.17
```

So when doing `bundle install` on Bundler 2.2.17 on TruffleRuby it fails with:
```
$ bundle install --verbose
Running `bundle install --verbose` with bundler 2.2.17
Found no changes, using resolution from the lockfile
[...]

PR #4625, opened 2021-05-26
Fix `bundle install` on truffleruby selecting incorrect variant for `sorbet-static` gem

deivid-rodriguez:

Bundler selecting incorrect variants on truffleruby.

Make sure we go through custom truffleruby logic: https://github.com/oracle/truffleruby/blob/72ce69ec079618a3c0513da968262a4484633834/lib/mri/rubygems/platform.rb#L44.

Fixes https://github.com/rubygems/rubygems/issues/4588.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -176,12 +176,36 @@ module Gem
     end
   end
 
+  require "rubygems/platform"
+
   class Platform
     JAVA  = Gem::Platform.new("java") unless defined?(JAVA)
     MSWIN = Gem::Platform.new("mswin32") unless defined?(MSWIN)
     MSWIN64 = Gem::Platform.new("mswin64") unless defined?(MSWIN64)
     MINGW = Gem::Platform.new("x86-mingw32") unless defined?(MINGW)
     X64_MINGW = Gem::Platform.new("x64-mingw32") unless defined?(X64_MINGW)
+  end
+
+  Platform.singleton_class.module_eval do
+    unless Platform.singleton_methods.include?(:match_spec?)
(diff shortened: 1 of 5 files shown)


eregon:
@deivid-rodriguez Do you need any help with this PR, anything I can do?

From a quick look at the CI there is some lint errors, and some other failures which I'm not sure are related to changes here.

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -174,20 +174,36 @@ def hash
...
     MINGW = Gem::Platform.new("x86-mingw32") unless defined?(MINGW)
     X64_MINGW = Gem::Platform.new("x64-mingw32") unless defined?(X64_MINGW)
+  end
 
-    undef_method :hash if method_defined? :hash
-    def hash
-      @cpu.hash ^ @os.hash ^ @version.hash
-    end
+  Platform.singleton_class.module_eval do
+    unless Platform.singleton_methods.include?(:match_spec?)
```
eregon, line 188: Indeed, they were all introduced at the same time in https://github.com/rubygems/rubygems/pull/3817/files#diff-b2eac061756123f15f8d8e41075e4b8466a60945cad4e594b4eadf241d1a4089
So to backfill them it sounds like indeed all 3 need to be backfilled.

deivid-rodriguez:
Last time I looked they seemed related, I'll try to find some time to look at it but you can have a look yourself if you want too.

deivid-rodriguez:
@eregon I had another look and I think I was wrong and the problem was indeed unrelated to this PR. And in any case, I think it should've been workarounded by #4736. So I rebased this and let's see how it goes.

eregon:
Thanks :)

PR #4629, opened 2021-05-27
Try running CI on mswin

deivid-rodriguez:

Because ruby-core is now running it, so they will introduce patches directly to keep it green that we will then revert when we sync upstream code. For example: https://github.com/ruby/ruby/commit/cc1ea6e0238a36c08a6255a6ac7a68c54ccafa4a.

Try running against mswin in CI.


MSP-Greg:
@deivid-rodriguez :thumbsup: Good idea!

JFYI, re the skipped 'extension' related tests, the ruby/setup-ruby repo installs JSON in it's CI, and mswin passes there, and has done so for a long time.  I've never looked into why there are issues with the RG tests.  I also installed it locally (`gem install json:2.2.0`), and it installed correctly.

Also, sometimes issues are mswin specific, other times they are Windows specific (both mingw & mswin).  Maybe consider also testing on mingw?

Lastly, when Ruby 3.1 is released, there may be an additional platform for Windows, possibly called 'ucrt'.

Thanks...

deivid-rodriguez:
The `mswin` entry don't really add much overhead to our CI and it will further discourage ruby committers from applying patches directly by reducing the CI failures they get, so I'll merge this to get back in sync with ruby-core copy.

MSP-Greg:
@deivid-rodriguez

I have rust installed on WSL2/Ubuntu, I just now installed it on Windows.  Haven't used it but...

There are two cargo targets available for x86_64 Windows, `x86_64-pc-windows-gnu`, and `x86_64-pc-windows-msvc`.

I suspect that `x86_64-pc-windows-gnu` is best for all the 'normal' Rubies, but the target `x86_64-pc-windows-msvc` might work better with the mswin build.

Also, [rustup](https://github.com/rust-lang/rustup/blob/master/.github/workflows/windows-builds-on-master.yaml) seems to be running CI with both.

So, just checking to see if you've tried it?

deivid-rodriguez:
If I've tried what, sorry?

PR #4642, opened 2021-06-01
regenerate lockfile

doodzik:

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -642,6 +642,8 @@ module Bundler
     desc "lock", "Creates a lockfile without installing"
     method_option "update", :type => :array, :lazy_default => true, :banner =>
       "ignore the existing lockfile, update all gems by default, or update list of given gems"
+    method_option "regenerate", :type => :boolean, :lazy_default => true, :banner =>
+      "regenerate the lockfile with the current version of bundler, doesn't update gems"
     method_option "local", :type => :boolean, :default => false, :banner =>
       "do not attempt to fetch remote gemspecs and use the local gem cache only"
     method_option "print", :type => :boolean, :default => false, :banner =>
--- a/bundler/lib/bundler/cli/lock.rb
+++ b/bundler/lib/bundler/cli/lock.rb
@@ -25,6 +25,7 @@ module Bundler
         Bundler::CLI::Common.ensure_all_gems_in_lockfile!(update)
         update = { :gems => update, :lock_shared_dependencies => options[:conservative] }
       end
+      definition_old = Bundler.definition
       definition = Bundler.definition(update)
 
       Bundler::CLI::Common.configure_gem_version_promoter(Bundler.definition, options) if options[:update]
@@ -47,6 +48,24 @@ module Bundler
       end
 
       definition.resolve_remotely! unless options[:local]
+
+      if options[:regenerate]
+        # re-read the Gemfile definitions
+        definition = Bundler.definition(true)
+        definition.gem_version_promoter.level = :patch
+        definition.resolve_remotely!
+
+        # resolve to a spec set
+        spec_set = definition.resolve
+        # update the version in the new spec_set to match the old versions
+        definition_old.specs.each do |spec|
+          gem_spec = spec_set[spec.name].first
+          if gem_spec
+            gem_spec.instance_variable_set(:@version, spec.version)
+            gem_spec.instance_variable_set(:@dependencies, spec.dependencies)
+          end
+        end
+      end
 
       if print
         puts definition.to_lock
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -115,7 +115,7 @@ module Bundler
       @disable_multisource = @locked_gem_sources.all?(&:disable_multisource?)
 
       unless @disable_multisource
-        msg = "Your lockfile contains a single rubygems source section with multiple remotes, which is insecure. You should run `bundle update` or generate your lockfile from scratch."
+        msg = "Your lockfile contains a single rubygems source section with multiple remotes, which is insecure. You should run `bundle lock --regenerate` or generate your lockfile from scratch."
 
         Bundler::SharedHelpers.major_deprecation 2, msg
 
--- a/bundler/spec/commands/lock_spec.rb
+++ b/bundler/spec/commands/lock_spec.rb
@@ -536,13 +536,12 @@ RSpec.describe "bundle lock" do
     bundle "lock --add-platform x86_64-linux", :artifice => "compact_index", :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }
   end
 
-  context "with multiple, duplicated sources, with lockfile in old format" do
+  describe "with --regenerate" do
     before do
       build_repo2 do
         build_gem "dotenv", "2.7.6"
 
         build_gem "oj", "3.11.3"
-        build_gem "oj", "3.11.5"
 
         build_gem "vcr", "6.0.0"
       end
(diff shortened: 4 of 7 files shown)


bundler/lib/bundler/cli.rb
```diff
@@ -642,6 +642,8 @@ def inject(name, version)
     desc "lock", "Creates a lockfile without installing"
     method_option "update", :type => :array, :lazy_default => true, :banner =>
       "ignore the existing lockfile, update all gems by default, or update list of given gems"
+    method_option "regenerate", :type => :boolean, :lazy_default => true, :banner =>
+      "regenerate the lockfile with the current version of bundler, doesn't update gems"
```
simi, line 646: Isn't this already possible using `bundle update --bundler`?

doodzik: Currently, the `bundle update --bundler` command only bumps the version in the `Gemfile.lock` file, but doesn't rebuild the lockfile. AFAIK the [`bundle update --bundler`](https://bundler.io/v2.0/man/bundle-update.1.html) command had been used to do that when [bundler upgraded from version 1 to 2](https://bundler.io/guides/bundler_2_upgrade.html#upgrading-applications-from-bundler-1-to-bundler-2). However, upgrading the lockfile without bumping the gem versions doesn't work for later versions, which we ("Shopify") ran into when we tried to upgrade our services to [adopt the bundler version that locks gem sources in the lockfile separately](https://github.com/rubygems/rubygems/pull/4381#issuecomment-827640511).

@pfeffer and I spent some time trying to fix the `bundle update --bundler` command not to resolve a gem version, but we couldn't make it work.
Screen Shot 2021-06-02 at 10 18 45 AM

For the current approach, we decided to move it into the lock command for two reasons:
1. The code of our current approach doesn't really fit into the update command because it has little overlap.
2. From a user perspective, the `Gemfile.lock` regeneration seems more related to the `lock` than the `update` command. The comments on this [stackoverflow topic](https://stackoverflow.com/questions/30635292/install-newer-version-of-bundler-with-bundler) highlights how users get confused by the intention of the `bundler update --bundler` command, which IMO suggests that the lockfile regeneration shouldn't be placed in the `update` command.

deivid-rodriguez: The approach here (if I understand it correctly) is brittle and might not always work because it assumes that the current versions of each gem will always be present in the sources resolved for the updated version of each gem. For example, it might happen that `bundle update` changed sources for a gem because a newer version of a dependant dropped the dependency on it.

This is why I feel a adapting a script like then one suggested in https://github.com/rubygems/rubygems/pull/4381#issuecomment-828992246 is best. We can even link to the gist in the deprecation message, warning that it's a convenience script but that the result might still need some tweaking.

deivid-rodriguez: Actually, I'm thinking that since we have fixed the remaining dependency confusion issues in bundler 2.2.19, and now we resolve sources for each gem name before actual resolution, I _think_ removing this line might actually do the trick now: https://github.com/rubygems/rubygems/blob/270f8f9f2ea39d2e35cb1063b88dcee26cfaf070/bundler/lib/bundler/definition.rb#L122, and will regenerate the lockfile with separated rubygems sources but with the same versions.

We do have to keep that logic in frozen mode but if removing it does the trick, it seems fine for the rest of the cases.

deivid-rodriguez: I tried this patch:

```diff
diff --git a/bundler/lib/bundler/definition.rb b/bundler/lib/bundler/definition.rb
index d0c8b90b78..569ed3a2bc 100644
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -112,7 +112,7 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
       end
 
       @locked_gem_sources = @locked_sources.select {|s| s.is_a?(Source::Rubygems) }
-      @disable_multisource = @locked_gem_sources.all?(&:disable_multisource?)
+      @disable_multisource = @locked_gem_sources.all?(&:disable_multisource?) || (sources.no_aggregate_global_source? && !Bundler.frozen_bundle?)
 
       unless @disable_multisource
         msg = "Your lockfile contains a single rubygems source section with multiple remotes, which is insecure. You should run `bundle update` or generate your lockfile from scratch."
```

But while it works for some simple cases, it seems to still upgrade some gems sometimes. I would need to investigate why.

pfeffer: We did try taking that line out and running `bundle update --bundler`, and it did split the sources into different groups, but it also bumped the versions on quite a few gems. 

We also ran the same command in `frozen` mode and this is the error message: 
```bash
❯ RUBYOPT=-rpry-byebug dbundle update --bundler
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the some_repo/Gemfile freeze
by running `bundle config unset frozen`.

The list of sources changed

You have added to the Gemfile:
* source: rubygems repository https://example.com/ruby/ or installed locally
* source: rubygems repository https://rubygems.org/ or installed locally

You have deleted from the Gemfile:
* source: rubygems repository https://example.com/ruby/, https://rubygems.org/ or installed locally
```

@deivid-rodriguez  do you think there's something else that we are missing?

paracycle: > This is why I feel a adapting a script like then one suggested in [#4381 (comment)](https://github.com/rubygems/rubygems/pull/4381#issuecomment-828992246) is best. We can even link to the gist in the deprecation message, warning that it's a convenience script but that the result might still need some tweaking.

I am not sure I understand what the difference is between parsing the output of `bundle update` and putting the versions back by manipulating `Gemfile.lock` as a text file (what the linked script does), and detecting the changed versions of gems after an update internally and setting the versions back to the old versions and writing out `Gemfile.lock` (what this PR does).

This PR should essentially be doing the same thing as the script you are linking to. Am I missing something?

deivid-rodriguez: @pfeffer Removing that line was just a quick hack to see if the idea was viable, but the more correct patch following that idea would be the one I mentioned at https://github.com/rubygems/rubygems/pull/4642#discussion_r644247887. That should avoid the issue you got in frozen mode, but still has the issue of some versions being unintentionally updated, which I'm currently investigating.

deivid-rodriguez: @paracycle Yeah, both solutions are incorrect in general, they are not different in that regard. It just feels wrong to ship a new flag to `bundle lock` that doesn't really work in general and that it's specific to this particular migration, so I prefer to link to a gist with the script and warn that it might get most of the job done but not all gems might get the correct source.

That said, I'm pretty confident this can be fixed in general using the idea I mentioned at https://github.com/rubygems/rubygems/pull/4642#discussion_r644247887. I'm currently testing one little extra tweak that might do the trick.

deivid-rodriguez: Yeah, I think the following patch should do the trick:

```diff
diff --git a/bundler/lib/bundler/definition.rb b/bundler/lib/bundler/definition.rb
index d0c8b90b78..569ed3a2bc 100644
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -112,7 +112,7 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
       end
 
       @locked_gem_sources = @locked_sources.select {|s| s.is_a?(Source::Rubygems) }
-      @disable_multisource = @locked_gem_sources.all?(&:disable_multisource?)
+      @disable_multisource = @locked_gem_sources.all?(&:disable_multisource?) || (sources.no_aggregate_global_source? && !Bundler.frozen_bundle?)
 
       unless @disable_multisource
         msg = "Your lockfile contains a single rubygems source section with multiple remotes, which is insecure. You should run `bundle update` or generate your lockfile from scratch."
diff --git a/bundler/lib/bundler/source_list.rb b/bundler/lib/bundler/source_list.rb
index 584d693dea..642380428a 100644
--- a/bundler/lib/bundler/source_list.rb
+++ b/bundler/lib/bundler/source_list.rb
@@ -185,6 +185,8 @@ def equal_sources?(lock_sources, replacement_sources)
     end
 
     def equal_source?(source, other_source)
+      return source.include?(other_source) if source.is_a?(Source::Rubygems) && other_source.is_a?(Source::Rubygems) && !merged_gem_lockfile_sections?
+
       source == other_source
     end
 
```

With that, `bundle install` should automatically update the lockfile format when run in non frozen mode.

Can you try it?

doodzik: > With that, bundle install should automatically update the lockfile format when run in non frozen mode.

I ran the changes on multiple repositories and all of them had issues resolving gems across different sources. The repos I ran the command on didn't have issues with installing/updating gems before so I assume the Gemfile is correctly configured.
```
master ❯ dbundle install
Fetching gem metadata from https://example_private_registry.com/gems/....
Resolving dependencies...
Bundler could not find compatible versions for gem "rails":
  In snapshot (Gemfile.lock):
    rails (= 5.2.5)

  In Gemfile:
    rails (~> 5.2.4)

    gem_from_private_source (~> 103.0.0) was resolved to 103.0.1, which depends on
      rails (>= 5.0.0)

```

deivid-rodriguez: Mmmmm... that's disappointing :sob:.

I don't know why that's happening. What's explained there seems exactly what we want to happen so I don't know why a resolution error is being raised. Furthermore, I don't know why I didn't get this on any tests (both automated and manual) I made whereas you're getting it in all your projects.

I'll look into it more, if you can share a reproducible example that would be helpful.

deivid-rodriguez: Hi again! I found another issue that _could_ be affecting this, I'm working on it at #4652. Would you be able to apply that patch on top of the suggested patch and see if that fixes things? A simpler version of #4652 which would get you the same effect would be:

```diff
diff --git a/bundler/lib/bundler/definition.rb b/bundler/lib/bundler/definition.rb
index d0c8b90b78..0e92261895 100644
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -917,7 +917,7 @@ def compute_requires
     end
 
     def additional_base_requirements_for_resolve
-      return [] unless @locked_gems
+      return [] # unless @locked_gems
       dependencies_by_name = dependencies.inject({}) {|memo, dep| memo.update(dep.name => dep) }
       @locked_gems.specs.reduce({}) do |requirements, locked_spec|
         name = locked_spec.name
```

pfeffer: Thank you so much for looking at this issue!
I tried adding the suggested change on top of the other changes in this thread on 4 different repos, and all of them had issues resolving gems with the error messages similar to what @doodzik posted above 😢

deivid-rodriguez: :disappointed:

deivid-rodriguez: Alright, third time is the charm! Can you try #4647 now, and verify whether it properly updates your lockfiles without bumping any versions? I'm even more confident this time because CI is now green :smiley:.

doodzik: @deivid-rodriguez I tried the changes on three of our repos that ran into edge cases with our tooling, and it worked flawlessly 🎉 Thank you for pushing this through 🙇 

On a bit of a related note, we are trying to prevent a service from being deployed with merged gem sections in the lockfile for security reasons (dependency confusion). I opened a PR that introduces a flag to make `bundle install` fail if that is the case, but I'm not sure if it is the best way of achieving it and how it fits into bundler's long-term strategy. https://github.com/rubygems/rubygems/pull/4672

doodzik:
closing in favour of #4647

bundler/lib/bundler/cli.rb
deivid-rodriguez: Awesome! I'll get to #4672 soon!

PR #4643, opened 2021-06-02
Rename test/rubygems/test_{case,utilities}.rb to avoid "test_" prefix

mame:
This changes "test/rubygems/test_case.rb" to "test/rubygems/helper.rb",
and "test/rubygems/test_utilities.rb" to "test/rubygems/utilities.rb".

The two files are a helper for tests, not test files. However, a file
starting with "test_" prefix is handled as a test file directly loaded
by test-unit because Rakefile specifies:

```
```

Directly loading test/rubygems/test_utilities.rb caused "uninitialized
constant Gem::TestCase". This issue was fixed by
59c682097197fee4052b47e4b4ab86562f3eaa9b, but the fix caused a
"circular require" warning because test_utilities.rb and test_case.rb
are now requiring each other.

Anyway, adding "test_" prefix to a test helper file is confusing, so
this changeset reverts the fix and solve the issue by renaming them.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -526,6 +526,7 @@ test/rubygems/good_rake.rb
 test/rubygems/grandchild_cert.pem
 test/rubygems/grandchild_cert_32.pem
 test/rubygems/grandchild_key.pem
+test/rubygems/helper.rb
 test/rubygems/installer_test_case.rb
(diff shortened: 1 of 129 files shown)


deivid-rodriguez (approved):
Nice catch, thanks so much @mame!

PR #4647, opened 2021-06-03
Auto-update insecure lockfile to split GEM source sections whenever possible

deivid-rodriguez:

It's currently hard to update to the new lockfile format without updating all your gems at the same time. Bundler should do this for you whenever possible.

Even if the lockfile has merged GEM sections in the lockfile, we can still use separated sources and auto-update the format unless we are in frozen mode or the Gemfile actually has multiple global sources.


deivid-rodriguez:
Merging since @doodzik confirmed it works for them and it makes bundler more secure by dealing with insecure lockfiles instead of just warning them.

PR #4651, opened 2021-06-04
Check requirements class before loading marshalled requirements

nobu:
Mitigate the security risk:
https://devcraft.io/2021/01/07/universal-deserialisation-gadget-for-ruby-2-x-3-x.html

Unexpected method (`Kernel#system` in the article) can be called by just loading marshaled data.

Check the classes of `@requirements` and its elements.

Maybe `fix_syck_default_key_in_requirements` is no longer used and can be removed, but not sure.


deivid-rodriguez:
Hi @nobu, thanks!

I think this is a good opportunity to move https://github.com/rubygems/rubygems/pull/3797 forward. I do think `fix_syck_default_key_in_requirements` can be removed at this point. There was some theoretical backwards compatibility concerns, but we never really clarified them, so if this is a security risk I think we can just remove the vulnerable code.

@hsbt What do you think?

deivid-rodriguez:
@nobu After having another look at this, even if we remove `fix_syck_default_key_in_requirements`, we are still vulnerable to this, with the difference that we would only need to check that `@requirements` is an `Array` upon marshal loading. Correct?

nobu:
Correct, this doesn't fix the risk totally, but just "mitigate" for old ruby versions.

deivid-rodriguez:
I do believe that we can remove all `syck` traces from `rubygems` at this point, and it does simplify this patch, so I included removing the `fix_syck_default_key_in_requirements` method in this PR.

deivid-rodriguez:
I do believe this shouldn't break anything, so I'll take this opportunity to clean things up and improve security at the same time. I don't think this is really exploitable so I tagged it just as an enhancement.

deivid-rodriguez:
Thanks @nobu!

PR #4653, opened 2021-06-05
Ignore any gemfiles in parent directories

nobu:

When `Gemfile` and its lockfile exist in any upper directories, e.g., `~/Gemfile.lock`, the lockfile's content is read and the `Gem::BundlerVersionFinder` related tests fail.

By setting `ENV["BUNDLE_GEMFILE"]` to an nonexistent file path under an empty directory, skip to traverse parent directories.

Because this is a bug-fix of tests, no new tests.

### DIFF
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -546,6 +546,10 @@ class Gem::TestCase < Test::Unit::TestCase
     Gem.pre_uninstall_hooks.clear
   end
 
+  def without_any_upwards_gemfiles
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (commented):
Sounds good!

Should we extract this logic to a helper at `Gem::TestCase`, like `with_non_existent_gemfile_to_avoid_upwards_gemfile_search(&block)` or something like that?

nobu:
As seems to be needed in nearly all tests, extracted to the method to be called in `setup`.

deivid-rodriguez (approved):
Thanks @nobu

PR #4658, opened 2021-06-07
Add better specification policy error description

ceritium:

Make easier find what key of metadata from `gemspec` is invalid. 

Print `metadata['#{key}']` for string type and length value checks like for link keys.

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -124,6 +124,7 @@ class Gem::SpecificationPolicy
     end
 
     metadata.each do |key, value|
+      entry = "metadata['#{key}']"
       if !key.kind_of?(String)
         error "metadata keys must be a String"
       end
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
I find this a useful developer experience improvement.

It would shorten the time it took for a user to correct their mistake.

lib/rubygems/specification_policy.rb
```diff
@@ -133,16 +134,16 @@ def validate_metadata
       end
 
       if !value.kind_of?(String)
-        error "metadata values must be a String"
+        error "#{entry} value must be a String"
       end
 
       if value.size > 1024
-        error "metadata value too large (#{value.size} > 1024)"
+        error "#{entry} value too large (#{value.size} > 1024)"
```
simi, line 141: What about to fix this as well? `...value is too large...`

olleolleolle: Ah, now I see, to improve the grammar of the wording, add "is". 👍

deivid-rodriguez (approved):
Thank you! :muscle:

PR #4672, opened 2021-06-14
[PoC] add flag to raise an error for a multisource lockfile

doodzik:

Bundler 2.2.18 fixed an issue regarding dependency confusion https://github.com/rubygems/rubygems/pull/4609. However, there isn't an easy way to make sure that a Gemfile.lock file adheres to the new format. We would like the option to fail any `bundle install` that encounters the old multisource lockfile format. This is particularly useful for our CI and CD environment since it would prevent a compromised service to be deployed.

I'm introducing a new flag `fail-on-multisource`, that checks whether multiple sources have been defined for the same gems. If that is the case a failure is being raised.

Is this the best method of solving it? I saw there is [some code behind a feature flag](https://github.com/rubygems/rubygems/blob/96e5cff3df491c4d943186804c6d03b57fcb459b/bundler/lib/bundler/dsl.rb#L463-L468) that requires sources to be specified in the Gemfile. If bundler 3 is planned on deprecating the old lockfile format maybe it makes sense to [raise an issue in the definition file](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/definition.rb#L117-L123) by providing a feature flag?

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -232,6 +232,8 @@ module Bundler
       "Specify a different path than the system default ($BUNDLE_PATH or $GEM_HOME).#{" Bundler will remember this value for future installs on this machine" unless Bundler.feature_flag.forget_cli_options?}"
     method_option "quiet", :type => :boolean, :banner =>
       "Only output warnings and errors."
+    method_option "fail-on-multisource", :type => :boolean, :banner =>
+      "Fails bundle install if lockfile contains multiple primary sources"
     method_option "shebang", :type => :string, :banner =>
       "Specify a different shebang executable name than the default (usually 'ruby')"
     method_option "standalone", :type => :array, :lazy_default => [], :banner =>
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -57,6 +57,13 @@ module Bundler
       definition = Bundler.definition
       definition.validate_runtime!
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
If I understand correctly (since now lockfiles can be easily upgraded now), what you want to prevent here is some developer using an insecure version of bundler accidentally reverting to the old format, and that insecure lockfile being deployed.

Similarly to #4642, I don't want to introduce a new feature just for this particular case. I think there are more generic features to achieve this. These is what I have in mind:

* Bundler version locking. A lockfile generated with bundler x.y.z should always be installed by bundler x.y.z. This is already the case if you make sure `bundler-x.y.z` is installed before running `bundle install`. But our plan is to make this automatic so that you don't need an extra `gem install bundler:x.y.z` step.

* A `--deployment` flag to `bundle check`, that would check whether running `bundle install` in non frozen mode would create changes in the lockfile and fail otherwise.

* A configuration to make bundler fail on warnings/deprecations. Could be even more fine-grained and allow per-warning/deprecation configuration.

I think these features would give you want you need but also be more useful in general.

However, I've been thinking more about this particular situation, I don't think it makes sense to sacrifice security for backwards compatibility. So I'm leaning towards completely avoiding installing insecure gemfiles and lockfiles, no matter what, and just fail when we encounter them.

deivid-rodriguez:
I'll close this in favour of the other alternatives I mentioned, thanks for bringing this to our attention!

Issue #4649, opened 2021-06-04
Allow some "backdoor" gems available under 'bundle exec` regardless of Gemfile


mame:
How about providing a "backdoor" way to make some special gems available even if they are not written in Gemfile?

I frequently use the `stackprof` gem to profile Ruby apps and gems. However, to use `stackprof` under `bundle exec`, I need to modify Gemfile to add `gem 'stackprof'` every time. After profiling and performance improvement, I need to be careful to revert the modification before commit. This repeated process is, honestly, annoying and error-prone to me. (If there is already a way to temporarily make some special gems available without Gemfile modification, sorry but let me know.)

My current motivation is `stackprof`, but this is not limited to the gem. There are some gems that we want to use only during development. Currently, @ko1 is creating a new `debug` gem, which we will want to use with no code modification. Also, this is related to https://bugs.ruby-lang.org/issues/16778; if it is easy and useful enough to make a gem "special", we may promote some default gems (for example, irb) to bundle gems.

My naive idea to use an environment variable. For example, `bundle exec` can make gems listed in `BUNDLER_BACKDOOR_GEMS`.

```
export BUNDLER_BACKDOOR_GEMS=stackprof,debug,irb
```

Another idea is `~/.bundle/config`, or something.

I know that my proposal reduces the environment reproducibility which is the main motivation of Bundler. Some users may use this backdoor in development environment, unintentionally write code that depends on the gems, commit and deploy the code, but the code does not work in production. So, this is a trade-off between development usability and production robustness. Thus, I call my proposal as "backdoor" to notice users that this feature is somewhat dangerous.

However, in recent practice, this kind of issues (unintentionally deploying code) will be typically visible in CI environment. Rather, in this regard, unintentionally committing a temporal modification of Gemfile to add `gem "stackprof"` or `gem "debug"` or something is much worse because it may pass the CI test and bring big performance overhead in production.

What do you think?

duckinator:
The entire purpose of `bundle exec` is to create an isolated environment. From `man bundle-exec`:

> **bundle exec** makes a number of changes to the shell environment, then
> executes the command you specify in full.
>
> [...]
>
>  It also modifies Rubygems:
>  •   disallow loading additional gems not in the bundle
> [...]

Interestingly, you can do `bundle exec gem install stackprof` and it still hides it during `bundle exec`. But the files are all there, meaning it is theoretically possible to allow access to it.

[...]

PR #4680, opened 2021-06-17
`bundle exec --allow-gems=GEMS` to enable GEMS regardless of Gemfile

mame:

I want to enable some tooling gems temporarily even if they are not written in Gemfile. See #4649 in detail.

This changeset adds `--allow-gems=GEMS` option to `bundle exec`.
The option makes the speficied gems available even if the gems are not
listed in Gemfile. This is supposed to be used for some tooling gems
(like stackprof) which is needed just temporarily.

If the gem specified by `--allow-gems` is not listed in Gemfile, the
installed latest version of the gem is picked up. If the gem is listed
in Gemfile, the option is just ignored and Gemfile's specification is
respected.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -469,6 +469,7 @@ module Bundler
 
     desc "exec [OPTIONS]", "Run the command in context of the bundle"
     method_option :keep_file_descriptors, :type => :boolean, :default => false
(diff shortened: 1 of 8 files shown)


mame:
@deivid-rodriguez proposed `ruby -rstackprof -rbundler/setup` which is much simpler than this PR, so I close this. Thanks.

PR #4681, opened 2021-06-17
Close then unlink tempfiles on Windows

nobu:

In ruby/ruby test actions, number of "leaked tempfile" messages are shown on Windows.

As Windows disallows removing open files, `Tempfile#unlink` fails silently before `#close`.
Close then unlink by `#close!` instead.


deivid-rodriguez (approved):
Thank you!

This is copied from minitest, so the PR should be proposed there too in order to  be good citizens :)

PR #4683, opened 2021-06-17
Fix multisource lockfile warning message

deivid-rodriguez:

This is a follow up to https://github.com/rubygems/rubygems/pull/4647.

In that PR, we fixed our logic to automatically update the lockfile format to have split GEM sections when necessary.

So, there's only two cases when this warning is printed now:

* When the Gemfile has multiple global sources. In this case, the warning is not really necessary because we already display a deprecation warning about the `Gemfile`, and once that's fixed, the lockfile format will be automatically upgraded.

* When we are in frozen mode, because we're not allowed to write to the lockfile.

This PR changes bundler so that the warning is not printed in the first case, and so that the warning gives proper resolution steps for the second case.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -112,7 +112,7 @@ module Bundler
       end
 
       @locked_gem_sources = @locked_sources.select {|s| s.is_a?(Source::Rubygems) }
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I added a couple of extra commits to try to improve the naming and make the logic easier to understand.

PR #4684, opened 2021-06-17
Revert lib/rubygems/test_case.rb and so on

nobu:

These files are a part of the public interface, and the removal has broken tests of rdoc.

This reverts commits:

* 2904fc8a4b725bc0ffac41d0fc0a521c928365cc

  Small tweak to play nice with packaging setups

* 6460e018df3741f0a70a8366a958105504d76e10

  Rename test/rubygems/test_{case,utilities}.rb to avoid "test_" prefix

* 0efb894c3b8930c0563b407a5e280be4562e6d40

  Remove no longer needed `RUBYGEMS_TEST_PATH` env variable

* 393f269e8feb87f10ba3dd6f1507624f55bb2626

  Resolve to missing `Gem::TestCase` issue with random order tests

* 8dfe1e30b5ed4a9006c74368a41bf2872af47836

  Remove no longer relevant commentsº

* c77868a555d05ea6699a23b73b32eac1cfd22494

  Require the new files in `test/` relatively

* aa390a3500ff1fe9803ba9ec982c5ea6edd4f85f

  Copy files specific to testing rubygems to `test`

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -380,6 +380,7 @@ lib/rubygems/install_default_message.rb
 lib/rubygems/install_message.rb
 lib/rubygems/install_update_options.rb
 lib/rubygems/installer.rb
+lib/rubygems/installer_test_case.rb
 lib/rubygems/installer_uninstaller_utils.rb
 lib/rubygems/local_remote_options.rb
 lib/rubygems/mock_gem_ui.rb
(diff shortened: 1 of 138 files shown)


hsbt:
I'm working to resolve rdoc case at https://github.com/ruby/rdoc/pull/815

> These files are a part of the public interface

This is true, But we didn't support the compatibility of `Gem::TestCase`. I also agree to hide `Gem::TestCase` from LOAD_PATH after new release.

PR #4687, opened 2021-06-21
Fix test: 'rack' version from other source needs to be higher

pvalena:
to give it higher priority. It shouldn't be selected either way, as the dependency 'depends_on_rack' is in scoped source. This is related to PR #4609. 

While backporting the fix for PR 4609 for Ruby 2.7 release (bundler 2.1.4), I did not see any test actually checking for correct behaviour (no test was failing without the fix in PR).

I've raised the version to higher one, to give the dependency in unwanted (non-scoped) source higher priority, therefore properly testing the issue.

### DIFF
--- a/bundler/spec/install/gemfile/sources_spec.rb
+++ b/bundler/spec/install/gemfile/sources_spec.rb
@@ -177,8 +177,9 @@ RSpec.describe "bundle install with gems on multiple sources" do
       context "and in another source" do
         before do
           # need this to be broken to check for correct source ordering
+          # version needs to be higher to give it higher priority
           build_repo gem_repo2 do
-            build_gem "rack", "1.0.0" do |s|
+            build_gem "rack", "1.0.1" do |s|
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
There are a few new tests checking for the correct behaviour. In particular, all the tests that were previously scoped to the `disable_multisource` setting, under the "when a top-level gem has an indirect dependency" context. In https://github.com/rubygems/rubygems/commit/32ac0ba0cdad8ffe522865196d1a2ec3bbf62e08, those tests were enabled by default without the need for that setting, since they test the behaviour being fixed.

The test you are changing was indeed passing before the fix too. I didn't modify it because I thought it served to illustrate that in previous versions dependency confusion was only possible if a gem with a higher version was pushed to the outtermost source, not if a gem with the same version was pushed.

pvalena:
Thanks!

You're right, I've made a mistake when backporting that test. Retrying the backport ([patch](https://git.io/JnPmw)), it does correctly [detect](https://git.io/JnPm7) the issue.

PR #4695, opened 2021-06-24
Fixes for the edge case when openssl library is missing

rhenium:

Some test cases from RubyGems are currently failing on a Ruby installation without OpenSSL support.

Most code requiring openssl library is properly handling that. Here is an attempt to fix the leftovers:

---
**Do not use openssl constants in Gem::Specification#to_ruby**

The attribute can be either a String indicating the path of the private
key file or a raw key object (an instance of OpenSSL::PKey::RSA).
Currently, the attribute is excluded from #to_ruby in the latter case.

The current code does not interact well when the openssl library is
missing. Evaluating "defined?(OpenSSL::PKey::RSA)" will attempt to
activate the openssl gem due to autoload, and that causes weird test
failures.

Since we know what type signing_key will be otherwise, which is String,
we can simply check that instead.

---
**Skip TestGemExtRakeBuilder#test_class_no_openssl_override if necessary**

It is a regression test for a bug where RubyGems' openssl.rb slipped
into LOAD_PATH. This requires the actual openssl library to be
available.

---
**Make "gem cert --help" available when OpenSSL is missing**

The entire Gem::Commands::CertCommand class is undefined when OpenSSL
is not available. This is not expected by Gem::CommandManager and
attempting to invoke the cert command fails with an unfriendly message:

        $ gem cert --help
        ERROR:  Loading command: cert (NameError)
                uninitialized constant Gem::Commands::CertCommand
            undefined method `deprecated?' for nil:NilClass

Let's tweak the OptionParser construction to not use OpenSSL constants
until they are necessary.

### DIFF
--- a/lib/rubygems/commands/cert_command.rb
+++ b/lib/rubygems/commands/cert_command.rb
@@ -7,37 +7,9 @@ class Gem::Commands::CertCommand < Gem::Command
     super 'cert', 'Manage RubyGems certificates and signing settings',
           :add => [], :remove => [], :list => [], :build => [], :sign => []
 
-    OptionParser.accept OpenSSL::X509::Certificate do |certificate_file|
-      begin
-        certificate = OpenSSL::X509::Certificate.new File.read certificate_file
-      rescue Errno::ENOENT
-        raise OptionParser::InvalidArgument, "#{certificate_file}: does not exist"
-      rescue OpenSSL::X509::CertificateError
-        raise OptionParser::InvalidArgument,
-          "#{certificate_file}: invalid X509 certificate"
-      end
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This is awesome @rhenium, thanks!

Currently we only have few realworld checks at `.github/workflows/install-rubygems.yml` without openssl, but if I recall correctly, the only reason for that was that I couldn't make the full test suite pass without openssl before I considered I was spending too much time on it.

I think at this point we might want to replace those checks by a full "no openssl" test run?

deivid-rodriguez (approved):
The comment I left is definitely not a blocker to get this in.

PR #4697, opened 2021-06-25
Allow setting `--otp` via `GEM_HOST_OTP_CODE`

CGA1123:
## What was the end-user or developer problem that led to this PR?

For release automation it can be nice to want to pass in the OTP via
some UI (e.g. a workflow_dispatch input on GitHub Actions) and then just
be able to run `rake release` to publish to rubygems.

## What is your fix for the problem, implemented in this PR?

This introduces a new environment variable `GEM_HOST_OTP_CODE` which
will take precendence over `--otp` command line options, allowing for
automated releases with OTP.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4704, opened 2021-06-26
Update api_key configuration precedence

CGA1123:

During review of #4697 it was realised that the precedence order for the `api_key` option was not as expected.
CLI option inputted by a user should take precedence over environment variables.

Updates the precedence order for where to look for an API key to be the
following:
- The api_key as passed in via the `--api-key` CLI option
- The api_key as configured for the current host in `.gems/credentials`
- The api_key as configured by GEM_HOST_API_KEY env var
- The api_key as configured in `.gems/credentials` for rubygems.org

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -40,12 +40,12 @@ module Gem::GemcutterUtilities
   # The API key from the command options or from the user's configuration.
 
   def api_key
-    if ENV["GEM_HOST_API_KEY"]
-      ENV["GEM_HOST_API_KEY"]
-    elsif options[:key]
+    if options[:key]
       verify_api_key options[:key]
     elsif Gem.configuration.api_keys.key?(host)
       Gem.configuration.api_keys[host]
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@CGA1123 Thanks! :)

Can you add a regression test to cover the fix?

deivid-rodriguez (commented):
@CGA1123 Thanks!

I was reviewing the PR and I noticed that the `--key` option to `gem push` and the `GEM_HOST_API_KEY` env var doesn't actually have the same meaning. While the `--key` option should hold a key _name_ to be picked up from your `credentials` file, the `GEM_HOST_API_KEY` env variable should hold an _actual API key_.

With this in mind, I'm not sure there's an actual bug in the priority here. It's a bit unclear what the user expectation should be in this situation, since while CLI flags should take precedence over ENV, the ENV configuration here is actually more specific than the CLI flag.

What do you think?

CGA1123:
I see, yes it is more specific and allows you to overwrite the key used ignoring the matched host 👍

PR #4712, opened 2021-07-01
Limit MacOS jobs

deivid-rodriguez:

After merging #4245, we got a CI failure with this log:

> 2021-07-01T10:08:49.5189122Z The agent pool assigned to this job has hit their MacOs concurrency limits
> 2021-07-01T10:08:49.5532238Z Found online and idle hosted runner(s) in the current repository's organization account that matches the required labels: 'macos-10.15'
> 2021-07-01T10:08:49.5532347Z Waiting for a hosted runner in 'organization' to pick this job...

The job ended up waiting too much until it timed out, apparently. 

My fix is to run less jobs on MacOS.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -17,7 +17,6 @@ jobs:
       matrix:
         os:
           - { name: Ubuntu, value: ubuntu-20.04 }
-          - { name: MacOS, value: macos-10.15 }
 
         ruby:
           - { name: ruby-2.3, value: 2.3.8 }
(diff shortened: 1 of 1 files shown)


MSP-Greg:
@deivid-rodriguez

I believe the problem is intermittent, and not due to anything a repo can control.  I work with one repo that has 28 macOS jobs, and I also run the same CI in my fork.  It passes most of the time.

I have seen the problem in other repos, a good reason to add timeouts to all workflow jobs...

deivid-rodriguez:
Thanks for the info @MSP-Greg!

I imagine the problem is not fully under our control, but I guess the less jobs we run on MacOS the less likely it is too happen. I prefer to start off with less jobs and see how it goes. Just the main rubies on bundler 2 mode seems more than enough for now. Our CI is very heavy already.

MSP-Greg:
@deivid-rodriguez

I'll leave the question of how many jobs to run up to you.  People get frustrated when CI takes too long, but then they also complain when a particular OS/Ruby combination isn't tested and has a problem.  Ran into one recently where all the 'latest' OS jobs ran, but someone raised an issue, and CI failed on Ruby 3.0+ on Ubuntu-18.04.  IOW, you can't win...

> the less jobs we run on MacOS the less likely it is too happen

I think the problem isn't really related to that, but more related to when the CI is run, as some days/times all CI seems to run  slower.  Anyway, thanks...

deivid-rodriguez:
Yeah, I understand it's a temporary issue, but since even when it's happening only some jobs seem to run into it, sometimes none, I guess the less jobs the less chance we get it? Anyways, the current combination seems ok to me as I said, so let's see how it goes.

Issue #3159, opened 2020-02-27
Improving code documentation


duckinator:
The RubyGems codebase lacks documentation in some areas. Let's try to fix that. :slightly_smiling_face: 

This issue is focusing on documenting classes, modules, and methods to help current contributors and make it easier for new contributors.

For discussing documenting how to use RubyGems itself and things of that sort, please open separate issues!

This is also intended as a kind of "meta issue", used mostly for coordination. If you want to discuss documenting a specific part of the codebase at-length, you can [join us on Slack](https://slack.bundler.io/)  or open a new issue about it and @-mention me in the issue.

**If you intend to work on documenting a particular file, leave a comment below saying so and I'll put your GitHub handle by the relevant file.** This helps avoid duplicated effort.

You can see [`lib/rubygems.rb`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems.rb) and [`lib/rubygems/util.rb`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/util.rb) for some good examples of how to document a class or method.

## Areas To Document

Here is the list of areas that need documentation improvements:

- [x] `lib/rubygems/deprecate.rb` (@duckinator, #4717)
  - [x] Update outdated information: It says it provides a single method, `deprecate`, but it actually provides `deprecated` + `deprecate_command`.
- [ ] `lib/rubygems/ext/builder.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/ext/cmake_builder.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/ext/configure_builder.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/ext/ext_conf_builder.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/ext/rake_builder.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/request/connection_pools.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
- [ ] `lib/rubygems/request_set/lockfile.rb`
  - [ ] Class/module comment explaining common usage.
  - [ ] Method/function documentation.
[...]

PR #4717, opened 2021-07-04
[internal docs] Improve documentation of lib/rubygems/deprecate.rb

duckinator:
This is part of #3159.

The documentation for `Gem::Deprecate` was extremely outdated.

I updated the documentation.

- [x] <del>Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes</del> (Nothing to do.)
- [x] Write <del>code</del> <ins>documentation</ins> to solve the problem

### DIFF
--- a/lib/rubygems/deprecate.rb
+++ b/lib/rubygems/deprecate.rb
@@ -1,23 +1,70 @@
 # frozen_string_literal: true
 ##
-# Provides a single method +deprecate+ to be used to declare when
-# something is going away.
+# Provides 3 methods for declaring when something is going away.
+#
+# +deprecate(name, repl, year, month)+:
+#     Indicate something may be removed on/after a certain date.
+#
+# +rubygems_deprecate(name, replacement=:none)+:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (commented):
Looks good!

Should we also add a note about the visibility of the module, and make it clear that the whole module is not intended for public usage?

deivid-rodriguez (approved):
My suggestion can be added separately, this PR just brings documentation inline with the current code, so makes sense to merge as is I think!

deivid-rodriguez:
One small thing. The `rubygems: documentation` label is used for changelog purposes, so that the pull request is included in the changelog of the next version under the "Documentation" section. As a general rule, we only include in the changelog PRs that affect end users, not PRs targeted to rubygems developers. In this case, this is just internal documentation that I don't think affects end users, so I think the label doesn't apply.

duckinator:
I was unaware labels were used in changelog generation. Good to know!

PR #4721, opened 2021-07-05
Remove `Gem::UriParsing` module and move logic to `Gem::UriParser` class

### LINKED ISSUE #4719: [code quality] Should Gem::UriParser be merged into Gem::UriParsing?
I was looking into improving documentation, as part of #3159, but ran into this, and I'd like input from other maintainers on how to approach it.

RubyGems has a pair of closely related files: [`lib/rubygems/uri_parsing`](https://github.com/rubygems/rubygems/blob/e17127b11e23b7f6b6613c92cff29fa0fba69618/lib/rubygems/uri_parsing.rb) (`Gem::UriParsing`) and [`lib/rubygems/uri_parser.rb`](https://github.com/rubygems/rubygems/blob/e17127b11e23b7f6b6613c92cff29fa0fba69618/lib/rubygems/uri_parser.rb) (`Gem::UriParser`).

The thing I realized is, basically: `Gem::UriParser` is only used by `Gem::UriParsing`, and `Gem::UriParsing` is only used by `Gem::RemoteFetcher`. It is also used exactly 4 times.

This means, in practice, we have two files totaling 72 lines, which could potentially be replaced with this:

```ruby
##
# Parses the #uri, raising if it's invalid.

def parse!(uri)
    require 'uri'
    raise Uri::InvalidURIError unless uri
    
    # Always escape URI's to deal with potential spaces and such
    # It should also be considered that source_uri may already be
    # a valid URI with escaped characters. e.g. "{DESede}" is encoded
    # as "%7BDESede%7D". If this is escaped again the percentage
    # symbols will be escaped.
    begin
      URI.parse(uri)
    rescue URI::InvalidURIError
      URI.parse(URI::DEFAULT_PARSER.escape(uri))
    end
end

##
# Parses the uri, returning the original uri if it's invalid.

def parse_uri(uri)
  return uri unless source_uri.is_a?(String)

  parse!(uri)
rescue URI::InvalidUriError
  uri
end
```

All of this basically leaves me with 3 questions:

1. Should this be consolidated into one file?
2. If so, should we completely replace the existing API, and have e.g. `Gem::UriParser.parse(uri)`? (The lengthiness of it seems less unreasonable to me, with the context that it's only used 4 times.)
2. If we should not completely replace the existing API, should it be changed to use `Gem::UriParser.parse(...)` directly, or should we stick to the `include Gem::UriParsing; parse_uri(...)` approach?

deivid-rodriguez:

These had the same purpose, so it's confusing to keep that duplication.

Move `Gem::UriParsing` logic to the `Gem::UriParser` class.

Closes #4719.


hsbt:
👍  I confirmed this is not used by other libraries.

```
hsbt@aluminium:~$ gem-codesearch "require \'rubygems\/uri_parsing\'"
/srv/gems/rubygems-update-3.2.21/lib/rubygems/remote_fetcher.rb:require 'rubygems/uri_parsing'
hsbt@aluminium:~$ gem-codesearch 'require \"rubygems\/uri_parsing\"'
hsbt@aluminium:~$
```

deivid-rodriguez:
Thanks! Yeah, I introduced it very recently, not sure what I was thinking... :sweat_smile:.

KellyHang (commented):
****

KellyHang (commented):
****

KellyHang (commented):
****

KellyHang (commented):
****

KellyHang (commented):
****

PR #4724, opened 2021-07-06
Spare meaningless warning on read-only bundle invocations

### LINKED ISSUE #3358: "`/` is not writable" message emitted on read-only bundle invocations
If bundler is run in a production environment (i.e., on an immutable filesystem and with `$HOME` unset), it helpfully prints out the following warning on every invocation (even those like `bundle exec` which should never write to the filesystem)

```
`/` is not writable.
Bundler will use `/tmp/bundler/home/username' as your home directory temporarily.
```

Setting `$HOME` to point to a writable temporary directory suppresses the message. I verified (using `strace`) that nothing is ever opened in that temporary directory.

I have two issues with this log message:

   1. Warning that `/` isn't writable on an action that will never write seems extraneous.
   2. Errors should be printed to stderr instead of stdout. There is never an acceptable care to print errors to stdout. Yes, I'm aware that I can export `$BUNDLE_ERROR_ON_STDERR` but does that really need to be opt-in for relatively-new warnings? Nobody could possibly have been relying on the broken behavior of writing errors to stdout for a warning that didn't even exist two minor versions ago.

This looks closely related to rubygems/bundler#5371 (which was closed but only for the case where $HOME is unset) and to have been proximally caused by rubygems/bundler#5385.

It would be swell if bundler (a) would not attempt to write when running `bundle exec`, and (b) would write this warning to stderr if it really feels the need to emit it.

```
% bundle version
Bundler version 1.16.3 (2018-07-17 commit 12e73cab7)
% bundle exec ruby --version
ruby 2.3.7p456 (2018-03-28 revision 63024) [x86_64-linux-gnu]
```

Let me know if you'd prefer that I re-open rubygems/bundler#5371 in favor of opening a new issue

deivid-rodriguez:

Sometimes bundler complains about home not being writable without any need, because bundler doesn't really need to write to the filesystem.

The problem is that resolving the user's home to figure out where the global settings file lives prints this warning, and bundler checks setting essentially every time it is run.

My fix is to ignore global settings when user's home is not set, and only raise when we need to write configuration but not in other cases. Using a "temporary home fallback" for settings doesn't make sense anyways, because settings are supposed to be permanent, but this temporary home is generated at a random location at `tmp` so whatever we would write there would not be accessible the next time bundler is run because we don't know the exact location.

Fixes #3358.

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -428,12 +428,8 @@ module Bundler
     def global_config_file
       if ENV["BUNDLE_CONFIG"] && !ENV["BUNDLE_CONFIG"].empty?
         Pathname.new(ENV["BUNDLE_CONFIG"])
-      else
-        begin
-          Bundler.user_bundle_path("config")
-        rescue PermissionError, GenericSystemCallError
-          nil
-        end
+      elsif Bundler.rubygems.user_home && !Bundler.rubygems.user_home.empty?
+        Pathname.new(Bundler.rubygems.user_home).join(".bundle/config")
       end
     end
 
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -236,8 +236,9 @@ module Bundler
         end
 
         if warning
-          user_home = tmp_home_path(warning)
-          Bundler.ui.warn "#{warning}\nBundler will use `#{user_home}' as your home directory temporarily.\n"
+          Bundler.ui.warn "#{warning}\n"
+          user_home = tmp_home_path
+          Bundler.ui.warn "Bundler will use `#{user_home}' as your home directory temporarily.\n"
           user_home
         else
           Pathname.new(home)
@@ -684,15 +685,13 @@ EOF
       Bundler.rubygems.clear_paths
     end
 
-    def tmp_home_path(warning)
+    def tmp_home_path
       Kernel.send(:require, "tmpdir")
       SharedHelpers.filesystem_access(Dir.tmpdir) do
         path = Bundler.tmp
         at_exit { Bundler.rm_rf(path) }
         path
       end
-    rescue RuntimeError => e
-      raise e.exception("#{warning}\nBundler also failed to create a temporary home directory':\n#{e}")
     end
 
     # @param env [Hash]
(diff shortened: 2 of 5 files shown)


deviant:
I've applied this locally and it's working great. Thanks!

bundler/lib/bundler/settings.rb
```diff
@@ -428,12 +428,8 @@ def printable_value(value, key)
     def global_config_file
       if ENV["BUNDLE_CONFIG"] && !ENV["BUNDLE_CONFIG"].empty?
         Pathname.new(ENV["BUNDLE_CONFIG"])
-      else
-        begin
-          Bundler.user_bundle_path("config")
-        rescue PermissionError, GenericSystemCallError
-          nil
-        end
+      elsif Bundler.rubygems.user_home && !Bundler.rubygems.user_home.empty?
```
ChrisBr, line 431: @deivid-rodriguez Hey 👋  It seems after this change these env variables are not picked up anymore.

https://github.com/rubygems/rubygems/blob/6a5a1a4af23526e24f21b558517bb8948a6e4ff5/bundler/lib/bundler.rb#L251-L259

Is this expected?

deivid-rodriguez: Mmmm... right :grimacing:.

I'm guessing you're setting `BUNDLE_USER_CONFIG` to something like `~/.config/bundler` to save the user's bundler configuration to something XDG-compliant, correct?

I broke that, yeah... :disappointed:. I checked these environment variables but I thought they were mostly related caches and missed they also served the change configuration's root (specially since we also have `BUNDLE_CONFIG` to change the configuration location...).

Sorry about that, I'll look into restoring this without reintroducing the original issue.

ChrisBr: > I'm guessing you're setting BUNDLE_USER_CONFIG to something like ~/.config/bundler to save the user's

That's right. We're using this in CI to setup a bundler config and use `BUNDLE_USER_CONFIG` to point to the config. 

Thanks a lot for looking into this 🙏

deivid-rodriguez: No problem! In the meantime you can use `BUNDLE_CONFIG` as a workaround. So if you were using `BUNDLE_USER_CONFIG=~/.config/bundler`, you can change it to `BUNDLE_CONFIG=~/.config/bundler/config`.

tomstuart: This bit me too. Is there an issue for it yet (I couldn’t find one)? If not, should I create one?

deivid-rodriguez: Yes, please, because I forgot to create one myself and I had forgotten about it.

tomstuart: Okay, thanks, I’ve opened #4793.

PR #4729, opened 2021-07-07
Rewind IO source to allow working with contents in memory

### LINKED ISSUE #4664: Unable to Extract Gem from StringIO
I've got some code that looks like:

```ruby
Gem::Package.new(StringIO.new("contents of gem file")).extract_files(gem_dir)
```

Unfortunately, this doesn't work since the IO object is read to the end in this call to [`Gem::Package#verify`](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package.rb#L374). In the resulting call to [`Gem::Package::TarReader#each`](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package.rb#L381), the reader is tripped up [here](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package/tar_reader.rb#L58).

I've come up with a few solutions, but didn't know if y'all had a preferred way of addressing this:

1. In [`Gem::Package::IOSource#with_read_io`](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package/io_source.rb#L33-L35), rewind after:
  ```ruby
  def with_read_io
    yield io
  ensure
    io.rewind
  end
```

2. In [`Gem::Package::TarReader#each`](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package/tar_reader.rb#L58), rewind at the beginning:
  ```ruby
  def each
    ...

    @io.rewind

    until @io.eof? do
      ...
  ```

3. In `Gem::Package`, rewind anytime reading happens.

Option 1 is my favorite since `Gem::Package::IOSource` already expects to start at the [beginning of the stream](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package/io_source.rb#L19-L21) and matches the behavior of [`Gem::Package::FileSource`](https://github.com/rubygems/rubygems/blob/v3.2.19/lib/rubygems/package/file_source.rb#L28-L30) (since a new IO object is created each time).

### PR DESCRIPTION
**Pull request by drcapulet**
This PR fixes the bug I reported in #4664.

Fixes #4664.

### DIFF
--- a/lib/rubygems/package/io_source.rb
+++ b/lib/rubygems/package/io_source.rb
@@ -32,10 +32,14 @@ class Gem::Package::IOSource < Gem::Package::Source # :nodoc: all
 
   def with_read_io
     yield io
+  ensure
+    io.rewind
   end
 
   def with_write_io
     yield io
+  ensure
+    io.rewind
   end
 
   def path
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -1145,6 +1145,13 @@ class TestGemPackage < Gem::Package::TarTestCase
     end
   end
 
+  def test_contents_from_io
+    io = StringIO.new Gem.read_binary @gem
+    package = Gem::Package.new io
+
+    assert_equal %w[lib/code.rb], package.contents
+  end
+
   def util_tar
     tar_io = StringIO.new
 

### CODE REVIEW

File: lib/rubygems/package/io_source.rb [hunk 1]
```diff
@@ -32,10 +32,14 @@ def present?
...
   def with_read_io
     yield io
+  ensure
+    io.rewind
   end
 
   def with_write_io
     yield io
+  ensure
+    io.rewind
```
deivid-rodriguez, line 42: As far I understand, rewinding after write is not related to this PR, and you just added it for consistency? Seems fine, just wanted to confirm.

deivid-rodriguez (approved):
Looks good to me, comment was just out of curiosity.

Thanks!

PR #4732, opened 2021-07-07
Fix incorrect warning about duplicated gems in the Gemfile

### LINKED ISSUE #4481: Spurious warning: Your Gemfile lists the gem my-private-gem (>= 0) more than once

If you have a gem that only exists in a non-default source in your gemspec listed as a development dependency, it does not appear to be possible to add it to your `Gemfile` without this warning appearing:

```
Your Gemfile lists the gem my-private-gem (>= 0) more than once.
You should probably keep only one of them.
Remove any duplicate entries and specify the gem only once.
While it's not a problem now, it could cause errors if you change the version of one of them later.
```

Clone and run bundle install:

https://github.com/aaronjensen/private-development-gem-warning-repro

`bundle install`

No warning

### What actually happened?

Warning

Bundler 2.2.15

deivid-rodriguez:

An incorrect warning when a gem is a development dependency of a local gemspec, and also specified in the Gemfile with compatible requirements.

# What is your fix for the problem, implemented in this PR?

Skip the warning in this case.

Fixes #4481.

### DIFF
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -102,28 +102,26 @@ module Bundler
       # if there's already a dependency with this name we try to prefer one
       if current = @dependencies.find {|d| d.name == dep.name }
         deleted_dep = @dependencies.delete(current) if current.type == :development
+        return if deleted_dep
 
         if current.requirement != dep.requirement
-          unless deleted_dep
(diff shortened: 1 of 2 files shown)


tf:
Do we know when this will be released? This bug is causing a lot of CI failures for me right now.

PR #4738, opened 2021-07-08
Bump vendored connection_pool to 2.2.5

deivid-rodriguez:

Nothing but I think we should try to benefit from upstream's work.

Bump vendored version to 2.2.5. I also included a copy of patches that we were using directly and proposed them upstream:
* https://github.com/mperham/connection_pool/pull/152
* https://github.com/mperham/connection_pool/pull/153.


deivid-rodriguez:
Both upstream patches have been merged, so now we're in sync with current `connection_pool`'s master branch.

deivid-rodriguez:
MacOS is very flaky. I think it's Github Actions environments fault.

I don't want to lose the coverage, but I would further reduce the matrix to only run against the oldest and the newest ruby supported. The cases were ruby version specific issues are found are very few, and when it happens it's always because of using newer ruby features not present in older versions. I don't recall any other cases at the moment. So I believe on Windows and MacOS maybe oldest+newest is fine.

deivid-rodriguez:
@simi Let me know if the proposal seems ok to you. I think it should further reduce flakyness, and also make CI feedback faster.

deivid-rodriguez:
Proposed at https://github.com/rubygems/rubygems/pull/4741.

simi:
@deivid-rodriguez makes sense. :+1:

PR #4739, opened 2021-07-08
Fix contradictory message about deletion of default gem

### LINKED ISSUE #4733: Contradictory output when trying to uninstall a default gem
This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

```
$ gem env version
3.2.22
```

# Observed behavior

```
gem uninstall bundler
Gem bundler-2.2.22 cannot be uninstalled because it is a default gem
...
Successfully uninstalled bundler-2.2.22
```

These two statements are contradictory.

Also, it is unclear what state the system has been left in.

# Expected behavior

`gem` should not print contradictory statements.

# Full output

```
gem uninstall bundler
Gem bundler-2.2.22 cannot be uninstalled because it is a default gem

You have requested to uninstall the gem:
	bundler-2.2.22

bundler-audit-0.8.0 depends on bundler (>= 1.2.0, < 3)
libyear-bundler-0.5.3 depends on bundler (>= 1.14, < 3)
rails-6.0.4 depends on bundler (>= 1.3.0)
If you remove this gem, these dependencies will not be met.
Continue with Uninstall? [yN]  y
Remove executables:
	bundle, bundler

in addition to the gem? [Yn]  Y
Removing bundle
Removing bundler
Successfully uninstalled bundler-2.2.22
gem uninstall bundler
Gem bundler-2.2.22 cannot be uninstalled because it is a default gem

### PR DESCRIPTION
**Pull request by jaredbeck**
[Fixes #4733]

@deivid-rodriguez I'm happy to write another test, or change the copy-writing. I'm just opening a PR to get initial feedback and make sure I'm on the right track. Thanks.

----

#4733

As discussed in #4733

### DIFF
--- a/lib/rubygems/uninstaller.rb
+++ b/lib/rubygems/uninstaller.rb
@@ -70,6 +70,9 @@ class Gem::Uninstaller
     # only add user directory if install_dir is not set
     @user_install = false
     @user_install = options[:user_install] unless options[:install_dir]
+
+    # Optimization: populated during #uninstall
+    @default_specs_matching_uninstall_params = []
   end
 
   ##
@@ -98,10 +101,8 @@ class Gem::Uninstaller
     default_specs, list = list.partition do |spec|
       spec.default_gem?
     end
-
-    default_specs.each do |default_spec|
-      say "Gem #{default_spec.full_name} cannot be uninstalled because it is a default gem"
-    end
+    warn_cannot_uninstall_default_gems(default_specs - list)
+    @default_specs_matching_uninstall_params = default_specs
 
     list, other_repo_specs = list.partition do |spec|
       @gem_home == spec.base_dir or
@@ -273,7 +274,7 @@ class Gem::Uninstaller
     end
 
     safe_delete { FileUtils.rm_r gemspec }
-    say "Successfully uninstalled #{spec.full_name}"
+    announce_deletion_of(spec)
 
     Gem::Specification.reset
   end
@@ -375,5 +376,35 @@ class Gem::Uninstaller
     e.spec = @spec
 
     raise e
+  end
+
+  private
+
+  def announce_deletion_of(spec)
+    name = spec.full_name
+    say "Successfully uninstalled #{name}"
+    if default_spec_matches?(spec)
+      say(
+        "There was both a regular copy and a default copy of #{name}. The " \
+          "regular copy was successfully uninstalled, but the default copy " \
+          "was left around because default gems can't be removed."
+      )
+    end
+  end
+
+  # @return true if the specs of any default gems are `==` to the given `spec`.
+  def default_spec_matches?(spec)
+    !default_specs_that_match(spec).empty?
+  end
+
+  # @return [Array] specs of default gems that are `==` to the given `spec`.
+  def default_specs_that_match(spec)
+    @default_specs_matching_uninstall_params.select {|default_spec| spec == default_spec }
+  end
+
+  def warn_cannot_uninstall_default_gems(specs)
+    specs.each do |spec|
+      say "Gem #{spec.full_name} cannot be uninstalled because it is a default gem"
+    end
   end
 end
--- a/test/rubygems/test_gem_uninstaller.rb
+++ b/test/rubygems/test_gem_uninstaller.rb
@@ -295,8 +295,15 @@ class TestGemUninstaller < Gem::InstallerTestCase
 
     uninstaller = Gem::Uninstaller.new spec.name, :executables => true
 
-    uninstaller.uninstall
-
+    ui = Gem::MockGemUi.new "1\ny\n"
+    use_ui ui do
+      uninstaller.uninstall
+    end
+    expected = "Successfully uninstalled default-2\n" \
+      "There was both a regular copy and a default copy of default-2. The " \
+      "regular copy was successfully uninstalled, but the default copy " \
+      "was left around because default gems can't be removed.\n"
+    assert_equal expected, ui.output
     assert_path_not_exist spec.gem_dir
   end
 

### CODE REVIEW

**Comment by jaredbeck** [comment]
Incidentally, I'm not able to run the full test suite locally. I get a bunch of errors like:

```
Errno::EPERM: Operation not permitted @ dir_s_mkdir - /Users/jaredbeck/code/rubygems/tmp/test_rubygems_20210708-4227-1kpzi0k
```

.. and then my corporate anti-malware freaks out.

I'll try running the test suite on a personal laptop and see how that goes.

File: lib/rubygems/uninstaller.rb [hunk 1]
```diff
@@ -100,7 +100,11 @@ def uninstall
     end
 
     default_specs.each do |default_spec|
-      say "Gem #{default_spec.full_name} cannot be uninstalled because it is a default gem"
+      say(
+        "#{default_spec.full_name} is a default gem. If there is a user " \
+          "copy, it will be uninstalled, but the default copy cannot be " \
+          "uninstalled."
+      )
```
deivid-rodriguez, line 107: Instead of adding conditionals to the message, since we can figure out whether there is a user copy or not, I would leave the message intact if there's no user copy, and don't print anything otherwise (because we'll give a message after we're doing with uninstall clarifying that case). So something like (completely untested):

```suggestion
       say "Gem #{default_spec.full_name} cannot be uninstalled because it is a default gem" unless list.include? {|s| s.full_name == default_spec.full_name }
```

lib/rubygems/uninstaller.rb
```diff
@@ -376,4 +380,19 @@ def safe_delete(&block)
...
+    name = spec.full_name
+    say "Successfully uninstalled #{name}"
+    if spec.default_gem?
+      say(
+        "#{name} is a default gem. A user copy has been deleted, but the " \
+          "default copy cannot be deleted. Please see " \
+          "https://github.com/rubygems/rubygems/issues/2112 which explains " \
+          "the difference between (and precedence of) user and default copies."
+      )
+    end
```
deivid-rodriguez, line 396: I think this is incorrect, because we partition the list of initial candidates between default specs and regular specs here:

https://github.com/rubygems/rubygems/blob/6a5a1a4af23526e24f21b558517bb8948a6e4ff5/lib/rubygems/uninstaller.rb#L98-L101

And we don't even try to uninstall default gems, so `spec.default_gem?` will always be `false`?

Regarding the message, I would keep the message as is for the standard case, and change it to something simpler like

"There was a regular copy and a default copy of #{name}. The regular copy was successfully uninstalled, but the default copy was left around because default gems can't be removed."

deivid-rodriguez (commented):
@jaredbeck Thanks for working on this, I left some feedback!

Regarding your problems running tests, that's weird. Can you try removing the local `tmp` folder, run `rake setup` and try again?

lib/rubygems/uninstaller.rb
jaredbeck: Good idea

lib/rubygems/uninstaller.rb
jaredbeck: Hopefully resolved by https://github.com/rubygems/rubygems/pull/4739/commits/ff8c129a9f0ceca27ad279f9b30ba4b8631124b5

jaredbeck:
Please let me know if this looks good and I will try to add a test for the new message. Thanks!

lib/rubygems/uninstaller.rb
```diff
@@ -376,4 +378,32 @@ def safe_delete(&block)
...
+    !default_specs_that_match(spec).empty?
+  end
+
+  # @return [Array] specs of default gems that are `==` to the given `spec`.
+  def default_specs_that_match(spec)
+    default_specs = []
+    Gem::Specification.each_spec([Gem.default_specifications_dir]) do |s|
+      default_specs << s
+    end
+    default_specs.select {|default_spec| spec == default_spec }
```
deivid-rodriguez, line 407: Maybe we could save default specs to an instance variable when they are first listed at the beginning, and use it here so that we don't need to look for them every time?

deivid-rodriguez (commented):
Just a minor comment, but looks good yeah! You can go ahead with the test :+1:.

deivid-rodriguez (approved):
Looks great to me! :heart:

jaredbeck:
Thanks, David. What's the next step? Do we need two reviewers?

deivid-rodriguez:
@jaredbeck Not necessary, I usually leave PRs for a couple days to give some time to other maintainers to have a look, but I merge after that otherwise!

test/rubygems/test_gem_uninstaller.rb
```diff
@@ -295,8 +295,15 @@ def test_uninstall_default_gem_with_same_version
...
     uninstaller = Gem::Uninstaller.new spec.name, :executables => true
 
-    uninstaller.uninstall
-
+    ui = Gem::MockGemUi.new "1\ny\n"
+    use_ui ui do
+      uninstaller.uninstall
+    end
+    expected = "Successfully uninstalled default-2\n" \
+      "There was both a regular copy and a default copy of default-2. The " \
```
simi, line 303: Do we use "copy" term at all in RubyGems?

deivid-rodriguez: "Installation" maybe? This is a very edge case situation, so the term is probably not used anywhere else, indeed.

deivid-rodriguez: What do you think @simi? Should we use "installation" or another alternative, or is "copy" ok even if it's not used anywhere else?

deivid-rodriguez: Alright, let's merge this and we can further tweak things as followup PR's.

deivid-rodriguez:
Thanks @jaredbeck!

jaredbeck:
Thanks for the quick reviews, everyone.

PR #4749, opened 2021-07-09
Make sure `bundle check` shows uniq missing gems

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

`bundle check` displaying duplicated missing gems.

## What is your fix for the problem, implemented in this PR?

Make sure `SpecSet#for` returns uniq gems.

Fixes #4746.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4750, opened 2021-07-11
Slightly speed up `bundler/setup`

deivid-rodriguez:

We should be making `bundler/setup` as fast as we can.

Refactor some stuff to reduce the work we need to do setup the bundle.

It fixes a small performance regression introduced by #4749 (still unreleased), but the final result should also be ~5-10% faster than the latest release.

On a brand new rails application, I get this:

```
$ hyperfine --warmup 1 'ruby -rbundler/setup -e1' 'BUNDLER_VERSION=2.2.23 ruby -rbundler/setup -e1' 
Benchmark #1: ruby -rbundler/setup -e1
  Time (mean ± σ):     205.1 ms ±   1.8 ms    [User: 185.7 ms, System: 21.6 ms]
  Range (min … max):   202.7 ms … 209.1 ms    14 runs
 
Benchmark #2: BUNDLER_VERSION=2.2.23 ruby -rbundler/setup -e1
  Time (mean ± σ):     219.4 ms ±   1.5 ms    [User: 196.3 ms, System: 25.4 ms]
  Range (min … max):   217.5 ms … 223.3 ms    13 runs
 
Summary
  'ruby -rbundler/setup -e1' ran
    1.07 ± 0.01 times faster than 'BUNDLER_VERSION=2.2.23 ruby -rbundler/setup -e1'
```  

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -240,17 +240,11 @@ module Bundler
     end
 
     def requested_specs
-      @requested_specs ||= begin
-        groups = requested_groups
-        groups.map!(&:to_sym)
-        specs_for(groups)
-      end
+      @requested_specs ||= specs_for(requested_groups)
     end
 
     def requested_dependencies
-      groups = requested_groups
-      groups.map!(&:to_sym)
-      dependencies_for(groups)
+      dependencies_for(requested_groups)
     end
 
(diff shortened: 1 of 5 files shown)


indirect (approved):
nice! 7% is a pretty solid win for something that runs so often 🎉

PR #4755, opened 2021-07-13
fix dangling empty plugin hooks

indirect:
it turns out that running `bundle plugin uninstall some-plugin` would remove that plugin from the list of hooks, but if the list of hooks for an event was now empty, we would serialize the empty array into yaml as an empty single bullet item. which would then get unserialized as a plugin with the name empty string. which we would then try to load and explode. 😬


deivid-rodriguez (approved):
Makes sense to me!

PR #4767, opened 2021-07-15
Fix interrupt handling in Bundler workers

### LINKED ISSUE #4764: `SharedHelpers.trap` assumes the prior handler for the signal is callable

`SharedHelpers.trap` attempts to preserve any existing handler for a signal by invoking `call` on it.

https://github.com/rubygems/rubygems/blob/e7bf6b42cae52e4c6f400dd346659711dfb2c31c/bundler/lib/bundler/shared_helpers.rb#L148-L153

However, `prior` can be a string, resulting in 

```
NoMethodError: undefined method `call' for "DEFAULT":String
```

Run this script:

```ruby
# frozen_string_literal: true

require 'bundler/inline'

gemfile true do # force installation so that SharedHelpers.trap('INT') is called in Bundler::Worker#initialize
  source 'https://rubygems.org'
  gem 'rake'
end

Process.kill 'INT', Process.pid
# => bundler/shared_helpers.rb:158:in `block in trap': undefined method `call' for "DEFAULT":String (NoMethodError)
```

`bundler/inline`

I get the standard Ruby `Interrupt` exception when I interrupt the script

### What actually happened?

If Bundler installed any gems, I get `NoMethodError`. If Bundler did not install any gems (and therefore did not trap SIGINT) then I get the correct `Interrupt`.

```
Bundler       2.2.23
Ruby          2.7.3p183 (2021-04-05 revision 6847ee089d7655b2a0eea4fee3133aeacd4cc7cc) [x86_64-darwin20]
  Full Path   /Users/andrew/.rubies/ruby-2.7.3/bin/ruby
  Config Dir  /Users/andrew/.rubies/ruby-2.7.3/etc
RubyGems      3.2.23
  Gem Home    /Users/andrew/.gem/ruby/2.7.3
  Gem Path    /Users/andrew/.gem/ruby/2.7.3:/Users/andrew/.rubies/ruby-2.7.3/lib/ruby/gems/2.7.0
  User Home   /Users/andrew
  User Path   /Users/andrew/.gem/ruby/2.7.0
  Bin Dir     /Users/andrew/.gem/ruby/2.7.3/bin
Tools         
  Git         2.32.0
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

haines:

Fixes #4764.

The existing interrupt handling using `SharedHelpers.trap` fails when the previous handler for a signal is not callable (for example, when it is the string "DEFAULT").

Instead, we now handle interrupts by aborting the process when worker threads are running, and restore the previous handler after worker threads are finished.

### DIFF
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -145,13 +145,6 @@ module Bundler
       Bundler.ui.warn message
     end
 
-    def trap(signal, override = false, &block)
-      prior = Signal.trap(signal) do
-        block.call
-        prior.call unless override
-      end
-    end
-
     def ensure_same_dependencies(spec, old_deps, new_deps)
       new_deps = new_deps.reject {|d| d.type == :development }
       old_deps = old_deps.reject {|d| d.type == :development }
--- a/bundler/lib/bundler/worker.rb
+++ b/bundler/lib/bundler/worker.rb
@@ -26,7 +26,7 @@ module Bundler
       @func = func
       @size = size
       @threads = nil
-      SharedHelpers.trap("INT") { abort_threads }
+      @previous_interrupt_handler = nil
     end
 
     # Enqueue a request to be executed in the worker pool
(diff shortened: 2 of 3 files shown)


deivid-rodriguez (commented):
Looks good to me, but if you were able to add the script in #4764 as a regression spec, it'd be awesome.

haines:
@deivid-rodriguez no problem, I've added some regression specs and double-checked that they fail before applying the changes. It's a little bit tricky to test the behaviour because it results in killing the process - I've opted to fork subprocesses and check that they exit in the expected manner, hope that looks sensible to you.

indirect:
nice work!

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks @haines!!

PR #4779, opened 2021-07-23
Deprecate Gemfile without an explicit global source

### LINKED ISSUE #4759: Deprecate `Gemfile` without an explicit global source

Gemfiles like this, specially now that rubygems sources are are no longer merged into a single source for security, are very confusing because they generate a different lockfile depending on the gems you have locally installed. This is because bundler always use an implicit global source that defaults to locally installed gems.

See https://github.com/rubygems/rubygems/issues/4757 for an example.

`bundle lock`.

A consistent lockfile regardless of the gems I have installed on my system globally.

### What actually happened?

A different lockfile depending on which gems I have installed before running `bundler`.

intuxicated:

Issue: #4759 

Deprecate Gemfile without an explicit global source 

Fixes #4759.

#### Example

Gemfile:
```gemspec
source 'https://rubygems.org' do
  gem "my-gem"
end
```

Execute the following command:
```bash
$ bundle install
```

Will **output** the following warning when using Bundler 2.x:
```
[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run.Instead, define a global source in your Gemfile like this: source "https://rubygems.org".
```

For Bundler 3.x will raise an **error** with the same message.

Screenshot:

Screen Shot 2021-07-25 at 3 21 36 am

### DIFF
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -83,18 +83,11 @@ module Bundler
       end
 
       Bundler::CLI::Common.output_fund_metadata_summary
-    rescue GemNotFound, VersionConflict => e
+    rescue GemNotFound => e
       if options[:local] && Bundler.app_cache.exist?
         Bundler.ui.warn "Some gems seem to be missing from your #{Bundler.settings.app_cache_path} directory."
       end
 
-      unless Bundler.definition.has_rubygems_remotes?
-        Bundler.ui.warn <<-WARN, :wrap => true
-          Your Gemfile has no gem server sources. If you need gems that are \
-          not already on your machine, add a line like this to your Gemfile:
-          source 'https://rubygems.org'
-        WARN
-      end
       raise e
     rescue Gem::InvalidSpecificationException => e
       Bundler.ui.warn "You have one or more invalid gemspecs that need to be fixed."
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -447,6 +447,13 @@ repo_name ||= user_name
     end
 
     def check_rubygems_source_safety
+      if @sources.implicit_global_source?
+        Bundler::SharedHelpers.major_deprecation 2, "This Gemfile does not include an explicit global source. " \
+          "Not using an explicit global source may result in a different lockfile being generated depending on " \
+          "the gems you have installed locally before bundler is run." \
+          "Instead, define a global source in your Gemfile like this: source \"https://rubygems.org\"."
+      end
+
       return unless @sources.aggregate_global_source?
 
       if Bundler.feature_flag.bundler_3_mode?
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -448,14 +448,20 @@ repo_name ||= user_name
 
     def check_rubygems_source_safety
       if @sources.implicit_global_source?
-        Bundler::SharedHelpers.major_deprecation 2, "This Gemfile does not include an explicit global source. " \
-          "Not using an explicit global source may result in a different lockfile being generated depending on " \
-          "the gems you have installed locally before bundler is run." \
-          "Instead, define a global source in your Gemfile like this: source \"https://rubygems.org\"."
+        implicit_global_source_warning
+      elsif @sources.aggregate_global_source?
+        multiple_global_source_warning
       end
+    end
 
-      return unless @sources.aggregate_global_source?
+    def implicit_global_source_warning
+      Bundler::SharedHelpers.major_deprecation 2, "This Gemfile does not include an explicit global source. " \
+        "Not using an explicit global source may result in a different lockfile being generated depending on " \
+        "the gems you have installed locally before bundler is run." \
+        "Instead, define a global source in your Gemfile like this: source \"https://rubygems.org\"."
+    end
 
+    def multiple_global_source_warning
       if Bundler.feature_flag.bundler_3_mode?
         msg = "This Gemfile contains multiple primary sources. " \
           "Each source after the first must include a block to indicate which gems " \
--- a/bundler/lib/bundler/source_list.rb
+++ b/bundler/lib/bundler/source_list.rb
@@ -37,6 +37,10 @@ module Bundler
       global_rubygems_source.multiple_remotes?
     end
 
+    def implicit_global_source?
+      global_rubygems_source.no_remotes?
+    end
+
     def add_path_source(options = {})
       if options["gemspec"]
         add_source_to_list Source::Gemspec.new(options), path_sources
--- a/bundler/spec/bundler/dsl_spec.rb
+++ b/bundler/spec/bundler/dsl_spec.rb
@@ -245,7 +245,7 @@ RSpec.describe Bundler::Dsl do
   describe "#check_primary_source_safety" do
     context "when a global source is not defined implicitly" do
       it "will raise a major deprecation warning" do
-        not_a_global_source = double("not-a-global-source", no_remotes?: true, multiple_remotes?: false)
+        not_a_global_source = double("not-a-global-source", :no_remotes? => true)
         allow(Bundler::Source::Rubygems).to receive(:new).and_return(not_a_global_source)
 
         warning = "This Gemfile does not include an explicit global source. " \
--- a/bundler/spec/bundler/dsl_spec.rb
+++ b/bundler/spec/bundler/dsl_spec.rb
@@ -241,4 +241,21 @@ RSpec.describe Bundler::Dsl do
       end
     end
   end
+
+  describe "#check_primary_source_safety" do
+    context "when a global source is not defined implicitly" do
+      it "will raise a major deprecation warning" do
+        not_a_global_source = double("not-a-global-source", no_remotes?: true, multiple_remotes?: false)
+        allow(Bundler::Source::Rubygems).to receive(:new).and_return(not_a_global_source)
+
+        warning = "This Gemfile does not include an explicit global source. " \
+          "Not using an explicit global source may result in a different lockfile being generated depending on " \
+          "the gems you have installed locally before bundler is run." \
+          "Instead, define a global source in your Gemfile like this: source \"https://rubygems.org\"."
+        expect(Bundler::SharedHelpers).to receive(:major_deprecation).with(2, warning)
+
+        subject.check_primary_source_safety
+      end
+    end
+  end
 end
--- a/bundler/spec/bundler/source_list_spec.rb
+++ b/bundler/spec/bundler/source_list_spec.rb
@@ -441,4 +441,19 @@ RSpec.describe Bundler::SourceList do
       source_list.remote!
     end
   end
+
+  describe "implicit_global_source?" do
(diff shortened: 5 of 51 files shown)


bundler/lib/bundler/dsl.rb
```diff
@@ -23,6 +23,7 @@ def self.evaluate(gemfile, lockfile, unlock)
 
     def initialize
       @source               = nil
+      @global_source_defined = false
```
intuxicated, line 26: Using `@global_source_defined` as a flag to check whether a global source has been defined or not. any suggestion for a better name?

The `@sources` also has a `global_rubygems_source` and `global_path_source` which could be used instead of `@global_source_defined` to check if a global source has been added or not but I felt this would make it more complicated because `global_rubygems_source` is memoized with a new instance of `Bundler::Source::Rubygems`. Fixing this issue will require more work and I wasn't just sure about it. Let me know what you think 🤔

https://github.com/intuxicated/rubygems/blob/592dbf31a3344b24470f1c0f07a4078036f7cb76/bundler/lib/bundler/source_list.rb#L11-L13

bundler/lib/bundler/dsl.rb
```diff
@@ -45,6 +46,15 @@ def eval_gemfile(gemfile, contents = nil)
       @gemfiles << expanded_gemfile_path
       contents ||= Bundler.read_file(@gemfile.to_s)
       instance_eval(contents.dup.tap{|x| x.untaint if RUBY_VERSION < "2.7" }, gemfile.to_s, 1)
+
+      unless @global_source_defined
```
intuxicated, line 50: `instance_eval` will load the Gemfile source so in the next step we will raise a warning if no global source has been defined. 

Let me know what you think about the warning message.

bundler/lib/bundler/dsl.rb
```diff
@@ -161,6 +171,7 @@ def source(source, *args, &blk)
       elsif block_given?
         with_source(@sources.add_rubygems_source("remotes" => source), &blk)
       else
+        @global_source_defined = true
```
intuxicated, line 174: Set the global source to true since we are adding a global source here.

bundler/spec/bundler/dsl_spec.rb
```diff
@@ -95,6 +95,33 @@
       expect { subject.eval_gemfile("Gemfile") }.
         to raise_error(Bundler::GemfileError, /There was an error evaluating `Gemfile`: ruby_version must match the :engine_version for MRI/)
     end
+
+    it "raises a warning when no global source has been defined" do
```
intuxicated, line 99: Let me know what you think about tests. Are these enough? 
I've also tried to follow the same approach in other tests here, but let me know if it's better to change anything.

bundler/lib/bundler/dsl.rb
deivid-rodriguez: @intuxicated This approach seems fine too, but I think we can use a similar approach to the deprecation that warns about multiple global sources. For consistency and also because I think it will make it easier to figure out all the code that can be cleaned up once the deprecation is turned into an error.

I would define a `no_remotes?` method in `Bundler::Source::Rubygems` similar to the `multiple_remotes?` method. Like

```ruby
def no_remotes?
  @remotes.size == 0
end
```

Then I would define an `implicit_global_source?` method that checks whether the global source is defined implicitly with the above fallback (`rubygems_aggregate_class.new("allow_local" => true)`) and thus has no remotes. So, similarly to the `aggregate_global_source?` method, this one would be defined as

```ruby
def implicit_global_source?
  global_rubygems_source.no_remotes?
end
```

Finally I would add a deprecation to the `check_rubygems_source_safety` that uses the `implicit_global_source?` method in `Bundler::SourceList` with the same style as the one for multiple global sources, but that also warns this case.

Let me know if this approach makes sense.

intuxicated: @deivid-rodriguez Thanks for the feedback. I'll refactor it.

bundler/lib/bundler/dsl.rb
```diff
@@ -447,6 +447,15 @@ def check_path_source_safety
     end
 
     def check_rubygems_source_safety
+      unless @sources.implicit_global_source?
```
deivid-rodriguez, line 450: It should be `if @sources.implicit_global_source?`, but the logic of the method is also reversed, so it works like this too thanks to the double negation :sweat_smile:.

bundler/lib/bundler/dsl.rb
```diff
@@ -447,6 +447,15 @@ def check_path_source_safety
...
+        Bundler::SharedHelpers.major_deprecation(2, <<-W.strip)
+Using Gemfile without a global source is deprecated
+and it will result in an error in the future Bundler versions.
+To fix this issue, simply define a global source in your Gemfile like this:
+source 'https://rubygems.org'
```
deivid-rodriguez, line 455: Let's use the same string concatenation style used in the warnings below, so that we can keep the message nicely intented and automatically wrapped by the shell.

Regarding the wording, I'd go with

> This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: `source "https://rubygems.org"`.

Finally, you can early return after printing the deprecation/error because the case is mutually exclusive with the multiple global sources case.

bundler/lib/bundler/source_list.rb
```diff
@@ -37,6 +37,10 @@ def aggregate_global_source?
       global_rubygems_source.multiple_remotes?
     end
 
+    def implicit_global_source?
+      !global_rubygems_source.no_remotes?
```
deivid-rodriguez, line 41: As pointed out in other review comment, I think the implementation here is inversed. It should be:

```suggestion
      global_rubygems_source.no_remotes?
```

bundler/spec/bundler/source_list_spec.rb
```diff
@@ -441,4 +441,19 @@
       source_list.remote!
     end
   end
+
+  describe "implicit_global_source?" do
+    context "when a global rubygem source provided" do
+      it "returns a truthy value" do
+        source_list.add_global_rubygems_remote("https://rubygems.org")
+
+        expect(source_list.implicit_global_source?).to be_truthy
```
deivid-rodriguez, line 450: As mentioned in other review comments, the expectations here are also reversed. This case should be falsey and the one below truthy.

deivid-rodriguez (changes requested):
Added a few comments, looking good!

There's a bunch of specs failing that need to be fixed. Normally the issue is that those specs expect empty error output which is no longer empty because they're now using deprecated behaviour and thus printing a warning. It should be an easy fix in most cases by adding a global source to the relevant gemfiles, I think `source "#{file_uri_for(gem_repo1)}"` should do the trick in most cases.

bundler/lib/bundler/dsl.rb
intuxicated: I was thinking about extracting the `check_rubygems_source_safety` logic into two methods.

- `check_implicit_global_source`
- `check_multiple_global_source`

What do you think?

deivid-rodriguez: Yeah, it makes sense! :+1:

intuxicated: I ended up separating the warnings part only :D you will see.

deivid-rodriguez:
Regarding the test failures you're now getting, in one I think you need to also modify the expected lockfile to include to global source we added, and the other one serves to point out that we now need to remove the following warning message that we printed up until now when there were no global explicit sources _and_ resolution had failed:

https://github.com/rubygems/rubygems/blob/90aed1065eec5a4a266147e9e3ab48d1f0290416/bundler/lib/bundler/cli/install.rb#L91-L97

The `has_rubygems_sources` in `Definition` can also be removed since it will be now unused.

And I think you can adapt the message expecting "Your Gemfile has no gem server sources" to expect the new deprecation message instead.

bundler/lib/bundler/dsl.rb
```diff
@@ -447,8 +447,25 @@ def check_path_source_safety
     end
 
     def check_rubygems_source_safety
-      return unless @sources.aggregate_global_source?
+      if @sources.implicit_global_source?
+        implicit_global_source_warning
 
+        return
```
deivid-rodriguez, line 453: I know I did suggest to early return here, but looking at the final version, an early return usually suggests and exceptional condition, whereas in this case both warnings are "simmetrical" (of equal importance). So I think an `elsif` clause is a better style fit. So

```ruby
      if @sources.implicit_global_source?
        implicit_global_source_warning
      elsif @sources.aggregate_global_source?
        multiple_global_source_warning
      end
```

bundler/spec/runtime/setup_spec.rb
```diff
@@ -1072,6 +1077,10 @@ def clean_load_path(lp)
     end
 
     it "evals each gemspec with a binding from the top level" do
+      gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+      G
+
```
deivid-rodriguez, line 1083: I think you're already "fixing" the Gemfile in the `before` hook, so you don't need to create a separate `Gemfile` here?

deivid-rodriguez (commented):
Looks almost ready, just a few final comments. One of them requires a separate PR that I will create now, so let's hold until I merge that.

Thank you! :heart_eyes:

bundler/spec/commands/install_spec.rb
```diff
@@ -589,12 +601,12 @@
...
 
       gemfile <<-G
-        gem 'rack'
+        source "#{file_uri_for(gem_repo1)}"
+        gem 'non-existing-gem'
       G
 
       bundle :install, :quiet => true, :raise_on_error => false
-      expect(err).to include("Could not find gem 'rack'")
-      expect(err).to_not include("Your Gemfile has no gem server sources")
```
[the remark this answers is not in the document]
deivid-rodriguez: #4781 was merged implementing the above, so you can now rebase this PR, and you don't need to mess with adapting this spec at all.

bundler/lib/bundler/cli/install.rb
```diff
@@ -88,13 +88,6 @@ def run
         Bundler.ui.warn "Some gems seem to be missing from your #{Bundler.settings.app_cache_path} directory."
       end
 
-      unless Bundler.definition.has_rubygems_remotes?
-        Bundler.ui.warn <<-WARN, :wrap => true
-          Your Gemfile has no gem server sources. If you need gems that are \
-          not already on your machine, add a line like this to your Gemfile:
-          source 'https://rubygems.org'
-        WARN
-      end
```
deivid-rodriguez, line 97: Something I also noticed is that rescuing `VersionConflict` above was probably only done to show this error message, so since we are removing it, we can now limit the rescue to the `GemNotFound` class.

deivid-rodriguez (approved):
I rebased the PR myself and added the small tweaks I suggested myself, since this work has conflicts with https://github.com/rubygems/rubygems/pull/4782 which I'm working on.

Thanks so much @intuxicated!! :heart:

PR #4781, opened 2021-07-25
Fix `--quiet` flag not printing warnings

deivid-rodriguez:

The problem is that the `--quiet` flag hides warnings for no good reason, and contrary to documentation. See https://github.com/rubygems/rubygems/pull/4779#discussion_r676125872.
 
Restore the behaviour prior to https://github.com/rubygems/bundler/commit/ca0676cb1c638e0b9747ea8c18f28adf82cc01de.

### DIFF
--- a/bundler/lib/bundler/cli/cache.rb
+++ b/bundler/lib/bundler/cli/cache.rb
@@ -9,7 +9,7 @@ module Bundler
     end
 
     def run
-      Bundler.ui.level = "error" if options[:quiet]
+      Bundler.ui.level = "warn" if options[:quiet]
       Bundler.settings.set_command_option_if_given :path, options[:path]
       Bundler.settings.set_command_option_if_given :cache_path, options["cache-path"]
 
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
@indirect I wanted to double check this with you.

I think we you introduced https://github.com/rubygems/bundler/commit/ca0676cb1c638e0b9747ea8c18f28adf82cc01de there was no way to turn off warnings and deprecations, so the idea made sense at the time to satisfy the use case of the original reporter: needed a way to silence the warning about bundler being run as root.

However, since we now have `silence_deprecations` and `silence_root_warning` configurations, I think it's no longer necessary and we can go back to how `--quiet` traditionally worked and is still documented: avoid non-essential output, but print warnings and errors.

indirect:
I think that's right! I would expect `--quiet` to output warnings and errors, basically analogous to `bundle install > /dev/null`. If we need it for some reason, I would expect `--silent` to output nothing... but we could also tell people to run `bundle install 2>&1 > /dev/null` if they want that 😅.

deivid-rodriguez:
Yeah, I also thought of `--silent` as a fully silent output if we happen to need that in the future :+1:. I'll merge this now then! Thanks @indirect!

PR #4789, opened 2021-07-28
gem which command now recognize packages like actioncable 

yuriShafet:
gem which command now recognize packages like actioncable that has entry point with underscore like action_cable.rb

1. Run `gem list`
Output:
```
$ gem list

*** LOCAL GEMS ***

activesupport (6.1.4)
bigdecimal (default: 1.3.4)
cmath (default: 1.0.0)
coderay (1.1.3)
concurrent-ruby (1.1.9)
csv (default: 1.0.0)

```
2. Run `gem which activesupport`
Output:
```
$ gem which activesupport
ERROR:  Can't find Ruby library file or shared library activesupport
```

After the fix:
```
$  gem which activesupport
/usr/local/bundle/gems/activesupport-5.2.6/lib/active_support.rb
```
Same goes for all packages that has "_" in entry point file like: actioncable, actionmailer, actionpack and more.

### DIFF
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -328,11 +328,21 @@ class Gem::BasicSpecification
   private
 
   def have_extensions?; !extensions.empty?; end
-
   def have_file?(file, suffixes)
     return true if raw_require_paths.any? do |path|
-      base = File.join(gems_dir, full_name, path.tap(&Gem::UNTAINT), file).tap(&Gem::UNTAINT)
-      suffixes.any? {|suf| File.file? base + suf }
+      base_directory =  File.join(gems_dir, full_name, path.tap(&Gem::UNTAINT))
+      base = File.join(base_directory, file).tap(&Gem::UNTAINT)
+      if Dir.exist?(base_directory)
+        candidate_entry_files = Dir.entries(base_directory)
+      else
+        candidate_entry_files = []
+      end
+      res = suffixes.any? {|suf| File.file? base + suf || have_entry_file_with_underscore?(file + suf, candidate_entry_files)}
+      if res == false
+        return  have_entry_file_with_underscore?(suffixes, base_directory, file)
+      else
+        return true
+      end
     end
 
     if have_extensions?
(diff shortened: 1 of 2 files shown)


yuriShafet:
Fixes #4789 4787

deivid-rodriguez:
@yuriShafet Sorry you already worked on this before I realized `rubygems` is actualy doing the right thing here.

`gem which` expects a requirable file, and it's supposed to tell you which package that file belongs to. As per `gem which` helps, it's useful if some file you're requiring is not doing what you expect. You can then use `gem which` to figure out the full path to the file being inspected so that you can read the code and figure out the issue.

But `activesupport` is not something you can require (`require "activesupport"` fails). Perhaps we could add some "did you mean?" behaviour here but making this "just work" seems like not the way to go to me.

What do you think? What's your use case for `gem which`?

deivid-rodriguez:
The new logic you're adding seems like it would be the desired behaviour in general, but these methods are used not only by `gem which` but also other places, and I'm not sure if the new logic makes sense in those other places. So if we want to go this way, I would at least move the logic to the `gem which` command itself and run it as a fallback _only_ when the command fails to find the given file.

Still, I'm not fully sure we want this, I'd like to hear more about your use case :thinking:

yuriShafet:
@deivid-rodriguez I agree with you. Indeed requiring `require "active_support"` will work fine  as well as `gem which active_support`. 

Actually the help description of `which` command proves that my assumption  that `list` output should be always found with `which`  is wrong.
```
  Summary:
    Find the location of a library file you can require
```
I think, ill investigate if changes to did_you_mean can do any good and meanwhile ill close the issue and PR.
What do you think?

deivid-rodriguez:
Sure! From my side, I think I good behaviour from the end user point of view would be:

```
$ gem which activesupport
ERROR:  Can't find Ruby library file or shared library "activesupport"
ERROR: I did find `/usr/local/bundle/gems/activesupport-5.2.6/lib/active_support.rb`, if "active_support" is what you meant
```

And again, this extra sure would need to only be run if we fail to find the exact argument given originally by the user.

PR #4814, opened 2021-08-03
Fix bin/bundle path suggestion for local setup

nickcharlton:

In trying to test out #4805 locally, I was confused by how to reference my local version of the bundler executable. In https://github.com/rubygems/rubygems/commit/ecf7dc53d975f64ade0d36d99f2be52c8f5df7ab this moved to the spec directory.

This changes the hint to use an alias that will reference the right path. We also need to make the file executable again.

### DIFF
--- a/bundler/doc/development/SETUP.md
+++ b/bundler/doc/development/SETUP.md
@@ -30,7 +30,7 @@ To work on Bundler, you'll probably want to do a couple of things:
 
 * Set up a shell alias to run Bundler from your clone, e.g. a Bash alias ([follow these instructions](https://www.moncefbelyamani.com/create-aliases-in-bash-profile-to-assign-shortcuts-for-common-terminal-commands/) for adding aliases to your `~/.bashrc` profile):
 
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (commented):
Nice!

I normally install both `bundler` & `rubygems` by running `ruby setup.rb` from the root of the repo, or just `bundler` by running `bin/rake install` from the `bundler/` folder. We should probably document those alternatives as well, and how to undo them, but we can tackle that separately.

This is a less invasive way of running the development version of `bundler` (although a bit less close to real usage), so probably worth keeping, and the instructions were indeed broken since that file was moved :+1:.

deivid-rodriguez:
Looks good, only comment is that since we're running the file through `ruby` now, we don't need to change its permissions, since unless we also add a shebang to it, it won't be runnable directly.

nickcharlton:
Thanks for merging this in! Sorry I didn't get to make the changes before you did. I appreciate it.

deivid-rodriguez:
No worries! :)

PR #4815, opened 2021-08-04
Avoid shared libraries not getting environment passed right after argv in memory when `bundle exec` is used

### LINKED ISSUE #4803: `bundler exec` does not pass AWS session into the share lib

Using ffi to load a shared library that does some AWS operations. But when calling fails because the right profile information is [not passed into library](https://github.com/ffi/ffi/issues/909).

Due to proprietary reasons i am not able to provide you with the shared library at the moment to test this. But I will see if I could create one for this reason and replicate the issue. I am unsure if the issue is with bundler or ffi.

BTW I get this error, I have tested with 2.1.5+ version of the bundler i have this issue. But this issue is not there in 2.0.0 and lower versions. Did not check with 2.0.0 - 2.1.5 versions.

`bundler exec ruby <script_name>`

AWS response is expected

### What actually happened?

`get_resp': NoCredentialProviders: no valid providers in chain (StandardError)
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed

## Environment

```
Bundler             2.2.24
  Platforms         ruby, x86_64-linux
Ruby                2.7.3p183 (2021-04-05 revision 6847ee089d7655b2a0eea4fee3133aeacd4cc7cc) [x86_64-linux]
  Full Path         /usr/share/rvm/rubies/ruby-2.7.3/bin/ruby
  Config Dir        /etc
RubyGems            3.1.6
  Gem Home          /usr/share/rvm/gems/ruby-2.7.3@ex1
  Gem Path          /usr/share/rvm/gems/ruby-2.7.3@ex1:/usr/share/rvm/rubies/ruby-2.7.3/lib/ruby/gems/2.7.0
  User Home         /home/sam
  User Path         /home/sam/.gem/ruby/2.7.0
  Bin Dir           /usr/share/rvm/gems/ruby-2.7.3@ex1/bin
Tools               
  Git               2.25.1
  RVM               1.29.12 (manual)
  rbenv             not installed
  chruby            not installed
  rubygems-bundler  (1.4.5)
```

## Bundler Build Metadata

```
Built At          2021-07-15
Git SHA           d78b1ee235
Released Version  true
```

## Bundler settings

```
build.ffi
  Set for the current user (/home/sam/.bundle/config): "--enable-system-libffi"
```

## Gemfile

```ruby
source 'https://rubygems.org'

...proprietary stuff here...

gem 'aws-sdk'
gem 'http-parser'
```

### Gemfile.lock

```
GIT
  <proprietary stuff here>
  specs:
    <proprietary stuff here>
      ffi (~> 1.15.3)
      ffi-compiler (>= 1.0, < 2.0)
      google-protobuf (~> 3.17.3)

deivid-rodriguez:

Due to the way `bundle exec` is passing `ENV` to the executed script, shared libraries can no longer assume that ENV should immediately follow ARGV in memory. For example, the `go` runtime makes this assumption.

Add a missing early return so that we don't replace the `ENV` twice, once with `ENV.replace` and another one manually through `ENV.clear` and then manually filling it. This was a clear bug, and `ENV.replace` does not cause the above issues.

Closes #4803.


deivid-rodriguez:
This seems to make Windows specs hang, not sure why.

deivid-rodriguez:
This is finally ready!

Issue #3831, opened 2020-07-14
undefined method `rubyforge_project=' for class `Gem::Specification' (NameError)


utkarsh2102:
Whilst trying to install rubygems manually, I tried running `ruby setup.rb`, as suggested in the README.md file.
Upon doing so, I got:

```
➜  rubygems git:(master) ruby setup.rb
Traceback (most recent call last):
	5: from setup.rb:23:in `<main>'
	4: from setup.rb:23:in `require'
	3: from /home/utkarsh/debian/ruby-team/rubygems/rubygems/lib/rubygems.rb:1347:in `<top (required)>'
	2: from /home/utkarsh/debian/ruby-team/rubygems/rubygems/lib/rubygems.rb:1347:in `require'
	1: from /usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb:90:in `<top (required)>'
/usr/lib/ruby/vendor_ruby/rubygems/defaults/operating_system.rb:91:in `<class:Specification>': undefined method `rubyforge_project=' for class `Gem::Specification' (NameError)
```

I believe this has been deprecated and shouldn't really be used :thinking: 
Am I missing something or is it really a bug!?

simi:
@utkarsh2102 you do mix system and custom ruby installations. It is not supported. Please install ruby manually (not by system package, for example by ruby-install or rbenv) and update rubygems manually or keep your ruby updated by system (not manually).

@deivid-rodriguez `operating_system.rb` seems really dangerous place actually. what about to rescue all errors during requiring and next to the stack-trace report user friendly message explaining that there are some problems at rubygems distributor side.

utkarsh2102:
Hi @simi,

> you do mix system and custom ruby installations. It is not supported. Please install ruby manually (not by system package, for example by ruby-install or rbenv) and update rubygems manually or keep your ruby updated by system (not manually).

Gah, I see. Thanks for the tip :+1: 
(we're working on packaging rubygems, btw)

simi:
@utkarsh2102 yeah, I know. This exact problem should be already fixed at https://salsa.debian.org/ruby-team/rubygems-integration/-/commit/39d3a082f568ec0b6c9acd10e54a4a31a08bd8b5.

utkarsh2102:
Aha, I see. Thanks \o/

hsbt:
Duplicate with https://github.com/rubygems/rubygems/issues/3068

deivid-rodriguez:
[...]

PR #4824, opened 2021-08-05
Enhance the error handling for loading the `rubygems/defaults/operating_system` file

### LINKED ISSUE #4585: Rescue errors when loading `operating_system.rb` and tell users to report the problem to the OS
My current problem is that we keep getting issues that we can't really fix, because they happen inside `operating_system.rb`, which is not under our control, but on the OS packager's side.

Let's at least rescue any errors inside `operating_system.rb` and display a message making it clear where the problem is and where it should be reported, as suggested by @simi at https://github.com/rubygems/rubygems/issues/3831#issuecomment-658382577, so that at least we can save people some time.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

intuxicated:

Issue #4585 

> My current problem is that we keep getting issues that we can't really fix, because they happen inside operating_system.rb, which is not under our control, but on the OS packager's side.
>
> Let's at least rescue any errors inside operating_system.rb and display a message making it clear where the problem is and where it should be reported, as suggested by @simi at #3831 (comment), so that at least we can save people some time.

When loading `rubygems/defaults/operating_system`
- we want to keep it silent if the raised exception is a LoadError
- we want to print a message in other cases and ask users to report the issue to their OS support.

Ruby 3 comes with special error handling for loading `rubygems` and it will show a warning when LoadError exception is raised for requiring 'rubygem'.
Because of that, we decided to leave the LoadError scenario as it is.
Reference: https://github.com/ruby/ruby/blob/d1998d8767affe58be0bd09ec536dae9198a7fbd/gem_prelude.rb#L1-L5

Closes #4585.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1339,7 +1339,12 @@ begin
 rescue LoadError
   # Ignored
 rescue Exception => e
-  raise e.class, "#{e.message}\nThis is not expected so please report this issue to your OS support and ask for help."
+  msg = "#{e.message}\n" \
+    "Loading the rubygems/defaults/operating_system.rb file caused an error. " \
+    "This file is owned by your OS, not by rubygems upstream. " \
+    "Please find out which OS package this file belongs to and follow the guidelines from your OS to report " \
+    "the problem and ask for help."
+  raise e.class, msg
 end
 
 begin
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1337,6 +1337,9 @@ begin
 
   require 'rubygems/defaults/operating_system'
 rescue LoadError
+  # Ignored
+rescue Exception => e
+  raise e.class, "#{e.message}\nThis is not expected so please report this issue to your OS support and ask for help."
 end
 
 begin
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1338,7 +1338,7 @@ begin
   require 'rubygems/defaults/operating_system'
 rescue LoadError
   # Ignored
-rescue Exception => e
+rescue StandardError => e
   msg = "#{e.message}\n" \
     "Loading the rubygems/defaults/operating_system.rb file caused an error. " \
     "This file is owned by your OS, not by rubygems upstream. " \
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -691,6 +691,7 @@ test/rubygems/test_kernel.rb
 test/rubygems/test_project_sanity.rb
 test/rubygems/test_remote_fetch_error.rb
 test/rubygems/test_require.rb
+test/rubygems/test_rubygems.rb
 test/rubygems/utilities.rb
 test/rubygems/wrong_key_cert.pem
 test/rubygems/wrong_key_cert_32.pem
--- a/test/rubygems/test_rubygems.rb
+++ b/test/rubygems/test_rubygems.rb
@@ -7,6 +7,8 @@ class GemTest < Gem::TestCase
   end
 
   def test_operating_system_other_exceptions
+    pend "does not apply to truffleruby" if RUBY_ENGINE == 'truffleruby'
+
     path = util_install_operating_system_rb <<-RUBY
       intentional synt'ax error
     RUBY
--- a/test/rubygems/test_rubygems.rb
+++ b/test/rubygems/test_rubygems.rb
@@ -10,12 +10,12 @@ class GemTest < Gem::TestCase
     pend "does not apply to truffleruby" if RUBY_ENGINE == 'truffleruby'
 
     path = util_install_operating_system_rb <<-RUBY
-      intentional synt'ax error
+      intentionally_not_implemented_method
     RUBY
 
     output = Gem::Util.popen(*ruby_with_rubygems_and_fake_operating_system_in_load_path(path), '-e', "'require \"rubygems\"'", {:err => [:child, :out]}).strip
     assert !$?.success?
-    assert_includes output, "unterminated string meets end of file (SyntaxError)"
+    assert_includes output, "undefined local variable or method `intentionally_not_implemented_method'"
     assert_includes output, "Loading the rubygems/defaults/operating_system.rb file caused an error. " \
     "This file is owned by your OS, not by rubygems upstream. " \
     "Please find out which OS package this file belongs to and follow the guidelines from your OS to report " \
--- a/test/rubygems/test_rubygems.rb
+++ b/test/rubygems/test_rubygems.rb
@@ -15,7 +15,11 @@ class GemTest < Gem::TestCase
 
     output = Gem::Util.popen(*ruby_with_rubygems_and_fake_operating_system_in_load_path(path), '-e', "'require \"rubygems\"'", {:err => [:child, :out]}).strip
     assert !$?.success?
-    assert_includes output, "This is not expected so please report this issue to your OS support and ask for help"
+    assert_includes output, "unterminated string meets end of file (SyntaxError)"
+    assert_includes output, "Loading the rubygems/defaults/operating_system.rb file caused an error. " \
+    "This file is owned by your OS, not by rubygems upstream. " \
+    "Please find out which OS package this file belongs to and follow the guidelines from your OS to report " \
+    "the problem and ask for help."
   end
 
   private
--- a/test/rubygems/test_rubygems.rb
+++ b/test/rubygems/test_rubygems.rb
@@ -1,0 +1,38 @@
+require_relative 'helper'
+
+class GemTest < Gem::TestCase
+  def test_rubygems_normal_behaviour
+    _ = Gem::Util.popen(*ruby_with_rubygems_in_load_path, '-e', "'require \"rubygems\"'", {:err => [:child, :out]}).strip
+    assert $?.success?
+  end
+
+  def test_operating_system_other_exceptions
+    path = util_install_operating_system_rb <<-RUBY
+      intentional synt'ax error
+    RUBY
+
+    output = Gem::Util.popen(*ruby_with_rubygems_and_fake_operating_system_in_load_path(path), '-e', "'require \"rubygems\"'", {:err => [:child, :out]}).strip
+    assert !$?.success?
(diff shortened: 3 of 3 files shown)


lib/rubygems.rb
```diff
@@ -1336,7 +1336,10 @@ def default_gem_load_paths
   # Defaults the operating system (or packager) wants to provide for RubyGems.
 
   require 'rubygems/defaults/operating_system'
-rescue LoadError
+rescue LoadError => e
+  raise e unless e.message.include? 'rubygems/defaults/operating_system'
```
deivid-rodriguez, line 1340: I think there's `LoadError#path` for this. Also, you can avoid generating a new exception and raise the original one with `raise` without arguments.

deivid-rodriguez (commented):
I think the problem with the test is that you're generating a `LoadError` "artificially", so it comes with `LoadError#path` as `nil`, which makes your code crash. Instead, I think we need to run something closer to real life:

* For a `LoadError` of the `rubygems/default/operating_system` file, I don't think we need to test anything because this is what happens most of the times: rubygems tries to load this file, there's no such file, so the error is ignored. I guess is something was broken here, we should be noticing through other specs, even from bundler specs.

* For a `LoadError` of a file inside `rubygems/defaults/operating_system` file, you can generate a `rubygems/defaults/operating_system` file that tries to load a file that doesn't exist, with something like `require "file_that_doesnt_exist"`, and check that rubygems also says that this is an issue on the OS packaging side.

* For a different error inside `rubygems/defaults/operating_system`, I think the test you wrote is fine.

Regarding the specific wording, I'm not too sure either. @simi any ideas?

intuxicated:
Ruby 3 comes with special error handling for `rubygems`.

https://github.com/ruby/ruby/blob/d1998d8767affe58be0bd09ec536dae9198a7fbd/gem_prelude.rb#L1-L5
```
begin
  require 'rubygems'
rescue LoadError
  warn "`RubyGems' were not loaded."
end if defined?(Gem)
```

That implementation is making it hard to test ruby 3.

deivid-rodriguez:
Nice find @intuxicated!

I wasn't aware that change made it into ruby-core, and I'm not sure I agree with it, but it already happened, so now we need to think how it affects this feature.

Basically that change means that now any errors while loading rubygems will be swallowed (including the ones we want to warn about here) and all the end user will see will be a "`RubyGems' were not loaded." warning message. That seems bad to me because later code is, I believe, quite likely to rely on being able to load gems, so it will error at a later stage and the error will be hard to figure out.

I believe what https://github.com/ruby/ruby/pull/2764 intended to warn is the case where `rubygems.rb` is not available at all, but I'm not sure it explicitly intended to silently swallow other `LoadError`s like this one. If that's the case, the solution would be to do something similar to what we're doing here in ruby-core and only swallow the error if the culprit is `rubygems` itself not being available for require:

```ruby
begin
  require 'rubygems'
rescue LoadError => e
 raise unless e.path == 'rubygems'
  warn "`RubyGems' were not loaded."
end if defined?(Gem)
```

We should probably summon @hsbt and @voxik to help us here.

deivid-rodriguez:
Something else I'm thinking is that, while technically possible, it's unlikely that a `rubygems/defaults/operating_system` file will ever result in a `LoadError`. At least, it has never happened in real life yet.

So maybe we can skip that part for the moment and rescue only other exceptions. That wouldn't result in any issues with tests since `LoadError`'s wouldn't be involved at all.

voxik:
1) I agree with the idea of rescuing all operating_system.rb exceptions, while I'll do everything to avoid their use (and I hope Fedora/RHEL was not the reason for this PR :angel:).
2) In years we are using operating_system.rb in Fedora/RHEL, we had no need to load any other file inside. I think the chances to need it are pretty low. The basic Ruby file operation libraries are probably preloaded by Ruby/RubyGems at that time. I don't know what else should be needed there.
3) If the gem_prelude.rb was more cautious about the `LoadError`, that would be probably good idea. Why there is nothing like `try_require` in Ruby, which would provide similar functionality?

> That seems bad to me because later code is, I believe, quite likely to rely on being able to load gems, so it will error at a later stage and the error will be hard to figure out.

I am not sure I understand the concern here. Are you concerned that did_you_mean needs RubyGems (it does not if I am not mistaken) or about something else?

deivid-rodriguez:
Alright @intuxicated, let's then forget about `LoadError` in this PR, and just rescue other errors :+1:.

Regarding my concerns, they were related to swallowing exceptions when loading `rubygems`, not related to `did_you_mean`. I am unaware of the situations where Fedora does not provide rubygems with ruby, so concerns are probably unfounded, but I was imagining someone installing this rubygems-less ruby and some packaged gem and then getting an error when trying to use the gem and not being able to associate the error to the fact that the ruby they are using comes without rubygems.

hsbt:
In basically, ruby core team don't intend to separate ruby interpreter and rubygems. So, this situation is not intentional behavior for ruby core.

But I understood to fedora/debian situation.

intuxicated:
I've run into a similar issue with `truffleruby`, it will return false when any error happened while loading rubygems.

To confirm the situation, I used the following command

```bash
$ ruby -v
truffleruby 21.2.0, like ruby 2.7.3, GraalVM CE Native [x86_64-darwin]

% bundler -v
Bundler version 2.2.25

$ gem -v
3.1.6

$ irb      
irb(main):001:0> require 'rubygems'
=> false

irb(main):002:0> require 'nonexisting-gem'
        from $HOME/.rbenv/versions/truffleruby-21.2.0/bin/irb:42:in `<main>'
        from <internal:core> core/kernel.rb:400:in `load'
        from <internal:core> core/kernel.rb:400:in `load'
        from /Users/daniel_niknam_personal/.rbenv/versions/truffleruby-21.2.0/lib/gems/gems/irb-1.3.3/exe/irb:11:in `<top (required)>'
        from <internal:core> core/throw_catch.rb:36:in `catch'
        from <internal:core> core/throw_catch.rb:36:in `catch'
        from <internal:core> core/kernel.rb:437:in `loop'
        from <internal:core> core/kernel.rb:325:in `eval'
        from (irb):2:in `<top (required)>'
        from $HOME/.rbenv/versions/truffleruby-21.2.0/lib/mri/rubygems/core_ext/kernel_require.rb:83:in `require'
<internal:core> core/kernel.rb:236:in `gem_original_require': cannot load such file -- nonexisting-gem (LoadError)
```

lib/rubygems.rb
```diff
@@ -1337,6 +1337,9 @@ def default_gem_load_paths
 
   require 'rubygems/defaults/operating_system'
 rescue LoadError
+  # Ignored
+rescue Exception => e
+  raise e.class, "#{e.message}\nThis is not expected so please report this issue to your OS support and ask for help."
```
intuxicated, line 1342: See https://github.com/rubygems/rubygems/pull/4824#issuecomment-894757680

lib/rubygems.rb
```diff
@@ -1337,6 +1337,9 @@ def default_gem_load_paths
 
   require 'rubygems/defaults/operating_system'
 rescue LoadError
+  # Ignored
```
intuxicated, line 1340: See https://github.com/rubygems/rubygems/pull/4824#issuecomment-894696805

deivid-rodriguez:
Yeah! Truffleruby completely disables this feature by shipping an emtpy file: https://github.com/oracle/truffleruby/blob/966f7385107085b60d13fe648f5496d543c5b5d5/lib/truffle/rubygems/defaults/operating_system.rb. So it does indeed not apply :+1:.

lib/rubygems.rb
```diff
@@ -1337,6 +1337,9 @@ def default_gem_load_paths
...
+rescue Exception => e
```
deivid-rodriguez, line 1341: In general, it's not recommended to `rescue Exception`, and it's best to let other exceptions go. I say let's do the same here and rescue only `StandardError`.

deivid-rodriguez: Regarding wording, I would say something like

> Loading the `rubygems/defaults/operating_system.rb` file caused an error. This file is owned by your OS, not by rubygems upstream. Please find out which OS package this file belongs to and follow the guidelines from your OS to report the problem and ask for help.

intuxicated: @deivid-rodriguez While I agree with you on best practices, I think it could be better for us to capture any Exception because we are not going to silence the error and it will be raised again with a new message.

The problem with using `StandardError` is it doesn't cover `SyntaxError` which I thought could be important to us since we don't own the `rubygems/defaults/operating_system`. However, I'm happy to change the code to rescue `StandardError` instead.

deivid-rodriguez: Yeah, I was thinking along those lines too.

In the case of `SyntaxError`, just like with `LoadError`, I think it's really unlikely that an OS is going to ship a ruby file with invalid syntax, so it feels fine to ignore that.

With other errors, likely say `NoMemoryError`, if they happen, we can't really be sure it's the fault of the `rubygems/defaults/operating_system` file, so I'd say it's better to let them be raised unchanged.

That's my reasoning for rescuing only `StandardError`:, it will cover the real life situations that have already happened, but no others that might or might not result in "false positives".

deivid-rodriguez (approved):
It looks good to me! I'll leave it there a couple of days in case there's more feedback, specially on wording.

voxik:
For a second, I thought that the message could be customizable via operating_system.rb, but that is probably silly idea :zany_face:

intuxicated:
@voxik 😄 I think it's a "nice to have" feature.

deivid-rodriguez:
Thanks @intuxicated! I'll merge this!

Allowing this message to be customized seems overkill to me :sweat_smile:.

matkoniecz:
Would it be feasible to list the absolute path in file system? Not just relative path? Base of that relative path is likely obvious for someone developing rubygems but for me is complete mystery and I would need to start digging in documentation (reported as #5334) 

Would it be feasible to just list absolute path to the problematic file in the error message?

deivid-rodriguez:
Good idea @matkoniecz, I'm happy to accept such a change.

PR #4827, opened 2021-08-06
Remove `RUBYGEMS_GEMDEPS` warning

deivid-rodriguez:

When setting the `RUBYGEMS_GEMDEPS` environment variable to allow skipping `bundle exec`, `bundler` will print a warning about potential incompatibility.

Initially the `RUBYGEMS_GEMDEPS` variable used a completely different (re)implementation of `bundler` functionality. That
implementation was not battle tested and could potentially differ in behaviour from what `bundler` does. That's why print a warning.

However, these days, all `rubygems` does when `RUBYGEMS_GEMDEPS` is set is to require `bundler/setup`, so there's no risk of any incompatibility, since that's just plain `bundler`.

Completely remove the warning.


simi (approved):

deivid-rodriguez:
Thank you for your reviews @simi :heart:

PR #4833, opened 2021-08-08
Remove MacOS specific gem layout

### LINKED ISSUE #2890: Default gems location as framework on macOS
In the case of installation as system bundled framework on macOS (when `RUBY_FRAMEWORK_VERSION` is defined), `Gem.default_dir` returns a  path derived from `RbConfig::CONFIG['sitedir']`.
Regarding the bundled framework on macOS Mojave, default gems are placed there but other vendor bundled gems are placed in `/usr/lib/ruby/gems/2.3.0/gems` (= `/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/lib/ruby/`).

I think that also the former gems including defaults should be placed in the latter location.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [x] Other

```
$ gem env version
PASTE HERE
```

deivid-rodriguez:

MacOS should properly configure Ruby. They should not expect us to maintain a different layout just for them.

Don't maintain a specific default layout just for MacOS.

Closes #2890.

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -462,20 +462,8 @@ By default, this RubyGems will install gem as:
       lib_dir = RbConfig::CONFIG[site_or_vendor]
       bin_dir = RbConfig::CONFIG['bindir']
     else
-      # Apple installed RubyGems into libdir, and RubyGems <= 1.1.0 gets
-      # confused about installation location, so switch back to
-      # sitelibdir/vendorlibdir.
-      if defined?(APPLE_GEM_HOME) and
-        # just in case Apple and RubyGems don't get this patched up proper.
-        (prefix == RbConfig::CONFIG['libdir'] or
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Ironically, they monkeypatch the code we maintain just for them :rofl: https://opensource.apple.com/source/ruby/ruby-145.40.1/patches/lib_rubygems_defaults.rb.diff.auto.html

hsbt (approved):

bronzdoc (approved):

deivid-rodriguez:
This will break `gem update --system` on ruby packaged by MacOS. Not really supported anyways, but there's no hurry to ship this, we can keep it in master and release it with the next major.

voxik:
Nice, they should use operating_system.rb as every other platform

PR #4841, opened 2021-08-13
Only check if descriptions *start with* FIXME/TODO

### LINKED ISSUE #4808: TODO should not be a forbidden word
This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [X] The command line `gem`
  - [X] Other

Today I found out that this .gemspec leads to gem being unable to create the .gem.

     DESCRIPTION = <<-EOF

    Does this make sense that we can not upload this when TODO is a substring ...

      EOF

      s.summary     = DESCRIPTION
      s.description = DESCRIPTION

The issue is that "TODO" as a substring leads to the following message:

    "FIXME" or "TODO" is not a description

My problem is: why does gem think that "TODO" is not a valid description?

I actually had a 12 lines long description and the last part was "long-term
stability is on the TODO list". Why is this a forbidden word?

I can understand it a little bit perhaps if the FIRST word is "TODO:" and
then this may be a feature. But having a substring at a later point 
become invalid as an error? Isn't that ... weird? I reworded this for my
gemspec, so gem works fine ... but I really don't think gem should 
prevent me via an error. I see no problem with "TODO" as part 
of a summary or description. Any other forbidden words to add? ;)

(I assume this was more aimed as a feature for forgetful people;
I just fail to see how my use case fits to that hardcoded block. 
I would propose that perhaps as a fix, TODO as a word is allowed
after some other words came prior to it. But to be fair, I actually 
do not even think that any restriction should be allowed where gem
itself is opinionated. It's very strange ... not very important either
but it confused me since it did not make sense that in sentence
line #11 I can not upload a gem anymore because TODO is 
a forbidden word ... )

duckinator:
Fixes #4808.

RubyGems outright forbids the words "FIXME" or "TODO" in a gem description.

I made RubyGems only check if the gem description _starts with_ "FIXME" or "TODO".

### DIFF
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -2781,6 +2781,20 @@ duplicate dependency on c (>= 1.2.3, development), (~> 1.2) use:
       end
 
       assert_equal %("#{f}" or "#{t}" is not a description), e.message
+
+      # Adding #{f} anywhere after the start of the description should be fine.
+      @a1.description = "(some description) #{f}"
+
+      assert_nothing_raised do
+        @a1.validate
+      end
+
+      # Adding #{t} anywhere after the start of the description should be fine.
+      @a1.description = "(some description) #{t}"
+
+      assert_nothing_raised do
+        @a1.validate
+      end
     end
   end
 
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -381,7 +381,7 @@ http://spdx.org/licenses or '#{Gem::Licenses::NONSTANDARD}' for a nonstandard li
   end
 
   LAZY = '"FIxxxXME" or "TOxxxDO"'.gsub(/xxx/, '')
-  LAZY_PATTERN = /FI XME|TO DO/x.freeze
+  LAZY_PATTERN = /^FI XME|^TO DO/x.freeze
   HOMEPAGE_URI_PATTERN = /\A[a-z][a-z\d+.-]*:/i.freeze
 
   def validate_lazy_metadata


test/rubygems/test_gem_specification.rb
```diff
@@ -2781,6 +2781,14 @@ def test_validate_description
       end
 
       assert_equal %("#{f}" or "#{t}" is not a description), e.message
+
+      # Allow these words anywhere after the start of the description.
+      # See https://github.com/rubygems/rubygems/issues/4808
+      @a1.description = "(some description) #{f}"
```
bronzdoc, line 2787: Don't we want to check if `TODO` or `FIXME` is in this string?

duckinator: Yes, and FIXME _is_ in that string -- that's what `#{f}` is. The words TODO/FIXME are obfuscated throughout this entire part of the codebase, presumably so `git grep -E 'TODO|FIXME'` to find areas that need work doesn't give false positives.

duckinator: See this part of the file: https://github.com/rubygems/rubygems/blob/0c50fd5c4ad7da55affece14d8307ec8aba78942/test/rubygems/test_gem_specification.rb#L2551-L2554

deivid-rodriguez: I think you also want to check the same thing with `#{t}` as well to cover the change in the corresponding regexp?

As a personal nitpick, I usually avoid linking to specific GitHub issues. If something needs a code comment, I'd rather explain things from scratch. Some reasons why I prefer that:

* A short explanation is usually about the same length as a GitHub link.
* It makes the code independent from where it's hosted. For example, when we moved the repository, we had to update hundreds of code links like this.
* I think a link might actually distract the reader and make her lose focus.
* If more information is really needed, the PR that introduced the code can be found through `git blame` or other tools easily.
* If we added a link to the closed issue every time we fixed an issue, the source code would be full of those links, and if we do it only in some cases, it's hard to explain why those cases are special.

It's not a strong opinion anyways, and I'm happy with leaving it if you think it's best to link to the full issue in this case!

duckinator: I'll add a check with `#{t}` and replace the link with a better description.

duckinator: Done. :tada:

duckinator:
@bronzdoc @deivid-rodriguez I think this should be good to go now.

deivid-rodriguez (approved):
Thank you!

PR #4834, opened 2021-08-10
Remove unused file

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Nothing, just an opportunity for cleanup.

## What is your fix for the problem, implemented in this PR?

Remove the file. It's no longer used since https://github.com/rubygems/rubygems/commit/b7f6270b9e5e15591cd679f60fa8f43e6763f3f5.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4842, opened 2021-08-13
Revert "Remove unused file"

soffieprincess09:
Reverts rubygems/rubygems#4834


bronzdoc:
@soffieprincess09 why do you think that PR should be reverted?

soffieprincess09:
coz i just want fix and better my my work

PR #4846, opened 2021-08-16
Add concurrency to workflows

intuxicated:

_Changes in this PR will not affect the end-user._

_Changes in this PR will not affect the end-user._

## Change

Add a cancel workflow which should be triggered by `git push` and will cancel previously triggered GitHub workflow/actions and only keeps the last one.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -9,7 +9,7 @@ on:
       - 3.2
 
 concurrency:
-  group: ${{ github.head_ref }}
+  group: ci-${{ github.ref }}-${{ github.workflow }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -8,6 +8,10 @@ on:
       - master
       - 3.2
 
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   bundler:
     name: Bundler ${{ matrix.bundler.name }} on ${{ matrix.os.name }} (${{ matrix.ruby.name }})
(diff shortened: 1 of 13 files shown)


deivid-rodriguez:
@intuxicated I was research on the current state of things on the github side, and I _think_ this might be supported now? Can you check https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#concurrency and verify whether it is what we need?

intuxicated:
Thanks, @deivid-rodriguez. I've seen it too.  GitHub documentation is saying:

> Note: Concurrency is currently in beta and subject to change.

So I thought we give this a try first and now given how hard it is to get this thing right, I think I'll give the `concurrency` option a try.

deivid-rodriguez (commented):
I think you need to use the per-job setting, otherwise all workflows but one will be cancelled from a given push'

intuxicated:
@deivid-rodriguez we can give it a try and see if that's actually what's going to happen but if I understand the document correctly, the `group: ${{ github.head_ref }}` should prevent that from happening. the `group` attribute is the key to find previous actions and `${{ github.head_ref }}` [should return](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) the source branch for the pull request.

deivid-rodriguez:
Yes, I think this way all workflows get grouped together for a given push, and only one is run and the others cancelled. That's why only the `lint` job was run for this PR, which is not what we want.

Instead, I think we want to either use the job-level workflow setting, or maybe add `github.workflow` to the `group` expresion, like `group: ci-${{ github.ref }}-${{ github.workflow }}`?

intuxicated:
We gave the job-level `concurrency` settings a try but no matter what group we set, every new job was getting canceled. GitHub has a note for job-level `concurrency` as follow:

> Note: When concurrency is specified at the job level, order is not guaranteed for jobs or runs that queue within 5 minutes of each other.

We think that could be the cause.

deivid-rodriguez (approved):
Let's try it, thanks @intuxicated!

PR #4864, opened 2021-08-19
Make plugin installation idempotent

### LINKED ISSUE #4863: Plugin installation is not idempotent
First install works

```
bundle plugin install diffend
Resolving dependencies...
Using bundler 2.2.26
Fetching diffend 0.2.49
Installing diffend 0.2.49
Installed plugin diffend
```

But retrying the install procedure after a successful install:

```
bundle plugin install diffend
Resolving dependencies...
Using bundler 2.2.26
Using diffend 0.2.49
Failed to install the following plugins: `diffend`. The underlying error was: plugins.rb was not found in the plugin..
 /home/mencio/.rbenv/versions/2.7.4/lib/ruby/site_ruby/2.7.0/bundler/plugin.rb:263:in `validate_plugin!'
 /home/mencio/.rbenv/versions/2.7.4/lib/ruby/site_ruby/2.7.0/bundler/plugin.rb:249:in `block in save_plugins'
```

_Originally posted by @mensfeld in https://github.com/rubygems/rubygems/issues/4730#issuecomment-901691154_

deivid-rodriguez:

When running `bundle plugin install` twice, the second time fails.

Fix is to guard for already installed plugins, and ignore those.

While working on it, I realized that plugin installation errors, while being printed, they were not resulting in a non-zero status code. I also fixed that issue that was actually preventing our specs from catching this, and also improved some error messages.

Fixes #4863.

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -246,9 +246,8 @@ module Bundler
     def save_plugins(plugins, specs, optional_plugins = [])
       plugins.each do |name|
         spec = specs[name]
-        validate_plugin! Pathname.new(spec.full_gem_path)
-        installed = register_plugin(name, spec, optional_plugins.include?(name))
-        Bundler.ui.info "Installed plugin #{name}" if installed
+
+        save_plugin(name, spec, optional_plugins.include?(name))
       end
     end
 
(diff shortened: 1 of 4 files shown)


mensfeld:
@deivid-rodriguez shouldn't a second install actually update the plugin if not the most recent version? there is no `update` in the plugin API now.

deivid-rodriguez:
Mmmmm, not sure honestly, but I would expect that if a compatible version is installed, no reinstallation happens.

I think updating should be implemented on its own command, or by passing explicit `--version` to `bundle plugin install`.

There seems to be more issues here, by the way:

```
$ bundle plugin uninstall diffend 
Uninstalled plugin diffend

$ bundle plugin install diffend 
Resolving dependencies...
Using bundler 2.3.0.dev
Installing diffend 0.2.49
Installed plugin diffend

$ bundle plugin install diffend --version=0.2.48
Resolving dependencies...
Using bundler 2.3.0.dev
Using diffend 0.2.48
```

I never installed 0.2.48 before :confused:.

mensfeld:
Another bug? :D

deivid-rodriguez:
Yeah.

deivid-rodriguez:
In any case, this PR changes an "error + do nothing" to "no error + do nothing" which seems strictly better. Specially given that before the error bundler prints something like "Using diffend 0.2.49", suggesting that the intended behaviour is not to update, but to do nothing if the plugin is already installed.

deivid-rodriguez:
The latest issue seems to be because only one version per plugin is allowed to be installed at the same time... So there's no currently a way to update a plugin other than uninstalling and then installing again. 

I think this would require to be completely reworked to have a better behaviour, but a quick way out for the time being would be to do what @mensfeld suggested and install the latest version (or the one requested) every time :man_shrugging:-

deivid-rodriguez:
Well, the above is actually wrong. Multiple plugins can be installed at the same time, so this should be fixable without a major rewrite.

deivid-rodriguez:
#4869 should fix the latest issue.

PR #4868, opened 2021-08-21
Redact credentails when printing URI

intuxicated:

When installing a gem from a source that contains credentials, they will be exposed on printing exceptions (with the `--debug` option) and on printing verbose logs.

- Redact credentials in `Gem::RemoteFetcher::FetchError` 
- Redact credentials in `Gem::Request`
- Redact credentials when printing error on installing gems


lib/rubygems/commands/install_command.rb
```diff
@@ -260,7 +261,8 @@ def show_install_errors(errors) # :nodoc:
     errors.each do |x|
       return unless Gem::SourceFetchProblem === x
 
-      msg = "Unable to pull data from '#{x.source.uri}': #{x.error.message}"
```
intuxicated, line 263: The `x.source.uri` could be a URI with a credential so `Gem::PrintableUri` is used here to make sure that sensitive information will not be exposed.

lib/rubygems/commands/install_command.rb
```diff
@@ -260,7 +261,8 @@ def show_install_errors(errors) # :nodoc:
...
+      printable_uri = Gem::PrintableUri.parse_uri(x.source.uri.clone)
+      msg = "Unable to pull data from '#{printable_uri}': #{x.error.message}"
```
intuxicated, line 265: I've run the command by myself and I didn't see any credential inside the `x.error.message` so I kept it as it is.

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+  def parse_uri
+    @original_uri = Gem::UriParser.parse_uri(@original_uri)
+    @uri = @original_uri.clone
+    redact_credential
+
+    self
+  end
+
+  def parsed_uri?
+    @uri.is_a? URI::Generic
```
intuxicated, line 25: the `URI` could return `URI::HTTP` or `URI::HTTPS`, etc, when parsing a URI string but every class is inheriting the `URI::Generic`. 
I was thinking to rename this method to `parsable_uri?` but I wasn't sure.

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def credential_redacted?
+    @credential_redacted
+  end
+
+  def original_password
```
intuxicated, line 32: This method is used only in `Gem::RemoteFetcher::FetchError` to redact the password from the message (if exists).

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+    @credential_redacted = true
+  end
+
+  def redactable_credential?
+    return false unless parsed_uri?
+
+    password? || oauth_basic? || token?
+  end
+
+  def password?
```
intuxicated, line 64: `http://user:password@example.com`

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+    return false unless parsed_uri?
+
+    !!@uri.password
+  end
+
+  def oauth_basic?
```
intuxicated, line 70: `http://token:x-oauth-basic@example.com`

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+    return false unless parsed_uri?
+
+    @uri.password == 'x-oauth-basic'
+  end
+
+  def token?
```
intuxicated, line 76: `http://token@example.com`

lib/rubygems/remote_fetcher.rb
```diff
@@ -21,19 +22,24 @@ class Gem::RemoteFetcher
   class FetchError < Gem::Exception
     ##
     # The URI which was being accessed when the exception happened.
+    def self.build(message, uri)
```
intuxicated, line 25: Extracting the `initialize` method logic into the `build` class method and keeping the `initialize` method just for initializing the class.

test/rubygems/test_gem_uri_parser.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'helper'
+require 'uri'
+require 'rubygems/uri_parser'
+
+class TestUriParser < Gem::TestCase
```
intuxicated, line 5: Just adding a few tests for `Gem::UriParser` class.

lib/rubygems/remote_fetcher.rb
```diff
@@ -21,19 +22,24 @@ class Gem::RemoteFetcher
...
 
-      uri = Gem::UriParser.parse_uri(uri)
+      new(message, uri.to_s, original_uri)
+    end
 
-      @original_uri = uri.dup
+    attr_accessor :uri, :original_uri
 
-      uri.password = 'REDACTED' if uri.respond_to?(:password) && uri.password
+    def initialize(message, uri, original_uri = nil)
```
intuxicated, line 38: The `Gem::RemoteFetcher::FetchError` is used in bundler too and if we break the class initialization signature, those tests for checking bundler backward compatibility with old RubyGems will break. 

The `original_uri` use to be generated inside the initializer.

lib/rubygems/remote_fetcher.rb
deivid-rodriguez: It feels a bit overkill to me to do this, specially given that it requires us to change all `FetchError` instantiations, and adapt the new `initialize` signature to still work with `bundler`. How about keep doing thins in `initialize`? If we also expose the original uri inside `Gem::PrintableUri`, the diff could be like this:

```diff
diff --git a/lib/rubygems/remote_fetcher.rb b/lib/rubygems/remote_fetcher.rb
index 8c9c9e6150..b9b1c3ac60 100644
--- a/lib/rubygems/remote_fetcher.rb
+++ b/lib/rubygems/remote_fetcher.rb
@@ -5,6 +5,7 @@
 require_relative 's3_uri_signer'
 require_relative 'uri_formatter'
 require_relative 'uri_parser'
+require_relative 'printable_uri'
 require_relative 'user_interaction'
 
 ##
@@ -25,15 +26,14 @@ class FetchError < Gem::Exception
     attr_accessor :uri, :original_uri
 
     def initialize(message, uri)
-      super message
-
-      uri = Gem::UriParser.parse_uri(uri)
+      uri = Gem::PrintableUri.parse_uri(uri)
 
-      @original_uri = uri.dup
+      message = message.sub(uri.original_password, 'REDACTED') if uri.original_password
 
-      uri.password = 'REDACTED' if uri.respond_to?(:password) && uri.password
+      super message
 
       @uri = uri.to_s
+      @original_uri = uri.original_uri.to_s
     end
 
     def to_s # :nodoc:
```

lib/rubygems/remote_fetcher.rb
deivid-rodriguez: See the previous comment.

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+      @uri.user = 'REDACTED'
+    elsif password?
+      @uri.password = 'REDACTED' if password?
+    end
+
+    @credential_redacted = true
+  end
+
+  def redactable_credential?
+    return false unless parsed_uri?
```
deivid-rodriguez, line 59: All these checks are maybe a bit too defensive? Maybe we can avoid calling `redact_credential` at all unless `parsed_uri?` and remove all these checks assuming we're always dealing with proper URIs. By the way, this makes me think that maybe `parsed_uri?` should be named `valid_uri?`. As far as I understand, if `GemParser.parse_uri` returns the original string, it means it was not given a valid uri, so we cannot further look for passwords or usernames, correct?

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def redact_credential
+    return unless redactable_credential?
+
+    if token?
+      @uri.user = 'REDACTED'
+    elsif oauth_basic?
+      @uri.user = 'REDACTED'
+    elsif password?
+      @uri.password = 'REDACTED' if password?
```
deivid-rodriguez, line 52: The trailing `if` seems unnecessary.

deivid-rodriguez (commented):
It's looking good, I left some comments!

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'uri'
```
simi, line 3: Do we need to vendor https://github.com/ruby/uri now as well?

deivid-rodriguez: I'm not fully sure, but `Gem::UriParser` requires `uri` lazily and things seem fine, so we should probably keep that pattern and move the require to be lazy here too.

We could also avoid using `uri` entirelly in this class by changing `valid_uri?` implementation to `!@uri.is_a?(String)`? If we do that, maybe `parsed_as_valid_uri?` would be a better name? Not sure.

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,70 @@
...
+require_relative 'uri_parser'
+
+class Gem::PrintableUri
+  def self.parse_uri(uri)
+    new(uri).parse_uri
+  end
+
+  def initialize(original_uri)
+    @credential_redacted = false
```
deivid-rodriguez, line 12: What's the purpose of `credential_redacted?`? I don't think it's used anywhere?

lib/rubygems/remote_fetcher.rb
```diff
@@ -21,17 +22,13 @@ class Gem::RemoteFetcher
   class FetchError < Gem::Exception
     ##
     # The URI which was being accessed when the exception happened.
-
```
deivid-rodriguez, line 24: The blank line may have been removed unintentionally.

deivid-rodriguez (commented):
It's looking great. I just observed that the whole looking keeping track of whether we have redacted something or not seems unnecessary at this point?

Although I was thinking about the whole URI validation and defensive progamming. If you add an invalid URI to `Gem.sources`, what kind of error do we get? If the URI gets validated at a moment prior to instantiating `Gem::PrintableUri`, we might be able to remove all `valid_uri?` checks and assume we're always dealing with valid URIs.

lib/rubygems/printable_uri.rb
intuxicated: Done [`9bee1e6` (#4868)](https://github.com/rubygems/rubygems/pull/4868/commits/9bee1e6b0b3531e6412a07337be3611a9250ed54)

lib/rubygems/printable_uri.rb
intuxicated: I have removed it. please take a look at this commit: [`04dc862` (#4868)](https://github.com/rubygems/rubygems/pull/4868/commits/04dc862f80c99c4ca599156c046cea18dd98fad8)

lib/rubygems/remote_fetcher.rb
intuxicated: Done [`6f39f98` (#4868)](https://github.com/rubygems/rubygems/pull/4868/commits/6f39f9802d374700869f836e34a3570db8e9d0a8)

lib/rubygems/printable_uri.rb
deivid-rodriguez: Great!

lib/rubygems/printable_uri.rb
deivid-rodriguez: `redactable_credential?` was unused too, but I meant the whole tracking of whether credentials have been redacted or not. I think `@credential_redacted` seems unnecessary at this point since the whole purpose of this class is to redact credentials? Maybe we can redact credentials at instantiation time and we don't need to track whether we have done that or not at all?

intuxicated: Yeah, make sense. I've removed it [`e085076` (#4868)](https://github.com/rubygems/rubygems/pull/4868/commits/e085076be272436af85df1e1a11991758c6ccb4f)

lib/rubygems/printable_uri.rb
intuxicated: I'm trying to figure out if this could be better:

```diff
@@ -20,10 +20,7 @@ def parse_uri
   end
 
   def valid_uri?
-    @uri.respond_to?(:user) &&
-      @uri.respond_to?(:user=) &&
-      @uri.respond_to?(:password) &&
-      @uri.respond_to?(:password=)
+    [:user, :user=, :password, :password=].all? { |method| @uri.respond_to?(method) }
   end

```

Thoughts?

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,60 @@
...
+  def initialize(original_uri)
+    @original_uri = original_uri
+  end
+
+  def parse_uri
+    @original_uri = Gem::UriParser.parse_uri(@original_uri)
+    @uri = @original_uri.clone
+    redact_credential if valid_uri?
+
+    self
```
deivid-rodriguez, line 19: Maybe have to return `self` here is a sign that we should do the redacting on `initialize`, remove `Gem::PrintableUri.parse_uri` and replace `Gem::PrintableUri.parse_uri` instances with `Gem::PrintableUri.new`. At this point I'm probably just nitpicking...

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,60 @@
...
+    new(uri).parse_uri
+  end
+
+  def initialize(original_uri)
+    @original_uri = original_uri
+  end
+
+  def parse_uri
+    @original_uri = Gem::UriParser.parse_uri(@original_uri)
+    @uri = @original_uri.clone
```
deivid-rodriguez, line 16: I'm not sure whether we want `clone` or `dup` here, but I see that similar code in `Gem::RemoteFetcher` uses `dup... :man_shrugging:.

deivid-rodriguez (approved):
Just left two more small comments that you can feel free to address, but it's very minor and probably just nitpicking from my side, so I'm approving this anyways!

Something else I'd like to investigate is whether we could minimize defensiveness from URI handling by validating URIs at a single place "close to user input" (reading `Gem.sources`), and then remove any further defensive code by assuming we're always dealing with valid URIs from that point.

lib/rubygems/printable_uri.rb
intuxicated: \cc @deivid-rodriguez

test/rubygems/test_gem_printable_uri.rb
```diff
@@ -0,0 +1,80 @@
+require_relative 'helper'
+require 'uri'
```
simi, line 2: Do we need to still require `uri` here?

intuxicated: Good question. I have a couple of tests to make sure we can pass a URI object to this class and it will work. but now I think it's unnecessary since we will be passing a string to this method.

intuxicated: Thanks, @simi. I've removed it.

lib/rubygems/printable_uri.rb
intuxicated: It's refactored to use `dup`.

lib/rubygems/printable_uri.rb
intuxicated: This could be only me being too cautious but I like to keep the `initialize` to be responsible only for initializing the object, meaning no external function call, etc. It's refactored to remove self by [`52ed495` (#4868)](https://github.com/rubygems/rubygems/pull/4868/commits/52ed49563a070cc052b412759c2740767c75ba63). Please take a look, if still what you described reads better to you, let me know.

test/rubygems/test_gem_uri_parser.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'helper'
+require 'uri'
```
simi, line 2: I assume this require can be removed as well.

intuxicated: For this one, I have required the `uri` module intentionally to test the correctness of the `Gem::UriParser.parse_uri`. Do you this test should be removed?

https://github.com/rubygems/rubygems/blob/4ceb2218dea88801ff98263ddada764139c2ac85/test/rubygems/test_gem_uri_parser.rb#L14-L16

simi: That makes sense. @deivid-rodriguez should we add `uri` at least to https://github.com/rubygems/rubygems/blob/master/dev_gems.rb to ensure it is present for tests?

lib/rubygems/printable_uri.rb
deivid-rodriguez: I'm fine with keeping `initialize` to the minimum, my concern was that the instance returned by `Gem::PrintableUri` is not really "usable" unless you redact credentials first, that's why it felt to me like `initialize`'s job. Returning `self` from `Gem::PrintableUri.parse_uri` didn't bother me per-se, but https://github.com/rubygems/rubygems/pull/4868/commits/52ed49563a070cc052b412759c2740767c75ba63 doesn't bother me either :)

I wonder if it would make sense to actually merge this credential redaction functionality and the parsing stuff off `Gem::UriParser` into a `Gem::Uri` class. In that case, you could instantiate a `Gem::Uri` which would be a wrapper around `URI`, which could be useful for some situations, but also redact credentials from it when needed.

Anyways, I think we can try these improvements separately?

test/rubygems/test_gem_uri_parser.rb
deivid-rodriguez: Yeah, it makes sense to me, at least to guarantee that we all use the same version of the `uri` gem for testing.

lib/rubygems/remote_fetcher.rb
```diff
@@ -25,13 +26,10 @@ class FetchError < Gem::Exception
...
     def initialize(message, uri)
-      super message
-
-      uri = Gem::UriParser.parse_uri(uri)
-
       @original_uri = uri.dup
+      uri = Gem::PrintableUri.parse_uri(uri)
 
-      uri.password = 'REDACTED' if uri.respond_to?(:password) && uri.password
+      super(uri.valid_uri? && uri.original_password ? message.sub(uri.original_password, 'REDACTED') : message)
```
deivid-rodriguez, line 32: Sorry I keep coming up with improvements. I think it'd be neat to move this to `Gem::PrintableUri`, to keep all redacting functionality together? Maybe something like

```ruby
def redact_credentials_from(text)
  return text unless valid_uri? && original_password?

  text.sub(uri.original_password, 'REDACTED')
end
```

lib/rubygems/printable_uri.rb
```diff
@@ -0,0 +1,61 @@
...
+    @original_uri.password
+  end
+
+  def to_s
+    @uri.to_s
+  end
+
+  private
+
+  def redact_credential
```
deivid-rodriguez, line 40: This is just some small naming comment, not sure about the grammar behind it, but I've always used `credentials` as plural, even for a single `<user>:<pass>` pair. Using `credential` feels a bit weird to me? Is it just me? Should we use `credential` everywhere?

deivid-rodriguez (commented):
Added a couple more comments that I think could be useful!

lib/rubygems/remote_fetcher.rb
deivid-rodriguez: By doing the above, we no longer need to expose an `original_password` method.

lib/rubygems/printable_uri.rb
deivid-rodriguez: I ended up merging the classes at https://github.com/rubygems/rubygems/pull/4868/commits/9581c2740a37ef12ac0351cb9f166e3d6aa448c1. I think it's simpler that way, let me know what you think @simi and @intuxicated.

test/rubygems/test_gem_uri_parser.rb
deivid-rodriguez: Done at https://github.com/rubygems/rubygems/pull/4868/commits/77e3717e2d17d80d167b5ebec3985db8e23e4394.

lib/rubygems/printable_uri.rb
deivid-rodriguez: After merging classes into `Gem::Uri` at https://github.com/rubygems/rubygems/pull/4868/commits/9581c2740a37ef12ac0351cb9f166e3d6aa448c1, this method is no longer used so no longer a concern.

lib/rubygems/remote_fetcher.rb
deivid-rodriguez: I did this as part of https://github.com/rubygems/rubygems/pull/4868/commits/9581c2740a37ef12ac0351cb9f166e3d6aa448c1.

intuxicated:
👌 @deivid-rodriguez

deivid-rodriguez:
Let's do this! I have some more changes removing some code to make the code less defensive, but I want to ship this first separately since it seems less risky!

deivid-rodriguez (approved):
Thanks @intuxicated!

sergei-fedorov:
Hello, @intuxicated! After updating to 3.2.27 version `gem install bundler --source https://secret-user@gems.host.com` started failing with `Unable to download data from https://REDACTED@gems.host.com/ - bad response Unauthorized 401`. We rolled back to `3.2.26` and now it works. Can there be a bug in this PR? Can it replace creds to `REDACTED` when trying access our gems repo?

intuxicated:
Hi @sergei-fedorov,
Sorry to hear that you had trouble with that. Let me have another look at this.

intuxicated:
@sergei-fedorov I just tested `3.2.27` by installing a gem from a private repository and it's working fine. Are you sure your server works as expected and credentials are correct?

sergei-fedorov:
@intuxicated I guess so, it works on versions less than `3.2.27`.

intuxicated:
@sergei-fedorov in that case, I'll suggest creating an issue so we can look into it. Also, it would be good if you could provide the output of the `gem install` command with additional `--verbose --debug` arguments in the issue.

sergei-fedorov:
@intuxicated here https://github.com/rubygems/rubygems/issues/4906. Thanks for paying attention to our problem.

Issue #4889, opened 2021-08-31
`bundle remove` does not update lockfile


indirect:
### Describe the problem as clearly as you can

When using `bundle add` to add a gem, the lockfile is updated. However, when using `bundle remove` to take a gem out of the Gemfile, the Gemfile.lock is not updated. I would expect `add` and `remove` to by symmetrical, and to both update the lockfile. Instead, I have to run `bundle lock` after running `bundle remove`.

### Post steps to reproduce the problem

```
bundle init
cat Gemfile.lock # no rake
bundle add rake
cat Gemfile.lock # rake is now present
bundle remove rake
cat Gemfile.lock # rake is still present (?) ← this is the error state
bundle lock
cat Gemfile.lock # rake no longer present
```

### Which command did you run?

`bundle remove rake`

### What were you expecting to happen?

I expected both `Gemfile` and `Gemfile.lock` to be updated to no longer contain `rake`.

### What actually happened?

Only `Gemfile` was updated.

### If not included with the output of your command, run `bundle env` and paste the output below

```
Bundler       2.2.21
  Platforms   ruby, universal-darwin-20
Ruby          2.6.3p62 (2019-04-16 revision 67580) [universal.arm64e-darwin20]
  Full Path   /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/ruby
  Config Dir  /Library/Ruby/Site
RubyGems      3.0.3
  Gem Home    /Library/Ruby/Gems/2.6.0
  Gem Path    /Users/andre/.gem/ruby/2.6.0:/Library/Ruby/Gems/2.6.0:/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/gems/2.6.0
  User Home   /Users/andre
  User Path   /Users/andre/.gem/ruby/2.6.0
  Bin Dir     /usr/local/bin
Tools         
  Git         2.30.1
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

Bundler Build Metadata

```
Built At          2021-06-23
Git SHA           0b5db72094
```

Bundler config
```
gem.changelog
  Set for the current user (/Users/andre/.bundle/config): true
gem.ci
  Set for the current user (/Users/andre/.bundle/config): "github"
gem.coc
  Set for the current user (/Users/andre/.bundle/config): true
gem.mit
  Set for the current user (/Users/andre/.bundle/config): true
gem.rubocop
[...]

PR #4891, opened 2021-09-01
Deprecate `--install` flag to `bundle remove` and trigger install by default

ceritium:
Closes https://github.com/rubygems/rubygems/issues/4889

`bundle remove` does not update lockfile.

- Trigger `install` command by default on `bundle remove`
- Remove option `--install` from the cli

### DIFF
--- a/bundler/spec/commands/remove_spec.rb
+++ b/bundler/spec/commands/remove_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe "bundle remove" do
     end
   end
 
-  context "when --install flag is specified" do
+  context "when --install flag is specified", :bundler => "< 3" do
     it "removes gems from .bundle" do
       gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
@@ -40,6 +40,7 @@ RSpec.describe "bundle remove" do
         bundle "remove rack"
 
         expect(out).to include("rack was removed.")
+        expect(the_bundle).to_not include_gems "rack"
         gemfile_should_be <<-G
           source "#{file_uri_for(gem_repo1)}"
         G
--- a/bundler/spec/support/builders.rb
+++ b/bundler/spec/support/builders.rb
@@ -35,6 +35,11 @@ module Spec
       build_repo gem_repo1 do
         FileUtils.cp rake_path, "#{gem_repo1}/gems/"
 
+        build_gem "coffee-script-source"
+        build_gem "git"
+        build_gem "puma"
+        build_gem "minitest"
+
         build_gem "rack", %w[0.9.1 1.0.0] do |s|
           s.executables = "rackup"
           s.post_install_message = "Rack's post install message"
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -184,6 +184,7 @@ module Bundler
     method_option "install", :type => :boolean, :banner =>
       "Runs 'bundle install' after removing the gems from the Gemfile"
     def remove(*gems)
+      SharedHelpers.major_deprecation(2, "The `--install` flag has been deprecated. `bundle install` is triggered by default.") if ARGV.include?("--install")
       require_relative "cli/remove"
       Remove.new(gems, options).run
     end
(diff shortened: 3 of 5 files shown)


bundler/spec/support/builders.rb
```diff
@@ -35,6 +35,11 @@ def build_repo1
       build_repo gem_repo1 do
         FileUtils.cp rake_path, "#{gem_repo1}/gems/"
 
+        build_gem "coffee-script-source"
+        build_gem "git"
+        build_gem "puma"
+        build_gem "minitest"
```
ceritium, line 41: I added those gems to make the "remove specs" pass with minimal changes. 

We can replace them with existing gems like `foo`, `thin`, and so on. What do you think?

deivid-rodriguez (commented):
Nice job! My only concern here, is that to be nice citizens we should deprecate the `--install` flag instead of abruptly removing it.

bundler/spec/commands/remove_spec.rb
```diff
@@ -13,21 +13,6 @@
...
-
-        gem "rack"
-      G
-
-      bundle "remove rack --install"
-
-      expect(out).to include("rack was removed.")
-      expect(the_bundle).to_not include_gems "rack"
-    end
-  end
```
deivid-rodriguez, line 29: We can keep this spec, but run it only on `:bundler => "< 3"`.

bundler/spec/remove/install_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    G
+  end
+
+  describe "with --install", :bundler => 2 do
+    it "shows a deprecation warning" do
+      bundle "remove rack --install"
+
+      expect(err).to include "[DEPRECATED] The `--install` flag has been deprecated. `bundle install` is triggered by default."
+    end
+  end
```
deivid-rodriguez, line 17: Can you move this spec to `spec/other/major_deprecation_spec.rb`. I like to keep all deprecation specs together.

deivid-rodriguez (commented):
Just a couple of minor comments, but other than that looks good!

ceritium:
@deivid-rodriguez, do you think it makes sense to add a pending test for bundler 3 like [`bundle show --outdated`](https://github.com/rubygems/rubygems/blob/master/bundler/spec/other/major_deprecation_spec.rb#L623)

bundler/spec/commands/remove_spec.rb
```diff
@@ -13,11 +13,10 @@
     end
   end
 
-  context "when --install flag is specified" do
+  context "when --install flag is specified", :bundler => "<3" do
     it "removes gems from .bundle" do
       gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
-
```
deivid-rodriguez, line 20: Whitespace nitpick, let's leave it as it was. It gave me an idea of a rubocop cop, like `Bundler/BlankLineAfterGlobalSource` :)

deivid-rodriguez (commented):
One final nitpick :stuck_out_tongue_closed_eyes:

deivid-rodriguez:
> @deivid-rodriguez, do you think it makes sense to add a pending test for bundler 3 like bundle show --outdated

Sure, yeah!

deivid-rodriguez (approved):
:+1:

ceritium:
@deivid-rodriguez  Do you need me to squash the commits or can you do it on merge?

deivid-rodriguez:
Actually, if you can do that it'd be awesome! I suspect our backport scripts require explicit merge commits, so we have "squash and merge" disabled.

bundler/spec/commands/remove_spec.rb
```diff
@@ -13,7 +13,7 @@
     end
   end
 
-  context "when --install flag is specified" do
+  context "when --install flag is specified", :bundler => "<3" do
```
deivid-rodriguez, line 16: Just notice another small inconsistency of missing a space between "<" and "3". You fix it while squashing if you want, otherwise it's no big deal!

ceritium: No problem

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks!

ceritium:
@deivid-rodriguez thanks to you. I learn a lot about how the test suite works inside bundler.

PR #4899, opened 2021-09-05
Fix triple spacing when generating lockfile

### LINKED ISSUE #3326: Lockfile generates 3 spaces
https://github.com/bundler/bundler/blob/d44d803357506895555ff97f73e60d593820a0de/lib/bundler/lockfile_generator.rb#L89 seems to generate lines with 3 spaces. 

E.g. The `BUNDLED WITH` section has 3 spaces before `2.0.1`

```
GEM
  remote: https://rubygems.org/
  specs:
    diff-lcs (1.3)
    rspec (3.8.0)
      rspec-core (~> 3.8.0)
      rspec-expectations (~> 3.8.0)
      rspec-mocks (~> 3.8.0)
    rspec-core (3.8.0)
      rspec-support (~> 3.8.0)
    rspec-expectations (3.8.2)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-mocks (3.8.0)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.8.0)
    rspec-support (3.8.0)

PLATFORMS
  ruby

DEPENDENCIES
  rspec

BUNDLED WITH
   2.0.1
```

Not sure if this is intentional or not ?

jmlntw:

The `BUNDLED WITH` section of the Gemfile.lock have an extra space before each line of content.

Remove this extra space from line 89 and close #3326.

There was a pull request on the old repository (rubygems/bundler#7236), but it was not merged to this repository. So I re-create this pull request.

### DIFF
--- a/bundler/lib/bundler/lockfile_generator.rb
+++ b/bundler/lib/bundler/lockfile_generator.rb
@@ -86,7 +86,7 @@ module Bundler
           out << "  #{key}: #{val}\n"
         end
       when String
-        out << "   #{value}\n"
+        out << "  #{value}\n"
       else
         raise ArgumentError, "#{value.inspect} can't be serialized in a lockfile"
       end
--- a/bundler/spec/bundler/definition_spec.rb
+++ b/bundler/spec/bundler/definition_spec.rb
@@ -78,7 +78,7 @@ RSpec.describe Bundler::Definition do
           foo!
 
         BUNDLED WITH
-           #{Bundler::VERSION}
+          #{Bundler::VERSION}
       G
     end
 
@@ -115,7 +115,7 @@ RSpec.describe Bundler::Definition do
           foo!
 
         BUNDLED WITH
-           #{Bundler::VERSION}
+          #{Bundler::VERSION}
       G
     end
 
@@ -143,7 +143,7 @@ RSpec.describe Bundler::Definition do
           only_java
 
         BUNDLED WITH
-           #{Bundler::VERSION}
+          #{Bundler::VERSION}
       G
     end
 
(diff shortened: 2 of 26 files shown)


simi:
@jmlntw :information_source: some specs do rely on exact lockfile format and are currently failing due to missing space.

jmlntw:
Specs are updated now.

simi:
@deivid-rodriguez any opinion on this?

deivid-rodriguez:
I was hoping to do this after we have bundler version locking in place, because otherwise this is going to generate undesired whitespace lockfile churn depending on which version of bundler generates the lockfile.

tomstuart:
Would you be willing to merge this now? I can rebase & reopen if that’d help.

deivid-rodriguez:
I think it should be ok, yeah.

deivid-rodriguez:
But let me ask for other maintainers opinions first. Given that we mostly have bundler version locking in place, it seems not too disruptive to me.

maco:
@deivid-rodriguez so, a year and a half later, is bundler version locking done and rebasing this possible?

deivid-rodriguez:
Hello! I don't have a clear opinion yet, sorry.

This is a very minor issue (some people find it annoying to see three spaces here), although the bad consequences of fixing it are also very minor (some people will find annoying that a space change shows up in the lockfile diffs).

We do have version locking in place now, but only for `bundle install`. I think that may be enough to avoid too many flip flops here, since once a lockfile is upgraded, everybody should get the new version and two spaces will always be used and not reverted back to three spaces.

So this may be ok, but I'm not fully convinced it needs fixing.

simi:
Is there any simple way to introduce this only to new Gemfiles? :thinking:

deivid-rodriguez:
We could, but in my opinion it would add unnecessary complexity for little benefit, and the potential issue would still be there. If someone uses that lockfile with an old bundler version that does not have "trampolining", the spaces would be reverted back to 3.

PR #4910, opened 2021-09-11
Fix `wordy` method in `SourceFetchProblem` changing the password of source

### LINKED ISSUE #4908: The wordy function in SourceFetchProblem will change the password of source

My current problem is when I was reviewing the code, I found that in the **errors.rb** file, the **wordy** function of class**SourceFetchProblem** is used to anonymizes **uri.password**. However, the value of **source** is assigned to @source during initialization. When the value of **@source.uri.password** is changed to **REDACTED**, the password of the **source** is also changed.
I know what's in the current code doesn't have any practical impact, but I think this usage leaves a risk for future evolution. For developers who are not familiar with the content, if the password in the **source** is processed after the **sourceFetchProblem.wordy** is invoked, for example, password verification, and the password is reset to REDACTED, security risks may exist.
Therefore, I suggest modifying the current anonymization implementation and using **@source = Marshal.load(Marshal.dump(source))** to assign values to reduce risks.

This is my test code:
```
  require 'rubygems/remote_fetcher'
  
  source = Gem::Source.new 'https://username:secret@gemsource.com'
  uri = 'https://user:secret@gemsource.org'
  error = Gem::RemoteFetcher::FetchError.new('There was an error fetching', uri)
  print("error：", error, "\n")
  print("用户密码(SourceFetchProblem处理前):", source.uri.password, "\n")
  sf = Gem::SourceFetchProblem.new source, error
  print("用户密码(SourceFetchProblem处理后):", source.uri.password, "\n")
  print("error：", error, "\n")
  print("sf.error：", error, "\n")
  print("wordy输出：", sf.wordy, "\n")
  print("source.uri(SourceFetchProblem处理后):", source.uri, "\n")
```

image

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X ] Other

Huangxiaodui:

Just a potential issue, but some error code was modifiying the original source which was not expected.

Do not change the password of the input parameter source during anonymization, instead create a copy of the URL for anonymization.

Fixes #4908.

### DIFF
--- a/lib/rubygems/errors.rb
+++ b/lib/rubygems/errors.rb
@@ -171,8 +171,7 @@ module Gem
     # An English description of the error.
 
     def wordy
-      @source.uri.password = 'REDACTED' unless @source.uri.password.nil?
-      "Unable to download data from #{@source.uri} - #{@error.message}"
+      "Unable to download data from #{Gem::Uri.new(@source.uri).redacted} - #{@error.message}"
     end
 
     ##
--- a/test/rubygems/test_gem_source_fetch_problem.rb
+++ b/test/rubygems/test_gem_source_fetch_problem.rb
@@ -23,4 +23,14 @@ class TestGemSourceFetchProblem < Gem::TestCase
 
     refute_match sf.wordy, 'secret'
   end
+
+  def test_source_password_no_redacted
+    source = Gem::Source.new 'https://username:secret@gemsource.com'
+    error  = RuntimeError.new 'test'
+
+    sf = Gem::SourceFetchProblem.new source, error
+    sf.wordy
+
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
@Huangxiaodui I pushed what I think it's a better fix, and also some fixes to your test which had a duplicated `end` making it not valid `ruby`.

Huangxiaodui:
@deivid-rodriguez Thanks for the modification, this is a much better solution.

deivid-rodriguez:
Great, thanks for your contribution!

Issue #3118, opened 2020-02-01
Include value of `Gem.source_date_epoch` in metadata of built gems.


duckinator:
We now (via #2882 and #3088) have builds that are reproducible by default _if you know the value of `Gem.source_date_epoch`/`Gem.source_date_epoch_string`_.

If built gems include the value of `Gem.source_date_epoch_string` in their metadata, we open up the possibility of making automated tools that could detect some instances of things like the [backdoored version of the `bootstrap-sass` gem](https://snyk.io/blog/malicious-remote-code-execution-backdoor-discovered-in-the-popular-bootstrap-sass-ruby-gem/) that was released in March 2019.

I'm not sure, however, where this metadata should be included.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [x] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

rubyFeedback:
This sounds really great - would be epic if we could get the whole of ruby (MRI) and default gem to fulfil reproducibility criteria e. g. debian's reproducible packages project ( https://wiki.debian.org/ReproducibleBuilds#Even_more ).

3.0 release will be on xmas so about ~10 months still left.

deivid-rodriguez:
@duckinator It seems that as long as the `Gem::Specification#date` is not explicitly set by the gem author, we set it to the date corresponding to `Gem.source_date_epoch`, so this should be already available by converting the date in the specification of the built gem back to a timestamp?

duckinator:
@deivid-rodriguez if that's the case, yeah, it should work. And it already being there is fantastic. I'll try to make a script that demonstrates using this to try to reproduce a build based on a gem file relatively soon. (I'd try it now, but it's after midnight and I have things to do tomorrow lol)

duckinator:
<details>
<summary>Rough first attempt at such a script, which doesn't seem to work yet</summary>

```ruby
#!/usr/bin/env ruby

# Rebuild a gem from source using `rake build`, with SOURCE_DATE_EPOCH
# set to a timestamp derived from the value of the `date` field in the gem
# metadata.

require 'date'
require 'digest'
require 'rubygems/package'

DATE_FORMAT = '%Y-%m-%d %H:%M:%S.%N Z'

def print_sha256(file)
  hash = Digest::SHA256.hexdigest(File.read(file))
[...]

PR #4913, opened 2021-09-12
Add "gem rebuild" command.

duckinator:

If a gem implements the requirements for reproducible builds, and I have a copy of the source, I want to be able to verify that the released version was built from that source.

I implemented the `gem rebuild` command. See the discussion in #3118 for details.

```
$ ls -a
./  ../
$ git clone https://github.com/duckinator/okay.git
Cloning into 'okay'...
remote: Enumerating objects: 483, done.
remote: Counting objects: 100% (169/169), done.
remote: Compressing objects: 100% (86/86), done.
remote: Total 483 (delta 87), reused 134 (delta 83), pack-reused 314
Receiving objects: 100% (483/483), 77.50 KiB | 3.52 MiB/s, done.
Resolving deltas: 100% (213/213), done.
$ gem rebuild -C ./okay okay 12.0.1
Fetching okay-12.0.1.gem
Downloaded okay version 12.0.1 as /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem.
WARNING:  open-ended dependency on cacert (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
  Successfully built RubyGem
  Name: okay
  Version: 12.0.1
  File: okay-12.0.1.gem

Built at: 2021-09-11 20:01:16 EDT (1631404876)
Original build saved to:   /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem
Reproduced build saved to: /usr/home/puppy/dev/rebuild-test/rebuild/new/okay-12.0.1.gem
Working directory: ./okay

Hash comparison:
  558bf6a6343c984f11021aff41c1a63ec07911c0658a408af6e4308b3e7764ea      /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem
  558bf6a6343c984f11021aff41c1a63ec07911c0658a408af6e4308b3e7764ea      /usr/home/puppy/dev/rebuild-test/rebuild/new/okay-12.0.1.gem

SUCCESS - original and rebuild hashes matched
$
```

Failed example:

```
$ ls -a
./  ../
$ git clone https://github.com/duckinator/okay.git
Cloning into 'okay'...
remote: Enumerating objects: 483, done.
remote: Counting objects: 100% (169/169), done.
remote: Compressing objects: 100% (86/86), done.
remote: Total 483 (delta 87), reused 134 (delta 83), pack-reused 314
Receiving objects: 100% (483/483), 77.50 KiB | 3.10 MiB/s, done.
Resolving deltas: 100% (213/213), done.
$ echo '#hi' >> okay/lib/okay.rb 
$ tail -n 2 okay/lib/okay.rb 
end
#hi
$ gem rebuild -C ./okay okay 12.0.1
Fetching okay-12.0.1.gem
Downloaded okay version 12.0.1 as /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem.
WARNING:  open-ended dependency on cacert (>= 0) is not recommended
  use a bounded requirement, such as '~> x.y'
  Successfully built RubyGem
  Name: okay
  Version: 12.0.1
  File: okay-12.0.1.gem

Built at: 2021-09-11 20:01:16 EDT (1631404876)
Original build saved to:   /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem
Reproduced build saved to: /usr/home/puppy/dev/rebuild-test/rebuild/new/okay-12.0.1.gem
Working directory: ./okay

Hash comparison:
  558bf6a6343c984f11021aff41c1a63ec07911c0658a408af6e4308b3e7764ea      /usr/home/puppy/dev/rebuild-test/rebuild/old/okay-12.0.1.gem
  f803f93864366942b9aed3e5f7d806c1bffccae5d59eff74f68dbd59d6811426      /usr/home/puppy/dev/rebuild-test/rebuild/new/okay-12.0.1.gem

FAILURE - original and rebuild hashes did not match
$
```

### DIFF
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -11,6 +11,10 @@ class Gem::Commands::RebuildCommand < Gem::Command
   def initialize
     super "rebuild", "Attempt to reproduce a build of a gem."
 
+    add_option "--diff", "If the files don't match, compare them using diffoscope." do |_value, options|
+      options[:diff] = true
+    end
+
     add_option "--force", "Skip validation of the spec." do |_value, options|
       options[:force] = true
     end
@@ -150,6 +154,14 @@ An example of reproducing a gem build:
       say "SUCCESS - original and rebuild hashes matched"
     else
       say "FAILURE - original and rebuild hashes did not match"
+
+      if options[:diff]
+        say
+        if system("diffoscope", old_file, new_file).nil?
+          alert_error "error: could not find `diffoscope` executable"
+        end
+      end
+
       terminate_interaction 1
     end
   end
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -3,6 +3,7 @@
 require "date"
 require "digest"
 require "fileutils"
+require "tmpdir"
 require_relative "../package"
 
 class Gem::Commands::RebuildCommand < Gem::Command
@@ -167,13 +168,9 @@ An example of reproducing a gem build:
   end
 
   def prep_dirs
-    rebuild_dir = File.expand_path("rebuild")
+    rebuild_dir = Dir.mktmpdir("gem_rebuild")
     old_dir = File.join(rebuild_dir, "old")
     new_dir = File.join(rebuild_dir, "new")
-
-    if File.directory?(rebuild_dir)
-      FileUtils.remove_dir(rebuild_dir)
-    end
 
     FileUtils.mkdir_p(old_dir)
     FileUtils.mkdir_p(new_dir)
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -4,9 +4,12 @@ require "date"
 require "digest"
 require "fileutils"
 require "tmpdir"
+require_relative "../gemspec_helpers"
 require_relative "../package"
 
 class Gem::Commands::RebuildCommand < Gem::Command
+  include Gem::GemspecHelpers
+
   DATE_FORMAT = "%Y-%m-%d %H:%M:%S.%N Z"
 
   def initialize
@@ -221,17 +224,6 @@ Please install RubyGems v#{rg_version} and try again.
     yield
   ensure
     ENV["SOURCE_DATE_EPOCH"] = old_sde
-  end
-
-  def find_gemspec(glob = "*.gemspec")
-    gemspecs = Dir.glob(glob).sort
-
-    if gemspecs.size > 1
-      alert_error "Multiple gemspecs found: #{gemspecs}, please specify one"
-      terminate_interaction(1)
-    end
-
-    gemspecs.first
   end
 
   def error_message(gem_name)
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -154,6 +154,9 @@ Please install RubyGems v#{rg_version} and try again.
         if system("diffoscope", old_file, new_file).nil?
           alert_error "error: could not find `diffoscope` executable"
         end
+      else
+        say
+        say "Pass --diff for more details (requires diffoscope to be installed)."
       end
 
       terminate_interaction 1
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -120,8 +120,10 @@ An example of reproducing a gem build:
 
   def get_timestamp(file)
     mtime = nil
-    Gem::Package::TarReader.new(File.open(file)) do |tar|
-      mtime = tar.seek("metadata.gz") {|f| f.header.mtime }
+    File.open(file, "rb") do |f|
+      Gem::Package::TarReader.new(f) do |tar|
+        mtime = tar.seek("metadata.gz") {|tf| tf.header.mtime }
+      end
     end
 
     mtime
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -55,35 +55,12 @@ This command assumes the gemspec can be built with the `gem build` command.
 If you use either `gem build` or `rake build`/`rake release` to build/release
 a gem, it is a potential candidate.
 
-If the gem includes top-level files that change frequently (e.g. Gemfile.lock),
-it may require more effort to reproduce a build. For example, it might require
-more precisely matched versions of Ruby, RubyGems, or Bundler to be used.
+You will likely need to match the RubyGems version used, since this is
+included in the Gem metadata.
 
-An example of reproducing a gem build:
-
-  $ pwd
-  /usr/home/puppy/test/rebuild
-  $ ls -a
-  ./  ../
-  $ git clone --branch v12.0.2 https://github.com/duckinator/okay.git
-  $ gem rebuild -C ./okay --gemspec okay.gemspec okay 12.0.2
-  Fetching okay-12.0.2.gem
-  Downloaded okay version 12.0.2 as /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem.
-    Successfully built RubyGem
-    Name: okay
-    Version: 12.0.2
-    File: okay-12.0.2.gem
-
-  Built at: 2023-08-31 21:39:02 EDT (1693532342)
-  Original build saved to:   /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
-  Reproduced build saved to: /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
-  Working directory: ./okay
-
-  Hash comparison:
-    38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
-    38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
-
-  SUCCESS - original and rebuild hashes matched
+If the gem includes lockfiles (e.g. Gemfile.lock) and similar, it will require
+more effort to reproduce a build. For example, it might require more precisely
+matched versions of Ruby and/or Bundler to be used.
     EOF
   end
 
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -83,6 +83,22 @@ matched versions of Ruby and/or Bundler to be used.
       download_gem(gem_name, gem_version, old_file)
     end
 
+    rg_version = rubygems_version(old_file)
+    unless rg_version == Gem::VERSION
+      alert_error <<-EOF
+You need to use the same RubyGems version #{gem_name} v#{gem_version} was built with.
+
+#{gem_name} v#{gem_version} was built using RubyGems v#{rg_version}.
+Gem files include the version of RubyGems used to build them.
+This means in order to reproduce #{gem_filename}, you must also use RubyGems v#{rg_version}.
+
+You're using RubyGems v#{Gem::VERSION}.
+
+Please install RubyGems v#{rg_version} and try again.
+      EOF
+      terminate_interaction 1
+    end
+
     source_date_epoch = get_timestamp(old_file).to_s
 
     if build_path = options[:build_path]
@@ -246,5 +262,9 @@ matched versions of Ruby and/or Bundler to be used.
     FileUtils.move(download_path, old_file)
 
     say "Downloaded #{gem_name} version #{gem_version} as #{old_file}."
+  end
+
+  def rubygems_version(gem_file)
+    Gem::Package.new(gem_file).spec.rubygems_version
   end
 end
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -97,12 +97,12 @@ matched versions of Ruby and/or Bundler to be used.
   private
 
   def sha256(file)
-    Digest::SHA256.hexdigest(File.read(file))
+    Digest::SHA256.hexdigest(Gem.read_binary(file))
   end
 
   def get_timestamp(file)
     mtime = nil
-    File.open(file, "rb") do |f|
+    File.open(file, Gem.binary_mode) do |f|
       Gem::Package::TarReader.new(f) do |tar|
         mtime = tar.seek("metadata.gz") {|tf| tf.header.mtime }
       end
--- a/lib/rubygems/commands/rebuild_command.rb
+++ b/lib/rubygems/commands/rebuild_command.rb
@@ -1,0 +1,262 @@
+# frozen_string_literal: true
+
+require "date"
+require "digest"
+require "fileutils"
+require_relative "../package"
+
+class Gem::Commands::RebuildCommand < Gem::Command
+  DATE_FORMAT = "%Y-%m-%d %H:%M:%S.%N Z"
+
+  def initialize
+    super "rebuild", "Attempt to reproduce a build of a gem."
+
+    add_option "--force", "Skip validation of the spec." do |_value, options|
+      options[:force] = true
+    end
+
+    add_option "--strict", "Consider warnings as errors when validating the spec." do |_value, options|
+      options[:strict] = true
+    end
+
+    add_option "--source GEM_SOURCE", "Specify the source to download the gem from." do |value, options|
+      options[:source] = value
+    end
+
+    add_option "--original GEM_FILE", "Specify a local file to compare against (instead of downloading it)." do |value, options|
+      options[:original_gem_file] = value
+    end
+
+    add_option "--gemspec GEMSPEC_FILE", "Specify the name of the gemspec file." do |value, options|
+      options[:gemspec_file] = value
+    end
+
+    add_option "-C PATH", "Run as if gem build was started in <PATH> instead of the current working directory." do |value, options|
+      options[:build_path] = value
+    end
+  end
+
+  def arguments # :nodoc:
+    "GEM_NAME      gem name on gem server\n" \
+    "GEM_VERSION   gem version you are attempting to rebuild"
+  end
+
+  def description # :nodoc:
+    <<-EOF
+The rebuild command allows you to (attempt to) reproduce a build of a gem
+from a ruby gemspec.
+
+This command assumes the gemspec can be built with the `gem build` command.
+If you use either `gem build` or `rake build`/`rake release` to build/release
+a gem, it is a potential candidate.
+
+If the gem includes top-level files that change frequently (e.g. Gemfile.lock),
+it may require more effort to reproduce a build. For example, it might require
+more precisely matched versions of Ruby, RubyGems, or Bundler to be used.
+
+An example of reproducing a gem build:
+
+  $ pwd
+  /usr/home/puppy/test/rebuild
+  $ ls -a
+  ./  ../
+  $ git clone --branch v12.0.2 https://github.com/duckinator/okay.git
+  $ gem rebuild -C ./okay --gemspec okay.gemspec okay 12.0.2
+  Fetching okay-12.0.2.gem
+  Downloaded okay version 12.0.2 as /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem.
+    Successfully built RubyGem
+    Name: okay
+    Version: 12.0.2
+    File: okay-12.0.2.gem
+
+  Built at: 2023-08-31 21:39:02 EDT (1693532342)
+  Original build saved to:   /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
+  Reproduced build saved to: /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
+  Working directory: ./okay
+
+  Hash comparison:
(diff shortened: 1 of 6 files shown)


duckinator:
TODO:

- [ ] add tests
- [ ] update the example in the help text (i changed the output format since i wrote it)

matiaskorhonen:
Does this work with cryptographically signed gems? (see https://guides.rubygems.org/security/)

duckinator:
@matiaskorhonen I haven't put in work to specifically support that, so I'd be surprised if it worked. I'm not sure what would be needed to support it.

duckinator:
Sorry for the slow movement on this &mdash; I've been struggling with basically-perpetual brain fog the last two months or so, and it's making it really difficult to actually work on the RubyGems codebase.

I did finally get this to the point that it has a test which fails _when expected to_ instead of passing unexpectedly, which is nice.

At the moment, the test fails because it tries to download something not on RubyGems.org. I tried using the `--original` flag to specify a local file to compare against but it just gave a different error instead (see the TODO near the end of `test/rubygems/test_gem_commands_rebuild_command.rb`). That's the next thing to figure out, I think.

duckinator:
Rebased off master.

duckinator:
Rebased off master (d4325048dce3ae4d5d63a3efa7a776a1c2d4d2eb).

I still need to resolve this: https://github.com/rubygems/rubygems/blob/fbe6fcc75ed61278ccca55fabbfd2edbf4d1eb29/test/rubygems/test_gem_commands_rebuild_command.rb#L114-L115

duckinator:
Rebased off master (c44fe8e).

duckinator:
Resolved linter errors.

duckinator:
I'm having trouble figuring out why the tests are failing, but the `gem rebuild` command itself is working:

```
~/test/rebuild$ ls -a
./  ../
~/test/rebuild$ git clone --branch v12.0.2 https://github.com/duckinator/okay.git
Cloning into 'okay'...
remote: Enumerating objects: 512, done.
remote: Counting objects: 100% (43/43), done.
remote: Compressing objects: 100% (43/43), done.
remote: Total 512 (delta 0), reused 0 (delta 0), pack-reused 469
Receiving objects: 100% (512/512), 73.46 KiB | 2.62 MiB/s, done.
Resolving deltas: 100% (240/240), done.
~/test/rebuild$ gem rebuild -C ./okay --gemspec okay.gemspec okay 12.0.2
Fetching okay-12.0.2.gem
Downloaded okay version 12.0.2 as /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem.
WARNING:  open-ended dependency on cacert (>= 0) is not recommended
  use a bounded requirement, such as "~> x.y"
  Successfully built RubyGem
  Name: okay
  Version: 12.0.2
  File: okay-12.0.2.gem

Built at: 2023-08-31 21:39:02 EDT (1693532342)
Original build saved to:   /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
Reproduced build saved to: /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
Working directory: ./okay

Hash comparison:
  38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
  38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem

SUCCESS - original and rebuild hashes matched
~/test/rebuild$
```

duckinator:
Problems that have been resolved:

- the test does not do `sleep 1` to ensure time differences anymore
- there is no longer two almost-identical complex functions
- the test gives the `gem rebase` command the original gem file, now

So at this point I think the _only_ problem is that the test doesn't actually run.

The fact that it's eating stderr is making this a lot more difficult to figure out.

deivid-rodriguez:
Thanks for the hard work @duckinator. Let me know if you want me to have a look at the failing tests!

duckinator:
@deivid-rodriguez if you could take a look, I'd definitely appreciate it. They both fail at the same spot (the `rebuild_gem_file = util_test_rebuild_gem(...)` line), but I have no idea why.

duckinator:
I _finally_ figured out what the problem is. (Or at least part of it, anyway.)

My code does `find_gemspec("#{gem_name}.spec")` when no gemspec is provided.

It turns out that `util_spec` creates `#{gem_name}-#{gem_version}.spec`.

So I can just specify the exact path and it should get past that.

(I don't want to even think about how much time I've spent debugging this...)

duckinator:
I am extremely excited because after TWO AND A HALF FREAKING YEARS, if CI passes, then this is FINALLY ready for a proper review.

deivid-rodriguez:
@duckinator So sorry I forgot to look into this, I'm glad that you managed to fix the remaining issues! 💪

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+    say
+    say "Hash comparison:"
+    say "  #{old_hash}\t#{old_file}"
+    say "  #{new_hash}\t#{new_file}"
+    say
+
+    if old_hash == new_hash
+      say "SUCCESS - original and rebuild hashes matched"
+    else
+      say "FAILURE - original and rebuild hashes did not match"
```
segiddins, line 150: should we add an option to maybe run [`diffoscope`](https://diffoscope.org) if present against the two files, to try and help reveal the differences between the two files?

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+    compare(source_date_epoch, old_file, new_file)
+  end
+
+  private
+
+  def sha256(file)
+    Digest::SHA256.hexdigest(File.read(file))
+  end
+
+  def get_timestamp(file)
```
segiddins, line 121: note to self: this is a sign we should embed the source date epoch used into the metadata of the built gemspec. additionally, ruby/rubygems versions would be helpful

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+    mtime = nil
+    Gem::Package::TarReader.new(File.open(file)) do |tar|
```
segiddins, line 123: I think this will leak the files, use the block form of `File.open` with `"rb"` ?

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+      terminate_interaction 1
+    end
+  end
+
+  def prep_dirs
+    rebuild_dir = File.expand_path("rebuild")
+    old_dir = File.join(rebuild_dir, "old")
+    new_dir = File.join(rebuild_dir, "new")
+
+    if File.directory?(rebuild_dir)
```
segiddins, line 160: this will cause errors if `rebuild` is a file and not a directory I think

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+  def with_source_date_epoch(source_date_epoch)
+    old_sde = ENV["SOURCE_DATE_EPOCH"]
+    ENV["SOURCE_DATE_EPOCH"] = source_date_epoch.to_s
+
+    yield
+  ensure
+    ENV["SOURCE_DATE_EPOCH"] = old_sde
+  end
+
+  def find_gemspec(glob = "*.gemspec")
```
segiddins, line 220: can this share an implementation with the `build` command?

lib/rubygems/commands/rebuild_command.rb
deivid-rodriguez: This seems like a good idea!

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+The rebuild command allows you to (attempt to) reproduce a build of a gem
+from a ruby gemspec.
+
+This command assumes the gemspec can be built with the `gem build` command.
+If you use either `gem build` or `rake build`/`rake release` to build/release
+a gem, it is a potential candidate.
+
+If the gem includes top-level files that change frequently (e.g. Gemfile.lock),
+it may require more effort to reproduce a build. For example, it might require
+more precisely matched versions of Ruby, RubyGems, or Bundler to be used.
```
deivid-rodriguez, line 55: I think, since the running RubyGems version is [included in metadata](https://github.com/rubygems/rubygems/blob/a7aa64a918a590efe0e8c4da0eba189a8aa5140d/lib/rubygems/specification.rb#L153), I think an exact RubyGems version match is always necessary at the moment?

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+  Original build saved to:   /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
+  Reproduced build saved to: /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
+  Working directory: ./okay
+
+  Hash comparison:
+    38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/old/okay-12.0.2.gem
+    38a8bd78ce10bc19189ead0b56fa490d03788a2f926a6481b2f1f5d5fa5ab75b      /usr/home/puppy/test/rebuild/rebuild/new/okay-12.0.2.gem
+
+  SUCCESS - original and rebuild hashes matched
+    EOF
```
deivid-rodriguez, line 82: Do we want to include the full output of the current implementation in the command documentation/description? I think it's likely to get out of date and become confusing?

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+    date = Time.at(source_date_epoch.to_i).strftime("%F %T %Z")
+
+    old_hash = sha256(old_file)
+    new_hash = sha256(new_file)
+
+    say
+    say "Built at: #{date} (#{source_date_epoch})"
+    say "Original build saved to:   #{old_file}"
+    say "Reproduced build saved to: #{new_file}"
+    say "Working directory: #{options[:build_path] || Dir.pwd}"
```
deivid-rodriguez, line 140: Should default to building things in `/tmp`? We print paths for inspection and that way we don't leave artifacts around in the working directory. Or maybe current directory's `tmp/` if it exists?

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,262 @@
...
+  def error_message(gem_name)
+    if gem_name
+      "Couldn't find a gemspec file matching '#{gem_name}' in #{Dir.pwd}"
+    else
+      "Couldn't find a gemspec file in #{Dir.pwd}"
+    end
+  end
+
+  def download_gem(gem_name, gem_version, old_file)
+    # This code was based loosely off the `gem fetch` command.
```
deivid-rodriguez, line 240: What's different with what `gem fetch` does? Could we extract some common method for both, or directly reuse `FetchCommand` here?

deivid-rodriguez (commented):
I like this!

I worry that having to use the exact same RubyGems version makes it harder to reproduce builds? If only that metadata field is different, things are safe, right? We could consider build with `rubygems_version` set to the one that originally built the gem, and if they match then, inform about that?

lib/rubygems/commands/rebuild_command.rb
segiddins: yes -- one thing we could do is error out ahead of time if the wrong gem version is used, or add a way to override the rubygems version serialized into the gemspec, the same way we allow overriding the source date epoch?

deivid-rodriguez: Yeah. I'd go with the latter, seems like it could make this more useful. Also I see some other differences in metadata, like some trailing spaces differences caused by the rubygems version divergence. We could also normalize those.

duckinator:
> I like this!

I'm glad. I've put a lot of effort into it. 😂 

> I worry that having to use the exact same RubyGems version makes it harder to reproduce builds?

It definitely does.

> If only that metadata field is different, things are safe, right?

Yes.

> We could consider build with rubygems_version set to the one that originally built the gem, and if they match then, inform about that?

Yeah, this is definitely an option worth considering. And if this is included in the metadata (as @segiddins suggested), we could check RubyGems version before trying the build and just print a message like "It is recommended to use RubyGems v$VERSION for rebuilding this gem" or something like that.

lib/rubygems/commands/rebuild_command.rb
duckinator: The main reason I can't use `FetchCommand#execute` is that I need a clear indicator of whether the lookup + download failed.

I'm basically using a modified version of this part, which downloads a single gem:

https://github.com/rubygems/rubygems/blob/bb66253/lib/rubygems/commands/fetch_command.rb#L73-L91

lib/rubygems/commands/rebuild_command.rb
duckinator: This is definitely a good idea.

lib/rubygems/commands/rebuild_command.rb
duckinator: Yeah, it looks like it's identical, so it can be shared.

Any idea where that should be put? It seems like shared things in `rubygems/commands/` just get dumped in `rubygems/*`, which seems like a pattern that may be worth worth breaking. Perhaps creating `rubygems/commands/shared/` for that purpose? (I'm okay with other names for such a directory, it's just first one that came to mind.)

<details><summary>details about shared things for `rubygems/commands/*`</summary>

For context, it appears that at least these files are only used in `rubygems/commands/` (and corresponding tests):

- `rubygems/gemcutter_utilities`
- `rubygems/install_message`
- `rubygems/install_default_message`
- `rubygems/install_update_options`
- `rubygems/installer`
- `rubygems/local_remote_options`
- `rubygems/query_utils`
- `rubygems/rdoc`

</details>

lib/rubygems/commands/rebuild_command.rb
duckinator: Done. :3

lib/rubygems/commands/rebuild_command.rb
duckinator: It probably makes sense to pull this out and include it (with some other information) on <https://guides.rubygems.org>.

lib/rubygems/commands/rebuild_command.rb
duckinator: It now uses a temporary directory, instead of the working directory.

lib/rubygems/commands/rebuild_command.rb
duckinator: Resolved by making `gem rebuild` use a temporary directory for output.

lib/rubygems/commands/rebuild_command.rb
duckinator: This was mainly there so I had a practical example to reference. I think <https://smallest.dog/resources/reproducible-gem-builds/> has taken over that purpose, however, so I'm removing it from the help text.

lib/rubygems/commands/rebuild_command.rb
duckinator: I updated the help text to be more correct.

duckinator:
@segiddins @deivid-rodriguez I've done a lot of cleanup on this today. From your reviews, this is all that's left:

1. `find_gemspec()` is redundant with the `gem build` command.
2. `download_gem()` and `FetchCommand#execute` could potentially be refactored to have a common helper function.
3. It does not bail early if the RubyGems version is wrong.

Do y'all think any of these should block merging this PR? And do you have any other feedback?

As for resolving those 3 remaining things:
- For 1 and 2, I'm not sure where to put the common functionality.
- For 3, I'm not sure how to take a path to a `.gem` file and get metadata from it.

duckinator:
Rebased off master (2e28234).

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,250 @@
...
+
+  private
+
+  def sha256(file)
+    Digest::SHA256.hexdigest(File.read(file))
+  end
+
+  def get_timestamp(file)
+    mtime = nil
+    File.open(file, "rb") do |f|
```
segiddins, line 105:
```suggestion
    File.open(file, Gem.binary_mode) do |f|
```

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,250 @@
...
+      build_gem(gem_name, source_date_epoch, new_file)
+    end
+
+    compare(source_date_epoch, old_file, new_file)
+  end
+
+  private
+
+  def sha256(file)
+    Digest::SHA256.hexdigest(File.read(file))
```
segiddins, line 100:
```suggestion
    Digest::SHA256.hexdigest(Gem.read_binary(file))
```

though ideally we'd use one of the helpers @martinemde wrote, so we can avoid reading the whole gem contents into memory at once

lib/rubygems/commands/rebuild_command.rb
segiddins: could it be extracted into a common `SpecFetching` module?

segiddins (commented):
> For 1 and 2, I'm not sure where to put the common functionality.

Just throw it into a new module somewhere

> For 3, I'm not sure how to take a path to a .gem file and get metadata from it.

`Gem::Package.new(path).spec` will give you the gemspec object from the .gem at path

duckinator:
It now bails early if the RubyGems version doesn't match:

```
puppy@orthrus:~/test/okay$ gem rebuild okay 12.0.2
Fetching okay-12.0.2.gem
Downloaded okay version 12.0.2 as /tmp/gem_rebuild20240131-79494-x86r8z/old/okay-12.0.2.gem.
ERROR:  You need to use the same RubyGems version okay v12.0.2 was built with.

okay v12.0.2 was built using RubyGems v3.5.0.dev.
Gem files include the version of RubyGems used to build them.
This means in order to reproduce okay-12.0.2.gem, you must also use RubyGems v3.5.0.dev.

Please install RubyGems v3.5.0.dev and try again.
puppy@orthrus:~/test/okay$
```

duckinator:
Rebased off master (5602c2b).

duckinator:
@segiddins @deivid-rodriguez I believe all that's left from your suggestions is:

- [ ] Extract `find_gemspec()` into a common module and have `RebuildCommand` and `BuildCommand` both use it.
- [ ] Extract `download_gem()` into a common module and have `RebuildCommand` and `BuildCommand` both use it.

I'm not feeling great atm, so I'm gonna call it a day. If you want to leave these two things for a follow-up PR, you can go ahead and merge this; otherwise, I'll try to get those changes done sometime in the next week or so.

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,270 @@
...
+      say "SUCCESS - original and rebuild hashes matched"
+    else
+      say "FAILURE - original and rebuild hashes did not match"
+
+      if options[:diff]
+        say
+        if system("diffoscope", old_file, new_file).nil?
+          alert_error "error: could not find `diffoscope` executable"
+        end
+      end
```
deivid-rodriguez, line 157: Should we add an `else` here

> To see package differences, use the `--diff` flag (needs `diffoscope` installed)

duckinator: I have mixed feelings about that. On one hand, it's good for discoverability; on the other hand, having the clear SUCCESS/FAILURE right at the end by default has proven really nice in practice.

duckinator: Added a message suggesting using `--diff` if you didn't use it.

duckinator:
Extracting common code from `RebuildCommand#download_gem` and `FetchCommand#execute` doesn't seem to make sense. There's only like 4 lines that are the same, and they're scattered throughout the two functions.

I believe everything else y'all have requested has been addressed.

duckinator:
Rebased off master (1aa7252). @segiddins @deivid-rodriguez this is ready for a final review. If y'all are happy with it and CI passes, it's ready to merge.

segiddins (approved):

deivid-rodriguez (approved):

duckinator:
Rebased off master (5956bfb). Will merge if CI passes. :+1:

lib/rubygems/commands/rebuild_command.rb
```diff
@@ -0,0 +1,264 @@
...
+require "digest"
+require "fileutils"
+require "tmpdir"
+require_relative "../gemspec_helpers"
+require_relative "../package"
+
+class Gem::Commands::RebuildCommand < Gem::Command
+  include Gem::GemspecHelpers
+
+  DATE_FORMAT = "%Y-%m-%d %H:%M:%S.%N Z"
```
nobu, line 13: What is this constant for?
Seems unused.

duckinator: @nobu good catch, it is unused. I opened #7546 to remove it. :+1:

duckinator:
@deivid-rodriguez I think we should release it. The plugin is basically a way to "backport" the functionality to old versions, and only gets updated by [a script that updates the plugin's files with ones from the RubyGems repo](https://github.com/duckinator/gem-rebuild/blob/main/bin/update-source-files).

The sooner the real version is out, the sooner the plugin becomes irrelevant.

nobu:
Is it intentional that the downloaded and rebuilt gem files are left under `gem_rebuild*` temporary directory?

deivid-rodriguez:
mmmm, I think it'd be better to clean after ourselves? Thoughts @duckinator?

duckinator:
leaving the files after `gem rebuild` was intentional, for local inspection, but the _test suite_ should be made to clean up after itself independently of that.

PR #4918, opened 2021-09-17
Use example.com in new gem template, since it will never have a potentially dangerous backing website

### LINKED ISSUE #4917: Potentially dangerous URL in Gem creation code generator
All y'all might want to change the code generator that creates the `.gemspec` upon running `bundle gem foobar` (as is given in the newbie instructions at https://bundler.io/guides/creating_gem.html): `foobar.gemspec` contains the example URL `https://mygemserver.com` which has apparently been domaingrabbed and which looks sort of shady. A good alternative choice might be [example.com](https://example.com). Okay, that's it, thanks for an awesome piece of software, bye!

fiveNinePlusR:
example.com is the canonical stand in for domain examples and will never have a backing website.

via https://www.rfc-editor.org/rfc/rfc2606.html

possible malicious website in gemspec template

Closes #4917.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -16,7 +16,7 @@ Gem::Specification.new do |spec|
 <%- end -%>
   spec.required_ruby_version = ">= <%= config[:required_ruby_version] %>"
 
-  spec.metadata["allowed_push_host"] = "TODO: Set to 'https://mygemserver.com'"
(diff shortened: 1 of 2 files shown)


hsbt:
@fiveNinePlusR Thanks. Can you fix the test failures?

fiveNinePlusR:
Not sure how to run the CI tests again but it should pass now.

fiveNinePlusR:
@hsbt Should i resubmit the PR to initiate tests or do you just want to trigger them?

deivid-rodriguez (approved):
Thank you! :heart_eyes:

simi (approved):

PR #4920, opened 2021-09-20
Require at least Ruby 2.6.0 for gems created with recent rubies

okuramasafumi:

When I generated a new gem, the required Ruby version was `2.4.0`.
I wanted to use latest features of Ruby, so I changed the required version to `2.6.0`, but `rubocop.yml` file also contains `2.4.0` string and it was annoying.

I changed the default required Ruby version from `2.4.0` to `2.6.0`.


bundler/lib/bundler/cli/gem.rb
```diff
@@ -68,7 +68,7 @@ def run
         :bundler_version  => bundler_dependency_version,
         :git              => use_git,
         :github_username  => github_username.empty? ? "[USERNAME]" : github_username,
-        :required_ruby_version => Gem.ruby_version < Gem::Version.new("2.4.a") ? "2.3.0" : "2.4.0",
+        :required_ruby_version => Gem.ruby_version < Gem::Version.new("2.6.a") ? "2.3.0" : "2.6.0",
```
simi, line 71: This change means for Ruby 2.4 and 2.5 it will fallback to 2.3 and that is breaking some specs. To keep the old logic (for pre 2.4), we can keep the `2.4.a` check (as before), but instead of using `2.6.0`, we can use your current Ruby version.

okuramasafumi: I see, but what if a user uses `2.5.0` for example?

deivid-rodriguez: I think as long as we also change the `:linter_version` accordingly using the same criteria, namely, if using an old ruby, use an old linter version too, this change should be fine? For example, here: https://github.com/rubygems/rubygems/blob/3387dbf66d3721f1a25ea60d6bab8fddaf257f74/bundler/lib/bundler/cli/gem.rb#L169

okuramasafumi: We could choose the latest available linter version for each Ruby version, like:
```ruby
case Gem.ruby_version
when -> v { v < Gem::Version.new("2.4.a") } then "0.81.0"
when -> v { v < Gem::Version.new("2.5.a") } then "1.12.1"
else
  "1.21.0"
end
```
If you agree we should do this, I'll add this snippet to this Pull Request. What do you think? @deivid-rodriguez

deivid-rodriguez: Sure, just make sure you use `1.12` and `1.21`. We use the`~>` operator in the template so that the result is flexible enough to not require frequent updates.

Also, style wise I would stick to `if/elsif/else` rather than `case` with procs.

okuramasafumi: @deivid-rodriguez I got it.
I'll apply the same strategy for `required_ruby_version` so that we can use the current Ruby version as a required Ruby version.

deivid-rodriguez: `required_ruby_version` uses the `>=` operator by default, so what we have currently seems fine?

okuramasafumi:
@deivid-rodriguez @simi Hi, I finally resolved the issue and CI is all green, thank you!

bundler/spec/commands/newgem_spec.rb
```diff
@@ -569,7 +569,13 @@ def create_temporary_dir(dir)
...
       bundle "gem #{gem_name}"
 
-      expect(generated_gemspec.required_ruby_version).to eq(Gem::Requirement.new(Gem.ruby_version < Gem::Version.new("2.4.a") ? ">= 2.3.0" : ">= 2.4.0"))
+      required_ruby_version = if Gem.ruby_version < Gem::Version.new("2.4.a") then ">= 2.3.0"
+      elsif Gem.ruby_version < Gem::Version.new("2.5.a") then ">= 2.4.0"
+      elsif Gem.ruby_version < Gem::Version.new("2.6.a") then ">= 2.5.0"
+      else
+        ">= 2.6.0"
+      end
+      expect(generated_gemspec.required_ruby_version).to eq(Gem::Requirement.new(required_ruby_version))
```
deivid-rodriguez, line 578: At this point it feels like we are just reimplementing the same logic we are testing inside the test itself. I think I'd change to spec to just check that a `required_rub_version` property is generated with a lower bound (starts with >=).

bundler/tool/bundler/standard_gems.rb.lock
```diff
@@ -63,4 +61,4 @@ DEPENDENCIES
   test-unit
 
 BUNDLED WITH
-   2.3.0.dev
+   2.2.24
```
deivid-rodriguez, line 64: Can you update the `update_locked_bundler` rake task at the root of the repsository and run it to update all lockfiles?

bundler/spec/commands/newgem_spec.rb
okuramasafumi: Thanks, I might misunderstand your intention, but I've rewritten that spec to simply check `>=`.

deivid-rodriguez: If you observe the previous logic here, and the logic you added inside the `required_ruby_version` method in `bundler/lib/bundler/cli/gem.rb`, it's exactly the same. That means that in order to test some logic, we are using exactly the same logic that we want to get tested. The defeats the point of having a test.

It'd be like using a spec like

`expect(sum(1, 1)).to eq(1 + 1)`

to test the method

```
def sum(a, b)
  a +  b
end
```

okuramasafumi: @deivid-rodriguez Could you check the current code? It checks using `>=` only.

okuramasafumi:
The CI got cancelled, but it should run green.

bundler/spec/commands/newgem_spec.rb
deivid-rodriguez: Yes, it looks good, thank you!

deivid-rodriguez (approved):
Thanks for the patch and your patience @okuramasafumi!

PR #4919, opened 2021-09-17
Fix redacted credentials being sent to gemserver

jdliss:
## What was the end-user or developer problem that led to this PR?

When using a `username:password` remote source for a private gem server, the credentials redaction process was modifying the original remote uri resulting in subsequent requests being rejected with a 401.  The issue is detailed here: https://github.com/rubygems/rubygems/issues/4906

Also fixed a small bug I stumbled onto that did not seem to be causing problems.  A `Gem::Uri` instance was being passed into `Gem::Uri.new` unnecessarily, which created a nested instance.  This did not seem to be causing issues because of the `method_missing` flow in the `Gem::Uri` class, but I updated the `Gem::Uri.new` call as this behavior did not seem like it was intended.

## What is your fix for the problem, implemented in this PR?

I was able to reproduce the issue and examine the web traffic using charles as detailed here: https://github.com/rubygems/rubygems/issues/4906#issuecomment-914492070
The credentials being sent to the gem server were `username:REDACTED` instead of `username:password`.

The issue was that the `Gem::Uri` class used shallow cloning when redacting the password for logging purposes, but this resulted in modifying the uri that would then go on to be used in the request.  To fix this issue I used `initialize_copy` to implement deep cloning to make sure the uri that is used for redacting credentials is a copy, not the original.

Thanks @pindell-matt for the help!

Fixes #4906.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #4929, opened 2021-09-23
Revert "Fix an issue causing nested Gem::Uri instances"

deivid-rodriguez:

Commit 6589f7bcc7a63a47cb73f58a290c1e1ac42bba99 caused issues since sometimes the `uri` passed to `Gem::Request` is not a `Gem::Uri` but a standard `URI`, so we can't call `#redacted` on it.

I discovered this when trying to release bundler 2.2.28 using the (still unreleased) 3.2.28 version of rubygems and getting the following crash:

```
bundler 2.2.28 built to pkg/bundler-2.2.28.gem.
    undefined method `redacted' for #<URI::HTTPS https://rubygems.org/api/v1/gems>
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/request.rb:194:in `perform_request'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/request.rb:152:in `fetch'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/remote_fetcher.rb:309:in `request'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/gemcutter_utilities.rb:234:in `request_with_otp'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/gemcutter_utilities.rb:105:in `rubygems_api_request'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/commands/push_command.rb:89:in `send_push_request'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/commands/push_command.rb:81:in `send_gem'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/commands/push_command.rb:66:in `execute'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/command.rb:323:in `invoke_with_build_args'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/command_manager.rb:178:in `process_args'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/command_manager.rb:147:in `run'
	/home/deivid/.rbenv/versions/3.0.2/lib/ruby/site_ruby/3.0.0/rubygems/gem_runner.rb:53:in `run'
	/home/deivid/.rbenv/versions/3.0.2/bin/gem:15:in `<main>'
```

Revert the commit, since in addition to causing issues, it was unrelated to the fix in #4919.

### DIFF
--- a/lib/rubygems/request.rb
+++ b/lib/rubygems/request.rb
@@ -191,7 +191,7 @@ class Gem::Request
     begin
       @requests[connection.object_id] += 1
 
-      verbose "#{request.method} #{@uri.redacted}"
+      verbose "#{request.method} #{Gem::Uri.new(@uri).redacted}"
 
       file_name = File.basename(@uri.path)
       # perform download progress reporter only for gems


deivid-rodriguez:
@jdliss I had to revert this change because it created some issues. The spirit of the change seems correct, but we need to make sure `Gem::Request` objects are always instantiated with `Gem::Uri` instances. Right now it seems we're passing `URI` and `Gem::URI`'s inconsistently.

olleolleolle (approved):
Excellent context in the commit message and description!

jdliss:
@deivid-rodriguez sorry about that! I thought about that when I made the change (https://github.com/rubygems/rubygems/commit/6589f7bcc7a63a47cb73f58a290c1e1ac42bba99) and did not make it clear in the PR that that could be an issue.  I saw (here)[https://github.com/rubygems/rubygems/blob/master/lib/rubygems/remote_fetcher.rb#L307] during my diagnosing of the issue that the `uri` being passed into the `Gem::Request` class was already a `Gem::Uri` instance - hence my change to remove the instantiation inside the `Gem::Request` class to avoid a doubly nested `Gem::Uri` instance which was happening.  I figured that the intended behavior of `Gem::Request` was that the passed in `uri` should be a `Gem::Uri` and so I made the change to update the tests and remove the `Gem::Uri` instantiation inside the class.  I'll be honest I don't know what the best practice is around type enforcement in ruby, but I can add some code to make sure the uri passed into the `Gem::Request` class is always instantiated as a `Gem::Uri` while avoiding nested `Gem::Uri` instances which should make it backwards compatible.  Should be a simple update I think.  If you'd like me to take a stab at that I'd be happy to hear your thoughts.

Anyway, thanks for catching that issue before it got released!

deivid-rodriguez:
Hei @jdliss, no worries, these things happen, and we caught it on time before it bit anyone anyways, so no big deal.

Type enforcement in ruby is not something very common in ruby, you usually rely on tests to catch this kind of issues, although there's new exciting stuff going on in that area: https://sorbet.org/.

I'm happy if you try to take a stab at avoiding those nested `Gem::Uri`'s in a way that doesn't cause issues :+1:.

jdliss:
@deivid-rodriguez I added a PR with some changes to allow for either `URI` or `Gem::Uri` instances to be passed into `Gem::Request`, while also removing (and preventing) the nested `Gem::Uri` bug.  Nested instances of `Gem::Uri` shouldn't cause any issues though because of the `method_missing` flow that's used, so if you feel this is not the correct way to handle this type of issue then feel free to close since I'm mostly fixing a non-issue.

https://github.com/rubygems/rubygems/pull/4933

PR #4932, opened 2021-09-23
Use pend instead of skip in the absence of git in test helper

lucaskanashiro:

In Debian, we build rubygems without git and almost 40 tests failed because the `skip` method is missing. Checking other cases, for instance where tests depending on cmake are skipped, the `skip` method is not used but `pend`.

Use `pend` instead of `skip` in `have_git?` method.

- [-] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes

### DIFF
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -595,7 +595,7 @@ class Gem::TestCase < Test::Unit::TestCase
   def have_git?
     return if in_path? @git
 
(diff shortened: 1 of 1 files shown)


hsbt (approved):

hsbt:
Thanks!

PR #4933, opened 2021-09-23
Prevent nested gem uri instances

jdliss:

I had previously fixed an issue ([now reverted](https://github.com/rubygems/rubygems/pull/4929)) where a `Gem::Uri` instance was being passed into `Gem::Uri.new` creating a nested instance.  This was happening because there are places where both `URI` and `Gem::Uri` instances are being passed into `Gem::Request`.  This bug was not causing any issues because of the `method_missing?` flow in the `Gem::Uri` class.  However, fixing the nested instance issue and treating all uri's passed into `Gem::Request` as `Gem::Uri` instances created issues when other parts of the code base pass in `URI`.  

As a stopgap measure, these changes will stop the nesting of `Gem::Uri` instances to prevent any possible issues that may arise from that, while also forcing all uri's passed into `Gem::Request` to be wrapped in a `Gem::Uri` class so that the code can be clearer about what the uri is expected to be.

`# Gem::Uri`
* If a `Gem::Uri` is passed into `Gem::Uri.new`, the `parse` method will return the `parsed_uri` in the passed in `Gem::Uri` instance.

`# Gem::Request`
* Always wrap the passed in uri in a `Gem::Uri` instance.

A better solution in my mind would probably be to go through the code base and make sure only one type of uri is passed into `Gem::Request` and add an argument error if the other type is used to prevent this kind of multi type handling from being necessary.  I'm not confident that I would be able to find all these places to solidify one pattern over the other, so as a secondary alternative this will allow for flexibility and should prevent issues where either type is used.  

Since the original nested `Gem::Uri` instance bug was not causing issues however, feel free to close and keep the current behavior (although I believe it should be fixed since it seems to be unintended).

### DIFF
--- a/lib/rubygems/request.rb
+++ b/lib/rubygems/request.rb
@@ -26,7 +26,7 @@ class Gem::Request
   end
 
   def initialize(uri, request_class, last_modified, pool)
-    @uri = uri
+    @uri = Gem::Uri.new(uri)
     @request_class = request_class
     @last_modified = last_modified
     @requests = Hash.new 0
@@ -191,7 +191,7 @@ class Gem::Request
     begin
       @requests[connection.object_id] += 1
 
-      verbose "#{request.method} #{Gem::Uri.new(@uri).redacted}"
+      verbose "#{request.method} #{@uri.redacted}"
 
       file_name = File.basename(@uri.path)
       # perform download progress reporter only for gems
--- a/lib/rubygems/uri.rb
+++ b/lib/rubygems/uri.rb
@@ -45,7 +45,7 @@ class Gem::Uri
 
   protected
 
-  # Add a protected reader for the cloned instance to access the original object's parsed uri
+  # A protected reader for other Gem::Uri instances to access @parsed_uri
   attr_reader :parsed_uri
 
   private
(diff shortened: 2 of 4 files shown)


olleolleolle:
Thanks for making this thoughtful change to the URI handling. Is this statement now correct: _With this change, the code reader should be able to assume that most `uri` locals are Gem::Uri instances._

jdliss:
Yes but only in the `Gem::Request` class.  I didn't change the uri behavior in other places.  From what I've seen in the code it seems like `Gem::Uri` should be preferred since it allows for logging with redacted credentials, but I think there are places where it's not used.  However it is safe to pass either into the `Gem::Request` class

jeromedalbert:
@jdliss Are you still interested in moving this PR forward?

jdliss:
Yes it's good to go as long as nothing's changed in the last couple years its been open 😄

jeromedalbert:
Well it looks like your PR has conflicts. 😅 Would you mind resolving them?

jdliss:
Ok after further review, there have been so many changes to the project since this PR was created that I do not feel confident in adding these changes.  Plus I think the bug I wanted to fix has been fixed anyway.  Feel free to close.

jeromedalbert:
Sounds good, looks like this issue can be closed @deivid-rodriguez, if you're OK with this resolution.

deivid-rodriguez:
Cool! I am ok with this resolution. It was only a small code improvement on top of the actual bug that @jdliss [did fix](https://github.com/rubygems/rubygems/pull/4919) at the time.

PR #4937, opened 2021-09-24
Only disallow FIXME/TODO for first word of gemspec description

duckinator:

As @nobu pointed out in 7890c98, #4841 — my fix for #4808 — used `^` in the regex, which matched every line. Using `\A` to match the beginning of the entire string makes more sense.

I used `\A` instead of `^`.


duckinator:
cc @deivid-rodriguez

deivid-rodriguez (approved):
Awesome! :tada:

Issue #4943, opened 2021-09-29
Non-deterministic behaviour with git sources using globs


riffraff:
### Describe the problem as clearly as you can

We've recently started experiencing a problematic behaviour: the git sources list in `Gemfile.lock` keeps changing when running `bundle` in different operating systems. 
This appears to be caused by the fact that we have now multiple gems which refer to the same repository and are loaded using the glob option, something like this 

```
git_source(:organization) do |repo_name|
  "git@github.organization/#{repo_name}.git"
end

gem 'foo', organization: 'private_repo', glob: "gems/foo/*.gemspec", ref: '123'
gem 'bar', organization: 'private_repo', glob: "gems/bar/*.gemspec", ref: '123' 
gem 'baz', organization: 'private_repo', glob: "gems/baz/*.gemspec", ref: '123' # depends on bar and foo
```

It appears bundler attempts to order sources in a deterministic way [sorting by their string representation](https://github.com/rubygems/rubygems/blob/d01d6ca84011b05251b31d496f75f6aff8f677c4/bundler/lib/bundler/source_list.rb#L109) but the string representation is the same for all three, i.e. "repo name @ git reference", so the order in which the sources were initialized gets through, and we get a non-deterministic order of the source stanzas in `Gemfile.lock`

By changing `lock_other_sources` to account for the glob
```
# Bundler::SourceList 
  def lock_other_sources
    (path_sources + git_sources + plugin_sources).sort_by do |source|
      glob = source.respond_to?(:glob) ? source.glob : ''
      [source.to_s, glob]
    end
  end
```

we are able to get a behaviour which is consistent across different operating systems, and results in a `Gemfile.lock` where the three sources are sorted by glob (bar -> baz -> foo).

I guess we could change `SourceList#to_s` but that seems like it's used in multiple ways, so maybe a better way to do this is to introduce a `SourceList#sort_key` method which would only be used for this goal. 
I can try to provide a patch if you're interested. 

### Post steps to reproduce the problem

It's pretty hard to offer a way to reproduce this as it happens with our internal repo, and you need both a linux and macOS machine, but I will try to setup some test case later if the problem is not clear.

Anyway, the various software versions were ruby 2.6.5p114, rubygems 3.2.28, bundler 2.2.28.

### Which command did you run?

`bundle`

### What were you expecting to happen?

Get the same `Gemfile.lock` in Linux and macOS 

### What actually happened?

The Gemfile.lock is different on the two machines

[...]

PR #4947, opened 2021-09-30
Include glob information in Bundler::Source::Git#to_s

riffraff:

This fixes the issue described in https://github.com/rubygems/rubygems/issues/4943, where multiple gems in a single repo identified by separate glob lead to non-deterministic lockfile generation.

The changes here make it so a glob is included in `Bundler::Source::Git#to_s` which in turn makes the Lockfile generation (more) deterministic when including multiple gems from the same git repo with different globs. 

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -42,7 +42,7 @@ module Bundler
         %w[ref branch tag submodules].each do |opt|
           out << "  #{opt}: #{options[opt]}\n" if options[opt]
         end
-        out << "  glob: #{@glob}\n" unless @glob == DEFAULT_GLOB
+        out << "  glob: #{@glob}\n" unless default_glob?
         out << "  specs:\n"
       end
 
@@ -75,12 +75,20 @@ module Bundler
             git_proxy.branch
           end
 
-          rev = " (at #{at}@#{shortref_for_display(revision)})"
+          rev = "at #{at}@#{shortref_for_display(revision)}"
         rescue GitError
           ""
         end
 
-        "#{@safe_uri}#{rev}"
+        specifiers = [rev, glob_for_display].compact
+        suffix =
+          if specifiers.any?
+            " (#{specifiers.join(", ")})"
+          else
+            ""
+          end
+
+        "#{@safe_uri}#{suffix}"
       end
 
       def name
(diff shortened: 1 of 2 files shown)


bundler/spec/bundler/source/git_spec.rb
```diff
@@ -24,5 +24,31 @@
...
+      let(:options) do
+        { "uri" => uri, "glob" => glob }
+      end
+
+      it "includes it" do
+        expect(subject.to_s).to eq "https://github.com/foo/bar.git (glob: bar/baz/*.gemspec)"
+      end
+    end
+
+    context "when the source has a reference" do
```
riffraff, line 39: this was not tested, and I spent some time before realizing there was an exception being swallowed, so I added a spec.

bundler/lib/bundler/source/git.rb
```diff
@@ -80,7 +80,7 @@ def to_s
           ""
         end
 
-        "#{@safe_uri}#{rev}"
+        "#{@safe_uri}#{rev}#{glob_for_display}"
```
riffraff, line 83: I'm not 100% sure of the format, this makes it so that both annotations show up independently e.g.

```
Using foo 0.1.1 from git@github.com:org/repo.git (at f334c1b@f334c1b) (glob: gems/foo/*.gemspec)
```
Maybe we should have a single set of parenthesis and separate them by comma instead

```
Using foo 0.1.1 from git@github.com:org/repo.git (at f334c1b@f334c1b, glob: gems/foo/*.gemspec)
```

riffraff:
@deivid-rodriguez this is the PR as promised, let me know if there's something I can do to improve it

bundler/lib/bundler/source/git.rb
deivid-rodriguez: Both options seem fine to maybe, but if I have to choose I think I'd go with the compact version although not a big deal.

deivid-rodriguez:
Hello! It looks great to me! Just need to agree with the format, from the two options you present I think I'd go with the latter, but I don't have a strong opinion.

riffraff:
I did a minor tweak to the second option, so it's more consistent (I had typed it wrong in my example), I hope this is ok, thanks for the support.

deivid-rodriguez (approved):
Looks great, thank you! ❤️

konalegi:
@deivid-rodriguez do you happen to know when this change is going to be released? This inconsistency is really painful for us :( Since the order is different for linux and os x developers.
Thanks.

PR #4956, opened 2021-10-04
Fix `assert_contains_make_command` on `make` defined by environment variable.

### LINKED ISSUE #4921: rubygems tests failing with customized environment variable "make"
I'm having a problem. When I was working with defined `make` environment variable, I faced this issue. In ruby/ruby, it is used in the following part.
https://github.com/ruby/ruby/blob/9770bf23b7a273246b9a6b084e79a8fb6fc1af11/.github/workflows/compilers.yml#L184

My current problem is 

On the latest master branch `3387dbf66d3721f1a25ea60d6bab8fddaf257f74`, the following tests fail with the defined `make` environment variable.

Here is the reproducer. I am using Fedora 34.

```
$ ruby -v
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]

$ export make='make -j2'

$ rake setup

$ ruby -Ilib:test test/rubygems/test_gem_ext_configure_builder.rb
Loaded suite test/rubygems/test_gem_ext_configure_builder
Started
F
========================================================================================================
Failure: test_self_build(TestGemExtConfigureBuilder): Expected make command "make -j2": make -j2 DESTDIR\=
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:299:in `assert_contains_make_command'
test/rubygems/test_gem_ext_configure_builder.rb:37:in `test_self_build'
========================================================================================================
.F
========================================================================================================
Failure: test_self_build_has_makefile(TestGemExtConfigureBuilder): Expected make command "make -j2": make -j2 DESTDIR\=
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:299:in `assert_contains_make_command'
test/rubygems/test_gem_ext_configure_builder.rb:76:in `test_self_build_has_makefile'
========================================================================================================

Finished in 0.050182784 seconds.
--------------------------------------------------------------------------------------------------------
3 tests, 16 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
33.3333% passed
--------------------------------------------------------------------------------------------------------
59.78 tests/s, 318.83 assertions/s
Coverage report generated for Unit Tests to /home/jaruga/var/git/rubygems/coverage. 1142 / 3755 LOC (30.41%) covered.
```

The reason of the failures are at test/rubygems/test_gem_ext_configure_builder.rb:37:in `test_self_build', the called `assert_contains_make_command` with 1st argument: `''` has a bug.

test/rubygems/test_gem_ext_configure_builder.rb:37

```
assert_contains_make_command '', output.shift
```

The used make command is to get from `MAKE` or `make` if it is defined according to `lib/rubygems/ext/builder.rb` method `make`.

```
  def self.make(dest_path, results, make_dir = Dir.pwd)
    ...
    make_program = ENV['MAKE'] || ENV['make'] || $1
    ...
```

Then in the testing method `assert_contains_make_command`, here is a logic to set empty string to make the `assert_contains_make_command` return true even when the 1st argument is empty string. However if the `make` environment variable has argument such as `-j2`,  the targets is not empty, therefore the `targets` array does not has empty string `''`, and the assertion fails. 

test/rubygems/helper.rb

```
  def parse_make_command_line(line)
    ...
        targets << '' if targets.empty?
    ...
```

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

```
$ gem env
PASTE HERE
```

The `gem` command is not used in this case. I tested this issue on the latest master branch `3387dbf66d3721f1a25ea60d6bab8fddaf257f74`.

junaruga:
This PR fixes #4921.

The `parse_make_command_line` in `assert_contains_make_command` fails to get
targets correctly, when the make command is set with make options by
environment variable such as `export make='make -j2'` at
lib/rubygems/ext/builder.rb::make. In this case, it's hard to get the make
targets identifying make options. So, just parse the make arguments,
and check if the arguments include the given target.

Enhance test/rubygems/test_gem_ext_ext_conf_builder.rb#test_class_build_env_make
to test `assert_contains_make_command` with the make command with options and
to pass on `export make='make -j2'.

The rubygems tests fails on `export make='make -j2'`.

The `parse_make_command_line` method currently does not get the make target considering the make options from the make command. So, I changed how to check a make target is included in the make command, parsing both make arguments (make options and make targets).

I also considered the case on the `make` environment variable is set such as `export make=`make -j2`. In this case, the test fails. So, I modified it.

I checked the code on my local.

```
$ bundler/bin/rubocop -a
Inspecting 388 files
....................................................................................................................................................................................................................................................................................................................................................................................................

388 files inspected, no offenses detected
```

### DIFF
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -246,7 +246,7 @@ class Gem::TestCase < Test::Unit::TestCase
   end
 
   def parse_make_command_line(line)
-    command, *args = line.shellsplit
+    args = line.sub(/^#{Regexp.escape make_command}/, "").shellsplit
 
     targets = []
     macros = {}
@@ -263,7 +263,7 @@ class Gem::TestCase < Test::Unit::TestCase
     targets << '' if targets.empty?
 
     {
-      :command => command,
+      :command => make_command,
       :targets => targets,
       :macros => macros,
     }
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -290,7 +290,6 @@ class Gem::TestCase < Test::Unit::TestCase
       make = parse_make_command_line(line)
 
       if make[:targets].include?(target)
-        yield make, line if block_given?
         true
       else
         false
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -245,16 +245,14 @@ class Gem::TestCase < Test::Unit::TestCase
     output.scan(/^#{Regexp.escape make_command}(?:[[:blank:]].*)?$/)
   end
 
-  def parse_make_command_line(line)
+  def parse_make_command_line_targets(line)
     args = line.sub(/^#{Regexp.escape make_command}/, "").shellsplit
 
     targets = []
-    macros = {}
 
     args.each do |arg|
       case arg
       when /\A(\w+)=/
-        macros[$1] = $'
       else
         targets << arg
       end
@@ -262,11 +260,7 @@ class Gem::TestCase < Test::Unit::TestCase
 
     targets << '' if targets.empty?
 
-    {
-      :command => make_command,
-      :targets => targets,
-      :macros => macros,
-    }
+    targets
   end
 
   def assert_contains_make_command(target, output, msg = nil)
@@ -287,9 +281,9 @@ class Gem::TestCase < Test::Unit::TestCase
     end
 
     assert scan_make_command_lines(output).any? {|line|
-      make = parse_make_command_line(line)
+      targets = parse_make_command_line_targets(line)
 
-      if make[:targets].include?(target)
+      if targets.include?(target)
         true
       else
         false
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -279,7 +279,7 @@ class Gem::TestCase < Test::Unit::TestCase
       )
     else
       msg = build_message(msg,
-        'Expected make command "%s": %s' % [
+        'Expected make command "%s", but was "%s"' % [
           ('%s %s' % [make_command, target]).rstrip,
           output,
         ]
--- a/test/rubygems/test_gem_ext_ext_conf_builder.rb
+++ b/test/rubygems/test_gem_ext_ext_conf_builder.rb
@@ -66,8 +66,11 @@ class TestGemExtExtConfBuilder < Gem::TestCase
     end
   end
 
-  def test_class_build_env_make
-    env_make = ENV.delete 'MAKE'
+  def test_class_build_env_MAKE
+    env_make = ENV.delete 'make'
+    ENV['make'] = nil
+
+    env_MAKE = ENV.delete 'MAKE'
     ENV['MAKE'] = 'anothermake'
 
     if java_platform?
@@ -89,7 +92,8 @@ class TestGemExtExtConfBuilder < Gem::TestCase
       assert_contains_make_command 'clean', output[4]
     end
   ensure
-    ENV['MAKE'] = env_make
+    ENV['MAKE'] = env_MAKE
+    ENV['make'] = env_make
   end
 
   def test_class_build_extconf_fail


test/rubygems/helper.rb
```diff
@@ -299,7 +297,7 @@ def assert_contains_make_command(target, output, msg = nil)
     assert scan_make_command_lines(output).any? {|line|
       make = parse_make_command_line(line)
 
-      if make[:targets].include?(target)
+      if target == '' || make[:arguments].include?(target)
```
junaruga, line 300: Sorry. The returning true always if the given make target is empty is not appropriate. I need to think a better logic.

junaruga: But the problem is that I have no idea about how to parse only the make targets skipping some patterns of the make options (`-jN`, `-i`, `-j N` or `--jobs=N`). Perhaps, we can just skip the tests when we find the "-" in the make command line, supposing it as the make options?

deivid-rodriguez:
How about the following approach: consider `-j` flag (or any other flags) part of the make command itself when parsing it.

```diff
diff --git a/test/rubygems/helper.rb b/test/rubygems/helper.rb
index 812142f7d2..c7be7a1118 100644
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -263,6 +263,8 @@ def parse_make_command_line(line)
 
     args.each do |arg|
       case arg
+      when /\A-/
+        command << " " << arg
       when /\A(\w+)=/
         macros[$1] = $'
       else
diff --git a/test/rubygems/test_gem_ext_ext_conf_builder.rb b/test/rubygems/test_gem_ext_ext_conf_builder.rb
index 70833f35cf..10a544cbbc 100644
--- a/test/rubygems/test_gem_ext_ext_conf_builder.rb
+++ b/test/rubygems/test_gem_ext_ext_conf_builder.rb
@@ -66,8 +66,11 @@ def test_class_build_rbconfig_make_prog
     end
   end
 
-  def test_class_build_env_make
-    env_make = ENV.delete 'MAKE'
+  def test_class_build_env_MAKE
+    env_make = ENV.delete 'make'
+    ENV['make'] = nil
+
+    env_MAKE = ENV.delete 'MAKE'
     ENV['MAKE'] = 'anothermake'
 
     if java_platform?
@@ -89,7 +92,8 @@ def test_class_build_env_make
       assert_contains_make_command 'clean', output[4]
     end
   ensure
-    ENV['MAKE'] = env_make
+    ENV['MAKE'] = env_MAKE
+    ENV['make'] = env_make
   end
 
   def test_class_build_extconf_fail
```

junaruga:
@deivid-rodriguez Sorry for my late response for your comment. OK. I am fine for your way. I see the PR fixes the case of `export make='make -j2'`, but doesn't fix the case of `export make='make -j 2'`. But the modification is better than the current one. I rebased the PR with your patch on the latest master, updating the commit message.

Thanks!

```
$ export make='make -j2'
$ ruby -Ilib:test test/rubygems/test_gem_ext_configure_builder.rb
Loaded suite test/rubygems/test_gem_ext_configure_builder
Started
...
Finished in 0.056902731 seconds.
--------------------------------------------------------------------------------------------------------
3 tests, 21 assertions, 0 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
100% passed
--------------------------------------------------------------------------------------------------------
52.72 tests/s, 369.05 assertions/s
Coverage report generated for Unit Tests to /home/jaruga/var/git/rubygems/coverage. 1145 / 3753 LOC (30.51%) covered.
```

```
$ export make='make -j 2'
$ ruby -Ilib:test test/rubygems/test_gem_ext_configure_builder.rb
Loaded suite test/rubygems/test_gem_ext_configure_builder
Started
F
========================================================================================================
Failure: test_self_build(TestGemExtConfigureBuilder): Expected make command "make -j 2": make -j 2 DESTDIR\=
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:301:in `assert_contains_make_command'
test/rubygems/test_gem_ext_configure_builder.rb:37:in `test_self_build'
========================================================================================================
.F
========================================================================================================
Failure: test_self_build_has_makefile(TestGemExtConfigureBuilder): Expected make command "make -j 2": make -j 2 DESTDIR\=
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:301:in `assert_contains_make_command'
test/rubygems/test_gem_ext_configure_builder.rb:76:in `test_self_build_has_makefile'
========================================================================================================

Finished in 0.049215118 seconds.
--------------------------------------------------------------------------------------------------------
3 tests, 16 assertions, 2 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
33.3333% passed
--------------------------------------------------------------------------------------------------------
60.96 tests/s, 325.10 assertions/s
Coverage report generated for Unit Tests to /home/jaruga/var/git/rubygems/coverage. 1145 / 3753 LOC (30.51%) covered.
```

deivid-rodriguez:
I pushed a different approach that should fix the remaining issues 👍.

PR #4960, opened 2021-10-05
Big speed up of `gem install` under Windows

deivid-rodriguez:

When I need to troubleshoot CI issues on Windows, it is very tedious because running specs is really really slow. Even running a single bundler spec takes over a minute.

I investigated the reason for this and I found that the culprit was a `gem install bundler-2.3.0.dev.gem` command that bundler specs run before the first spec to make sure specs run against a development version of bundler installed as a regular gem.

The `gem install` command was taking over 40 seconds on my VM (not sure why this particular `gem install` is so slow, realworld  `gem install` is faster).

I investigated the culprit here and I found that package extraction was causing a lot of `File.realpath` and `File.lstat` calls that were very very slow on Windows.
  
My fix is to make a series of refactorings to the file extraction methods in `Gem::Package` to minimize the number of this calls we need to make. This area is a bit sensitive because things are done they way they are done to prevent symlink attack vulnerabilities, but I think my refactorings are safe (and don't break regression tests for these vulnerabilities).

The results are summarized as follow:

### Running a single bundler spec

To test this first I set the `RGV` environment variable to point to the local version of rubygems using

```
PS V:\rubygems> [Environment]::SetEnvironmentVariable("RGV", "..",  [System.EnvironmentVariableTarget]::Process)
```

#### Without this PR

```
PS V:\rubygems\bundler> ruby bin/rspec spec/commands/install_spec.rb:821
Run options:
  include {:locations=>{"./spec/commands/install_spec.rb"=>[821]}}
  exclude {:truffleruby=>true, :jruby=>true, :readline=>true, :permissions=>true, :no_color_tty=>true, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.33.0", :realworld=>true, :sudo=>true}

Randomized with seed 62519
.

Finished in 1 minute 4.83 seconds (files took 0.79805 seconds to load)
1 example, 0 failures

Randomized with seed 62519
```

#### With this PR

```
PS V:\rubygems\bundler> ruby bin/rspec spec/commands/install_spec.rb:821
Run options:
  include {:locations=>{"./spec/commands/install_spec.rb"=>[821]}}
  exclude {:truffleruby=>true, :jruby=>true, :readline=>true, :permissions=>true, :no_color_tty=>true, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.33.0", :realworld=>true, :sudo=>true}

Randomized with seed 26393
.

Finished in 30.3 seconds (files took 0.77234 seconds to load)
1 example, 0 failures
```

### Running a realworld `gem install` command

To test this, first I modified the `bin/gem` script like this:

```diff
diff --git a/bin/gem b/bin/gem
index 133774ff11..a157c38631 100755
--- a/bin/gem
+++ b/bin/gem
@@ -5,9 +5,16 @@
 # See LICENSE.txt for permissions.
 #++
 
-require 'rubygems'
-require 'rubygems/gem_runner'
+require "benchmark"
 
-args = ARGV.clone
+puts Benchmark.measure {
+  require 'rubygems'
+  require 'rubygems/gem_runner'
 
-Gem::GemRunner.new.run args
+  args = ARGV.clone
+
+  begin
+    Gem::GemRunner.new.run args
+  rescue Exception
+  end
+}
```
#### Without this PR

```
PS V:\rubygems> ruby -Ilib bin/gem install activerecord
Successfully installed activerecord-6.1.4.1
1 gem installed
  1.625000   2.812000   4.437000 (  5.704987)
```

#### With this PR

```
PS V:\rubygems> ruby -Ilib bin/gem install activerecord
Successfully installed activerecord-6.1.4.1
1 gem installed
  1.172000   1.765000   2.937000 (  4.016889)
```


deivid-rodriguez:
Note that this actually speeds up `gem install` in general, not only on Windows. It's just more noticeable there. For example, `gem install activerecord` seems to be around 10-20% faster with this PR:

```
$ hyperfine --prepare 'git checkout master' --command-name 'master' 'ruby -Ilib bin/gem install activerecord' --prepare 'git checkout speed_up_gem_install' --command-name 'PR' 'ruby -Ilib bin/gem install activerecord'
Benchmark #1: master
  Time (mean ± σ):      1.355 s ±  0.197 s    [User: 663.9 ms, System: 125.3 ms]
  Range (min … max):    1.007 s …  1.648 s    10 runs
 
Benchmark #2: PR
  Time (mean ± σ):      1.136 s ±  0.145 s    [User: 625.2 ms, System: 97.5 ms]
  Range (min … max):    0.991 s …  1.388 s    10 runs
 
Summary
  'PR' ran
    1.19 ± 0.23 times faster than 'master'
```

deivid-rodriguez:
I completely changed the approach here. Instead of allowing the installation of symlinks inside gems that point outside of the installation folder, and then checking that those symlinks are not exploited to install files outside of the installation folder, I changed the code to prevent the installation of symlinks that point outside of the installation folder altogether. With that in place, there's no need to check for "mkdir safety" anymore and we can use `FileUtils.mkdir_p` directly.

The performance improvements are similar to the initial version of this PR but this approach feels simpler to me.

deivid-rodriguez:
@MSP-Greg Can you verify whether this makes any difference on a real Windows machine?

ahorek:
I can confirm there's a significant difference on a real windows machine

```
gem install activerecord

before
  5.157000   4.250000   9.407000 (  8.412141)

after
  5.219000   3.031000   8.250000 (  7.389975)
```

simi:
## 3.2.28

```
PS F:\tmp\rubygems> Measure-Command { gem install --install-dir=test-328 activerecord }

Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 37
Milliseconds      : 425
Ticks             : 374251843
TotalDays         : 0,000433161855324074
TotalHours        : 0,0103958845277778
TotalMinutes      : 0,623753071666667
TotalSeconds      : 37,4251843
TotalMilliseconds : 37425,1843
```

## speed_up_gem_install
```
PS F:\tmp\rubygems> Measure-Command { gem install --install-dir=test-speedup activerecord }


Days              : 0
Hours             : 0
Minutes           : 0
Seconds           : 34
Milliseconds      : 385
Ticks             : 343854883
TotalDays         : 0,000397980188657407
TotalHours        : 0,00955152452777778
TotalMinutes      : 0,573091471666667
TotalSeconds      : 34,3854883
```

deivid-rodriguez:
Thanks @simi! Nice idea to test on an isolated `--install-dir`, I guess that forces resolving, downloading and installing all dependencies, making it slow. Still confirms a noticiable improvement :)

deivid-rodriguez:
For what it's worth, I found what was causing bundler specs to be so slow on my Windows VM: I mount the rubygems clone in my host machine as a shared folder inside the VM. The filesystem mapping necessary for this makes things really slow. Cloning the repo from scratch on a regular folder inside the VM makes things fast again :tada:

#### Without this PR

```
PS C:\Users\deivi\rubygems\bundler> ruby bin/rspec spec/commands/install_spec.rb:821
Run options:
  include {:locations=>{"./spec/commands/install_spec.rb"=>[821]}}
  exclude {:truffleruby=>true, :jruby=>true, :readline=>true, :permissions=>true, :no_color_tty=>true, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.33.0", :realworld=>true, :sudo=>true}

Randomized with seed 38097
.

Finished in 7.92 seconds (files took 0.45736 seconds to load)
1 example, 0 failures

Randomized with seed 38097
```


#### With this PR

```
PS C:\Users\deivi\rubygems\bundler> ruby bin/rspec spec/commands/install_spec.rb:821
Run options:
  include {:locations=>{"./spec/commands/install_spec.rb"=>[821]}}
  exclude {:truffleruby=>true, :jruby=>true, :readline=>true, :permissions=>true, :no_color_tty=>true, :ruby_repo=>false, :bundler=>"!= 2", :git=>"!= 2.33.0", :realworld=>true, :sudo=>true}

Randomized with seed 58076
.

Finished in 4.66 seconds (files took 0.48285 seconds to load)
1 example, 0 failures

Randomized with seed 58076
```

Issue #4939, opened 2021-09-26
Bundle install error on rails


Stephanie041996:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/usr/local/bin/bundle install`

- **What did you expect to happen?**

  I expected Bundler to...

- **What happened instead?**

  Instead, what happened was...

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried...

- **Have you read our issues document, https://github.com/rubygems/rubygems/blob/master/bundler/doc/contributing/ISSUES.md?**

  ...

## Backtrace

```
Errno::EACCES: Permission denied @ rb_sysopen - /Library/Ruby/Gems/2.6.0/gems/rails-6.1.4.1/README.md
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:408:in `initialize'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:408:in `open'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:408:in `block (2 levels) in extract_tar_gz'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package/tar_reader.rb:65:in `each'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:389:in `block in extract_tar_gz'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:522:in `block in open_tar_gz'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:519:in `wrap'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:519:in `open_tar_gz'
  /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/rubygems/package.rb:388:in `extract_tar_gz'
[...]

Issue #4938, opened 2021-09-26
Unfortunately, an unexpected error occurred, and Bundler cannot continue.


silenceJT:
### Describe the problem as clearly as you can

I'm using ruby 3.0.2p107 and using macOS bigsur 11.6 on M1 chip , I am trying to install mysql2 gem and it shows error occured.

### Post steps to reproduce the problem

1. `bundle install mysql2`
2. Then I tried solutions from others like:
`brew install openssl`
Then export
`export LDFLAGS="-L/usr/local/opt/openssl/lib"`
`export CPPFLAGS="-I/usr/local/opt/openssl/include"`
Then (I have already installed MySQL from Homebrew)
`sudo gem install mysql2 -v '0.5.3' --source 'https://rubygems.org/' -- --with-cppflags=-I/usr/local/opt/openssl/include --with-ldflags=-L/usr/local/opt/openssl/lib`
Then run bundle
`bundle install`

### What actually happened?
It then gives me the error message:

Errno::EACCES: Permission denied @ rb_sysopen - /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/mysql2-0.5.3/CHANGELOG.md
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `initialize'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `open'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `block (2 levels) in extract_tar_gz'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package/tar_reader.rb:65:in `each'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:405:in `block in extract_tar_gz'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:534:in `block in open_tar_gz'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:531:in `wrap'
  /Users/jiataowu/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:531:in `open_tar_gz'
[...]

Issue #4888, opened 2021-08-30
Permission issue when installing nio4r gem


cguess:
### Describe the problem as clearly as you can

Attempting to install nio4r throws a permission error when trying to access workflows.

### Post steps to reproduce the problem

---
MacOS 12 Beta (latest)
MacBook Pro M1
Ruby 3.0.2 via rbenv
---

`bundle install`  when setting up a Rails system.

### Which command did you run?

I ran the command `/Users/username/.rbenv/versions/3.0.2/bin/bundle install`

### What were you expecting to happen?

Properly install, which works on MacOS 11, Windows 10 Linux subsystem and Raspberry Pi OS (latest)

### What actually happened?

Errno::EACCES: Permission denied @ rb_sysopen - /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/nio4r-2.5.8/.github/workflows/workflow.yml
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `initialize'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `open'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:424:in `block (2 levels) in extract_tar_gz'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package/tar_reader.rb:65:in `each'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:405:in `block in extract_tar_gz'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:534:in `block in open_tar_gz'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:531:in `wrap'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:531:in `open_tar_gz'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:404:in `extract_tar_gz'
  /Users/cwg15/.rbenv/versions/3.0.2/lib/ruby/3.0.0/rubygems/package.rb:384:in `block (2 levels) in extract_files'
[...]

PR #4965, opened 2021-10-07
Give better errors for some permission issues

deivid-rodriguez:

Sometimes bundler fails to do what it needs to do regarding some file operations like creating the gem folder where a gem will be installed, or deleting the previous installation. In some of those cases, bundler fails to report a proper error and instead prints a bug report template with a different error at random places.

The fix is to properly detect where the actual problem happens, and give a proper error.

Closes https://github.com/rubygems/rubygems/issues/4888.
Closes https://github.com/rubygems/rubygems/issues/4938.
Closes https://github.com/rubygems/rubygems/issues/4939.
Closes https://github.com/rubygems/rubygems/issues/4981.

### DIFF
--- a/bundler/lib/bundler/errors.rb
+++ b/bundler/lib/bundler/errors.rb
@@ -79,6 +79,10 @@ module Bundler
       case @permission_type
       when :create
         "executable permissions for all parent directories and write permissions for `#{parent_folder}`"
+      when :delete
+        permissions = "executable permissions for all parent directories and write permissions for `#{parent_folder}`"
+        permissions += ", and the same thing for all subdirectories inside #{@path}" if File.directory?(@path)
+        permissions
       else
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
This a weird MacOS failure (only happens on 2.5, 2.6, and 2.7, but not on 3.0) that I need to investigate. But in addition to that, I'm going to tweak the fix.

I think it's fine that bundler tries to automatically fix permissions so that it's able to delete an existing gem folder before reinstalling a gem. But this shouldn't be the default `FileUtils.rm_rf` behaviour, since it should just be a wrapper on top of `rm -rf` and not deviate from the standard shell behavior.

So for now, I won't be changing `fileutils` at all. Instead, I will check whether the folder has actually been removed after `FileUtils.rm_rf` and if not, retry with fixed permissions.

PR #4979, opened 2021-10-10
Avoid loading the `digest` gem unnecessarily

deivid-rodriguez:
OpenSSL includes what we need.

Some double loading issues related to the `digest` gem. See comments at https://github.com/ruby/digest/commit/16172612d56ac42f57e5788465791329303ac5d0.

Use `OpenSSL::Digest` instead.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -274,10 +274,9 @@ namespace 'blog' do
   checksums = ''
 
   task 'checksums' => 'package' do
-    require 'digest'
     require 'net/http'
     Dir['pkg/*{tgz,zip,gem}'].each do |file|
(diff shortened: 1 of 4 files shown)


voxik:
Just FTR, Ruby still enables build with disabled OpenSSL support. I wonder what is the impact? I thought that although OpenSSL was disabled, the Digest would be still available.

Also, if RubyGems stance was that OpenSSL is mandatory, then the code should be probably audited, because I guess there is more conditionals in test suite and what not.

deivid-rodriguez:
I believe any parts of rubygems using `digest` require `openssl` anyways, so I don't think this should have any unintended impact on those openssl-less usages of `rubygems`.

PR #4983, opened 2021-10-13
Keep loaded features

nobu:

As `save_loaded_features` removes the rest libraries other than this test directory, ordinary libraries loaded from files under rubygems also removed, and often causes constant redefinition warnings.

Now `$LOADED_FEATURES` list is being maintained by `setup` and `teardown` and, only libaries under the temporary directory will be removed.


deivid-rodriguez:
Thanks @nobu!

PR #4989, opened 2021-10-15
Add a custom SHA1 digest implementation to no longer depend on the digest gem before we know which version to activate

casperisfine:
Ref: https://bugs.ruby-lang.org/issues/17873

Ruby 3.1. is making `digest` a default gem, because of this if bundler
depends on it, it might cause already activated issues.

This currently happens if your Gemfile contains a git gem, because
the GitSource makes an SHA1 of the repository URL.

Currently it is very complicated if not impossible to use a gem that has `digest` in its dependencies if you also have a gitst gem.

```ruby
source "https://rubygems.org"

gem "net-imap"
gem "bootsnap", github: "Shopify/bootsnap"
```

```
$ ruby -v 
ruby 3.1.0dev (2021-10-06T06:42:37Z master d53493715c) [x86_64-darwin20]
$ bundle
Using msgpack 1.4.2
Using bootsnap 1.9.1 from https://github.com/Shopify/bootsnap.git
Using bundler 2.2.25
Using digest 3.0.0
Using io-wait 0.1.0
Using timeout 0.2.0
Using net-protocol 0.1.1
Using strscan 3.0.0
Using net-imap 0.2.2
Bundle complete! 2 Gemfile dependencies, 9 gems now installed.
$ bundle exec ruby -r bundler/setup -r digest -e 'p :hello'
gems/bundler-2.2.25/lib/bundler/runtime.rb:309:in `check_for_activated_spec!': You have already activated digest 3.1.0.pre2, but your Gemfile requires digest 3.0.0. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
	from gems/bundler-2.2.25/lib/bundler/runtime.rb:25:in `block in setup'
	from gems/bundler-2.2.25/lib/bundler/spec_set.rb:136:in `each'
	from gems/bundler-2.2.25/lib/bundler/spec_set.rb:136:in `each'
	from gems/bundler-2.2.25/lib/bundler/runtime.rb:24:in `map'
	from gems/bundler-2.2.25/lib/bundler/runtime.rb:24:in `setup'
	from gems/bundler-2.2.25/lib/bundler.rb:149:in `setup'
	from gems/bundler-2.2.25/lib/bundler/setup.rb:10:in `block in <top (required)>'
	from gems/bundler-2.2.25/lib/bundler/ui/shell.rb:136:in `with_level'
	from gems/bundler-2.2.25/lib/bundler/ui/shell.rb:88:in `silence'
	from gems/bundler-2.2.25/lib/bundler/setup.rb:10:in `<top (required)>'
	from <internal:/opt/rubies/3.1.0-dev/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from <internal:/opt/rubies/3.1.0-dev/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
```

I found https://github.com/Solistra/ruby-digest which is under public domain and could be vendored. It conveniently implement the 3 hash algorithms used by bundler. 

This is a rough first draft, it could certainly be cleaned a bit more. 

@deivid-rodriguez how does this sound?

### DIFF
--- a/bundler/lib/bundler/digest.rb
+++ b/bundler/lib/bundler/digest.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+# This code was extracted from https://github.com/Solistra/ruby-digest which is under public domain
+module Bundler
+  module Digest
+    # The initial constant values for the 32-bit constant words A, B, C, D, and
+    # E, respectively.
+    SHA1_WORDS = [0x67452301, 0xEFCDAB89, 0x98BADCFE, 0x10325476, 0xC3D2E1F0].freeze
+
+    # The 8-bit field used for bitwise `AND` masking. Defaults to `0xFFFFFFFF`.
+    SHA1_MASK = 0xFFFFFFFF
+
+    class << self
+      def sha1(string)
+        unless string.is_a?(String)
+          raise TypeError, "can't convert #{string.class.inspect} into String"
+        end
+
+        buffer = string.b
+
+        words = SHA1_WORDS.dup
+        generate_split_buffer(buffer) do |chunk|
+          w = []
+          chunk.each_slice(4) do |a, b, c, d|
+            w << (((a << 8 | b) << 8 | c) << 8 | d)
+          end
+          a, b, c, d, e = *words
+          (16..79).each do |i|
+            w[i] = SHA1_MASK & rotate((w[i-3] ^ w[i-8] ^ w[i-14] ^ w[i-16]), 1)
+          end
+          0.upto(79) do |i|
+            case i
+            when  0..19
+              f = ((b & c) | (~b & d))
+              k = 0x5A827999
+            when 20..39
+              f = (b ^ c ^ d)
+              k = 0x6ED9EBA1
+            when 40..59
+              f = ((b & c) | (b & d) | (c & d))
+              k = 0x8F1BBCDC
+            when 60..79
+              f = (b ^ c ^ d)
+              k = 0xCA62C1D6
+            end
+            t = SHA1_MASK & (SHA1_MASK & rotate(a, 5) + f + e + k + w[i])
+            a, b, c, d, e = t, a, SHA1_MASK & rotate(b, 30), c, d # rubocop:disable Style/ParallelAssignment
+          end
+          mutated = [a, b, c, d, e]
+          words.map!.with_index {|word, index| SHA1_MASK & (word + mutated[index]) }
+        end
+
+        words.pack("N*").unpack("H*").first
+      end
+
+      private
+
+      def generate_split_buffer(string, &block)
+        size   = string.bytesize * 8
+        buffer = string.bytes << 128
+        buffer << 0 while buffer.size % 64 != 56
+        [size].pack("Q").bytes.reverse_each {|b| buffer << b }
+        buffer.each_slice(64, &block)
+      end
+
+      def rotate(value, spaces)
+        value << spaces | value >> (32 - spaces)
+      end
+    end
+  end
+end
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -60,6 +60,7 @@ bundler/lib/bundler/dep_proxy.rb
 bundler/lib/bundler/dependency.rb
 bundler/lib/bundler/deployment.rb
 bundler/lib/bundler/deprecate.rb
+bundler/lib/bundler/digest.rb
 bundler/lib/bundler/dsl.rb
 bundler/lib/bundler/endpoint_specification.rb
 bundler/lib/bundler/env.rb
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -43,6 +43,7 @@ module Bundler
   autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
   autoload :DepProxy,               File.expand_path("bundler/dep_proxy", __dir__)
   autoload :Deprecate,              File.expand_path("bundler/deprecate", __dir__)
+  autoload :Digest,                 File.expand_path("bundler/digest", __dir__)
   autoload :Dsl,                    File.expand_path("bundler/dsl", __dir__)
   autoload :EndpointSpecification,  File.expand_path("bundler/endpoint_specification", __dir__)
   autoload :Env,                    File.expand_path("bundler/env", __dir__)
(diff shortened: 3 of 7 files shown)


deivid-rodriguez:
@casperisfine Yes, the explanation of the issue makes sense and the solution seems like the way to go.

The scary part is to start depending on a bunch of code from a random public repository. But as long as we properly audit the code, it should be fine. Did you do that?

As a side note, we could also workaround the issue by using `OpenSSL::Digest` instead of `Digest` directly, thanks to `openssl` not activating `digest` as a gem. But at https://github.com/ruby/digest/commit/16172612d56ac42f57e5788465791329303ac5d0#commitcomment-57698673 we agreed that was an `openssl` issue, which has already been fixed at https://github.com/ruby/openssl/pull/463. So it would still be an issue if ruby 3.1 ships an `openssl` version with that patch. And in any case, it would just trick bundler into not detecting the problem, but you would still get redefinition warnings and your application would be using two different copies of the gem at the same time. Well, actually, you might not even get any warnings due to https://github.com/ruby/digest/commit/16172612d56ac42f57e5788465791329303ac5d0, but the problem will still be there and I believe https://github.com/ruby/digest/commit/16172612d56ac42f57e5788465791329303ac5d0 should be reverted anyways.

simi:
Just to ensure I do understand it well, is this is the problematic code doing MD5 too early?

https://github.com/rubygems/rubygems/blob/e4dcb33f6d0f922f8551c9a82887e5b4323a6e24/bundler/lib/bundler/source/rubygems/remote.rb#L29

deivid-rodriguez:
Also, the particular use of hashing algorithms here (at leas the one that's causing trouble), it's not really crypto-related. It's used to digest the url for git repositories into something that can be used as a file name on disk. So I think it should pose no issue if the digest could be reversed into the original uri. We could even stop using it and do something different like replacing slashes with other characters that can be used in filenames if it wasn't for backwards compatibility concerns.

casperisfine:
What if the vendored code is in a namespace like `Bundler::CryptoUnsafeDigest`, would that solves the concerns of this ever be used in a crypto context?

byroot:
I suppose there's some corner cases in deployment mode, but yes I prefer that too. I'll do that then.

deivid-rodriguez:
Actually I think let's stick to the vendoring we were planning, since we know it's backwards compatible for sure. I want to better understand those edge cases before doing this kind of "hopefully backwards compatible" changes. My understanding is that it will be backwards compatible even in deployment mode, since deployment mode is only strict about the lockfile and the lockfile is not involved here. But I want to make sure and I'm not sure when I'll get to that and I don't want to block this.

Also, regarding the vendoring, since the library we're copying here is not maintained since years ago, it's not even packaged as a gem, and we'll be heavily simplifying it, I think it's better to not put it under `vendor` but somewhere else. With proper attribution, of course, but without commitment to contributing back any changes

simi:
@deivid-rodriguez that will fail to do offline install for valid bundles if I understand it well. Better to keep it compatible as much as possible.

casperisfine:
> Maybe it does not matter for this usage

It definitely doesn't. The use case here is to digest repository urls (e.g. `"https://github.com/rubygems/rubygems.git"`) with SHA1.

casperisfine:
@simi yes, I expect CI to go green now.

casperisfine:
I'm sorry to be that guy, but 3.1.0-preview1 is around the corner (likely some time next week), it would be nice if people who test it wouldn't run into that bug.

simi:
@casperisfine Since RubyGems is crucial part of Ruby distribution itself, I think it should be possible to wait with release to not release broken version.

bundler/lib/bundler/digest.rb
```diff
@@ -0,0 +1,71 @@
...
+              f = (b ^ c ^ d)
+              k = 0x6ED9EBA1
+            when 40..59
+              f = ((b & c) | (b & d) | (c & d))
+              k = 0x8F1BBCDC
+            when 60..79
+              f = (b ^ c ^ d)
+              k = 0xCA62C1D6
+            end
+            t = SHA1_MASK & (SHA1_MASK & rotate(a, 5) + f + e + k + w[i])
```
nobu, line 46: As `+` is higher precedence than `&`, this equals `SHA1_MASK & (SHA1_MASK & (rotate(a, 5) + f + e + k + w[I]))`.
I don't think bit-and with the same value twice has meaning.
Maybe `SHA1_MASK & ((SHA1_MASK & rotate(a, 5)) + f + e + k + w[I])` is intended?

simi: [rolling own crypto is hard](https://github.com/rubygems/rubygems/pull/4989#issuecomment-944224157) :) Anyway isn't this used mostly for caching? There could be problem, but not visible to clients (except collision case).

deivid-rodriguez: Yes, sure, what I'm saying is that it seems better to not change the behaviour without fully understanding the bug. If @nobu can provide a string incorrectly digested with the current code, then of course we should fix it.

byroot: We have a test suite that ensure this vendored version produce the same hashes than the stdlib digest. If we had a bug here we'd have widely different output.

It's safe to assume there's no bug, just a small ineficiency.

simi: @byroot this is the CI covering this.

```ruby
["foo", "skfjsdlkfjsdf", "3924m", "ldskfj"].each do |payload|
  expect(subject.sha1(payload)).to be == stdlib.hexdigest(payload)
end
```

In theory it is still possible some other strings could be hashed wrongly. Indeed if @nobu can provide one, we need to fix this.

---

As you can see rolling own crypto and maintaining it is hard. I still think merging this was not good idea (even it is great work @byroot :heart:, I'm not complaining about the code itself).

byroot: The above test is enough to cover that code, hash functions have widely different outputs.

Also it's not hard to understand that `a & a == a`. Let's not be dogmatic here.

> I still think merging this was not good idea 

What else would have you done? `digest` was gemified, by design `bundler` can't rely on gems before the full gemfile is activated.

We talked about using something else than SHA1 for this, and IIRC it's not even for crypto but just mangling, but we ruled it out as it would have broken backward compatibility.

deivid-rodriguez: We needed a solution for a problem, @byroot came up with this one, and this works well. I don't see the issue here, to. be honest. Let's add this simplification and move on, no?

deivid-rodriguez: https://github.com/rubygems/rubygems/pull/5595

simi: > Let's add this simplification and move on, no?

Sure, I'll check if there is any way to move out of MD5 (and digest at all) and we can start slowly deprecate that somehow. :pray:

PR #5003, opened 2021-10-21
Re-enable `default_ignores` option for standard

searls:
I am not sure why this flag was turned off (it wasn't explained in my commit message in 0365dc852767ae589376a7aad1fb129738e408b0 or in my PR in #4411). 

Whatever the reason, without `default_ignores` turned on, most default CI configurations will immediately fail, as they most likely vendor and cache their dependencies under `vendor`, which will cause standard to run against all the vendored gems and (most likely) fail. I think we should remove this before this feature is released.


deivid-rodriguez:
@searls I'm unsure why but I missed to release this change with the previous release... I'll release it tomorrow, sorry about that.

PR #5009, opened 2021-10-24
Use latest simplecov on supported rubies.

simi:
IMHO using branching in Gemfile is not the best practice (since lockfile doesn't reflect that), but this is only for really old Rubies and during development it should not cause any harm. WDYT?

simplecov 0.18+ supports only Ruby >= 2.5. Sadly ruby-trunk is not compatible with simplecov 0.17.


deivid-rodriguez:
@simi I'm going to update this PR myself because sometimes I run into an error related to using old simplecov and it's a bit annoying 😅, ok?

simi:
@deivid-rodriguez updated

deivid-rodriguez (approved):
Thank you @simi!

PR #5010, opened 2021-10-25
Warn on empty or open required_ruby_version specification attribute.

simi:
This is just concept what we can do to improve situation of filled in `required_ruby_version` value to prevent problems described in https://github.com/rubygems/rfcs/issues/34.

I have used placeholder text for now. Feel free to suggest better. We can also use different text for different situations.

:information_source: `required_ruby_version` is already part of `bundle gem` template based on this logic.

https://github.com/rubygems/rubygems/blob/2f48d60afe050862258212d0c68530f39fb5fdfa/bundler/lib/bundler/cli/gem.rb#L407-L414

Later on we can move this attribute to required. It could be also possible to warn on "potentially misused values". For example when newer Ruby is used to build gem (then the minimal), we can inform user to ensure it is good idea to check if gem is still compatible with minimal Ruby version.

related to https://github.com/rubygems/guides/pull/295, https://github.com/rubygems/rfcs/issues/34 and https://github.com/rubygems/rfcs/pull/26

/cc @marcandre

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -103,6 +103,8 @@ class Gem::SpecificationPolicy
 
     validate_dependencies
 
+    validate_required_ruby_version
+
     validate_extensions
 
     validate_removed_attributes
@@ -224,6 +226,12 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
     end
     if warning_messages.any?
       warning_messages.each {|warning_message| warning warning_message }
+    end
+  end
+
+  def validate_required_ruby_version
+    if @specification.required_ruby_version.requirements == [Gem::Requirement::DefaultRequirement]
+      warning "make sure you specify the oldest ruby version constraint (like \">= 3.0\") that you want your gem to support by setting the `required_ruby_version` gemspec attribute"
     end
   end
 
--- a/test/rubygems/specifications/rubyforge-0.0.1.gemspec
+++ b/test/rubygems/specifications/rubyforge-0.0.1.gemspec
@@ -1,14 +1,15 @@
 # frozen_string_literal: true
 
 Gem::Specification.new do |s|
-  s.name              = "rubyforge"
-  s.version           = "0.0.1"
-  s.platform          = "ruby"
-  s.require_paths     = ["lib"]
-  s.summary           = "A very bar gem"
-  s.authors           = ["unknown"]
-  s.license           = "MIT"
-  s.homepage          = "http://example.com"
-  s.files             = ["README.md"]
-  s.rubyforge_project = "abc"
+  s.name                  = "rubyforge"
+  s.version               = "0.0.1"
+  s.platform              = "ruby"
+  s.require_paths         = ["lib"]
+  s.summary               = "A very bar gem"
+  s.authors               = ["unknown"]
+  s.license               = "MIT"
+  s.homepage              = "http://example.com"
+  s.files                 = ["README.md"]
+  s.rubyforge_project     = "abc"
+  s.required_ruby_version = ">= 1.9.3"
 end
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
What's left to do here?

simi:
Warning text is currently just a placeholder. Feel free to suggest better.

lib/rubygems/specification_policy.rb
```diff
@@ -219,6 +221,15 @@ def validate_dependencies # :nodoc:
     end
   end
 
+  def validate_required_ruby_version
+    op, dep_version = @specification.required_ruby_version.requirements.first
+    segments = dep_version.segments
+
+    if (op == '>' || op == '>=') && segments == [0]
+      warning "please specify minimal required ruby for your RubyGem"
```
deivid-rodriguez, line 229: I would go with something like

```suggestion
      warning "make sure you specify the oldest ruby version that you want your gem to support by setting the `required_ruby_version` gemspec attribute
```

lib/rubygems/specification_policy.rb
```diff
@@ -219,6 +221,15 @@ def validate_dependencies # :nodoc:
     end
   end
 
+  def validate_required_ruby_version
+    op, dep_version = @specification.required_ruby_version.requirements.first
+    segments = dep_version.segments
+
+    if (op == '>' || op == '>=') && segments == [0]
```
doodzik, line 228: I have a hard time understanding what is happening in this conditional. Does the `segments == [0]` check if the version hasn't been set yet? I wonder if moving the check into a variable with a descriptive name would make it a bit easier to follow.

deivid-rodriguez: Indeed!

This is cleaner

```diff
diff --git a/lib/rubygems/specification_policy.rb b/lib/rubygems/specification_policy.rb
index a23a363e9f..b30665b9d8 100644
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -222,10 +222,7 @@ def validate_dependencies # :nodoc:
   end
 
   def validate_required_ruby_version
-    op, dep_version = @specification.required_ruby_version.requirements.first
-    segments = dep_version.segments
-
-    if (op == '>' || op == '>=') && segments == [0]
+    if @specification.required_ruby_version.requirements == [Gem::Requirement::DefaultRequirement]
       warning "please specify minimal required ruby for your RubyGem"
     end
   end
```

The same "obscure check" is done in other validations, so we can probably extract a method with the cleaner approach and use it everywhere.

simi: done

doodzik (commented):
LGTM

hsbt (approved):

hsbt:
@simi Can you merge this? I'd like to merge for Ruby 3.3.0-preview2.

simi:
I'm out for next two weeks. Feel free to merge.

Dne út 8. 8. 2023 12:20 uživatel Hiroshi SHIBATA ***@***.***>
napsal:

lib/rubygems/specification_policy.rb
```diff
@@ -224,6 +226,12 @@ def validate_dependencies # :nodoc:
     end
   end
 
+  def validate_required_ruby_version
+    if @specification.required_ruby_version.requirements == [Gem::Requirement::DefaultRequirement]
+      warning "make sure you specify the oldest ruby version that you want your gem to support by setting the `required_ruby_version` gemspec attribute"
```
deivid-rodriguez, line 231: I wonder if this wording could lead to people setting `required_ruby_version = "<min_supported_version>"` instead of ``required_ruby_version = ">= <min_supported_version>"`? And that would actually prevent installation on any rubies, exception for the minimum supported one? 🤔

Should we clarify how it's supposed to be set?

simi:
rebased @hsbt

lib/rubygems/specification_policy.rb
simi: I have updated the message to make it clear.

hsbt (approved):
@simi Thank you!

Issue #3260, opened 2020-01-18
Drop support for ruby 2.3


deivid-rodriguez:
I'd like to drop support for ruby 2.3, maybe in the upcoming 2.2 release, maybe right after.

deivid-rodriguez:
This was fixed by #6107 🎉

hsbt:
🎉

PR #5021, opened 2021-10-26
Add require parameter to `bundle add`

### LINKED ISSUE #5017: Feature request: Set `require` parameter in `bundle add`

When calling `bundle add`, there's no option to set the `require` parameter. For example, when adding `view_component` to a project, it's best to declare it in the Gemfile as `gem 'view_component', require: 'view_component/engine'`. For the majority of cases, it may also be ideal to set `require` to `false`.

Yes.

boardfish:

`bundle add` does not allow setting the `require` parameter in a gem declaration. #5017

Adds an option to the CLI, and updates the part of the code that writes the gem declaration to the Gemfile to include the `require` parameter.

Closes #5017.

I can't find tests for the `CLI` and `Injector` classes.

TODO:

- [x] Ensure that if a boolean value is passed, it isn't surrounded by quotes

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -367,6 +367,7 @@ module Bundler
     method_option "version", :aliases => "-v", :type => :string
     method_option "group", :aliases => "-g", :type => :string
     method_option "source", :aliases => "-s", :type => :string
+    method_option "require", :aliases => "-r", :type => :string, :banner => "Adds require path to gem. Provide false, or a path as a string."
     method_option "git", :type => :string
     method_option "branch", :type => :string
     method_option "skip-install", :type => :boolean, :banner =>
--- a/bundler/lib/bundler/injector.rb
+++ b/bundler/lib/bundler/injector.rb
@@ -113,8 +113,9 @@ module Bundler
         source = ", :source => \"#{d.source}\"" unless d.source.nil?
         git = ", :git => \"#{d.git}\"" unless d.git.nil?
         branch = ", :branch => \"#{d.branch}\"" unless d.branch.nil?
+        require_path = ", :require => #{convert_autorequire(d.autorequire)}" unless d.autorequire.nil?
 
-        %(gem #{name}#{requirement}#{group}#{source}#{git}#{branch})
+        %(gem #{name}#{requirement}#{group}#{source}#{git}#{branch}#{require_path})
       end.join("\n")
     end
 
@@ -268,6 +269,12 @@ module Bundler
 
     def show_warning(message)
       Bundler.ui.info Bundler.ui.add_color(message, :yellow)
+    end
+
+    def convert_autorequire(autorequire)
+      autorequire = autorequire.first
+      return autorequire if autorequire == "false"
+      autorequire.inspect
     end
   end
 end
--- a/bundler/spec/commands/add_spec.rb
+++ b/bundler/spec/commands/add_spec.rb
@@ -68,6 +68,18 @@ RSpec.describe "bundle add" do
     end
   end
 
+  describe "with --require" do
+    it "adds the require param for the gem" do
+      bundle "add 'foo' --require=foo/engine"
+      expect(bundled_app_gemfile.read).to match(%r{gem "foo",(?: .*,) :require => "foo\/engine"})
+    end
+
+    it "converts false to a boolean" do
+      bundle "add 'foo' --require=false"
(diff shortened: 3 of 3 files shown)


boardfish:
Thanks for the guidance on where the specs should live. I've written some and made sure everything's covered.

I also realized `Dependency` is already aware of an autorequire, so I figured it best to use that.

bundler/lib/bundler/injector.rb
```diff
@@ -269,5 +270,13 @@ def cross_check_for_errors(gemfile_path, original_deps, removed_deps, initial_ge
     def show_warning(message)
       Bundler.ui.info Bundler.ui.add_color(message, :yellow)
     end
+
+    def convert_autorequire(autorequire)
+      require 'json'
+      autorequire
+        .yield_self { |r| (r.one? ? r.first : r) }   # Unwrap array
+        .yield_self { |r| r == "false" ? false : r } # Convert false to boolean
```
doodzik, line 278: `yield_self` has been [introduced in ruby 2.5 ](https://www.ruby-lang.org/en/news/2017/12/25/ruby-2-5-0-released/) and [bundler currently supports ruby 2.3 and up](https://github.com/rubygems/rubygems/issues/3260). So I think it is best to change the code to support ruby 2.3.

boardfish:
This is ready to be reviewed again @doodzik. Thanks for checking it out!

bundler/lib/bundler/dependency.rb
```diff
@@ -7,7 +7,7 @@
 module Bundler
   class Dependency < Gem::Dependency
     attr_reader :autorequire
-    attr_reader :groups, :platforms, :gemfile, :git, :branch
+    attr_reader :groups, :platforms, :gemfile, :git, :branch, :autorequire
```
deivid-rodriguez, line 10: This is already defined in the line above, so it's causing method redefinition warnings that make some tests fail.

deivid-rodriguez (changes requested):

boardfish:
@deivid-rodriguez I've removed it now - hopefully the workflows should pass now.

bundler/lib/bundler/cli.rb
```diff
@@ -366,6 +366,7 @@ def binstubs(*gems)
     method_option "version", :aliases => "-v", :type => :string
     method_option "group", :aliases => "-g", :type => :string
     method_option "source", :aliases => "-s", :type => :string
+    method_option "require", :aliases => "-r", :type => :string
```
simi, line 369: Would it be possible to add banner with info about possible values (path or `false`)?

bundler/lib/bundler/injector.rb
```diff
@@ -269,5 +270,12 @@ def cross_check_for_errors(gemfile_path, original_deps, removed_deps, initial_ge
     def show_warning(message)
       Bundler.ui.info Bundler.ui.add_color(message, :yellow)
     end
+
+    def convert_autorequire(autorequire)
+      require 'json'
+      autorequire = autorequire.one? ? autorequire.first : autorequire # Unwrap array
+      autorequire = false if autorequire == "false"                    # Convert false to boolean
+      autorequire.to_json                                              # Wrap in quotes if necessary
```
deivid-rodriguez, line 278: I think we don't really need `json` here. Would this be enough?

```ruby
    def convert_autorequire(autorequire)
      autorequire = autorequire.first
      return autorequire if autorequire == "false"
      autorequire.inspect
    end
```

boardfish: It passes the tests, so it'd seem so 👍 

The ethos of this project has been pretty interesting relative to other stuff I've worked on. Things like being quite slim on dependencies and ensuring support for all versions have meant I've had to think a little differently, and that's refreshing!

boardfish:
Next round of feedback has been resolved 👍

simi (approved):

deivid-rodriguez:
Looks good to me! Since the PR has stood there for a while, can you rebase just to make sure it's all good with latest master? And since you're at it, can you squash everything to one commit?

boardfish:
All sorted!

deivid-rodriguez (approved):
Thanks!

boardfish:
@deivid-rodriguez The auto-merge hasn't gone through as I'm not authorized to do so. Could you merge this in please?

deivid-rodriguez:
Oh, no worries, it's because one CI check is still running. It'll get automerged in a bit.

Issue #4694, opened 2021-06-23
Bundler dependency confusion issues not resolved?


pvalena:
### Describe the problem as clearly as you can

Following up on my successful test backport (#4687), I've tried to run the test[*] with Bundler 2.2.20 and Bundler 2.2.21 (but maybe in not in a very smart way). The test fails, although it was supposed to be fixed.

This was prompted with a failing shell [reproducer](https://gist.github.com/pvalena/d6113818216902ea1c5f3d428928df66), from one of my collegues, which AFAICT is doing exactly the same thing, as in the Bundler test, but it is failing with Bundler 2.2.18+.

[*] Test name: bundle install with gems on multiple sources with source affinity when a top-level gem has an indirect dependency and the dependency is in both the top-level and a pinned source installs the dependency from the top-level source without warning

### Post steps to reproduce the problem

I've taken Ruby 2.7.3 sources, and replaced the Bundler in them with Bundler 2.20.21. Then I've run the backported test: 

https://gist.github.com/08e9423766b83a8fb0e5e1a64fa4ca74

### Additional

Is there something I'm doing wrong, or is the test really failing?

I've run the test test  from rubygems git repository as well, and it did succeed (with Ruby 2.7), although I had to do a [workaround](https://git.io/JnHHo) to make it work.

deivid-rodriguez:
I'll investigate tomorrow. In principle it seems like your reproducer indeed suffers from the issue. I think it's related to the gems being served by servers using the old marshalled index, instead of more modern dependency APIs. In that case, the issue is still present because, indeed.

Buju77:
I'm not sure if my issue is related to yours, but starting with bundler 2.2.18 and up, our `bundle update` suddenly fails, while 2.2.17 works perfectly fine.

we are getting dependency issues like:

```
Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.

Bundler could not find compatible versions for gem "google-api-client":
  In snapshot (Gemfile.lock):
    google-api-client (>= 0.38.0)

  In Gemfile:
    google-api-client (~> 0.38)

    fastlane (~> 2.150) was resolved to 2.181.0, which depends on
      google-api-client (>= 0.37.0, < 0.39.0)
```

All 3 different version requirements looks fine for me and should match 0.38.0. But why does bundler 2.2.18+ now suddenly throw an error?

deivid-rodriguez:
[...]

PR #5029, opened 2021-10-27
Print a warning for a potential confusion from the indirect dependencies.

junaruga:
Print warning when a confusion by the indirect dependencies can happen.

This PR is a fix for the https://github.com/rubygems/rubygems/issues/4694#issuecomment-871371684 (a possible dependency confusion from indirect dependencies to print a warning message.

I added the logic to print a warning in the case.

This PR is working in progress. I will prepare the reproducer using `gem server` (non dependency API) soon. I have some questions before writing the tests.

### Q1

I have not implemented the following logic to check the size of the index. Could you tell me how to check the size?

> I guess this is not too bad if the index is small, so we could allow that and fix the issue up to a certain index size, and print a warning otherwise?

### Q2

What do you think about adding the following code to raise an error as a default, and print a warning if the `bundle config` key `allow_dependency_confusion` (`BUNDLE_ALLOW_DEPENDENCY_CONFUSION` environment variable) is available as a compatibility?  Because maybe we want to stop the use case that triggers [the security issue](https://bundler.io/blog/2021/02/15/a-more-secure-bundler-we-fixed-our-source-priorities.html).

```
$ diff --git a/bundler/lib/bundler/definition.rb b/bundler/lib/bundler/definition.rb
index 7414a5e7d4..d8dddb9f59 100644
@@ -537,6 +537,9 @@ def precompute_source_requirements_for_indirect_dependencies?
         msg << "Your bundle cannot resolve indirect dependencies because "
         msg << reasons.join(" and ")
         msg << "."
+
+        raise ProductionError, msg unless Bundler.settings[:allow_dependency_confusion]
+
         Bundler.ui.warn msg
       end
```

### Q3

The added logic is
in `lib/bundler/definition.rb#precompute_source_requirements_for_indirect_dependencies?`
in `lib/bundler/definition.rb#source_requirements`
in `lib/bundler/definition.rb#resolve`
in `lib/bundler/definition.rb#resolve_remotely!`

However technically after running `sources.remote!` in `resolve_remotely!`, the Bundler is ready to print the warning (or raise an error) referring the `sources` object. So, I wondered if we could add the new logic to print the warning after `sources.remote!` in the `resolve_remotely!` instead of `source_requirements`. And I think the logic is the remote flow specific (`@remote == true` ). What do you think?

```
    def resolve_remotely!
      @remote = true
      sources.remote!

      # Add a new logic here? 

      resolve
    end
```

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -777,8 +777,6 @@ module Bundler
     end
 
     def precompute_source_requirements_for_indirect_dependencies?
-      return false unless @remote && !sources.aggregate_global_source?
-
       if sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
         true
       else
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -777,7 +777,27 @@ module Bundler
     end
 
     def precompute_source_requirements_for_indirect_dependencies?
-      sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
+      return false unless @remote && !sources.aggregate_global_source?
+
+      if sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
+        true
+      else
+        non_dependency_api_warning
+        false
+      end
+    end
+
+    def non_dependency_api_warning
+      non_api_sources = sources.non_global_rubygems_sources.reject(&:dependency_api_available?)
+      non_api_source_names = non_api_sources.map {|d| "  * #{d}" }.join("\n")
+
+      msg = String.new
+      msg << "Your Gemfile contains scoped sources that don't implement a dependency API, namely:\n\n"
+      msg << non_api_source_names
+      msg << "\n\nUsing the above gem servers may result in installing unexpected gems. " \
+        "To resolve this warning, make sure you use gem servers that implement dependency APIs, " \
+        "such as gemstash or geminabox gem servers."
+      Bundler.ui.warn msg
     end
 
     def current_platform_locked?
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -304,8 +304,8 @@ RSpec.describe Bundler::Definition do
     context "when all the scoped sources implement a dependency API" do
       let(:non_global_rubygems_sources) do
         [
-          double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
-          double("non-global-source-1", :dependency_api_available? => true, :to_s => "b"),
+          double("non-global-source-0", "dependency_api_available?":true, to_s:"a"),
+          double("non-global-source-1", "dependency_api_available?":true, to_s:"b"),
         ]
       end
 
@@ -319,9 +319,9 @@ RSpec.describe Bundler::Definition do
     context "when some scoped sources do not implement a dependency API" do
       let(:non_global_rubygems_sources) do
         [
-          double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
-          double("non-global-source-1", :dependency_api_available? => false, :to_s => "b"),
-          double("non-global-source-2", :dependency_api_available? => false, :to_s => "c"),
+          double("non-global-source-0", "dependency_api_available?":true, to_s:"a"),
+          double("non-global-source-1", "dependency_api_available?":false, to_s:"b"),
+          double("non-global-source-2", "dependency_api_available?":false, to_s:"c"),
         ]
       end
 
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -289,6 +289,61 @@ RSpec.describe Bundler::Definition do
     end
   end
 
+  describe "#precompute_source_requirements_for_indirect_dependencies?" do
+    before do
+      allow(Bundler::SharedHelpers).to receive(:find_gemfile) { Pathname.new("Gemfile") }
+    end
+
+    let(:sources) { Bundler::SourceList.new }
+    subject { Bundler::Definition.new(nil, [], sources, []) }
+
+    context "when remote and does not have multiple global sources" do
+      before do
+        subject.instance_variable_set(:@remote, true)
+        allow(sources).to receive(:aggregate_global_source?).and_return(false)
+        allow(sources).to receive(:non_global_rubygems_sources).and_return(non_global_rubygems_sources)
+      end
+
+      context "when all the scoped sources contain a dependency API" do
+        let(:non_global_rubygems_sources) do
+          [
+            double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
+            double("non-global-source-1", :dependency_api_available? => true, :to_s => "b"),
+          ]
+        end
+
+        it "will not raise a warning" do
+          expect(subject).not_to receive(:non_dependency_api_warning)
+
+          expect(subject.send(:precompute_source_requirements_for_indirect_dependencies?)).to be_truthy
+        end
+      end
+
+      context "when scoped sources do not contain a dependency API" do
+        let(:non_global_rubygems_sources) do
+          [
+            double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
+            double("non-global-source-1", :dependency_api_available? => false, :to_s => "b"),
+            double("non-global-source-2", :dependency_api_available? => false, :to_s => "c"),
+          ]
+        end
+
+        it "will raise a warning" do
+          expect(Bundler.ui).to receive(:warn).with(<<-W.strip)
+Your Gemfile contains scoped sources that don't implement a dependency API, namely:
+
+  * b
+  * c
+
+Using the above gem servers may result in installing unexpected gems. To resolve this warning, make sure you use gem servers that implement dependency APIs, such as gemstash or geminabox gem servers.
+          W
+
+          expect(subject.send(:precompute_source_requirements_for_indirect_dependencies?)).to be_falsy
+        end
+      end
+    end
+  end
+
   def mock_source_list
     Class.new do
       def all_sources
--- a/spec/bundler/definition_spec.rb
+++ b/spec/bundler/definition_spec.rb
@@ -297,49 +297,45 @@ RSpec.describe Bundler::Definition do
     let(:sources) { Bundler::SourceList.new }
     subject { Bundler::Definition.new(nil, [], sources, []) }
 
-    context "when remote and does not have multiple global sources" do
-      before do
-        subject.instance_variable_set(:@remote, true)
-        allow(sources).to receive(:aggregate_global_source?).and_return(false)
-        allow(sources).to receive(:non_global_rubygems_sources).and_return(non_global_rubygems_sources)
+    before do
+      allow(sources).to receive(:non_global_rubygems_sources).and_return(non_global_rubygems_sources)
+    end
+
+    context "when all the scoped sources implement a dependency API" do
+      let(:non_global_rubygems_sources) do
+        [
+          double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
+          double("non-global-source-1", :dependency_api_available? => true, :to_s => "b"),
+        ]
       end
 
-      context "when all the scoped sources contain a dependency API" do
-        let(:non_global_rubygems_sources) do
-          [
-            double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
-            double("non-global-source-1", :dependency_api_available? => true, :to_s => "b"),
-          ]
-        end
+      it "returns true without warning" do
+        expect(subject).not_to receive(:non_dependency_api_warning)
 
-        it "will not raise a warning" do
-          expect(subject).not_to receive(:non_dependency_api_warning)
+        expect(subject.send(:precompute_source_requirements_for_indirect_dependencies?)).to be_truthy
+      end
+    end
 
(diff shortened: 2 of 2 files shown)


junaruga:
@deivid-rodriguez 

Regarding Q1, Q2, OK. I can agree with you.

Regarding Q3, what methods (the old, insecure one, or the new one) are you talking about? Are those in `definition.rb`?

> we need to know which method we should use to resolve sources (the old, insecure one, or the new one).

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,30 @@ def materialize(dependencies)
...
+      reasons = []
+
+      if @remote
+        unless sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
+          reasons << "any of the involved gem servers does not implement a dependency API"
+          precompute = false
+        end
+
+        if sources.aggregate_global_source?
+          reasons << "the Gemfile has multiple global sources"
```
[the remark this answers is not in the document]
junaruga: OK. It seems we already have 2 warnings on the master branch in the case.

https://github.com/rubygems/rubygems/blob/4411e77290d8d7965b08080353642cb168ced006/bundler/lib/bundler/dsl.rb#L471-L474

https://github.com/rubygems/rubygems/blob/4411e77290d8d7965b08080353642cb168ced006/bundler/lib/bundler/cli/install.rb#L196

---

On the master commit: `4411e77290d8d7965b08080353642cb168ced006`, I executed the `ruby bundler/spec/support/bundle.rb install`, modifying the `Gemfile` into the following one in my reproducer.

```
$ cat Gemfile
source "http://127.0.0.1:8801"
source "http://127.0.0.1:8802"

gem "a"
```

I see 2 warnings.

```
+ ruby /home/jaruga/git/rubygems/bundler/spec/support/bundle.rb install
[DEPRECATED] Your Gemfile contains multiple primary sources. Using `source` more than once without a block is a security risk, and may result in installing unexpected gems. To resolve this warning, use a block to indicate which gems should come from the secondary source.
Fetching source index from http://127.0.0.1:8802/
Fetching source index from http://127.0.0.1:8801/
Resolving dependencies...
Using bundler 2.3.0.dev
Fetching a 0.0.2
Installing a 0.0.2
Bundle complete! 1 Gemfile dependency, 2 gems now installed.
Bundled gems are installed into `./app`
Warning: the gem 'a' was found in multiple sources.
Installed from: http://127.0.0.1:8801/
Also found in:
  * http://127.0.0.1:8802/
You should add a source requirement to restrict this gem to your preferred source.
For example:
    gem 'a', :source => 'http://127.0.0.1:8801/'
Then uninstall the gem 'a' (or delete all bundled gems) and then install again.
```

junaruga:
Regarding Q3, I understood it now, and I agree with you.

> Essentially, which branch here we take:

bundler/lib/bundler/definition.rb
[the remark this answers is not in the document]
deivid-rodriguez: Indeed, the first warning is generic about the deprecation of multiple Gemfiles sources. This new warning would be equivalent two that. The last warning lets you know that there's a potential dependency confusion in this particular case. So if you have a Gemfile with multiple sources but with no ambiguity on any indirect dependencies, you still get the first warning, but not the second.

I think it's fine as is, but we don't need a third warning.

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,30 @@ def materialize(dependencies)
     end
 
     def precompute_source_requirements_for_indirect_dependencies?
-      @remote && sources.non_global_rubygems_sources.all?(&:dependency_api_available?) && !sources.aggregate_global_source?
+      precompute = true
+      reasons = []
+
+      if @remote
+        unless sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
+          reasons << "any of the involved gem servers does not implement a dependency API"
```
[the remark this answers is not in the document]
junaruga: OK. I will change the message printing the servers not implementing dependency APIs. I would like to use the "do not" rather than "don't", because the "do not" is more often used in the source code. It seems the "don't" is used in the source comment.

```
$ grep -r " do not " bundler/lib/ | wc -l
41

$ grep -r " don't " bundler/lib/ | wc -l
36
```

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,30 @@ def materialize(dependencies)
...
+
+        if sources.aggregate_global_source?
+          reasons << "the Gemfile has multiple global sources"
+          precompute = false
+        end
+      end
+
+      unless precompute
+        msg = String.new
+        msg << "Your bundle cannot resolve indirect dependencies because "
```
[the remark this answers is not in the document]
junaruga: OK. I will change it better.

junaruga:
@deivid-rodriguez I rebased this PR considering your review.

Now the warning message is like this.
```
Your Gemfile contains scoped sources rubygems repository http://127.0.0.1:8802/ or installed locally whose servers do not implement a dependency API. Using the `source` is a security risk, and may result in installing unexpected gems. To resolve this warning, use the sources whose servers implement the dependency API.
```

About how to warn and the content of the message, I referred warning methods in `dsl.rb` such as

https://github.com/rubygems/rubygems/blob/3613737ef061ddc6ca941a13b9c01478e0bb625d/bundler/lib/bundler/dsl.rb#L464-L476

What do you think?
Now I will try to write the tests as my next step.

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,25 @@ def materialize(dependencies)
...
+      precompute && !sources.aggregate_global_source?
+    end
+
+    def non_depeneency_api_warning
+      non_api_sources = sources.non_global_rubygems_sources.reject(&:dependency_api_available?)
+      non_api_source_names = non_api_sources.map(&:to_s).join(", ")
+
+      msg = "Your Gemfile contains scoped sources #{non_api_source_names} whose servers do not " \
+        "implement a dependency API. Using the `source` is a security risk, and may result in installing " \
+        "unexpected gems. To resolve this warning, use the sources whose servers implement the dependency API."
```
junaruga, line 537: I thought "Your Gemfile contains scoped sources ... whose servers do not implement a dependency API" is more clear for what users did. And it was hard for me to understand the message "any of the ... gem servers don't implement a dependency API" you suggested. So, I applied the message "Your Gemfile contains ...". But feel free to suggest the better message to me. Thanks.

deivid-rodriguez: I like the change from "any of the .. gems servers" to "Your Gemfiles contains...". But I think we can still improve the message. How about

> Your Gemfile contains scoped sources that don't implement a dependency API, namely:
> * rubygems repository http://127.0.0.1:8802/ or installed locally
>
> Using the above gem servers may result in installing unexpected gems. To resolve this warning, make sure you use gem servers that implement dependency APIs, such as gemstash or geminabox gem servers.

Thinking about it, the warning is a bit tricky since the end user might not be responsible for the maintenance of the gem servers themselves. But I don't have better alternatives. Maybe we should implement the size check directly so that at least this doesn't show for people just playing around with small servers with no dependency API.

junaruga: OK, your suggestion is more readable. I rebased to print the exact warning message you suggest.

```
+ ruby /home/jaruga/git/rubygems/bundler/spec/support/bundle.rb install
Your Gemfile contains scoped sources that don't implement a dependency API, namely:

  * rubygems repository http://127.0.0.1:8802/ or installed locally

Using the above gem servers may result in installing unexpected gems. To resolve this warning, make sure you use gem servers that implement dependency APIs, such as gemstash or geminabox gem servers.
Fetching source index from http://127.0.0.1:8802/
Fetching source index from http://127.0.0.1:8801/
Resolving dependencies...
Using bundler 2.3.0.dev
Fetching a 0.0.2
Installing a 0.0.2
Fetching b 1.0.0
Installing b 1.0.0
Bundled gems are installed into `./app`
```

And of course if the `non_api_source_names.size` is more than 2, the message like this.

```
Your Gemfile contains scoped sources that don't implement a dependency API, namely:

  * aaa
  * bbb

Using the above gem servers may result in installing unexpected gems. To resolve this warning, make sure you use gem servers that implement dependency APIs, such as gemstash or geminabox gem servers.
```

voxik: What about being blunt about the issue and refer to the CVE?

junaruga: Currently there is no `CVE-` words in the source code of the RubyGems and Bundler. The `CHANGELOG.md` and `bundler/CHANGELOG.md` and commit messages only include the `CVE-` words. So, I added the CVE info `CVE-2020-36327` to the commit message now.

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,29 @@ def materialize(dependencies)
     end
 
     def precompute_source_requirements_for_indirect_dependencies?
-      @remote && sources.non_global_rubygems_sources.all?(&:dependency_api_available?) && !sources.aggregate_global_source?
+      precompute = false
+      if @remote
+        if sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
+          precompute = true
+        else
+          non_depeneency_api_warning
```
deivid-rodriguez, line 525:
```suggestion
          non_dependency_api_warning
```

bundler/lib/bundler/definition.rb
```diff
@@ -517,7 +517,29 @@ def materialize(dependencies)
...
+        end
+      end
```
deivid-rodriguez, line 527: I think maybe the implementation is simpler like this:

```ruby
def precompute_source_requirements_for_indirect_dependencies?
  return false unless @remote && !sources.aggregate_global_source?

  if sources.non_global_rubygems_sources.all?(&:dependency_api_available?)
    true
  else
    non_dependency_api_warning
    false
  end
end
````

**Review by deivid-rodriguez — COMMENTED** [review]
I think you can proceed with the tests! Thanks!

File: bundler/lib/bundler/definition.rb (hunk 5 above)
Comment (junaruga) [reply bundler/lib/bundler/definition.rb:525]: Oh good catch!

File: bundler/lib/bundler/definition.rb (hunk 6 above)
Comment (junaruga) [reply bundler/lib/bundler/definition.rb:527]: OK. It's simpler. I will update the logic with your suggestion.

Note previously the `sources.non_global_rubygems_sources.all?(&:dependency_api_available?)` is evaluated earlier than `!sources.aggregate_global_source?`. Now in the new logic, the `!sources.aggregate_global_source?` is evaluated earlier than `sources.non_global_rubygems_sources.all?(&:dependency_api_available?)`.

Comment (deivid-rodriguez) [reply bundler/lib/bundler/definition.rb:527]: Indeed. I don't think these methods have side effects so in principle it shouldn't make a difference.

File: bundler/spec/bundler/definition_spec.rb [hunk 7]
```diff
@@ -277,6 +277,59 @@
...
+          [
+            double("non-global-source-0", :dependency_api_available? => true, :to_s => "a"),
+            double("non-global-source-1", :dependency_api_available? => false, :to_s => "b"),
+            double("non-global-source-2", :dependency_api_available? => false, :to_s => "c"),
+          ]
+        end
+
+        it "will raise a warning" do
+          expect(subject.send(:precompute_source_requirements_for_indirect_dependencies?)).to be_falsy
+          expect(subject).to receive(:non_dependency_api_warning)
```
junaruga, line 319: @deivid-rodriguez I added the tests. But the test fail in this line. RSpec 3's `receive` method doesn't count as expected. Do you know what's wrong in my code?

```
$ bin/rspec spec/bundler/definition_spec.rb:322

  1) Bundler::Definition#precompute_source_requirements_for_indirect_dependencies? when remote and does not have multiple global_source when scoped sources do not contain a dependency API will raise a warning
     Failure/Error: expect(subject).to receive(:non_dependency_api_warning)
...
           expected: 1 time with any arguments
           received: 0 times with any arguments
```

deivid-rodriguez: Not sure, I think you need to setup the expectation before actually calling the method, or use `have_received` if you do it after?

junaruga: @deivid-rodriguez I found the reason. As you said, the expectation logic should be before the actual calling. and not only that, for example the following code don't work at the line `expect(Bundler.ui).to receive(:warn)`. Because the first line `expect(subject).to receive(:non_dependency_api_warning)` hides the received count of nested method `Bundler.ui.warn` actually called in the `non_dependency_api_warning`. So, we cannot use the `expect`, `receive` for the nested 2 methods in the one `it` block.

```
expect(subject).to receive(:non_dependency_api_warning)
expect(Bundler.ui).to receive(:warn) # <= can't count the received method in the `non_dependency_api_warning`.
expect(subject.send(:precompute_source_requirements_for_indirect_dependencies?)).to be_falsy
```

junaruga:
I am ready for the review including the tests now.

junaruga:
Why don't you merge this PR? Do you have anything to review?

deivid-rodriguez:
I will review it when I get time. Mainly I want to evaluate whether we can/should add a size check before merging this, and get a safe warning-less behavior by default for small size servers.

hsbt:
👋  @junaruga, I'd like to merge this PR with a small simplification.

I removed the `@remote` and `aggregate_global_source?` guards to preserve the existing behavior of the method and only add the warning. Are you okay with merging this version?

junaruga:
@hsbt 👋🏼 Thank you for your work on this PR. Yes, it's okay with merging with your changes. It's also okay to squash my commit with your other commits or open another PR closing this PR if you would prefer.

Issue #1629, opened 2016-05-29
Different behaviour of fetch and install with platform specific gems


toy:
When checking how rubygems behave with platform specific gems, I've noticed different behaviour of fetch and install commands. Given my platform is `x86_64-darwin-15`, using geminabox and given there are originally following gems uploaded:
- abc-0.1.0
- abc-0.1.0-darwin
- abc-0.1.0-x86_64-darwin
- abc-0.1.0-x86_64-darwin-14
- abc-0.1.0-x86_64-darwin-15

I fetched/installed a gem, then removed the one that was fetched/installed from server and repeated.
For install the order was:
- abc-0.1.0-darwin
- abc-0.1.0-x86_64-darwin
- abc-0.1.0-x86_64-darwin-15
- abc-0.1.0
- ERROR:  Could not find a valid gem 'abc' (>= 0) …

For fetch the order was:
- abc-0.1.0-x86_64-darwin-15
- abc-0.1.0
- abc-0.1.0-x86_64-darwin
- abc-0.1.0-darwin
- ERROR:  Could not find a valid gem 'abc' (>= 0) …

This issue is related to:
- [ ] Network problems
- [ ] Installing a library
- [ ] Publishing a library
- [x] The command line `gem`
- [ ] Other

Here are my current environment details:

```
$ gem env version
2.6.4
```

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
@toy is this still an issue for you, i apologize for not looking at this earlier :sweat_smile:

MSP-Greg:
@bronzdoc

This issue may be something similar to PR #2312, in that RubyGems is doing odd things when a gem has several platform releases.  I didn't know that some Darwin gems were pre-compiled...

toy:
@bronzdoc Rechecked using ruby 2.4.4, rubygems 2.7.7 on darwin 16, besides need to change 14/15 to 15/16 the behaviour is still identical

MSP-Greg:
@toy

I'm a Windows type, and know almost nothing about MacOS.  Using your first 'group',  which gems can be properly installed on your system?

On Windows, the first gem `abc-0.1.0` would be `platform ruby`, and only one of the additional gems would match.  Would more than one match on MacOS?

Thanks, Greg

toy:
[...]

PR #5037, opened 2021-10-29
Fix `gem install` vs `gem fetch` inconsistency

deivid-rodriguez:

Running `gem fetch` and `gem install` against a `geminabox` server results in different fetched/installed gems when there are platform specific gems involved.

Use the same sort criteria in `gem fetch` used elsewhere.

Fixes https://github.com/rubygems/rubygems/issues/1629.

### DIFF
--- a/lib/rubygems/commands/fetch_command.rb
+++ b/lib/rubygems/commands/fetch_command.rb
@@ -60,7 +60,7 @@ then repackaging it.
         specs_and_sources = filtered unless filtered.empty?
       end
 
-      spec, source = specs_and_sources.max_by {|s,| s.version }
+      spec, source = specs_and_sources.max_by {|s,| s }
 
       if spec.nil?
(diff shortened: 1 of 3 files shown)


lib/rubygems/commands/fetch_command.rb
```diff
@@ -60,7 +60,7 @@ def execute
         specs_and_sources = filtered unless filtered.empty?
       end
 
-      spec, source = specs_and_sources.max_by {|s,| s.version }
+      spec, source = specs_and_sources.sort_by {|s,| s }.last
```
toy, line 63: I think `max_by` should still work:
```suggestion
      spec, source = specs_and_sources.max_by {|s,| s }
```

deivid-rodriguez: Totally, not sure what I was thinking!

PR #5039, opened 2021-10-30
Fix system man pages no longer working after bundler overrides `MANPATH`

### LINKED ISSUE #5038: Bundler incorrectly sets MANPATH when a gem contains a man/ directory
I noticed that when I had the [kramdown](https://github.com/gettalong/kramdown) gem in my `Gemfile`, `man` commands would mysteriously not be able to find system man pages (ex: `system("man ls")` or `IO.popen("man ls")`). 

## Steps To Reproduce

Control to make sure there's a man page for the `ls` command:

```shell
$ man ls
```

Without Bundler or kramdown (note: `command` bypasses any shell aliases):

```shell
$ command ruby -e 'system("man ls")'
```

With Bundler and kramdown loaded:

```shell
$ bundle exec ruby -e 'system("man ls")'
```

### Gemfile:

```ruby
source 'https://rubygems.org/'

gem 'kramdown', '2.3.1'
```

## Actual Output

Without Bundler or kramdown:

```
*opens the man page for ls*
```

With Bundler and kramdown loaded:

```
No manual entry for ls
```

## Expected Output

The commands should display the man page for `ls`.

## Versions

* ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
* rubygems 3.2.22
* Bundler version 2.2.24

Reproducible on ruby 2.x as well.

## Additional Information

On a hunch, I decided to test whether `$MANPATH` was being overrode:

```shell
$ command ruby -e 'system("echo $MANPATH")'

$ bundle exec ruby -e 'system("echo $MANPATH")'
/home/postmodern/test/ruby/bundler/vendor/bundle/ruby/3.0.0/gems/kramdown-2.3.1/man
```

It appears that because the kramdown gem contains a top-level `man/` directory, bundler is adding the directory to `MANPATH`. Since `MANPATH` is not set, this actually overrides the `MANPATH` causing the `man` command to not search `/usr/share/man` or `/usr/local/share/man`, thus preventing any Ruby code from opening or reading man pages installed on the system.

deivid-rodriguez:

When a gem is shipping man pages, bundler overrides MANPATH in a way that system man pages no longer work.

Leave ":" in front of the MANPATH. I found it on https://askubuntu.com/a/693612/100732.

EDIT: In the end, we're appending ":" to `MANPATH` to still give priority to man pages from gems, while defaulting to system man pages otherwise.

Fixes #5038.

### DIFF
--- a/bundler/lib/bundler/runtime.rb
+++ b/bundler/lib/bundler/runtime.rb
@@ -265,7 +265,7 @@ module Bundler
 
       return if manuals.empty?
       Bundler::SharedHelpers.set_env "MANPATH", manuals.concat(
-        ENV["MANPATH"].to_s.split(File::PATH_SEPARATOR)
+        ENV["MANPATH"] ? ENV["MANPATH"].to_s.split(File::PATH_SEPARATOR) : [""]
       ).uniq.join(File::PATH_SEPARATOR)
     end
 
--- a/bundler/spec/runtime/setup_spec.rb
+++ b/bundler/spec/runtime/setup_spec.rb
@@ -745,41 +745,68 @@ end
     expect(err).to be_empty
   end
 
-  describe "$MANPATH" do
-    before do
+  context "when the user has `MANPATH` set", :man do
+    before { ENV["MANPATH"] = "/foo#{File::PATH_SEPARATOR}" }
+
+    it "adds the gem's man dir to the MANPATH" do
       build_repo4 do
         build_gem "with_man" do |s|
           s.write("man/man1/page.1", "MANPAGE")
         end
       end
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo4)}"
+        gem "with_man"
+      G
+
+      run "puts ENV['MANPATH']"
+      expect(out).to eq("#{default_bundle_path("gems/with_man-1.0/man")}#{File::PATH_SEPARATOR}/foo")
     end
+  end
 
-    context "when the user has one set" do
-      before { ENV["MANPATH"] = "/foo#{File::PATH_SEPARATOR}" }
+  context "when the user does not have `MANPATH` set", :man do
+    before { ENV.delete("MANPATH") }
 
-      it "adds the gem's man dir to the MANPATH" do
-        install_gemfile <<-G
-          source "#{file_uri_for(gem_repo4)}"
-          gem "with_man"
-        G
+    it "adds the gem's man dir to the MANPATH, leaving : in the end so that system man pages still work" do
+      build_repo4 do
+        build_gem "with_man" do |s|
+          s.write("man/man1/page.1", "MANPAGE")
+        end
 
-        run "puts ENV['MANPATH']"
-        expect(out).to eq("#{default_bundle_path("gems/with_man-1.0/man")}#{File::PATH_SEPARATOR}/foo")
+        build_gem "with_man_overriding_system_man" do |s|
+          s.write("man/man1/ls.1", "LS MANPAGE")
+        end
       end
-    end
 
-    context "when the user does not have one set" do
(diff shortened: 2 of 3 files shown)


bundler/spec/runtime/setup_spec.rb
```diff
@@ -757,28 +757,28 @@ def clean_load_path(lp)
...
-      it "adds the gem's man dir to the MANPATH" do
+      it "appends the gem's man dir to the MANPATH" do
         install_gemfile <<-G
           source "#{file_uri_for(gem_repo4)}"
           gem "with_man"
         G
 
         run "puts ENV['MANPATH']"
-        expect(out).to eq("#{default_bundle_path("gems/with_man-1.0/man")}#{File::PATH_SEPARATOR}/foo")
+        expect(out).to eq("/foo#{File::PATH_SEPARATOR}#{default_bundle_path("gems/with_man-1.0/man")}")
```
simi, line 767: If I understand it well, this now appends (instead of original prepending) bundle man path. Is that really intended?

deivid-rodriguez: You have a point. The trick as documented on the stackoverflow page is to prepend `:` to the `MANPATH`, but maybe it also works by appending it? I will try tomorrow. If appending `:` works, it will be the best behavior and it would make things fully backwards compatible.

If appending `:` doesn't work, we can still do better and keep the current behavior when the `MANPATH` is set.

I will improve this 👍.

postmodern: Confirmed prepending `:` works.

```
MANPATH=/home/postmodern/test/ruby/bundler/vendor/bundle/ruby/3.0.0/gems/kramdown-2.3.1/man man ls
No manual entry for ls
MANPATH=:/home/postmodern/test/ruby/bundler/vendor/bundle/ruby/3.0.0/gems/kramdown-2.3.1/man man ls
```

deivid-rodriguez: Appending `:` seems to also work, so that should be the way to go 👍.

deivid-rodriguez: I pushed a new version, it should be better now. But maybe I should add a `man ls` test to double check this works.

bundler/lib/bundler/runtime.rb
```diff
@@ -265,7 +265,7 @@ def setup_manpath
 
       return if manuals.empty?
       Bundler::SharedHelpers.set_env "MANPATH", manuals.concat(
-        ENV["MANPATH"].to_s.split(File::PATH_SEPARATOR)
+        ENV["MANPATH"] ? ENV["MANPATH"].to_s.split(File::PATH_SEPARATOR) : [""]
```
simi, line 268: My idea was to enhance `MANPATH`:

1. if `MANPATH` is empty, assign to `manuals` with `:` prepended
2. if `MANPATH` is not empty, just append `manuals`

deivid-rodriguez: If understand this trick correctly, prepending `:` makes system man pages have precedence over gem man pages while appending it make gem man pages have precedence. So by appending, a gem overriding a potentially globally installed executable would get its man pages properly displayed.

But this is just a guess on how I would intiutively expect this to work, I need to try it and verify it.

deivid-rodriguez: Ok, I added some more specs to verify that this works exactly as we want, and they confirm appending `:` to the man path when not already set does the trick.

I skipped `MANPATH` specs on Windows because I don't think Windows has `man`?

simi: There is no `man` on Windows, at least no by default.

simi:
:muscle:

PR #5040, opened 2021-11-01
Use OpenSSL constants for error codes.

voxik:
This fixes the following test error testing against OpenSSL 3.x:

~~~
  2) Failure:
TestGemRequest#test_verify_certificate_extra_message [/builddir/build/BUILD/ruby-3.0.2/test/rubygems/test_gem_request.rb:358]:
<"ERROR:  SSL verification error at depth 0: invalid CA certificate (24)\n" +
"ERROR:  Certificate  is an invalid CA certificate\n"> expected but was
<"ERROR:  SSL verification error at depth 0: invalid CA certificate (79)\n" +
"ERROR:  Certificate  is an invalid CA certificate\n">.
~~~

Where the root cause is this OpenSSL commit:

https://github.com/openssl/openssl/commit/1e41dadfa7b9f792ed0f4714a3d3d36f070cf30e

It seems that OpenSSL upstream considers the constant value just an implementation detail and therefore this changes the test case to follow the suite.

### DIFF
--- a/test/rubygems/test_gem_request.rb
+++ b/test/rubygems/test_gem_request.rb
@@ -354,30 +354,36 @@ class TestGemRequest < Gem::TestCase
 
   def test_verify_certificate
     pend if Gem.java_platform?
+
+    error_number = OpenSSL::X509::V_ERR_OUT_OF_MEM
+
     store = OpenSSL::X509::Store.new
     context = OpenSSL::X509::StoreContext.new store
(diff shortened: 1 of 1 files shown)


voxik:
I should have noted, that while there is just one test failure, I have fixed both test cases which used similar pattern and could fail any time in the future.

doodzik (approved):
LGTM

deivid-rodriguez (approved):
👍

PR #5049, opened 2021-11-04
Skip to set DESTDIR option when it's not provided for mswin platform

hsbt:

`make DESTDIR= install` is different with `make install`. Ex. It broke the installation of the bundled gems with Ruby 3.1 in mswin environment.

Skip to set `DESTDIR=` options when it's not provided.


deivid-rodriguez (approved):
It'd be nice to be able to reproduce and test this in CI, but if it's too hard I guess it's fine as is. The change seems harmless.

Issue #2370, opened 2018-07-28
--regenerate-binstubs argument to setup.rb does not respect --destdir


Roguelazer:
I'm packing a new versions of rubygems for a distribution and it appears that some time between the last time I did this (2.6.14) and now (2.7.7), the `--regenerate-binstubs` argument became the default; this argument does not respect the `--destdir` argument and always attempts to write into the actual prefix. This makes it difficult to package rubygems into an RPM. Packaging works correctly if `--no-regenerate-binstubs` is passed, but I feel like I ought to be regenerating binstubs.

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

segiddins:
I believe the `--destdir` argument is only meant to control where _rubygems itself_ gets installed, and not where the underlying gems get put?

Roguelazer:
Yes, this is an issue occuring while installing rubygems itself with `/path/to/ruby setup.rb --prefix=/path/to/prefix --destdir=$HOME/rpmbuild/BUILDROOT/blah/`

tjouan:
> I'm packing a new versions of rubygems for a distribution and it appears that some time between the last time I did this (2.6.14) and now (2.7.7), the --regenerate-binstubs argument became the default; this argument does not respect the --destdir argument and always attempts to write into the actual prefix.

  It was added in v2.7.0 (909b5fb81), but at the time the option itself was
dead code: using `--no-regenerate-binstubs` had no effect, and the related code
was always executed. Also if I recall correctly, there was another issue
breaking `--destdir` option when `Gem::USE_BUNDLER_FOR_GEMDEPS` was `true` and
it was fixed in v2.7.1 (not 100 % sure, but 5d5f0c39 and 76945185 might be
related).

  In the original patch I used, I had changed the default value to `false`
because I think this should be the default behavior. But since this feature was
already released, in the end I only submitted a patch to make
`--no-regenerate-binstubs` work: #2099. So starting with v2.7.3, we *must*
specify `--no-regenerate-binstubs` when using `--destdir`. Alternatively,
patching `Gem::USE_BUNDLER_FOR_GEMDEPS` to be `false` allows using `--destdir`
alone if I'm correct (which would be done when building a package).

  Then in v2.7.5 you can just set `DONT_USE_BUNDLER_FOR_GEMDEPS=yes`
environment variable instead of patching, and again `--destdir` might work
without `--no-regenerate-binstubs`.

[...]

PR #5051, opened 2021-11-04
Fix `ruby setup.rb` when `--prefix` is passed

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

The different parts of the default rubygems & bundler installer would be installed at inconstent locations when both `--destdir` and `--prefix` were passed, because `--prefix` was not being considered everywhere.

## What is your fix for the problem, implemented in this PR?

Fix is to make sure all target folders consider `--prefix`.

Closes  https://github.com/rubygems/rubygems/issues/3604.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #5052, opened 2021-11-05
Fix binstub and plugin regeneration when `--destdir` is used

deivid-rodriguez:

When `--destdir` is passed to `setup.rb`, rubygem still rewrites binstubs and plugins outside of destdir

Pass the proper bin dir to the pristine commands, so that the proper stubs are regenerated.

Fixes https://github.com/rubygems/rubygems/issues/2370.

NOTE: This depends on #5051 and it's built on top of it.


deivid-rodriguez:
I'm merging this PR with #5053 because I realized they are fixing the same thing (never try to write outside of `--destdir`).

PR #5062, opened 2021-11-09
Use `shellsplit` instead of array of strings for git push

Maimer:

When using the command:
```ruby
Rake.application.invoke_task('release:source_control_push[--no-verify origin]')
```
It now results in this error starting with bundler version 2.2.30+:
```
Tagged v1.0.0.
Untagging v1.0.0 due to error.
rake aborted!
Running `git push --no-verify\ origin refs/heads/master` failed with the following output:

error: unknown option `no-verify origin'
```

At my organization every developer has pre-push git hooks that need to be bypassed when releasing a gem on the `master` branch.

The fix is to use `shellsplit` instead of an array of strings for the parameters used for the shell command to git push. In this [commit](https://github.com/rubygems/rubygems/commit/97241e0ea42c5f055cd7858a2360b0ec1e3e786c#diff-6e5b8451af5a3bc880efc58d5629a1e70a13d604257aebc314af6c0236b6418c), the change was made and released in version 2.2.29.

In the previous method, the rake command above would produce the following command:
```ruby
sh(['git', 'push', '--no-verify', 'origin', 'refs/heads/master'])
```
In the new method, due to the string interpolation in the array of strings, it returns the following incorrect command:
```ruby
sh(['git', 'push', '--no-verify origin', 'refs/heads/master'])
```

### DIFF
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -129,8 +129,8 @@ module Bundler
 
     def git_push(remote = nil)
       remote ||= default_remote
-      sh(%W[git push #{remote} refs/heads/#{current_branch}])
-      sh(%W[git push #{remote} refs/tags/#{version_tag}])
+      sh("git push #{remote} refs/heads/#{current_branch}".shellsplit)
+      sh("git push #{remote} refs/tags/#{version_tag}".shellsplit)
       Bundler.ui.confirm "Pushed git commits and release tag."
     end
 


deivid-rodriguez:
Thanks @Maimer. Any chance you can write a regression spec for this?

I don't want to break this again for you in the future, and it's very unintuitive that a parameter named `remote` can be used to inject `git` CLI flags. specially since apparently `git` does not accept remote names with spaces. Just saying this to justify that someone may break this again in the future unless we add a spec.

PR #5068, opened 2021-11-11
Unlock dependencies that no longer match lockfile

deivid-rodriguez:

When you change a dependency `foo` in your `Gemfile` with a new requirement that no longer matches the version of the dependency in the lockfile, running `bundle install` will result in an error that tells you to run `bundle update foo` instead.
 
The current outcome is not terrible but it's not great either. Ideally `bundler` should automatically unlock the dependency when it detects this situation instead of printing an error. This is what this PR implements.

NOTE: In addition to fixing the issue, I also added several simplifications in the same area of the code that I was working on.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -656,15 +656,8 @@ module Bundler
     end
 
     def converge_dependencies
-      frozen = Bundler.frozen_bundle?
       (@dependencies + locked_dependencies).each do |dep|
-        locked_source = @locked_deps[dep.name]
-        # This is to make sure that if bundler is installing in deployment mode and
-        # after locked_source and sources don't match, we still use locked_source.
-        if frozen && !locked_source.nil? &&
-            locked_source.respond_to?(:source) && locked_source.source.instance_of?(Source::Path) && locked_source.source.path.exist?
-          dep.source = locked_source.source
-        elsif dep.source
+        if dep.source
           dep.source = sources.get(dep.source)
         end
       end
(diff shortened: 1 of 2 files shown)


indirect (approved):
This looks like a nice change! Convenient that we can now generally link dependencies to specific sources. :) I have one potential question, and it's about whether this might cause a performance regression in the frozen case—it looks like the codepath that got deleted might have been a special case to speed up frozen mode.

In retrospect, we probably need a benchmarking suite for install/lock/update/add/remove with and without freeze so we can graph how fast they are over time. Maybe that's a good GSoC project for next summer. 😅

deivid-rodriguez:
Thanks @indirect!

Is it [4ca4407](https://github.com/rubygems/rubygems/pull/5068/commits/4ca440730413e2d73435d150fdc36fdb8c1b6260) you're concerned about? My understanding is that block of code changes the source of some path dependencies by overwriting the source specified in the Gemfile with the one in the lockfile. I'm not sure why we did that, but I don't think it could cause a performance regression.

It was first added by https://github.com/rubygems/bundler/pull/4617, and the spec added with that PR has been untouched since and it's still passing. My understanding is that this code was unnecessary at best, if not incorrect, because in frozen mode the sources in the lockfile should match strictly the sources in the Gemfile and if they differ, we should error out, not fake that they are the same.

indirect:
Yeah, your analysis seems right to me. Let's do this. 👍🏻

deivid-rodriguez:
Great, hopefully sneaking in these small refactorings doesn't cause any regressions!

PR #5070, opened 2021-11-11
Don't corrupt lockfile when user moves a gem that's already in the lockfile to an incorrect source by mistake

### LINKED ISSUE #5067: Incorrectly specifying a gem under a source where it doesn't exist generates an incorrect lockfile if the gem is cached

Issue was first reported by @sambostock at https://github.com/rubygems/rubygems/issues/4875#issuecomment-910414293.

As stated in the title, incorrectly specifying a gem under a source where it doesn't exist generates an incorrect lockfile if the gem had been previously cached from the correct source.

Furthermore, this lockfile cannot be properly installed on a fresh machine because it's completely incorrect.

Yes, it happens on the latest bundler.

A sample repository to reproduce this was provided at: https://github.com/sambostock/rubygems-rubygems-4875-demo.

If you have a Gemfile and `Gemfile.lock` like the following, where `some_private_gem` lives at `http://localhost:9292` and depends on `example`, which lives at `https://rubygems.org`:

```ruby
source 'https://rubygems.org'

source 'http://localhost:9292' do
  gem 'some_private_gem'
end
```

```
GEM
  remote: http://localhost:9292/
  specs:
    example (1.0.2)
    some_private_gem (0.1.0)
      example (~> 1.0)

GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  arm64-darwin-20

DEPENDENCIES
  example!
  some_private_gem!

BUNDLED WITH
   2.3.0.dev
```

If you mistakenly edit the lockfile like this

```diff
diff --git a/app/Gemfile b/app/Gemfile
index aa3930e..ae9231a 100644
--- a/app/Gemfile
+++ b/app/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 source 'http://localhost:9292' do
   gem 'some_private_gem'
+  gem 'example'
 end
```

Then `bundle install` will generate an incorrect lockfile like the following that will always error out on machines where `example` is not already installed or where `bundler` needs to re-resolve for some reason:

```
GEM
  remote: http://localhost:9292/
  specs:
    example (1.0.2)
    some_private_gem (0.1.0)
      example (~> 1.0)

GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  arm64-darwin-20
  x86_64-darwin-20

DEPENDENCIES
  example!
  some_private_gem!

BUNDLED WITH
   2.3.0.dev
```

This incorrect lockfile is generated _only_ when the `example` gem from rubygems.org was previously present in cache.

Initial investigation shows that the problem is that the current gem cache does not discriminate sources.

`bundle install` on a `Gemfile` where the source for a gem is incorrect.

An error.

### What actually happened?

"Success" with an incorrect lockfile getting generated.

deivid-rodriguez:

A mistake by a user editing a Gemfile could make bundler generate a bad lockfile, and a situation quite tricky to workaround.

At the beginning of resolution, where we make sure that all top level dependencies are available, make sure we ignore locked specs.

Fixes #5067.

### DIFF
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -19,7 +19,7 @@ module Bundler
     end
 
     def full_name
-      if platform == Gem::Platform::RUBY || platform.nil?
+      if platform == Gem::Platform::RUBY
         "#{@name}-#{@version}"
       else
         "#{@name}-#{@version}-#{platform}"
@@ -61,7 +61,7 @@ module Bundler
     def to_lock
       out = String.new
 
-      if platform == Gem::Platform::RUBY || platform.nil?
+      if platform == Gem::Platform::RUBY
         out << "    #{name} (#{version})\n"
       else
         out << "    #{name} (#{version}-#{platform})\n"
@@ -105,7 +105,7 @@ module Bundler
     end
 
     def to_s
-      @__to_s ||= if platform == Gem::Platform::RUBY || platform.nil?
+      @__to_s ||= if platform == Gem::Platform::RUBY
         "#{name} (#{version})"
       else
         "#{name} (#{version}-#{platform})"
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -84,7 +84,7 @@ module Bundler
         else
           ruby_platform_materializes_to_ruby_platform? ? self : Dependency.new(name, version)
         end
-        platform_object = ruby_platform_materializes_to_ruby_platform? ? Gem::Platform.new(platform) : Bundler.local_platform
+        platform_object = ruby_platform_materializes_to_ruby_platform? ? platform : Bundler.local_platform
         candidates = source.specs.search(search_object)
         same_platform_candidates = candidates.select do |spec|
           MatchPlatform.platforms_match?(spec.platform, platform_object)
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -75,7 +75,17 @@ module Bundler
       out
     end
 
-    def __materialize__
+    def materialize_for_installation
+      __materialize__(ruby_platform_materializes_to_ruby_platform? ? platform : Bundler.local_platform)
+    end
+
+    def materialize_for_resolution
+      return self unless Gem::Platform.match_spec?(self)
+
+      __materialize__(platform)
+    end
+
+    def __materialize__(platform)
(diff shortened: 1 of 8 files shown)


doodzik (approved):

doodzik:
This is such a good change 🙌

bundler/lib/bundler/definition.rb
```diff
@@ -385,8 +385,8 @@ def ensure_equivalent_gemfile_and_lockfile(explicit_flag = false)
 
       # Check if it is possible that the source is only changed thing
       if (new_deps.empty? && deleted_deps.empty?) && (!new_sources.empty? && !deleted_sources.empty?)
-        new_sources.reject! {|source| (source.path? && source.path.exist?) || equivalent_rubygems_remotes?(source) }
-        deleted_sources.reject! {|source| (source.path? && source.path.exist?) || equivalent_rubygems_remotes?(source) }
+        new_sources.reject! {|source| source.path? && source.path.exist? }
+        deleted_sources.reject! {|source| source.path? && source.path.exist? }
```
doodzik, line 385: \nit Would it make sense to add a dedicated method for the `source.path? && source.path.exist?` check on source? IMO this would make the code a bit more readable.

bundler/lib/bundler/definition.rb
```diff
@@ -716,6 +716,10 @@ def converge_specs(specs)
           if s.source.include?(effective_source)
             deps << dep
           else
+            if effective_source.dup.specs.search_all(s.name).empty?
```
doodzik, line 719: \nit the check can probably be moved into a separate method so it doesn't expose as many internals.

bundler/lib/bundler/definition.rb
deivid-rodriguez: Yes, I actually had some more refactorings on gemfile vs lockfile equivalence, and I was planning on removing this whole block of code, since I don't think it's needed. But I decided to leave them out of this PR for now to keep it smaller.

bundler/lib/bundler/definition.rb
deivid-rodriguez: Yes, I will do that, thanks! :+1:

deivid-rodriguez: I wasn't fully satisfied with checking this at this point, so I'm working on a simpler solution. I marked this a WIP for now.

bundler/lib/bundler/definition.rb
deivid-rodriguez: I created #5120 that completely removes this block of code.

PR #5078, opened 2021-11-17
Fix downgrading dependencies by changing the `Gemfile` and running `bundle update`

### LINKED ISSUE #5066: `bundle lock --update` and  `bundle update` cannot be used to downgrade a dependency since 2.2.22
I want to change the version requirement in my gemfile to downgrade to a previous version. I don't want to change the versions of unrelated gems (so running `bundle update` to update everything is not an option).

In Bundler ≤2.2.21, I can do that by altering the requirement and running `bundle lock --update <gem name>` or `bundle update <gem name>`.

In Bundler ≥2.2.22, I get an error because the gem's dependencies are not downgraded to match.

### Gemfile

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

gem "activerecord", ENV.fetch("ACTIVERECORD_VERSION")
```

### Commands to run

```console
$ ACTIVERECORD_VERSION="~> 6.1.0" bundle install
$ ACTIVERECORD_VERSION="~> 6.0.0" bundle lock --update activerecord
```

(alternatively run `bundle update` rather than `bundle lock --update` in the second command)

### Expected result (Bundler ≤2.2.21)

Gemfile.lock is updated with activerecord downgraded to version 6.0.4.1. For `bundle update` only,  Bundler prints a warning:

```
Note: activerecord version regressed from 6.1.4.1 to 6.0.4.1
```

### Actual result (Bundler ≥2.2.22)

Gemfile.lock is not updated, and Bundler fails with an error:

```
Bundler could not find compatible versions for gem "activesupport":
  In snapshot (Gemfile.lock):
    activesupport (>= 6.1.4.1)

  In Gemfile:
    activerecord (~> 6.0.0) was resolved to 6.0.1.rc1, which depends on
      activesupport (= 6.0.1.rc1)
```

Apart from the failure, it's also weird that it's saying that `~> 6.0.0` was resolved to `6.0.1.rc1` rather than `6.0.4.1` - as well as not being the latest version, I don't think that `6.0.1.rc1` should match the version requirement at all?

### Environment

```
Bundler       2.2.31
Ruby          3.0.2p107 (2021-07-07 revision 0db68f023372b634603c74fca94588b457be084c) [x86_64-darwin20]
  Full Path   /Users/andrew/.rubies/ruby-3.0.2/bin/ruby
  Config Dir  /Users/andrew/.rubies/ruby-3.0.2/etc
RubyGems      3.2.31
  Gem Home    /Users/andrew/.gem/ruby/3.0.2
  Gem Path    /Users/andrew/.gem/ruby/3.0.2:/Users/andrew/.rubies/ruby-3.0.2/lib/ruby/gems/3.0.0
  User Home   /Users/andrew
  User Path   /Users/andrew/.gem/ruby/3.0.0
  Bin Dir     /Users/andrew/.gem/ruby/3.0.2/bin
Tools         
  Git         2.33.1
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

### Bundler Build Metadata

```
Built At          2021-11-08
Git SHA           2505ef8972
Released Version  true
```

deivid-rodriguez:

Due to a recent regression, one cannot downgrade the version constraint of a gem in the Gemfile, and then run `bundle update` to get it resolved and installed.

My fix involves resolving the same set of locked specs that it's used as a base when resolving normally, but change them to be `>= <locked_version>` instead of exact requirements.

By using the same logic used to converged locked specifications, we get this edge case fixed, but maybe more.

Fixes #5066.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -691,21 +691,15 @@ module Bundler
     # generated
     def converge_locked_specs
       deps = []
-
-      # Build a list of dependencies that are the same in the Gemfile
-      # and Gemfile.lock. If the Gemfile modified a dependency, but
-      # the gem in the Gemfile.lock still satisfies it, this is fine
-      # too.
-      @dependencies.each do |dep|
-        if satisfies_locked_spec?(dep)
-          deps << dep
-        end
-      end
-
       converged = []
       @locked_specs.each do |s|
         # Replace the locked dependency's source with the equivalent source from the Gemfile
         dep = @dependencies.find {|d| s.satisfies?(d) }
+
+        if dep && (!dep.source || s.source.include?(dep.source))
+          deps << dep
+        end
+
         s.source = (dep && dep.source) || sources.get(s.source) unless multisource_allowed?
 
         # Don't add a spec to the list if its source is expired. For example,
@@ -761,10 +755,6 @@ module Bundler
       end
 
       resolve
-    end
-
-    def satisfies_locked_spec?(dep)
-      @locked_specs[dep].any? {|s| s.satisfies?(dep) && (!dep.source || s.source.include?(dep.source)) }
     end
 
     def metadata_dependencies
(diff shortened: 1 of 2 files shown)


simi (approved):

deivid-rodriguez:
Thanks @simi!

PR #5089, opened 2021-11-24
newgem templ: Avoid Float 3.0 -> "3" in GH Action

olleolleolle:

This is a very small change, but which avoids a Float-to-String loss of characters.

An example (not from this project) of what it can look like before this change, in this picture:

bild

Read more details, if you like: https://github.com/actions/runner/issues/849

This change avoids a YAML Float-to-String conversion, which turns a 3.0 into a "3". That can make names of builds less clear.

In order to use this new capability, I added a "name" descriptor to the matrix-created Job.

**Maybe**: Perhaps it is also wise to chop off the patch version of the version string, so that only "3.0" is listed, there? **Thought together**: No, let's not.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         ruby:
-          - '<%= RUBY_VERSION %>'
+          - '<%= RUBY_VERSION.split(".").take(2).join(".") %>'
 
     steps:
     - uses: actions/checkout@v2
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Hei!

Regarding quoting the version, definitely 👍. Regarding dropping patch level, I personally prefer exact versions (more determinism), but I understand not everyone prefers that. So not so sure about it. I guess we could make it configurable, not sure.

olleolleolle:
> Hei!
> 
> Regarding quoting the version, definitely 👍. Regarding dropping patch level, I personally prefer exact versions (more determinism), but I understand not everyone prefers that. So not so sure about it. I guess we could make it configurable, not sure.

Cool, let me drop that part of the patch - the ruby/setup-ruby GitHub Action conveniently picks the latest-known in that series.

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -10,11 +10,11 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-
+    name: Ruby ${{ matrix.ruby }}
```
deivid-rodriguez, line 13: CI is complaining about the trailing whitespace.

```suggestion
    name: Ruby ${{ matrix.ruby }}
```

simi (approved):

PR #5090, opened 2021-11-24
Don't unlock dependencies when running `bundle install` after changing global source

deivid-rodriguez:

If you change the global source in your `Gemfile`, and then run `bundle install`, bundler will unlock all dependencies not pinned to non global sources. Instead, `bundle install` should behave conservatively as it always does and try to preserve existing locked versions.

This is admittedly a small bug, since it's not very common to change the global source. But it also allows us to delete code, which makes me happy. Bug was introduced recently, by the way, with 2.2.18 (when dependency confusion issues were fixed).

Previously in this particular case bundler would update the locked source to the new global source, but don't add the spec to the set of converged specs because it would exit the loop early. Instead, do update the source in this case, and remove the early exit of the loop so that the spec gets locked.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -163,10 +163,6 @@ module Bundler
       end
     end
 
-    def multisource_allowed?
-      @multisource_allowed
-    end
-
     def resolve_only_locally!
       @remote = false
       sources.local_only!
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
Since I was at it and it will be useful for future work, I refactored the way sources for locked specs are set, so that they are set once correctly, and don't need to be amended right before resolving. This also allowed me to remove a fair bit of code.

PR #5110, opened 2021-11-29
Let original `Errno::EACCES` error be raised in compact index updater

### LINKED ISSUE #5105: Bundler 2.2.32: misleading error message for Errno::EACCES

When bundler catches an `Errno::EACCES` error, it presents the following error message, which is actually misleading in cases where `Errno::EACCES` is raised by a firewall rule *NOT* a temp directory permissions issue:

```
There was an error while trying to write to `Bundler does not have write access
to create a temp directory within C:/Users/Simon/AppData/Local/Temp. Bundler
must have write access to your systems temp directory to function properly. `.
It is likely that you need to grant write permissions for that path.
```

This may be specific to Windows OS, I'm not sure if other platforms also raise `Errno::EACCES` when blocked by a firewall.

Yes, using Bundler 2.2.32.

1. On Windows, setup a firewall rule to block `ruby.exe` from initiating outgoing connections. On my system, I use TinyWall, but other firewall apps may perform similarly.
2. Run any `bundle` command that requires internet access
3. The error message shown above appears
4. After disabling the firewall block and allowing the application through, the above error message resolves

`bundle install`

To see an error message about internet connectivity.

### What actually happened?

It told me that my temp folder permissions were incorrect.

## Environment

```
Bundler       2.2.32
  Platforms   ruby, x64-mingw32
Ruby          3.0.2p107 (2021-07-07 revision 0db68f023372b634603c74fca94588b457be084c) [x64-mingw32]
  Full Path   d:/Ruby/v30-x64/bin/ruby.exe
  Config Dir  C:/ProgramData
RubyGems      3.2.32
  Gem Home    d:/Ruby/v30-x64/lib/ruby/gems/3.0.0
  Gem Path    C:/Users/-----/.gem/ruby/3.0.0;d:/Ruby/v30-x64/lib/ruby/gems/3.0.0
  User Home   C:/Users/------
  User Path   C:/Users/-------/.gem/ruby/3.0.0
  Bin Dir     d:/Ruby/v30-x64/bin
Tools
  Git         2.30.0.windows.1
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2021-11-23
Git SHA           20d4957649
Released Version  true
```

deivid-rodriguez:

Apparently this error is sometimes a false positive, not due to any problem when creating a temporary directory.

* This block of code already wraps file operations with `SharedHelpers.filesystem_access`, which rescues and re-raises more friendly errors.
* Creating a temporary directory through `tmpdir` does not raise `Errno::EACCES`. Instead it looks for several potential candidates (in ENV or the CWD as the last resort) with proper permissions to be the temporary directory, and if none is found, it raises an `ArgumentError`.
* Finally, this rescue block apparently leads to some false positives when firewall is blocking the ruby executable on Windows, or at least that's what we've got reported.

In any case, I think it's best to let the original error be raised.

Closes #5105.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -76,11 +76,6 @@ module Bundler
 
           update(local_path, remote_path, :retrying)
         end
-      rescue Errno::EACCES
-        raise Bundler::PermissionError,
-          "Bundler does not have write access to create a temp directory " \
-          "within #{Dir.tmpdir}. Bundler must have write access to your " \
-          "systems temp directory to function properly. "
       rescue Zlib::GzipFile::Error
         raise Bundler::HTTPError
       end
--- a/bundler/spec/bundler/compact_index_client/updater_spec.rb
+++ b/bundler/spec/bundler/compact_index_client/updater_spec.rb
@@ -36,16 +36,6 @@ RSpec.describe Bundler::CompactIndexClient::Updater do
     end
   end
 
-  context "when bundler doesn't have permissions on Dir.tmpdir" do
-    it "Errno::EACCES is raised" do
-      allow(Bundler::Dir).to receive(:mktmpdir) { raise Errno::EACCES }
-
-      expect do
-        updater.update(local_path, remote_path)
-      end.to raise_error(Bundler::PermissionError)
-    end
-  end
-
   context "when receiving non UTF-8 data and default internal encoding set to ASCII" do
     let(:response) { double(:response, :body => "\x8B".b) }
 


katrkatcloud:
Let origanal error be added

deivid-rodriguez:
The new error message as reported by @SimonEast in the original issue is already an improvement, and rescuing this error here felt weird anyways. So I'll merge this and if @SimonEast wants to help us digging more into why this `Errno::EACCES` error is being raised on Windows, we can improve this in future PRs.

PR #5118, opened 2021-12-01
Ignore to generate the documentation from vendored libraries

hsbt:

When users generated the documentation of rubygems and bundler by rdoc, They got the duplicated entries with the standard library of Ruby. 

But I'm not sure we enabled the documentation of non-standard libraries like `connection_pool`. Maybe we will ignore only the following librraries.

* optparse
* tsort
* fileutils
* tmpdir
* uri

Ignore to generate them.


deivid-rodriguez (commented):
Looks good, but you need to run `rake manifest` to update the `Manifest.txt` file.

PR #5120, opened 2021-12-01
Improve gemfile-lockfile source equivalence errors

deivid-rodriguez:

When checking whether a `Gemfile` is equivalent to the `Gemfile.lock` in frozen mode, `bundler` prints some messages about sources being added or removed from the Gemfile that I believe could be improved. Specially since we have a single global source, I don't think it makes much sense to talk about "This source was added to the Gemfile" or "This source was removed from the Gemfile". Instead, I think it's simpler to say "This dependency's source has changed from this source to this other source".

My fix implements the above, which allows to remove a fair amount of new unneecessary code.

### DIFF
--- a/bundler/lib/bundler/cli/update.rb
+++ b/bundler/lib/bundler/cli/update.rb
@@ -66,7 +66,7 @@ module Bundler
 
       if locked_gems = Bundler.definition.locked_gems
         previous_locked_info = locked_gems.specs.reduce({}) do |h, s|
-          h[s.name] = { :spec => s, :version => s.version, :source => s.source.to_s }
+          h[s.name] = { :spec => s, :version => s.version, :source => s.source.identifier }
           h
         end
       end
@@ -95,7 +95,7 @@ module Bundler
           end
 
           locked_source = locked_info[:source]
-          new_source = new_spec.source.to_s
+          new_source = new_spec.source.identifier
           next if locked_source != new_source
 
           new_version = new_spec.version
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -634,15 +634,13 @@ module Bundler
     end
 
     def converge_dependencies
+      changes = false
+
       @dependencies.each do |dep|
         if dep.source
           dep.source = sources.get(dep.source)
         end
-      end
 
-      changes = false
-
-      @dependencies.each do |dep|
         unless locked_dep = @locked_deps[dep.name]
           changes = true
           next
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -396,9 +396,7 @@ module Bundler
       end
 
       added.concat new_deps.map {|d| "* #{pretty_dep(d)}" } if new_deps.any?
-      if deleted_deps.any?
-        deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" }
-      end
+      deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" } if deleted_deps.any?
 
       both_sources = Hash.new {|h, k| h[k] = [] }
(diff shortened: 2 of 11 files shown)


sly7-7:
Hello, 

This morning, we encountered an issue after deployment, during app launch:
```
 Message from application: You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the /mnt/www/webapps/heimdall/staging/api/releases/20211208082325/Gemfile freeze
by running `bundle config unset deployment`.

* tzinfo-data from `no specified source` to `rubygems repository https://rubygems.org/`
 (Bundler::ProductionError)
  /mnt/www/webapps/.rvm/rubies/ruby-3.0.0/lib/ruby/site_ruby/3.0.0/bundler/definition.rb:396:in `ensure_equivalent_gemfile_and_lockfile'
```

Turns out by removing the tzinfo-data gem from the gemfile (which seems to be useless) resolved the issue, but the exact same Gemfile/Gemfile.lock couple deployed and launched yesterday with bundler 2.2.32 did throw the error. So I just wanted to ask, if this was kind of "sleeping issue" rising with this update, or if this commit may have introduced an issue.

deivid-rodriguez:
Hello @sly7-7, thanks for letting me know about this, it definitely looks related to this PR! And sorry things broke 😞.

The intention of this PR was only to give better error messages when an error is raised due to `Gemfile` and `Gemfile.lock` files not being equivalent in frozen mode. But bundler somehow considering gemfile-lockfile couples as not equivalent that previous versions considered as equivalent is definitely some unintended side effect that I do not understand.

Any chance you can provide me with a minimal example to reproduce this?

sly7-7:
Yeah, this seems really related to the `tzinfo-data` gem.

Considering this Gemfile

```
# frozen_string_literal: true
  
source 'https://rubygems.org' do
    ruby '~> 3.0.0'
    gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
end
```

then `bundle` (in dev mode)

```
[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".
Fetching gem metadata from https://rubygems.org/............
Resolving dependencies...
Using bundler 2.2.33
Bundle complete! 1 Gemfile dependency, 1 gem now installed.
```

then `bundle config --local deployment true`

then `bundle`

```
[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".

You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

If this is a development machine, remove the /Users/sma01/work/sandbox/bug-bundler/Gemfile freeze
by running `bundle config unset deployment`.

* tzinfo-data from `no specified source` to `rubygems repository https://rubygems.org/`
* 
```

sly7-7:
🤯  this works if I get rid of the deprecation warning, ie
```Gemfile
# frozen_string_literal: true
  
source 'https://rubygems.org'

gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
```

deivid-rodriguez:
Oh, right, I think I missed this deprecated case 😬, let me have a look at it.

deivid-rodriguez:
I created #5152 to fix this issue.

sly7-7:
@deivid-rodriguez Thank you :). It was such an edge case... If I took care of the deprecation warning about global source, I would never encountered it.

deivid-rodriguez:
Yeah, no problem!

PR #5125, opened 2021-12-02
Skip installation if resolved version is already installed as a default gem

### LINKED ISSUE #5098: `gem install bundler` could render the default bundler executables unusable

Bundler is failing to execute commands with Ruby 3.0.3 and Bundler 2.2.32

https://github.com/rails/spring/runs/4328629672?check_suite_focus=true

But it passes with Ruby 3.0.2 and same version of Bundler

https://github.com/rails/spring/runs/4328604987?check_suite_focus=true

Yes, those are the newest versions

`bundle check` is failing

deivid-rodriguez:

Running `gem install <name>:<version>` where `<name>-<version>` is a default gem has the potential to break things because of little divergences on how ruby-core installs default gems vs how rubygems installs default gems.
    
For example, the `bundler` gem installed by ruby-core has its executables installed on a "libexec" folder inside the gem installation folder (it's the only thing included there because the other files belong in the stdlib core folder). However, when installed as a regular gem this folder is called "exe". If we allow overwriting the default gem with a regular installation, the "libexec" folder will be renamed to "exe", but the default gemspec will still point to the original "libexec" folder. This can cause issues when changing `GEM_PATH` and `GEM_HOME` (ignoring the regular gem installed), because the "libexec" folder will no longer be found.

The easiest way to prevent this mess seems to skip installation of gems if the same resolved version is already installed as a default gem.

Closes #5098.

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -197,6 +197,14 @@ You can use `i` command instead of `install`.
 
     request_set = dinst.resolve_dependencies name, req
 
+    resolved_spec = request_set.always_install.first
+    matching_spec = Gem::Dependency.new(resolved_spec.name, resolved_spec.version).matching_specs.first
+
+    if matching_spec && matching_spec.default_gem?
+      say "Skipping installation of #{resolved_spec.full_name} because it's already a default gem"
+      return
+    end
+
     if options[:explain]
       say "Gems to install:"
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This broke some stuff, apparently. I'm glad we have such a comprehensive end to end test suite because I thought everything was going to be green.

PR #5128, opened 2021-12-02
Avoid error reporting relative path when validating frozen

yaauie:
When invoking `Bundler::Runtiume::setup` on a frozen bundle, and there is no
relative path from PWD to our default lockfile, our attempt to pre-generate a
helpful message results in an `ArgumentError`.

This is more likely to occur on Windows, for example if PWD is on the `C:`
volume and our project is on the `D:` volume, our application entry-point is
invoked using an absolute path, and our application invokes `Bundler::setup`.

A customer invoking Logstash 7.15.x using an absolute path `D:\logstash-7.15.0\bin\logstash` from a shell whose PWD was on the `C:\` volume resulted in an `ArgumentError` propagating up from `Bundler::Definition#ensure_equivalent_gemfile_and_lockfile`. Logstash invokes `Bundler` directly in code with `:frozen` attribute set to `true`.

Stdlib's `Pathname#relative_path_from` raises an `ArgumentError` if it cannot give a relative path.

Since Bundler's usage doesn't actually require a relative path but is only meant to provide helpful guidance to the user should our invocation eventually fail, it makes sense to rescue the `ArgumentError` and simply use the absolute path when a relative path can't be built.

I would appreciate guidance on how to write a test for this. `Bundler::Definition#ensure_equivalent_gemfile_and_lockfile` does not appear to be covered at all in unit tests, and in any case there are likely to be OS-specific instructions for what constitutes a different base (on my POSIX machine running jruby, I cannot get `Pathname#relative_path_from` to produce the same `ArgumentError`).

### DIFF
--- a/bundler/lib/bundler/cli/check.rb
+++ b/bundler/lib/bundler/cli/check.rb
@@ -29,7 +29,7 @@ module Bundler
         Bundler.ui.warn "Install missing gems with `bundle install`"
         exit 1
       elsif !Bundler.default_lockfile.file? && Bundler.frozen_bundle?
-        Bundler.ui.error "This bundle has been frozen, but there is no #{Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)} present"
+        Bundler.ui.error "This bundle has been frozen, but there is no #{SharedHelpers.relative_lockfile_path} present"
         exit 1
       else
         Bundler.load.lock(:preserve_unknown_sections => true) unless options[:"dry-run"]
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
Normally bundler uses end to end specs. I would grep specs for this specific error message and use the result as a base to reproduce this particular realworld situation (or a simplified version of it).

The changes look good to me, by the way!

deivid-rodriguez:
Actually one of our Windows specs just reproduced this error. See #6989.

So I went ahead and edited this PR so that this similar error you were running into is also prevented.

deivid-rodriguez:
Thanks!

PR #5133, opened 2021-12-04
Add login & logout alias for the signin & signout commands

colby-swandale:

I can never remember the commands to "sign into RubyGems" & "logout of RubyGems" the first time. I always try using `login` & `logout` first, but then have to lookup the command to remember that it's actually `signin` & `signout`.

I wish to add an alias to allow both options so people don't have to remember which set is specifically needed.

### DIFF
--- a/lib/rubygems/command_manager.rb
+++ b/lib/rubygems/command_manager.rb
@@ -73,7 +73,9 @@ class Gem::CommandManager
   ].freeze
 
   ALIAS_COMMANDS = {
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Honestly I never confused these myself, and I don't recall the last time I run `gem signing` anyways, because `gem push` and friends automatically login for you.

That said, I won't block this if it helps making @colby-swandale happy :)

PR #5136, opened 2021-12-05
Allow using `Gem::Version` without loading the rest of rubygems

deivid-rodriguez:

Some users of `--disable-gems` might want to use `Gem::Version` for feature checking for example without loading the rest of RubyGems.

See https://bugs.ruby-lang.org/issues/18376 or https://github.com/mperham/connection_pool/issues/158.

Add a couple of tiny changes to make this possible.

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -149,6 +149,15 @@
 # For the last example, single-digit versions are automatically extended with
 # a zero to give a sensible result.
 
+# Our code style opens classes directly without opening the intermediate
+# modules. This works because tha main entrypoint `rubygems.rb`, which defines
+# the root `Gem` module, is usually required first. But in this case we want to
+# allow using `Gem::Version` without loading the rest of rubygems, so we
+# explicit define the `Gem` placeholder module first.
+module Gem; end
+
+require_relative "deprecate"
+
 class Gem::Version
   autoload :Requirement, File.expand_path('requirement', __dir__)
 
--- a/test/rubygems/test_project_sanity.rb
+++ b/test/rubygems/test_project_sanity.rb
@@ -13,7 +13,7 @@ class TestProjectSanity < Gem::TestCase
   end
 
   def test_require_rubygems_package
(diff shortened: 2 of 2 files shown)


lib/rubygems/version.rb
```diff
@@ -149,6 +149,10 @@
 # For the last example, single-digit versions are automatically extended with
 # a zero to give a sensible result.
 
+module Gem; end
```
simi, line 152: Any reason to not define Gem module in "cascade"?

```ruby
module Gem
  class Version
  # ...
```

deivid-rodriguez: I wanted to keep the diff small instead of having to change the indentation of the whole file. Also, directly opening the class is how other all other files are defined, so it seemed more consistent. That said, I don't mind changing it at all.

simi: For me, it would be confusing to just find this definition at the top of the file. At least small comment would make it easier to understand.

lib/rubygems/version.rb
```diff
@@ -149,6 +149,15 @@
 # For the last example, single-digit versions are automatically extended with
 # a zero to give a sensible result.
 
+# Our code style opens classes directly without opening the intermediate
+# modules. This works because tha main entrypoint `rubygems.rb`, which defines
+# the root `Gem` module, is usually required first. But in this case we want to
+# allow using `Gem::Version` without loading the rest of rubygems, so we
+# explicit define the `Gem` placeholder module first.
```
eregon, line 156: The `module Gem; end` will load all of RubyGems if there is `autoload :Gem, 'rubygems'` defined.
That can be detected with `if Object.autoload?(:Gem)`, but anyway I think there is nothing sensible we can do except loading the whole RubyGems on `require 'rubygems/version'` for that case.

This means, this approach does not work as uintended when `autoload :Gem, 'rubygems'` is used to load RubyGems lazily.

simi: Do you have any example of autoloading Gem lazily? When is that useful?

eregon: TruffleRuby does it, because it's a huge gain on startup time: https://eregon.me/blog/2019/04/24/how-truffleruby-startup-became-faster-than-mri.html#autoloading-rubygems

PR #5139, opened 2021-12-06
Revert stuff about `Gem::Version` availability

deivid-rodriguez:

No problem, really, but just reverting a couple of things to keep affected parties happy.

* Revert #5112, because the upstream change was reverted. I'll leave a "won't fix" note at the related ticket.
* Partially revert #5136, because a better solution to provide built-in version comparison functionality is being discussed.
 

deivid-rodriguez:
@simi Let's revert this until this discussion is more matured, ok?

simi:
ok

PR #5140, opened 2021-12-06
Fix upgrade crashing when multiple versions of `fileutils` installed

deivid-rodriguez:

If you have multiple versions of `fileutils` installed on a ruby where fileutils is a default gem, upgrading rubygems partially fails with an error like this.

```
$ ruby setup.rb             
  Successfully built RubyGem
  Name: bundler
  Version: 2.3.0.dev
  File: bundler-2.3.0.dev.gem
    undefined method `name' for nil:NilClass
```
 
The issue is tricky.  There's 3 relevant moments:

* When first loading the `update` command,  `tmpdir` is required. This require causes the `Gem::Specification.stubs` array to be loaded, because since there are more than one version of `fileutils`, which is a `tmpdir` dependency, `rubygems` needs to figure out which one to require, and needs to load stubs from disk for that.
* Later on, when installing the upgraded default bundler gem, the upgrader first deletes the existing default bundler gemspec from disk before installing the upgraded version.
* Finally, when plugins are being regenerated specification stubs are materialized into actual `Gem::Specification`'s, but since one of them is no longer on disk, it results on a `nil` specification being loaded, causing a runtime error later on when calling a specification method on it.
    
I can think of several ways of fixing this issue:
* Delete the existing spec after installing the new one, not before.
* Require `tmpdir` lazily.
* Don't materialize specs at all when regenerating plugins, since I don't think it's needed. All we need is their versions and names, which stubs already provide.
 
The PR implements the second and third solution. Since I was at it, I removed the specification mutex that I think it was no longer necessary after the third solution. In any case, it was added to support multiple concurrent `gem install` commands, but that has never really been supported, because the gem installer first completely wipes out any existing installation, and then unpackages files from scratch, and that's obviously not thread safe.

As a consequence of this PR, upgrading rubygems should be faster. On my system, it shows a consistent 3% speed up but that's with no gems installed, I guess it could be bigger with many gems installed.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -42,6 +42,14 @@ jobs:
         run: |
           ruby -Ilib -S rake install 2> errors.txt || (cat errors.txt && exit 1)
           test ! -s errors.txt || (cat errors.txt && exit 1)
+      - name: Check downgrading
+        run: gem update --system 3.2.32
+      - name: Check installing fileutils
+        run: gem install fileutils
+      - name: Check installing with upgraded fileutils
+        run: |
+          ruby -Ilib -S rake install 2> errors.txt || (cat errors.txt && exit 1)
+          test ! -s errors.txt || (cat errors.txt && exit 1)
       - name: Run a local rubygems command
         run: gem list bundler
         env:
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -12,8 +12,6 @@ class Gem::Commands::SetupCommand < Gem::Command
   ENV_PATHS = %w[/usr/bin/env /bin/env].freeze
 
   def initialize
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
@simi We got some flaky failure in CI and this PR is very likely to fix it so I'm going to sneak it into the release if you're ok.

deivid-rodriguez:
Actually, I thought this could save us including some regression in this release, but https://github.com/rubygems/rubygems/pull/5050 which I was thinking could've introduced this issue, has already been released. So it's ok to ship this later.

Also, [the flaky failure I'm talking about](https://github.com/rubygems/rubygems/runs/4445320581?check_suite_focus=true( is against an older rubygems, so fixing it would require further patches to bundler.

So to sum up, this PR would be good to merge, but there's no rush to release it.

deivid-rodriguez:
Thanks!!

PR #5153, opened 2021-12-08
Revision is now considered within the eql? comparator for a Source::Git object

george-ma:
Preface: this PR is a proof of concept and is mainly just to gain insight in the design decision behind why `revision` isn't currently included in the `eql?` comparator for a `Source::Git` object.

## Description of the problem

This PR will add the `revision` attribute to the `eql?` method for a `Source::Git` object. We have a work around for this on our application, but are looking to upstream this fix through this PR. 

For `Source::Git` objects, the [eql?](
https://github.com/rubygems/rubygems/blob/ea8a61cf621fab4df6848f9a80eba0c8a0ce60da/bundler/lib/bundler/source/git.rb#L53-58) method does not consider `revision` when comparing two `Source::Git` objects. Hence, when two gemfile.lock with different revisions are compared the `eql?` method will return true. Is `revision` being ignored intentionally?

## Reproducible case
In one of our Rails applications, we use the following on the `Gemfile`:

```ruby
gem "rails", github: "rails/rails", branch: "main"
```

That translates to a Git source section that looks like this, on the `Gemfile.lock`:

**Gemfile.lock (v1)**

```ruby
GIT
  remote: https://github.com/rails/rails.git
  revision: a7320e6de7dcab0ee33adc652934998ef5b7324a
  branch: main
  specs:
    actioncable (7.0.0.alpha2)
      actionpack (= 7.0.0.alpha2)
...
```
This would under the current implementation, be equal when converted to a `Source::Git` object to another `Gemfile.lock` with a different revision under the same gem.

**Gemfile.lock (v2)**

```ruby
GIT
  remote: https://github.com/rails/rails.git
  revision: dea5603dffc19db6fbf94d506f087549c014efec
  branch: main
  specs:
    actioncable (7.0.0.alpha2)
      actionpack (= 7.0.0.alpha2)
...
```

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -47,14 +47,19 @@ module Bundler
       end
 
       def hash
-        [self.class, uri, ref, branch, name, version, glob, submodules].hash
+        [self.class, uri, ref, branch, name, version, glob, submodules, safe_revision].hash
       end
 
       def eql?(other)
         other.is_a?(Git) && uri == other.uri && ref == other.ref &&
           branch == other.branch && name == other.name &&
           version == other.version && glob == other.glob &&
-          submodules == other.submodules
+          submodules == other.submodules && eql_revision?(other)
+      end
+
+      def eql_revision?(other)
+        return true if safe_revision.nil? || other.safe_revision.nil?
+        safe_revision == other.safe_revision
       end
 
       alias_method :==, :eql?
@@ -230,6 +235,12 @@ module Bundler
 
       def app_cache_dirname
         "#{base_name}-#{shortref_for_path(cached_revision || revision)}"
+      end
+
+      def safe_revision
+        revision
+      rescue StandardError
+        nil
       end
 
       def revision


deivid-rodriguez:
Can you add a spec so that I can understand what this is fixing? Bundler never compares two lockfiles sources, it only compares lockfile sources with Gemfile sources, and Gemfiles don't have the concept of "locked revision", so I don't understand why we would need this.

george-ma:
> Bundler never compares two lockfile sources, it only compares lockfile sources with Gemfile sources, and Gemfiles don't have the concept of "locked revision"
Okay with this in mind I think this answers my question of why `revision` is not being considered within the `eql?` method. I don't think this change should be upstreamed in this case, thanks Deivid!

deivid-rodriguez:
Alright! I'm still happy to know more about why you need to do this, in case we con do something about it upstream. As you can see in the CI, the patch as proposed breaks a lot of things.

george-ma:
Sure! For context we have a gem-bumping bot (similar to [dependabot](https://github.com/dependabot)) for our core monolith that would create PRs to bump `rails`. We have a method that checks to see if any gems were bumped comparing two `gemfile.lock` files and that would check `dependencies`, `platform`, etc. and use the `eql?` method as described above which is where this was encountered. Just mainly wanted to get feedback to whether this could be upstreamed or not as we were unsure within the team.

deivid-rodriguez:
Alright, and where these different `Gemfile.lock` files come from? Is there like a main lockfile that gets updated normally through `bundle update` and then other lock files are checked against the main lockfile to see if they need updates? I guess you don't want the secondary lock files to go past the revision of the main lockfile, and that's why you don't run `bundle update` directly on them? Or maybe it has nothing to do with this and I'm just making this up 🤣.

deivid-rodriguez:
I'll close this, by the way :)

Issue #5114, opened 2021-11-30
`bundle doctor` misreports all shared libraries as missing on new MacOS versions


deivid-rodriguez:
### Describe the problem as clearly as you can

Running `bundle doctor` on new MacOS versions, you'll get something like this:

```
The following gems are missing OS dependencies:
 * eventmachine: /usr/lib/libSystem.B.dylib
 * eventmachine: /usr/lib/libc++.1.dylib
 * ffi: /usr/lib/libSystem.B.dylib
 * ffi: /usr/lib/libffi.dylib
 * http_parser.rb: /usr/lib/libSystem.B.dylib
 * nokogiri: /usr/lib/libSystem.B.dylib
 * nokogiri: /usr/lib/liblzma.5.dylib
 * nokogiri: /usr/lib/libobjc.A.dylib
 * racc: /usr/lib/libSystem.B.dylib
 * sassc: /usr/lib/libSystem.B.dylib
 * sassc: /usr/lib/libc++.1.dylib
```

The problem is that `bundle doctor` on MacOS uses `otool -L` under the hood, but that's no longer reliable to detect missing shared libraries. According to [macOS Big Sur release notes](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_0_1-release-notes):

> New in macOS Big Sur 11.0.1, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

I guess we should either remove this check on newer MacOS versions, or use something different like ffi.

### Did you try upgrading rubygems & bundler?

 Yes.

### Post steps to reproduce the problem

Running `bundle doctor` on an app with gems using c-extensions on recent MacOS versions.

### Which command did you run?

`bundle doctor`.

### What were you expecting to happen?

No false positives.

### What actually happened?

It tells all shared libraries are missing.

deivid-rodriguez:
Closed by #5173.

PR #5173, opened 2021-12-17
Use `Fiddle` in `bundle doctor` to check for dynamic library presence

alexeevit:

#5114

In the `doctor` command we use `File.exists?` to check if the dynamic library presents. But in macOS Big Sur 11.0.1 a built-in dynamic linker cache of all system-provided libraries was introduced. So the dynamic libraries are stored not on the filesystem but in the cache.

I replaced the `File.exists?` with the `Fiddler.dlopen` as it was proposed in the initial issue.

### DIFF
--- a/bundler/spec/commands/doctor_spec.rb
+++ b/bundler/spec/commands/doctor_spec.rb
@@ -49,8 +49,8 @@ RSpec.describe "bundle doctor" do
       doctor = Bundler::CLI::Doctor.new({})
       expect(doctor).to receive(:bundles_for_gem).exactly(2).times.and_return ["/path/to/rack/rack.bundle"]
       expect(doctor).to receive(:dylibs).exactly(2).times.and_return ["/usr/lib/libSystem.dylib"]
-      allow(File).to receive(:exist?).with("/usr/lib/libSystem.dylib").and_return(true)
-      expect { doctor.run }.not_to(raise_error, @stdout.string)
+      allow(Fiddle).to receive(:dlopen).with("/usr/lib/libSystem.dylib").and_return(true)
+      expect { doctor.run }.not_to raise_error
       expect(@stdout.string).to be_empty
     end
 
@@ -58,7 +58,7 @@ RSpec.describe "bundle doctor" do
       doctor = Bundler::CLI::Doctor.new({})
       expect(doctor).to receive(:bundles_for_gem).exactly(2).times.and_return ["/path/to/rack/rack.bundle"]
       expect(doctor).to receive(:dylibs).exactly(2).times.and_return ["/usr/local/opt/icu4c/lib/libicui18n.57.1.dylib"]
-      allow(File).to receive(:exist?).with("/usr/local/opt/icu4c/lib/libicui18n.57.1.dylib").and_return(false)
+      allow(Fiddle).to receive(:dlopen).with("/usr/local/opt/icu4c/lib/libicui18n.57.1.dylib").and_raise(Fiddle::DLError)
       expect { doctor.run }.to raise_error(Bundler::ProductionError, strip_whitespace(<<-E).strip), @stdout.string
         The following gems are missing OS dependencies:
          * bundler: /usr/local/opt/icu4c/lib/libicui18n.57.1.dylib
--- a/bundler/lib/bundler/cli/doctor.rb
+++ b/bundler/lib/bundler/cli/doctor.rb
@@ -2,6 +2,7 @@
 
 require "rbconfig"
 require "shellwords"
+require "fiddle"
 
 module Bundler
   class CLI::Doctor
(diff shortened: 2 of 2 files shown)


deivid-rodriguez (commented):
Thanks @alexeevit. CI broke for non MacOS rubies, I think we should only use this if all the other options are not available, i.e., only on recent MacOS.

alexeevit:
> Thanks @alexeevit. CI broke for non MacOS rubies, I think we should only use this if all the other options are not available, i.e., only on recent MacOS.

It actually should work for all OS. I'll try to figure it out.

deivid-rodriguez:
Oh, I think I run into that a while ago while trying out the `asdf` version manager and trying to run bundler specs under it. I think it's worth reporting a ticket. I will have a look at the failures here as soon as I can.

alexeevit:
> Oh, I think I run into that a while ago while trying out the `asdf` version manager and trying to run bundler specs under it. I think it's worth reporting a ticket. I will have a look at the failures here as soon as I can.

Oh ok, that's cool. I'll create a ticket and try to run the specs with another ruby version manager.

deivid-rodriguez:
I had a look. First I recommend to add this patch, because the current failure shows no information about what when wrong

```diff
diff --git a/bundler/spec/commands/doctor_spec.rb b/bundler/spec/commands/doctor_spec.rb
index 860b638f06..1d746a889e 100644
--- a/bundler/spec/commands/doctor_spec.rb
+++ b/bundler/spec/commands/doctor_spec.rb
@@ -50,7 +50,7 @@
       expect(doctor).to receive(:bundles_for_gem).exactly(2).times.and_return ["/path/to/rack/rack.bundle"]
       expect(doctor).to receive(:dylibs).exactly(2).times.and_return ["/usr/lib/libSystem.dylib"]
       allow(File).to receive(:exist?).with("/usr/lib/libSystem.dylib").and_return(true)
-      expect { doctor.run }.not_to(raise_error, @stdout.string)
+      expect { doctor.run }.not_to raise_error
       expect(@stdout.string).to be_empty
     end
 
```

Regarding the error itself, I think you need to find a new way to test this because the current test was based off the previous implementation. It was based on stubbing the path existence to return `true` (the line previous to the line being modified in the above diff) but the new implementation no longer relies on a file to exist.

simi:
:thinking: `fiddle` is not present in older Ruby distributions (it is present for 2.5+ only). We should make it optional or wait unless support for older Rubies is dropped (https://github.com/rubygems/rubygems/pull/3398).

alexeevit:
> 🤔 `fiddle` is not present in older Ruby distributions (it is present for 2.5+ only). We should make it optional or wait unless support for older Rubies is dropped (#3398).

Yep, but I added it as a dependency. Doesn't this fix the problem?

simi:
@alexeevit sadly not, that Gemfile is just for bundler releasing purpose, it is not installing gems on client side.

deivid-rodriguez:
Yes, I think it's fine to use `fiddle` consistently on newer rubies, but if it's not available, we should fallback to the current solution of checking whether the dynamic library exists as a file on disk.

alexeevit:
> but if it's not available, we should fallback to the current solution 

@deivid-rodriguez and what about the case of modern macOS and old ruby? Just leave a warning message?

deivid-rodriguez:
Yes, that sounds good, tell the user that she needs to install `fiddle` for the check to work.

alexeevit:
Ok, I've just checked, maybe `fiddle` does not support old ruby versions, but ruby 2.0.0 includes Fiddle.

```
$ docker run -ti ruby:2.0.0-slim irb
irb(main):001:0> require 'fiddle'
=> true
irb(main):002:0> Fiddle.dlopen('/lib/x86_64-linux-gnu/libc.so.6')
=> #<Fiddle::Handle:0x0055d20b0613e0>
```

@deivid-rodriguez I think we should not have the fallback then?

bundler/spec/commands/doctor_spec.rb
```diff
@@ -40,29 +40,39 @@
       allow(File).to receive(:readable?).with(unwritable_file) { true }
     end
 
+    let(:existing_dylib_path) do
```
alexeevit, line 43: I need some help here. The idea is to really check the existence of 100% existed system libraries but I'm not sure what libraries are 100% present in the OSs.

deivid-rodriguez:
Oh I see, then this is indeed easier and we shouldn't need the fallback, yeah.

bundler/spec/commands/doctor_spec.rb
deivid-rodriguez: I guess the test could be made more realistic, but I think for now it could do to change the unit test to adapt to the new code, and stub `Fiddle.dlopen` in a similar way we used to stub `File.exist?` before?

simi:
@alexeevit Ahh, you're right. It is just not installed as a gem, but still present for older rubies.

bundler/spec/commands/doctor_spec.rb
alexeevit: Yep, fixed and all the commits are squished

deivid-rodriguez (approved):
Looks good to me! Also tried it manually on MacOS Monterrey and it gives the proper result now.

deivid-rodriguez:
Thank you! 💪

PR #5175, opened 2021-12-18
Add cargo builder for rust extensions

ianks:
After a couple years of stagnation, and thanks to the help from @qubitrenegade and @briankung finally got around to building a Rust gem extension builder for Rubygems.

### How it works

**tl;dr** [here is a fully working example of a Rust gem](https://github.com/rubygems/rubygems/tree/f91ee8a6465e220f61795d539f7e901b224561fb/test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example), and an excellent [blog post](https://briankung.dev/2022/01/31/sneak-preview-writing-ruby-gem-native-extensions-in-rust/) from @briankung

1. If a `Cargo.toml` file is detected, the new `Gem::Ext::CargoBuilder` is invoked.
2. We invoke `cargo rustc` and handle the linking, and all of the gymnastics around that
3. Move the compiled dylibs to their expected locations for Ruby (i.e. `my_rust_extension.{so,bundle,dll}`)

### Why the extconf?

~My reason is this: In order for ruby to successfully load a ruby extension (via `dlopen`), its first check to see that `ruby_xmalloc` matches on the built library. This ensures that the extension is only loaded if it was compiled with the same version of Ruby.~

We no longer need to do this since we are correctly handling the linking process using Cargo.

Hopefully this is on the right track, would love to get some feedback and get this merged at some point!

closes #4903


ianks:
Though: It would be nice if this thing  “just worked” with rake-compiler. That would enable cross-compilation and what not. I think in order to do that, it might be best to do the whole cargo compilation logic extconf.rn. Although I’m not sure if that would work…

lib/rubygems/ext/builder.rb
```diff
@@ -119,10 +119,13 @@ def builder_for(extension) # :nodoc:
     when /configure/ then
       Gem::Ext::ConfigureBuilder
     when /rakefile/i, /mkrf_conf/i then
-      @ran_rake = true
+      @ran_rake = tru
```
briankung, line 122: Is this a typo?

bjorn3:
> First, the crate is built (either staticlib or cdylib crates will work`)

Linking two rust staticlibs together will cause symbol conflicts because each contains a copy of the standard library. If you are linking only a single rust staticlib into a dylib you should probably ensure that all mangled symbols and all symbols used by rust itself are not exported from the dylib (rustc normally handles this for you when compiling cdylibs). Using two rust cdylibs in the same program on the other hand is fine.

ianks:
So do you think it's best to raise an error when staticlib is used @bjorn3?

bjorn3:
I think that would be the best option indeed. I presume it would be easier to forbid staticlibs for now and maybe allow them in the future with appropriate safety guards than to allow them now and possibly have to backtrack if it causes issues.

ianks:
> I think that would be the best option indeed. I presume it would be easier to forbid staticlibs for now and maybe allow them in the future with appropriate safety guards than to allow them now and possibly have to backtrack if it causes issues.

Another option might be to add `-C metadata=GEM_NAME,GEM_VERSION` to prevent symbol conflicts. Thoughts?

bjorn3:
That is not enough. The standard library doesn't get recompiled when specifying `-Cmetadata`. In addition there are a couple of functions in the standard library that aren't mangled at all. (for example `__rust_alloc`, `rust_eh_personality` or `__rust_foreign_exception`)

bjorn3:
> Also, it seems the rust docs recommend staticlib for this type of use case:
>
> > This format is recommended for use in situations such as linking Rust code into an existing non-Rust application because it will not have dynamic dependencies on other Rust code.
> 
> Thoughts?

This recommendation makes sense when only a single rust library is linked. Unfortunately it breaks down as soon as multiple rust libraries are linked.

Another argument against staticlibs is that they don't have an easy way to specify the required dynamic libraries to link against. To get this list you need to pass `--print native-static-libs` to the rustc invocation producing the staticlib. This will result in a note like `note: native-static-libs: -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc` listing all arguments you need to pass to the linker.

ianks:
>

ianks:
> > Also, it seems the rust docs recommend staticlib for this type of use case:
> > > This format is recommended for use in situations such as linking Rust code into an existing non-Rust application because it will not have dynamic dependencies on other Rust code.
> > 
> > 
> > Thoughts?
> 
> This recommendation makes sense when only a single rust library is linked. Unfortunately it breaks down as soon as multiple rust libraries are linked.
> 
> Another argument against staticlibs is that they don't have an easy way to specify the required dynamic libraries to link against. To get this list you need to pass `--print native-static-libs` to the rustc invocation producing the staticlib. This will result in a note like `note: native-static-libs: -lgcc_s -lutil -lrt -lpthread -lm -ldl -lc` listing all arguments you need to pass to the linker.

Thank you so much for this info! Going to raise unless we detect a cdylib.

ianks:
@bjorn3 Added a patch to raise unless a dylib is found

Also added full integration spec, and things seem to be working well. Hopefully we can get some more eyes on this 👀

simi:
This is on my radar for whole time. I'm just waiting for proper time to start reviewing this. Few initial ideas, questions.

Are there any gems we can check using this already?

Would it be possible to integrate this into `bundle gem` template as new argument? Currently it is possible to generate C extension skeleton with `bundle gem example --ext`. Maybe we can extend `--ext` to support argument (C or RUST, and fallback to C when nothing is present) or add another argument like `--rust-ext`. It would be super useful for testing. But we can also add this later.

Can we make Rust extension testing optional (enabled only when Rust tools are around) so our test suite doesn't rely on Rust being set on the dev. machine?

iliabylich:
I'm so glad to see this being (hopefully) a part of rubygems. I spent enormous amount of time building it manually [here](https://github.com/lib-ruby-parser/ruby-bindings). In short: I followed a very similar path. Rust library (actually, its C bindings, but it doesn't matter) is compiled statically and included with ` -Lpath -lname` with a a bunch of glue code that is based on `RbConfig` data.

Compiling Rust lib to dynamic library seems fine, but actually I don't really understand the reason behind it. What is the conflict of standard libraries here? There's a single crate that (once compiled) includes a single version of Rust stdlib, and dylibs are allowed to share code and symbols. IIRC it is even possible to `dlopen` the same library multiple times (I think I've seen toy implementation of a singleton in a global variable that is different for every loaded dylib because each has its own`.bss`). There's nothing bad about it, but this extra `dlopen` feels a bit redundant. Rust stdlib cannot be extracted to a separate shared library anyway (like `libstdc++`), because, well, ABI is unstable.

But actually maybe I'm wrong and maybe it's easier to build it this way: Rust dylib doesn't know about Ruby dependencies and it's 100% valid on its own, it's isolated and self-contained. Ruby dylib that wraps it is also linked with `libruby.[so,bundle]` and other things that Ruby needs on different platforms. It is going to be small and transparent, and so things like Rust <-> C LTO that could be used with static lib don't bring that much value.

Quick question (sorry if it doesn't belong to this PR directly): is there a way to customize generated Makefile? Or will it run `cargo build` every time you run `rake compile`? I think it's ok for now to rely on cargo dependency tracking, just wondering what's the plan.

ianks:
> Are there any gems we can check using this already? (edit: I see there is some list at https://bugs.ruby-lang.org/issues/15759, I'll check on the approach in there.)

Yes I've been referencing (thermite)[https://github.com/malept/thermite] (big thanks to @malept for this!)

> Quick question (sorry if it doesn't belong to this PR directly): is there a way to customize generated Makefile? Or will it run cargo build every time you run rake compile? I think it's ok for now to rely on cargo dependency tracking, just wondering what's the plan.

The generated makefile actually doesn't run `cargo build`, that's done in the builder. The makefile is just there to compile the ultra-thin C wrapper so we can easily re-use the ruby build pipeline if that makes sense

> Also, what about targets? I think it's irrelevant for linux/mac, but on windows x86_64-pc-windows-gnu should be used, I'm not even sure if dlopen from mingw (that is used for Ruby on windows builds) can load .dll compiled with msvc backend. Because of that, I think it's worth passing --target explicitly.

I was thinking about that, but passing the target is a non-trival thing unfortunately. The mapping from 'RbConfig::CONFIG['arch']` to Rust target-triple is edge casey. One unfortunate this is I don't have a lot of windows experience nor environment to test one. Would love some help with this though!

ianks:
I updated the PR to no longer build a dummy C wrapper, linking libruby with `LDFLAGS` seems sufficient 🎊

ianks:
Looks like rutie wants to compile against a static ruby in CI 🤔

ianks:
https://github.com/danielpclark

> Looks like rutie wants to compile against a static ruby in CI 🤔

@danielpclark I’m a bit stumped on this one. Could you take a look to see what may be causing this?

ianks:
@simi or someone, can you allow CI runs for me :)

simi:
@ianks running now

danielpclark:
@ianks do you have a specific build to point at that's using the static Ruby?  And is the ENV info shown or the state of `RbConfig::CONFIG` ?

One issue I found when implementing Rutie is that it had trouble finding the `libruby.so` file in its original location even if correctly providing the right C build args.  To get around this issue [I link the `libruby.so` file if possible](https://github.com/danielpclark/rutie/blob/master/build.rs#L174-L176) in the cargo dependency sub-directory within the current project `target/*/deps`.  I believe this is an issue with the Rust C linker not handling a defined path when libruby is provided... I don't know if they've gotten around to fixing that particular issue over the past few years.

ianks:
> @ianks do you have a specific build to point at that's using the static Ruby? And is the ENV info shown or the state of `RbConfig::CONFIG` ?
> 
> One issue I found when implementing Rutie is that it had trouble finding the `libruby.so` file in its original location even if correctly providing the right C build args. To get around this issue [I link the `libruby.so` file if possible](https://github.com/danielpclark/rutie/blob/master/build.rs#L174-L176) in the cargo dependency sub-directory within the current project `target/*/deps`. I believe this is an issue with the Rust C linker not handling a defined path when libruby is provided... I don't know if they've gotten around to fixing that particular issue over the past few years.

I've run into this too on [rb-sys](https://github.com/ianks/rb-sys). On *nix systems, I'm using `-rpath` to get around this. On windows, it's proved a bit more challenging.

ianks:
@simi can u run again??

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -67,21 +67,34 @@ def validate_cargo_build!(dir)
...
+    flag = arg[0..1]
+    val = arg[2..-1]
+
+    case flag
+    when "-L"
+      ["-L", "native=#{val}"]
+    when "-l"
+      ["-l", "dylib=#{val}"]
+    when "-F"
+      ["-F", "framework=#{val}"]
```
bjorn3, line 85: These are rustc flags, right? The `-F` here should be `-l` I think.

ianks:
Ok, after banging my head against the screen for a couple hours, I feel really good about this PR on both shared and static versions of ruby.

Not so confident on Windows, but ideally we could get this shipped without 100% windows support testing and have someone who actually knows windows give a shot at it… Thoughts?

ianks:
@eregon if i wanted this to work well with truffle / graal. Would it help to [`--emit llvm-bc`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--emit-specifies-the-types-of-output-files-to-generate) when we compile on Truffle?

iliabylich:
Thanks a lot for all this work! May I ask a few questions?

1. What's the reason behind linking with static `libruby.a`? As I know it's not even available on GitHub CI, and most Ruby distributions don't have it, instead dynamic `libruby.{so,bundle}` are used everywhere. Is it to build an "embeddable" Ruby? Should it be supported at all? Do other builders support it? From what I understand it can't be used to compile a dynamically loaded `.so` extension because it literally mirrors functionality of CRuby. Or, to be even more precise, it is a static archive of everything from CRuby. My hunch is that using it together with common Ruby distribution would make `.so` extension to use its own version of Ruby VM (that is not bootstrapped because ... nobody knows about it except for extension). Actually, I'm not 100% sure.
2. I like that there's no explicit dependency on crates with Ruby C API wrappers, but are there plans to promote any? C API changes over time. For example, as I know, "Variable Width Allocation" project that is going to be enabled-by-default in Ruby 3.1 will change the layout of many structs like `RString`, `RArray` and friends, and these structs are exposed in C API. New layout (these structs are exposed in Ruby include headers) means that there must be a a way to switch between old/new APIs. Do crates like this even exist?
3. Is it correct to link against `libruby.so`? Passing `-Wl,-undefined,dynamic_lookup` (`RUSTFLAGS="-Clink-args=-Wl,-undefined,dynamic_lookup` on linux/mac, unsupported on MinGW) is enough to bind them lately at runtime (and `libruby.so` is always loaded in the first place, that's basically the main purpose of the `ruby` executable). What I'm trying to say is that final `.so` library doesn't have to depend on anything from Ruby distribution. It can, but I guess it's easier to not do that.

For my extension I used the following command to build a `output.so` file from `input.o` (on MacOS):
<details>
  <summary>Collapsed</summary>

```sh
$ gcc \
  -fdeclspec \
  -dynamic \
  -bundle \
  -o output.so \
  input.o \
  -L~/.rvm/rubies/ruby-3.0.0/lib \
  -L. \
  -L/usr/local/opt/libyaml/lib \
  -L/usr/local/opt/libksba/lib \
  -L/usr/local/opt/readline/lib \
  -L/usr/local/opt/zlib/lib \
  -L/usr/local/opt/openssl@1.1/lib \
  -fstack-protector-strong \
  -L/usr/local/lib \
  -L/usr/local/opt/libyaml/lib \
  -L/usr/local/opt/libksba/lib \
  -L/usr/local/opt/readline/lib \
  -L/usr/local/opt/zlib/lib \
  -L/usr/local/opt/openssl@1.1/lib \
  -Wl,-undefined,dynamic_lookup \
  -Wl,-multiply_defined,suppress
```
</details>

(^as you can see there are no `-l` flags, but these `-Ldir` flags are mixed in `RbConfig::CONFIG['LDFLAGS']` with linker flags, so .. it's better to include everything)

In my case I used a script based on `rbconfig` (just like this PR does) that prints this:
<details>
  <summary>Collapsed</summary>

```ruby
win = false
OS_SPECIFIC_OPTIONS =
    case RbConfig::CONFIG['host_os']
    when /mingw/
        win = true
        ['-m64', '-luserenv']
    else
        []
    end

[
    ENV['CC'] || RbConfig::CONFIG['CC'],
    "-o output.bundle",
    'input.o',

    ENV['EXTRA_CFLAGS'] || '',

    '-L' + RbConfig::CONFIG['libdir'],
    RbConfig::CONFIG['LDFLAGS'],
    RbConfig::CONFIG['DLDFLAGS'],

    *OS_SPECIFIC_OPTIONS,
    win ? RbConfig::CONFIG['LIBRUBYARG'] : '',
    RbConfig::CONFIG['LIBS'],
].join(' ')
```
</details>

I had a slightly different setup where Rust library is compiled to a static archive and linked together with `main.o` (in C, has `Init_<ext_name>` fn) using a script posted above. I didn't pass any flags to Rust library, everything can be inferred from `rbconfig`.

On Windows: I'm definitely not an expert, but as I know Windows doesn't have a concept of rpath, a single `$PATH` is used instead, and like I mentioned above MinGW (that must be used instead of MSVC) doesn't support lazy symbol binding. Luckily, `libruby.so` (yes, mingw outputs `.so` files instead of `.dll`) is located in the same directory where `ruby.exe` is stored, and so if `ruby.exe` is in your `PATH` you can link with `libx64-msvcrt-ruby300.so` by its name. Of course, it's also a part of `RbConfig::CONFIG['LIBRUBYARG']`. Also, last time I checked it was necessary to pass `-m64 -luserenv` to compiler, I don't why it's not in `rbconfig`.

I have a repo with 100% hand-written scripts [compile.rb](https://github.com/lib-ruby-parser/ruby-bindings/blob/master/scripts/compile.rb) and [link.rb](https://github.com/lib-ruby-parser/ruby-bindings/blob/master/scripts/link.rb). They are ugly but it [kinda works](https://github.com/lib-ruby-parser/ruby-bindings/actions/runs/1530731258) (for example, check output of the "compile and test" step of the `compile-and-test-mac` flow, you can still view logs of calling cc/linker).

Thanks again and sorry for such a long read!

eregon:
> @eregon if i wanted this to work well with truffle / graal. Would it help to [`--emit llvm-bc`](https://doc.rust-lang.org/rustc/command-line-arguments.html#--emit-specifies-the-types-of-output-files-to-generate) when we compile on Truffle?

I don't know Rust well, we should experiment and test what works.
There are docs about Rust on GraalVM at https://www.graalvm.org/reference-manual/llvm/Compiling/#running-rust

I think instead the GraalVM LLVM toolchain should be used.
FWIW, if `mkmf` is required then the GraalVM LLVM toolchain (`clang`/`clang++`/etc) is prepended to PATH: https://github.com/oracle/truffleruby/blob/39d740bd3eb8abafe737656055850fc7b3a4d1dc/lib/mri/mkmf.rb#L52.
But I guess `rustc` doesn't use clang/clang++/etc and there is no way to tell it to use a given LLVM toolchain?

bjorn3:
> But I guess rustc doesn't use clang/clang++/etc and there is no way to tell it to use a given LLVM toolchain?

Rustc is directly linked to LLVM. If you want it to use a different LLVM version you will have to recompile it.

> Is there a way to tell cargo to emit a native library which has both bitcode and native code?

I don't think so. There is `-Clinker-plugin-lto` to make rustc emit llvm bitcode instead of object files and then let the linker perform LTO, but the LLVM version used by rustc is likely newer than the one used by graal. (rustc often updates before a new LLVM version is officially released) LLVM doesn't guarantee that loading bitcode produced by a newer LLVM versions load by older LLVM versions and rustc disables workarounds for bugs in older LLVM versions when linking against a newer LLVM version. Rustc itself does know how to produce libraries containing both bitcode and object files, but only for rlibs which are for use with rustc only. In addition cargo explicitly disables this feature using `-Cembed-bitcode=no` to save disk space and compile times. I don't think there is a way to make cargo enable it.

eregon:
Having only the bitcode should work, having both is preferred as documented in https://www.graalvm.org/reference-manual/llvm/Compiling/#file-format (s/executable/library/ also holds for those docs) because then it's possible to track native library dependencies.
Can we pass `--emit=llvm-bc` to `cargo` for `RUBY_ENGINE == 'truffleruby'`?

bjorn3:
Using `--emit=llvm-bc` will only emit llvm bitcode for the current crate. Not for any of it's dependencies and not for the standard library. It also won't emit the llvm bitcode for the so called "allocator shim" which is used when allocating memory. Using `-Clinker-plugin-lto` and then passing in a custom linker that collects the paths to all bitcode files is probably the safest option, though I think even that won't result in bitcode for the standard library I think unless you recompile the standard library with `-Clinker-plugin-lto` too.

ianks:
> Thanks a lot for all this work! May I ask a few questions?
> 
> 1. What's the reason behind linking with static `libruby.a`? As I know it's not even available on GitHub CI, and most Ruby distributions don't have it, instead dynamic `libruby.{so,bundle}` are used everywhere. Is it to build an "embeddable" Ruby? Should it be supported at all? Do other builders support it? From what I understand it can't be used to compile a dynamically loaded `.so` extension because it literally mirrors functionality of CRuby. Or, to be even more precise, it is a static archive of everything from CRuby. My hunch is that using it together with common Ruby distribution would make `.so` extension to use its own version of Ruby VM (that is not bootstrapped because ... nobody knows about it except for extension). Actually, I'm not 100% sure.

My goal was to make the compilation process fully compatible by what the defaults supported in extconf.rb. From what I gather, there are a segment of Ruby users who cannot used dylibs.

By default, dylibs are used unless ruby is not compiled with `--enable-shared` or the env var `RUBY_STATIC` is set. Github actions version of Ruby, for example, would used the dylib approach by default.

The static library is not actually linked, the functions are defined at runtime so it uses the API of the current ruby.

> 2. I like that there's no explicit dependency on crates with Ruby C API wrappers, but are there plans to promote any? C API changes over time. For example, as I know, "Variable Width Allocation" project that is going to be enabled-by-default in Ruby 3.1 will change the layout of many structs like `RString`, `RArray` and friends, and these structs are exposed in C API. New layout (these structs are exposed in Ruby include headers) means that there must be a a way to switch between old/new APIs. Do crates like this even exist?

This is something that would be supported in a different crate. I'm currently working on [rb-sys](https://github.com/ianks/rb-sys) to auto-generate C bindings, and there is also [rutie](https://github.com/danielpclark/rutie).

> 3. Is it correct to link against `libruby.so`? Passing `-Wl,-undefined,dynamic_lookup` (`RUSTFLAGS="-Clink-args=-Wl,-undefined,dynamic_lookup` on linux/mac, unsupported on MinGW) is enough to bind them lately at runtime (and `libruby.so` is always loaded in the first place, that's basically the main purpose of the `ruby` executable). What I'm trying to say is that final `.so` library doesn't have to depend on anything from Ruby distribution. It can, but I guess it's easier to not do that.

We don't actually `-l'ink libruby. We actually pass the same flags to the linker that `mkmf` generates by default. So you are correct, in the case of native extensions, linking ruby is not necessary.

ianks:
> Using `--emit=llvm-bc` will only emit llvm bitcode for the current crate. Not for any of it's dependencies and not for the standard library. It also won't emit the llvm bitcode for the so called "allocator shim" which is used when allocating memory. Using `-Clinker-plugin-lto` and then passing in a custom linker that collects the paths to all bitcode files is probably the safest option, though I think even that won't result in bitcode for the standard library I think unless you recompile the standard library with `-Clinker-plugin-lto` too.

Ok this looks a bit more entailed than I originally thought. We can hold off on this functionality for this PR then.

ianks:
Regarding explicitly passing the `--target` option, as @iliabylich mentioned, I took a stab at what it might look like. However, I cannot find an exhaustive list to map `RbConfig::CONFIG['target']` to the rust target triples. If we build a confident mapping, I think it would be wise to explicitly specify the `--target`.

<details>
  <summary>POC diff</summary>
  
```diff
diff --git a/lib/rubygems/ext/cargo_builder.rb b/lib/rubygems/ext/cargo_builder.rb
index e74af7f11..b4afe18d2 100644
--- a/lib/rubygems/ext/cargo_builder.rb
+++ b/lib/rubygems/ext/cargo_builder.rb
@@ -12,10 +12,43 @@ def initialize(dir)
     end
   end
 
-  attr_reader :spec
+  class TargetTriple
+    attr_reader :rb_config
+
+    def initialize(rb_config = RbConfig::CONFIG)
+      @rb_config = rb_config
+    end
+
+    def to_s
+      "#{cpu}-#{vendor}-#{os}"
+    end
+
+    def cpu
+      raw = rb_config['host_cpu']
+
+      return 'aarch64' if vendor == 'apple' && raw == 'arm64'
+
+      raw
+    end
+
+    def vendor
+      rb_config['target_vendor']
+    end
+
+    def os
+      raw_os = rb_config['target_os']
+
+      return 'darwin' if raw_os.include?('darwin')
+
+      raw_os
+    end
+  end
+
+  attr_reader :spec, :target_triple
 
   def initialize(spec)
     @spec = spec
+    @target_triple = TargetTriple.new
   end
 
   def build(extension, dest_path, results, args=[], lib_dir=nil, cargo_dir=Dir.pwd)
@@ -53,7 +86,7 @@ def build_crate(extension, dest_path, results, args, lib_dir, cargo_dir)
   end
 
   def validate_cargo_build!(dir)
-    dylibs = Dir.glob(File.join(dir, 'release', "lib#{spec.name}.{so,dylib}"))
+    dylibs = Dir.glob(File.join(dir, target_triple.to_s, 'release', "lib#{spec.name}.{so,dylib}"))
 
     raise DylibNotFoundError.new(dir) if dylibs.empty?
 
@@ -110,9 +143,10 @@ def cargo_rustc_args(dest_dir)
       '--lib',
       '--release',
       '--frozen',
+      '--target', target_triple.to_s,
       '--',
       *dynamic_linker_flags,
-      *libruby_flags_to_link_modifiers(libruby_arg),
+      *libruby_flags_to_link_modifiers(libruby_arg)
     ]
   end
```

</details>

eregon:
> Using `--emit=llvm-bc` will only emit llvm bitcode for the current crate. Not for any of it's dependencies and not for the standard library. It also won't emit the llvm bitcode for the so called "allocator shim" which is used when allocating memory. Using `-Clinker-plugin-lto` and then passing in a custom linker that collects the paths to all bitcode files is probably the safest option, though I think even that won't result in bitcode for the standard library I think unless you recompile the standard library with `-Clinker-plugin-lto` too.

We probably don't need to have the stdlib as bitcode, based on the example from https://www.graalvm.org/reference-manual/llvm/Compiling/#running-rust. Typically for C extensions only that C extension is compiled as bitcode, and everything else can be only available as native libraries.
However Rust has a concept of packaging, and it seems possible crate dependencies could also access the Ruby C API, so the requirement for TruffleRuby is that anything using (diretly or indirectly) the Ruby C API must be available as bitcode.

Maybe the rb-sys/rutie crates could automatically enable `--emit=llvm-bc`/`-Clinker-plugin-lto` for themselves and for all crates depending on them?

bjorn3:
You can set the `RUSTFLAGS` env var when invoking cargo which applies to all crates being built (just not the standard library), but it isn't possible to set rustc arguments for all crates dependent on your a specific crate.

ianks:
So overall I believe this PR is in a merge-able state. The comments on https://github.com/rubygems/rubygems/pull/5175#issuecomment-1002321696 still apply, mainly in regards to more thorough testing on windows. Would love some feedback from a windows dev if possible.

@simi, can you trigger a CI build 🙇🏻

asaaki:
First thanks for the effort here, really looking forward for a nicer Ruby+Rust experience! 👏🏻 

For the folks who might want to help with the failing windows build:

From the lengthy output this is the noteworthy error:
> c:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: unrecognized option '--eh-frame-hdr'

I have tried to search for easy solutions, but the internet was not very helpful so far.
For other cross compilation targets (like AVR or ARM) it is sometimes suggested to create a new target.json with `"eh-frame-header": false`, but that feels a bit overkill to me (I don't think we really need to create a customized target just for Ruby on Windows).

There's apparently no easy way to simply disable the EH (exception handling) frames as far as I have seen so far.
And I also have no deep experience with the mingw toolchain to help you further.

asaaki:
Ah, one thing which is odd, that the rust lib is build with `stable-x86_64-pc-windows-msvc` and I'm actually not sure if MSVC and mingw work well together anyway.

As per https://rust-lang.github.io/rustup/installation/windows.html

> […] Which version of Rust you need depends largely on what C/C++ libraries you want to interoperate with: for interop with software produced by Visual Studio use the MSVC build of Rust; **for interop with GNU software built using the MinGW/MSYS2 toolchain use the GNU build.**

So I believe the toolchain for Rust needs to be changed to `stable-x86_64-pc-windows-gnu` for the Windows builds.
And with some luck that should already fix the issue. 🤞🏻

.github/workflows/windows-rubygems.yml
```diff
@@ -33,6 +33,9 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby.value }}
           bundler: none
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
asaaki, line 38: If you want to confirm my suspicion, try this:

```suggestion
          toolchain: stable
          target: x86_64-pc-windows-gnu
```

ianks:
> Ah, one thing which is odd, that the rust lib is build with `stable-x86_64-pc-windows-msvc` and I'm actually not sure if MSVC and mingw work well together anyway.
> 
> As per https://rust-lang.github.io/rustup/installation/windows.html
> 
> > […] Which version of Rust you need depends largely on what C/C++ libraries you want to interoperate with: for interop with software produced by Visual Studio use the MSVC build of Rust; **for interop with GNU software built using the MinGW/MSYS2 toolchain use the GNU build.**
> 
> So I believe the toolchain for Rust needs to be changed to `stable-x86_64-pc-windows-gnu` for the Windows builds. And with some luck that should already fix the issue. 🤞🏻

Good catch. I think it's still borking on the `eh-frame-hdr` option. I'm gonna try explicitly setting the linker flavor to see if that fixes it. https://github.com/ianks/rubygems/runs/4708209515?check_suite_focus=true#step:5:230

ianks:
Still cannot get the `--eh-frame-hdr` option to go away 🤔

test/rubygems/test_gem_ext_cargo_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+    content = @fixture_dir.join('Cargo.toml').read.gsub("cdylib", "staticlib")
+    File.write(File.join(@ext, 'Cargo.toml'), content)
+
+    output = []
+
+    Dir.chdir @ext do
+      ENV.update(@rust_envs)
+      spec = Gem::Specification.new 'rust_ruby_example', '0.1.0'
+      builder = Gem::Ext::CargoBuilder.new(spec)
+      assert_raises(Gem::Ext::CargoBuilder::DylibNotFoundError) do
```
hsbt, line 39: Use `assert_raise` instead of `assert_raises`. rubygems and ruby core don't use minitest.

test/rubygems/test_gem_ext_cargo_builder.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  def test_build_fail
+    skip_jruby!
+
+    output = []
+
+    FileUtils.rm(File.join(@ext, 'src/lib.rs'))
+
+    error = assert_raises Gem::InstallError do
```
hsbt, line 80: Use `assert_raise` too.

hsbt (changes requested):

asaaki:
@ianks from one of your latest CI runs I can see that at least the error changed now:
- https://github.com/ianks/rubygems/runs/4710304341?check_suite_focus=true#step:5:223
> note: c:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find : Invalid argument
 >          collect2.exe: error: ld returned 1 exit status

That's at least a change, but not sure if an improvement. 😅 
At this point I'm not sure if I can help any further.

asaaki (commented):
Windows issue - attempt to fix it

-----

Readded the Ruby linker config, but the error was now around "undefined reference to `rb_define_module_function'" and others.
Which kinda makes sense, though I wonder if the flag "-undefined,dynamic_lookup" even works.

The slightly bigger error looks like that:

```
"cargo rustc --target-dir D:/a/rubygems/rubygems/tmp/test_rubygems_20220105-4592-mka8hd/prefix --manifest-path D:/a/rubygems/rubygems/tmp/test_rubygems_20220105-4592-mka8hd/ext/Cargo.toml --lib --release --locked -- -C linker\\=gcc -C link-arg\\=-Wl,-undefined,dynamic_lookup -C link-arg\\=-Wl,--warn-unresolved-symbols -C link_arg\\=-pipe -C link_arg\\=-s -C link_arg\\=-fstack-protector-strong -C link_arg\\=-Wl,--enable-auto-image-base,--enable-auto-import",
"   Compiling rust_ruby_example v0.1.0 (D:\\a\\rubygems\\rubygems\\tmp\\test_rubygems_20220105-4592-mka8hd\\ext)\n" +
 "error: linking with `gcc` failed: exit code: 1\n" +
"  |\n" +
<snip>
"          c:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: D:/a/rubygems/rubygems/tmp/test_rubygems_20220105-4592-mka8hd/prefix\\release\\deps\\rust_ruby_example.rust_ruby_example.5f8e78c8-cgu.2.rcgu.o:rust_ruby_example.:(.text+0x365): undefined reference to `rb_define_module_function'\n" +
 "          collect2.exe: error: ld returned 1 exit status\n" +
 "          \n" +
 "  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified\n" +
 "  = note: use the `-l` flag to specify native libraries to link\n" +
 "  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)\n" +
 "\n" +
 "error: could not compile `rust_ruby_example` due to previous error\n"]
```

(whole output of my tinkering: https://github.com/asaaki/rubygems/runs/4718504606?check_suite_focus=true#step:5:360)

It might be just the right linker flag incantation needed, but I tried now several things and nothing worked so far (`--warn-unresolved-symbols`, `--unresolved-symbols=…`, `--allow-shlib-undefined`), it keeps telling me, that it's unhappy about those refs/symbols.

_Some googling later …_

It might be possible that it is not possible to do actual dynamic linking under Windows at all:

> On Windows you can't have dynamic_lookup, each symbols must be linked to some library. 

— comment response to https://stackoverflow.com/q/68847350/653173

If that is true we must resolve the symbols, meaning probably linking some Ruby stuff into it.

_Some tinkering later …_

That's what I did next (see `ruby_linking_args`). Now it does compile, but then it crashes during require in the `test_build_cdylib` stage, with a segfault and pages of trace info.
One example: https://github.com/asaaki/rubygems/runs/4721519646?check_suite_focus=true#step:5:218

I have to stop now, but maybe someone else has an idea how to take it from here.

Note: "CI Driven Development" is a painfully slow process, but I didn't want to screw up my system to reflect the environment of the GitHub workflow. 😬

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -0,0 +1,178 @@
...
+  def cargo_rustc_args(_dest_dir)
+    [
+      "--lib",
+      "--release",
+      "--locked",
+      "--",
+      *platform_specific_rustc_args,
+      *rustc_dynamic_linker_flags,
+    ]
+  end
```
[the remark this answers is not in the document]
ianks: I think I may have solved this by taking a closer looker at what mkmf generate on windows. The main difference is on windows, it seems the librubyarg is used:

https://github.com/rubygems/rubygems/pull/5175/files#diff-b217e7ceb9bd2292a36835e7927ac2dd197490c1d1222475b0be3b5fcea7ee91R67

ianks:
I can see the finish line! We are compiling on windows, just some small things to make sure we install the lib correctly. Follow up ci here: https://github.com/ianks/rubygems/pull/1

ianks:
We’re passing windows on ruby 2.4!! 🎊 

Not sure why newer versions are segfaulting though. @asaaki can you  take a look?

asaaki:
I can give it a try, but I guess at this point I do have to set up a local dev environment for that.
Will probably take some time, maybe this weekend.

ianks:
After some digging, I'm wondering if the segfaults on mingw are related to https://github.com/msys2/MINGW-packages/pull/6932#issuecomment-688233761

ianks:
Looks like windows checks are ✅

I used the rb-sys crate to build the bindings, and the segfaults disappeared.

Huge milestone. Want to get this across the finish line next week. 😳

ianks:
Ok so I am soooo close to getting this done. There are a few failures which I cannot quite grok.

- [ ] [Install Rubygems on Ubuntu (jruby-9.3)](https://github.com/ianks/rubygems/runs/4915243905?check_suite_focus=true) - looks unrelated?
- [ ] [Bundler on Windows (ruby-3.0)](https://github.com/ianks/rubygems/runs/4915243575?check_suite_focus=true) - looks unrelated?
- [ ] [Bundler on Ruby 3.0](https://github.com/ianks/rubygems/runs/4915243508?check_suite_focus=true) - looks unrelated?
- [ ] [Rubygems on ruby-core](https://github.com/ianks/rubygems/runs/4915243225?check_suite_focus=true) - need to make sure libruby is installed

Can someone help me get items 1-4 passing? Or let me know whether or not I can skip them? Maybe @simi or another maintainer?

test/rubygems/test_gem_ext_cargo_builder.rb
hsbt: This line was reverted to use `assert_raises` again.

test/rubygems/test_gem_ext_cargo_builder.rb
hsbt: same

deivid-rodriguez:
Master CI is green, and we don't have any intermittent failures at the moment, so all failures should be related. I think you can fix the ones about `shellworlds` by moving the `require "shellwords"` to where it's used, so that it's not required by default.

ianks:
So it looks like [ruby 3.1 installer for windows uses the new ucrt runtime](https://rubyinstaller.org/2021/12/31/rubyinstaller-3.1.0-1-released.html), which seems to be related to a [failure about missing libclang](https://github.com/rubygems/rubygems/runs/4977329501?check_suite_focus=true)

@luislavena do you have any insights about ways to fix this? namely, do we just need to install `mingw-w64-ucrt-x86_64-clang`?

luislavena:
Hello @ianks, really amazing work you have put here! ❤️ ❤️ ❤️ 

Has been a while since I was involved with RubyInstaller and I'm catching up with the recent switch to ucrt. The best person to respond these questions will be @larskanis, who is maintaining newer versions of the project.

Your last point appears to be correct, seems to be missing `libclang.dll` which is provided by that package:

https://packages.msys2.org/package/mingw-w64-ucrt-x86_64-clang

Not sure how to install these packages in CI or if libclang should be included by default in RubyInstaller DevKit. Leaving that to Lars to comment on.

Again, amazing work! 👏🏽

Have a great weekend!

.github/workflows/windows-rubygems.yml
```diff
@@ -41,6 +41,10 @@ jobs:
         with:
           toolchain: ${{ matrix.cargo.toolchain }}-${{ matrix.cargo.target }}
           default: true
+      - name: Install deps for Rust
+        if: ${{ matrix.ruby.name != "3.1" }}
+        run: |
+          pacman -S mingw-w64-x86_64-clang
```
larskanis, line 47: It's necessary to remove the `mingw-w64-x86_64-clang` package and then install the `ucrt` one like here: https://github.com/larskanis/rubygems/runs/4985946989?check_suite_focus=true#step:5:2

So this should do the trick:
```
  pacman --remove --cascade mingw-w64-x86_64-clang
  pacman --sync --noconfirm --needed $ENV:MINGW_PACKAGE_PREFIX-clang
```

The necessity to remove the non-ucrt package is something that should be fixed (in clang-sys if I understand the code right). It's possible to have both ucrt and non-ucrt packages installed and one shouldn't disturb the other one.

ianks: Wow thank you for the quick, and thorough response! Will give this a shot and report back.

ianks:
Alright, with the latest patches I'm hoping CI will be fully green. The only thing that has not been completed is support for mswin (mingw is fully supported). It seems like mswin support can be added later on. Would love to see this shipped whenever the community is comfortable!

hsbt (approved):
@ianks Thanks for your great work. It's a big step for Rust and Ruby.

simi:
I'll take a look this week. :muscle: Good job indeed.

hsbt:
> I'll take a look this week.

@simi Can we merge this now?

hsbt:
@ianks Can you check to https://github.com/rubygems/rubygems/runs/5199421392?check_suite_focus=true#step:6:35 ?

It seems to fail dependency resolution of `rb-sys` or `rust-bindgen`.

simi:
@hsbt I got a look already, did some testing and I need to polish and submit my comments now.

ianks:
Pushing the fix up now. Keep in mind `rb-sys` is waiting on https://github.com/rust-lang/rust-bindgen/pull/2122 so once that is merged I'll update the cargo deps.

simi (commented):
I left some minor comments in the code. Awesome work. I have created follow-up branch to add support to `bundle gem` skeleton generator. That way it should be super simple to bootstrap new gem. I tried to keep it close to common C extension structure.

https://github.com/rubygems/rubygems/commit/548189dd018f2e1a66525c168ada9868f8100f2e

Example of usage:

```
[retro@retro  bundler (cargo-newgem *%=)]❤ exe/bundle gem --ext=rust pivo
Creating gem 'pivo'...
MIT License enabled in config
Initializing git repo in /home/retro/code/work/oss/rubygems/bundler/pivo
      create  pivo/Gemfile
      create  pivo/lib/pivo.rb
      create  pivo/lib/pivo/version.rb
      create  pivo/sig/pivo.rbs
      create  pivo/pivo.gemspec
      create  pivo/Rakefile
      create  pivo/README.md
      create  pivo/bin/console
      create  pivo/bin/setup
      create  pivo/.gitignore
      create  pivo/test/test_helper.rb
      create  pivo/test/test_pivo.rb
      create  pivo/LICENSE.txt
      create  pivo/Cargo.toml
      create  pivo/ext/pivo.rs
Gem 'pivo' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html

[retro@retro  bundler (cargo-newgem *%=)]❤ cd pivo/
```

From this point you can package the gem and install.

```
[retro@retro  pivo (master +)]❤ cargo generate-lockfile
    Updating git repository `https://github.com/ianks/rb-sys`
    Updating crates.io index
    Updating git repository `https://github.com/ianks/rust-bindgen`

[retro@retro  pivo (master +%)]❤ git add Cargo.lock # needs to be in git to be packed into gem
[retro@retro  pivo (master +%)]❤ gem build --force
  Successfully built RubyGem
  Name: pivo
  Version: 0.1.0
  File: pivo-0.1.0.gem

[retro@retro  pivo (master +%)]❤ ruby -I../../lib ../../bin/gem install --local pivo-0.1.0.gem
Successfully installed pivo-0.1.0
Parsing documentation for pivo-0.1.0
Installing ri documentation for pivo-0.1.0
Done installing documentation for pivo after 0 seconds
1 gem installed

[retro@retro  pivo (master +%)]❤ irb
irb(main):001:0> require 'pivo'
=> true
irb(main):002:0> Pivo.hello('world')
=> "Hello world!"
```

I have included also initial simple rake task for local development (similar to the one generated to C extensions).

```ruby
# frozen_string_literal: true

require "bundler/gem_tasks"
require "rake/testtask"

Rake::TestTask.new(:test) do |t|
  t.libs << "test"
  t.libs << "lib"
  t.test_files = FileList["test/**/test_*.rb"]
end

task :compile do
  sh "cargo build"
  sh "cp target/debug/libpivo_ext.#{RbConfig::CONFIG["SOEXT"]} lib/pivo_ext.#{RbConfig::CONFIG["SOEXT"]}"
end

CLOBBER << "lib/pivo_ext.#{RbConfig::CONFIG["SOEXT"]}"

task default: %i[clobber compile test]
```

```
[retro@retro  pivo (master +%)]❤ rake compile
cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.03s
cp target/debug/libpivo_ext.so lib/pivo_ext.so

[retro@retro  pivo (master *+%)]❤ rake test # it fails by default, but successfully loads compiled extension in lib directory
Run options: --seed 16963

# Running:

F.

Finished in 0.000932s, 2145.7292 runs/s, 2145.7292 assertions/s.

  1) Failure:
TestPivo#test_it_does_something_useful [/home/retro/code/work/oss/rubygems/bundler/pivo/test/test_pivo.rb:11]:
Expected false to be truthy.

2 runs, 2 assertions, 1 failures, 0 errors, 0 skips
rake aborted!

Tasks: TOP => test
```

---

I have almost no experience with Rust. It would be nice to review example code I wrote at https://github.com/rubygems/rubygems/commit/548189dd018f2e1a66525c168ada9868f8100f2e#diff-0121b86bd080a75ec219e1e41566ca43596b3743b9ea63e3347dc21fd5e6ae26R1-R31.

It is currently missing also validation for rust_package gemspec entry. We can also ensure `Cargo.lock` file is present in gem if `Cargo.toml` file is present. We can do that in gemspec validations.

ianks:
@simi Thanks for you review! I have addressed your comments and I think they helped a lot.

Instead of `rust_package`, I used the name `cargo_crate_name` to be more consistent with Rust's wording.

Are there any other things I should address??

simi:
@ianks I think once CI is green, we can merge this and continue with https://github.com/rubygems/rubygems/commit/548189dd018f2e1a66525c168ada9868f8100f2e as a next step.

ianks:
@peterzhu2118 Can you give your input on the best option here for Rust extensions?

peterzhu2118:
@ianks and I had a conversation about this through private DMs. To summarize the conclusion we came with:

- Using `rust-bindgen`, `rb-sys` is able to get the value of the [`RUBY_ABI_VERSION`](https://github.com/ruby/ruby/blob/bec492c77ed7659cafd2447cd042acde489c8d28/include/ruby/internal/abi.h#L25) macro defined in `abi.h`.
- Using the value obtained from this macro, it can define an exported `ruby_abi_version` function. This will get ABI checking working.

simi:
What does it mean? Is change at rb-sys needed? Do we need to scope rb-sys per Ruby version? Or is that evaluated during the compilation time? Who is going to handle this? Would that be backwards compatible for older Rubies?

---

:pray: _@ianks @peterzhu2118 feel free to discuss the stuff in the public at RubyGems Slack next-time so I can follow and there is no need to ask most likely obvious questions. Link can be found at the bottom of https://bundler.io/._

ianks:
> What does it mean? Is change at rb-sys needed? Do we need to scope rb-sys per Ruby version? Or is that evaluated during the compilation time? Who is going to handle this? Would that be backwards compatible for older Rubies?

We do not need to scope rb-sys per version. I think the best way to handle this is to have rb-sys generate the `ruby_api_version` function export. That way, users do not have to worry about the internal implementation details when using the crate. `rb-sys` already provides all of the functionality provided in `ruby.h`, so I think it makes sense for it to be added there

ianks:
Update: following in the [footsteps of pgx](https://github.com/zombodb/pgx/blob/master/pgx/src/lib.rs#L245), I've setup a macro for rb-sys named `rb_module_magic!()` which does the setup needed to export the `ruby_abi_version` function needed for Ruby 3.2+ (thanks for the guidance @peterzhu2118!). Gem authors will do something like this:

```rust
#[macro_use]
extern crate rb_sys;

rb_module_magic()!;

#[allow(non_snake_case)]
#[no_mangle]
pub extern "C" fn Init_my_gem() {
    // ...
}
```

This will set everything up needed to load the extension at runtime.

ianks:
[ruby-core](https://github.com/rubygems/rubygems/actions/workflows/ruby-core.yml) 🎊

simi:
done :pray:

simi:
Let me please check the code once more time. I think there is some chance to DRY some stuff shared across builders.

hsbt (approved):

ianks:
Leaving this here as a note if anyone wants to give cross-compilation a shot after this is merged... I think the mapping to target triples looks something like this:

```ruby
    MAPPING = Hash[
      "x86-mingw32" => "i686-pc-windows-gnu",
      "x64-mingw-ucrt"  => "x86_64-pc-windows-gnu",
      "x64-mingw32"  => "x86_64-pc-windows-gnu",
      "x86-linux"  => "i686-unknown-linux-gnu",
      "x86_64-linux" => "x86_64-unknown-linux-gnu",
      "arm-linux"  => "arm-unknown-linux-gnueabihf",
      "aarch64-linux" => "aarch64-unknown-linux-gnu",
      "x86_64-darwin" => "x86_64-apple-darwin",
      "arm64-darwin" => "aarch64-apple-darwin",
    ].freeze
```

hsbt:
@simi I hope to merge this before Ruby 3.2.0-preview1 release. Can we refactor/improve builders after merging this?

simi:
Great job everyone. I'll open additional PR with `bundler gem --ext=rust` support and various improvements.

simi:
Btw. I would consider this still experimental feature for now.

hsbt:
@simi Thanks!

ianks:
Awesome!! Thanks for the help everyone ❤️ 

Now onto making cross-compilation with rake-compiler as breezy as possible, so rubygems users don’t have to have a cargo tool chain installed!

pynixwang:
how about dsl
```
#[ruby_module(BigDecimal)]
mod bigdecimal {
}
```

ilyazub:
> how about dsl
> 
> ```
> #[ruby_module(BigDecimal)]
> mod bigdecimal {
> }
> ```

@pynixwang Looks like the [`rutie`](https://github.com/danielpclark/rutie) and macros from the [`wasmer-ruby`](https://github.com/wasmerio/wasmer-ruby/blob/4cbecd79faf32da7d4cb953a4a0474c2b01130e9/crates/rutie-derive/README.md#rust-defined-ruby-types). ([Usage example of `wasmer-ruby` macros](https://github.com/wasmerio/wasmer-ruby/blob/bcc8a79fa32cc656db063daa785edeba901c7772/crates/wasmer/src/lib.rs))

```rust
#[rubyclass(module = "Wasmer")]
pub struct Baz {
    inner: i32,
};

#[rubymethods]
impl Foo {
    pub fn bar(&self, baz: &Baz) -> RubyResult<…> {
        let inner = &baz.inner;

        …
    }
}
```

While `rutie` is ready for use, macros from `wasmer-ruby` are not exported. Help is much appreciated: https://github.com/danielpclark/rutie/issues/145.

PR #5177, opened 2021-12-19
Print warning when running Bundler on potentially problematic RubyGems & Ruby combinations

deivid-rodriguez:

Users should move on and stop using old rubies, but we shouldn't abruptly drop support in `bundler`. Although doing this is generally fine to do for gems on minor versions, bundler is a bit special, because many scripts include `gem install bundler` commands, and old rubygems versions (which old rubies use by default) didn't have the ability to figure out that the latest bundler does not support the current ruby, so those `gem install bundler` commands will start failing as soon as a version of `bundler` no longer supporting old rubies is released.

In order to make sure people get ready for this, start by printing a warning when `bundler` is run on old rubies, and explain which steps users should be taking.

### DIFF
--- a/bundler/exe/bundle
+++ b/bundler/exe/bundle
@@ -18,6 +18,15 @@ end
 # Workaround for non-activated bundler spec due to missing https://github.com/rubygems/rubygems/commit/4e306d7bcdee924b8d80ca9db6125aa59ee4e5a3
 gem "bundler", Bundler::VERSION if Gem.rubygems_version < Gem::Version.new("2.6.2")
 
+if Gem.rubygems_version < Gem::Version.new("3.2.3") && Gem.ruby_version < Gem::Version.new("2.6.a") && !ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"]
+  Bundler.ui.warn \
+    "Your RubyGems version (#{Gem::VERSION})) has a bug that prevents " \
+    "`required_ruby_version` from working for Bundler. Any scripts that use " \
+    "`gem install bundler` will break as soon as Bundler drops support for " \
+    "your Ruby version. Please upgrade RubyGems to avoid future breakage " \
+    "and silence this warning by running `gem update --system 3.2.3`"
+end
+
 if File.exist?(base_path)
   require_relative "../lib/bundler/friendly_errors"
 else
--- a/bundler/exe/bundle
+++ b/bundler/exe/bundle
@@ -18,16 +18,6 @@ end
 # Workaround for non-activated bundler spec due to missing https://github.com/rubygems/rubygems/commit/4e306d7bcdee924b8d80ca9db6125aa59ee4e5a3
 gem "bundler", Bundler::VERSION if Gem.rubygems_version < Gem::Version.new("2.6.2")
 
-# Check if an older version of bundler is installed
-$LOAD_PATH.each do |path|
-  next unless path =~ %r{/bundler-0\.(\d+)} && $1.to_i < 9
-  err = String.new
-  err << "Looks like you have a version of bundler that's older than 0.9.\n"
-  err << "Please remove your old versions.\n"
-  err << "An easy way to do this is by running `gem cleanup bundler`."
-  abort(err)
-end
-
 if File.exist?(base_path)
   require_relative "../lib/bundler/friendly_errors"
 else
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -239,7 +239,7 @@ module Bundler
       current  = File.expand_path(SharedHelpers.pwd).tap{|x| x.untaint if RUBY_VERSION < "2.7" }
 
       until !File.directory?(current) || current == previous
-        if ENV["BUNDLE_SPEC_RUN"]
+        if ENV["BUNDLER_SPEC_RUN"]
           # avoid stepping above the tmp directory when testing
           gemspec = if ENV["GEM_COMMAND"]
             # for Ruby Core
--- a/bundler/spec/commands/config_spec.rb
+++ b/bundler/spec/commands/config_spec.rb
@@ -425,27 +425,27 @@ E
 
   describe "subcommands" do
     it "list" do
-      bundle "config list"
-      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\nspec_run\nSet via BUNDLE_SPEC_RUN: \"true\""
+      bundle "config list", :env => { "BUNDLE_FOO" => "bar" }
+      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\nfoo\nSet via BUNDLE_FOO: \"bar\""
 
-      bundle "config list", :parseable => true
-      expect(out).to eq "spec_run=true"
+      bundle "config list", :env => { "BUNDLE_FOO" => "bar" }, :parseable => true
+      expect(out).to eq "foo=bar"
     end
 
     it "list with credentials" do
       bundle "config list", :env => { "BUNDLE_GEMS__MYSERVER__COM" => "user:password" }
-      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\ngems.myserver.com\nSet via BUNDLE_GEMS__MYSERVER__COM: \"user:[REDACTED]\"\n\nspec_run\nSet via BUNDLE_SPEC_RUN: \"true\""
+      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\ngems.myserver.com\nSet via BUNDLE_GEMS__MYSERVER__COM: \"user:[REDACTED]\""
 
       bundle "config list", :parseable => true, :env => { "BUNDLE_GEMS__MYSERVER__COM" => "user:password" }
-      expect(out).to eq "gems.myserver.com=user:password\nspec_run=true"
+      expect(out).to eq "gems.myserver.com=user:password"
     end
 
     it "list with API token credentials" do
       bundle "config list", :env => { "BUNDLE_GEMS__MYSERVER__COM" => "api_token:x-oauth-basic" }
-      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\ngems.myserver.com\nSet via BUNDLE_GEMS__MYSERVER__COM: \"[REDACTED]:x-oauth-basic\"\n\nspec_run\nSet via BUNDLE_SPEC_RUN: \"true\""
+      expect(out).to eq "Settings are listed in order of priority. The top value will be used.\ngems.myserver.com\nSet via BUNDLE_GEMS__MYSERVER__COM: \"[REDACTED]:x-oauth-basic\""
 
       bundle "config list", :parseable => true, :env => { "BUNDLE_GEMS__MYSERVER__COM" => "api_token:x-oauth-basic" }
-      expect(out).to eq "gems.myserver.com=api_token:x-oauth-basic\nspec_run=true"
+      expect(out).to eq "gems.myserver.com=api_token:x-oauth-basic"
     end
 
     it "get" do
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -72,6 +72,7 @@ RSpec.configure do |config|
     require_relative "support/rubygems_ext"
     Spec::Rubygems.test_setup
     ENV["BUNDLER_SPEC_RUN"] = "true"
+    ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"] = "true"
     ENV["BUNDLE_USER_CONFIG"] = ENV["BUNDLE_USER_CACHE"] = ENV["BUNDLE_USER_PLUGIN"] = nil
     ENV["RUBYGEMS_GEMDEPS"] = nil
     ENV["XDG_CONFIG_HOME"] = nil
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -71,7 +71,7 @@ RSpec.configure do |config|
 
     require_relative "support/rubygems_ext"
     Spec::Rubygems.test_setup
-    ENV["BUNDLE_SPEC_RUN"] = "true"
+    ENV["BUNDLER_SPEC_RUN"] = "true"
     ENV["BUNDLE_USER_CONFIG"] = ENV["BUNDLE_USER_CACHE"] = ENV["BUNDLE_USER_PLUGIN"] = nil
     ENV["RUBYGEMS_GEMDEPS"] = nil
     ENV["XDG_CONFIG_HOME"] = nil


bundler/exe/bundle
```diff
@@ -7,39 +7,26 @@ Signal.trap("INT") do
...
 
-# Check if an older version of bundler is installed
-$LOAD_PATH.each do |path|
-  next unless path =~ %r{/bundler-0\.(\d+)} && $1.to_i < 9
-  err = String.new
-  err << "Looks like you have a version of bundler that's older than 0.9.\n"
-  err << "Please remove your old versions.\n"
-  err << "An easy way to do this is by running `gem cleanup bundler`."
-  abort(err)
+if RUBY_VERSION < "2.6.0.a" && !ENV["BUNDLE_NO_OLD_RUBIES_WARNING"]
```
simi, line 15: Can we document this at https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/man/bundle-config.1.ronn#list-of-available-keys?

deivid-rodriguez: I wasn't sure to make this official, the only reason I added it is because otherwise making CI green is going to be really painful, since we constantly check output.

I guess if we want to make it official we can mention in in the warning and that would be enough?

simi: sure

deivid-rodriguez: I changed the name so that the environment variable is not listed with settings, since it's not a setting. Do you think it's necessary to advertise the environment variable, then?

deivid-rodriguez:
Marking as ready, pending the specific copy of the warning to be decided.

bundler/exe/bundle
```diff
@@ -18,14 +18,15 @@ end
...
-$LOAD_PATH.each do |path|
-  next unless path =~ %r{/bundler-0\.(\d+)} && $1.to_i < 9
-  err = String.new
-  err << "Looks like you have a version of bundler that's older than 0.9.\n"
-  err << "Please remove your old versions.\n"
-  err << "An easy way to do this is by running `gem cleanup bundler`."
-  abort(err)
+if RUBY_VERSION < "2.6.0.a" && !ENV["BUNDLER_NO_OLD_RUBIES_WARNING"]
+  Bundler.ui.warn \
+    "Running #{Bundler::VERSION} on Ruby #{RUBY_VERSION} is deprecated, " \
```
simi, line 23: :thinking: This could print wrong info on non CRuby implementations. For example on JRuby 9.2.20.1 (released 1st of December this year) `RUBY_VERSION` is `2.5.7` and AFAIK it is not EOL.

Maybe we should check if CRuby is used and print this detailed info. And fallback to more general message when on other implementation.

@enebo @eregon any ideas?

deivid-rodriguez: Oh, nice catch! The good news is that `JRuby` is shipping a recent enough rubygems version, so `gem install bundler` won't break there when we drop support for old rubies. So I think it would be fine to only bug CRuby users.

deivid-rodriguez: Well, thanks actually incorrect, that improvement was released with rubygems 3.2.3, and jruby 9.2.20.1 ships with 3.1.6. So... forget about what I said.

deivid-rodriguez: I'm thinking of skipping the warning on either Ruby 2.5 or JRuby for now. Not sure if JRuby plans to release one more 9.2 version. If they could release one last version including updated RubyGems, the would be ideal and we could skip this warning on JRuby.

eregon: cc @headius 
The part about "reached its End of Life and no longer receives security fixes" is only true for CRuby.

I feel "EOL Rubies" here are used as a reason to not support them in Bundler anymore, but I think the real reason is Bundler doesn't want to support older Ruby versions because it occurs extra maintenance efforts (that makes sense).
I feel it's not Bundler's place to warn about EOL Rubies, the last gem which did this (incorrectly) was very annoying: https://github.com/e2/ruby_dep

IMHO, if gems want a minimum version, they should (just) use `required_ruby_version` (and a new major or minor version).
I get a warning is useful here given the issue you mention with RubyGems.
So I think the message should be more about Bundler's choice for a minimum Ruby version than any Ruby EOL or anything like that.

eregon: i.e., I think the message should be like "Bundler x.y will require Ruby >= a.b; Due to a bug in RubyGems < c.d `gem install bundler` will not work correctly and `gem install bundler:'< 2.3.0'` is needed. Please update your setup to use bundler < 2.3.0 when running on Ruby x.y.z".

deivid-rodriguez: You are right about the maintenance reason to drop support, but not using insecure versions is also another reason, much easier to explain to users.

Your suggestion feels too complicated to me, specially since users might actually be running a proper rubygems version when they get it. We could skip the warning in that case, but I think we want all users of old rubies to read this, because they might still have scripts, or CI configurations subject to the `gem install bundler` problem.

I feel focusing on old rubies is simpler to explain and understand.

deivid-rodriguez: > I feel it's not Bundler's place to warn about EOL Rubies, the last gem which did this (incorrectly) was very annoying: https://github.com/e2/ruby_dep

The way I see it, part of the idea behind doing this is to be annoying 😆. Be a bit annoying now, so that we avoid being really annoying later when a lot of things break when we drop support.

eregon: > Your suggestion feels too complicated to me, specially since users might actually be running a proper rubygems version when they get it.

I think my suggestion is simply the right thing to do, anything else feels inappropriate (it's not Bundler's job to tell people about old Ruby versions, they're very likely already aware).

> We could skip the warning in that case

I think it's important to only warn for this specific case and no other.

> , but I think we want all users of old rubies to read this, because they might still have scripts, or CI configurations subject to the `gem install bundler` problem.

Those people already know they are on old versions, they'll just be annoyed to see a warning they typically can't do anything about (i.e., if they already have a recent enough RubyGems).
If they plan to migrate to a newer Ruby it's unlikely Bundler will be the only/main reason for it.

> I feel focusing on old rubies is simpler to explain and understand.

It's also completely hiding the real issue and reason for this warning.
For me it's a clear case of don't assume users are dumb, let them make their choices, explain the problem clearly, and don't use something else as an excuse.

Bundler wants to drop support for Ruby < 2.6, that's fine, but Bundler should own it and explain why Bundler warns here (and not just change `required_ruby_version`), not use something which is not the real reason.

eregon: > The way I see it, part of the idea behind doing this is to be annoying laughing. Be a bit annoying now, so that we avoid being really annoying later when a lot of things break when we drop support.

It's good to warn when people need to do something about it and to avoid breakage. It's bad to warn in other cases.
And the reason needs to be clear, "you're using EOL Rubies" will I think only annoy people but they typically won't do anything about it if that's the main thing mentioned in the warning.
"Bundler x.y will drop support for Ruby < 2.6", that's something everyone can understand and makes sense, there is no "you should not use EOL Rubies" (I've tried the EOL argument for setup-ruby, it's always a waste of time, some people made the choice to use an EOL Ruby, the only thing that works is to set required Ruby versions for gems).

deivid-rodriguez: > I think my suggestion is simply the right thing to do, anything else feels inappropriate (it's not Bundler's job to tell people about old Ruby versions, they're very likely already aware).

I can see how your suggestions very usually feel like the right thing to do to you 🤣. It happens to all of us.

Anyways, you have a point here, I'll try to reword the warning to be easy to understand and yet properly explain things.

eregon: :D
Another way to phrase this:
What is the most likely to get users to change to `gem install bundler:'< 2.3.0'` when needed?
* To tell them they use an EOL Ruby version and more things
* To tell them it'll break with Bundler x.y if they don't do it due to an issue in RubyGems a.b

eregon: It should suggest `< 2.3.0` as that then addresses the warning.

> Also, should we show the message for users running a "safe" version of rubygems? Maybe we can skip it there?

Yes I believe we should skip it if there is no need to change anything and it just works, and just verify it indeed works as expected on those RubyGems versions (probably already checked in tests?).

deivid-rodriguez: If they are already running a higher version of `bundler`, that won't address the warning because they will most likely keep using that version by default. I think the best thing to suggest is actually `gem update --system 3.3.0`. That will address the issue and make the warning not reappear since they will now have a modern enough rubygems version. Suggesting an upgrade is also better for us maintainers than suggesting a downgrade.

simi: > IMHO, if gems want a minimum version, they should (just) use required_ruby_version (and a new major or minor version).

That's actually next step once deprecation message would be around long enough. This message is just developer friendly semi-step in between.

- `<=` 2.2.x series is the last one supported on old Rubies without any warning
- `=` 2.3.x is last series supporting old Rubies, but with warning (maybe before 2.4+ release we can release last 2.3 without this warning)
- `>=` 2.4.x will drop support for old Rubies (and if possible will make it clear in `required_ruby_version`)

deivid-rodriguez: @simi What do you think about focusing on the rubygems version rather than on the ruby version? I think it's better because it's more clear, and also it's independent from whether `RUBY_VERSION` is considered unsupported or not for specific implementations.

How about something like this

> Your rubygems version (x.y.z) has a bug that makes dropping support for old rubies in bundler through bumping the `required_ruby_version` field in the bundler gemspec unsafe for any scripts including `gem install bundler`. That means that, if those scripts are using rubygems x.y.z, they will break as soon as bundler releases a version dropping support for your ruby version (a.b.c). Please upgrade rubygems through `gem update --system` to avoid future breakage.

deivid-rodriguez: Why downgrading a software, and thus giving up on all improvements and bug fixes, would be better than upgrading a different software, and thus opting in to all improvements and bug fixes? We run a full rubygems CI for every commit against all supported versions, including ruby 2.3, ruby 2.4, and ruby 2.5.

I think you should add support for upgrading rubygems in ruby/setup-ruby, and ideally ship a version higher than 3.2.3 on old rubies, so that users that do `gem install bundler` are safe.

eregon: > Why downgrading a software, ... would be better than upgrading a different software

Because either way Bundler < 2.3 is needed on those Rubies, so anyway there is a downgrade or pinning involved.
And then the difference is installing a new RubyGems including its new features/fixes/changes/bugs or not.

(what will happen when Bundler 2.3.0 is installed with RubyGems 3.3.0? Does it ignore that gem, or does it use it because it's more recent and so would still warn and would need uninstall as well to not warn?)

If people are on EOL Rubies I'd think they prefer to change as little as possible.
Also people might not like to update RubyGems (e.g. I never do it, and TruffleRuby doesn't support it because it's slow in CI, introducing instability/harder to reproduce, and too risky for too little gains, and some CIs might also not support updating RubyGems because it means writing to the Ruby prefix). I guess you already know how I feel about `gem update --system` installing Bundler in `site_ruby`.

> I think you should add support for upgrading rubygems in ruby/setup-ruby

I disagree, but people can do it themselves and probably in the future it'll possible to still use bundler-caching with that (by having a separate action for bundler caching).

> ideally ship a version higher than 3.2.3 on old rubies, so that users that do gem install bundler are safe.

I can already say this will not happen because it'd be a lot of work and it would differ for people just installing Ruby locally which would be a reproducibility issue.

deivid-rodriguez: Why would we print a message like

> Versions of software X older than a.b.c have a bug that might affect you, please downgrade software Y to a version older than d.e.f

rather than

> Versions of software X older than a.b.c have a bug that might affect you, please upgrade software X to a version higher than or equal to a.b.c

Rephrasing your own arguments, it feels like another case of assuming your users are dumb and hiding the real motivations for things.

eregon: I guess you wrote this before my reply. I think my reply before addresses that, either way it's a downgrade of Bundler to < 2.3, hence the first is clearer but also a smaller and safer change.

deivid-rodriguez: > Because either way Bundler < 2.3 is needed on those Rubies, so anyway there is a downgrade or pinning involved.

That doesn't make any sense, if we print this warning when running bundler 2.3 on those rubies, Bundler < 2.3 is definitely not needed.

> If people are on EOL Rubies I'd think they prefer to change as little as possible.

Yet, the people we're targeting here are installing the latest bundler. I don't think this kind of assumptions are accurate.

> Also people might not like to update RubyGems (e.g. I never do it, and TruffleRuby doesn't support it because it's slow in CI, introducing instability/harder to reproduce, and too risky for too little gains, and some CIs might also not support updating RubyGems because it means writing to the Ruby prefix). I guess you already know how I feel about gem update --system installing Bundler in site_ruby.

Alright! Now I see your real motivation for recommending downgrading bundler rather than upgrading rubygems. It seems that you're not following your own motto of being honest to users and not hiding your actual motivations 😅.

Anyways, I think we're going to start going round in circles at this point. My last proposal seems fine to me, if @simi and other maintainers agree I think it's fine to move on with that.

deivid-rodriguez: > I disagree, but people can do it themselves and probably in the future it'll possible to still use bundler-caching with that (by having a separate action for bundler caching).

> I can already say this will not happen because it'd be a lot of work and it would differ for people just installing Ruby locally which would be a reproducibility issue.

Alright, I respect your decisions about what to do with `ruby/setup-ruby`. Thanks for maintaining it by the way, it's great.

eregon: > That doesn't make any sense, if we print this warning when running bundler 2.3 on those rubies, Bundler < 2.3 is definitely not needed.

It is to avoid the warning and also make it clear to use a bundler version supporting that old Ruby version.
One subtlety is by updating RubyGems you'd be able to use bundler >= 2.3 and < 2.4, I didn't consider that.
Not sure how much that really matters.
Either way Bundler >= 2.4 can't be used on those older Rubies, hence Bundler will be effectively pinned one way or another.

> Alright! Now I see your real motivation for recommending downgrading bundler rather than upgrading rubygems. It seems that you're not following your own motto of being honest to users and not hiding your actual motivations 

I don't think I was hiding anything, I have many reasons to never update RubyGems, I just tried to present the ones relevant to this discussion. Not being able to reproduce when e.g. `ruby-build/install x.y.z` is maybe the biggest of them in this context.

I also remember when people were updating RubyGems for Bundler 2 even though it wasn't necessary, and I wrote https://eregon.me/blog/2020/01/13/a-migration-path-to-bundler2.html#the-rubygems-requirement-of-bundler-2 about it. I remember various issues and confusing docs on TravisCI related to that.
I think it would be wiser to not only suggest people to update RubyGems, but tell them to either update RubyGems or pin Bundler to < 2.3.0 (for old Rubies).
(For me the second is better but I can see others have different opinions on the matter)

bundler/exe/bundle
```diff
@@ -18,14 +18,16 @@ end
...
+if Gem::VERSION < "3.2.3" && RUBY_VERSION < "2.6.a" && !ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"]
+  Bundler.ui.warn \
+    "Your rubygems version (#{Gem::VERSION})) has a bug that makes dropping " \
+    "support for old rubies in bundler through bumping the " \
+    "`required_ruby_version` field in the bundler gemspec unsafe for any " \
+    "scripts including `gem install bundler`. That means that, if those " \
+    "scripts are using rubygems #{Gem::VERSION}, they will break as soon " \
+    "as bundler releases a version dropping support for your ruby version. " \
+    "Please upgrade rubygems by running `gem update --system` to avoid future " \
+    "breakage and silence this warning."
```
eregon, line 30:
```suggestion
    "breakage and silence this warning, or use `gem install bundler:'< 2.3.0'`."
```
The new message is much better to explain the issue, thanks :)

bundler/exe/bundle
deivid-rodriguez: Well, as per https://github.com/rubygems/rubygems/pull/5177#discussion_r772481039 you're suggesting a solution different from the straightforward one (upgrade rubygems to fix a bug in rubygems), and you're not explaining the reasons behind it (upgrading rubygems is "bad").

I'm leaning towards only suggesting a rubygems upgrade. I think any maintainer would sympathize with preferring to suggest upgrading the software that you maintain that to suggest downgrading the software that you also maintain. The bundler downgrade is also more complicated, not only conceptually, but because it doesn't remove the warning by itself, you also need to uninstall the currently running bundler, and that might not be straightforward for some users, so it would probably need further clarification.

bundler/exe/bundle
enebo: I was late to the party but it looks like this is a reasonable message and spells out the two main options.  I agree with @eregon that if you are explicitly pinned to an older version of Ruby (whether from JRuby still supporting it or other reasons) you would find the warning annoying.  I also agree it is nice to give a heads up that something will break at some future point.

bundler/exe/bundle
```diff
@@ -18,14 +18,17 @@ end
...
 
-# Check if an older version of bundler is installed
-$LOAD_PATH.each do |path|
-  next unless path =~ %r{/bundler-0\.(\d+)} && $1.to_i < 9
-  err = String.new
-  err << "Looks like you have a version of bundler that's older than 0.9.\n"
-  err << "Please remove your old versions.\n"
-  err << "An easy way to do this is by running `gem cleanup bundler`."
-  abort(err)
+if Gem::VERSION < "3.2.3" && RUBY_VERSION < "2.6.a" && !ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"]
```
indirect, line 21: I think we need to use Gem::Versions here. For example, `"3.10.0" < "3.2.3" => true`.

bundler/exe/bundle
```diff
@@ -18,14 +18,17 @@ end
...
+  Bundler.ui.warn \
+    "Your rubygems version (#{Gem::VERSION})) has a bug that makes dropping " \
+    "support for old rubies in bundler through bumping the " \
+    "`required_ruby_version` field in the bundler gemspec unsafe for any " \
+    "scripts including unconstrained `gem install bundler` commands. That " \
+    "means that, if those scripts are using rubygems #{Gem::VERSION}, they " \
+    "will break as soon as bundler releases a version dropping support for " \
+    "your Ruby version. Please upgrade rubygems to a safe version by running " \
+    "`gem update --system 3.2.3` to avoid future breakage and silence this " \
+    "warning."
```
indirect, line 31:
```suggestion
  Bundler.ui.warn \
    "Your RubyGems version (#{Gem::VERSION})) has a bug that prevents " \
    "`required_ruby_version` from working for Bundler. Any scripts that use " \
    "`gem install bundler` will break as soon as Bundler drops support for " \
    "your Ruby version. Please upgrade RubyGems to avoid future breakage " \
    "and silence this warning by running `gem update --system 3.2.3`"
```

Tried to get straight to the problem while keeping your wording. Open to pushback, though. :)

indirect (approved):
Thanks for setting this up! I think this will reduce future support by a lot. 👍🏻

bundler/exe/bundle
deivid-rodriguez: Nice catch, I'll update!

bundler/exe/bundle
deivid-rodriguez: I really like it!

bundler/exe/bundle
deivid-rodriguez: Addressed!

bundler/exe/bundle
deivid-rodriguez: The recommendation of downgrading bundler requires further actions to work, so it will need further clarifications. And I'm not a fan of recommending downgrading the software that I help maintain anyways. Educated users can do this if they please even if we don't directly recommended, so I think it's fine as is.

bundler/exe/bundle
deivid-rodriguez: Let's leave this as is, if the need to document the environment variable appears later, we can always do that in later iterations.

deivid-rodriguez:
@indirect Good to go now? Feel free to enable auto merge if you are satisfied with this.

deivid-rodriguez:
I'm reading the `gem update --system` code now, and I'm pretty sure it still suffers from the same issue already fixed for unconstrained `gem install` :(

Luckily we recommended a constrained `gem update --system 3.2.3` in the warning text, and unconstrained `gem update --system` is much less common that `gem install bundler` in scripts/CI anyways.

Providing a nice upgrade path for this other issue feels much harder.

One idea that comes to mind is to update the warning text to also mention that `gem update --system` commands need to be updated to be constrained, and to add some docs/blog post about this to get people ready for next year.

Another idea would be to keep support for old rubies in rubygems for now, while dropping it in bundler.

And of course, the issue needs to be fixed, but the problem is to propagate it to old rubies.

PR #5181, opened 2021-12-19
Remove hard errors when matching major bundler not found

### LINKED ISSUE #2671: Updating rubygems triggers irreversible bundler upgrade

### LINKED ISSUE #2800: Ability to disable all Bundler version checks

### LINKED ISSUE #3275: Bundler 2.1.0 raising error when it doesn't need to

### LINKED ISSUE #3283: Bundler 2.1.0 breaks bundle install in projects with Gemfile.lock or older Rubies

### LINKED ISSUE #3334: "bundle --help" not working when Gemfile.lock contains "BUNDLED WITH    2.0.0" but I don't have it installed

deivid-rodriguez:

Since we will now have auto-installing and auto-switching to the `BUNDLED WITH` version in the lockfike, we can completely remove these hard error from RubyGems.
 
Remove all the code that we no longer need.

Fixes #2671.
Supersedes #2696.
Closes #2800.
Fixes #3275.
Fixes #3283.
Fixes #3334.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -272,9 +272,6 @@ module Gem
 
     unless spec = specs.first
       msg = "can't find gem #{dep} with executable #{exec_name}"
-      if dep.filters_bundler? && bundler_message = Gem::BundlerVersionFinder.missing_version_message
-        msg = bundler_message
-      end
       raise Gem::GemNotFoundException, msg
     end
 
--- a/lib/rubygems/bundler_version_finder.rb
+++ b/lib/rubygems/bundler_version_finder.rb
@@ -2,48 +2,18 @@
 
 module Gem::BundlerVersionFinder
   def self.bundler_version
-    version, _ = bundler_version_with_reason
+    v = ENV["BUNDLER_VERSION"]
 
-    return unless version
+    v ||= bundle_update_bundler_version
+    return if v == true
 
-    Gem::Version.new(version)
+    v ||= lockfile_version
+    return unless v
+
+    Gem::Version.new(v)
   end
 
-  def self.bundler_version_with_reason
-    if v = ENV["BUNDLER_VERSION"]
-      return [v, "`$BUNDLER_VERSION`"]
-    end
-    if v = bundle_update_bundler_version
-      return if v == true
-      return [v, "`bundle update --bundler`"]
-    end
-    v, lockfile = lockfile_version
-    if v
-      return [v, "your #{lockfile}"]
-    end
-  end
-
-  def self.missing_version_message
-    return unless vr = bundler_version_with_reason
-    <<-EOS
-Could not find 'bundler' (#{vr.first}) required by #{vr.last}.
-To update to the latest version installed on your system, run `bundle update --bundler`.
-To install the missing version, run `gem install bundler:#{vr.first}`
-    EOS
-  end
-
-  def self.compatible?(spec)
-    return true unless spec.name == "bundler".freeze
-    return true unless bundler_version = self.bundler_version
-
-    spec.version.segments.first == bundler_version.segments.first
-  end
-
-  def self.filter!(specs)
-    return unless bundler_version = self.bundler_version
-
-    specs.reject! {|spec| spec.version.segments.first != bundler_version.segments.first }
-
+  def self.prioritize!(specs)
     exact_match_index = specs.find_index {|spec| spec.version == bundler_version }
     return unless exact_match_index
 
@@ -68,12 +38,10 @@ To install the missing version, run `gem install bundler:#{vr.first}`
   private_class_method :bundle_update_bundler_version
 
   def self.lockfile_version
-    return unless lockfile = lockfile_contents
-    lockfile, contents = lockfile
-    lockfile ||= "lockfile"
+    return unless contents = lockfile_contents
     regexp = /\n\nBUNDLED WITH\n\s{2,}(#{Gem::Version::VERSION_PATTERN})\n/
     return unless contents =~ regexp
-    [$1, lockfile]
+    $1
   end
   private_class_method :lockfile_version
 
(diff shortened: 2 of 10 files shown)


deivid-rodriguez:
I think this is a nice step to unblock a potential bundler 3 major release, and allows bundler 2.3 to properly handle lock files locked to bundler 1.17, for example, even if that version of bundler is not installed, thanks to our recent version switching feature.

So since I didn't get feedback I'm going to merge it.

simi (approved):

deivid-rodriguez:
Thank you @simi!!

deivid-rodriguez:
I'll try fix CI and then merge this.

PR #5182, opened 2021-12-20
Make `bundle update --bundler` actually lock to the latest bundler version (even if not yet installed)

### LINKED ISSUE #3343: Cannot update bundler version to be older using `bundler update --bundler
If you are trying to use a Gemfile.lock with a `BUNDLED WITH` that indicates bundler 2.x and try to use it on a system that does not have bundler 2.x then you'll get this error:

```
$ bundle
	2: from /Users/rschneeman/.gem/ruby/2.6.0/bin/bundle:23:in `<main>'
	1: from /Users/rschneeman/.rubies/ruby-2.6.0/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/Users/rschneeman/.rubies/ruby-2.6.0/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.0) required by your /private/tmp/default_ruby/Gemfile.lock. (Gem::GemNotFoundException)
```

If you try to follow the first instruction to update the Gemfile.lock you get this error:

```
$ bundle update --bundler
You must use Bundler 2 or greater with this lockfile.
```

## Repro

```
docker run -it --rm ruby:2.6 bash
```

```
gem install bundler -v 2.0.1
touch Gemfile
bundle install
gem uninstall bundler -v 2.0.1 -x
bundle install
bundle update --bundler
```

output:

```
root@f1fbb993272b:/# bundle install
  2: from /usr/local/bundle/bin/bundle:23:in `<main>'
  1: from /usr/local/lib/ruby/2.6.0/rubygems.rb:302:in `activate_bin_path'
/usr/local/lib/ruby/2.6.0/rubygems.rb:283:in `find_spec_for_exe': Could not find 'bundler' (2.0.1) required by your /Gemfile.lock. (Gem::GemNotFoundException)
To install the missing version, run `gem install bundler:2.0.1`
root@f1fbb993272b:/# bundle update --bundler
You must use Bundler 2 or greater with this lockfile.
```

### LINKED ISSUE #5172: `bundle update --bundler` should update bundler to the latest available version
Currently it only updates the lockfile to the latest installed version.

deivid-rodriguez:

The problem is that upgrading your application to use the latest Bundler is not easy. You have to first run `gem install bundler`, and then run `bundle update --bundler`. Running the latter should upgrade Bundler without having to run the former.
 
Fix is to implement this feature but checking available versions remotely and installing the latest available. That unless a specific version is specified as a target (`bundle update --bundler <version>`), in which case, we first check whether that version is already installed locally and switch to it in that case, or directly install it and switch otherwise.

Fixes #3343.
Fixes #5172.


deivid-rodriguez:
This is green now! I'll add some specs tomorrow.

deivid-rodriguez:
Alright, I fixed a bunch of bugs and added specs. If CI agrees, this one should be ready!

bundler/lib/bundler/self_manager.rb
```diff
@@ -9,46 +9,48 @@ class SelfManager
...
-        "Installing Bundler #{lockfile_version} and restarting using that version."
+      if spec.nil?
+        Bundler.ui.warn "Your lockfile is locked to a version of bundler (#{lockfile_version}) that doesn't exist at https://rubygems.org/. Going on using #{current_version}"
+        return
+      end
 
       spec.source.install(spec)
     rescue StandardError => e
       Bundler.ui.trace e
       Bundler.ui.warn "There was an error installing the locked bundler version (#{lockfile_version}), rerun with the `--verbose` flag for more details. Going on using bundler #{current_version}."
```
deivid-rodriguez, line 48: I wonder if we should now remove this rescue clause, given that the two errors we imagined could happen (invalid version in lockfile, and `https://rubygems.org` not in `Gem.sources`) are now either handled gracefully or no longer an issue.

I guess I'll leave this there for a few more versions and if nobody reports any issues remove it.

deivid-rodriguez:
It was more painful than I expected, but it's all green now!

PR #5184, opened 2021-12-20
Improve release scripts to support releasing minors & majors

deivid-rodriguez:

Since we manage changelings automatically, we have never released a minor.

Update the scripts to also handle minor and major versions.

Minor and major versions are handled similarly to patch version in regards to the changelog, but are cut directly from the main branch and don't need any cherry-picking.

Since we're actually shipping some changes that could be considered mildly incompatible, I'm not making any distinction semantically between minors and majors. The only different is the subjective assessments of how breaking and important they are.

### DIFF
--- a/util/changelog.rb
+++ b/util/changelog.rb
@@ -56,13 +56,7 @@ class Changelog
     @version = Gem::Version.new(version)
     @file = File.expand_path(file)
     @config = YAML.load_file("#{File.dirname(file)}/.changelog.yml")
-    @level = if @version.segments[1..2] == [0, 0]
-               :major
-             elsif @version.segments[2] == 0
-               :minor
-             else
-               :patch
-             end
+    @level = @version.segments[2] != 0 ? :patch : :minor
   end
 
   def release_notes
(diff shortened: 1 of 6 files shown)


util/changelog.rb
```diff
@@ -56,13 +56,7 @@ def initialize(file, version)
...
     @file = File.expand_path(file)
     @config = YAML.load_file("#{File.dirname(file)}/.changelog.yml")
-    @level = if @version.segments[1..2] == [0, 0]
-               :major
-             elsif @version.segments[2] == 0
-               :minor
-             else
-               :patch
-             end
+    @level = @version.segments[2] != 0 ? :patch : :minor
```
indirect, line 59: Does this actually support majors, like the PR title says? (It's fine if not, just wanted to clarify for myself so I would know)

deivid-rodriguez: Basically, as this is coded, there's no difference between major and minor, I should've named the level `:minor_or_major` to make this clear. We allow any types of changes in majors and minors, the only difference is subjective, major should include major breaking changes, but minors are allowed to ship small breaking changes.

I did this because we have a few changes in rubygems that can be considered breaking, but they affect small unimportant cases in my opinion. Check the changelog, if we don't agree to ship them now, I think the easiest would be to revert them from the main branch.

indirect: Works for me, thanks for explaining! 👍🏻

deivid-rodriguez: Great, I'm going to clarify the naming in a separate PR.

PR #5190, opened 2021-12-21
Drop Ruby 2.3, 2.4, and 2.5 support from bundler

deivid-rodriguez:

We want to drop support for old rubies. Main advantages being a smaller CI matrix, and thus easier to get green. For now I proposed to do this in Bundler only, since I want to figure out a nice upgrade path for RubyGems.

This SHOULD NOT be backported to any stable branches, which in particular, should keep testing the full matrix of rubies.

Drop support.


deivid-rodriguez:
Updating only bundler does not seem viable, since we want to keep releases in sync, and rubygems update also update bundler. Will figure out the best upgrade path for rubygems and then drop support everywhere.

PR #5191, opened 2021-12-21
Fix `gem update --system` crashing when latest version not supported

deivid-rodriguez:

Currently, if we released a new rubygems version dropping support for old rubies, `gem update --system` would start crashing calling a `version` method on `nil`.

This first solution is not perfect (it could be improved by falling back to updating as far as possible), but at least avoids the crash, still exists with a 0 status, and still shows an error about the problem. It matches the current behavior of `gem update <gem>` when the latest version of `<gem>` has this issue.
 
### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -288,7 +288,9 @@ command to remove old versions.
 
     installed_gems = Gem::Specification.find_all_by_name 'rubygems-update', requirement
     installed_gems = update_gem('rubygems-update', version) if installed_gems.empty?
-    version        = installed_gems.first.version
+    return if installed_gems.empty?
+
+    version = installed_gems.first.version
 
     install_rubygems version
   end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
CI detected an issue in our version switcher :/. I'll have a look now and try to make a patch release if necessary.

deivid-rodriguez:
No idea how this PR would affect this, by the way, and if it doesn't, why CI didn't catch this before.

deivid-rodriguez:
Oh, I see, I get it now. This is a bug in our re'exec code in the super weird edge case where Bundler is not run directly but through Ruby. It didn't manifest before because "Bundler 3 specs" only run in the main branch, not in the stable branch, and they cause the trampoline feature to be triggered because our development lock files are locked to bundler 2.3.0, but the running bundler is 3.0.0 in this mode.

This is just some weirdness of our specs and could hardly considered a supported case, but we should definitely not crash like that, so I'll fix it in some way.

deivid-rodriguez:
The CI issue has been fixed separately and this is now green. But it includes one enhancement and one bug fix, so I'll split the enhancement out.

deivid-rodriguez:
The enhancement was split to #5230.

PR #5193, opened 2021-12-21
Re-exec prepending command with `Gem.ruby` if `$PROGRAM_NAME` is not executable

deivid-rodriguez:

We have some failures in bundler 3 mode.

The problem is caused by the re-execing feature being triggered in bundler 3 mode, because our development lock files are locked to bundler 2.

There's two issues:

* Running bundler as `ruby <stub_not_directly_executable_that_loads_bundler>` is not handled properly.
* When reexecing to `bundler` from within a `bundler` context, and passing some custom bundler variables like we are doing in our test helper when setting up development dependencies (for example, `BUNDLE_GEMFILE`), we now need to reset the backup (`BUNDLER_ORIG_BUNDLE_GEMFILE`). This is due to bundler re-execing using `Bundler.with_original_env`, which seems conceptually correct, but I'm afraid it could cause issues for end users doing similar things. 

This PR fixes the first issue by prepending `Gem.ruby` to the re-exec'ed command if `$PROGRAM_NAME` is not executable.

It also fixes the second issue to get CI green by properly resetting the previous environment variable backup when running bundler in a subprocess with modified environment variables. I'm worried about this second issue potentially biting end users, and I'm considering not using `Bundler.with_original_env` when re-execing bundler. But I'm not 100% sure so I won't be taking any further action there for now.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -26,7 +26,7 @@ jobs:
           - { name: "2.6", value: 2.6.9 }
           - { name: "2.7", value: 2.7.5 }
           - { name: "3.0", value: 3.0.3 }
-          - { name: jruby-9.3, value: jruby-9.3.1.0 }
+          - { name: jruby-9.3, value: jruby-9.3.2.0 }
           - { name: truffleruby-21, value: truffleruby-21.2.0 }
         openssl:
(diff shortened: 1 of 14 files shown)


deivid-rodriguez:
Alright, I think I found an easy fix for the bundler binstub thing. I'll push it and see if it works.

PR #5195, opened 2021-12-22
Avoid unnecessary `shellwords` require on newer rubygems

deivid-rodriguez:

Just some small spec issues.

Fix a typo and relax the condition to run some specs, so that they are checked again.


deivid-rodriguez:
This fixes a flaky test failure but changes code in `lib`, I believe it could technically also fix an end user issue, so I'll ship this as an small enhancement.

Issue #5192, opened 2021-12-21
Test failures with Ruby 3.1 due to OpenSSL changes


voxik:
Testing with `ruby 3.1.0dev (2021-12-21 master cdb7d699d0) [x86_64-linux]`, I observe the following test failures:

~~~
  1) Error:
TestGemCommandsCertCommand#test_execute_build_ec_key:
OpenSSL::PKey::PKeyError: pkeys are immutable on OpenSSL 3.0
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/security.rb:430:in `public_key='
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/security.rb:430:in `get_public_key'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/security.rb:405:in `create_cert'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/security.rb:460:in `create_cert_self_signed'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/security.rb:451:in `create_cert_email'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/commands/cert_command.rb:157:in `build_cert'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/commands/cert_command.rb:143:in `build'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/commands/cert_command.rb:123:in `block in execute'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/commands/cert_command.rb:122:in `each'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/commands/cert_command.rb:122:in `execute'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/test/rubygems/test_gem_commands_cert_command.rb:326:in `block in test_execute_build_ec_key'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/user_interaction.rb:46:in `use_ui'
    /builddir/build/BUILD/ruby-3.1.0-cdb7d699d0/lib/rubygems/user_interaction.rb:69:in `use_ui'
[...]

PR #5196, opened 2021-12-22
Fix compatibility with OpenSSL 3.0

rhenium:

Gem::Security uses `openssl`'s features that are incompatible with OpenSSL 3.0. As reported in #5192, it can be observed as a test failure.

This only affects Ruby 3.1 (or later) since stdlib `openssl` in previous versions doesn't support OpenSSL 3.0.

RubyGems uses the following two methods. They modify the receiver object which is not possible if `openssl` is built against OpenSSL 3.0.

 - `OpenSSL::PKey::EC#public_key=`
 - `OpenSSL::PKey::EC#generate_key`

Fixes https://github.com/rubygems/rubygems/issues/5192.

### DIFF
--- a/lib/rubygems/security.rb
+++ b/lib/rubygems/security.rb
@@ -424,6 +424,8 @@ module Gem::Security
   # Gets the right public key from a PKey instance
 
   def self.get_public_key(key)
+    # Ruby 3.0 (Ruby/OpenSSL 2.2) or later
+    return OpenSSL::PKey.read(key.public_to_der) if key.respond_to?(:public_to_der)
     return key.public_key unless key.is_a?(OpenSSL::PKey::EC)
 
     ec_key = OpenSSL::PKey::EC.new(key.group.curve_name)
--- a/lib/rubygems/security.rb
+++ b/lib/rubygems/security.rb
@@ -531,7 +531,7 @@ module Gem::Security
     raise Gem::Security::Exception,
           "incorrect signing key for re-signing " +
           "#{expired_certificate.subject}" unless
-      expired_certificate.public_key.to_pem == get_public_key(private_key).to_pem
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
This is great @rhenium, thanks so much, and also for all the work on getting the `openssl` gem compatible with OpenSSL 3.0!

voxik:
@hsbt can this be pulled into Ruby 3.1 please?

hsbt:
@voxik After releasing RubyGems 3.3.1 from https://github.com/rubygems/rubygems/pull/5200, I will do that.

lib/rubygems/security.rb
```diff
@@ -490,9 +492,13 @@ def self.create_key(algorithm)
       when 'rsa'
         OpenSSL::PKey::RSA.new(RSA_DSA_KEY_LENGTH)
       when 'ec'
-        domain_key = OpenSSL::PKey::EC.new(EC_NAME)
-        domain_key.generate_key
-        domain_key
+        if RUBY_VERSION >= "2.4.0"
```
voxik, line 495: Is this version check necessary these days? Wouldn't this branch be enough, because Ruby 2.4 is not supported anyway.

voxik:
> @voxik After releasing RubyGems 3.3.1 from #5200, I will do that.

I have not seen the PR yet :+1:  I just saw some reverts in Ruby repository yesterday, so just wondered. Thx a lot.

lib/rubygems/security.rb
deivid-rodriguez: It is still supported. We can drop it once we drop ruby 2.3 support, so the explicit check for the version is useful so that we don't forget to cleanup.

PR #5205, opened 2021-12-22
Improve error message when `BUNDLED WITH` version does not exist

deivid-rodriguez:

We're installing gems using rubygems installer, which makes installation dependent on what's on `Gem.sources` and does not easily allow us to show a nice error message when the `BUNDLED WITH` version does not exist.

Install the bundler gem in the same way other gems are installed from bundler. This allows to easily fix the above issues, but also to use bundler custom gem installer (inherits from `Gem::Installer`) which for example, does not call `Gem::Specification.reset` after installing gems or does checksum verification. In general, this way feels better.


bundler/lib/bundler/self_manager.rb
```diff
@@ -26,15 +26,27 @@ def install_locked_bundler_and_restart_with_it_if_needed
...
     rescue StandardError => e
       Bundler.ui.trace e
       Bundler.ui.warn "There was an error installing the locked bundler version (#{lockfile_version}), rerun with the `--verbose` flag for more details. Going on using bundler #{current_version}."
     else
       restart_with_locked_bundler
     end
 
+    def fetch_spec_for(bundler_dep)
+      source = Bundler::Source::Rubygems.new("remotes" => "https://rubygems.org")
+      source.remote!
```
deivid-rodriguez, line 45: This needs to be done unless the `--local` flag was given to `bundle install`. I'll improve this tomorrow.

deivid-rodriguez: This issue is already present and there are more pressing things, so I will reticket it.

deivid-rodriguez:
This improves the error message when locked version does not exist, but also makes the bundler gem installation consistent with how all other gems are installed by `bundler`, and independent from how RubyGems `Gem.sources` are configured. So I'll merge it.

simi (approved):

PR #5212, opened 2021-12-23
Try fix file locking issues on Windows

deivid-rodriguez:

We're getting a quite frequent flaky test failure on Windows.

I think we should try to acquire the file lock before trying to read stub specifications, to avoid accessing a file that's already opened. Let's see if CI agrees.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -163,16 +163,6 @@ module Gem
     specifications/default
   ].freeze
 
-  ##
-  # Exception classes used in a Gem.read_binary +rescue+ statement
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Alright, CI passed twice here so I'll cross fingers and assume this fixes the Windows issues our CI is showing. Since this issue doesn't exist in a released version, I will not advertise this in the changelog, although I'm not sure how JRuby + Windows behaves, since we don't run the full test suite there and I had to skip the locking there because of a JRuby issue (I will report it latter).

PR #5226, opened 2021-12-26
Update some specs to pass ruby-head CI

deivid-rodriguez:

These specs were monkeypatching `RUBY_VERSION`, but that obviously doesn't change the running ruby to behave any different.

The removal of some features, in particular, `String#untaint`, made these specs fail, because untaint is no longer available under ruby-core and Bundler calls `untaint` when `RUBY_VERSION` is less than "2.7", which these specs were overwriting it to be.

Rewrite these specs to not overwrite `RUBY_VERSION` but still test the same things.

NOTE: I also included in this PR an optimization to repository checkout in the workflow that tests Bundler against different versions of RubyGems.

### DIFF
--- a/.github/workflows/older-rubygems-bundler.yml
+++ b/.github/workflows/older-rubygems-bundler.yml
@@ -43,8 +43,10 @@ jobs:
       RUBYOPT: --disable-gems
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
GitHub is having network connectivity issues: https://github.com/actions/virtual-environments/issues/4807.

But this seems fine, so I'm merging it.

PR #5228, opened 2021-12-26
Fix longstanding `gem install` failure on JRuby

deivid-rodriguez:

On JRuby, sometimes we get the following error in CI when running a realworld test that checks that `gem install rails` succeeds:

```
    undefined method `ignored=' for nil:NilClass
	/home/runner/.rubies/jruby-9.3.2.0/lib/ruby/stdlib/rubygems/stub_specification.rb:193:in `to_spec'
	org/jruby/RubyArray.java:2642:in `map'
	/home/runner/.rubies/jruby-9.3.2.0/lib/ruby/stdlib/rubygems/specification.rb:758:in `_all'
	/home/runner/.rubies/jruby-9.3.2.0/lib/ruby/stdlib/rubygems/specification.rb:956:in `each'
	org/jruby/RubyEnumerable.java:1710:in `any?'
	/home/runner/.rubies/jruby-9.3.2.0/lib/ruby/stdlib/rubygems/resolver/activation_request.rb:111:in `installed?'
	/home/runner/.rubies/jruby-9.3.2.0/lib/ruby/stdlib/rubygems/request_set.rb:173:in `block in install'
```

I'm not sure how this error is happening, but I think there's no need to copy the `@ignored` instance variable when materializing stub specifications. This instance variable is used to not print a warning about missing extensions more than once for each gem upon gem activation, but as far as I can see, it's only used by methods that work on specification stubs. Once specifications are materialized, I think it can be safely ignored.

### DIFF
--- a/lib/rubygems/stub_specification.rb
+++ b/lib/rubygems/stub_specification.rb
@@ -190,9 +190,6 @@ class Gem::StubSpecification < Gem::BasicSpecification
               end
 
     @spec ||= Gem::Specification.load(loaded_from)
-    @spec.ignored = @ignored if @spec
-
-    @spec
   end
 
   ##
--- a/test/rubygems/test_gem_stub_specification.rb
+++ b/test/rubygems/test_gem_stub_specification.rb
@@ -180,22 +180,6 @@ class TestStubSpecification < Gem::TestCase
     assert bar.to_spec
   end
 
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Some JRuby users are also running into this, so I'm going to tentatively merge this.

Issue #5221, opened 2021-12-26
`bundle update --bundler` does nothing


mattbrictson:
### Describe the problem as clearly as you can

`bundle update --bundler` does not change the `BUNDLED WITH` version in the lock file. The lock file remains unchanged.

### Did you try upgrading rubygems & bundler?

Yes.

```
$ gem install bundler
Successfully installed bundler-2.3.3
1 gem installed
$ gem update --system
Latest version already installed. Done.
```

### Post steps to reproduce the problem

1. Start with a lock file that has `BUNDLED WITH` of an older version (e.g. 2.2.3).
2. Install the latest version of bundler with `gem install bundler` (as of this writing, 2.3.3).
3. Run `bundle update --bundler`.

```
$ tail -n2 Gemfile.lock 
BUNDLED WITH
   2.2.3
$ gem install bundler
Successfully installed bundler-2.3.3
1 gem installed
$ bundle update --bundler
...
Bundle updated!
```

### Which command did you run?

`bundle update --bundler`

### What were you expecting to happen?

I am expecting the `Gemfile.lock` to be updated with a new `BUNDLED WITH` version of 2.3.3, since that is the latest version I have installed.

```
$ gem list -e bundler

*** LOCAL GEMS ***

bundler (2.3.3, 2.2.32, 2.2.3)
```

### What actually happened?

The `Gemfile.lock` is not changed. It remains with a `BUNDLED WITH` version of 2.2.3.

```
$ tail -n2 Gemfile.lock 
BUNDLED WITH
   2.2.3
```

## Environment

```
Bundler       2.2.3
  Platforms   ruby, x86_64-darwin-20
Ruby          3.0.3p157 (2021-11-24 revision 3fb7d2cadc18472ec107b14234933b017a33c14d) [x86_64-darwin20]
  Full Path   /Users/mbrictson/.rbenv/versions/3.0.3/bin/ruby
  Config Dir  /Users/mbrictson/.rbenv/versions/3.0.3/etc
RubyGems      3.3.3
  Gem Home    /Users/mbrictson/.rbenv/versions/3.0.3/lib/ruby/gems/3.0.0
  Gem Path    /Users/mbrictson/.gem/ruby/3.0.0:/Users/mbrictson/.rbenv/versions/3.0.3/lib/ruby/gems/3.0.0
  User Home   /Users/mbrictson
  User Path   /Users/mbrictson/.gem/ruby/3.0.0
[...]

PR #5238, opened 2021-12-29
Prepare rubygems 3.3.4 and bundler 2.3.4

deivid-rodriguez:
I'm still working on some more `bundle update --bundler` improvements and fleshing out some regressions, but I don't want to delay the fix for #5221 any longer.


MarcusSaviour (commented):
Good

Issue #5166, opened 2021-12-14
Let `gem fetch` to understand "foo:1.0.0" version syntax


voxik:
This behavior is inconsistent:

~~~
$ gem install mini_mime:1.0.0
Fetching mini_mime-1.0.0.gem
Successfully installed mini_mime-1.0.0
Parsing documentation for mini_mime-1.0.0
Installing ri documentation for mini_mime-1.0.0
Done installing documentation for mini_mime after 0 seconds
1 gem installed

$ gem fetch mini_mime:1.0.0
ERROR:  Could not find a valid gem 'mini_mime:1.0.0' (>= 0) in any repository
ERROR:  Possible alternatives: mini_mime

$ gem -v
3.2.22
~~~

ximenasandoval:
Hi! I'd like to work on this issue

deivid-rodriguez:
Hi @ximenasandoval, that's great! It's much appreciated :) If you get stuck let me know.

PR #5242, opened 2021-12-29
Let `gem fetch` understand `<gem>:<version>` syntax and `--[no-]suggestions` flag

ximenasandoval:

When running `gem install <gem>:<version>` we can install the given gem at the given version, but running `gem fetch <gem>:<version>` is not supported and makes this an inconsistent behavior, as described in #5166.

The problem was in the `get_all_gem_names` function, which was taking the string `<gem>:<version>` as the entire name of the gem.
To fix it, now we get the names of the gems alongside the specified versions with `get_all_gem_names_and_versions` (similar to the `install` command) and check if there are multiple gems to be fetched and the `--version` flag to output an error message. 

### DIFF
--- a/lib/rubygems/commands/fetch_command.rb
+++ b/lib/rubygems/commands/fetch_command.rb
@@ -18,6 +18,10 @@ class Gem::Commands::FetchCommand < Gem::Command
     add_version_option
     add_platform_option
     add_prerelease_option
+
+    add_option '--[no-]suggestions', 'Suggest alternates when gems are not found' do |value, options|
+      options[:suggest_alternate] = true
+    end
   end
 
   def arguments # :nodoc:
@@ -42,15 +46,27 @@ then repackaging it.
     "#{program_name} GEMNAME [GEMNAME ...]"
   end
 
+  def check_version # :nodoc:
+    if options[:version] and options[:version] != Gem::Requirement.default and
+         get_all_gem_names.size > 1
+      alert_error "Can't use --version with multiple gems. You can specify multiple gems with" \
+                  " version requirements using `gem fetch 'my_gem:1.0.0' 'my_other_gem:~>2.0.0'`"
+      terminate_interaction 1
+    end
+  end
+
   def execute
-    version = options[:version] || Gem::Requirement.default
+    check_version
+    version = options[:version]
 
     platform  = Gem.platforms.last
-    gem_names = get_all_gem_names
+    gem_names = get_all_gem_names_and_versions
 
-    gem_names.each do |gem_name|
-      dep = Gem::Dependency.new gem_name, version
+    gem_names.each do |gem_name, gem_version|
+      gem_version ||= version
+      dep = Gem::Dependency.new gem_name, gem_version
       dep.prerelease = options[:prerelease]
+      suppress_suggestions = options[:suggest_alternate]
 
       specs_and_sources, errors =
         Gem::SpecFetcher.fetcher.spec_for_dependency dep
@@ -63,12 +79,10 @@ then repackaging it.
       spec, source = specs_and_sources.max_by {|s,| s }
 
       if spec.nil?
-        show_lookup_failure gem_name, version, errors, options[:domain]
+        show_lookup_failure gem_name, gem_version, errors, suppress_suggestions, options[:domain]
         next
       end
-
       source.download spec
-
       say "Downloaded #{spec.full_name}"
     end
   end
--- a/lib/rubygems/commands/fetch_command.rb
+++ b/lib/rubygems/commands/fetch_command.rb
@@ -8,7 +8,12 @@ class Gem::Commands::FetchCommand < Gem::Command
   include Gem::VersionOption
 
   def initialize
-    super 'fetch', 'Download a gem and place it in the current directory'
+    defaults = {
+      :suggest_alternate => true,
+      :version           => Gem::Requirement.default,
+    }
+
+    super 'fetch', 'Download a gem and place it in the current directory', defaults
 
     add_bulk_threshold_option
     add_proxy_option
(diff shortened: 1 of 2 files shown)


simi:
Nice work! I was testing this and found one edge-case reporting wrong error message when gem is yanked or version doesn't exist.

```
[retro@retro  rubygems (fetch-with-colon-version %)]❤ gem fetch ABO:0.0.1
ERROR:  Could not find a valid gem 'ABO' (>= 0) in any repository
ERROR:  Possible alternatives: ABO, AllAboutFiles, aboardly, aboie, aboisvert_aws, abokifx_indexer, abongo, abort_if, abot-info, abot-whitelist
```

`gem install` reports valid version in error message

```
[retro@retro  rubygems (fetch-with-colon-version %)]❤ gem install ABO:0.0.1
ERROR:  Could not find a valid gem 'ABO' (= 0.0.1) (required by 'ABO' (0.0.1)) in any repository
ERROR:  Possible alternatives: ABO, AllAboutFiles, aboardly, aboie, aboisvert_aws, abokifx_indexer, abongo, abort_if, abot-info, abot-whitelist
```

ximenasandoval:
Thanks! I'll look into it and fix it 🙌🏽

simi:
It can be fixed by using `gem_version` at `show_lookup_failure gem_name, gem_version, errors, options[:domain]`.

ximenasandoval:
Thank you! I'll add it to the pull request 😄

simi:
@ximenasandoval can you please add test also for this case to ensure error message reports proper version?

ximenasandoval:
Sure! Should a flag for `--[no-]suggestions` be added? I noticed it is in the `install` command but not in the `fetch` one

simi:
@ximenasandoval feel free to add that one as well, but not needed to get this merged IMHO.

simi (approved):

deivid-rodriguez:
I'm in the middle of moving home, so I won't be able to have a look at this in a few days. But if this looks good to you @simi please go ahead and merge!

lib/rubygems/commands/fetch_command.rb
```diff
@@ -42,15 +46,27 @@ def usage # :nodoc:
     "#{program_name} GEMNAME [GEMNAME ...]"
   end
 
+  def check_version # :nodoc:
+    if options[:version] and options[:version] != Gem::Requirement.default and
```
deivid-rodriguez, line 50: I noticed that `check_version` in the install command does not check whether `options[:version]` was passed at all, and instead set `Gem::Requirement.default` as its the default value in `initialize`. I guess we can do the same for consistency.

test/rubygems/test_gem_commands_fetch_command.rb
```diff
@@ -157,4 +157,101 @@ def test_execute_version
...
+    assert_equal expected, @ui.error
+  end
+
+  def test_execute_nonexistent_hint_disabled
+    spec_fetcher do |fetcher|
+      fetcher.spec 'foo', 1
+    end
+
+    @cmd.options[:args] = %w[foo:2]
+    @cmd.options[:suggest_alternate] = true
```
deivid-rodriguez, line 245: Isn't the `suggest_alternate` logic inverted? I would expect suggestions to be enabled if `:suggest_alternate` is set.

lib/rubygems/commands/fetch_command.rb
```diff
@@ -18,6 +18,10 @@ def initialize
     add_version_option
     add_platform_option
     add_prerelease_option
+
+    add_option '--[no-]suggestions', 'Suggest alternates when gems are not found' do |value, options|
+      options[:suggest_alternate] = true
```
deivid-rodriguez, line 23: I think this should be assigned to `value`, and pass a default `true` value when calling `super` above.

deivid-rodriguez (changes requested):
I left one comment to make things more consistent with the install command and some others because I think the `--no-suggestions` logic is inverted.

test/rubygems/test_gem_commands_fetch_command.rb
simi: I was confused as well, but I did some testing and it works as inteded (and the logic is similar to one in install). :confused:

deivid-rodriguez: Actually it works _almost_ as intended. While `gem install <non-existent-gem> --suggestions` gives suggestions, `gem fetch <non-existent-gem> --suggestions` does not.

ximenasandoval: You're right! I was confused at first but now it's clear, I will fix it!

lib/rubygems/commands/fetch_command.rb
```diff
@@ -42,15 +51,27 @@ def usage # :nodoc:
...
+         get_all_gem_names.size > 1
+      alert_error "Can't use --version with multiple gems. You can specify multiple gems with" \
+                  " version requirements using `gem fetch 'my_gem:1.0.0' 'my_other_gem:~>2.0.0'`"
+      terminate_interaction 1
+    end
+  end
+
   def execute
+    check_version
     version = options[:version] || Gem::Requirement.default
```
deivid-rodriguez, line 65: I think the right hand side of the `||` is no longer necessary now.

deivid-rodriguez (commented):
@ximenasandoval I left one little final comment but other than that this is ready from my side, thanks so much for your work on this ❤️.

If you can rebase and squash this PR into two commits (one adding support for the `name:version` syntax and tests, and another one adding support for the `--[no-]suggestiosn` flag and tests), that's awesome. But it's not strictly necessary, I have already nitpicked enough for a first contribution 😅.

ximenasandoval:
No problem! I'll squash and rebase the branch 😁  and also fix the `version = options[:version]` line

ximenasandoval:
Sorry for the double push, I forgot to fix the `version = options[:version]` line but now is fixed 😅

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks so much for your work @ximenasandoval ❤️.

voxik:
Nice :clap: Looking forward to try this. Thx a lot

PR #4156, opened 2020-12-15
Add `rake build:checksum` task to build cheksums for a gem package

LuiGiovanni:
## What was the end-user or developer problem that led to this PR?
This is resolving issue number #3386 made by @svoop.

## What is your fix for the problem, implemented in this PR?
I implemented checksum for all gems in the bundler/pkg directory, this implementation may be found inside the bundler namespace starting at line 405.

Closes #3386.

## Make sure the following tasks are checked

- [x] Describe the problem/feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #5250, opened 2022-01-03
Fix `rake build:checksum` writing checksum of package path, not package contents

nobu:

`rake build:checksum` calculates the checksum of the full pathname of the built gem file, not its content.
This means the file has a wrong value and checks by downloaded users will always fail.
Fix up #4156.

Calculate the checksum of the content.


deivid-rodriguez (approved):
Oooops nice catch!

PR #5258, opened 2022-01-04
Support for pre flag in `bundle update`/`bundle lock`

### LINKED ISSUE #5207: Bundle update a prerelease version without specifying a prerelease version in the Gemfile

Hello all and happy holidays!

I've been digging a bit into the fact that `bundle update` doesn't support a `--pre` flag.

I've found multiple issues that are vaguely related, though they generally fall into two categories that aren't quite what I'm looking for:
* ones related to the `--pre` flag to the `gem` command: https://github.com/rubygems/bundler/issues/1198, https://github.com/rubygems/rubygems/issues/129, https://github.com/rubygems/rubygems/issues/3474
* ones related to specifying version in the gemfile: https://github.com/rubygems/bundler/issues/2938 & https://guides.rubygems.org/patterns/#prerelease-gems

I see the general sentiment is that if you pass `--pre` to `gem install`, you are taking matters into your own hands a bit. There doesn't seem to be a way to do so for a bundled gemfile though without specifying a version.

I'm most curious if there was a discussion/issue/rejected PR that I missed or if this is something y'all would be open to. If it is a feature that sounds appealing, I'd be happy to take a stab at it, but didn't want to dig in too far if the idea has already been discussed.

Yes

Reproduction of `bundle update --pre gem_name` is easy since the flag isn't supported and thus errors, though I'm struggling to find a public gem that whose latest version is a pre-release gem that has non-pre-release versions as well. Then, I'd expect to be able to reproduce by running
```sh
$ bundle add gem_name
$ bundle update --pre gem_name
```
. If this more complete reproduction step is desired, I'm happy to dig in more next week!

`$ bundle update --pre gem_name`

I expected bundler to update the version to the latest version, either only looking at pre-release versions or ignoring the release status of a version. I'm okay with either outcome, but let me know what y'all think it should behave like. The first would make it act more like the SEMVER flags, but the second is a bit more flexible and maybe more intuitive?

### What actually happened?

```sh
$ bundle update --pre gem_name
Unknown switches "--pre"
```

<details><summary>Probably not super relevant, but details under the fold</summary>

## Environment

```
Bundler       2.3.1
Ruby          2.6.3p62 (2019-04-16 revision 67580) [x86_64-linux]
  Full Path   /home/siggy/.rbenv/versions/2.6.3/bin/ruby
  Config Dir  /home/siggy/.rbenv/versions/2.6.3/etc
RubyGems      3.3.1
  Gem Home    /home/siggy/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0
  Gem Path    /home/siggy/.gem/ruby/2.6.0:/home/siggy/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0
  User Home   /home/siggy
  User Path   /home/siggy/.gem/ruby/2.6.0
  Bin Dir     /home/siggy/.rbenv/versions/2.6.3/bin
Tools         
  Git         2.34.1
  RVM         not installed
  rbenv       rbenv 1.1.1
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2021-12-22
Git SHA           7b38ca2f84
Released Version  true
```

## Bundler settings

```
github.com
  Set for your local app (/home/siggy/.bundle/config): "siegfault:[REDACTED]"
  Set for the current user (/home/siggy/.bundle/config): "siegfault:[REDACTED]"
```
</details>

siegfault:

More thoroughly described in https://github.com/rubygems/rubygems/issues/5207, but essentially allowing bumps to prerelease versions of gems without specifying a version in the lockfile.

Passing this flag allows bumping to the current version, even if that
version is prerelease. This is a superset of the major flag since it
behaves the same but also considers prerelease versions.

The main alternative I see would be having pre work in concert with the currently existing flags. This would mean passing `--pre --major` would work as in this PR, but `--pre --minor` would prefer not to bump major versions.

Assuming we want the current behavior, are there any other commands I should update, or any other places I should add specs?

Closes #5207.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -293,7 +293,7 @@ module Bundler
     method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
     method_option "pre", :type => :boolean, :banner =>
-      "If updating, allow consideration of prerelease gems"
+      "Always choose the highest allowed version when updating gems, regardless of prerelease status"
     method_option "strict", :type => :boolean, :banner =>
       "Do not allow any gem to be updated past latest --patch | --minor | --major"
     method_option "conservative", :type => :boolean, :banner =>
@@ -671,7 +671,7 @@ module Bundler
     method_option "major", :type => :boolean, :banner =>
       "If updating, prefer updating to next major version (default)"
     method_option "pre", :type => :boolean, :banner =>
-      "If updating, allow consideration of prerelease gems"
+      "If updating, always choose the highest allowed version, regardless of prerelease status"
     method_option "strict", :type => :boolean, :banner =>
       "If updating, do not allow any gem to be updated past latest --patch | --minor | --major"
     method_option "conservative", :type => :boolean, :banner =>
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -292,6 +292,8 @@ module Bundler
       "Prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
+    method_option "pre", :type => :boolean, :banner =>
+      "If updating, allow consideration of prerelease gems"
     method_option "strict", :type => :boolean, :banner =>
       "Do not allow any gem to be updated past latest --patch | --minor | --major"
     method_option "conservative", :type => :boolean, :banner =>
@@ -668,6 +670,8 @@ module Bundler
       "If updating, prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "If updating, prefer updating to next major version (default)"
+    method_option "pre", :type => :boolean, :banner =>
+      "If updating, allow consideration of prerelease gems"
     method_option "strict", :type => :boolean, :banner =>
       "If updating, do not allow any gem to be updated past latest --patch | --minor | --major"
     method_option "conservative", :type => :boolean, :banner =>
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -100,11 +100,11 @@ module Bundler
 
         if major?
           a <=> b
-        elsif either_version_older_than_locked(a, b, locked_version)
+        elsif either_version_older_than_locked?(a, b, locked_version)
           a <=> b
-        elsif segments_do_not_match(a, b, :major)
+        elsif segments_do_not_match?(a, b, :major)
           b <=> a
-        elsif !minor? && segments_do_not_match(a, b, :minor)
+        elsif !minor? && segments_do_not_match?(a, b, :minor)
           b <=> a
         else
           a <=> b
@@ -113,11 +113,11 @@ module Bundler
       post_sort(result, package.unlock?, locked_version)
     end
 
-    def either_version_older_than_locked(a, b, locked_version)
+    def either_version_older_than_locked?(a, b, locked_version)
       locked_version && (a.version < locked_version || b.version < locked_version)
     end
 
-    def segments_do_not_match(a, b, level)
+    def segments_do_not_match?(a, b, level)
       index = [:major, :minor].index(level)
       a.segments[index] != b.segments[index]
     end
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -74,12 +74,11 @@ module Bundler
 
       specs.select do |spec|
         gsv = spec.version
-        lsv = locked_version
 
         must_match = minor? ? [0] : [0, 1]
 
-        matches = must_match.map {|idx| gsv.segments[idx] == lsv.segments[idx] }
-        matches.uniq == [true] ? (gsv >= lsv) : false
+        all_match = must_match.all? {|idx| gsv.segments[idx] == locked_version.segments[idx] }
+        all_match && gsv >= locked_version
       end
     end
 
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -70,19 +70,16 @@ module Bundler
 
     def filter_dep_specs(specs, package)
       locked_version = package.locked_version
+      return specs if locked_version.nil? || major?
 
       specs.select do |spec|
-        if locked_version && !major?
-          gsv = spec.version
-          lsv = locked_version
+        gsv = spec.version
+        lsv = locked_version
 
-          must_match = minor? ? [0] : [0, 1]
+        must_match = minor? ? [0] : [0, 1]
 
-          matches = must_match.map {|idx| gsv.segments[idx] == lsv.segments[idx] }
-          matches.uniq == [true] ? (gsv >= lsv) : false
-        else
-          true
-        end
+        matches = must_match.map {|idx| gsv.segments[idx] == lsv.segments[idx] }
+        matches.uniq == [true] ? (gsv >= lsv) : false
       end
     end
 
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -8,6 +8,7 @@ module Bundler
   # to the resolution engine to select the best version.
   class GemVersionPromoter
     attr_reader :level
+    attr_accessor :pre
 
     # By default, strict is false, meaning every available version of a gem
     # is returned from sort_versions. The order gives preference to the
@@ -28,6 +29,7 @@ module Bundler
     def initialize
       @level = :major
       @strict = false
+      @pre = false
     end
 
     # @param value [Symbol] One of three Symbols: :major, :minor or :patch.
@@ -66,6 +68,11 @@ module Bundler
       level == :minor
     end
 
+    # @return [bool] Convenience method for testing value of pre variable.
+    def pre?
+      pre == true
+    end
+
     private
 
     def filter_dep_specs(specs, package)
@@ -86,7 +93,7 @@ module Bundler
       locked_version = package.locked_version
 
       result = specs.sort do |a, b|
-        unless locked_version && package.prerelease_specified?
+        unless locked_version && (package.prerelease_specified? || pre?)
           a_pre = a.prerelease?
           b_pre = b.prerelease?
 
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -13,7 +13,7 @@ module Bundler
     # By default, strict is false, meaning every available version of a gem
     # is returned from sort_versions. The order gives preference to the
     # requested level (:patch, :minor, :major) but in complicated requirement
-    # cases some gems will by necessity by promoted past the requested level,
+    # cases some gems will by necessity be promoted past the requested level,
     # or even reverted to older versions.
     #
     # If strict is set to true, the results from sort_versions will be
--- a/bundler/spec/commands/update_spec.rb
+++ b/bundler/spec/commands/update_spec.rb
@@ -1471,7 +1471,10 @@ RSpec.describe "bundle update conservative" do
         build_gem "foo", %w[1.5.1] do |s|
           s.add_dependency "bar", "~> 3.0"
         end
-        build_gem "bar", %w[2.0.3 2.0.4 2.0.5 2.1.0 2.1.1 3.0.0]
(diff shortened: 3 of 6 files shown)


deivid-rodriguez:
Thanks so much for working on this @siegfault!

> The main alternative I see would be having pre work in concert with the currently existing flags. This would mean passing --pre --major would work as in this PR, but --pre --minor would prefer not to bump major versions.

I like this idea, also allowing `--pre` without any existing flags which would work like already proposed by this PR.

siegfault:
Definitely, thanks for being on top of this! I'm switching computers and wanted to push my progress, but will hopefully have something more ready to view next week

bundler/lib/bundler/cli.rb
```diff
@@ -290,6 +290,8 @@ def install
       "Prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
+    method_option "pre", :type => :boolean, :banner =>
```
bf4, line 293: definitely nice to have parity with `gem list "\A$GEM_NAME\z" --all --prerelease --remote --source "$GEM_SOURCE"` (which is what I was looking for when I found this issue/pr)

siegfault:
As an update since this has been sitting awhile without movement (:sweat_smile:):

I'm pushing up a change that feels fairly close to what we want so I wanted to open up some questions.

## Questions

1. Maybe it's just me, but the effect of the `sort` within `GemVersionPromoter#sort_dep_specs` was really hard for me to wrap my head around. I ended up not needing to modify it and there is some documentation for it within the class (maybe elsewhere too?), but would you like me to (a) expand the docs or (b) make the code more obvious, perhaps by using `sort_by` as in https://github.com/rubygems/rubygems/blob/23c3622eb4cbe1b716255fec6ef5a022d97bab77/lib/rubygems/resolver.rb#L281-L292?
2. I've added some tests in `bundler/spec/commands/update_spec.rb` and there should probably be additions in `bundler/spec/bundler/gem_version_promoter_spec.rb`, but are there any other places that come to mind that I should add tests?
3. Would you rather I revert and/or break out any smaller changes into their own PRs (`map {}.uniq == [true]` -> `all? {}`, adding the `?` to interogative methods, etc.)?

## GemVersionPromoter Sort

The sort in `GemVersionPromoter` was difficult for me to tease out, but I believe it's essentially:
1. Prerelease versions (unless the current version is already a prerelease)
2. Downgrades
3. Upgrades, organized by number of major(/minor?) upgrades descending and version ascending
4. Current version (if the current version is locked)

To clarify (3), given:
* `--minor` flag
* current version `4.0`
* versions `[..., 4.0, 5.0, 5.1, 6.0, 6.1]`

, we want to group upgrades by how many major bumps there are (`[6.X, 5.X, 4.X]`). This gives us `[6.0, 6.1, 5.0, 5.1, 4.1]`.

bundler/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Unreleased
+
+## Enhancements
+
+  - Add `pre` option to `bundle update` [#5258](https://github.com/rubygems/rubygems/pull/5258)
+
```
deivid-rodriguez, line 6: Our changelog is managed automatically, so this should be removed.

bundler/lib/bundler/gem_version_promoter.rb
```diff
@@ -78,8 +85,8 @@ def filter_dep_specs(specs, package)
 
           must_match = minor? ? [0] : [0, 1]
 
-          matches = must_match.map {|idx| gsv.segments[idx] == lsv.segments[idx] }
-          matches.uniq == [true] ? (gsv >= lsv) : false
+          all_match = must_match.all? {|idx| gsv.segments[idx] == lsv.segments[idx] }
+          all_match && gsv >= lsv
```
deivid-rodriguez, line 89: Nice! Much easier to understand. Normally I extract these little refactorings/improvements to their own commit, but introduce them with the same PR where I come up with them. But I'm happy if you move them to a separate PR too!

By the way, another improvement here would be to return early if `locked_version & !major?` is not met, rather than checking that every time in the `select`.

bundler/spec/commands/update_spec.rb
```diff
@@ -1474,6 +1477,26 @@
...
+        bundle "update --minor foo bar --pre"
+
+        expect(the_bundle).to include_gems "foo 1.5.1", "bar 3.1.0.pre", "qux 1.0.0"
+      end
+
+      it "major preferred" do
+        bundle "update --major foo bar --pre"
+
+        expect(the_bundle).to include_gems "foo 2.0.0.pre", "bar 4.0.0.pre", "qux 1.0.0"
+      end
```
deivid-rodriguez, line 1498: I would also add a spec for `--pre` alone, which should work like `--major`, correct? 

How does `update bar --pre` currently work? Does it unlock `foo` to `2.0.0.pre`, so that it can also upgrade `bar` to `4.0.0.pre`? I _think_ that would be the best behavior?

deivid-rodriguez (commented):
Thanks @siegfault, it's looking very good!

Regarding improvements to gem version promoter, both of your suggestions look nice, happy to take any improvements there. Also I made some changes to this class while introducing PubGrub, and I think it can use some more renaming now since for example, `sort_spec_deps` no longer makes sense as a name since the class now deals with resolution packages, not dependencies.

siegfault:
Thanks for the feedback! I've moved the smaller changes to their own commits, taking your early return suggestion, reverted the changelog bit, and all bundler specs pass for me locally (outside 6 that fail for me on master as well).

With `sort_versions` being so simple and straightforward now (compared to when I first dug into this), it could make sense to let the specs for `GemVersionPromoter` test the public interface. I've pushed up a change with that, but I'm happy to revert it if we want to keep things as is. The diff is a little noisy since the args are expected in reverse order.

Either way, I'll work on adding some tests for `--pre` in `gem_version_promoter_spec`, likely tomorrow

bundler/spec/commands/update_spec.rb
siegfault: Yup, exactly - it works just like `--major`, I think due to https://github.com/rubygems/rubygems/blob/83ba6cff27bd98b79d6a9194f3fd34c2ec87d6f7/bundler/lib/bundler/gem_version_promoter.rb#L30

deivid-rodriguez:
Thanks @siegfault!

Looking forward to those specs, reviewed the other stuff and feels ready to me!

Thanks so much for the changes to version promoter specs, their current shape was a pain to work with when refactoring this class.

siegfault:
Re: the `GemVersionPromoter` specs, I've made a few more changes. I think there are five considerations for these tests (level, pre, strict, locked, and whether the current version is a prerelease version, though the last one overlaps with pre and didn't seem to behave how I expected under test). Rather than write out the 16 (/32 if the last consideration is real) combinations, I wrote most with independent tests for each value. The existing combined tests were maintained (level vs strict) because these seem the most interrelated. Let me know if you'd prefer a different resolution on any of those or if you'd prefer more of the build up (hidden within `sorted_versions` to be within the tests. Thanks so much!

deivid-rodriguez:
@siegfault So sorry I dropped the ball here, I will make sure to include this PR in the final 2.4.0 release before the end of the year. Could you fix the lint failures?

siegfault:
@deivid-rodriguez no worries at all! I know end of year can be pretty busy. My fault for not rerunning rubocop after making the spec changes. Just pushed a change that should fix the hash rockets

deivid-rodriguez (commented):
@siegfault Thanks so much for the fine work on this, and for your patience. This patch looks great to me ❤️.

However, the specs added in the last commit raises some questions, since the current behavior fees quite surprising and I suspect we may want to tweak the implementation instead of codifying the current behavior in specs. I also have some minor comments about wording.

Since I don't want to block this whole PR on that, how about temporarily removing the last commit and merging the rest of the PR as is, and then discussing further the desired behavior of the gem promoter on a separate PR.

bundler/lib/bundler/cli.rb
```diff
@@ -668,6 +670,8 @@ def inject(name, version)
       "If updating, prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "If updating, prefer updating to next major version (default)"
+    method_option "pre", :type => :boolean, :banner =>
+      "If updating, allow consideration of prerelease gems"
```
deivid-rodriguez, line 674: Just noticed that even though the PR mentions only `bundle update`, we're also adding this flag to `bundle lock`? That seems nice! Should we add a simple spec that runs `bundle lock --pre`? Just to make sure that flag is properly channeled through all the way to the version promoter.

siegfault: Good call! Added specs very similar to the ones for bundle update. I'll update the PR description

bundler/lib/bundler/cli.rb
```diff
@@ -292,6 +292,8 @@ def install
       "Prefer updating only to next minor version"
     method_option "major", :type => :boolean, :banner =>
       "Prefer updating to next major version (default)"
+    method_option "pre", :type => :boolean, :banner =>
+      "If updating, allow consideration of prerelease gems"
```
deivid-rodriguez, line 296: Prerelease gems are already considered when dependency specifications mention prereleases, for example if your `Gemfile` or a transitive gemspec specifies something like `rails >= 7.0.0.beta1`, or when they are the only possibility.

Also, the `If updating` seems redundant in the case of `bundle update`, right?

How could we word this flag's help message to clarify the above?

Maybe

```suggestion
      "Always choose the highest allowed version when updating gems, even if it's a prerelease"
```

deivid-rodriguez (commented):
Sorry I'm always seeing some little things right before merging 😅.

bundler/lib/bundler/cli.rb
siegfault: :laughing: no problem! Totally with you on this needing clarification.

What I think is important to convey is that prerelease gems are typically (outside the case you mention) deprioritized and that this flag allows avoiding that. Avoiding deprioritization is a nice double negative that I'm having trouble avoiding, but tweaking your suggestion slightly to `Always chose the highest allowed version when updating gems, regardless of prerelease status` is closer.

What do you think about `"Prioritize versions for selection independent of whether they're prerelease"`? Not loving it (especially prioritize versions for selection seems perhaps too low level for here, but it seems closest to what I imagine is important to convey. Otherwise, happy to go with the suggested text!

deivid-rodriguez: My rationale of mentioning "highest allowed version" is that without being clear on how things are ordered, talking about "prioritizing" feels a bit meaningless.

So from your two suggestions, I'd go with `Always chose the highest allowed version when updating gems, regardless of prerelease status`. I'm neutral between that one, my suggestion above, or some other variant like `Always chose the highest allowed version when updating gems, regardless of whether it's a prerelease`.

So just update with your preferred wording along those lines, and let's ship this!

siegfault: Sounds good! I kept the `if updating` in the lock version and removed the `when updating gems` since all the other similar descriptions in `lock` started that way, but let me know if you'd prefer consistency between the two `--pre` descriptions.

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks so much @siegfault ❤️

siegfault:
Of course! Thanks so much for helping get it merged! I'll open up that follow up spec PR at some point soon

deivid-rodriguez:
Hei @siegfault, just a note since I stumbled on this today. I think there's a bug in the different criteria considered by the gem version promoter, not sure if it's the same behavior that surprised you when writing specs:

Prereleases are NOT considered even when the dependency specification includes a prerelease segment, if the dependency is not locked. To me that doesn't make sense, because I could create a `Gemfile` with say `gem "rails", ">= 7.0.0.beta1"` and I'd expect prereleases considered there even if there's no lock file yet.

I think a patch like this would fix that:

```diff
diff --git a/bundler/lib/bundler/gem_version_promoter.rb b/bundler/lib/bundler/gem_version_promoter.rb
index f08fceb7e2..d281f46eeb 100644
--- a/bundler/lib/bundler/gem_version_promoter.rb
+++ b/bundler/lib/bundler/gem_version_promoter.rb
@@ -93,7 +93,7 @@ def sort_dep_specs(specs, package)
       locked_version = package.locked_version
 
       result = specs.sort do |a, b|
-        unless locked_version && (package.prerelease_specified? || pre?)
+        unless package.prerelease_specified? || pre?
           a_pre = a.prerelease?
           b_pre = b.prerelease?
 
```

I may propose this at some point but if you get to proposing these specs first and you agree, feel free to propose it yourself too together with the specs!

PR #5260, opened 2022-01-05
Use thor-1.2.1

hsbt:

Thor-1.2.1 fixes the regression with `insert_into_file`.

https://github.com/rails/thor/releases/tag/v1.2.1

Run `rake vendor:thor[v1.2.1]`

But I wonder why the current version is `1.1.0`. I think we already updated thor-1.2.0 with this diff.

### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -171,12 +171,9 @@ Automatiek::RakeTask.new("molinillo") do |lib|
   end
 end
 
-# We currently include the official version as of
-# https://github.com/rails/thor/commit/5c666b4c25e748e57eec2d529d94c5059030979e
-# with a change on top to support did_you_mean-1.6.0.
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
> But I wonder why the current version is 1.1.0. I think we already updated thor-1.2.0 with this diff.

It because I manually backported the fix to the `did_you_mean` deprecation messages before it was incorporated upstream. So that we could ship bundler without these warnings without having to wait for upstream Thor to release them.

When I do that, I add a note about it in the automatiek task:

https://github.com/rubygems/rubygems/blob/fbd8de30a681aa5963b1f2fa2335daaa03ee3716/bundler/Rakefile#L174-L186

Can you remove the note now and change the version there to 1.2.1?

Thanks @hsbt!

hsbt:
Ah, I see. Thanks for explanation.

PR #5266, opened 2022-01-07
Don't activate `yaml` gem from RubyGems

deivid-rodriguez:

I'm not sure there's any realworld user problem related to this. My developer problem is that sometimes I need to write some specs that work on an isolated default gem location, and I have to artificially install a bunch of default gems to that location, because Bundler & RubyGems use them internally.

See this abominable spec for example

https://github.com/rubygems/rubygems/blob/a9f9341e08478426ca3aa7829b5be4534cee06c1/bundler/spec/install/gems/standalone_spec.rb#L111-L149

So the less default gems we use internally, the easier is to write this kind of spec.

If I understand correctly, activating the `yaml` gem only defines `YAML` as an alias to `Psych`. Previously we used `Psych` and `YAML` inconsistently. I think it's best to use `Psych` consistently.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -606,17 +606,10 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   def self.load_yaml
     return if @yaml_loaded
 
-    begin
-      # Try requiring the gem version *or* stdlib version of psych.
-      require 'psych'
-    rescue ::LoadError
-      # If we can't load psych, that's fine, go on.
-    else
-      require_relative 'rubygems/psych_additions'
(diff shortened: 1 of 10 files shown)


indirect:
The alias is a leftover from before Psych, when YAML was sometimes Syck. Sometimes it's nice to make progress. :)

deivid-rodriguez:
Aaah I see! Yeah I now remember this `yaml` gem support different "engines" (Syck, Psych): https://github.com/ruby/yaml/commit/5a1367c70e740de5cb92fd869753c9324977157b.

So this seems good?

deivid-rodriguez:
Seems harmless, so let's do it.

PR #5273, opened 2022-01-10
Create minitest file to underscored path in "bundle gem" command with dashed gem name

unasuke:

In "bundle gem" command with dashed name gem (e.g. foo-bar) generates `test/test_foo/bar.rb`, but this file contains undefined class `TestFoo` and moreover, does not include in "bundle exec rake test" target.

Therefore, intentially the first test after gem created is fail, but in case of gem name contains dash character is not.

### step to reproduce

```
$ bundler --version
Bundler version 2.3.4

$ bundle gem foo-bar
Creating gem 'foo-bar'...
MIT License enabled in config
Code of conduct enabled in config
Changelog enabled in config
RuboCop enabled in config
Initializing git repo in /tmp/foo-bar
      create  foo-bar/Gemfile
      create  foo-bar/lib/foo/bar.rb
      create  foo-bar/lib/foo/bar/version.rb
      create  foo-bar/sig/foo/bar.rbs
      create  foo-bar/foo-bar.gemspec
      create  foo-bar/Rakefile
      create  foo-bar/README.md
      create  foo-bar/bin/console
      create  foo-bar/bin/setup
      create  foo-bar/.gitignore
      create  foo-bar/test/test_helper.rb
      create  foo-bar/test/test_foo/bar.rb
      create  foo-bar/.github/workflows/main.yml
      create  foo-bar/LICENSE.txt
      create  foo-bar/CODE_OF_CONDUCT.md
      create  foo-bar/CHANGELOG.md
      create  foo-bar/.rubocop.yml
Gem 'foo-bar' was successfully created. For more information on making a RubyGem visit https://bundler.io/guides/creating_gem.html

$ cd foo-bar
$ sed -i -e 's/gemspec/# gemspec/' Gemfile
$ bundle install
Resolving dependencies...
Using rake 13.0.6
Using ast 2.4.2
Using parallel 1.21.0
Using minitest 5.15.0
Using parser 3.1.0.0
Using regexp_parser 2.2.0
Using rexml 3.2.5
Using ruby-progressbar 1.11.0
Using unicode-display_width 2.1.0
Using bundler 2.3.4
Using rainbow 3.0.0
Using rubocop-ast 1.15.1
Using rubocop 1.24.1
Bundle complete! 3 Gemfile dependencies, 13 gems now installed.

$ bundle exec rake test
Run options: --seed 12319

# Running:



Finished in 0.005437s, 0.0000 runs/s, 0.0000 assertions/s.

0 runs, 0 assertions, 0 failures, 0 errors, 0 skips # not fail!

$ ls -1 test/**/*
test/test_foo/bar.rb
test/test_helper.rb

test/test_foo:
bar.rb

$ mv test/test_foo/bar.rb test/test_foo_bar.rb
$ bundle exec rake test
/tmp/foo-bar/test/test_foo_bar.rb:5:in `<top (required)>': uninitialized constant TestFoo (NameError)
        from /home/unasuke/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader.rb:21:in `require'
        from /home/unasuke/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader.rb:21:in `block in <main>'
        from /home/unasuke/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader.rb:6:in `select'
        from /home/unasuke/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader.rb:6:in `<main>'
rake aborted!
/home/unasuke/.rbenv/versions/3.0.2/bin/bundle:23:in `load'
/home/unasuke/.rbenv/versions/3.0.2/bin/bundle:23:in `<main>'
Tasks: TOP => test
```

The change doings...
(when "bundle gem foo-bar" called)

* create `test/test_foo_bar.rb`
* define `TestFooBar` class in `test/test_foo_bar.rb`

### DIFF
--- a/bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
+++ b/bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
@@ -2,7 +2,7 @@
 
 require "test_helper"
 
-class Test<%= config[:constant_name] %> < Minitest::Test
+class Test<%= config[:constant_name].gsub('::', '') %> < Minitest::Test
   def test_that_it_has_a_version_number
     refute_nil ::<%= config[:constant_name] %>::VERSION
   end
--- a/bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
+++ b/bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
@@ -2,7 +2,7 @@
 
 require "test_helper"
 
-class Test<%= config[:constant_name].gsub('::', '') %> < Minitest::Test
+class <%= config[:minitest_constant_name] %> < Minitest::Test
   def test_that_it_has_a_version_number
     refute_nil ::<%= config[:constant_name] %>::VERSION
   end
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -696,18 +696,22 @@ RSpec.describe "bundle gem" do
     end
 
     context "gem.test setting set to rspec and --test is set to minitest" do
+      let(:underscored_require_path) { require_path.tr("/", "_") }
+
       before do
         bundle "config set gem.test rspec"
         bundle "gem #{gem_name} --test=minitest"
       end
 
       it "builds spec skeleton" do
-        expect(bundled_app("#{gem_name}/test/test_#{require_path}.rb")).to exist
+        expect(bundled_app("#{gem_name}/test/test_#{underscored_require_path}.rb")).to exist
         expect(bundled_app("#{gem_name}/test/test_helper.rb")).to exist
       end
     end
 
     context "--test parameter set to minitest" do
+      let(:underscored_require_path) { require_path.tr("/", "_") }
+
       before do
         bundle "gem #{gem_name} --test=minitest"
       end
@@ -722,7 +726,7 @@ RSpec.describe "bundle gem" do
       end
 
       it "builds spec skeleton" do
-        expect(bundled_app("#{gem_name}/test/test_#{require_path}.rb")).to exist
+        expect(bundled_app("#{gem_name}/test/test_#{underscored_require_path}.rb")).to exist
         expect(bundled_app("#{gem_name}/test/test_helper.rb")).to exist
       end
 
@@ -731,11 +735,11 @@ RSpec.describe "bundle gem" do
       end
 
       it "requires 'test_helper'" do
-        expect(bundled_app("#{gem_name}/test/test_#{require_path}.rb").read).to include(%(require "test_helper"))
+        expect(bundled_app("#{gem_name}/test/test_#{underscored_require_path}.rb").read).to include(%(require "test_helper"))
       end
 
       it "creates a default test which fails" do
-        expect(bundled_app("#{gem_name}/test/test_#{require_path}.rb").read).to include("assert false")
+        expect(bundled_app("#{gem_name}/test/test_#{underscored_require_path}.rb").read).to include("assert false")
       end
     end
 
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -30,6 +30,8 @@ RSpec.describe "bundle gem" do
 
   let(:minitest_test_file_path) { "test/test_mygem.rb" }
 
+  let(:minitest_test_class_name) { "class TestMygem < Minitest::Test" }
+
   before do
     sys_exec("git config --global user.name 'Bundler User'")
     sys_exec("git config --global user.email user@example.com")
@@ -736,6 +738,10 @@ RSpec.describe "bundle gem" do
         expect(bundled_app("#{gem_name}/#{minitest_test_file_path}").read).to include(%(require "test_helper"))
       end
 
+      it "defines valid test class name" do
+        expect(bundled_app("#{gem_name}/#{minitest_test_file_path}").read).to include(minitest_test_class_name)
+      end
+
       it "creates a default test which fails" do
         expect(bundled_app("#{gem_name}/#{minitest_test_file_path}").read).to include("assert false")
       end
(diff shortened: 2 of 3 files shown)


simi:
:thinking: If I understand it well, we should keep it consistent and `foo-bar/lib/foo/bar.rb` file should have test in `foo-bar/test/foo/bar_test.rb`.

unasuke:
Um, fix the Rakefile is the more better way maybe, right? 

https://github.com/rubygems/rubygems/blob/a9f9341e08478426ca3aa7829b5be4534cee06c1/bundler/lib/bundler/templates/newgem/Rakefile.tt#L6-L13

hsbt:
`minitest` uses `test_` prefix. But I think `_test` suffix is popular usage like Rails. I have no strong opinion to stay the the current implementation. I'm okay to migrate to `_test.rb` like @simi 's suggestion.

deivid-rodriguez:
Note that we made the reverse change very recently: https://github.com/rubygems/rubygems/pull/3893. I do like the `_test` default better, but we have already changed the default once, so I think we should make a decision and stick to it.

In any case, this bug is different to that, isn't it?

> I think it's more better to generate `test/foo/test_bar.rb`. But it's bit hassle. I want to ask maintainer's opinion.

I also think this should be the ideal fix. By "hassle" you mean that this alternative fix is more complicated to implement than the one you proposed here, right?

simi:
The problem is when using "namespaced" gem name.

```
      create  foo-bar/test/test_helper.rb
      create  foo-bar/test/test_foo/bar.rb
```

It creates test file without `test` prefix/suffix, but it creates "namespace" directory with `test` prefix. It should be one of

```
      create  foo-bar/test/foo/test_bar.rb
OR
      create  foo-bar/test/foo/bar_test.rb
```

deivid-rodriguez:
Yeah, sure. What I'm saying is that we should go with 

```
create  foo-bar/test/foo/test_bar.rb
```

to avoid changing again the default that just changed very recently.

simi:
Suffix makes more sense IMHO, sorry for missing #3893. Anyway I think any variant is OK. Let's go with `foo-bar/test/foo/test_bar.rb` for now.

unasuke:
I addressed! b3a876f

bundler/spec/commands/newgem_spec.rb
```diff
@@ -1355,6 +1359,8 @@ def create_temporary_dir(dir)
 
     let(:require_relative_path) { "gem" }
 
+    let(:minitest_test_file_path) { "test/test/test_gem.rb" }
```
simi, line 1362: Calling test gem "test-gem" was really bad idea :) It is really confusing here. Anyway we can change this later. It is not problem of this PR.

bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
```diff
@@ -2,7 +2,7 @@
 
 require "test_helper"
 
-class Test<%= config[:constant_name] %> < Minitest::Test
+class Test<%= config[:constant_name].gsub('::', '') %> < Minitest::Test
```
simi, line 5: :thinking: Is there any test covering this change?

deivid-rodriguez (approved):
Looks good to me!

bundler/lib/bundler/templates/newgem/test/minitest/test_newgem.rb.tt
deivid-rodriguez: Good question, actually.

unasuke: Um, missing. I will add a test case for the test file containing "class TestFooBar < Minitest::Test" in the body.

unasuke: Add here 5f698fc

deivid-rodriguez: It seems ok now @simi, go ahead and merge if it looks good to you!

bundler/spec/commands/newgem_spec.rb
```diff
@@ -1355,6 +1367,10 @@ def create_temporary_dir(dir)
...
+
+    let(:minitest_test_class_name) { "class TestTestGem < Minitest::Test" }
```
simi, line 1372: @deivid-rodriguez @hsbt is this expected testing class name? That's one of the benefits for using test as a suffix, since you can just copy the original class name and append "Test" to it. Prepending doesn't work well using the same logic, mainly when "namespaced" class is used.

deivid-rodriguez: I see, you were expecting `Test::TestGem`, right? I'm not really sure what people use since I never used the `test_` prefix, but yeah you have a point. Hopefully @hsbt or @unasuke can clarify.

unasuke: @simi  @deivid-rodriguez I'm confusing because gem name contains "test". 🤣 

To clarify, I'm using "foo-bar"  as a gem name for the below examples.

`bundle gem foo-bar --test=minitest` generates class `TestFoo::Bar` in `test/test_foo/bar.rb` (by latest bundler)

This patch generates `TestFooBar` class instead.

What name of class do you expect to generate in "bundle gem" command?

deivid-rodriguez: I think I would expect `Foo::TestBar` in `test/foo/test_bar.rb`?

unasuke: I commited c946ee6d . @simi @deivid-rodriguez is it expected one?

bundler/.ruby-version
```diff
@@ -0,0 +1 @@
+3.0.2
```
simi, line 1: Can you remove this file please? Is that committed by accident?

deivid-rodriguez:
Alright let's do this, thanks so much for your patience @unasuke!

PR #5275, opened 2022-01-10
Let `Version#<=>` accept a String

amatsuda:

Currently, `Gem::Version#<=>` accepts Gem::Version instance only. But in my experience, the major use case for comparing gem versions is that we just want to ask via String. And I want to do this in a simpler way.

For instance, when we need to know if Bundler version is version 1 or 2:

```ruby
bundler = Gem.loaded_specs['bundler']
```

we have to write some clumsy code with `Gem::Version.new` like this.

```ruby
if bundler.version > Gem::Version.new('2')
  ...
end
```

So here's a patch that extends `Version#<=>` to accept a String parameter. It's nothing but just a syntax sugar, but the previous example became much easier to read and write now.

```ruby
if bundler.version > '2'
  ...
end
```

Even after this patch, if an invalid version String is given:
```ruby
if bundler.version > 'hello'
  ...
end
```

it raises an ArgumentError `comparison of Gem::Version with String failed`. This behavior is unchanged.

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -341,9 +341,11 @@ class Gem::Version
   # Compares this version with +other+ returning -1, 0, or 1 if the
   # other version is larger, the same, or smaller than this
   # one. Attempts to compare to something that's not a
-  # <tt>Gem::Version</tt> return +nil+.
+  # <tt>Gem::Version</tt> or a valid version String return +nil+.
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
I like this, I would've found this useful myself many times.

deivid-rodriguez:
Thank you!

PR #5277, opened 2022-01-11
Don't activate `yaml` gem from Bundler

deivid-rodriguez:

No real end user issue, just a developer problem being fixed: the less default gems we use internally, the better for us. It also allows for simplifying code and cleaning up.

Simplify `psych` activation and remove legacy code.

This PR is equivalent to #5266 but for Bundler.


deivid-rodriguez:
I'm actually running into an issue with Psych loading over at #3963 which I think this PR will fix, so I will merge it.

PR #5283, opened 2022-01-13
Test against Ruby 3.1

deivid-rodriguez:

Our CI should now support Ruby 3.1 on all OS, so we can start testing.

Add Ruby 3.1 to matrices, and also simplify some things to not let the matrix grow too big:

* Set Ruby 2.7 as the minimum tested Ruby on Windows, MacOS, and Bundler 3. It used to be Ruby 2.5 or Ruby 2.6, but Ruby 2.6 will reach its end of life in March, so we can probably also drop support for it next year together with 2.3, 2.4 and 2.5.
* When testing bundler against old RubyGems, only test against the oldest series supported on each version, not against the newest release. Bundler's master is already tested against RubyGems master, I think Bundler master against the latest RubyGems release provides little value, and it has never detected any issues that the master run did not detect.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -29,6 +29,7 @@ jobs:
           - { name: ruby-2.6, value: 2.6.9 }
           - { name: ruby-2.7, value: 2.7.5 }
           - { name: ruby-3.0, value: 3.0.3 }
+          - { name: ruby-3.1, value: 3.1.0 }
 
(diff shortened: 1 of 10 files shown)


simi:
:pray: for the CI

deivid-rodriguez:
Not fully green but not too bad either, only a single spec failure in Bundler on Ruby 3.1 + Windows.

PR #5291, opened 2022-01-15
Fix regression with old marshaled specs having null `required_rubygems_version`

### LINKED ISSUE #5290: bundler-2.3.5: undefined method `satisfied_by?' for nil:NilClass

On bundler 2.3.5, in my linux docker container, I see the following error report.  The Gemfile is below and as you can see it uses a single gem, from a third-party gem server.  I've been trying to dig in, and one of the dependencies (pdf-writer) seems to create this nil reference error.  I can't make this happen in my normal Mac environment though, so I don't really understand it beyond this.  There's a Dockerfile down beyond to make it simpler to recreate. If you need to me to try anything else, feel free to suggest things.

Might be related to #4877 

```
NoMethodError: undefined method `satisfied_by?' for nil:NilClass
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/lazy_specification.rb:95:in `block in __materialize__'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/lazy_specification.rb:92:in `select'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/lazy_specification.rb:92:in `__materialize__'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/spec_set.rb:75:in `block in materialize'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/spec_set.rb:72:in `map!'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/spec_set.rb:72:in `materialize'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/definition.rb:468:in `materialize'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/definition.rb:190:in `specs'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/installer.rb:243:in `ensure_specs_are_compatible!'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/installer.rb:83:in `block in run'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/process_lock.rb:12:in `block in lock'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/process_lock.rb:9:in `open'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/process_lock.rb:9:in `lock'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/installer.rb:71:in `run'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/installer.rb:23:in `install'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/cli/install.rb:62:in `run'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/cli.rb:255:in `block in install'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/settings.rb:131:in `temporary'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/cli.rb:254:in `install'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/vendor/thor/lib/thor.rb:392:in `dispatch'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/cli.rb:31:in `dispatch'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/vendor/thor/lib/thor/base.rb:485:in `start'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/cli.rb:25:in `start'
  /usr/share/gems/gems/bundler-2.3.5/exe/bundle:48:in `block in <top (required)>'
  /usr/share/gems/gems/bundler-2.3.5/lib/bundler/friendly_errors.rb:103:in `with_friendly_errors'
  /usr/share/gems/gems/bundler-2.3.5/exe/bundle:36:in `<top (required)>'
  /usr/bin/bundle:23:in `load'
  /usr/bin/bundle:23:in `<main>'
```

## Environment

```
Bundler       2.3.5
Ruby          2.7.4p191 (2021-07-07 revision a21a3b7d23704a01d34bd79d09dc37897e00922a) [x86_64-linux]
  Full Path   /usr/bin/ruby
  Config Dir  /etc
RubyGems      3.1.6
  Gem Home    /usr/share/gems
  Gem Path    /root/.gem/ruby:/usr/share/gems:/usr/local/share/gems
  User Home   /root
  User Path   /root/.gem/ruby
  Bin Dir     /usr/bin
OpenSSL
  Compiled    OpenSSL 1.1.1g FIPS  21 Apr 2020
  Loaded      OpenSSL 1.1.1k  FIPS 25 Mar 2021
  Cert File   /etc/pki/tls/cert.pem
  Cert Dir    /etc/pki/tls/certs
Tools
  Git         2.31.1
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2022-01-12
Git SHA           a13d015fcb
Released Version  true
```

### Gemfile

```ruby
source "https://rubygems.org"

gem "ruport", "=1.7.0.3", :source => "https://rubygems.manageiq.org"
```

### Gemfile.lock

```
<No /test/Gemfile.lock found>
```

This uses the latest version of bundler 2.3.5, and rubygems 3.1.6.  I can't seem to update rubygems because `gem update --system` seems to just hang forever.

Note that bundler 2.3.4 works fine.

This Dockerfile is the simplest reproducer I could get.  It will fail on the `bundle install` line when you try to `docker build` it.  If you comment out that line, it will drop you into a bash prompt where you can type `bundle install` manually.  You can also change the version of bundler to 2.3.4 to see that it works correctly.

```dockerfile
FROM registry.access.redhat.com/ubi8/ubi:8.4

RUN dnf -y module enable ruby:2.7 && dnf -y install ruby-devel
RUN gem install bundler:2.3.5

WORKDIR "test"

RUN echo -e 'source "https://rubygems.org"\n\n\
gem "ruport", "=1.7.0.3", :source => "https://rubygems.manageiq.org"'\
> Gemfile

RUN bundle install

ENTRYPOINT ["/bin/bash"]
```

`bundle install`

Not to blow up

### What actually happened?

It blew up

deivid-rodriguez:

I made [a commit](https://github.com/rubygems/rubygems/commit/ab0a1c16c9e92e175e50e2e386105725a5c7cc87) that failed to consider that some old marshaled specs may have `required_rubygems_version` set to `nil`.

Add a fallback for this edge case.

Fixes #5290.

### DIFF
--- a/bundler/lib/bundler/remote_specification.rb
+++ b/bundler/lib/bundler/remote_specification.rb
@@ -27,6 +27,13 @@ module Bundler
       @platform = _remote_specification.platform
     end
 
+    # A fallback is included because the original version of the specification
+    # API didn't include that field, so some marshalled specs in the index have it
+    # set to +nil+.
+    def required_rubygems_version
+      @required_rubygems_version ||= _remote_specification.required_rubygems_version || Gem::Requirement.default
+    end
+
     def full_name
       if platform == Gem::Platform::RUBY || platform.nil?
         "#{@name}-#{@version}"
--- a/bundler/spec/install/gemfile/sources_spec.rb
+++ b/bundler/spec/install/gemfile/sources_spec.rb
@@ -1443,4 +1443,57 @@ RSpec.describe "bundle install with gems on multiple sources" do
       L
     end
   end
+
+  context "when default source includes old gems with nil required_rubygems_version" do
+    before do
+      build_repo2 do
+        build_gem "ruport", "1.7.0.3" do |s|
+          s.add_dependency "pdf-writer", "1.1.8"
+        end
+      end
+
+      build_repo gem_repo4 do
+        build_gem "pdf-writer", "1.1.8"
+      end
+
+      path = "#{gem_repo4}/#{Gem::MARSHAL_SPEC_DIR}/pdf-writer-1.1.8.gemspec.rz"
+      spec = Marshal.load(Bundler.rubygems.inflate(File.binread(path)))
+      spec.instance_variable_set(:@required_rubygems_version, nil)
+      File.open(path, "wb") do |f|
+        f.write Gem.deflate(Marshal.dump(spec))
+      end
+
+      gemfile <<~G
+        source "https://localgemserver.test"
+
+        gem "ruport", "= 1.7.0.3", :source => "https://localgemserver.test/extra"
+      G
+    end
+
+    it "handles that fine" do
+      bundle "install", :artifice => "compact_index_extra", :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }
+
+      expect(lockfile).to eq <<~L
+        GEM
+          remote: https://localgemserver.test/
+          specs:
+            pdf-writer (1.1.8)
+
+        GEM
+          remote: https://localgemserver.test/extra/
+          specs:
+            ruport (1.7.0.3)
+              pdf-writer (= 1.1.8)
+
+        PLATFORMS
+          #{specific_local_platform}
+
+        DEPENDENCIES
+          ruport (= 1.7.0.3)!
+
+        BUNDLED WITH
+           #{Bundler::VERSION}
+      L
+    end
+  end
 end
--- a/bundler/spec/install/gems/resolving_spec.rb
+++ b/bundler/spec/install/gems/resolving_spec.rb
@@ -101,8 +101,6 @@ RSpec.describe "bundle install with install-time dependencies" do
   end
 
   it "installs gems with a dependency with no type" do
-    skip "incorrect data check error" if Gem.win_platform?
-
     build_repo2
 
     path = "#{gem_repo2}/#{Gem::MARSHAL_SPEC_DIR}/actionpack-2.3.2.gemspec.rz"
@@ -110,7 +108,7 @@ RSpec.describe "bundle install with install-time dependencies" do
     spec.dependencies.each do |d|
       d.instance_variable_set(:@type, :fail)
     end
-    File.open(path, "w") do |f|
+    File.open(path, "wb") do |f|
       f.write Gem.deflate(Marshal.dump(spec))
     end
 


bundler/spec/realworld/double_check_spec.rb
```diff
@@ -37,4 +37,12 @@
     ruby cmd
     ruby cmd
   end
+
+  it "deals with old marshaled specs with nil required_rubygems_version" do
+    install_gemfile <<~G
+      source "https://rubygems.org"
+
+      gem "ruport", "=1.7.0.3", :source => "https://rubygems.manageiq.org"
```
Fryguy, line 45: I'm not sure this is a good idea in specs, only because there's a potential it could go away in the future (unless this doesn't actually hit the live system? I see some recorded http things below).  For right now though, it's probably fine.

For testing purposes does rubygems have a third party package registry. If not, maybe Github package registry could be used for that purpose and dedicated to testing rubygems/bundler?

deivid-rodriguez: @Fryguy Correct, this doesn't hit the live system, it uses the recorded responses. Normally I tried to avoid realworld specs like this one, I only use them when I cannot find another way to test what I want to test. In this case, the tricky part was that our current spec setup does not support setting up multiple gem servers when some of them support new APIs and some of them don't. But I will give it another try.

Regarding a testing package registry,  I guess it could make sense although it feels slightly overkill for now given how little use we make of this realworld test harness.

Fryguy: That's fair. As the owner of that gem server, we have no intention of removing anything, since we use that for production builds, so for now it's fine, but who knows what could happen in the future.

Thanks for the quick fix!

deivid-rodriguez: I gave it another try and I managed to write a better spec, it should be ok now!

PR #5292, opened 2022-01-17
Avoid writing private registry credentials to logs

### LINKED ISSUE #5281: private registry credentials are written to the logs when `bundle install` fails due to `Bundler::PermissionError` error

username and password of a private registry that was configured as an environment variable like `BUNDLE_GITHUB__COM` are visible when the `bundle install` command have errors:
```
Command result error lines for "bundle install":
Read error line #1: [DEPRECATED] Your Gemfile contains multiple primary sources. Using `source` more than once without a block is a security risk, and may result in installing unexpected gems. To resolve this warning, use a block to indicate which gems should come from the secondary source.
Read error line #2: Retrying download gem from https://sami:notmyrealpassword@example.com due to error (2/4): Bundler::PermissionError There was an error while trying to write to `/usr/local/ruby/2.7.5/lib/ruby/gems/2.7.0/cache/faraday-em_synchrony-1.0.0.gem`. It is likely that you need to grant write permissions for that path.
Read error line #3: Retrying download gem from https://sami:notmyrealpassword@example.com due to error (2/4): Bundler::PermissionError There was an error while trying to write to `/usr/local/ruby/2.7.5/lib/ruby/gems/2.7.0/cache/faraday-em_http-1.0.0.gem`. It is likely that you need to grant write permissions for that path.
```

yes, using bundler 2.3.4 and ruby 2.7.5

install `bundler` with root permissions.
Configure environment variable for holding the credentials information, such that: `export BUNDLE_EXAMPLE__COM=username:password`
Then add a dependency to a private registry and include it in your Gemfile with multiple sources:
 ```
source 'https://rubygems.org'
source 'https://private.example.com'
gem 'private_gem', '~> 1.0.0'
```
run `bundle install` (not in root)

`bundle install`

The logs will not present the real password

### What actually happened?

We can see the real password of the private registry login credentials printed on the logs.

samisalamiws:

username and password of a private registry that was configured as an environment variable like `BUNDLE_GITHUB__COM` are visible when the bundle install command has errors

We duplicated the printed URI to the logs and escaped the username/password from this duplicated URI to be **FILTERED**

Fixes #5281.

### DIFF
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -494,7 +494,13 @@ module Bundler
       uri = Bundler.settings.mirror_for(uri)
       fetcher = gem_remote_fetcher
       fetcher.headers = { "X-Gemfile-Source" => spec.remote.original_uri.to_s } if spec.remote.original_uri
-      Bundler::Retry.new("download gem from #{uri}").attempts do
+
+      uri_for_print = uri.dup
+
+      uri_for_print.user = "**FILTERED**" if uri_for_print.user
+      uri_for_print.password = "**FILTERED**" if uri_for_print.password
+
+      Bundler::Retry.new("download gem from #{uri_for_print}").attempts do
         gem_file_name = spec.file_name
         local_gem_path = File.join cache_dir, gem_file_name
         return if File.exist? local_gem_path
@@ -517,7 +523,7 @@ module Bundler
         end
       end
     rescue Gem::RemoteFetcher::FetchError => e
-      raise Bundler::HTTPError, "Could not download gem from #{uri} due to underlying error <#{e.message}>"
+      raise Bundler::HTTPError, "Could not download gem from #{uri_for_print} due to underlying error <#{e.message}>"
     end
 
     def gem_remote_fetcher
--- a/bundler/spec/bundler/rubygems_integration_spec.rb
+++ b/bundler/spec/bundler/rubygems_integration_spec.rb
@@ -43,7 +43,6 @@ RSpec.describe Bundler::RubygemsIntegration do
 
   describe "#download_gem" do
     let(:bundler_retry) { double(Bundler::Retry) }
-    let(:uri) { Bundler::URI.parse("https://foo.bar") }
     let(:cache_dir) { "#{Gem.path.first}/cache" }
     let(:spec) do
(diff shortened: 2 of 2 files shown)


deivid-rodriguez (commented):
Thank you! We have similar logic in RubyGems, so we should probably reuse that. I think something like, `Gem::Uri.new(uri).redacted` should do the trick.

The tricky part is that it's only a recent RubyGems addition, so you need to backport the functionality it manually for older RubyGems that don't have the feature. That means define some `Bundler` class that duplicates the logic and that gets used only when RubyGems does not define `Gem::Uri`.

mensfeld:
@deivid-rodriguez would you mind approving the PR runs please? :pray:

deivid-rodriguez:
Done!

deivid-rodriguez:
Answering some questions asked privately to me by @mensfeld. Please ask public questions, so that everyone can benefit from the answers and leave their opinion, and others can share the support load with me 🙏.

> 1. can we use `Gem.const_defined?(:NoAliasYAMLTree)`  to check for the `Uri` presence? I don't want to have to figure out the exact versions and match those but rather just check for the object presence

Fine with me (I think you meant `Gem.const_defined?(:Uri)`?). But I think it’s good to also figure out where `Gem::Uri` was first shipped and add a note about it, so that we don’t forget to remove the backwards compatibility code we add now when the time comes.

> 2. Can Sami promote this issue as a CVE one?

We've had this same issue reported and fixed in several other places in the past, and potentially more dangerous, like `bundle env`? Do you think it's worth changing this criteria now? I don't have a strong opinion but I think it's fine to handle this in the same way we've handled it in the past.

> 3. btw the Uri you pointed to does not redact the user, shouldn't we redact user name as well aside from the password? in stupd cases like admin/admin it may be a security problem to have it in the logs (though not sure here)

In my opinion not redacting the user name is fine. If someone has admin/admin as their user/password combination, printing the user name to a user screen sounds like the smallest of her problems 😆.

timon:
> In my opinion not redacting the user name is fine.

The credential string is not necessary `user:password`, it could be just `token` instead

deivid-rodriguez:
Closing in favor of #8222. I opened a separate PR but I think I don't have permission to push to your fork (maybe because you created a PR from the master branch, not sure).

PR #5296, opened 2022-01-17
Use random order by default for RSpec

santib:

As a dev, when creating gems sometimes I forget to set the order of the tests to be random.

I think there is consensus in the community that tests can't depend on their order, so that's why I'm proposing to add `--order random` by default when creating the `.rspec` file on a new gem.

- [ ] Write [tests] (https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes - I didn't find tests for the rest of the configs, so I guess they are not needed?

### DIFF
--- a/bundler/lib/bundler/templates/newgem/rspec.tt
+++ b/bundler/lib/bundler/templates/newgem/rspec.tt
@@ -1,3 +1,4 @@
---format documentation
 --color
+--format documentation
+--order random
 --require spec_helper


deivid-rodriguez:
Hello! 👋

I think this is an interesting debate. There's definitely consensus that tests should not depend on the order they are run, but I don't think there's consensus on whether tests should be run by default in random order or not. The proof that there's no consensus is that RSpec doesn't do this by default, right?

I definitely don't want my tests to be order dependent, but since running tests in random order might or might not catch test ordering side-effects, I'm not sure I want order dependency issues to come up at random times (potentially at times different from the time where the issue is first introduced). So perhaps I would prefer to not use `--order random` by default, but only run with that flags explicitly to help catching ordering issues, like once a day or something.

So, I'm not totally sure about this, but I believe a discussion to change the default should take place at the RSpec repository. What do you think?

santib:
Hey @deivid-rodriguez good point, I agree with you. I assume that most people use `--order random` to eventually catch these issues but I've no clue why isn't this a default on `rspec` (there might be a reason I guess?). I can ask this on the `rspec` repo and I agree to have `bundler` respect whatever each gem chooses as its default.

Thanks 👍

deivid-rodriguez:
No problem! Feel free to link the discussion to this issue, I find it interesting and may participate :)

PR #5303, opened 2022-01-23
Document 128 magic number

### LINKED ISSUE #5165: Document magic numbers in exit code handling in our test helpers
We use a some magic number when dealing with exit codes in our help that runs subprocesses via `Open3.popen3`.

Here: https://github.com/rubygems/rubygems/blob/f62cad4818cc333b40d506d859ca53531f2f25f6/bundler/spec/support/subprocess.rb#L39-L54

We should probably add a link to https://tldp.org/LDP/abs/html/exitcodes.html and move it to a constant/variable with some non magical name.

_Originally posted by @doodzik in https://github.com/rubygems/rubygems/pull/4992/files#r734699098_

gustavothecoder:
We use a some magic number when dealing with exit codes in our help that runs sub processes via `Open3.popen3` (#5165). 

I moved the line of code where the magic number was used (`128 + status.termsig`) to a function and used the function name to describe the exit code handling.

Fixes #5165.

### DIFF
--- a/bundler/spec/commands/exec_spec.rb
+++ b/bundler/spec/commands/exec_spec.rb
@@ -775,9 +775,7 @@ RSpec.describe "bundle exec" do
       end
       let(:expected_err) { "" }
       let(:exit_code) do
-        # signal mask 128 + plus signal 15 -> TERM
-        # this is specified by C99
-        128 + 15
+        exit_status_for_signal(Signal.list["TERM"])
       end
 
       it "runs" do
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -199,7 +199,7 @@ module Spec
         command_execution.exitstatus = if status.exited?
           status.exitstatus
         elsif status.signaled?
-          128 + status.termsig
+          exit_status_for_signal(status.termsig)
         end
       end
 
@@ -558,6 +558,11 @@ module Spec
         false
       end
       port
+    end
+
+    def exit_status_for_signal(signal_number)
+      # For details see: https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts
+      128 + signal_number
     end
 
     private


deivid-rodriguez:
@gustavothecoder This seems fine, but, could you check something for me first? Can you try completely reverting https://github.com/rubygems/rubygems/commit/814e293e23fbb9452fa23bb5aa42389ec5e5d731, and see if any spec breaks?

gustavothecoder:
> @gustavothecoder This seems fine, but, could you check something for me first? Can you try completely reverting [814e293](https://github.com/rubygems/rubygems/commit/814e293e23fbb9452fa23bb5aa42389ec5e5d731), and see if any spec breaks?

After the revert that spec broke:
```
rspec ./spec/commands/exec_spec.rb:783 # [TEST_ENV_NUMBER=7] bundle exec `load`ing a ruby file instead of `exec`ing the executable exits by SignalException runs
```

deivid-rodriguez:
Aha! That's the 128 magic number again there, so we can probably reuse this new helper there.

https://github.com/rubygems/rubygems/blob/7c774fbc8497979e119465e7020620f1c3464763/bundler/spec/commands/exec_spec.rb#L777-L781

Also, I would use `Signal.list["TERM"]` to avoid the magic number "15" too.

Finally, I would rename the helper method to `exit_status_for_signal(signal_number)`, and point to some documentation that's not bash specific (apparently this is a standard for most shells). For example, https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts.

gustavothecoder:
> Aha! That's the 128 magic number again there, so we can probably reuse this new helper there.
> 
> https://github.com/rubygems/rubygems/blob/7c774fbc8497979e119465e7020620f1c3464763/bundler/spec/commands/exec_spec.rb#L777-L781
> 
> Also, I would use `Signal.list["TERM"]` to avoid the magic number "15" too.
> 
> Finally, I would rename the helper method to `exit_status_for_signal(signal_number)`, and point to some documentation that's not bash specific (apparently this is a standard for most shells). For example, https://en.wikipedia.org/wiki/Exit_status#Shell_and_scripts.

@deivid-rodriguez adjustments made : )

simi:
CI has same problem as at https://github.com/rubygems/rubygems/pull/5321#issuecomment-1024950338. Probably not related to this PR.

deivid-rodriguez (approved):
Looks good to me, just missing a rebase! Thank you @gustavothecoder

gustavothecoder:
> Looks good to me, just missing a rebase! Thank you @gustavothecoder

Made. Sorry for the delay.

deivid-rodriguez:
Thanks so much @gustavothecoder!

Issue #4478, opened 2021-03-20
`force_ruby_platfom` is ignored if platform is set in `Gemfile.lock`


ojab:
### Describe the problem as clearly as you can

If platform is set in `Gemfile.lock` — gems for platform in `Gemfile.lock` are installed after setting `force_ruby_platfom`.

### Post steps to reproduce the problem

linux x86_64, ruby-3.0.0 from rbenv.

### Which command did you run?

```
+ bundle init
Writing new Gemfile to /tmp/test/Gemfile
+ bundle add nokogiri
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Using racc 1.5.2
Using bundler 2.2.15
Using nokogiri 1.11.2 (x86_64-linux)
Updating files in vendor/cache
+ bundle config set --local force_ruby_platform true
+ bundle install
Using bundler 2.2.15
Using racc 1.5.2
Using nokogiri 1.11.2 (x86_64-linux)
Updating files in vendor/cache
Bundle complete! 1 Gemfile dependency, 3 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
+ bundle info nokogiri
  * nokogiri (1.11.2)
	Summary: Nokogiri (鋸) makes it easy and painless to work with XML and HTML from Ruby.
	Homepage: https://nokogiri.org
	Documentation: https://nokogiri.org/rdoc/index.html
	Source Code: https://github.com/sparklemotion/nokogiri
	Changelog: https://nokogiri.org/CHANGELOG.html
	Bug Tracker: https://github.com/sparklemotion/nokogiri/issues
	Path: /home/ojab/.rbenv/versions/3.0.0/lib/ruby/gems/3.0.0/gems/nokogiri-1.11.2-x86_64-linux
```

### What were you expecting to happen?

gems for ruby platform will be downloaded and installed.

### What actually happened?

gems for `x86_64-linux` is used

## Environment

```
Bundler       2.2.15
  Platforms   ruby, x86_64-linux
Ruby          3.0.0p0 (2020-12-25 revision 95aff214687a5e12c3eb57d056665741e734c188) [x86_64-linux]
  Full Path   /home/ojab/.rbenv/versions/3.0.0/bin/ruby
  Config Dir  /home/ojab/.rbenv/versions/3.0.0/etc
RubyGems      3.2.15
[...]

PR #5304, opened 2022-01-24
Fix `force_ruby_platform` ignored when lockfile includes current specific platform

deivid-rodriguez:

When lockfile includes the current real specific platform, the `force_ruby_platform` configuration is ignored. Bundler should force re-resolve using the RUBY platform instead.
 
Add the proper condition to force a re-resolve because of a new platform being considered (RUBY).

Fixes https://github.com/rubygems/rubygems/issues/4478.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -496,6 +496,7 @@ module Bundler
 
     def current_ruby_platform_locked?
       return false unless generic_local_platform == Gem::Platform::RUBY
+      return false if Bundler.settings[:force_ruby_platform] && !@platforms.include?(Gem::Platform::RUBY)
 
       current_platform_locked?
     end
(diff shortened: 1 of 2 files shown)


ojab:
How it would interact with frozen/deployment? I guess it would fail?

deivid-rodriguez:
Correct, this will result in the RUBY platform being added to the lockfile. Since frozen does not permit changing the lockfile, it will fail.

ojab:
Heh, IMHO ideally we should somehow try to install the same versions as we have in the lockfile (just for selected platform) and fail if they're not available, but this PR is also a proper fix, just with different behavior. And I don't think that `IMHO` version is doable without way more complicated changes.

deivid-rodriguez:
I see, yeah, that would be another approach. And I think it's slightly better than this one. But I believe this one is an improvement over what we have now.

PR #5305, opened 2022-01-24
Enable mfa on specific keys during gem signin

aellispierce:
Part of: https://github.com/rubygems/rubygems.org/issues/2755
Follow up to https://github.com/rubygems/rubygems.org/pull/2922

**What does this change do:**
- Requests the users mfa level from the server on gem signin
- If the user has an mfa level of `ui_only` or `ui_and_gem_sign`, asks the user if they would like to enable mfa on the key they are creating
- If the user has an mfa level of `ui_and_api`, turns mfa on for the key without asking 

**Why is this necessary:**

MFA can be enabled on specific API keys through the UI (ref: https://github.com/rubygems/rubygems.org/pull/2846), so a user should also be able to enable it on keys that they create during `gem signin` in the CLI.

We only want to ask a user if they would like to enable mfa on new keys if they have account mfa levels of `ui_only` or `ui_and_gem_sign`. Users that have MFA disabled or those that have it  enabled for `ui_and_api ` should not be prompted, as it should be auto enabled or disabled for those levels.

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -260,8 +260,8 @@ module Gem::GemcutterUtilities
     else
       say "Please select scopes you want to enable for the API key (y/n)"
       API_SCOPES.each do |scope|
-        selected = ask "#{scope} [y/N]: "
-        scope_params[scope] = true if selected =~ /^[yY](es)?$/
+        selected = ask_yes_no("#{scope}", false)
+        scope_params[scope] = true if selected
       end
       say "\n"
     end
@@ -279,8 +279,8 @@ module Gem::GemcutterUtilities
     mfa_level = get_user_mfa_level(email, password)
     params = {}
     if mfa_level == "ui_only" || mfa_level == "ui_and_gem_signin"
-      selected = ask "Would you like to enable MFA for this key? [Y/n]"
-      params["mfa"] = true unless selected =~ /^[nN](o)?$/
+      selected = ask_yes_no("Would you like to enable MFA for this key? (strongly recommended)")
+      params["mfa"] = true if selected
     end
     params
   end
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -165,12 +165,14 @@ module Gem::GemcutterUtilities
     key_name     = get_key_name(scope)
     scope_params = get_scope_params(scope)
     mfa_params   = get_mfa_params(email, password)
+    all_params   = scope_params.merge(mfa_params)
+
 
     response = rubygems_api_request(:post, "api/v1/api_key",
                                     sign_in_host, scope: scope) do |request|
       request.basic_auth email, password
       request["OTP"] = otp if otp
-      request.body = URI.encode_www_form({ name: key_name }.merge(scope_params, mfa_params))
+      request.body = URI.encode_www_form({ name: key_name }.merge(all_params))
     end
 
     with_response response do |resp|
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require_relative 'remote_fetcher'
 require_relative 'text'
+require 'json'
 
 ##
 # Utility methods for using the RubyGems API.
@@ -163,12 +164,13 @@ module Gem::GemcutterUtilities
 
     key_name     = get_key_name(scope)
     scope_params = get_scope_params(scope)
+    mfa_params   = get_mfa_params(email, password)
 
     response = rubygems_api_request(:post, "api/v1/api_key",
                                     sign_in_host, scope: scope) do |request|
       request.basic_auth email, password
       request["OTP"] = otp if otp
-      request.body = URI.encode_www_form({ name: key_name }.merge(scope_params))
+      request.body = URI.encode_www_form({ name: key_name }.merge(scope_params, mfa_params))
     end
 
     with_response response do |resp|
@@ -265,6 +267,27 @@ module Gem::GemcutterUtilities
     end
 
     scope_params
+  end
+
+  def get_mfa_params(email, password)
+    mfa_level = get_user_mfa_level(email, password)
+    params = {}
+    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_sign"
+      selected = ask "Would you like to enable MFA for this key? [y/N]"
+      params["mfa"] = true if selected =~ /^[yY](es)?$/
+    elsif mfa_level == "ui_and_api"
+      params["mfa"] = true
+    end
+    params
+  end
+
+  def get_user_mfa_level(email, password)
+    response = rubygems_api_request(:get, "api/v1/profile") do |request|
+      request.basic_auth email, password
+    end
+    with_response response do |resp|
+      JSON.parse(resp.body)["mfa"]
+    end
   end
 
   def get_key_name(scope)
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 require_relative 'remote_fetcher'
 require_relative 'text'
-require 'json'
 
 ##
 # Utility methods for using the RubyGems API.
@@ -283,11 +282,12 @@ module Gem::GemcutterUtilities
   end
 
   def get_user_mfa_level(email, password)
-    response = rubygems_api_request(:get, "api/v1/profile/me") do |request|
+    response = rubygems_api_request(:get, "api/v1/profile/me.yaml") do |request|
       request.basic_auth email, password
     end
     with_response response do |resp|
-      JSON.parse(resp.body)["mfa"]
+      body = Gem::SafeYAML.load clean_text(resp.body)
+      body["mfa"]
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,27 @@ def get_scope_params(scope)
...
   end
 
+  def get_mfa_params(email, password)
+    mfa_level = get_user_mfa_level(email, password)
+    params = {}
+    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_sign"
+      selected = ask "Would you like to enable MFA for this key? [y/N]"
+      params["mfa"] = true if selected =~ /^[yY](es)?$/
+    elsif mfa_level == "ui_and_api"
+      params["mfa"] = true
```
jenshenny, line 279: > If the user has an mfa level of `ui_and_api`, turns mfa on for the key without asking

I forgot to update the UI PR - the behaviour on the UI was changed so that we don't explicitly enable MFA if a user has `ui_and_api` ([discussion](https://github.com/rubygems/rubygems.org/pull/2846#discussion_r757738763))

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,27 @@ def get_scope_params(scope)
     scope_params
   end
 
+  def get_mfa_params(email, password)
+    mfa_level = get_user_mfa_level(email, password)
+    params = {}
+    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_sign"
```
jenshenny, line 275:
```suggestion
    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_signin"
```
I think it's [`ui_and_gem_signin`](https://github.com/rubygems/rubygems.org/blob/master/app/models/user.rb#L61)

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,27 @@ def get_scope_params(scope)
...
+      selected = ask "Would you like to enable MFA for this key? [y/N]"
```
jenshenny, line 276: To have consistency with the UI (and to increase the use of MFA), maybe we can have the default as Y instead of N.

lib/rubygems/gemcutter_utilities.rb
aellispierce: Doh! Updated and adjusted the test so a future mistake like this will cause a failure

lib/rubygems/gemcutter_utilities.rb
aellispierce: Ah gotcha! Updated

lib/rubygems/gemcutter_utilities.rb
aellispierce: Fantastic idea. Updated in https://github.com/rubygems/rubygems/pull/5305/commits/0b636f690275d9e4a1072e9a7b88ec18bb83d9dd

aellispierce:
Now that https://github.com/rubygems/rubygems.org/pull/2922 is merged, this is ready for review

jenshenny:
cc: @sonalkr132

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +270,25 @@ def get_scope_params(scope)
...
+    params = {}
+    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_signin"
+      selected = ask "Would you like to enable MFA for this key? [Y/n]"
+      params["mfa"] = true unless selected =~ /^[nN](o)?$/
+    end
+    params
+  end
+
+  def get_user_mfa_level(email, password)
+    response = rubygems_api_request(:get, "api/v1/profile/me") do |request|
```
simi, line 284: Any idea what's going to happen when endpoint fails (for example when is not implemented at other than rubygems.org gemservers)?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require_relative 'remote_fetcher'
 require_relative 'text'
+require 'json'
```
simi, line 4: I'm not sure it is possible to rely on `json` presence. It is already required at S3 signer. But for rubygems.org response there is already `YAML` endpoint used. I think we can try to keep it the same here.

https://github.com/rubygems/rubygems/blob/bec5fdf524c0ccd3d3e561db11455e4844049497/lib/rubygems/commands/owner_command.rb#L64-L79

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,28 @@ def get_scope_params(scope)
     scope_params
   end
 
+  def get_mfa_params(email, password)
+    return {} unless self.host == Gem::DEFAULT_HOST
```
simi, line 273: Including this new code, this check is done already three times in this class. What about to refactor it into separated private method `default_host?`?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,28 @@ def get_scope_params(scope)
...
+
+    mfa_level = get_user_mfa_level(email, password)
+    params = {}
+    if mfa_level == "ui_only" || mfa_level == "ui_and_gem_signin"
+      selected = ask "Would you like to enable MFA for this key? [Y/n]"
```
simi, line 278: Asking for `Y/n` value is happening already in `get_scope_params`. Do you think it would be possible to refactor this logic into "helper" private method within this class (`ask_yes_no`)?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -267,6 +269,28 @@ def get_scope_params(scope)
...
+      params["mfa"] = true unless selected =~ /^[nN](o)?$/
+    end
+    params
+  end
+
+  def get_user_mfa_level(email, password)
+    response = rubygems_api_request(:get, "api/v1/profile/me.yaml") do |request|
+      request.basic_auth email, password
+    end
```
simi, line 287: :thinking: I think newline would fit in here to separate those two blocks.

simi (commented):
I have left my final concerns regarding code style and possible improvements.

Regarding limitation to rubygems.org only for now, I'm not sure this is something we should do in the future. But for initial round I'm ok with this limitation. We can later just check interest of open-source gem hosts (like geminabox and gemstash) to implement related endpoint.

lib/rubygems/gemcutter_utilities.rb
aellispierce: In `get_scope_params` we're asking for `N/y` (as in No is the default) but in this method we're asking `Y/n` (as in Yes is the default). So they're slightly different. I didn't see a super clean way of merging the two into one method without having conditionals that made it more confusing than being separate

aellispierce:
Hmm I think the failing test might be a network issue. I see errors in there like `Bundler::HTTPError: Could not reach host index.rubygems.org. Check your network connection and try again.` and `SocketError: Failed to open TCP connection to index.rubygems.org:443`. I think re-running might solve those but i don't have that power.

simi:
@aellispierce restart has fixed the CI indeed.

lib/rubygems/gemcutter_utilities.rb
simi: There is actually `ask_yes_no` method in `user_interaction.rb`. Can we use this one to keep it simple? It solves also another problems like fail hard on non-TTY frontend. I suggest to use that one and provide no default. That way it will fail as expected when user input is not available. Feel free to update the message to make it clear this is highly recommended to enable.

We should update the other (previously added) ask for input in this class as well to use the `ask_yes_no` method.

aellispierce:
Thank you @simi !

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -258,15 +260,42 @@ def get_scope_params(scope)
     else
       say "Please select scopes you want to enable for the API key (y/n)"
       API_SCOPES.each do |scope|
-        selected = ask "#{scope} [y/N]: "
-        scope_params[scope] = true if selected =~ /^[yY](es)?$/
+        selected = ask_yes_no("#{scope}", false)
```
simi, line 263: Is there any particular reason to use `false` as a default? If I understand it well on non-TTY environments it will fail anyway (since the other `ask_yes_no`). We can provide on default and update the original message to highly recommend to reduce the scope of the key to the minimum according to the API key purpose. Maybe we can also list all scopes in this message to make it clear which scopes are you going to be asked like `Please select scopes (scope1, scope2,)` and I think where is no need for the `(y/n)` part in the message.

aellispierce: @simi Providing a default makes it so the user can simply hit "enter" instead of actually choosing yes or no. Otherwise, if you don't provide a response it will ask you the question again. Since there are so many options, I find I can get through the sign on process much more quickly with having a default. (one key press for enter instead of two, one for `n` and then enter)

> Maybe we can also list all scopes in this message to make it clear which scopes are you going to be asked like Please select scopes (scope1, scope2,)
There are seven scopes we ask about, so I feel it may be a bit overwhelming to list them all at once.

Either way, the scopes section here is not a new addition, this is the current behavior. So if we'd like to change it significantly I would suggest that be in a separate PR to give it the attention it deserves

simi: OK, seems reasonable.

simi (approved):

deivid-rodriguez:
I will review this in the next two weeks 👍.

deivid-rodriguez (approved):
Looks good to me!

PR #5321, opened 2022-01-29
Remove encoding pragma from specification.rb

simi:
- it is not used since it is not at the top of the file
- it is not useful anymore

requested by @nobu at https://github.com/rubygems/rubygems/commit/0a0663ff6a341c62edbaf5c3143024f3bef04076#r65077065

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-# -*- coding: utf-8 -*-
+#
 #--
 # Copyright 2006 by Chad Fowler, Rich Kilmer, Jim Weirich and others.
 # All rights reserved.


simi:
:thinking: CI fail seems unrelated to this change. I tried to revert the change, push and the same failure was still present.

deivid-rodriguez:
Yes, it seems some change in rspec-expectations 3.10.2 that broke our build. While our regular specs respect our lockfiles (locked to rspec-expectations 3.10.1), the ruby-core run does not. I'll try to fix ruby-core so that this run is not affected by third party releases either.

deivid-rodriguez:
Actually the culprit seems `rspec-mock-3.10.3`, I'm trying to lock our dependencies also when running the `ruby-core` CI job at #5324 to not run into something like this again in the future.

deivid-rodriguez (approved):
Can you rebase this @simi?

simi:
Done @deivid-rodriguez.

PR #5322, opened 2022-01-31
Expect to receive a non-keywords hash

nobu:
`RSpec::Mocks::ArgumentListMatcher#args_match?` fails when a non-keywords hash is passed while a keyword hash is expected.

Since a few days ago, `test-bundler` is failing on GitHub Actions.
I'm not sure why this didn't happened previously and what triggered it.

Make a non-keyword hash an explicit hash.

### DIFF
--- a/bundler/spec/bundler/rubygems_integration_spec.rb
+++ b/bundler/spec/bundler/rubygems_integration_spec.rb
@@ -47,7 +47,7 @@ RSpec.describe Bundler::RubygemsIntegration do
 
     it "successfully downloads gem with retries" do
       expect(Bundler.rubygems).to receive(:gem_remote_fetcher).and_return(fetcher)
-      expect(fetcher).to receive(:headers=).with("X-Gemfile-Source" => "https://foo.bar")
+      expect(fetcher).to receive(:headers=).with({ "X-Gemfile-Source" => "https://foo.bar" })
       expect(Bundler::Retry).to receive(:new).with("download gem from #{uri}/").
         and_return(bundler_retry)
       expect(bundler_retry).to receive(:attempts).and_yield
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Hei @nobu, thanks for this. The culprit seems a change in `rspec-mocks-3.10.3`, the upstream issue should be https://github.com/rspec/rspec-mocks/issues/1457.

I think it's unclear whether this will be fixed in the RSpec side or not, so before adapting our code I'm working on the alternative fix of making sure that our dependencies are locked also when running specs from ruby-core. Our regular specs were not affected by this because we are currently locked to `rspec-mocks-3.10.2`, but ruby-core does not consider our lockfiles and uses the latest version instead. I would like to change this to make things less brittle if that's ok with you.

simi:
There is another spec failing as well which can be fixed the same way.

```
Failures:

  1) Bundler::RubygemsIntegration#fetch_all_remote_specs when a rubygems source mirror is set sets the 'X-Gemfile-Source' header containing the original source
     Failure/Error: fetcher.headers = { "X-Gemfile-Source" => remote.original_uri.to_s } if remote.original_uri

       #<Double "gem_remote_fetcher"> received :headers= with unexpected arguments
         expected: ({"X-Gemfile-Source"=>"http://zombo.com"})
              got: ({"X-Gemfile-Source"=>"http://zombo.com"})
     # ./spec/bundler/bundler/rubygems_integration_spec.rb:82:in `block (4 levels) in <top (required)>'
     # ./spec/bundler/spec_helper.rb:101:in `block (4 levels) in <top (required)>'
     # ./spec/bundler/spec_helper.rb:101:in `block (3 levels) in <top (required)>'
     # ./spec/bundler/support/helpers.rb:352:in `block in with_gem_path_as'
     # ./spec/bundler/support/helpers.rb:366:in `without_env_side_effects'
     # ./spec/bundler/support/helpers.rb:347:in `with_gem_path_as'
     # ./spec/bundler/spec_helper.rb:100:in `block (2 levels) in <top (required)>'
```

IMHO the problem is in unclear message of problem

```
         expected: ({"X-Gemfile-Source"=>"http://zombo.com"})
              got: ({"X-Gemfile-Source"=>"http://zombo.com"})
```

seems like it is the same, but it is not

I'll open an issue at rspec-mocks side to improve the message.

deivid-rodriguez:
Yeah, the message is definitely confusing 😆.

hsbt:
This issue was solved by https://github.com/ruby/ruby/pull/5512

simi:
IMHO it wasn't really solve by Ruby 5512. 5512 just made bundler not picking latest rspec, but fallback to locked gems with rspec-mocks:3.10.2. If we decide to upgrade to newer version, this will popup again.

https://github.com/rubygems/rubygems/blob/15006ea01a3d2fbaca57bf5a297dbb532fadc0bf/bundler/tool/bundler/dev_gems.rb.lock#L30

deivid-rodriguez:
That's correct, the issue will pop up again once we update to rspec-mocks 3.10.3. I think we should report it upstream (the closest report seems https://github.com/rspec/rspec-mocks/issues/1457) to figure out whether we need to address this ourselves, or whether they will fix this upstream.

simi:
:handshake:

PR #5327, opened 2022-02-01
If GEM_HOME isn't set AND default gem home isn't writable, set `Gem.paths.home` to `Gem.user_dir`.

### LINKED ISSUE #1394: Make ruby gem install to user-install by default

duckinator:

RubyGems should not do a system install by default, for the reasons discussed in #1394, #2847, and #4031.

This was originally based on #2847 by @hsbt.

Fixes #1394.
Replaces #2847.

Revise the logic for setting `Gem.paths.home` such that the priority is:

1. `GEM_HOME`, if specified.
2. `Gem.default_dir`, if it is writable OR it doesn't exist and its parent directory is writable.
3. `Gem.user_dir`, in all other cases.

This is what the output looks like now, if `GEM_HOME` is not defined AND `Gem.default_dir` can't be used:

```shellsession
puppy@orthrus$ gem install boop
Defaulting to user installation because default GEM_HOME (/usr/local/lib/ruby/gems/3.1) is not writable.
Successfully installed boop-0.5.0
1 gem installed
puppy@orthrus$ 
```

Previously this situation would've resulted in a permissions error.

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -324,7 +324,7 @@ class Gem::Command
       @when_invoked.call options
     else
       if Gem.paths.auto_user_install && !options[:install_dir] && !options[:user_install]
-        self.ui.say "Defaulting to user installation because default GEM_HOME (#{Gem.default_dir}) is not writable."
+        self.ui.say "Defaulting to user installation because default installation directory (#{Gem.default_dir}) is not writable."
       end
 
       execute
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -323,6 +323,10 @@ class Gem::Command
     elsif @when_invoked
       @when_invoked.call options
     else
+      if Gem.paths.auto_user_install && !options[:install_dir] && !options[:user_install]
+        self.ui.say "Defaulting to user installation because default GEM_HOME (#{Gem.default_dir}) is not writable."
+      end
+
       execute
     end
   ensure
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -19,6 +19,10 @@ class Gem::PathSupport
   attr_reader :spec_cache_dir # :nodoc:
 
   ##
+  # Whether or `Gem.paths.home` defaulted to a user install.
+  attr_reader :auto_user_install
+
+  ##
   #
   # Constructor. Takes a single argument which is to be treated like a
   # hashtable, or defaults to ENV, the system environment.
@@ -38,9 +42,8 @@ class Gem::PathSupport
       @home = normalize_home_dir(Gem.default_dir)
     else
       # If `GEM_HOME` is not set AND we can't use `Gem.default_dir`,
-      # default to a user installation and print a message about this.
-      puts "Defaulting to user installation because default GEM_HOME (#{Gem.default_dir}) is not writable."
-
+      # default to a user installation and set `@auto_user_install`.
+      @auto_user_install = true
       @home = normalize_home_dir(Gem.user_dir)
     end
 
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -24,19 +24,24 @@ class Gem::PathSupport
   # hashtable, or defaults to ENV, the system environment.
   #
   def initialize(env)
-    @home = default_home_dir(env)
+    # Current implementation of @home, which is exposed as `Gem.paths.home`:
+    # 1. If `env["GEM_HOME"]` is defined in the environment: `env["GEM_HOME"]`.
+    # 2. If `Gem.default_dir` is writable OR it does not exist and it's parent
+    #    directory is writable: `Gem.default_dir`.
+    # 3. Otherwise: `Gem.user_dir`.
 
-    # If @home (aka Gem.paths.home) exists, but we can't write to it,
-    # fall back to Gem.user_dir (the directory used for user installs).
-    if File.exist?(@home) && !File.writable?(@home)
-      warn "The default GEM_HOME (#{@home}) is not" \
-            " writable, so rubygems is falling back to installing" \
-            " under your home folder. To get rid of this warning" \
-            " permanently either fix your GEM_HOME folder permissions" \
-            " or add the following to your ~/.gemrc file:\n" \
-            "    gem: --install-dir #{Gem.user_dir}"
+    if env.key?("GEM_HOME")
+      @home = normalize_home_dir(env["GEM_HOME"])
+    elsif File.writable?(Gem.default_dir) || \
+          (!File.exist?(Gem.default_dir) && File.writable?(File.expand_path("..", Gem.default_dir)))
 
-      @home = Gem.user_dir
+      @home = normalize_home_dir(Gem.default_dir)
+    else
+      # If `GEM_HOME` is not set AND we can't use `Gem.default_dir`,
+      # default to a user installation and print a message about this.
+      puts "Defaulting to user installation because default GEM_HOME (#{Gem.default_dir}) is not writable."
+
+      @home = normalize_home_dir(Gem.user_dir)
     end
 
     @path = split_gem_path env["GEM_PATH"], @home
@@ -52,9 +57,7 @@ class Gem::PathSupport
   # The default home directory.
   # This function was broken out to accommodate tests in `bundler/spec/commands/doctor_spec.rb`.
 
-  def default_home_dir(env)
-    home = env["GEM_HOME"] || Gem.default_dir
-
+  def normalize_home_dir(home)
     if File::ALT_SEPARATOR
       home = home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
     end
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -32,6 +32,19 @@ class Gem::PathSupport
 
     @home = expand(@home)
 
+    # If @home (aka Gem.paths.home) exists, but we can't write to it,
+    # fall back to Gem.user_dir (the directory used for user installs).
+    if File.exist?(@home) && !File.writable?(@home)
+      warn "The default GEM_HOME (#{@home}) is not" \
+            " writable, so rubygems is falling back to installing" \
+            " under your home folder. To get rid of this warning" \
+            " permanently either fix your GEM_HOME folder permissions" \
+            " or add the following to your ~/.gemrc file:\n" \
+            "    gem: --install-dir #{Gem.user_dir}"
+
+      @home = Gem.user_dir
+    end
+
     @path = split_gem_path env["GEM_PATH"], @home
 
     @spec_cache_dir = env["GEM_SPEC_CACHE"] || Gem.default_spec_cache_dir
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -19,7 +19,7 @@ class Gem::PathSupport
   attr_reader :spec_cache_dir # :nodoc:
 
   ##
-  # Whether or `Gem.paths.home` defaulted to a user install.
+  # Whether `Gem.paths.home` defaulted to a user install or not.
   attr_reader :auto_user_install
 
   ##
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -24,13 +24,7 @@ class Gem::PathSupport
   # hashtable, or defaults to ENV, the system environment.
   #
   def initialize(env)
-    @home = env["GEM_HOME"] || Gem.default_dir
-
-    if File::ALT_SEPARATOR
-      @home = @home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
-    end
-
-    @home = expand(@home)
+    @home = default_home_dir(env)
 
     # If @home (aka Gem.paths.home) exists, but we can't write to it,
     # fall back to Gem.user_dir (the directory used for user installs).
@@ -53,6 +47,20 @@ class Gem::PathSupport
   end
 
   private
+
+  ##
+  # The default home directory.
+  # This function was broken out to accommodate tests in `bundler/spec/commands/doctor_spec.rb`.
+
+  def default_home_dir(env)
+    home = env["GEM_HOME"] || Gem.default_dir
+
+    if File::ALT_SEPARATOR
+      home = home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
+    end
+
+    expand(home)
+  end
 
   ##
   # Split the Gem search path (as reported by Gem.path).
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -56,10 +56,6 @@ class Gem::PathSupport
 
   private
 
-  ##
-  # The default home directory.
-  # This function was broken out to accommodate tests in `bundler/spec/commands/doctor_spec.rb`.
-
   def normalize_home_dir(home)
     if File::ALT_SEPARATOR
       home = home.gsub(File::ALT_SEPARATOR, File::SEPARATOR)
--- a/lib/rubygems/path_support.rb
+++ b/lib/rubygems/path_support.rb
@@ -30,15 +30,12 @@ class Gem::PathSupport
   def initialize(env)
     # Current implementation of @home, which is exposed as `Gem.paths.home`:
     # 1. If `env["GEM_HOME"]` is defined in the environment: `env["GEM_HOME"]`.
-    # 2. If `Gem.default_dir` is writable OR it does not exist and it's parent
-    #    directory is writable: `Gem.default_dir`.
+    # 2. If `Gem.default_dir` is writable: `Gem.default_dir`.
     # 3. Otherwise: `Gem.user_dir`.
 
     if env.key?("GEM_HOME")
       @home = normalize_home_dir(env["GEM_HOME"])
-    elsif File.writable?(Gem.default_dir) || \
-          (!File.exist?(Gem.default_dir) && File.writable?(File.expand_path("..", Gem.default_dir)))
-
+    elsif File.writable?(Gem.default_dir)
       @home = normalize_home_dir(Gem.default_dir)
     else
       # If `GEM_HOME` is not set AND we can't use `Gem.default_dir`,
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -159,9 +159,9 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
     FileUtils.chmod 0o755, @gemhome
   end
 
-  def test_auto_user_install_unless_gem_home_writable
+  def test_auto_install_dir_unless_gem_home_writable
     if Process.uid.zero?
-      pend("test_auto_user_install_unless_gem_home_writable test skipped in root privilege")
+      pend("test_auto_install_dir_unless_gem_home_writable test skipped in root privilege")
       return
     end
 
@@ -174,14 +174,14 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
 
     @cmd.handle_options %w[]
 
-    refute @cmd.options[:user_install]
+    assert_not_equal Gem.paths.home, Gem.user_dir
 
     FileUtils.chmod 0o755, @userhome
     FileUtils.chmod 0o000, @gemhome
 
     Gem.use_paths @gemhome, @userhome
 
-    @cmd.install_update_options.include?(:user_install)
+    assert_equal Gem.paths.home, Gem.user_dir
   ensure
     FileUtils.chmod 0o755, @gemhome
   end
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -151,7 +151,7 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
 
       Gem.use_paths @gemhome, @userhome
 
-      assert_raise(Gem::FilePermissionError, Errno::EACCES) do
+      assert_raise(Gem::FilePermissionError) do
         Gem::Installer.at(@gem, @cmd.options).install
       end
     end
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -176,14 +176,14 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
 
     refute @cmd.options[:user_install]
 
-    FileUtils.chmod 0755, @userhome
-    FileUtils.chmod 0000, @gemhome
+    FileUtils.chmod 0o755, @userhome
+    FileUtils.chmod 0o000, @gemhome
 
     Gem.use_paths @gemhome, @userhome
 
     @cmd.install_update_options.include?(:user_install)
   ensure
-    FileUtils.chmod 0755, @gemhome
+    FileUtils.chmod 0o755, @gemhome
   end
 
   def test_vendor
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -159,6 +159,33 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
     FileUtils.chmod 0o755, @gemhome
   end
 
+  def test_auto_user_install_unless_gem_home_writable
+    if Process.uid.zero?
+      pend("test_auto_user_install_unless_gem_home_writable test skipped in root privilege")
+      return
+    end
+
+    @spec = quick_gem "a" do |spec|
+      util_make_exec spec
+    end
+
+    util_build_gem @spec
+    @gem = @spec.cache_file
+
+    @cmd.handle_options %w[]
+
+    refute @cmd.options[:user_install]
+
+    FileUtils.chmod 0755, @userhome
+    FileUtils.chmod 0000, @gemhome
+
+    Gem.use_paths @gemhome, @userhome
+
+    @cmd.install_update_options.include?(:user_install)
+  ensure
+    FileUtils.chmod 0755, @gemhome
+  end
+
   def test_vendor
     vendordir(File.join(@tempdir, "vendor")) do
       @cmd.handle_options %w[--vendor]
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -165,6 +165,9 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
       return
     end
 
+    orig_gem_home = ENV["GEM_HOME"]
+    ENV.delete("GEM_HOME")
+
     @spec = quick_gem "a" do |spec|
       util_make_exec spec
     end
@@ -179,11 +182,12 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
     FileUtils.chmod 0o755, @userhome
     FileUtils.chmod 0o000, @gemhome
 
-    Gem.use_paths @gemhome, @userhome
+    Gem.use_paths nil, @userhome
 
     assert_equal Gem.paths.home, Gem.user_dir
   ensure
     FileUtils.chmod 0o755, @gemhome
+    ENV["GEM_HOME"] = orig_gem_home if orig_gem_home
   end
 
   def test_vendor
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -151,7 +151,7 @@ class TestGemInstallUpdateOptions < Gem::InstallerTestCase
 
       Gem.use_paths @gemhome, @userhome
 
-      assert_raise(Gem::FilePermissionError) do
+      assert_raise(Gem::FilePermissionError, Errno::EACCES) do
         Gem::Installer.at(@gem, @cmd.options).install
       end
     end
--- a/bundler/spec/commands/doctor_spec.rb
+++ b/bundler/spec/commands/doctor_spec.rb
@@ -40,8 +40,9 @@ RSpec.describe "bundle doctor" do
       allow(File).to receive(:readable?).with(unwritable_file) { true }
 
       # The following 2 lines are for `Gem::PathSupport#initialize`.
-      allow(File).to receive(:exist?).with(Gem.paths.send(:default_home_dir))
-      allow(File).to receive(:writable?).with(Gem.paths.send(:default_home_dir))
+      allow(File).to receive(:exist?).with(Gem.default_dir)
+      allow(File).to receive(:writable?).with(Gem.default_dir)
+      allow(File).to receive(:writable?).with(File.expand_path("..", Gem.default_dir))
     end
 
     it "exits with no message if the installed gem has no C extensions" do
--- a/bundler/spec/commands/doctor_spec.rb
+++ b/bundler/spec/commands/doctor_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe "bundle doctor" do
       allow(File).to receive(:writable?).with(unwritable_file) { true }
       allow(File).to receive(:readable?).with(unwritable_file) { true }
 
-      # The following 2 lines are for `Gem::PathSupport#initialize`.
+      # The following lines are for `Gem::PathSupport#initialize`.
       allow(File).to receive(:exist?).with(Gem.default_dir)
       allow(File).to receive(:writable?).with(Gem.default_dir)
       allow(File).to receive(:writable?).with(File.expand_path("..", Gem.default_dir))
--- a/bundler/spec/commands/doctor_spec.rb
+++ b/bundler/spec/commands/doctor_spec.rb
@@ -38,6 +38,10 @@ RSpec.describe "bundle doctor" do
       allow(stat).to receive(:uid) { Process.uid }
       allow(File).to receive(:writable?).with(unwritable_file) { true }
       allow(File).to receive(:readable?).with(unwritable_file) { true }
+
+      # The following 2 lines are for `Gem::PathSupport#initialize`.
+      allow(File).to receive(:exist?).with(Gem.paths.send(:default_home_dir))
+      allow(File).to receive(:writable?).with(Gem.paths.send(:default_home_dir))
     end
 
     it "exits with no message if the installed gem has no C extensions" do
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -185,9 +185,24 @@ module Gem::InstallUpdateOptions
   # Default options for the gem install and update commands.
 
   def install_update_options
-    {
+    default_options = {
       :document => %w[ri],
     }
+
+    # If Gem.paths.home exists, but we can't write to it,
+    # fall back to a user installation.
+    if File.exist?(Gem.paths.home) && !File.writable?(Gem.paths.home)
+      default_options[:user_install] = true
+
+      alert_warning "The default GEM_HOME (#{Gem.paths.home}) is not" \
+                    " writable, so rubygems is falling back to installing" \
+                    " under your home folder. To get rid of this warning" \
+                    " permanently either fix your GEM_HOME folder permissions" \
+                    " or add the following to your ~/.gemrc file:\n" \
+                    "    gem: --user-install"
+    end
+
+    default_options
   end
 
   ##
--- a/lib/rubygems/install_update_options.rb
+++ b/lib/rubygems/install_update_options.rb
@@ -185,24 +185,9 @@ module Gem::InstallUpdateOptions
   # Default options for the gem install and update commands.
 
   def install_update_options
-    default_options = {
+    {
       :document => %w[ri],
     }
-
-    # If Gem.paths.home exists, but we can't write to it,
-    # fall back to a user installation.
-    if File.exist?(Gem.paths.home) && !File.writable?(Gem.paths.home)
-      default_options[:user_install] = true
-
-      alert_warning "The default GEM_HOME (#{Gem.paths.home}) is not" \
-                    " writable, so rubygems is falling back to installing" \
-                    " under your home folder. To get rid of this warning" \
-                    " permanently either fix your GEM_HOME folder permissions" \
-                    " or add the following to your ~/.gemrc file:\n" \
-                    "    gem: --user-install"
-    end
-
-    default_options
   end
 
   ##
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -196,9 +196,10 @@ class Gem::Installer
     end
 
     if @gem_home == Gem.user_dir
-      # If we get here, then:
-      # 1. `--user-install` was specified, or
-      # 2. GEM_HOME was not writable, and `Gem::PathSupport` fell back to `Gem.user_dir`.
+      # If we get here, then one of the following likely happened:
+      # - `--user-install` was specified
+      # - `Gem::PathSupport#home` fell back to `Gem.user_dir`
+      # - GEM_HOME was manually set to `Gem.user_dir`
 
       check_that_user_bin_dir_is_in_path
     end
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -193,6 +193,13 @@ class Gem::Installer
       @gem_home = Gem.user_dir
       @bin_dir = Gem.bindir gem_home unless options[:bin_dir]
       @plugins_dir = Gem.plugindir(gem_home)
+    end
+
+    if @gem_home == Gem.user_dir
+      # If we get here, then:
+      # 1. `--user-install` was specified, or
+      # 2. GEM_HOME was not writable, and `Gem::PathSupport` fell back to `Gem.user_dir`.
+
       check_that_user_bin_dir_is_in_path
     end
   end


duckinator:
All of the test failures appear to be a Bundler test which can't find the `bundler` executable, e.g.:

```
 8) [TEST_ENV_NUMBER=1] when a gem has no architecture still installs correctly
     Failure/Error:
               raise <<~ERROR
       
                 Invoking `#{cmd}` failed with output:
                 #{command_execution.stdboth}
               ERROR

     RuntimeError:


       Commands:
       $ gem build rcov
       Successfully built RubyGem
         Name: rcov
         Version: 1.0
         File: rcov-1.0-mswin32.gem
       # $? => 0

       $ gem generate_index
       Generating Marshal quick index gemspecs for 48 gems
       ................................................
       Complete
       Generated Marshal quick index gemspecs: 0.012s
       Generating specs index
       Generated specs index: 0.001s
       Generating latest specs index
       Generated latest specs index: 0.000s
       Generating prerelease specs index
       Generated prerelease specs index: 0.000s
       Compressing indices
       Compressed indices: 0.000s
       # $? => 0

       Invoking `/opt/hostedtoolcache/Ruby/2.3.8/x64/bin/ruby -I/home/runner/work/rubygems/rubygems/bundler/spec -r/home/runner/work/rubygems/rubygems/bundler/spec/support/artifice/windows.rb -r/home/runner/work/rubygems/rubygems/bundler/spec/support/hax.rb /home/runner/work/rubygems/rubygems/bundler/tmp/1/gems/system/bin/bundle install` failed with output:
       /opt/hostedtoolcache/Ruby/2.3.8/x64/bin/ruby: No such file or directory -- /home/runner/work/rubygems/rubygems/bundler/tmp/1/gems/system/bin/bundle (LoadError)
     # ./spec/support/helpers.rb:207:in `sys_exec'
     # ./spec/support/helpers.rb:124:in `bundle'
     # ./spec/install/gemfile/platform_spec.rb:528:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:101:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:101:in `block (3 levels) in <top (required)>'
     # ./spec/support/helpers.rb:352:in `block in with_gem_path_as'
     # ./spec/support/helpers.rb:366:in `without_env_side_effects'
     # ./spec/support/helpers.rb:347:in `with_gem_path_as'
     # ./spec/spec_helper.rb:100:in `block (2 levels) in <top (required)>'
     # ./spec/support/rubygems_ext.rb:97:in `load'
     # ./spec/support/rubygems_ext.rb:97:in `gem_load_and_activate'
```

Not entirely sure how to fix that. Gonna call it a day since it's ~1am here.

ekohl:
Slight nit: the PR title and commit message use `readable` while the code checks for `writeable`.

duckinator:
@ekohl good catch, thank you!

duckinator:
Rebased off master (https://github.com/rubygems/rubygems/commit/427c047e701eabd164f2ce8195feb4ae6f2c802d).

duckinator:
`ruby bin/rspec ./spec/cache/platform_spec.rb` will reproduce one of the failures and runs pretty quick.

Fails due to not finding `bundle` with this PR, passes on master.

Using that + some other info I've figured out to try to compare the paths with/without this change.

duckinator:
I'm not making any real progress on this, because I can't reproduce the CI-related problems locally.

simi:
@duckinator have you tried https://github.com/mxschmitt/action-tmate (better to do that in private fork) for remote debug?

duckinator:
@simi i have not. i'll look into it the next time i'm working on this. :+1:

deivid-rodriguez:
@duckinator I had a look at this.

The problem is that you are adding the check too early, and at that point the `GEM_HOME` might not even exist yet, even if it's a writable location. For example, when doing `gem install foo --install-dir local-non-existing-path`. This currently creates the local-non-existing-path folder and installs `foo` there. With this patch, it will ignore the `--install-dir` argument and install to the default user location instead, since `File.writable?("local-non-existing-path")` is `false`.

I see two approaches:

* Continue with you approach, but check whether the gem home _exists_ and it's not writable.
* Try to install, rescue permission errors, and retry in the default user location in that case. I think this is the error than we would need to react to: https://github.com/rubygems/rubygems/blob/125415593ead9ab69a9f0bb5392c9d7ec61b1f51/lib/rubygems/installer.rb#L718-L722

duckinator:
@deivid-rodriguez thank you for digging into this. I think I'll try the first of your approaches and see how it goes. (I think a slightly-more-complex conditional is far cleaner than a rescue.)

deivid-rodriguez:
Ok now CI is green :)

Does this work when an explicit, and already existing, `--install-dir` is passed? The expected outcome would be to NOT fallback to `--user-install`.

duckinator:
> Does this work when an explicit, and already existing, `--install-dir` is passed? The expected outcome would be to NOT fallback to `--user-install`.

Will check this next time I'm working on RubyGems/Bundler stuff. I'm thinking I should probably actually add a test for that.

duckinator:
Rebased off master, looking into `--install-dir` not falling back to `--user-install`.

iuriguilherme:
FYI, in my environment, GEM_HOME is an empty string. With that, bundler will attempt to use system path (root owned), even though user-level gems folder exists (because `gem install --user-install` was used at least once).

This is the case until I force the folder I want for the current user with `bundle config set --global ~/.local/share/gem`

duckinator:
Rebased off master again.

duckinator:
@iuriguilherme this PR is only touching RubyGems (not Bundler), so would affect e.g. `gem install boop` but not `bundle install`.

Having Bundler respect `--user-install` being set for RubyGems has been brought up a few times, and I split it out into its own issue at #5682.

deivid-rodriguez:
@duckinator Just to clarify, this PR is still a WIP, right?

duckinator:
@deivid-rodriguez yeah, still need to determine the answer to this thing you asked before (and probably write a test for it):

> Does this work when an explicit, and already existing, --install-dir is passed? The expected outcome would be to NOT fallback to --user-install.

Everything else should be good.

duckinator:
I updated the title to be more explicit about it being a WIP. (Annoyingly, I can't find a way to convert an existing PR to a draft on GitHub.)

duckinator:
(I find it odd that the "Convert to draft" button is under the _Reviewers_ section, but oh well. It's a lot clearer that it's a WIP now.)

deivid-rodriguez:
> @deivid-rodriguez yeah, still need to determine the answer to this thing you asked before (and probably write a test for it):

Alright, that's what I suspected, thanks for clarifying!

simi:
> Not mentioning that it is hard to predict what would be the behavior with [configuration](https://src.fedoraproject.org/rpms/ruby/blob/3f106c188ea6368c950ba7d610594f910773a25f/f/operating_system.rb) we are using on Fedora. I can't see any test case which would try to cover all the changes which are possible to do via operating_system.rb.

I think (since it is really free playground) you need to test your changes in `operating_system.rb` on your own. There is no way to test all combinations, since you can do anything in that file. :thinking:

simi:
Also @voxik we can make this opt-out if you would like to disable it in `operating_system.rb`, since I understand this is not expected from system `ruby`.

voxik:
> > Not mentioning that it is hard to predict what would be the behavior with [configuration](https://src.fedoraproject.org/rpms/ruby/blob/3f106c188ea6368c950ba7d610594f910773a25f/f/operating_system.rb) we are using on Fedora. I can't see any test case which would try to cover all the changes which are possible to do via operating_system.rb.
> 
> I think (since it is really free playground) you need to test your changes in `operating_system.rb` on your own. There is no way to test all combinations, since you can do anything in that file. thinking

Well, I have seen changes, where operating_system.rb was completely forgotten. I understand, that the changes allowed in `operating_system.rb` might be complex, but acknowledgement of its existence and best effort testing would be certainly welcome.

> Also @voxik we can make this opt-out if you would like to disable it in operating_system.rb, since I understand this is not expected from system ruby.

My concern is general, not Fedora specific. There should be one install location, which does not change.

Take `make install`, which is standard way to install user compiled content, as an example. If you call `make install` as regular user, it will probably fail. It is very likely needed to use `sudo` explicitly or specify `DESTDIR`. It does not try to be clever and figure out some alternatives. And that is completely fine IMO.

eregon:
@voxik There has been extensive discussion about it. If you really want to reopen that even though it has been decided then you will have to read all of it and argue its conclusions. Basically, `sudo gem install` is almost always a mistake. So we want something better when we can't write to the default gem home. `--user-install` is an already existing functionality and it makes sense to reuse it for that. It already warns about potential pitfalls (e.g. the gem home bin/ not being in PATH). Manually specifying the GEM_HOME is a recipe to cause problems, I would never recommend that to anyone.

Also I think this PR is not the place to discuss that (forks the discussion in yet another place).

voxik:
@eregon I don't disagree. I am just saying that the default should be changed and there should not be this "fallback" magic. It won't help to anybody. It just going to extend the pain.

IOW, if we really want to make things simpler, then the PR should remove multiple lines of code instead of adding more.

> Also I think this PR is not the place to discuss that (forks the discussion in yet another place).

I agree even with this, but there is not right place if this very PR starts with references to 6 tickets.

duckinator:
@voxik my understanding is that you're suggesting to just default to `--user-install` all the time. Is that correct?

It looks like [Fedora sets up `operating_system.rb` so root installs globally but `--user-install` is the default for everyone else](https://src.fedoraproject.org/rpms/ruby/blob/3f106c188ea6368c950ba7d610594f910773a25f/f/operating_system.rb#_95), and it's [been being used in the wild for at least 4 years](https://src.fedoraproject.org/rpms/ruby/c/c2db65dbc7245bf1e78266ac2c2b1c7e7718dd15).

The main reason for making it conditional on if `GEM_HOME` is not writable is so end-users can customize the install location. I do agree that having an explicit default would be a lot less confusing. If there is an equivalent to `GEM_HOME` for user installs, always defaulting to `--user-install` and suggesting people use _that_ variable instead might be worth looking into.

My brain's kinda fuzzy right now, so I'm not comfortable offering a concrete suggestion atm. But I see merit in the points you're raising and think it's worth trying to figure out if there's a way to solve this problem while _simplifying_ RubyGems instead of _complicating_ it.

duckinator:
@voxik I'm calling it a day for RubyGems things, and have other work to focus on for at least the rest of this week (and possibly also most of next week), but you made a good point regarding this complicating things when we should be simplifying them and it's making me reconsider this approach.

voxik:
> @voxik my understanding is that you're suggesting to just default to `--user-install` all the time. Is that correct?

Right

> It looks like [Fedora sets up `operating_system.rb` so root installs globally but `--user-install` is the default for everyone else](https://src.fedoraproject.org/rpms/ruby/blob/3f106c188ea6368c950ba7d610594f910773a25f/f/operating_system.rb#_95), and it's [been being used in the wild for at least 4 years](https://src.fedoraproject.org/rpms/ruby/c/c2db65dbc7245bf1e78266ac2c2b1c7e7718dd15).

`--user-install` is used for 4 years. But for past 10 years, the user home directory was used for gem installation, just the methods were different:

https://src.fedoraproject.org/rpms/ruby/c/117278abd02a5bf1a482753a93fbe47c52752f6f?branch=rawhide

All the methods has it own quirks, but the `--user-install` is the most standardized way.

BTW I would not be opposed, if Fedora changed the way gems are installed for root user. The current status is mostly *keep doing what Ruby does*. After all, I think (more the 10 years ago :wink:) that `/usr` (of course there is exception for `/usr/local`) on Fedora should be managed via Fedora means and Ruby(Gems) could do better.

The thing is that RubyGems nicely supports multiple gem directories, but unfortunately on top of this there is a lot of exceptions. One is `--user-install`, but then there are exceptions for default gems, bundled gems and then there is Bundler. But I would treat those as an exception. Does Ruby (during build) need to install default gems? Fine, but use `--install-dir` for that purpose. Does Ruby (during build) need to install bundled gems? Again, use `--install-dir` + maybe some other adjustments. But don't put the exceptions into RubyGems.

> The main reason for making it conditional on if `GEM_HOME` is not writable is so end-users can customize the install location.

The current proposal suffers from the assumption, that the fallback location (user home) is writable. But what if it is not writable for whatever reason? Or what if the main directory is not writable, but the subdirectories are? Then there will be displayed two error messages. Somebody could say it is unlikely, but who knows.

> I do agree that having an explicit default would be a lot less confusing. If there is an equivalent to `GEM_HOME` for user installs, always defaulting to `--user-install` and suggesting people use _that_ variable instead might be worth looking into.
> 
> My brain's kinda fuzzy right now, so I'm not comfortable offering a concrete suggestion atm. But I see merit in the points you're raising and think it's worth trying to figure out if there's a way to solve this problem while _simplifying_ RubyGems instead of _complicating_ it.

🫶

eregon:
`--user-install` all the time is begging for gem sharing issues, as I've said multiple times on https://github.com/rubygems/rubygems/pull/2847#issuecomment-596082611 and later comments. The issue is `--user-install` cannot choose a proper location without risks of invalid gem home sharing based on the many variants a Ruby can have like configure flags (e.g. --enable-shared, -DRUBY_DEBUG, etc), architecture (e.g. https://github.com/postmodern/chruby/issues/422#issuecomment-1232791999), etc.

There are 3 cases:
* user is root, then it doesn't matter, anything works
* Ruby is installed in writable user directory, then using the default gem home is by far the safest and it's already the default (unless a Ruby switcher messes that up and has bugs due to that (chruby/RVM), thank you rbenv for doing the right thing). This is the most common case for Ruby developers.
* Ruby is installed in a directory not writable by the user. Then installing gems can either be installed globally with sudo (basically always a mistake) or installed a user-writable location (I'm considering `gem install` and not Bundler here). The home of a user is very much expected to be writable by that user, anything else is nonsense and deserves fixing. Hence `--user-install` is good and useful for that case. It's still not as safe as using the default gem home but that's not a possibility in that setup.

duckinator:
Rebased, and did a minor refactor of the test so it bails at the start instead of part-way through.

duckinator:
Last change should've fixed the linting error. (It was related to using single quotes when it wasn't necessary.)

duckinator:
The 3 problems that @voxik and @eregon mentioned which stand out to me are:

- this setup can be difficult to reason about, and it would be nice to _remove_ instead of _add_ complexity if possible
- the current system assumes the fallback directory is writable, and this may cause problems
- `--user-install` as it works now is prone to certain gem home sharing problems, and this makes using `--user-install` all the time risky

However, this has been being discussed for 7 years, so to me the best course of action seems to be:

1. Get this PR merged, likely waiting for the next major version bump to do so.
2. Independently of that, try to make `--user-install` avoid the problems @eregon mentioned to whatever extent is possible.
3. Once _that_ is done, we can revisit the idea of using `--user-install` by default all the time.

@voxik @eregon what're your thoughts on that^?

@deivid-rodriguez do you agree this would need a major version bump?



(As an aside: @voxik and @eregon, I really appreciate you two being so consistently patient when I disappear for multiple months mid-conversation due to health issues. I know it can be frustrating to be on the receiving end of that.)

MSP-Greg:
I use `--user-install`, as I overwrite my Windows installs daily.  See [Gem::Ext::ExtConfBuilder](https://github.com/ruby/ruby/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L46).

Am I correct that v4 changes would always place compiled code in platform specific folders?  Or, rather than compiled code in the `gems` folder, it would be in the `extensions` folder?  example path:
```
extensions/x64-mingw-ucrt/3.2.0+3/commonmarker-0.23.6
```

Paths like the above are platform specific.  @eregon ?

eregon:
I agree with merging this PR.

I think making `--user-install` the default all the time would be too incompatible and have very little gains, so probably there gain/risk ratio is very low.

duckinator:
Rebased off master (6b23fb294084e7f9cba815fe4426652b357ad0ee).

I'm going to mark this as ready for review. I think the best course of action is what I described in https://github.com/rubygems/rubygems/pull/5327#issuecomment-1333335423:

1. Get this PR merged.
2. Independently of that, try to make `--user-install` avoid the problems @eregon mentioned to whatever extent is possible.
3. If/when 2 is resolved, revisit the idea of using `--user-install` by default all the time.

duckinator:
> Does this work when an explicit, and already existing, `--install-dir` is passed? The expected outcome would be to NOT fallback to `--user-install`.

I finally remembered to check this. `--user-install` and `--install-dir` are mutually exclusive &mdash; you get an error saying `Use --install-dir or --user-install but not both` if you specify them both.

However, this PR as-is adds `--user-install` to the result of `Gem::InstallUpdateOptions#install_update_options`, which means that you could never use `--install-dir`. I'll need to dig into this more to find a better implementation.

hsbt (approved):
I'm +1 to merge this.

duckinator:
I also just now discovered a problem with `--user-install`, which we REALLY should resolve before adopting `--user-install` more widely: `--user-install` is straight-up ignored by `gem uninstall`.

```
puppy@orthrus.fox:~/test$ gem --version
3.4.8
puppy@orthrus.fox:~/test$ gem install --user-install boop
Fetching boop-0.5.0.gem
WARNING:  You don't have /home/puppy/.gem/ruby/3.0/bin in your PATH,
          gem executables will not run.
Successfully installed boop-0.5.0
1 gem installed
puppy@orthrus.fox:~/test$ gem uninstall --user-install boop
ERROR:  While executing gem ... (Gem::InstallError)
    boop is not installed in GEM_HOME, try:
        gem uninstall -i /usr/home/puppy/.gem/ruby/3.0 boop
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/uninstaller.rb:124:in `uninstall'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:195:in `uninstall'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:181:in `uninstall_gem'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:175:in `block in uninstall_specific'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:167:in `each'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:167:in `uninstall_specific'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:133:in `execute'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/command.rb:327:in `invoke_with_build_args'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/command_manager.rb:252:in `invoke_command'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/command_manager.rb:192:in `process_args'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/command_manager.rb:150:in `run'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/gem_runner.rb:51:in `run'
        /usr/local/bin/gem:10:in `<main>'
puppy@orthrus.fox:~/test$
```

Given that the suggested fix is `-i` (aka `--install-dir`), that makes the `--user-install`/`--install-dir` conflict even more important to resolve before this PR gets merged.

duckinator:
Converted this back to a draft since that `gem uninstall` problem _has_ to be fixed before this can be viable.

duckinator:
Opened #6457 for the `gem uninstall --user-install GEM_NAME` problem. This is blocked by that issue.

voxik:
My opinion on the PR has not changed. I don't thing that fallback is right solution.

And the more I think about this the more I believe that RubyGems should get to the basics. And that is to ensure that RubyGems properly supports multiple gem locations, AKA `GEM_PATH`, and having one of the paths, AKA `GEM_HOME`, writable for all changes. If the change was not applicable for `GEM_HOME`, RubyGems should fail.

IOW I don't perceive the `--user-install` doing anything else then changing `GEM_HOME`.

voxik:
BTW in the context of #6457, I need to clarify that the `GEM_HOME` is important mainly as a installation target. Because it should not be big deal for RubyGems to uninstall a gem from whatever `GEM_PATH` as long as it is writable.

duckinator:
@voxik yeah, at this point I'm thinking like you suggested might be the best option.

What're your thoughts on this approach?:

1. make `--user-install` control `GEM_HOME` instead of the weirdness it does now
2. optionally: in a subsequent minor release, fall back to `--user-install` if the GEM_HOME from the environment isn't writable
3. in a subsequent major release, make it the unconditional default

voxik:
> @voxik yeah, at this point I'm thinking like you suggested might be the best option.
> 
> What're your thoughts on this approach?:
> 
>     1. make `--user-install` control `GEM_HOME` instead of the weirdness it does now

That ^^ would be sweet.

BTW there is one distinction for `--user-install` and that is installation of binaries into `~/bin` (and that also similarly applies to system wide gems). But that IMHO should be handled as and extension to the basic use case.

duckinator:
Using ~/bin frankly has always felt wrong to me, as many people use that for custom scripts. That's screwed me over a few times.

I'm considering the idea of having --user-install be equivalent to setting GEM_HOME to a directory in ~/.local/rubygems and also put executables somewhere in there, or something like that. Not awake enough to be confident in that idea, and not tested it at all, though.

kenyon:
Maybe do the same things as `pip install --user`: https://pip.pypa.io/en/stable/user_guide/#user-installs

duckinator:
@kenyon that's actually a very interesting idea. It means user installs would be basically changing the prefix (when installed via the system package manager, this is `/usr` on most Linux distros and `/usr/local` on FreeBSD) to `~/.local`.

It might make sense to dig into this idea more. If we go this route, I'm leaning towards this being how it's implemented:

1. Add a RubyGems config option to specify the prefix.
2. Add a `--user` flag (for consistency with pip) that sets the prefix to `~/.local`.
3. Deprecate `--user-install`, printing a warning suggesting `--user` and explaining the differences between them.
4. Remove `--user-install`, changing the warning to an error.

I suspect this approach would also resolve longstanding problems with `--user-install` and compiled extensions because it'd use the same file layout as normal, just under `~/.local` instead of `/usr` or `/usr/local`. @MSP-Greg I think you were the one who brought these issues to my attention -- do you know if this addresses those problems?

@voxik @deivid-rodriguez what're your thoughts on this? It seems like a real candidate for fixing user installs while also simplifying the codebase, which I think is the ideal end result here.


<details>
<summary>side note</summary>

Interestingly, https://github.com/rubygems/rubygems/discussions/5878#discussioncomment-3481502 implies that user installs on Ubuntu already use `~/.local`:

> When I issued gem install --user-install bundler in a Ubuntu system, ~/.local/share/gem was created.

Not sure what to make of that.

</details>

kenyon:
@duckinator note that `gem install --user` already is equivalent to `gem install --user-install`, because you can use the shortest unambiguous abbreviation, so even `gem install --us` does the same thing.

indirect:
FYI, ~/.local/share is the default XDG_DATA_HOME. The [XDG basedir spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is yet another take on "how do we organize user-level data", distinct from GEM_HOME/GEM_PATH. It says that the location for user-local binaries must be ~/.local/bin, and distributions that support the XDG spec should add that directory to any user's PATH.

I think this PR is mergable as long as automatic `--user-install` is disabled by passing `--install-dir`. After that, I like the idea for `--user-install` to just set GEM_HOME if we can make that work.

duckinator:
Alright, I'll dig in to the user-install vs install-dir conflict today.

After that I'll work on an RFC about making user-install less magical and we can go from there.

duckinator:
As best I can tell, the code where this PR sets `--user-install` can't access the flags passed to `gem`. Which means it can't detect if `--install-dir` is used.

I determined that modifying `Gem::PathSupport#initialize` to overwrite `Gem.paths.home` directly works better, but I'm a lot less confident in this approach and would appreciate someone taking another look at it and trying it out. I'm also not sure if just using `warn` is appropriate there, or if I should be doing something else. (cc @deivid-rodriguez)

duckinator:
(The test has not been updated to match how it is implemented, so this is NOT ready to be merged.)

duckinator:
I'm not sure how or why, but the `test_user_install_disabled_read_only` test broke despite me not touching it.

And, oddly, the failure is that it raises an `Errno::EACCES` instead of the expected `Gem::FilePermissionError`.

```
<Gem::FilePermissionError> expected but was
<Errno::EACCES(<Permission denied @ rb_sysopen - /home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230929-2222-1e7hicb/gemhome/cache/a-2.gem>)
/home/runner/work/rubygems/rubygems/lib/rubygems/package/file_source.rb:30:in `initialize'
/home/runner/work/rubygems/rubygems/lib/rubygems/package/file_source.rb:30:in `open'
/home/runner/work/rubygems/rubygems/lib/rubygems/package/file_source.rb:30:in `with_read_io'
/home/runner/work/rubygems/rubygems/lib/rubygems/package.rb:616:in `verify'
/home/runner/work/rubygems/rubygems/lib/rubygems/package.rb:597:in `spec'
/home/runner/work/rubygems/rubygems/lib/rubygems/installer.rb:288:in `spec'
/home/runner/work/rubygems/rubygems/lib/rubygems/installer.rb:722:in `verify_spec'
/home/runner/work/rubygems/rubygems/lib/rubygems/installer.rb:912:in `pre_install_checks'
/home/runner/work/rubygems/rubygems/lib/rubygems/installer.rb:303:in `install'
/home/runner/work/rubygems/rubygems/test/rubygems/test_gem_install_update_options.rb:155:in `block in test_user_install_disabled_read_only'
```

Looking into that atm.

duckinator:
With _and_ without my change, `Gem.paths.home` is a `String`.

So I have no clue why it changed. However, I determined this should fix all the test failures:

```patch
diff --git a/test/rubygems/test_gem_install_update_options.rb b/test/rubygems/test_gem_install_update_options.rb
index 30ad370510..8c58520577 100644
--- a/test/rubygems/test_gem_install_update_options.rb
+++ b/test/rubygems/test_gem_install_update_options.rb
@@ -151,7 +151,7 @@ def test_user_install_disabled_read_only

       Gem.use_paths @gemhome, @userhome

-      assert_raise(Gem::FilePermissionError) do
+      assert_raise(Gem::FilePermissionError, Errno::EACCES) do
         Gem::Installer.at(@gem, @cmd.options).install
       end
     end
```

So my last commit (3c42edd) adds this change.

Assuming CI passes, this is ready for review. 👍

duckinator:
As a quick note, I made a discovery just now.

_If `$HOME/.gem` does not exist,_ then `--user-install` defaults to `#{Gem.data_home}/gem/ruby/<VERSION>` and the bin directory is simply `#{Gem.data_home}/gem/ruby/<VERSION/bin`.

This, combined with `--user` being shorthand for `--user-install` means that _if you delete `~/.gem`,_ then `--user` (aka `--user-install`) already does exactly what was discussed between @kenyon and myself.

This means the list of things I mentioned in [an earlier comment on March 26 2023](https://github.com/rubygems/rubygems/pull/5327#issuecomment-1484233728) can be simplified to:

1. Merge this PR once it's ready, making RubyGems set `Gem.paths.home` to `Gem.user_dir` (same as `--user-install` does _if `~/.gem` doesn't exist_).
2. Optionally, eventually (maybe with 4.0?) simplify `--user-install` by removing legacy support for `~/.gem` (and probably deprecate it for a while beforehand, if we do so).

duckinator:
There were some `bundle doctor` specs that this broke. I split out a helper function (`Gem::PathSupport#default_home_dir`), and made `bundle doctor` specs allow `File.exist?()` and `File.writable?()` to be called with that value as an argument, and it has been fixed.

In addition, there was a problem where `gem install` would not show the "You don't have ... in your PATH" warning when `Gem.paths.home` was automatically set to `Gem.user_dir` (as opposed to that being done because of `--user-install`). I added a commit that fixes this.

duckinator:
I realized my implementation was not allowing `GEM_HOME` to unconditionally override `Gem.paths.home` (and imo, if you explicitly set `GEM_HOME`, this code should not be touching it). It also was rather hard to understand the code anyway, so I used this as an opportunity to simplify the logic.

I revised it so that the way `Gem.paths.home` is set is now:

1. If `GEM_HOME` is defined in the environment: `env["GEM_HOME"]`.
2. If `Gem.default_dir` is writable OR it does not exist and it's parent directory is writable: `Gem.default_dir`.
3. Otherwise: `Gem.user_dir`.

Additionally, I realized that in practice the only reason anyone would have to explicitly add `--user-install` is _to get rid of the very long warning_. So I replaced it with a single-line note, based on what pip did. I also think it being a _warning_ was inappropriate, because nothing is necessarily _wrong_ &mdash; it's just useful information to know.

This is what the output looks like now, if `Gem.default_dir` can't be used AND `GEM_HOME` is not defined:

```shellsession
puppy@orthrus$ gem install boop
Defaulting to user installation because default GEM_HOME (/usr/local/lib/ruby/gems/3.1) is not writable.
Successfully installed boop-0.5.0
1 gem installed
puppy@orthrus$ 
```

duckinator:
I consolidated all the useful information I had into the original description. Then I hid comments that were outdated, resolved, or duplicated information from the description.

I'm hoping GitHub eventually hides all of those all under the "$N hidden items" thing, to get rid of that _5 pages_ of scrolling between the original description and this comment. 😂

duckinator:
I updated it so that the message is suppressed if you use `--install-dir` or `--user-install` (since those mean it has no effect).

<details>
<summary>Example output from gem install/update/uninstall commands, with or without `--install-dir`, as of f88cf80</summary>

```
$ gem install boop -v 0.4.0
Defaulting to user installation because default GEM_HOME (/usr/local/lib/ruby/gems/3.1) is not writable.
Fetching boop-0.4.0.gem
Successfully installed boop-0.4.0
1 gem installed
$ gem update boop
Defaulting to user installation because default GEM_HOME (/usr/local/lib/ruby/gems/3.1) is not writable.
Updating installed gems
Updating boop
Fetching boop-0.5.0.gem
Successfully installed boop-0.5.0
Gems updated: boop
$ gem uninstall boop
Defaulting to user installation because default GEM_HOME (/usr/local/lib/ruby/gems/3.1) is not writable.

Select gem to uninstall:
 1. boop-0.4.0
 2. boop-0.5.0
 3. All versions
> 3
Successfully uninstalled boop-0.4.0
Remove executables:
        boop

in addition to the gem? [Yn]
Removing boop
Successfully uninstalled boop-0.5.0
$ gem install --install-dir x boop
Fetching boop-0.5.0.gem
Successfully installed boop-0.5.0
1 gem installed
$ gem uninstall --install-dir x boop
Remove executables:
        boop

in addition to the gem? [Yn]
Removing boop
Successfully uninstalled boop-0.5.0
$
```

</details>

duckinator:
I think this is ready for a full review, finally! :3 

@deivid-rodriguez @hsbt could you two review this when you have the time? (A lot has changed since y'all last provided feedback on it.)

---

@voxik I believe my changes should have addressed the concerns you've raised. Basically: `--install-dir` and `--user-install` are highest priority, then `GEM_HOME`, and then the auto-determined directory. This means you can always use `--install-dir`, `--user-install`, and/or `GEM_HOME` to make it fully deterministic, same as before. I believe this PR, as it is now, actually brings us in line with how Fedora does it. 🙂

lib/rubygems/path_support.rb
```diff
@@ -41,6 +56,18 @@ def initialize(env)
 
   private
 
+  ##
+  # The default home directory.
+  # This function was broken out to accommodate tests in `bundler/spec/commands/doctor_spec.rb`.
```
deivid-rodriguez, line 61: I feel this is the kind of comment that gets easily out of date as code evolves and ends up being confusing. The extraction of this method feels neat anyways, so I'd say we don't need the comment. Also, I don't get the "The default home directory" part.

lib/rubygems/path_support.rb
```diff
@@ -18,20 +18,35 @@ class Gem::PathSupport
...
+    # Current implementation of @home, which is exposed as `Gem.paths.home`:
+    # 1. If `env["GEM_HOME"]` is defined in the environment: `env["GEM_HOME"]`.
+    # 2. If `Gem.default_dir` is writable OR it does not exist and it's parent
+    #    directory is writable: `Gem.default_dir`.
+    # 3. Otherwise: `Gem.user_dir`.
+
+    if env.key?("GEM_HOME")
+      @home = normalize_home_dir(env["GEM_HOME"])
+    elsif File.writable?(Gem.default_dir) || \
+          (!File.exist?(Gem.default_dir) && File.writable?(File.expand_path("..", Gem.default_dir)))
```
deivid-rodriguez, line 40: I think if you extract a helper method `does_not_exist_but_parent_is_writable(<directory>)`, the code become self explainable and you don't really need the comment above?

deivid-rodriguez: Also, `File.writable?` on directories seems to not do what one would expect:

```
$ mkdir foo
$ chmod -x foo
$ touch foo/bar
touch: cannot touch 'foo/bar': Permission denied
$ ruby -e 'puts File.writable?("foo")'
true
```

So, for directories, I think you need to check all permission bits: `File.writable? && File.readable? && File.executable?`.

deivid-rodriguez: BTW, no idea how this works on Windows.

test/rubygems/test_gem_install_update_options.rb
```diff
@@ -151,14 +151,45 @@ def test_user_install_disabled_read_only
 
       Gem.use_paths @gemhome, @userhome
 
-      assert_raise(Gem::FilePermissionError) do
+      assert_raise(Gem::FilePermissionError, Errno::EACCES) do
```
deivid-rodriguez, line 154: May no longer be needed after the latest changes?

lib/rubygems/path_support.rb
```diff
@@ -18,20 +18,35 @@ class Gem::PathSupport
   # Directory with spec cache
   attr_reader :spec_cache_dir # :nodoc:
 
+  ##
+  # Whether or `Gem.paths.home` defaulted to a user install.
```
deivid-rodriguez, line 22:
```suggestion
  # Whether `Gem.paths.home` defaulted to a user install or not.
```

lib/rubygems/command.rb
```diff
@@ -323,6 +323,10 @@ def invoke_with_build_args(args, build_args)
     elsif @when_invoked
       @when_invoked.call options
     else
+      if Gem.paths.auto_user_install && !options[:install_dir] && !options[:user_install]
+        self.ui.say "Defaulting to user installation because default GEM_HOME (#{Gem.default_dir}) is not writable."
```
deivid-rodriguez, line 327: Could this mislead users into thinking they have set `GEM_HOME` (the env variable)? Maybe

```suggestion
        self.ui.say "Defaulting to user installation because default installation directory (#{Gem.default_dir}) is not writable."
```
?

deivid-rodriguez (commented):
Add a few comments, but looks really good. Nice job!

lib/rubygems/path_support.rb
duckinator: Oh, right. idk about read permissions, but I do know that you need to be able to _write_ and _execute_ a directory to create it on Linux, and I believe most (if not all?) Unixes and Unix-likes.

Maybe this is why the original test for this was just `!File.exist?(Gem.default_dir)` instead of `!File.exist?(Gem.default_dir) && File.writable?(File.expand_path("..", Gem.default_dir))`.

I'll revisit this logic when I next work on this.

lib/rubygems/path_support.rb
duckinator: As if to prove your point: This comment is _already_ out of date. 😅 

Literally none of it is true anymore (including the "default home directory" part, which is why you don't get it).

test/rubygems/test_gem_install_update_options.rb
duckinator: I'll check if this is still needed. Still not sure what caused it to be needed, honestly.

lib/rubygems/path_support.rb
duckinator: Fixed in 6dc41e1.

test/rubygems/test_gem_install_update_options.rb
duckinator: Currently testing this now. Unfortunately, I can only reproduce it in CI, so it'll be about an hour before I have an update. 😅

duckinator: You were right, I was able to undo that change without issue. 👍

lib/rubygems/path_support.rb
duckinator: Looks like this is the only remaining problem, of the things you brought up.

deivid-rodriguez: Yes, there's also the refactoring I mentioned in the first comment in this review thread, but not a blocker at all!

hsbt (approved):
:+1: We can try this now.

duckinator:
Okay, so at this point it's just a matter of addressing the [`File.writable?` weirdness](https://github.com/rubygems/rubygems/pull/5327#discussion_r1341974304) that @deivid-rodriguez pointed out, and then we're good to go.

I'll try to get that done today!

duckinator:
I couldn't find a cross-platform way to reliably check _"Can I create this directory?"_ so I changed back to the simpler behavior that was originally used in both this PR and #2847: just a basic `if File.writable?(Gem.default_dir)`.

In retrospect I'm not convinced the more complex condition actually solves a problem.

@deivid-rodriguez @hsbt if y'all are okay with my latest change, then this is good to go.

duckinator:
Rebased off master (e6bd3cb).

hsbt (approved):

lib/rubygems/path_support.rb
```diff
@@ -18,20 +18,32 @@ class Gem::PathSupport
...
+    elsif File.writable?(Gem.default_dir)
+      @home = normalize_home_dir(Gem.default_dir)
+    else
+      # If `GEM_HOME` is not set AND we can't use `Gem.default_dir`,
+      # default to a user installation and set `@auto_user_install`.
+      @auto_user_install = true
+      @home = normalize_home_dir(Gem.user_dir)
     end
 
-    @home = expand(@home)
```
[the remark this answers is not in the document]
duckinator: I'm not opposed to this being refactored, but honestly I've been working on this off and on for 20 months and I just want it to be done. 😂 

Between that and not having much time to work on RubyGems/Bundler stuff for the next week or so, I'm gonna go ahead and merge this so the functionality is in place. A follow-up PR is always an option. 🙂

simi:
:partying_face: @duckinator last question to answer is how to release this. Would be minor bump enough?

deivid-rodriguez:
Nice job @duckinator!

duckinator:
This should ONLY affect RubyGems behavior in situations where it would not work at all before, so I think a minor version bump is sufficient. 

If this changes behavior when GEM_HOME is set, or when the default gem home is writable, that's a bug and should be treated as such.

voxik:
Playing with this PR, I have realized one more issue. On Fedora, we simply never wont `gem` installed gems to be installed in Ruby directory. Ruby directory is managed by RPM, period. `gem` command cannot touch such content, otherwise we risk inconsistent system state. However, this PR does not address this unfortunately.

And actually, I believe that Ruby(Gems) should disallow intermixing gems shipped with Ruby to mix together with gems installed by `gem` command. This should not be the default behavior.

duckinator:
It seems like my attempt to limit the changes to situations that would've previously been errors wound up causing more problems than it solved. I'm beginning to wonder if the "right" solution might be to only install globally if it's specifically requested. Something along these lines:

1. If `GEM_HOME` is specified, use that.
2. If `GEM_GLOBAL_INSTALL` is specified, use `Gem.default_dir` (the thing using `RbConfig::CONFIG["rubylibprefix"]`, where you don't want users to mess around).
3. Otherwise, use `Gem.user_dir`.

This would obviously need some more input from other RubyGems maintainers. But it feels like the root of the problem is trying to automatically choose between two options, so I'm wondering if the fix may be to just ... not have one option be considered at all unless someone goes out of their way to enable it. Because then the behavior most people have to care about is just "`GEM_HOME` is specified, or `Gem.user_dir`", which is a lot easier to reason about. But you'd be able to specify `GEM_GLOBAL_INSTALL=true` when packaging things (like in #7083) in situations where you _do_ want to have `gem` play with the actual Ruby installation directory.

@voxik Do you think this would solve your problems with this PR + adequately address your concern wrt intermixing gems shipped with Ruby + gems installed via the `gem` command?

voxik:
Since you have mentioned `Gem.default_dir`, I think I should open separate ticket and take a better look on where gems are stored. Because this to me look as piling one exception above other unfortunately.

voxik:
> > However, this PR does not address this unfortunately.
> 
> Why not? I assume RPM directories are not writable by the user, isn't it?

It depends who the user is. `root` is also user. On Fedora, we have choose that if the user is root, the `gem install` modifies some systemwide gem directory. Specifically `/usr/local/share/gems`. More details can be found [here](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb#_101) and this might be also useful information:

~~~
$ ruby -r rbconfig -e "p Gem.default_dirs"
{:system=>{:bin_dir=>"/usr/bin", :gem_dir=>"/usr/share/gems", :ext_dir=>"/usr/lib64/gems"}, :local=>{:bin_dir=>"/usr/local/bin", :gem_dir=>"/usr/local/share/gems", :ext_dir=>"/usr/local/lib64/gems"}}
~~~

But generally, it is not really important if the location is user home or some other directory and under what conditions it is writable. The point is that there is no reason why RubyGems should not be flexible and configurable in this regard. And the heuristics baked into RubyGems collides with the configuration we would like to use for Fedora.

And also please note that we have special use case, which is building of RPM itself. For that case, we are creating fake root, where the directories don't necessarily exists etc. This can be seen again in our [operating_system.rb](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb#_126)

> For various reasons I mentioned previously the safest is to put user-installed gems under the ruby prefix

I don't disagree. But this is not necessarily case of Linux distribution. For Fedora, I see Ruby as building piece to be able to integrate other applications. Everything distributed via RPM. `gem install` or `bundler install` is just to add extensions which are not available in the distribution. Now the question is system upgrades and Ruby major bumps. But this is always problematic case if something such as `~/.gem` directory is used. For Fedora, there is no easy solution no matter what.

eregon:
Right, for `root` you need to keep the logic in [operating_system.rb](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/operating_system.rb). So this doesn't address Fedora's case entirely on its own, but it also doesn't make anything worse.
For non-root I think you might be able to use this new functionality and not alter the gem path.

deivid-rodriguez:
I just found that rubyinstaller2 is already implementing this same functionality: https://github.com/oneclick/rubyinstaller2/blob/1037499985fcaff2c471e88ed254a7b698e508f6/resources/files/operating_system.rb.

The approach in there is probably more bulletproof than ours by the way.

@larskanis I think you may be able to get rid of all that for Ruby 3.3?

PR #5331, opened 2022-02-02
Improve internal development dependency management

deivid-rodriguez:

I wanted to report an issue to `rspec-mocks`, but it's currently a bit cumbersome to reproduce because our current `rake update` task is very limited, since:

* It only updates lockfiles, it doesn't install any dependencies.
* It doesn't support updating specific dependencies.

So the instructions to repro the issue would be

> Run `rake update`, then run `rake setup`, and all development dependencies are updated, not just rspec-mocks.

With this changes, I can tell `rspec-mocks` maintainers to run `rake "update_dev_dep[rspec-mocks]"` and the issue will be ready to reproduce.


simi:
:+1: 

Btw. what kind of an issue are you going to report?

deivid-rodriguez:
The spec issues we get when upgrading to rspec-mocks 3.10.3. Basically ask whether it's an intentional change that we should address ourselves, or a regression that they plan to fix themselves. If it's the latter, we don't need to take any action and can wait for 3.10.4.

simi:
I think it is intentional change.

https://github.com/rspec/rspec-mocks/commit/fefabc4de9e4fbf7b7d859478d9a3794d4f4f107#r65706463

I'll take a look how to improve the confusing assertion message.

deivid-rodriguez:
Yeah, you're most likely right, it's probably easiest to adapt the code and move on.

By the way, in https://github.com/rspec/rspec-mocks/pull/1394#issuecomment-755835402 @mame was already aware of the confusing error message and said he already had something coded to show a better message.

simi:
Good yes, I'll ask @mame in there.

PR #5335, opened 2022-02-07
Don't silently persist `BUNDLE_WITH` and `BUNDLE_WITHOUT` envs locally

### LINKED ISSUE #3708: Calling bundle install with BUNDLE_WITH runtime env var persists it in .bundle/config

### LINKED ISSUE #4835: BUNDLE_WITHOUT environment variable is lower priority than .bundle/config

The `BUNDLE_WITHOUT` environment variable is lower priority than a cached value in `.bundle/config`, leading to unexpected behaviour.

```
# ruby via rbenv
❯ ruby --version
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin20]
❯ bundle --version
Bundler version 2.2.25
```

Given the following `Gemfile`:

```ruby
source "https://rubygems.org"

group :test, :development do
    gem "rspec"
end
```

I run `BUNDLE_WITHOUT="test" bundle install` which still installs `rspec` as I forgot to specify `BUNDLE_WITHOUT=test:development`

So, I update my command to exclude both of those groups, but the gems are still installed:

```ruby
❯ BUNDLE_WITHOUT="test:development" bundle install
Using bundler 2.2.25
Using diff-lcs 1.4.4
Using rspec-support 3.10.2
Using rspec-core 3.10.1
Using rspec-expectations 3.10.1
Using rspec-mocks 3.10.2
Using rspec 3.10.0
Bundle complete! 1 Gemfile dependency, 7 gems now installed.
Gems in the group 'test' were not installed.
```

The issue here is that `BUNDLE_WITHOUT` was cached after my first run in `.bundle/config`:

```bash
❯ cat .bundle/config
---
BUNDLE_PATH: "vendor/bundle"
BUNDLE_WITHOUT: "test"
```

The solution is to remove the `BUNDLE_WITHOUT` line in `.bundle/config` and re-run your command.

```bash
BUNDLE_WITHOUT="test" bundle install
BUNDLE_WITHOUT="test:development" bundle install
```

Bundler would exclude both groups on the second run

### What actually happened?

```
❯ BUNDLE_WITHOUT="test:development" bundle install
Using bundler 2.2.25
Using diff-lcs 1.4.4
Using rspec-support 3.10.2
Using rspec-core 3.10.1
Using rspec-expectations 3.10.1
Using rspec-mocks 3.10.2
Using rspec 3.10.0
Bundle complete! 1 Gemfile dependency, 7 gems now installed.
Gems in the group 'test' were not installed.
```

## Environment

```
Bundler       2.2.25
Ruby          2.7.2p137 (2020-10-01 revision 5445e0435260b449decf2ac16f9d09bae3cafe72) [x86_64-darwin20]
  Full Path   /Users/michael/.rbenv/versions/2.7.2/bin/ruby
  Config Dir  /Users/michael/.rbenv/versions/2.7.2/etc
RubyGems      3.1.4
  Gem Home    /Users/michael/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0
  Gem Path    /Users/michael/.gem/ruby/2.7.0:/Users/michael/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0
  User Home   /Users/michael
  User Path   /Users/michael/.gem/ruby/2.7.0
  Bin Dir     /Users/michael/.rbenv/versions/2.7.2/bin
Tools         
  Git         2.30.1 (Apple Git-130)
  RVM         not installed
  rbenv       rbenv 1.1.2
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2021-07-30
Git SHA           7f0f257c7a
Released Version  true
```

## Bundler settings

```
path
  Set for your local app (/private/tmp/til/.bundle/config): "vendor/bundle"
without
  Set for your local app (/private/tmp/til/.bundle/config): [:development]
```

### Gemfile

```ruby
source "https://rubygems.org"

group :test, :development do
    gem "rspec"
end
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    diff-lcs (1.4.4)
    rspec (3.10.0)
      rspec-core (~> 3.10.0)
      rspec-expectations (~> 3.10.0)
      rspec-mocks (~> 3.10.0)
    rspec-core (3.10.1)
      rspec-support (~> 3.10.0)
    rspec-expectations (3.10.1)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.10.0)
    rspec-mocks (3.10.2)
      diff-lcs (>= 1.2.0, < 2.0)
      rspec-support (~> 3.10.0)
    rspec-support (3.10.2)

PLATFORMS
  x86_64-darwin-20

DEPENDENCIES
  rspec

BUNDLED WITH
   2.2.25
```

deivid-rodriguez:

`BUNDLE_WITH` and `BUNDLE_WITHOUT` environment variables, are silently persisted to local configuration.

Fix is to refactor `bundle install` option handling to not do this.

This PR reintroduces #3666, which was reverted by https://github.com/rubygems/rubygems/pull/4154, because it should no longer have any unexpected consequences, since configuration is now local by default if inside an application.

Fixes #3708.
Fixes #4835.

### DIFF
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -154,7 +154,7 @@ module Bundler
       # need to nil them out first to get around validation for backwards compatibility
       Bundler.settings.set_command_option :without, nil
       Bundler.settings.set_command_option :with,    nil
-      Bundler.settings.set_command_option :without, without - with
+      Bundler.settings.set_command_option :without, without
       Bundler.settings.set_command_option :with,    with
     end
 
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -140,21 +140,11 @@ module Bundler
 
       check_for_group_conflicts_in_cli_options
 
-      with = options.fetch(:with, [])
-      with |= Bundler.settings[:with].map(&:to_s)
-      with -= options[:without] if options[:without]
-      with = nil if options[:with] == []
-
-      without = options.fetch(:without, [])
-      without |= Bundler.settings[:without].map(&:to_s)
-      without -= options[:with] if options[:with]
-      without = nil if options[:without] == []
-
       # need to nil them out first to get around validation for backwards compatibility
       Bundler.settings.set_command_option :without, nil
       Bundler.settings.set_command_option :with,    nil
-      Bundler.settings.set_command_option :without, without
-      Bundler.settings.set_command_option :with,    with
+      Bundler.settings.set_command_option :without, options[:without]
+      Bundler.settings.set_command_option :with,    options[:with]
     end
 
     def normalize_settings
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -140,16 +140,15 @@ module Bundler
 
       check_for_group_conflicts_in_cli_options
 
-      Bundler.settings.set_command_option :with, nil if options[:with] == []
-      Bundler.settings.set_command_option :without, nil if options[:without] == []
-
       with = options.fetch(:with, [])
       with |= Bundler.settings[:with].map(&:to_s)
       with -= options[:without] if options[:without]
+      with = nil if options[:with] == []
 
       without = options.fetch(:without, [])
       without |= Bundler.settings[:without].map(&:to_s)
       without -= options[:with] if options[:with]
+      without = nil if options[:without] == []
 
       # need to nil them out first to get around validation for backwards compatibility
       Bundler.settings.set_command_option :without, nil
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -154,13 +154,11 @@ module Bundler
       options[:with]    = with
       options[:without] = without
 
-      unless Bundler.settings[:without] == options[:without] && Bundler.settings[:with] == options[:with]
-        # need to nil them out first to get around validation for backwards compatibility
-        Bundler.settings.set_command_option :without, nil
-        Bundler.settings.set_command_option :with,    nil
-        Bundler.settings.set_command_option :without, options[:without] - options[:with]
-        Bundler.settings.set_command_option :with,    options[:with]
-      end
+      # need to nil them out first to get around validation for backwards compatibility
+      Bundler.settings.set_command_option :without, nil
+      Bundler.settings.set_command_option :with,    nil
+      Bundler.settings.set_command_option :without, options[:without] - options[:with]
+      Bundler.settings.set_command_option :with,    options[:with]
     end
 
     def normalize_settings
@@ -184,7 +182,7 @@ module Bundler
 
       Bundler.settings.set_command_option_if_given :clean, options["clean"]
 
-      normalize_groups
+      normalize_groups if options[:without] || options[:with]
 
       options[:force] = options[:redownload]
     end
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -151,14 +151,11 @@ module Bundler
       without |= Bundler.settings[:without].map(&:to_s)
       without -= options[:with] if options[:with]
 
-      options[:with]    = with
-      options[:without] = without
-
       # need to nil them out first to get around validation for backwards compatibility
       Bundler.settings.set_command_option :without, nil
       Bundler.settings.set_command_option :with,    nil
-      Bundler.settings.set_command_option :without, options[:without] - options[:with]
-      Bundler.settings.set_command_option :with,    options[:with]
+      Bundler.settings.set_command_option :without, without - with
+      Bundler.settings.set_command_option :with,    with
     end
 
     def normalize_settings
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -135,9 +135,6 @@ module Bundler
     end
 
     def normalize_groups
-      options[:with] &&= options[:with].join(":").tr(" ", ":").split(":")
-      options[:without] &&= options[:without].join(":").tr(" ", ":").split(":")
-
       check_for_group_conflicts_in_cli_options
 
       # need to nil them out first to get around validation for backwards compatibility
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -367,7 +367,7 @@ module Bundler
 
     def to_array(value)
       return [] unless value
-      value.split(":").map(&:to_sym)
+      value.tr(" ", ":").split(":").map(&:to_sym)
     end
 
     def array_to_s(array)
--- a/bundler/spec/install/gemfile/groups_spec.rb
+++ b/bundler/spec/install/gemfile/groups_spec.rb
@@ -91,16 +91,7 @@ RSpec.describe "bundle install with groups" do
         expect(the_bundle).to include_gems "rack 1.0.0", :groups => [:default]
       end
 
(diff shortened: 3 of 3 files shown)


nsmmrs:
For anyone else who might be scratching their heads today (I hope it's OK to comment here):

We have a custom environment, `staging`, running on Heroku, and this bugfix resulted in crashing on `bundle exec` with "Could not find [all of the development/test gems] in any of the sources".

The reason was that our image is built with `BUNDLE_WITHOUT` being set via `ARG`, which is not persisted to the final image. So our image only worked in `staging` because of the Bundler bug which was silently persisting that configuration.

Luckily, fixing this was as easy as running:
```sh
heroku config:set BUNDLE_WITHOUT=development:test --remote staging
```

deivid-rodriguez:
Oh, that's right! First time I fixed this, people reported something similar, but in that case they were configuring `bundle config set without development:test` (which was unintentionally global), and then thanks to this bug, that global configuration was getting copied locally and making things work. Since I also changed `bundle config` to be local now, I didn't expect that issue to be hit again, but it does break your case.

I'm sorry about it, hopefully not too many people is bitten by it, and hopefully your fix will be useful to them!

nsmmrs:
@deivid-rodriguez No problem! This was definitely a problem of our own creation. 😅 Thanks for the bugfix!

gugaiz:
After this change, when deploying to our staging environment on AWS Beanstalk, the app will not start because we are getting  "Could not find [all of the development/test gems] in any of the sources" even though we have the env variable BUNDLE_WITHOUT set to "development:test"

deivid-rodriguez:
`BUNDLE_WITH` needs to be set both when installing and when starting the application. Is that the case?

gugaiz:
BUNDLE_WITHOUT is set as an environment variable on the OS so it is present on both (installing and starting)

deivid-rodriguez:
Well, then I don't know what's happening, sorry. If you give me instructions to reproduce this I can try to help.

gugaiz:
Trying to figure out what could be happening... so far I found that adding the following debug line into `bundler/definition.rb:482:in 'materialize'` function 

```
puts "BUNDLE_WITHOUT=#{ENV['BUNDLE_WITHOUT']}"
```
I can see that `BUNDLE_WITHOUT` variable is not set but all other env variables are set, ie if on bash I do
 
```
$ printenv|grep BUNDLE_
BUNDLE_WITHOUT=test:development
```
I also noticed that if I run `bundle info spring` I get the following result
```
$ bundle info spring
BUNDLE_WITHOUT=test:development
Could not find gem 'spring', because it's in the group 'development', configured to be ignored.
```
meaning that it is set in that case.

It seems to be an issue related to beanstalk environments. I will create a ticket there.

nsmmrs:
@gugaiz On Heroku, the build stage ENV is configured separately from the normal application ENV. AWS may have a similar setup.

PR #5336, opened 2022-02-08
Fix typo in major_ deprecation message

svoop:
Using the newer "gems" file names convention, the [lockfile is expected to have the `.rb` extension replaced with `.locked`](https://github.com/rubygems/rubygems/blob/a9e305432955f56b847881187a7e01a97b6901e3/bundler/lib/bundler/shared_helpers.rb#L29)

### DIFF
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -141,7 +141,7 @@ module Bundler
       end
       return unless multiple_gemfiles
       message = "Multiple gemfiles (gems.rb and Gemfile) detected. " \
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Thanks! There are some specs using the old message, so you'll need to update those too.

svoop:
@deivid-rodriguez Ah, darn, that's what happens if you go "let's try that easy in browser edit and PR feature of GitHub for a change". :smile: I'll fix this later today!

deivid-rodriguez:
Hehe, right. Sometimes we assert for too specific error messages, but it's hard to find the right balance there 😅.

svoop:
I'll close this one and do a proper PR from a "real" fork.

PR #5337, opened 2022-02-09
Suggest quicker `bundle add` for installation in `README.md` generated by `bundle gem` 

bradgessler:
Reduce the number of steps required to install a gem from two steps to one by using `bundle add`

I keep seeing 2-step installation instructions for installing gems on README's all around the internet. `bundler add` has been around long enough now where it seems safe to suggest to people installing a gem to use that command.

This will save thousands of people-hours per year and have the advantage of locking the dependency with a `~>` by default, which will save even more time.

Updated the `README.md.tt` doc to tell people how to install a gem via `bundle add` instead of the old two-step method.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/README.md.tt
+++ b/bundler/lib/bundler/templates/newgem/README.md.tt
@@ -6,17 +6,11 @@ TODO: Delete this and the text above, and describe your gem
 
 ## Installation
 
-Add this line to your application's Gemfile:
+Install the gem and add to the application's Gemfile by executing:
 
-```ruby
-gem '<%= config[:name] %>'
-```
+    $ bundle add <%= config[:name] %>
 
-And then execute:
-
-    $ bundle install
-
-Or install it yourself as:
+If bundler is not being used to manage dependencies, install the gem by executing:
 
     $ gem install <%= config[:name] %>
 


deivid-rodriguez (approved):

bundler/lib/bundler/templates/newgem/README.md.tt
```diff
@@ -6,15 +6,9 @@ TODO: Delete this and the text above, and describe your gem
 
 ## Installation
 
-Add this line to your application's Gemfile:
+Add this line to your application's Gemfile by executing:
```
deivid-rodriguez, line 9: `bundle add` not only adds the gem to the `Gemfile`, but also installs it. Can we clarify that somehow?

deivid-rodriguez (commented):
I noticed some minor thing.

bundler/lib/bundler/templates/newgem/README.md.tt
bradgessler: I changed to `Add this line to your application's Gemfile and install by executing:`

bradgessler: Actually I went a step further and attempted to describe the difference between installing via bundler vs gem. I also removed *your application* since it may not actually be the persons' application.

```md
## Installation

Install the gem and add to the application's Gemfile by executing:

    $ bundle add <%= config[:name] %>

If bundler is not being used to manage dependencies, install the gem by executing:

    $ gem install <%= config[:name] %>
```

simi (approved):

paddor:
Funny how I just used it this week for the first time and now I see this in RubyWeekly. How long has it been around? 😅

deivid-rodriguez:
`bundle add`? Not sure, quite a while!

jon-sully:
I'll be curious to see how much adoption this ends up getting — while I've used `bundle add` before, I tend to agree with Justin's take in https://github.com/testdouble/standard/pull/402#issuecomment-1075369113 both because I tend to see folks wanting to keep their Gemfile in alphabetical order but also because a lot of folks add a comment after each gem declaration with a high level description (which I think is good practice)

indirect:
I strongly prefer `bundle add` followed by whatever edits you want—it's much faster to `bundle add` and then edit your Gemfile than it is to look up the current version of a gem and type a version constraint by hand.

bradgessler:
> I'll be curious to see how much adoption this ends up getting — while I've used `bundle add` before, I tend to agree with Justin's take in https://github.com/testdouble/standard/pull/402#issuecomment-1075369113 both because I tend to see folks wanting to keep their Gemfile in alphabetical order but also because a lot of folks add a comment after each gem declaration with a high level description (which I think is good practice)

I encourage you to read https://bundler.io/man/bundle-add.1.html, note the gaps in your concerns, and enhance `bundle add` to insert alphabetically or have a comment CLI flag that automatically inserts the comment above the Gemfile entry when added. Those are all good ideas.

It doesn't bother me that much that maintainers, like Justin, don't want to use `bundle add`. I'd guess they feel the change is pedantic and it's not an established norm yet, and they're not wrong.

Between this PR and making people more aware of `bundle add` in newsletters like Ruby Weekly or some of the newer Rails docs, I'm confident we'll get to our "add a gem with one command" nirvana in the near future.

deivid-rodriguez:
I think when you first add a gem to your application you don't necessarily care about the `Gemfile` style. You might just want to try it out, see if it works for your application, or whatever. `bundle add` gets the job done quickly for everyone. Those who want to keep a curated `Gemfile` (which I really like and recommend) still need to do similar edits than before, so I don't think this really changes the experience for them.

jon-sully:
Great (and pragmatic) thoughts, @bradgessler. 👍

simi:
:information_source: I left some comments regarding problem with commands modifying `Gemfile` at https://github.com/rubygems/rfcs/pull/14#issuecomment-1070034054.

PR #5346, opened 2022-02-09
Add clarification for bundle-config "with" option

izenk:

The `with` configuration option only affects _optional_ Gemfile groups. Other groups are installed regardless of `with`, unless `without` is configured. However, documentation does not mention this.

Update documentation to mention expected behavior.

Closes https://github.com/rubygems/bundler-site/issues/597.


deivid-rodriguez:
Thanks @izenk, looks good! I think you just need to run `bin/rake man:build` from Bundler folder in your git clone to properly regenerate man pages.

izenk:
@deivid-rodriguez  Should I commit all regenerated pages?

deivid-rodriguez (approved):

PR #5353, opened 2022-02-12
Add newline to validate_platforms! message when platform is missing

win93:

When I run bundle install with BUNDLE_DEPLOYMENT=true in the environment
on a different platform than I usually do development, I get the
following output to the console (wrapped exactly as shown):

    Your bundle only supports platforms ["x86_64-darwin-19"] but your local platform
    is x86_64-linux. Add the current platform to the lockfile with `bundle lock
    --add-platform x86_64-linux` and try again.

Because the way the message wraps, its not as simple as copying the
suggested command to the clipboard because it contains a newline:

    $ bundle lock
    Writing lockfile to [...]/Gemfile.lock
    $ --add-platform x86_64-linux

Adding a newline right before the command forces the command in the
error message to be on the same line, which facilitates copy-pasting the
command in the message.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -433,7 +433,7 @@ module Bundler
 
       raise ProductionError, "Your bundle only supports platforms #{@platforms.map(&:to_s)} " \
         "but your local platform is #{Bundler.local_platform}. " \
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
I want to work soon on the bigger change of making lockfiles generated outside of linux also work on Linux when possible, which would completely avoid this error altogether in most cases, but this one seems like an easy improvement over what we have now, so I say let's do it.

PR #5396, opened 2022-03-12
Unhide RDoc documentation from top level `Gem` module

nobu:

The `Gem` module is not listed in the “Class and Module Index” at all.

Move `:stopdoc:` directive at the top level in lib/rubygems/compatibility.rb in the module block.
If this is at the top level, it stops the documentation of the entire module, but not only the part in this file.

### DIFF
--- a/lib/rubygems/compatibility.rb
+++ b/lib/rubygems/compatibility.rb
@@ -1,5 +1,4 @@
 # frozen_string_literal: true
-# :stopdoc:
 
 #--
 # This file contains all sorts of little compatibility hacks that we've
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
We discussed something somewhat similar at https://github.com/rubygems/rubygems/pull/5299, any ideas to prevent this kind of issue from being reintroduced?

deivid-rodriguez (approved):

nobu:
Although `rdoc --verbose` shows a `module Gem (undocumented)` line, this is buried in a lot of similar messages.
I guess rdoc may be possible to warn when already documented class/module is ignored.
Or, simple `git grep '^# *:stopdoc:' -- '*.rb'` may help in many cases.

deivid-rodriguez:
I run `rdoc --verbose` and I get three "module Gem (undocumented)" messages with or without this PR.

nobu:
I think this is a bug of rdoc, a `:stopdoc:` in a file should not affect the same class/module in other files.

deivid-rodriguez:
@nobu I'm looking into this and I'm not seeing the missing `Gem` module. When I run `rdoc` and open the generated docs, the `Gem` module _is_ listed under "Class and Module Index". 🤔

nobu:
Snapshot of the part of https://docs.ruby-lang.org/en/master/: image
The next to `GC::Profiler` is `Gem::AvailableSet`, there should be `Gem` between them.

PR #5399, opened 2022-03-14
[DOC] Enable Gem::Package example

nobu:

The document of `Gem::Package` is not shown in the generated file.
https://docs.ruby-lang.org/en/master/Gem/Package.html

Other code must not be between the doc and class definition.


nobu:
Sorry if this is intentional since that is an example but not a document.

deivid-rodriguez (approved):
Exactly the same issue as https://github.com/rubygems/rubygems/pull/5299, but for `Gem::Package` instead of `Gem::Version`.

PR #5410, opened 2022-03-21
Relax a test to test `bin/bundle update --bundler` to pass on ruby/ruby

junaruga:
This change to relax the test is to pass `make test-bundler` on ruby/ruby,
where the `system_bundler_version` (= `Bundler::VERSION`) is not the latest version on the RubyGems.
See <https://bugs.ruby-lang.org/issues/18643>.

The test can fail in `make test-bundler` (`make test-bundler BUNDLER_SPECS="commands/binstubs_spec.rb"`) in ruby/ruby, where the `Bundler::VERSION` is not the latest version in RubyGems.

I relaxed the matching criteria not depending on the version number.

I tried to use a regular expression seeing the RSpec `include` method document,
https://relishapp.com/rspec/rspec-expectations/v/3-10/docs/built-in-matchers/include-matcher

```
expect(out).to include %r(Using bundler \w+\n)
```

However I see there is a syntax error `no implicit conversion of Regexp into String`. So, I am just using a string matching.


bundler/spec/commands/binstubs_spec.rb
```diff
@@ -241,7 +241,7 @@
         it "falls back to the latest installed bundler" do
           FileUtils.rm bundled_app_lock
           sys_exec "bin/bundle install", :env => { "DEBUG" => "1" }
-          expect(out).to include "Using bundler #{system_bundler_version}\n"
+          expect(out).to include "Using bundler "
```
deivid-rodriguez, line 244: This is not the spec that's failing, it's the one at line 229, right?

deivid-rodriguez: To fix the spec in line 229, the fix is not good because if we change the expectation, then the spec no longer checks what it's supposed to check (that bundler actually "trampolines" to a higher version). Instead we should parse the version in the "Using bundler x.y.z\n" message, and check that the version is `>= system_bundler_version`. I think that would get the spec to pass on ruby's stable branch while still testing what we want to test here.

deivid-rodriguez (changes requested):

bundler/spec/commands/binstubs_spec.rb
junaruga: Oh,,, sorry yes you are right..

junaruga: > Oh,,, sorry yes you are right..

Rebased with the line 229 modification.

> To fix the spec in line 229, the fix is not good because if we change the expectation, then the spec no longer checks what it's supposed to check (that bundler actually "trampolines" to a higher version). Instead we should parse the version in the "Using bundler x.y.z\n" message, and check that the version is >= system_bundler_version. I think that would get the spec to pass on ruby's stable branch while still testing what we want to test here.

That's a good idea. Let me try to update this PR.

junaruga:
Rebased with the implementation to compare the Bundler version. The `bin/rubocop` on Ruby 3.1.1 shows an issue maybe coming from rubocop itself. Hope the `rubocop` test passes on CI.

```
$ ruby -v
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x86_64-linux]

$ bin/rubocop spec/commands/binstubs_spec.rb
Inspecting 1 file
An error occurred while Layout/BlockAlignment cop was inspecting /home/jaruga/var/git/rubygems/bundler/spec/commands/binstubs_spec.rb:3:0.
To see the complete backtrace run rubocop -d.
.

1 file inspected, no offenses detected

1 error occurred:
An error occurred while Layout/BlockAlignment cop was inspecting /home/jaruga/var/git/rubygems/bundler/spec/commands/binstubs_spec.rb:3:0.
Errors are usually caused by RuboCop bugs.
Please, report your problems to RuboCop's issue tracker.
https://github.com/rubocop-hq/rubocop/issues

Mention the following information in the issue report:
0.80.1 (using Parser 3.1.0.0, running on ruby 3.1.1 x86_64-linux)
```

bundler/spec/commands/binstubs_spec.rb
```diff
@@ -226,7 +226,10 @@
 
         it "calls through to the latest bundler version" do
           sys_exec "bin/bundle update --bundler", :env => { "DEBUG" => "1" }
-          expect(out).to include %(Using bundler #{system_bundler_version}\n)
+          word_found = /Using bundler ([\w\.]+)\n/.match(out)
```
deivid-rodriguez, line 229: Looks good now! Only nitpick from my side is that I don't like the `word_found` name too much, I would use something a bit more descriptive like `using_bundler_line` or something. And maybe also using a named capture could make it more readable

```diff
diff --git a/bundler/spec/commands/binstubs_spec.rb b/bundler/spec/commands/binstubs_spec.rb
index aea897f03e..64aa4cbbc9 100644
--- a/bundler/spec/commands/binstubs_spec.rb
+++ b/bundler/spec/commands/binstubs_spec.rb
@@ -226,9 +226,9 @@
 
         it "calls through to the latest bundler version" do
           sys_exec "bin/bundle update --bundler", :env => { "DEBUG" => "1" }
-          word_found = /Using bundler ([\w\.]+)\n/.match(out)
-          expect(word_found).to_not be_nil
-          latest_version = word_found[1]
+          using_bundler_line = /Using bundler (?<version>[\w\.]+)\n/.match(out)
+          expect(using_bundler_line).to_not be_nil
+          latest_version = using_bundler_line[:version]
           expect(Gem::Version.new(latest_version)).to be >= Gem::Version.new(system_bundler_version)
         end
 
```

junaruga: OK. Let me fix it.

junaruga: I think RubyGems supports Ruby >= 2.3. seeing the `.rubocop.yml` file. Seeing Ruby 2.3 `Regexp` document, I didn't know `Regexp#match` returns a named capture, while `Regexp#named_captures` is explained on this purpose.
https://docs.ruby-lang.org/en/2.3.0/Regexp.html#method-i-match

deivid-rodriguez: I did a quick check and RubyGems code base does not really use named captures at all at the moment, so let's just ignore that suggestion, sorry for the noise.

junaruga: Searching by `\.march`, I found 2 cases using maybe `String#match` https://docs.ruby-lang.org/en/2.3.0/String.html#method-i-match .

```
$ grep -r '\.match' bundler/spec/
...
bundler/spec/support/helpers.rb:            version = gem_name.match(/\Abundler-(?<version>.*)\z/)[:version] if gem_name != "bundler"
...
bundler/spec/support/builders.rb:        File.basename(file, ".gem").match(/\A(?<gem_name>[^-]+)-.*\z/)[:gem_name]
```

```
$ grep -r '\.named_captures' bundler/spec/
  => empty
```

The variable name `word_found` is used in some parts in the tests.

```
$ grep -r word_found bundler/spec/
bundler/spec/commands/binstubs_spec.rb:          word_found = /Using bundler ([\w\.]+)\n/.match(out)
bundler/spec/commands/binstubs_spec.rb:          expect(word_found).to_not be_nil
bundler/spec/commands/binstubs_spec.rb:          latest_version = word_found[1]
bundler/spec/quality_es_spec.rb:      next unless word_found = pattern.match(line)
bundler/spec/quality_es_spec.rb:      failing_line_message << "#{filename}:#{number.succ} contiene '#{word_found}'. Esta palabra tiene un significado subjetivo y es mejor obviarla en textos técnicos."
bundler/spec/quality_es_spec.rb:      next unless word_found = specific_pronouns.match(line)
bundler/spec/quality_es_spec.rb:      failing_line_message << "#{filename}:#{number.succ} contiene '#{word_found}'. Use pronombres más genéricos en la documentación."
bundler/spec/quality_spec.rb:      next unless word_found = pattern.match(line)
bundler/spec/quality_spec.rb:      failing_line_message << "#{filename}:#{number.succ} has '#{word_found}'. Avoid using these kinds of weak modifiers."
bundler/spec/quality_spec.rb:      next unless word_found = specific_pronouns.match(line)
bundler/spec/quality_spec.rb:      failing_line_message << "#{filename}:#{number.succ} has '#{word_found}'. Use more generic pronouns in documentation."
```

Considering this, if you suggest something, I am happy to consider modifying the PR's code.

deivid-rodriguez: Well, `word_found` makes sense to me when regexp matching words, but in this case not so much. So my only suggestion would be to change that to a better variable name. The named captures is not necessary. As you point out, it's only used very scarcely in bundler's code.

junaruga: OK I will change `word_found` to `using_bundler_line`.

junaruga: No problem at all! I am happy on it.

Issue #5405, opened 2022-03-18
Unable to marshal gemspecs with YAML


jessebs:
### Describe the problem as clearly as you can

This is a doozy and I'm not exactly sure what caused this to pop up - it could be bundler 2.3.9 or there could have been a change on our build stack that brought this to light.

When dependency resolution is done through a full index (downloading gemspec.rz files), there is a condition within the published gemspecs that prevents parsing and prevents bundler from resolving dependencies. 

### Did you try upgrading rubygems & bundler?

Yes - both are the latest

### Post steps to reproduce the problem

Create a gemfile
```
source 'https://rubygems.org'
gem 'pry', '= 0.4.7'
```

Then run:
`bundle install --full-index --verbose`

You'll see the error.  It works without full-index

The problem comes down to line containing: `Bundler.load_marshal Bundler.rubygems.inflate(downloader.fetch(uri).body)` in lib/bundler/fetcher.rb

It looks like a very simple resolution is to `require 'yaml'` in the fetcher.rb file

### Which command did you run?

`bundle install --full-index --verbose`

### What were you expecting to happen?

The gem is installed

### What actually happened?

```
Running `bundle install --full-index --verbose` with bundler 2.3.9
Found no changes, using resolution from the lockfile
The definition is missing ["pry-0.4.7"]
HTTP GET https://index.rubygems.org/versions
HTTP 304 Not Modified https://index.rubygems.org/versions
Fetching source index from https://rubygems.org/
Need to query more than 500 gems. Downloading full index instead...
HTTP GET https://rubygems.org/quick/Marshal.4.8/pry-0.4.7.gemspec.rz
HTTP 200 OK https://rubygems.org/quick/Marshal.4.8/pry-0.4.7.gemspec.rz
Gemspec ["pry", #<Gem::Version "0.4.7">] contained invalid data.
Your network or your gem server is probably having issues right now.
Bundler::HTTPError: Gemspec ["pry", #<Gem::Version "0.4.7">] contained invalid data.
Your network or your gem server is probably having issues right now.
/Users/ME/.rvm/gems/ruby-2.7.1@pry_test/gems/bundler-2.3.9/lib/bundler/fetcher.rb:112:in `rescue in fetch_spec'
  /Users/ME/.rvm/gems/ruby-2.7.1@pry_test/gems/bundler-2.3.9/lib/bundler/fetcher.rb:98:in `fetch_spec'
[...]

PR #5416, opened 2022-03-22
More helpul reporting of marshal loading issues

deivid-rodriguez:

This rescue is hiding a real bundler issue inside a network error. See #5405.

It's more helpful to get a proper bug report, with stack trace and so on. I don't think network issues result in invalid marshal data, at least that has never been reported to us.

Remove this error handling.


simi:
Actually I have faced `Your network or your gem server is probably having issues right now.` error recently using private gem server which was really in issues and it was enough to find out the problem. Seeing the stacktrace would make it unclear what has happened (if the problem is on the client or server side). Would it be possible to keep both? Inform about unexpected response and also include the stacktrace for debug?

deivid-rodriguez:
Can you detail these server issues? I think there's two potential issues here:

* Gem server having issues and server "garbage" / incomplete responses from the marshaled endpoint. In this case, the client will see a `TypeError` because of `Marshal.load` not understanding the marshal format of the response. I think this is fine, so I can restore the "server is having issues message" to still be printed in that case?

* The one being "unrescued" here where the server response is correct, but the marshaled ruby code can't be evaluated due to some issue, in this case, a client issue. This kind of problem, even if it could potentially be some server issue that let invalid marshaled specs be uploaded, seems better not to hide I believe.

So I think I can restore the "server is having issues" message, but not raise it in this particular case. How about that?

deivid-rodriguez:
I updated the PR with my alternative proposal.

simi:
If I remember well the server just timed out.

deivid-rodriguez:
> If I remember well the server just timed out.

And you still got this error, I would've expected a timeout error to be raised before trying to un marshal the response, maybe it was a similar wording raised from somewhere else?

Anyways, this version is good enough to me, so let's do it!

simi:
I'm sadly not able to constantly reproduce this on my side. I'll take a look once it happens again, but I would rather to not face this problem :).

PR #5420, opened 2022-03-23
Prepare rubygems 3.3.10 and bundler 2.3.10

deivid-rodriguez:
It's release day!

@simi two questions. Cool with releasing #5175 now, right? Also, may I also sneak #5416 into the release?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+# 3.3.10 / 2022-03-23
+
+## Enhancements:
+
+* Installs bundler 2.3.10 as a default gem.
+
+## Documentation:
+
+* Enable `Gem::Package` example in RDoc documentation. Pull request #5399
+  by nobu
+* Unhide RDoc documentation from top level `Gem` module. Pull request
+  #5396 by nobu
+
 # 3.3.9 / 2022-03-09
 
 ## Enhancements:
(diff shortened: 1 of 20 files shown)


simi:
I'm ok to release https://github.com/rubygems/rubygems/pull/5175, but I would not promote it for now unless `bundler gem` integration is done, which is going to introduce Rust extension gem structure. The gem structure in tests (which is the only reference implementation I'm aware of) is currently simplified and it will be changed as well in the `bundler gem` integration to follow the structure.

deivid-rodriguez:
I see, so we can release it without a changelog entry? I think there's no rush to ship it, so we can also just delay it a little bit until we can tweak the `bundle gem` generated skeleton. What do you think?

deivid-rodriguez:
Oh, ok, so we release it normally but don't further promote it for now 👍.

deivid-rodriguez:
Actually, the ruby-core job seems broken due to #5175. Mind skipping it for this release and I'll work on making it green for 3.3.11?

simi:
Feel free to skip it for now if needed.

_Btw. @hsbt planned to include it into Ruby 3.2.preview1 if I remember well. We need to fix the problems anyway soon._

deivid-rodriguez:
Yes, I will address it for next release. But I think it won't block 3.2.preview1 anyways, because this issue seems specific to our stable branch, and I think 3.2.preview1 is cut from Ruby's master branch, which seems fine.

simi:
There are some troubles already AFAIK - https://github.com/ruby/ruby/pull/5669.

PR #5423, opened 2022-03-23
Remove `tmpdir` dependency

deivid-rodriguez:

Sometimes Bundler has issues when creating a temporary directory because it can't find a proper tmp home. Not really our issue but I think we can avoid reports about this by not using `tmpdir` at all.

Remove the `tmpdir` dependency.


deivid-rodriguez:
I'd like to do this for Bundler 2.4. Any concerns? Diff is best viewed with https://github.com/rubygems/rubygems/pull/5423/files?w=1.

indirect (approved):

PR #5427, opened 2022-03-24
Improve Bundler daily CI

deivid-rodriguez:

The `stringio-3.0.0` gem has issues when installed with latest Ruby head. I think `stringio-3.0.1` is fine.

Also there are some specs that can't really pass on ruby-head, so I skipped them. See commit messages for details.

Upgrade the stringio version installed by the standalone spec, and skip the others.


deivid-rodriguez:
This PR should leave a single failure on our daily Bundler CI. The failure is a real issue, and a bit tricky to address, so I will handle it on a separate PR.

Issue #5431, opened 2022-03-26
Creating CAPITAL_NAME gem should produce warning


okuramasafumi:
### Describe the problem as clearly as you can

https://guides.rubygems.org/name-your-gem/
Here in the guide, it says:

> DON’T USE UPPERCASE LETTERS

However, we CAN generate capital-named gem with `bundle gem CAPITAL_NAME` command.

### Did you try upgrading rubygems & bundler?

I confirmed the problem with `Bundler version 2.3.10`

### Post steps to reproduce the problem

`bundle gem CAPITAL_NAME`

### Which command did you run?

`Bundler version 2.3.10`

### What were you expecting to happen?

Printing warning indicating the guide page (https://guides.rubygems.org/name-your-gem/)

### What actually happened?

No warning was produced.

## Environment

```
Bundler       2.3.10
  Platforms   ruby, x86_64-darwin-19
Ruby          3.0.2p107 (2021-07-07 revision 0db68f023372b634603c74fca94588b457be084c) [x86_64-darwin-19]
  Full Path   /Users/okuramasafumi/.rubies/ruby-3.0.2/bin/ruby
  Config Dir  /Users/okuramasafumi/.rubies/ruby-3.0.2/etc
RubyGems      3.3.5
  Gem Home    /Users/okuramasafumi/.gem/ruby/3.0.2
  Gem Path    /Users/okuramasafumi/.gem/ruby/3.0.2:/Users/okuramasafumi/.rubies/ruby-3.0.2/lib/ruby/gems/3.0.0
  User Home   /Users/okuramasafumi
  User Path   /Users/okuramasafumi/.gem/ruby/3.0.0
  Bin Dir     /Users/okuramasafumi/.gem/ruby/3.0.2/bin
Tools
  Git         2.35.1
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

## Bundler Build Metadata

```
Built At          2022-03-23
Git SHA           4bbb70e7de
Released Version  true
```

## Bundler settings

```
build.pg
  Set for the current user (/Users/okuramasafumi/.bundle/config): "--with-pg-config=/Applications/Postgres.app/Contents/Versions/13/bin/pg_config"
gem.changelog
  Set for the current user (/Users/okuramasafumi/.bundle/config): true
gem.ci
  Set for the current user (/Users/okuramasafumi/.bundle/config): "github"
gem.coc
  Set for the current user (/Users/okuramasafumi/.bundle/config): true
gem.linter
[...]

PR #5432, opened 2022-03-27
Raise error when gem contains capital letters

okuramasafumi:
The problem is described in https://github.com/rubygems/rubygems/issues/5431
The guide ( https://guides.rubygems.org/name-your-gem/ ) says
> DON’T USE UPPERCASE LETTERS
so `bundle gem` command should respect it.

Issue #5431 

It now raises error when gem name includes capital letters.
@simi said warning might be enough, but I thought we don't want to proceed with uppercase letters anyway, so raising error should be fine.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -389,8 +389,8 @@ module Bundler
     end
 
     def ensure_safe_gem_name(name, constant_array)
-      if name =~ /^\d/
-        Bundler.ui.error "Invalid gem name #{name} Please give a name which does not start with numbers."
+      if name =~ /(^\d)|[A-Z]/
+        Bundler.ui.error "Invalid gem name #{name} Please give a name which does not start with numbers nor include capital letters."
         exit 1
       end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@simi Are you ok with an error or you prefer a warning? Honestly I would be fine with both.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -389,8 +389,8 @@ def bundler_dependency_version
     end
 
     def ensure_safe_gem_name(name, constant_array)
-      if name =~ /^\d/
-        Bundler.ui.error "Invalid gem name #{name} Please give a name which does not start with numbers."
+      if name =~ /(^\d)|[A-Z]/
+        Bundler.ui.error "Invalid gem name #{name} Please give a name which does not start with numbers nor include capital letters."
```
simi, line 393: `only capital letters`?

deivid-rodriguez:
I guess a warning for now is probably best.

bundler/lib/bundler/cli/gem.rb
okuramasafumi: @simi This regex matches when there's only one capital letter.

simi:
Other level of this would be to make this warning in specification_policy as well. Anyway this seems like good start.

hsbt:
@okuramasafumi I will merge this for warning at first.

Issue #5411, opened 2022-03-21
The test/rubygems/test_gem_ext_cargo_builder trigger the errors when cargo is not installed


junaruga:
I'm having a problem.

My current problem is the `test/rubygems/test_gem_ext_cargo_builder.rb` finishes with error when the `cargo` command is not installed. I think the pending is ok, but the error is problem.

```
$ git clean -fdx

$ rake setup

$ rake test
...
some errors and pendings for cargo.
...

$ ruby -Ilib:test:bundler/lib test/rubygems/test_gem_ext_cargo_builder.rb -n /test_build_staticlib/
Loaded suite test/rubygems/test_gem_ext_cargo_builder
Started
P 
===============================================================================================================================
Pending: test_build_staticlib(TestGemExtCargoBuilder): cargo not present
test/rubygems/test_gem_ext_cargo_builder.rb:15:in `setup'
===============================================================================================================================
E
===============================================================================================================================
Error: test_build_staticlib(TestGemExtCargoBuilder): TypeError: no implicit conversion of nil into String
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:460:in `chdir'
/home/jaruga/var/git/rubygems/test/rubygems/helper.rb:460:in `teardown'
===============================================================================================================================

Finished in 0.013538263 seconds.
-------------------------------------------------------------------------------------------------------------------------------
1 tests, 0 assertions, 0 failures, 1 errors, 1 pendings, 0 omissions, 0 notifications
0% passed
-------------------------------------------------------------------------------------------------------------------------------
73.86 tests/s, 0.00 assertions/s
```

This error doesn't happen when the `cargo` is installed and running the following commands.

```
$ git clean -fdx
$ rake setup
$ rake test
```

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [ ] The command line `gem`
  - [X] Other

Here are my current environment details:

Fedora, Ruby 3.1.1, the latest upstream fe96fb6e2ac5a8b6df5e852470d11fa854301eca .

```
$ ruby -Ilib bin/gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.4.0.dev
  - RUBY VERSION: 3.1.1 (2022-02-18 patchlevel 18) [x86_64-linux]
  - INSTALLATION DIRECTORY: /home/jaruga/local/ruby-3.1.1-test-rubygems/lib/ruby/gems/3.1.0
[...]

PR #5437, opened 2022-03-29
WIP: Fix tests in case of running each relative test file.

junaruga:

Some tests fails when running with a specific relative test file. We should avoid using `__FILE__` in each `test_*` method, because a value of `__FILE__` in the `test_*.rb` is different executed from between `rake test` and `ruby -Ilib:test:bundler/lib test/path/to/test_*.rb`.

When running `rake test`, the value of `__FILE__` is the absolute path. When running `ruby -Ilib:test:bundler/lib test/rubygems/test_foo.rb`, the value of `__FILE__` is the relative path. Note that in the `test_*` method, the working directory (`Dir.pwd`) is `/path/to/tmp/test_rubygems_*`.

```
$ ruby -Ilib:test:bundler/lib test/rubygems/test_gem_ext_cargo_builder.rb -n /test_build_staticlib/
...
Error: test_build_staticlib(TestGemExtCargoBuilder): Errno::ENOENT: No such file or directory @ rb_check_realpath_internal - /home/jaruga/var/git/rubygems/tmp/test_rubygems_20220328-1989428-x5pim5/test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example
...
test/rubygems/test_gem_ext_cargo_builder.rb:27:in `setup_rust_gem'
...
```

You can see https://github.com/rubygems/rubygems/issues/5411 for details.

I did set static file paths by not using `__FILE__`. Add a new test to run each relative test file isolatedly.

I am still working on it to fix other test files to pass the case of running each relative test file.

### DIFF
--- a/.github/workflows/ubuntu-rubygems.yml
+++ b/.github/workflows/ubuntu-rubygems.yml
@@ -44,6 +44,9 @@ jobs:
       - name: Run Test
         run: rake test
         if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
+      - name: Run Test isolatedly
+        run: rake test:isolated
+        if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
       - name: Run Test (JRuby)
         run: JRUBY_OPTS=--debug rake test
         if: startsWith(matrix.ruby.name, 'jruby')
--- a/test/rubygems/path.rb
+++ b/test/rubygems/path.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'pathname'
+
+def source_root
+  @source_root ||= Pathname.new("../..").expand_path(__dir__)
+end
+
+def test_dir
+  @test_dir ||= source_root.join('test')
+end
+
+def test_rubygems_dir
+  @test_rubygems_dir ||= test_dir.join('rubygems')
+end
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -567,6 +567,7 @@ test/rubygems/invalidchild_key.pem
 test/rubygems/package/tar_test_case.rb
 test/rubygems/packages/ascii_binder-0.1.10.1.gem
 test/rubygems/packages/ill-formatted-platform-1.0.0.10.gem
+test/rubygems/path.rb
 test/rubygems/plugin/exception/rubygems_plugin.rb
 test/rubygems/plugin/load/rubygems_plugin.rb
 test/rubygems/plugin/standarderror/rubygems_plugin.rb
--- a/Rakefile
+++ b/Rakefile
@@ -63,6 +63,15 @@ Rake::TestTask.new do |t|
   t.test_files = FileList['test/**/test_*.rb']
 end
 
+namespace 'test' do
+  desc 'Run each test isolatedly by specifying the relative test file path'
+  task 'isolated' do
+    FileList['test/**/test_*.rb'].each do |file|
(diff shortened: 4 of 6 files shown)


junaruga:
Any maintainers, could you review about the way to fix this issue for the current PR? If you agree on this way, I will work on other test files that have this issue too. The files maybe be the ones listed at https://github.com/rubygems/rubygems/issues/5411#issuecomment-1081921621 .

.github/workflows/ubuntu-rubygems.yml
```diff
@@ -44,6 +44,9 @@ jobs:
       - name: Run Test
         run: rake test
         if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
+      - name: Run Test isolatedly
+        run: rake test:isolated
+        if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
```
junaruga, line 49: We may not need to run this test in all the CRubies. Running on the latest stable CRuby might be enough good like this.

```
diff --git a/.github/workflows/ubuntu-rubygems.yml b/.github/workflows/ubuntu-rubygems.yml
index 5bc7ff8365..29cf6440f7 100644
--- a/.github/workflows/ubuntu-rubygems.yml
+++ b/.github/workflows/ubuntu-rubygems.yml
@@ -46,7 +46,7 @@ jobs:
         if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
       - name: Run Test isolatedly
         run: rake test:isolated
-        if: "!startsWith(matrix.ruby.name, 'truffleruby') && !startsWith(matrix.ruby.name, 'jruby')"
+        if: "!startsWith(matrix.ruby.name, '3.1')"
       - name: Run Test (JRuby)
         run: JRUBY_OPTS=--debug rake test
         if: startsWith(matrix.ruby.name, 'jruby')
```

simi:
> Any maintainers, could you review about the way to fix this issue for the current PR? If you agree on this way, I will work on other test files that have this issue too. The files maybe be the ones listed at [#5411 (comment)](https://github.com/rubygems/rubygems/issues/5411#issuecomment-1081921621) .

Hello. I'm ok with this solution. This blocked me to continue working on Rust builder improvements.

I'm not super sure we need that isolated setup on CI. @deivid-rodriguez is CI overlord and can make decision. Isn't there any other way to enforce `__FILE__` is not used in specs?

test/rubygems/path.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'pathname'
+
+def source_root
```
simi, line 5: What about to keep it in module (similar to Bundler setup) to not "pollute" global (actually Kernel) "namespace"?

deivid-rodriguez:
Let me have a look at this now.

deivid-rodriguez (commented):
I think this is due to https://bugs.ruby-lang.org/issues/16978 (`__FILE__` not using realpaths), which is not an easy fix in a backwards compatible way as many scripts out there rely on things like

```ruby
if __FILE__ == $0
  do_something
end
```

The fix proposed here (rely on `__dir__` rather than on `__FILE__`) makes sense to me.

But I think we can update locations in place instead of adding a helper file and enforce proper `__dir__` usages by enabling the [`Style/ExpandPathArguments` RuboCop cop](https://docs.rubocop.org/rubocop/cops_style.html#styleexpandpatharguments).

Regarding a test that runs test files in isolation, it seems fine to me too, and I agree it's enough to run it only on the latest Ruby.

PR #5456, opened 2022-04-06
Avoid another crash in test `teardown`

deivid-rodriguez:

Confusing errors in CI when some problem happens during test setup. See for example https://github.com/rubygems/rubygems/actions/runs/2104061533/attempts/1. This is quite similar to #5424.

Avoid the crash in `teardown` by making sure the `@tempdir` variable is set as early as possible.

### DIFF
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -308,6 +308,9 @@ class Gem::TestCase < Test::Unit::TestCase
 
     FileUtils.mkdir_p @tmp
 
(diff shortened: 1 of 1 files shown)


indirect (approved):

deivid-rodriguez:
I had to remove the change to avoid `Dir.chdir` because it broke tests on JRuby for some reason. I'll investigate it, but I removed it from this PR for now since it was not related anyways.

PR #5459, opened 2022-04-07
Misc cargo builder improvements

ianks:
Hey there!

After battle testing the cargo code which [making Rust play nice with rake-compiler-dock](https://github.com/ianks/rb-rust-gem), I've come up with some improvements that will fix some bugs, and make the development experience a bit better all around. Here's some of the things I adjusted

1. Make sure we can find binaries when `CARGO_BUILD_TARGET` is detected
2. Iron out some kinks when parsing link flags, namely the `-l:namespec` form
3. Add a configurable `runner` for the extension builder, so I can print out logs more easily when debugging / developing
4. Set all of the `RbConfig::CONFIG` values and environment variables when invoking `cargo rustc`, so gem extensions do not have to figure out have to invoke the correct ruby binary to print out the correct config

With all of these changes, I am able to cross compile Rust extensions on the following platforms (ruby 2.4, 2.5, 2.6, 2.7, 3.0, 3.1):

- x86_64-linux
- aarch64-linux
- arm-linux
- x86_64-darwin
- arm64-darwin
- x64-mingw32
- x64-mingw-ucrt


lib/rubygems/ext/cargo_builder.rb
```diff
@@ -4,26 +4,26 @@
...
 
-  def initialize(spec)
+  def initialize(spec, runner: nil)
     @spec = spec
+    @runner = runner || self.class.method(:run)
   end
 
   def build(_extension, dest_path, results, args = [], lib_dir = nil, cargo_dir = Dir.pwd)
     require "rubygems/command"
+    require "rubygems/ext/cargo_builder/link_flag_converter"
```
deivid-rodriguez, line 16: Can we use relative requires for internal files, to be consistent with the rest of the code?

deivid-rodriguez: Thanks! Could you also change the previous `require`? It was not introduced here but in #5175, but I happened to notice it just now.

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -4,26 +4,26 @@
 # over the `cargo rustc` command which takes care of building Rust code in a way
 # that Ruby can use.
 class Gem::Ext::CargoBuilder < Gem::Ext::Builder
-  attr_reader :spec
+  attr_reader :spec, :runner
 
-  def initialize(spec)
+  def initialize(spec, runner: nil)
     @spec = spec
+    @runner = runner || self.class.method(:run)
```
simi, line 11: I think we should not introduce per builder special arguments. For now, I would prefer to remover runner. We can reintroduce it later if needed in base builder class.

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -4,26 +4,26 @@
...
+    build_crate(_extension, dest_path, results, args, cargo_dir)
     ext_path = rename_cdylib_for_ruby_compatibility(dest_path)
     finalize_directory(ext_path, dest_path, lib_dir, cargo_dir)
     results
   end
 
-  private
-
-  def build_crate(dest_path, results, args, cargo_dir)
+  def build_crate(_extension, dest_path, results, args, cargo_dir)
```
simi, line 26: I wasn't able to find out what's the reason for the `_extension` argument since it is not used. Can we remove it?

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -127,42 +145,19 @@ def rustc_dynamic_linker_flags(dest_dir)
...
-    if link_name.include?("ruby")
-      # Specify the lib kind and give it the name "ruby" for linking
-      kind = ruby_static? ? "static" : "dylib"
-
-      ["-l", "#{kind}=ruby:#{link_name}"]
-    else
-      ["-l", link_name]
-    end
+  def ldflag_to_link_modifier(arg)
+    LinkFlagConverter.call(arg)
```
simi, line 160: I think there is no need to create this additional class just to hide one method. Also I'm not big fan of generic `call` methods. IMHO the purpose of `call` method naming convention was to be able to mimic `proc` and its `call` method, which is not used in here. Is there any problem to revive the method back to this class?

test/rubygems/test_gem_ext_cargo_builder.rb
```diff
@@ -58,7 +58,7 @@ def test_build_cdylib
 
     output = output.join "\n"
 
-    bundle = File.join(@dest_path, "release/rust_ruby_example.#{RbConfig::CONFIG['DLEXT']}")
+    bundle = Dir[File.join(@dest_path, "**/release/rust_ruby_example.#{RbConfig::CONFIG['DLEXT']}")].first
```
simi, line 61: Any info on this change?

lib/rubygems/ext/cargo_builder.rb
ianks: I want to be able to isolate it for more robust testing, so having it in a separate class is nice for that. Is there another method name besides `call` you would prefer??

lib/rubygems/ext/cargo_builder.rb
ianks: OK, fair enough. FWIW, The main reason for this is I would like to be able to configure the builder a bit more when cross compiling, and development.

I can probably get away with just inheriting instead of passing in the runner:

```ruby
class DevBuilder < Gem::Ext::CargoBuilder
  def self.run(...)
end
```

ianks: Actually, would you be OK with something like this given I remove the special arguments from the builder?

```ruby
builder = Gem::Ext::CargoBuilder.new(spec)
builder.profile = :dev
```

lib/rubygems/ext/builder.rb
```diff
@@ -48,7 +48,7 @@ def self.make(dest_path, results, make_dir = Dir.pwd)
     end
   end
 
-  def self.run(command, results, command_name = nil, dir = Dir.pwd)
+  def self.run(command, results, command_name = nil, dir = Dir.pwd, env = {})
```
ianks, line 51: When cross-compiling with multiple ruby versions present, it's necessary to pass information about `RbConfig` to cargo so we can ensure proper linking. I found using environment variables is a simple and effective way to do this.

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -4,48 +4,67 @@
...
+  end
 
-    ENV["RUBY_STATIC"] = "true" if ruby_static? && !given_ruby_static
+  def build_env
+    build_env = rb_config_env
+    build_env["RUBY_STATIC"] = "true" if ruby_static? && ENV.key?('RUBY_STATIC')
+    build_env
+  end
 
+  def cargo_command(cargo_dir, dest_path, args = [])
```
ianks, line 43: I want to expose the actual cargo command to run, so it can be used in a Makefile later.

test/rubygems/test_gem_ext_cargo_builder.rb
ianks: Good catch, it was accidental. I just reverted it.

lib/rubygems/ext/cargo_builder.rb
ianks: I renamed the method to `convert_to_cargo_flag`, does that work for you??

simi: looks ok :heart:

simi: > I want to be able to isolate it for more robust testing, so having it in a separate class is nice for that. Is there another method name besides `call` you would prefer??

Since the purpose of the class was to convert something, I would (personally) prefer to use `convert` method for example.

ianks:
Not sure why, but it looks like the Windows actions were interrupted causing CI to ❌. @simi could you trigger a re-run?

test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example/Cargo.toml
```diff
@@ -7,4 +7,4 @@ crate-type = ["cdylib"]
 
 [dependencies]
 # Needed until bindgen has the `allowlist_file` feature
```
simi, line 9: Is this comment still valid?

test/rubygems/test_gem_ext_cargo_builder_unit.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+require_relative 'test_gem_ext_cargo_builder'
+require 'rubygems/ext'
+
+class TestGemExtCargoBuilderUnit < Gem::TestCase
+  def test_cargo_command_passes_args
+    skip_unsupported_platforms!
```
simi, line 9: Isn't it possible to move this to `setup` since it is shared for all tests? Does skipping of tests work in there?

test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example/Cargo.toml
ianks: yes it is, will update once it isn't, which is soon hopefully since it's merged into main!

test/rubygems/test_gem_ext_cargo_builder_unit.rb
ianks: done!

ianks: tried this, and unfortunately it did not work :/

simi: I wasn't sure, sorry for confusion.

test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example/Cargo.toml
ianks: Updated

simi: Nice!

lib/rubygems/ext/builder.rb
simi: If I understand it well, this argument is not used directly by `RubyGems` now. Can you confirm?

deivid-rodriguez:
@ianks Sorry for dropping the ball here, I will do my best to get this reviewed and shipped with the next release!

lib/rubygems/ext/cargo_builder.rb
deivid-rodriguez: Not sure what @simi thanks, but it feels ok to me.

lib/rubygems/ext/builder.rb
deivid-rodriguez: As far as I understand, all RubyGems does is pass on the given environment to `cargo`. This feels nit to me, because it's more explicit than setting up the environment globally just so it is inherited by the `cargo` subprocess.

ianks:
@deivid-rodriguez Should be good to go.

test/rubygems/test_gem_ext_cargo_builder.rb
```diff
@@ -57,15 +55,35 @@ def test_build_cdylib
...
 
     output = output.join "\n"
-
     bundle = File.join(@dest_path, "release/rust_ruby_example.#{RbConfig::CONFIG['DLEXT']}")
 
-    require(bundle)
+    assert_match "Finished release [optimized] target(s)", output
+    assert_ffi_handle bundle, 'Init_rust_ruby_example'
+  rescue Exception => e
+    pp output if output
```
simi, line 63: Is this intended to be left in here?

test/rubygems/test_gem_ext_cargo_builder/custom_name/Cargo.toml
```diff
@@ -6,5 +6,4 @@ version = "0.1.0"
 crate-type = ["cdylib"]
 
 [dependencies]
-# Needed until bindgen has the `allowlist_file` feature
-rb-sys = { git = "https://github.com/ianks/rb-sys", tag = "v0.9.0" }
+rb-sys = "0.9.3"
```
simi, line 9: :muscle: finally :)

simi:
I have left minor comments. Otherwise looks good to me.

test/rubygems/test_gem_ext_cargo_builder.rb
ianks: Yes, it really does make all of the difference when debugging. I can remove it if you prefer though!

lib/rubygems/ext/builder.rb
ianks: RubyGems does not use this, other than Cargo builder. It allow us to pass env variables without setting them globally for the entire process.

test/rubygems/test_gem_ext_cargo_builder.rb
deivid-rodriguez: Looks strange to me too, but I can see how it's useful, and the code was already there, so ok!

deivid-rodriguez (approved):
If @simi is fine with this, I'm fine too :)

PR #5469, opened 2022-04-13
Add modern rubies as valid platform values in Gemfile DSL

BobbyMcWho:

Attempting to limit a gem version to a specific version on ruby 3.0, realized that :ruby_30 was not a supported platform. 

Add modern rubies to the platforms and known Major versions.


simi:
Hello, thanks for keeping this list updated. It seems super easy to get this list out-of-date. Any plans @deivid-rodriguez on keeping this updated? Maybe we can use some CI mojo and ensure all CI versions have proper entry in those lists at least.

deivid-rodriguez:
It seems easy enough to add an extra test (in addition to the hardcode list of platforms that we check to be defined) to check that also the running ruby has a corresponding platform value defined?

simi:
@deivid-rodriguez ok to merge this first?

deivid-rodriguez:
Sure!

PR #5470, opened 2022-04-13
Fix formatting in docs

peterzhu2118:

rdoc uses + for typewriter font rather than backticks. This causes incorrect formatting in the generated docs.

Use + instead of backticks.


deivid-rodriguez (approved):
Thanks, will try to remember this in the future, since I guess there's no easy way to automatically enforce this? Maybe a RuboCop cop, `NoBackticksInDocumentationHeader` or something could help.

peterzhu2118:
Unfortunately I don't know of a way to detect this automatically (using either rubocop or rdoc).

PR #5472, opened 2022-04-15
Stop considering `RUBY_PATCHLEVEL` for resolution

deivid-rodriguez:

Bundler still has some code that considers `RUBY_PATCHLEVEL` during resolution. However, this number is no longer meaningful since Ruby 2.1 when Ruby started release patch level versions. Now the `RUBY_VERSION` properly identifies a given Ruby release so it's no longer necessary to consider the patchlevel.

This generally doesn't cause any issues, but the patchlevel does leak to some error messages, which makes them quite confusing. For example, if you set `ruby "3.1.1"` in your `Gemfile`, you may end up with an error message including the following

```
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    Ruby (~> 3.1.1.0)

  ...
```

My fix is to stop considering patchlevels for modern rubies. With this patch, the above error now reads better

```
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    Ruby (= 3.1.1)

  ...
``` 

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -746,17 +746,8 @@ module Bundler
 
     def ruby_version_requirements(ruby_version)
       return [] unless ruby_version
-      if ruby_version.patchlevel
-        [ruby_version.to_gem_version_with_patchlevel]
-      else
-        ruby_version.versions.map do |version|
-          requirement = Gem::Requirement.new(version)
-          if requirement.exact?
-            "~> #{version}.0"
-          else
-            requirement
-          end
-        end
+      ruby_version.versions.map do |version|
+        Gem::Requirement.new(version)
       end
     end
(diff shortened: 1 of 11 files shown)


indirect (approved):
Nice work! This seems like a good change. And cleaning up code that only matters on Ruby versions we have stopped supporting always seems worthwhile. :)

deivid-rodriguez:
Thanks @indirect, it was a bit tricky to make this backwards compatible, but it should be ok now. I extracted some changes to a separate PR to RubyGems at #5473. I'll rebase this one after merging #5473 and CI should be fine.

deivid-rodriguez:
Nevermind, I need to introduce changes in RubyGems and Bundler at the same time, so I closed the other PR. Will merge after CI.

PR #5474, opened 2022-04-16
Improve Ruby version resolution conflicts

deivid-rodriguez:

Right now, if Bundler can't find a version of a requirement whose `required_ruby_version` satisfies the running Ruby version, it will print a resolution error. However, this error is quite confusing because even if there's no `ruby` directive in the `Gemfile`, it will read like this:

```
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    activeadmin was resolved to 2.11.2, which depends on
      Ruby (>= 3.1)

    Ruby
```

There's two issues here:

* The error mentions that the requirement on the Ruby version is set in the `Gemfile`. However, that's not always the case.
* The error does not mention the current Ruby version at all, which is the real culprit.

This PR fixes the issue by:

* Always passing the current Ruby version explicitly to the resolver, so that errors always include that.
* Customizing Molinillo's error building logic to not mention the `Gemfile` at all when the conflict happened with a "metadata" dependency. This requires an extra feature to Molinillo which I will propose separately upstream.

Now the error reads like this

```
Bundler found conflicting requirements for the Ruby version:
  In Gemfile:
    activeadmin was resolved to 2.11.2, which depends on
      Ruby (>= 3.1)

  Current Ruby version:
    Ruby (= 2.7.5)
```
 
### DIFF
--- a/bundler/Rakefile
+++ b/bundler/Rakefile
@@ -159,9 +159,11 @@ end
 load "../util/automatiek.rake"
 
 # We currently ship Molinillo master branch as of
-# https://github.com/CocoaPods/Molinillo/commit/b4a510349e6893486b9eb44a71734c05a8034256,
-# with one change on top to require tsort relatively to use our vendored
-# version.
+# https://github.com/CocoaPods/Molinillo/commit/b4a510349e6893486b9eb44a71734c05a8034256
+# plus the following patches:
+# * Require `tsort` relatively to use our vendored version.
+# * Allow full customization of Molinillo's conflict error message
+#   (https://github.com/CocoaPods/Molinillo/pull/164).
 desc "Vendor a specific version of molinillo"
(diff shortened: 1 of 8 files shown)


deivid-rodriguez:
I created an upstream PR to Molinillo: https://github.com/CocoaPods/Molinillo/pull/164.

And since I was at it I also updated vendored Molinillo and TSort to latest versions which made me create a few PRs upstream. I will update this PR once everything is in sync.

deivid-rodriguez:
This patch to Molinillo seems harmless to me, so I will merge this. Hopefully it will be upstreamed soon.

deivid-rodriguez:
This patch to Molinillo seems harmless to me, so I will merge this. Hopefully it will be upstreamed soon.

PR #5475, opened 2022-04-16
Less error swallowing when installing gems

deivid-rodriguez:

I'm working on some changes to `fileutils` to not swallow meaningful errors from `FileUtils.rm_f` and `FileUtils.rm_rf` at https://github.com/ruby/fileutils/pull/58.

Trying that PR against RubyGems tests made some errors (mostly likely typos where `FileUtils.rm_f` instead of `FileUtils.rm_rf` was used).

Use the proper call so that the folders are actually removed, and fix the issues caused by that.

While debugging those errors, I also removed one `rescue` that felt unnecessary and helped me track down the original culprit of the error.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -285,8 +285,6 @@ class Gem::Installer
 
   def spec
     @package.spec
-  rescue Gem::Package::Error => e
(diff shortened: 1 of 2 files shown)


simi (approved):

deivid-rodriguez:
Thanks @simi!

PR #5481, opened 2022-04-20
Increase timeout when testing RubyGems on Windows and MacOS

deivid-rodriguez:

Since we added it, this job has been sporadically timing out. See for example https://github.com/rubygems/rubygems/runs/6094008084?check_suite_focus=true.

. Given the little frequency when it happens, I guess 5 more minutes is enough.

### DIFF
--- a/.github/workflows/macos-rubygems.yml
+++ b/.github/workflows/macos-rubygems.yml
@@ -40,4 +40,4 @@ jobs:
         run: rake setup
       - name: Run Test
         run: rake test
-    timeout-minutes: 15
+    timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I think Ruby 3.1 is somewhat slower (at least setting it up takes about 1 minute vs a few seconds of other rubies). But it turns out that timeout are happening across all rubies. I'm pretty sure that this is due to adding support for rust extensions since I also noticed locally that the test suite is much slower now.

Let's relax the timeouts as needed.

deivid-rodriguez:
This just made some jobs get cancelled at #5486. I'll merge to see if it helps.

Issue #5422, opened 2022-03-23
NoMethodError: undefined method `full_name' for nil:NilClass bug that should have been fixed


jugdx:
Hello there 👋🏼 ,

### Context 

I post here an error that should have been fixed since December last year, but it seems that we can still have it.
[Here](https://github.com/rubygems/rubygems/pull/5092) is the fix I'm talking about.
This should have been integrated in the [2.2.33](https://github.com/rubygems/rubygems/releases/tag/bundler-v2.2.33) version. We are using the [2.3.9](https://github.com/rubygems/rubygems/releases/tag/bundler-v2.3.9) one so the fix should still be there...

### Additional information

This is for an iOS application using `Fastlane`, we have lots of script in Ruby + Shell + YML for Github Actions.
We are migrating from Bitrise to Github so I'm testing our `Release` flow.
The error happen when we execute the `bundle update fastlane`.

I'm posting into the following toggles all the details I have about this.

<details>
<summary>Part of the Release.yml (Github Action)</summary>

```yml
generate_release:
    name: 'Create new release'
    runs-on: macos-11
    outputs:
      jiraIssueKey: ${{ env.ISSUE_KEY }} # Env Variable set at the end of the release flow, via Fastlane's "_release_ci" lane
      rcNumber: ${{ env.RC_NUMBER }} # Env Variable set at the end of the release flow, via Fastlane's "_release_ci" lane

    steps:
      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
      - uses: actions/checkout@v2

      - name: Setup SSH Agent
        uses: webfactory/ssh-agent@v0.5.3
        with:
          ssh-private-key: ${{ secrets.SSH_KEY_XX }} 

      - name: Setup PHP
        uses: shivammathur/setup-php@v2
        with:
          php-version: '7.4'

      # Runs a single command using the runners shell
      - name: Setup Ruby
        uses: ruby/setup-ruby@v1
        with:
          bundler-cache: true

      - name: Setup Project
        run: |
          ci/_setup.sh
          ci/set-slack-target.sh

      - name: Execute release flow
        env:
          RELEASE_VERSION: ${{ github.event.inputs.releaseVersion }}
          RELEASE_OWNER_EMAIL: ${{ github.event.inputs.releaseOwnerEmail }}
        run: ci/_release.sh

      - name: Publish Artifacts
        uses: actions/upload-artifact@v2
        with:
          name: artifacts-appstore
          path: iOS-submission/

[...]

PR #5490, opened 2022-04-22
Fix check for missing spec in check_for_unmet_dependencies

Eusebius1920:

See #5422

Add a check that catches the condition and handles it properly, see https://github.com/rubygems/rubygems/issues/5422#issuecomment-1106434053

### DIFF
--- a/bundler/lib/bundler/installer/parallel_installer.rb
+++ b/bundler/lib/bundler/installer/parallel_installer.rb
@@ -121,7 +121,12 @@ module Bundler
 
       unmet_dependencies.each do |spec, unmet_spec_dependencies|
         unmet_spec_dependencies.each do |unmet_spec_dependency|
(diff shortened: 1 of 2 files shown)


Eusebius1920:
- [x]  code linting should be fine (rubocop does not complain)
- [x] fixed / enhanced existing tests to comply with changes

- [ ] Running the test-suite locally atm, but I already can see some Failures

sadfuzzy:
could this be merged?

deivid-rodriguez:
Hi!

We never followed up on this because the original issue was no longer reproducible, although improving the error message in case the problem comes up again won't hurt.

What problem are you having?

deivid-rodriguez:
Actually I'm pretty sure that with recent changes in our we materialize specs, this can no longer be triggered. So I'm going to close this with a big thanks to @Eusebius1920 for starting this work and helping figuring out root causes.

@sadfuzzy Please open a ticket with whatever issue you're having and we can figure it out from there!

Thanks both!!

PR #5497, opened 2022-04-26
Clarify description of owner-flags

kronn:

I wanted to add a coworker as owner to a gem and was confused by the help-output. In the end, I tried to add the new owner by rubygems-handle and it worked. The docs and description only mention using an e-mail address.

I remember that other experienced developers also confused that and were asking for e-mails when the handle would have been sufficient.

I changed the help-output to clarify what is accepted as argument to the respective flags. I did not write tests because the behavior has not been changed.

### DIFF
--- a/lib/rubygems/commands/owner_command.rb
+++ b/lib/rubygems/commands/owner_command.rb
@@ -15,6 +15,10 @@ The owner command lets you add and remove owners of a gem on a push
 server (the default is https://rubygems.org). Multiple owners can be
 added or removed at the same time, if the flag is given multiple times.
 
+The supported user identifiers are dependant on the push server.
+For rubygems.org, both e-mail and handle are supported, even though the
+user identifier field is called "email".
+
 The owner of a gem has the permission to push new versions, yank existing
 versions or edit the HTML page of the gem.  Be careful of who you give push
 permission to.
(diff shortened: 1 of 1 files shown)


simi:
Hello. Well spotted.

This really depends on implementation of gemserver. RubyGems.org is friendly enough to support both handle and email (since this method is used -> https://github.com/rubygems/rubygems.org/blob/a4634ee7626e837f9b776f966c8b97b55f72ec8b/app/models/user.rb#L80). 

That's probably the reason for having so general description in there.

Would it be possible to rewrite it to mention you need to specifiy user identifier and also mention RubyGems.org supports handle and email?

kronn:
> That's probably the reason for having so general description in there.

I found the naming "email" a bit misleading, though I understand now that this comes from the API. 

> Would it be possible to rewrite it to mention you need to specifiy user identifier and also mention RubyGems.org supports handle and email?

I generalized the description a bit more and added a note about the specific handling in the push server "rubygems.org".

kronn:
Oh noez! Timeout with this check:

    bundler / Bundler 2 on MacOS (ruby-3.1) (pull_request) Cancelled after 60m

Is there anything I can do?

deivid-rodriguez:
@simi Does this address your concerns? It seems like it does, so I plan to merge it.

simi (approved):

PR #5513, opened 2022-05-04
Update man page for `require` option in `bundle add` command

nobuyo:

The `bundle add -h` does not mentioned to `require` option with bundler 2.3.12.

I've updated man page for `bundle add`.

- [ ] ~Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes~
- [ ] ~Write code to solve the problem~


deivid-rodriguez (approved):
Thank you!

PR #5519, opened 2022-05-09
Cleanup old legacy code

deivid-rodriguez:

This code was first added a long time ago (Bundler 1.0.8), because we were using a deprecated RubyGems method that was printing warnings that confused users. The relevant commit is https://github.com/rubygems/bundler/commit/fb3a08191f78c0b56742526f4fc454291aea8988.

The old method is now long gone since https://github.com/rubygems/rubygems/commit/1bff87f83b8de70a13cae92087fdc636cb4549d2 (RubyGems 1.6.0, 11 years ago), so this is now dead code. Also the tests (at least in their current wording) no longer test anything meaningful since the method producing the old warnings is gone, so I went ahead and removed them too.

### DIFF
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -283,11 +283,7 @@ module Bundler
 
           e = Gem::LoadError.new(message)
           e.name = dep.name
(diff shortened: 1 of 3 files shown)


hsbt (approved):
👍

PR #5521, opened 2022-05-09
Fully lock rake version in CI

deivid-rodriguez:

Our CI and local development commands would inconsistently be run through `bin/rake` or `rake`. This is because we can't use `bin/rake` for tasks that install dependencies, because the exact `rake` version they need might not yet be installed.

This PR adds some magic to `bin/rake`, when tasks to install dependencies are run, so that the locked version of `rake` is also installed automatically if not present.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -61,13 +61,13 @@ jobs:
       - name: Install graphviz (MacOS)
         run: brew install graphviz
         if: matrix.bundler.value == '' && matrix.os.name == 'MacOS'
+      - name: Prepare dependencies
+        run: |
+          bin/rake spec:parallel_deps
+        working-directory: ./bundler
       - name: Replace version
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
I was testing some Bundler stuff on Ruby 2.4 (which includes rake 12 by default), and got the following error when running `bin/rake spec:parallel_deps`:

> We couldn't activate rake (~> 13.0). Run `gem install rake:'~> 13.0'`

With this PR, it just works, so I will merge it.

deivid-rodriguez:
I was testing some Bundler stuff on Ruby 2.4 (which includes rake 12 by default), and got the following error when running `bin/rake spec:parallel_deps`:

> We couldn't activate rake (~> 13.0). Run `gem install rake:'~> 13.0'`

With this PR, it just works, so I will merge it.

simi (approved):

Issue #5364, opened 2022-02-18
Bundler's inline `install` option misdocumented


ngan:
According to the [source documentation](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/inline.rb#L11-L13), the `install` argument specifies whether or not installation of missing gems should happen or not. However, looking at the code, it looks like installation will [_always_ happen for missing gems](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/inline.rb#L61). The only thing that the flag does is determines whether to [quiet the installation message or not](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/inline.rb#L60).

Considering that it already [recognizes a `quiet` option](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/inline.rb#L37), perhaps we should just fix the logic to:

```ruby
if install && definition.missing_specs?
```

ngan:
@deivid-rodriguez does this make sense? Would be happy to open PR.

deivid-rodriguez:
@ngan Sorry I never left feedback here.

Since `gemfile(&gemfile)` already behaves essentially the same as `gemfile(true, &gemfile)`, wouldn't it make more sense to deprecate the `install` parameter, since `bundler/inline` installs what's necessary regardless?

The patch you suggest changes (misdocumented, that's true) behavior because people used to `gemfile(&gemfile)` auto-installing as necessary will no longer get that behavior.

tonyaraujop:
Hello there @ngan have you started to work on this? If not, I would like to help.

ngan:
Hey @tonyaraujop I started this but didn't have time to finish. It's fairly straightforward. I think where I got stuck was how to how to show the deprecation message. Summer distract me 😓 

Feel free to have a go at it!

tonyaraujop:
Thanks.

tonyaraujop:
Hey @ngan I looked in files and found `deprecate.rb` under `lib/rubygems` that has a method `rubygems_deprecate` which `Indicate something will be removed in the next major RubyGems version, and (optionally) a replacement for it.` So I just added this line `rubygems_deprecate(install)` just under the line `definition.validate_runtime!`. Do you think this is enough? I tried to test in `inline_spec.rb` and when I use `rake test`, all tests passes. I feel that something is wrong but I can tell where.

jeromedalbert:
[...]

PR #5529, opened 2022-05-10
Handle inline gemfile activated gem conflicts

jonathanhefner:

When using an inline gemfile that performs installation (i.e. `gemfile(true) {}`), newly installed gems can conflict with gems already activated by Bundler.

For example, when running [this Rails bug report template](https://github.com/rails/rails/blob/88c0e2c9507f0725bdffea9ada405f4d36756d42/guides/bug_report_templates/action_controller_gem.rb) with `strscan` v3.0.1 (but not v3.0.2) installed, `strscan` v3.0.1 will be activated, then v3.0.2 will be installed as a transitive dependency, then Bundler will try to activate v3.0.2 and raise an error.  If you then run the script a 2nd time, it will work (because it immediately activates `strscan` v3.0.2, installed by the 1st run).

This is currently causing the [Rails build to fail](https://buildkite.com/rails/rails/builds/86305#9be4d24a-2944-4660-b5f0-03aac7992131/1158-1265) (see also rails/rails#45052).

I was working on this solution, but then I realized that [`@definition.dependencies`](https://github.com/jonathanhefner/rubygems/commit/98be36a95b450522f61354e78fae8cc55f792d50#diff-4a2872c36bdaa433a9d117937197a04fe83603a9c045349677dba5a9e1faf2a2R292) is not recursive.  I am not familiar enough with the Bundler codebase to know what the best solution would be.  Here are possible some alternatives:

1. Don't install gems that have already been activated.  This solution is difficult because the installation process itself activates gems.  Namely, [`Bundler::Installer.install`](https://github.com/rubygems/rubygems/blob/3f39b2fc2dcd68e0b7f0adc84cfa08c2794747be/bundler/lib/bundler/inline.rb#L63) appears to activate `strscan`.  (In the version of Bundler I had installed, it was via an eventual call to [`require_relative "psyched_yaml"`](https://github.com/rubygems/rubygems/blob/84fd3ca03e9b1bc9d32c08cf7468bc2bf8e936cc~1/bundler/lib/bundler/rubygems_integration.rb#L108).)

2. Change the [error message](https://github.com/rubygems/rubygems/blob/3f39b2fc2dcd68e0b7f0adc84cfa08c2794747be/bundler/lib/bundler/runtime.rb#L294-L302) to suggest re-running the script when an inline gemfile performs installation.

3. Downgrade the error to a warning when the spec is for a gem that was just installed by an inline gemfile.

### DIFF
--- a/bundler/spec/runtime/inline_spec.rb
+++ b/bundler/spec/runtime/inline_spec.rb
@@ -355,6 +355,26 @@ RSpec.describe "bundler/inline#gemfile" do
     expect(out).to include("BUNDLE_GEMFILE is empty")
   end
 
+  it "uses already activated specs if they satisfy dependencies" do
+    script <<-RUBY
+      gemfile do
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack", "0.9.1"
+      end
+
+      gemfile(true) do
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack", "<= 1.0.0"
+      end
+
+      puts RACK
+    RUBY
+
+    expect(out).to include("Installing rack 1.0.0")
+    expect(out.lines.last).to eq("0.9.1")
+    expect(err).to be_empty
+  end
+
   it "does not error out if library requires optional dependencies" do
     Dir.mkdir tmp("path_without_gemfile")
 
--- a/bundler/lib/bundler/inline.rb
+++ b/bundler/lib/bundler/inline.rb
@@ -68,7 +68,7 @@ def gemfile(install = false, options = {}, &gemfile)
       end
 
       runtime = Bundler::Runtime.new(nil, definition)
-      runtime.setup.require
+      runtime.setup(:allow_compatible_activated_specs => true).require
     end
   ensure
     if bundler_module
--- a/bundler/lib/bundler/runtime.rb
+++ b/bundler/lib/bundler/runtime.rb
@@ -9,13 +9,14 @@ module Bundler
       @definition = definition
     end
 
-    def setup(*groups)
+    def setup(*groups, allow_compatible_activated_specs: false)
       @definition.ensure_equivalent_gemfile_and_lockfile if Bundler.frozen_bundle?
 
       # Has to happen first
       clean_load_path
 
       specs = @definition.specs_for(groups)
+      specs = exclude_specs_with_compatible_activated_spec(specs) if allow_compatible_activated_specs
 
       SharedHelpers.set_bundle_environment
       Bundler.rubygems.replace_entrypoints(specs)
@@ -285,6 +286,18 @@ module Bundler
       end
 
       output
+    end
+
+    def exclude_specs_with_compatible_activated_spec(specs)
+      dependencies_by_name = (@definition.dependencies + specs.flat_map(&:dependencies)).group_by(&:name)
+
+      specs.reject do |spec|
+        next unless activated_spec = Bundler.rubygems.loaded_specs(spec.name)
+
+        dependencies_by_name[spec.name]&.all? do |dependency|
+          dependency.requirement.satisfied_by?(activated_spec.version)
+        end
+      end
     end
 
     def check_for_activated_spec!(spec)


deivid-rodriguez (commented):
Hello @jonathanhefner, thanks for starting this, this one is tricky!

This kind of errors are somewhat common in auto-install mode (both with inline Gemfiles and also with the `BUNDLE_AUTO_INSTALL` setting), because there's quite a bit of code that gets run before `Bundler.setup` time that may activate gems.

We try very hard to not depend on any gems, but sometimes we can't avoid it. In particular, this one [was easy to fix](https://github.com/ruby/psych/pull/534), however, until that's released and all users pick up the new version, it won't have any effect here, so we should provide some workaround.

There's no great solution to this and your approach looks like a good compromise. To make it work, you would have to verify all dependencies of all resolved specs, and verify that none of them is incompatible with the activated version (your current approach only looks into top level gem file dependencies, as you noticed).

bundler/spec/runtime/inline_spec.rb
```diff
@@ -355,6 +355,26 @@ def confirm(msg, newline = nil)
...
+    script <<-RUBY
+      gemfile do
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack", "0.9.1"
+      end
+
+      gemfile(true) do
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack", "<= 1.0.0"
+      end
```
jonathanhefner, line 368: `gemfile(true)` shouldn't be necessary here, but when using only `gemfile` (no `(true)`), I can't get Bundler to (re)resolve `gem "rack", "<= 1.0.0"` to `1.0.0`.  (i.e. It resolves to `0.9.1` even when `allow_compatible_activated_specs` is false.)

deivid-rodriguez:
Hei @jonathanhefner!

After having a second look at this, I realized that having to extend `Bundler::Runtime#setup` signature just to fix this admittedly weird edge case feels not right to me.

I came up with another approach that feels better. From `bundler/inline`, rescue this kind of issue and in that case, retry resolution with an added requirement on the exact spec that got activated by Bundler while installing gems. This is [a proof of concept](https://github.com/rubygems/rubygems/commit/b9a9e34fa76b96391d06a5e5e347cbcb3567417b).

What do you think?

jonathanhefner:
@deivid-rodriguez My knowledge about Bundler's internals is limited, but that also seems like a reasonable approach.

It would run once for each conflict (vs once for all conflicts in this PR), but that probably doesn't matter.

Are there other circumstances in which a `Gem::LoadError` could be raised?  Should there be guard like `raise unless install` or `raise unless activated_version` (or both)?

Minor nits:

* You could move `runtime = Bundler::Runtime.new(nil, definition)` inside `begin` and out from `rescue`, since you are reassigning `definition` in `rescue`.
* `Gem.loaded_specs[e.name]` => `Gem.loaded_specs[name]`.
* `Bundler will now retry to re-resolve using` => `Bundler will now try to re-resolve using` or `Bundler will now retry resolving with`.

deivid-rodriguez:
Hi @jonathanhefner! Nice comments.

> Are there other circumstances in which a `Gem::LoadError` could be raised?

Uggh, that's a nice catch! I though I was only wrapping the `runtime.setup` call, which only sets up the `$LOAD_PATH`, so I don't think it can raise `Gem::LoadError` except for this situation. But since I'm also wrapping `.require`, I'm not so sure it's safe. I think we can scope this block to `runtime.setup` only, and leave the `require` step for "after success". I will update this.

> Should there be guard like `raise unless install` or `raise unless activated_version` (or both)?

I thought about the latter, to avoid possibly entering an infinite loop, but my conclusion was that a second raise for the same reason is not possible because I'm explicitly locking the resolution to the gem version that was activated, which is exactly the check that can raise this error.

Regarding `raise unless install`, with the current code, the installer might actually be invoked even if `install` is not passed, so I think we should do this recovery regardless? In fact, we might [deprecate the `install` flag](https://github.com/rubygems/rubygems/issues/5364) because it's misleading.

> Minor nits:

Yes, all three suggestions make sense to me, I will apply them, thanks!

jonathanhefner:
> ...only wrapping the `runtime.setup` call, which only sets up the `$LOAD_PATH`, so I don't think it can raise `Gem::LoadError` except for this situation. ...I think we can scope this block to `runtime.setup` only...

My knee-jerk reaction is "circumstances may change in the future."  In other words, what would happen if `runtime.setup` raises `Gem::LoadError` for some other reason in the future?

I feel like including `raise unless activated_version` (or, more specifically, `raise unless Gem.loaded_specs[name]`) would be safer, even if we only wrap `runtime.setup`.  Or, we could raise a more specific error [in `check_for_activated_spec!`](https://github.com/rubygems/rubygems/blob/3f39b2fc2dcd68e0b7f0adc84cfa08c2794747be/bundler/lib/bundler/runtime.rb#L301), and rescue that.

But, again, my knowledge of Bundler internals is limited, so take my opinion with a grain of salt.

deivid-rodriguez:
`Bundler::Runtime#setup` is not meant or allowed to ever load/activate any gems, only setup the `$LOAD_PATH`. That should be a well established contract I believe. Also, note that this is not a standard `LoadError`, it inherits from it to signal exactly this kind of situation. But maybe it makes sense to still be conservative, I'll think about it.

hsbt:
Closing favor of #5535

PR #5532, opened 2022-05-12
Memoize some #hash calls

ngan:

`require 'bundler/setup` is slow for large Gemfiles (eg 500+ gems)

Memoize some `#hash` implementations since they don't change and we do it a lot for large gem files.

**Before**
```
➜  time ruby -e 'require "bundler/setup"' 
1.49s user 0.11s system 97% cpu 1.645 total
➜  time ruby -e 'require "bundler/setup"'
1.50s user 0.12s system 96% cpu 1.668 total
➜ time ruby -e 'require "bundler/setup"'
1.50s user 0.12s system 96% cpu 1.678 total
```

**After**
```
➜  time ruby -e 'require "bundler/setup"'
0.95s user 0.12s system 94% cpu 1.136 total
➜  time ruby -e 'require "bundler/setup"'
0.95s user 0.12s system 94% cpu 1.125 total
➜ time ruby -e 'require "bundler/setup"'
0.96s user 0.12s system 95% cpu 1.128 total
```

About 30% less time.

### DIFF
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -35,7 +35,7 @@ module Bundler
     end
 
     def hash
-      identifier.hash
+      @__hash ||= identifier.hash
     end
 
     ##
--- a/bundler/lib/bundler/plugin/api/source.rb
+++ b/bundler/lib/bundler/plugin/api/source.rb
@@ -221,7 +221,7 @@ module Bundler
         # docstring for `==` method, i.e. two methods equal by above comparison
         # should have same hash.
         def hash
-          [self.class, uri].hash
+          @__hash = [self.class, uri].hash
         end
 
         # A helper method, not necessary if not used internally.
--- a/bundler/lib/bundler/resolver/spec_group.rb
+++ b/bundler/lib/bundler/resolver/spec_group.rb
@@ -67,7 +67,7 @@ module Bundler
       end
 
       def hash
-        name.hash ^ version.hash ^ sorted_activated_platforms.hash ^ source.hash
+        @__hash ||= name.hash ^ version.hash ^ sorted_activated_platforms.hash ^ source.hash
       end
 
       protected
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -177,10 +177,12 @@ module Gem
     class Requirement
       module CorrectHashForLambdaOperator
         def hash
-          if requirements.any? {|r| r.first == "~>" }
-            requirements.map {|r| r.first == "~>" ? [r[0], r[1].to_s] : r }.sort.hash
-          else
-            super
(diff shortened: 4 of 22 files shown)


bundler/lib/bundler/lazy_specification.rb
```diff
@@ -35,7 +35,7 @@ def eql?(other)
     end
 
     def hash
-      identifier.hash
+      @__hash ||= identifier.hash
```
ngan, line 38: This is the money maker right here.

ngan:
@deivid-rodriguez here's a small change for a pretty decent performance improvement. I tried memoizing more things in rubygems but decided not to since I'm unsure how useful it'd be.

deivid-rodriguez (commented):
This looks very good, thanks for this. A 30% improvement for a one line change feels pretty great.

Do any of the other changes have any effect at all on your application? I'm asking because it might be best to just change "the money maker" here. Other changes involve vendored code from Molinillo and code monkey patched from RubyGems, so we probably should not change those without changing the upstream too. Also, I'm not sure changing `sources` has any effect given that the source list is usually short and sources are not compared often.

deivid-rodriguez:
I tried this patch on `rails/rails` Gemfile and it leads to a pretty nice 8% `bundler/setup` time improvement! Also verified that only the change in `LazySpecification#hash` contributes to that improvement.

bundler/lib/bundler/lazy_specification.rb
ngan: Was following this pattern: https://github.com/rubygems/rubygems/blob/0aa81b10c8f6c20d03747400d4386514416d7244/bundler/lib/bundler/lazy_specification.rb#L116

ngan:
Yeah, I got a little carried away. Happy to just memorize the lazy spec only. It’s late for me so I won’t be able to get to it until tomorrow. Feel free to modify this pr and land! It’s a very simple change. Otherwise I can get to it tomorrow.

ngan:
Closing this out in favor of https://github.com/rubygems/rubygems/pull/5533

deivid-rodriguez:
So, we can confirm that #5533 does beat the 30% improvement of this PR in your application, and that this patch no longer has an effect in performance after #5533?

deivid-rodriguez:
Thank you so much for your investigation and patch, by the way!! ❤️ I obviously credited you on #5533 since I wouldn't have found the culprit without this hint!

ngan:
No problem. @technicalpickles and I will find some more time and more hints for you 😄

There’s actually a problem with the latest released version though so we haven’t been able to upgrade. We’re unable to do a bundle install from scratch. I’ll take a closer look today to see if it’s something on our end or it’s because of the last speed up changes.

ngan:
> Ouch, yeah, let me know what you find! :)

@deivid-rodriguez I started an issue: https://github.com/rubygems/rubygems/issues/5538

PR #5533, opened 2022-05-13
Improve `bundler/setup` performance again

deivid-rodriguez:

The performance of `bundler/setup` is still slow.

On a [different patch](https://github.com/rubygems/rubygems/pull/5532), it was noticed by @ngan that we are calling `LazySpecification#hash` many times, and simply memoizing that led to a very considerable performance improvement in his app.

I noticed though that we shouldn't be calling `LazySpecification#hash` that many times, and I located the culprit at `SpecSet#for` where we were deduplicating the partial aggregated result on every iteration. It is enough to do it just once at the end.

This leads on a 12% speedup on Rails repository Gemfile vs the previous 8% I was getting from memoizing `LazySpecification#hash`. Also, after this patch memoizing `LazySpecification#hash` has no effect in performance anymore.

### DIFF
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -24,7 +24,7 @@ module Bundler
 
         specs_for_dep = spec_for_dependency(dep, match_current_platform)
         if specs_for_dep.any?
-          specs += specs_for_dep
+          specs.concat(specs_for_dep)
 
           specs_for_dep.first.dependencies.each do |d|
             next if d.type == :development
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -24,7 +24,7 @@ module Bundler
 
         specs_for_dep = spec_for_dependency(dep, match_current_platform)
         if specs_for_dep.any?
-          match_current_platform ? specs += specs_for_dep : specs |= specs_for_dep
+          specs += specs_for_dep
 
           specs_for_dep.first.dependencies.each do |d|
             next if d.type == :development
@@ -39,6 +39,8 @@ module Bundler
       if spec = lookup["bundler"].first
         specs << spec
       end
+
+      specs.uniq! unless match_current_platform
 
       check ? true : specs
     end


bundler/lib/bundler/spec_set.rb
```diff
@@ -24,7 +24,7 @@ def for(dependencies, check = false, match_current_platform = false)
 
         specs_for_dep = spec_for_dependency(dep, match_current_platform)
         if specs_for_dep.any?
-          match_current_platform ? specs += specs_for_dep : specs |= specs_for_dep
+          specs += specs_for_dep
```
technicalpickles, line 27: I was looking at this line of code too after chatting with @ngan . `concat` would be faster since it adds to the existing array without creating a new array. 

```suggestion
          specs += specs_for_dep
```

It'd also use less memory, which I only mention because one of the flamegraphs I was looking at did show GC happening during `bundler/setup

bundler/lib/bundler/spec_set.rb
```diff
@@ -40,6 +40,8 @@ def for(dependencies, check = false, match_current_platform = false)
         specs << spec
       end
 
+      specs.uniq! unless match_current_platform
```
technicalpickles, line 43: I was also considering a `Set` for `specs`, but I'm not familiar enough with the semantics. It also doesn't have methods I could see to add several items in one go.

bundler/lib/bundler/spec_set.rb
deivid-rodriguez: Yes, I actually tried `concat` too, but it didn't show a significant speed up on my testing, so I left it out for now to keep the patch more focused. But we can add it yeah.

bundler/lib/bundler/spec_set.rb
deivid-rodriguez: We used to use `Set`, but we removed it in order to support using the `set` gem inside `Gemfile`'s.

bundler/lib/bundler/spec_set.rb
deivid-rodriguez: Added! It leads to about 10% memory saving on `rails/rails` Gemfile!

technicalpickles:
Compared this to 2.3.9 and 2.3.13 with our Gemfile:

```
# output is bundler version, and the output of `Benchmark.measure`
2.3.9
  2.789033   0.243557   3.032590 (  3.830557)

2.3.13
  1.328539   0.277676   1.606215 (  2.631212)

2.4.0.dev
  0.923131   0.097992   1.021123 (  1.024666)
```

So, looks good!

PR #5534, opened 2022-05-13
Memoize Bundler::LazySpecification#hash to improve performance

technicalpickles:

In local testing, we found `require 'bundler/setup'` took 1.8 seconds to load.
I started profiling and looking at flamegraphs for this code, and found
a surprising amount of time in this method.

Memoizing this specific method reduced time by 0.7 to 1.1 seconds, a 38%
improvement.

cc @ngan who I paired with on this

### DIFF
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -35,7 +35,8 @@ module Bundler
     end
 
     def hash
-      identifier.hash
+      # memoize this to help performance when called repeatedly for hash operations
+      @hash ||= identifier.hash
     end
 
     ##


technicalpickles:
I'm attaching the stackprof data as JSON, which can be viewed with [speedscope](https://github.com/jlfwong/speedscope)


[bundler-test.txt](https://github.com/rubygems/rubygems/files/8688131/bundler-test.txt)
(_is JSON despite name_)

deivid-rodriguez:
Yes! And also I think #5533 should make this unnecessary, can you verify?

technicalpickles:
Confirmed that https://github.com/rubygems/rubygems/pull/5533 would address this case 🚀

PR #5535, opened 2022-05-13
Gracefully handle gem activation conflicts in inline mode

deivid-rodriguez:

Bundler's "inline" mode resolves gems, installs them, and activates them on the fly. This is handy but in particular our install code still uses some gems (like `psych`). When that happens, the activated version of those gems might not match the resolved version in the inline `Gemfile`, causing a conflict.

The real fix to this issue is to completely stop depending on any gems. That's ongoing work that has been going on for years, but it's hard specially since new libraries get gemified.

This add some ad-hoc code to `bundler/inline` to recover from this kind of issue, by retrying with an extra requirement on the exact gem version that got activated, to ensure that there's no mismatch the second time.

This is a less invasive (I think) alternative to #5529.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -519,7 +519,7 @@ module Bundler
     end
 
     def write_lock(file, preserve_unknown_sections)
-      return if Definition.no_lock
+      return if Definition.no_lock || file.nil?
 
       contents = to_lock
 
--- a/bundler/lib/bundler/inline.rb
+++ b/bundler/lib/bundler/inline.rb
@@ -60,8 +60,24 @@ def gemfile(install = false, options = {}, &gemfile)
         end
       end
 
-      runtime = Bundler::Runtime.new(nil, definition)
-      runtime.setup.require
+      begin
+        runtime = Bundler::Runtime.new(nil, definition).setup
+      rescue Gem::LoadError => e
+        name = e.name
+        version = e.requirement.requirements.first[1]
+        activated_version = Gem.loaded_specs[name].version
+
+        Bundler.ui.info \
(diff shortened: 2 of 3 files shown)


bundler/lib/bundler/definition.rb
```diff
@@ -281,7 +281,7 @@ def groups
     end
 
     def lock(file, preserve_unknown_sections = false)
-      return if Definition.no_lock
+      return if Definition.no_lock || file.nil?
```
jonathanhefner, line 284: I think this removes the need for `file &&` on line 301, no? (Unless `file` can also be `false` for some reason?)

deivid-rodriguez: Indeed, thanks!

hsbt:
@deivid-rodriguez I rebase this today. Can you review this with my additional changes?

deivid-rodriguez:
Of course, I will have a look later today 👍.

bundler/lib/bundler/definition.rb
deivid-rodriguez: This comment still makes sense, we can now stop using `file_exists?(file)` a few lines below and do `File.exist?(file)` instead since the `nil` check is no longer needed. After that,  the `file_exists?` method has a single caller, so can probably be removed and change `lockfile_exists` (the only caller) to instead do `lockfile && File.exist?(lockfile)`.

deivid-rodriguez:
Looks good to me! Not sure why the ruby-core job failed the new test, I can look into it tomorrow.

hsbt:
@deivid-rodriguez I skip this example at ruby repo. This example is not working because ruby repo don't have installed gemspec. We may put fake system installation path and gemspec for this example.

Issue #5240, opened 2021-12-29
Some bundler specs are failing with `No such plugin: ruby` if `asdf` version manager is used


alexeevit:
### Describe the problem as clearly as you can

In specs, there are executions of bundle install with specific environment variables. Somehow, it writes `No such plugin: ruby` to the STDERR. However, if I run the same command manually in the terminal (I pause the spec to have the `bundle` binary), STDERR is empty.

### Did you try upgrading rubygems & bundler?

I use the master version

### Which command did you run?

`bin/rspec spec/install/gemspecs_spec.rb:11`

### What were you expecting to happen?

Successfully passed test

### What actually happened?

The test failed with the next error:
```
     Failure/Error: expect(err).to be_empty


       Commands:
       $ gem generate_index
       Generating Marshal quick index gemspecs for 48 gems
       ................................................
       Complete
       Generated Marshal quick index gemspecs: 0.007s
       Generating specs index
       Generated specs index: 0.000s
       Generating latest specs index
       Generated latest specs index: 0.000s
       Generating prerelease specs index
       Generated prerelease specs index: 0.000s
       Compressing indices
       Compressed indices: 0.000s
       # $? => 0

       $ /Users/vyacheslav/.asdf/installs/ruby/2.7.4/bin/ruby -I/Users/vyacheslav/dev/hobby/rubygems/bundler/spec -r/Users/vyacheslav/dev/hobby/rubygems/bundler/spec/support/artifice/fail.rb -r/Users/vyacheslav/dev/hobby/rubygems/bundler/spec/support/hax.rb /Users/vyacheslav/dev/hobby/rubygems/bundler/tmp/1/gems/system/bin/bundle install
       Fetching source index from file:///Users/vyacheslav/dev/hobby/rubygems/bundler/tmp/1/gems/remote2/
       Resolving dependencies...
       Using bundler 2.3.0.dev
       Fetching yaml_spec 1.0
       Installing yaml_spec 1.0
       Bundle complete! 1 Gemfile dependency, 2 gems now installed.
       Use `bundle info [gemname]` to see where a bundled gem is installed.
       No such plugin: ruby
       # $? => 0
       expected `"No such plugin: ruby".empty?` to be truthy, got false
     # ./spec/install/gemspecs_spec.rb:17:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:100:in `block (4 levels) in <top (required)>'
[...]

PR #5557, opened 2022-05-20
Support running specs when using `asdf` version manager

deivid-rodriguez:

Contributors using `asdf` as their version manager run into issues.

The problem is the `asdf-ruby` plugin sets `RUBYLIB` to require some code to reshim after installing gems. This interferes with our specs.
    
My fix is to reset `RUBYLIB` for specs, making sure to leave "internal entries" since ruby-core needs them.

Fixes https://github.com/rubygems/rubygems/issues/5240.

### DIFF
--- a/bundler/spec/spec_helper.rb
+++ b/bundler/spec/spec_helper.rb
@@ -78,10 +78,19 @@ RSpec.configure do |config|
     ENV["XDG_CONFIG_HOME"] = nil
     ENV["GEMRC"] = nil
 
+    extend(Spec::Helpers)
+
+    # Ruby-core needs RUBYLIB set, leave any "internal" RUBYLIB entries to avoid
+    # resetting that, but reset anything else to avoid external stuff like asdf
+    # hooks from interfering with our specs
+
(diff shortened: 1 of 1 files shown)


simi:
Can't be this solved in `asdf` actually?

deivid-rodriguez:
I don't think so, I think asdf relies on this for shim regeneration, but this is not particular to `asdf` anyways, but to any environment with `RUBYLIB` set. I think it's best that our specs are resilient to that, just like we reset other environment variables to make our test environment as independent as possible.

deivid-rodriguez:
I recommend https://github.com/asdf-community/asdf-direnv anyways, that completely removes the need for shims anyways :)

deivid-rodriguez:
Well, this seems to have broken ruby-core specs, so I'll revert it and reevaluate later.

deivid-rodriguez:
Thinking about this again, I think Rbenv does something similar to asdf regarding shims, but Rbenv does not rely on `RUBYLIB`, or at least has always worked fine for developing RubyGems, so there's probably something that can be improved in the asdf side indeed.

PR #5565, opened 2022-05-25
Exclude `Rakefile` file from new gems

nobu:

Often unused files after installation are included in gem files.

Exclude files for `rake`.

All `[A-Z]\w+file` may be unnecessary.


deivid-rodriguez:
As a compromise (and because it fixes CI since the line is no longer too long for RuboCop), I changed this to only exclude `Rakefile`, which is the one file that gets generated by default.

nobu:
I didn't consider the case extensions using Rakefile.
I'll close this.

PR #5592, opened 2022-06-02
Refine digest

nobu:

Actually nothing.
bundler/lib/bundler/digest.rb implementation seems literally translated from C code, and not like ruby code. 

Get rid of byte operations as possible.

### DIFF
--- a/bundler/lib/bundler/digest.rb
+++ b/bundler/lib/bundler/digest.rb
@@ -43,7 +43,7 @@ module Bundler
               f = (b ^ c ^ d)
               k = 0xCA62C1D6
             end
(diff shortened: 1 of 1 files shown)


djberg96:
Agree with @deivid-rodriguez I don't think it looks all that more readable. I'm also curious why it even exists when we've had Digest::SHA1 in the stdlib for a long time. What am I missing?

deivid-rodriguez:
@djberg96 See https://github.com/rubygems/rubygems/pull/4989 for why we need to do this.

deivid-rodriguez:
Closing due to lack of feedback.

PR #5595, opened 2022-06-02
Improve our SHA1 implementation

### LINKED ISSUE #5591: As `+` is higher precedence than `&`, this equals `SHA1_MASK & (SHA1_MASK & (rotate(a, 5) + f + e + k + w[I]))`.
As `+` is higher precedence than `&`, this equals `SHA1_MASK & (SHA1_MASK & (rotate(a, 5) + f + e + k + w[I]))`.
I don't think bit-and with the same value twice has meaning.
Maybe `SHA1_MASK & ((SHA1_MASK & rotate(a, 5)) + f + e + k + w[I])` is intended?

_Originally posted by @nobu in https://github.com/rubygems/rubygems/pull/4989#discussion_r887406201_

deivid-rodriguez:

It was noted that there's a redundant bitwise AND, and also I found that the bitwise AND here

https://github.com/rubygems/rubygems/blob/7e5765a66c9fe5187b167f619f34db5db121f2df/bundler/lib/bundler/digest.rb#L50

was not covered by any specs.

Fix both issues by removing the redundant `&` and by adding more specs from the RFC.

Closes #5591.

### DIFF
--- a/bundler/lib/bundler/digest.rb
+++ b/bundler/lib/bundler/digest.rb
@@ -43,7 +43,7 @@ module Bundler
               f = (b ^ c ^ d)
               k = 0xCA62C1D6
             end
-            t = SHA1_MASK & (SHA1_MASK & rotate(a, 5) + f + e + k + w[i])
+            t = SHA1_MASK & rotate(a, 5) + f + e + k + w[i]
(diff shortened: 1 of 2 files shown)


byroot (commented):
👍

simi (approved):

Issue #5597, opened 2022-06-03
Unresolved or ambiguous specs during Gem::Specification.reset warning with plugin


djberg96:
ruby 3.1.1p18 on Mac, rubygems 3.3.15 (full env below)

I'm seeing a bunch of warnings when I try to use a local rubygems plugin. There is nothing special about the plugin, it's just a clock skew checker that you can see at https://github.com/djberg96/skew_guard/blob/main/lib/rubygems_plugin.rb. A regular `gem install` does not emit any warnings:

```
>RUBYOPT=-Ilib gem install sys-admin
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1245:in `reset': WARN: Unresolved or ambiguous specs during Gem::Specification.reset: (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1247:in `block in reset':       rspec (~> 3.9) (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1251:in `block in reset':       Available/installed versions of this gem: (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1252:in `block (2 levels) in reset':       - 3.11.0 (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1252:in `block (2 levels) in reset':       - 3.10.0 (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1255:in `reset': WARN: Clearing out unresolved specs. Try 'gem cleanup <gem>' (StructuredWarnings::StandardWarning)
/Users/dberger/.rbenv/versions/3.1.1/lib/ruby/site_ruby/3.1.0/rubygems/specification.rb:1256:in `reset': Please report a bug if this causes problems. (StructuredWarnings::StandardWarning)
```

I tried uninstalling and/or cleaning first, but that had no effect.

```
>gem env
RubyGems Environment:
[...]

PR #5601, opened 2022-06-06
Remove unnecessary string concatenation from `Gem::Specification.reset`

djberg96:
Sort of a followup to https://github.com/rubygems/rubygems/issues/5597, in which we discover that the reset method specification.rb has a goofy string construction that was originally written that way just to avoid syntax highlighting.

We don't need the word "WARN" (or in this case "W" + "ARN"), it's already using Ruby's `warn`.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1240,8 +1240,7 @@ class Gem::Specification < Gem::BasicSpecification
     clear_load_cache
     unresolved = unresolved_deps
     unless unresolved.empty?
-      w = "W" + "ARN"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Can we keep the "WARN" part and just remove the code weirdness? Some specs rely on it, so I would focus on what we want to fix here.

djberg96:
@deivid-rodriguez Ok, updated. Pretty sure the one failed check is unrelated.

deivid-rodriguez (approved):
Rebased, squashed, and set to auto merge after CI. Thank you!

PR #5606, opened 2022-06-07
Try make MacOS CI more stable

deivid-rodriguez:

MacOS CI seems to be running slower at the moment.

Relax some limits so that it still runs green.

### DIFF
--- a/bundler/spec/realworld/slow_perf_spec.rb
+++ b/bundler/spec/realworld/slow_perf_spec.rb
@@ -17,6 +17,6 @@ RSpec.describe "bundle install with complex dependencies", :realworld => true do
 
     duration = Time.now - start_time
 
-    expect(duration.to_f).to be < 12 # seconds
+    expect(duration.to_f).to be < 18 # seconds
   end
 end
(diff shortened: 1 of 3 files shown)


bundler/spec/realworld/slow_perf_spec.rb
```diff
@@ -17,6 +17,6 @@
 
     duration = Time.now - start_time
 
-    expect(duration.to_f).to be < 12 # seconds
+    expect(duration.to_f).to be < 18 # seconds
```
djberg96, line 20: This test just strikes me as being too unreliable in general. Maybe `skip if ENV['CI']`?

deivid-rodriguez: Yes, indeed, but if it's unreliable in CI I don't see why it wouldn't be locally too. This spec was originally added to cover a performance issue where this Gemfile would take more than 10 minutes to resolve, so I think having something relaxed still covers us from a real regression here where the test would most likely fail consistently.

deivid-rodriguez: I will merge this for now, but we can keep discussing this.

PR #5613, opened 2022-06-08
Add support for `bundle gem --rust` command

ianks:

Despite the recent support of Rust/Cargo as a supported language for developing Ruby gems (#5175), it's still too hard to scaffold out a Rust gem. It should be as easy to scaffold a Rust extension as it is a C extension.

This PR adds the ability to scaffold a Rust gem by running: `bundle gem --rust`. A few key points about this integration are:

1. Works perfectly with `rake-compiler` via the `rb_sys` gem
2. Opted to use the [magnus](https://github.com/matsadler/magnus) crate, as I believe it offers the most correct and easy to use bindings for Ruby, plus it supports interop with [rb-sys](https://github.com/oxidize-rb/rb-sys). With magnus, gem authors can avoid writing unsafe code when authoring a new gem in Rust.
3. Scaffolded gems make use of the [rb-allocator](https://github.com/oxidize-rb/rb-sys/tree/main/crates/rb-allocator), so Rust memory usage is properly reported to the Ruby GC.
4. I prefer using the `--rust` flag over something like `--ext=rust`. It's much nicer to read and write.

Looking forward to everyone's feedback!


simi:
Damn, I have missed one. I have code prepared for `--ext=rust` (backwards compatible) which seems better to me. I tried to keep it as close as possible with C extension (using `ext` folder, ...). Would you mind to check that one (https://github.com/rubygems/rubygems/commit/548189dd018f2e1a66525c168ada9868f8100f2e) as well? It creates `"ext/<%= config[:name] %>.rs"`, but I wasn't sure how to keep it as close as possible to common Rust package.

Maybe we can just keep the standard Rust (Cargo) structure, but in `ext` or `ext-rust` subfolder. WDYT? That way for `bundle gem newgem --ext=rust` you'll find this structure (next to common gem one):

```
ext/src/newgem.rs
ext/Cargo.toml
```

So actual changing directory to `ext`, you'll get the original `Cargo` feeling. Also we need to add generated dummy tests for nice and smooth experience. Those were last 2 blockers for me to open PR, but due to lack of time and ideas I wasn't able to move it forward.

Btw. we can make `--rust` alias to `--ext=rust` if needed.

simi:
@ianks do you plan to continue on this or should I take over? I would like to merge with my branch and test.

simi:
@ianks friendly pingie. I'll have time this weekend, I can try to finish it if welcomed.

ianks:
@simi can you take it from here? things got busy for me!!

ianks:
@simi let me know if you run into any issues!

ankane:
Hi, excited to see this included in Bundler!

I just tried moving [a project](https://github.com/ankane/tokenizers-ruby) to this directory structure and think there may be a potential issue.

From what I can tell, the Cargo builder always places the shared library in `lib`. However, the `rake-compiler` configuration places it in `lib/#{gem_name}` (which is what the `require_relative` expects).

This may be something worth changing or making configurable with the Cargo builder.

ianks:
For others that stumble on this, check out the structure I setup for [`wasmtime-rb`](https://github.com/bytecodealliance/wasmtime-rb), I think it serves as a good standard

ankane:
Hmm, it doesn't seem to install correctly when using `spec.extensions = ["ext/Cargo.toml"]` and Ruby 3.2.0-preview2.

On installation, it creates `lib/wasmtime-rb.bundle` on Mac, but `lib/wasmtime.rb` has `require "wasmtime/ext"`, so it fails with:

```text
cannot load such file -- wasmtime/ext (LoadError)
```

ianks:
> Edit: require "wasmtime-rb" should work in this case, but for a gem name like hello, you'll end up with lib/hello.rb and lib/hello.bundle and have to use require "hello.#{RbConfig::CONFIG["DLEXT"]}" to require it.

Yeah you end up having to use a different name each. It's even a bit more entailed because for cross compiled gems you have to load based on the Ruby version.

deivid-rodriguez:
This PR can be closed now, correct?

simi:
Yes

Dne st 21. 12. 2022 18:45 uživatel David Rodríguez ***@***.***>
napsal:

deivid-rodriguez:
Thanks for getting this started @ianks!

PR #5618, opened 2022-06-09
feat: add Bundler.nicer_setup method

### LINKED ISSUE #5617: export `bundler/setup` as method in `bundler` to allow customizing groups

Using

```rb
require "bundler/setup"
```

gives the nice UI when any dependency is missing.

However it is hardcoded to setup all groups.

If I want to setup only `:default` group but keep the nice UI for the bundler errors, I would need to rip-off the code from `bundler/setup` file:

```rb
require "bundler"
begin
  Bundler.ui.silence { Bundler.setup(:default) }
rescue Bundler::BundlerError => e
  Bundler.ui.error e.message
  Bundler.ui.warn e.backtrace.join("\n") if ENV["DEBUG"]
  if e.is_a?(Bundler::GemNotFound)
    Bundler.ui.warn "Run `bundle install` to install missing gems."
  end
  exit e.status_code
end
```

Could we export the `bundler/setup` code to e.g. `Bundler.friendly_setup(*groups)`. So it would be possible to just have this code:
```rb
require "bundler"
Bundler.friendly_setup(:default)
```

Alternatively, should `bundler/setup` respect `BUNDLER_WITH` / `BUNDLER_WITHOUT` environment variables?

DocX:
Moves code from bundler/setup to Bundler.nicer_setup method so it can be reused with different groups

Fixes #5617 

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -187,6 +187,25 @@ module Bundler
       setup(*groups).require(*groups)
     end
 
+    # Same as +Bundler.setup+ but catches errors and prints nicer
+    # message in terminal
+    def nicer_setup(*groups)
+      if STDOUT.tty? || ENV["BUNDLER_FORCE_TTY"]
+        begin
+          Bundler.ui.silence { Bundler.setup(*groups) }
+        rescue Bundler::BundlerError => e
+          Bundler.ui.error e.message
+          Bundler.ui.warn e.backtrace.join("\n") if ENV["DEBUG"]
+          if e.is_a?(Bundler::GemNotFound)
+            Bundler.ui.warn "Run `bundle install` to install missing gems."
+          end
+          exit e.status_code
+        end
+      else
+        Bundler.ui.silence { Bundler.setup(*groups) }
+      end
+    end
+
     def load
       @load ||= Runtime.new(root, definition)
     end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I don't like the `Bundler.nicer_setup` or `Bundler.friendly_setup` names, they kind of imply that `Bundler.setup` is not nice of friendly and I don't like that 😅.

But I don't have any good ideas for a better name... Only thing that comes to mind would be `Bundler.setup!` where the bang would imply that the method "exits hardly" (`ExitError`), so errors raised from it if any are not fine grained and not supposed to be rescued.

simi:
Maybe `graceful_setup`?

DocX:
The whole "nicer" method uses lot of `Bundler.ui` - so would it make sense to move it as a whole to `Bundler.ui`? E.g. `Bundler.ui.setup(*groups)`?

Another idea would be to overload the already existing `Bundler.setup` method with kwarg: `setup(*groups, print_errors: false)`? However it feels bit forceful and it would make the code likely not very readable :/

deivid-rodriguez:
Sorry I didn't reply earlier. I was thinking this code is actually quite similar to what `Bundler.with_friendly_errors` does. Maybe it can be incorporated into that helper, and reduce the amount of code we keep at `bundler/setup.rb`. I will have a look at that.

PR #5632, opened 2022-06-16
Fix extension paths in generated standalone script

citrusmoose:

The paths for extensions of gems would contain the hardcoded ruby version on which the extension was built.  That's inconsistent with the parent ruby version directory.  It also makes the generated script incompatible with different ruby versions.

I replaced the static ruby version with runtime ruby version like the parent version directory.


deivid-rodriguez (approved):
Looks good to me, thanks a lot @citrusmoose!

PR #5638, opened 2022-06-21
Fix daily Bundler CI

deivid-rodriguez:

After merging #5632, our daily Bundler CI against ruby 3.2.0-dev started failing.

The reason is that ruby 3.2.0-dev is built without shared libraries enabled, and that distinction is part of the directory name where extensions are installed (it's `3.2.0+1-static` for this Ruby), because statically and non-statically built extensions are incompatible with each other.

My fix is to amend #5632 to add support for this. RubyGems has `Gem.extension_api_version` for this, but I believe the standalone script is supposed to work without RubyGems so that required a bit of extra work.

Link to a passing daily CI is here: https://github.com/rubygems/rubygems/runs/6989422115?check_suite_focus=true.

### DIFF
--- a/bundler/lib/bundler/installer/standalone.rb
+++ b/bundler/lib/bundler/installer/standalone.rb
@@ -12,6 +12,7 @@ module Bundler
       end
       File.open File.join(bundler_path, "setup.rb"), "w" do |file|
         file.puts "require 'rbconfig'"
+        file.puts define_path_helpers
         file.puts reverse_rubygems_kernel_mixin
         paths.each do |path|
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
By the way, this issue should've been caught by our `ruby-core` CI workflow, but... it's skipped there:

https://github.com/rubygems/rubygems/blob/ecc8e895b69063562b9bf749b353948e051e4171/bundler/spec/install/gems/standalone_spec.rb#L193

I will propose another patch to skip this spec in ruby-core so something similar will be caught earlier in the future.

PR #5644, opened 2022-06-22
Document `gem env` argument aliases and add `gem env userhome`

### LINKED ISSUE #5366: Support: "gem env NAME" such as "gem env INSTALLATION_DIRECTORY"
First my proposed syntax:

    gem env INSTALLATION_DIRECTORY
    gem env installation_directory

In other words, you can use ALL CAPS or without caps.

The above command would then output this:

    /root/.gem

Or whatever else is at that, e. g. the user home dir + .gem.

Why do I propose this? After all we could use "gem env"
and parse the output. This is what I am doing right now,
so I don't have an issue with it as-is. But I'd like to get
this fine-tuned information without having to parse on
my own.

Right now I want to copy all bin/ files of one gem into 
the current working directory, if a user makes use of
a special flag. But the bin/ files are not available under
lib/ so I need some way to find out which bin/ files
exist. The .gemspec is not available all the time either,
so my idea was simply to add a .yml file that contains
all entries under bin/ and then simply copy these files
as-is. For this I need to know the user prefix (from
installation directory), and then iterate over that
yaml file, and simply copy all found bin/ files into the
current working directory (think of restricted campus
sites where the bin/ files may be available but no 
sufficient permission was given to the user, not even
for a home dir, so just /tmp/ is available or something
like that. Here I want to give users a simple way to
batch-copy all bin/ entries and use my gem files
that way too, as a fall-back option).

Not sure if this has a chance to be accepted, but I wanted
to suggest it nonetheless.

The same pattern should be true for other keywords,
in particular all CAP entries, such as USER INSTALLATION
DIRECTORY or RUBY EXECUTABLE. Rather than the
" " I propose to replace these with "_" for the purpose of
this command here.

So:

    gem env USER_INSTALLATION_DIRECTORY

would just feedback where the user installation directory is at.
Thanks for reading!

deivid-rodriguez:

Nothing really, just making `gem env` goodies more visible, and also adding support for `gem env userhome`.

Implement the above.

Also made `gem env home` and `gem env path` more prominent since they read very well :)

Closes #5366.

### DIFF
--- a/lib/rubygems/commands/environment_command.rb
+++ b/lib/rubygems/commands/environment_command.rb
@@ -8,8 +8,8 @@ class Gem::Commands::EnvironmentCommand < Gem::Command
 
   def arguments # :nodoc:
     args = <<-EOF
-          gemdir          display the path where gems are installed
-          gempath         display path used to search for gems
+          home            display the path where gems are installed
+          path            display path used to search for gems
           version         display the gem format version
           remotesources   display the remote gem servers
           platform        display the supported gem platforms
--- a/lib/rubygems/commands/environment_command.rb
+++ b/lib/rubygems/commands/environment_command.rb
@@ -10,6 +10,7 @@ class Gem::Commands::EnvironmentCommand < Gem::Command
     args = <<-EOF
           home            display the path where gems are installed. Aliases: gemhome, gemdir, GEM_HOME
           path            display path used to search for gems. Aliases: gempath, GEM_PATH
+          user_gemhome    display the path where gems are installed when `--user-install` is given. Aliases: user_gemdir
           version         display the gem format version
           remotesources   display the remote gem servers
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Any better ideas for `gem env userhome`? I'm afraid it could be confused with `ENV["HOME"]`.

Maybe `user_install_dir` or something? But in that case we might want to introduce `install_dir` as yet another alias for `gem env home`. Currently we already have `gem env home`, `gem env gemhome`, `gem env gemdir` and `gem env GEM_HOME` that print the same thing.

Maybe don't introduce it at all, but seems handy to me given that we already have shortcut for the other common RubyGems paths.

deivid-rodriguez:
I went with `gem env user_gemhome` and `gem env user_gemdir` to avoid the above potential confusion.

PR #5649, opened 2022-06-24
Add support for platform `:x64_mingw` to correctly lookup "x64-mingw-ucrt"

johnnyshields:
Fixes https://github.com/rubygems/rubygems/issues/5269

I've implemented it in such a way that platform `:x64_mingw` resolves to either `x64-mingw32` or `x64-mingw-ucrt`.

Note [this comment from larskanis](https://github.com/rubygems/rubygems/issues/5269#issuecomment-1010869946):

> [deivid-rodriguez](https://github.com/deivid-rodriguez) > So, as an alternative to adding a new platform, you'd prefer that the current x64_mingw platform value resolved to urct variants, right?
[larskanis](https://github.com/larskanis) >Yes. The platforms x64-mingw32 and x64-mingw-ucrt are ABI incompatible - that's why we introduced a new platform string for RUBY_PLATFORM and Rubygems. But there is little to no difference between them, when it comes to enabling/disabling gems from a Gemfile.

PR #5655, opened 2022-06-26
Make platform `universal-mingw32` match "x64-mingw-ucrt"

johnnyshields:
This is a follow-up to https://github.com/rubygems/rubygems/pull/5649.

@larskanis please confirm this change is correct.

In order to do this change, we need to modify `Gem::Platform#===` with the following special logic:

```
    # universal-mingw32 matches x64-mingw-ucrt
    return true if (@cpu == 'universal' or other.cpu == 'universal') and
                   @os.start_with?('mingw') and other.os.start_with?('mingw')
```

The reason we can't so something simpler like adding `.start_with?('mingw')` to the existing OS conditions is that it will cause `x64-mingw32` to match with `x64-mingw-ucrt`, which we don't want.

### DIFF
--- a/bundler/spec/resolver/platform_spec.rb
+++ b/bundler/spec/resolver/platform_spec.rb
@@ -337,6 +337,14 @@ RSpec.describe "Resolving platform craziness" do
         should_resolve_as %w[thin-1.2.7-x64-mingw-ucrt]
       end
     end
+
+    if Gem.rubygems_version >= Gem::Version.new("3.3.18")
+      it "finds universal-mingw gems on x64-mingw-ucrt" do
+        platform "x64-mingw-ucrt"
+        dep "win32-api"
+        should_resolve_as %w[win32-api-1.5.1-universal-mingw32]
+      end
+    end
   end
 
   describe "with conflicting cases" do
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -159,6 +159,10 @@ class Gem::Platform
   def ===(other)
     return nil unless Gem::Platform === other
 
+    # universal-mingw32 matches x64-mingw-ucrt
+    return true if (@cpu == 'universal' or other.cpu == 'universal') and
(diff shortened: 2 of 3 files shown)


larskanis:
`win32-api` is the only gem that I know of, which is using this `universal` CPU prefix. The RubyInstaller-3.1 currently doesn't install the `universal-mingw32` version right now, but the `ruby` version instead. To fix this is what I understand as the intention of this PR. Right?

Implementation wise it looks correct to me, but I don't know the bundler/rubygems internals enough, to know if there are any more changes necessary.

johnnyshields:
@larskanis 

> To fix this is what I understand as the intention of this PR. Right?

Yes, correct. Thanks!

johnnyshields:
@deivid-rodriguez this is ready for your review.

deivid-rodriguez:
Alright, thanks, I'll review this in the following 2 weeks and try ship it with the next release!

johnnyshields:
@deivid-rodriguez one spec didn't run for some reason, but this is good to go.

deivid-rodriguez:
I don't mind introducing this if we decide it's practical/helpful, I just want to make sure it's really necessary.

deivid-rodriguez:
I think you're right @johnnyshields. I will approve this but will let it rest some days in case somebody wants to chime in.

deivid-rodriguez (approved):

bundler/spec/resolver/platform_spec.rb
```diff
@@ -337,6 +337,14 @@
         should_resolve_as %w[thin-1.2.7-x64-mingw-ucrt]
       end
     end
+
+    if Gem.rubygems_version > Gem::Version.new("3.3.999")
```
deivid-rodriguez, line 341: Could you change this to `>= 3.3.18`? Just to match what we do in other places, keeping the exact version where the feature was introduced.

johnnyshields: Done

deivid-rodriguez:
Thanks to you!

PR #5662, opened 2022-06-28
Document how to run rake setup under regular user

graywolf-at-work:
When running rake setup under regular user, it fails with permission
errors since it tries to install into global gem directory. Running it
under root is a workaround, however in that case bundler complains that
it should not run under root.

Setting GEM_HOME is a working solution, so it should be documented.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,14 @@ here: https://guides.rubygems.org/contributing/
 
     rake setup
 
+NOTE: If this fails with permission related errors, you're most likely using a
+global Ruby installation (like the one packaged by your OS), which sets
+`GEM_HOME` to a location regular users can't write to. Consider using a Ruby
+version manager like RVM, Rbenv or Chruby (which install Ruby to a location
+regular users can write to). Alternatively, consider setting `GEM_HOME`
+environment variable to a writable location with something like
+`export GEM_HOME=/tmp/rubygems.gems` and try again.
+
 > Optionally you can configure git hooks with: rake git_hooks
 
 To run commands like `gem install` from the repo:


CONTRIBUTING.md
```diff
@@ -31,6 +31,11 @@ here: https://guides.rubygems.org/contributing/
 
     rake setup
 
+If this fails with permission related errors, consider setting `GEM_HOME`
+environment variable and then try again:
+
+    export GEM_HOME=/tmp/rubygems.gems
+
```
deivid-rodriguez, line 38: This will be fixed by #5327, but I'm happy to provide workarounds in the mean time.

But I'd rather recommend using a "per-user" Ruby installation like those provided by RMV, Rbenv, or Chruby. We could say something like

> NOTE: If this fails with permission related errors, you're most likely using a global Ruby installation (like the one packaged by your OS), which sets `GEM_HOME` to a location regular users can't write to. Consider using a Ruby version manager like RVM, Rbenv or Chruby (which install Ruby to a location regular users can write to). Alternatively, consider setting `GEM_HOME` environment variable to a writable location with something like `export GEM_HOME=/tmp/rubygems.gems` and try again.

simi:
If I remember well, there is nothing special about `rake setup`. It is similar to `gem install rubygems-update` package. It depends on your system setup where it gets installed. For system maintained ruby you should never install gems using `gem` command (including this `rake setup`).

deivid-rodriguez:
I still think it's nice to add this advice (and #5327) to support casual contributors that are not every day Ruby developers (I guess every day Ruby developers who are using system maintained Ruby know what they are doing and don't really need this advice).

simi:
> > If I remember well, there is nothing special about `rake setup`. It is similar to `gem install rubygems-update` package. It depends on your system setup where it gets installed. For system maintained ruby you should never install gems using `gem` command (including this `rake setup`).
> 
> My problem with rake setup is that it does not accept --user flag, the gem command does. So I though documenting this is useful for people like me, who just want to sent few pull requests and need to figure out how to run tests locally.

By your description is still seems to me you're mixing system Ruby with custom RubyGems and custom gems directory. That's unsupported installation and is not recommended.

deivid-rodriguez:
This recommendation is going to be short-lived because this will just work after #5327. So I guess we can close this "in favor" of #5327.

deivid-rodriguez:
By the way @simi I understand your concerns and that this is not recommended for regular users (although I don't think in this case there's no custom RubyGems being mixed, just custom `GEM_HOME`).

I think it's acceptable to inform people willing to contribute to RubyGems (as opposed to end users) about this workaround, and take the chance to make it clear what our preferred setup is.

But as I said, I'm ok to close this too since this won't be necessary anymore soon.

simi:
Ahh, this is not related to `ruby setup.rb` at all. Sorry, I have probably misunderstood the original problem then.

deivid-rodriguez:
That's correct @simi. Do you think this note is worth being added or it's maybe too distracting for most users?

Issue #508, opened 2013-03-11
Additional Slashes in paths breaks gem installation


dekz:
When install an gem into `/a/b//c/gems` Gem complains and [fails this condition](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/package.rb#L374) all is well if the path is changed to `/a/b/c/gems`. This is an annoyance when using tools like RVM which may have an entire environment setup with a double slashed path.

`destination` is correct `/a/b/c/gems` whereas `destination_dir` is `/a/b//c/gems`. As it seems ruby takes into account double slashed paths in either `File.join` or `File.expand_path`

copiousfreetime:
This appears to be fixed with d3ac81a and 1.8.7 is no longer supported.

PR #5663, opened 2022-06-28
Fix test_install_location_extra_slash on alpine

### LINKED ISSUE #5652: Failing test on alpine linux

When running tests from current master, one test is failing on alpine linux:

```
Failure: test_install_location_extra_slash(TestGemPackage)
/home/x/rubygems/test/rubygems/test_gem_package.rb:744:in `test_install_location_extra_slash'
     741:
     742:     destination = package.install_location file, destination
     743:
  => 744:     assert_equal File.join(@destination, 'foo', 'file.rb'), destination
     745:     refute destination.tainted? if RUBY_VERSION < '2.7'
     746:   end
     747:
<"/home/x/rubygems/tmp/test_rubygems_20220624-280-oqo8ih/extract/foo/file.rb"> expected but was
<"//home/x/rubygems/tmp/test_rubygems_20220624-280-oqo8ih/extract/foo/file.rb">

diff:
? //home/x/rubygems/tmp/test_rubygems_20220624-280-oqo8ih/extract/foo/file.rb
```

I plan to look into this and provide a patch, creating the issue mostly for record keeping.

No

```
podman run --rm alpine:3.16 sh -c 'apk add alpine-sdk ruby ruby-dev ruby-rdoc ruby-rake && adduser -D x && su - x -c "git clone https://github.com/rubygems/rubygems" && cd ~x/rubygems && rake setup && su - x -c "cd rubygems && rake test"'
```

```
rake test
```

I was expecting the tests to pass.

### What actually happened?

One test failed.

```
RubyGems Environment:
  - RUBYGEMS VERSION: 3.3.7
  - RUBY VERSION: 3.1.2 (2022-04-12 patchlevel 20) [x86_64-linux-musl]
  - INSTALLATION DIRECTORY: /usr/lib/ruby/gems/3.1.0
  - USER INSTALLATION DIRECTORY: /root/.local/share/gem/ruby/3.1.0
  - RUBY EXECUTABLE: /usr/bin/ruby
  - GIT EXECUTABLE:
  - EXECUTABLE DIRECTORY: /usr/bin
  - SPEC CACHE DIRECTORY: /root/.local/share/gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - x86_64-linux-musl
  - GEM PATHS:
     - /usr/lib/ruby/gems/3.1.0
     - /root/.local/share/gem/ruby/3.1.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => false
     - :bulk_threshold => 1000
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /usr/local/sbin
     - /usr/local/bin
     - /usr/sbin
     - /usr/bin
     - /sbin
     - /bin
```

graywolf-at-work:
Under POSIX behavior of leading // is implementation defined. Musl does
preserve it in realpath, glibc does not. That means the test was failing
when executed on alpine linux. Original issue #508 was about // in the
path, not about leading ones. When executed in such environment, the
test will still test what it should when the explicit mangling of the
path is not done.

Fixes #5652

### DIFF
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -737,9 +737,7 @@ class TestGemPackage < Gem::Package::TarTestCase
     file = 'foo//file.rb'.dup
     file.taint if RUBY_VERSION < '2.7'
 
-    destination = @destination.sub '/', '//'
-
-    destination = package.install_location file, destination
+    destination = package.install_location file, @destination
 
     assert_equal File.join(@destination, 'foo', 'file.rb'), destination
     refute destination.tainted? if RUBY_VERSION < '2.7'


test/rubygems/test_gem_package.rb
```diff
@@ -738,6 +738,7 @@ def test_install_location_extra_slash
     file.taint if RUBY_VERSION < '2.7'
 
     destination = @destination.sub '/', '//'
+    destination = ?/ + destination if destination.start_with? '//'
```
deivid-rodriguez, line 741: But now this test always tests with 3 slashes, no? Because the previous line will always return a path with two leading slashes?

Can we instead detect that the test is running on Alpine and add the extra slash just in that case, with a comment? I think it would make more explicit what's going on.

deivid-rodriguez: I'm pretty sure it does indeed not matter. But we'll be changing the test to test something different from the original, so I think we should at least double check the original issue that led into adding the test, and make sure that using 3 double slashes covers the original issue just as fine.

Even if the commit message is well detailed, I think adding a comment may prevent someone from breaking the test in the future thinking two slashes are enough 😅, given that we don't really test on Alpine or any other musl libc distribution. But if you don't feel it's needed, it's fine with me.

Also, regarding `?/`, today I learned. Can we use something more commonly used like `'/'`?

deivid-rodriguez: I was checking the code and it was introduced to fix https://github.com/rubygems/rubygems/issues/508. In that issue, leading extra slashes were not the issue, it was double slashes inside gem paths, and I don't think that's a thing in real life either. I guess the original issue came from something RVM was doing when merging paths that resulted in duplicated `//` in the middle of `GEM_HOME` or something.

So maybe the easiest thing to do here is to remove the line above and pass `@destination` directly to `#install_location`. That has no chances of being reverted, needs no comment, and the test is still checking normalization of inner double slashes which seems to work consistently across all libc implementations.

deivid-rodriguez: To be clear, this would be my suggested alternative patch

```diff
diff --git a/test/rubygems/test_gem_package.rb b/test/rubygems/test_gem_package.rb
index 34660bdccd..e3dce87c95 100644
--- a/test/rubygems/test_gem_package.rb
+++ b/test/rubygems/test_gem_package.rb
@@ -737,9 +737,7 @@ def test_install_location_extra_slash
     file = 'foo//file.rb'.dup
     file.taint if RUBY_VERSION < '2.7'
 
-    destination = @destination.sub '/', '//'
-
-    destination = package.install_location file, destination
+    destination = package.install_location file, @destination
 
     assert_equal File.join(@destination, 'foo', 'file.rb'), destination
     refute destination.tainted? if RUBY_VERSION < '2.7'
```

deivid-rodriguez: > Note: Technically we could replace last / with //, but not sure it's worth the effort. We are basically testing realpath at that point.

Yes, I was exactly looking into that, I was thinking of `@destination.reverse.sub('/', '//').reverse`, but then I realized that the test didn't really need this.

deivid-rodriguez (approved):
Thank you!

PR #5667, opened 2022-06-28
Run linters before tests and compilation

skryukov:

Currently, `bundler gem` generates `Rakefile` with default task running linters as the last step, which is not the best UX.

This PR updates `Rakefile.tt` to make linters run first.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/Rakefile.tt
+++ b/bundler/lib/bundler/templates/newgem/Rakefile.tt
@@ -27,17 +27,6 @@ require "rspec/core/rake_task"
 RSpec::Core::RakeTask.new(:spec)
 
 <% end -%>
-<% if config[:linter] == "rubocop" -%>
-<% default_task_names << :rubocop -%>
-require "rubocop/rake_task"
-
-RuboCop::RakeTask.new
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (commented):
I'm curious why you think this is best. I think it's more standard to do it the current way? Normally you want to focus first on things passing and you can leave style tweaks till after tests are passing? If you want direct style feedback, normally people integrate that with their editor. I guess this reversed order could make sense for test suites that take very long, but I'm not sure it makes sense for empty gems?

skryukov:
I think if you are working on the feature and running only tests, you should probably run `rake test/spec`, since failed linter at the end is not what you are looking for in the process.

The other thing is that we use the same default rake task in CI templates, and for CI, it's definitely better to fail fast.

We can add a special `:ci` task and reorder things in it instead, that also might work 🤔

deivid-rodriguez:
Since there was no further feedback, and I'm not convinced this default is better than the one we have now, I'll close this. I feel there will be some people who will not like the new default, so I think it's best to not change anything here.

Feel free to keep discussing if something does not make sense here.

PR #5669, opened 2022-06-28
Upgrade `rb-sys` to 0.9.15

ianks:
resolves #5668


hsbt:
@ianks Thanks! I will coordinate the supported Ruby versions with rubygems and `rb-sys`.

PR #5680, opened 2022-06-30
Bump `rb-sys` to 0.9.18

ianks:
Addressing https://github.com/rubygems/rubygems/actions/runs/2586659865

### DIFF
--- a/.github/workflows/daily-rubygems.yml
+++ b/.github/workflows/daily-rubygems.yml
@@ -3,6 +3,7 @@ name: daily-rubygems
 on:
   schedule:
     - cron: '0 0 * * *'
+  workflow_dispatch:
 
 jobs:
   daily_rubygems:
--- a/test/rubygems/test_gem_ext_cargo_builder/custom_name/Cargo.lock
+++ b/test/rubygems/test_gem_ext_cargo_builder/custom_name/Cargo.lock
@@ -160,9 +160,9 @@ dependencies = [
 
 [[package]]
 name = "rb-sys"
-version = "0.9.15"
+version = "0.9.18"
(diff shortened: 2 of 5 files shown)


ianks:
I tested this branch against ruby head locally (`ruby 3.2.0dev (2022-07-01T00:08:25Z master c684b1aa26) [arm64-darwin21]`), and the tests pass. I also made the `daily-rubygems` GH action be trigger-able from the actions UI. Let's get this merged so I can test it against CI.

deivid-rodriguez:
@ianks Seems ok to allow manually triggering this workflow, but what I usually do when I need to work on it on a PR is changing the workflow directly like this:

```diff
diff --git a/.github/workflows/daily-rubygems.yml b/.github/workflows/daily-rubygems.yml
index d0b96b5c9c..2eccdad9bf 100644
--- a/.github/workflows/daily-rubygems.yml
+++ b/.github/workflows/daily-rubygems.yml
@@ -1,8 +1,7 @@
 name: daily-rubygems
 
 on:
-  schedule:
-    - cron: '0 0 * * *'
+  pull_request:
 
 jobs:
   daily_rubygems:
```

So that I can verify it's all green, and then once I'm sure I revert that tweak for merging.

Can you try it?

deivid-rodriguez:
I'm not sure what happened here because the ruby-head CI only failed for one day without us making any changes: https://github.com/rubygems/rubygems/actions/workflows/daily-rubygems.yml.

In any case, the update to `rb-sys` was already proposed by dependabot automatically and merged, so this no longer necessary.

I'll close this since I don't think we need the `workflow-dispatch:` change either, but I don't have a strong opinion, so if we feel it's best to introduce it, please propose a separate PR.

Thanks @ianks! ❤️

PR #5681, opened 2022-07-01
Delete partial file and re-raise on Errno::ENOSPC.

### LINKED ISSUE #1664: gem install stops after full disk error
I  have installed Gemfile for one of older versions of rails tutorials and ran out of disk space in process.
After making space, I got greeted with this message:

```
ERROR:  Error installing childprocess:
invalid gem: package metadata is missing in /home/marko/.gem/ruby/2.1.9/cache/childprocess-0.5.9.gem
```

Problem was that cached .gem file had file size 0. It fixed itself after I deleted all 0 size files in cache dir. Gem should automatically reload files in cache if they are 0 size or corrupt. Or at least ask you to do so.

duckinator:

If you have some free space, but not enough for the entire `.gem` file, `gem install SOME_GEM` leaves a corrupt file behind.

Closes #1664.

If `Gem.write_binary(path, data)` fails with `Errno::ENOSPC`, delete any partial file and re-raise the exception.


duckinator:
I'm unsure what the appropriate way to test code which is only triggered by running out of disk space.

I did, however, confirm that it works in practice.

<details>
<summary>Without this patch</summary>

```
PS F:\> rm .\boop-0.5.0.gem
PS F:\> ruby --version
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x64-mingw-ucrt]
PS F:\> gem --version
3.3.17
PS F:\> gem install boop
Fetching boop-0.5.0.gem
#<Thread:0x00000236dd4a3f10 C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/request_set.rb:167 run> terminated with exception (report_on_exception is true):
C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:774:in `write': No space left on device @ io_write - F:/boop-0.5.0.gem (Errno::ENOSPC)
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:774:in `block in write_binary'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:789:in `block in open_file'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:782:in `open'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:782:in `open_file'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:773:in `write_binary'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/remote_fetcher.rb:292:in `cache_update_path'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/remote_fetcher.rb:146:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/source.rb:208:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/resolver/specification.rb:112:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/request_set.rb:173:in `block (2 levels) in install'
ERROR:  While executing gem ... (Errno::ENOSPC)
    No space left on device @ io_write - F:/boop-0.5.0.gem
PS F:\> dir


    Directory: F:\


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         4/15/2018   1:51 AM                efi
d-----         6/30/2022   7:54 PM                build_info
d-----         6/30/2022   7:54 PM                doc
d-----         6/30/2022   7:54 PM                specifications
d-----         6/30/2022   8:39 PM                cache
d-----         6/30/2022   8:39 PM                extensions
d-----         6/30/2022   8:39 PM                gems
d-----         6/30/2022   8:39 PM                plugins
-a----         6/30/2022   8:09 PM           5412 aaaaa.txt
-a----         6/30/2022   8:40 PM              0 boop-0.5.0.gem


PS F:\>
```

</details>

<details>
<summary>With this patch</summary>

```
PS F:\> rm .\boop-0.5.0.gem
PS F:\> ruby --version
ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [x64-mingw-ucrt]
PS F:\> gem --version
3.4.0.dev
PS F:\> gem install boop
Fetching boop-0.5.0.gem
#<Thread:0x000002249e989260 C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/request_set.rb:167 run> terminated with exception (report_on_exception is true):
C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:774:in `write': No space left on device @ io_write - F:/boop-0.5.0.gem (Errno::ENOSPC)
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:774:in `block in write_binary'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:792:in `block in open_file'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:785:in `open'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:785:in `open_file'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems.rb:773:in `write_binary'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/remote_fetcher.rb:292:in `cache_update_path'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/remote_fetcher.rb:146:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/source.rb:208:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/resolver/specification.rb:112:in `download'
        from C:/Ruby31-x64/lib/ruby/site_ruby/3.1.0/rubygems/request_set.rb:173:in `block (2 levels) in install'
ERROR:  While executing gem ... (Errno::ENOSPC)
    No space left on device @ io_write - F:/boop-0.5.0.gem
PS F:\> dir


    Directory: F:\


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         4/15/2018   1:51 AM                efi
d-----         6/30/2022   7:54 PM                build_info
d-----         6/30/2022   7:54 PM                doc
d-----         6/30/2022   7:54 PM                specifications
d-----         6/30/2022   8:39 PM                cache
d-----         6/30/2022   8:39 PM                extensions
d-----         6/30/2022   8:39 PM                gems
d-----         6/30/2022   8:39 PM                plugins
-a----         6/30/2022   8:09 PM           5412 aaaaa.txt


PS F:\>
```

</details>

lib/rubygems.rb
```diff
@@ -773,6 +773,10 @@ def self.write_binary(path, data)
     open_file(path, 'wb') do |io|
       io.write data
     end
+  rescue Errno::ENOSPC => e
+    # If the write failed, delete the file -- it's *guaranteed* to be corrupted.
+    File.delete(path) if File.exist?(path)
```
deivid-rodriguez, line 778: I think an `Errno::ENOSPC` error would _always_ leave a corrupted file so the `File.exist?` check is not necessary?

lib/rubygems.rb
```diff
@@ -773,6 +773,10 @@ def self.write_binary(path, data)
...
+    raise e
```
deivid-rodriguez, line 779: When raising the same error rescued, I generally prefer to let the original exception flow, rather than raising a duplicated one.

```suggestion
    raise
```

deivid-rodriguez (commented):
Looks good, couple of comments!

I also wrote a spec for you, let me know what you think :)

```diff
diff --git a/test/rubygems/test_gem_installer.rb b/test/rubygems/test_gem_installer.rb
index 46bb6965c7..726ac30b04 100644
--- a/test/rubygems/test_gem_installer.rb
+++ b/test/rubygems/test_gem_installer.rb
@@ -2219,6 +2219,37 @@ def test_write_spec_writes_cached_spec
     assert_equal @spec, eval(File.read(@spec.spec_file))
   end
 
+  def test_leaves_no_empty_cached_spec_when_no_more_disk_space
+    @spec = setup_base_spec
+    FileUtils.rm @spec.spec_file
+    assert_path_not_exist @spec.spec_file
+
+    @spec.files = %w[a.rb b.rb c.rb]
+
+    installer = Gem::Installer.for_spec @spec
+    installer.gem_home = @gemhome
+
+    File.class_eval do
+      alias_method :original_write, :write
+
+      def write(data)
+        raise Errno::ENOSPC
+      end
+    end
+
+    assert_raises Errno::ENOSPC do
+      installer.write_spec
+    end
+
+    assert_path_not_exist @spec.spec_file
+  ensure
+    File.class_eval do
+      remove_method :write
+      alias_method :write, :original_write # rubocop:disable Lint/DuplicateMethods
+      remove_method :original_write
+    end
+  end
+
   def test_dir
     installer = setup_base_installer
 
```

lib/rubygems.rb
duckinator: While poking around with this bug, I encountered exactly two scenarios:

1. If you have space left, but not enough for the whole file, it will be corrupt.
2. If you have perfectly filled your hard drive, it will fail with `Errno::ENOSPC` but not create a file.

The `File.delete(path)` addresses # 1 and the `if File.exist?(path)` conditional addresses # 2.

lib/rubygems.rb
duckinator: Sounds good to me. I'll push a change removing the ` => e` and changing `raise e` to `raise`. 👍

duckinator:
> Looks good, couple of comments!

One of those I explained my reasoning for, and the other I addressed. 👍 

> I also wrote a spec for you, let me know what you think :)

Your test is checking for the existence of a gemspec, but only the cached `.gem` file is relevant. I did confirm, however, that since `File.open` is what creates the file, stubbing out `File.write` is a good approach.

So I think we just need to figure out how to reference the `.gem` file placed in the gem cache, as opposed to the gemspec, and I can change the `assert_path_not_exist` call accordingly.

deivid-rodriguez:
Oh, I see. Well the test covers the change perfectly anyways, even if it does not represent the realworld test case you run into. It fails without the patch and passes with it, it just tests another instance of the same bug, so I think it's fine as is and we don't need anything else.

lib/rubygems.rb
deivid-rodriguez: I think case 2 is weird, what raises `Errno::ENOSPC`? I assume it's not `File.open` because otherwise it would leave an empty file, but I'm not sure what it is then. Ideally I'd like to cover case 2 with a spec too, that's why I'm asking. Can you reproduce scenario 2 with `/dev/full?

duckinator:
@deivid-rodriguez the reason I noticed it was doing something different was that it failed for me both with _and_ without my changes:

```
Failure: test_leaves_no_empty_cached_spec_when_no_more_disk_space(TestGemInstaller): <"C:/Users/puppy/Desktop/dev/rubygems/tmp/test_rubygems_20220702-8712-bo3pt3/gemhome/specifications/a-2.gemspec"> was expected to not exist
./test/rubygems/test_gem_installer.rb:2244:in `test_leaves_no_empty_cached_spec_when_no_more_disk_space'
```

I'll look into this again when I have some free time.

deivid-rodriguez:
Ok, let me know!

lib/rubygems.rb
duckinator: My understanding is:

- in scenario 1, the error is raised by `io.write`
- in scenario 2, the error is raised by `File.open`

If `File.open` successfully creates the file, there was enough space. And then `io.write` can fail, giving a corrupt file.

If `File.open` _does not_ successfully create the file, it'd raise some error &mdash; and `Errno::ENOSPC` would make sense. This would result in an error, but no file.

duckinator: When I work on this PR next, I'll try to reproduce.

duckinator:
I rebased this earlier and made another attempt at writing a test for this, but it's proving a bit challenging, heh.

deivid-rodriguez:
Hei @duckinator! I can see how this can be challenging 😅. We already have a test for one of the two scenarios being fixed here (case 1 above). I would be happy with either:

* Ship the other fix (case 2) as well without a test as long as we can at least verify it's fixing the issue with some manual steps.
* Ship the fix for case 1 and its test for now and leave the other for a separate PR so that at least we don't block half of this improvement.

What do you think?

duckinator:
@deivid-rodriguez Is the test you're referring to the one you mentioned in [your review back in july](https://github.com/rubygems/rubygems/pull/5681#pullrequestreview-1025804134)? Looking at it again, I'm not sure why it didn't work when I tried it before &mdash; the code all makes sense when I look at it, so I'm wondering if there's some subtle detail that's off as opposed to it being fundamentally wrong. I'll poke at it again when I pick this back up.

A bit more detail about the code, btw: The only code specific to scenario 2 is the `if File.exist?(path)`. Without that, this PR addresses scenario 1, but for scenario 2 changes the error from `Errno::ENOSPC` to an error about trying to delete a nonexistent file. So honestly I think just not having a test for that is _probably_ okay?

So when I next pick this up I'll poke at integrating your test (or some variant thereof) again, and try to actually _finally_ get this PR wrapped up.

I won't be able to get to this again until sometime next week. If I don't have an update here by next Wednesday (November 9th), feel free to ping me here as a reminder.

**Note for future me:** There's details on how to reproduce this over at https://github.com/rubygems/rubygems/issues/1664#issuecomment-536096584.

deivid-rodriguez:
Oh, yes, I mean that test, yeah. I actually thought we had already incorporated it to this PR!

deivid-rodriguez:
Take care @duckinator 🤗

duckinator:
Rebased and added a test. I confirmed that keeping the test and reverting the fix causes it the test fail.

I think this is ready for review, @deivid-rodriguez!

duckinator:
Accidentally used `assert_raises` instead of `assert_raise` &mdash; that should be fixed now.

duckinator:
Last force-push was removing an @-mention of @deivid-rodriguez from a commit message, since [doc/development/PULL_REQUESTS.md says to avoid that](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages).

deivid-rodriguez (approved):
Thanks for working on this @duckinator! 💪 

I think you're write and the second cases we're protecting against here could also be raised (the `Errno::ENOSPC` potentially raised by `File.open`). I think that could happen if he filesystem has run out of i-nodes, but I haven't been able to reproduce it. In any case, it seems fine to ship that without a test.

Also, I think this may also fix the underlying root cause for https://github.com/rubygems/rubygems/issues/5981?

duckinator:
> I think that could happen if he filesystem has run out of i-nodes, but I haven't been able to reproduce it.

On Linux, if you have less than the block size (frequently 4096 bytes) available and try to write a file, it fails because that's not enough to even create an empty file.

> In any case, it seems fine to ship that without a test.

So this is good to go, then? If so feel free to merge it. It'll likely be a week or two at least before I get back to RubyGems stuff, and there's no reason for it to sit around that long if it's not needed.

> Also, I think this may also fix the underlying root cause for https://github.com/rubygems/rubygems/issues/5981?

It's definitely a likely cause of that, yeah.

PR #5694, opened 2022-07-06
Fix TruffleRuby no longer installing some lockfiles correctly

### LINKED ISSUE #5691: Bundler 2.3.17 installs precompiled CRuby gems on TruffleRuby

```ruby
source 'https://rubygems.org'

gem 'nokogiri'
```

With the latest truffleruby-dev (from ruby-build):
```
$ bundle
Resolving dependencies...
Using bundler 2.3.17
Using racc 1.6.0
Using mini_portile2 2.8.0
Using nokogiri 1.13.6 (x86_64-linux)
Bundle complete! 1 Gemfile dependency, 4 gems now installed.
```
That is incorrect, it must not be `Using nokogiri 1.13.6 (x86_64-linux)`, it must be `Installing nokogiri 1.13.6 with native extensions` (i.e., compiling from source).
Trying to use it results in errors (expected since it's using the CRuby-precompiled gem):
```
$ ruby -rnokogiri -e0
<internal:core> core/kernel.rb:293:in `require_relative': /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/nokogiri-1.13.6-x86_64-linux/lib/nokogiri/3.0/nokogiri.so: undefined symbol: rb_eTypeError (RuntimeError)
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/nokogiri-1.13.6-x86_64-linux/lib/nokogiri/extension.rb:7:in `<top (required)>'
	from <internal:core> core/kernel.rb:293:in `require_relative'
	from /home/eregon/.rubies/truffleruby-dev/lib/gems/gems/nokogiri-1.13.6-x86_64-linux/lib/nokogiri.rb:10:in `<top (required)>'
	from <internal:core> core/kernel.rb:234:in `gem_original_require'
	from <internal:/home/eregon/.rubies/truffleruby-dev/lib/mri/rubygems/core_ext/kernel_require.rb>:160:in `require'
	from <internal:core> core/unbound_method.rb:18:in `bind_call'
	from <internal:core> core/kernel.rb:272:in `require'
<internal:core> core/kernel.rb:236:in `gem_original_require': cannot load such file -- nokogiri (LoadError)
	from <internal:/home/eregon/.rubies/truffleruby-dev/lib/mri/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from <internal:core> core/unbound_method.rb:18:in `bind_call'
	from <internal:core> core/kernel.rb:272:in `require'
```

2.3.16 and 2.2.32 work fine (after cleaning up/reinstalling truffleruby-dev to remove the bad precompiled nokogiri gem):
```
$ bundle   
Resolving dependencies...
Using bundler 2.3.16
Using racc 1.6.0
Using mini_portile2 2.8.0
Installing nokogiri 1.13.6 with native extensions
Bundle complete! 1 Gemfile dependency, 4 gems now installed.
```

Yes, that's what causes the problem.

## Environment

```
Bundler       2.3.16
  Platforms   ruby
Ruby          3.0.3p0 (2022-07-05 revision be96ee562944f863e7e72ed750f336b939e5d9c3) [x86_64-linux]
RubyGems      3.2.32
  Gem Path    /home/eregon/.gem/truffleruby/3.0.3.7:/home/eregon/.rubies/truffleruby-dev/lib/gems
  User Home   /home/eregon
  User Path   /home/eregon/.gem/truffleruby/3.0.3.7
Tools         
  Git         2.35.3
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

## Bundler Build Metadata

```
Built At          2022-06-15
Git SHA           324ee6e542
Released Version  true
```

I thought this could relate to https://github.com/rubygems/rubygems/pull/4049 but that's not released yet, right @deivid-rodriguez?
cc @andrykonchin who first saw this issue.

deivid-rodriguez:

Until 2.3.17, would always use the RUBY platform for Truffleruby, so the RUBY platform would always be used in lockfiles generated on Truffleruby.

From 2.3.18, Bundler will consider `Gem::Platform.local` as Truffleruby's platform, just like CRuby, and use that inside lockfiles.

But we need to properly deal with old lockfiles that used the RUBY platform, because currently Bundler tries to install the most specific variant available and that's usually not expected on Truffleruby.

My fix is to use TruffleRuby platform selection logic (according to the gem name) not only for resolution but also for materialization, to make sure we never materialize a lockfile into an incorrect platform specific gem.

Closes #5691.
 
### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -74,6 +74,7 @@ bundler/lib/bundler/fetcher/compact_index.rb
 bundler/lib/bundler/fetcher/dependency.rb
 bundler/lib/bundler/fetcher/downloader.rb
 bundler/lib/bundler/fetcher/index.rb
+bundler/lib/bundler/force_platform.rb
 bundler/lib/bundler/friendly_errors.rb
 bundler/lib/bundler/gem_helper.rb
 bundler/lib/bundler/gem_helpers.rb
--- a/bundler/lib/bundler/dependency.rb
+++ b/bundler/lib/bundler/dependency.rb
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 
 require "rubygems/dependency"
+require_relative "force_platform"
 require_relative "shared_helpers"
 require_relative "rubygems_ext"
 
 module Bundler
   class Dependency < Gem::Dependency
+    include ForcePlatform
+
     attr_reader :autorequire
     attr_reader :groups, :platforms, :gemfile, :git, :github, :branch, :ref, :force_ruby_platform
 
@@ -158,18 +161,6 @@ module Bundler
       super
     rescue NoMethodError
       requirement != ">= 0"
-    end
-
-    private
-
-    # The `:force_ruby_platform` attribute is `false` by default, except for
-    # TruffleRuby. TruffleRuby generally needs to force the RUBY platform
-    # variant unless the name is explicitly allowlisted.
-
-    def default_force_ruby_platform
-      return false unless Bundler.current_ruby.truffleruby?
-
-      !Gem::Platform::REUSE_AS_BINARY_ON_TRUFFLERUBY.include?(name)
     end
   end
 end
--- a/bundler/lib/bundler/force_platform.rb
+++ b/bundler/lib/bundler/force_platform.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Bundler
+  module ForcePlatform
+    private
+
+    # The `:force_ruby_platform` value used by dependencies for resolution, and
+    # by locked specifications for materialization is `false` by default, except
+    # for TruffleRuby. TruffleRuby generally needs to force the RUBY platform
+    # variant unless the name is explicitly allowlisted.
+
+    def default_force_ruby_platform
+      return false unless Bundler.current_ruby.truffleruby?
+
+      !Gem::Platform::REUSE_AS_BINARY_ON_TRUFFLERUBY.include?(name)
+    end
+  end
+end
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -1,13 +1,16 @@
 # frozen_string_literal: true
 
+require_relative "force_platform"
 require_relative "match_platform"
 
 module Bundler
   class LazySpecification
+    include ForcePlatform
     include MatchPlatform
 
     attr_reader :name, :version, :dependencies, :platform
-    attr_accessor :source, :remote, :force_ruby_platform
+    attr_writer :force_ruby_platform
+    attr_accessor :source, :remote
 
     def initialize(name, version, platform, source = nil)
       @name          = name
@@ -16,6 +19,7 @@ module Bundler
       @platform      = platform || Gem::Platform::RUBY
       @source        = source
       @specification = nil
+      @force_ruby_platform = nil
     end
 
     def full_name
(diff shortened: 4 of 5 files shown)


eregon:
Thank you!
Could you also test that an existing lockfile generated by CRuby and so containing a local platform and even the nokogiri platform-specific gem name installs correctly on truffleruby? (unless that's already tested?)

For instance we initially noticed the issue of #5691 by `bundle install` in `stimulus_reflex` which has a checked-in Gemfile.lock:
https://github.com/stimulusreflex/stimulus_reflex/blob/master/Gemfile.lock
with things like:
```
...
    nokogiri (1.13.6-x86_64-darwin)
      racc (~> 1.4)
    nokogiri (1.13.6-x86_64-linux)
      racc (~> 1.4)
...
PLATFORMS
  x86_64-darwin-19
  x86_64-linux
...
```

deivid-rodriguez:
Nice catch. If I had to guess I'd say it doesn't currently work as expected. I will find some time to verify that and try to fix it if necessary.

deivid-rodriguez:
I had a look and this is indeed still broken. I'm not sure how to best fix it so I created [a PR](https://github.com/rubygems/rubygems/pull/5711) to revert the whole new approach to supporting TruffleRuby using `:force_ruby_platform`, so that we can get things back to working. I might retry the idea in the future.

eregon:
Maybe Bundler could detect the platforms changed, and in that case keep gem versions the same but re-resolve which variant of each gem to use?
Also I would have thought the `:force_ruby_platform` should force a ruby platform even if the lockfile says otherwise (a bit similar to someone adding `force_ruby_platform: true` in their `Gemfile`).

deivid-rodriguez:
> Maybe Bundler could detect the platforms changed, and in that case keep gem versions the same but re-resolve which variant of each gem to use?

How can we detect that platforms changed? Unless TruffleRuby reports its own platform, I'm not sure how to do it.

> Also I would have thought the :force_ruby_platform should force a ruby platform even if the lockfile says otherwise (a bit similar to someone adding force_ruby_platform: true in their Gemfile).

Yes, I expect this too, but the logic is complicated and since I want to release today I had no time to fix it. I can retry "the new way" in the future, but for now I just want things to work again.

eregon:
> How can we detect that platforms changed?

I'm not sure either, but for instance we could detect "only 'ruby' platform" vs "other platforms than 'ruby'".

> Unless TruffleRuby reports its own platform (and we record it in the lockfile)

I'm not sure if this would help, but I can consider it if it would help.

> Yes, I expect this too, but the logic is complicated and since I want to release today I had no time to fix it. I can retry "the new way" in the future, but for now I just want things to work again.

OK.

deivid-rodriguez:
> I'm not sure if this would help, but I can consider it if it would help.

In the case you reported, the current platform is `x86_64-linux` and that's in the lockfile, so Bundler thinks there's no need to re-resolve due to platform changes. In the previous implementation, TruffleRuby reports itself as "ruby", so Bundler thinks it needs to re-resolve due to platform changes and that's why things work. But with the "force_ruby_platform" implementation, TruffleRuby also reports the "real platform", x86_64-linux, so things no longer work as expected.

I could've tried to keep the "force_ruby_platform" changes but still let TruffleRuby be reported as "ruby", but again, I just wanted to fix things for now so I reverted the two commits involved.

I will make a proposal if I feel we need TruffleRuby to report something specific as its platform when I retry this in the future 👍.

PR #5650, opened 2022-06-24
Add platform alias `:windows` to Gemfile DSL

johnnyshields:
You'll typically see this in the Gemfile of a Windows user:

```ruby
platforms :mswin, :mswin64, :mingw, :x64_mingw do
  gem 'tzinfo-data'
  gem 'wdm', '>= 0.1.0'
end
```

This is quite silly. Let's make it possible to do this:

```ruby
platforms :windows do
  gem 'tzinfo-data'
  gem 'wdm', '>= 0.1.0'
end
```

In addition, this PR includes some changes to Bunder.current_ruby. I'll break those out into another PR probably.

This PR includes the changes from #5649. It does not change/break any existing platform behavior (aside from what has changed in #5649.)

PR #5699, opened 2022-07-09
Remove windows extra methods

johnnyshields:
[Recently we added](https://github.com/rubygems/rubygems/pull/5650) `CurrentRuby#windows?` and Ruby-version specific methods e.g. `CurrentRuby#windows_30?`

This PR removes the following methods from `CurrentRuby` class:

```
CurrentRuby#mswin?
CurrentRuby#mswin64?
CurrentRuby#mingw?
CurrentRuby#x64_mingw?

and version-specific CurrentRuby#mingw_30? etc
```

These methods are not used internally and I doubt they are widely-used externally. They can certainly be replaced with `windows?`. Maybe good to merge this for Bundler 4.0 to be on the safe side.

### DIFF
--- a/bundler/lib/bundler/current_ruby.rb
+++ b/bundler/lib/bundler/current_ruby.rb
@@ -29,15 +29,11 @@ module Bundler
     KNOWN_PLATFORMS = %w[
       jruby
       maglev
-      mingw
       mri
-      mswin
-      mswin64
       rbx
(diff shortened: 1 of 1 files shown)


johnnyshields:
Hmm... these are still needed to handle `platforms` macro in legacy Bundler versions. Closing PR.

PR #5700, opened 2022-07-09
Make lockfiles generated on macOS include a lock for Linux by default

### LINKED ISSUE #4269: Bundler 2.2.3 generated Gemfile.lock on a mac cannot be used on linux out of the box

Here's an example repo: https://github.com/schneems/bundler_223_fails_deploy. It has one dependency `rake` and the Gemfile.lock was generated by running `bundle install` on my mac. When I try to use this Gemfile/Gemfile.lock combo it fails:

```
docker run -it --rm ruby:2.7 bash
git clone https://github.com/schneems/bundler_223_fails_deploy
cd bundler_223_fails_deploy
export BUNDLE_WITHOUT='development:test'
export BUNDLE_DEPLOYMENT=1
export BUNDLE_PATH=vendor/bundle
bundle install -j4
```

Then

```
 bundle exec rake -P
bundler: failed to load command: rake (/bundler_223_fails_deploy/vendor/bundle/ruby/2.7.0/bin/rake)
Gem::Exception: can't find executable rake for gem rake. rake is not currently included in the bundle, perhaps you meant to add it to your Gemfile?
  /usr/local/lib/ruby/2.7.0/bundler/rubygems_integration.rb:374:in `block in replace_bin_path'
  /usr/local/lib/ruby/2.7.0/bundler/rubygems_integration.rb:402:in `block in replace_bin_path'
  /bundler_223_fails_deploy/vendor/bundle/ruby/2.7.0/bin/rake:23:in `<top (required)>'
```

I expect that a Gemfile.lock produced on mac can also work on linux

### What actually happened?

The gem appears to install, but then it cannot be loaded.

### Extra context 

Either adding `ruby` to the `PLATFORMS` section or `x86_64-linux` fixes the issue, but this is not immediately clear.

<details>
  <summary>bundle env (on my mac)</summary>
## Environment

```
Bundler       2.2.3
  Platforms   ruby, x86_64-darwin-19
Ruby          3.0.0p0 (2020-12-25 revision 95aff214687a5e12c3eb57d056665741e734c188) [x86_64-darwin19]
  Full Path   /Users/rschneeman/.rubies/ruby-3.0.0/bin/ruby
  Config Dir  /Users/rschneeman/.rubies/ruby-3.0.0/etc
RubyGems      3.2.3
  Gem Home    /Users/rschneeman/.gem/ruby/3.0.0
  Gem Path    /Users/rschneeman/.gem/ruby/3.0.0:/Users/rschneeman/.rubies/ruby-3.0.0/lib/ruby/gems/3.0.0
  User Home   /Users/rschneeman
  User Path   /Users/rschneeman/.gem/ruby/3.0.0
  Bin Dir     /Users/rschneeman/.gem/ruby/3.0.0/bin
Tools         
  Git         2.28.0
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

## Bundler Build Metadata

```
Built At          2020-12-22
Git SHA           29dc3c8398
Released Version  true
```

## Bundler settings

```
jobs
  Set for the current user (/Users/rschneeman/.bundle/config): 8
build.nokogiri
  Set for the current user (/Users/rschneeman/.bundle/config): "--use-system-libraries"
gem.test
  Set for the current user (/Users/rschneeman/.bundle/config): "rspec"
gem.mit
  Set for the current user (/Users/rschneeman/.bundle/config): true
gem.coc
  Set for the current user (/Users/rschneeman/.bundle/config): true
--path
  Set for the current user (/Users/rschneeman/.bundle/config): "vendor/"
```

### Gemfile

```ruby
source 'https://rubygems.org'
gem 'rake'
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    rake (13.0.3)

PLATFORMS
  x86_64-darwin-19

DEPENDENCIES
  rake

BUNDLED WITH
   2.2.3
```

</details>

<details>
  <summary>bundle env (on docker)</summary>
## Environment

```
Bundler       2.1.4
Ruby          2.7.2p137 (2020-10-01 revision 5445e0435260b449decf2ac16f9d09bae3cafe72) [x86_64-linux]
  Full Path   /usr/local/bin/ruby
  Config Dir  /usr/local/etc
RubyGems      3.1.4
  Gem Home    /usr/local/bundle
  Gem Path    /root/.gem/ruby/2.7.0:/usr/local/lib/ruby/gems/2.7.0:/usr/local/bundle
  User Home   /root
  User Path   /root/.gem/ruby/2.7.0
  Bin Dir     /usr/local/bundle/bin
Tools
  Git         2.20.1
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2021-01-11
Git SHA           unknown
Released Version  false
```

## Bundler settings

```
jobs
  Set for your local app (/usr/local/bundle/config): 4
without
  Set for your local app (/usr/local/bundle/config): [:development, :test]
  Set via BUNDLE_WITHOUT: [:development, :test]
app_config
  Set via BUNDLE_APP_CONFIG: "/usr/local/bundle"
deployment
  Set via BUNDLE_DEPLOYMENT: true
silence_root_warning
  Set via BUNDLE_SILENCE_ROOT_WARNING: true
path
  Set via BUNDLE_PATH: "vendor/bundle"
```

### Gemfile

```ruby
source 'https://rubygems.org'
gem 'rake'
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    rake (13.0.3)

PLATFORMS
  x86_64-darwin-19

DEPENDENCIES
  rake

BUNDLED WITH
   2.2.3
```
</details>

### LINKED ISSUE #5338: Bundler platform locking causes problems out-of-box, on CI, on Heroku etc.

I'm hopeful we can come up with a compromise to a significant pain point that's worsening over time, not getting better.

In Bundler 2.2, `PLATFORMS` in `Gemfile.lock` only includes the local machine's architecture when that machine causes a `Gemfile.lock` to be built (but confusingly, not if it is merely updated). Bundler lockfiles cease to become architecture independent, even though Ruby is supposed to be. Any attempt to install using that lockfile on any other architecture fails unless new architectures are manually added. This even includes developers upgrading e.g. macOS versions from 11 to 12 and similar, since their platform includes a Darwin kernel version. On Heroku, this broke the world for a while, as the 2.2 approach appeared to [not be backwards compatible with Bundler 2.1](https://www.moncefbelyamani.com/understanding-the-gemfile-lock-file/) that Heroku provided at the time.

Developers are forced to constantly remember to check, double check and where necessary hand-maintain their Gemfile.lock's `PLATFORMS`, to nanny things through CI (e.g. macOS development, AWS Linux CI) or just allow for the fact that they might not be on the same precise e.g. macOS version as their peers, or might be using Arch Linux, or a different CPU (e.g. M1), or whatever.

There is presently no clean way around this. Unfrozen is too permissive and has unrelated side effect of mutable Ruby gem versions, rather than _only_ dealing with native extension platforms. Bundler should not be locking itself to the arbitrary architecture of the local developer machine IMHO; its job is to lock gem versions. I am aware of [this comment](https://github.com/rubygems/rubygems/issues/4269#issuecomment-758564690) but it doesn't really seem to resonate; the "safety" argument talks about a gem which doesn't have a particular architecture variant and installs then runs on CI presumably building "generic" native extensions, CI passes, then subsequently an architecture-specific variant is pushed but it's buggy and this gets deployed _after_ the CI run to Production. If the timing of things were thus, it'd still happen regardless of what `PLATFORMS` might say - it'd still build generic on CI because no arch-specific version existed. Meanwhile, if malicious code is being pushed to RubyGems, then there are far bigger issues at play and far greater concern than specific platform variant locking.

A reasonable solution IMHO might be to provide a configuration option that allows alternative platforms, as if the running platform was included in the `PLATFORMS` array, without needing to modify `Gemfile.lock`. That could be a `bundle config` item or env var. That's not a question of just forcing to `ruby`, the resolution of which allegedly is buggy or unsafe and is the whole reason why the new 2.2 behaviour was introduced anyway - though if Bundler **guaranteed and formally documented** that forcing to `ruby` was still fully supported, perhaps that would be acceptable (now that it works again, [thanks to this fix](https://github.com/rubygems/rubygems/pull/5304)!). Otherwise we're building on sand.

Yes. Bundler 2.2.x is working as designed. I do not believe this design is helpful, with widespread community posts across forums and blogs indicating the developer pain it has inadvertently introduced.

Cause a `Gemfile.lock` to be generated under Bundler 2.2.x on one architecture. Try to `bundle install` on another. This will fail, even though Ruby and gems all work just fine on the target architecture. In Bundler 2.1 and all earlier versions, the gems would install.

* If the gem is not compatible with the chosen platform, the native extension build will fail, just as it always has.
* If the gem can build on the chosen platform, then it should do so, and not have Bundler fail the whole thing.

### What actually happened?

Bundler refuses to install the gems:

```
Your bundle only supports platforms ["x86_64-darwin-21"] but your local platform
is x86_64-linux. Add the current platform to the lockfile with `bundle lock
--add-platform x86_64-linux` and try again.
```

deivid-rodriguez:

Locking only the specific platform in lockfiles causes issues in CI platforms or Heroku, that run on Linux. This issue only appeared after we started locking the specific platform in the lockfile. Previously, although platform specific resolution would not always be correct, the general case would just work because platform specific variants where chosen at install time (they were not recorded in the lockfile at all).

My fix is to, after resolving for the current platform like we do now, check whether the resolution found is also compatible with Linux (the 99% case), and if so, add the linux platform and linux specific variants to the lockfile.

This way, new lockfiles generated on Darwin should just work on Linux.

For example, for the following `Gemfile`, 

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

# gem "rails"

gem "nokogiri", "~> 1.13"
```

running `bundle install` on my MacOS laptop results in

```
$ bundle install
Resolving dependencies...
Using racc 1.6.0
Using bundler 2.4.0.dev
Using nokogiri 1.13.6 (arm64-darwin)
```

generating the following `Gemfile.lock` file, that should just work on Linux

```
GEM
  remote: https://rubygems.org/
  specs:
    nokogiri (1.13.6-arm64-darwin)
      racc (~> 1.4)
    nokogiri (1.13.6-x86_64-linux)
      racc (~> 1.4)
    racc (1.6.0)

PLATFORMS
  arm64-darwin-21
  x86_64-linux

DEPENDENCIES
  nokogiri (~> 1.13)

BUNDLED WITH
   2.4.0.dev
```

Closes #4269.
Closes #5338.

### DIFF
--- a/bundler/lib/bundler/current_ruby.rb
+++ b/bundler/lib/bundler/current_ruby.rb
@@ -43,7 +43,7 @@ module Bundler
     ].freeze
 
     def ruby?
-      return true if Bundler::GemHelpers.generic_local_platform == Gem::Platform::RUBY
+      return true if Bundler::GemHelpers.generic_local_platform_is_ruby?
 
       !windows? && (RUBY_ENGINE == "ruby" || RUBY_ENGINE == "rbx" || RUBY_ENGINE == "maglev" || RUBY_ENGINE == "truffleruby")
     end
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -84,7 +84,7 @@ module Bundler
       @new_platform = nil
       @removed_platform = nil
 
-      if lockfile && File.exist?(lockfile)
+      if lockfile_exists?
         @lockfile_contents = Bundler.read_file(lockfile)
         @locked_gems = LockfileParser.new(@lockfile_contents)
         @locked_platforms = @locked_gems.platforms
@@ -302,6 +302,10 @@ module Bundler
       end
     end
 
+    def should_complete_platforms?
+      !lockfile_exists? && generic_local_platform_is_ruby? && !Bundler.settings[:force_ruby_platform]
+    end
+
     def spec_git_paths
       sources.git_sources.map {|s| File.realpath(s.path) if File.exist?(s.path) }.compact
     end
@@ -491,6 +495,10 @@ module Bundler
 
     private
 
+    def lockfile_exists?
+      lockfile && File.exist?(lockfile)
+    end
+
     def resolver
       @resolver ||= Resolver.new(resolution_packages, gem_version_promoter)
     end
@@ -567,11 +575,12 @@ module Bundler
     end
 
     def start_resolution
-      result = resolver.start
+      result = SpecSet.new(resolver.start)
 
       @resolved_bundler_version = result.find {|spec| spec.name == "bundler" }&.version
+      @platforms = result.complete_platforms!(platforms) if should_complete_platforms?
 
-      SpecSet.new(SpecSet.new(result).for(dependencies, false, @platforms))
+      SpecSet.new(result.for(dependencies, false, @platforms))
     end
 
     def precompute_source_requirements_for_indirect_dependencies?
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -592,7 +592,7 @@ module Bundler
     end
 
     def current_ruby_platform_locked?
-      return false unless generic_local_platform == Gem::Platform::RUBY
+      return false unless generic_local_platform_is_ruby?
       return false if Bundler.settings[:force_ruby_platform] && !@platforms.include?(Gem::Platform::RUBY)
 
       current_platform_locked?
--- a/bundler/lib/bundler/gem_helpers.rb
+++ b/bundler/lib/bundler/gem_helpers.rb
@@ -34,6 +34,11 @@ module Bundler
     end
     module_function :local_platform
 
+    def generic_local_platform_is_ruby?
+      generic_local_platform == Gem::Platform::RUBY
+    end
+    module_function :generic_local_platform_is_ruby?
+
     def platform_specificity_match(spec_platform, user_platform)
       spec_platform = Gem::Platform.new(spec_platform)
 
--- a/bundler/lib/bundler/gem_helpers.rb
+++ b/bundler/lib/bundler/gem_helpers.rb
@@ -114,8 +114,6 @@ module Bundler
 
     def same_deps(spec, exemplary_spec)
       same_runtime_deps = spec.dependencies.sort == exemplary_spec.dependencies.sort
-      return same_runtime_deps unless spec.is_a?(Gem::Specification) && exemplary_spec.is_a?(Gem::Specification)
-
       same_metadata_deps = spec.required_ruby_version == exemplary_spec.required_ruby_version && spec.required_rubygems_version == exemplary_spec.required_rubygems_version
       same_runtime_deps && same_metadata_deps
     end
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -7,8 +7,8 @@ module Bundler
     include MatchPlatform
     include ForcePlatform
 
-    attr_reader :name, :version, :dependencies, :platform
-    attr_accessor :source, :remote, :force_ruby_platform
+    attr_reader :name, :version, :platform
+    attr_accessor :source, :remote, :force_ruby_platform, :dependencies
 
     def initialize(name, version, platform, source = nil)
       @name          = name
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -8,12 +8,14 @@ module Bundler
     include ForcePlatform
 
     attr_reader :name, :version, :platform
-    attr_accessor :source, :remote, :force_ruby_platform, :dependencies
+    attr_accessor :source, :remote, :force_ruby_platform, :dependencies, :required_ruby_version, :required_rubygems_version
 
     def initialize(name, version, platform, source = nil)
       @name          = name
       @version       = version
       @dependencies  = []
+      @required_ruby_version = Gem::Requirement.default
+      @required_rubygems_version = Gem::Requirement.default
       @platform      = platform || Gem::Platform::RUBY
       @source        = source
       @force_ruby_platform = default_force_ruby_platform
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -10,6 +10,14 @@ module Bundler
     attr_reader :name, :version, :platform
     attr_accessor :source, :remote, :force_ruby_platform, :dependencies, :required_ruby_version, :required_rubygems_version
 
+    def self.from_spec(s)
+      lazy_spec = new(s.name, s.version, s.platform, s.source)
+      lazy_spec.dependencies = s.dependencies
+      lazy_spec.required_ruby_version = s.required_ruby_version
+      lazy_spec.required_rubygems_version = s.required_rubygems_version
+      lazy_spec
+    end
+
     def initialize(name, version, platform, source = nil)
       @name          = name
       @version       = version
--- a/bundler/lib/bundler/lockfile_parser.rb
+++ b/bundler/lib/bundler/lockfile_parser.rb
@@ -251,8 +251,7 @@ module Bundler
 
         version = Gem::Version.new(version)
         platform = platform ? Gem::Platform.new(platform) : Gem::Platform::RUBY
-        @current_spec = LazySpecification.new(name, version, platform)
-        @current_spec.source = @current_source
+        @current_spec = LazySpecification.new(name, version, platform, @current_source)
         @current_source.add_dependency_names(name)
 
         @specs[@current_spec.full_name] = @current_spec
--- a/bundler/lib/bundler/resolver/spec_group.rb
+++ b/bundler/lib/bundler/resolver/spec_group.rb
@@ -25,11 +25,8 @@ module Bundler
 
       def to_specs(force_ruby_platform)
         @specs.map do |s|
-          lazy_spec = LazySpecification.new(name, version, s.platform, source)
+          lazy_spec = LazySpecification.from_spec(s)
           lazy_spec.force_ruby_platform = force_ruby_platform
-          lazy_spec.dependencies = s.dependencies
-          lazy_spec.required_ruby_version = s.required_ruby_version
-          lazy_spec.required_rubygems_version = s.required_rubygems_version
           lazy_spec
         end
       end
--- a/bundler/lib/bundler/resolver/spec_group.rb
+++ b/bundler/lib/bundler/resolver/spec_group.rb
@@ -27,7 +27,7 @@ module Bundler
         @specs.map do |s|
           lazy_spec = LazySpecification.new(name, version, s.platform, source)
           lazy_spec.force_ruby_platform = force_ruby_platform
-          lazy_spec.dependencies.replace s.dependencies
+          lazy_spec.dependencies = s.dependencies
           lazy_spec
         end
       end
--- a/bundler/lib/bundler/resolver/spec_group.rb
+++ b/bundler/lib/bundler/resolver/spec_group.rb
@@ -28,6 +28,8 @@ module Bundler
           lazy_spec = LazySpecification.new(name, version, s.platform, source)
           lazy_spec.force_ruby_platform = force_ruby_platform
           lazy_spec.dependencies = s.dependencies
+          lazy_spec.required_ruby_version = s.required_ruby_version
+          lazy_spec.required_rubygems_version = s.required_rubygems_version
           lazy_spec
         end
       end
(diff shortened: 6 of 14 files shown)


indirect:
This is very much a new idea and I might be wrong, but should we print a message when we do this? “Automatically adding linux-x86 to the lockfile for future deployments.” or something? Also, what if I deploy to linux-arm machines and I don’t want to automatically lock to linux-x86? Can we provide an option to disable this, or configure the platform that is always automatically added?

deivid-rodriguez:
Yeah, I figured the desire to disable this could come up but then I thought that this doesn't really affect end users in any way, every thing should work just as it worked before, so I'm not sure people anybody will want to disable this or configure it in any way. The only reason would be "lockfile aesthetics"?

Anyways, I'm not opposed to adding a configuration to toggle this off. I wouldn't like to allow configuring the specific extra platform because Linux is the 99% case and because there's a bit of confusion regarding platforms in the codebase and I'm really not sure how this will behave for other platforms (in particular Windows and Java).

I would like to figure things out before committing to adding things, but this seems like a handy, mostly transparent, improvement. Basically, it's just running `bundle lock --add-platform x86_64-linux` for our users so things just work for them.

deivid-rodriguez:
Regarding the message, I think it goes together with configuring this. If we add a configuration option, then logging something definitely makes sense. But if not, I wouldn't log anything because then maybe someone will want to disable it just because they saw the message 😆.

Also, something that makes me hesitate allowing this to be configured is that my initial plan was to "just add all platforms where the resolution is valid". So allowing this to be configured feels like an opposite direction.

Finally, if we configure this, I think there's two options:

* A standard configuration option.
* A Gemfile DSL for the platforms to be locked.

deivid-rodriguez:
Well, after a bit more thought, users are probably going to request this anyways, and making this a configuration that defaults to linux seems simple enough. Plus logging a message like the one you mentioned makes things also clear for users. Plus @schneems already suggested exactly this solution at https://github.com/rubygems/rubygems/issues/4269#issuecomment-759076800.

So I think let's implement both of your suggestions @indirect.

deivid-rodriguez:
Sorry for too many message but also I just noticed that I did a bit of diagonal reading of your first message and missed this part 😳

> what if I deploy to linux-arm machines 

So definitely this is not just lockfile aesthetics and makes sense to allow configuring it 👍. I will work on it!

fauno:
>> what if I deploy to linux-arm machines 
>
> So definitely this is not just lockfile aesthetics and makes sense to
> allow configuring it 👍. I will work on it!

also `x86_64-linux-musl`!

indirect:
@deivid-rodriguez If you think it makes sense, that sounds good to me. :)

pond:
As others have pointed out, there's a question of _which_ Linux.

I don't understand why `ruby` isn't just added as a fallback, since this has historically worked "everywhere" for Heroku, CI systems etc., but would still allow for the local developer's lock to be added to.

I still also don't see why the local developer machine's architecture should have any bearing on lock file whatsoever since it says nothing about anywhere else that the gems might be installed and the whole point of a Gemfile is to be a _platform-independent_ way to specify you required Ruby gem versions, but that's another debate.

indirect:
I have no idea where you got the idea that the entire point of a Gemfile is to be platform independent, but for the record: it is not.

Gemfiles are usually able to be applied to multiple platforms, but sometimes they can’t be. The generic ruby platform is often workable, but sometimes isn’t. The lockfile in particular cannot provide Bundler’s basic guarantee of “the same code” as soon as two platforms are involved—Heroku has never actually “just worked” this entire time, it has just been true that the most common case didn’t run into any of the cross platform issues.

indirect:
(Bundler creating an additional lock resolution for each platform is the only solution we have for this set of problems, and it is why we now have multiple platforms in the lockfile.)

deivid-rodriguez:
The time RubyGems allowed versions of gems precompiled for specific architectures, lockfile stopped being platform independent (and also stopped working properly in some cases). I know you @pond did not run into these issues, but that doesn't mean everything worked perfectly.

Anyways, adding back the "ruby" platform as a fallback for this case is an option. We would have to warn (and recommend adding the specific platform instead) when this "generic" platform gets actually used for materializing the lockfile into actual gems. This is because by doing this we would lose the guarantee mentioned by @indirect that in frozen mode the same lockfile always deploys the exact same set of gems. But at least things would "work" again by default.

This two solutions are not mutually exclusive, and the current one is on the one hand, simpler to implement, and on the other hand, it already covers the most common cases. So I'd go with this one and see how it goes.

pond:
> I have no idea where you got the idea that the entire point of a Gemfile is to be platform independent, but for the record: it is not.

Well, OK; that's surprising. I had this idea that Ruby and its gems were meant to run on more than one operating system and architecture. The only mechanism in this ecosystem that lets us specify exact versions of gems and their dependencies is Bundler. If `Gemfile.lock` is not platform-independent, the entire bundled Ruby language and its gems will not be platform independent either. Of course there will be limits - up to whatever is reasonably possible within Bundler developer time and technology constraints. An example would be native components building against locally included or system libraries, for example, which gem authors have been free to decide on their own and which therefore Bundler has no control over; or even a Gem that's specifically bound to (say) a Windows-only x86 library and is itself, by virtue of native extensions, explicitly not cross-platform. But those are the edge cases, not the major use cases.

> Gemfiles are usually able to be applied to multiple platforms, but sometimes they can’t be. The generic ruby platform is often workable, but sometimes isn’t.

So as a suggestion or discussion point - can we have Bundler write the `ruby` platform for the majority use case where it's workable and the `Gemfile.lock` _is_ cross-platform? It would only lock to a developer's local machine - though I still struggle to see how that is ever meaningful - if one of the edge cases you mention arise? Could it not _always_ write the `ruby` entry under platforms _in addition to_ the local machine maybe?

After all, what's the lockfile for?

* If I specify the precise semver-orientated version constraints that I require in my `Gemfile`, then there would never be a reason to commit a `Gemfile.lock` to any source control system, nor would I need to send it to any deployed system; the `Gemfile` would be enough and lockfiles can go away completely.
* If the `Gemfile` version constraints are insufficient for a developer to be sure that their tested software would be deployed and running under the same versions of associated libraries (gems), then a `Gemfile.lock` _must_ be sent so that that _the developer has the same gem versions as the **deployed** software_ in which case the developer's local machine architecture and operating system is irrelevant to that developer's intent unless it just happens to be precisely identical to the deployed system.

...though of course:

* We can't ever be sure of precisely the same behaviour even if only because of operating system differences, so usually the best you can do is have a CI system that matches and run tests there; so...
* The lockfile in such cases is invaluable because it makes sure that CI and deployment use precisely the same gem versions; but then...
* The platform architecture that would need to be specified therein is _once again **clearly not** the local developer's machine architecture_ and Bundler writing such by default is very unhelpful.

pond:
(Also as a note, the engagement on this discussion by the community & maintainers is excellent & helps everyone understand what's being done and way a lot more; it'll help in future too as this thread can be pointed to; so, thanks for that!)

deivid-rodriguez:
> If I specify the precise semver-orientated version constraints that I require in my Gemfile, then there would never be a reason to commit a Gemfile.lock to any source control system, nor would I need to send it to any deployed system; the Gemfile would be enough and lockfiles can go away completely.

Yes, if you specified exact requirements for ALL of your dependencies (including all transitive dependencies), and if the Gemfile DSL was fully complete and allowed you to specific the platforms your application should work on. I guess you can imagine why nobody would do this (even if we made it possible by expanding the Gemfile DSL).

> If the Gemfile version constraints are insufficient for a developer to be sure that their tested software would be deployed and running under the same versions of associated libraries (gems), then a Gemfile.lock must be sent so that that the developer has the same gem versions as the deployed software in which case the developer's local machine architecture and operating system is irrelevant to that developer's intent unless it just happens to be precisely identical to the deployed system.

Well, not only that. The `Gemfile.lock`, specially in "frozen" mode which makes this intent very clear, is also there to ensure that two different deployments of the same lockfile result in exactly the same code being deployed. And by locking only the generic "ruby" platform, this is not possible. The way Bundler worked before, one deployment could result in the nokogiri-1.13.1 gem being installed, and a subsequent deployment installing the nokogiri-1.13.1-x86_64-linux gem instead, or viceversa.

The above is in addition to the resolution not guaranteed to be correct if specific platforms are not recorded locked in the `Gemfile.lock` file, because the same version of a gem, but for a different platform, may bring requirements incompatible with the rest of gems recorded in the `Gemfile.lock` file. I guess this last issue could be improved by adding additional version resolution work at install time, but that defeats another point of having a lockfile: avoid having to resolve gem dependencies every time.

Anyways, with this feature you should be able to specify not only `x86_64-linux`, but also `ruby` as an additional platform by default if you really want to, so that should fix things for you without needing anything else. But still the extra improvement I suggested above of also adding the "ruby" platform by default (but deprecating it by warning when it gets actually used) might not be a bad idea. Food for future discussion, I'll now focus on getting this one ready :)

indirect:
Updated name to make it slightly clearer that we will resolve and lock for linux, but we have no idea if your bundle will _work_ on linux. 😅

deivid-rodriguez:
This is ready for a review now!

ansonhoyt:
Thank you! This solves one of our regular first-deploy issues! ❤️

PR #5710, opened 2022-07-12
Ask gem scope details on gem signin

jenshenny:

From https://github.com/rubygems/rubygems.org/issues/2755, interest has been expressed regarding being able to scope an API key to a gem. This has been implemented in the UI https://github.com/rubygems/rubygems.org/pull/2944.

This PR adds this behaviour in the CLI. PR to add `rubygem_name` to the rubygems.org endpoint https://github.com/rubygems/rubygems.org/pull/3136.

Adds the following prompt
```
Enter the name of the gem you want to scope this key to (strongly recommended).
Gem scope [All gems]: rails
```
if...
1. the user enables one or more of the following scopes (push rubygem, yank rubygem, add owner, remove owner)
2. user is on default host (I'm assuming other hosts aren't compatible yet)

the input would be passed in a param `rubygem_name` to the creation endpoint.

### Alternatives
I thought about having the user select from a list of their gems but it's difficult to display correctly if the user has many gems. Furthermore, the only endpoint that returns the gems someone owns requires an API key with the `index_rubygems `scope which can't work because the command is creating an API key. So if we were going to go down this path, we would need to allow for basic auth to the endpoint.


lib/rubygems/gemcutter_utilities.rb
```diff
@@ -269,6 +271,24 @@ def get_scope_params(scope)
     scope_params
   end
 
+  def get_rubygem_scope_params(scope_params)
+    rubygem_scope_params = {}
+    return rubygem_scope_params unless gem_scope_compatible_api_scope_enabled?(scope_params)
+
+    say "Enter the name of the gem you want to scope this key to."
+    rubygem_name = ask "Gem scope [All gems]: "
```
kevinlinxc, line 279: Is this signalling that the default behaviour/just hitting enter is to scope to all gems? I don't use many CLIs so I'm wondering if this is the standard format for declaring this behaviour, or if something like [Default: all gems] would be better

simi (approved):

simi:
Looks good to me, but I would prefer to get second review before merge if possible.

jeromedalbert:
@jenshenny Are you still interested in moving this PR forward?

PR #5495, opened 2022-04-25
Raise if ruby platform is forced and there are no ruby variants

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

We have been reported a case where Bundler generates a lockfile when using `force_ruby_platform` that then makes Bundler crash when run on a different platform.

The case was reported by @Stex here: https://github.com/rubygems/rubygems/issues/5422#issuecomment-1106472389.

When using `force_ruby_platform` on a `Gemfile` like this:

```ruby
source "https://rubygems.org"

gem "sorbet", "0.5.9889"
```

the following `Gemfile.lock`file is generated

```

GEM
  remote: https://rubygems.org/
  specs:
    sorbet (0.5.9889)
      sorbet-static (= 0.5.9889)
    sorbet-static (0.5.9889-universal-darwin-14)
    sorbet-static (0.5.9889-universal-darwin-15)
    sorbet-static (0.5.9889-universal-darwin-16)
    sorbet-static (0.5.9889-universal-darwin-17)
    sorbet-static (0.5.9889-universal-darwin-18)
    sorbet-static (0.5.9889-universal-darwin-19)
    sorbet-static (0.5.9889-universal-darwin-20)
    sorbet-static (0.5.9889-universal-darwin-21)
    sorbet-static (0.5.9889-x86_64-linux)

PLATFORMS
  ruby

DEPENDENCIES
  sorbet (= 0.5.9889)

BUNDLED WITH
   2.3.12
```

And using that lockfile under a M1 docker container (which has platform "aarch64-linux") leads to a crash because no compatible gem can be found for that arch.

Closes #5422.

## What is your fix for the problem, implemented in this PR?

We should give a better error message when dealing with this `Gemfile.lock` file anyways, but in my opinion, the above lockfile should not be generated at all because the user is explicitly saying "I don't want platform specific gems" by specifying `force_ruby_platform`, and sorbet-static can't satisfy that.

I prepared a patch to implement the above and  was really happy because it seemed to work well and it led to removing this line:

https://github.com/rubygems/rubygems/blob/2faada63ae65d8559bcba9e083f3c35ba6c18fdc/bundler/lib/bundler/match_platform.rb#L18

Which was the specific line giving @lloeki issues when trying to add non gnu libc support at https://github.com/rubygems/rubygems/pull/4488, so this could also unblock that.

[...]

PR #5711, opened 2022-07-12
Fix TruffleRuby no longer able to install lockfiles generated with other implementations

deivid-rodriguez:

Since #5495, TruffleRuby is no longer able to install lockfiles generated with other implementations. Specifically when they include platform specific gems, usually not compatible with TruffleRuby.

I was no able to find a good solution that allows us to keep all the recent changes in the approach to supporting TruffleRuby by using `:force_ruby_platform`. And I see no easy way of doing so. I think the only way to make this work cleanly would be to get TruffleRuby to report its own platform, so that we can act accordingly, and can trigger a fresh re-resolve if the lockfile was never bundled before with TruffleRuby before.

Anyways, I'm not able to implement or make a proposal for the above right now, so for the moment I reverted our new approach to supporting TruffleRuby and restored what we were doing before.


bundler/spec/runtime/platform_spec.rb
```diff
@@ -281,38 +281,67 @@
...
+      gem "platform_specific"
     G
 
     lockfile <<-L
       GEM
-        remote: #{file_uri_for(gem_repo4)}/
+        remote: #{file_uri_for(gem_repo1)}/
         specs:
-          libv8 (1.0)
+          platform_specific (1.0-x86-darwin-100)
```
eregon, line 330: Do we need a simulate_platform around or it isn't necessary here?

deivid-rodriguez: This is precisely to test the case where we have a lockfile generated on a different platform. Since we never test on `x86-darwin-100` (I don't think that exists at all), I think it works as is. If we wrapped things with `simulate_platform` the spec would change its meaning I believe?

deivid-rodriguez: Well we could use `simulate_platform` just to run `bundle install` to generate the lockfile. But to make specs simpler, focus and what we want to test and don't run a lot of extra bundler process I usually favor writing whatever lockfile we're having problems with directly.

eregon: OK, makes sense

PR #5720, opened 2022-07-14
Drop support for old `Gem::Specification` versions

nobu:

`specification_version` method was added before RubyGems 1.0, and `add_runtime_dependency` method was before 1.2. 
These seem aged enough to remove.

Just remove the conditionals.


deivid-rodriguez (approved):
This makes sense to me and I don't see any obvious risk. I could imagine downgrading RubyGems to one of these ancient versions while keeping the same installed specifications could make this break, but I'm pretty sure it's not possible to downgrade that far.

PR #5750, opened 2022-07-22
Use `SystemExit#status` as `exit_code`

nobu:

None, but no reasons to manage `Gem::SystemExitException#exit_code` and `SystemExit#status` separately.

Make `exit_code` an alias of `status`.


nobu:
It was `attr_accessor`, but seems never assigned after initialized.

deivid-rodriguez (approved):
Sounds good!

PR #5759, opened 2022-07-24
Add `Bundler.settings[:only]` to install gems of the specified groups

### LINKED ISSUE #4048: Ability to install only some gem groups (a subset of everything in the Gemfile)

The `--with` flag, or some other mechanism, such as an `--only` flag, should allow installing only specified groups.

The use-case for me is that when running Github actions I only want to install a few gems and installing all of them for every CI run costs time and money.

Doing negative selection (`--without=default,development`) doesn't feel like an elegant solution, since it stops working if more groups are added.

1. Create this Gemfile

```ruby
# Gemfile
source 'https://rubygems.org'

gem 'bcrypt', '~> 3.1'
gem 'excon', '~> 0.75'

group :development do
  gem 'spring', '~> 1.4'
end

group :development, :linting do
  gem 'rubocop', :require => false
  gem 'brakeman', :require => false
end
```

2. Run `bundle install --with=linting`

I expected `bundle install --with=linting` to only install `rubocop` and `brakeman`.

### What actually happened?

All gems were installed.

## Prior Art

This PR (with quite a lot of activity and upvotes) is basically about the same thing: https://github.com/rubygems/bundler/pull/4907

It contains a bunch of arguments for and against this functionality.

r7kamura:

This resolves #4048.

For example, our company's Rails application has the following `Gemfile`:

```ruby
source 'https://rubygems.org'

# ...

group(
  :group1,
  :group2,
  :group3,
  :group4,
  :group5,
  :group6,
  :group7,
) do
  # ...
end

group(
  :group1,
  :group2,
  :group3,
  :group5
) do
  # ...
end

group :rubocop do
  gem 'erbcop', require: false
  gem 'rubocop', require: false
  gem 'rubocop-performance', require: false
  gem 'rubocop-rails', require: false
  gem 'rubocop-rspec', require: false
  gem 'slimcop', require: false
end

group :danger do
  gem 'danger'
end
```

and the CI workflow to run `rubocop` is as follows:

```yaml
# .github/workflows/rubocop.yml
on:
  pull_request:
  push:
    branches:
      - main
jobs:
  rubocop:
    runs-on: ubuntu-latest
    env:
      BUNDLE_WITHOUT: danger:group1:group2:group3:group4:group5:group6:group7
    steps:
      - uses: actions/checkout@v3
      - uses: ruby/setup-ruby@v1
        with:
          bundler-cache: true
          cache-version: rubocop
      - run: bundle exec rubocop --format progress --format github
```

In the above CI workflow to run RuboCop, we only want to install gems related to RuboCop (to avoid installing unrelated 300+ gems).

To accomplish this, we need to enumerate the names of the groups we do not need in `BUNDLER_WITHOUT`. However, this would duplicate the code that enumerates such groups when preparing several workflow files, and becomes a problem when a new gem group is added to the Gemfile.

Having multiple Gemfiles for each workflow, (such as `Gemfile`, `Gemfile-rubocop`, `Gemfile-danger`, etc.) at least eliminates the need to list groups in `BUNDLER_WITHOUT`, but it creates new problems, such as code duplication and version mismatches due to forgetting to update some of their `*.lock` files.

You may be surprised to see Gemfile with so many groups, but if you look at real-world examples of large old Rails applications, this kind of code is surprisingly common.

In this PR, I added `Bundler.settings[:only]` Array option to only install gems of the specified groups.

This option can be used as follows:

```
bundle config set only rubocop
bundle install
```

or

```
BUNDLE_ONLY=rubocop bundle install
```

For example, the workflow code I mentioned above can be replaced as follows:

```diff
 env:
-  BUNDLE_WITHOUT: danger:group1:group2:group3:group4:group5:group6:group7
+  BUNDLE_ONLY: rubocop
```

## Other information

I tried it locally and confirmed that it works as expected.

image

### DIFF
--- a/bundler/lib/bundler/man/bundle-config.1
+++ b/bundler/lib/bundler/man/bundle-config.1
@@ -74,6 +74,10 @@ Creates a directory (defaults to \fB~/bin\fR) and place any executables from the
 In deployment mode, Bundler will \'roll\-out\' the bundle for \fBproduction\fR use\. Please check carefully if you want to have this option enabled in \fBdevelopment\fR or \fBtest\fR environments\.
 .
 .TP
+\fBonly\fR
+A space\-separated list of groups to install only gems of the specified groups\.
+.
+.TP
 \fBpath\fR
 The location to install the specified gems to\. This defaults to Rubygems\' setting\. Bundler shares this location with Rubygems, \fBgem install \.\.\.\fR will have gem installed there, too\. Therefore, gems installed without a \fB\-\-path \.\.\.\fR setting will show up by calling \fBgem list\fR\. Accordingly, gems installed to other locations will not get listed\.
 .
@@ -218,6 +222,9 @@ The following is a list of all configuration keys and their purpose\. You can le
 .
 .IP "\(bu" 4
 \fBno_prune\fR (\fBBUNDLE_NO_PRUNE\fR): Whether Bundler should leave outdated gems unpruned when caching\.
+.
+.IP "\(bu" 4
+\fBonly\fR (\fBBUNDLE_ONLY\fR): A space\-separated list of groups to install only gems of the specified groups\.
 .
 .IP "\(bu" 4
 \fBpath\fR (\fBBUNDLE_PATH\fR): The location on disk where all gems in your bundle will be located regardless of \fB$GEM_HOME\fR or \fB$GEM_PATH\fR values\. Bundle gems not found in this location will be installed by \fBbundle install\fR\. Defaults to \fBGem\.dir\fR\. When \-\-deployment is used, defaults to vendor/bundle\.
--- a/bundler/spec/commands/install_spec.rb
+++ b/bundler/spec/commands/install_spec.rb
@@ -1000,6 +1000,26 @@ RSpec.describe "bundle install with gem sources" do
     end
   end
 
+  context "with only option" do
+    before do
+      bundle "config set only a:b"
+    end
+
+    it "installs only gems of the specified groups" do
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rails"
+        gem "rack", group: :a
+        gem "rake", group: :b
+        gem "yard", group: :c
+      G
+
+      expect(out).to include("Installing rack")
+      expect(out).to include("Installing rake")
+      expect(out).not_to include("Installing yard")
+    end
+  end
+
   context "with a symlinked configured as bundle path and a gem with symlinks" do
     before do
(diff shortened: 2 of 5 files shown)


deivid-rodriguez (approved):
I'm good with adding this. The maintenance overhead is small and there's many people who have requested it over the years.

Also I did not have a clear opinion on how this should interact with `BUNDLE_WITH` and `BUNDLE_WITHOUT`. I think the way you implemented it makes sense, and it's probably not needed to raise conflicts when combining these options.

I'll let this sit here for a few days to get approval from others, given this has been quite controversial over time.

deivid-rodriguez:
I really think this is fine. Merging!

sandstrom:
Awesome @r7kamura and @deivid-rodriguez! 🎉 

We have this exact problem in our app, and have used one of the aforementioned workarounds. But this is much better! 🏅

bundler/spec/commands/install_spec.rb
```diff
@@ -1000,6 +1000,26 @@ def run
...
 
+  context "with only option" do
+    before do
+      bundle "config set only a:b"
+    end
+
+    it "installs only gems of the specified groups" do
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rails"
```
morgoth, line 1011: @r7kamura would this gem be installed with `config set only a:b`, thus would gems without any group be installed?

r7kamura: `rails` gem will implicitly belong to the default group, so it will not be installed if you use `bundle config set a:b`. If you specify `bundle config set a:b:default`, the `rails` gem will also be installed.

It would have been better to explicitly include that type of expectation.

```ruby
expect(out).not_to include("Installing rails")
```

bundler/lib/bundler/man/bundle-config.1
```diff
@@ -220,6 +224,9 @@ The following is a list of all configuration keys and their purpose\. You can le
 \fBno_prune\fR (\fBBUNDLE_NO_PRUNE\fR): Whether Bundler should leave outdated gems unpruned when caching\.
 .
 .IP "\(bu" 4
+\fBonly\fR (\fBBUNDLE_ONLY\fR): A space\-separated list of groups to install only gems of the specified groups\.
```
eregon, line 227: Is it intentional this is space-separated instead of `:`-separated like BUNDLE_WITH/WITHOUT?
Also the other comments on this PR assume it would be `:`-separated so that might be better/more intuitive?

deivid-rodriguez: Maybe this is a documentation issue. Implementation and tests here seem to use `:`.

Issue #5715, opened 2022-07-13
Do not consider platform gems as candidates when `required_ruby_version` is incompatible


ianks:
### Describe the problem as clearly as you can

When selecting a spec to use when a platform spec is available, currently Bundler will choose the platform matching gem even when it is not compatible with the `required_ruby_version`.

The consequences of this means that it's impossible to publish precompiled gems that fallback to compiling from source code on newer Ruby versions. At Shopify, this makes it so we have to use a bunch of hacks to test gems against `ruby-head`.

Here's a more concrete example of the issue:

```ruby
it "can fallback to a source gem when platform gems are incompatible with current ruby version" do
  setup_multiplatform_gem_with_source_gem

  source = file_uri_for(gem_repo2)

  gemfile <<~G
    source "#{source}"

    gem "my-precompiled-gem"
  G

  # simulate lockfile which includes both a precompiled gem with:
  # - Gem the current platform (with imcompatible ruby version)
  # - A source gem with compatible ruby version
  lockfile <<-L
    GEM
      remote: #{source}/
      specs:
        my-precompiled-gem (3.0.0)
        my-precompiled-gem (3.0.0-#{Bundler.local_platform})

    PLATFORMS
      ruby
      #{Bundler.local_platform}

    DEPENDENCIES
      my-precompiled-gem

    BUNDLED WITH
        #{Bundler::VERSION}
  L

  expect { install_gemfile(gemfile) }.to_not raise_error
end

def setup_multiplatform_gem_with_source_gem
  build_repo2 do
    build_gem("my-precompiled-gem", "3.0.0")
    build_gem("my-precompiled-gem", "3.0.0") do |s| 
      s.platform = Bundler.local_platform

      # purposely unresolvable
      s.required_ruby_version = ">= 1000.0.0"
    end
  end
end
```

In the above test, Bundler should choose `my-precompiled-gem (3.0.0)`. Today, it will pick `my-precompiled-gem (3.0.0-#{Bundler.local_platform})` and simply raise an error in `ensure_specs_are_compatible!`.

cc: @flavorjones

deivid-rodriguez:
Right, yeah. I'll fix this.

flavorjones:
Some prior context on related/same issue: https://github.com/rubygems/rubygems/issues/4012

deivid-rodriguez:
Yes, I believe they are very similar issues but in that case it was a "resolution from scratch" issue, while in this case it's a "lockfile issue".

[...]

PR #5766, opened 2022-07-25
Fix `ruby` DSL requirement matching for head and prerelease rubies

deivid-rodriguez:

While working on #5715, my changes made this bug surface.

With the following `Gemfile`

```ruby
source "https://rubygems.org"

ruby ">= 3.2.0"
```

And

```
$ ruby -v
ruby 3.2.0preview1 (2022-04-03 master f801386f0c) [arm64-darwin21]
```

Currently Bundler is happy

```
$ bundle
The Gemfile specifies no dependencies
Resolving dependencies...
Bundle complete! 0 Gemfile dependencies, 1 gem now installed.
```

My fix is to pass the resolver an actual string for the current Ruby version that considers prereleases.

After this fix, the previous case gives the expected error:

```
$ bundle
Your Ruby version is 3.2.0.preview1, but your Gemfile specified >= 3.2.0
```

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -765,7 +765,7 @@ module Bundler
 
     def metadata_dependencies
       @metadata_dependencies ||= [
-        Dependency.new("Ruby\0", RubyVersion.system.gem_version),
+        Dependency.new("Ruby\0", Gem.ruby_version),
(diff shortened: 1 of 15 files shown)


deivid-rodriguez:
Merging since it unblocks the fix for #5715.

PR #5774, opened 2022-07-25
Use main as default branch for Bundler specs

deivid-rodriguez:

Main is a better terminology than master for representing the default branch of repositories.

Unify terminology in Bundler specs.


simi (approved):

simi:
I would prefer to keep it inline with real git default branch of repo, but if you prefer this one, :white_check_mark: .

Btw. "official" Ruby answer to this is here - https://bugs.ruby-lang.org/issues/17828#note-2.

> I think master without corresponding slave is still OK (e.g. master branch in git).

indirect:
Yeah, let's move the GitHub default branch at the same time. 👍🏻

simi:
@indirect there is already discussion regarding this. https://github.com/rubygems/rubygems/discussions/5145

indirect:
Sorry, my comment wasn't very clear. I'm in favor of this change, for consistency in the test suite, and to make sure we are testing cases our users will see. I'll go leave my thoughts about the repo branch in the discussion ticket.

deivid-rodriguez:
Thanks! I will move this forward if CI permits it. We can tackle later the renaming of the main branch.

PR #5775, opened 2022-07-26
Add deprecation notice of bundle console

tnir:

`bundler` CLI has warned users on running `bundle console` due to https://github.com/rubygems/bundler/pull/4741 and https://github.com/rubygems/bundler/pull/7295, but Bundler Website and Bundler man does not warn them.

As per https://github.com/rubygems/bundler/pull/4741 and https://github.com/rubygems/bundler/pull/7295, adds the deprecation notice of `bundler console` in man `bundle(1)`.

### DIFF
--- a/bundler/lib/bundler/man/bundle.1
+++ b/bundler/lib/bundler/man/bundle.1
@@ -81,7 +81,7 @@ Show the source location of a particular gem in the bundle
 Show all of the outdated gems in the current bundle
 .
 .TP
-\fBbundle console(1)\fR
+\fBbundle console(1)\fR (deprecated)
 Start an IRB session in the current bundle
 .
 .TP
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
👍

simi:
Actually I was thinking about un-deprecating this command since I'm using it often. It suggest to be replaced by `bin/console` generated by the `bundle gem`, but that is not often available and `bundle console` is the only way to open console quickly.

Usually my usecase is to quickly prototype some code starting with `Gemfile` to fetch deps and opening the console via `bundle console`.

tnir:
This is simply a follow-up PR of rubygems/bundler#7295. Please reduce bikeshedding discussions as https://github.com/rubygems/rubygems/blob/2af2520b4a7ab1c6eb1fdc3d2ef4d8c062d96ad7/bundler/UPGRADING.md?plain=1#L77-L83 repeats.

I think you can use https://github.com/rubygems/rfcs/issues.

tnir:
What is blocking this PR, which is a simple follow up of rubygems/bundler#7295, to get merged then?

deivid-rodriguez:
Nothing really, I didn't merge it because I thought we might revert it soon if we undeprecate `bundle console`, but it's unclear when someone will get to reviewing that, so I guess we can merge this for the time being.

PR #5782, opened 2022-07-28
Introduce man for aliased bundle-package(1)

### LINKED ISSUE #5781: Miss bundle-package(1)

bundle-package(1) is no more available after #7249

cf. https://github.com/rubygems/bundler-site/issues/783

Yes.

- Bundler 2.2 (bundler-2.2.33)
- or Bundler 2.3 (bundler-2.3.19)

- `open https://bundler.io/v2.3/man/bundle-package.1.html`
   - `curl https://bundler.io/v2.3/man/bundle-package.1.html`
- `open https://bundler.io/v2.2/man/bundle-package.1.html`
   - `curl https://bundler.io/v2.2/man/bundle-package.1.html`

Consistent with the results from `bundle _2.3.19_ help package`.

### What actually happened?

404.

Omitted.

tnir:

Closes #5781

Follows up https://github.com/rubygems/bundler/pull/7249

No man/Website for `bundle-package(1)` has been available since Bundler 2.2.0.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -126,6 +126,8 @@ bundler/lib/bundler/man/bundle-open.1
 bundler/lib/bundler/man/bundle-open.1.ronn
 bundler/lib/bundler/man/bundle-outdated.1
 bundler/lib/bundler/man/bundle-outdated.1.ronn
+bundler/lib/bundler/man/bundle-package.1
+bundler/lib/bundler/man/bundle-package.1.ronn
 bundler/lib/bundler/man/bundle-platform.1
 bundler/lib/bundler/man/bundle-platform.1.ronn
 bundler/lib/bundler/man/bundle-pristine.1
--- a/bundler/lib/bundler/man/bundle-package.1
+++ b/bundler/lib/bundler/man/bundle-package.1
@@ -1,0 +1,13 @@
+.\" generated with Ronn/v0.7.3
+.\" http://github.com/rtomayko/ronn/tree/0.7.3
+.
+.TH "BUNDLE\-PACKAGE" "1" "July 2022" "" ""
+.
+.SH "NAME"
+\fBbundle\-package\fR \- Renamed command\. See \fBbundle\-cache(1)\fR
+.
+.SH "SYNOPSIS"
+\fBbundle package\fR
+.
+.SH "DESCRIPTION"
+This command was renamed to \fBbundle cache\fR in Bundler 2\.1 and now is an alias of \fBbundle\-cache(1)\fR \fIbundle\-cache\.1\.html\fR\.
--- a/bundler/lib/bundler/man/bundle-package.1.ronn
+++ b/bundler/lib/bundler/man/bundle-package.1.ronn
@@ -1,0 +1,11 @@
+bundle-package(1) -- Renamed command. See `bundle-cache(1)`
+===========================================================
+
+## SYNOPSIS
+
+`bundle package`
+
+## DESCRIPTION
+
+This command was renamed to `bundle cache` in Bundler 2.1 and now is an alias
+of [`bundle-cache(1)`](bundle-cache.1.html).
--- a/bundler/lib/bundler/man/index.txt
+++ b/bundler/lib/bundler/man/index.txt
@@ -17,6 +17,7 @@ bundle-list(1)        bundle-list.1
 bundle-lock(1)        bundle-lock.1
 bundle-open(1)        bundle-open.1
 bundle-outdated(1)    bundle-outdated.1
+bundle-package(1)     bundle-package.1
 bundle-platform(1)    bundle-platform.1
 bundle-pristine(1)    bundle-pristine.1
 bundle-remove(1)      bundle-remove.1


deivid-rodriguez:
Does running `bundle package` actually warn?

When I went through this, I think my conclusion was that we DO NOT want to deprecate `bundle package`, just make it an alias to `bundle cache`, because the whole point was reducing duplicated functionality (they had almost 100% the same code).

Since `bundle_package.html` was an existing page before (documenting a first class command), I think it makes sense to keep it, but simply mention that it's a `bundle cache` alias, or maybe just redirect it to `bundle_cache.html`, not sure.

tnir:
No, `bundle package` CLI does not warn.

> Since bundle_package.html was an existing page before (documenting a first class command), I think it makes sense to keep it, but simply mention that it's a bundle cache alias, or maybe just redirect it to bundle_cache.html, not sure.

What do you mean `bundl_package.html`? `It looks like bundl_package.html` never exists to me. So `https://github.com/rubygems/bundler/blob/2-0-stable/man/bundle-package.ronn` or https://bundler.io/v2.1/man/bundle-package.1.html?

I am not sure if `package` was/is/will be deprecated.

User experience as follows looks bad to me.

- Visit https://bundler.io/v2.1/man/bundle-package.1.html
- Select v2.2 or v2.3 on sidebar.
- You will see 404.

To address this problem, I would say we should add `bundle-package.1.ronn` to this repository or `bundle-package.1.html.erb` to bundler-site. I prefer the former.

deivid-rodriguez:
Sorry, I meant https://bundler.io/v2.1/man/bundle-package.1.html.

So, you don't like redirecting to https://bundler.io/v2.3/man/bundle-cache.1.html, given that they are now aliases? We could add a section on the very top one the man page to specify command aliases, and add the alternatives to the command synopsis, or even to the title to clarify.

tnir:
In terms of user experience on Bundler Web site, I am fine just if we can implement it so that a user go back and forth between `v{2.2,2.3}/man/bundle-cache.1.html` and `v{2.1,2.0,1.17}/man/bundle-package.1.html`.

But, based on my experience in UNIX culture, a sudden removal of the existing man page(s) looks weird. I could not rationalize this kind so I have googled 100 times but I could not find a good explanation yet.

For example, `npm ci` has `clean-install` alias (and many other aliases) but there has been never `npm clean-install` page. So this looks more acceptable to me, too.

https://github.com/npm/cli/blob/58cc362ceec82ddeb16311d3fb574908294ac789/docs/content/commands/npm-ci.md#L7-L17

I am an old person.

deivid-rodriguez:
So, what would be your suggestion? Keep two duplicated pages? Add a dummy bundle-package man page like you did here, but explaining the `bundle package` is now an alias to `bundle cache`? From those two options, I'd go with the duplication if a man page must exist.

tnir:
My suggestion was the PR, but as the first step, I am still fine if simply going with #5785.

tnir:
Pushing back this to draft at the moment in favor of #5785.

deivid-rodriguez:
> My suggestion was the PR, but as the first step, I am still fine if simply going with https://github.com/rubygems/rubygems/pull/5785.

Oh, ok. I see you updated the PR to mention that the command was renamed, that makes sense too. I'm fine with either approach, or both.

bundler/lib/bundler/man/bundle-cache.1.ronn
```diff
@@ -4,6 +4,7 @@ bundle-cache(1) -- Package your needed `.gem` files into your application
 ## SYNOPSIS
 
 `bundle cache`
+`bundle package`
```
tnir, line 7: I think we need to add another break:

```suggestion
`bundle cache`

`bundle package`
```

Otherwise we will get a single line with `bundle cache bundle package` in man.

tnir (commented):
Fix it.

bundler/lib/bundler/man/bundle-cache.1.ronn
tnir: Done in 6080e96

bundler/lib/bundler/man/bundle-cache.1
```diff
@@ -9,6 +9,9 @@
 .SH "SYNOPSIS"
 \fBbundle cache\fR
 .
+.P
+\fBbundle package\fR
+.
```
tnir, line 14: This looks the best in constraint with `ronn` to me...

tnir:
@deivid-rodriguez I see. What do you think about the current changes?

deivid-rodriguez:
I'm thinking, currently `bundle package --help` directly shows `bundle cache` man page. So how do users get to see this man page?

tnir:
For example, Debian team updates their own packages from "our" man pages, and publishes them on site (but it looks like only bullseye at this moment), and Ubuntu team succeeds and publishes them for all LTS versions at https://manpages.ubuntu.com/cgi-bin/search.py?q=bundle-package.

For example, Ubuntu 20.04 LTS users upgrade their OS from 20.04 to 22.04, they expect to have one in 22.04 unless the command was deleted.

- bundle-package(1) in Ubuntu 20.04: https://manpages.ubuntu.com/manpages/focal/en/man1/bundle-package.1.html

deivid-rodriguez:
I see. I'm still not sure we should add this page just to satisfy Debian. A man page is supposed to been actually read by someone, not just to fit some consistency rules of packagers, no? How can Debian users see this page? Running `man bundle-package` or something?

How about symlinking `bundle package` man page to `bundle cache` man page? Would that work? I think if there were any Debian users actually reading this man page, it would be more user friendly because we would directly display to them the information they need, instead of forcing them to manually run a second man command to see what they want to see.

deivid-rodriguez:
I understand but to me that seems like packaging work to be done. Make sure the `bundle package` man page for new bundler versions redirect to the `bundle cache` man page. Perhaps we can hint this to them by symlinking one to the other?

deivid-rodriguez:
I think it's best to not introduce this, so I will close this. But let's keep discussing if you disagree, and we can reopen if we reach other conclusion.

tnir:
I am now okay to close this PR since now I am not sure which sub-command should be the default at all.

Issue #5435, opened 2022-03-28
Add more details to error if gem version in `Gemfile.lock` is not compatible with current ruby version


ShockwaveNN:
### Describe the problem as clearly as you can

This issue is a result of https://github.com/grosser/parallel_tests/issues/846#issuecomment-1080782798

`parallel_tests` dropped support of `ruby-2.6` in `parallel_tests-2.8.0`
But if you Gemfile.lock frozen on newer version error message for `bundler install` looks like this version is `yanked`, not like `your ruby is not supporting this version of gem`

### Did you try upgrading rubygems & bundler?

Yes, latest version 

### Post steps to reproduce the problem

1. Create `Gemfile` with content:
```
source 'https://rubygems.org'

gem "parallel_tests"
```
2. Create `Gemfile.lock` with `parallel_tests` locked on 3.8.0 (which do not support ruby-2.6):
```
GEM
  remote: https://rubygems.org/
  specs:
    parallel (1.22.1)
    parallel_tests (3.8.0)
      parallel

PLATFORMS
  x86_64-linux

DEPENDENCIES
  parallel_tests

BUNDLED WITH
   2.3.10

```
3. Create `Dockerfile`:
```
FROM ruby:2.6

RUN gem update --system
ADD Gemfile .
ADD Gemfile.lock .
RUN bundle install

```

4. run `docker build .`

### What were you expecting to happen?

Message like
```
You've trying to install `parallel_tests` on `ruby-2.6`, but `parallel_tests` has `.required_ruby_version = '>= 2.7.0'` in its sourced
```

### What actually happened?

```
Your bundle is locked to parallel_tests (3.8.0) from rubygems repository
https://rubygems.org/ or installed locally, but that version can no longer be
found in that source. That means the author of parallel_tests (3.8.0) has
removed it. You'll need to update your bundle to a version other than
parallel_tests (3.8.0) that hasn't been removed in order to install.
```

The problem with this message, that it's looked like `parallel_tests (3.8.0) ` was yanked
I've spend a hour understanding why my CI on `ruby-2.6` is failing and not failing on `ruby-3.1` and how one gem may be yanked only for some ruby version

## Environment

```
Bundler       2.3.10
  Platforms   ruby, x86_64-linux
Ruby          2.6.9p207 (2021-11-24 revision 67954) [x86_64-linux]
  Full Path   /usr/local/bin/ruby
  Config Dir  /usr/local/etc
RubyGems      3.3.10
[...]

PR #5784, opened 2022-07-28
Don't discard candidates matching Ruby metadata

### LINKED ISSUE #5715: Do not consider platform gems as candidates when `required_ruby_version` is incompatible

deivid-rodriguez:

Sometimes a `Gemfile.lock` file will include a variant that matches the current Ruby, but Bundler will discard it because of not being the most platform specific one.

Bundler materializes lockfiles into real specifications in two steps. First, it filters the dependencies it needs (for example, if `BUNDLE_WITHOUT` is set or things like that), and chooses a set of specs from those. Then it converts those lockfile specifications into real specifications.

From these two steps, only the second one considered Ruby metadata, while both steps considered platform. This resulted on these candidates already discarded by the time we try to materialize specifications.

This PR merges these two steps into a single pass, so that we consider metadata and platforms at the same time. This also speeds up `require "bundler/setup"` time in our example big Gemfile by about 2-3%.

```
✗ hyperfine 'BUNDLER_VERSION=2.4.0.dev ruby -rbundler/setup -e1' 'BUNDLER_VERSION=2.3.19 ruby -rbundler/setup -e1'  
Benchmark 1: BUNDLER_VERSION=2.4.0.dev ruby -rbundler/setup -e1
  Time (mean ± σ):     135.5 ms ±   1.0 ms    [User: 112.1 ms, System: 22.1 ms]
  Range (min … max):   134.2 ms … 138.4 ms    21 runs
 
Benchmark 2: BUNDLER_VERSION=2.3.19 ruby -rbundler/setup -e1
  Time (mean ± σ):     138.9 ms ±   0.5 ms    [User: 115.6 ms, System: 21.9 ms]
  Range (min … max):   138.1 ms … 139.9 ms    21 runs
 
Summary
  'BUNDLER_VERSION=2.4.0.dev ruby -rbundler/setup -e1' ran
    1.03 ± 0.01 times faster than 'BUNDLER_VERSION=2.3.19 ruby -rbundler/setup -e1'
```

Closes #5715.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -524,10 +524,8 @@ module Bundler
         end
       end
 
-      unless specs["bundler"].any?
-        bundler = sources.metadata_source.specs.search(Gem::Dependency.new("bundler", VERSION)).last
-        specs["bundler"] = bundler
-      end
+      bundler = sources.metadata_source.specs.search(Gem::Dependency.new("bundler", VERSION)).last
+      specs["bundler"] = bundler
 
       specs
     end
--- a/bundler/lib/bundler/gem_helpers.rb
+++ b/bundler/lib/bundler/gem_helpers.rb
@@ -44,6 +44,12 @@ module Bundler
 
     def select_best_platform_match(specs, platform)
       matching = specs.select {|spec| spec.match_platform(platform) }
+
+      sort_best_platform_match(matching, platform)
+    end
+    module_function :select_best_platform_match
+
+    def sort_best_platform_match(matching, platform)
       exact = matching.select {|spec| spec.platform == platform }
       return exact if exact.any?
 
@@ -52,7 +58,7 @@ module Bundler
 
       sorted_matching.take_while {|spec| same_specificity(platform, spec, exemplary_spec) && same_deps(spec, exemplary_spec) }
     end
-    module_function :select_best_platform_match
+    module_function :sort_best_platform_match
 
     class PlatformMatch
       def self.specificity_score(spec_platform, user_platform)
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -96,12 +96,11 @@ module Bundler
         else
           source.specs.search(self)
         end
-        best_installable_candidate = candidates.reverse.find do |spec|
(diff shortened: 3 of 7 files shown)


deivid-rodriguez:
As a side effect, this PR also implements a different improvement, namely, auto fixing lockfiles generated with different Ruby versions automatically, instead of raising errors. That, of course, unless we are in frozen mode and the lockfile can't be modified. This improvement was already planed since https://github.com/rubygems/rubygems/issues/5435#issuecomment-1080966063.

PR #5789, opened 2022-07-31
RuboCop: Add Performance/StringInclude cop

sunny:

Launching RuboCop performance cops on an app’s binstubs returns minor performance suggestions. This led me to suggest them back to rubygems.

Minor performance benefits and cleanups. Including:

- Replaces some regexp matches by `String#include?`.
- Adds the [`Performance/StringInclude`](https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancestringinclude) cop to the project.
- Ensures RuboCop looks at the binstub template files as well.
- Fixes pull-request documentation, since `bin/rubocop` does not exist _at root_.

- [x] ~Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes~


simi:
Hello @sunny. Is there any measurable difference for common commands like `gem/bundle install`?

bundler/doc/development/PULL_REQUESTS.md
```diff
@@ -10,7 +10,7 @@ Before you submit a pull request, please remember to do the following:
 
 Make sure the code formatting and styling adheres to the guidelines. We use RuboCop for this. Lack of formatting adherence will result in automatic GitHub Actions build failures.
 
-      $ bin/rubocop -a
+      $ rake rubocop
```
deivid-rodriguez, line 13: Actually there's `bin/rubocop` at `bundler/` folder in the repo, which is where you usually work when working on Bundler. Should we instead clarify that?

bundler/lib/bundler/templates/Executable.bundler
```diff
@@ -62,8 +62,9 @@ m = Module.new do
 
   def bundler_requirement
     @bundler_requirement ||=
-      env_var_version || cli_arg_version ||
-        bundler_requirement_for(lockfile_version)
+      env_var_version ||
+      cli_arg_version ||
+      bundler_requirement_for(lockfile_version)
```
deivid-rodriguez, line 67: This seems just personal preference, not RuboCop related, right?

bundler/lib/bundler/templates/Executable.standalone
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env <%= Bundler.settings[:shebang] || RbConfig::CONFIG["ruby_install_name"] %>
+# frozen_string_literal: true
```
deivid-rodriguez, line 2: This one would be a different cop, right? I guess could be proposed separately.

lib/rubygems/commands/rdoc_command.rb
```diff
@@ -86,8 +86,9 @@ def execute
       begin
         doc.generate
       rescue Errno::ENOENT => e
-        e.message =~ / - /
-        alert_error "Unable to document #{spec.full_name}, #{$'} is missing, skipping"
+        match = / - /.match(e.message)
+        alert_error "Unable to document #{spec.full_name}, " \
+                    " #{match.post_match} is missing, skipping"
```
deivid-rodriguez, line 91: This one seems a different cop too, let's remove it from this PR.

deivid-rodriguez (commented):
Left a few comments, mostly for stuff unrelated to this PR's purposes. Other than that this seems good to me. Regardless of performance improvements, this is more readable Ruby code anyways.

bundler/doc/development/PULL_REQUESTS.md
sunny: Oh, I see! Indeed, this wasn’t very clear to me.

Perhaps by prefixing a `cd …` it would be enough for newcomers to understand?

```suggestion
      $ cd bundler
      $ bin/rubocop -a
```

bundler/lib/bundler/templates/Executable.bundler
sunny: Actually, now that the templates are correctly reviewed by RuboCop there’s a `Layout/MultilineOperationIndentation` offense that pops up.

This is what the autocorrect on this file suggests:

```diff
  def bundler_requirement
     @bundler_requirement ||=
       env_var_version || cli_arg_version ||
-        bundler_requirement_for(lockfile_version)
+      bundler_requirement_for(lockfile_version)
   end
```

I thought we could balance this small loss in clarity by splitting this to an extra line. Would RuboCop’s autocorrect suggestion be more in the style of the rest of the project?

bundler/lib/bundler/templates/Executable.standalone
sunny: This also came up as an offense now that the template is correctly covered by RuboCop:

```
bundler/lib/bundler/templates/Executable.standalone:1:1: C: [Correctable] Style/FrozenStringLiteralComment: Missing frozen string literal comment.
#!/usr/bin/env <%= Bundler.settings[:shebang] || RbConfig::CONFIG["ruby_install_name"] %>
^
```

I could split this in two pull-requests, though: one that fixes the RuboCop coverage of the template files. And another one, based on the first one, that introduces the performance cop. What do you think?

lib/rubygems/commands/rdoc_command.rb
sunny: Unfortunately RuboCop thinks this is also `Performance/StringInclude` offense. It suggests calling `e.message.includes?(" - ")` since it doesn’t know about the following `$'`.

Would you rather we use `rubocop:disable Performance/StringInclude` around the method, instead?

sunny:
Hi @simi & @deivid-rodriguez! Thanks for reviewing this and for your time spent on RubyGems. Much appreciated.

> Is there any measurable difference for common commands like gem/bundle install?

I don’t think this would make any noticeable improvement except for rare cases where users would be calling these commands thousands of times. Perhaps this PR shouldn’t be labeled as "performance" but rather under style guide, good practices or clarity, instead.

bundler/doc/development/PULL_REQUESTS.md
deivid-rodriguez: Indeed, that's a good idea! 👍

bundler/lib/bundler/templates/Executable.bundler
deivid-rodriguez: Aaaah I see! I think from the available alternatives to fix that offense I'd probably go with putting everything in the same line. But your initial edit seems fine too.

bundler/lib/bundler/templates/Executable.standalone
deivid-rodriguez: I think two pull requests is a good idea yeah. Or at least separate commits.

lib/rubygems/commands/rdoc_command.rb
deivid-rodriguez: Aha! No, I think your alternative is better than what we have. I don't like using globals like this.

bundler/lib/bundler/templates/Executable.standalone
sunny: I went for two separate commits, then: one for correctly calling the cops on the binstubs (plus the minor documentation change), and one that introduces `Performance/StringInclude` and its fixes.

deivid-rodriguez (approved):

simi:
:partying_face:

PR #5811, opened 2022-08-03
Include backtrace with crashes by default

deivid-rodriguez:

As opposed to Bundler, which reports a lot of very useful information when it crashes, every time a user reports a RubyGems crash, I have to tell them to rerun the command with the `--backtrace` flag, because errors without the `--backtrace` flag are very unhelpful and lack context. And sometimes, depending on where the error happened, rerunning might not be that easy.

My fix is to start showing backtraces with errors by default.

### DIFF
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -39,7 +39,7 @@ require "rbconfig"
 class Gem::ConfigFile
   include Gem::UserInteraction
 
-  DEFAULT_BACKTRACE = false
+  DEFAULT_BACKTRACE = true
   DEFAULT_BULK_THRESHOLD = 1000
   DEFAULT_VERBOSITY = true
   DEFAULT_UPDATE_SOURCES = true
--- a/test/rubygems/test_gem_config_file.rb
+++ b/test/rubygems/test_gem_config_file.rb
@@ -35,7 +35,7 @@ class TestGemConfigFile < Gem::TestCase
   def test_initialize
     assert_equal @temp_conf, @cfg.config_file_name
 
-    assert_equal false, @cfg.backtrace
+    assert_equal true, @cfg.backtrace
     assert_equal true, @cfg.update_sources
     assert_equal Gem::ConfigFile::DEFAULT_BULK_THRESHOLD, @cfg.bulk_threshold
     assert_equal true, @cfg.verbose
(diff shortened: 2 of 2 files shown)


simi:
I'm not big fan of backtrace, since it hides the most important info when long (since it is usually at the top and it can be scrolled over in console). Personally I would prefer to store it to some temp file and explain to copy content of the file into the issue template.

deivid-rodriguez:
Note that this is for unhandled errors, handled errors should print something useful to the user without backtraces.

But when RubyGems crashes, you can't even see where the error happened, errors are so confusing. For years I had to repeate the same pattern of telling reporters to rerun commands with `--backtrace` because otherwise I'm in the dark.

Also, we print backtraces in Bundler and it works great.

Finally, RubyGems is required first thing by Ruby, so backtraces of RubyGems errors are usually small.

I think storing the backtrace in temporary file is overkill, harder for us to implement, and for users to get reports ready.

If you are still not convinced, I'll just revert this, but honestly I strongly think this is a good move.

deivid-rodriguez:
This is a quick search of errors in closed issues:

```
$ gem install --user-install rdoc-data
Fetching: rdoc-data-4.1.0.gem (100%)
ERROR: While executing gem ... (Errno::EACCES)
Permission denied @ rb_sysopen - C:/Program Files/Ruby23/lib/ruby/gems/2.3.0/cache/rdoc-data-4.1.0.gem
```

```
$ gem.cmd uninstall strscan -v '3.0.1'                                                                                                                                                           
ERROR:  While executing gem ... (Errno::ENOTEMPTY)                                                                                                                                                                   
Directory not empty @ dir_s_rmdir - C:/Users/Alexandr.Evstigneev/.gem/ruby/3.0.0/gems/strscan-3.0.1  
```

```
$ sudo gem update --system
ERROR:  While executing gem ... (Errno::ENOENT)
    No such file or directory @ dir_chdir - /usr/lib/ruby/gems/3.0.0/gems/rubygems-update-3.3.13
```

They're just not helpful like that!

simi:
> If you are still not convinced, I'll just revert this, but honestly I strongly think this is a good move.

I have no strong opinion actually, let's try this one. :muscle:

deivid-rodriguez:
Ok, great, thanks @simi :)

PR #5817, opened 2022-08-05
Enable `Style/AndOr` cop in RubyGems

deivid-rodriguez:

In #4488 I need to backport the whole `Gem::Platform#===` method to Bundler, so that platform fixes work across more RubyGems versions. However, code style in RubyGems and Bundler differ a lot because RubyGems uses "English" logical operators. This makes it a bit harder to keep the duplicated code in sync.

In addition to that, I always found it very weird how RubyGems uses these operators rather than the symbolic ones for plain logical operations.

Enable `Style/AndOr` in RubyGems.

In theory the auto-correction of this cop is unsafe due to the different precedence, but this cop seems to have been battle tested a lot and the corrections are really smart. I will give it a manual review if this is accepted though (and if CI passes, of course).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -160,6 +160,9 @@ Style/MutableConstant:
 Style/NilComparison:
   Enabled: true
 
+Style/Not:
+  Enabled: true
+
 Style/BlockDelimiters:
   Enabled: true
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -128,6 +128,10 @@ Performance/StringReplacement:
 Security/Open:
   Enabled: true
 
+Style/AndOr:
+  Enabled: true
+  EnforcedStyle: always
+
 Style/Dir:
   Enabled: true
 
--- a/.rubocop_bundler.yml
+++ b/.rubocop_bundler.yml
@@ -14,6 +14,7 @@ AllCops:
   Exclude:
     - bin/*
     - pkg/**/*
+    - tmp/**/*
     - util/**/*
     - bundler/tmp/**/*
     - bundler/lib/bundler/vendor/**/*
--- a/bin/update_rubygems
+++ b/bin/update_rubygems
@@ -7,7 +7,7 @@
 
 require "rubygems"
 
-if ARGV.include? "-h" or ARGV.include? "--help"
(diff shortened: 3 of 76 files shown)


simi:
Is there a plan to add it to https://github.com/rubygems/rubygems/blob/master/.git-blame-ignore-revs after merging?

simi (approved):

deivid-rodriguez:
> Is there a plan to add it to https://github.com/rubygems/rubygems/blob/master/.git-blame-ignore-revs after merging?

🤷‍♂️, I think we should set up a rule for this, otherwise we'll be discussing this too many times. I think a criteria for adding it would be:
* Style change
* Net change of the commit (additions + deletions) to be at least 1000.

simi:
IMHO any "cosmetic" bulk change should be added to `.git-blame-ignore-revs`.

deivid-rodriguez:
We just need to define "bulk" 😄

deivid-rodriguez:
In my opinion, cosmetic changes in git history are fine, the problem is when they appear every single time you use `git blame` (and I'm not even bothered personally by that that much either, I just proposed `.git-blame-ignore-revs` to make a patch like the one to change quote styles more likely to be accepted).

deivid-rodriguez:
Also, I think RuboCop cops with unsafe auto-corrections should never qualify as cosmetic. Imagine what it would be like to figure it out if this patch introduced a bug, and the commit was added to `.git-blame-ignore-revs` 😅.

indirect (approved):

simi (approved):

simi:
Indeed this is not cosmetic change.

deivid-rodriguez:
I reviewed the patch manually and made two modifications:

* The autocorrected code started mixing `not` with symbolic operators which felt weird and needed extra parenthesis to work. So I enabled the [`Style/Not` cop](https://docs.rubocop.org/rubocop/cops_style.html#stylenot) too.
* The autocorrected code was introducing many unnecessary parentheses, given that `==`, `===`, `>=`, and so on still have higher precedence than `&&` and `||`. So I removed them unnecessary parentheses, keeping the code closer to what it was.

This should be ready now from my side!

hsbt:
👍

PR #5256, opened 2022-01-04
Fix metadata requirements being bypassed when custom gem servers are used

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Sometimes, when using custom gem servers, some gems may end up being installed that are not compatible with the running ruby or rubygems version.

## What is your fix for the problem, implemented in this PR?

My fix to download the full gemspec from the marshaled index when we need to check metadata requirements for gem sources not implementing the compact index API, so that resolution is still correct.

Fixes #5251.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #5826, opened 2022-08-09
Restore previous performance of private RubyGems servers

### LINKED ISSUE #5385: Bundler fetches gemspec for every version of every private gem when using github as a source as of 2.3.5

Bundler fetches gemspec for every valid version of gem that uses github has a source as of bundler 2.3.5. Version 2.3.4 does not exhibit this problem.

Yes

Ruby 3.1.1
Bundler 2.3.8

```rb
source "https://rubygems.org"
git_source(:github) { |repo| "https://github.com/#{repo}.git" }

source "https://rubygems.pkg.github.com/my-org" do
  gem "some-gem", require: false
end
```

```sh
bundle config set --local path ./gems
bundle install
```

Gems are installed relatively quickly

### What actually happened?

Gems take an incredibly long time to install and they show gemspec fetches for every version of every private gem when running in verbose. If I specify a version constraint, it limits the gemspecs fetched, but we don't specify constraints typically so this is a non-starter for us.

See redacted log:

[log.txt](https://github.com/rubygems/rubygems/files/8197368/log.txt)

## Environment

```
Bundler       2.3.8
Ruby          3.1.1p18 (2022-02-18 revision 53f5fc4236a754ddf94b20dbb70ab63bd5109b18) [arm64-darwin-21]
  Full Path   /Users/username/.asdf/installs/ruby/3.1.1/bin/ruby
  Config Dir  /Users/username/.asdf/installs/ruby/3.1.1/etc
RubyGems      3.3.8
  Gem Home    /Users/username/.asdf/installs/ruby/3.1.1/lib/ruby/gems/3.1.0
  Gem Path    /Users/username/.gem/ruby/3.1.0:/Users/username/.asdf/installs/ruby/3.1.1/lib/ruby/gems/3.1.0
  User Home   /Users/username
  User Path   /Users/username/.gem/ruby/3.1.0
  Bin Dir     /Users/username/.asdf/installs/ruby/3.1.1/bin
Tools         
  Git         2.35.1
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2022-02-23
Git SHA           4e812b9ca5
Released Version  true
```

## Bundler settings

```
disable_local_branch_check
  Set for the current user (/Users/username/.bundle/config): true
gem.coc
  Set for the current user (/Users/username/.bundle/config): false
gem.mit
  Set for the current user (/Users/username/.bundle/config): false
gem.test
  Set for the current user (/Users/username/.bundle/config): false
https://rubygems.pkg.github.com/my-org/
  Set for the current user (/Users/username/.bundle/config): "username:[REDACTED]"
jobs
  Set for the current user (/Users/username/.bundle/config): 8
path
  Set for your local app (/Users/username/Source/project/api/.bundle/config): "./gems"
rubygems.pkg.github.com
  Set for the current user (/Users/username/.bundle/config): "username:[REDACTED]"
```

### Gemfile

```ruby
source "https://rubygems.org"
git_source(:github) { |repo| "https://github.com/#{repo}.git" }

source "https://rubygems.pkg.github.com/my-org" do
  gem "some-gem", require: false
end
```

### Gemfile.lock

```
<No /Users/username/Source/project/api/Gemfile.lock found>
```

deivid-rodriguez:

Since #5256, resolution for private gem servers not implementing the compact index API need to fetch marshaled gemspecs for ALL gems considered by resolution, in order to fetch their metadata requirements. This is painfully slow.

My fix is to restore the old optimistic approach of resolving without metadata, and then only if the result is not valid, leverage the work done in #5840 to retry resolution excluding the specs with conflicting metadata requirements.

I added a spec to make sure that we're no longer downloading all gemspecs and also tested this in real life on a private Github server.

Closes #5385.

~I leave this as WIP for now until I merge #5840, since this one builds on that.~ #5840 merged.


vosechu:
@deivid-rodriguez Thank you so much! This fixed some problems we were tearing our hair out about! Really appreciate all your hard work!!

deivid-rodriguez:
No problem, I realized it was a painful thing and there are two many private servers on old APIs. Glad it got fixed!

PR #5852, opened 2022-08-21
Make `gem install` support non gnu libc linux

deivid-rodriguez:

`gem install` may end up installing musl-specific variants on non musl platforms.

Make platform comparison and initialization support must and other non gnu lib linux.

NOTE: This is extracted from #4488, but focusing on the `gem install` issues. `bundle install` issues will be handled in #4488 by backporting these changes to Bundler.

Fixes https://github.com/rubygems/rubygems/issues/3174.

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -151,10 +151,17 @@ class Gem::Platform
   ##
   # Does +other+ match this platform?  Two platforms match if they have the
   # same CPU, or either has a CPU of 'universal', they have the same OS, and
-  # they have the same version, or either has no version.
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Let's do this!

lloeki:
Thank you David!

ntkme:
A tip for those who are publishing linux-gnu or linux-musl specific gems: make sure you add `spec.required_rubygems_version = '>= 3.3.21'` to your gem spec to prevent it from being used by lower version of rubygems.

rubygems < 3.3.21 have the following problems:not only allow incorrect platform installation (e.g. allow linux-gnu )

PR #5853, opened 2022-08-21
Introduce bundle-plugin(1) man

tnir:

While `bundle plugin` was introduced before Bundler 2.3 and partial document for it has been available on https://bundler.io, now man for `bundle-plugin(1)` is included in Bundler gem.

Introduces bundle-plugin(1) man. While there was https://bundler.io/v2.3/bundle_plugin.html, the new man is fully rewritten.

Replaces https://github.com/rubygems/bundler-site/pull/860

### DIFF
--- a/bundler/lib/bundler/man/bundle-plugin.1.ronn
+++ b/bundler/lib/bundler/man/bundle-plugin.1.ronn
@@ -1,0 +1,62 @@
+bundle-plugin(1) -- Manage Bundler plugins
+==========================================
+
+## SYNOPSIS
+
+`bundle plugin` install PLUGINS [--source=<SOURCE>] [--version=<version>]
+                              [--git|--local_git=<git-url>] [--branch=<branch>] [--ref=<rev>]
+
+`bundle plugin` uninstall PLUGINS
+
+`bundle plugin` list
+
+`bundle plugin` help [COMMAND]
+
+## DESCRIPTION
+
+You can install, uninstall, and list plugin(s) with this command to extend functionalities of Bundler.
+
+## SUB-COMMANDS
+
+### install
+
+Install the given plugin(s).
+
+* `bundle plugin install bundler-graph`:
+  Install bundler-graph gem from RubyGems.org. The global source, specified in source in Gemfile is ignored.
+
+* `bundle plugin install bundler-graph --source https://example.com`:
+  Install bundler-graph gem from example.com. The global source, specified in source in Gemfile is not considered.
+
+* `bundle plugin install bundler-graph --version 0.2.1`:
+  You can specify the version of the gem via `--version`.
+
+* `bundle plugin install bundler-graph --git https://github.com/rubygems/bundler-graph`:
+  Install bundler-graph gem from Git repository. `--git` can be replaced with `--local-git`. You cannot use both `--git` and `--local-git`. You can use standard Git URLs like:
+
+  * `ssh://[user@]host.xz[:port]/path/to/repo.git`
+  * `http[s]://host.xz[:port]/path/to/repo.git`
+  * `/path/to/repo`
+  * `file:///path/to/repo`
+
+  When you specify `--git`/`--local-git`, you can use `--branch` or `--ref` to specify any branch, tag, or commit hash (revision) to use. When you specify both, only the latter is used.
+
+### uninstall
+
+Uninstall the plugin(s) specified in PLUGINS.
+
+### list
+
+List the installed plugins and available commands.
+
+No options.
+
+### help
+
+Describe subcommands or one specific subcommand.
+
+No options.
+
+## SEE ALSO
+
+* [How to write a Bundler plugin](https://bundler.io/guides/bundler_plugins.html)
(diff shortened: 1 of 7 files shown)


bundler/spec/plugins/install_spec.rb
```diff
@@ -32,7 +32,7 @@
   it "shows help when --help flag is given" do
     bundle "plugin install --help"
 
-    expect(out).to include("bundle plugin install PLUGINS    # Install the plugin from the source")
+    expect(out).to include("bundle plugin install PLUGINS")
```
deivid-rodriguez, line 35: @tnir I think this failure is expected, because previously, since we had no man page, we defaulted to [Thor](https://github.com/rails/thor)'s autogenerated help text. We should be good updating the text to what the manual page prints now.

deivid-rodriguez: Also if you're having trouble with the output, you might want to use the `with_fake_man` helper, which stubs the `man` executable to just print the argument given to it.

tnir: This fails as `MANWIDTH` is `80` or so in the CI while that is 300 in my local. `bundle  plugin  install  PLUGINS` is included both Thor-based and the man introduced in this PR.

```
bundle  plugin  install  PLUGINS  [--source=SOURCE] [--version=version]
       [--git|--local_git=git-url] [--branch=branch] [--ref=rev]
```

As I have no idea on testing multiple spaces between words added by justification easily, I chose different parts from the updated help messages in the latest commit.

deivid-rodriguez: That sounds fine. The other time I checked for this man output I checked the man header: https://github.com/rubygems/rubygems/blob/6b3a5a9ab0453463381a8164efb6298ea9eb776f/.github/workflows/install-rubygems.yml#L77

bundler/lib/bundler/man/bundle-plugin.1.ronn
```diff
@@ -0,0 +1,61 @@
...
+`bundle plugin` list
+
+`bundle plugin` help [COMMAND]
+
+## DESCRIPTION
+
+You can install, uninstall, and list plugin(s) with this command to extend functionalities of Bundler.
+
+## SUB-COMMANDS
+### install
```
deivid-rodriguez, line 20: I know this file is not meant to be read directly, but I think it's customary to add a blank line between headers?

bundler/lib/bundler/man/bundle-plugin.1.ronn
```diff
@@ -0,0 +1,61 @@
...
+
+Install the plugin(s) from the source.
```
deivid-rodriguez, line 22: Maybe just "Install the given plugin(s)"? When reading the examples below, I thought the one showing `--version` felt off because the header didn't mention it.

deivid-rodriguez (commented):
I only had two very minor comments, but looks good anyways!

bundler/lib/bundler/man/bundle-plugin.1.ronn
tnir: Makes sense to me.

bundler/lib/bundler/man/bundle-plugin.1.ronn
tnir: Suggestion for the sentence was applied. Per `--version`, the first line in the header includes it, doesn't it?
```
$ MANWIDTH=48 MANPAGER=head exe/bundle plugin install --help
BUNDLE-PLUGIN(1)              BUNDLE-PLUGIN(1)

NAME
       bundle-plugin - Manage Bundler plugins

SYNOPSIS
       bundle     plugin    install    PLUGINS
       [--source=SOURCE]   [--version=version]
       [--git|--local_git=git-url]
       [--branch=branch] [--ref=rev]
```

tnir (commented):
@deivid-rodriguez Thanks. Two suggestions were applied. The last one looks misunderstanding to me. Back to you.

bundler/lib/bundler/man/bundle-plugin.1.ronn
deivid-rodriguez: Yes, in the synopsis, I meant in the subcommand description.

Since it only talked about source ("Install the plugin(s) from the source."), but didn't mention "version", and then described different commands using "source related options", I was not sure if there was going to be a different section describing other options. That's why I suggested to make the description more generic.

bundler/lib/bundler/man/bundle-plugin.1.ronn
deivid-rodriguez: It doesn't seem applied actually, I meant a blank line between `## SUB-COMMANDS` and `### install`. But this is too minor, let's ignore it, if we want to have this kind of consistency we should use something like https://github.com/markdownlint/markdownlint. Disregard this suggestion.

bundler/lib/bundler/man/bundle-plugin.1.ronn
deivid-rodriguez: Actually it doesn't seem applied?

tnir: Sorry, cb899acec274c48e5ad66f655d0963eb0a1d8778 was not actually pushed. Just did it.

deivid-rodriguez (approved):
Thank you!

PR #5854, opened 2022-08-21
Improve `bundle config` command synopsis

### LINKED ISSUE #4924: Bundle config docs don't mention that you can set without using set

The `bundle config` docs [here](https://bundler.io/man/bundle-config.1.html) suggest that the options to bundle config are 

```
bundle config [list|get|set|unset] [name [value]]
```

They fail to document the fact that you can run `bundle config name value` and have it perform the same behavior as `bundle config set name value`.

It is unclear if this is a bug in the behavior of `config` or a bug in documentation. I'm assuming documentation. 

### Reproduction 

1. run `bundle config foo bar` 
2. run `bundle config get foo` 
3. confirm that "bar" is returned 
4. note that this is an unexpected result, as this is undocumented behavior

### `bundle env` output: 

I think the only relevant bit here is that I'm using bundler 2.2.27

### commentary: 
Personally, I'd prefer it if config returned an error if you tried to use it without providing one of the supported verbs. Ambiguous behavior is rarely a good thing.

deivid-rodriguez:

It's not clear from the docs that `bundle config NAME` gets the configured value for `NAME` and that `bundle config NAME VALUE` sets the `NAME` config to `VALUE`.

Make it clear by expanding the command synopsis.

Fixes #4924.

### DIFF
--- a/bundler/lib/bundler/man/bundle-config.1.ronn
+++ b/bundler/lib/bundler/man/bundle-config.1.ronn
@@ -3,7 +3,13 @@ bundle-config(1) -- Set bundler configuration options
 
 ## SYNOPSIS
 
-`bundle config` [list|get|set|unset] [<name> [<value>]]
+`bundle config` list
+
+`bundle config`[ get] NAME
+
+`bundle config`[ set] NAME VALUE
+
+`bundle config` unset NAME
 
 ## DESCRIPTION
 
--- a/bundler/lib/bundler/man/bundle-add.1
+++ b/bundler/lib/bundler/man/bundle-add.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-ADD" "1" "July 2022" "" ""
+.TH "BUNDLE\-ADD" "1" "August 2022" "" ""
 .
 .SH "NAME"
 \fBbundle\-add\fR \- Add gem to the Gemfile and run bundle install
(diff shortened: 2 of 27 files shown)


bundler/lib/bundler/man/bundle-config.1.ronn
```diff
@@ -3,7 +3,13 @@ bundle-config(1) -- Set bundler configuration options
 
 ## SYNOPSIS
 
-`bundle config` [list|get|set|unset] [<name> [<value>]]
+`bundle config` list
+
+`bundle config`[ get] NAME
```
tnir, line 8:
```
`bundle config` [get] NAME
```
How about this?

bundler/lib/bundler/man/bundle-config.1.ronn
```diff
@@ -3,7 +3,13 @@ bundle-config(1) -- Set bundler configuration options
...
+
+`bundle config`[ set] NAME VALUE
```
tnir, line 10:
```
`bundle config` [set] NAME VALUE
```

tnir (commented):
@deivid-rodriguez Some comments while this PR was already merged.

bundler/lib/bundler/man/bundle-config.1.ronn
tnir: PR in #5856

deivid-rodriguez: It was intentional, since if you omit the optionally part (`[ get]`), you're left with just a single space (`bundle config NAME`). I think I saw this style in man pages for other tools, and I like it, but if you prefer the other I don't really mind too much.

deivid-rodriguez: I'm sure I got this from some man page, but the most common tools don't use it, so let's kill it.

PR #5875, opened 2022-08-25
Fix resolution hanging on musl platforms

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

After recent musl support was added, Bundler started hanging in musl platforms. I identified the issue where valid candidates were being filtered out because their platform was specified as a string, and thus `Gem::Platform.match_spec?` which under the hood ends up calling `Gem::Platform#===` would return `nil`, because it does not support comparing platforms to strings.

## What is your fix for the problem, implemented in this PR?

I think we may want to add support for passing strings here, but the more focused fix was to make sure that Bundler always passes proper `Gem::Platform` instances to this method.

Fixes #5871.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #5881, opened 2022-08-27
Improve lockfile dependency unlocking

### LINKED ISSUE #3288: Setting a gem to a new version range shouldn't fail when setting a specific version in that range passes
Depending on your `Gemfile.lock`, when changing `rails` from `~> 5.2` to `~> 6.0`, you may see something similar to the following when running `bundle update rails`:

```
Bundler could not find compatible versions for gem "rails":
  In Gemfile:
    rails (~> 6.0)
    data_migrate was resolved to 6.1.0, which depends on
      rails (>= 5.0)
```

This message is very confusing, since it doesn't seem like there's a conflict there...

However, this goes away if you change `~> 6.0` to `6.0.1`. So, for some reason using a specific version instead of a range is resolving differently. This seems like maybe a UX problem that should be addressed. 

So really the "bug" here is that if you set rails `~> 6.0`, it's only updating the gem you requested (and direct dependencies of that gem which would change), but it doesn't update other things which depend on direct dependencies of that gem. But if you set `rails` to `6.0.1`, it will update those gems even if you didn't ask it to.

Script to reproduce:
https://gist.github.com/iwz/247640c88af7e5bcd962a51654d904f9

### LINKED ISSUE #5559: Update rails with --conservative

On our team we instructed people to use `--conservative` when upgrading gem to avoid unexpected upgrades. When upgrading rails however, as it is a bundling gem installing other gems, the fail to get the expected result with `bundle update --conservative rails`.

The proper command should be `bundle update --conservative rails activesupport actionpack actionview activemodel activerecord actionmailer activejob actioncable activestorage actionmailbox actiontext railties`

I would expect `bundle update --conservative rails` to upgrade also dependent gems.

### What actually happened?

I get the following error:

```
bundle update --conservative rails
Resolving dependencies................
Bundler could not find compatible versions for gem "actionmailer":
  In snapshot (Gemfile.lock):
    actionmailer (= 6.1.4.7)

  In Gemfile:
    devise_invitable was resolved to 2.0.2, which depends on
      actionmailer (>= 5.0)

    rails (~> 7) was resolved to 7.0.0, which depends on
      actionmailer (= 7.0.0)

```

Whereas this one works:

```
bundle update --conservative rails activesupport actionpack actionview activemodel activerecord actionmailer activejob actioncable activestorage actionmailbox actiontext railties
Resolving dependencies........................................
[SKIPED]
Bundle updated!
````

I don't know how we can achieve this though, should we add a `bundle` attribute to gems? or check gems with no ruby files in the `files`? or just make a hard coded rule for rails (and other gems with this pattern)?

### PR DESCRIPTION
**Pull request by deivid-rodriguez**

Currently Bundler tries to avoid downgrading locked gems by setting lower bound requirements on them when you run `bundle update`. This works fine in general, but sometimes Gemfile changes can introduce new requirements that conflict with what’s already locked in the lockfile. These situations end up with resolution conflicts, and with the recommendation of removing the Gemfile.lock file and run `bundle install`.

Best explained with an example adapted from our own test suite.

Let’s say you have the following Gemfile and Gemfile.lock file

```ruby
source “https://rubygems.org”

gem "jekyll-feed", "~> 0.12"
```

```
GEM
  remote: https://rubygems.org
  specs:
    jekyll-feed (0.16.0)

PLATFORMS
  arm-darwin-21

DEPENDENCIES
  jekyll-feed

BUNDLED WITH
   2.3.21
```

And then let’s so you add `gem "github-pages", "~> 226”` to it, and run `bundle update`

The result is this:

```
Bundler could not find compatible versions for gem "jekyll-feed":
  In snapshot (Gemfile.lock):
    jekyll-feed (>= 0.16.0)

  In Gemfile:
    jekyll-feed (~> 0.12)

    github-pages (~> 226) was resolved to 226, which depends on
      jekyll-feed (= 0.15.1)

Deleting your Gemfile.lock file and running `bundle install` will rebuild your snapshot from scratch, using only the gems in your Gemfile, which may resolve the conflict.
```

Something similar happens when you run `bundle install` after changing the lockfile. But in that case, Bundler does not use lower bound requirements from the lockfile, but exact dependencies. And similarly, conflicts like the above are raised.

Another similar issue was reported at #3288.

Instead, Bundler should be able to detect that it needs to to let `jekyll-feed` be downgraded to 0.15.1, just like it would happen if you delete the lockfile and resolve again.

This is what this PR implements. Bundler will try to be conservative for starters, but if it finds conflicts, it will try unlocking locked gems until those conflicts are fixed, or all locked gems are unlocked.

Fixes #3288.
Fixes #5559.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -161,6 +161,7 @@ bundler/lib/bundler/plugin/source_list.rb
 bundler/lib/bundler/process_lock.rb
 bundler/lib/bundler/remote_specification.rb
 bundler/lib/bundler/resolver.rb
+bundler/lib/bundler/resolver/base.rb
 bundler/lib/bundler/resolver/spec_group.rb
 bundler/lib/bundler/retry.rb
 bundler/lib/bundler/ruby_dsl.rb
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -485,7 +485,7 @@ module Bundler
       @resolver ||= begin
         last_resolve = converge_locked_specs
         remove_ruby_from_platforms_if_necessary!(dependencies)
-        Resolver.new(source_requirements, last_resolve, gem_version_promoter, additional_base_requirements_for_resolve, platforms)
+        Resolver.new(source_requirements, last_resolve, gem_version_promoter, additional_base_requirements_for_resolve(last_resolve), platforms)
       end
     end
 
@@ -878,9 +878,9 @@ module Bundler
       end
     end
 
-    def additional_base_requirements_for_resolve
+    def additional_base_requirements_for_resolve(last_resolve)
       return [] unless @locked_gems && unlocking? && !sources.expired_sources?(@locked_gems.sources)
-      converge_specs(@originally_locked_specs).map do |locked_spec|
+      converge_specs(@originally_locked_specs - last_resolve).map do |locked_spec|
         Dependency.new(locked_spec.name, ">= #{locked_spec.version}")
       end.uniq
     end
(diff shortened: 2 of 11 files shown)


deivid-rodriguez:
As a bonus, this PR also unblocks #5875.

The one liner PR there only breaks a single spec, which is testing specifically for this feature (I have no idea how it was passing before 😆).

PR #5883, opened 2022-08-29
Deprecate old :mswin, etc. platforms and replace with :windows

johnnyshields:
In Gemfiles, the following platforms should simply be replaced with `:windows` as a catch-all:
- `:mswin`
- `:mswin64`
- `:mingw32`
- `:x64_mingw`

By way of analogy, Gemfiles don't allow selecting different Linux / Darwin architectures, so there's no reason to allow this for Windows either. Using `:windows` will automatically select the appropriate gem architecture when running `bundle install`.

This PR includes both an update to manpage and a deprecation warning.

This will require users to change their Gemfile to use `:windows` when upgrading to Bundler 3. If Bundler 3 we will be released soon, we may want to leave these platform options until Bundler 4 to give users enough time to migrate. Another option would be to make the legacy options synonymous with `:windows` in Bundler 3.


johnnyshields:
@deivid-rodriguez

deivid-rodriguez (commented):
Thanks @johnnyshields! This looks good to me in principle, I guess fine grained Windows platforms are not useful for anybody, and `mswin` has not actually be fine grained for a long time and nobody complained anyways.

This PR implements two different things:

* Document the new `:windows` platform options and add support for per-version variants like `:windows26 `. This seems like a no-brainer, just a follow up to #5650 and we should include it with the next release.

* Deprecate old fine grained Windows platforms. I'm generally positive to this but I'd like it to have its own PR and change log entry. Could you split this out to a separate PR? I can do it too if you don't feel like doing it, no worries.

johnnyshields:
@deivid-rodriguez sure, here's a PR with the manpage only: https://github.com/rubygems/rubygems/pull/5890

I wasn't 100% sure with what you want so feel free to adjust that PR to your liking.

johnnyshields:
Ah wait you wanted the manpage + the windows versions together. Hang on I'll do that...

deivid-rodriguez:
I just want deprecations separated. The other two (docs + windows version) I don't mind whether they go together or separate.

johnnyshields:
OK https://github.com/rubygems/rubygems/pull/5890 ready for review

johnnyshields:
@deivid-rodriguez this is ready for review.

deivid-rodriguez:
I think this is going to be a bit noisy, because default Rails generators use these deprecated values. Should we create a PR upstream first to generate brand new app's Gemfiles using `:windows` if available?

johnnyshields:
> I think this is going to be a bit noisy, because default Rails generators use these deprecated values.

This will only warn once per process, and of course its very easy to fix in one's Gemfile, so I think it's probably not too noisy.

> Should we create a PR upstream first to generate brand new app's Gemfiles using :windows if available?

I assume "upstream" is referring to Rails? I think we can do independently of this PR. The deprecation warning will help nudge that forward :)

johnnyshields:
@deivid-rodriguez can you also check the CHANGELOG for when `:windows` platform alias was released? I think you may have forgotten to add it.

johnnyshields:
Rails PR here: https://github.com/rails/rails/pull/46009. I think this is a bit of a chicken-and-egg situation tho, probably this PR should be merged first.

simi:
:information_source: We can also make it soft-deprecated (no warning printed) for some time or at Rails side check which version of bundler is used during `rails g`. We had already many annoyed users with various deprecations printed even once per runtime in the history :(

simi:
@deivid-rodriguez what about versioning? Should we bump minor version with this change? Deprecations on patch bumps are super annoying in my :eyes:.

deivid-rodriguez:
> @deivid-rodriguez what about versioning? Should we bump minor version with this change? Deprecations on patch bumps are super annoying in my 👀.

We can wait until 2.4.0 to ship this, and so give us some time to update Rails generators, how about that?

> ℹ️ We can also make it soft-deprecated (no warning printed) for some time or at Rails side check which version of bundler is used during rails g. We had already many annoyed users with various deprecations printed even once per runtime in the history :(

I think ideally Rails would check Bundler version and only generate `:windows` if the running Bundler supports it.

> @deivid-rodriguez can you also check the CHANGELOG for when :windows platform alias was released? I think you may have forgotten to add it.

It was released as Bundler 2.3.22, it's in the changelog https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md (although I'm not sure why GitHub is no longer displaying it 🤔).

johnnyshields:
@deivid-rodriguez 
> We can wait until 2.4.0 to ship this, and so give us some time to update Rails generators, how about that?
Sounds good.

@simi
> We can also make it soft-deprecated (no warning printed) for some time or at Rails side check which version of bundler is used during rails g.
Ah, fair point that this will affect users on the existing released versions of Rails. Since it will only be shown for Windows users (`rails g` won't include the lines for non-Windows users), and there is also a patch for the new version of Rails, I think the log error is probably tolerable.

johnnyshields:
@deivid-rodriguez now that Ruby 3.2 is released, I think this is good to merge.

johnnyshields:
@deivid-rodriguez please merge.

deivid-rodriguez:
Sorry for dropping the ball here. So what's the plan? Merging now and releasing the deprecation when?

deivid-rodriguez:
I've been shipping small deprecations with patch releases but this one does not seem small, so I guess we can wait till the end of the year, if that's ok.

johnnyshields:
@deivid-rodriguez I think there's no problem to release this immediately. The deprecations are simply warnings, they don't remove any functionality. The actual functionality can be removed in the next major upgrade in 1-2 years.

hsbt:
wait. `mswin` and `mingw*` is completely different platform. I know RubyInstaller is majority on Windows, But It's only `mingw*` platform, not `mswin`. 

Is there strong reason they are deprecated?

deivid-rodriguez:
> @deivid-rodriguez I think there's no problem to release this immediately. The deprecations are simply warnings, they don't remove any functionality. The actual functionality can be removed in the next major upgrade in 1-2 years.

Yeah, I understand. They still introduce the warnings, which are annoying. And they will affect a lot of Gemfiles (every Gemfile generated with Rails for years). If we're not in a hurry, I think we can introduce the deprecation in the next minor or major release.

> Is there strong reason they are deprecated?

I think there's no strong reason, but I believe nobody uses this differentiation. I've never seen a Gemfile scoped only to one of these platforms but not the others.

johnnyshields:
@hsbt this PR doesnt deprecate mswin, mingw etc as platforms in the Gemfile.lock, it only removes them for the purpose of gem groups in the Gemfile. As @deivid-rodriguez notes, no one is using this differentiation today (infact, the :mswin group even includes :mingw and apparently no one noticed.)

indirect:
So far, Bundler has only deprecated something when it was a significant UX improvement, or addressed a source of many bugs and tons of user confusion. I'm not convinced that forcing users to change `mswin` to `windows` is an improvement of any kind. Can we just leave the existing labels unchanged, without deprecation, and make all of them aliases for the new all-inclusive `windows`?

johnnyshields:
@indirect Im fine with that approach. :mswin is already an alias for :windows.

indirect:
Okay, summarizing just to make sure we are on the same page: everything in this PR stays the same except for using the old names doesn't print a deprecation message, and all of the existing names (plus the new name) run the same automatic-selection codepath.

If that is changed, I think I'm okay merging this whenever.

hsbt:
We might not look `mswin` user today. I say technical side.

https://bugs.ruby-lang.org/issues/19325#note-6 In Ruby core team, these platforms are treated as different things. I always have concerns people said `mingw` as Windows platform.

deivid-rodriguez:
I think this PR does not do what @indirect is asking at the moment. This PR only introduces a the deprecation, but does not change behavior.

The proposed alternative would be changing the behavior of `:mingw` and others to behave like `:mswin` and `:windows`. I'm also ok with that, because I believe nobody is using this differentation, but just noting it.

johnnyshields:
I've rebased and re-raised this PR here: https://github.com/rubygems/rubygems/pull/6391

PR #5884, opened 2022-08-29
Fix `gem info` with explicit `--version`

### LINKED ISSUE #5755: gem info don't work with flag version
RUBYGEMS VERSION: 3.3.7

I trying to get information from rubygems about author license etc. 

```
$ gem info rails --remote --version '7.0.2.2'

*** REMOTE GEMS ***

rails (7.0.3.1)
    Author: David Heinemeier Hansson
    Homepage: https://rubyonrails.org

    Full-stack web application framework.
```

why 7.0.3.1? where are the information about 7.0.2.2 ?

tonyaraujop:

Bug related to gem info using flag version not being applied correctly, as mentioned in #5755.

~First, it was added an error handler if gem version not exist. Second, it will be added a fix to the bug itself~

Make sure we request all versions from API when a specific version is specified, and then select only the one matching it.

Fixes #5755.

### DIFF
--- a/lib/rubygems/query_utils.rb
+++ b/lib/rubygems/query_utils.rb
@@ -151,7 +151,7 @@ module Gem::QueryUtils
       fetcher.detect(specs_type) { true }
     else
       fetcher.detect(specs_type) do |name_tuple|
-        name === name_tuple.name
+        name === name_tuple.name && options[:version].satisfied_by?(name_tuple.version)
       end
     end
 
@@ -159,7 +159,7 @@ module Gem::QueryUtils
   end
 
   def specs_type
-    if options[:all]
+    if options[:all] || options[:version].specific?
       if options[:prerelease]
         :complete
       else
--- a/test/rubygems/test_gem_commands_info_command.rb
+++ b/test/rubygems/test_gem_commands_info_command.rb
@@ -40,4 +40,30 @@ class TestGemCommandsInfoCommand < Gem::TestCase
     assert_match %r{#{@gem.summary}\n}, @ui.output
     assert_match "", @ui.error
   end
+
+  def test_execute_with_version_flag
+    spec_fetcher do |fetcher|
+      fetcher.spec "coolgem", "1.0"
+      fetcher.spec "coolgem", "2.0"
+    end
+
+    @cmd.handle_options %w[coolgem --remote --version 1.0]
+
+    use_ui @ui do
+      @cmd.execute
+    end
+
+    expected = <<-EOF
+
+*** REMOTE GEMS ***
+
+coolgem (1.0)
+    Author: A User
+    Homepage: http://example.com
+
+    this is a summary
+    EOF
+
+    assert_equal expected, @ui.output
+  end
 end


lib/rubygems/query_utils.rb
```diff
@@ -147,6 +147,12 @@ def show_remote_gems(name)
 
     fetcher = Gem::SpecFetcher.fetcher
 
+    raise Gem::MissingSpecError.new(
+      name.to_s.split(":").last.delete(")"),
```
deivid-rodriguez, line 151: What is this trying to do? Does `name` not have a correct value here?

tonyaraujop: `name` is a Regexp, and is like this `(?i-mx:coolgem)`

deivid-rodriguez: Oh I see, this is complicated because this mixing is reused by `gem info`, `gem search` and `gem list`, and while the former takes an "exact argument" by default, the latter interpret the argument as a pattern. So `gem list rails` lists all gems including "rails" in their names.

The `--version` argument makes sense only in "exact mode", so you should check whether `options[:exact]` has been passed, and in that case, `name` should be a plain string.

tonyaraujop: This is the test:
```ruby
  def test_execute_remote_version
    spec_fetcher do |fetcher|
      fetcher.spec "coolgem", "4.1.2"
      fetcher.spec "coolgem", "4.1.1"
    end

    @cmd.handle_options %w[coolgem --remote --version 4.1.1]

    use_ui @stub_ui do
      @cmd.execute
    end

    expected = <<-EOF

  *** REMOTE GEMS ***

  coolgem (4.1.1)
      EOF

    assert_equal expected, @stub_ui.output
  end
  ```
  
  And this is the output for `options`
  ```ruby
  {:domain=>:remote, :details=>false, :versions=>true, :installed=>nil,
:version=>#<Gem::Requirement:0x00007f5652e75ea0 @requirements=[["=", #<Gem::Version "4.1.1">]]>,
:explicit_prerelease=>false, :prerelease=>false, :args=>["coolgem"]}
  ```
  It sems that `options[:exact]` is nil

Comment (deivid-rodriguez) [reply lib/rubygems/query_utils.rb:151]: I see, for the test yes. But in real life `gem info` sets `exact` by default: https://github.com/rubygems/rubygems/blob/1bb66b849f7bd120f77b68adef3ab4178bf27696/lib/rubygems/commands/info_command.rb#L19.

So we should probably update all `gem info` tests to set `exact` explicitly (unless testing the behavior of `exact = false` explicitly), so that they better represent realworld use.

Can you also commit your WIP test, by the way?

Comment (tonyaraujop) [reply lib/rubygems/query_utils.rb:151]: I see...Sure, I'll commit right now.

Comment (tonyaraujop) [reply lib/rubygems/query_utils.rb:151]: Hey @deivid-rodriguez what this ternary actually does? Should not be `options[:exact].nil?`? 

https://github.com/rubygems/rubygems/blob/75e7ff5f0a39ae4d2b4dd06c449c4c92037fda87/lib/rubygems/query_utils.rb#L64

Comment (tonyaraujop) [reply lib/rubygems/query_utils.rb:151]: Also, I've noticed that when `options[:prerelease] == true` the command works as it should.

Comment (deivid-rodriguez) [reply lib/rubygems/query_utils.rb]: It seems that this code always works with regexps . So when the exact option is specified, an "exact regexp" is used. So if you type `gem info rails`, then the code will check that `name =~ /\Arails\Z/`, which is essentially equivalent to `name == "rails"`.

Regarding the `prerelease` option, it does not work for me:

### Without prerelease (WRONG)

```
$ gem info rails --remote --version 7.0.2  

*** REMOTE GEMS ***

rails (7.0.3.1)
    Author: David Heinemeier Hansson
    Homepage: https://rubyonrails.org

    Full-stack web application framework.

```

### With prerelease (WRONG TOO)

```
$ gem info rails --remote --version 7.0.2 --prerelease  

*** REMOTE GEMS ***

rails (7.0.0.rc3, 7.0.0.rc2, 7.0.0.rc1, 7.0.0.alpha2, 7.0.0.alpha1, 6.1.0.rc2, 6.1.0.rc1, 6.0.3.rc1, 6.0.2.rc2, 6.0.2.rc1, 6.0.1.rc1, 6.0.0.rc2, 6.0.0.rc1, 6.0.0.beta3, 6.0.0.beta2, 6.0.0.beta1, 5.2.4.rc1, 5.2.3.rc1, 5.2.2.rc1, 5.2.1.rc1, 5.2.0.rc2, 5.2.0.rc1, 5.2.0.beta2, 5.2.0.beta1, 5.1.7.rc1, 5.1.5.rc1, 5.1.4.rc1, 5.1.3.rc3, 5.1.3.rc2, 5.1.3.rc1, 5.1.2.rc1, 5.1.0.rc2, 5.1.0.rc1, 5.1.0.beta1, 5.0.6.rc1, 5.0.5.rc2, 5.0.5.rc1, 5.0.4.rc1, 5.0.2.rc1, 5.0.1.rc2, 5.0.1.rc1, 5.0.0.rc2, 5.0.0.rc1, 5.0.0.racecar1, 5.0.0.beta4, 5.0.0.beta3, 5.0.0.beta2, 5.0.0.beta1.1, 5.0.0.beta1, 4.2.10.rc1, 4.2.9.rc2, 4.2.9.rc1, 4.2.8.rc1, 4.2.7.rc1, 4.2.6.rc1, 4.2.5.rc2, 4.2.5.rc1, 4.2.4.rc1, 4.2.3.rc1, 4.2.1.rc4, 4.2.1.rc3, 4.2.1.rc2, 4.2.1.rc1, 4.2.0.rc3, 4.2.0.rc2, 4.2.0.rc1, 4.2.0.beta4, 4.2.0.beta3, 4.2.0.beta2, 4.2.0.beta1, 4.1.16.rc1, 4.1.15.rc1, 4.1.14.rc2, 4.1.14.rc1, 4.1.13.rc1, 4.1.12.rc1, 4.1.10.rc4, 4.1.10.rc3, 4.1.10.rc2, 4.1.10.rc1, 4.1.9.rc1, 4.1.6.rc2, 4.1.6.rc1, 4.1.2.rc3, 4.1.2.rc2, 4.1.2.rc1, 4.1.0.rc2, 4.1.0.rc1, 4.1.0.beta2, 4.1.0.beta1, 4.0.13.rc1, 4.0.10.rc2, 4.0.10.rc1, 4.0.6.rc3, 4.0.6.rc2, 4.0.6.rc1, 4.0.4.rc1, 4.0.1.rc4, 4.0.1.rc3, 4.0.1.rc2, 4.0.1.rc1, 4.0.0.rc2, 4.0.0.rc1, 4.0.0.beta1, 3.2.15.rc3, 3.2.15.rc2, 3.2.15.rc1, 3.2.14.rc2, 3.2.14.rc1, 3.2.13.rc2, 3.2.13.rc1, 3.2.9.rc3, 3.2.9.rc2, 3.2.9.rc1, 3.2.8.rc2, 3.2.8.rc1, 3.2.7.rc1, 3.2.4.rc1, 3.2.3.rc2, 3.2.3.rc1, 3.2.2.rc1, 3.2.0.rc2, 3.2.0.rc1, 3.1.5.rc1, 3.1.4.rc1, 3.1.2.rc2, 3.1.2.rc1, 3.1.1.rc3, 3.1.1.rc2, 3.1.1.rc1, 3.1.0.rc8, 3.1.0.rc6, 3.1.0.rc5, 3.1.0.rc4, 3.1.0.rc3, 3.1.0.rc2, 3.1.0.rc1, 3.1.0.beta1, 3.0.13.rc1, 3.0.12.rc1, 3.0.10.rc1, 3.0.9.rc5, 3.0.9.rc4, 3.0.9.rc3, 3.0.9.rc1, 3.0.8.rc4, 3.0.8.rc2, 3.0.8.rc1, 3.0.7.rc2, 3.0.7.rc1, 3.0.6.rc2, 3.0.6.rc1, 3.0.5.rc1, 3.0.4.rc1, 3.0.0.rc2, 3.0.0.rc, 3.0.0.beta4, 3.0.0.beta3, 3.0.0.beta2, 3.0.0.beta, 2.3.9.pre, 2.3.8.pre1)
    Author: David Heinemeier Hansson
    Homepage: https://rubyonrails.org

    Full-stack web application framework.
```

File: lib/rubygems/query_utils.rb [hunk 2]
```diff
@@ -147,6 +147,11 @@ def show_remote_gems(name)
 
     fetcher = Gem::SpecFetcher.fetcher
 
+    gem_name, gem_versions = from_fetcher(fetcher).slice(:name, :versions).values
+    option_version = options[:version].requirements[0][1]
+
+    raise Gem::MissingSpecError.new(gem_name, option_version) unless gem_versions.include?(Gem::NameTuple.new(gem_name, option_version))
+
```
deivid-rodriguez, line 154: Not sure what the above is trying to do, but what I think we want is:

* Let the index be complete if version is passed, so that we not only get the latest spec, and can filter later the proper version that has been requested. I think you want to add something like this at the beginning of the `execute` method:

```ruby
options[:all] = true if options[:version].specific?
```

The above is just a hacky way to check whether a specific version has been passed, because the default value for version is an open requirement (>= 0).

Then in `show_remote_gems` you probably want to change the filtering to be like this:

```diff
     spec_tuples = if name.nil?
       fetcher.detect(specs_type) { true }
     else
       fetcher.detect(specs_type) do |name_tuple|
-        name === name_tuple.name
+        name === name_tuple.name && options[:version].satisfied_by?(name_tuple.version)
       end
     end
```

tonyaraujop: > Not sure what the above is trying to do, but what I think we want is:

This throws an error if the version passed do not exist.

test/rubygems/test_gem_commands_query_command.rb
```diff
@@ -841,6 +841,45 @@ def test_execute_exact_multiple
...
+    @cmd.handle_options %w[coolgem --remote --version 4.1.3]
+
+   
+    assert_raise Gem::MissingSpecError do
+      use_ui @stub_ui do
+        @cmd.execute
+      end
+      raise "Could not find 'coolgem' (4.1.3) among 2 total gem(s)"
+    end
+  end
```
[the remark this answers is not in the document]
deivid-rodriguez: Yeah, but the other file actually instantiates a `Gem::Command::InfoCommand` which is what we want to test here.

lib/rubygems/query_utils.rb
deivid-rodriguez: I see. I say let's remove that part for now and focus on getting the two specs you added passing without breaking other things.

tonyaraujop: I've updated files and tests...when you have a chance to look, let me know please.

test/rubygems/test_gem_commands_info_command.rb
```diff
@@ -40,4 +40,30 @@ def test_execute
     assert_match %r{#{@gem.summary}\n}, @ui.output
     assert_match "", @ui.error
   end
+
+  def test_execute_with_multiple_versions
```
deivid-rodriguez, line 44:
```suggestion
  def test_execute_with_version_flag
```

deivid-rodriguez (commented):
Just two tiny comments. If you can address them, rebase and squash everything to a single commit, I think this is ready!! Thanks so much for your patience! 🎉

tonyaraujop:
Sure, no problem at all. Happy to help.

deivid-rodriguez (approved):
Great work @tonyaraujop, thanks!!

PR #5889, opened 2022-08-31
Support non gnu libc arm-linux-eabi platforms

ntkme:

#5852 made `linux` implicitly `linux-gnu`, however this logic does not apply to arm 32-bit variants. E.g. Currently `arm-linux-eabi` is not implicit `arm-linux-gnueabi`. 

This PR enhances #5852 to support version `eabi`, `eabihf`, etc. used by 32-bit arm variants.


deivid-rodriguez (approved):
Makes sense to me!

ntkme:
Looks like one of the runner might have lost connectivity and caused a CI test to stuck and attempt to cancel that job also got stuck and eventually timed out. May need a rerun of that failed job.

PR #5895, opened 2022-09-02
Introduce bundle-version(1) man

tnir:

We miss bundle-version(1) man, meaning that users have one generated with Thor while we have had `https://bundler.io/v2.3/bundle_version.html` on the [website](https://bundler.io/).

Options indicated in `bundle help version` (see the below) do not work in `bundle version`:

```
$ bundle help version
Usage:
  bundle version

Options:
      [--no-color]                 # Disable colorization in output
  -r, [--retry=NUM]                # Specify the number of times you wish to attempt network commands
  -V, [--verbose], [--no-verbose]  # Enable verbose output mode

Prints the bundler's version information
```

Relates to https://github.com/rubygems/bundler-site/issues/879

Introduces bundle-version(1) man while we have had `https://bundler.io/v2.3/bundle_version.html` on the [website](https://bundler.io/).

### DIFF
--- a/bundler/lib/bundler/man/bundle-version.1.ronn
+++ b/bundler/lib/bundler/man/bundle-version.1.ronn
@@ -1,0 +1,24 @@
+bundle-version(1) -- Prints Bundler version information
+=======================================================
+
+## SYNOPSIS
+
+`bundle version`
+
+## DESCRIPTION
+
+Prints Bundler version information.
+
+## OPTIONS
+
+No options.
+
+## EXAMPLE
+
+Print the version of Bundler with build date and commit hash of the in the Git source.
+
+    bundle version
+
+shows `Bundler version 2.3.21 (2022-08-24 commit d54be5fdd8)` for example.
+
+cf. `bundle --version` shows `Bundler version 2.3.21`.
--- a/bundler/spec/commands/help_spec.rb
+++ b/bundler/spec/commands/help_spec.rb
@@ -23,8 +23,8 @@ RSpec.describe "bundle help" do
   end
 
   it "still outputs the old help for commands that do not have man pages yet" do
-    bundle "help version"
-    expect(out).to include("Prints the bundler's version information")
+    bundle "help fund"
+    expect(out).to include("Lists information about gems seeking funding assistance")
   end
 
   it "looks for a binary and executes it with --help option if it's named bundler-<task>" do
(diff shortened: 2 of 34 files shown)


bundler/spec/commands/help_spec.rb
```diff
@@ -22,9 +22,9 @@
     expect(out).to match(/bundle-install/)
   end
 
-  it "still outputs the old help for commands that do not have man pages yet" do
-    bundle "help version"
-    expect(out).to include("Prints the bundler's version information")
+  it "still outputs the old help for commands that do not have man pages yet", :bundler => "< 3" do
+    bundle "help console"
+    expect(out).to include("Opens an IRB session with the bundle pre-loaded")
```
tnir, line 27: Note: `bundle help console` is the very last command which depends on Thor-generated man pages and this command is to be removed in Bundler 3 so we can test this case only in Bundler 2.x.

tnir:
@deivid-rodriguez What do you think?

bundler/spec/commands/help_spec.rb
deivid-rodriguez: I think `bundle help fund` also does not have a specific man page, so we could use that instead?

bundler/lib/bundler/man/bundle-version.1.ronn
```diff
@@ -0,0 +1,24 @@
+bundle-version(1) -- Prints the bundler's version information
+=============================================================
+
+## SYNOPSIS
+
+`bundle version`
+
+## DESCRIPTION
+
+Prints the bundler's version information.
```
deivid-rodriguez, line 10:
```suggestion
Prints Bundler version information.
```

I think the above reads slightly better, what do you think?

deivid-rodriguez (approved):
Small comment on wording, but looks great otherwise!

bundler/lib/bundler/man/bundle-version.1.ronn
tnir: Yes, makes sense to me. That sentence just came from:
https://github.com/rubygems/rubygems/blob/ff67cb0fba3526fa411c0fdf73269f22c46c7d9a/bundler/lib/bundler/cli.rb#L520
so updating the sentence in `cli.rb` as well.

deivid-rodriguez: Oh, I see. Well, thanks for updating it!

deivid-rodriguez (approved):

PR #5896, opened 2022-09-03
Use a more accurate source code uri in gemspec

dduugg:
Updates the bundler metadata source_code_uri
Also, since `metadata` support was [introduced](https://guides.rubygems.org/specification-reference/) in rubygems 2.0, and the required rubygems version is ">= 2.5.2", the check for `metadata=` support has been removed.

Locating the `bundler` source code

Fix the link to the bundler source code within the bundler gemspec


deivid-rodriguez (approved):
Awesome, thank you!

PR #5897, opened 2022-09-03
Feature: `bundle add` supports `--path` option

flavorjones:

As a developer of a commonly used gem (nokogiri), I frequently want to test downstream libraries against the version under development. To do this, I want to check out that downstream library in a sibling directory and add a line like this to the `Gemfile`:

```ruby
gem "nokogiri", path: "../nokogiri"
```

There's currently no good way to do this with the `bundle add` command, and so I have to do something like:

```sh
if grep nokogiri Gemfile ; then
  sed -i 's/\(.*nokogiri.*\)/\1, path: ".."/' Gemfile
else
  echo "gem 'nokogiri', path: '..'" >> Gemfile
fi
bundle install
```

I'd prefer to do something like:

```sh
bundle remove nokogiri || true
bundle add nokogiri --path="../nokogiri"
```

[Here one place](https://github.com/sparklemotion/nokogiri/blob/186edd0f3fab60ea57ecf48e0bf502c29d350eed/.github/workflows/downstream.yml#L64-L68) I'm doing this in the real world, and [here's another](https://github.com/rails/tailwindcss-rails/pull/197/files#diff-5cdc711f62bf5479ce06503ca175423c6bc94c414b774a2a6e88294d71d2a94aR21-R22).

I've followed the example in c3e54ac and:

- add spec coverage for `bundle add --path=PATH`
- add a `--path` option to the `add` command in `lib/bundler/cli.rb`
- add a `#path` attribute to `Dependency`
- handle the path attribute in `Injector`
- update the man page for "bundle add"

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -372,6 +372,7 @@ module Bundler
     method_option "group", :aliases => "-g", :type => :string
     method_option "source", :aliases => "-s", :type => :string
     method_option "require", :aliases => "-r", :type => :string, :banner => "Adds require path to gem. Provide false, or a path as a string."
+    method_option "path", :type => :string
     method_option "git", :type => :string
     method_option "github", :type => :string
     method_option "branch", :type => :string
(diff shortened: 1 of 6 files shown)


deivid-rodriguez (approved):
Handy! Thank you ❤️

simi (approved):

Issue #1161, opened 2015-02-05
Gem install minitest on NFS fails


EHadoux:
Hi, 

installing ```minitest```, ```hoe``` or whatever gems from seattlerb on a nfs partition fails, e.g.:
```
/home/hadouxe/testruby/bin/gem install minitest
ERROR:  While executing gem ... (Errno::EACCES)
    Permission denied @ rb_sysopen - /home/hadouxe/.gems/gems/minitest-5.5.1/History.txt
```

This seems independent to rbenv (the example above does not use it).
You can see at seattlerb/minitest#435 all the tests and why I am opening this issue here.
I honestly do not know if the bug comes from minitest or rubygems.

Anyways, thank you very much.

jirutka:
IMHO the bug comes from this [hoe](https://github.com/seattlerb/hoe) crap, that is what the seattlerb gems have in common.

EHadoux:
I guess it is. Is there any ways to skip the minitest part in ruby installation?
It seems mandatory and that prevents me from installing new rubies.

zenspider:
@jirutka crap? "humble" opinion? interesting...

zenspider:
@drbrain, I believe this to be a bug in rubygems' tar support. I've verified that it is NOT an issue with NFS by manually untarring a minitest gem on one of the affected hosts w/ NFS and then used `gem unpack` to reproduce. Details below:

```
minitestgem@ssh:~$ rm -rf xxx; mkdir xxx; tar zxf data.tar.gz -C xxx
minitestgem@ssh:~$ ls -l xxx
total 68
-r--r--r-- 1 minitestgem minitestgem  1918 2015-01-10 01:35 design_rationale.rb
-r--r--r-- 1 minitestgem minitestgem 32911 2015-01-10 01:35 History.txt
drwxrwxr-x 4 minitestgem minitestgem    65 2015-02-07 14:20 lib
-r--r--r-- 1 minitestgem minitestgem   617 2015-01-10 01:35 Manifest.txt
-r--r--r-- 1 minitestgem minitestgem  2003 2015-01-10 01:35 Rakefile
-r--r--r-- 1 minitestgem minitestgem 19794 2015-01-10 01:35 README.txt
drwxrwxr-x 3 minitestgem minitestgem    29 2015-02-07 14:20 test
minitestgem@ssh:~$ rm -rf xxx
[...]

PR #5906, opened 2022-09-06
Mask the file mode when extracting files

kddnewton:
When extracting files from the tarball, a mode is retrieved from
the header. Occasionally you'll encounter a gem that was packaged
on a system whose permission bits result in a value that is larger
than the value that File.chmod will allow (anything >= 2^16). In
that case the extraction fails with a RangeError, which is pretty
esoteric.

If you extract the tarball with the tar and gunzip utilities, the
file permissions end up being just the bottom 16 bits masked off
from the original value. I've mirrored that behavior here. Per the
tar spec:

> Modes which are not supported by the operating system restoring
> files from the archive will be ignored.

I think that basically means what I've done here.

You can test this by trying to extract the files for `Bluebie-legs-0.6.2`,
which was the first gem that I found (alphabetically) that had this
issue. `cameroncox-slicehost-tools-0.0.3` also had this issue, with
a slightly different set of permissions.

Cannot extract gems files with permissions too large for an unsigned
short.

Mask out the bottom 16 bits only.

I did not end up writing tests for this because I'm not entirely sure how I would. The existing tests all pass, which is good. I suppose I could attempt to package a gem such as the failure cases and vendor it into the test files, but I'm not sure it's worth the trouble.


deivid-rodriguez:
Nice!

We do vendor some problematic gem packages for the shake of testing:

```
$ fd gem$ test
test/rubygems/packages/ascii_binder-0.1.10.1.gem
test/rubygems/packages/ill-formatted-platform-1.0.0.10.gem
```

`ascii_bindir-0.1.10.1.gem` is actually a realworld gem. I don't think it'd be a big deal to vendor one of these too.

kddnewton:
@deivid-rodriguez alright I've vendored one of the problematic gems 👍

deivid-rodriguez:
The failures mean you need to run ' rake manifest` and commit the result.

deivid-rodriguez:
The new test seems to be failing in TruffleRuby, can you have a look?

kddnewton:
@deivid-rodriguez I've updated the commit message with the fix. Basically there was an issue with buffered writes happening to the filesystem and CRuby and TruffleRuby had differing behavior. It should be consistent now.

kddnewton:
For context, here's the issue where this behavior was introduced: https://github.com/rubygems/rubygems/issues/1161.

deivid-rodriguez (approved):
Wow, thanks for the fix and the thorough investigation!

kddnewton:
Oh gosh now it's failing on windows... okay I'll keep looking.

kddnewton:
Oh wait it looks like maybe I should be skipping this test on windows. @deivid-rodriguez does that sound right? I'm seeing a lot of pending "chmod not supported" stuff in the test file?

deivid-rodriguez:
I think file permissions in Windows are not very well defined, I think it's fine to skip 👍.

kddnewton:
Nice I think this is finally good to go

deivid-rodriguez:
Thank you! ❤️

Issue #5910, opened 2022-09-07
Feature proposal: Bundler: Add warning to the lockfile contents warning not to hand-edit it


zofrex:
I see beginners (and some non-beginners) make this mistake quite frequently of hand-editing the lockfile to make changes. I think a warning message in the lockfile itself to not do this could be extremely effective in pointing people in the right direction.

This might look something like this:

```
# Warning: Do not edit this file by hand! You should only update this file by running Bundler

GEM
  remote: https://rubygems.org/
  specs:
    actioncable (7.0.3.1)
      actionpack (= 7.0.3.1)
      activesupport (= 7.0.3.1)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
... etc ...
```

zofrex:
Based on [the PoC](https://github.com/rubygems/rubygems/pull/5911) the smoothest way to add this was as a new section in the lockfile, which would look something like this:

```
EDIT WARNING
  Do not edit this file by hand!
  You should only update this file by running Bundler
  e.g. `bundle install` or `bundle update`
  See https://bundler.io for more information test

GEM
  remote: https://rubygems.org/
  specs:
    actioncable (7.0.3.1)
      actionpack (= 7.0.3.1)
      activesupport (= 7.0.3.1)
```

Questions:

1. Are we happy with it being a regular lockfile section? (Constraints: The first line cannot start with whitespace and needs to be constant over time, i.e. probably shouldn't contain any part of the message as we may want to reword it later. Subsequent lines need to start with whitespace. By convention, but not a hard constraint, it would be all caps.)
2. What name should the section have? For simplicity we would ideally not change this again as it is machine-readable, so let's pick one that everyone is really happy with? I think ideally this needs to be fairly human readable (as it will be the first line of the warning) but also not something that might ever collide with any future additions (so just "WARNING" might not be such a great name, for example).
3. What should the warning say? This is less important to settle on than the header as this can be updated easily in the future, but it's still worth discussing what this should contain. Is it a hard "do not edit by hand ever" or is this a grey area, where it's ok sometimes? What are the best things to suggest instead, or resources to point people to?
4. Should this message be added to existing lockfiles, or only to newly-created lockfiles? @simi suggested only newly created ones, but I think there would be more value in adding it to existing lockfiles too: a lot of projects already have lockfiles and developers editing them by hand, it would be valuable to nudge them onto the right path.

deivid-rodriguez:
[...]

PR #5911, opened 2022-09-07
PoC: Add warning to lockfile not to hand-edit it

zofrex:

Proposed implementation of #5910, read that ticket for motivation

* Add a new section to the lockfile, "EDIT WARNING", at the top, using the existing section formatting so it plays nicely with the existing parsing etc
* Add a warning inside this section explaining not to edit the lockfile by hand and what to do instead
* Write this section out both when creating new lockfiles, and updating existing ones
* Don't write this section out when updating an existing lockfile that does not already contain it (in other words, only add it to new lockfiles, and ones that already have it)

This results in the following being added to the top of new lockfiles:

```
EDIT WARNING
  Do not edit this file by hand!
  You should only update this file by running Bundler
  e.g. `bundle install` or `bundle update`
  See https://bundler.io for more information test

GEM
  remote: https://rubygems.org/
  specs:
    actioncable (7.0.3.1)
      actionpack (= 7.0.3.1)
      activesupport (= 7.0.3.1)
... etc ...
```

### Up for debate

* @simi suggested only adding the message to newly created lockfiles, but I think there would be more value in adding it to existing lockfiles too? A lot of projects already have lockfiles and developers editing them by hand, it would be nice to stop that.
* The name of the section: For simplicity we would ideally not change this again as it is machine-readable, so let's pick one that everyone is really happy with? I think ideally this needs to be fairly human readable (as it will be the first line of the warning) but also not something that might ever collide with any future additions (so just "WARNING" might not be such a great name, for example)
* The contents of the warning: This is less important to settle on than the header as this can be updated easily in the future, but it's still worth discussing what this should contain. Is it a hard "do not edit by hand ever" or is this a grey area, where it's ok sometimes? What are the best things to suggest instead, or resources to point people to?

### Todo

Write tests. Fix existing tests this breaks. This is just a proof of concept to get the conversation started, once we're happy with the shape it's in I'll add & fix tests.


zofrex:
Closing as it has served its purpose for now, have summarised the key questions in [the feature discussion](https://github.com/rubygems/rubygems/issues/5910) to continue the conversation there.

PR #5920, opened 2022-09-09
Give a better error when trying to write the lock file on a read-only filesystem

### LINKED ISSUE #5893: Bundler 2.3.21 attempts to update PLATFORMS in Gemfile.lock on bundle exec, fails on read-only filesystem

We run our apps with a read-only filesystem in a Docker container.

In many of our apps' Gemfile.lock, we have a `PLATFORMS` section that lists various platforms we use for development and deployment; we do not typically maintain this list ourselves. Example:

```
PLATFORMS
  x86_64-darwin-17
  x86_64-darwin-19
  x86_64-darwin-20
  x86_64-linux-musl
```

As of Bundler 2.3.21, when we run our app using `bundle exec`, it attempts to write to Gemfile.lock to add the platform relevant for the current system (e.g. linux on the server). Because we're running in a read only file system, this fails with:

`Errno::EROFS: Read-only file system @ rb_sysopen - /app/Gemfile.lock`

**As a workaround**, we have found that removing other platforms and adding the `ruby` platform to Gemfile.lock prevents the issue, but we don't really understand why, or if this is a good general solution.

Yes :) 

Running `bundle exec` commands does not attempt to write to Gemfile.lock or any other files.

### What actually happened?

`Errno::EROFS: Read-only file system @ rb_sysopen - /app/Gemfile.lock`

Can provide this if needed.

deivid-rodriguez:

If `bundle exec` tries to write to the lock file on a read-only filesystem, Bundler will fail but it's unclear why.

Suggest a solution to fix the problem.

Closes #5893.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -412,39 +412,8 @@ module Bundler
 
       raise ProductionError, "Frozen mode is set, but there's no lockfile" unless lockfile_exists?
 
-      added =   []
-      deleted = []
-      changed = []
-
-      added.concat @new_platforms.map {|p| "* platform: #{p}" }
-      deleted.concat @removed_platforms.map {|p| "* platform: #{p}" }
-
-      added.concat new_deps.map {|d| "* #{pretty_dep(d)}" } if new_deps.any?
-      deleted.concat deleted_deps.map {|d| "* #{pretty_dep(d)}" } if deleted_deps.any?
-
-      both_sources = Hash.new {|h, k| h[k] = [] }
-      current_dependencies.each {|d| both_sources[d.name][0] = d }
-      current_locked_dependencies.each {|d| both_sources[d.name][1] = d }
-
-      both_sources.each do |name, (dep, lock_dep)|
-        next if dep.nil? || lock_dep.nil?
-
-        gemfile_source = dep.source || default_source
-        lock_source = lock_dep.source || default_source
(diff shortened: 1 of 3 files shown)


lyricsboy (commented):
This looks like a good improvement to me! It would definitely have helped my understanding of the problem.

deivid-rodriguez:
Thanks! I will eventually loop back and investigate why this broke CI.

bundler/lib/bundler/shared_helpers.rb
```diff
@@ -115,10 +115,6 @@ def filesystem_access(path, action = :write, &block)
       raise NoSpaceOnDeviceError.new(path, action)
     rescue Errno::ENOTSUP
       raise OperationNotSupportedError.new(path, action)
-    rescue Errno::EEXIST, Errno::ENOENT
-      raise
-    rescue SystemCallError => e
-      raise GenericSystemCallError.new(e, "There was an error #{[:create, :write].include?(action) ? "creating" : "accessing"} `#{path}`.")
```
segiddins, line 121: I'm a bit concerned about removing this fallback. Instead, what if we raised a new exception on `Errno::EROFS` that `write_lock` could rescue

deivid-rodriguez: Yes, I was just thinking the same, sounds good 👍.

deivid-rodriguez:
This now rescues a specific Bundler exception for read-only situations 👍

PR #5922, opened 2022-09-11
Add 'call for update' to RubyGems install command.

simi:

Promote update of RubyGems, since it is not currently best practice to update local installations.

Mostly based on similar features in yarn and gh CLI. More info at https://bundler.slack.com/archives/C0HJMKWH4/p1639683555027700.

I have added small banner to `gem install` command which is printed when all following conditions are met.

- `gem install` command was successful
- you have not opted-out this feature by `prevent_update_suggestion` entry in `gemrc` file
- you have not opted-out this feature by `RUBYGEMS_PREVENT_UPDATE_SUGGESTION` environment variable
- you have interactive terminal session active
- you're not running rubygems dev/pre-release version (to not be annoying during development of RubyGems)
- there is no `disable_system_update_message` set (since `gem update --system` is blocked when set)
- CI is not detected
- `gemrc` file is writable (empty one will be created during check to ensure if not present yet)
- message wasn't shown already in last 7 days (to not be annoying, we can decrease this time later once battle-tested)

I tried to avoid any links in the message to make it futrue-proof (since link can expire easily). 

Plan for later is add similar to `bundle install` and `bundle update` commands.

## Example

```
[retro@retro  rubygems (call-for-update %=)]❤ ruby -Ilib -S bin/gem install ABO
Successfully installed ABO-0.0.2
Parsing documentation for ABO-0.0.2
Done installing documentation for ABO after 0 seconds
1 gem installed

A new release of RubyGems is available: 3.2.0 → 3.3.22!
Run `gem update --system 3.3.22` to update your installation.

[retro@retro  rubygems (call-for-update %=)]❤ 
```

---

closes https://github.com/rubygems/rubygems/issues/1620, https://github.com/rubygems/rubygems/pull/3096

### DIFF
--- a/lib/rubygems/update_suggestion.rb
+++ b/lib/rubygems/update_suggestion.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+##
+# Mixin methods for Gem::Command to promote available RubyGems update
+
+module Gem::UpdateSuggestion
+  # list taken from https://github.com/watson/ci-info/blob/7a3c30d/index.js#L56-L66
+  CI_ENV_VARS = [
+    "CI", # Travis CI, CircleCI, Cirrus CI, Gitlab CI, Appveyor, CodeShip, dsari
+    "CONTINUOUS_INTEGRATION", # Travis CI, Cirrus CI
+    "BUILD_NUMBER", # Jenkins, TeamCity
+    "CI_APP_ID", "CI_BUILD_ID", "CI_BUILD_NUMBER", # Applfow
+    "RUN_ID" # TaskCluster, dsari
+  ].freeze
+
+  ONE_WEEK = 7 * 24 * 60 * 60
+
+  ##
+  # Message to promote available RubyGems update with related gem update command.
+
+  def update_suggestion
+    <<-MESSAGE
+
+A new release of RubyGems is available: #{Gem.rubygems_version} → #{Gem.latest_rubygems_version}!
+Run `gem update --system #{Gem.latest_rubygems_version}` to update your installation.
+
+    MESSAGE
+  end
+
+  ##
+  # Determines if current environment is eglible for update suggestion.
+
+  def eglible_for_update?
+    # explicit opt-out
+    return false if Gem.configuration[:prevent_update_suggestion]
+    return false if ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"]
+
+    # focus only on human usage of final RubyGems releases
+    return false unless Gem.ui.tty?
+    return false if Gem.rubygems_version.prerelease?
+    return false if Gem.disable_system_update_message
+    return false if ci?
+
+    # check makes sense only when we can store of last try
+    # otherwise we will not be able to prevent annoying update message
+    # on each command call
+    return unless Gem.configuration.config_file_writable?
+
+    # load time of last check, ensure the difference is enough to repeat the suggestion
+    check_time = Time.now.to_i
+    last_update_check = Gem.configuration[:last_update_check] || 0
+    return false if (check_time - last_update_check) < ONE_WEEK
+
+    # compare current and latest version, this is the part where
+    # latest rubygems spec is fetched from remote
+    (Gem.rubygems_version < Gem.latest_rubygems_version).tap do |eglible|
+      if eglible
+        # store the time of last successful check into config file
+        Gem.configuration[:last_update_check] = check_time
+        Gem.configuration.write
+      end
+    end
+  rescue # don't block install command on any problem
+    false
+  end
+
+  def ci?
+    CI_ENV_VARS.any? {|var| ENV.include?(var) }
+  end
+end
--- a/lib/rubygems/update_suggestion.rb
+++ b/lib/rubygems/update_suggestion.rb
@@ -53,10 +53,11 @@ Run `gem update --system #{Gem.latest_rubygems_version}` to update your installa
 
     # compare current and latest version, this is the part where
     # latest rubygems spec is fetched from remote
-    if (Gem.rubygems_version < Gem.latest_rubygems_version)
+    (Gem.rubygems_version < Gem.latest_rubygems_version).tap do |eglible|
       # store the time of last successful check into state file
       Gem.configuration.last_update_check = check_time
-      return true
+
+      return eglible
     end
   rescue # don't block install command on any problem
     false
--- a/lib/rubygems/update_suggestion.rb
+++ b/lib/rubygems/update_suggestion.rb
@@ -41,24 +41,22 @@ Run `gem update --system #{Gem.latest_rubygems_version}` to update your installa
     return false if Gem.disable_system_update_message
     return false if ci?
 
-    # check makes sense only when we can store of last try
-    # otherwise we will not be able to prevent annoying update message
+    # check makes sense only when we can store timestamp of last try
+    # otherwise we will not be able to prevent "annoying" update message
     # on each command call
-    return unless Gem.configuration.config_file_writable?
+    return unless Gem.configuration.state_file_writable?
 
     # load time of last check, ensure the difference is enough to repeat the suggestion
     check_time = Time.now.to_i
-    last_update_check = Gem.configuration[:last_update_check] || 0
+    last_update_check = Gem.configuration.last_update_check
     return false if (check_time - last_update_check) < ONE_WEEK
 
     # compare current and latest version, this is the part where
     # latest rubygems spec is fetched from remote
-    (Gem.rubygems_version < Gem.latest_rubygems_version).tap do |eglible|
-      if eglible
-        # store the time of last successful check into config file
-        Gem.configuration[:last_update_check] = check_time
-        Gem.configuration.write
-      end
+    if (Gem.rubygems_version < Gem.latest_rubygems_version)
+      # store the time of last successful check into state file
+      Gem.configuration.last_update_check = check_time
+      return true
     end
   rescue # don't block install command on any problem
     false
--- a/test/rubygems/test_gem_update_suggestion.rb
+++ b/test/rubygems/test_gem_update_suggestion.rb
@@ -27,7 +27,7 @@ class TestUpdateSuggestion < Gem::TestCase
     original_config, Gem.configuration[:prevent_update_suggestion] = Gem.configuration[:prevent_update_suggestion], nil
     original_env, ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"] = ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"], nil
     original_disable, Gem.disable_system_update_message = Gem.disable_system_update_message, nil
-    Gem.configuration[:last_update_check] = nil
+    Gem.configuration.last_update_check = 0
 
     Gem.ui.stub :tty?, tty do
       Gem.stub :rubygems_version, rubygems_version do
@@ -61,10 +61,10 @@ class TestUpdateSuggestion < Gem::TestCase
     with_eglible_environment(cmd: @cmd) do
       Time.stub :now, 123456789 do
         assert @cmd.eglible_for_update?
-        assert_equal Gem.configuration[:last_update_check], 123456789
+        assert_equal Gem.configuration.last_update_check, 123456789
 
         # test last check is written to config file
-        assert File.read(Gem.configuration.config_file_name).match("last_update_check: 123456789")
+        assert File.read(Gem.configuration.state_file_name).match("123456789")
       end
     end
   end
@@ -122,7 +122,7 @@ class TestUpdateSuggestion < Gem::TestCase
 
   def test_eglible_for_update_unwrittable_config
     with_eglible_environment(ci: true, cmd: @cmd) do
-      Gem.configuration.stub :config_file_writable?, false do
+      Gem.configuration.stub :state_file_writable?, false do
         refute @cmd.eglible_for_update?
       end
     end
@@ -130,7 +130,7 @@ class TestUpdateSuggestion < Gem::TestCase
 
   def test_eglible_for_update_notification_delay
     with_eglible_environment(cmd: @cmd) do
-      Gem.configuration[:last_update_check] = Time.now.to_i
+      Gem.configuration.last_update_check = Time.now.to_i
       refute @cmd.eglible_for_update?
     end
   end
--- a/test/rubygems/test_gem_update_suggestion.rb
+++ b/test/rubygems/test_gem_update_suggestion.rb
@@ -9,6 +9,9 @@ class TestUpdateSuggestion < Gem::TestCase
 
     @cmd = Gem::Command.new "dummy", "dummy"
     @cmd.extend Gem::UpdateSuggestion
+    @start_time = 1_000_000
+    @minute = 60 * 60
+    @week = 7 * 24 * @minute
   end
 
   def with_eglible_environment(**params)
@@ -22,12 +25,13 @@ class TestUpdateSuggestion < Gem::TestCase
     rubygems_version: Gem::Version.new("1.2.3"),
     latest_rubygems_version: Gem::Version.new("2.0.0"),
     ci: false,
+    reset_last_update_check: true,
     cmd:
   )
     original_config, Gem.configuration[:prevent_update_suggestion] = Gem.configuration[:prevent_update_suggestion], nil
     original_env, ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"] = ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"], nil
     original_disable, Gem.disable_system_update_message = Gem.disable_system_update_message, nil
-    Gem.configuration.last_update_check = 0
+    Gem.configuration.last_update_check = 0 if reset_last_update_check
 
     Gem.ui.stub :tty?, tty do
       Gem.stub :rubygems_version, rubygems_version do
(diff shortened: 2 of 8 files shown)


lib/rubygems/update_suggestion.rb
```diff
@@ -0,0 +1,70 @@
...
+    last_update_check = Gem.configuration[:last_update_check] || 0
+    return false if (check_time - last_update_check) < ONE_WEEK
+
+    # compare current and latest version, this is the part where
+    # latest rubygems spec is fetched from remote
+    (Gem.rubygems_version < Gem.latest_rubygems_version).tap do |eglible|
+      if eglible
+        # store the time of last successful check into config file
+        Gem.configuration[:last_update_check] = check_time
+        Gem.configuration.write
```
skipkayhil, line 60: Would it be possible to store this somewhere other than the configuration file? 

Libraries like the `update-notifier` npm package write to a file in `~/.config` directory, which is undesired because people version control it. I [opened an issue there](https://github.com/yeoman/update-notifier/issues/180) and the maintainer suggested `XDG_DATA_HOME` would be better, [npm uses a cache directory](https://github.com/npm/cli/blob/e95017a07b041cbb3293e659dad853f76462c108/lib/utils/update-notifier.js#L27)

indirect: I believe `Gem.configuration` already respects `XDG_DATA_HOME` if it is set.

indirect:
Very nice work, thanks for setting this up!

lib/rubygems/update_suggestion.rb
simi: I have used configuration file as the simplest option to open discussion. What about `Gem.data_home`?

https://github.com/rubygems/rubygems/blob/41ce5ce11d0e76a964a8e8a43501a7e389fbf58d/lib/rubygems/defaults.rb#L146-L148

We can create separate file like `last_update_check` and fill it simply with the timestamp number.

simi: Alternatively `Gem.cache_home` respecting `XDG_CACHE_HOME` could be the choice.



> I believe `Gem.configuration` already respects `XDG_DATA_HOME` if it is set.

If I understand it well, Gem.configuration respects `XDG_CONFIG_HOME` and that's probably the subject to put into version control.

skipkayhil: > If I understand it well, `Gem.configuration` respects `XDG_CONFIG_HOME` and that's probably the subject to put into version control.

Yep, exactly! Either `cache_home` or `data_home` would be fine with me, as long as its outside of `config_home` 😄

simi: :thinking: Looking at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html again, it seems `XDG_STATE_HOME` is directory we're looking for.

> $XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used.

> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

:information_source: I have raised question at freedesktop IRC channel to confirm. :pray: Once done, I'll update this PR.

lib/rubygems/update_suggestion.rb
```diff
@@ -0,0 +1,70 @@
...
+# Mixin methods for Gem::Command to promote available RubyGems update
+
+module Gem::UpdateSuggestion
+  # list taken from https://github.com/watson/ci-info/blob/7a3c30d/index.js#L56-L66
+  CI_ENV_VARS = [
+    "CI", # Travis CI, CircleCI, Cirrus CI, Gitlab CI, Appveyor, CodeShip, dsari
+    "CONTINUOUS_INTEGRATION", # Travis CI, Cirrus CI
+    "BUILD_NUMBER", # Jenkins, TeamCity
+    "CI_APP_ID", "CI_BUILD_ID", "CI_BUILD_NUMBER", # Applfow
+    "RUN_ID" # TaskCluster, dsari
```
deivid-rodriguez, line 13: Interesting!

I think we should leverage these ENV vars here: https://github.com/rubygems/rubygems/blob/7de840c32e4a98f885613e6953b1a6eada6f48a3/bundler/lib/bundler/fetcher.rb#L223-L236 so that we can improve a CI env detection at https://ecosystem.rubytogether.org!

deivid-rodriguez (commented):
Awesome improvement! And great careful user-focused approach 😍

lib/rubygems/update_suggestion.rb
simi: I can somehow merge that list. I was thinking to keep this functionality in RubyGems itself if possible and from Bundler just check if this functionality is present and reuse as much as possible. But I'm afraid I'll need to keep this list in Bundler only for the original purpose just to not mess with metrics.

deivid-rodriguez: I think it makes sense to merge both lists and keep it in RubyGems, although as usual, if we want Bundler accuracy when reporting this to be the same across RubyGems versions, then we'll need to backport it to Bundler and keep the list duplicated in Bundler for a while.

lib/rubygems/update_suggestion.rb
piegamesde: Hi, as the person who brought `XDG_STATE_HOME` into the spec I'd say that folder is the best fit for your file. However, it really is a close call and it putting it into `XDG_DATA_HOME` would be equally fine. So I'd say just pick one of them without overthinking it too much, there is no "wrong" here.

simi: Thanks for the suggestion @piegamesde. I had nice chat also at xdg mail list (https://lists.freedesktop.org/archives/xdg/2022-September/014627.html, that's probably where we got your attention) and I decided to go with `XDG_STATE_HOME` here. :muscle:

deivid-rodriguez: I also agree that `XDG_STATE_HOME` is the best fit!

deivid-rodriguez:
@simi Is this still a WIP or is it ready for review?

simi:
> @simi Is this still a WIP or is it ready for review?

It is awaiting `XDG_STATE_HOME` change, I'll do my best to finish it this week. Anyway I think new minor version bump will be needed. If I remember well, there is also other case (Windows platform deprecation) awaiting minor version bump.

deivid-rodriguez:
Thanks, that's what I thought but I wanted to make sure :)

lib/rubygems/update_suggestion.rb
simi: Updated to use `$XDG_STATE_HOME/gem/last_update_check` file with timestamp.

indirect (approved):
👍🏻 🙌🏻

lib/rubygems/update_suggestion.rb
simi: I have not tackled this yet, my plan is to provide next PR to make same "call to update" feature for `bundle install`. I'll try to share as much as possible and consolidate those 2 CI env lists. For example looking randomly at first 3 entries, they all support `CI` env var as well. We can probably compact this list. Thinking about this again, I think I can do that in separated PR actually.

https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
https://circleci.com/docs/built-in-environment-variables/
https://docs.semaphoreci.com/ci-cd-environment/environment-variables/

deivid-rodriguez: Sounds good! I created https://github.com/rubygems/rubygems/issues/6038 to keep track of that.

test/rubygems/test_gem_update_suggestion.rb
```diff
@@ -0,0 +1,137 @@
...
+  end
+
+  def test_eglible_for_update_on_ci
+    with_eglible_environment(ci: true, cmd: @cmd) do
+      refute @cmd.eglible_for_update?
+    end
+  end
+
+  def test_eglible_for_update_unwrittable_config
+    with_eglible_environment(ci: true, cmd: @cmd) do
```
deivid-rodriguez, line 124: Here we're not actually testing the code to detect unwritability, because we will early exit when checking the ci case due to `ci: true` above, correct? I guess that should be removed and probably some extra assertion about `state_file_writable?` actually being called?

lib/rubygems/update_suggestion.rb
```diff
@@ -0,0 +1,68 @@
...
+    # load time of last check, ensure the difference is enough to repeat the suggestion
+    check_time = Time.now.to_i
+    last_update_check = Gem.configuration.last_update_check
+    return false if (check_time - last_update_check) < ONE_WEEK
+
+    # compare current and latest version, this is the part where
+    # latest rubygems spec is fetched from remote
+    if (Gem.rubygems_version < Gem.latest_rubygems_version)
+      # store the time of last successful check into state file
+      Gem.configuration.last_update_check = check_time
```
deivid-rodriguez, line 58: I think there's a small divergence between naming and logic here?

If I understand correctly, this is not checking the last time we checked whether RubyGems is out of date, but the last time we check whether RubyGems is out of date, and it actually ended up being out of date. I think we should either:

* Change the configuration name to something that makes this more clear.
* Actually check this just once a week, regardless of the result of the check.

I think I'd go with the latter? If the user is using `gem install` a lot of times, and has an up to date RubyGems, it seems enough to only go through the overhead of checking for updates once a week?

deivid-rodriguez (commented):
Just a couple of comments. Also I wonder whether this "call for update" message would make sense in other commands other than `gem install`. But regardless of the answer to that, `gem install` seems like a good place to start!

This is awesome by the way!

lib/rubygems/update_suggestion.rb
simi: Ahh, you're right. I remember the older code was working in other way. This was missed in latest refactoring. I'll restore old behaviour and add test to ensure it works this way.

test/rubygems/test_gem_update_suggestion.rb
simi: well spotted, fixed

lib/rubygems/update_suggestion.rb
simi: fixed now, test `test_eglible_for_update_is_not_annoying_when_new_version_is_released` will fail now if update timestamp is stored only on successful check (old behaviour)

test/rubygems/test_gem_update_suggestion.rb
deivid-rodriguez: Nice. I was also suggesting somehow switching to mocks, not just stubs. If I'm understanding correct, if we have a mechanism that patches methods, like `Gem.configuration.stub :state_file_writable?, false`, but also makes sure the methods are actually called, then this test issue would not have been introduced in the first place because the original version was not calling `Gem.configuration.state_file_writable?` at all.

In any case, totally out of scope of this PR!

lib/rubygems/update_suggestion.rb
deivid-rodriguez: Nice! Particularly impressed by the readibility and quality of your tests, they feel like reading documentation 👏 👏 🙌

deivid-rodriguez (approved):
This looks good to me and I think it's an impressive piece of work! 💪

How should we ship this? Standard patch release, or wait for new years?

I'd go with shipping it right now with the next patch, but happy to do otherwise! End of year is so close anyways!

test/rubygems/test_gem_update_suggestion.rb
simi: I'm not sure about the historical reasons for rolling out custom simple stub library, but since we control test dependencies using Gemfile now, maybe we can sneak in some mock library. I can explore this if welcomed. Indeed out of this PRs scope.

simi:
Thanks for the kind words.

I was looking around and it seems similar changes were released in other projects on minor release. I would prefer the same, but I'm not sure if we can just upgrade minor version simply at any time. If that's problem, I think we can roll it out actually with next patch version.

The scope is super limited and it would be great to get some community testing before new Ruby release, since if I understand it well, that's how majority of RubyGems installations is upgraded (when upgrading Ruby version). I do aim to prepare bundler support for same feature as well before Ruby 3.2 release. :pray:

test/rubygems/test_gem_update_suggestion.rb
deivid-rodriguez: I think it was because of wanting to get rid of MiniTest and TestUnit not providing exact replacements, so we rolled our own. Feel free to explore using a mock library, sure!

deivid-rodriguez:
Let's ship this with RubyGems 3.4 then. I expect to release than at the beginning of December, so we have about a month to test before final Ruby release. Merging this now!

simi:
:thinking: @enebo @eregon is it ok to recommend update of RubyGems for JRuby and TruffleRuby? We can scope this also to CRuby only for now.

eregon:
Thank you for asking.
I think it's not good at least for TruffleRuby because TruffleRuby no-ops on `gem update --system` because it generally causes more problems than it solves and it's very slow (there is `--force-update` to force it).
So it's going to be confusing, and I think better not have it on TruffleRuby.

https://github.com/oracle/truffleruby/commit/859069ef94951042916ebf8cda6c807b9bb81b7f is an example of a change I needed to make to RubyGems, and that of course gets lost if Rubygems gets updated.
Of course I plan to contribute this patch here, or change TruffleRuby launchers to no longer need the patch (a clear loss in functionality so probably not), but there will likely always be a gap in between.

---

IMHO any such message is just annoying no matter for which Ruby, but that's my personal opinion.
Kind of similar to the super annoying bug where bundle would constantly updated the BUNDLED WITH section years ago.
Also it would be terribly annoying if any stdlib/default gem/other gem you use would warn if you're not using the latest version.
But I guess I won't convince you about that anyway.

---

BTW this message should also not be shown if where RubyGems is installable is not writable. For instances this message is not actionable for a Ruby installed by apt/dnf under /usr (isn't it?). Unless that's already handled?

simi:
@eregon I'll update to print out message only on CRuby for now.

> IMHO any such message is just annoying no matter for which Ruby, but that's my personal opinion.
Kind of similar to the super annoying bug where bundle would constantly updated the BUNDLED WITH section years ago.
Also it would be terribly annoying if any stdlib/default gem/other gem you use would warn if you're not using the latest version.
But I guess I won't convince you about that anyway.

We're doing our best to keep it as friendly as possible. There is no impact on the `Gemfile.lock` for now. Since even you update RubyGems, thanks to auto-switch in Bundler it is going to use locked bundler for resolving.

> BTW this message should also not be shown if where RubyGems is installable is not writable. For instances this message is not actionable for a Ruby installed by apt/dnf under /usr (isn't it?). Unless that's already handled?

It is recommended to prevent any `gem update --system` using `Gem.disable_system_update_message` (already implemented in Debian for [example](https://salsa.debian.org/ruby-team/rubygems-integration/-/blob/master/lib/rubygems/defaults/operating_system.rb#L7-14)). It respects this option and doesn't show any message in this kind of installation. Same for CIs, non-tty envs, ... Feel free to check initial description of PR for all cases this is skipped. And in the end, the message pops just once per week now and could be completely disabled by gemrc entry or ENV var.

If you think 1 week is still annoying, we can double that. I would be ok with that.

We're following same pattern as similar CLI tools do already for years. Personally I had no problem with npm/terraform/gh cli printing out those messages from time to time. It is good notification of thinking about the upgrade (not only locally, but also on project side).

eregon:
> It is recommended to prevent any `gem update --system` using `Gem.disable_system_update_message`

Nice, I think I can use this in `rubygems/defaults/truffleruby.rb` so then there is no need to change anything in RubyGems.
I find the Debian message interesting, they seem to be of the opinion of "don't update when there is no need" which I agree with. IMHO frequently updating to latest makes the development environment less stable, which is why I don't like these messages.

> If you think 1 week is still annoying, we can double that. I would be ok with that.

I think no need to change, and probably would be good to have opinions from more people. I know I'm rather biased on this matter.

simi:
> Nice, I think I can use this in rubygems/defaults/truffleruby.rb so then there is no need to change anything in RubyGems.
I find the Debian message interesting, they seem to be of the opinion of "don't update when there is no need" which I agree with. IMHO frequently updating to latest makes the development environment less stable, which is why I don't like these messages.

OK, I'm going to keep it as is.

simi:
> ...is it ok to recommend update of RubyGems for JRuby? We can scope this also to CRuby only for now.

/cc @headius

enebo:
@simi I somewhat agree with @eregon opinion on these messages and that I personally would only want to see a message like this if it fixed something I should-not/cannot ignore like a security update.  If I am using an older version of something I typically have a reason for it.  

Rubygems release stride is also pretty quick so I think people will be seeing this message once every couple of weeks (RG averages about 2 releases a month) from just new releases.  If you change the re-notification period to once every two weeks the odds are higher they will only see this once per RG udpate. 

If we do not want to see these by default on JRuby we can just put ```Gem.disable_system_update_message``` in our defaults file right?  If so then we will figure out which way we want to go with presenting this warning after it is released.

schneems:
👋🏻 hi. As the main maintainer for the Ruby Buildpack for Heroku I care about RubyGems quite a bit. Our policy (https://devcenter.heroku.com/articles/ruby-support) is currently:

> Heroku supports the following Ruby versions and the associated Rubygems. A supported version means that you can expect our tools and platform to work with a given version. It also means you can receive technical support. Here are our supported Ruby versions:

Basically we never re-install the rubygem. If someone wants a newer rubygems version they need to upgrade to a newer Ruby. This is great for consistency but is somewhat annoying in cases where there are large updates on RubyGems but there's no associated release of MRI. Also the version support between RubyGems and MRI somewhat differs.

With the addition of this prompt I can see a few paths forward:

- Do nothing: Always an option
- Disable the warning via env var: Same behavior as today
- Upgrade all RubyGems versions to the latest: Seems bad, hard to guarantee consistency
- Introduce a `gem update --not-greedy` flag or something that would only bump patch level
- Let the customer define a rubygems version.

For this last option "Let the customer define a rubygems version." ideally it would be pinned to their local development environment in the same way that `RUBY VERSION` is in the `Gemfile.lock` and isn't yet another config that needs to be updated that is vendor specific that everyone will forget about until it breaks.

I think my two favorites would be `gem update --not-greedy` (naming TBD) or to introduce the specific RubyGems version into the Gemfile.lock (or maybe both).

kaporn1990:
sudo gem install cocoapods

simi:
@schneems can you explain how this affects Ruby Buildpack for Heroku? Message is skipped on non-tty environments and by doing local upgrade there is no change to any of the files (including `Gemfile.lock`). Bundler currently supports auto-switch/install, and even you upgrade RubyGems or Bundler, `bundle` commands are going to automatically use locked `Bundler`. Personally I don't see any reason to not use latest RubyGems with non-EOL CRuby.

Apart of that, `gem` command on its own is not directly used by users for installing gems often, since most of the times `bundle install` takes over. That leads me to other change we can do, make update banner into `gem push` command instead of `gem install`. That's independent command and it should be best practice to use latest RubyGems to push new gem if possible.

simi:
> sudo gem install cocoapods

?

simi:
@enebo By setting `Gem.disable_system_update_message` you'll get any `gem update --system` call blocked as well. When set, it doesn't make any sense to promote upgrade. I'm not sure that's what you're looking for.

schneems:
I missed that it's only for non-tty environments. I guess i'm using this as an opportunity to talk about what our support SHOULD be as i'm re-writing the Ruby Buildpack in Rust and nows the time to make changes. 

One big change we're making is relying on the bundler version in the Gemfile.lock instead of stripping it out.

https://github.com/heroku/buildpacks-ruby/tree/schneems/fresh-start-rust#known-differences-against-herokuheroku-buildpack-ruby

Feel free to disregard the rest, as it's mostly for your information. I'm mostly latching on to this conversation to have a related but different one. 

> Bundler currently supports auto-switch/install, and even you upgrade RubyGems or Bundler, bundle commands are going to automatically use locked Bundler. Personally I don't see any reason to not use latest RubyGems with non-EOL CRuby.

The only guarantee of behavior is specific bundler, ruby, and rubygems version. Right now we lock down on Ruby, in my upcoming buildpack, we'll lock to the customer's bundler version, that just leaves RubyGems as the odd duck out. When Bundler 2 came out there were TONS of problems with it's RubyGems and Ruby version support. It basically caused a 3 dimensional matrix of possible problem cases and we had no capacity to mediate the problem because we have no way of decoupling Ruby version from RubyGems version (at the moment).

> Personally I don't see any reason to not use latest RubyGems with non-EOL CRuby.

We don't [provide] support for EOL Ruby versions, but we don't remove them either. The oldest version of Ruby on the Heroku platform right now is 2.4.10. So whatever we put in the buildpack needs to play nice with 2.4.10 or we need to otherwise have code that says "update rubygems version".

Generally speaking there's never a case where customers won't need a specific version. RubyGems and Bundler and Ruby all have bugs in them, especially new releases. I try to be wary that (most) blanket decisions I make for my customers can be opted out of in some way. So if I move forward with auto upgrading everyone to the latest, I would still want some way to opt out of that behavior and I would want that to be supported by community tooling.

enebo:
@simi Yeah.  I misread that.  I guess we will just ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"] if we decide not to provide this with the rubygems we ship with.   Looking at the PR there are other ways too.  Whatever we do we want to be able to make someone be able to opt into this if they want it (if we decide to disable it by default).  Thanks for the ping on this.

simi:
@schneems 

> I missed that it's only for non-tty environments.

It is actually only for TTY environments if I understand it well.

> I guess i'm using this as an opportunity to talk about what our support SHOULD be as i'm re-writing the Ruby Buildpack in Rust and nows the time to make changes.

Feel free to ping us on Slack to get in sync about your expectations and your plans.

> One big change we're making is relying on the bundler version in the Gemfile.lock instead of stripping it out.

That's seems a good step. If I understand it well, the challenge currently is to detect proper Ruby. Once detected, you can install latest bundler (or keep the Ruby default if already supporting auto-install/switch feature) and let the bundler do its best to manage Gemfile.lock with proper version itself.

simi:
> @simi Yeah. I misread that. I guess we will just ENV["RUBYGEMS_PREVENT_UPDATE_SUGGESTION"] if we decide not to provide this with the rubygems we ship with. Looking at the PR there are other ways too. Whatever we do we want to be able to make someone be able to opt into this if they want it (if we decide to disable it by default). Thanks for the ping on this.

@enebo If welcomed, I can add also special config for this so there is no need to manipulate ENV just because of this.

enebo:
@simi That's ok.  Sorry I meant to reply to this a few days ago and it got lost in the tabs

PR #5923, opened 2022-09-12
CI - Windows JRuby - use pre-installed JDK 17

MSP-Greg:
Some Windows Actions jobs are using `setup-java` to install Java.  Actions Windows images typically have the most recent 'LTS' version installed.  Adjust `JAVA_HOME` to use it, and remove `setup-java`.

Note that you must abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md) to contribute to this project.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -101,17 +101,13 @@ jobs:
           - { name: jruby-9.3, value: jruby-9.3.4.0 }
     steps:
       - uses: actions/checkout@v3
+      # set JDK version, see https://github.com/actions/runner-images
+      - run: echo JAVA_HOME=$env:JAVA_HOME_17_X64 >> $env:GITHUB_ENV
       - name: Setup ruby
         uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby.value }}
           bundler: none
-      - name: Setup java
-        uses: actions/setup-java@v3.4.1
-        with:
-          distribution: temurin
-          java-version: 18.0.0
-        if: startsWith(matrix.ruby.name, 'jruby')
       - name: Install rubygems
         run: ruby setup.rb
         shell: bash
--- a/.github/workflows/windows-bundler.yml
+++ b/.github/workflows/windows-bundler.yml
@@ -32,17 +32,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
+      # set JDK version, see https://github.com/actions/runner-images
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Setting up Java is fast, and not causing any issues, and the reason is fully locked is that changes in Java caused our CI to break in the past.

So I don't understand why we need to downgrade and unpin this dependency.

Can you explain the motivation @MSP-Greg? Thanks!

MSP-Greg:
> Can you explain the motivation

Well, another dependabot PR in my fork - https://github.com/MSP-Greg/rubygems/pull/11 and the email.

All three of the 'latest' images on Actions have Java 8,11, & 17 installed (see https://github.com/actions/runner-images).

17 is the most recent 'LTS' version (see https://en.wikipedia.org/wiki/Java_version_history).  I suspect that the most recent 'LTS' version is the most common 'recent' version in the wild.  The Actions images would seem to support that.  And the Actions images are similar (identical?) to Azure, which is probably the most used Windows cloud.

If one wants to run tests locally, installing Java 18 seems a PITA.

I am not a regular user of JRuby, but I have both Windows & Ubuntu versions installed.  I don't have a strong opinion on this, so close if you prefer.

deivid-rodriguez:
> Well, another dependabot PR in my fork - https://github.com/MSP-Greg/rubygems/pull/11 and the email.

That's bad, that's a Dependabot issue I believe where forks don't get dependabot disabled by default, hopefully will be fixed soon. Can you disable dependabot manually?

I don't mind which version of Java we use but I'd like to keep it locked to avoid issues like the one that led me to locking the dependency. Using the default Github Actions image version is not enough since that will change when GitHub rolls out new images. That's in fact how CI broke last time.

> I don't have a strong opinion on this, so close if you prefer.

Ok, yes, I'll close it!

MSP-Greg:
> Can you disable dependabot manually?

Not sure.  It's on my TODO, I think at one point one couldn't.  New code writing tasks keep jumping ahead of it.  I'm old, so I have a very low tolerance re email noise.

> I'd like to keep it locked to avoid issues

I did look at the history, but I didn't see if the issues were the test code or the actual lib code.

I often work with master/main branches/builds (my default Rubies are master).  Re locking, 'production' is one thing, but doing so in CI often leads to a lot of changes whenever one updates them.  Leaving them 'loose' often has a slower need for changes.  What kind of rain do you like, a short torrential blast or a longer steady drizzle?

Re closing, no problem.

PR #5939, opened 2022-09-19
Fix `Gem::Platform.match` not handling String argument properly

### LINKED ISSUE #5938: Gem::Platform.match no longer behaves properly with a String argument

When passing a string to `Gem::Platform.match`, the behavior of the method is the same as before 1b9f7f5 introduced a fix for musl platforms.

Although `Gem::Platform#=~` properly coerces the `other` argument, the `#match` method, when passed a `String`, will instead invoke `String#=~`.

The interesting bit can be seen in the docstring for `String#=~` (from https://ruby-doc.org/core-3.1.2/String.html#method-i-3D~)

> If the given object is not a Regexp, returns the value returned by object =~ self.

thereby avoiding the intended behavior change of 1b9f7f5

Present in 3.3.22

On a `x86_64-linux-musl` system (e.g., `ruby:3.1-alpine` image):

```ruby
#! /usr/bin/env ruby

Gem::VERSION # => "3.3.22"
Gem::Platform.local # => #<Gem::Platform:0x00007fda5ca0e0b8 @cpu="x86_64", @os="linux", @version="musl">

gem_platform_string = "x86_64-linux"
gem_platform = Gem::Platform.new(gem_platform_string)

Gem::Platform.match(gem_platform_string) # => false
Gem::Platform.match(gem_platform) # => true
```

```
RubyGems Environment:
  - RUBYGEMS VERSION: 3.3.22
  - RUBY VERSION: 3.1.2 (2022-04-12 patchlevel 20) [x86_64-linux-musl]
  - INSTALLATION DIRECTORY: /usr/local/bundle
  - USER INSTALLATION DIRECTORY: /root/.local/share/gem/ruby/3.1.0
  - RUBY EXECUTABLE: /usr/local/bin/ruby
  - GIT EXECUTABLE: 
  - EXECUTABLE DIRECTORY: /usr/local/bundle/bin
  - SPEC CACHE DIRECTORY: /root/.local/share/gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /usr/local/etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - x86_64-linux-musl
  - GEM PATHS:
     - /usr/local/bundle
     - /root/.local/share/gem/ruby/3.1.0
     - /usr/local/lib/ruby/gems/3.1.0
  - GEM CONFIGURATION:
     - :update_sources => true
     - :verbose => true
     - :backtrace => true
     - :bulk_threshold => 1000
     - "install" => "--no-document"
     - "update" => "--no-document"
  - REMOTE SOURCES:
     - https://rubygems.org/
  - SHELL PATH:
     - /usr/local/bundle/bin
     - /usr/local/sbin
     - /usr/local/bin
     - /usr/sbin
     - /usr/bin
     - /sbin
     - /bin
```

### Proposed fix

I think the right thing to do here is to coerce the argument to a `Gem::Platform` object, so something like:

```patch
diff --git a/lib/rubygems/platform.rb b/lib/rubygems/platform.rb
index 06de5de..1dacc59 100644
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -22,6 +22,7 @@ def self.match(platform)
   end
 
   def self.match_platforms?(platform, platforms)
+    platform = Gem::Platform.new(platform) unless platform.is_a?(Gem::Platform)
     platforms.any? do |local_platform|
       platform.nil? ||
         local_platform == platform ||
```

I couldn't figure out how to write a nice test for this, but hopefully I've provided enough detail?

flavorjones:
Previously 9eead86 introduced non-commutativity of platforms, and later commit 1b9f7f50 changed the behavior of `Gem::Platform.match` to ensure the callee of `#=~` was the gem platform.

However, when the platform argument is a String, then the callee and argument of `#=~` are flipped (see docs for `String#=~`), which works against the fix from 1b9f7f50.

Closes #5938

See #5938 

Coerce the `Gem::Platform.match` argument to a `Gem::Platform` object

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -22,6 +22,7 @@ class Gem::Platform
   end
 
   def self.match_platforms?(platform, platforms)
+    platform = Gem::Platform.new(platform) unless platform.is_a?(Gem::Platform)
     platforms.any? do |local_platform|
       platform.nil? ||
         local_platform == platform ||
--- a/test/rubygems/test_gem_platform.rb
+++ b/test/rubygems/test_gem_platform.rb
@@ -452,6 +452,13 @@ class TestGemPlatform < Gem::TestCase
     assert_equal 1, result.scan(/@version=/).size
   end
 
+  def test_gem_platform_match_with_string_argument
+    util_set_arch "x86_64-linux-musl"
+
+    assert(Gem::Platform.match(Gem::Platform.new("x86_64-linux")), "should match Gem::Platform")
+    assert(Gem::Platform.match("x86_64-linux"), "should match String platform")
+  end
+
   def assert_local_match(name)
     assert_match Gem::Platform.local, name
   end


deivid-rodriguez (approved):
Looks good, thanks @flavorjones!

PR #5947, opened 2022-09-21
Fix `bundle init` not respecting umask in generated gem's Gemfile

heyvito:

On a clean Ruby 3 installation (through Homebrew), I noticed that `Gemfile`s created through `bundler init` had different permission than the one provided by umask.
This was due to Bundler copying the template from the source directory (which has a mode `0444`) instead of piping the template's contents into the destination Gemfile:

From my installation:

```
$ pwd
/opt/homebrew/opt/ruby/lib/ruby/3.1.0/bundler/templates

$ stat Gemfile
16777234 11017035 -r--r--r-- 1 victorgama admin 0 76 "May 19 22:51:10 2022" "Apr 12 08:11:17 2022" "May 19 22:51:12 2022" "Apr 12 08:11:17 2022" 4096 8 0 Gemfile
```

Searching this repo's issues/pull requests I was unable to find anything discussing this specific behaviour.

Instead of leveraging `FileUtils#cp`, the fix leverages `File#open` on both the template and destination file, and `IO#copy_stream` to respect the current umask.

> Notice: Maintainers, I couldn't come up with a spec for this specific scenario.

### DIFF
--- a/bundler/spec/commands/init_spec.rb
+++ b/bundler/spec/commands/init_spec.rb
@@ -7,6 +7,29 @@ RSpec.describe "bundle init" do
     expect(bundled_app_gemfile).to be_file
   end
 
+  context "with a template with permission flags not matching current process umask" do
+    let(:template_file) do
+      gemfile = Bundler.preferred_gemfile_name
+      templates_dir.join(gemfile)
+    end
+
+    let(:target_dir) { bundled_app("init_permissions_test") }
+
+    around do |example|
+      old_chmod = File.stat(template_file).mode
+      FileUtils.chmod(old_chmod | 0o111, template_file) # chmod +x
+      example.run
+      FileUtils.chmod(old_chmod, template_file)
+    end
+
+    it "honours the current process umask when generating from a template" do
+      FileUtils.mkdir(target_dir)
+      bundle :init, :dir => target_dir
+      generated_mode = File.stat(File.join(target_dir, "Gemfile")).mode & 0o111
+      expect(generated_mode).to be_zero
+    end
+  end
+
   context "when a Gemfile already exists" do
     before do
       create_file "Gemfile", <<-G
--- a/bundler/lib/bundler/cli/init.rb
+++ b/bundler/lib/bundler/cli/init.rb
@@ -32,7 +32,11 @@ module Bundler
           file << spec.to_gemfile
         end
       else
-        FileUtils.cp(File.expand_path("../templates/#{gemfile}", __dir__), gemfile)
+        File.open(File.expand_path("../templates/#{gemfile}", __dir__), "r") do |template|
+          File.open(gemfile, "wb") do |destination|
+            IO.copy_stream(template, destination)
+          end
+        end
       end
 
       puts "Writing new #{gemfile} to #{SharedHelpers.pwd}/#{gemfile}"
--- a/bundler/spec/support/path.rb
+++ b/bundler/spec/support/path.rb
@@ -312,6 +312,10 @@ module Spec
       source_root.join("tool/bundler")
     end
 
+    def templates_dir
+      lib_dir.join("bundler", "templates")
+    end
+
     extend self
   end
 end


hsbt:
@heyvito Thanks. Can you add the additional test for this?

heyvito:
@hsbt I'd love to! Would you be so kind as to help me with that?

See, I thought about generating a Gemfile template without the write bit set, and asserting its presence on the generated Gemfile.
However, there's little room to do that, since intercepting `File#expand_path` to return the new path wouldn't work, so I'm not really sure what to do in this case (and how you folks usually handle those cases in this codebase). Any advice?

heyvito:
Considered introducing an environment variable to replace that `File#expand_path` whenever it is set, but I usually avoid changing code just to support tests.

The second approach I though about using is to have a spec change the default template's permissions to add execute bits, checking whether the Gemfile created by `init` does not have those extra bits, and finally reverting the template's mode. This option wouldn't require further changes to the CLI:

```ruby
  context "with a template with permission flags not matching current process umask" do
    let(:template_file) do
      gemfile = Bundler.preferred_gemfile_name
      File.expand_path("../../lib/bundler/templates/#{gemfile}", __dir__)
    end

    let(:target_dir) { bundled_app("init_permissions_test") }

    before do
      @old_chmod = File.stat(template_file).mode
      FileUtils.chmod(@old_chmod | 0o111, template_file) # chmod +x
    end

    after do
      FileUtils.chmod(@old_chmod, template_file)
      FileUtils.rm_rf(target_dir)
    end

    it "honours the current process umask when generating from a template" do
      FileUtils.mkdir(target_dir)
      bundle :init, :dir => target_dir
      generated_mode = File.stat(File.join(target_dir, "Gemfile")).mode & 0o111
      expect(generated_mode).to be_zero
    end
  end
```

deivid-rodriguez:
Your idea seems fine in principle, did you check that the above spec actually catches the issue though? I _think_ specs by default uses a copy of Bundler installed at `tmp` and not the actual source files.

heyvito:
Hi there, @deivid-rodriguez! No, I didn't run those specs, just kinda drafted them to check with you folks whether it made sense.

I'll implement those and see how they behave, but you're right: specs run against a copy under `tmp`! Thank you so much for pointing this out! Definitely saved me some time! 💖

heyvito:
I apologise for the double-post, but just to let you know that I just added specs (with minimal changes, thanks @deivid-rodriguez for the heads up!), and everything seems to be fine. ✨

I'll just rebase master against this branch, and wait for CI results.

bundler/spec/commands/init_spec.rb
```diff
@@ -7,6 +7,32 @@
...
+    let(:target_dir) { bundled_app("init_permissions_test") }
+
+    before do
+      @old_chmod = File.stat(template_file).mode
+      FileUtils.chmod(@old_chmod | 0o111, template_file) # chmod +x
+    end
+
+    after do
+      FileUtils.chmod(@old_chmod, template_file)
+      FileUtils.rm_rf(target_dir)
```
deivid-rodriguez, line 25: All these directories are removed automatically by our global teardown, so we generally don't do this particular specific cleanup.

Also, I would instead use an `around` hook, so that you don't need to keep the previous mode around in an instance variable.

deivid-rodriguez (commented):
A couple of small comments in the spec, but looks good. Also nice that we're now dropping the `fileutils` dependency from the `init` command.

bundler/spec/commands/init_spec.rb
heyvito: Hey @deivid-rodriguez! Sorry the delay for address this! Busy week.
I'll remove the cleanup, and replace `before`/`after` hooks with an `around`; completely forgot about it.

I'll force-push and mark this as resolved right after. Thank you for reviewing, and for the suggestions!

heyvito:
My pleasure! I'll try to keep tabs on issues and try to contribute more in the future. 🌻
I appreciate the attention and kindness, Deivid! Have a great one. :)

PR #5960, opened 2022-09-30
Migrate our resolver engine to PubGrub

### LINKED ISSUE #3296: Small improvement to dependency resolution conflict messages
#### Feature Request  Summary

An improvement to error messages for dependency resolution with an extra line like "In particular ..." (explained below).

#### Why?
When upgrading gems, I ran into incompatibility message like this one:

```
Bundler could not find compatible versions for gem "railties":

[followed by "depends on" messages]
```

The dependency conflict messages require manual review, which is pretty painful to me.  If I make a mistake parsing the error messages, then it wastes even more time.  The dependency error messages prints all of the dependency specs for all the gem requirement (apologies if I've butchered nomenclature here); however, they could call out the dependencies that actually conflict.

### Example

I started with this Gemfile that resolves with Bundler 2.0.0.

<details>
<summary>Gemfile</summary>
<div>
<pre>
<code>
source 'https://rubygems.org'
ruby '2.4.5'

gem 'rails', '4.2.10'

gem 'devise', '\~> 4.0.3'
gem 'jquery-rails', '\~> 4.0.5'
</code>
</pre>
</div>
</details>

<details>
<summary>Gemfile.lock</summary>
<pre>
<code>
GEM
  remote: https://rubygems.org/
  specs:
    actionmailer (4.2.10)
      actionpack (= 4.2.10)
      actionview (= 4.2.10)
      activejob (= 4.2.10)
      mail (~> 2.5, >= 2.5.4)
      rails-dom-testing (~> 1.0, >= 1.0.5)
    actionpack (4.2.10)
      actionview (= 4.2.10)
      activesupport (= 4.2.10)
      rack (~> 1.6)
      rack-test (~> 0.6.2)
      rails-dom-testing (~> 1.0, >= 1.0.5)
      rails-html-sanitizer (~> 1.0, >= 1.0.2)
    actionview (4.2.10)
      activesupport (= 4.2.10)
      builder (~> 3.1)
      erubis (~> 2.7.0)
      rails-dom-testing (~> 1.0, >= 1.0.5)
      rails-html-sanitizer (~> 1.0, >= 1.0.3)
    activejob (4.2.10)
      activesupport (= 4.2.10)
      globalid (>= 0.3.0)
    activemodel (4.2.10)
      activesupport (= 4.2.10)
      builder (~> 3.1)
    activerecord (4.2.10)
      activemodel (= 4.2.10)
      activesupport (= 4.2.10)
      arel (~> 6.0)
    activesupport (4.2.10)
      i18n (~> 0.7)
      minitest (~> 5.1)
      thread_safe (~> 0.3, >= 0.3.4)
      tzinfo (~> 1.1)
    arel (6.0.4)
    bcrypt (3.1.12)
    builder (3.2.3)
    concurrent-ruby (1.1.4)
    crass (1.0.4)
    devise (4.0.3)
      bcrypt (~> 3.0)
      orm_adapter (~> 0.1)
      railties (>= 4.1.0, < 5.1)
      responders
      warden (~> 1.2.3)
    erubis (2.7.0)
    globalid (0.4.2)
      activesupport (>= 4.2.0)
    i18n (0.9.5)
      concurrent-ruby (~> 1.0)
    jquery-rails (4.0.5)
      rails-dom-testing (~> 1.0)
      railties (>= 4.2.0)
      thor (>= 0.14, < 2.0)
    loofah (2.2.3)
      crass (~> 1.0.2)
      nokogiri (>= 1.5.9)
    mail (2.7.1)
      mini_mime (>= 0.1.1)
    mini_mime (1.0.1)
    mini_portile2 (2.4.0)
    minitest (5.11.3)
    nokogiri (1.10.1)
      mini_portile2 (~> 2.4.0)
    orm_adapter (0.5.0)
    rack (1.6.11)
    rack-test (0.6.3)
      rack (>= 1.0)
    rails (4.2.10)
      actionmailer (= 4.2.10)
      actionpack (= 4.2.10)
      actionview (= 4.2.10)
      activejob (= 4.2.10)
      activemodel (= 4.2.10)
      activerecord (= 4.2.10)
      activesupport (= 4.2.10)
      bundler (>= 1.3.0, < 2.0)
      railties (= 4.2.10)
      sprockets-rails
    rails-deprecated_sanitizer (1.0.3)
      activesupport (>= 4.2.0.alpha)
    rails-dom-testing (1.0.9)
      activesupport (>= 4.2.0, < 5.0)
      nokogiri (~> 1.6)
      rails-deprecated_sanitizer (>= 1.0.1)
    rails-html-sanitizer (1.0.4)
      loofah (~> 2.2, >= 2.2.2)
    railties (4.2.10)
      actionpack (= 4.2.10)
      activesupport (= 4.2.10)
      rake (>= 0.8.7)
      thor (>= 0.18.1, < 2.0)
    rake (12.3.2)
    responders (2.4.0)
      actionpack (>= 4.2.0, < 5.3)
      railties (>= 4.2.0, < 5.3)
    sprockets (3.7.2)
      concurrent-ruby (~> 1.0)
      rack (> 1, < 3)
    sprockets-rails (3.2.1)
      actionpack (>= 4.0)
      activesupport (>= 4.0)
      sprockets (>= 3.0.0)
    thor (0.20.3)
    thread_safe (0.3.6)
    tzinfo (1.2.5)
      thread_safe (~> 0.1)
    warden (1.2.7)
      rack (>= 1.0)

PLATFORMS
  ruby

DEPENDENCIES
  devise (~> 4.0.3)
  jquery-rails (~> 4.0.5)
  rails (= 4.2.10)

RUBY VERSION
   ruby 2.4.5p335

BUNDLED WITH
   2.0.0
</pre>
</code>
</details>

Now, I want to update from Rails 4.2.10 to Rails 5.1.7.

<details>
<summary>Gemfile</summary>
<pre>
<code>
source 'https://rubygems.org'
ruby '2.4.5'

gem 'rails', '5.1.7'

gem 'devise', '\~> 4.0.3'
gem 'jquery-rails', '\~> 4.0.5'

</code>
</pre>
</details>

<details>
<summary>$ bundle install</summary>
<pre>
<code>
Resolving dependencies...
Bundler could not find compatible versions for gem "railties":
  In snapshot (Gemfile.lock):
    railties (= 4.2.10)

  In Gemfile:
    devise (~> 4.0.3) was resolved to 4.0.3, which depends on
      railties (>= 4.1.0, < 5.1)

    jquery-rails (~> 4.0.5) was resolved to 4.0.5, which depends on
      railties (>= 4.2.0)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      railties (= 5.1.7)

    devise (~> 4.0.3) was resolved to 4.0.3, which depends on
      responders was resolved to 2.4.0, which depends on
        railties (>= 4.2.0, < 5.3)

</code>
</pre>
</details>

When I first encountered this, I didn't know how to interpret these messages, but I followed the advice here to run `bundle update`.

<details>
<summary>$ bundle update</summary>
<pre>
<code>
Resolving dependencies...
Bundler could not find compatible versions for gem "railties":
  In Gemfile:
    devise (~> 4.0.3) was resolved to 4.0.3, which depends on
      railties (>= 4.1.0, < 5.1)

    jquery-rails (~> 4.0.5) was resolved to 4.0.5, which depends on
      railties (>= 4.2.0)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      railties (= 5.1.7)
</code>
</pre>
</details>

However, I still see compatibility errors.

If I'm patient and observant, I know that devise is limiting the ability to upgrade Rails, so I bump Devsie to ~> 4.3.0.

<details>
<summary>$ bundle update</summary>
<pre>
<code>
Resolving dependencies...
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      actionpack (= 5.1.7)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      sprockets-rails (>= 2.0.0) was resolved to 3.2.1, which depends on
        actionpack (>= 4.0)

Bundler could not find compatible versions for gem "activesupport":
  In Gemfile:
    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      activejob (= 5.1.7) was resolved to 5.1.7, which depends on
        globalid (>= 0.3.6) was resolved to 0.4.2, which depends on
          activesupport (>= 4.2.0)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      activesupport (= 5.1.7)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      sprockets-rails (>= 2.0.0) was resolved to 3.2.1, which depends on
        activesupport (>= 4.0)

Bundler could not find compatible versions for gem "rails-dom-testing":
  In Gemfile:
    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      actionpack (= 5.1.7) was resolved to 5.1.7, which depends on
        rails-dom-testing (~> 2.0)

    jquery-rails (~> 4.0.5) was resolved to 4.0.5, which depends on
      rails-dom-testing (~> 1.0)

Bundler could not find compatible versions for gem "railties":
  In Gemfile:
    devise (~> 4.3.0) was resolved to 4.3.0, which depends on
      railties (>= 4.1.0, < 5.2)

    jquery-rails (~> 4.0.5) was resolved to 4.0.5, which depends on
      railties (>= 4.2.0)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      railties (= 5.1.7)
</code>
</pre>
</details>

Now, with more conflict messages, patience, and observance, I know that jquery-rails is the limiting gem here.

If I bump jquery-rails to ~> 4.1.0, then this will resolve.

The pain point is reviewing and understanding the conflict messages.

##### Improvemenet

I dug into the code and found an approach to adding some slightly more enlightening messages. The foundation for dependency resolution error messages is already present, but the messages can be extended with some conditionally printed statements to call out the resolution conflict.  If this is a welcome change, I can open a PR.

With the approach I found, the above error messages could include an extra line like the "In particular" statements here:

<details>
<summary>$ bundle update # (update the rails gem)</summary>
<pre>
<code>
Bundler could not find compatible versions for gem "railties":
  In snapshot (Gemfile.lock):
    railties (= 4.2.10)

  In Gemfile:
    devise (~> 4.0.3) was resolved to 4.0.3, which depends on
      railties (>= 4.1.0, < 5.1)

    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      railties (= 5.1.7)

      In particular, railties (= 5.1.7) is not satisfied by 4.2.10

</code>
</pre>
</details>

<details>
<summary>$ bundle update # (update the devise gem)</summary>
<pre>
<code>
Resolving dependencies...
Bundler could not find compatible versions for gem "rails-dom-testing":
  In Gemfile:
    rails (= 5.1.7) was resolved to 5.1.7, which depends on
      actionpack (= 5.1.7) was resolved to 5.1.7, which depends on
        rails-dom-testing (~> 2.0)

    jquery-rails (~> 4.0.5) was resolved to 4.0.5, which depends on
      rails-dom-testing (~> 1.0)

      In particular, rails-dom-testing (~> 1.0) is not satisfied by 2.0.3
</code>
</pre>
</details>

Perhaps a future iteration could suggest to the user which gem is limiting the resolution - so, something like "Try upgrading jquery-rails."

### LINKED ISSUE #5689: Very slow dependency resolving

The step to resolve the project dependencies with bundler takes multiple hours.

I tried it with 2.3.17 and a couple of other older releases of bundler:

```
$ bundle --version
Bundler version 2.3.17
```

With this Gemfile:

```
source ENV['GEM_SOURCE'] || 'https://rubygems.org'

  gem "puppet-module-posix-default-r2.7", '~> 1.0', require: false
  gem "puppet-module-posix-dev-r2.7", '~> 1.0',     require: false
  gem "github_changelog_generator",                 require: false
  gem "puppet-module-posix-system-r2.7", '~> 1.0',  require: false
  gem "librarian-puppet",                           require: false
```

I also tried it inside a Docker container:

```
FROM ruby:2.7.6-slim

ADD Gemfile /root/Gemfile

RUN gem install --version '= 2.3.17' bundler && \
        bundle --version && \
        cd /root && \
        bundle config set --local path vendor/bundle && \
        bash -c 'time bundle install'
```

The build inside the Docker container fails, because of missing build dependencies, however bundler mostly spends resolving dependencies there and the final time output, after it fails to build is:

```
real    332m27.595s
user    331m56.310s
sys     0m21.727s
```

So over 5 hours.

Which is the same on my host, where the build of those dependencies succeeds.

```
bundle config set --local path vendor/bundle
bundle install
```

With the `Gemfile` posted above in place.

Resolving dependencies in under 5 minutes.

### What actually happened?

Resolving dependencies in over 5 hours.

I used the docker container, so the environment should be clear. I can however post it as well if required...

deivid-rodriguez:

Our current resolve engine has performance issues where sometimes it will hang during resolution due to an excessive number of resolution steps. In addition to that, its resolution error messages tend to be confusing and too verbose, not highlighting the real resolution culprits.

My fix is to migrate Bundler to use [pub_grub](https://github.com/jhawthorn/pub_grub) instead.

Closes #3296.
Closes #5689.


voxik:
I understand this is WIP, but seeing the new resolver as part of Bundler instead of RubyGems does not seems to be right.

BTW I also wonder, why there needs to be new resolver. For instance, there is libsolv used by DNF on Fedora. Of course, it is not Ruby, but it has Ruby bindings at least.

simi:
> I understand this is WIP, but seeing the new resolver as part of Bundler instead of RubyGems does not seems to be right.

Can you explain why it is not right?

> BTW I also wonder, why there needs to be new resolver. For instance, there is libsolv used by DNF on Fedora. Of course, it is not Ruby, but it has Ruby bindings at least.

Per my understanding it should be much faster. Some info is at https://github.com/gel-rb/gel#why-should-i-use-gel.

deivid-rodriguez:
> I understand this is WIP, but seeing the new resolver as part of Bundler instead of RubyGems does not seems to be right.

I also don't get what's not right about it. I guess you mean we should also migrate the RubyGems resolver, and eventually keep everything in RubyGems? I guess, but no way that should be a blocker for improvements like this one. It's very close to ready, by the way, I'm in the process of upstreaming changes needed to pub_grub and it's progressing smoothly :)

> BTW I also wonder, why there needs to be new resolver. For instance, there is libsolv used by DNF on Fedora. Of course, it is not Ruby, but it has Ruby bindings at least.

There are issues with our current resolver that have been around for ages. In particular that its error messages are confusing, and that somethings it's not smart enough and takes forever to find a solution. PubGrub is state of the art with respect to the version solving problem and many package managers are adopting it for good reasons, have a look at https://nex3.medium.com/pubgrub-2fb6470504f too.

voxik:
> I guess you mean we should also migrate the RubyGems resolver, and eventually keep everything in RubyGems?

Yep, this is the end result I'd like to see. I think that Bundler should be extension to RubyGems, not the other way around. But it seems we are on the same page after all :)

> BTW I also wonder, why there needs to be new resolver.

Just to explain, I was not defending Molinillo. I was just pointing out there are already other dependency resolvers. However, the [PIP discussion](https://github.com/pypa/pip/issues/7406) might be enlightening, why e.g. libsolv might not be the right choice.

deivid-rodriguez:
Makes sense!

The "why there needs to be a new resolver" and then suggesting an alternative resolver confused me 😅. I do defend Molinillo by the way, it has done the job so far, and I'm sure it provided a much better alternative to what we had when it was first introduced. But it's time to move on to something better!

The PIP link was an interesting read, thanks! The outcome confused me a bit because it sounds like they went with something else that's not Pub Grub but I tend to recall pub grub style messages from running pip... 🤔

voxik:
> The PIP link was an interesting read, thanks! The outcome confused me a bit because it sounds like they went with something else that's not Pub Grub but I tend to recall pub grub style messages from running pip... thinking

I think they have some (probably implementation specific) [concerns](https://github.com/pypa/pip/issues/7406#issuecomment-578151915), which disqualified the implementation, not the algorithm.

deivid-rodriguez:
This PR is now ready for a review! There are a few changes proposed upstream, but I don't think we need to wait for them to move this PR forward.

indirect:
Thanks for persisting on this and working through it! I'm excited about the idea of being able to shared algorithm and library improvements across Bundler, RubyGems, Gel, Dart, and all the other pubgrub implementors and users.

hsbt:
@deivid-rodriguez Do you(we?) have a plan to migrate Molinillo to PubGrub on RubyGems's resolver in the future? (Sorry to my late concern)

bundler/spec/install/gemfile/sources_spec.rb
```diff
@@ -371,7 +371,15 @@
...
-        expect(err).to include("Could not find gem 'missing', which is required by gem 'depends_on_missing', in any of the sources.")
+        expect(err).to end_with <<~E.strip
+          Could not find compatible versions
+
+          Because every version of depends_on_missing depends on missing >= 0
+            and no versions in any of the sources satisfy missing >= 0,
+            every version of depends_on_missing is forbidden.
+          So, because Gemfile depends on depends_on_missing >= 0,
+            version solving has failed.
+        E
```
indirect, line 382: I'm not sure if this kind of message is clearer or not. "No versions satisfy `gem` >= 0" seems like an unclear way of saying "we could not find a gem named `gem` in any sources or installed locally", to me.

indirect (commented):
I've built and installed Bundler from this PR and I'm going to try it out for the next day or two and report back! In the meantime, I have one question about the failure messages. Overall, it's looking good!

deivid-rodriguez:
@hsbt Yes, migrating RubyGems resolver to pub grub as well should be our goal, yeah. I think it should not be as painful as this migration, but it will require some work. I can open a separate ticket about that if that's ok?

bundler/spec/install/gemfile/sources_spec.rb
deivid-rodriguez: I think there may be room for improvement here.

 This is original pub grub message:

```
Because every version of depends_on_missing depends on missing >= 0
  and no versions satisfy missing >= 0,
  every version of depends_on_missing is forbidden.
So, because Gemfile depends on depends_on_missing >= 0,
  version solving has failed.
```

I added a small tweak to mention the specific source for the gem when no versions could be found, hence the **in any of the sources** extra bit. Note that this case is particularly unspecific ("in any of the sources")  because it's a spec for the legacy multiple global sources mode. Normally, error messages will include a specific source ("in `https://rubygems.org` or installed locally").

I figured mentioning the source could sometimes help figuring out why no matching versions were found, since that's we already do today.

So, your suggestion would be something like the following?

```
Because every version of depends_on_missing depends on missing >= 0
  and missing >= 0 could not be found in any of the sources,
  every version of depends_on_missing is forbidden.
So, because Gemfile depends on depends_on_missing >= 0,
  version solving has failed.
```

indirect: Yes, exactly, something like that. I might say "cannot be used" instead of "forbidden"? But that's basically what I was thinking. :)

deivid-rodriguez: Great! I pushed updated messages.

The "forbidden" part is a bit trickier because pub grub we'd need to make sure it reads well in every situation, and it may get a bit complex. See https://github.com/jhawthorn/pub_grub/blob/9d8e9a5d2ce781e3967cffe1cc9ca3838613b457/lib/pub_grub/incompatibility.rb#L55-L102.

hsbt:
> I can open a separate ticket about that if that's ok?

Of course! I'm also excited to open the door to migrate to PubGrub on RubyGems and Bundler.

deivid-rodriguez:
I created a tracking issue at https://github.com/rubygems/rubygems/issues/6042.

Fryguy:
I'll try this out on ManageIQ - we have a _crazy_ Gemfile with git based gems, plugins, custom gem servers :)

bundler/Rakefile
```diff
@@ -139,29 +139,39 @@ end
...
+#   customization of the "root package", which we call "Gemfile" in our error
+#   messages.
+#
+# * https://github.com/jhawthorn/pub_grub/pull/16, for easier access to pub_grub
+#   errors, in order to retry resolution with more flexible constraints when it
+#   fails.
+#
+# * https://github.com/jhawthorn/pub_grub/pull/17, for easier customization of
+#   error messages to include sources and better metadata dependency conflict
+#   explanations.
```
Fryguy, line 155: Is this goal to get these patches merged and released before bundler is updated or are you planning on carrying these patches?

deivid-rodriguez: I would like to get these patches merged, but I documented them for now in case review takes a while and we decide to move on. They are mostly cosmetic patches, the one that [more fundamentally changes resolution](https://github.com/jhawthorn/pub_grub/pull/13) has already been merged. So I think we're good even if we decide to move on before these are merged/reviewed upstream, we were also patching Molinillo after all :)

deivid-rodriguez: All patches are now merged! 🎉

deivid-rodriguez:
I tried ManageIQ Gemfile since I was too curious 😅.

Turns out pub grub is actually slower in this particular case. I did not recall another case where this version of pub grub is actually slower than Molinillo, but I knew it would happen 😅.

In any case, most of the time Bundler spends here is downloading external repositories, and the resolution itself seems fast with both resolvers. My numbers for a full `bundle lock` are about ~30 seconds with Molinillo and ~34 seconds with pub grub.

Still seems good enough for this "pathological case", but we can keep it in mind when trying to speed things up.

I also noticed that [my patch to pub grub](https://github.com/jhawthorn/pub_grub/pull/13) makes a significant difference because without it, `bundle lock` takes ~52 seconds, so that's good.

bundler/spec/install/gemfile/sources_spec.rb
deivid-rodriguez: @indirect Do you think it's worth re-ticketing the suggestion to avoid "forbidden" in favor of other wording? I guess we'd need to change two cases:

#### With a universal constraint:

Change

> every version of depends_on_missing is forbidden

to 

> depends_on_missing cannot be used

#### With another constraint:

Change

> depends_on_missing >= 2.0 is forbidden

to

> depends_on_missing >= 2.0 cannot be used

I'm happy to open an issue at pub_grub upstream.

The proposed missing feels better to me but I don't know why, so hard to justify changing. Maybe "forbidden" is too strong of a word?

jamiemccarthy:
This works great on Vox Media's 260-line Gemfile for our main public Rails webserver. `bundle update --local` runs successfully at about the same speed (maybe a few percent faster) but more importantly when it can't find a solution pub_grub has so much better output describing the incompatibilities.

simi:
♾️ I did some quick test on our Gemfile (210 lines, but nothing fancy) and everything works well. Lockfile is without change. I have not measured the time yet, but I can try if welcomed.

bundler/spec/install/gemfile/sources_spec.rb
indirect: A separate ticket sounds fine to me. I think the reason “forbidden” can possibly be a little misleading is because we often use that word around permissions or authorization—so it might give the impressions that the user could use that gem if only they had the right permission. Instead, we are trying to say “we can’t do that because it is a logical impossibility” which has a different feel.

deivid-rodriguez: Understood! Created an issue at https://github.com/jhawthorn/pub_grub/issues/18.

Fryguy:
> I tried ManageIQ Gemfile since I was too curious 😅.

Thank you! Glad to see that there were no major issues.  30s to 34s is pretty close, so I don't think that's a problem.

> In any case, most of the time Bundler spends here is downloading external repositories

git repositories are very expensive both in time to fetch and in storage space.  I believe there are some wins here but I never got a chance to dig in.  After the switch to PubGrub maybe we can look into this.

deivid-rodriguez:
@indirect How did your testing go?

simi:
I forgot to mention our Gemfile is using mutliple sources (rubygems.org as a main global one and some gems are scoped to githug packages source). Everything works well.

indirect:
@deivid-rodriguez it looks good from my end! I didn’t see any issues while working on a few smaller projects this week.

indirect (approved):

simi (approved):

deivid-rodriguez:
Great! Going in then!

krzysiek1507:
Hey! Which release this change will be included in?

deivid-rodriguez:
We'll release this with Bundler 2.4.0 in a few days/weeks.

duckinator:
I am _incredibly_ excited about this! Thank you so much for making it happen, @deivid-rodriguez! ❤️

PR #5978, opened 2022-10-12
Map 'universal' to the real arch in Bundler for prebuilt gem selection

Bo98:

#4234 - Bundler may choose the wrong prebuilt gem architecture for universal Ruby builds

The fix is to adjust `Gem:Platform.local` to change the `universal` CPU to the real architecture, so it will pick the correct prebuilt gem. If the gem is available as a universal build, it will still be able to pick that.

I did so while retaining the value of `extensions_dir`. This is so the default extension directory for that Ruby is unchanged (still `universal-darwin-XX`), which will retain compatibility should anything be executed outside of Bundler.

This fix can be ported to Rubygems too if desired. I personally don't need to however.

In regards to "arm64e", I suppose an alternative fix there would've been to incorporate that in `Gem::Platform` somewhere, like how `i386` is mapped to `x86`. Technically arm64e is a different ABI to arm64, but they are interoperable and this the different is largely irrelevant for now because there should be no `arm64e-darwin*` gems on rubygems.org yet as only Apple-provided system binaries can run that architecture at the moment without jumping through boot arg hoops.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -310,7 +310,7 @@ module Gem
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
   class Specification
-    if /^universal\.(?<arch>.*?)-/ =~ RUBY_PLATFORM
+    if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
       local_platform = Platform.local
       if local_platform.cpu == "universal"
         ORIGINAL_LOCAL_PLATFORM = local_platform.to_s.freeze
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -308,6 +308,28 @@ module Gem
     end
   end
 
+  # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
+  class Specification
+    if /^universal\.(?<arch>.*?)-/ =~ RUBY_PLATFORM
+      local_platform = Platform.local
+      if local_platform.cpu == "universal"
+        ORIGINAL_LOCAL_PLATFORM = local_platform.to_s.freeze
+
+        local_platform.cpu = if arch == "arm64e" # arm64e is only permitted for Apple system binaries
+          "arm64"
+        else
+          arch
+        end
+
+        def extensions_dir
+          Gem.default_ext_dir_for(base_dir) ||
+            File.join(base_dir, "extensions", ORIGINAL_LOCAL_PLATFORM,
+                      Gem.extension_api_version)
+        end
+      end
+    end
+  end
+
   require "rubygems/util"
 
   Util.singleton_class.module_eval do
--- a/bundler/spec/support/helpers.rb
+++ b/bundler/spec/support/helpers.rb
@@ -432,6 +432,14 @@ module Spec
       pristine_system_gems :bundler
     end
 
+    def simulate_ruby_platform(ruby_platform)
+      old = ENV["BUNDLER_SPEC_RUBY_PLATFORM"]
+      ENV["BUNDLER_SPEC_RUBY_PLATFORM"] = ruby_platform.to_s
+      yield
+    ensure
+      ENV["BUNDLER_SPEC_RUBY_PLATFORM"] = old
+    end
+
     def simulate_platform(platform)
       old = ENV["BUNDLER_SPEC_PLATFORM"]
       ENV["BUNDLER_SPEC_PLATFORM"] = platform.to_s
--- a/bundler/spec/install/gemfile/platform_spec.rb
+++ b/bundler/spec/install/gemfile/platform_spec.rb
@@ -75,6 +75,81 @@ RSpec.describe "bundle install across platforms" do
     expect(the_bundle).to include_gems "platform_specific 1.0 RUBY"
   end
 
+  context "on universal Rubies" do
+    before do
+      build_repo4 do
+        build_gem "darwin_single_arch" do |s|
+          s.platform = "ruby"
+          s.write "lib/darwin_single_arch.rb", "DARWIN_SINGLE_ARCH = '1.0 RUBY'"
+        end
+        build_gem "darwin_single_arch" do |s|
+          s.platform = "arm64-darwin"
+          s.write "lib/darwin_single_arch.rb", "DARWIN_SINGLE_ARCH = '1.0 arm64-darwin'"
+        end
+        build_gem "darwin_single_arch" do |s|
+          s.platform = "x86_64-darwin"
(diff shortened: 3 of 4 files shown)


indirect:
I am worried that this will break on Apple Silicon hardware, which does not have a "real CPU". You can easily run the universal system ruby in either x86 or arm mode on the same machine, simply by charging the ARCH env var.

Installing a gem binary into the "universal" directory that only contains one of x86 or arm will cause errors when the universal ruby is run using the other arch.

Bo98:
> Installing a gem binary into the "universal" directory that only contains one of x86 or arm will cause errors when the universal ruby is run using the other arch.

How extensions are compiled is to my understanding controlled by `RbConfig::CONFIG["ARCH_FLAG"]`, which can be overriden by the envs `ARCHFLAGS` and `RC_ARCHS`. So changing `Gem::Platform.local` should not change this at all. The issue you describe already happens on Apple's Ruby, because they default to an empty `ARCH_FLAG`, which means to just use the compiler default (i.e. the current arch).

This behaviour hasn't changed, nor has the "universal" directory. It was already universal before this change for < Ruby 3.1 and remains so. Changing this directory doesn't seem safe on a Bundler level since RubyGems won't understand it, which is why I'm avoiding it.

Prebuilt gems are not installed into the extensions directory so that isn't a concern for them.

indirect:
Okay, that's fair. Gems do build to a single arch today, and it can definitely cause problems. It seems like it would be ideal for universal gems to build to a universal arch, but that's probably not worth it given Apple's plans to remove Ruby from macOS entirely. Thanks for explaining!

Bo98:
Yeah the extensions directory is a bit tricky. I see use cases for both a univesral directory (Apple always ship "fat" binaries) and arch specific directories (double compile is slow and compiler isn't guaranteed to support every arch).

I suppose the ideal solution would be something dynamic so that both style of directories could be read, but I'm hoping to avoid that discussion for this PR at least. Particularly since it is a separate issue, is beyond what Bundler can really control and isn't near as broken as installing prebuilts is. If Apple do drop their Ruby which they strictly want to be universal binaries, and Ruby 3.1 is defaulting to arch-specific directories for everyone else, then it'll naturally resolve itself over time.

deivid-rodriguez:
@Bo98 Could you add some specs here? We have a `spec/support/hax.rb` that we use to test against RubyGems modified in certain ways, I think you could leverage that here.

Bo98:
Thanks for the pointers. I've added one here.

Bo98:
I was unable to reproduce the issue locally - the test passes on a local build of Ruby master.

I'll try create a second branch and poke around the workflow to see what might be different and why.

deivid-rodriguez:
The ruby-core job is a bit strange, but it's there because otherwise the ruby-core team will revert our changes when upstream if they break their CI. So we try to run something as close as possible to what they run to prevent this.

deivid-rodriguez:
I can try have a look too if needed!

Bo98:
Found it: https://github.com/ruby/ruby/blob/c6f439a6a8df582416e756d7511aa4d9c72071a9/template/fake.rb.in#L38-L52

I'm guessing we could check `/^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)`.

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -308,6 +308,28 @@ def match_platforms?(platform, platforms)
...
   end
 
+  # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
+  class Specification
+    if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
+      local_platform = Platform.local
+      if local_platform.cpu == "universal"
+        ORIGINAL_LOCAL_PLATFORM = local_platform.to_s
+
+        local_platform.cpu = if arch == "arm64e" # arm64e is only permitted for Apple system binaries
```
deivid-rodriguez, line 318: I guess we also need a test simulating a `arm64e` platform.

bundler/spec/support/helpers.rb
```diff
@@ -432,6 +432,14 @@ def simulate_new_machine
       pristine_system_gems :bundler
     end
 
+    def simulate_ruby_platform(ruby_platform)
+      old = ENV["BUNDLER_SPEC_RUBY_PLATFORM"]
+      ENV["BUNDLER_SPEC_RUBY_PLATFORM"] = ruby_platform.to_s
+      yield if block_given?
+    ensure
+      ENV["BUNDLER_SPEC_RUBY_PLATFORM"] = old if block_given?
+    end
```
deivid-rodriguez, line 441: This helper is never currently used with a block, right? I would remove the unused code paths if that's the case.

Actually, in my opinion the cleanest version of this helper is with a block, so I'd change the helper to only support passing a block and change the tests that use it to the block version.

deivid-rodriguez (commented):
Added a couple of comments, but looks good!

bundler/spec/support/helpers.rb
Bo98: Yeah, this was copied from `simulate_platform`. Adjusted now.

bundler/spec/support/builders.rb
```diff
@@ -138,6 +138,19 @@ def build_repo1
...
+        end
+        build_gem "darwin_single_arch" do |s|
+          s.platform = "arm64-darwin"
+          s.write "lib/darwin_single_arch.rb", "DARWIN_SINGLE_ARCH = '1.0 arm64-darwin'"
+        end
+        build_gem "darwin_single_arch" do |s|
+          s.platform = "x86_64-darwin"
+          s.write "lib/darwin_single_arch.rb", "DARWIN_SINGLE_ARCH = '1.0 x86_64-darwin'"
+        end
+
```
deivid-rodriguez, line 153: Sorry to bother you again @Bo98, just one more request. Since this gem is only used in the particular tests added here, I think it makes the tests more easily readable to move the creation of the dummy gem there. What I normally do is using an empty repository with only those gems, like this:

```ruby
build_repo4 do
  # definitions of the dummy gems here
end
```

And then use `file_uri_for(gem_repo4)` in test Gemfile and lock file.

deivid-rodriguez (commented):
Just one final comment @Bo98!

deivid-rodriguez:
Rebased, I think the previous failure was just a flaky one.

deivid-rodriguez (approved):

Bo98:
Hmm I guess I'm unlucky.

The disparity between CRuby and JRuby will be because the former is Rails 7 and the latter is Rails 6 (until JRuby 9.4 is out).

Rails 6 uses `webpacker` which seems to have a very slow install time.

We could use `--skip-webpack-install`, though that flag will only be recognised under Rails 6. `--skip-javascript` is an alternative, and is recognised on both, but does skip more than strictly necessary.

Bo98:
Workflow tweak turns out quite clean: #6049

Issue #5454, opened 2022-04-06
Security features tests are failing on RHEL 9 / c9s


jackorp:
Tests related to security features of RubyGems are currently failing during Ruby build in RHEL 9 / CentOS 9 stream.

RHEL 9 / c9s have introduced stricter crypto policies and with that, the SHA1 is disabled by default.
The following tests are failing in this environment, setting the crypto policies to LEGACY instead of DEFAULT makes the tests pass.

What I found the failures have in common are loading certificate fixtures that are using SHA1 in one way or another.

The failures:
```
  1) Failure:
TestGemRequest#test_configure_connection_for_https_ssl_ca_cert [/builddir/build/BUILD/ruby-3.1.1/test/rubygems/test_gem_request.rb:124]:
Failed assertion, no message given.
  2) Failure:
TestGemRequest#test_configure_connection_for_https [/builddir/build/BUILD/ruby-3.1.1/test/rubygems/test_gem_request.rb:105]:
Failed assertion, no message given.
  3) Failure:
TestGemSecurityPolicy#test_check_chain_invalid [/builddir/build/BUILD/ruby-3.1.1/test/rubygems/test_gem_security_policy.rb:108]:
<"invalid signing chain: certificate /CN=invalidchild/DC=example was not issued by /CN=child/DC=example"> expected but was
<"invalid signing chain: certificate /CN=child/DC=example was not issued by /CN=nobody/DC=example">.
  4) Failure:
TestGemSecurityPolicy#test_verify_signatures_missing [/builddir/build/BUILD/ruby-3.1.1/test/rubygems/test_gem_security_policy.rb:493]:
<"missing signature for data.tar.gz"> expected but was
<"certificate /CN=nobody/DC=example was not issued by /CN=nobody/DC=example">.
  5) Error:
TestGemPackageTarWriter#test_add_file_signer:
Gem::Security::Exception: certificate /CN=nobody/DC=example was not issued by /CN=nobody/DC=example
    /builddir/build/BUILD/ruby-3.1.1/lib/rubygems/security/policy.rb:101:in `check_cert'
    /builddir/build/BUILD/ruby-3.1.1/lib/rubygems/security/policy.rb:141:in `check_root'
    /builddir/build/BUILD/ruby-3.1.1/lib/rubygems/security/policy.rb:247:in `verify'
    /builddir/build/BUILD/ruby-3.1.1/lib/rubygems/security/signer.rb:154:in `sign'
[...]

PR #5982, opened 2022-10-13
Use SHA256 in test certificates

jackorp:

CentOS 9 Stream / RHEL 9 (Fedora will follow as well) are disabling SHA1 digest in default allowed crypto configuration. Most operations with certificates (such as signing content with them) fail. Operations with SHA1 also fail in the test suite ran as part of the whole Ruby test suite.

Fixes https://github.com/rubygems/rubygems/issues/5454

Regenerate certificates using SHA256, using `util/create_certs.rb` with SHA256 as the hash instead of SHA1.

I have corrected most failures that cropped up as a result of regenerated certificates, but there are 2 tests that still fail because they are using SHA1:
`TestGemPackage#test_build_auto_signed_encrypted_key` and `TestGemPackage#test_build_signed_encrypted_key`

They are using a certificate and key that I am not sure where they are generated (or how): `test/rubygems/data/gem-public_cert.pem` and `gem-private_key.pem`

I'll try to add these certificates to the `util/create_certs.rb` tool.

- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes (We are fixing tests themselves here)


jackorp:
Set as draft to have preview of the work, and I'd like to update certs also for the `TestGemPackage#test_build_signed_encrypted_key`

deivid-rodriguez:
This fixes #5454 right, @jackorp? Thanks!

@simi Do you have time to review this? I saw that you self-assigned that issue, and also I've seen you working on other OpenSSL things around, so maybe you have better context.

jackorp:
Yes, it fixes that issue. I have run builds to confirm this, and those builds passed.

simi (approved):

deivid-rodriguez:
Thanks so much @jackorp!

PR #5991, opened 2022-10-17
Prepare RubyGems 3.3.24 and Bundler 2.3.24

deivid-rodriguez:
It's release day!


deivid-rodriguez:
It's not actually release day, but there's a regression many people is running into on Windows, so I want to get the fix out as soon as possible.

simi (approved):

simi:
Windows CI failed for 3.1, but it seems it would be enough to restart since the trouble is in `setup-ruby` action. :thinking:

deivid-rodriguez:
I restarted that, and also a new smoke test that I added to generate a new Rails app on Windows, which failed (I suspect) due to some JRuby issue. If this one keeps failing like that too much, I'll just disable it on JRuby.

PR #6002, opened 2022-10-20
Allow upcoming JRuby to pass keywords to Kernel#warn

enebo:

rubtygems monkey patches Kernel#warn making it impossible for JRuby to pass any Ruby warning category features (or uplevel).

jruby-head (which will be JRuby 9.4.0.0) can now properly process the keywords to Kernel#warn.  I cannot think of any capability based test for this so I constrained it using a version guard.  Only JRuby will ever hit the version guard.

- [ x] Describe the problem / feature
- [ x] Write code to solve the problem
- [ x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

JRuby internal tests do not work with this monkey patch stripping off keywords.  I did not make a test for this change in this repository but JRuby itself has lots of test coverage for the behavior of warn.

### DIFF
--- a/lib/rubygems/core_ext/kernel_warn.rb
+++ b/lib/rubygems/core_ext/kernel_warn.rb
@@ -16,7 +16,7 @@ if RUBY_VERSION >= "2.5" && !Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES
 
     module_function define_method(:warn) {|*messages, **kw|
       unless uplevel = kw[:uplevel]
-        if Gem.java_platform?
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):

deivid-rodriguez:
Thanks @enebo!

PR #6013, opened 2022-10-25
Fix bad unfreeze recommendation

deivid-rodriguez:

If failing because lockfile changed while in frozen mode, Bundler recommends to run a command in development to temporarily unfreeze Bundler and regenerate the lockfile.

That recommendation does not make sense, however, if Bundler is frozen through ENV, and it currently results in a meaningless advice reading:

```
If this is a development machine, remove the /path/to/Gemfile freeze by running ``
```

This commit fixes the issue by avoiding displaying any recommendation in this case.

### DIFF
--- a/bundler/spec/commands/update_spec.rb
+++ b/bundler/spec/commands/update_spec.rb
@@ -613,6 +613,12 @@ RSpec.describe "bundle update" do
       expect(err).to match(/You are trying to install in deployment mode after changing.your Gemfile/m).
         and match(/freeze \nby running `bundle config unset deployment`./m)
     end
+
+    it "should not suggest any command to unfreeze bundler if frozen is set through ENV" do
+      bundle "update", :all => true, :raise_on_error => false, :env => { "BUNDLE_FROZEN" => "true" }
+      expect(err).to match(/You are trying to install in deployment mode after changing.your Gemfile/m)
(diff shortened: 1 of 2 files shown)


bundler/spec/commands/update_spec.rb
```diff
@@ -613,6 +613,12 @@
       expect(err).to match(/You are trying to install in deployment mode after changing.your Gemfile/m).
         and match(/freeze \nby running `bundle config unset deployment`./m)
     end
+
+    it "should not suggest any command to unfreeze bundler is freezing is set through ENV" do
```
simi, line 617: Should be `...if freezing...`?

deivid-rodriguez: Maybe this better?

```suggestion
    it "should not suggest any command to unfreeze bundler is frozen is set through ENV" do
```

simi: There is `is` twice, it seems to me one of the `is` should be actually `if`. But maybe I miss something.

deivid-rodriguez: Oh, sorry, no no, I misread it! Will fix, thanks!

simi (approved):

PR #6014, opened 2022-10-25
Handle development dependencies duplicated in gemspec vs Gemfile

deivid-rodriguez:

If a Gemfile duplicates a development dependency also defined in a local gemspec with a different requirement, the requirement in the local gemspec will be silently ignored.

This surprised me.

I think we should either:

* Make sure both requirements are considered, like it happens for runtime dependencies (I added a spec to illustrate the current behavior here).

* Add a warning that the requirement in the gemspec will be ignored.

I think the former is slightly preferable, but it may cause some bundle's that previously resolved to no longer resolver.

I went with the latter but the more I think about it, the more this seems like it should behave like the former.

### DIFF
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -102,38 +102,43 @@ module Bundler
 
       # if there's already a dependency with this name we try to prefer one
       if current = @dependencies.find {|d| d.name == dep.name }
-        deleted_dep = @dependencies.delete(current) if current.type == :development
+        if current.requirement != dep.requirement
+          current_requirement_open = current.requirements_list.include?(">= 0")
 
-        unless deleted_dep
-          if current.requirement != dep.requirement
-            return if dep.type == :development
+          if current.type == :development
+            @dependencies.delete(current)
 
+            unless current_requirement_open || dep.type == :development
+              Bundler.ui.warn "A gemspec development dependency (#{dep.name}, #{current.requirement}) is being overridden by a Gemfile dependency (#{dep.name}, #{dep.requirement}).\n" \
+                              "This behaviour may change in the future. Please remove either of them, or make sure they both have the same requirement\n" \
+            end
+          else
             update_prompt = ""
 
             if File.basename(@gemfile) == Injector::INJECTED_GEMS
-              if dep.requirements_list.include?(">= 0") && !current.requirements_list.include?(">= 0")
+              if dep.requirements_list.include?(">= 0") && !current_requirement_open
                 update_prompt = ". Gem already added"
               else
                 update_prompt = ". If you want to update the gem version, run `bundle update #{current.name}`"
 
-                update_prompt += ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive." unless current.requirements_list.include?(">= 0")
+                update_prompt += ". You may also need to change the version requirement specified in the Gemfile if it's too restrictive." unless current_requirement_open
               end
             end
 
             raise GemfileError, "You cannot specify the same gem twice with different version requirements.\n" \
-                            "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement})" \
-                             "#{update_prompt}"
-          elsif current.source != dep.source
-            return if dep.type == :development
-            raise GemfileError, "You cannot specify the same gem twice coming from different sources.\n" \
-                            "You specified that #{dep.name} (#{dep.requirement}) should come from " \
-                            "#{current.source || "an unspecified source"} and #{dep.source}\n"
-          else
-            Bundler.ui.warn "Your Gemfile lists the gem #{current.name} (#{current.requirement}) more than once.\n" \
-                            "You should probably keep only one of them.\n" \
-                            "Remove any duplicate entries and specify the gem only once.\n" \
-                            "While it's not a problem now, it could cause errors if you change the version of one of them later."
+                           "You specified: #{current.name} (#{current.requirement}) and #{dep.name} (#{dep.requirement})" \
+                           "#{update_prompt}"
           end
+        elsif current.source != dep.source
+          return if dep.type == :development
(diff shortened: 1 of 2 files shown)


simi:
Thinking about this again after hitting this today (https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/716) I would suggest warning as a good start in here.

bundler/lib/bundler/dsl.rb
```diff
@@ -102,38 +102,43 @@ def gem(name, *args)
...
 
-        unless deleted_dep
-          if current.requirement != dep.requirement
-            return if dep.type == :development
+          if current.type == :development
+            @dependencies.delete(current)
 
+            unless current_requirement_open || dep.type == :development
+              Bundler.ui.warn "A gemspec development dependency (#{dep.name}, #{current.requirement}) is being overriden by a Gemfile dependency (#{dep.name}, #{dep.requirement}).\n" \
+                              "Please remove either of them, or make sure they both have the same requirement\n" \
```
simi, line 113: Would it be wise to announce this will change later?

deivid-rodriguez: I updated the message, what do you think?

simi: looks better, :muscle:

simi (approved):

simi:
What about release? Is patch bump ok since it just introduces warning? :thinking:

deivid-rodriguez:
I think patch should be fine, yes?

simi:
> I think patch should be fine, yes?

I'm ok with patch.

sambostock:
👋 Something in the last couple days broke Ruby HEAD CI in Shopify/job-iteration

 - Shopify/job-iteration#436 from November 12 [passes](https://github.com/Shopify/job-iteration/actions/runs/6842124678/job/18603251833?pr=436)
 - Shopify/job-iteration#437 from November 15 [fails](https://github.com/Shopify/job-iteration/actions/runs/6880158446/job/18713744117?pr=437)
 
Neither of those PRs touches dependencies, so it must be something upstream, and I'm not positive, but this PR seems related.

In `job-iteration`, we have a CI matrix in GitHub Actions between Ruby versions and various Gemfiles which specify different version requirements for `activejob` and `activerecord`.

```ruby
# gemfiles/rails_edge.gemfile (for example)
eval_gemfile "../Gemfile"

gem "activejob", github: "rails/rails", branch: "main"
gem "activerecord", github: "rails/rails", branch: "main"
```
```ruby
# Gemfile
# ...
gemspec
# ...
```
```ruby
# job-iteration.gemspec
# ...
  spec.add_development_dependency("activerecord")
  spec.add_dependency("activejob", ">= 5.2")
end
```

The error we're now seeing is

```
...
> bundle install
/home/runner/.rubies/ruby-head/bin/bundle config --local path /home/runner/work/job-iteration/job-iteration/vendor/bundle
/home/runner/.rubies/ruby-head/bin/bundle lock

[!] There was an error parsing `rails_edge.gemfile`: You cannot specify the same gem twice coming from different sources.
You specified that activerecord (>= 0) should come from an unspecified source and https://github.com/rails/rails. Bundler cannot continue.

 #  from /home/runner/work/job-iteration/job-iteration/gemfiles/rails_edge.gemfile:6
 #  gem "activejob", github: "rails/rails", branch: "main"
 >  gem "activerecord", github: "rails/rails", branch: "main"
Error: The process '/home/runner/.rubies/ruby-head/bin/bundle' failed with exit code 4
```

I've tentatively opened Shopify/job-iteration#438 to change our approach, in case the way we were doing it was technically incorrect (and we'll probably merge it even if it isn't), but I wanted to flag this so we can identify if this PR is the cause of the error or not, and if it is, to surface that it may not be a safe change to make as a patch version update.

simi:
@sambostock can you extract and share minimal steps to reproduce this? There should be warning, not failure.

sambostock:
@simi I've created a [minimal repo](https://github.com/sambostock/gemfile-gemspec-conflicts-demo) which [reproduces the failure in this CI run](https://github.com/sambostock/gemfile-gemspec-conflicts-demo/actions/runs/6885295820/job/18729224164).

deivid-rodriguez:
Thanks for the report, will look into this before releasing this PR 👍.

tagliala:
Probably I run into the same issue here:

https://github.com/ifad/chronomodel/actions/runs/7021550132/job/19103883939

The use case is similar to the above one, we're using specific `gemfiles` to test across several rails versions

deivid-rodriguez:
Reproduced! Looking into it now.

deivid-rodriguez:
#7215 fixes this. Thanks for diligently reporting the problem so early ❤️.

WesleyW:
Hi @deivid-rodriguez!

I'm running into the same issue as [this comment](https://github.com/rubygems/rubygems/pull/6014#issuecomment-1830357417). We're using [Appraisal](https://github.com/thoughtbot/appraisal/) to test that our gem works with multiple versions of a gem (e.g. Rails). This means our Gemfile + gemspec look something like this when we're testing:
```
# Gemfile
gem "rails", "~> 7.0.8"
```
```
  spec.add_development_dependency 'rails'
```
OR like this:
```
# Gemfile
gem "rails", "~> 7.0.8"
```
```
  spec.add_development_dependency 'rails', '>= 5'
```


From the wording of this issue, it seems like our use case is discouraged, but it should just warn rather than error outright. Could you clarify what the best practice for us would be or if this could be fixed?

deivid-rodriguez:
@WesleyW Seems like a similar bug, I'll try to fix it for final ruby release.

deivid-rodriguez:
I think in your case the warning is intended (at least when both Gemfile and gemspec specify some requirement). The way to silence the warning would be to remove either of them.

deivid-rodriguez:
Fixed by #7319, thanks for letting us know!

WesleyW:
Amazing! Tried it out locally and confirmed it works. Thank you for such a quick fix!

Issue #4047, opened 2020-11-04
Pure ruby replacement for `git ls-files` in the bundle gem generated gemspec


robotdana:
This can make git an unexpected & barely-necessary dependency of many gems.

e.g. https://github.com/rubygems/bundler/issues/5810

I have two gems which i'd be happy to copy into rubygems/bundler somehow to avoid this dependency on git
- https://github.com/robotdana/git_ls (parse the binary .git/index file in pure ruby)
- https://github.com/robotdana/fast_ignore (reads all .gitignore rules in pure ruby and walks the file system checking against them, useful when .git/index doesn't exist for some reason. it's also pretty quick)

Is this something the bundler team would be at all interested in me doing?

simi:
@robotdana It would be great to find alternative having no external dependency. But how should that work in your case? If I understand this well, your idea is to vendor those gems into rubygems and use them in default gemspecs. What about older rubygems versions? AFAIK it will make that new gemspec incompatible with old versions.

deivid-rodriguez:
Hi @robotdana! Thanks for bringing this up.

Yes, I would definitely be interested. I guess the idea would be to vendor this code inside rubygems and the use it in the generated gemspec if available, right?

Something I was thinking though after I saw https://github.com/rubygems/rubygems/pull/4033 is that maybe "respecting git" is not a good idea after all. If I got the history of this issue correctly, we introduced `git ls-files` in the generated gemspec to avoid newbie users from pushing heavy gems full of "garbage files" like generated files. However, `git ls-files` does not prevent "garbage" from being included on a gem as that pull request shows. In other words, you might want some files to be source controlled, but not included with your gem. Conversely, you might want some generated files to be actually included. For example, in the case of gems including precompiled extensions.

So, I'm leaning towards providing a good explicit enough default, like `Dir["lib/**/*.rb"]`, and also `Dir["ext/**/*.c"]` if `bundle gem` is run with the `--ext` flag.

Thoughts?

robotdana:
@simi, regarding the backward compatibility thing i was imagining at least temporarily some kind of `` if defined?(new thing) new thing; else `git ls-files` ``

@deivid-rodriguez I think a sensible simple explicit default would be ideal, and is what i generally do, even on the only gems in the world that can theoretically FastIgnore, i didn't bother suggesting it because it felt like the team was set on respecting git

[...]

PR #6019, opened 2022-10-26
Remove git from new gem gemspec template

abrahamparayil:
I work with the Debian Ruby team to package various ruby software for Debian. When we build gems, having git in gempsec would mean that we would need to git as a build dependency for every gem which isn't quite necessary. Until now Debian contributors have been making PRs to upstream maintainers to remove git and use something like `Dir`. The upstream maintainers were all very kind and understanding, so they merge the changes.

Most of these maintainers use this method to specify files in gemspec because it is how it was by default in the files generated with `bundle gem`. I am aware that even if we change the template, a developer can just use git and in that case we (Debian contributors) can patch it when we use it. That being said not using git in the default files created by `bundle gem` would prevent it from getting into new gems and save us a bunch of time. [More info](https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#packaginggemspecgit)

Lastly, specifying individual files/folders would mean that only necessary files (that an end user would care about) are shipped in the gem, which is overall a good thing.

Please let me know what you folks think! Thanks for the amazing work.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -23,12 +23,7 @@ Gem::Specification.new do |spec|
   spec.metadata["changelog_uri"] = "TODO: Put your gem's CHANGELOG.md URL here."
 
   # Specify which files should be added to the gem when it is released.
-  # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files = Dir.chdir(__dir__) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      (f == __FILE__) || f.match(%r{\A(?:(?:bin|test|spec|features)/|\.(?:git|travis|circleci)|appveyor)})
-    end
-  end
+  spec.files = Dir["lib/**/*.rb", "LICENSE.txt", "README.md"]
   spec.bindir = "exe"
   spec.executables = spec.files.grep(%r{\Aexe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]


simi:
There is alternative approach at https://github.com/rubygems/rubygems/issues/4047. Maybe we can try to push that one forward.

pboling:
I prefer this approach, as it is a simpler change, and makes the behavior clear and explicit.  Zero magic.

pboling:
Reading through rubygems source, for an unrelated reason, and noticed that internally Rubygems does this:
```
  s.files = File.read("Manifest.txt").split
```
And ships with a `Manifest.txt` file listing everything to be included.  If it is good enough for Rubygems, why shouldn't it be good enough for all of us?  I think splitting the responsibility of tracking files into a Manifest file is a common, widely understood, and beneficial practice, as it would also allow simple integrations with automated tooling that likes to know file names to look at.

deivid-rodriguez:
I also prefer this approach, but there are different points of view between maintainers as you can read at #4047. We do all agreed that the alternative approach in #4047 would be better over what we have now. However, I think the complexity of implementing that approach made it stall.

indirect:
One thing I am not seeing discussed is any of the reasons why we started using git to list files in the first place, so I would like to add that in, since I was there.

We decided to shell out to git to get a list of committed files because that was the closest thing we had to "which files, out of the files on disk, are actually supposed to go into the gem".

The reason we decided against `Manfest.txt` is because it must be updated every time a file is added or removed, either by a human or by a script. If it needs to be updated by a human, it will be missed, and then bugfix releases will be required. If it is updated by a script, we can just call the script in the gemspec, and it will be run automatically when the gem is built.

The reason we decided against `Dir["lib/**/*.rb"]` and other similar systems is that it's just too hard to know exactly which files should be included. Many gems include artifacts or other files that don't end with `.rb`, and many gems also have generated files or other artifacts that should not be included in the built gem. One outcome of a simple automatic file list that I have seen is a gem that includes every previous version inside itself, because the glob included every .gem the author built as they worked on new versions.

Beyond simply "why did we decide to use git", it seems to me like there are several separate things getting discussed here:

1) The Debian project wants to be able to build their own .gems, ignoring the .gem files released by authors, on machines that do not have git installed.

The Debian project can choose to do that if they want to, but I don't think we should take any action to support that choice. Authors do not publish the source of their packages to rubygems.org, they publish actual released versions to rubygems.org, and those released versions do not require git. IMO, is not our problem to solve the problem of the Debian project refusing to use released packages, created by package authors, that do not depend on git.

2) Some Ruby developers want to be able to check out and use gems without git installed.

It's valid to want this. It is also valid for gem authors to want to use git when they develop their gems. Anyone who doesn't want to use git while developing gems should absolutely go ahead and not use git. I do not think the git-removers should be asking the git-users to remove git or otherwise make changes or do work simply to accommodate the git-removers. It's open source! If you don't like git, you can make a version without git, that's the Beauty Of Open Source™.

3) Other Ruby developers find it distasteful or surprising that building a gem depends on a tool that is not written in Ruby.

This is maybe the most convincing argument for me personally, and I would be fine with the solution proposed in #4047, to vendor a gitignore parser written in Ruby and to use that to select files to include in the gem.

indirect:
(Since the implementation in this PR is not going to be accepted, I'm going to close the PR. That said, we do agree with the idea of removing the dependency on git, and I look forward to that happening at some point in the future.)

abrahamparayil:
@indirect Thanks for the detailed reply. I am currently working on implementing a solution mentioned in #4047 and will create a new PR for the same.

Issue #5942, opened 2022-09-20
Security discrepancies between rake task and documentation


pboling:
### Describe the problem as clearly as you can

The documentation on Security for Rubygems is not congruent with the rake task for `rake build:checksum`, indeed the documentation doesn't seem to be aware of the `checksum` task at all.

#### Discrepancies

When I mention "Docs" I am referring to this article: https://guides.rubygems.org/security/

1. Docs say to put the `checksum` files in a `checksum` directory, but the `build:checksum` task puts them in a `checksums` (plural) directory.
2. Docs say to use a script to generate an SHA512 checksum, and do not mention the availability of the `rake build:checksum` task.
3. Docs and `rake build:checksum` task both refer to SHA512 checksums, but the checksum created for reference and display on rubygems.org is SHA256.  Perhaps the rake task should generate both SHA256 and SHA512?
4. Checksum files creates by the rake task have an empty line at the end of the file, while when following the example script in the documentation an empty line does not result at the end of the checksum file.  This may not matter much, but it is inconsistent, and, when running checksums, a change in the file, and thus a potential change in the git history, can be disconcerting, and waste brain cycles.

### Did you try upgrading RubyGems?

Yes.
```
$ gem -v
3.3.22
```

### Run `gem env` and paste the output below

```
$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.3.22
  - RUBY VERSION: 3.1.2 (2022-04-12 patchlevel 20) [x86_64-darwin21]
  - INSTALLATION DIRECTORY: /Users/pboling/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0
  - USER INSTALLATION DIRECTORY: /Users/pboling/.gem/ruby/3.1.0
  - RUBY EXECUTABLE: /Users/pboling/.asdf/installs/ruby/3.1.2/bin/ruby
  - GIT EXECUTABLE: /usr/local/bin/git
  - EXECUTABLE DIRECTORY: /Users/pboling/.asdf/installs/ruby/3.1.2/bin
  - SPEC CACHE DIRECTORY: /Users/pboling/.gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /Users/pboling/.asdf/installs/ruby/3.1.2/etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - x86_64-darwin-21
  - GEM PATHS:
[...]

PR #6022, opened 2022-10-27
Add sha256 checksum to rake build:checksum

pboling:
… in addition to the existing SHA512

- Until now Rubygems.org has publicly displayed `SHA256` checksum for published gems, but has only created the `SHA512` checksum for the package via `rake build:checksum` task
- This will allow more gem authors to easily verify integrity of published gems with an existing Rubygems.org feature
- See #5942 for details on the problem

Create a SHA256 checksum, in addition to the current SHA512 checksum.

Note that I've changed the Gem helper method from `build_checksum` to `build_checksums`.  Certainly could keep it the same if that's a problem, but it felt good to be grammatically correct with it.  Not sure if internal or external API.

- [x] Tests - Have been updated

### DIFF
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -2,13 +2,12 @@
 
 require_relative "../bundler"
 require "shellwords"
-require "digest/sha2"
 
 module Bundler
   class GemHelper
     CHECKSUMS = {
-      "sha256" => ::Digest::SHA256,
-      "sha512" => ::Digest::SHA512,
+      "sha256" => "::Digest::SHA256",
+      "sha512" => "::Digest::SHA512",
     }.freeze
     include Rake::DSL if defined? Rake::DSL
 
@@ -109,8 +108,9 @@ module Bundler
 
     def build_checksums(built_gem_path = nil)
       built_gem_path ||= build_gem
-      CHECKSUMS.each do |extension, type|
-        write_checksum(built_gem_path, extension, type)
+      require "digest/sha2"
+      CHECKSUMS.each do |extension, digest_klass|
+        write_checksum(built_gem_path, extension, Object.const_get(digest_klass))
       end
     end
 
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -2,9 +2,14 @@
 
 require_relative "../bundler"
 require "shellwords"
+require "digest/sha2"
 
 module Bundler
   class GemHelper
+    CHECKSUMS = {
+      "sha256" => ::Digest::SHA256,
+      "sha512" => ::Digest::SHA512,
+    }.freeze
     include Rake::DSL if defined? Rake::DSL
 
     class << self
@@ -47,9 +52,9 @@ module Bundler
         built_gem_path = build_gem
       end
 
-      desc "Generate SHA512 checksum if #{name}-#{version}.gem into the checksums directory."
+      desc "Build #{name}-#{version}.gem into pkg directory, then generate SHA256 & SHA512 checksums in checksums directory."
       task "build:checksum" => "build" do
-        build_checksum(built_gem_path)
+        build_checksums(built_gem_path)
       end
 
       desc "Build and install #{name}-#{version}.gem into system gems."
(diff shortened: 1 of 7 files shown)


pboling:
I am working on a PR updating rubygems.org documentation to be concordant with `rake build:checksum` to address the other parts of #5942 .

bundler/lib/bundler/gem_helper.rb
```diff
@@ -2,9 +2,14 @@
 
 require_relative "../bundler"
 require "shellwords"
+require "digest/sha2"
```
deivid-rodriguez, line 5: Can we keep lazily loading `digest/sha2`?

deivid-rodriguez (commented):
Thanks for fixing this, @pboling! I just have one request regarding lazily loading `digest/sha`, so that it's only loaded when this task is used.

bundler/lib/bundler/gem_helper.rb
pboling: lazy loading prevents using this map:
```
    CHECKSUMS = {
      "sha256" => ::Digest::SHA256,
      "sha512" => ::Digest::SHA512,
    }.freeze
```
But I can use strings and const_get instead.

bundler/lib/bundler/gem_helper.rb
```diff
@@ -102,19 +106,25 @@ def install_gem(built_gem_path = nil, local = false)
...
     end
 
-    def build_checksum(built_gem_path = nil)
+    def build_checksums(built_gem_path = nil)
       built_gem_path ||= build_gem
-      SharedHelpers.filesystem_access(File.join(base, "checksums")) {|p| FileUtils.mkdir_p(p) }
-      file_name = "#{File.basename(built_gem_path)}.sha512"
       require "digest/sha2"
-      checksum = ::Digest::SHA512.file(built_gem_path).hexdigest
+      CHECKSUMS.each do |extension, digest_klass|
```
deivid-rodriguez, line 112: Super minor but I'm thinking, do you think the `CHECKSUMS` constant buys us much, over

```ruby
write_checksum(built_gem_path, "sha256", Digest::SHA256)
write_checksum(built_gem_path, "sha512", Digest::SHA512)
```

bundler/lib/bundler/gem_helper.rb
```diff
@@ -102,19 +106,25 @@ def install_gem(built_gem_path = nil, local = false)
       Bundler.ui.confirm "#{name} (#{version}) installed."
     end
 
-    def build_checksum(built_gem_path = nil)
+    def build_checksums(built_gem_path = nil)
       built_gem_path ||= build_gem
-      SharedHelpers.filesystem_access(File.join(base, "checksums")) {|p| FileUtils.mkdir_p(p) }
```
deivid-rodriguez, line 107: Why was this moved inside the `write_checksums` method? It should be enough to run it just once I think?

deivid-rodriguez (commented):
Just to small additional comments, but it's most ready from my side. Thanks so much for working on this!

bundler/lib/bundler/gem_helper.rb
pboling: Good point.  Oversight.

bundler/lib/bundler/gem_helper.rb
pboling: It doesn't get us much, agreed.

pboling:
I've created a gem that does this same concept, except it supports back to Ruby 2.2.  I expect it can be used as an alternative when people are stuck on older versions of Ruby, as I often find myself.

I tried pushing the gem to RubyGems.org, but the push was rejected.

```
❯ bundle exec rake release
["gem", "build", "-V", "/Users/pboling/src/rubygems/gem_checksums/gem_checksums.gemspec"]
gem_checksums 1.0.0 built to pkg/gem_checksums-1.0.0.gem.
["git", "diff", "--exit-code"]
["git", "diff-index", "--quiet", "--cached", "HEAD"]
["git", "tag"]
["git", "tag", "-m", "Version 1.0.0", "v1.0.0"]
Tagged v1.0.0.
["git", "rev-parse", "--abbrev-ref", "HEAD"]
["git", "config", "--get", "branch.main.remote"]
["git", "rev-parse", "--abbrev-ref", "HEAD"]
["git", "push", "origin", "refs/heads/main"]
["git", "push", "origin", "refs/tags/v1.0.0"]
Pushed git commits and release tag.
["gem", "push", "/Users/pboling/src/rubygems/gem_checksums/pkg/gem_checksums-1.0.0.gem"]
Code:   261990
RubyGems.org cannot process this gem.
Please try rebuilding it and installing it locally to make sure it's valid.
```

Perhaps if I change the name to `germ_checksums`?  It's odd because I already have another, very popular, gem called `gem_bench`, but maybe there is a rule in place guarding the `gem_*` prefix for new gem pushes?

deivid-rodriguez:
That's interesting, the error message is definitely not great because you just built the gem successfully 🤔

PR #6025, opened 2022-10-28
Load "bundler/setup" in lib/rubygems.rb

mame:

Currently, we cannot specify the version of did_you_mean by using Gemfile.

For example, Ruby master bundles with did_you_mean 1.6.1 by default:

```
$ ruby -e 'p DidYouMean::VERSION'
"1.6.1"
```

Consider that we want to use did_you_mean 1.5.0 with this version of ruby:

```
$ cat Gemfile
source "https://rubygems.org"
gem "did_you_mean", "= 1.5.0"
```

But the attempt fails:

```
$ bundle exec ruby -e 'p DidYouMean::VERSION'
/home/mame/work/ruby/local/lib/ruby/gems/3.2.0+3/gems/bundler-2.3.19/lib/bundler/runtime.rb:308:in `check_for_activated_spec!': You have already activated did_you_mean 1.6.1, but your Gemfile requires did_you_mean 1.5.0. Since did_you_mean is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports did_you_mean as a default gem. (Gem::LoadError)
...
```

This issue is not only with did_you_mean, but also with error_highlight and syntax_suggest which are automatically loaded at the interpreter startup.

This example is specifying an older version of did_you_mean, but typically you will want to specify a newer version. Actually, in https://github.com/rails/rails/pull/45818, I wanted to use error_highlight 0.4.0 with Ruby 3.1. (Note that Ruby 3.1 bundles with error_highlight 0.3.0.)

The cause of this problem is that `bundle exec` makes the interpreter load `bundler/setup` using `RUBYOPT=-rbundler/setup`, but this load is too late.

When ruby is invoked via `bundle exec`, rubygems.rb should `require "bundler/setup"` (i.e., before the special gems are loaded).

More specifically: `bundle exec` sets an environment variable `BUNDLER_SETUP`, and rubygems.rb requires the variable if defined.

(I am not particular about the name of this environment variable. I will be happy to change it if you prefer another name.)

See also: https://bugs.ruby-lang.org/issues/19089

### DIFF
--- a/bundler/spec/bundler/shared_helpers_spec.rb
+++ b/bundler/spec/bundler/shared_helpers_spec.rb
@@ -247,7 +247,7 @@ RSpec.describe Bundler::SharedHelpers do
     end
 
     shared_examples_for "ENV['BUNDLER_SETUP'] gets set correctly" do
-      it "ensures bundler/setup is set in ENV['BUNDLE_SETUP']" do
+      it "ensures bundler/setup is set in ENV['BUNDLER_SETUP']" do
         subject.set_bundle_environment
         expect(ENV["BUNDLER_SETUP"]).to eq("#{source_lib_dir}/bundler/setup")
       end
--- a/bundler/spec/bundler/shared_helpers_spec.rb
+++ b/bundler/spec/bundler/shared_helpers_spec.rb
@@ -246,6 +246,13 @@ RSpec.describe Bundler::SharedHelpers do
       end
     end
 
+    shared_examples_for "ENV['BUNDLER_SETUP'] gets set correctly" do
+      it "ensures bundler/setup is set in ENV['BUNDLE_SETUP']" do
+        subject.set_bundle_environment
+        expect(ENV["BUNDLER_SETUP"]).to eq("#{source_lib_dir}/bundler/setup")
+      end
+    end
+
     shared_examples_for "ENV['RUBYLIB'] gets set correctly" do
       let(:ruby_lib_path) { "stubbed_ruby_lib_dir" }
 
--- a/bundler/lib/bundler/environment_preserver.rb
+++ b/bundler/lib/bundler/environment_preserver.rb
@@ -7,6 +7,7 @@ module Bundler
       BUNDLE_BIN_PATH
       BUNDLE_GEMFILE
       BUNDLER_VERSION
+      BUNDLER_SETUP
       GEM_HOME
       GEM_PATH
       MANPATH
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -284,6 +284,7 @@ module Bundler
       Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", exe_file
       Bundler::SharedHelpers.set_env "BUNDLE_GEMFILE", find_gemfile.to_s
       Bundler::SharedHelpers.set_env "BUNDLER_VERSION", Bundler::VERSION
+      Bundler::SharedHelpers.set_env "BUNDLER_SETUP", File.expand_path("setup", __dir__)
     end
 
     def set_path
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -1348,3 +1348,5 @@ Gem::Specification.load_defaults
 require_relative "rubygems/core_ext/kernel_gem"
 require_relative "rubygems/core_ext/kernel_require"
 require_relative "rubygems/core_ext/kernel_warn"
+
+require ENV["BUNDLER_SETUP"] if ENV["BUNDLER_SETUP"]


jeremyevans:
Is it possible to fix this in bundler instead of rubygems, by having bundler set `RUBYLIB` in addition to `RUBYOPT`?  If so, I would prefer that approach, instead of modifying rubygems to treat bundler specially.

bundler/spec/bundler/shared_helpers_spec.rb
```diff
@@ -246,6 +246,13 @@
       end
     end
 
+    shared_examples_for "ENV['BUNDLER_SETUP'] gets set correctly" do
+      it "ensures bundler/setup is set in ENV['BUNDLE_SETUP']" do
```
nobu, line 250:
```suggestion
      it "ensures bundler/setup is set in ENV['BUNDLER_SETUP']" do
```

mame:
> Is it possible to fix this in bundler instead of rubygems, by having bundler set `RUBYLIB` in addition to `RUBYOPT`?

@jeremyevans I don't think so. What we need here is doing `require "bundler/setup"` *before* `require "error_highlight"` and `require "did_you_mean"` in gem_prelude.rb. `RUBYOPT` is evaluated after gem_prelude.rb, so I don't think any use of `RUBYOPT` will solve this.

`RUBYLIB` may make it easier to resolve the path to `bundler/setup`, but anyway, it is necessary to call `require "bundler/setup"` in rubygems.rb, I think.

jeremyevans:
> > Is it possible to fix this in bundler instead of rubygems, by having bundler set `RUBYLIB` in addition to `RUBYOPT`?
> 
> @jeremyevans I don't think so. What we need here is doing `require "bundler/setup"` _before_ `require "error_highlight"` and `require "did_you_mean"` in gem_prelude.rb. `RUBYOPT` is evaluated after gem_prelude.rb, so I don't think any use of `RUBYOPT` will solve this.

Assuming bundler sets `RUBYLIB` and includes the path to `bundler`, `error_highlight`, `did_you_mean`, I would think that would fix the requiring of `error_highlight` and `did_you_mean`.  Is there a reason it wouldn't work?  It appears to work:

```
$ ruby -e 'p DidYouMean::VERSION'
"1.6.1"
$ RUBYLIB=/usr/local/lib/ruby/gems/3.1/gems/did_you_mean-1.5.0/lib/ ruby -e 'p DidYouMean::VERSION'
"1.5.0"
```

mame:
Thank you, I think I understand your proposal.

I am not that familiar with the code of bundler, so I could be wrong, but I think this requires a big redesign of the implementation strategy of bundler.

Here is my understanding. Currently `bundle exec` does not interpret Gemfile, but simply sets `RUBYOPT=-rbundler/setup` and invokes the ruby command. Also, `bundler/setup` interprets Gemfile, but does not set `RUBYLIB`, instead, overrides `Kernel#require` function to load the specified gem version. (Again, I could be wrong!)

To implement your proposal, I think we first need to rewrite the bundle exec to interpret Gemfile, and also to make the Ruby interpreter load the specified gem version by using `RUBYLIB` instead of overriding `Kernel#require`. It might be theoretically possible, but this would be a substantial change for bundler and would probably result in a number of minor incompatibilities.

hsbt (approved):

deivid-rodriguez:
@hsbt I'm not convinced this is the **best solution** to this problem, but if we're going with this, we should at least remove the following exemption from our specs to at least prove that this is **a solution**:

https://github.com/rubygems/rubygems/blob/ee2f8398324af4bc1b95f7565ce2fda98126e026/bundler/spec/runtime/setup_spec.rb#L1305-L1308

mame:
> @hsbt I'm not convinced this is the **best solution** to this problem

In the long term, the best solution is to fix the fundamental mechanism of bundler as Jeremy suggested, IMO. I think it would require much time and effort, though. I hope this fix is a good compromise as a short-term solution.

eregon:
Isn't this also a potential security concern/issue? It means if BUNDLER_SETUP is set in the environment (by anything, including a Ruby parent process, etc), it makes loading RubyGems load a attacker-specified file.

deivid-rodriguez:
I think we can at least improve this feature by leveraging the `Gem.use_gemdeps` feature that optionally requires `bundler/setup` when the `RUBYGEMS_GEMDEPS` environment variable is set, instead of introducing a new environment variable.

deivid-rodriguez:
My idea was to call `Gem.use_gemdeps` unconditionally after loading `lib/rubygems.rb`. That method already has logic to require `bundler/setup` if `RUBYGEMS_GEMDEPS` is set.

However, we used to do that in the past but caused some issues when running `bundler` itself, so I moved it to the RubyGems generated binstus to be able to skip it for Bundler at https://github.com/rubygems/rubygems/pull/4532.

I need to allocate some time to find a solution for this, I'm not really happy about the current approach.

hsbt:
> I'm not really happy about the current approach.

I see. The current implementation for only 3.2.0-preview3.

The ruby core have a plan to release all of stable versions next week. If we can fixed this and release RG 3.2.x and 3.3.x, we will merge this to the next stable versions.

But I'm not familier with `Gem.use_gemdeps` and its problem. We will forcus to the next preview version of 3.2.0 and next stable releases in 2023. Thanks for yoru comment.

deivid-rodriguez:
I'm saw a first issue that seems due to this: https://github.com/ruby/ruby/commit/e5a0abc5dedfd011986b16e8f8cf5cda476984c9.

Could we revert this for the time being until we find a solution we're all comfortable with?

mame:
I have created a PR to fix the issue in erb: https://github.com/rubygems/rubygems/pull/6080

mensfeld:
> Isn't this also a potential security concern/issue? It means if BUNDLER_SETUP is set in the environment (by anything, including a Ruby parent process, etc), it makes loading RubyGems load a attacker-specified file.

It could be. Hard for me to assess the risk level, but this introduced a risk factor. Adding this to ENV and planting a seed script to do something is one thing I could think of atm.

PR #6026, opened 2022-10-28
github source should default to secure protocol

jasonkarns:
Bundler 2 switched to secure https here https://github.com/rubygems/rubygems/commit/c2e81f8ff63613871cc8b52653c5e176f8dafde3

Insecure protocols should be avoided to prevent MITM attacks.

I discovered the differing implementation that was fixed "upstream" in bundler itself. I haven't encountered a _user_ side problem to demonstrate this; it was merely a source-code detection.

Simple fix is the same as before: use secure protocol https and bring in-line consistently with bundler.

### DIFF
--- a/bundler/tool/bundler/release_gems.rb.lock
+++ b/bundler/tool/bundler/release_gems.rb.lock
@@ -60,6 +60,7 @@ PLATFORMS
   universal-java-11
   universal-java-18
   x86_64-darwin-20
+  x86_64-darwin-21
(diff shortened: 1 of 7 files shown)


hsbt:
👍 Can you also update related tests?

hsbt:
> it appears that I'm the first to add macOS Monterey to the lockfiles?

It's no problem. we added it by the first touch developper.

Issue #6016, opened 2022-10-25
`bundle init` should respect `BUNDLE_GEMFILE`


dentarg:
### Describe the problem as clearly as you can

```
$ bundle -v
Bundler version 2.3.24

$ BUNDLE_GEMFILE=foo bundle init
Writing new Gemfile to /tmp/Gemfile
```

I was expecting `/tmp/foo` to be created as `bundle config --help` says this about `BUNDLE_GEMFILE`

>The name of the file that bundler should use as the Gemfile. This location of this file also sets the root of the project, which is used to resolverelative paths in the Gemfile, among other things. By default, bundler will search up from the current working directory until it finds a Gemfile.

gustavothecoder:
Can I try to resolve this problem?

dentarg:
Feel free :)

manuraj17:
Can this issue be considered as closed with this PR https://github.com/rubygems/rubygems/pull/6046 @deivid-rodriguez @dentarg ?

deivid-rodriguez:
Indeed, thanks for noticing @manuraj17!

PR #6046, opened 2022-11-07
Add `--gemfile` flag to `bundle init` to configure gemfile name to generate 

gustavothecoder:

When running `bundle init` with `BUNDLE_GEMFILE`, the gemfile name is still the original (`Gemfile`) instead of the custom defined in `BUNDLE_GEMFILE`.

For more details, see the issue #6016.

The main problem here is that `Bundler#preferred_gemfile_name` doesn't consider the `BUNDLE_GEMFILE` config, so my first change was adjust this method to use the `BUNDLE_GEMFILE` when necessary. **I changed the method to give priority to the `init_gems_rb` config**, so if user uses both `init_gems_rb` and `BUNDLE_GEMFILE`, the first one will be used. I decide to make this way to preserve the original behavior, but I'm not sure if this approach is ideal and I would like to know what you guys think. I think that we'll need to update the docs with this behavior, by the way.

The second change I made was to adjusting the `init` command to use some template when `BUNDLE_GEMFILE` is used. This was necessary because the template is chosen based on the gemfile name and the current implementation doesn't consider the case where the gemfile has a custom name. **I noticed that all available templates has the same content, is it feasible to merge them into a single template?** Because that would simplify the code a bit.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -156,6 +156,7 @@ module Bundler
       dependency listed in the gemspec file to the newly created Gemfile.
     D
     method_option "gemspec", :type => :string, :banner => "Use the specified .gemspec to create the Gemfile"
+    method_option "gemfile", :type => :string, :banner => "Use the specified name for the gemfile instead of 'Gemfile'"
     def init
       require_relative "cli/init"
       Init.new(options.dup).run
--- a/bundler/lib/bundler/cli/init.rb
+++ b/bundler/lib/bundler/cli/init.rb
@@ -32,7 +32,7 @@ module Bundler
           file << spec.to_gemfile
         end
       else
-        File.open(File.expand_path("../templates/#{gemfile}", __dir__), "r") do |template|
+        File.open(File.expand_path("../templates/Gemfile", __dir__), "r") do |template|
           File.open(gemfile, "wb") do |destination|
             IO.copy_stream(template, destination)
           end
@@ -45,7 +45,7 @@ module Bundler
     private
 
     def gemfile
-      @gemfile ||= Bundler.preferred_gemfile_name
+      @gemfile ||= options[:gemfile] || Bundler.preferred_gemfile_name
     end
   end
 end
--- a/bundler/lib/bundler/man/bundle-init.1.ronn
+++ b/bundler/lib/bundler/man/bundle-init.1.ronn
@@ -16,6 +16,8 @@ created [`Gemfile(5)`][Gemfile(5)].
 
 * `--gemspec`:
   Use the specified .gemspec to create the [`Gemfile(5)`][Gemfile(5)]
+* `--gemfile`:
+  Use the specified name for the gemfile instead of `Gemfile`
 
 ## FILES
 
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -205,7 +205,6 @@ bundler/lib/bundler/templates/Executable
 bundler/lib/bundler/templates/Executable.bundler
 bundler/lib/bundler/templates/Executable.standalone
 bundler/lib/bundler/templates/Gemfile
-bundler/lib/bundler/templates/gems.rb
 bundler/lib/bundler/templates/newgem/CHANGELOG.md.tt
 bundler/lib/bundler/templates/newgem/CODE_OF_CONDUCT.md.tt
 bundler/lib/bundler/templates/newgem/Cargo.toml.tt
--- a/bundler/lib/bundler/man/bundle-add.1
+++ b/bundler/lib/bundler/man/bundle-add.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-ADD" "1" "January 2023" "" ""
+.TH "BUNDLE\-ADD" "1" "February 2023" "" ""
 .
 .SH "NAME"
 \fBbundle\-add\fR \- Add gem to the Gemfile and run bundle install
--- a/bundler/lib/bundler/man/bundle-binstubs.1
+++ b/bundler/lib/bundler/man/bundle-binstubs.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-BINSTUBS" "1" "January 2023" "" ""
+.TH "BUNDLE\-BINSTUBS" "1" "February 2023" "" ""
 .
 .SH "NAME"
 \fBbundle\-binstubs\fR \- Install the binstubs of the listed gems
--- a/bundler/lib/bundler/man/bundle-cache.1
+++ b/bundler/lib/bundler/man/bundle-cache.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CACHE" "1" "January 2023" "" ""
+.TH "BUNDLE\-CACHE" "1" "February 2023" "" ""
 .
 .SH "NAME"
 \fBbundle\-cache\fR \- Package your needed \fB\.gem\fR files into your application
--- a/bundler/lib/bundler/man/bundle-check.1
+++ b/bundler/lib/bundler/man/bundle-check.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CHECK" "1" "January 2023" "" ""
+.TH "BUNDLE\-CHECK" "1" "February 2023" "" ""
 .
 .SH "NAME"
 \fBbundle\-check\fR \- Verifies if dependencies are satisfied by installed gems
--- a/bundler/lib/bundler/man/bundle-clean.1
+++ b/bundler/lib/bundler/man/bundle-clean.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CLEAN" "1" "January 2023" "" ""
+.TH "BUNDLE\-CLEAN" "1" "February 2023" "" ""
 .
 .SH "NAME"
 \fBbundle\-clean\fR \- Cleans up unused gems in your bundler directory
--- a/bundler/lib/bundler/man/bundle-config.1
+++ b/bundler/lib/bundler/man/bundle-config.1
@@ -1,7 +1,7 @@
 .\" generated with Ronn/v0.7.3
 .\" http://github.com/rtomayko/ronn/tree/0.7.3
 .
-.TH "BUNDLE\-CONFIG" "1" "January 2023" "" ""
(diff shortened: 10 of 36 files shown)


deivid-rodriguez:
Thank you @gustavothecoder!

Any opinions on what the behavior should be if the `BUNDLE_GEMFILE` set already exists? Should we keep the current behavior with a warning in that case? @dentarg what do you think?

bundler/lib/bundler/cli/init.rb
```diff
@@ -32,20 +32,36 @@ def run
...
+      template_folder = "../templates/"
+      if File.exist?(template_folder + gemfile)
+        template_folder + gemfile
+      else
+        template_folder + "Gemfile"
+      end
+    end
+
+    def print_feedback
+      # We need this because when the key BUNDLE_GEMFILE is seted #gemfile will
```
simi, line 61: :scissors: `s/seted/set/g` :question:

gustavothecoder: I don't understand, could you explain in another way, please?

gustavothecoder:
> Thank you @gustavothecoder!
> 
> Any opinions on what the behavior should be if the `BUNDLE_GEMFILE` set already exists? Should we keep the current behavior with a warning in that case? @dentarg what do you think?

You mean when the user sets the `BUNDLE_GEMFILE` using the `bundle config`? I think we should use the `BUNDLE_GEMFILE` defined in runtime.

dentarg:
> Any opinions on what the behavior should be if the BUNDLE_GEMFILE set already exists? Should we keep the current behavior with a warning in that case? @dentarg what do you think?

I think we should exit with error saying the file already exists, like we do today

deivid-rodriguez:
Some environments have a fixed `BUNDLE_GEMFILE`, and on top of that, when `bundle exec` is run, `BUNDLE_GEMFILE` is fixed to the current Gemfile for the subprocess. Also, Bundler generated binstubs set `BUNDLE_GEMFILE` right at the top.

I'm afraid this may break `bundle init` processes inside user scripts due to the above.

That's why I was suggesting a warning if `BUNDLE_GEMFILE` has a non default, existing, value, and fallback to the current behavior of creating a file named `Gemfile`.

deivid-rodriguez:
Yes, in these simple examples feeding `BUNDLE_GEMFILE` directly, it feels a bit dumb, but sometimes things are messy/complicated and it's actually hard to tell how/where/what set/exported `BUNDLE_GEMFILE` in the first place.

bundler/lib/bundler/cli/init.rb
gustavothecoder: Sorry, I'm still learning English (my main language is Portuguese), I'll correct it.

gustavothecoder:
Just to confirm if I understand correctly. When `BUNDLE_GEMFILE` already exists and is different from the default gemfile name, we'll print a warning and create another gemfile with the default name (`Gemfile`), that's it?

deivid-rodriguez:
@gustavothecoder Yes, that's my suggestion.

gustavothecoder:
@deivid-rodriguez I think it's working fine now:

https://user-images.githubusercontent.com/57065994/200973024-922389f8-ab7d-4aaa-9cee-59ebaa05f039.mp4

deivid-rodriguez:
Looks good!

I was also thinking that warnings should be actionable, so we should clearly explain to users how to fix the warning.

The solution would be to pass `BUNDLE_GEMFILE=Gemfile` explicitly to the subcommand, but depending on the situation it might not be so easy, maybe we should add a explicit `--gemfile` flag to `bundle init`?

Or maybe this is all a bit overkill and I'm overthinking this. I'd like to hear opinions from other maintainers.

dentarg:
> maybe we should add a explicit `--gemfile` flag to `bundle init`?

That makes sense to me. I have actually looked for it, and I almost thought I found it 😆 but then I noticed that it said `--gemspec` 😓

gustavothecoder:
> maybe we should add a explicit `--gemfile` flag to `bundle init`?

That makes sense for me too, but, are we going to support both `BUNDLE_GEMFILE` and `--gemfile`?

@deivid-rodriguez what do you think of this message?: "Bundler will name the gemfile "Gemfile" because foo already exists at /some/dir/foo. To silence this warning, run `bundle init --gemfile Gemfile`."

deivid-rodriguez:
@gustavothecoder Sorry for the lack of feedback.

Based on the discussion, I think the easiest approach here may be to add a `--gemfile` option and don't change anything regarding `BUNDLE_GEMFILE` here, i.e., ignore it like we've been doing till now?

gustavothecoder:
> @gustavothecoder Sorry for the lack of feedback.
> 
> Based on the discussion, I think the easiest approach here may be to add a `--gemfile` option and don't change anything regarding `BUNDLE_GEMFILE` here, i.e., ignore it like we've been doing till now?

That make sense. I'll change the fix.

gustavothecoder:
@deivid-rodriguez do you think we can improve the implementation?

deivid-rodriguez:
Sorry for the lack of feedback here @gustavothecoder, I'll try to review this soon.

bundler/lib/bundler/cli/init.rb
```diff
@@ -27,25 +28,43 @@ def run
...
+                        "#{gemfile_name} already exists at #{File.expand_path(gemfile_name)}. " \
+                        "To silence this warning, run `bundle init --gemfile Gemfile`."
+        "Gemfile"
+      else
+        gemfile_name
+      end
+    end
+
+    def template_path
+      template_folder = "../templates/"
```
K-S-A, line 62:
```suggestion
      template_file = "../templates/#{@gemfile}"
      return template_file if File.exist?(template_file)
      
      "../templates/Gemfile"
```

gustavothecoder: Thanks for the review!

bundler/lib/bundler/cli/init.rb
```diff
@@ -27,25 +28,41 @@ def run
...
 
-    def gemfile
-      @gemfile ||= Bundler.preferred_gemfile_name
+    def preferred_gemfile_name
+      gemfile_name = options[:gemfile] || Bundler.preferred_gemfile_name
+
+      if File.exist?(gemfile_name) && gemfile_name != "Gemfile"
+        Bundler.ui.warn "Bundler will name the gemfile \"Gemfile\" because " \
+                        "#{gemfile_name} already exists at #{File.expand_path(gemfile_name)}. " \
+                        "To silence this warning, run `bundle init --gemfile Gemfile`."
```
deivid-rodriguez, line 54: Sorry for going around in circles with this, but I'm not sure this deprecation makes sense anymore, if we add a CLI flag? Maybe it's better to keep the current behavior of always generating a file named `Gemfile`, regardless of `BUNDLE_GEMFILE`, unless `--gemfile` is given.

bundler/lib/bundler/cli.rb
```diff
@@ -156,6 +156,7 @@ def self.handle_no_command_error(command, has_namespace = $thor_runner)
       dependency listed in the gemspec file to the newly created Gemfile.
     D
     method_option "gemspec", :type => :string, :banner => "Use the specified .gemspec to create the Gemfile"
+    method_option "gemfile", :type => :string, :banner => "Use the specified gemfile instead of Gemfile"
```
deivid-rodriguez, line 159: I'd try to make it more clear that you're supposed to give this a _name_ for the Gemfile to be generated (that the command never expects a file that already exists).

```suggestion
    method_option "gemfile", :type => :string, :banner => "Use the specified name for the gemfile instead of 'Gemfile'"
```

bundler/lib/bundler/cli/init.rb
deivid-rodriguez: Was this a bug, or something else? If a bug that wasn't caught by CI, would be nice to add an extra spec.

gustavothecoder: No, it's not a bug. It's just a simpler way to set the template file.

deivid-rodriguez: Thanks for clarifying!

bundler/lib/bundler/cli/init.rb
gustavothecoder: I like the idea of having only one way of doing things. So it makes sense for me to ignore `BUNDLE_GEMFILE` and only consider `--gemfile`.

deivid-rodriguez: Alright, in that case I think maybe all the changes in this file can be removed and only change 

```ruby
      @gemfile ||= Bundler.preferred_gemfile_name
```

to

```ruby
      @gemfile ||= options[:gemfile] || Bundler.preferred_gemfile_name
```

gustavothecoder: OK

gustavothecoder: Done.

bundler/lib/bundler/cli/init.rb
```diff
@@ -27,25 +28,28 @@ def run
...
       end
 
-      puts "Writing new #{gemfile} to #{SharedHelpers.pwd}/#{gemfile}"
+      puts "Writing new #{@gemfile} to #{SharedHelpers.pwd}/#{@gemfile}"
     end
 
     private
 
-    def gemfile
-      @gemfile ||= Bundler.preferred_gemfile_name
```
deivid-rodriguez, line 48: Any reason for moving this to `initialize` and change every reference from using an accessor to using the instance variable directly?

deivid-rodriguez (commented):
Looks mostly ready, just one more comment.

Thanks for your infinite patience here, I think it was this long discussion because this approach finally seems the simplest way to go to me!

bundler/lib/bundler/cli/init.rb
gustavothecoder: This was needed in the last approach, I just forgot to remove 😅

I'll remove right now.

bundler/lib/bundler/cli/init.rb
```diff
@@ -32,7 +32,7 @@ def run
           file << spec.to_gemfile
         end
       else
-        File.open(File.expand_path("../templates/#{gemfile}", __dir__), "r") do |template|
+        File.open(File.expand_path(template_path, __dir__), "r") do |template|
```
deivid-rodriguez, line 35: We currently have two gemfile templates: `gems.rb` and `Gemfile`, but they have the same content. I think we can use `../templates/Gemfile` unconditionally here and remove the `gems.rb` template and the `template_path` method?

gustavothecoder: We can. I'll do it.

bundler/lib/bundler/man/bundle-init.1.ronn
```diff
@@ -16,6 +16,8 @@ created [`Gemfile(5)`][Gemfile(5)].
 
 * `--gemspec`:
   Use the specified .gemspec to create the [`Gemfile(5)`][Gemfile(5)]
+* `--gemfile`:
+  Use the specified gemfile instead of the [`Gemfile(5)`][Gemfile(5)]
```
deivid-rodriguez, line 20: I'll clarify this help to make it explicit that the flag takes a explicit name like in https://github.com/rubygems/rubygems/pull/6046#discussion_r1085935335, but I'll do it quickly myself while giving this a final rebase.

deivid-rodriguez (approved):
Looks great, just one more small comment but I'll fix it myself when rebasing this! Thanks so much @gustavothecoder!

PR #6049, opened 2022-11-08
Speed up JRuby Rails test

Bo98:
This skips the Rails 6 default `webpacker:install` step, which is slow, invokes yarn installs and not particularly relevant to anything on the RubyGems side.

This does not remove webpacker as a Ruby dependency, so the dependency tree tested here is unchanged.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -97,8 +97,8 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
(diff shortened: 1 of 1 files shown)


Bo98:
There's still a 2 minute hang during the `bundle install`, but it should overall not be near the 10 minute limit now.

deivid-rodriguez (approved):
Thanks!

I'm quite tempted to remove this check on JRuby, but let's try this first!

PR #6050, opened 2022-11-08
Fix '--force' option documentation of 'bundle clean'

syohex:

`--force` option documentation of `bundle clean` is wrong. Because `--path` option is no longer used. The following code checks if system location is used or not.

https://github.com/syohex/rubygems/blob/clean-doc/bundler/lib/bundler/cli/clean.rb#L19

Fix the documentation according to the current implementation.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -620,7 +620,7 @@ module Bundler
     method_option "dry-run", :type => :boolean, :default => false, :banner =>
       "Only print out changes, do not clean gems"
     method_option "force", :type => :boolean, :default => false, :banner =>
-      "Forces clean even if system location is used"
+      "Forces cleaning up unused gems even if Bundler is configured to use globally installed gems. As a consequence, removes all system gems except for the ones in the current application."
     def clean
       require_relative "cli/clean"
       Clean.new(options.dup).run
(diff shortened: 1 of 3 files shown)


bundler/lib/bundler/cli.rb
```diff
@@ -620,7 +620,7 @@ def self.source_root
     method_option "dry-run", :type => :boolean, :default => false, :banner =>
       "Only print out changes, do not clean gems"
     method_option "force", :type => :boolean, :default => false, :banner =>
-      "Forces clean even if --path is not set"
+      "Forces clean even if system location is used"
```
deivid-rodriguez, line 623: I think "system location" may still not be 100% clear. Maybe we could elaborate a bit more, like

> Forces cleaning up unused gems even if Bundler is configured to install and look for gems globally, and as a consequence, all system gems except for the one in the current application would be removed.

Perhaps other maintainers can propose a better wording that explains this more concisely.

deivid-rodriguez (commented):
Thanks for catching this. Still not sure if the new wording could still be more clear, but this is already an improvement!

bundler/lib/bundler/cli.rb
deivid-rodriguez: @syohex Thanks for proposing this even if not familiar with Bundler. I tweaked my suggested warning a bit, can you update this PR to use that?

syohex: @deivid-rodriguez I have updated at 7c54dc56f651f93d37f8b6329e893211d5ca2e9b

deivid-rodriguez (approved):
Thank you ❤️

PR #6052, opened 2022-11-09
Don't set BUNDLER_ORIG environment variables

peterzhu2118:

Bundler creates a backup of environment variables by creating new environment variables prefixed with `BUNDLER_ORIG_`. This caused Ruby test suite's environment variable leak checker to report lots of created/changed environment variables (suppressed in commit ruby/ruby@6d835901575d58e7db404665801a1c455ee982a8). 

This PR stops creating `BUNDLER_ORIG_` environment variables and instead relies on storing backups inside Ruby at boot.

However, this implementation causes a breaking change when spawning child processes. For example:

```ruby

require "bundler"

ENV["BUNDLE_BIN_PATH"] = "foobar"

script = <<~RUBY
  require "bundler"

  Bundler.with_original_env do
    p ENV["BUNDLE_BIN_PATH"]
  end
RUBY

puts `bundle exec ruby -e '#{script}'`
```

Before this change it outputs `nil` while after this change it outputs `"foobar"`.


hsbt:
👋  Dropping the `BUNDLER_ORIG_*` vars breaks the cross-process semantics of `with_original_env`/`with_unbundled_env` (your own example shows it).

Since this is really about quieting the test helper's env-leak checker, and `ruby/ruby` has lived fine with the suppression for ~4 years, the compatibility cost feels too steep.

Closing it, but thanks for your investigation.

PR #6068, opened 2022-11-18
Improve some resolution messages when some platform gems are missing

deivid-rodriguez:

This should've been part of the initial PubGrub work, but I took a shortcut to get a green baseline and in the end I forgot to get the related spec passing the way I wanted.

To spec is the following. The first message is the old one given by Molinillo, the second is the one given by PubGrub.

https://github.com/rubygems/rubygems/blob/7ff32ade207cb35ce4d46b8d571d1b4e1545792c/bundler/spec/install/gemfile/specific_platform_spec.rb#L365-L382

The idea is to give a hint like we did before when resolution failed because we couldn't find specs matching the requirements for some resolution platform. The current message does not give any hint about this, so it's confusing.

My fix is to allow further customization of PubGrub incompatibilities to be able to give an extra hint about this edge case.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -168,6 +168,7 @@ bundler/lib/bundler/remote_specification.rb
 bundler/lib/bundler/resolver.rb
 bundler/lib/bundler/resolver/base.rb
 bundler/lib/bundler/resolver/candidate.rb
+bundler/lib/bundler/resolver/incompatibility.rb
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
Ready!

simi (approved):

deivid-rodriguez:
Thanks @simi!

PR #6071, opened 2022-11-19
Fx link in bundle-platform man page

chtitux:

At the end of the page https://bundler.io/man/bundle-platform.1.html , the link to `bundle-lock(1)` points to `https://bundler.io/man/bundle-lock.1.ronn`, which is not correct.

Internal links should use `.html` extension, instead of `.ronn`.


deivid-rodriguez (approved):
Thanks so much for this @chtitux!

PR #6087, opened 2022-11-29
Use latest JRuby in CI

deivid-rodriguez:

No real problem, but it's good to keep dependencies up to date.

Use latest JRuby in CI.


deivid-rodriguez:
This PR found two [JRuby](https://github.com/jruby/jruby/issues/7499) [bugs](https://github.com/jruby/jruby/issues/7500). They are already fixed in JRuby so we should be good to upgrade when they release the next version.

deivid-rodriguez:
Still getting at least one of those bugs after upgrading to JRuby 9.4.1.0.

deivid-rodriguez:
I _think_ 9.4.2.0 will do the trick! 🤞

PR #4469, opened 2021-03-18
Fix git source cache being used as the install location 

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Bundler app-cache of git gems behaving unexpectedly.

## What is your fix for the problem, implemented in this PR?

Make cache for git gems behave like a proper cache. This means that now git gems are properly installed in presence of a project gem cache (for example, extension gems get properly compiled). Since the cache now acts as a cache, we no longer need to keep full clones in cache and a bare clone is enough.

Fixes #3204.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

Issue #3204, opened 2017-02-02
bundle install is not installing project vendor cached git gems, but using the cache as the install location


bughit:
My understanding is that the cache is only supposed to be used as a source for installation of gems, and not function as the install location, from which the code is ultimately loaded.

The normal git gem system install location is `~/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bundler/`, so when I bundle install (normally, without --path or --deployment) a gem that is cached in the project's vendor/cache I expect it to end up in `~/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bundler/`

It doesn't, and `bundle show` shows its location in `$PROJECT/vendor/cache/`.

If it's removed from  `$PROJECT/vendor/cache/`  and bundle reinstalled, it does get installed to `~/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bundler/` but `bundle show` still shows `$PROJECT/vendor/cache/`

Is this expected behavior?  Does not seem like a good idea.

dentarg:
I just ran into this, so I checked if it was still a problem with the latest released Bundler, 2.1.4 (and RubyGems 3.1.2), and it is. [Recorded the steps](https://github.com/dentarg/puma-vagrant/commits/master/vendoring_master_app)  to reproduce [in git](https://github.com/dentarg/puma-vagrant/tree/a83aa96eabb6a316636b412ef0cc6f84f06e0c58/vendoring_master_app).

I understand that this problem is somewhat of a niche use-case, but just wanted to post some up-to-date info on this, seeing the issue hasn't received any comments in three years.

thedanbob:
I've been banging my head against this bug for two days. Running a second `bundle install` doesn't fix the problem for me. I've only found two ways to properly install a git gem in deployment mode:

1. Manually create a .gem file and copy it into vendor/cache, then run `bundle install`
2. Manually copy the gem directory from vendor/cache to vendor/bundle/ruby/2.x.0/bundler/gems/

Oddly enough I don't have a problem with bundler 1.x, `bundle install --deployment --local` installs git gems correctly for me. Only 2.1.4 is broken.

deivid-rodriguez:
I had a quick look. Could you try [this patch](https://github.com/rubygems/rubygems/compare/try_fix_git_cache_issue) and see if it fixes the issue?

thedanbob:
You're my hero! `bundle install` and `bundle package` work as expected with that patch.

[...]

PR #6088, opened 2022-11-30
Respect --no-install option for git: sources

KJTsanaktsidis:

Currently, the --no-install option to `bundle package` is totally ignored for git sources. This can have very strange effects if you have:

- a git-sourced gem,
- with native extensions,
- whose extconf.rb script depends on another gem,
- which is installed from Rubygems in the gemfile.

In that circumstance, `bundle package --no-install --all` will download the Rubygems dependencies to `vendor/cache` but NOT install them. It will also check out the git gems to `vendor/cache` (good), and attempt to build their native extensions (bad!).

The native extension build will fail because the extconf.rb script crashes, since the dependency it needs is missing.

I implemented a fix for this in `source/git.rb` as an early return if `Bundle.settings[:no_install]`, since this is analogous to what's happening in `source/rubygems.rb`.

I do admit though the whole thing is a little strange though - an "install" method that.... proceeds to look at a global flag to not install anything.

At this point, if you think this is the right solution for the problem, I can whip up a test and then get this merged. If you think there's a better way to factor this package functionality though I'm all ears.

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -173,6 +173,7 @@ module Bundler
       end
 
       def install(spec, options = {})
+        return if Bundler.settings[:no_install]
         force = options[:force]
 
         print_using_message "Using #{version_message(spec, options[:previous_spec])} from #{self}"
--- a/bundler/spec/cache/git_spec.rb
+++ b/bundler/spec/cache/git_spec.rb
@@ -220,4 +220,57 @@ RSpec.describe "bundle cache with git" do
       expect(the_bundle).to include_gem "foo 1.0"
     end
   end
+
+  it "respects the --no-install flag" do
+    git = build_git "foo", &:add_c_extension
+    ref = git.ref_for("main", 11)
+
+    gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem "foo", :git => '#{lib_path("foo-1.0")}'
+    G
+    bundle "config set cache_all true"
+
+    # The algorithm for the cache location for a git checkout is
+    # in Bundle::Source::Git#cache_path
+    cache_path_name = "foo-1.0-#{Digest(:SHA1).hexdigest(lib_path("foo-1.0").to_s)}"
+
+    # Run this test twice. This is because materially different codepaths
+    # will get hit the second time around.
+    # The first time, Bundler::Sources::Git#install_path is set to the system
+    # wide cache directory bundler/gems; the second time, it's set to the
+    # vendor/cache directory. We don't want the native extension to appear in
+    # either of these places, so run the `bundle cache` command twice.
+    2.times do
+      bundle :cache, "all-platforms" => true, :install => false
+
+      # it did _NOT_ actually install the gem - neither in $GEM_HOME (bundler 2 mode),
+      # nor in .bundle (bundler 3 mode)
+      expect(Pathname.new(File.join(default_bundle_path, "gems/foo-1.0-#{ref}"))).to_not exist
+      # it _did_ cache the gem in vendor/
+      expect(bundled_app("vendor/cache/foo-1.0-#{ref}")).to exist
+      # it did _NOT_ build the gems extensions in the vendor/ dir
+      expect(Dir[bundled_app("vendor/cache/foo-1.0-#{ref}/lib/foo_c*")]).to be_empty
+      # it _did_ cache the git checkout
+      expect(default_cache_path("git", cache_path_name)).to exist
+      # And the checkout is a bare checkout
+      expect(default_cache_path("git", cache_path_name, "HEAD")).to exist
+    end
+
+    # Subsequently installing the gem should compile it.
+    # _currently_, the gem gets compiled in vendor/cache, and vendor/cache is added
+    # to the $LOAD_PATH for git extensions, so it all kind of "works". However, in the
+    # future we would like to stop adding vendor/cache to the $LOAD_PATH for git extensions
+    # and instead treat them identically to normal gems (where the gem install location,
+    # not the cache location, is added to $LOAD_PATH).
+    # Verify that the compilation worked and the result is in $LOAD_PATH by simply attempting
+    # to require it; that should make sure this spec does not break if the load path behaviour
+    # is changed.
+    bundle :install, :local => true
+    ruby <<~R, :raise_on_error => false
+      require 'bundler/setup'
+      require 'foo_c'
+    R
+    expect(last_command).to_not be_failure
+  end
 end


deivid-rodriguez:
I think the underlying root cause of this is related to https://github.com/rubygems/rubygems/issues/3204?

KJTsanaktsidis:
_oof_. That issue seems like a lot. Sorry you had to deal with that :heart: 

So yeah, @thedanbob already noticed this: https://github.com/rubygems/rubygems/issues/3204#issuecomment-797170223

> I figured out why the test didn't fail: bundle cache --no-install actually does install git gems (and only git gems), and to the correct location of vendor/bundle/ruby/2.5.0/bundler/gems/. So there's no "Fetching' in the later bundle install because it's already installed!
>
> I guess I should open another issue for bundle cache not respecting the --no-install flag for git gems

--------------------------

I think fixing the "the install path is the same as the cache path" is a necessary but not sufficient condition. If my reading of what the rubygems source is doing is correct, this is what I think the expected behaviour _should_ be...

(apologies in advance for my wording here, I'm not too familiar with the correct Rubygems terminology)

For `bundle install`:
* Normally: Do a bare clone of the git repo from the remote to the system cache directory
* If `--local` is passed in: Assume there is a bare clone in the `vendor/cache` or system cache directory already
* git clone the gem from the bare clone to the install path (which will be either the system path or a path from the `--path` setting). Check out the correct ref.
* Build native extensions - this is done by cd'ing into the system gem path (or `--path` setting) and doing the `ruby extconf.rb; make` dance. (but see below!). Native extensions (that weren't precompiled) should never wind up in `vendor/cache`
* The `.git` directory is then deleted (not sure about this - but it seems like we _could_, because we know for sure the ref that was checked out from the directory name, and apparently from that issue I take it people don't want to copy large `.git` directories in situations where they're 

For `bundle exec`:
* The installed path of the gem, either from the system directory of from the `--path` setting, should be added to `$LOAD_PATH`. The `vendor/cache` or system cache directory should NOT be added there.

For `bundle` cache`:
* Nothing should happen unless the `--all` setting is present.
* The bare clone should be copied from the system cache dir to `vendor/cache` (or cloned there, if it wasn't already in the system cache dir).
* If `--no-install` is present, stop here
* Otherwise, proceed with the install steps - git clone the gem from `vendor/cache` to the `--path` setting or system gem dir, and build native extensions in that location as for `bundle install`.

Sound about right?

------------------------------------------------------------

Question about native gems:

AFAICT native extensions installed by bundler wind up in _two_ places - both in `ruby/x.y.0/gems/$GEM_NAME/lib/` and `ruby/x.y.0/extensions/$PLATFORM/x.y.0/$GEM_NAME/`. _Both_ seem to get added to `$LOAD_PATH`:

```
kj@kj-thinkpad p % find vendor/ -iname '*.so'
vendor/bundle/ruby/3.0.0/gems/puma-6.0.0/ext/puma_http11/puma_http11.so
vendor/bundle/ruby/3.0.0/gems/puma-6.0.0/lib/puma/puma_http11.so
vendor/bundle/ruby/3.0.0/gems/nio4r-2.5.8/ext/nio4r/nio4r_ext.so
vendor/bundle/ruby/3.0.0/gems/nio4r-2.5.8/lib/nio4r_ext.so
vendor/bundle/ruby/3.0.0/extensions/x86_64-linux/3.0.0/puma-6.0.0/puma/puma_http11.so
vendor/bundle/ruby/3.0.0/extensions/x86_64-linux/3.0.0/nio4r-2.5.8/nio4r_ext.so

irb(main):005:0> $LOAD_PATH
=>
["/home/kj/.rbenv/rbenv.d/exec/gem-rehash",
 "/home/kj/p/vendor/bundle/ruby/3.0.0/gems/puma-6.0.0/lib",
 "/home/kj/p/vendor/bundle/ruby/3.0.0/extensions/x86_64-linux/3.0.0/puma-6.0.0",
 "/home/kj/p/vendor/bundle/ruby/3.0.0/gems/nio4r-2.5.8/lib",
 "/home/kj/p/vendor/bundle/ruby/3.0.0/extensions/x86_64-linux/3.0.0/nio4r-2.5.8",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/gems/3.0.0/gems/bundler-2.2.33/lib",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/site_ruby/3.0.0",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/site_ruby/3.0.0/x86_64-linux",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/site_ruby",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/vendor_ruby/3.0.0",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/vendor_ruby/3.0.0/x86_64-linux",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/vendor_ruby",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/3.0.0",
 "/home/kj/.rbenv/versions/3.0.4/lib/ruby/3.0.0/x86_64-linux"]
```

Should they actually wind up in both places? Any idea what's going on here?

-----------------------------------------------------------

Question about `--deployment`:

Does the `--deployment` mode do anything other than imply `--local --path` vendor/bundle`? That's my reading of the doc, but I'm not 100% sure (we don't use this at $DAYJOB because we ship docker images...)

deivid-rodriguez:
@KJTsanaktsidis Sorry for the lack of feedback! Let's start with this patch, which seems correct and indeed a related but separate issue.

More issues with caching git sources should get fixed by https://github.com/rubygems/rubygems/pull/4469. Once we introduced the fix here and that PR we can reevaluate if there's still anything missing/incorrect.

Can you add a test for this patch? You can find related specs at `bundler/spec/commands/cache_spec.rb`.

KJTsanaktsidis:
Hey @deivid-rodriguez, @Julzerator and I put our heads together in the new year and put together a test for this fix. PTAL - thanks!

deivid-rodriguez:
Hei @KJTsanaktsidis, thanks for working on this.

That spec seems to also pass without the fix in this PR. Can you write it so that it fails without this change, and passes with it?

deivid-rodriguez:
@KJTsanaktsidis still interested?

KJTsanaktsidis:
Sorry @deivid-rodriguez  - this totally fell off my radar. I think I wrote a much better test now, WDYT?

deivid-rodriguez (approved):
Looks great! Thanks for the thorough comments and test, hopefully we can remove the second comment block once #4469 is out!

deivid-rodriguez:
Ooops, unfortunately the spec needs to be adapted to also work in "bundler 3" mode, where Bundler will install to `.bundle` by default, not to the default `GEM_HOME`. I think you can use the `default_bundle_path` helper for this.

KJTsanaktsidis:
OK, think I got it working - should be good to go when you run the workflow. Thanks!

deivid-rodriguez (approved):
It seems good now. There were two lint issues but I fixed them myself to not delay this fix anymore!

KJTsanaktsidis:
Great, thanks for your help, sorry it took me so long!

deivid-rodriguez:
Thanks to you for the fix!

PR #6091, opened 2022-11-30
Clean with make clean using ENV

eloyesp:
This is an alternative implementation for #4017 following the conversations with the team.

It adds an undocumented environment variable that allow a custom make task to be run after the install (actually anything but "clean" will break).

    RUBYGEMS_POST_MAKE_TASK=clean gem install xyzl

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -40,7 +40,9 @@ class Gem::Ext::Builder
       env << "sitelibdir=%s" % sitedir
     end
 
-    ["clean", "", "install"].each do |target|
+    steps = ["clean", "", "install", ENV["RUBYGEMS_POST_MAKE_TASK"]]
+
+    steps.compact.each do |target|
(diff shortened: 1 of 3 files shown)


simi:
> It adds an undocumented environment variable that allow a custom make task to be run after the install (actually anything but "clean" will break).

We can clearly document it as well, we just need to find out proper place for that.

simi:
> It adds an undocumented environment variable that allow a custom make task to be run after the install (actually anything but "clean" will break).

Thinking about this again, it is not true anything but "clean" will break. But this makes it global used for all gems installed. There is no easy way to specify this option per gem. Anyway it is just for initial testing.

eloyesp:
"clean" is a hardcoded value that is allowed to fail, using anything else might break the build for any gem. For example, someone recomended using "distclean" instead of "clean", and that might work for some gems and make the build fail entirely on other gem.

deivid-rodriguez:
I think this landed through the other (unfortunately more abrupt 😅) proposal, so should no longer be relevant. Thanks @eloyesp for moving this forward!

PR #6107, opened 2022-12-09
Drop support for Ruby 2.3, 2.4, 2.5 and RubyGems 2.5, 2.6, 2.7

deivid-rodriguez:

We've been warning when running Bundler on old RubyGems if running on Ruby 2.3, Ruby 2.4 and Ruby 2.5 for a year.

Hopefully that has given most folks sufficient head ups about our intention, although I can still imagine some disruption in some cases. Yet, we need to do this at some point, let's use new year's release to do it.

Drop support for Ruby 2.3, 2.4 and 2.5, and RubyGems 2.5, 2.6 and 2.7.

There's a bunch of follow up PRs after this like ~RuboCop style,~ bumping development dependencies, or bumping vendored gems. But I'll do that as follow up PRs.

EDIT: I'm also normalizing RuboCop styles in this PR to make CI happy.


bundler/exe/bundle
```diff
@@ -15,10 +15,7 @@ else
   require "bundler"
 end
 
-# Workaround for non-activated bundler spec due to missing https://github.com/rubygems/rubygems/commit/4e306d7bcdee924b8d80ca9db6125aa59ee4e5a3
-gem "bundler", Bundler::VERSION if Gem.rubygems_version < Gem::Version.new("2.6.2")
-
-if Gem.rubygems_version < Gem::Version.new("3.2.3") && Gem.ruby_version < Gem::Version.new("2.6.a") && !ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"]
+if Gem.rubygems_version < Gem::Version.new("3.2.3") && Gem.ruby_version < Gem::Version.new("3.0.a") && !ENV["BUNDLER_NO_OLD_RUBYGEMS_WARNING"]
```
deivid-rodriguez, line 18: Ruby 3.0.0 ships with RubyGems 3.2.3. So next year we can drop support for Ruby 2.6 and Ruby 2.7 and we'll no longer need this warning.

bundler/lib/bundler/source/path.rb
```diff
@@ -225,7 +225,7 @@ def generate_bin(spec, options = {})
         # Some gem authors put absolute paths in their gemspec
         # and we have to save them from themselves
         spec.files = spec.files.map do |p|
-          next p unless p =~ /\A#{Pathname::SEPARATOR_PAT}/
```
simi, line 228: :thinking: When touching this line, maybe it is good time to rename block var `p` to not shadow `Kernel#p`. Maybe to `path`?

deivid-rodriguez: Sure, why not.

simi (approved):

PR #6110, opened 2022-12-11
Fix display of previous gem version when previously downloaded already

zarqman:

`bundle update` is no longer showing the green ` (was: x.y.z)` suffix for updated gems that were already installed (eg: by another app).

This displayed in bundler <= 2.3.13 and went missing in 2.3.14.

This PR reenables the "was..." suffix, which matches what was displayed both in bundler <= 2.3.13 and as still displayed in the current version for newly downloaded gems and for sources other than Rubygems.

In <= 2.3.13 and again with this PR:
```
Using rails 7.0.4 (was 7.0.3)
```
In 2.3.14+:
```
Using rails 7.0.4
```

## Background

A number of changes surrounding tracking of the previous gem version were made in #5530, which was part of bundler 2.3.14. It appears those changes simply missed passing the previous spec info to the newly enhanced `version_message` method in the described case.

### DIFF
--- a/bundler/lib/bundler/source/rubygems.rb
+++ b/bundler/lib/bundler/source/rubygems.rb
@@ -145,7 +145,7 @@ module Bundler
         end
 
         if installed?(spec) && !force
-          print_using_message "Using #{version_message(spec)}"
+          print_using_message "Using #{version_message(spec, options[:previous_spec])}"
           return nil # no post-install message
         end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Nice catch @zarqman, thanks for fixing this!

I added a regression spec to make sure I don't introduce the same mistake again in the future.

PR #6112, opened 2022-12-11
Bump vendored deps

deivid-rodriguez:

No real user problem, but I figured our new year's release is a good moment to bump these.

Bump versions of vendored fileutils, connection_pool, and uri. All of them were pinned back due to having to support old rubies. Now we should be able to upgrade since we have dropped support.


indirect:
Thanks for doing this work!

hsbt (approved):

PR #6114, opened 2022-12-12
Clean up Indexer build files in tests

nobu:

test/rubygems/test_gem_indexer.rb leaves 5 temporary directories fore each runs.

* In `teardown`, remove the directory created at `setup`.
* In tests creating other indexers, clean those directories.


test/rubygems/test_gem_indexer.rb
```diff
@@ -179,6 +190,8 @@ def test_generate_index_modern_back_to_back
 
     assert_indexed @indexerdir, "latest_specs.#{@marshal_version}"
     assert_indexed @indexerdir, "latest_specs.#{@marshal_version}.gz"
+  ensure
+    FileUtils.rm_rf(indexer.directory)
```
deivid-rodriguez, line 194: If something failed before the `indexer` variable is defined, this would result in a runtime error in `ensure`, right? Really loving to be able to use `ensure` without explicit begin, but in this case the correct thing would be to only wrap the code after `indexer` is defined?

test/rubygems/test_gem_indexer.rb
```diff
@@ -31,15 +31,26 @@ def setup
...
+    indexer1 = Gem::Indexer.new @indexerdir
+    assert_predicate indexer1, :build_modern
 
-    indexer = Gem::Indexer.new @indexerdir, :build_modern => true
-    assert indexer.build_modern
+    indexer2 = Gem::Indexer.new @indexerdir, :build_modern => true
+    assert_predicate indexer2, :build_modern
+  ensure
+    FileUtils.rm_rf(indexer1.directory)
+    FileUtils.rm_rf(indexer2.directory)
```
deivid-rodriguez, line 53: Samething about making sure `indexer1` and `indexer2` are defined when this `ensure` block is run.

test/rubygems/test_gem_indexer.rb
```diff
@@ -31,15 +31,26 @@ def setup
     @indexer = Gem::Indexer.new(@indexerdir)
   end
 
+  def teardown
+    FileUtils.rm_rf(@indexer.directory)
+  ensure
+    super
+  end
```
deivid-rodriguez, line 38: Here we have the same issue again about `@indexer` being potentially undefined when `teardown` is run. Since we don't have `around` hooks available, I think I would extract a `with_indexer(dir)` method that creates the proper directories, instantiates the index, yields to the passed block, and deletes the directory afterwards. And explicitly wrap every test that needs that with it.

deivid-rodriguez (commented):
I left a few comments, thanks for making our tests more polite :)

deivid-rodriguez (approved):
I would've probably left `Gem::Indexer` instantiation out of the `with_indexer` method so that I can avoid the guard of `if build_directory` in the `ensure` block.

But that's super minor nit, and I admit defensive code comes handy sometimes, so looks good!

PR #6127, opened 2022-12-14
Use safe navigation operator

nobu:

None.

Replaced `var && var.method` using `&.`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -179,6 +179,10 @@ Style/BlockDelimiters:
 Style/PercentLiteralDelimiters:
   Enabled: true
 
+Style/SafeNavigation:
+  Enabled: true
+  MaxChainLength: 1
+
 Style/StringLiterals:
   EnforcedStyle: double_quotes
 
--- a/bundler/lib/bundler/cli/platform.rb
+++ b/bundler/lib/bundler/cli/platform.rb
@@ -10,7 +10,7 @@ module Bundler
     def run
       platforms, ruby_version = Bundler.ui.silence do
         locked_ruby_version = Bundler.locked_gems && Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
-        gemfile_ruby_version = Bundler.definition.ruby_version && Bundler.definition.ruby_version.single_version_string
+        gemfile_ruby_version = Bundler.definition.ruby_version&.single_version_string
         [Bundler.definition.platforms.map {|p| "* #{p}" },
          locked_ruby_version || gemfile_ruby_version]
       end
--- a/bundler/lib/bundler/cli/platform.rb
+++ b/bundler/lib/bundler/cli/platform.rb
@@ -8,12 +8,12 @@ module Bundler
     end
 
     def run
-      platforms, ruby_version = Bundler.ui.silence do
-        locked_ruby_version = Bundler.locked_gems && Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
-        gemfile_ruby_version = Bundler.definition.ruby_version&.single_version_string
-        [Bundler.definition.platforms.map {|p| "* #{p}" },
-         locked_ruby_version || gemfile_ruby_version]
+      ruby_version = if Bundler.locked_gems
+        Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
+      else
+        Bundler.definition.ruby_version&.single_version_string
       end
+
       output = []
 
       if options[:ruby]
@@ -23,6 +23,8 @@ module Bundler
           output << "No ruby version specified"
         end
       else
+        platforms = Bundler.definition.platforms.map {|p| "* #{p}" }
+
         output << "Your platform is: #{Gem::Platform.local}"
         output << "Your app has gems that work on these platforms:\n#{platforms.join("\n")}"
 
--- a/bundler/lib/bundler/gem_helper.rb
+++ b/bundler/lib/bundler/gem_helper.rb
@@ -21,7 +21,7 @@ module Bundler
 
       def gemspec(&block)
         gemspec = instance.gemspec
-        block.call(gemspec) if block
+        block&.call(gemspec)
         gemspec
       end
     end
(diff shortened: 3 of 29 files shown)


deivid-rodriguez (commented):
Could you enable https://docs.rubocop.org/rubocop/1.40/cops_style.html#stylesafenavigation so that we keep code consistent going forward?

lib/rubygems/resolver/molinillo/lib/molinillo/resolution.rb
```diff
@@ -713,8 +713,7 @@ def filtered_possibility_set(vertex)
       # @return [Object] the locked spec named `requirement_name`, if one
       #   is found on {#base}
       def locked_requirement_named(requirement_name)
-        vertex = base.vertex_named(requirement_name)
-        vertex && vertex.payload
+        base.vertex_named(requirement_name)&.payload
```
deivid-rodriguez, line 716: Molinillo's files are vendored code, so we shouldn't modify them directly. You can propose the change [upstream](https://github.com/CocoaPods/Molinillo).

deivid-rodriguez:
> Can gem_spec.respond_to?(:default_gem?) be false here?

I think it cannot anymore, but CI will tell.

Can you enable https://docs.rubocop.org/rubocop/1.40/cops_style.html#stylesafenavigation in configuration at `.rubocop.yml` and `.rubocop_bundler.yml`?

simi (approved):

deivid-rodriguez (commented):
Looks good to me once changes in Molinillo are reverted 👍.

bundler/spec/support/artifice/compact_index_wrong_dependencies.rb
```diff
@@ -6,7 +6,7 @@ class CompactIndexWrongDependencies < CompactIndexAPI
   get "/info/:name" do
     etag_response do
       gem = gems.find {|g| g.name == params[:name] }
-      gem.versions.each {|gv| gv.dependencies.clear } if gem
+      gem&.versions&.each {|gv| gv.dependencies.clear }
```
deivid-rodriguez, line 9: RuboCop limits `ChainLength` for this cop to 2. Personally, I think more than one already becomes messy, so I'd go with `ChainLength: 1`.

If others are happy with the default though, then no problem.

deivid-rodriguez (commented):
One final comment, just want to be sure before we commit to this style since I'm really not used to it.

nobu:
Sorry to be late.
I had a cold since Thursday.

bundler/lib/bundler/cli/platform.rb
```diff
@@ -9,8 +9,8 @@ def initialize(options)
 
     def run
       platforms, ruby_version = Bundler.ui.silence do
-        locked_ruby_version = Bundler.locked_gems && Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
```
deivid-rodriguez, line 12: Mmmm this is the kind of change I wanted to avoid with `MaxChainLength: 1`. Should we revert it?

bundler/lib/bundler/gem_helper.rb
```diff
@@ -218,13 +217,13 @@ def sh_with_status(cmd, &block)
...
         status = $?
-        block.call(outbuf) if status.success? && block
+        block&.call(outbuf) if status.success?
         [outbuf, status]
       end
     end
 
     def gem_key
-      Bundler.settings["gem.push_key"].to_s.downcase if Bundler.settings["gem.push_key"]
+      Bundler.settings["gem.push_key"]&.to_s&.downcase
```
deivid-rodriguez, line 226: Another one with that double chain I wanted to avoid.

deivid-rodriguez (commented):
I added to small comments regarding `MaxChainLength`.

Hope you feel better now @nobu.

bundler/lib/bundler/cli/platform.rb
nobu: How about splitting?
```ruby
        locked_ruby_version = Bundler.locked_gems&.ruby_version
        locked_ruby_version = locked_ruby_version&.gsub(/p\d+\Z/, "")
```

deivid-rodriguez: If think this is the simplest

```ruby
ruby_version = if Bundler.locked_gems
  Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
else
  Bundler.definition.ruby_version&.single_version_string
end
```

And then below replace

```ruby
locked_ruby_version || gemfile_ruby_version
```

with

```ruby
ruby_version
```

nobu: Then `ruby_version` will be shared with the outer scope.

Is `ruby_version` ok by just assigned?
```ruby
      ruby_version = nil
      platforms = Bundler.ui.silence do
        ruby_version = if Bundler.locked_gems
          Bundler.locked_gems.ruby_version&.gsub(/p\d+\Z/, "")
        else
          Bundler.definition.ruby_version&.single_version_string
        end
        Bundler.definition.platforms.map {|p| "* #{p}" }
      end
```

nobu: BTW, I can't find a way to align `else` and `end` with the LHS of the assignment, in ruby-mode.el.

deivid-rodriguez: It seems fine! Actually, the `Bundler.ui.silence` block was introduced a long time ago at https://github.com/rubygems/bundler/commit/33ed52e703b3fa9337787344f0425063b33a2f91, and I believe we probably don't need it anymore, so we can assign both variables directly?

nobu: I wasn't sure if it has any effects now.
And `platforms` seems unused in `if options[:ruby]`.

deivid-rodriguez (approved):
Looks good @nobu. I will merge this after releasing. Because I have to draw the line somewhere! :)

.rubocop.yml
```diff
@@ -179,6 +179,12 @@ Style/BlockDelimiters:
 Style/PercentLiteralDelimiters:
   Enabled: true
 
+Style/SafeNavigation:
+  Enabled: true
+  MaxChainLength: 1
+  Exclude:
+    - 'util/automatiek/**/*'
```
deivid-rodriguez, line 186: Why did we exclude `automatiek`? Automatiek used to be a vendored library, but we no longer upstream any patches so it should be fine to normalize it too.

deivid-rodriguez (commented):
Small question about automatiek.

.rubocop.yml
nobu: I didn't know it.
I excluded it since it seemed like vendored.

deivid-rodriguez (approved):

PR #6133, opened 2022-12-15
Clean built artifacts after building extensions

### LINKED ISSUE #3958: Cleanup object files on gems with extensions

deivid-rodriguez:

Currently installing extensions leaves around intermediary files that may end up taking a lot of disk space.

Add a clean step after installing the extension.

This PR is based on the useful work by @eloyesp at #4017.

Fixes #3958.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -17,7 +17,7 @@ class Gem::Ext::Builder
     $1.downcase
   end
 
-  def self.make(dest_path, results, make_dir = Dir.pwd, sitedir = nil)
+  def self.make(dest_path, results, make_dir = Dir.pwd, sitedir = nil, targets = ["clean", "", "install"])
     unless File.exist? File.join(make_dir, "Makefile")
       raise Gem::InstallError, "Makefile not found"
     end
@@ -40,7 +40,7 @@ class Gem::Ext::Builder
       env << "sitelibdir=%s" % sitedir
     end
 
-    ["clean", "", "install"].each do |target|
+    targets.each do |target|
       # Pass DESTDIR via command line to override what's in MAKEFLAGS
       cmd = [
         *make_program,
--- a/lib/rubygems/ext/ext_conf_builder.rb
+++ b/lib/rubygems/ext/ext_conf_builder.rb
@@ -55,6 +55,8 @@ class Gem::Ext::ExtConfBuilder < Gem::Ext::Builder
         destent = ent.class.new(dest_path, ent.rel)
         destent.exist? || FileUtils.mv(ent.path, destent.path)
       end
+
+      make dest_path, results, extension_dir, tmp_dest_relative, ["clean"]
     ensure
       ENV["DESTDIR"] = destdir
     end
--- a/test/rubygems/test_gem_installer.rb
+++ b/test/rubygems/test_gem_installer.rb
@@ -1584,6 +1584,41 @@ gem 'other', version
     end
   end
 
+  def test_install_extension_clean_intermediate_files
+    pend "extensions don't quite work on jruby" if Gem.java_platform?
+    @spec = setup_base_spec
+    @spec.require_paths = ["."]
+    @spec.extensions << "extconf.rb"
+
+    File.write File.join(@tempdir, "extconf.rb"), <<-RUBY
+      require "mkmf"
+      CONFIG['CC'] = '$(TOUCH) $@ ||'
+      CONFIG['LDSHARED'] = '$(TOUCH) $@ ||'
+      $ruby = '#{Gem.ruby}'
+      create_makefile("#{@spec.name}")
+    RUBY
+
+    # empty depend file for no auto dependencies
+    @spec.files += %W[depend #{@spec.name}.c].each do |file|
+      write_file File.join(@tempdir, file)
+    end
+
+    shared_object = "#{@spec.name}.#{RbConfig::CONFIG["DLEXT"]}"
+    extension_file = File.join @spec.extension_dir, shared_object
+    intermediate_file = File.join @spec.gem_dir, shared_object
+
+    assert_path_not_exist extension_file, "no before installing"
+    use_ui @ui do
+      path = Gem::Package.build @spec
+
+      installer = Gem::Installer.at path
+      installer.install
+    end
+
+    assert_path_exist extension_file, "installed"
+    assert_path_not_exist intermediate_file
+  end
+
   def test_installation_satisfies_dependency_eh
     installer = setup_base_installer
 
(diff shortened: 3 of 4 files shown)


lib/rubygems/ext/builder.rb
```diff
@@ -17,7 +17,7 @@ def self.class_name
     $1.downcase
   end
 
-  def self.make(dest_path, results, make_dir = Dir.pwd, sitedir = nil)
+  def self.make(dest_path, results, make_dir = Dir.pwd, targets = ["clean", "", "install"])
```
eloyesp, line 20: I don't know what `sitedir` mean... is it being used?

Anyway, I would not modify the signature of the method this way, may be adding a keyword argument to make sure to prevent misunderstandings.

lib/rubygems/ext/ext_conf_builder.rb
```diff
@@ -39,35 +31,22 @@ def self.build(extension, dest_path, results, args=[], lib_dir=nil, extension_di
 
       ENV["DESTDIR"] = nil
 
-      make dest_path, results, extension_dir, tmp_dest_relative
-
-      full_tmp_dest = File.join(extension_dir, tmp_dest_relative)
+      rel_dest_path = Pathname.new(dest_path).relative_path_from(Pathname.new(extension_dir))
```
eloyesp, line 34: I prefer the `Pathname(dest_path)` syntax, without `.new`, it's cleaner and easier to read.

eloyesp (commented):
I like these simplifications a lot, just wanted to add that there might be a secondary bug that was hidden by the build artifacts. 

That is: when bundle install a git sourced gem, it adds some non-existing paths to the load path, I've added a check for that on `Bundler::Runtime#setup` breaking multiple escenarios.

The problem is that I'm not entirely sure how rubygems and bundler interact when building extensions on git sourced gems.

Hope my comments help somehow.

lib/rubygems/ext/builder.rb
deivid-rodriguez: The `sitedir` parameter was only introduced very recently at https://github.com/rubygems/rubygems/commit/dea41fa2dcce8a6c11c7c92c73db7a66a7d7ed51. I'm aware our extension building code is quite complicated, but it used to be worse 😅.

The above change was released with RubyGems 3.3.17, quite recently, so I don't think removing changing it with another positional parameter is going to break anything.

That said, I want to get to a green base line first, then figure out the details.

lib/rubygems/ext/ext_conf_builder.rb
deivid-rodriguez: I actually find the more explicit `Pathname.new` easier to pase, I guess because it's more common Ruby. And it's also the style we generally use in the rest of the codebase. That said, I was also considering to not introduce a dependency on `pathname` just for this, and instead modify the existing `get_relative_path` helper. Will figure this out later.

deivid-rodriguez:
> That is: when bundle install a git sourced gem, it adds some non-existing paths to the load path, I've added a check for that on `Bundler::Runtime#setup` breaking multiple escenarios.

Yeah, building extensions for git gems is a big hack at the moment, and it indeed has the consequence of some paths to be added to the `$LOAD_PATH` that don't exist. That's actually what prevents us to refactor things further and also stop copying build artifacts to `lib_dir`, because that's apparently what makes extensions for `git` gems actually work.

We should refactoring things further to just be able to `make, make install, make clean` to install extensions, no matter the source. But it sounds too much for this PR.

lib/rubygems/ext/builder.rb
deivid-rodriguez: This refactorings broke Windows 2.6 and Windows 2.7 for some reason. I really don't want to investigate that so I removed those simplifications and will be opening a separate PR that can be merged by the end of next year when we will drop Ruby 2.6 and Ruby 2.7 support.

deivid-rodriguez:
After investigating the issues here, I think this is safe to do.

I could only see this breaking if someone is intentionally changing the `$LOAD_PATH` to point to where the artifacts live (normally the `ext/` folder of an installed gem). These artifacts are already installed in two places right now (the `lib` folder of the installed gem, and the `extensions` folder), and both are already in the `$LOAD_PATH` by default. So it makes no sense that someone did this, and if they did, it can't hardly be considered a breaking change in my opinion. Also, cleanup is currently limited to `extconf` extension builder, so the scope is even more limited.

So from my side this is ready.

eloyesp:
Just one more question, will this impact bundler installs? Is bundler installing  gems using the `extconf` builder or is it running his own builder?

deivid-rodriguez:
Yes, Bundler delegates to RubyGems to install extensions, so it will use whatever builder each gem uses. If gems define their extension through an `extconf.rb` file, then Bundler will use that builder from RubyGems.

simi:
Thanks for digging in this @deivid-rodriguez. I did other review of the build artifact handling. I agree this seems safe. I did quick test to see impact on sassc issue https://github.com/sass/sassc-ruby/issues/200 and bundle size went from 65MB to 25MB.

simi:
Thanks @eloyesp for pushing this forward!

lipidity:
Hi 👋 

Running `make clean` after installing an extension breaks https://github.com/ohler55/ox because it removes the built artifacts which are needed at runtime.

simi:
> 

https://github.com/rubygems/rubygems/issues/6205#issuecomment-1365622769

simi:
> Hi wave
> 
> Running `make clean` after installing an extension breaks https://github.com/ohler55/ox because it removes the built artifacts which are needed at runtime.

https://github.com/rubygems/rubygems/issues/6205#issuecomment-1365622769

hsbt:
`appsignal-ruby` was fixed at https://github.com/appsignal/appsignal-ruby/pull/907.

lipidity:
@simi thanks, I should have linked to https://github.com/ohler55/ox/issues/303

hsbt:
@lipidity No. It works. Please confirm [my comment](https://github.com/ohler55/ox/issues/303#issuecomment-1370506686)

lipidity:
I was wrong about ox still being broken. Confirmed that it works.

PR #6137, opened 2022-12-16
Deprecate `sitedir` parameter to `Gem::Ext::Builder.make`

deivid-rodriguez:
<!--
Thanks so much for the contribution!

Extension building code is complicated.

Simplifying things. We were doing what `make install` should be doing ourselves. By passing a relative dest_path and feeding that as `sitelibdir` and `sitearchdir`, we can get `make install` to install there directly, rather than getting `make install` to install to a temporary location, and then manually copying files in there to final final installation path.

This PR can't be merged at the moment because for some reason Ruby 2.6 and 2.7 on Windows don't like it. It sounds like some bug recently fixed since Ruby 3.0 and 3.1 are fine.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -19,7 +19,14 @@ class Gem::Ext::Builder
     $1.downcase
   end
 
-  def self.make(dest_path, results, make_dir = Dir.pwd, sitedir = nil, targets = ["clean", "", "install"])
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
I should probably refresh the full context of this PR in my mind, but I think it provides a nice simplification, and it will be unblocked by #7116 since only problems CI found were on Windows + Ruby 2.6 and 2.7.

martinemde:
You don't _have to_ change the signature of that method. We just ignore the 3rd arg, maybe print a deprecation warning?

deivid-rodriguez:
I made this backwards compatible now, I think. Should be ready for a review.

martinemde (approved):
Awesome

PR #6141, opened 2022-12-17
Use safe Marshal deserialization for dependency API response.

simi:
:information_source: I'm looking at introducing real-word spec to cover the real response from Dependency API as well.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -39,6 +39,16 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
+  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash].freeze
+  SAFE_MARSHAL_ERROR = "Unexpected class %s present in marshaled data. Only %s are allowed.".freeze
+  SAFE_MARSHAL_PROC = proc do |object|
+    object.tap do
+      unless SAFE_MARSHAL_CLASSES.include?(object.class)
+        raise TypeError, format(SAFE_MARSHAL_ERROR, object.class, SAFE_MARSHAL_CLASSES.join(", "))
+      end
+    end
+  end
+
   autoload :Definition,             File.expand_path("bundler/definition", __dir__)
   autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
   autoload :Deprecate,              File.expand_path("bundler/deprecate", __dir__)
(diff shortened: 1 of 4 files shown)


bundler/lib/bundler.rb
```diff
@@ -39,6 +39,16 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
+  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash].freeze
```
simi, line 42: :thinking: I'm not sure why `TrueClass` is needed, but it is passed to proc even not part of the structure.

```ruby
irb(main):013:0> simple_structure = { "name" => [:abc] }
=> {"name"=>[:abc]}
irb(main):014:0> Marshal.load(Marshal.dump(simple_structure), ->(object) { object.tap { puts object.class }})
TrueClass
String
Symbol
Array
Hash
=> {"name"=>[:abc]}
```

simi: I was able to strip this down to string deserialization.

```ruby
Marshal.load(Marshal.dump('string'), ->(o) { o.tap { puts o.class } })
String
TrueClass
```

indirect (approved):
This seems good for the dependency API! Can we also use SafeMarshal for loading `Marshal.4.8.specs.gz` and `latest_specs`? Could be in another PR if that's easier.

simi:
@indirect those includes `Gem::Version`, I'll open another PR to handle those as well.

bundler/lib/bundler.rb
composerinteralia: In case it's useful, I think `TrueClass` is related to the encoding. That seems to be what the ET at the end of the dump is about:

```rb
Marshal.dump('string')
#=> "\x04\bI\"\vstring\x06:\x06ET"
```

whereas we get `false` in this case:

```rb
Marshal.dump('string.encode!('ASCII'))
#=> "\x04\bI\"\vstring\x06:\x06EF"
```

or a string for other encodings:

```rb
Marshal.dump("string".encode!('ISO-8859-1'))
#=> "\x04\bI\"\vstring\x06:\rencoding\"\x0FISO-8859-1"
```

railsfactory-pavan:
Hi Team, I am also facing the same issue while installing the gem for my application I need help from your end.

 RUBY VERSION
   ruby 2.6.5p114

BUNDLED WITH
   2.4.12
```

--- TEMPLATE END ----------------------------------------------------------------

Unfortunately, an unexpected error occurred, and Bundler cannot continue.

First, try this link to see if there are any existing issue reports for this error:
https://github.com/rubygems/rubygems/search?q=TypeError++Unexpected+class+FalseClass+present+in+marshaled+data.+Only+Symbol%2C+TrueClass%2C+String%2C+Array%2C+Hash%2C+Gem++Version%2C+Gem++Specification+are+allowed.&type=Issues

If there aren't any reports for this error yet, please fill in the new issue form located at https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md, and copy and paste the report template above in there.

PR #6146, opened 2022-12-18
Enhance `bundle open` command to allow opening subdir/file of gem

yoka:

I had an development workflow halt when vscode + remote containers combo was used. The scenario was following:

I wanted to open up specific gems specific file, but as remote containers use filesystem from within the docker container I could not use bundle open command to be able to see within a specific file of a gem. 

Even though remote containers does not allow easy "mount n+1 dirs" mechanisms, it would have been sufficient enough to be able to scope my bundle open command to a specific file that my tests were failing at.

I had the expectation that `bundle open` command would have allowed to open specific file with some syntax, but my expectations were not met.

As the first intuitive command to open a specific file was to try `bundle open GEM/some/file.rb` this is what was implemented.

I did try out adding second argument too, so that `bundle open GEM some/file.rb` would have worked, but that felt kind of off.

So, wrote the tests for the core of this enhancement and tests that make sure all the other fuzz matching & "did you mean" features would work just like they've been working before this enhancement.

- [ x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [ x] Write code to solve the problem
- [ x] Make sure you follow the [current code style]

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -509,6 +509,7 @@ module Bundler
     subcommand "config", Config
 
     desc "open GEM", "Opens the source directory of the given bundled gem"
+    method_option "path", :type => :string, :banner => "Open relative path of the gem source."
     def open(name)
       require_relative "cli/open"
       Open.new(options, name).run
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -509,7 +509,7 @@ module Bundler
     subcommand "config", Config
 
     desc "open GEM", "Opens the source directory of the given bundled gem"
-    method_option "path", :type => :string, :banner => "Open relative path of the gem source."
+    method_option "path", :type => :string, :lazy_default => "", :banner => "Open relative path of the gem source."
     def open(name)
       require_relative "cli/open"
       Open.new(options, name).run
--- a/bundler/lib/bundler/cli/open.rb
+++ b/bundler/lib/bundler/cli/open.rb
@@ -10,6 +10,7 @@ module Bundler
     end
 
     def run
+      raise InvalidOption, "Cannot specify `--path` option without a value" if !@path.nil? && @path.empty?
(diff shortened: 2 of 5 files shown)


simi:
> I did try out adding second argument too, so that bundle open GEM some/file.rb would have worked, but that felt kind of off.

Can you elaborate? That seems much better to me. :thinking:

deivid-rodriguez:
Thanks for coding this enhancement!

I agree with @simi. If it's because of clarity, we could add a specific flag for it, like `bundle open foo --path README.md`?

Also using `GEM/path` could potentially cause ambiguity is we introduced [scoped gems](https://github.com/rubygems/rfcs/pull/40) in the future.

yoka:
> Thanks for coding this enhancement!
> 
> I agree with @simi. If it's because of clarity, we could add a specific flag for it, like `bundle open foo --path README.md`?
> 
> Also using `GEM/path` could potentially cause ambiguity is we introduced [scoped gems](https://github.com/rubygems/rfcs/pull/40) in the future.

Alright fellas, --path opt sounds reasonable to me. I'll make the necessary changes.

bundler/lib/bundler/cli.rb
```diff
@@ -509,6 +509,7 @@ def exec(*args)
     subcommand "config", Config
 
     desc "open GEM", "Opens the source directory of the given bundled gem"
+    method_option "path", :type => :string, :banner => "Open relative path of the gem source."
```
deivid-rodriguez, line 512: Minor thing, but Thor currently passes `path="path"` if you specify just `--path`, which is probably never what you want 🤷‍♂️.

As a workaround, we can pass `:lazy_default => ""`, and then check this in the command and raise a missing value error?

deivid-rodriguez (commented):
Looks great, just added a small comment about handling a missing value.

bundler/lib/bundler/cli.rb
yoka: Makes sense, will check this

deivid-rodriguez (approved):
🙌

simi (approved):

PR #6149, opened 2022-12-19
Add --ext=rust support for simple Rust extension.

simi:
Supporting just raw Rust without magnus. I would like to add additional extension skeleton using [magnus](https://github.com/matsadler/magnus) (like `--ext=magnus`).

based on combination of:

- https://github.com/oxidize-rb/rb-sys/tree/main/examples/rust_reverse
- https://github.com/rubygems/rubygems/pull/5613
- https://github.com/oxidize-rb/oxi-test

I would like to add another PR as followup to do something at least "almost usual" in generated extensions (for example define "hello" method and return "hello world") for both C and Rust variants (and potentially `magnus` as well once introduced).

@ianks feel free to review :pray:


bundler/lib/bundler/cli/gem.rb
```diff
@@ -190,14 +190,23 @@ def run
...
           "ext/newgem/newgem.h.tt" => "ext/#{name}/#{underscored_name}.h",
           "ext/newgem/newgem.c.tt" => "ext/#{name}/#{underscored_name}.c"
         )
       end
 
+      if extension == "rust"
+        templates.merge!(
+          "Cargo.toml.tt" => "Cargo.toml",
+          "ext/newgem/Cargo.toml.tt" => "ext/#{name}/Cargo.toml",
+          "ext/newgem/extconf-rust.rb.tt" => "ext/#{name}/extconf.rb",
```
hsbt, line 205: How about `extconf-rs.rb.tt` instead of `extconf-rust.rb.tt`?

bundler/lib/bundler/cli/gem.rb
```diff
@@ -190,14 +190,23 @@ def run
...
+          "ext/newgem/src/lib.rs.tt" => "ext/#{name}/src/lib.rs",
```
hsbt, line 206: Should we use `newgem.rs.tt` same as C extension?

hsbt (commented):
I added comment for cometic.

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Why do you think `extconf-rs.rb.tt` is better, @hsbt?

simi: In case we will onboard `magnus` as proposed, it would be easier to keep it in line with `--ext` value using current `extconfig-$EXT.rb.tt` and pattern.

bundler/lib/bundler/cli/gem.rb
simi: I was thinking about renaming folder `ext/newgem` to `ext/newgem-c` and `ext/newgem-rust` or similar, but not all files related to extensions are in `ext` folder :thinking:.

bundler/lib/bundler/cli/gem.rb
hsbt: > it would be easier to keep it in line with `--ext` value using current `extconfig-$EXT.rb.tt` and pattern.

👍  Thanks. We should keep it.

bundler/lib/bundler/cli/gem.rb
simi: Ahh, you mean for the file itself. `src/lib.rs` is standard Cargo name of Rust library entry point.

https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html

> After we run cargo new, we use ls to see what Cargo creates. In the project directory, there’s a Cargo.toml file, giving us a package. There’s also a src directory that contains main.rs. Open Cargo.toml in your text editor, and note there’s no mention of src/main.rs. Cargo follows a convention that src/main.rs is the crate root of a binary crate with the same name as the package. Likewise, Cargo knows that if the package directory contains src/lib.rs, the package contains a library crate with the same name as the package, and src/lib.rs is its crate root. Cargo passes the crate root files to rustc to build the library or binary.

> Here, we have a package that only contains src/main.rs, meaning it only contains a binary crate named my-project. If a package contains src/main.rs and src/lib.rs, it has two crates: a binary and a library, both with the same name as the package. A package can have multiple binary crates by placing files in the src/bin directory: each file will be a separate binary crate.

ianks:
This looks good for the most part. Great work @simi 

Although I do really think we should default to Magnus (with the rb-sys-interop feature). [Defaulting to rb-sys is not something I would encourage for gem authors](https://github.com/oxidize-rb/rb-sys/tree/main/crates/rb-sys#%EF%B8%8F-notice). Are you open to doing that?

ianks:
So essentially the CLI interface I’d recommend would be like this:

- `ext=rust` -> Magnus
- `ext=rb-sys` -> rb-sys

simi:
> So essentially the CLI interface I’d recommend would be like this:
> 
> * `ext=rust` -> Magnus
> * `ext=rb-sys` -> rb-sys

What about following?

- ext=magnus -> Magnus
- ext=rb-sys -> rb-sys

simi:
I'm going to submit `magnus` implementation and we can decide on the naming later. We can use generic names like `rust` and `rust-minimal` or similar, so the underlying implementation is not important.

bundler/lib/bundler/cli/gem.rb
hsbt: Thanks, I see. `lib.rs` is reasonable.

hsbt:
> * ext=magnus -> Magnus
> * ext=rb-sys -> rb-sys

Can we alias `rust` to `rb-sys`? I'm not sure which is the better recommendation for the Rubyists. I'm okay to alias `magnus`.

ianks:
> > * ext=magnus -> Magnus
> > * ext=rb-sys -> rb-sys
> 
> Can we alias `rust` to `rb-sys`? I'm not sure which is the better recommendation for the Rubyists. I'm okay to alias `magnus`.

I'd prefer not to, because really gem authors should be using a high-level / safe library like `magnus` when writing gems. With `magnus` they also have the ability to opt-in to the low-level / unsafe bindings from `rb-sys` (via the `rb-sys-interop` feature) when they need to do, but the default should be safe.

simi:
@ianks ok, for now I decided to just got with `--ext=rust` and in docs mention currently it will generate code based on magnus. That should be enough to be able to change this in the future. I have used magnus crate only for now (following magnus README) to keep it as simple as possible.

Next to those changes, I have updated CIs to be able to compile Rust by default (when selected). I have used diferent approach per CI since each work in different way. At some of them I have opted-in to install latest RubyGems and Bundler, since bundler can now auto-switch. That seems the easiest option for now to keep it compatible accross various Rubies.

simi:
Also I have added `required_rubygems_version` constraint to rust extension gemspec, since older rubygems can't handle Rust extensions.

bundler/lib/bundler/cli/gem.rb
```diff
@@ -415,6 +425,10 @@ def open_editor(editor, file)
       thor.run(%(#{editor} "#{file}"))
     end
 
+    def rust_builder_required_rubygems_version
+      "3.3.11"
```
deivid-rodriguez, line 429: Should we also fail to generate the gem if the user does not have a RubyGems version installed that knows Rust?

If we did this, we could also install the version of Bundler (like we do know) and RubyGems (to make sure it's up to date enough) that the user is running in CI configuration, to make sure things work in CI just like in their local environment.

deivid-rodriguez: Actually, every RubyGems version that knows Rust also installs a default Bundler version that can auto-switch, so if we do this, I think we can completely avoid the `gem install bundler` step in CI configuration.

simi: The CI setup is just example, I would not like to spend a lot of time to make it bullet-proof. Anyway the problem I have faced was related to older Ruby using default RubyGems on CI with no Rust builder support.

IMHO it would be enough to make a comment in the gemspec explaning why this entry is in there.

deivid-rodriguez: Not sure if my suggestion was clear.

I'm suggesting to add a CI step that runs `gem update --system #{Gem.rubygems_version}`, so that we keep generating CI configuration with deterministic versions (like we did until now), and also with an modern enough RubyGems version that has Rust builder support.

With that we keep the CI configuration using fixed RubyGems versions instead of rolling RubyGems versions, and also guarantee that RubyGems version is modern enough for Rust.

The above would work best if we refused to generate Rust gems that are not going to work anyways because the user does not have an up to date enough RubyGems, because otherwise we would be also generating CI configuration that does not work with Rust either.

I do like to encourage deterministic CI configurations like we've done until now, but I guess it's no big deal if it's necessary to unlock them to make this work without a lot of extra effort.

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -23,5 +23,8 @@ jobs:
       with:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
+<%- if config[:ext] == 'rust' -%>
+        rubygems: 'latest' # for best Rust extension support
```
ianks, line 27: We should use https://github.com/oxidize-rb/actions/blob/main/setup-ruby-and-rust/readme.md for CI on rust if possible. It’s well supported and this current configuration will probably not work in all scenarios, it also will be slow due to no caching etc

simi:
Can we keep CI improvement discussion out if this PR?

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Can we at least keep the CI untouched for the Ruby case, i.e., using fixed RubyGems and Bundler versions? That should be an straightforward change, happy to do it myself!

simi:
@deivid-rodriguez I have reverted all changes into CI templates for this PR. I'll continue on improving those in following PR keeping in mind all suggestions from comments in here.

deivid-rodriguez (approved):
Awesome, thanks @simi!

ianks:
Looks good to me! Thanks so much @simi, this will have a huge impact. Let's give @matsadler a chance to chime in as well :)

simi:
@ianks cool, we would like to release this soon before Ruby 3.2 (to be included in there). I'm going to prepare followup (based on this one) for CI thingies meanwhile.

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
simi: @ianks there is only v1 tag and that doesn't support needed `rubygems` option. Would it be possible to tag new version?

matsadler:
No comments from me apart from this is very cool.

Let me know if you'd like a hand with any documentation, like if you want examples of how to do something in particular with Magnus.

simi:
Should I rebase and merge?

hsbt:
You can merge this without `squash`, I think.

hsbt (approved):

simi:
@deivid-rodriguez can you please check https://github.com/rubygems/rubygems/pull/6149/commits/68537383487afabc5793f35716a4e515ac7d6a67 is what you suggested? I can squash and merge one approved.

deivid-rodriguez (approved):
❤️

bundler/lib/bundler/cli/gem.rb
deivid-rodriguez: Resolved now! 💪

PR #6151, opened 2022-12-19
Validate Cargo.lock is present for Rust based extensions.

simi:
When rust based extension is created (for example by https://github.com/rubygems/rubygems/pull/6149), you need to run cargo to generate lock file and ensure lock is added to the gem, otherwise gem will fail to compile later when being installed.

Let's add error if it is missing during specification validation.


deivid-rodriguez (approved):

simi:
@ianks does this make sense?

PR #6161, opened 2022-12-19
Fix resolver hang

deivid-rodriguez:

TODO.

TODO.


deivid-rodriguez:
Unfortunately I'm not able to reliably reproduce this hang anymore. So I'm giving up on this and will wait for user reports after release to verify this.

PR #6168, opened 2022-12-20
Update generated CI scripts to be able to compile Rust extension.

simi:
- lock current RubyGems
- keep it as simple/same as possible for now

### DIFF
--- a/bundler/lib/bundler/templates/newgem/circleci/config.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/circleci/config.yml.tt
@@ -3,8 +3,20 @@ jobs:
   build:
     docker:
       - image: ruby:<%= RUBY_VERSION %>
+<%- if config[:ext] == 'rust' -%>
+    environment:
+      RB_SYS_FORCE_INSTALL_RUST_TOOLCHAIN: 'true'
+<%- end -%>
     steps:
       - checkout
+<%- if config[:ext] == 'rust' -%>
+      - run:
+          name: Install Rust/Cargo dependencies
+          command: apt-get update && apt-get install -y clang
+      - run:
+          name: Install a RubyGems version that can compile rust extensions
+          command: gem update --system '<%= ::Gem.rubygems_version %>'
+<%- end -%>
       - run:
           name: Run the default task
           command: |
--- a/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
@@ -18,10 +18,20 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+<%- if config[:ext] == 'rust' -%>
+    - name: Set up Ruby & Rust
+      uses: oxidize-rb/actions/setup-ruby-and-rust@main
+      with:
+        ruby-version: ${{ matrix.ruby }}
+        bundler-cache: true
+        cargo-cache: true
+        rubygems: '<%= ::Gem.rubygems_version %>'
+<%- else -%>
     - name: Set up Ruby
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
+<%- end -%>
     - name: Run the default task
       run: bundle exec rake
--- a/bundler/lib/bundler/templates/newgem/gitlab-ci.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/gitlab-ci.yml.tt
@@ -2,9 +2,17 @@ default:
   image: ruby:<%= RUBY_VERSION %>
 
   before_script:
+<%- if config[:ext] == 'rust' -%>
+    - apt-get update && apt-get install -y clang
+    - gem update --system '<%= ::Gem.rubygems_version %>'
+<%- end -%>
     - gem install bundler -v <%= Bundler::VERSION %>
     - bundle install
 
 example_job:
+<%- if config[:ext] == 'rust' -%>
+  variables:
+    RB_SYS_FORCE_INSTALL_RUST_TOOLCHAIN: 'true'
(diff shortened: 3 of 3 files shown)


bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
```diff
@@ -18,10 +18,21 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+<%- if config[:ext] == 'rust' -%>
+    - name: Set up Ruby & Rust
+      uses: oxidize-rb/actions/setup-ruby-and-rust@0d21ce09c5500315bc61815440f8c4211530f413
```
ianks, line 23: Let's not lock it to a hash, otherwise users won't get critical updates. We already do something similar with `ruby/setup-ruby@v1`, since the `v1` tag actually receives updates.

```suggestion
      uses: oxidize-rb/actions/setup-ruby-and-rust@main
```

simi: This is not recommended by [OpenSSF](https://securityscorecards.dev/). It is recommended to point to paritcular fixed version (like SHA in here) and triage each update manually (with assitance of dependabot or similar tool).

Anyway we already use tagged actions for now, I would be ok to move to tagged one if released. At some point we will move to particular fixed versions as well for all actions at once.

bundler/lib/bundler/templates/newgem/circleci/config.yml.tt
```diff
@@ -3,11 +3,21 @@ jobs:
...
       - checkout
+<%- if config[:ext] == 'rust' -%>
+      - run:
+          name: Instal Rust/Cargo dependencies
+          command: apt-get update && apt-get install -y clang
+<%- end -%>
       - run:
           name: Run the default task
           command: |
+            gem update --system '<%= ::Gem.rubygems_version %>'
```
deivid-rodriguez, line 20: The concern I was raising in your other PR is that (unless we enforce strictly the RubyGems >= 3.3.11 requirement and refuse to generate this file otherwise), this may not be modern enough to include the rust extensions feature.

Because of this, maybe we should go with a fixed high enough version like 3.3.26?

simi: You'll not be able to do anything useful locally as well wen generated by older RubyGems. Do we really need to handle those edge cases? Alt. I can check if local version is enough and prevent generator to provide Rust skeleton, since it will not be able to compile it.

simi: It would be possible to make it clear you need to install latest RubyGems to get best support.

deivid-rodriguez: > Alt. I can check if local version is enough and prevent generator to provide Rust skeleton, since it will not be able to compile it.

Yes, this is exactly what I've been suggesting.

simi: Ahh, finally I do follow. I'll update.

deivid-rodriguez: Sorry if I failed to explain my idea properly 🙏

simi: this should be handled at https://github.com/rubygems/rubygems/commit/68537383487afabc5793f35716a4e515ac7d6a67 now

deivid-rodriguez: Awesome @simi, thanks!

bundler/lib/bundler/templates/newgem/github/workflows/main.yml.tt
simi: @deivid-rodriguez an idea what to do now? Should we use main tag for now? To promote OpenSSF best practices, we will need to move all actions to point to particular version later anyway.

deivid-rodriguez: Main sounds good!

bundler/lib/bundler/templates/newgem/circleci/config.yml.tt
deivid-rodriguez, line 20: I'm good with this change, but not sure if it could be controversial for other people since it increases the maintenance burden of the workflow and it doesn't really provide any more determinism (the default RubyGems version is also fixed).

So maybe we should restrict it to Rust gems, which are the ones actually needing this?

deivid-rodriguez (commented):
Just a comment on the changes that apply also to gem skeletons that don't include Rust extensions.

bundler/lib/bundler/templates/newgem/circleci/config.yml.tt
simi: I don't see any problem with this and there is a little benefit, when using Ruby version like 3.1, you'll get stable rubygems installed for all 3.1.x versions. Anyway feel free to change if that's the only stuff blocking and merge. I'm leaving for today @deivid-rodriguez.

deivid-rodriguez: The only problem is that now gems generated with `--ext=c` or `--no-ext` (default) have two version numbers that need to update, while before there would be only the Bundler version in there.

Since this PR is focusing on generating CI scripts for `--ext=rust`, and those are the only ones that need an up to date RubyGems, I think it'd be best to keep other kind of gems untouched in this PR.

deivid-rodriguez: Asked @simi privately whether the above made sense and he said yes so I tweaked the PR myself 👍.

deivid-rodriguez (approved):

PR #6169, opened 2022-12-21
Fix Bundler not choosing the platform specific version for libv8-node and other allowlisted gems

### LINKED ISSUE #6165: Bundler 2.3.26 ignores the precompiled libv8-node on x86_64-linux on TruffleRuby

Given this Gemfile:
```ruby
source 'https://rubygems.org'
                  
gem 'mini_racer'
```

If I use CRuby 3.1.3 and Bundler 2.3.26, `bundle lock` gives:
```
GEM
  remote: https://rubygems.org/
  specs:
    libv8-node (16.10.0.0-x86_64-linux)
    mini_racer (0.6.3)
      libv8-node (~> 16.10.0.0)

PLATFORMS
  x86_64-linux

DEPENDENCIES
  mini_racer

BUNDLED WITH
   2.3.26
```
which is expected and looks fine (and `bundle install` just fetches the binary gem).

Now if I switch to TruffleRuby, here I use truffleruby-dev (from ruby-build).
With Bundler 2.3.7 (which TruffleRuby currently comes with), it works although it chooses x86_64-linux-musl which is a bit weird (#5508).
```
$ rm Gemfile.lock
$ bundle lock
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    libv8-node (16.10.0.0)
    mini_racer (0.6.3)
      libv8-node (~> 16.10.0.0)

PLATFORMS
  ruby

DEPENDENCIES
  mini_racer

BUNDLED WITH
   2.3.7
$ bundle install
Fetching gem metadata from https://rubygems.org/..
Using bundler 2.3.7
Fetching libv8-node 16.10.0.0 (x86_64-linux-musl)
Installing libv8-node 16.10.0.0 (x86_64-linux-musl)
Fetching mini_racer 0.6.3
Installing mini_racer 0.6.3 with native extensions
Bundled gems are installed into `/home/eregon/.bundle`
```

But with the latest Bundler, Bundler 2.3.26 it fails, i.e., it doesn't use the precompiled gem:
```
$ gem i bundler
Fetching bundler-2.3.26.gem
Successfully installed bundler-2.3.26
1 gem installed
$ rm Gemfile.lock
$ bundle lock
$ cat Gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    libv8-node (16.10.0.0)
    mini_racer (0.6.3)
      libv8-node (~> 16.10.0.0)

PLATFORMS
  ruby

DEPENDENCIES
  mini_racer

BUNDLED WITH
   2.3.26
$ bundle install
Fetching gem metadata from https://rubygems.org/..
Using bundler 2.3.26
Installing libv8-node 16.10.0.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
```
Just in case:
```
$ ruby -ve 'p Gem::Platform.local'
truffleruby 23.0.0-dev-bdd53f99, like ruby 3.1.2, GraalVM CE Native [x86_64-linux]
#<Gem::Platform:0x768 @cpu="x86_64", @os="linux", @version=nil>
```

This is causing issues while updating TruffleRuby to Ruby 3.1.3 which uses Bundler 2.3.26.

Yes I tried latest Bundler.

### What actually happened?

See above.

## Environment

```
Bundler       2.3.26
  Platforms   ruby
Ruby          3.1.2p0 (2022-12-19 revision bdd53f99afd36be2eda0aa823db7230b7ad972e8) [x86_64-linux]
RubyGems      3.3.7
  Gem Path    /home/eregon/.gem/truffleruby/3.1.2.4:/home/eregon/.rubies/truffleruby-dev/lib/gems
  User Home   /home/eregon
  User Path   /home/eregon/.gem/truffleruby/3.1.2.4
Tools         
  Git         2.37.2
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

## Bundler Build Metadata

```
Built At          2022-11-17
Git SHA           23ec5b8501
Released Version  true
```

## Bundler settings

```
path
  Set for the current user (/home/eregon/.bundle/config): "/home/eregon/.bundle"
```

### Gemfile

```ruby
source 'https://rubygems.org'
                  
gem 'mini_racer'
```

### Gemfile.lock

```
<No /home/eregon/code/truffleruby-ws/truffleruby/Gemfile.lock found>
```

deivid-rodriguez:

This was supposed to have been fixed, but Bundler was still not able to choose the precompiled version of those gems that truffleruby has deemed as having precompiled versions that work fine with truffleruby.

This reintroduces the approach in https://github.com/rubygems/rubygems/pull/4049/commits/6085de09c711f072ddd9fa06d125a6da65ebc501, which was reverted at https://github.com/rubygems/rubygems/pull/5711 because it caused some issues. Unfortunately the revert created the original problem again, due to a problem with the spec initially added that was not able to properly cover the fix.

Reintroduce the same approach of setting a proper default value for the `:force_ruby_platform` dependency attribute, and to locked specifications. This approach no longer seem to cause any issues.

Also fix our specs to properly cover the issue.

Fixes #6165.

- [x] ~Write~ Fix [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -16,6 +16,7 @@ require "rubygems/specification"
 require "rubygems/source"
 
 require_relative "match_metadata"
+require_relative "force_platform"
 require_relative "match_platform"
 
 # Cherry-pick fixes to `Gem.ruby_version` to be useful for modern Bundler
@@ -153,12 +154,16 @@ module Gem
   end
 
   class Dependency
+    include ::Bundler::ForcePlatform
+
     attr_accessor :source, :groups
 
     alias_method :eql?, :==
 
     def force_ruby_platform
-      false
+      return @force_ruby_platform if defined?(@force_ruby_platform) && !@force_ruby_platform.nil?
+
+      @force_ruby_platform = default_force_ruby_platform
     end
 
     def encode_with(coder)
@@ -276,6 +281,10 @@ module Gem
 
         without_gnu_nor_abi_modifiers
       end
+    end
+
+    if RUBY_ENGINE == "truffleruby" && !defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)
+      REUSE_AS_BINARY_ON_TRUFFLERUBY = %w[libv8 libv8-node sorbet-static].freeze
     end
   end
 
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -73,6 +73,7 @@ bundler/lib/bundler/fetcher/compact_index.rb
 bundler/lib/bundler/fetcher/dependency.rb
 bundler/lib/bundler/fetcher/downloader.rb
 bundler/lib/bundler/fetcher/index.rb
+bundler/lib/bundler/force_platform.rb
 bundler/lib/bundler/friendly_errors.rb
 bundler/lib/bundler/gem_helper.rb
 bundler/lib/bundler/gem_helpers.rb
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -463,7 +463,7 @@ EOF
     end
 
     def local_platform
-      return Gem::Platform::RUBY if settings[:force_ruby_platform] || Gem.platforms == [Gem::Platform::RUBY]
+      return Gem::Platform::RUBY if settings[:force_ruby_platform]
       Gem::Platform.local
     end
 
--- a/bundler/lib/bundler/dependency.rb
+++ b/bundler/lib/bundler/dependency.rb
@@ -7,7 +7,7 @@ require_relative "rubygems_ext"
 module Bundler
   class Dependency < Gem::Dependency
     attr_reader :autorequire
-    attr_reader :groups, :platforms, :gemfile, :path, :git, :github, :branch, :ref, :force_ruby_platform
+    attr_reader :groups, :platforms, :gemfile, :path, :git, :github, :branch, :ref
 
     ALL_RUBY_VERSIONS = ((18..27).to_a + (30..31).to_a).freeze
     PLATFORM_MAP = {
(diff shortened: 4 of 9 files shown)


bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -277,6 +282,10 @@ def normalized_linux_version_ext
         without_gnu_nor_abi_modifiers
       end
     end
+
+    if RUBY_ENGINE == "truffleruby" && !defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)
+      REUSE_AS_BINARY_ON_TRUFFLERUBY = %w[libv8 libv8-node sorbet-static].freeze
+    end
```
deivid-rodriguez, line 288: Later I will propose that truffleruby moves this constant to the `defaults/truffleruby` file, so that eventually we no longer need to keep it ourselves. We currently need to keep it ourselves for our own tests, and to work properly with upgraded copies of RubyGems that don't include it like the "blessed copy" of RubyGems included with truffleruby does.

eregon: Should I do a PR to truffleruby to have
```ruby
class Gem::Platform
  remove_const :REUSE_AS_BINARY_ON_TRUFFLERUBY if const_defined?(:REUSE_AS_BINARY_ON_TRUFFLERUBY)
  REUSE_AS_BINARY_ON_TRUFFLERUBY = %w[libv8 libv8-node sorbet-static]
end
```
in `defaults/truffleruby` then?
I can do that soon.

Or where you thinking of something else?

deivid-rodriguez: Yup, that's what I was thinking.

eregon: Depending on https://github.com/rubygems/rubygems/pull/6238#issuecomment-1372313641 we might need to remove the constant

deivid-rodriguez: Yep, you shouldn't need `remove_const` because of that.

Issue #4058, opened 2020-11-12
Promote `-C` to a global flag


deivid-rodriguez:
I noticed that the `-C` flag, currently only usable with `gem build`, could probably be useful as a global flag. For example, it could be helpful for `gem install` too as per [#448 (comment)](https://github.com/rubygems/rubygems/issues/448#issuecomment-716509271).

I propose to promote it to a global flag that can be used for any command.

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

I will abide by the [code of conduct](https://github.com/rubygems/rubygems/blob/master/CODE_OF_CONDUCT.md).

bronzdoc:
What other commands could benefit from the `-C` flag besides the `install` command?

deivid-rodriguez:
I think a global `-C` flag can be useful in general for scripting, but I don't have specific ideas of other commands that could use it. I you don't think this is useful globally, we can then add `-C` for `gem install` only. I don't have a very strong opinion, really.

deivid-rodriguez:
Let's add it for `gem install` only to fix the referenced issue.

PR #6180, opened 2022-12-22
Add global flag (`-C`) to change execution directory

### LINKED ISSUE #448: $ gem install should look for dependencies in the same directory as the dependant gem

gustavothecoder:

`gem install`, when installing a `.gemspec` file directly, looks for dependencies of the gemspec in other gemspec files present in the CWD. This works when the target `.gemspec` is in the CWD but not if it's somewhere else.

So if I have `a.gemspec` and `b.gemspec`, where `b` is a dependency if `a`, at folder `foo/bar`, then `cd foo/bar; gem install a.gemspec` works but `gem install foo/bar/a.gemspec` does not work.

This PR does not fix the issue per se, but introduces a `-C` flag to simplify making this work, by running, `gem -C foo/bar a.gemspec`. So we'll make this PR tentatively close #448 and wait for feedback in case someone still needs #448 itself after this.

As discussed before (#448 and #4058), and explained above, this solution introduces a flag (`-C`) that gives users the ability to change the execution directory from the current one to another. This flag was inspired by [git -C](https://git-scm.com/docs/git#Documentation/git.txt--Cltpathgt) and [ruby -C](https://linux.die.net/man/1/ruby).

Although the problem reported in #448 is related to `gem install`, implementing `-C` as a global flag provides consistency with other command line tools and exposes this feature to other commands.

`gem build` already has this same option, so I added a deprecation warning for it.

Closes #448.

### DIFF
--- a/lib/rubygems/command.rb
+++ b/lib/rubygems/command.rb
@@ -630,7 +630,11 @@ RubyGems is a package manager for Ruby.
   Usage:
     gem -h/--help
     gem -v/--version
-    gem command [arguments...] [options...]
+    gem [global options...] command [arguments...] [options...]
+
+  Global options:
+    -C PATH                      run as if gem was started in <PATH>
+                                 instead of the current working directory
 
   Examples:
     gem install rake
--- a/lib/rubygems/commands/build_command.rb
+++ b/lib/rubygems/commands/build_command.rb
@@ -26,6 +26,9 @@ class Gem::Commands::BuildCommand < Gem::Command
     add_option "-C PATH", "Run as if gem build was started in <PATH> instead of the current working directory." do |value, options|
       options[:build_path] = value
     end
+    deprecate_option "-C",
+                     version: "4.0",
+                     extra_msg: "-C is a global flag now. Use `gem -C PATH build GEMSPEC_FILE [options]` instead"
   end
 
   def arguments # :nodoc:
--- a/test/rubygems/test_gem_command_manager.rb
+++ b/test/rubygems/test_gem_command_manager.rb
@@ -126,6 +126,46 @@ class TestGemCommandManager < Gem::TestCase
     @command_manager.unregister_command :crash
   end
 
+  def test_process_args_with_c_flag
+    custom_start_point = File.join @tempdir, "nice_folder"
+    FileUtils.mkdir_p custom_start_point
+
+    execution_path = nil
+    use_ui @ui do
+      @command_manager[:install].when_invoked do
+        execution_path = Dir.pwd
+        true
+      end
+      @command_manager.process_args %W[-C #{custom_start_point} install net-scp-4.0.0.gem --local]
+    end
+
+    assert_equal custom_start_point, execution_path
+  end
+
+  def test_process_args_with_c_flag_without_path
+    use_ui @ui do
+      assert_raise Gem::MockGemUi::TermError do
+        @command_manager.process_args %w[-C install net-scp-4.0.0.gem --local]
+      end
+    end
+
+    assert_match(/install isn't a directory./i, @ui.error)
+  end
+
+  def test_process_args_with_c_flag_path_not_found
+    custom_start_point = File.join @tempdir, "nice_folder"
+    FileUtils.mkdir_p custom_start_point
+    custom_start_point.tr!("_", "-")
+
+    use_ui @ui do
+      assert_raise Gem::MockGemUi::TermError do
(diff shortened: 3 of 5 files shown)


lib/rubygems/command.rb
```diff
@@ -630,7 +630,7 @@ def wrap(text, width) # :doc:
   Usage:
     gem -h/--help
     gem -v/--version
-    gem command [arguments...] [options...]
+    gem [-C PATH] command [arguments...] [options...]
```
gustavothecoder, line 633: Is there a better way to document this type of option?

lib/rubygems/commands/build_command.rb
```diff
@@ -26,6 +26,8 @@ def initialize
     add_option "-C PATH", "Run as if gem build was started in <PATH> instead of the current working directory." do |value, options|
       options[:build_path] = value
     end
+    deprecate_option "-C",
+                     extra_msg: "-C is a global flag now. Use `gem -C PATH build GEMSPEC_FILE [options]` instead"
```
gustavothecoder, line 30: If you guys want, I can add a version here.

test/rubygems/test_gem_command_manager.rb
```diff
@@ -126,6 +126,46 @@ def test_run_crash_command
...
+    custom_start_point.tr!("_", "-")
+
+    use_ui @ui do
+      assert_raise Gem::MockGemUi::TermError do
+        @command_manager.process_args %W[-C #{custom_start_point} install net-scp-4.0.0.gem --local]
+      end
+    end
+
+    assert_match(/#{custom_start_point} isn't a directory./i, @ui.error)
+  end
```
gustavothecoder, line 167: My test approach only cares about the `-C` behavior. Is this sufficient or we need to write test to all comands?

lib/rubygems/command.rb
deivid-rodriguez: I guess we could add a "Global options:" section below "Usage:"? What do you think?

lib/rubygems/commands/build_command.rb
deivid-rodriguez: Let's say 4.0 I guess.

test/rubygems/test_gem_command_manager.rb
deivid-rodriguez: What you did seems fine to me!

lib/rubygems/command.rb
gustavothecoder: Good idea.

deivid-rodriguez:
Thanks @gustavothecoder, it looks very good!

I'm releasing RubyGems 3.4.0 now, but will merge & release this one with a future release (RubyGems 3.4.1 hopefully!).

deivid-rodriguez (commented):
Looks good, only thing I'm missing is maybe a test for the deprecation of `gem build -C`.

gustavothecoder:
> Looks good, only thing I'm missing is maybe a test for the deprecation of `gem build -C`.

Added!

deivid-rodriguez (approved):
😍

deivid-rodriguez:
@simi Any opinion on this? In particular, you've been against shipping deprecations with patch level releases. I _believe_ for this relatively minor usage it's fine, but if not, we can split this PR, add the enhancement to the next version, and leave the deprecation for next year's release.

simi:
> @simi Any opinion on this? In particular, you've been against shipping deprecations with patch level releases. I _believe_ for this relatively minor usage it's fine, but if not, we can split this PR, add the enhancement to the next version, and leave the deprecation for next year's release.

This seems ok to me.

simi (approved):

deivid-rodriguez:
Awesome, thanks @simi and @gustavothecoder! ❤️

deivid-rodriguez:
This PR will close the oldest issue we currently have opened 💪.

PR #6187, opened 2022-12-24
Allow Bundler to run on old RubyGems + Ruby 2.7 without warnings

deivid-rodriguez:

We may have been a little too aggressive here. Users running Ruby 2.7 which is still supported for a few months are getting this warning.

Don't show the warning on Ruby 2.7. We can reintroduce this once Ruby 2.7 reaches end of life.

### DIFF
--- a/bundler/exe/bundle
+++ b/bundler/exe/bundle
@@ -15,7 +15,7 @@ else
   require "bundler"
 end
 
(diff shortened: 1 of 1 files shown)


indirect:
I think this is good, and it is ok to trail ruby-core a little bit on support. In the old days, we supported all the way back to Ruby 1.8, mostly because it benefits us to be able to upgrade even very old users to new bundler version fixes or improvements. 👍🏻

PR #6196, opened 2022-12-25
Hide `bundle lock --update --bundler` from changelog

deivid-rodriguez:

Just noticed the `bundle lock --update --bundler` does not work as advertised.

No fix yet, but I think it's best to not advertise the feature for now since it doesn't really work.


deivid-rodriguez:
I'm going to give fixing this a try.

PR #6206, opened 2022-12-27
Renable realworld specs skipped for some time

deivid-rodriguez:

I noticed that we still have some realworld specs checking output from Molinillo. However, we had not noticed because these specs had been disable for some time now. The reason they have been disabled is that I set them to run only when a specific patch level ruby is used (Ruby 2.7.3, and Ruby 3.0.3), so they got disabled when we started using other patch levels. I should've used a more flexible range instead (2.7, and 3.0). 

My initial fix is to reenable these specs.

But I wonder if they are worth the complexity of this setup. They does not seem to have caught any issues with the migration to Pub Grub and for a while I've tried to avoid adding realworld specs for testing the resolver and make an effort to reduce tests for fixes to something that can be reproduced with a simpler test setup.


deivid-rodriguez:
I'll introduce this to completely get rid of any traces from Molinillo. We can let these realworld specs that are only valid for some rubies go when we finally drop support for those rubies. At that point, the new resolver should be properly battle tested, so it should be fine.

Issue #883, opened 2014-04-14
Use different separator then \0 for stub line


voxik:
Commit https://github.com/rubygems/rubygems/commit/652e31942e807ae4625d2cc1eb771cd296c8d1ed introduced stub lines and is using \0 as a item separator. This is unfortunate choice, since it confuses some tools. Namely, diff cannot do diff of such files, complaining about binary files, gedit opens such file just as a garbage. If .gemspec is considered text file and ruby source code, it should not contain any NULL characters.

copiousfreetime:
This should be changeable since it only affects the `# stub:` lines in specifications. See:
* https://github.com/rubygems/rubygems/blob/master/lib/rubygems/stub_specification.rb
* https://github.com/rubygems/rubygems/blob/master/lib/rubygems/specification.rb#L2471-L2472

bronzdoc:
What separator do you recomend @voxik? are you still interested in this feature?

voxik:
Dunno, semicolon? Of course I still consider NULL character in text file as a bad idea. Not sure if the tools were fixed ...

deivid-rodriguez:
I guess we could use `File::PATH_SEPARATOR` since this separator seems to be used for separating paths?

voxik:
> I guess we could use `File::PATH_SEPARATOR` since this separator seems to be used for separating paths?

Isn't it different for *NIX/Windows?

deivid-rodriguez:
Yes, does it matter?

voxik:
It will make sharing the .gemspec between OSes harder?

I don't know how often this might be real issue, but there were times when I developed on Windows and deployed on Solaris. So I can imagine I would (use Bundler to) vendor the gems and I had a troubles. Dunno. That is just speculation.

deivid-rodriguez:
Maybe? Note that these are just installed gemspec stubs, so I don't think we'd want to support share those. The only alternative I can think if is... `\0` because all the other candidates are valid characters on a filesystem path... right? So changing this would be exchanging a tiny minor annoyance with another tiny minor annoyance?

MSP-Greg:
Re `File::PATH_SEPARATOR`, Windows Ruby and almost all recent Windows versions work fine with a forward-slash.  The only problem happens when string comparisons are done in code.  You probably already know this, but a reminder never hurts...

deivid-rodriguez:
[...]

PR #6209, opened 2022-12-27
Change stub line separator to `:`

gustavothecoder:

#883

As discussed in the linked issue, my fix is to change the stub line separator to `:`. 

It's important to know that **this fix affects only the new gemspecs**, the ones already generated still have `\0` as path separator. Still, I think this fix solves #883, because if the user wants to diff the gemspec or open it with a text editor, he just needs to reinstall de gem.

### Before
```bash
$ diff json-2.6.3.gemspec json-schema-3.0.0.gemspec
Binary files json-2.6.3.gemspec and json-schema-3.0.0.gemspec differ
```
### After
```bas
$ diff json-2.6.3.gemspec json-schema-3.0.0.gemspec
2,3c2
< # stub: json 2.6.3 ruby lib
< # stub: ext/json/ext/generator/extconf.rb:ext/json/ext/parser/extconf.rb:ext/json/extconf.rb
---
> # stub: json-schema 3.0.0 ruby lib
6,7c5,6
<   s.name = "json".freeze
<   s.version = "2.6.3"
---
>   s.name = "json-schema".freeze
>   s.version = "3.0.0"

# more differences...
```

_*It was necessary to reinstall both gems to make the diff works._

Originally, the `foo-0.0.1-x86-mswin32.gemspec` has a stub line with multiple paths: `# stub: foo 0.0.1 mswin32 lib\0lib/f oo/ext`. I changed the `\0` to `:`.


simi:
You can force diff using `diff --text` to not be "clever". There was no need for this in last 4 years.

Also it was considered to use `:` in the end (https://github.com/rubygems/rubygems/issues/883#issuecomment-701269570) if this change is needed.

eregon:
I think this would be nice to solve, and it seems a fairly straightforward change.

test/rubygems/test_gem_stub_specification.rb
```diff
@@ -15,11 +15,11 @@ def setup
   end
 
   def test_initialize
-    assert_equal "foo", @foo.name
-    assert_equal Gem::Version.new("0.0.1"), @foo.version
+    assert_equal "foo",                        @foo.name
```
voxik, line 18: Wouldn't be better instead of adding the alignment here to remove it elsewhere? Looking the very same diff side by side [here](https://github.com/rubygems/rubygems/pull/6209/files#diff-c41f72b806d91dc3bf46c805909ef4d091e63a35cbae5fd2349c3d6dafac50b7R36)the lines are wrapped just because of the spaces and the end result is less readable.

Anyway, style changes like this should be separate commit IMHO.

deivid-rodriguez: I agree with both: I prefer the non aligned style, and I prefer not mixing in style changes 👍.

deivid-rodriguez:
I think this PR should be fine too. @simi do you have any concerns? I also think this is no big deal, but it seems an easy fix.

simi:
What about scenario when you install gem amd than downgrade RubyGems?

My concern is simple, it causes currently zero to almost no harm. I'm afraid to fix problem just to close old issue with no activity.

deivid-rodriguez:
Good point with regards to downgrading!

I'm introducing `--regenerate-default-specifications` for that at #4199, like I did with plugins when I sped up RubyGems at https://github.com/rubygems/rubygems/pull/3108. This PR could pick up the implementation for that from #4199.

That said, I don't really have a strong opinion here, if you think there's some risk that makes this not worth effort it, we can just close this.

deivid-rodriguez:
> My concern is simple, it causes currently zero to almost no harm.

Sure, no harm of course, but it's true that text files with a single binary character are a bit annoying. Of course, only power users like ourselves that may mess with the RubyGems installation will benefit from this, but we're users too :)

test/rubygems/test_gem_stub_specification.rb
gustavothecoder: I removed the style change from my commit.

simi:
> That said, I don't really have a strong opinion here, if you think there's some risk that makes this not worth it, we can just close this.

> Sure, no harm of course, but it's true that text files with a single binary character are a bit annoying. Of course, only power users like ourselves that may mess with the RubyGems installation will benefit from this, but we're users too :)

Let me summarize my thinking of this change.

- yes, I agree it is (unfortunately) wrongly picked path separator
- separator change is useful only for "hackers" and there is usually simple "bandaid" (like using `diff --text` to get diff)
- there is no benefit for "end-users", but there is chance things get broken on their side
  - it is not backwards compatible
  - there is no automated migration (next to manual fresh `--regenerate-default-specifications`)
- this issue was opened by @gustavothecoder to help triaging old issues, but not to solve any real problem for the reporter (btw. thanks @gustavothecoder for doing this, it is super helpful :heart: )

Considering all this I would be ok to introduce this in RubyGems 4 with some migration (regenerating specs on update) and ensure RubyGems 4 can't be downgraded (or make migration both ways). But I'm afraid this could cause unexpected troubles for users on minor upgrade with no simple way of fixing it.

deivid-rodriguez:
> Considering all this I would be ok to introduce this in RubyGems 4 with some migration (regenerating specs on update) and ensure RubyGems 4 can't be downgraded (or make migration both ways). But I'm afraid this could cause unexpected troubles for users on minor upgrade with no simple way of fixing it.

Sounds good! Note that I already handled a very similar migration with the plugin change layout via automatic regeneration on update, and automatic revert to previous format on downgrade. We got no reports about anything related that I can remember so I believe that approach worked fine.

eregon:
> but not to solve any real problem for the reporter

I disagree, the following may sound like complaining but I just want to explain how annoying and how much a problem it is.
I would say it's a real problem and I already commented on that issue (I'm not the reporter because the issue was already reported). For instance every time anyone needs to look at an installed gemspec with multiple native exts, they need to find a workaround with the tool to actually be able to read it (some tools don't have such a workaround, a null byte is basically illegal in a text file). It looks like a binary file to every tool including git repositories and BitBucket (e.g., no diff shown in PRs and no direct workaround for that, number of bytes vs lines for `git log --stat` when reviewing a Ruby update, etc). So at least it is a real nuisance I see multiple times a year easily.
Also the \0 makes it look pretty confusing on GitHub: https://github.com/oracle/truffleruby/blob/master/lib/gems/specifications/default/json-2.6.1.gemspec

Maybe Ruby should even disallow \0 bytes even in comments, I wouldn't be surprised if that can cause security issues in some cases (e.g., C code cannot rely on strlen() to know the end of that snippet of Ruby, wouldn't be surprised if some C-ext or C code in CRuby does that).

eregon:
My general thought on this fast-to-access data is it should be in proper binary files (because that's a lot faster to get the relevant data), as mentioned in https://github.com/rubygems/rubygems/pull/4199#issuecomment-1207212755. We should not use non-text bytes inside comments, that just looks like a hack, in addition to being inefficient and a lot more complex to update.
If it was separate binary files, then if the format was incompatible between 2 different RubyGems version it would be very easy to deal with: just load the gemspec and regenerate the corresponding binary file, that could be all automatic.

simi:
@eregon by `but not to solve any real problem for the reporter` I was refering to @gustavothecoder's motivation to open this PR.

hsbt:
I decided not to change the stub line format: it is machine-read only, so its readability is not a reason to change it. See #883.

Issue #6210, opened 2022-12-28
No live threads left. Deadlock?


Priyanka328:
--- ERROR REPORT TEMPLATE -------------------------------------------------------
    
    ```
    fatal: No live threads left. Deadlock?
    10 threads, 10 sleeps current:0x000055ae80cb1d70 main thread:0x000055ae7fea1ff0
    * #<Thread:0x000055ae7feceea0 sleep_forever>
       rb_thread_t:0x000055ae7fea1ff0 native:0x00007fd4ee229dc0 int:0
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/worker.rb:42:in `pop'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/worker.rb:42:in `deq'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer/parallel_installer.rb:173:in `process_specs'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer/parallel_installer.rb:133:in `install_with_worker'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer/parallel_installer.rb:89:in `call'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer/parallel_installer.rb:67:in `call'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer.rb:244:in `install_in_parallel'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer.rb:201:in `install'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/installer.rb:89:in `block in run'
       /home/dockeruser/.local/share/autoproj/gems/ruby/2.7.0/gems/bundler-2.4.1/lib/bundler/process_lock.rb:12:in `block in lock'
[...]

Issue #6188, opened 2022-12-24
Bundler 2.4.0 fails to resolve dependencies


jgraichen:
### Describe the problem as clearly as you can

Bundler dependency solving fails to resolve `ffi` in the following excerpt from a production application. It does work fine with `2.3.14`.

It seems to wrongly collapse dependency version constraints when there are multiple alternatives.

### Did you try upgrading rubygems & bundler?

This happens after upgrading to the latest version `2.4.0`.

```console
$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.4.0
  - RUBY VERSION: 3.1.2 (2022-04-12 patchlevel 20) [x86_64-linux]

$ gem list bundler

*** LOCAL GEMS ***

bundler (2.4.0, 2.3.14, 2.3.11, 2.3.8)
```

### Post steps to reproduce the problem

Take the following Gemfile snippet from a Rails application:

```ruby
source 'https://rubygems.org'

gem 'listen', '~> 3.3'
gem 'ffi', platforms: [:mingw, :x64_mingw, :mswin]
```

With `2.4.0`, bundler fails to resolve dependencies:

```console
# bundle _2.4.0_ install
Fetching gem metadata from https://rubygems.org/............
Resolving dependencies...
Could not find compatible versions

Because listen >= 3.3.0, < 3.3.1 depends on rb-inotify >= 0.9.10, < 1.A
  and listen >= 3.3.1, < 3.3.2 depends on rb-inotify >= 0.9.10, < 1.A,
  listen >= 3.3.0, < 3.3.2 requires rb-inotify >= 0.9.10, < 1.A.
And because listen >= 3.3.2, < 3.3.3 depends on rb-inotify >= 0.9.10, < 1.A
  and listen >= 3.3.3, < 3.3.4 depends on rb-inotify >= 0.9.10, < 1.A,
  listen >= 3.3.0, < 3.3.4 requires rb-inotify >= 0.9.10, < 1.A.
And because listen >= 3.3.4, < 3.4.0 depends on rb-inotify >= 0.9.10, < 1.A
  and listen >= 3.4.0, < 3.4.1 depends on rb-inotify >= 0.9.10, < 1.A,
  listen >= 3.3.0, < 3.4.1 requires rb-inotify >= 0.9.10, < 1.A.
And because listen >= 3.4.1, < 3.5.0 depends on rb-inotify >= 0.9.10, < 1.A
[...]

PR #6215, opened 2022-12-29
Speed up resolution by properly merging incompatibility ranges

deivid-rodriguez:

Sometimes PubGrub is very slow, like with this Gemfile:

https://github.com/rubygems/rubygems/blob/e1540e4bf64adcb9c98494e79f692762b3027f4d/bundler/spec/realworld/edgecases_spec.rb#L359-L519

Sometimes it also hangs while propagating incompatibilities, like with the Gemfile reported [here](https://github.com/rubygems/rubygems/issues/6210#issuecomment-1367360624).

While PubGrub's builtin package source relies only on comparing requirement strings for merging incompatibility ranges, our custom package source relies on comparing full `VersionContraint` instances.

This patch should speed things up. As an example, the nix Gemfile above takes on my machine about 30 seconds to resolve with this patch, and about 7 minutes without it.

It should also give more compact explanations. For example, the weird explanation (with a lot of tiny version ranges) reported at https://github.com/rubygems/rubygems/issues/6188 should now be much simpler (although that case now resolves fine since the issue is fixed, but just to give an example).

### DIFF
--- a/bundler/lib/bundler/vendor/pub_grub/lib/pub_grub/version_constraint.rb
+++ b/bundler/lib/bundler/vendor/pub_grub/lib/pub_grub/version_constraint.rb
@@ -20,6 +20,10 @@ module Bundler::PubGrub
         range.eql?(other.range)
     end
 
+    def ==(other)
+      package == other.package && range == other.range
+    end
+
     class << self
       def exact(package, version)
         range = VersionRange.new(min: version, max: version, include_min: true, include_max: true)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I proposed the change upstream to pub_grub at https://github.com/jhawthorn/pub_grub/pull/21. Setting this as ready!

indirect (approved):
awesome 🙌🏻

PR #6217, opened 2022-12-30
Improve documentation about `Kernel` monkeypatches

nobu:

https://bugs.ruby-lang.org/issues/19285

> It seems we merge all of the docs for stdlib and core into one for docs.ruby-lang.org, this creates weirdness like [Kernel](https://docs.ruby-lang.org/en/master/Kernel.html) which includes monkey patches for rubygems for example.
> 
> > RubyGems adds the gem method to allow activation of specific gem versions and overrides the require method on Kernel to make gems appear as if they live on the $LOAD_PATH. See the documentation of these methods for further detail.

This seems an implementation detail but not for users.

Hide this comment from RDoc.

### DIFF
--- a/lib/rubygems/core_ext/kernel_gem.rb
+++ b/lib/rubygems/core_ext/kernel_gem.rb
@@ -1,9 +1,4 @@
 # frozen_string_literal: true
-##
-# RubyGems adds the #gem method to allow activation of specific gem versions
-# and overrides the #require method on Kernel to make gems appear as if they
-# live on the <code>$LOAD_PATH</code>.  See the documentation of these methods
-# for further detail.
 
 module Kernel
 
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -119,10 +119,6 @@ module Gem
   # to avoid deprecation warnings in Ruby 2.7.
   UNTAINT = RUBY_VERSION < "2.7" ? :untaint.to_sym : proc {}
 
-  # When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
-  KERNEL_WARN_IGNORES_INTERNAL_ENTRIES = RUBY_ENGINE == "truffleruby" ||
-                                         (RUBY_ENGINE == "ruby" && RUBY_VERSION >= "3.0")
-
   ##
   # An Array of Regexps that match windows Ruby platforms.
 
@@ -1348,7 +1344,16 @@ end
 Gem::Specification.load_defaults
 
 require_relative "rubygems/core_ext/kernel_gem"
-require_relative "rubygems/core_ext/kernel_require"
-require_relative "rubygems/core_ext/kernel_warn"
+
+path = File.join(__dir__, "rubygems/core_ext/kernel_require.rb")
+# When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
+if RUBY_ENGINE == "truffleruby" ||
+   (RUBY_ENGINE == "ruby" && RUBY_VERSION >= "3.0")
+  file = "<internal:#{path}>"
+else
+  require_relative "rubygems/core_ext/kernel_warn"
+  file = path
+end
+eval File.read(path), nil, file
 
 require ENV["BUNDLER_SETUP"] if ENV["BUNDLER_SETUP"] && !defined?(Bundler)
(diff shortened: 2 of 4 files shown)


deivid-rodriguez:
Should we make this appear at https://docs.ruby-lang.org/en/master/Kernel.html#method-i-require instead? It does seem useful there to me.

nobu:
Already `lib/rubygems/core_ext/kernel_require.rb` has the description, but rdoc ignores it as it is in a here doc string.

deivid-rodriguez:
Ah I see! Do you know of any workaround for that?

nobu:
Just an idea.
```diff
diff --git i/lib/rubygems.rb w/lib/rubygems.rb
index 73a9848bf..70e51a996 100644
--- i/lib/rubygems.rb
+++ w/lib/rubygems.rb
@@ -119,10 +119,6 @@
   # to avoid deprecation warnings in Ruby 2.7.
   UNTAINT = RUBY_VERSION < "2.7" ? :untaint.to_sym : proc {}
 
-  # When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
-  KERNEL_WARN_IGNORES_INTERNAL_ENTRIES = RUBY_ENGINE == "truffleruby" ||
-                                         (RUBY_ENGINE == "ruby" && RUBY_VERSION >= "3.0")
-
   ##
   # An Array of Regexps that match windows Ruby platforms.
 
@@ -1348,7 +1344,16 @@
 Gem::Specification.load_defaults
 
 require_relative "rubygems/core_ext/kernel_gem"
-require_relative "rubygems/core_ext/kernel_require"
+
+path = File.join(__dir__, "rubygems/core_ext/kernel_require.rb")
+# When https://bugs.ruby-lang.org/issues/17259 is available, there is no need to override Kernel#warn
+if RUBY_ENGINE == "truffleruby" ||
+   (RUBY_ENGINE == "ruby" && RUBY_VERSION >= "3.0")
+  file = "<internal:#{path}>"
+else
   require_relative "rubygems/core_ext/kernel_warn"
+  file = path
+end
+module_eval File.read(path), file
 
 require ENV["BUNDLER_SETUP"] if ENV["BUNDLER_SETUP"] && !defined?(Bundler)
diff --git i/lib/rubygems/core_ext/kernel_require.rb w/lib/rubygems/core_ext/kernel_require.rb
index 8064d813e..267cfcd33 100644
--- i/lib/rubygems/core_ext/kernel_require.rb
+++ w/lib/rubygems/core_ext/kernel_require.rb
@@ -17,8 +17,6 @@
     private :gem_original_require
   end
 
-  file = Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES ? "<internal:#{__FILE__}>" : __FILE__
-  module_eval <<'RUBY', file, __LINE__ + 1 # rubocop:disable Style/EvalWithLocation
   ##
   # When RubyGems is required, Kernel#require is replaced with our own which
   # is capable of loading gems on demand.
@@ -168,7 +166,6 @@
       end
     end
   end
-RUBY
 
   private :require
 
diff --git i/lib/rubygems/core_ext/kernel_warn.rb w/lib/rubygems/core_ext/kernel_warn.rb
index f0f173c0b..1f4c77f04 100644
--- i/lib/rubygems/core_ext/kernel_warn.rb
+++ w/lib/rubygems/core_ext/kernel_warn.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-if !Gem::KERNEL_WARN_IGNORES_INTERNAL_ENTRIES
-
 module Kernel
   rubygems_path = "#{__dir__}/" # Frames to be skipped start with this path.
 
@@ -50,4 +48,3 @@
     original_warn.bind(self).call(*messages, **kw)
   }
 end
-end
```

nobu:
Also `:doc:` directive  is needed to display `require` in kernel_require.rb.

deivid-rodriguez (approved):
It feels really nice, not only to have this documented at the right place, but also to no longer have our whole `require` redefinition wrapped in a heredoc.

lib/rubygems/core_ext/kernel_gem.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
-##
+#--
 # RubyGems adds the #gem method to allow activation of specific gem versions
 # and overrides the #require method on Kernel to make gems appear as if they
 # live on the <code>$LOAD_PATH</code>.  See the documentation of these methods
 # for further detail.
+#++
```
deivid-rodriguez, line 7: I think we can completely get rid of the above paragraph, actually. The same information is provided at `#gem` and `#require` documentation, and feels out of context here.

deivid-rodriguez:
There was no further feedback here, so I'm merging this, thanks so much @nobu!

PR #6218, opened 2022-12-30
Get `Gem::Specification#extensions_dir` documented

deivid-rodriguez:

We should make `Gem::Specification#extensions_dir` documented at https://guides.rubygems.org/specification-reference/.

Unfortunately the scripts for generating the site are [a bit weird](https://github.com/rubygems/guides/blob/cade28179dc3b47836660a39bf0b69d7020dca6f/rdoc/generator/template/jekdoc/classpage.rhtml).

This change changes `Gem::Specification` hopefully in a harmless way, so that the method appears documented on our site.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -531,13 +531,6 @@ class Gem::Specification < Gem::BasicSpecification
   attr_reader :required_rubygems_version
 
   ##
-  # The version of RubyGems used to create this gem.
-  #
-  # Do not set this, it is set automatically when the gem is packaged.
-
-  attr_accessor :rubygems_version
-
-  ##
   # The key used to sign this gem.  See Gem::Security for details.
 
   attr_accessor :signing_key
@@ -722,6 +715,21 @@ class Gem::Specification < Gem::BasicSpecification
 
   def test_files=(files) # :nodoc:
     @test_files = Array files
+  end
+
+  ######################################################################
+  # :section: Read-only attributes
+
+  ##
+  # The version of RubyGems used to create this gem.
+
+  attr_accessor :rubygems_version
+
+  ##
+  # The path where this gem installs its extensions.
+
+  def extensions_dir
+    @extensions_dir ||= super
   end
 
   ######################################################################
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -351,9 +351,8 @@ module Gem
         end
 
         def extensions_dir
-          Gem.default_ext_dir_for(base_dir) ||
-            File.join(base_dir, "extensions", ORIGINAL_LOCAL_PLATFORM,
-                      Gem.extension_api_version)
+          @extensions_dir ||=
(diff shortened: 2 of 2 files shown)


eloyesp:
This is great, having a single documented way to access the `extensions_dir` should solve most of our issues and would create an important API. To make sure that we follow the contract, we should also make use of this API within the test suite, I'm thinking on those two specs that kept failing when [I were working on this](https://github.com/rubygems/rubygems/pull/4017#issuecomment-1352516109). That way we make sure that we follow the contract that we offer.

deivid-rodriguez:
Thanks for the encouragement @eloyesp!

I agree we should use our API ourselves internally, however, as the trouble you run into at the time proves, we're not there yet because git sources in Bundler currently rely on extensions being installed into `lib`.

We should fix this, but it's not an easy one... 😢. Currently git sources work but they are a bit of a hack. I think effectively that means we won't be able to stop installing into `lib` for a while, unfortunately.

That said, I think we can start promoting and documenting our preferred API right away.

deivid-rodriguez:
I'm going to include this with next release if that's ok.

deivid-rodriguez:
@simi Any concerns about this, or suggested alternative?

This does bring the slight functional change of getting this directory memoized. It seems to me it shouldn't have any side effects, but I'm good at being wrong 😄. The motivation for this is to "force" the method to be picked up as a method in `Gem::Specification` by the scripts in https://github.com/rubygems/guides. I briefly looked into changing the scripts to also look in the parent class, but it seemed not straightforward so I went with this alternative.

lib/rubygems/specification.rb
```diff
@@ -726,6 +719,21 @@ def test_files=(files) # :nodoc:
...
+  ##
+  # The version of RubyGems used to create this gem.
+
+  attr_accessor :rubygems_version
+
+  ##
+  # The path where this gem installs its extensions.
+
+  def extensions_dir
+    @extensions_dir ||= super
```
simi, line 734: Any reason to not use just `super` and make it clear in comment the only purpose of this re-definition is spec reference completeness. It should be possible to avoid this commend in reference guide using `#--` directive (https://ruby.github.io/rdoc/RDoc/Markup.html#class-RDoc::Markup-label-Documenting+Source+Code).

deivid-rodriguez: There was no strong reason, I just thought that memoizing would not heart, but your idea makes this patch less invasive and more focused on what we want to achieve. I will update it as suggested, thanks!

simi: I'm trying to be more conservative to not introduce additional problems for extensions now. :pray:

simi:
@deivid-rodriguez Looking at this again, it seems safe. I have rebased the branch, looking for CI now :pray:. Let's do the final discussion once it is :green_circle:.

deivid-rodriguez:
Sorry I dropped the ball here @simi. You meant that my initial version is ok, and should I still update the PR as you suggested?

deivid-rodriguez:
Ping @simi on my previous question :)

simi:
> Sorry I dropped the ball here @simi. You meant that my initial version is ok, or should I still update the PR as you suggested?

@deivid-rodriguez Both works for me. Once you make this ready for review, it is ready to merge IMHO.

PR #6219, opened 2022-12-30
Let resolver deal with legacy gems with equivalent version and different dependencies

deivid-rodriguez:

Some legacy gems could have equivalent versions like 1.10.rc2 and 1.10.0.rc2 with different dependencies.

This makes our new resolver crash.

My fix is to make our deduplication of version candidates deal with this.

Not totally sold we should do anything about this though.

### DIFF
--- a/bundler/lib/bundler/resolver.rb
+++ b/bundler/lib/bundler/resolver.rb
@@ -215,7 +215,7 @@ module Bundler
 
     def all_versions_for(package)
       name = package.name
-      results = (@base[name] + @all_specs[name]).uniq(&:full_name)
+      results = (@base[name] + @all_specs[name]).uniq {|spec| [spec.version.hash, spec.platform] }
       locked_requirement = base_requirements[name]
       results = filter_matching_specs(results, locked_requirement) if locked_requirement
 
--- a/bundler/spec/commands/install_spec.rb
+++ b/bundler/spec/commands/install_spec.rb
@@ -1076,4 +1076,31 @@ RSpec.describe "bundle install with gem sources" do
       G
     end
   end
+
+  context "when a gem has equivalent versions with inconsistent dependencies" do
+    before do
+      build_repo4 do
+        build_gem "autobuild", "1.10.rc2" do |s|
+          s.add_dependency "utilrb", ">= 1.6.0"
+        end
+
+        build_gem "autobuild", "1.10.0.rc2" do |s|
+          s.add_dependency "utilrb", ">= 2.0"
+        end
+      end
+    end
+
+    it "does not crash unexpectedly" do
+      gemfile <<~G
+        source "#{file_uri_for(gem_repo4)}"
+
+        gem "autobuild", "1.10.rc2"
+      G
+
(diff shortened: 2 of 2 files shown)


indirect (approved):
Even if we chose to instead yank every gem like this from rubygems.org, we would still need to implement some kind of solution for this issue that is not a deadlock—other gem sources like S3 or gemstash or geminabox could easily be out there serving up this same kind of gem version conflict, happily working for all users until they upgrade to Bundler 2.4.

Because of that, I am in favor of merging this PR, following the principle of "if it worked before it should keep working, so that users stay confident they can upgrade to newer bundler versions".

(I am a little bit worried about adding 500kb to the git repo just for this one test, but not worried enough to block the PR over it.)

bundler/spec/support/artifice/vcr_cassettes/realworld/index.rubygems.org/info/hoe-yard/GET/response
```diff
@@ -0,0 +1,34 @@
...
+x-frame-options: SAMEORIGIN
+x-xss-protection: 1; mode=block
+x-content-type-options: nosniff
+x-download-options: noopen
+x-permitted-cross-domain-policies: none
+referrer-policy: strict-origin-when-cross-origin
+etag: "fcba05bc4488cd21065b564db5c0e047"
+last-modified: Mon, 14 Dec 2020 16:36:07 GMT
+cache-control: max-age=60, public
+content-security-policy: default-src 'self'; font-src 'self' https://fonts.gstatic.com; img-src 'self' https://secure.gaug.es https://gravatar.com https://secure.gravatar.com https://*.fastly-insights.com; object-src 'none'; script-src 'self' https://secure.gaug.es https://www.fastly-insights.com; style-src 'self' https://fonts.googleapis.com; connect-src 'self' https://s3-us-west-2.amazonaws.com/rubygems-dumps/ https://*.fastly-insights.com https://fastly-insights.com https://api.github.com; form-action 'self'; frame-ancestors 'self'
```
simi, line 14: Btw. I do wonder why we do set these for `index.rubygems.org`. I'll take a look.

indirect: my guess is that it is probably a Fastly VCL header that is applied to all domains

deivid-rodriguez:
You're totally right @indirect.

Regarding the added size, I actually managed to reduce this to a "regular spec" but I needed to create a lot of candidates for this `autobuild` gem, and I couldn't figure out how to reduce it further. As it was, I didn't like the spec because it was not minimal at all and it was actually slower than this realworld spec, due to the overhead of creating that many gems.

I will give it another try to reduce this even further, if that's ok. I agree to do something about this but I think it can wait a bit.

indirect:
That makes sense! I am ok with adding the spec as-is if it's not clear how to reduce the size more. 👍🏻

deivid-rodriguez:
I managed to write a more focused spec! 💪

PR #6220, opened 2022-12-31
Remove stray word in `bundle config` man page.

markdoliner:
There was an extra word ("with") in this sentence.

I only modified the `.ronn` file and not `bundle-config.1`. Please let me know if you would like me to modify the latter, too.


deivid-rodriguez:
We do need to run `bin/rake man:build` from the `bundler/` folder in the repo, to get the other page regenerated. But I think you just contributed this change from the UI, so I can do it myself if necessary.

Thanks for fixing this ❤️.

deivid-rodriguez (approved):

PR #4202, opened 2020-12-26
feat(CI): add a GitHub Action to check spelling

jbampton:
- Run on pull request only
- Using https://github.com/client9/misspell
- Pass in `-w` to autocorrect.

## What was the end-user or developer problem that led to this PR?

Need to have correct spelling

## What is your fix for the problem, implemented in this PR?

Add a Golang based spell checker to the CI to automatically check for mistakes.

From my experience it won't find all mistakes in all file types.

## Make sure the following tasks are checked

- [X] Describe the problem / feature
- [X] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [X] Write code to solve the problem
- [X] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6224, opened 2023-01-01
Fix several typos

jdufresne:


deivid-rodriguez (commented):
Thanks!

Can you contribute fixes to `bundler/lib/bundler/vendor/thor/` files to the [upstream repository](https://github.com/rails/thor)?

Also, how did you find all these? We used to have a spell checker in CI, but it never found any issues and I removed it.

jdufresne:
> Can you contribute fixes to bundler/lib/bundler/vendor/thor/ files to the [upstream repository](https://github.com/rails/thor)?

Yup! I'll work on moving those upstream.

I have removed them from the PR here so this only includes non-vendored files.

> Also, how did you find all these? We used to have a spell checker in CI, but it never found any issues and I removed it.

I used [codespell](https://github.com/codespell-project/codespell). I believe Rails also uses this tool.

jdufresne:
Thor PR opened at:
https://github.com/rails/thor/pull/806

simi:
Should we onboard this tool as well?

https://github.com/rails/rails/blob/c49b8270e18375d64a045da2662f4d31e23726cf/.github/workflows/lint.yml#L25-L42

deivid-rodriguez:
Definitely!

For what it's worth, we did have [spell checking in CI](https://github.com/rubygems/rubygems/pull/4202) for about a year, but [I removed it](https://github.com/rubygems/rubygems/commit/f1bd461ac7d77901099c0a3b7e2113e82a90d362) because the one time the CI check showed as failed, it was because the action failed to install, not because of detecting any typos.

This one seems very easy to setup: https://github.com/codespell-project/actions-codespell. We could try if it detects a fair amount of the typos being fixed here.

jdufresne:
I went ahead and added a GitHub actions for codespell. This should catch future misspellings.

.codespellignore
```diff
@@ -0,0 +1,14 @@
+crate
+falsy
+filetest
+fo
+hax
+msdos
```
simi, line 6: :older_adult: :)

.github/workflows/spell.yml
```diff
@@ -0,0 +1,19 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  spell:
+    name: Check spelling
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
simi, line 18: Can you lock both actions to given SHA with tag comment like it is done at

https://github.com/rubygems/rubygems/blob/75ec11612de0818c1efb83db017484c65b2c8ea4/.github/workflows/daily-bundler.yml#L23

:question:

jdufresne: Sure thing. Done in the latest push.

deivid-rodriguez (approved):
Thanks so much, this is great!

PR #6230, opened 2023-01-03
Add Ruby 3.2 to CI

deivid-rodriguez:

Ruby 3.2 has been released for a few weeks now. We should test against it.

Add Ruby 3.2 to our CI workflows.


simi:
:thinking: seems `setup-ruby` action needs to be updated

PR #6238, opened 2023-01-05
Upstream RubyGems changes from TruffleRuby

eregon:
So these changes will be in RubyGems master too and there is fewer changes to RubyGems in TruffleRuby.


eregon:
https://github.com/rubygems/rubygems/actions/runs/3847681232/jobs/6554455173
```
/home/runner/work/rubygems/rubygems/lib/rubygems/platform.rb:48:in `const_missing': uninitialized constant Gem::Platform::REUSE_AS_BINARY_ON_TRUFFLERUBY (NameError)
	from /home/runner/work/rubygems/rubygems/lib/rubygems/platform.rb:48:in `match_gem?'
	from /home/runner/work/rubygems/rubygems/lib/rubygems/platform.rb:40:in `match_spec?'
	from /home/runner/work/rubygems/rubygems/lib/rubygems/specification.rb:857:in `block in stubs_for_pattern'
	from <internal:core> core/array.rb:1164:in `select'
	from <internal:core> core/array.rb:1164:in `select!'
	from /home/runner/work/rubygems/rubygems/lib/rubygems/specification.rb:857:in `stubs_for_pattern'
	from /home/runner/work/rubygems/rubygems/lib/rubygems/specification.rb:845:in `stubs_for'
	from /home/runner/work/rubygems/rubygems/lib/rubygems/dependency.rb:276:in `matching_specs'
	from /home/runner/work/rubygems/rubygems/lib/rubygems.rb:257:in `find_spec_for_exe'
	from /home/runner/work/rubygems/rubygems/lib/rubygems.rb:284:in `activate_bin_path'
	from /home/runner/.rubies/truffleruby-22.3.0/bin/rake:44:in `<main>'
```
which makes sense.

I see two ways to fix this:
* Use `if RUBY_ENGINE == "truffleruby" && defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)` in platform.rb
* Do the same as https://github.com/rubygems/rubygems/blob/6ba790cfd434692412fcd79859202a2245eba7d2/bundler/lib/bundler/rubygems_ext.rb#L286-L288 but in rubygems.rb just after `require "rubygems/defaults/#{RUBY_ENGINE}"`

@deivid-rodriguez What do you think?

deivid-rodriguez:
I'd rather not upstream anything for now until that change is released.

eregon:
I didn't think of that.
But it seems problematic if we just wait and change nothing in the PR because someone using e.g. truffleruby 22.3 and doing `gem update --system --force-update` will actually break RubyGems in an unrecoverable way.
Granted, it's not frequent to update RubyGems on TruffleRuby but we want to leave the possibility and it can be useful in some cases.
And a big part of this upstreaming is to make updating RubyGems on TruffleRuby not undo the changes in truffleruby's rubygems' copy.

Long story short, I think we need one of these two alternatives otherwise updating RubyGems in TruffleRuby would break older releases.
The first way wouldn't install `libv8 libv8-node sorbet-static` after updating RubyGems on truffleruby 22.3, but that's a small issue compared to RubyGems failing to load entirely.

hsbt (approved):

eregon:
I would like to revisit this.
I rebased the PR and used the second way.
The first way would not work because platform.rb is loaded before defaults/truffleruby.rb.
cc @deivid-rodriguez 

I will merge https://github.com/oracle/truffleruby/pull/2819 after this is merged.

eregon:
The only alternative I see is we would automatically "patch" (i.e. extra logic when requiring) `rubygems/platform.rb` in TruffleRuby. But patches are quite messy/surprising, so this PR seems a much better and clearer solution.

lib/rubygems.rb
```diff
@@ -1336,6 +1336,17 @@ def default_gem_load_paths
...
 end
 
+# TruffleRuby >= 24 defines REUSE_AS_BINARY_ON_TRUFFLERUBY in defaults/truffleruby.
+# However, TruffleRuby < 24 defines REUSE_AS_BINARY_ON_TRUFFLERUBY directly in its copy
+# of lib/rubygems/platform.rb, so it is not defined if RubyGems is updated (gem update --system).
+# Instead, we define it here in that case, similar to bundler/lib/bundler/rubygems_ext.rb.
+# We must define it here and not in platform.rb because platform.rb is loaded before defaults/truffleruby.
+class Gem::Platform
+  if RUBY_ENGINE == "truffleruby" && !defined?(REUSE_AS_BINARY_ON_TRUFFLERUBY)
+    REUSE_AS_BINARY_ON_TRUFFLERUBY = %w[libv8 libv8-node sorbet-static].freeze
```
hsbt, line 1346: Should we move these list to TruffleRuby side? It's hard to maintain this list by RubyGems/Bundler team.

martinemde: If I understand correctly, the list of gems is in truffleruby, this is only a fallback if rubygems gets updated.

martinemde:
@eregon I'm trying to understand this a bit more. Are these special gems compiled specifically for truffleruby? Are they different than normal ruby gems? What is their platform?

Would it make sense to add a unique platform somehow, like "truffleruby-" and then allow rubygems to resolve these like normal gems? Why is it different than java platform?

lib/rubygems.rb
eregon: The list is already on TruffleRuby's side.
But currently it's in the vendored copy of lib/mri/rubygems/platform.rb, which means in RubyGems' CI or when installing a new RubyGems it's gone.
https://github.com/oracle/truffleruby/pull/2819/files is fixing that by moving the definition to lib/truffle/rubygems/defaults/truffleruby.rb.
So this code is there because we need it for TruffleRuby < 24, as the comment says.

It's also the same code as in bundler/lib/bundler/rubygems_ext.rb, which is already in master.

eregon:
@martinemde These are precompiled gems which do not rely on the Ruby ABI, so they can be reused on TruffleRuby safely. And in fact reusing them precompiled is the only good option, hence why this is done.
For instance `sorbet-static` only exists as a precompiled/binary gem. There is no ruby-gem-platform variant of it: https://rubygems.org/gems/sorbet-static/versions
It's a just a binary, independent of the Ruby ABI.
So the only way it can work on TruffleRuby is to reuse those precompiled gems for sorbet-static.

For libv8 libv8-node there is a ruby-gem-platform variant of it, but it's very difficult and slow to build so basically it's almost as if it didn't exist. And for mini_racer and execjs, the main users of libv8* gems, we actually use Graal.js so we don't actually need those dependencies, but there is no way to specify "dependency for CRuby, skip it on TruffleRuby", at least currently.

In general TruffleRuby declares only the ruby gem platform, because it would be a bug to reuse gems precompiled against the CRuby ABI (https://github.com/rubygems/rubygems/issues/2945), since TruffleRuby has a different ABI (e.g. modified headers, different macros, different struct fields, etc).

> Would it make sense to add a unique platform somehow, like "truffleruby-" and then allow rubygems to resolve these like normal gems?

Yes, this mechanism could be useful in general, for example to declare different dependencies for `mini_racer` (i.e. only depend on libv8-node for the ruby-gem-platform, and not for the truffleruby-gem-platform).
There is some discussion about that in https://github.com/rubygems/rubygems/discussions/5988
As you can see it's still very much undecided and needs more discussion and significant changes.

> Why is it different than java platform?

The java platform should really be named the jruby platform, it means such gems contain `.jar` files which depend on JRuby APIs.
We have no such thing in TruffleRuby currently, the only APIs of TruffleRuby are all Ruby APIs, so the ruby-gem-platform + a check for RUBY_ENGINE when requiring the gem is enough for that.

But it would be useful to have something like truffleruby+NATIVE_PLATFORM gem platforms. The current NATIVE_PLATFORM gem platforms should actually be cruby+NATIVE_PLATFORM since they are cruby-specific (for most precompiled gems). That means we could ship precompiled C extensions for truffleruby too.

As said above and in https://github.com/rubygems/rubygems/discussions/5988, it can also be useful to have a truffleruby platform without tying to a native platform. Basically all combinations are useful.

lib/rubygems.rb
hsbt: Is there any reason why not backport https://github.com/oracle/truffleruby/pull/2819 to TruffleRuby 23?

eregon: It's too late to backport such changes to 23.1.2, and there won't be more 23.x releases after 23.1.2.

Also because this fallback code is already at https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/rubygems_ext.rb#L297-L299 I don't think it's meaningful to make things more complicated & brittle and avoid it here.

segiddins (approved):

martinemde (approved):

eregon:
Thank you for the approvals, could a maintainer merge this PR? :)

PR #6240, opened 2023-01-05
Fix version ranges incorrectly handling platforms

### LINKED ISSUE #6237: bundler adds additional ruby vesrion requirement < 3.2.dev on gem.

Bundler adds ruby version requirements (`< 3.2.dev`) to gems which prevents installation of some gems.

I use the new ruby 3.2.0

```shell
$ ruby --version
ruby 3.2.0 (2022-12-25 revision a528908271) [x86_64-linux]
$ gem --version
3.4.1
$ bundler --version
Bundler version 2.4.2
```

```shell
echo "3.2.0" > .ruby-version
echo -e "source 'https://rubygems.org'\ngem 'nokogiri', '1.13.10'" > Gemfile
bundle install
```
This result in 
```shell
Resolving dependencies...

Because nokogiri >= 1.13.10 (x86_64-linux), < 1.14.0.rc1 depends on Ruby >= 2.6, < 3.2.dev
  and Gemfile depends on nokogiri = 1.13.10,
  Ruby >= 2.6, < 3.2.dev is required.
So, because current Ruby version is = 3.2.0,
  version solving has failed.
```

```
bundle install
```

I expected nokogiri to be installed

### What actually happened?

It did not install and claimed that nokogiri required ruby version < 3.2.dev

````shell
$ bundle env
## Environment

```
Bundler       2.4.2
Ruby          3.2.0p0 (2022-12-25 revision a528908271c678360d2d8ca232c178e7cdd340b4) [x86_64-linux]
  Full Path   /home/jarl/.rbenv/versions/3.2.0/bin/ruby
  Config Dir  /home/jarl/.rbenv/versions/3.2.0/etc
RubyGems      3.4.1
  Gem Home    /home/jarl/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0
  Gem Path    /home/jarl/.gem/ruby/3.2.0:/home/jarl/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0
  User Home   /home/jarl
  User Path   /home/jarl/.gem/ruby/3.2.0
  Bin Dir     /home/jarl/.rbenv/versions/3.2.0/bin
Tools         
  Git         2.34.1
  RVM         not installed
  rbenv       rbenv 1.2.0-52-g61747c0
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-01-01
Git SHA           2cd3ed45bf
Released Version  true
```

## Gemfile

### Gemfile

```ruby
source 'https://rubygems.org'
gem 'nokogiri', '1.13.10'
```

### Gemfile.lock

```
<No /home/jarl/tmp/bundler-demo/Gemfile.lock found>
```

````

### PR DESCRIPTION
**Pull request by deivid-rodriguez**

In certain cases, for example, when exactly pinning `nokogiri` in a `Gemfile` to `1.13.10`, on Ruby 3.2, Bundler would not be able to properly fallback to the generic Ruby version when necessary (like here).

We did have some specs checking this, but not this particular situation. For example, in a similar situation, but using an open requirement instead of an exact requirement, Bundler does properly fallback, and we do have tests for that.

For this kind of platform fallback, we use the trick of considering `1.13.10` and `1.13.10-x86_64-linux` as separate versions, and the latter being "higher" than the former, in the sense that it's considered for resolution first even if they both have the same version.

However, we were considering these two separate resolution candidates when retrieving all possible versions for a gem, but not when creating version ranges for dependencies.

This PR ammends version range creation for dependencies, so that the ranges created properly include both generic variants and platform specific variants.

Fixes #6237.

### DIFF
--- a/bundler/spec/bundler/resolver/candidate_spec.rb
+++ b/bundler/spec/bundler/resolver/candidate_spec.rb
@@ -3,8 +3,19 @@
 RSpec.describe Bundler::Resolver::Candidate do
   it "compares fine" do
     version1 = described_class.new("1.12.5", :specs => [Gem::Specification.new("foo", "1.12.5") {|s| s.platform = Gem::Platform::RUBY }])
-    version2 = described_class.new("1.12.5")
+    version2 = described_class.new("1.12.5") # passing no specs creates a platform specific candidate, so sorts higher
 
-    expect(version1 >= version2).to be true
+    expect(version2 >= version1).to be true
+
+    expect(version1.generic! == version2.generic!).to be true
+    expect(version1.platform_specific! == version2.platform_specific!).to be true
+
+    expect(version1.platform_specific! >= version2.generic!).to be true
+    expect(version2.platform_specific! >= version1.generic!).to be true
+
+    version1 = described_class.new("1.12.5", :specs => [Gem::Specification.new("foo", "1.12.5") {|s| s.platform = Gem::Platform::RUBY }])
+    version2 = described_class.new("1.12.5", :specs => [Gem::Specification.new("foo", "1.12.5") {|s| s.platform = Gem::Platform::X64_LINUX }])
+
+    expect(version2 >= version1).to be true
   end
 end
--- a/bundler/lib/bundler/resolver.rb
+++ b/bundler/lib/bundler/resolver.rb
@@ -337,7 +337,8 @@ module Bundler
 
     def requirement_to_range(requirement)
       ranges = requirement.requirements.map do |(op, version)|
-        ver = Resolver::Candidate.new(version)
+        ver = Resolver::Candidate.new(version).generic!
+        platform_ver = Resolver::Candidate.new(version).platform_specific!
 
         case op
         when "~>"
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: bundler/spec/bundler/resolver/candidate_spec.rb [hunk 1]
```diff
@@ -3,8 +3,8 @@
 RSpec.describe Bundler::Resolver::Candidate do
   it "compares fine" do
     version1 = described_class.new("1.12.5", :specs => [Gem::Specification.new("foo", "1.12.5") {|s| s.platform = Gem::Platform::RUBY }])
-    version2 = described_class.new("1.12.5")
+    version2 = described_class.new("1.12.5", :specs => [Gem::Specification.new("foo", "1.12.5") {|s| s.platform = Gem::Platform::X64_LINUX }])
```
simi, line 6: Is there any reason to replace this spec instead of keeping this one and adding new one 🤔 ?

deivid-rodriguez: The spec as written previously no longer works, the result is reversed.

The explanation for that is that previously comparison would take into account whether associated specs had been given originally when instantiating the candidate or not. I realized that this did not make sense and was incorrect. Other than the version, the comparison should only take into account whether the candidate is platform specific or not (platform specific candidates should sort higher than generic candidates).

Because of changing the above logic, the result is reversed because passing `:specs => []` now essentially means creating a platform specific candidate, because the platforms of the passed specs (empty) are not all `Gem::Platform::RUBY`. And platform specific candidates sort higher.

But the above felt just an implementation detail, the resolver in practice either instantiates candidates by passing associated `:specs` to `initialize` or by explicitly making them "generic" or "platform specific" using `generic!` or `platform_specific!` after instantiation.

That's why I replaced the spec instead of adding a new one. But I restored the original spec now (changing the expectation) and added a few more.

simi: Thanks for detailed explanation.

simi (approved):

bundler/spec/bundler/resolver/candidate_spec.rb
deivid-rodriguez: Great, I'm now preparing an upstream patch for pub_grub.

deivid-rodriguez:
Upstream PR created: https://github.com/jhawthorn/pub_grub/pull/22.

flavorjones:
Thank you, @deivid-rodriguez !

PR #6241, opened 2023-01-05
Fix pointing GitHub sources to PRs

### LINKED ISSUE #6239: Can't fetch GitHub gem with pull request URL starting with 2.4.0

Starting with 2.4.0, gems that use the provided `:github` source with a pull request URL (#5126) are not able to be fetched anymore.

Yes, I ran  `gem update --system` and `bundle update --bundler` and that failed on my repo, then I bisected by running `bundle _2.3.24_ update --bundler` (the version used in that repo) and `bundle _2.4.0_ update --bundler` started failing.

You can use this repro script:

```ruby
# bug.rb
gem "bundler", ARGV.first != "bug" ? "2.3.26" : "2.4.0"
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  gem "reline", github: "https://github.com/ruby/reline/pull/488"
end
```

Running `gem install bundler:2.3.26 bundler:2.4.0` then `ruby bug.rb` works (using 2.3.26), but not `ruby bug.rb bug` (using 2.4.0).

Originally it was a simple `bundle install`, running in a clean environment where the repo wasn't already fetched. If the Git repo is already fetched and cached, installing with 2.4.0 works fine.

Even if the PR is closed, we should be able to fetch its ref.

### What actually happened?

The remote ref couldn't be fetched.

deivid-rodriguez:

Recent refactorings in git sources to use shallow clones when possible broke pointing to GitHub PRs.

Git sources are now more strict and require that the parameter the `:branch` to be an actual branch, and the parameter to `:tag` to be an actual branch. For arbitrary refs, which is what pointing to GitHub pull requests use under the hood, `ref` should be used.

This PR changes Bundler to use `ref` internally for handling PR URLs given as the value for the `:github` parameter.

Fixes #6239.


etiennebarrie (commented):
It works for me. Thanks for the quick fix!

Maybe there's some kind of integration test missing showing that `gem "foo", github: "https://github.com/foo/foo/pull/123"` works in its entirety, from the DSL to the fetching and installing.

deivid-rodriguez:
Definitely, otherwise we would've caught this. It did not seem super easy though, so I decided to pass on adding a spec here.

deivid-rodriguez:
I added a realworld spec in the end, just to get the feature covered.

PR #6252, opened 2023-01-07
Skip setting `BUNDLER_SETUP` on Ruby 2.6

### LINKED ISSUE #6251: uninitialized constant DidYouMean::VERSION for Ruby 2.6

The long story is we are trying to fix Rails 6.1 CI on Ruby 2.6: https://buildkite.com/rails/rails/builds/92250#01858a31-3eac-44d2-b718-d2d1e3cea717

The issue is that `DidYouMean` somehow ends up as an empty module. It gets required by the prelude (I added a `puts` in `did_you_mean.rb` to confirm that) but then none of its methods or constants are defined by the time any user code runs.

I've confirmed the problem exists on latest Bundler (2.4.3)/Rubygems, and that there is no issue on older Bundler (2.3.17 works fine)

```Dockerfile
FROM ruby:2.6.10
RUN gem update --system && gem install bundler -v 2.3.17
WORKDIR /app
RUN touch Gemfile
RUN bundle install
# build fails if you remove below line
RUN sed -i s/2.4.3/2.3.17/g Gemfile.lock
RUN echo 'puts DidYouMean::VERSION' > test.rb
RUN bundle exec ruby test.rb
```

`docker build -t bundler-issue .` with above Dockerfile

The build should succeed, and 1.3.0 should be printed for the last RUN

### What actually happened?

```
docker build -t bundler-issue .
Sending build context to Docker daemon  570.3MB
Step 1/7 : FROM ruby:2.6.10
 ---> b71cca836fff
Step 2/7 : RUN gem update --system && gem install bundler -v 2.3.17
 ---> Using cache
 ---> 6d5426d58f40
Step 3/7 : WORKDIR /app
 ---> Using cache
 ---> 45eb305e4501
Step 4/7 : RUN touch Gemfile
 ---> Using cache
 ---> 104144143c22
Step 5/7 : RUN bundle install
 ---> Using cache
 ---> 081c22fb05d0
Step 6/7 : RUN echo 'puts DidYouMean::VERSION' > test.rb
 ---> Running in a57b74cb6ea0
Removing intermediate container a57b74cb6ea0
 ---> 4aa2719dd185
Step 7/7 : RUN bundle exec ruby test.rb
 ---> Running in 04220de79a69
test.rb:1:in `<main>': uninitialized constant DidYouMean::VERSION (NameError)
The command '/bin/sh -c bundle exec ruby test.rb' returned a non-zero code: 1
```

## Environment

```
Bundler       2.4.3
Ruby          2.6.10p210 (2022-04-12 revision 67958) [x86_64-linux]
  Full Path   /usr/local/bin/ruby
  Config Dir  /usr/local/etc
RubyGems      3.4.3
  Gem Home    /usr/local/bundle
  Gem Path    /root/.gem/ruby/2.6.0:/usr/local/lib/ruby/gems/2.6.0:/usr/local/bundle
  User Home   /root
  User Path   /root/.gem/ruby/2.6.0
  Bin Dir     /usr/local/bundle/bin
Tools
  Git         2.30.2
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-01-06
Git SHA           ed1f27f75c
Released Version  true
```

## Bundler settings

```
app_config
  Set via BUNDLE_APP_CONFIG: "/usr/local/bundle"
silence_root_warning
  Set via BUNDLE_SILENCE_ROOT_WARNING: true
```

### Gemfile

```ruby

```

### Gemfile.lock

```
GEM
  specs:

PLATFORMS
  x86_64-linux

DEPENDENCIES

BUNDLED WITH
   2.4.3
```

deivid-rodriguez:

Setting `BUNDLER_SETUP` on `bundle exec` so that RubyGems can setup the bundle early in the subprocess breaks loading DidYouMean on Ruby 2.6, since it was a bundled (not default) gem there, and thus gets "erased" by `bundler/setup`.
 
Skip this feature on Ruby 2.6.

Closes #6251.

### DIFF
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -284,7 +284,7 @@ module Bundler
       Bundler::SharedHelpers.set_env "BUNDLE_BIN_PATH", exe_file
       Bundler::SharedHelpers.set_env "BUNDLE_GEMFILE", find_gemfile.to_s
       Bundler::SharedHelpers.set_env "BUNDLER_VERSION", Bundler::VERSION
-      Bundler::SharedHelpers.set_env "BUNDLER_SETUP", File.expand_path("setup", __dir__)
+      Bundler::SharedHelpers.set_env "BUNDLER_SETUP", File.expand_path("setup", __dir__) unless RUBY_VERSION < "2.7"
     end
 
     def set_path
--- a/bundler/spec/bundler/shared_helpers_spec.rb
+++ b/bundler/spec/bundler/shared_helpers_spec.rb
@@ -248,6 +248,8 @@ RSpec.describe Bundler::SharedHelpers do
 
     shared_examples_for "ENV['BUNDLER_SETUP'] gets set correctly" do
       it "ensures bundler/setup is set in ENV['BUNDLER_SETUP']" do
+        skip "Does not play well with DidYouMean being a bundled gem instead of a default gem in Ruby 2.6" if RUBY_VERSION < "2.7"
+
         subject.set_bundle_environment
         expect(ENV["BUNDLER_SETUP"]).to eq("#{source_lib_dir}/bundler/setup")
       end


simi:
:thinking: Was https://github.com/rubygems/rubygems/pull/6025 just forgotten? I think we all agreed that was mistake to merge.

deivid-rodriguez:
@mame and @hsbt were fine with merging and shipping it. Nobody answered my request for a revert, and I couldn't find a strong reason to block it unilaterally other than "it feels wrong", so I did not proceed with the revert myself. So, to answer your question, no, I personally did not forget.

At this point I'm not sure we should revert the feature. People may be already using it, since we have released it. That's why I went with reverting it for Ruby 2.6 to fix this particular issue as the least controversial thing to do.

deivid-rodriguez:
I will merge this, since whatever we discuss around this issue should not block fixing #6251.

hsbt (approved):

hsbt:
Thanks. I try to replace it with other approach.

Issue #6268, opened 2023-01-12
`--ext=rust` unexpected build and install behavior


steakknife:
Related but different from #6204 which also appears to be a problem.

### Describe the problem as clearly as you can

`bundle gem example --ext=rust` doesn't mirror build and install behavior of native C extensions.

### Did you try upgrading rubygems & bundler?

- rubygems 3.4.3
- bundler 2.4.3

### Post steps to reproduce the problem

<details><summary>MCE Dockerfile</summary>

```
FROM quay.io/centos/centos:stream9

RUN dnf install -y ruby-devel rubygem-irb '@Development Tools' clang
RUN echo 'gem: --no-document' >~/.gemrc
RUN gem update --system
RUN curl https://sh.rustup.rs -sSf | sh -s -- --profile complete -y --no-modify-path
ENV PATH "/root/.cargo/bin:/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
WORKDIR /root
RUN bundle gem example --ext=rust --test=rspec --linter=rubocop
WORKDIR /root/example
RUN cargo generate-lockfile
RUN git add -A
RUN sed -i 's/TODO[^"]*//;/_uri/d' example.gemspec
RUN bundle
# Note: cargo unexpectedly builds a native lib here when a regular ruby gem should be built
RUN rake build
# Note: cargo doesn't run here
RUN rake install
WORKDIR /tmp
CMD ["irb -rexample"]
```
</details>

### Which command did you run?

`docker build .`

### What were you expecting to happen?

1. Rust native ext compiles at user-level `gem install` time.

2. If a user doesn't have rust, it's the same as not having a C toolchain. Maintainers can push precompiled `--platform` flavors of a release to rubygems, or not. Don't try to do parts of both, and especially not for a `--platform=ruby` gem.

### What actually happened?

```
/usr/share/gems/gems/irb-1.3.5/lib/irb/init.rb:376: warning: LoadError: cannot load such file -- /usr/share/gems/gems/example-0.1.0/lib/example/example
```

1. Compiled rust native ext at maintainer `gem build` time.
2. Installed itself into the *original source directory* without asking.
3. Looked for native ext in the wrong directory.

~~### Notes~~

~~1. `Rakefile` includes problematic ` ext.lib_dir = "lib/example"`. Deployed native exts live in `ext/`.~~

[...]

Issue #6204, opened 2022-12-26
`bundle gem --ext=rust` generates gem that fails to load when installed


ankane:
### Describe the problem as clearly as you can

The gem created by `bundle gem --ext=rust` fails to load when installed.

### Did you try upgrading rubygems & bundler?

Yes.

### Post steps to reproduce the problem

With Ruby 3.2:

```sh
bundle gem hello --ext=rust
cd hello
# manually remove TODOs in gemspec
bundle exec rake compile
git add .
git commit -m "First commit"
bundle exec rake build
gem install pkg/hello-0.1.0.gem
ruby -e 'require "hello"'
```

### Which command did you run?

See above.

### What were you expecting to happen?

Gem to load.

### What actually happened?

Gem fails to load, since it tries to load `lib/hello/hello.bundle` (on Mac), but the Cargo builder creates `lib/hello.bundle`.

```text
/Users/user/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/hello-0.1.0/lib/hello.rb:4:in `require_relative': cannot load such file -- /Users/user/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/hello-0.1.0/lib/hello/hello (LoadError)
    from /Users/user/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/hello-0.1.0/lib/hello.rb:4:in `<top (required)>'
    from <internal:/Users/user/.rbenv/versions/3.2.0/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:160:in `require'
    from <internal:/Users/user/.rbenv/versions/3.2.0/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:160:in `rescue in require'
    from <internal:/Users/user/.rbenv/versions/3.2.0/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:149:in `require'
    from -e:1:in `<main>'
<internal:/Users/user/.rbenv/versions/3.2.0/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- hello (LoadError)
    from <internal:/Users/user/.rbenv/versions/3.2.0/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
    from -e:1:in `<main>'
```

## Environment

```
Bundler       2.4.1
[...]

PR #6269, opened 2023-01-12
Fix `--ext=c` generating buggy code

steakknife:
CentOS, Fedora, RHEL and so forth break without this.

Also, `Rakefile` doesn't need those options.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/ext/newgem/newgem.c.tt
+++ b/bundler/lib/bundler/templates/newgem/ext/newgem/newgem.c.tt
@@ -2,8 +2,20 @@
 
 VALUE rb_m<%= config[:constant_array].join %>;
 
+VALUE rb_<%= config[:constant_array].join %>_hello(VALUE self, VALUE str);
+
 void
 Init_<%= config[:underscored_name] %>(void)
 {
   rb_m<%= config[:constant_array].join %> = rb_define_module(<%= config[:constant_name].inspect %>);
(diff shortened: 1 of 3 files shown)


steakknife:
Where the hack is the squash rebase button in Microsoft Github?

hsbt:
https://github.com/puma/puma/issues/2371 seems not related this. Can you explain it?

steakknife:
@hsbt I misread puma/puma#2381 about offtopic puma/puma#2371. It's puma/puma#2381.

steakknife:
This also fixes #6268 and #6204

steakknife:
I fixed both `bundle gem --ext=c foo` and `bundle gem --ext=rust bar` on several machines (CentOS 9 Stream x86_64, Ubuntu 22 LTS x86_64, and macOS 13 arm) I ran test benches against.

It's not ideal and probably not doing everything right, but it's closer to tried-and-true based on what already exists. I'm sure I'll see it's 100% wrong after I sleep.

deivid-rodriguez:
This should've been fixed by #6298, so good to close, right?

deivid-rodriguez:
(Thanks for your work regardless @steakknife)

steakknife:
@deivid-rodriguez No worries. I'll give it a shot to validate.

steakknife:
It's still broken and it pollutes `lib/` with a binary shared object.

PR #6284, opened 2023-01-18
Update bundle-exec man page to not use deprecated `Bundler.with_clean_env`

technicalpickles:
`with_clean_env` has been deprecated in `with_unbundled_env`. It already generates a deprecation warning when it uses, but this man page was still referring to it.

If a user follows the instructions from `bundle exec --help` and uses `Bundler.with_clean_env`, they will get a deprecated message to use `Bundler.with_unbundled_env`

Update the .ronn and .man pages


deivid-rodriguez (approved):
We should probably document `Bundler.with_original_env`, and the differences between the two (in my experience `Bundler.with_original_env` is _normally_ what users want).

But this first pass replacing the deprecated helper with the non-deprecated equivalent already improves things, so good to go as is!

PR #6285, opened 2023-01-18
Fix packages for external platforms being introduced in lockfile when Bundler retries resolution

deivid-rodriguez:

After we started ignoring prereleases and only considering them if we can't resolve without them at #6246, we got reports that Bundler was sometimes introducing gems for external platforms in the lockfile. For example, when using `gem "tzinfo-data", platform: :windows` on non Windows platforms.

The problem was that while on initial resolution we were properly ignoring these dependencies, they were not being properly ignored when Bundler had to resort to re-resolving including prereleases (and that's the case of Rails main branch).

The fix is to properly remove the dependency so that it's not considered on retries either.

Another improvement we can do here, but I'll leave that for another PR, is to not exclude prereleases from gems configured to use a gemspec/path source. But this bug fix is independent from that enhancement.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -487,7 +487,7 @@ module Bundler
     end
 
     def expanded_dependencies
-      @expanded_dependencies ||= dependencies + metadata_dependencies
+      dependencies + metadata_dependencies
     end
 
     def resolution_packages
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -551,7 +551,7 @@ module Bundler
     end
 
     def start_resolution
-      result = resolver.start(expanded_dependencies)
+      result = resolver.start
 
       SpecSet.new(SpecSet.new(result).for(dependencies, false, @platforms))
     end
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I improved the initial approach. Instead of removing the requirement from the resolver when it's for a different platform, when avoid passing it to the resolver at all in the first place.

zzak:
@deivid-rodriguez Out of curiosity, Is there a way we can bump rubygems in the Rails CI to verify this fixes our builds?

indirect:
@zzak add a step at the beginning that runs `gem install --system`? or check the docs for the ruby-setup GitHub action to see if you can request a specific rubygems version.

yahonda:
FYI Rails CI executes `gem update --system`.

https://github.com/rails/buildkite-config/blob/cb5f3064e5e28c931782a2782fb170e0844d6916/Dockerfile#L7

https://buildkite.com/rails/rails/builds/92828#0185ccd1-a6fe-4c92-b3c8-805510d0edfe/152-168

deivid-rodriguez:
I see that you already pinned Bundler & RubyGems 👍.

Trying this PR is not as easy since the PR is not released nor even merged, but trying it should be a matter of cloning this repo, checking out this branch, and running `cd bundler; bin/rake install`. Or without installing anything, from the rails repo root, run `~/path/to/cloned/rubygems/bundler/exe/bundle`.

PR #6291, opened 2023-01-20
Relax slow perf spec

deivid-rodriguez:

Sometimes this spec fails randomly on Ruby 3.2.

I guess it's harder on Ruby 3.2. Relax it a bit to avoid flakes.


indirect:
ooh you have reminded me about my dream of one day having an actual per-commit performance benchmark for Bundler. I should get that onto our list!

PR #6292, opened 2023-01-21
Explicitly mark `eval_gemfile` as undocumented

sambostock:
> **Warning**
> **This PR and rubygems/bundler-site#1081 are mutually exclusive. At most one of them should be merged.**
> I personally think documenting it is the better approach, but I have opened this as the alternative.

It is unclear if `eval_gemfile` should be used or not. On the one hand, it is public, and shows up in the generated API documentation. On the other hand, it is not explicitly called out in the guides.

Despite this, many projects use it, as noted in rubygems/bundler-site#1081.

While I _personally_ think it should be part of the public API (as it is already in use by many consumers) and would rather see rubygems/bundler-site#1081 merged, and this PR closed, if it is indeed considered private, it should be explicitly marked as undocumented, so this is clear to consumers.

- ~Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes~

### DIFF
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -40,7 +40,7 @@ module Bundler
       add_git_sources
     end
 
-    def eval_gemfile(gemfile, contents = nil)
+    def eval_gemfile(gemfile, contents = nil) # :nodoc:
       expanded_gemfile_path = Pathname.new(gemfile).expand_path(@gemfile&.parent)
       original_gemfile = @gemfile
       @gemfile = expanded_gemfile_path


simi:
Hello again! Thanks for opening this. I think both PRs actually could be closed, but we can provide more info to documentation to explain the current state and also do some global documentation changes to start properly documenting Bundler code. Let me explain.

`Bundler` itself doesn't utilize `nodoc` and it doesn't respect public/private scopes for client usage. Introducing this one (first) `nodoc` in here (thanks to this) doesn't make any sense IMHO. There are clearly some public APIs like `Bundler.require` or `Bundler.setup`.  To fix this problem it would be great to start with `nodoc` of all and specifically select supported calls to be exposed in `public` scope in documentation. Currently everything is public and documented as you can see at https://www.rubydoc.info/gems/bundler/2.4.4/Bundler, which is definitely wrong.

On the other side, API exposed in `Gemfile` is not related to actual "code scoping". There is DSL documented at https://bundler.io/v2.4/man/gemfile.5.html and you should use only methods specified in there. It is not related to `Bundler` code and it doesn't inherit any public/private scoping.

Does that make sense? @deivid-rodriguez any ideas?

sambostock:
Thank you for the detailed response, @simi!

> Bundler itself doesn't utilize `nodoc` and it doesn't respect public/private scopes for client usage.

My mistake. I had run `git grep :nodoc:` which turned up a bunch of results under the `bundler` directory, but I failed to notice they're all under `bundler/lib/bundler/vendor`.

Given the stance that `eval_gemfile` should not be used, are there recommended alternatives for the usages I listed in https://github.com/rubygems/bundler-site/pull/1081#issue-1551660825?

> - Loading the main gemfile from alternate gemfiles, used to test against specific gem versions
>   - https://github.com/Shopify/maintenance_tasks/blob/6a95544479979ea38151b72b9d2cb99e8e381c67/gemfiles/rails_6_0.gemfile#L5
>   - https://github.com/socketry/falcon/blob/a0e70287ba5168545cab00b980c01a1a6fa0f5ac/gems/rack-v1.rb#L3
> - Loading a per-dev gemfile that isn't tracked in version control
>   - https://github.com/rubocop/rubocop/blob/3a542a96fb59c254d68348c54e04af8fdcab7b4e/Gemfile#L30-L31
>   - https://github.com/dry-rb/dry-validation/blob/ce18224a1c05a8fe78c1a729b804a12acdebbbb5/Gemfile#L5
>   - https://github.com/rom-rb/rom/blob/7fb82cf7ffa86805d9c5499a4ecc64d5d3c20f14/Gemfile#L7
>   - The debated usage in Rails, mentioned above (see rails/rails#47033 & rails/rails#47098)
> - Other usage (didn't investigate)
>   - https://github.com/rspec/rspec-rails/blob/40261bb72875c00a6e4a0ca2ac697b660d4e8d9c/example_app_generator/generate_app.rb#L47-L48
>   - https://github.com/oracle/truffleruby/blob/dd8609b306c1558e6071d4e76424da82e1481325/lib/mri/bundler/injector.rb#L35

zzak:
It seems like there is a need for this kind of use-case, so in my own humble opinion, would like to see something like this added to bundler. Although I realize this is not the place to make feature requests, just sayin' :)

PR #6298, opened 2023-01-23
Include directory in CargoBuilder install path

matsadler:

This is an attempt to address issue #6204 "`bundle gem --ext=rust` generates gem that fails to load when installed".

The `CargoBuilder` is installing Rust extensions in `lib/gemname.dlext`, but the Rust extension template is set up expecting the extension to be installed in `lib/gemname/gemname.dlext`.

I don't think a concrete decision was made in that ticket, but my reading was that `lib/gemname/gemname.dlext` *is* the desired destination, and the bug is in `CargoBuilder`, not the Rust extension template.

Furthermore, it seems the consensus was that while this would be a breaking change to `CargoBuilder`, it has not ever been announced as stable, so it's fine to make the change.

Currently `CargoBuilder` installs extensions to `lib/gemname.dlext`, these changes update that to `lib/dirname/cratename.dlext`.

`dirname` is sourced from the extension path, dropping the first and last parts, for example "ext/foo/bar/Cargo.toml" becoming "foo/bar".

`cratename` is sourced by calling out to the `cargo metadata` command.

Using `dirname` as the first part should better match the directory structure and is analogous to how C extensions are built. Users have control over this as they can nest directories however they want.

Using `cratename` instead of `gemname` makes `CargoBuilder` more reliable, as it better detects the location of the extension once cargo has compiled it to move to the final install location. This also provides a benefit to users as they have more control over the crate name, and can choose to give it a different name to the gem if they wish. And it allows for multiple Rust extensions in a single gem, which might not be a common use case, but it matches what can be done with C extensions.
There is also benefit to users of [Magnus](http://github.com/matsadler/magnus) (as used by the Rust extension template) as the autogenerated `Init` function generated by Magnus will match the crate name, not the gem name, potentially saving confusing debugging.

As I was making these changes I wanted to be able to only call `#cargo_crate_name` once, to avoid multiple calls out to `cargo`. To do this I refactored `#build`, inlining a number of helper methods only used a single time. Hopefully this makes the flow of `#build` easier to understand.

I also updated the gem spec validation to allow `Cargo.lock` to be anywhere in the gem, as requiring it to be in the root seemed to be a frustrating limitation.

I updated the `CargoBuilder` tests, and reused the `custom_name` test extending it to cover the new behaviour. As I was updating the tests I found two files in the fixtures that seemed to be unused, so removed them.


simi:
Thanks for the detailed info. If I understand it well, `lib/gemname/gemname.dlext` is the same path as for C extensions. I think that's the way to go for now.

simi:
Windows CO seems unhappy. Any ideas how to fix that? Anything I can help with (I have access to Windows machine if needed)?

matsadler:
So part of this change is calling out to `cargo metadata` to get the crate name. It's possible there's more than one crate, so that returns the data like:
``` json
{
  "packages": [
    {"name": "foo", "manifest_path": "path/to/foo/Cargo.toml", ...},
    {"name": "bar", "manifest_path": "path/to/bar/Cargo.toml", ...},
  ]
}
```
We know the path to the crate we're trying to build, as it's passed as the first argument to `build`, so my code matches them up and gets the name that way.

The error on Windows is happening right around this bit of code, and my guess is that Ruby and cargo aren't agreeing on the exact format of the paths on Windows, so they aren't matching.

I've added a better error message in this case, which should hopefully confirm that's the issue. If it is it's probably just a case of figuring out how to normalise the paths.

doaamegehahed:
Yes, lost my password

dddd

في الجمعة، ٢٧ يناير ٢٠٢٣, ٧:٥١ ص Mat Sadler ***@***.***> كتب:

simi:
@matsadler feel free to check CI now

simi:
TL;DR

```
  looking for: D:/a/rubygems/rubygems/tmp/test_rubygems_20230127-4224-1i30256/ext/Cargo.toml
...
  found:
  rust_ruby_example at D:\a\rubygems\rubygems\tmp\test_rubygems_20230127-4224-1i30256\ext\Cargo.toml

```

full error

```
Error: test_build_cdylib(TestGemExtCargoBuilder):
  Gem::InstallError: failed to determine cargo package name
  
  looking for: D:/a/rubygems/rubygems/tmp/test_rubygems_20230127-4224-1i30256/ext/Cargo.toml
  
  found:
  rust_ruby_example at D:\a\rubygems\rubygems\tmp\test_rubygems_[202](https://github.com/rubygems/rubygems/actions/runs/4021593599/jobs/6918549856#step:8:203)30127-4224-1i30256\ext\Cargo.toml
D:/a/rubygems/rubygems/lib/rubygems/ext/cargo_builder.rb:229:in `cargo_crate_name'
D:/a/rubygems/rubygems/lib/rubygems/ext/cargo_builder.rb:24:in `build'
D:/a/rubygems/rubygems/test/rubygems/test_gem_ext_cargo_builder.rb:35:in `block in test_build_cdylib'
     32:     Dir.chdir @ext do
     33:       ENV.update(@rust_envs)
     34:       builder = Gem::Ext::CargoBuilder.new
  => 35:       builder.build "Cargo.toml", @dest_path, output
     36:     end
     37: 
     38:     output = output.join "\n"
D:/a/rubygems/rubygems/test/rubygems/test_gem_ext_cargo_builder.rb:32:in `chdir'
D:/a/rubygems/rubygems/test/rubygems/test_gem_ext_cargo_builder.rb:32:in `test_build_cdylib'
```

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -134,42 +184,63 @@ def ruby_static?
...
+    end
 
-  def cargo_dylib_path(dest_path)
-    prefix = so_ext == "dll" ? "" : "lib"
-    path_parts = [dest_path]
-    path_parts << ENV["CARGO_BUILD_TARGET"] if ENV["CARGO_BUILD_TARGET"]
-    path_parts += ["release", "#{prefix}#{cargo_crate_name}.#{so_ext}"]
-    File.join(*path_parts)
-  end
+    # cargo metadata output is json, but yaml is a superset of json, so we can
```
simi, line 223: :thinking: If I understand it well, YAML is actually not JSON superset, even it is super close. Maybe it would be better to avoid this comment, but make it clear YAML is used to parse JSON since it is compatible in this case.

lib/rubygems/ext/cargo_builder.rb
```diff
@@ -6,30 +6,60 @@
...
-    require "shellwords"
 
-    build_crate(dest_path, results, args, cargo_dir)
-    validate_cargo_build!(dest_path)
-    rename_cdylib_for_ruby_compatibility(dest_path)
-    finalize_directory(dest_path, lib_dir, cargo_dir)
-    results
-  end
+    # Where's the Cargo.toml of the crate we're building
+    cargo_toml = File.join(cargo_dir, "Cargo.toml")
```
simi, line 22: So it seems (thanks to failing specs message now) this produces wrong path on Windows.

```
D:/a/rubygems/rubygems/tmp/test_rubygems_20230127-4224-1i30256/ext/Cargo.toml
```

Since actually proper one is present int `cargo metadata` output.

```
D:\a\rubygems\rubygems\tmp\test_rubygems_20230127-4224-1i30256\ext\Cargo.toml
```

This seems to be known Ruby limitation. I'll check on Windows machine if there's any way to produce proper Windows path in here. By looking at the [other code](https://github.com/rubygems/rubygems/blob/be1cc4ae73e5836ed620a02284560040936e6ceb/lib/rubygems/installer.rb#L690), this could work (not tested).

```ruby
cargo_toml = cargo_toml.tr(File::ALT_SEPARATOR, File::SEPARATOR) if File::ALT_SEPARATOR
```

lib/rubygems/ext/cargo_builder.rb
matsadler: I've updated the wording in this comment to avoid saying YAML is a superset of JSON, and to point out we're working with a known format that we fix with the `--format-version 1` option to `cargo`.

The format is specified here if you're interested: https://doc.rust-lang.org/cargo/commands/cargo-metadata.html#output-format

lib/rubygems/ext/cargo_builder.rb
matsadler: I think that `tr(ALT_SEPARATOR ...` trick should work. I've pushed a changed using that approach.

simi:
@matsadler seems ok. Is this ready to merge for now?

matsadler:
Yep, if there’s no more comments then I’m happy for this to be merged.

PR #5844, opened 2022-08-17
Deprecate `install` parameter to `bundler/inline`

tonyaraujop:
## What was the end-user or developer problem that led to this PR?

Deprecate the install parameter of Bundler's inline option mentioned in #5364 

## What is your fix for the problem, implemented in this PR?

It was added `rubygems_deprecate(install)` method to deprecate `install` argument.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6304, opened 2023-01-24
Inline gemfile: change `install` to keyword argument

tbates-redarc:

PR #5844 removes the `install` positional argument, since it didn't really do anything as the gems would be installed silently anyway. It would be nice to be able to control whether or not the gems are installed automatically.

- This PR builds on the previous one to add a keyword `install` argument.
- It also converts the other keyword options to first-class keyword arguments - though I'm not sure if this creates problems with older versions of Ruby.
- Finally, it makes the default `ui.level` to be `"confirm"` which causes messages to be printed to the console only if packages are installed, but not otherwise.

### DIFF
--- a/bundler/lib/bundler/inline.rb
+++ b/bundler/lib/bundler/inline.rb
@@ -42,6 +42,9 @@ def gemfile(install = false, options = {}, &gemfile)
     Bundler.instance_variable_set(:@bundle_path, Pathname.new(Gem.dir))
     old_gemfile = ENV["BUNDLE_GEMFILE"]
     Bundler::SharedHelpers.set_env "BUNDLE_GEMFILE", "Gemfile"
+    Bundler::SharedHelpers.major_deprecation 2, "The optional install parameter to the `gemfile(install = false, &block)` helper is getting"\
+    " removed because regardless of what you pass in there, it still installs missing gems."\
+    " Remove the explicit `install` parameter to get rid of this message.", :print_caller_location => true if install
 
     Bundler::Plugin.gemfile_install(&gemfile) if Bundler.feature_flag.plugins?
     builder = Bundler::Dsl.new
--- a/bundler/lib/bundler/inline.rb
+++ b/bundler/lib/bundler/inline.rb
@@ -10,6 +10,13 @@
 #
 # @param install [Boolean] whether gems that aren't already installed on the
 #                          user's system should be installed.
+#                          Defaults to `true`.
+#
+# @param ui [Bundler::UI::*] `Bundler.ui` component to use.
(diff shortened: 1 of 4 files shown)


deivid-rodriguez:
So, your proposal is an alternative to #5844 where instead of deprecating anything, we fix the problem and make installation "on the fly" really optional, correct?

Not having to go through a deprecation sounds appealing, but I wonder if we really need this, since this has never actually worked as expected and nobody complained for years 🤔

tbates-redarc:
We still need to deprecate the positional `install` argument, but instead of removing it completely I am proposing changing it to a keyword argument (that actually works).

You're right that anyone that was trying to use it was probably mislead as to whether it was doing what was intended; in my testing I found that if `install` was false it would reinstall cached gems but not download new ones, or not refresh the source, or something like that.

My concern is that people might have thought it was doing (some of) what they expected. For example, see [this comment](https://github.com/rubygems/rubygems/pull/5529/files#r871825388), or [this StackOverflow post](https://stackoverflow.com/a/44596129) where the author's comment says:
```ruby
# passing true here does the install; in real scripts we
# pass a boolean based on a --install flag.
gemfile(true) do
  # ...
end
```

jeromedalbert:
> My concern is that people might have thought it was doing (some of) what they expected. For example, see [this comment](https://github.com/rubygems/rubygems/pull/5529/files#r871825388), or [this StackOverflow post](https://stackoverflow.com/a/44596129) where the author's comment says:

In both links people are using `gemfile(true)` because they want installation to happen (even though in practice, as we know, it always installs).

So this leads me to think think that people only need bundler inline with install true. To confirm that assumption, let's expand the sample size by looking at occurences of true vs false in the wild:

- `gemfile do`: [4.9k files](https://github.com/search?q=%22gemfile+do%22&type=code) on GitHub repos.
- `gemfile(true)`: [1.1k files](https://github.com/search?q=%22gemfile%28true%29%22&type=code) on GitHub repos.
- `gemfile(false)`: a measly [14 files](https://github.com/search?q=%22gemfile%28false%29%22&type=code) on GitHub repos. Of those lone 14 use cases, nobody seemed to have complained. 😅

So I tend to agree with David. Since virtually noone uses `gemfile(false)`, and noone complained for years, and `gemfile(false)` has never actually worked as expected, and people overwhelmingly use bundler inline for installing, my vote goes for deprecating `install` and eventually removing it. I believe https://github.com/rubygems/rubygems/pull/5844 should be favored instead.

deivid-rodriguez:
Cool @jeromedalbert, let's do that, then. Thanks for the patch and discussion @tbates-redarc, and apologies for dropping the ball for so long!

PR #6305, opened 2023-01-24
Fix `bundler/inline` not skipping installation when gems already there, when used more than once

tbates-redarc:

Using `bundler/inline` multiple times causes gems mentioned in the second and subsequent uses to be reinstalled on every invocation, even if they were already installed.

- Added tests to show the problem.
- Added a call to `Bundler.reset!` to `gemfile` method which fixes the problem, but I am not sure if this is the correct approach.
- ~~I had another problem that I can't work out how to write a test for, where extensions were not being compiled correctly when installed in this manner. Wrapping the body of the `gemfile` method in `Bundler.with_unbundled_env` fixed it.~~ (Moved to #6306)

### DIFF
--- a/bundler/lib/bundler/inline.rb
+++ b/bundler/lib/bundler/inline.rb
@@ -31,6 +31,7 @@
 #
 def gemfile(install = false, options = {}, &gemfile)
   require_relative "../bundler"
+  Bundler.reset!
 
   opts = options.dup
   ui = opts.delete(:ui) { Bundler::UI::Shell.new }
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Thanks for fixing these things! Can you move the fix related to extensions to a separate PR? That way we can ship the first fix which seems ready to go with a test, and then focus on reproducing/fixing the other issue.

tbates-redarc:
Moved second commit to #6306 and fixed unit tests.

deivid-rodriguez (approved):
Looks good to me!

PR #6308, opened 2023-01-25
Don't undo require decorations made by other gems

### LINKED ISSUE #5263: Bundler removes require decorations made by other gems
I am not sure how to address this in a generic way. Let me open this issue at least to start a conversation about it.

Bundler [hard-resets `Kernel#require`](https://github.com/rubygems/rubygems/blob/a9f9341e08478426ca3aa7829b5be4534cee06c1/bundler/lib/bundler/rubygems_integration.rb#L251-L259) this way:

```ruby
def reverse_rubygems_kernel_mixin
  # Disable rubygems' gem activation system
  kernel = (class << ::Kernel; self; end)
  [kernel, ::Kernel].each do |k|
    if k.private_method_defined?(:gem_original_require)
      redefine_method(k, :require, k.instance_method(:gem_original_require))
    end
  end
end
```

Therefore, if you have other decorations below RubyGems, they are just erased. Who decorates `Kernel#require`? Zeitwerk and Bootsnap do, for example ([here](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/kernel.rb), and [here](https://github.com/Shopify/bootsnap/blob/master/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb)).

For instance, that means that if a Ruby program loads Zeitwerk first, and then Bundler, Zeitwerk doesn't work. This is a synthetic, minimal reproduction of the issue:

```ruby
require 'tmpdir'

require 'zeitwerk'

require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'zeitwerk'
end

Dir.mktmpdir do |dir|
  Dir.chdir(dir)

  Dir.mkdir('lib')
  File.write('lib/foo.rb', 'Foo = 1')

  loader = Zeitwerk::Loader.new
  loader.push_dir('lib')
  loader.on_load("Foo") { p :CALLED }
  loader.setup

  Foo
end
```

If you run that as is, the `on_load` callback is not called. If you comment out line 3, it is called (as expected).

The problem is not only the callback, the purpose of that script is to show you something concrete that we can run, Zeitwerk is broken in more fundamental ways without its `Kernel#require` decoration in place.

And in turn, this is not only about Zeitwerk, but about how Bundler could be better coordinated with other gems decorating `Kernel#require` in general. If RubyGems is so special that you need to hijack it, that may be a very special case. The libraries even live in the same repository, they can be as coupled as they need to.

But what about other libraries?

(PS: Fortunately anything run via `bundle exec` is not subject to this because Bundler is loaded first.)

deivid-rodriguez:

If a gem decorates require, depending on the order things are loaded, Bundler may end up not only removing `RubyGems` require decorations, but also those extra decorations.

My fix is, instead of overwriting `require` with the original version, let RubyGems require be aware of Bundler having taken over, and delegate to standard require in that case.

Fixes #5263.

### DIFF
--- a/bundler/lib/bundler/installer/standalone.rb
+++ b/bundler/lib/bundler/installer/standalone.rb
@@ -84,13 +84,17 @@ module Bundler
 
     def reverse_rubygems_kernel_mixin
       <<~END
-      kernel = (class << ::Kernel; self; end)
-      [kernel, ::Kernel].each do |k|
-        if k.private_method_defined?(:gem_original_require)
-          private_require = k.private_method_defined?(:require)
-          k.send(:remove_method, :require)
-          k.send(:define_method, :require, k.instance_method(:gem_original_require))
-          k.send(:private, :require) if private_require
+      if Gem.respond_to?(:discover_gems_on_require=)
+        Gem.discover_gems_on_require = false
+      else
+        kernel = (class << ::Kernel; self; end)
+        [kernel, ::Kernel].each do |k|
+          if k.private_method_defined?(:gem_original_require)
+            private_require = k.private_method_defined?(:require)
+            k.send(:remove_method, :require)
+            k.send(:define_method, :require, k.instance_method(:gem_original_require))
+            k.send(:private, :require) if private_require
+          end
         end
       end
       END
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -227,10 +227,14 @@ module Bundler
 
     def reverse_rubygems_kernel_mixin
       # Disable rubygems' gem activation system
-      kernel = (class << ::Kernel; self; end)
-      [kernel, ::Kernel].each do |k|
-        if k.private_method_defined?(:gem_original_require)
-          redefine_method(k, :require, k.instance_method(:gem_original_require))
+      if Gem.respond_to?(:discover_gems_on_require=)
+        Gem.discover_gems_on_require = false
+      else
+        kernel = (class << ::Kernel; self; end)
+        [kernel, ::Kernel].each do |k|
+          if k.private_method_defined?(:gem_original_require)
+            redefine_method(k, :require, k.instance_method(:gem_original_require))
+          end
         end
       end
     end
--- a/bundler/spec/runtime/setup_spec.rb
+++ b/bundler/spec/runtime/setup_spec.rb
@@ -1519,4 +1519,29 @@ end
       expect(err).to be_empty
     end
   end
+
+  it "does not undo the Kernel.require decorations", :rubygems => ">= 3.4.6" do
+    install_gemfile "source \"#{file_uri_for(gem_repo1)}\""
+    script = bundled_app("bin/script")
+    create_file(script, <<~RUBY)
+      module Kernel
+        module_function
+
+        alias_method :require_before_extra_monkeypatches, :require
+
+        def require(path)
+          puts "requiring \#{path} used the monkeypatch"
+
+          require_before_extra_monkeypatches(path)
+        end
+      end
+
+      require "bundler/setup"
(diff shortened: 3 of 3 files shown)


simi:
The change itself looks ok to me. If I understand it well, instead of defining custom `require` back and forth, it is currently no-op in RubyGems if Bundler is loaded thanks to the new configuration being set super early. Btw. what about [standalone](https://github.com/rubygems/rubygems/blob/39448b198896a019c2ba375545aad943213efe2c/bundler/lib/bundler/installer/standalone.rb#L85-L97)? Wouldn't it benefit from this as well?

But I do struggle to follow the testing changes. Part of the `test/rubygems/test_gem.rb` was moved into `test/rubygems/bundler_test_gem.rb` to not be part of `rake test` anymore and `t.libs << "bundler/lib"` was removed as well. But at the same time, this `bundler_test_gem.rb` file is run in `rake test:isolated` (explicitly) and there is no problem to keep `bundler/lib` in lib path for all isolated test file runs aka `sh Gem.ruby, "-Ilib:test:bundler/lib", file`.

I did quick testing locally, and tests are passing even when file is renamed back to be under `test_` prefix, but there are some warnings (caused probably by double bundler require).

@deivid-rodriguez would you mind to explain a little why this change is needed? Is `rake test:isolated` the only way for now to run all RubyGems tests? Would it make sense to mention that in [CONTRIBUTING.md](https://github.com/rubygems/rubygems/blob/37274f0fe7449d5ff1fdcd8c8769d95ad91e8f00/CONTRIBUTING.md#running-tests)?

deivid-rodriguez:
Good point about standalone, I'll update to include it!

The change is needed to play nicer with gems that decorate `require`, you can see the full story in the issue this is closing, and the attached spec as an example.

I should've commented on test changes before setting this as ready. The problem is that RubyGems test suite, as it works now, loads `bundler`. This is problematic since Bundler monkeypatches RubyGems, so we end up testing Bundler, not RubyGems. See for example, https://github.com/rubygems/rubygems/pull/5956/commits/d20b4d1950610dd1bd7a953db8c54f400a6be8c1, and that PR in general where I tried to improve the situation.

In this case, it's biting us again because during RubyGems test execution, Bundler is loaded and reverts the require monkeypatch and then depending on order, the tests checking the "automatic activation of gems on require" start failing since the behavior is no longer there. You can see the failures without test changes here: https://github.com/rubygems/rubygems/actions/runs/4007644021/jobs/6880765362.

The solution is exactly what you observed: ensure that our main suite does not load Bundler at all by moving the specs that load Bundler to a separate file that does not get run by default. We still get the functionality that loads Bundler tested through `rake test:isolated` where the side effects of it does not cause any problems because they run... in isolation. I think this is a good change regardless of this patch, and I can propose it as a separate PR if necessary.

I'm not sure why this didn't bite us with the previous approach, but I think this kind of isolation is helpful regardless so I'd rather not look deeper into that.

I do not think it's worth mentioning this in CONTRIBUTING because these few tests exercise a very particular functionality (auto require of `bundler/setup` when `RUBYGEMS_GEMDEPS` env variable is set) that hardly ever changes, and doesn't usually break when changing RubyGems. But I can mention it if you feel strongly. We also have many CI checks not run by default with `rake test` that could be mentioned but in general I feel it would be distracting and not helpful for most cases.

simi (approved):

deivid-rodriguez:
I added the change also to standalone! I believe this is ready, but I believe this is technically a new feature in RubyGems and a bug fix in Bundler that leverages the new feature. So let me split the PR so that each one get their proper entry in the respective changelogs.

simi (approved):
Feel free to merge once ready.

deivid-rodriguez:
I extracted #6319 and will leave this one for the fix in Bundler.

PR #6309, opened 2023-01-26
Add gem exec command

segiddins:

Lack of an easy way to run executables from gems that may or may not be installed.

See https://github.com/rubygems/rfcs/pull/45

### DIFF
--- a/lib/rubygems/commands/exec_command.rb
+++ b/lib/rubygems/commands/exec_command.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 require_relative "../command"
 require_relative "../dependency_installer"
+require_relative "../gem_runner"
 require_relative "../package"
 require_relative "../version_option"
 
@@ -60,7 +61,10 @@ to the same gem path as user-installed gems.
     check_executable
 
     print_command
-    if options[:conservative]
+    if options[:gem_name] == "gem" && options[:executable] == "gem"
+      Gem::GemRunner.new.run options[:args]
+      return
+    elsif options[:conservative]
       install_if_needed
     else
       install
@@ -135,15 +139,19 @@ to the same gem path as user-installed gems.
     activate!
   end
 
-  def install
-    gem_name = options[:gem_name]
-    gem_version = options[:version]
-
+  def set_gem_exec_install_paths
     home = File.join(Gem.dir, "gem_exec")
 
     ENV["GEM_PATH"] = ([home] + Gem.path).join(Gem.path_separator)
     ENV["GEM_HOME"] = home
     Gem.clear_paths
+  end
+
+  def install
+    set_gem_exec_install_paths
+
+    gem_name = options[:gem_name]
+    gem_version = options[:version]
 
     install_options = options.merge(
       minimal_deps: false,
--- a/lib/rubygems/commands/exec_command.rb
+++ b/lib/rubygems/commands/exec_command.rb
@@ -1,0 +1,186 @@
+# frozen_string_literal: true
+require_relative "../command"
+require_relative "../package"
+require_relative "../version_option"
+
+class Gem::Commands::ExecCommand < Gem::Command
+  include Gem::VersionOption
+
+  def initialize
+    super "exec", "Run a command from a gem", {
+      version: Gem::Requirement.default,
+    }
+
+    add_platform_option
+    add_version_option
+    add_prerelease_option "to be installed"
+
+    add_option "-g", "--gem GEM", "run the executable from the given gem" do |value, options|
+      options[:gem_name] = value
+    end
+
+    add_option(:"Install/Update", "--conservative",
+      "Prefer the most recent installed version, ",
+      "rather than the latest version overall") do |value, options|
+      options[:conservative] = true
+    end
+  end
+
+  def arguments # :nodoc:
+    "COMMAND  the executable command to run"
+  end
+
+  def defaults_str # :nodoc:
+    "--version '#{Gem::Requirement.default}'"
+  end
+
+  def description # :nodoc:
+    <<-EOF
+    EOF
+  end
+
+  def usage # :nodoc:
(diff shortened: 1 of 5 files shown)


lib/rubygems/commands/exec_command.rb
```diff
@@ -0,0 +1,204 @@
...
+
+  def arguments # :nodoc:
+    "COMMAND  the executable command to run"
+  end
+
+  def defaults_str # :nodoc:
+    "--version '#{Gem::Requirement.default}'"
+  end
+
+  # TODO(segiddins): add a proper description!
```
segiddins, line 37: - [ ] TODO here to add a description for the command

indirect (approved):
This is so much faster than I expected. Nice work.

One open question: when you run `gem exec rails` or `gem exec rspec`, the actual command is coming from a dependency of the gem with the name (railties and rspec-core, respectively). When I tested, it seemed like `gem exec rspec` did not fetch the newest version of rspec-core, and I had to run `gem exec -g rspec-core rspec` to get the newest version installed and then run. Should we expect the latest version to be installed? I think maybe we should?

segiddins:
@indirect try again ;)

indirect:
It works! Nice job. :)

While testing it out, I noticed this:

```
❯ gem uninstall rspec-core
Gem 'rspec-core' is not installed
❯ gem exec -g rspec-core rspec
No examples found.
❯ gem exec -g rspec-core --conservative rspec
No examples found.
```

1. So with exec gems installing to a different location that isn't managed by `gem install` and `gem uninstall`... how does the end user handle issues installing gems? 

While testing out issues installing gems, I ran into this:

```
❯ gem exec -g tailwindcss-rails tailwindcss
[success output]

❯ gem exec -g tailwindcss-rails --platform x86_64-darwin tailwindcss
[building native extensions blah blah blah]
/Users/andre/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/dependency.rb:314:in `to_specs': Could not find 'mini_portile2' (~> 2.8.0) among 252 total gem(s) (Gem::MissingSpecError)
Checked in 'GEM_PATH=/Users/andre/.gem/ruby/3.2.0:/Users/andre/.rubies/ruby-3.2.0/lib/ruby/gems/3.2.0' , execute `gem env` for more information
        from /Users/andre/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/dependency.rb:326:in `to_spec'
        from /Users/andre/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_gem.rb:57:in `gem'
        from extconf.rb:430:in `process_recipe'
        from extconf.rb:838:in `<main>'
```

2. I think that's a bug in the `--platform` switch? Maybe? Bundler does install mini_portile2 as a dependency of nokogiri on x86_64-darwin.

segiddins:
Should be fixed now! (and the `--platform` option removed, because it breaks caching and honestly is a bad experience)

indirect:
Cool, thank you!

indirect:
Oh right, looking back over my previous post, I still have a question: with exec gems installing to a different location that isn't managed by gem install and gem uninstall... 

1. how does the user uninstall a gem installed with exec? is that possible? 
2. how does the end user handle gem installations that fail? do they just get the same error messages as for gem install?

indirect:
and I guess also
3. is there any html or other written documentation that needs to be updated for this change?

segiddins:
1. how does the user uninstall a gem installed with exec? is that possible?
  it would be possible by setting GEM_HOME to the `gem_exec` gem home and then running `gem uninstall`
2. yeah, should be the same error messages since we're using the same installation code as `gem install` is
3. I don't think so? the command has a `--help` output

indirect:
I didn’t think of this until just now, but does `gem exec gem uninstall foo` work to remove a gem after installing it via `gem exec foo`? if so I think that’s a good solution to any possible questions raised by the alternate install directory.

segiddins:
It does not work, since `gem` isn't provided by any gems

```
❯ ruby -Ilib -S gem exec --verbose gem
running gem exec with:
           version: >= 0
        executable: gem
          gem_name: gem

HEAD https://index.rubygems.org/
200 OK
GET https://index.rubygems.org/info/gem
200 OK
ERROR:  Could not find a valid gem 'gem' (>= 0) in any repository
GET https://rubygems.org/latest_specs.4.8.gz
304 Not Modified
ERROR:  Possible alternatives: zxygemm, 42_gem, AABeginnerTestGem, AccountGem, Alonso_gem, AndyFirstGem, Backup_GEM_2, BeverageMaker, BidManagerLoggerGem, BoardGameGem
```

indirect:
Ok, I think I’m going to request to special case `gem exec gem` (sorry), because then no one has to manually set GEM_HOME to work with the exec install dir using the commands they already know.

segiddins:
OK, `gem exec gem` now works!

indirect:
Okay, this looks good to me! Going to restart the jobs that aren't passing.

simi:
:thinking: Sadly there's gem called gem already pushed to rubygems.org (https://rubygems.org/gems/gem). That's going to conflict with `gem exec gem` if I understand it well.

segiddins:
@indirect `--pre` already was in there & supported! But I added some tests to cover it.

segiddins:
All builds are finally green!!

indirect:
🎉

johnnyshields:
Should yank/delete the gem called "gem". That seems like a (minor) security risk to me.

indirect:
The gem called `gem` has been handed over to the RubyGems team. If we were to yank it, the name would immediately open up for anyone else to take over, which would create a completely new security risk that does not exist today. We're not going to yank it.

martinemde:
I submitted rubygems/rubygems.org#3569 to address this concern.

Mark24Code:
very helpful! 🎉

PR #6310, opened 2023-01-26
Update the list of SPDX licence identifiers

segiddins:
Update the list of SPDX identifiers by running `ruby util/generate_spdx_license_list.rb`, also updating that script so it generates rubocop-approved code.


hsbt (approved):
Nice catch 👍

simi:
:+1: Would be any automation in here welcomed? We can start with something simple like monthly reminder to regenerate list on Slack in maintainers channel.

segiddins:
We could probably have an action re-run the script and open a PR if there's any diff?

deivid-rodriguez:
It would make sense to me to run something like that on a periodic (monthly?) basis, yeah!

PR #6312, opened 2023-01-27
Don't warn on `bundle binstubs --standalone --all`

composerinteralia:

Prior to this commit `bundle binstubs --standalone --all` would output a warning about not being able to generate a standalone binstub for bundler.

This warning predates the `--all` option, and I don't think it makes sense in this context. The warning makes good sense when explicitly trying to generate a bundler standalone binstub with `bundle binstubs bundler --standalone`, since that command won't do what the user might have expected. But `--all` is not specifically asking for bundler, and having it report each time that the bundler binstubs could not be generated does not seem particularly helpful. The only way to make that warning go away would be to stop using `--standalone --all`.

This commit skips the warning when running with the `--all` option.

### DIFF
--- a/bundler/lib/bundler/cli/binstubs.rb
+++ b/bundler/lib/bundler/cli/binstubs.rb
@@ -40,7 +40,11 @@ module Bundler
         end
 
         if options[:standalone]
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Makes total sense to me!

PR #6317, opened 2023-01-30
Add test for old lockfile with new ruby version

flavorjones:
## Summary

We want to be able to `bundle install` on `ruby-head` starting from a `Gemfile.lock` that was generated on the latest stable release. This currently works, but has broken a few times over the past few years. This PR adds a test to prevent regressions to this workflow:

> Given an existing application using native gems (e.g., nokogiri)
> And a lockfile generated with a stable ruby version
> When we run `bundle install` with ruby-head to test against that ruby version
> Then bundler should fall back to the generic ruby platform gem

I couldn't find existing test coverage of this exact scenario. I see a few tests that are _close_, but use `bundle update` or `bundle lock` as the action, not `bundle install`, so I wanted to add a more explicit test. 

Would the Bundler maintainers be willing to formally support this use case?

Note that this test has been passing since 45931ac9 / 4ff54191 (bundler 2.4.3).

(This PR is co-authored by @byroot.)

For a couple years now, Shopify has been continuously testing our applications against ruby-head. This brings a lot of value to us as it allows to deal with breaking changes in advance, but more importantly we believe it brings a lot of value to the community. This practice allows us to report bugs quickly after they are introduced, which stabilizes the final release; it also gives us the opportunity to fix compatibility issues upstream in common gems way before the yearly Ruby release in December.

To achieve those goals, we need to test with the same gem versions, so regenerating the `Gemfile.lock` from scratch or running `bundle update` aren't quite the right solution.

Over the last few years our testing pipelines have been broken a few times in surprising ways because of improvements and changes to how Bundler handle platforms. We'd like to add a test for this workflow to ensure that it continues working, if the Bundler team is open to supporting it.

Most recently, this workflow was broken in Bundler 2.4.1 (possibly earlier? I didn't bisect), we saw this error message after a bundler install on ruby-head:

```
Because xxxx >= 0.1.23 (arm64-darwin, x86_64-darwin, x86_64-linux) depends
on Ruby >= 3.1, < 3.3.dev
  and Gemfile depends on xxxx ~> 0.1.23,
  Ruby >= 3.1, < 3.3.dev is required.
So, because current Ruby version is = 3.3.0.dev,
```

This was quickly fixed in Bundler 2.4.3 (thank you for that!), but the test added doesn't _explicitly_ cover this use case (unless I'm misunderstanding).

- [ ] ~Write code to solve the problem~ _Already passes!_

### DIFF
--- a/bundler/spec/install/gemfile/specific_platform_spec.rb
+++ b/bundler/spec/install/gemfile/specific_platform_spec.rb
@@ -821,6 +821,64 @@ RSpec.describe "bundle install with specific platforms" do
     bundle :install
   end
 
+  it "automatically fixes the lockfile if the specific platform is locked and we move to a newer ruby version for which a native package is not available" do
+    #
+    # Given an existing application using native gems (e.g., nokogiri)
+    # And a lockfile generated with a stable ruby version
+    # When want test the application against ruby-head and `bundle install`
+    # Then bundler should fall back to the generic ruby platform gem
+    #
+    simulate_platform "x86_64-linux" do
+      build_repo4 do
+        build_gem "nokogiri", "1.14.0"
+        build_gem "nokogiri", "1.14.0" do |s|
+          s.platform = "x86_64-linux"
+          s.required_ruby_version = "< #{Gem.ruby_version}"
+        end
+      end
+
+      gemfile <<~G
+        source "#{file_uri_for(gem_repo4)}"
+
+        gem "nokogiri", "1.14.0"
+      G
+
+      lockfile <<~L
+        GEM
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Absolutely, if this kind of thing broke again, it'd be another regression, so whatever we can do to avoid my overlooks is welcome! 👍

As per the failures, I think you may want to wrap the whole test with a `simulate_platform "x86_64-linux" do ... end` block so that it passes regardless of CI platform.

flavorjones:
> wrap the whole test

Will do!

deivid-rodriguez:
Thanks @flavorjones and @byroot!

Issue #5317, opened 2022-01-28
`bundler/setup.rb` script generates by `bundle install --standalone` uses absolute paths for `:path` gem dependencies.


prakhunov:
### Describe the problem as clearly as you can

We package our Ruby applications in a Docker image that gets deployed but since we use a bunch of private git repos, and due to it being a monorepo we also use gems installed via `:path`.

On bundler 2.2.29 (JRuby 9.3 default) ,and even updating to bundler 2.3.6, the `bundler/setup.rb` script has absolute paths to the gems that are installed via `:path`. For additional info `:git` doesn't have this issue. This of course breaks in the resulting docker image because those paths don't exist.

On bundler 2.1.4 this was working fine. 

### Did you try upgrading rubygems & bundler?
Yes I tried 2.3.6.

### Post steps to reproduce the problem

1. Create a project with a Gemfile similar to the following:

```ruby
  gem "dep_a", :path => "../dep_a" # relative path for dep_a
```

2. Run the following:

```bash
bundle package --all
bundle install --standalone --deployment --local
```

The resulting `vendor/cache/bundler/setup.rb` will use the absolute path to `dep_a` which breaks if this is copied to another system or docker.

### Which command did you run?

`bundle install --standalone`

### What were you expecting to happen?
If I have relative paths in my Gemfile, I expected the paths to look like this (as they did in 2.1.4)

```ruby
$:.unshift "#{path}/../../cache/dep_a
```

instead it looks like:

```
$:.unshift "/home/username/main_project/vendor/cache/dep_a"
```

### What actually happened?

The result in the file looks like:

```
/home/username/project/vendor/cache/dep_a
```

### If not included with the output of your command, run `bundle env` and paste the output below

The only difference is during the `bunde package` phase it says the dependencies are using a relative path,
and during the `bundle install --standalone` phase it uses the absolute path in the output.

deivid-rodriguez:
I actually have a vague recollection that this was done intentionally but I can't recall why now. Could you bisect the specific pull request that caused this? Thanks.

prakhunov:
@deivid-rodriguez 

I have not done a bisect, but I'm pretty sure it is this:

https://github.com/rubygems/rubygems/issues/3505

Good old: https://www.hyrumslaw.com/ at work.

deivid-rodriguez:
[...]

PR #4792, opened 2021-07-29
Make script generated by `bundle install --standalone` resilient to moving the application to a differently nested folder when `path` sources are used

intuxicated:
## What was the end-user or developer problem that led to this PR?

When a `Gemfile` using path sources, and `bundle install --standalone` is run, the generated script won't work if the application is moved to a folder with a different "folder depth".

## What is your fix for the problem, implemented in this PR?

The fix is, if path sources point to an absolute path, add it to `$LOAD_PATH` directly, so that the generated script doesn't depend on how depely nested the application is with respect to the path source.  

Fixes #4758.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6327, opened 2023-01-31
Preserve relative path sources in standalone setup

composerinteralia:

Fixes https://github.com/rubygems/rubygems/issues/5317

When specifying gems with the `:path` option, `bundler --standalone` generates a `bundler/setup.rb` with absolute paths for those gems.

This makes good sense when the `:path` is absolute (the behavior was explicitly added in https://github.com/rubygems/rubygems/pull/4792 to handle absolute paths).

But when the `:path` option is relative, it is not always desirable to have absolute paths in `bundler/setup.rb`. For example, I work on an application that runs `bundler --standalone` in CI to build a deploy artifact, but because the root directory in CI does not match the root in production we are unable to use the generated bundler/setup.rb directly.

As suggested in https://github.com/rubygems/rubygems/issues/5317#issuecomment-1024438825, this commit reverts back to bundler-setup-relative paths in if the `:path` option is relative. It preserves the current behavior of using absolute paths only when the `:path` option is absolute.

### DIFF
--- a/bundler/lib/bundler/installer/standalone.rb
+++ b/bundler/lib/bundler/installer/standalone.rb
@@ -52,7 +52,7 @@ module Bundler
 
     def gem_path(path, spec)
       full_path = Pathname.new(path).absolute? ? path : File.join(spec.full_gem_path, path)
-      if spec.source.instance_of?(Source::Path)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
This is great, thanks so much for tackling this!

PR #6329, opened 2023-02-01
Fix inconsistent behavior of zero byte files in archive

martinemde:

When working on unpacking gems for rubygems/rfcs#44 I found that one of our test `.gem` files has a zero byte file in it. This is perfectly valid, but caused inconsistent behavior when calculating SHA digests or mime types.

The TarReader in the gem can extract the file, but `TarReader::Entry#read` returns `nil` when it should return a blank string.

This required me to make exceptions when the output of read was nil, when the real solution is that `#read` should behave just like any other reference implementation of read.

I made a few new tests that verify the behavior of a 0 byte file in a tar archive. I used `StringIO` as a reference for how `Entry` should behave.

I also locally verified that this is how an open `File` object behaves.

I will submit the following fixes:
- `#read` after already reading to the end of the file should return a blank string.
- `#read` and `#readpartial` should return an empty string if 0 bytes are requested. This was true already, except when the `pos` is at the end of the file. Both `File#readpartial(0)` and `StringIO#readpartial(0)` continue to return `""` even if `readpartial(1)` would EOF.
- `#readpartial` only raises EOF if you attempt to read at least 1 byte past EOF when the `pos` is already at EOF.

I did not fix one inconsistency: `StringIO#readpartial()` (no arguments) gives an argument error. You must supply a length. `TarReader::Entry#readpartial` defaults the maxlen to nil and then calculates the remaining length. I didn't change this behavior since I'm not able to check that every usage of `#readpartial` does include an argument.

The changes I did make are unlikely to break anything since anyone working around our behavior to account for its inconsistency would instead receive a blank string as if the file was empty. Given that you already have the file name, the file size, and any other settings like filemode, it's reasonable to send back a blank string if that is the actual contents of the file.

### DIFF
--- a/lib/rubygems/package/tar_reader/entry.rb
+++ b/lib/rubygems/package/tar_reader/entry.rb
@@ -130,9 +130,10 @@ class Gem::Package::TarReader::Entry
   def read(len = nil)
     check_closed
 
-    return nil if @read >= @header.size
-
     len ||= @header.size - @read
+
+    return nil if len > 0 && @read >= @header.size
+
     max_read = [len, @header.size - @read].min
 
     ret = @io.read max_read
@@ -144,9 +145,10 @@ class Gem::Package::TarReader::Entry
   def readpartial(maxlen = nil, outbuf = "".b)
     check_closed
 
-    raise EOFError if @read >= @header.size
-
     maxlen ||= @header.size - @read
+
+    raise EOFError if maxlen > 0 && @read >= @header.size
+
     max_read = [maxlen, @header.size - @read].min
 
     @io.readpartial(max_read, outbuf)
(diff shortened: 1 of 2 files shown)


indirect (approved):
This looks reasonable to me! I can confirm that `File.read` works the way you describe, returning an empty string for a file with 0 length instead of `nil`.

I'll wait a day or two to see if @deivid-rodriguez, @hsbt, or @simi have any feedback before merging.

deivid-rodriguez (approved):
It also makes sense to me, yeah!

So if I understood correctly before this fix RubyGems was unable to unpack/install gems including an empty file?

martinemde:
When `TarReader::Entry#read` is used to extract the gem, the `nil` goes into `File#write` which accepts `nil`.

Here's a quick test:

```ruby
irb(main):002:0> File.open('emptytest.txt', 'wb') { |f| f.write nil }
=> 0
irb(main):003:0> File.exist?('emptytest.txt')
=> true
irb(main):004:0> File.read('emptytest.txt')
=> ""
```

I think rubygems worked fine and this just fixes anyone reading directly from the TarReader instead of writing to a file.

deivid-rodriguez:
Aaah I see. Yup I did make similar testing, was just curious whether `gem install` actually relied on this or not. Thanks again!

PR #6330, opened 2023-02-01
Better deal with circular dependencies

### LINKED ISSUE #6328: Bundler: Could not find compatible versions for standalone_migrations + rack combo

Gemfile of:

```
# frozen_string_literal: true

source "https://rubygems.org"

gem "rack", "~> 3.0"
gem "standalone_migrations"
```

Running:

```
bundle install
```

Shows:

```
Resolving dependencies...

Because standalone_migrations >= 0 depends on itself
  and Gemfile depends on standalone_migrations >= 0,
  version solving has failed.
```

RubyGems: 3.3.16
Bundler: 2.4.6

As above.

As above.

Installation to find latest versions of strong_migrations + Rails gems that fufil the criteria.

### What actually happened?

```
Resolving dependencies...

Because standalone_migrations >= 0 depends on itself
  and Gemfile depends on standalone_migrations >= 0,
  version solving has failed.
```

## Environment

```
Bundler       2.4.6
Ruby          3.1.2p20 (2022-04-12 revision 4491bb740a9506d76391ac44bb2fe6e483fec952) [arm64-darwin-21]
  Full Path   /Users/ryan.bigg/.asdf/installs/ruby/3.1.2/bin/ruby
  Config Dir  /Users/ryan.bigg/.asdf/installs/ruby/3.1.2/etc
RubyGems      3.3.16
  Gem Home    /Users/ryan.bigg/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0
  Gem Path    /Users/ryan.bigg/.gem/ruby/3.1.0:/Users/ryan.bigg/.asdf/installs/ruby/3.1.2/lib/ruby/gems/3.1.0
  User Home   /Users/ryan.bigg
  User Path   /Users/ryan.bigg/.gem/ruby/3.1.0
  Bin Dir     /Users/ryan.bigg/.asdf/installs/ruby/3.1.2/bin
Tools         
  Git         2.37.1 (Apple Git-137.1)
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-01-31
Git SHA           598c04d54a
Released Version  true
```

## Bundler settings

```
gem.changelog
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
gem.ci
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
gem.coc
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
gem.linter
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
gem.mit
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
gem.test
  Set for the current user (/Users/ryan.bigg/.bundle/config): false
local.fat_zebra_models
  Set for the current user (/Users/ryan.bigg/.bundle/config): "/Users/ryan.bigg/code/fz/fat-zebra-models"
```

### Gemfile

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

gem "rack", "~> 3.0"
gem "standalone_migrations"
```

### Gemfile.lock

```
<No /Users/ryan.bigg/code/playground/bundler-24/Gemfile.lock found>
```

deivid-rodriguez:

When a dependency includes a version with circular dependencies, and the resolver ends up considering it, it gives up with a confusing error.

Ignore these ill-defined versions.

Fixes #6328.

### DIFF
--- a/bundler/lib/bundler/resolver.rb
+++ b/bundler/lib/bundler/resolver.rb
@@ -38,7 +38,7 @@ module Bundler
 
       @all_specs = Hash.new do |specs, name|
         specs[name] = source_for(name).specs.search(name).reject do |s|
-          s.dependencies.any? {|d| d.name == name } # ignore versions that depend on themselves
+          s.dependencies.any? {|d| d.name == name && !d.requirement.satisfied_by?(s.version) } # ignore versions that depend on themselves incorrectly
         end.sort_by {|s| [s.version, s.platform.to_s] }
       end
 
@@ -57,7 +57,7 @@ module Bundler
           { root_version => root_dependencies }
         else
           Hash.new do |versions, version|
-            versions[version] = to_dependency_hash(version.dependencies, @packages)
+            versions[version] = to_dependency_hash(version.dependencies.reject {|d| d.name == package.name }, @packages)
           end
         end
       end
--- a/bundler/lib/bundler/resolver.rb
+++ b/bundler/lib/bundler/resolver.rb
@@ -37,7 +37,9 @@ module Bundler
       root_version = Resolver::Candidate.new(0)
 
       @all_specs = Hash.new do |specs, name|
-        specs[name] = source_for(name).specs.search(name).sort_by {|s| [s.version, s.platform.to_s] }
+        specs[name] = source_for(name).specs.search(name).reject do |s|
+          s.dependencies.any? {|d| d.name == name } # ignore versions that depend on themselves
+        end.sort_by {|s| [s.version, s.platform.to_s] }
       end
 
       @sorted_versions = Hash.new do |candidates, package|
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Spec added!

deivid-rodriguez:
Actually let me try a slightly better approach.

deivid-rodriguez:
Alright so the new approach is:

* If the self-referencing dependency is simply redundant, ignore the dependency but consider the version. This makes the case in #6328 now resolve to the self-referencing version, which seems correct since it's the highest one that meets all requirements.

* If the self-referencing dependency is incompatible with itself, then we completely ignore the version as a candidate for resolution. I've never seen such an ill-defined gem, but I believe it's currently possible to push such gem, so I handled this just in case.

We should probably raise a validation warning on `gem build` for the first case above, and an error for the latter.

deivid-rodriguez:
I also proposed a similar change [upstream](https://github.com/jhawthorn/pub_grub/pull/25).

simi:
:thinking: I can take over to introduce gemspec validations once merged.

simi (approved):

PR #6335, opened 2023-02-03
Warn on self referencing gemspec dependency.

simi:
- it makes no sense and can break resolving

@deivid-rodriguez as promised at https://github.com/rubygems/rubygems/pull/6330#issuecomment-1413891433

It warns only for now. Feel free to ping me if error makes more sense in here, I can quickly update. As usual I would appreciate any suggestion on the warn/error message.

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -173,6 +173,7 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
   end
 
   ##
+  # Checks that the gem does not depend on itself.
   # Checks that dependencies use requirements as we recommend.  Warnings are
   # issued when dependencies are open-ended or overly strict for semantic
   # versioning.
@@ -180,6 +181,10 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
   def validate_dependencies # :nodoc:
     warning_messages = []
     @specification.dependencies.each do |dep|
+      if dep.name == @specification.name # warn on self reference
+        warning_messages << "Self referencing dependency is unnecessary and strongly discouraged."
+      end
+
       prerelease_dep = dep.requirements_list.any? do |req|
         Gem::Requirement.new(req).prerelease?
       end
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -2677,6 +2677,23 @@ duplicate dependency on c (>= 1.2.3, development), (~> 1.2) use:
     end
   end
 
+  def test_validate_self_referencing_dependencies
+    util_setup_validate
+
+    Dir.chdir @tempdir do
+      @a1.add_runtime_dependency @a1.name, "1"
+
+      use_ui @ui do
+        @a1.validate
+      end
+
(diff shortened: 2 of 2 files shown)


lib/rubygems/specification_policy.rb
```diff
@@ -173,13 +173,18 @@ def validate_duplicate_dependencies # :nodoc:
   end
 
   ##
+  # Checks that dependencies are not self referencing itself.
```
deivid-rodriguez, line 176: This sentence feels weird to me, "self" and "itself" seem to add some redundancy. But I don't have a great alternative, maybe

```suggestion
  # Checks that no dependencies reference the gem itself.
```

but doesn't feel great it. A native speaker should word this better.

lib/rubygems/specification_policy.rb
```diff
@@ -173,13 +173,18 @@ def validate_duplicate_dependencies # :nodoc:
...
   # Checks that dependencies use requirements as we recommend.  Warnings are
   # issued when dependencies are open-ended or overly strict for semantic
   # versioning.
 
   def validate_dependencies # :nodoc:
     warning_messages = []
     @specification.dependencies.each do |dep|
+      if dep.name == @specification.name # warn on self reference
+        warning_messages << "Self referencing dependency is highly not recommended and can cause resolving issues."
```
deivid-rodriguez, line 185: I'd go with

```suggestion
        warning_messages << "Self referencing dependency is unnecessary and not recommended."
```

deivid-rodriguez (commented):
Couple of wording suggestions, but I native speaker would do a much better review.

lib/rubygems/specification_policy.rb
jchestershopify: Or even "strongly discouraged" instead of "not recommended".

lib/rubygems/specification_policy.rb
jchestershopify: Is the meaning here that the gem shouldn't refer to itself?

lib/rubygems/specification_policy.rb
simi: > Or even "strongly discouraged" instead of "not recommended".

Yes, like this one https://rubygems.org/gems/standalone_migrations/versions/2.0.4.

lib/rubygems/specification_policy.rb
simi: > Or even "strongly discouraged" instead of "not recommended".

Yes, like this one https://rubygems.org/gems/standalone_migrations/versions/2.0.4.

deivid-rodriguez: I like this

```suggestion
  # Checks that the gem does not depend on itself.
```

indirect: Yep, I like that also. 👍🏻

simi:
Thanks for all suggestions. Comment and message updated.

jchestershopify (commented):
LGTM.

deivid-rodriguez (approved):
Thanks for this @simi!

PR #6339, opened 2023-02-04
Exclude gemspec file itself from gem (take 2)

amatsuda:
This patch follows up f444478eaccf036fdc46407c5b82e0f3ed5a2ad5 by @nobu 

There used to be a commit f444478eaccf036fdc46407c5b82e0f3ed5a2ad5 that intended to exclude the gemspec file from gem package, but the gemspec file is in fact still included in gems that we build today, even though the test included in that commit (that asserts it "does not include the gemspec file in files") is passing.

This was because:
1. The patch was incomplete, and it failed to actually exclude the gemspec file
2. The test code was referencing a wrong gemspec which of course *does not include* the newly generated gemspec file during the test. So the test wasn't really testing anything

This patch fixes both broken test and the newgem template. Now the gemspec file should be properly excluded as intended.


deivid-rodriguez (approved):
Makes total sense, thanks for fixing this @amatsuda!

PR #6344, opened 2023-02-05
Simplify the gem package file filter in the gemspec template

orien:

When using `bundle gem` to create a new project, the generated gemspec uses a complex regular expression in the gem package file filter. This complexity will impede many developers  if they need to make adjustments here.

I propose to replace the regular expression with simple string comparisons. I think this will be much more approachable for many developers.

Example of a generated gemspec:

```ruby
  spec.files = Dir.chdir(__dir__) do
    `git ls-files -z`.split("\x0").reject do |f|
      f.start_with?(*%w[my_project.gemspec bin test spec features .git .circleci appveyor])
    end
  end
```

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(__dir__) do
     `git ls-files -z`.split("\x0").reject do |f|
-      (File.expand_path(f) == __FILE__) || f.match(%r{\A(?:(?:bin|test|spec|features)/|\.(?:git|circleci)|appveyor)})
+      (File.expand_path(f) == __FILE__) || f.start_with?(*%w[bin test spec features .git .circleci appveyor])
     end
   end
   spec.bindir = "exe"


deivid-rodriguez:
Merging since there was no further feedback and I like this.

bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -29,7 +29,7 @@ Gem::Specification.new do |spec|
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(__dir__) do
     `git ls-files -z`.split("\x0").reject do |f|
-      (File.expand_path(f) == __FILE__) || f.match(%r{\A(?:(?:bin|test|spec|features)/|\.(?:git|circleci)|appveyor)})
+      f.start_with?(*%w[<%= config[:name] %>.gemspec bin test spec features .git .circleci appveyor])
```
simi, line 32: Any reason to not keep `__FILE__`? That's bulletproof also for later rename of the gem.

deivid-rodriguez: I think the point of this PR is to simplify this list and make it more approachable, and in my opinion a hardcoded gemspec name helps with that.

I think renaming a gem is something super rare that's going to be messy anyways, so I'd take the simpler version even if it makes renaming a gem need one more edit.

deivid-rodriguez: But no big deal, if you prefer to keep it, let's keep it. Switching to `start_with?` already improves things.

simi: IMHO renaming is quite common for pushed gems, since often you'll find out push was rejected due to name is already reserved or it is in reserved "range". We can do the opposite, and make it actual specification validation. Feel free to push this forward.

deivid-rodriguez: I see. Sounds a bit painful, since you have to rename files/requires/modules/classes, so I guess it's best to not make it more painful. In any case, if they are pushing right after generation to reserve the name, probably easier to run `bundle gem` again. If they're not, I guess second time people they will check availability beforehand 😅.

Anyways, let me amend it to make this a strict non-controversial improvement.

PR #6346, opened 2023-02-06
Add Ruby 3.2 and 3.3 platforms to Gemfile DSL

composerinteralia:

Ruby 3.2 platforms are not yet supported in the Gemfile DSL:
 
```
gem "abc", platform: :ruby_32
```

Along the same lines as https://github.com/rubygems/rubygems/pull/5469, this adds support for Ruby 3.2 platforms.


composerinteralia:
Seeing https://github.com/rubygems/rubygems/pull/5469#issuecomment-1098428186, I played around with a test to catch this earlier in the future, and ended up with something like this:


```rb
    it "allows platforms matching the running Ruby version" do
      platform = "ruby_#{RbConfig::CONFIG["MAJOR"]}#{RbConfig::CONFIG["MINOR"]}"
      subject.gem("foo", :platform => platform)
    end
```

Do you think that is worth adding? Or is there a better way of getting at that?

deivid-rodriguez:
Nice, I like it! It should make our daily CI against ruby-head fail, and then fixed after adding `:ruby_33`.

deivid-rodriguez (approved):
Approving the current changeset since it seems good to introduce right away, but feel free to also add the extra spec on this PR if you prefer!

composerinteralia:
Confirmed the test fails on 3.3: https://github.com/rubygems/rubygems/actions/runs/4106956899/jobs/7085847793

I'll add those platforms and then this should be ready to go.

deivid-rodriguez (approved):
Thanks a bunch!

composerinteralia:
https://github.com/rubygems/rubygems/actions/runs/4107455290/jobs/7086969404 looks like it succeeded, but is stuck 😕

deivid-rodriguez:
Indeed, it seems like an actions glitch?

composerinteralia:
This time https://github.com/rubygems/rubygems/actions/runs/4112632902/jobs/7097788892 got cancelled for some reason 😢 

I'll rebase to trigger this one more time 🤞🏻

deivid-rodriguez:
Oh boy, sorry about that. Our CI has become a little bit unstable since adding Ruby 3.2 :(

simi:
Finally 🟢 💪.

PR #6352, opened 2023-02-06
Use `RbSys::ExtensionTask` when creating new rust gems

ianks:
This PR makes it so `bundle gem --ext=rust` will use [`RbSys::ExtensionTask`](https://www.rubydoc.info/gems/rb_sys/0.9.63/RbSys/ExtensionTask). This is a thin wrapper over `Rake::ExtensionTask`, which is tailored for Rust and addresses a number of shortcomings of using the default configuration. Namely:

- Removes duplicative `target` directories (reducing compilation times and cache bloat)
- Uses `cargo metadata` to find extension dir (similar to @matsadler PR).
- Adds some helpers for compiling debug-friendly builds

### Links

- [See more in the PR here](https://github.com/oxidize-rb/rb-sys/pull/144)
- [PR integrating this into wasmtime-rb](https://github.com/bytecodealliance/wasmtime-rb/pull/132)

### DIFF
--- a/bundler/lib/bundler/templates/newgem/Rakefile.tt
+++ b/bundler/lib/bundler/templates/newgem/Rakefile.tt
@@ -41,6 +41,15 @@ require "standard/rake"
 <% if config[:ext] -%>
 <% default_task_names.unshift(:compile) -%>
 <% default_task_names.unshift(:clobber) unless config[:ext] == 'rust' -%>
+<% if config[:ext] == 'rust' -%>
+require "rb_sys/extensiontask"
+
+task build: :compile
+
+RbSys::ExtensionTask.new(<%= config[:name].inspect %>) do |ext|
+  ext.lib_dir = "lib/<%= config[:namespaced_path] %>"
+end
+<% else -%>
 require "rake/extensiontask"
 
 task build: :compile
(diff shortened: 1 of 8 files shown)


hsbt (approved):
👍  I confirmed to usage for `rb_sys/extensiontask` via https://github.com/oxidize-rb/rb-sys/tree/main/gem#rbsysextensiontask

bundler/lib/bundler/templates/newgem/Rakefile.tt
```diff
@@ -41,13 +41,23 @@ require "standard/rake"
 <% if config[:ext] -%>
 <% default_task_names.unshift(:compile) -%>
 <% default_task_names.unshift(:clobber) unless config[:ext] == 'rust' -%>
+<% if config[:ext] == 'rust' -%>
+require "rb_sys/extensiontask"
+
+task build: :compile
+
+RbSys::ExtensionTask.new(<%= config[:name].inspect %>) do |ext|
```
y-yagi, line 49: @ianks 
Don't need to pass `gemspec` here? According to the document, it requires here. 
https://github.com/oxidize-rb/rb-sys/tree/main/gem#rbsysextensiontask  

If my understanding is correct, it is needed to check some data in the `gemspec`.
https://github.com/oxidize-rb/rb-sys/blob/dfc51c60bef89925bfded1713e9ec02c5dfb639d/gem/lib/rb_sys/extensiontask.rb#L139-L149

y-yagi: Thanks. I understood the intent. I was curious because using `RbSys::ExtensionTask` without `gemspec` raises an error now.  It seems it's not intended so I will create a patch to `rb-sys`.

ianks: TyReleased in v0.9.69

PR #6355, opened 2023-02-07
Avoid crashing when installing from a corrupted lockfile

composerinteralia:

I did a bad thing (script that edits the Gemfile.lock directly) and ended up with a Gemfile.lock that was completely missing some indirect dependencies. While this is my fault and an error is reasonable, I noticed that the error got progressively less friendly in recent versions of bundler.

Something similar came up in https://github.com/rubygems/rubygems/issues/6210, and this commit would have helped with that case as well (although we've already handled that a different way with #6219).

## Previous behavior

Back on Bundler 2.2.23, a corrupt lockfile like this would cause a helpful error:

```
Unable to find a spec satisfying minitest (>= 5.1) in the set. Perhaps the lockfile is corrupted?
```

Bundler 2.2.24 to 2.3.26 gave a helpful (if inaccurate, since an old Bundler was not necessarily at fault) warning:

```
Warning:
Your lockfile was created by an old Bundler that left some things out.
Because of the missing DEPENDENCIES, we can only install gems one at a time,
instead of installing 16 at a time.
You can fix this by adding the missing gems to your Gemfile, running bundle
install, and then removing the gems from your Gemfile.
The missing gems are:
* minitest depended upon by activesupport
```

But then continued on and crashed while trying to report the unmet dependency:

```
--- ERROR REPORT TEMPLATE -------------------------------------------------------

NoMethodError: undefined method `full_name' for nil:NilClass
lib/bundler/installer/parallel_installer.rb:127:in `block (2 levels) in check_for_unmet_dependencies'

...
```

Bundler 2.4.0 and up crash as above when jobs=1, but crash even harder when run in parallel:

```
--- ERROR REPORT TEMPLATE -------------------------------------------------------

fatal: No live threads left. Deadlock?
3 threads, 3 sleeps current:0x00007fa6b6704660 main thread:0x00007fa6b6704660
* #<Thread:0x000000010833b130 sleep_forever>
   rb_thread_t:0x00007fa6b6704660 native:0x0000000108985600 int:0

* #<Thread:0x0000000108dea630@Parallel Installer Worker #0 tmp/1/gems/system/gems/bundler-2.5.0.dev/lib/bundler/worker.rb:90 sleep_forever>
   rb_thread_t:0x00007fa6b67f67c0 native:0x0000700009a62000 int:0

* #<Thread:0x0000000108dea4a0@Parallel Installer Worker #1 tmp/1/gems/system/gems/bundler-2.5.0.dev/lib/bundler/worker.rb:90 sleep_forever>
   rb_thread_t:0x00007fa6b67f63c0 native:0x0000700009c65000 int:0

<internal:thread_sync>:18:in `pop'
tmp/1/gems/system/gems/bundler-2.5.0.dev/lib/bundler/worker.rb:42:in `deq'

...
```

Note that these crashes only happen when installing local. When installing from a remote we end up getting an `APIResponseMismatchError` via https://github.com/rubygems/rubygems/blob/2f48d60afe050862258212d0c68530f39fb5fdfa/bundler/lib/bundler/remote_specification.rb#L55

This commit fixes the confusing thread deadlock crash by detecting if dependencies are missing such that we'll never be able to enqueue. When that happens we treat it as a failure so the install can finish.

That gets us back to the `NoMethodError`, which this commit fixes by using a different warning in the case where no spec is found.

### DIFF
--- a/bundler/lib/bundler/installer/parallel_installer.rb
+++ b/bundler/lib/bundler/installer/parallel_installer.rb
@@ -47,6 +47,13 @@ module Bundler
         dependencies.all? {|d| installed_specs.include? d.name }
       end
 
+      # Check whether spec's dependencies are missing, which can indicate a
+      # corrupted lockfile
+      def dependencies_missing?(all_specs)
+        spec_names = all_specs.map(&:name)
+        dependencies.any? {|d| !spec_names.include? d.name }
+      end
+
       # Represents only the non-development dependencies, the ones that are
       # itself and are in the total list.
       def dependencies
@@ -115,7 +122,12 @@ module Bundler
 
       unmet_dependencies.each do |spec, unmet_spec_dependencies|
         unmet_spec_dependencies.each do |unmet_spec_dependency|
-          warning << "* #{unmet_spec_dependency}, dependency of #{spec.full_name}, unsatisfied by #{@specs.find {|s| s.name == unmet_spec_dependency.name && !unmet_spec_dependency.matches_spec?(s.spec) }.full_name}"
+          found = @specs.find {|s| s.name == unmet_spec_dependency.name && !unmet_spec_dependency.matches_spec?(s.spec) }
+          if found
+            warning << "* #{unmet_spec_dependency}, dependency of #{spec.full_name}, unsatisfied by #{found.full_name}"
+          else
+            warning << "* #{unmet_spec_dependency}, dependency of #{spec.full_name} but missing from lockfile"
+          end
         end
       end
 
@@ -212,6 +224,8 @@ module Bundler
         if spec.dependencies_installed? @specs
           spec.state = :enqueued
           worker_pool.enq spec
+        elsif spec.dependencies_missing? @specs
+          spec.state = :failed
         end
       end
     end
--- a/bundler/spec/lock/lockfile_spec.rb
+++ b/bundler/spec/lock/lockfile_spec.rb
@@ -1221,6 +1221,44 @@ RSpec.describe "the lockfile format" do
       and include("Either installing with `--full-index` or running `bundle update rack_middleware` should fix the problem.")
   end
 
+  it "errors gracefully on a corrupt lockfile" do
+    build_repo4 do
+      build_gem "minitest-bisect", "1.6.0" do |s|
+        s.add_dependency "path_expander", "~> 1.1"
+      end
+
+      build_gem "path_expander", "1.1.1"
+    end
+
+    gemfile <<~G
+      source "#{file_uri_for(gem_repo4)}"
+      gem "minitest-bisect"
+    G
+
+    # Corrupt lockfile (completely missing path_expander)
+    lockfile <<~L
+      GEM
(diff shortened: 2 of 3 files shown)


composerinteralia:
Oops, some failures from changing the wording a bit. I'll look into those.

bundler/lib/bundler/installer/parallel_installer.rb
```diff
@@ -110,12 +117,17 @@ def check_for_unmet_dependencies
 
       warning = []
       warning << "Your lockfile doesn't include a valid resolution."
-      warning << "You can fix this by regenerating your lockfile or trying to manually editing the bad locked gems to a version that satisfies all dependencies."
```
composerinteralia, line 113: I don't think `trying to manually editing` is quite grammatically correct. `trying to manually edit` might work as well, but I didn't think `trying to` helped much here.

This is not strictly related to the rest of this PR, so I can do it in a separate PR if it is distracting.

deivid-rodriguez: Yeah, if you don't mind extracting this, we can advertise your two PRs as different enhancements in the changelog!

deivid-rodriguez:
Thank you, looking good. Normally I try to automatically re-resolve and fix the lockfile if we detect it's corrupted. But it's tricky sometimes, so changing a bad error message with a good error message is great!

bundler/lib/bundler/installer/parallel_installer.rb
composerinteralia: 👍🏻 Will do. 

That'll also give me another chance at that windows Ruby 3.2 job 😅

bundler/lib/bundler/installer/parallel_installer.rb
```diff
@@ -115,7 +122,12 @@ def check_for_unmet_dependencies
 
       unmet_dependencies.each do |spec, unmet_spec_dependencies|
         unmet_spec_dependencies.each do |unmet_spec_dependency|
-          warning << "* #{unmet_spec_dependency}, depended upon #{spec.full_name}, unsatisfied by #{@specs.find {|s| s.name == unmet_spec_dependency.name && !unmet_spec_dependency.matches_spec?(s.spec) }.full_name}"
+          found = @specs.find {|s| s.name == unmet_spec_dependency.name && !unmet_spec_dependency.matches_spec?(s.spec) }
+          if found
+            warning << "* #{unmet_spec_dependency}, depended upon #{spec.full_name}, unsatisfied by #{found.full_name}"
```
composerinteralia, line 127: This will conflict with https://github.com/rubygems/rubygems/pull/6357, so might make sense to merge that one first

composerinteralia:
> Thank you, looking good. Normally I try to automatically re-resolve and fix the lockfile if we detect it's corrupted. But it's tricky sometimes, so changing a bad error message with a good error message is great!

Played around with that a bit and it does seems possible. I can do that in a couple weeks when I get back from a vacation.

deivid-rodriguez:
That'd be awesome! For now enjoy your vacation ❤️

PR #6357, opened 2023-02-08
Improve wording of unmet dependencies warning

composerinteralia:

I noticed some small grammar problems in the unmet dependencies warning while working on https://github.com/rubygems/rubygems/pull/6355

`trying to manually editing` doesn't seem quite grammatically correct. We could change it to `trying to manually edit` (is that a split infinitive?), but I don't think `trying to` adds much here so I've removed it instead so `editing` is the verb.

For the list of dependencies, the wording before this commit seemed to reverse the dependency. "B, depended on A" sounds like B depends on A (or did in the past but doesn't anymore?), but that's not correct. I think there's a missing word: "B, depended on by A", but I find "B, dependency of A" a bit nicer.

### DIFF
--- a/bundler/lib/bundler/installer/parallel_installer.rb
+++ b/bundler/lib/bundler/installer/parallel_installer.rb
@@ -110,12 +110,12 @@ module Bundler
 
       warning = []
       warning << "Your lockfile doesn't include a valid resolution."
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
💯

PR #6374, opened 2023-02-13
Bundler lockfile checksums (continued)

mercedesb:
This is based off of #5808.

This mitigates issues such as cache poisoning, or dependency confusion. If the remote gem server has gems compromised via those issues, then the application installing such gems can use the locked checksum to warn
there is a potential issue. e.g. [GHSA-2jmx-8mh8-pm8w](https://github.com/rubygems/rubygems.org/security/advisories/GHSA-2jmx-8mh8-pm8w)

See also discussion in https://github.com/rubygems/rubygems/issues/3379

Lock the checksum of each gem file in the lockfile, in a new CHECKSUMS section

Note only installed gems will lock the checksum. Gem that are not matching the current local platform will not lock its checksum (because it was never resolved) - however if they had a previous locked checksum it should be preserved)

After talking to @indirect, I added some additional functionality to what was already in #5808 

1. Use the checksum provided by the server if we have it. This provides security knowing if the gem you downloaded matches the gem on the server
2. Calculate the checksum from the gem on disk. This provides security knowing if the gem has changed between installs which is not as helpful as 1 but still a valuable security feature.
3. In some cases, neither 1 or 2 is possible in which case we don't put anything in the checksum and we maintain functionality as it is today.

I think my biggest outstanding question here is how should `bundle lock` behave? When should it add checksums and when shouldn't it?

My second question is I might have updated too many spec expectations? I think the answer to the previous question will shed some light on that.

There are still some failing specs but hopefully this moves the conversation along.


indirect:
> I think my biggest outstanding question here is how should bundle lock behave? When should it add checksums and when shouldn't it?

I think `bundle lock` should add any checksums that it has from the EndpointSpecifications, or even add checksums for any .gem files that happen to already be present on disk, but it should not download any .gem files. I'm open to hearing arguments for other approaches, but I'm pretty sure that's the closest match to the way that the lock command works today.

mercedesb:
Hi @segiddins and @deivid-rodriguez! @indirect told me to reach out and see if either of you has the availability to help me with the final 6 failing specs that I'm just super stuck on. 

The first couple I have absolutely no idea. 

And for the `lock_spec` failures, it seems like we might need some way to skip materializing LazySpecs when we haven't fetched the gems. I thought I could check for `definition.unlocking?` but that doesn't seem to do what I want.

- rspec ./spec/commands/clean_spec.rb:891 bundle clean keeps git extension directories when excluded by group
- rspec ./spec/cache/gems_spec.rb:279 bundle cache when previously cached doesn't remove gems with mismatched :rubygems_version or :date
- rspec ./spec/install/gemfile/specific_platform_spec.rb:778 bundle install with specific platforms does not remove ruby if gems for other platforms, and not present in the lockfile, exist in the Gemfile
- rspec ./spec/commands/lock_spec.rb:312 bundle lock updates the bundler version in the lockfile without re-resolving
- rspec ./spec/commands/lock_spec.rb:842 bundle lock when an update is available does not implicitly update
- rspec ./spec/commands/lock_spec.rb:898 bundle lock when an update is available accounts for changes in the gemfile

deivid-rodriguez:
Hei @mercedesb, thanks for the work! I'll try to find some time next week to look at this 👍.

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,42 @@
...
+    def match_spec?(spec)
+      name == spec.name &&
+        version == spec.version &&
+        platform.to_s == spec.platform.to_s
+    end
+
+    def to_lock
+      out = String.new
+
+      if platform == Gem::Platform::RUBY
```
segiddins, line 30: nit that totally isn't introduced by this PR:
this pattern is repeated a bunch (e.g. LazySpecification#to_lock), and it would be good to have a single method to format these strings and do the `platform == Gem::Platform::RUBY` dance, etc

bundler/lib/bundler/lazy_specification.rb
```diff
@@ -76,6 +76,19 @@ def to_lock
       out
     end
 
+    def materialize_for_checksum
```
segiddins, line 79: if this takes a block, add a `(&blk)` parameter for clarity?

bundler/lib/bundler/lockfile_generator.rb
```diff
@@ -65,6 +66,22 @@ def add_dependencies
       end
     end
 
+    def add_checksums
+      out << "\nCHECKSUMS\n"
+      definition.resolve.sort_by(&:full_name).each do |spec|
+        checksum = spec.to_checksum if spec.respond_to?(:to_checksum)
```
segiddins, line 72: when would the spec not respond to `to_checksum` ?

bundler/lib/bundler/lockfile_generator.rb
```diff
@@ -65,6 +66,22 @@ def add_dependencies
...
+        if spec.is_a?(LazySpecification)
+          spec.materialize_for_checksum do |materialized_spec|
+            checksum ||= materialized_spec.to_checksum if materialized_spec&.respond_to?(:to_checksum)
+          end
+        end
+
+        checksum ||= definition.locked_checksums.find {|c| c.match_spec?(spec) }
```
segiddins, line 79: this is going to be quadratic -- if we make a hash keyed by full name, we can avoid looping through the full array of locked checksums

bundler/lib/bundler/remote_specification.rb
```diff
@@ -97,12 +97,48 @@ def git_version
...
+    end
+
     private
 
     def to_ary
       nil
     end
 
+    def _fetched_gem
+      @_fetched_gem ||= begin
```
segiddins, line 129: are you concerned about keeping the entire gem contents in memory if it's cached on the remote spec?

bundler/lib/bundler/remote_specification.rb
```diff
@@ -97,12 +97,48 @@ def git_version
...
+        fetcher = gem_remote_fetcher
+        fetcher.headers = { "X-Gemfile-Source" => remote.original_uri.to_s } if remote&.original_uri
+        fetcher.download(self, remote.uri)
+      end
+    end
+
+    def gem_remote_fetcher
```
segiddins, line 136: I think there's a method in rubygems_integration.rb we can use for this.

Ooooh actually I think we can use `download_gem` to download the gem, to the cache path, and then checksum the cached file?

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Bundler
+  class Checksum
+    attr_reader :name, :version, :platform
+    attr_accessor :checksum
+
+    SHA256 = /\Asha256-([a-z0-9]{64}|[A-Za-z0-9+\/=]{44})\z/.freeze
+
+    def initialize(name, version, platform, checksum = nil)
```
segiddins, line 10: i think it makes sense for `checksum` here to be an array of `checksums`, so we have the flexibility to add more hashing algorithms down the road

bundler/lib/bundler/remote_specification.rb
```diff
@@ -97,12 +97,48 @@ def git_version
...
 
+    # we don't get the checksum from a server like we could with EndpointSpecs
+    # calculating the checksum from the file on disk still provides some measure of security
+    # if it changes from install to install, that is cause for concern
+    def checksum
+      @checksum ||= begin
+        return unless _fetched_gem
+
+        File.open(_fetched_gem) do |f|
+          digest = Bundler::SharedHelpers.digest(:SHA256).new
```
segiddins, line 108: can we share this with the code in `RubyGemsGemInstaller#validate_bundler_checksum` ?

bundler/spec/commands/install_spec.rb
```diff
@@ -545,6 +545,8 @@
 
          DEPENDENCIES
 
+         CHECKSUMS
```
segiddins, line 548: do we want to print this section when it's empty?

lib/rubygems/specification.rb
```diff
@@ -763,6 +763,13 @@ def test_files=(files) # :nodoc:
 
   attr_accessor :specification_version
 
+  ##
+  # The checksum for this gem
+  #
+  # Do not set this, it is set when the gemspec is indexed (if possible).
```
segiddins, line 769: if we don't want it to be set, should we make it an `attr_reader` and use a different method to write to the ivar?

lib/rubygems/specification.rb
```diff
@@ -2716,4 +2723,22 @@ def flatten_require_paths # :nodoc:
   def raw_require_paths # :nodoc:
     @require_paths
   end
+
+  # if we don't get the checksum from the server
+  # calculating the checksum from the file on disk still provides some measure of security
+  # if it changes from install to install, that is cause for concern
+  def to_checksum
+    return Bundler::Checksum.new(name, version, platform, "sha256-#{checksum}") if checksum
```
segiddins, line 2731: is it safe for RubyGems to be referencing the just-introduced bundler constants?

bundler/tool/bundler/dev_gems.rb.lock
```diff
@@ -35,8 +35,8 @@ PLATFORMS
   ruby
   universal-java-11
   universal-java-18
-  x64-mingw-ucrt
   x64-mingw32
+  x64-unknown
```
segiddins, line 39: why is this changing?

bundler/spec/support/artifice/helpers/compact_index.rb
```diff
@@ -80,7 +80,7 @@ def gems(gem_repo = default_gem_repo)
               CompactIndex::Dependency.new(d.name, reqs)
             end
             checksum = begin
-                         Digest(:SHA256).file("#{gem_repo}/gems/#{spec.original_name}.gem").base64digest
+                         Digest(:SHA256).file("#{gem_repo}/gems/#{spec.original_name}.gem").hexdigest
```
segiddins, line 83: why is this changing?

bundler/spec/support/checksums.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Spec
+  module Checksums
+    class ChecksumsBuilder
```
segiddins, line 5: nice!

kuahyeow:
Thank you @mercedesb for moving this forward ! :)

bundler/spec/support/artifice/helpers/compact_index.rb
mercedesb: Because the actual compact index api uses a hexdigest so we want to use a hexdigest for testing

https://github.com/rubygems/rubygems.org/blob/00ca6264ba74727a24b6021e18937ca7ba5fd91b/app/models/gem_info.rb#L110

bundler/spec/commands/install_spec.rb
mercedesb: 🤷‍♀️ I'm not sure. Maybe it's helpful to have an obvious way to see that we were unable to get checksums? I could go either way

bundler/lib/bundler/lockfile_generator.rb
mercedesb: `LazySpecification` doesn't respond to `to_checksum`

bundler/lib/bundler/remote_specification.rb
mercedesb: Once I updated to use `download_gem`, I don't think this is a problem anymore

bundler/tool/bundler/dev_gems.rb.lock
mercedesb: I have no idea, I'll investigate in the history where this came from

lib/rubygems/specification.rb
mercedesb: This is a good question. I don't know, I didn't realize that it could be a problem so let me know if I need to remove it. If so, I think we'll need to go back to the drawing board b/c there are a lot of specs that resolve to `Gem::Specification` and so if we can't put a checksum on this class then I'm not totally sure which direction to go

deivid-rodriguez: We can customize/monkey-patch RubyGems classes in Bundler, but we do it in `lib/bundler/rubygems_ext.rb` to make them independent of the specific RubyGems version being installed.

bundler/lib/bundler/lockfile_generator.rb
segiddins: So it turns out it actually does, which is causing spec failures when git gems are being excluded

segiddins:
I just rebased this & did a not-small refactor in the last commit, but it got all the specs passing on my machine!

indirect:
woohoo, nice progress

deivid-rodriguez:
Wow nice, sorry I've been failing to get to this!

segiddins:
@deivid-rodriguez @indirect the specs here should finally be passing!

indirect:
Amazing! I'm very excited about shipping this.

@segiddins would you be interested in writing a blog post about this feature? I think it's a great candidate both for "look at how great this new feature is" and also "look at how interesting the process of getting this working was" 😂

indirect (approved):

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -21,6 +44,7 @@ class LockfileParser
       Gem::Version.create("1.10") => [BUNDLED].freeze,
       Gem::Version.create("1.12") => [RUBY].freeze,
       Gem::Version.create("1.13") => [PLUGIN].freeze,
+      Gem::Version.create("2.5.0") => [CHECKSUMS].freeze,
```
deivid-rodriguez, line 47: It makes sense to ship this with 2.5.0.

What happens to existing lockfiles not including this section? Are we keeping the current behavior in that case, and limiting this to fresh generated lockfiles, or this will fill in the checksums when Bundler 2.5 is run on them?

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -63,18 +87,26 @@ def initialize(lockfile)
       @dependencies = {}
       @state        = nil
       @specs        = {}
+      @lockfile_path = begin
+        Bundler.default_lockfile.relative_path_from(SharedHelpers.pwd)
+      rescue GemfileNotFound
+        "Gemfile.lock"
```
deivid-rodriguez, line 93: Is this new rescue fixing some issue that would be worth extracting to a separate PR? If not, is it necessary?

bundler/spec/bundler/definition_spec.rb
```diff
@@ -159,6 +167,9 @@
         DEPENDENCIES
           only_java
 
+        CHECKSUMS
+          only_java (1.1-java)
```
deivid-rodriguez, line 171: Why no checksum here? Because the lockfile has not yet been installed?

bundler/spec/cache/gems_spec.rb
```diff
@@ -283,6 +283,7 @@
         :rubygems_version => "1.3.2"
       simulate_new_machine
 
+      pending "Causes checksum mismatch exception"
```
deivid-rodriguez, line 286: Overlook?

bundler/spec/commands/clean_spec.rb
```diff
@@ -905,7 +905,7 @@ def should_not_have_gems(*gems)
     bundle :lock
     bundle "config set without development"
     bundle "config set path vendor/bundle"
-    bundle "install"
+    bundle "install", :verbose => true
```
deivid-rodriguez, line 908: For easier debugging I guess?

bundler/spec/commands/install_spec.rb
deivid-rodriguez: Had the same thought, you ended up deciding empty section was better?

bundler/spec/commands/install_spec.rb
```diff
@@ -888,16 +892,16 @@ def run
   context "with missing platform specific gems in lockfile" do
     before do
       build_repo4 do
-        build_gem "racc", "1.5.2"
+        build_gem "racca", "1.5.2"
```
deivid-rodriguez, line 895: Can we explain this? Commit message mentions a conflict with the default gem? Is this a test only issue, or could that happen in real life?

bundler/spec/commands/lock_spec.rb
```diff
@@ -98,6 +118,9 @@
       DEPENDENCIES
         foo
 
+      CHECKSUMS
+        #{checksum_for_repo_gem repo, "foo", "1.0", :empty => true}
```
deivid-rodriguez, line 122: Is this `empty` parameter needed? Will that just lock `foo (1.0)`?

bundler/spec/commands/update_spec.rb
```diff
@@ -527,6 +539,10 @@
       expect(the_bundle).to include_gems("activesupport 6.0.4.1", "tzinfo 1.2.9")
       expect(lockfile).to eq(expected_lockfile)
 
+      # needed because regressing to versions already present on the system
+      # won't add a checksum
+      expected_lockfile = expected_lockfile.gsub(/ sha256-[a-f0-9]+$/, "")
```
deivid-rodriguez, line 544: Why? Is it because the original package may not be available? Could we store the original checksum of installed packages somewhere, so that we can resurrect that on this kind of downgrades to installed versions?

Does the same thing happens when upgrading to installed versions? I feel this kind of flip flop is going to be surprising?

bundler/spec/commands/update_spec.rb
```diff
@@ -535,7 +551,7 @@
       lockfile original_lockfile
       bundle "lock --update"
       expect(the_bundle).to include_gems("activesupport 6.0.4.1", "tzinfo 1.2.9")
-      expect(lockfile).to eq(expected_lockfile)
+      expect(lockfile).to eq expected_lockfile
```
deivid-rodriguez, line 554: No big deal, but seems like an unnecessary style update.

deivid-rodriguez (commented):
This is great. I won't have time to review it in depth until I get back from holidays, but I gave it a quick pass.

bundler/lib/bundler/lockfile_parser.rb
segiddins: Same as any of the other sections that have been added in this list -- bundler will not add the section if run with `--frozen` or `--deployment`, or if not unlocking anything

bundler/lib/bundler/lockfile_parser.rb
segiddins: necessary because I wanted to pull out the call to `Bundler.default_lockfile`, which was already being used in the conflict message, but I expanded it to include in the `source` messages for parsed checksums, but it raises when no gemfile is found (which was actually hit in a couple of tests when parsing a lockfile)

bundler/spec/bundler/definition_spec.rb
segiddins: correct, and we don't use the compact index for `file:` sources :(

bundler/lib/bundler/lockfile_parser.rb
deivid-rodriguez: Under which situations do we parse a lockfile without an accompanying Gemfile? I guess that happened on unit specs for the parser, but not in real life? To clarify, I'm good with this, just curious.

bundler/spec/commands/install_spec.rb
segiddins: good question, I don't 100% remember

bundler/spec/commands/lock_spec.rb
segiddins: yeah, it omits the actual checksum

bundler/lib/bundler/lockfile_parser.rb
segiddins: yeah, that. and also my random scripts lying around that parse lockfiles :)

bundler/spec/commands/lock_spec.rb
deivid-rodriguez: I'd remove this parameter then and check the real content for consistency, because I've seen both styles used around.

martinemde (commented):
Looks awesome. I had a few questions and thoughts.

martinemde:
Hey, I opened #6917 with some fixes and refactors targeting this as base branch. Feel free to take what you like or merge if you think it's worth it. I was struggling to capture my actual feedback in the PR review (because I wanted to remove more than half of Checksum class)

martinemde:
I tried to rebase, but I ended up with a weird history. Pushed the original branch back up before rebase. Sorry for the noise.

bundler/spec/commands/lock_spec.rb
martinemde: I replaced all these with no_checksum or something similar so they are (mostly) uniform.

martinemde:
Strange, this was passing on the branch and now fails after merge.

martinemde:
Currently if you install a gem using `bundle install` and the gem already exists locally, it will install the gem without fetching anything remotely, then write the checksum of that gem to the Gemfile.lock.

I believe this behavior is incorrect. If we use the local gem checksum as an authority, then go to deploy, we could conceivably "verify" the install of a bad gem because the dev that installed it last had it installed already.

Since the command to download all checksums from the source doesn't quite exist, there will be a lot of people who initially install this and end up with a bunch of (hopefully fine but dubiously sourced) checksums stored that won't be updated until that gem is fetched during an update (at least with current feature set in this PR).

indirect:
@martinemde can you recap what's left on this PR and when you expect to be able to get to it? Thanks!

martinemde:
@indirect @segiddins I think this is ready for one more pass of reviews. Test should be passing and I’m not adding bundle lock changes here.

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,228 @@
+# frozen_string_literal: true
+
+module Bundler
+  class Checksum
+    DEFAULT_ALGORITHM = "sha256"
+    DEFAULT_BLOCK_SIZE = 16_384
+    private_constant :DEFAULT_BLOCK_SIZE
```
segiddins, line 7: should `DEFAULT_ALGORITHM` be private? and also maybe an array?

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    class << self
+      def from_gem(io, pathname, algo = DEFAULT_ALGORITHM)
+        digest = Bundler::SharedHelpers.digest(algo.upcase).new
+        digest << io.read(DEFAULT_BLOCK_SIZE) until io.eof?
```
segiddins, line 12: readpartial into a pre-allocated buffer?

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,228 @@
...
+        Checksum.new(algo, digest.hexdigest!, Source.new(:gem, pathname))
+      end
+
+      def from_api(digest, source_uri)
+        # transform the bytes from base64 to hex
+        hexdigest = digest.length == 44 ? digest.unpack("m0").first.unpack("H*").first : digest
```
segiddins, line 18: note: add a comment to use `unpack1` once we drop older rubies, it avoids the array allocation

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,228 @@
...
+
+        if hexdigest.length != 64
+          raise ArgumentError, "#{digest.inspect} is not a valid SHA256 hexdigest nor base64digest"
+        end
+
+        Checksum.new(DEFAULT_ALGORITHM, hexdigest, Source.new(:api, source_uri))
+      end
+
+      def from_lock(lock_checksum, lockfile_location)
+        algo, digest = lock_checksum.strip.split("-")
```
segiddins, line 28:
```suggestion
        algo, digest = lock_checksum.strip.split("-", 2)
```

bundler/lib/bundler/checksum.rb
```diff
@@ -0,0 +1,228 @@
...
+      #   rake (10.3.2) sha256-abc123 from #{to_s}
+      def to_s
+        case type
+        when :lock
+          "the lockfile CHECKSUMS at #{location}"
+        when :gem
+          "the gem at #{location}"
+        when :api
+          "the API at #{location}"
+        when
```
segiddins, line 119:
```suggestion
        else
```

bundler/lib/bundler/lazy_specification.rb
```diff
@@ -20,11 +20,15 @@ def initialize(name, version, platform, source = nil)
     end
 
     def full_name
-      @full_name ||= if platform == Gem::Platform::RUBY
-        "#{@name}-#{@version}"
-      else
-        "#{@name}-#{@version}-#{platform}"
-      end
+      @full_name ||= name_tuple.full_name
```
segiddins, line 23: this will cause a name_tuple to be allocated which feels unnecessary to me

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -65,18 +89,26 @@ def initialize(lockfile)
...
 
       if lockfile.match?(/<<<<<<<|=======|>>>>>>>|\|\|\|\|\|\|\|/)
-        raise LockfileError, "Your lockfile contains merge conflicts.\n" \
-          "Run `git checkout HEAD -- #{SharedHelpers.relative_lockfile_path}` first to get a clean lock."
+        raise LockfileError, "Your #{@lockfile_path} contains merge conflicts.\n" \
+          "Run `git checkout HEAD -- #{@lockfile_path}` first to get a clean lock."
       end
 
-      lockfile.split(/(?:\r?\n)+/) do |line|
+      lockfile.split(/((?:\r?\n)+)/).each_slice(2) do |line, whitespace|
```
segiddins, line 104: hmmm I think chaining the `each_slice` off of the `split` might cause the `split` to allocate an array, and the `each_slice` might (I'm not sure) allocate arrays

```
         split block   213.751k memsize (    74.015k retained)
                         1.585k objects (   701.000  retained)
                        50.000  strings (    50.000  retained)
    split each slice   243.360k memsize (    54.240k retained)
                         2.378k objects (   702.000  retained)
                        50.000  strings (    50.000  retained)

Comparison:
         split block:     213751 allocated
    split each slice:     243360 allocated - 1.14x more
         split block   207.000  i/100ms
    split each slice   183.000  i/100ms
         split block      2.072k (± 0.7%) i/s -     10.557k in   5.095838s
    split each slice      1.822k (± 0.9%) i/s -      9.150k in   5.023257s

Comparison:
         split block:     2071.8 i/s
    split each slice:     1821.7 i/s - 1.14x  slower
```

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -176,6 +212,31 @@ def parse_dependency(line)
...
+    def parse_checksum(line)
+      return unless line =~ NAME_VERSION
+
+      spaces = $1
+      return unless spaces.size == 2
+      name = $2
+      version = $3
+      platform = $4
+      checksums = $6
+      return unless checksums
```
segiddins, line 224: nit: if this return is done before referencing `$2-4`, we can avoid allocating those strings in the short-circuit case

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -176,6 +212,31 @@ def parse_dependency(line)
...
+
+      version = Gem::Version.new(version)
+      platform = platform ? Gem::Platform.new(platform) : Gem::Platform::RUBY
+      full_name = Gem::NameTuple.new(name, version, platform).full_name
+      # Don't raise exception if there's a checksum for a gem that's not in the lockfile,
+      # we prefer to heal invalid lockfiles
+      return unless spec = @specs[full_name]
+
+      checksums.split(",").each do |lock_checksum|
```
segiddins, line 233:
```suggestion
      checksums.split(",") do |lock_checksum|
```

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -359,6 +359,27 @@ def extensions_dir
     end
   end
 
+  require "rubygems/name_tuple"
+
+  class NameTuple
+    def self.new(name, version, platform="ruby")
```
segiddins, line 365: should this be done in `initialize` instead of `new`?

segiddins (commented):
Great teamwork here @martinemde, thanks for picking this up & getting it to the finish line!

bundler/lib/bundler/checksum.rb
martinemde: oops, haha.

bundler/lib/bundler/lazy_specification.rb
martinemde: Reverting. Should I do something els with `lock_name` below too?

bundler/lib/bundler/lockfile_parser.rb
martinemde: Updated. Let me know if you thing this is better.

bundler/lib/bundler/checksum.rb
martinemde: I'll make it private which will protect our ability to change it later if we need an array.

bundler/lib/bundler/rubygems_ext.rb
martinemde: Certain rubies (we won't name names) segfault when I do that. Same if I redeclare `hash` on this object. This is the least segfaulty way since it doesn't undef or overwrite a method.

bundler/spec/commands/update_spec.rb
martinemde, line 554:
```suggestion
      expect(lockfile).to eq(expected_lockfile)
```

PR #6376, opened 2023-02-14
Use daily check and enabled auto-merge feature for dependabot

hsbt:

We often received [dependabot updates](https://github.com/rubygems/rubygems/pulls/app%2Fdependabot) every week and triage manually.

It's important for preventing supply chain attack. But We can merge them with CI status check automatically because it's only effect to github actions workflow, not gem dependencies.

I added auto-merge workflow with GitHub PAT. Currently, I set my PAT to `RUBYGEMS_GITHUB_TOKEN` for organizational secrets. It's good to use https://github.com/bundlerbot instead of my PAT. /cc @indirect 

@deivid-rodriguez @simi How about this?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,12 +3,12 @@ updates:
   - package-ecosystem: 'github-actions'
     directory: '/'
     schedule:
-      interval: 'weekly'
+      interval: 'daily'
   - package-ecosystem: 'cargo'
     directory: '/test/rubygems/test_gem_ext_cargo_builder/custom_name/'
     schedule:
-      interval: 'weekly'
+      interval: 'daily'
   - package-ecosystem: 'cargo'
     directory: '/test/rubygems/test_gem_ext_cargo_builder/rust_ruby_example/'
     schedule:
-      interval: 'weekly'
+      interval: 'daily'
--- a/.github/workflows/dependabot_automerge.yml
+++ b/.github/workflows/dependabot_automerge.yml
@@ -1,0 +1,26 @@
+# from https://github.com/gofiber/swagger/blob/main/.github/workflows/dependabot_automerge.yml
+name: Dependabot auto-merge
+on:
+  pull_request_target:
+
+jobs:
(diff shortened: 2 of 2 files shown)


hsbt:
I only check `^rubygems*|^bundler*` statuses.

deivid-rodriguez:
So if a new action release breaks a workflow that's not `rubygems-*` or `bundler-*`, will we merge it even if it breaks the workflow? I don't personally feel we need this, specially now that we have merge queues, so no rebasing or anything is needed, just click on "Merge when ready". But if we want to do it, we should use a method that has the same CI requirements as regular PRs for getting merged.

hsbt:
> So if a new action release breaks a workflow that's not `rubygems-*` or `bundler-*`, will we merge it even if it breaks the workflow?

There is no strong reason to use only `rubygems-*` and `bundler-*`. It's enough to check with CI statuses at first.

I'm +1 to cover all statuses now. I update target regexp.

deivid-rodriguez:
How does this work with Merge Queue by the way?

hsbt:
I don't know that yet. I try to check how work this with merge queue feature on other repository.

simi:
I'm a little sceptic to introduce any auto-merge to critical repository like this. I'm triaging all updates manually (I do it anyway since I do maintain multiple repos). Sometimes I leave some PRs also for the others to not be the only one, anyway if this becomes annoying, I can try better and handle all update PRs.

On the other side, it is not clear to me what's the risk in here. In theory, automated PR can update the code or steal CI secrets without notice. But even with stolen secrets it is not possible (currently) to release new version for or distribute the code in any other way. :thinking: Maybe I'm too conservative only.

deivid-rodriguez:
I'm neutral. I don't think auto-merge buys us much here, since it's just a few PRs every week which as @simi said we review anyways. But I also don't think this poses any serious security concern either. So if anyone feels strongly about this, please move this forward.

hsbt:
I withdraw this proposal.

PR #6384, opened 2023-02-15
Issue 6353: safe load all marshaled data

### LINKED ISSUE #6353: Expand SafeMarshal to cover all loading of marshaled data in RubyGems and Bundler
In https://github.com/rubygems/rubygems/pull/6141, we landed a `SafeMarshal` class to deserialize data returned from the dependency API. Let's expand use of `SafeMarshal` to include every place in RubyGems and Bundler where we load marshaled files, to remove the possibility of an attack by a malicious server.

mercedesb:

Resolves #6353. 

In https://github.com/rubygems/rubygems/pull/6141, we landed a SafeMarshal class to deserialize data returned from the dependency API. We want to expand the use of SafeMarshal to include every place in RubyGems and Bundler where we load marshaled files, to remove the possibility of an attack by a malicious server.

Use the `safe_load_marshal` method when fetching Marshal.specs.4.8.gz and when inflating gem specifications.

Made the legacy `Bundler.load_marshal` method private so that others will be less tempted to use it directly.

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -525,12 +525,6 @@ EOF
       load_marshal(data, :marshal_proc => SAFE_MARSHAL_PROC)
     end
 
-    def load_marshal(data, marshal_proc: nil)
-      Marshal.load(data, marshal_proc)
-    rescue TypeError => e
-      raise MarshalError, "#{e.class}: #{e.message}"
-    end
-
     def load_gemspec(file, validate = false)
       @gemspec_cache ||= {}
       key = File.expand_path(file)
@@ -618,6 +612,12 @@ EOF
     end
 
     private
+
+    def load_marshal(data, marshal_proc: nil)
+      Marshal.load(data, marshal_proc)
+    rescue TypeError => e
+      raise MarshalError, "#{e.class}: #{e.message}"
+    end
 
     def eval_yaml_gemspec(path, contents)
       Kernel.require "psych"
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -39,7 +39,7 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
-  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash].freeze
+  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash, Gem::Version].freeze
   SAFE_MARSHAL_ERROR = "Unexpected class %s present in marshaled data. Only %s are allowed."
   SAFE_MARSHAL_PROC = proc do |object|
     object.tap do
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -39,7 +39,7 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
-  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash, Gem::Version].freeze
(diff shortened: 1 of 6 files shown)


bundler/spec/bundler/bundler_spec.rb
```diff
@@ -5,9 +5,9 @@
 
 RSpec.describe Bundler do
   describe "#load_marshal" do
-    it "loads any data" do
```
segiddins, line 8: we should probably keep the test around

bundler/lib/bundler.rb
```diff
@@ -39,7 +39,7 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
-  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash].freeze
+  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash, Gem::Version, Gem::Specification].freeze
```
segiddins, line 42: can we add a test that covers loading these two classes?

bundler/lib/bundler/fetcher.rb
```diff
@@ -102,11 +102,11 @@ def fetch_spec(spec)
       uri = Bundler::URI.parse("#{remote_uri}#{Gem::MARSHAL_SPEC_DIR}#{spec_file_name}.rz")
       if uri.scheme == "file"
         path = Bundler.rubygems.correct_for_windows_path(uri.path)
-        Bundler.load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
+        Bundler.safe_load_marshal Bundler.rubygems.inflate(Gem.read_binary(path))
```
segiddins, line 105: let's add a test that has an exception raised when attempting to load an unexpected class through this codepath? and same with the one in `RubyGemsIntegration.fetch_specs`

segiddins (approved):
Exciting! I think it would just be good to add a few tests here to prevent regressions, and make sure the user experience when loading a disallowed class is solid.

bundler/lib/bundler.rb
simi: I'm not 100% sure it is good idea to allow those classes everywhere. Initial effort was to limit structures parsed from Dependency API. I don't see any reason why to allow `Gem::Version` and `Gem::Specification` as valid objects coming from Dependency API.

Are those needed to read gemspecs? :thinking: Are those the only valid gemspecs classes? I think at least `Gem::Requirement` is valid for gemspecs. Also I'm not sure this doesn't change across the versions.

Thinking about this, wouldn't make sense to make special loader for gemspecs?

indirect: I believe the idea is that we need to avoid any classes that contain a remote execution path. If we are sure Gem::Specification does not allow remote execution, I think it is ok to allow that class everywhere—if a server returns it from the Dependency API, there will definitely be an exception somewhere else in the code.

It might make sense to add a special loader just for gemspecs if that makes it easier to separate out, but personally I would be ok with one SafeMarshal loader as long as the list of allowed classes does not include any exploitable classes.

mercedesb: Happy to go whatever direction is decided here

simi: I have no big preference. The only concern is still regarding `Gem::Requirement` being added. I would be really surprised see specs passing without `Gem::Requirement` in this list. Are we missing some code coverage? Or am I totally missing something in here and it is not needed? :thinking:

mercedesb: I do have some linting issues to fix, but all the tests pass without `Gem::Requirement` in this list. However, I'm not familiar enough with things to say whether that's a b/c there is missing test coverage or whether it is not needed.

simi:
@indirect are you 100% `Gem::Requirement` is not needed in the list? :thinking:

PR #6387, opened 2023-02-16
Add debugging docs

mercedesb:

Figuring out how to set up test cases and debug for a new maintainer is hard and there aren't existing docs to refer to. I had a call with @indirect and he walked me through how he debugs the Bundler code. This doc is a starting point for how to debug various test cases.

- Add DEBUGGING.md to document how to debug Bundler

### DIFF
--- a/bundler/doc/development/DEBUGGING.md
+++ b/bundler/doc/development/DEBUGGING.md
@@ -9,7 +9,7 @@ Make sure you've followed the [development setup](SETUP.md) docs before trying t
 Your best option is to print debug in the test suite. Put `puts` statements anywhere you want to see an object or variable and you'll see your `puts` in the console output.
 
 ```ruby
-puts "@defintion: #{@defintion}"
+puts "@definition: #{@definition}"
 puts "specification.class.name: #{specification.class.name}"
 puts "spec.method(:to_checksum).source_location: #{spec.method(:to_checksum).source_location}"
 # etc
--- a/bundler/doc/development/DEBUGGING.md
+++ b/bundler/doc/development/DEBUGGING.md
@@ -34,9 +34,9 @@ To learn more about using IRB, [TODO: link to doc]
 
 Interactive debugging is like REPL + the ability to advance the code execution line by line.
 
-When testing Bundler locally, you can use any debugger that you are comfortable with for interactive debugging. 
+When testing Bundler locally, you can use any debugger that you are comfortable with for interactive debugging.
 
-[`debug`](https://github.com/ruby/debug) and [`pry-byebug`](https://github.com/deivid-rodriguez/pry-byebug) are common favorites. `debug` has been included with Ruby since v3.1. 
+[`debug`](https://github.com/ruby/debug) and [`pry-byebug`](https://github.com/deivid-rodriguez/pry-byebug) are common favorites. `debug` has been included with Ruby since v3.1.
 
 You just need your chosen debugger gem installed globally. Then you will need to require it on the command line before running your local Bundler.
 
@@ -92,7 +92,7 @@ bin/parallel_rspec
 
 Then you'll find built gems in the `bundler/tmp` directory, e.g. `bundler/tmp/1/gems/remote1/`
 
-You can set up your Gemfile 
+You can set up your Gemfile with a file source pointing to the built gems from your test run.
 
 ```ruby
 # frozen_string_literal: true
@@ -102,4 +102,4 @@ source "file:///[path to repo's bundler directory]/tmp/1/gems/remote1/"
 gem "rack", '=0.9.1'
 ```
 
-And then you can test in the same way as you did in the previous example when fetching the gem from the RubyGems source.
+Then you can test in the same way as you did in the previous example when fetching the gem from the RubyGems source.
--- a/bundler/doc/development/DEBUGGING.md
+++ b/bundler/doc/development/DEBUGGING.md
@@ -4,9 +4,11 @@ Debugging Bundler can be challenging, don't be discouraged 🤗.
 
 Make sure you've followed the [development setup](SETUP.md) docs before trying to debug.
 
-## Debugging in tests
+## Print debugging
 
-Your best option is to print debug in the test suite. Put `puts` statements anywhere you want to see an object or variable and you'll see your `puts` in the console output.
+The easiest way to debug is to print debug. Put `puts` statements anywhere in the code that you want to see an object or variable and you'll see your `puts` in the console output.
+
+This can be especially helpful when running tests.
 
 ```ruby
 puts "stacktrace: #{caller_locations(0).join("\n")}"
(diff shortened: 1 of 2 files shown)


simi:
:thinking: Usually I just put `binding.irb` in test and run that one test file or case individually. For integration testing I just run `ruby setup.rb` to install dev version locally or even run `ruby -Ilib  bin/gem` locally to test `gem` command.

bundler/doc/development/DEBUGGING.md
```diff
@@ -0,0 +1,76 @@
+# Debugging
+
+Debugging Bundler can be challenging, don't be discouraged 🤗.
+
+Make sure you've followed the [development setup](SETUP.md) docs before trying to debug.
+
+## Debugging in tests
+
+Your best option is to print debug in the test suite. Put `puts` statements anywhere you want to see an object or variable and you'll see your `puts` in the console output.
```
deivid-rodriguez, line 9: I mainly use `puts` debugging too! Sometimes printing `caller_locations(0).join("\n")` helps to see what the stacktrace is at a given point.

bundler/doc/development/DEBUGGING.md
```diff
@@ -0,0 +1,76 @@
...
+When testing Bundler locally, you can use pry for interactive debugging.
+
+The easiest way to test locally is to set up a directory with a Gemfile and run your Bundler shell alias (see the [development setup](SETUP.md) docs for instructions on setting up the alias).
+
+We recommend putting this directory inside of `tmp` so that your local tests don't accidentally get committed.
+
+```bash
+cd tmp
+mkdir [name of directory for local testing] && cd [name of directory for local testing]
+dbundle init
```
deivid-rodriguez, line 31: I normally point directly to my working copy with something like `/path/to/rubygems/bundle/exe/bundle`. These days once you land on a copy of Bundler, you never get out of it because we always use `require_relative`, so messing with the `$LOAD_PATH` should not be necessary.

deivid-rodriguez (commented):
I added a couple of comments. Debugging tips never hurt, thanks for documenting this!

bundler/doc/development/DEBUGGING.md
mercedesb: Wonderful, I'll add that to the list of examples

bundler/doc/development/DEBUGGING.md
mercedesb: Ah so should we change the alias instructions in the setup doc?

bundler/doc/development/DEBUGGING.md
```diff
@@ -0,0 +1,76 @@
...
+
+```ruby
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem 'tiny_css'
+```
+
+Put a `binding.pry` anywhere you want the debugger to break. Run your Bundler shell alias.
```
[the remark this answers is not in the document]
mercedesb: Because `binding.irb` is a REPL and not a debugger 🤷‍♀️ It's nice, but you can't really interactive debug. I can add that to these docs as a tool to use though

mercedesb:
> For integration testing I just run `ruby setup.rb` to install dev version locally or even run `ruby -Ilib  bin/gem` locally to test gem command.

> I normally point directly to my working copy with something like `/path/to/rubygems/bundle/exe/bundle`

It sounds like there are multiple ways to test locally which don't match up with what we currently recommend to new maintainers in [SETUP.md](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/SETUP.md). Is the alias a totally bad idea? And should we update that doc too?

bundler/doc/development/DEBUGGING.md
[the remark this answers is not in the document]
simi: That's it. Some info is at https://github.com/rubygems/rubygems/blob/030e285881ac99608fe8a23121b80ce1ce14ed6e/bundler/doc/development/SETUP.md#debugging-with-pry. Instead of this "complex" setup, you can do just `binding.irb` and you have basic debug console. For debug console, you can use `binding.break` (you need to `require 'debug'` somehow from [debug](https://github.com/ruby/debug) gem, which is distributed with Ruby if I remember well and there is no need to install additional dependency).

bundler/doc/development/DEBUGGING.md
deivid-rodriguez: I don't mind the current recommendation to set an alias, and it should also work just fine, I don't think you need to change anything. Just wanted to mention the way I do it :)

bundler/doc/development/DEBUGGING.md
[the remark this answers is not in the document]
simi: You can install debug gem the same way you can install `pry-byebug` for older Rubies. It supports Ruby 2.6+. 

Anyway the whole discussion is good sign that this setup is not simple. Experienced developers can use their favourite tools, there is almost nothing specific. And for beginners it would be maybe better to point to some generic Ruby debugging tutorial where basics like puts debugging and real debuggers are explained. :thinking: 

On the other way, if you think @indirect this kind of guide is valuable in here, let's start somewhere. I'm ok with anything that works. I was just about share my experience.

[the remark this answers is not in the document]
mercedesb: Based on this convo, I reorganized the doc slightly and tried to add just enough info to get new maintainers started but would like to link out to other places for more learning. Do we care who/what we link to? Here are some options (not in the doc yet)

- REPL debugging more info: https://st0012.dev/whats-new-in-ruby-3-2-irb
- Prints debugging more info: https://medium.com/@calielrushton/puts-prints-and-more-printing-to-the-console-in-ruby-7830cdaace0a
- Interactive debugging more info: https://github.com/ruby/debug#how-to-use

bundler/doc/development/DEBUGGING.md
```diff
@@ -0,0 +1,105 @@
...
+
+## REPL debugging
+
+REPL (or Read-Eval-Print Loop) is a way to interact with your code.
+
+With REPL you can look at the values of objects and variables, the stack trace, where methods are defined etc.
+
+To use REPL, place a `binding.irb` wherever you'd like to take a look around. An interactive `irb` console will open when your code gets to your breakpoint.
+
+To learn more about using IRB, [TODO: link to doc]
```
simi, line 31: Maybe `https://github.com/ruby/irb#the-bindingirb-breakpoint`?

bundler/doc/development/DEBUGGING.md
simi: I think everyone have different debugging habits :). I was using `ruby -Ilib -s ./exe/bundle -- --version` and even today it is usually enough to just call `./exe/bundle`. 

But indeed, `dbundle` is mentioned at various places. It is ok to keep it in here as well.

simi (approved):

PR #6388, opened 2023-02-17
TarReader::Entry fix for rewind when reading data.tar.gz in a gem

martinemde:

Zlib::GzipReader doesn't support #pos= so you can't rewind a TarReader::Entry that is wrapping a Zlib::GzipReader.

Gem::Package#open_tar_gz is used in many places to read the data.tar.gz contained within a .gem file. This normally works if you never rewind or try to use TarReader#seek. TarReader::Entry wraps the gzio, but attempting to rewind or seek within the gzip fails by `NoMethodError: undefined method 'pos=' for #<Zlib::GzipReader>`.

This would not be encountered during normal extract of a gem, but does happen when someone uses the methods on Gem::Package to try to access individual files within the gem, or to rewind an entry that's already been read.

I don't know if this really needs to be fixed in rubygems, but I'm pushing it upstream because I fixed it when parsing gems in memory. If this is useful, I can push up a few more fixes to the TarReader that help with seeking to individual files in the data.tar.gz, or whatever else I encounter that could be fixed more easily in rubygems itself.

The only solution supported by Zlib::GzipReader (that I can find) is to rewind and read back to the @orig_pos. This works correctly as shown in the test. This test fails with the above NoMethodError without this fix.

### DIFF
--- a/test/rubygems/test_gem_package_tar_reader_entry.rb
+++ b/test/rubygems/test_gem_package_tar_reader_entry.rb
@@ -216,14 +216,12 @@ class TestGemPackageTarReaderEntry < Gem::Package::TarTestCase
     tgz = util_gzip(@tar)
 
     Zlib::GzipReader.wrap StringIO.new(tgz) do |gzio|
-      begin
-        header = Gem::Package::TarHeader.from gzio
-        entry = Gem::Package::TarReader::Entry.new header, gzio
+      header = Gem::Package::TarHeader.from gzio
+      entry = gem::package::tarreader::entry.new header, gzio
 
-        assert_equal @contents, entry.read
-        entry.rewind
-        assert_equal @contents, entry.read, "read after rewind failed"
-      end
+      assert_equal @contents, entry.read
+      entry.rewind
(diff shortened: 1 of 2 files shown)


test/rubygems/test_gem_package_tar_reader_entry.rb
```diff
@@ -211,4 +211,19 @@ def test_zero_byte_file_readpartial
   ensure
     close_util_entry(zero_entry) if zero_entry
   end
+
+  def test_read_from_gzip_io
+    tgz = util_gzip(@tar)
+
+    Zlib::GzipReader.wrap StringIO.new(tgz) do |gzio|
+      begin
```
simi, line 219: :thinking: Any reason to wrap this into `begin` block?

martinemde: Oops, leftover from an `ensure` that was used in the first try at writing this test. I'll fix.

martinemde:
Closing in favor of #6390

PR #6179, opened 2022-12-22
Ask user to otp at webauthn verification url

aellispierce:
_**NOTE:** This is being merged into a feature branch, not main, so that work may continue to be developed in small pieces as we built out webauthn for the CLI_

## What was the end-user or developer problem that led to this PR?
Users with MFA enabled would like to use security devices (yubikey, touch id, etc) for logging in with WebAuthn on the CLI

## What is your fix for the problem, implemented in this PR?

- For users with MFA enabled, we request a webauthn verification url from rubygems.org on sensitive actions (sign in, yank, push, add/remove owners). 
- If the user _does not_ have a webauthn device enabled, there is no change in the user interaction. The api would respond with a 403 and we would continue to ask for a regular OTP code as normal.
- For users that do have webauthn enabled, the api would respond with the webauthn verification url. Users would then verify at that link, receive an OTP code, and then input that here. 
- This is **just** the first step, receiving the link and displaying it to users, the actual verification will happen in follow up PRs on the client and CLI. 

.org feature branch: https://github.com/rubygems/rubygems.org/pull/3298
Matching .org work: https://github.com/rubygems/rubygems.org/pull/3305

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6389, opened 2023-02-17
Create Webauthn Listener to retrieve OTP code from browser

jenshenny:
_This is being merged into a feature branch, not main, so that work may continue to be developed in small pieces as we built out webauthn for the CLI_

Part of implementing Webauthn support for the CLI https://github.com/rubygems/rubygems.org/issues/3453. 

Pivoting from the approach from https://github.com/rubygems/rubygems/pull/6179 as when a user authenticates their security device using a unique link to the browser, the browser will display an OTP code to paste into the CLI. This PR removes the pasting of the code by creating a listener to receive the code.

If the user has MFA enabled and has a webauthn device, a socket will be opened on the user's machine. A unique link will be shown to the user with the localhost port of the socket. 

After authentication, the server will redirect back to the socket with the OTP code (eg. `http://localhost:45673?code=Uvh6T57tkWuUnWYo`). The code will be used as the otp option in API requests.

Screenshot 2023-02-16 at 10 19 26 PM

Note:  if the user has TOTP and Webauthn enabled, the webauthn way of authentication is shown. As a follow up, a fallback option to OTP will be provided.

### Possible responses when receiving the code from the Gem host
1. 200 OK: OTP code was successfully retrieved.
2. 204 No Content: If the request was an OPTIONS request. Since the request was originally sent on the client side of RubyGems.org, a preflight CORS request is sent since this is a cross origin.
3. 400 Bad Request: URL doesn't have code param (eg. `GET http://localhost:45673`)
Screenshot 2023-02-16 at 10 28 48 PM

4. 404 Not Found: URL having a bad path (not the root path) (eg. `GET http://localhost:45673/lol?code=Uvh6T57tkWuUnWYo`)
Screenshot 2023-02-16 at 10 29 42 PM

5. 405 Method Not Allowed: URL was not a GET/OPTIONS request (eg. `POST http://localhost:45673?code=Uvh6T57tkWuUnWYo`)

## What should reviewers focus on
- If it works with the RubyGems.org changes (TBD)
- Is there an easier way to build socket responses using the std lib?
- Any edge cases missing when receiving a request from the Gem host
- Response bodies both in content type (should we respond in another format like json) and the actual content. Right now we are sending plain text "success" and "missing code parameter" for bad request.

### Testing
1. Rubygems.org code isn't completed yet. You can test this using this branch https://github.com/Shopify/rubygems.org/pull/130. Make sure to setup a user with TOTP and Webauthn.
2. Run `RUBYGEMS_HOST=http://localhost:3000 ruby -Ilib bin/gem signin`
3. Try the same with gem push, yank and owner -a/-r

### Remaining TODOs:

- [x] Add access control headers for 40x responses
- [x] Look into JRuby failing test


indirect:
I like this strategy! Thanks for continuing to iterate on this. 🎉

lib/rubygems/webauthn_listener.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+require_relative "webauthn_listener/response/response_ok"
+require_relative "webauthn_listener/response/response_no_content"
+require_relative "webauthn_listener/response/response_bad_request"
+require_relative "webauthn_listener/response/response_not_found"
+require_relative "webauthn_listener/response/response_method_not_allowed"
+
+##
+# The WebauthnListener class retrieves an OTP after a user successfully WebAuthns with the Gem host.
+# An instance opens a socket from the provided server and listens for a request from the Gem host.
```
jchestershopify, line 10: What does "provided server" mean in this context?

lib/rubygems/webauthn_listener/response.rb
```diff
@@ -0,0 +1,75 @@
...
+#   - ResponseNotFound
+#   - ResponseMethodNotAllowed
+#
+# Example:
+#   socket = TCPSocket.new(host, port)
+#   Gem::WebauthnListener::ResponseOk.send(socket, host)
+#
+
+class Gem::WebauthnListener
+  class Response
```
jchestershopify, line 21: Should we use the [`HTTPResponseFactory`](https://github.com/rubygems/rubygems/blob/1424c26861c30719afd7519db65cda9f75d4d7fa/test/rubygems/utilities.rb#L166-L187) somehow?

jchestershopify (commented):
Generally LGTM. Some questions.

lib/rubygems/webauthn_listener/response.rb
jenshenny: Oh I think we can... instead of building a raw string response, it might be better to create a HTTPResponse object and use that to send to the Gem host 👍  I'll try it out

lib/rubygems/webauthn_listener.rb
jenshenny: The TCPServer that is passed in to [`fetch_otp_from_server` ](https://github.com/rubygems/rubygems/pull/6389/files/0cdc4165ca85b15949c43d4912f05af250145ac8#diff-c67e7c6f9528cf72a398423b241dd30bd42b696ef1e8d326311abf2e89e2c2d1R39). I'll update the docs to specify the type of server that should be accepted.

lib/rubygems/webauthn_listener/response.rb
```diff
@@ -0,0 +1,160 @@
...
+
+    private
+
+    def code
+      raise NotImplementedError
+    end
+
+    def reason_phrase
+      raise NotImplementedError
+    end
```
jchestershopify, line 56: I think a comment that these are intended to be overridden in child classes would be helpful. I was very confused by them until I read further down.

jchestershopify (commented):
One suggestion, but LGTM!

test/rubygems/utilities.rb
```diff
@@ -186,6 +186,41 @@ def self.create(body:, code:, msg:, headers: {})
...
+#
+# Example:
+#
+#   # Sends a get request to http://localhost:5678
+#   Gem::FakeBrowser.get URI("http://localhost:5678")
+#
+# See RubyGems' tests for more examples of FakeBrowser.
+#
+
+class Gem::FakeBrowser
```
simi, line 201: What about `TestBrowser` or `MockBrowser`?

simi (approved):
Looks good!

test/rubygems/utilities.rb
jenshenny: I took `FakeBrowser` from `FakeFetcher` above. I do think `MockBrowser` is a better name so I changed it to that 👍

jenshenny:
@simi @indirect could one of you merge this PR if there isn't anything blocking? I don't have merge access 😞

indirect:
Sounds good!

PR #6390, opened 2023-02-17
Add TarReader::Entry#seek to seek within the tar file entry

martinemde:

Supersedes #6388, which can be closed if this solution is preferred.

`TarReader::Entry` partially implements some of the `IO` methods necessary for wrapping in `Zlib::GzipReader`. However, if you try to `#rewind` the gzip_io or `#seek` within the tar that is yielded by it, you get `NoMethodError`.

`Gem::Package#open_tar_gz` is used in many places to read the data.tar.gz contained within a .gem file. On a single pass, this normally works and is obviously working fine for `Gem::Package`. 

This PR only adds features and does not fix any bugs that would be affecting rubygems currently. It would allow TarReader to be used more generally to seek within a gem's files in memory. 

This PR combines some of the existing seeking code in TarReader#each and implements a full version of `IO#seek` in `TarReader::Entry`.

Closing an entry now seeks to the end of the entry within the tar file io stream, including skipping the padding. This is required because, for example, Zlib::GzipReader closes the data.tar.gz TarReader::Entry when done reading. TarReader#each may need to continue. Now "closing" an Entry (which is a sub-io of the main tar `io`) means "move to io position to the very end of this entry and then prevent altering the io."

TarReader#each is simplified, allowing it to just open an entry and then close it.

Uses seek for `Entry#rewind` which fixes `Zlib::GzipReader.wrap(<TarReader::Entry>)` which could not `#rewind` previously. 

### DIFF
--- a/lib/rubygems/package/tar_reader.rb
+++ b/lib/rubygems/package/tar_reader.rb
@@ -53,39 +53,11 @@ class Gem::Package::TarReader
   def each
     return enum_for __method__ unless block_given?
 
-    use_seek = @io.respond_to?(:seek)
-
     until @io.eof? do
       header = Gem::Package::TarHeader.from @io
       return if header.empty?
-
       entry = Gem::Package::TarReader::Entry.new header, @io
-      size = entry.header.size
-
       yield entry
-
-      skip = (512 - (size % 512)) % 512
-      pending = size - entry.bytes_read
-
-      if use_seek
-        begin
-          # avoid reading if the @io supports seeking
-          @io.seek pending, IO::SEEK_CUR
-          pending = 0
-        rescue Errno::EINVAL
-        end
-      end
-
-      # if seeking isn't supported or failed
-      while pending > 0 do
-        bytes_read = @io.read([pending, 4096].min).size
-        raise UnexpectedEOF if @io.eof?
-        pending -= bytes_read
-      end
-
-      @io.read skip # discard trailing zeros
-
-      # make sure nobody can use #read, #getc or #rewind anymore
       entry.close
     end
   end
--- a/lib/rubygems/package/tar_reader/entry.rb
+++ b/lib/rubygems/package/tar_reader/entry.rb
@@ -9,6 +9,20 @@
 
 class Gem::Package::TarReader::Entry
   ##
+  # Creates a new tar entry for +header+ that will be read from +io+
+  # If a block is given, the entry is yielded and then closed.
+
+  def self.open(header, io, &block)
+    entry = new header, io
+    return entry unless block_given?
+    begin
+      yield entry
+    ensure
+      entry.close
+    end
+  end
+
+  ##
   # Header for this tar entry
 
   attr_reader :header
@@ -21,6 +35,7 @@ class Gem::Package::TarReader::Entry
     @header = header
     @io = io
     @orig_pos = @io.pos
+    @end_pos = @orig_pos + @header.size
     @read = 0
   end
 
@@ -39,7 +54,14 @@ class Gem::Package::TarReader::Entry
   # Closes the tar entry
 
   def close
+    return if closed?
+    # Seek to the end of the entry if it wasn't fully read
+    seek(0, IO::SEEK_END)
+    # discard trailing zeros
+    skip = (512 - (@header.size % 512)) % 512
+    @io.read(skip)
     @closed = true
+    nil
   end
 
   ##
@@ -117,6 +139,14 @@ class Gem::Package::TarReader::Entry
     bytes_read
   end
 
+  ##
+  # Seek to the position in the tar entry
+
+  def pos=(new_pos)
+    seek(new_pos, IO::SEEK_SET)
+    new_pos
+  end
+
   def size
     @header.size
   end
@@ -158,12 +188,61 @@ class Gem::Package::TarReader::Entry
   end
 
   ##
+  # Seeks to +offset+ bytes into the tar file entry
+  # +whence+ can be IO::SEEK_SET, IO::SEEK_CUR, or IO::SEEK_END
+
+  def seek(offset, whence = IO::SEEK_SET)
+    check_closed
+
+    new_pos =
+      case whence
+      when IO::SEEK_SET then @orig_pos + offset
+      when IO::SEEK_CUR then @io.pos + offset
+      when IO::SEEK_END then @end_pos + offset
+      else
+        raise ArgumentError, "invalid whence"
+      end
+
+    if new_pos < @orig_pos
+      new_pos = @orig_pos
+    elsif new_pos > @end_pos
+      new_pos = @end_pos
+    end
+
+    pending = new_pos - @io.pos
+
+    if @io.respond_to?(:seek)
+      begin
+        # avoid reading if the @io supports seeking
+        @io.seek new_pos, IO::SEEK_SET
+        pending = 0
+      rescue Errno::EINVAL
+      end
+    end
+
+    # if seeking isn't supported or failed
+    # negative seek requires that we rewind and read
+    if pending < 0
+      @io.rewind
+      pending = new_pos
+    end
+
+    while pending > 0 do
+      size_read = @io.read([pending, 4096].min).size
+      raise UnexpectedEOF if @io.eof?
+      pending -= size_read
+    end
+
+    @read = @io.pos - @orig_pos
+
+    0
+  end
+
+  ##
   # Rewinds to the beginning of the tar file entry
 
   def rewind
     check_closed
-
-    @io.pos = @orig_pos
-    @read = 0
+    seek(0, IO::SEEK_SET)
   end
 end
--- a/test/rubygems/test_gem_package_tar_reader_entry.rb
+++ b/test/rubygems/test_gem_package_tar_reader_entry.rb
@@ -10,8 +10,7 @@ class TestGemPackageTarReaderEntry < Gem::Package::TarTestCase
 
     @tar = String.new
     @tar << tar_file_header("lib/foo", "", 0, @contents.size, Time.now)
-    @tar << @contents
-    @tar << "\0" * (512 - (@tar.size % 512))
+    @tar << tar_file_contents(@contents)
 
     @entry = util_entry @tar
   end
@@ -21,8 +20,35 @@ class TestGemPackageTarReaderEntry < Gem::Package::TarTestCase
     super
   end
 
-  def close_util_entry(entry)
-    entry.instance_variable_get(:@io).close!
+  def test_open
+    io = TempIO.new @tar
+    header = Gem::Package::TarHeader.from io
+    retval = Gem::Package::TarReader::Entry.open header, io do |entry|
+      entry.getc
+    end
+    assert_equal "a", retval
+    assert_equal @tar.size, io.pos, "should have read to end of entry"
+  end
+
+  def test_open_closes_entry
+    io = TempIO.new @tar
+    header = Gem::Package::TarHeader.from io
+    entry = nil
+    Gem::Package::TarReader::Entry.open header, io do |e|
+      entry = e
+    end
+    assert entry.closed?
+    assert_raise(IOError) { entry.getc }
+  end
+
+  def test_open_returns_entry
+    io = TempIO.new @tar
+    header = Gem::Package::TarHeader.from io
+    entry = Gem::Package::TarReader::Entry.open header, io
+    refute entry.closed?
+    assert_equal ?a, entry.getc
+    assert_nil entry.close
+    assert entry.closed?
   end
 
   def test_bytes_read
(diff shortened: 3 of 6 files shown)


lib/rubygems/package/tar_reader/entry.rb
```diff
@@ -117,6 +138,13 @@ def pos
     bytes_read
   end
 
+  ##
+  # Seek to the position in the tar entry
+
+  def pos=(new_pos)
+    seek(new_pos, IO::SEEK_SET)
```
simi, line 145: Sholdn't this return `new_pos` similar `IO#seek` return value?

lib/rubygems/package/tar_reader/entry.rb
```diff
@@ -39,6 +54,12 @@ def bytes_read
   # Closes the tar entry
 
   def close
+    return true if closed?
```
simi, line 57: :thinking: `close` should return `nil`, but since it already returns `true`, IMHO it is ok to keep it as is. I can open additional PR to inline this with `IO#close` later if needed.

lib/rubygems/package/tar_reader/entry.rb
martinemde: That seems right for `#pos=`. I couldn't get seek to return anything but 0 no matter what I did.

lib/rubygems/package/tar_reader/entry.rb
martinemde: Makes sense. I preserved the old behavior which would have returned true. Makes more sense to behave like IO#close. I'll update it here.

lib/rubygems/package/tar_reader/entry.rb
simi: Clear, `seek` returns 0 constantly as expected. I was thinking about making it explicit to keep it clear, so later if `pos=` could be potentially changed, `return new_pos` will not depend on the of implementation. Anyway I'm ok to keep it as is if that seems simpler for you @martinemde.

martinemde: For reference: 
```ruby
irb(main):002:0> f = File.open('README.md')
=> #<File:README.md>
irb(main):003:0> f.seek(10)
=> 0
irb(main):004:0> f.pos
=> 10
irb(main):005:0> f.pos = 20
=> 20
irb(main):006:0> f.close
=> nil
```

simi:
Looks good to me! Can you please squash commits @martinemde?

simi:
@deivid-rodriguez would you mind to take a look as well? This feature is partially extracted from https://github.com/rubygems/rubygems.org/pull/3454 implementing https://github.com/rubygems/rfcs/pull/44. :pray:

simi (approved):

lib/rubygems/package/tar_reader/entry.rb
```diff
@@ -157,13 +187,62 @@ def readpartial(maxlen = nil, outbuf = "".b)
     outbuf
   end
 
+  ##
+  # Seeks to +offset+ bytes into the tar file entry
+  # +whence+ can be IO::SEEK_SET, IO::SEEK_CUR, or IO::SEEK_END
```
deivid-rodriguez, line 192: Nice word, "whence", I didn't know it :)

lib/rubygems/package/tar_reader/entry.rb
```diff
@@ -157,13 +187,62 @@ def readpartial(maxlen = nil, outbuf = "".b)
...
+
+  def seek(offset, whence = IO::SEEK_SET)
+    check_closed
+
+    new_pos =
+      case whence
+      when IO::SEEK_SET then @orig_pos + offset
+      when IO::SEEK_CUR then @io.pos + offset
+      when IO::SEEK_END then @orig_pos + @header.size + offset
```
deivid-rodriguez, line 201: Maybe I would extract `@end_pos = @orig_pos + @header.size` in `initialize`, just to bring nice symmetry to this logic.

test/rubygems/test_gem_package_tar_reader_entry.rb
```diff
@@ -177,6 +207,22 @@ def test_rewind
     assert_equal char, @entry.getc
   end
 
+  def test_seek
+    @entry
```
deivid-rodriguez, line 211: Seems unneeded?

test/rubygems/test_gem_package_tar_reader_entry.rb
```diff
@@ -211,4 +257,67 @@ def test_zero_byte_file_readpartial
...
+      assert_equal contents2, entry.read, "second read after rewind should read same contents"
+    end
+  end
+
+  def test_seek_in_gzip_io_with_non_zero_offset
+    skip = @tar.size
+    contents2 = ("0".."9").to_a.join * 100
+    @tar << tar_file_header("lib/bar", "", 0, contents2.size, Time.now)
+    @tar << contents2
+    @tar << "\0" * (512 - (@tar.size % 512))
```
deivid-rodriguez, line 299: Saw this logic in a few places, we could move it to a `util_tar_contents` method or something like that.

deivid-rodriguez (commented):
I added a few small comments, but this looks really nice & solid, and the server feature it enables is cool :)

lib/rubygems/package/tar_reader/entry.rb
simi: TIL as well

lib/rubygems/package/tar_reader.rb
```diff
@@ -98,8 +70,16 @@ def each
   def rewind
     if @init_pos == 0
       @io.rewind
-    else
+    elsif @io.respond_to? :pos=
```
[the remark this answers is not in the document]
martinemde: Oh good catch for tests. I'll add a test.

The issue is when reading the data.tar.gz with TarReader, it calls `#pos=` on the gzip_io which doesn't support `#pos=` and can only rewind and read.

lib/rubygems/package/tar_reader.rb
```diff
@@ -117,6 +97,8 @@ def seek(name) # :yields: entry
 
     return yield found
   ensure
+    # we must rewind because #each is interrupted by #find
+    # which leaves the io at the wrong position.
```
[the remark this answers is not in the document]
martinemde: I think maybe my comment is not functional. I think I just left it here so I could explain this problem:

Ideally `TarReader#each` would close each entry after yielding it. In fact, I wrote the code to do exactly this at first, using `Entry.open` to auto-close the entry when iteration is done.. This makes sense because as soon as each entry is out of scope, any modification to that entry can alter the position of the underlying shared IO. 

The problem is that TarReader mixes in Enumerable, and some Enumerable methods _return_ one or many of the objects in the enumerable. This would be like returning the `file`  inside a `File.open` block. The Entry _should_ be closed once the block exits. The structure of `TarReader#seek` contradicts this, since the entry lives longer than the block, so `#each` cannot close the block. The `#rewind` after seek is a fix for the io position being indeterminant after the method call.

If I could, I would remove Enumerable from TarReader and add back some of the methods in a way that allows each to close entries when the position moves beyond. TarReader is more like a stream than an enumerable.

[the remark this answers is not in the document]
martinemde: I think I failed to explain why exactly this method breaks this. Find `returns` from within the iteration, halting it part way through. If `#each` were to `ensure; entry.close` then `TarReader.seek` would be broken. Using `Enumerable#find` here exits the iteration, preventing the call to entry.close because it's not `ensure`d.

lib/rubygems/package/tar_reader/entry.rb
martinemde: I have to give credit to GitHub Co-Pilot which filled in this word for me. I wouldn't have used such fancy wording without its help. 😆

lib/rubygems/package/tar_reader/entry.rb
martinemde: nice, I'll do that.

test/rubygems/test_gem_package_tar_reader_entry.rb
martinemde: oops, thank you.

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
martinemde: Ok, I had intended to test it here but my test case doesn't trigger it: [test_seek_in_gem_data](https://github.com/rubygems/rubygems/pull/6390/files#diff-a59526deeebec23c8cd531e267ef8fd6b9f2d7116be838d00a49d51938484dedR114-R121)

I literally can't trigger it any more, so I'm going to revert this portion of the PR.

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
martinemde: tl;dr I may follow up with another PR that tries to address this issue. For now I'm ignoring it and I removed the extra comment.

test/rubygems/test_gem_package_tar_reader_entry.rb
martinemde: Updated. Looks a lot better. Thanks!

martinemde:
Updated based on feedback.

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
deivid-rodriguez: I'm not clear whether the plan is to add a test, or revert this portion of the PR? 🤔

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
deivid-rodriguez: Like with the previous comment, I think you may have marked this as resolved, but forgot to push the changes actually resolving it?

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
martinemde: I reverted this part.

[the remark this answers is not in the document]
martinemde: Ack, apparently I didn't. Oops.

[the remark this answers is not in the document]
martinemde: Ok, fixed and pushed. I swear I reverted this when I pushed the changes, but I lost the change somehow. I should be back to normal now. The failure case is where the tar reader is created from an IO that is not at the beginning, and also doesn't support `pos=` which is very much an edge case.

lib/rubygems/package/tar_reader.rb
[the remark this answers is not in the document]
deivid-rodriguez: @martinemde Just double checking whether you actually planned to revert the extra comment and also somehow lost that change or whether you changed your mind and decided to keep it.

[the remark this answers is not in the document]
martinemde: I think I just reverted the changes to this file before committing. Thanks for catching these.

simi:
:information_source: @deivid-rodriguez I have removed auto-merge since there is pending squash.

deivid-rodriguez:
Oh, sorry, forgot that there was still the squash to be done. Thanks for catching it!

simi:
@martinemde CI seems happy, awaiting squash.

lib/rubygems/package/tar_reader.rb
```diff
@@ -53,39 +53,11 @@ def close
...
-      while pending > 0 do
-        bytes_read = @io.read([pending, 4096].min).size
-        raise UnexpectedEOF if @io.eof?
-        pending -= bytes_read
-      end
-
-      @io.read skip # discard trailing zeros
-
-      # make sure nobody can use #read, #getc or #rewind anymore
       entry.close
```
segiddins, line 61: should this be in an `begin; yield entry; ensure; entry.close; end` block?

lib/rubygems/package/tar_reader/entry.rb
```diff
@@ -8,6 +8,21 @@
...
 class Gem::Package::TarReader::Entry
+  ##
+  # Creates a new tar entry for +header+ that will be read from +io+
+  # If a block is given, the entry is yielded and then closed.
+
+  def self.open(header, io, &block)
+    entry = new header, io
+    return entry unless block_given?
+    begin
+      yield entry
```
segiddins, line 19: for consistency with `IO.open`, should the return value of the block be returned from `open` ?

lib/rubygems/package/tar_reader.rb
martinemde: It can't without making a breaking change because of the way each is exited during the iteration by seek (and because Enumerable is included). If I could make a breaking change I would ensure close.

lib/rubygems/package/tar_reader/entry.rb
martinemde: Hmm, I think I was imitating File.open here but I'm not certain now. The nil is clearly intentional but I don't remember why for sure. I can check tomorrow.

martinemde: You're completely right. Updated and added a test.

lib/rubygems/package/tar_reader.rb
martinemde: I think the solution is to remove Enumerable from TarReader since most of enumerable would produce broken already-closed entries. That's a notable breaking change on a minor part of rubygems though. Bumping a major or even minor version for it seems overzealous and leaving it as is probably won't hurt anyone.

deivid-rodriguez (approved):

deivid-rodriguez:
Everything should be addressed now, so let me pull this in! Thank you @martinemde!

PR #6391, opened 2023-02-18
Alias CurrentRuby#mswin?, mswin64?, mingw?, x64_mingw? to #windows?

johnnyshields:
This is a redux of https://github.com/rubygems/rubygems/pull/5883. See that PR for discussion thread.

This PR makes Bundler's platform values `:mswin, :mswin64, :mingw, :x64_mingw` equivalent to the recently added `:windows`. In other words, specifying any of these platform values will now bundle the specified gem on **any** Windows system, which is generally the desired behavior.

*(There is no practical use case where I would want to Bundle a gem such as Nokogiri only on `:mswin`, but not on `:mingw`.)*

**This PR does not affect how gems are built.** Binary gems can/should still be built to target specific Windows architectures such as `x64-mingw-ucrt`.

For comparison, this change makes Windows platforms much closer to how other platforms e.g. `:ruby`, `:jruby` behave today, where a catch-all platform is used in the Gemfile rather than specifying different gems for different Mac/Linux versions.

This PR also updates the manpage to indicate that users should be using `windows` in place of the others.

Lastly, this PR does not add any deprecation warnings when bundling.

### DIFF
--- a/bundler/spec/install/gemfile/platform_spec.rb
+++ b/bundler/spec/install/gemfile/platform_spec.rb
@@ -550,7 +550,7 @@ RSpec.describe "bundle install with platform conditionals" do
     gemfile <<-G
       source "#{file_uri_for(gem_repo1)}"
 
-      gem "rack", :platform => [:windows, :mingw, :mswin, :x64_mingw, :jruby]
+      gem "rack", :platform => [:windows, :mswin, :mswin64, :mingw, :x64_mingw, :jruby]
     G
 
     bundle "install"
--- a/bundler/spec/runtime/platform_spec.rb
+++ b/bundler/spec/runtime/platform_spec.rb
@@ -365,7 +365,7 @@ RSpec.describe "Bundler.setup with multi platform stuff" do
     simulate_windows do
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
-        gem "nokogiri", :platforms => [:mingw, :mswin, :x64_mingw, :jruby]
+        gem "nokogiri", :platforms => [:windows, :mswin, :mswin64, :mingw, :x64_mingw, :jruby]
         gem "platform_specific"
       G
 
--- a/bundler/lib/bundler/current_ruby.rb
+++ b/bundler/lib/bundler/current_ruby.rb
@@ -71,10 +71,10 @@ module Bundler
     def windows?
       Gem.win_platform?
     end
-    alias mswin? windows?
-    alias mswin64? windows?
-    alias mingw? windows?
-    alias x64_mingw? windows?
+    alias_method :mswin?, :windows?
+    alias_method :mswin64?, :windows?
+    alias_method :mingw?, :windows?
+    alias_method :x64_mingw?, :windows?
 
     (KNOWN_MINOR_VERSIONS + KNOWN_MAJOR_VERSIONS).each do |version|
       trimmed_version = version.tr(".", "")
--- a/bundler/lib/bundler/current_ruby.rb
+++ b/bundler/lib/bundler/current_ruby.rb
@@ -71,26 +71,10 @@ module Bundler
     def windows?
       Gem.win_platform?
     end
-
-    def mswin?
-      # For backwards compatibility
-      windows?
-
-      # TODO: This should correctly be:
-      # windows? && Bundler.local_platform != Gem::Platform::RUBY && Bundler.local_platform.os == "mswin32" && Bundler.local_platform.cpu == "x86"
-    end
-
-    def mswin64?
-      windows? && Bundler.local_platform != Gem::Platform::RUBY && Bundler.local_platform.os == "mswin64" && Bundler.local_platform.cpu == "x64"
-    end
-
-    def mingw?
-      windows? && Bundler.local_platform != Gem::Platform::RUBY && Bundler.local_platform.os == "mingw32" && Bundler.local_platform.cpu != "x64"
-    end
-
-    def x64_mingw?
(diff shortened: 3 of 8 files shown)


simi (approved):
Code-wise looks good to me. I think we should wait for reviews of @deivid-rodriguez @indirect and @hsbt as well before merging. :pray:

indirect:
This seems good to me. I would love to improve rubygems to the point where this kind of “solution” isn’t needed, but this is an improvement on what we have now. 👍🏻

mohits:
Hi! I was poking around in the rubygems source code for a post on native gems that I am writing, and I was reminded of this issue that I had come across earlier.

https://github.com/rubygems/rubygems/blob/0f3d7db4580d839ec7f142adc0d5e0a681eee0a1/lib/rubygems/ext/builder.rb#L29 has this code (lib/rubygems/ext/builder.rb):

```ruby
      make_program_name = (RUBY_PLATFORM.include?("mswin")) ? "nmake" : "make"
```

Would this be impacted with the change being proposed here?

For my setup with RubyInstaller, I have:
* RUBY_PLATFORM = "x64-mingw-ucrt"
* nmake = not recognised
* where make => C:\Ruby31-x64\msys64\usr\bin\make.exe

Sorry for the noise if I misunderstood.

indirect:
Thank you for speaking up! If I am understanding this PR correctly, it will not change the underlying RUBY_PLATFORM, only how Bundler allows users to request gems inside the gemfile while on any Windows version of Ruby.

johnnyshields:
@mohits what @indirect has said is correct. This PR only really affects the following in your `Gemfile`:

```ruby
platforms :mswin do
  gem 'some_gem'
end
```

This syntax will now bundle the gem on any mswin/mingw Windows platform.

Your `RUBY_PLATFORM` value will not change, and you can still use it to make fine-grained distinctions between `mswin64`, `mingw-ucrt-x64` etc.

By way of analogy, `platform :windows` is now a lot more like `platform :ruby`, however, you can still use `RUBY_PLATFORM` to make distinction between different flavors of Mac, Linux, etc.

mohits:
Thanks for the clarification @johnnyshields and @indirect.

bundler/spec/install/gemfile/platform_spec.rb
```diff
@@ -550,7 +550,7 @@
     gemfile <<-G
       source "#{file_uri_for(gem_repo1)}"
 
-      gem "rack", :platform => [:windows, :mingw, :mswin, :x64_mingw, :jruby]
+      gem "rack", :platform => [:windows, :mswin, :mswin64, :mingw, :x64_mingw, :jruby]
```
deivid-rodriguez, line 553: What's this change for?

bundler/spec/runtime/platform_spec.rb
```diff
@@ -365,7 +365,7 @@
     simulate_windows do
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
-        gem "nokogiri", :platforms => [:mingw, :mswin, :x64_mingw, :jruby]
+        gem "nokogiri", :platforms => [:windows, :mswin, :mswin64, :mingw, :x64_mingw, :jruby]
```
deivid-rodriguez, line 368: What's this change for?

deivid-rodriguez (commented):
I'm good with this, and happy to merge now.

I'd wait for the next minor or major to release it though, since this is not a bug fix nor new feature, just code cleanup, and it's technically a breaking change in case someone is somehow relying on this distinction.

bundler/spec/install/gemfile/platform_spec.rb
johnnyshields: `:mswin64` looked like it was unintentionally missing here, as all other windows platforms were there.

bundler/spec/runtime/platform_spec.rb
johnnyshields: Same reason as https://github.com/rubygems/rubygems/pull/6391#discussion_r1120877604

johnnyshields:
@deivid-rodriguez I think this is fine to go on a minor release. It is very unlikely to affect any windows users in practice.

bundler/spec/install/gemfile/platform_spec.rb
johnnyshields: Hmm I think these should still be left in the spec, as we want to ensure they all still work even though they are considered deprecated.

deivid-rodriguez: But the spec seems completely unrelated to Windows

> it "does not print a warning when a dependency is unused on a platform different from the current one"

That's why I'd either leave only `:windows` there to somehow link the change to this PR, or leave this spec completely untouched.

Anyways, it's no big deal really.

johnnyshields: Meh, lets do the way I have in this PR. Err on the side of more coverage.

deivid-rodriguez: I don't see this as more coverage, but I'm not going to fight over this. Approving.

deivid-rodriguez (approved):

deivid-rodriguez:
Let's just wait for a final approval from @hsbt here, since he expressed some concerns over at #5883.

hsbt (approved):
I still have concern about `mswin` and `mingw` differences.

But we also don't handle `platform` with various UNIX platforms like `linux`, `freebsd` and others. I'm okay to mix `mswin` and `mingw` to `windows` platform. 

We can reraise my concerns when anyone have new issue.

PR #6394, opened 2023-02-19
[DOC] Remove (newly unneeded) remarks about aliases

BurdetteLamar:
Rdoc now notes aliases.

### DIFF
--- a/bundler/lib/bundler/vendor/fileutils/lib/fileutils.rb
+++ b/bundler/lib/bundler/vendor/fileutils/lib/fileutils.rb
@@ -192,8 +192,6 @@ module Bundler::FileUtils
   #
   #   Bundler::FileUtils.pwd # => "/rdoc/fileutils"
   #
-  # Bundler::FileUtils.getwd is an alias for Bundler::FileUtils.pwd.
-  #
   # Related: Bundler::FileUtils.cd.
   #
   def pwd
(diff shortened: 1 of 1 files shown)


simi:
:wave: @BurdetteLamar. Thanks for the PR. Is there any reason to pull in only this change? `fileutils` are vendored and could be updated using `rake vendor:fileutils`. It will pull in version specified in Rake file (currently `v1.7.0`). For example this change is also missing https://github.com/ruby/fileutils/commit/8af16ba57c4e68a7ee40e665c4b09fadbf9acc5e compared to current fileutils master.

Would it make sense to wait for new release of fileutils and then pull in all changes?

BurdetteLamar:
> 👋 @BurdetteLamar. Thanks for the PR. Is there any reason to pull in only this change? `fileutils` are vendored and could be updated using `rake vendor:fileutils`. It will pull in version specified in Rake file (currently `v1.7.0`). For example this change is also missing [ruby/fileutils@8af16ba](https://github.com/ruby/fileutils/commit/8af16ba57c4e68a7ee40e665c4b09fadbf9acc5e) compared to current fileutils master.
> 
> Would it make sense to wait for new release of fileutils and then pull in all changes?

Agnostic on that.  Whatever y'all think best.

deivid-rodriguez:
Unless there's an immediate benefit of pulling from master now (which we have done many times and it would be fine too), I think we can update this PR next time fileutils is released and pull everything.

Thanks for this @BurdetteLamar, by the way! :)

BurdetteLamar:
> Unless there's an immediate benefit of pulling from master now (which we have done many times and it would be fine too), I think we can update this PR next time fileutils is released and pull everything.
> 
> Thanks for this @BurdetteLamar, by the way! :)

Works for me!

PR #6399, opened 2023-02-21
Auto-heal on corrupted lockfile with missing deps

composerinteralia:
## What was the end-user or developer problem that led to this PR?

See https://github.com/rubygems/rubygems/pull/6355 for the original problem.

## What is your fix for the problem, implemented in this PR?

https://github.com/rubygems/rubygems/pull/6355 turned a crash into a nicer error. This commit auto-heals the corrupt lockfile instead.

Since in this particular case (a corrupt Gemfile.lock with missing dependencies) the LazySpecification will not have accurate dependency information, we have to materialize the SpecSet to determine there are missing dependencies. Since we've already got some incomplete-spec-auto-healing here, I've treated this as another brand of incomplete spec.

## Make sure the following tasks are checked

- [X] Describe the problem / feature
- [X] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [X] Write code to solve the problem
- [X] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6400, opened 2023-02-22
Auto-heal on corrupted lockfile with missing deps

composerinteralia:

See https://github.com/rubygems/rubygems/pull/6355 for the original problem.

https://github.com/rubygems/rubygems/pull/6355 turned a crash into a nicer error. This commit auto-heals the corrupt lockfile instead.

In this particular case (a corrupt Gemfile.lock with missing dependencies) the LazySpecification will not have accurate dependency information, so we have to materialize the SpecSet to determine there are missing dependencies. We've already got a way to handle this, via `SpecSet#incomplete_specs`, but it wasn't quite working for this case because we'd get to `@incomplete_specs += lookup[name]` and `lookup[name]` would be empty for the dependency.

With this commit we catch it a bit earlier, marking the spec containing the missing dependency as incomplete.

(I had originally opened https://github.com/rubygems/rubygems/pull/6399, but misunderstood a few things there. I closed it and then force pushed to the branch, preventing me from reopening it.)

### DIFF
--- a/bundler/lib/bundler/installer/parallel_installer.rb
+++ b/bundler/lib/bundler/installer/parallel_installer.rb
@@ -47,13 +47,6 @@ module Bundler
         dependencies.all? {|d| installed_specs.include? d.name }
       end
 
-      # Check whether spec's dependencies are missing, which can indicate a
-      # corrupted lockfile
-      def dependencies_missing?(all_specs)
-        spec_names = all_specs.map(&:name)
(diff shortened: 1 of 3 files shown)


composerinteralia:
🤦🏻‍♂️ Still failing. I'll take a look, but feel free to tell me if my general approach is nonsense.

deivid-rodriguez (approved):
Looks great! Thanks for the nice follow up!

simi (approved):

Issue #6280, opened 2023-01-17
NoMethodError: undefined method `metadata' for nil:NilClass when running `bundle install`


quiro91:
### Describe the problem as clearly as you can

Running `bundle install` does not work as expected after upgrading to the latest bundler version (previous was 2.1.4). This is the output I get when running `bundle install`:

```
NoMethodError: undefined method `metadata' for nil:NilClass
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli/common.rb:23:in `block in output_fund_metadata_summary'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli/common.rb:23:in `count'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli/common.rb:23:in `output_fund_metadata_summary'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli/update.rb:119:in `run'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli.rb:310:in `block in update'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/settings.rb:131:in `temporary'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/cli.rb:309:in `update'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
  /Users/lorenzo.quiroli/.rbenv/versions/3.2.0/lib/ruby/gems/3.2.0/gems/bundler-2.4.4/lib/bundler/vendor/thor/lib/thor/invocation.rb:127:in `invoke_command'
[...]

PR #6409, opened 2023-02-27
Avoid NoMethodError with malformed Gemfile.lock.

duckinator:

If a Gemfile.lock has a mismatch between `specs:` and `DEPENDENCIES`, printing funding information runs what boils down to `nil.metadata`.

Before printing funding information, make sure there's something to print.

### DIFF
--- a/bundler/lib/bundler/cli/common.rb
+++ b/bundler/lib/bundler/cli/common.rb
@@ -20,7 +20,7 @@ module Bundler
       current_dependencies = definition.requested_dependencies
       current_specs = definition.specs
 
-      count = current_dependencies.count {|dep| current_specs[dep.name].first.metadata.key?("funding_uri") }
+      count = current_dependencies.count {|dep| current_specs[dep.name]&.first&.metadata&.key?("funding_uri") }
 
       return if count.zero?
 
--- a/bundler/spec/commands/install_spec.rb
+++ b/bundler/spec/commands/install_spec.rb
@@ -1103,4 +1103,39 @@ RSpec.describe "bundle install with gem sources" do
       expect(err).to include("Could not find compatible versions")
     end
   end
+
+  context "with a lockfile that has mismatched 'specs:' and 'DEPENDENCIES' sections" do
+    it "does not throw a NoMethodError" do
+      build_repo4 do
+        build_gem "indirect_dependency", "1.2.3" do |s|
+          s.metadata["funding_uri"] = "https://example.com/donate"
+        end
+
+        build_gem "direct_dependency", "4.5.6" do |s|
+          s.add_dependency "indirect_dependency", ">= 0"
+        end
+      end
+
+      lockfile <<-G
+        GEM
+          remote: #{file_uri_for(gem_repo4)}/
+          specs:
+
+        PLATFORMS
+          ruby
+
+        DEPENDENCIES
+          direct_dependency
+
+        BUNDLED WITH
+           #{Bundler::VERSION}
+      G
+
+      install_gemfile <<-G
+        source "#{file_uri_for(gem_repo4)}"
+
+        gem "direct_dependency"
+      G
+    end
+  end
 end


duckinator:
I've not figured out how to write a test for this yet. However, I have confirmed the fix manually using the testcase from https://github.com/rubygems/rubygems/issues/6280#issuecomment-1409600004:

```
puppy@orthrus.fox:~/dev/rubygems/rg-6280$ ls -a
./            ../           Gemfile       Gemfile.lock
puppy@orthrus.fox:~/dev/rubygems/rg-6280$ cat Gemfile
source "https://rubygems.org"
gem "fastlane"
puppy@orthrus.fox:~/dev/rubygems/rg-6280$ cat Gemfile.lock 
GEM
  remote: https://rubygems.org/
  specs:

PLATFORMS
  ruby

DEPENDENCIES
  fastlane

BUNDLED WITH
   2.4.4
puppy@orthrus.fox:~/dev/rubygems/rg-6280$ bundle _2.5.0.dev_ install
Using bundler 2.5.0.dev
Bundle complete! 1 Gemfile dependency, 1 gem now installed.
Bundled gems are installed into `./.bundle/gems`
puppy@orthrus.fox:~/dev/rubygems/rg-6280$
```

duckinator:
Added a test. Since I added the test after the fix, I also verified that the test fails without the fix and passes with it.

duckinator:
Figured out the cause of the failing tests. Should be good now.

duckinator:
Some CI tasks are getting this error during the "Setup ruby" stage, and I'm not sure what's up with that:

```
 Error: Error: Unexpected HTTP response: 404
    at /home/runner/work/_actions/ruby/setup-ruby/d3c9825d67b0d8720afdfdde5af56c79fdb38d16/dist/index.js:8135:25
    at Generator.next (<anonymous>)
    at fulfilled (/home/runner/work/_actions/ruby/setup-ruby/d3c9825d67b0d8720afdfdde5af56c79fdb38d16/dist/index.js:8045:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

deivid-rodriguez:
@duckinator I understand that this fixes the crash, and having a test is also great. However, what it does is actually respecting the `Gemfile.lock` (empty) instead of the `Gemfile`, which I believe is not what we want because it may cause headaches to the developer where she will think everything is fine while gems in `Gemfile` are being actually ignored!

Instead, we should at least print a warning, or better, respect the `Gemfile` and regenerate the `Gemfile.lock` accordingly. I think we should improve the test to make sure that the `Gemfile.lock` is regenerate to include `direct_dependency` under SPECS, and get that to pass.

deivid-rodriguez:
I think this alternative fix should do the trick

```diff
diff --git a/bundler/lib/bundler/spec_set.rb b/bundler/lib/bundler/spec_set.rb
index 7478bd9ca2..25664e707f 100644
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -39,7 +39,7 @@ def for(dependencies, check = false, platforms = [nil])
             deps << [d, dep[1]]
           end
         elsif check
-          @incomplete_specs += lookup[name]
+          @incomplete_specs += lookup[name].any? ? lookup[name] : [LazySpecification.new(name, nil, nil)]
         end
       end
 
```

The fix conflicts with #6400, but it should be fine to adapt the fix to it, or viceversa.

duckinator:
@deivid-rodriguez your fix definitely works better than mine. It avoids the crash, but does so by regenerating `Gemfile.lock`. I'll update the test to check for an updated lockfile, and then add your fix. 👍

duckinator:
@deivid-rodriguez I tried your fix and it causes another test to fail. I made a separate PR from a different branch over at #6423, since the fix and test are both different. (Whichever PR winds up merged, the other can just be closed.)

duckinator:
I'm closing this in favor of #6423. After changing that one to be based off #6400, it works by regenerating `Gemfile.lock` (instead of ignoring the Gemfile), but doesn't break other tests.

PR #6410, opened 2023-02-27
Ensure safe load marshal is also used when loading gemspecs

segiddins:

`Bundler.safe_load_marshal` wasn't catching "unsafe" objects that were loaded from a `Gem::Specification`

My fix ensures that the string loaded in `Gem::Specification._load` goes through a safe-load method. In RubyGems, this is done directly, while in Bundler it is monkey-patched in as a preamble to the `_load` method.

### DIFF
--- a/lib/rubygems/util.rb
+++ b/lib/rubygems/util.rb
@@ -112,4 +112,50 @@ module Gem::Util
     end
   end
 
+  class DefaultKey
+  end
+
+  class PrivateType
+  end
+
+  require "date"
+  SAFE_MARSHAL_CLASSES = [
+    Array,
+    Date,
+    Float,
+    Hash,
+    Integer,
+    String,
+    Symbol,
+    Time,
+
+    TrueClass,
+    FalseClass,
+
+    NilClass,
+
+    Gem::Dependency,
+    Gem::NameTuple,
+    Gem::Platform,
+    Gem::Requirement,
+    Gem::Specification,
+    Gem::Version,
+
+    DefaultKey,
+    PrivateType,
+  ].freeze
+  SAFE_MARSHAL_ERROR = "Unexpected class %s present in marshaled data. Only %s are allowed."
+  SAFE_MARSHAL_PROC = proc do |object|
+    object.tap do
+      unless SAFE_MARSHAL_CLASSES.include?(object.class)
+        raise TypeError, format(SAFE_MARSHAL_ERROR, object.class, SAFE_MARSHAL_CLASSES.join(", "))
+      end
+    end
+  end
+
+  private_constant :SAFE_MARSHAL_CLASSES, :SAFE_MARSHAL_ERROR, :SAFE_MARSHAL_PROC
+
+  def self.safe_load_marshal(data)
+    Marshal.load(data, SAFE_MARSHAL_PROC)
+  end
 end
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "bundler/vendored_fileutils"
+require "date"
 require "pathname"
 require "rbconfig"
 
@@ -39,7 +40,31 @@ module Bundler
   environment_preserver.replace_with_backup
   SUDO_MUTEX = Thread::Mutex.new
 
-  SAFE_MARSHAL_CLASSES = [Symbol, TrueClass, String, Array, Hash, Gem::Version, Gem::Specification].freeze
+  SAFE_MARSHAL_CLASSES = [
+    Array,
+    Date,
+    Float,
+    Hash,
+    Integer,
+    String,
+    Symbol,
+    Time,
+
+    TrueClass,
+    FalseClass,
+
+    NilClass,
+
+    Gem::Dependency,
+    Gem::NameTuple,
+    Gem::Platform,
+    Gem::Requirement,
+    Gem::Specification,
+    Gem::Version,
+
+    Gem::Util::DefaultKey,
+    Gem::Util::PrivateType,
+  ].freeze
   SAFE_MARSHAL_ERROR = "Unexpected class %s present in marshaled data. Only %s are allowed."
   SAFE_MARSHAL_PROC = proc do |object|
     object.tap do
@@ -48,6 +73,8 @@ module Bundler
       end
     end
   end
+
+  private_constant :SAFE_MARSHAL_CLASSES, :SAFE_MARSHAL_ERROR, :SAFE_MARSHAL_PROC
 
   autoload :Definition,             File.expand_path("bundler/definition", __dir__)
   autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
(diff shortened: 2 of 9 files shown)


lib/rubygems/util.rb
```diff
@@ -112,4 +112,50 @@ def self.correct_for_windows_path(path)
...
+      unless SAFE_MARSHAL_CLASSES.include?(object.class)
+        raise TypeError, format(SAFE_MARSHAL_ERROR, object.class, SAFE_MARSHAL_CLASSES.join(", "))
+      end
+    end
+  end
+
+  private_constant :SAFE_MARSHAL_CLASSES, :SAFE_MARSHAL_ERROR, :SAFE_MARSHAL_PROC
+
+  def self.safe_load_marshal(data)
+    Marshal.load(data, SAFE_MARSHAL_PROC)
```
eregon, line 159: I don't think this is really safe because the Proc gets called after an object has been created.
That means its `class.allocate` has been called and also `#initialize` has been called, which could be arbitrary code and e.g. does affect `ObjectSpace.each_object` and potentially the object is even referenced by other objects already if  `#initialize` stores `self` somewhere.

lib/rubygems/util.rb
```diff
@@ -112,4 +112,50 @@ def self.correct_for_windows_path(path)
     end
   end
 
+  class DefaultKey
+  end
+
+  class PrivateType
+  end
+
+  require "date"
```
eregon, line 121: This seems a problem (RubyGems shouldn't load as few stdlibs as possible), but it could probably be worked around easily (e.g., check `obj.class.name` instead)

lib/rubygems/util.rb
indirect: Can we build a test for the kind of "unsafeness" we are trying to guard against? Is it just preventing execution of `class.allocate` and `#initialize` on classes outside the safe list?

simi:
I spend some time digging deeper into this problem. I have found out few individual problems we should solve.

# 1. unexpected unsafe `Marshal.load` call

**TL;DR -> merge the `_load` fix and onboard [Security/MarshalLoad](https://docs.rubocop.org/rubocop/cops_security.html#securitymarshalload)** 

In https://github.com/rubygems/rubygems/pull/6141 I have missed there is additional `_load` in `Gem::Specification` calling `Marshal.load`. That was partially revealed (not the cause, but the missed protection) at https://github.com/rubygems/rubygems/pull/6384#discussion_r1107915914. I have wrongly assumed the problem is different. To sum up this, it is super simple to miss those `marshal_load`/`_load` calls across Bundler and RubyGems.

To prevent this in the future, I suggest to offboard safe marshal into custom modules like `Gem::SafeMarshal` and `Bundler::SafeMarshal` in separated files and onboard [Security/MarshalLoad](https://docs.rubocop.org/rubocop/cops_security.html#securitymarshalload) rubocop rule in both codebases with exception of those two files with bold comment explaining why those two files are excluded from this list and adding new file to the list can potentially result into security problem. "Deep copy" hack used in the code at few random places is luckily allowed by this rule.

# 2. unexpected class deserialization

**TL;DR -> fixed by 1.**

This was originally addressed at https://github.com/rubygems/rubygems/pull/6141 and now fixed by fixing the `Gem::Specification#_load` special case as well. Combined with https://github.com/rubygems/rubygems/pull/6384 and RuboCop rule to prevent unexepcted `Marshal.load` it should be enough to prevent deserialization anything other than allowed in the list in the whole codebase. :pray:

# 3. unexpected instance variables deserialization

**TL;DR -> add minimal `marshal_load`/`marshal_dump` to all `Gem::*` classes in allowed list and add test to enforce this**

Even individual allowed classes could lead into unexpected state after deserialization.

```ruby
Gem::Requirement.new(['A'])
#=> /home/retro/.rubies/ruby-3.1.3/lib/ruby/site_ruby/3.1.0/rubygems/requirement.rb:106:in `parse': Illformed requirement ["A"] (Gem::Requirement::BadRequirementError)
```

There is simple validation on example `Gem::Requirement` class. This can be locally bypassed using `allocate`. `Marshal#load` would not care about this "validation" when serialized (`allocate` is actually useful to craft the payload, but understanding the Marshal format, anyone can craft the payload even manually without `allocate` at all.

```ruby
modified_requirement = Gem::Requirement.allocate.tap {|r| r.instance_variable_set(:@requirements, ['A'])}
#=> Gem::Requirement.new(["A "])
Gem::Util.safe_load_marshal(Marshal.dump(modified_requirement))
#=> Gem::Requirement.new(["A "])

# it could be also chained using all classes in the allowed list

modified_requirement = Gem::Requirement.allocate.tap {|r| r.instance_variable_set(:@requirements, [Gem::Specification.new])}
#=> Gem::Requirement.new(["#<Gem::Specification name= version=> "])
modified_requirement.requirements
#=> 
# [Gem::Specification.new do |s|                                        
#    s.name = nil                                                       
#    s.version = nil                                                    
#    s.installed_by_version = Gem::Version.new("0")                     
#    s.date = Time.utc(2023, 3, 1)                                      
#    s.require_paths = ["lib"]                                          
#    s.rubygems_version = "3.5.0.dev"                                   
#    s.specification_version = 4                                        
#    s.summary = nil                                                    
#    end]                 
```

My examples are useless and not explaining any security problem (it will just result into unexpected crashes). But similar kind of clever combination of unexpected assignments resulted into [security problem](https://github.com/rubygems/rubygems/commit/141c2f4388f0f6f81e4d420d73961dbd68f5c08f) already. And there is no simple way to be 100% safe in here. Maybe (I have no experience with this) some kind of fuzzing could to help reveal similar dangerous combinations. :thinking:.

What we can do in here IMHO is at least ensure all custom classes in the allowed list have implmenented at least `marshal_load`/`marshal_dump` methods to limit access to instance variables of related classes. Imagine some kind of improvement will land into `Gem::Dependency` (currently missing those methods) memoizing source of the dependency.

```ruby
class Gem::Dependency
  def source_url
    @source_url ||= something.to_load_source_url
  end
end
```

During code review, it would be super easy to miss this is potentially security problem, since any instance variable can be modified during loading the gemspec. And also it is just plain `String`, which is part of the allow list. This somehow combined in clever way together "in theory" can lead to unexpected problems including security issues.

I suggest to add at least tests checking all custom classes in allow list implements `marshal_load`/`marshal_dump` and add comments around to explain security importance of those methods.

---

:thinking:  I got also alternative idea (outside of the scope of this PR or even short-term solution) to build some kind of safe "marshal evaluator" (could be just bunch of minimal stubs for needed classes) and evalute gemspecs outside of the main process, turn them into some simple structure and share back to reconstruct gemspec in safe way using just basic types.

indirect:
@simi thank you for this thorough research! what you have written matches my understanding of the situation, and I think your suggested approach makes sense to remediate the problem for now.

In the long run, I hope we can switch to JSON or some other non-executable serialization format. 😅

lib/rubygems/util.rb
eregon: > Can we build a test for the kind of "unsafeness" we are trying to guard against? Is it just preventing execution of `class.allocate` and `#initialize` on classes outside the safe list?

That's exactly what Marshal cannot do and my point above.

PR #6413, opened 2023-02-27
Fix `gems.rb` lockfile for bundler version lookup in template

TonyCTHsu:

When I create a rails 7 project with `rails new my_app --minimal` , there are a couple of files under `bin` provided from bundler.  `bin/bundle` caught my attention with 

```
  def lockfile
    lockfile =
      case File.basename(gemfile)
      when "gems.rb" then gemfile.sub(/\.rb$/, gemfile)
      else "#{gemfile}.lock"
      end
    File.expand_path(lockfile)
  end
```
This line: `when "gems.rb" then gemfile.sub(/\.rb$/, gemfile)`

When provided with `gems.rb`, the lockfile ends up become `gemsgems.rb`, which is very likely that the file is missing and this [expression](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/templates/Executable.bundler#L57) is returning `false` and prevent reading bundler version from the lockfile. 

 I noticed that the filename(`Gemfile`/`gems.rb`, `Gemfile.lock`/`gems.locked`) is probably not that important, since the content is the same, but Bundler has a [specific mapping](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/bundler_version_finder.rb#L68) for `gems.rb` to `gems.locked` I think should be respected.

The fix changes the lockfile lookup for `gems.rb` during bundler version lookup. Respect the bundler version specified in `gems.locked`. 

### DIFF
--- a/bundler/lib/bundler/templates/Executable.bundler
+++ b/bundler/lib/bundler/templates/Executable.bundler
@@ -47,7 +47,7 @@ m = Module.new do
   def lockfile
     lockfile =
       case File.basename(gemfile)
-      when "gems.rb" then gemfile.sub(/\.rb$/, gemfile)
+      when "gems.rb" then gemfile.sub(/\.rb$/, ".locked")
       else "#{gemfile}.lock"
       end
     File.expand_path(lockfile)
--- a/bundler/spec/commands/binstubs_spec.rb
+++ b/bundler/spec/commands/binstubs_spec.rb
@@ -166,6 +166,21 @@ RSpec.describe "bundle binstubs <gem>" do
           end
         end
 
+        context "and the version is newer when given `gems.rb` and `gems.locked`" do
+          before do
+            gemfile bundled_app("gems.rb"), gemfile
+            lockfile bundled_app("gems.locked"), lockfile.gsub(system_bundler_version, "999.999")
+          end
+
+          it "runs the correct version of bundler" do
+            sys_exec "bin/bundle install", :env => { "BUNDLE_GEMFILE" => "gems.rb" }, :raise_on_error => false
+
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/templates/Executable.bundler
```diff
@@ -47,7 +47,7 @@ m = Module.new do
   def lockfile
     lockfile =
       case File.basename(gemfile)
-      when "gems.rb" then gemfile.sub(/\.rb$/, gemfile)
+      when "gems.rb" then gemfile.sub(/\.rb$/, ".locked")
```
TonyCTHsu, line 50: After the change, the tests all passed. But I do not really know how write a test for this. Would love to have maintainer's help on this.

simi: :thinking: This was added probably here https://github.com/rubygems/rubygems/commit/cdcda68e79ef3d6b5df78ce41066f15d7703395e#diff-fc635269f10cd2189b47bf5ae555ee4bf14882724b5871383904cd8a07de69ddR34. :older_adult: 

Ad testing, have you checked https://github.com/rubygems/rubygems/blob/master/bundler/spec/runtime/executable_spec.rb

TonyCTHsu: 👋 @simi , thanks for the tip! I have added some tests which are failing on `master` and green on this branch!

simi (approved):

deivid-rodriguez (approved):
Looks great, nice catch!

Do you mind rebasing and squashing the commits into one. I like to have the fix and the test to illustrate/exercise it in the same commit.

PR #6414, opened 2023-02-27
Add bulk formatting commits into .git-blame-ignore-revs.

simi:
I have picked few bulk ones. There are also smaller ones touching 2 or 3 files around probably not worth it to add.


hsbt (approved):

deivid-rodriguez (approved):
Never been bothered by any of these myself, and I usually don't mind skipping an occasional style commit anyways, but the `Style/StringLiterals` one seems big enough to ignore.

The others I would not add them because they're small. I draw the line around 400 or 500 net change (number of additions + number of deletions) but this is of course subjective so happy to add those too if you find that helpful.

Issue #6418, opened 2023-02-27
Gem.state_home memoizes using @data_home


nevans:
### Describe the problem as clearly as you can

While debugging an unrelated configuration issue of my own, I noticed the following:

https://github.com/rubygems/rubygems/blob/0f3d7db4580d839ec7f142adc0d5e0a681eee0a1/lib/rubygems/defaults.rb#L160-L162

It seems to me that `@data_home` here *must* be a mistake, right?  And, looking at the commit that created that method (0fbc4ace8a) and at the tests:
https://github.com/rubygems/rubygems/blob/0f3d7db4580d839ec7f142adc0d5e0a681eee0a1/test/rubygems/helper.rb#L366
both seem to confirm that `@state_home` was intended.

### Did you try upgrading RubyGems?

This is visible on the current `master` branch.

### Post steps to reproduce the problem

I attempted to use a codespace so I could submit a new failing test along with simply changing `@data_home` to `@state_home`.  After `rake setup`, `rake test` gave a number of unrelated errors.  This may be an incompatibility issue in my dotfiles.  Unfortunately, I don't have time to dive deeper into why that wasn't working right now, so I gave up on that for today.  Sorry about that... I'll try to figure out my issue before I submit any PRs, but I'm hoping this might be a simple fix for someone who already has a working dev env.

### What were you expecting to happen?

`def state_home = @state_home ||= (xdg_state_env || xdg_state_default)`

### What actually happened?

`def state_home = @data_home ||= (xdg_state_env || xdg_state_default)`

simi:
Yup, that's mistake. And what's even worse, it is my mistake at https://github.com/rubygems/rubygems/commit/0fbc4ace8a664649e76bf616895a41a86e0710f8. I'll open PR. Thanks for the info.

nevans:
I had an almost identical XDG_DATA vs XDG_CACHE copy/paste bug in one of my config files (IIRC: bashrc or vimrc or maybe both) for months before I discovered it, and cache vs data is even worse than state vs data.  So you'll receive no judgement from me! 😉

simi:
feel free to review @nevans https://github.com/rubygems/rubygems/pull/6420 :pray:

PR #6420, opened 2023-02-28
Use proper memoized var name for Gem.state_home.

simi:
~I wasn't able to find any useful way to test this, thus I decided to onboard related RuboCop rule excluding all violations for now (it is just few of them). I can update this PR if needed @deivid-rodriguez to try to update those files or I can do that later in separated PR.~

PR with Rubocop change will follow.

- fixes https://github.com/rubygems/rubygems/issues/6418

### DIFF
--- a/lib/rubygems/defaults.rb
+++ b/lib/rubygems/defaults.rb
@@ -158,7 +158,7 @@ module Gem
   # The path to standard location of the user's state directory.
 
   def self.state_home
-    @data_home ||= (ENV["XDG_STATE_HOME"] || File.join(Gem.user_home, ".local", "state"))
+    @state_home ||= (ENV["XDG_STATE_HOME"] || File.join(Gem.user_home, ".local", "state"))
   end
 
   ##
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1654,6 +1654,30 @@ class TestGem < Gem::TestCase
     ENV["SOURCE_DATE_EPOCH"] = old_epoch
   end
 
+  def test_data_home_default
+    expected = File.join(@userhome, ".local", "share")
+    assert_equal expected, Gem.data_home
+  end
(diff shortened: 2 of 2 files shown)


nevans:
@simi I like that rubocop rule.  🙂

Re: tests; Although test/helper.rb clearly resets various instance vars and env vars in setup, I didn't see any tests for the env-var-derived or defaults on state_home or data_home (etc).  Perhaps something as simple as:

```ruby
# this should pass already, before PR #6420
def test_data_home_default
  expected = File.join(@userhome, ".local", "share")
  assert_equal expected, Gem.data_home
end

def test_data_home_from_env
  ENV["XDG_DATA_HOME"] = expected = "/test/data/home"
  assert_equal expected, Gem.data_home
end

def test_state_home_default
  Gem.instance_variable_set :@state_home, nil
  Gem.data_home # memoize @data_home, to demonstrate GH-6418
  expected = File.join(@userhome, ".local", "state")
  assert_equal expected, Gem.state_home
end

def test_state_home_from_env
  Gem.instance_variable_set :@state_home, nil 
  Gem.data_home # memoize @data_home, to demonstrate GH-6418
  ENV["XDG_STATE_HOME"] = expected = "/test/state/home"
  assert_equal expected, Gem.state_home
end
```

What do you think?  Would that work?

deivid-rodriguez:
@nevans' tests look good to me! And yeah, I generally prefer fixing all offenses when introducing a cop, specially if there's only a handful of them like here.

simi:
OK, thanks @nevans for the tests. I wasn't sure similar testing makes sense, but looking at it, I don't see any problem to add it to current test suite.

@deivid-rodriguez Since the change is covered by test right now, I have removed the Rubocop rule from this PR. I'll add it as a separate one with all fixes.

deivid-rodriguez (approved):

PR #6438, opened 2023-03-07
Avoid calling String#dup in Gem::Version#marshal_dump

segiddins:

We're doing unnecessary work when marshal dumping versions, since the provided string is never returned directly to the user
Might potentially save a second every time RubyGems.org creates a specs index

Use the ivar directly, instead of calling the method that calls `@version.dup`


simi:
> Might potentially save a second every time RubyGems.org creates a specs index

Sounds great!

indirect (approved):

deivid-rodriguez (approved):

PR #6439, opened 2023-03-07
Put extension_dir at first on full_require_paths when using C ext gem

hsbt:

When we set `GEM_HOME` environment variable to `~/.gem` and use ruby version manager like `rbenv`, `chruby`, We always run `gem pristine` command for rebuild C extensions.

Because https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L47 install build artifact to `lib_dir` without `arch` and ruby version. 

I reverse orer of `Gem::Specification#full_require_paths` with C ext gems. 

before $LOAD_PATH:

```
    [
      "/Users/hsbt/.local/share/gem/gems/oj-3.14.2/lib",
      "/Users/hsbt/.local/share/gem/extensions/arm64-darwin-22/3.3.0+0-static/oj-3.14.2",
      "/Users/hsbt/.local/share/rbenv/versions/3.3.0-dev/lib/ruby/site_ruby/3.3.0+0",
      (snip)
      "/Users/hsbt/.local/share/rbenv/versions/3.3.0-dev/lib/ruby/3.3.0+0/arm64-darwin22"
    ]
```

after $LOAD_PATH:

```
    [
      "/Users/hsbt/.local/share/gem/extensions/arm64-darwin-22/3.3.0+0-static/oj-3.14.2",
      "/Users/hsbt/.local/share/gem/gems/oj-2.14.2/lib",
      "/Users/hsbt/.local/share/rbenv/versions/3.3.0-dev/lib/ruby/site_ruby/3.3.0+0",
      (snip)
      "/Users/hsbt/.local/share/rbenv/versions/3.3.0-dev/lib/ruby/3.3.0+0/arm64-darwin22"
    ]
```

## Further work

I need to more investigate effect of this change. current gem implementation mixed `.rb` and `.so` like https://github.com/ohler55/oj/blob/develop/lib/oj.rb#L12. `oj` case is safe for me. 

### DIFF
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -145,15 +145,17 @@ class Gem::BasicSpecification
   def full_require_paths
     @full_require_paths ||=
     begin
-      full_paths = raw_require_paths.map do |path|
-        File.join full_gem_path, path.tap(&Gem::UNTAINT)
-      end
-
-      full_paths << extension_dir if have_extensions?
+      full_paths = []
 
       # We should search `extension_dir` first. because lib_dir mixed with
       # several architecture and ruby versions is a mess.
-      full_paths.reverse
+      full_paths << extension_dir if have_extensions?
+
+      full_paths = full_paths + raw_require_paths.map do |path|
+        File.join full_gem_path, path.tap(&Gem::UNTAINT)
+      end
+
+      full_paths
     end
   end
 
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -151,7 +151,9 @@ class Gem::BasicSpecification
 
       full_paths << extension_dir if have_extensions?
 
-      full_paths
+      # We should search `extension_dir` first. because lib_dir mixed with
+      # several archtecture and ruby versions is a mess.
+      full_paths.reverse
     end
   end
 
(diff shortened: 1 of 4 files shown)


byroot:
The pattern that worries me here would be:

```ruby
# in lib/mylib.rb
require 'mylib.so'
```

With this change I fear that `require "mylib"` will load the `.so` first, and the `.rb` never.


I think this patten was popular a while ago. I don't have an example in mind but it should be possible to find some by searching for `require *.so` in on the gem codesearch server.

hsbt:
@byroot Thanks for your comment. I also worried your case. I know this pattern is used widely for injecting pure ruby logic into C extension on gem.

I'll do this proposal carefully for not breaking gem ecosystem.

hsbt:
or deprecate to copy build artifact to `lib` directory.

https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/ext_conf_builder.rb#L47

mohits:
@hsbt - will that affect fat gems?

hsbt:
@mohits Nothing. Because fat gem like nokogiri, grpc have separated binaries each Ruby versions under their `lib` directory. RubyGems will not build their library under `extension_dir`.

hsbt:
I re-consider this and withdraw this because this may have large effect for ruby ecosystem. see https://github.com/rubygems/rubygems/pull/6463.

PR #6444, opened 2023-03-08
Detect extension files under full_required_paths

hsbt:

When we use this methods with local gemspec, we don't handle build status of extension correctly. So We need to find extension files in require_paths.

Example with ruby/erb repository:

```
$ bundle exec irb
Ignoring erb-4.0.2 because its extensions are not built. Try: gem pristine erb --version 4.0.2
>>
```

I added additional detection logic for above situation. `Gem::Specification` find extension files from `full_required_paths`.

But `Gem::Specification` couldn't access extension name from `extconf.rb`. So we find them with heuristic way. We should improve it or implement `gem.build_complete` file feature to `rake-compiler`.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -12,6 +12,8 @@ require_relative "stub_specification"
 require_relative "platform"
 require_relative "util/list"
 
+require "rbconfig"
+
 ##
 # The Specification class contains the information for a gem.  Typically
 # defined in a .gemspec file or a Rakefile, and looks like this:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@hsbt This sounds like a change pretty deep down in RubyGems core logic, could we add a test to verify what this is fixing? I thought that this was just a WIP and you were still considering the different alternative solutions (like making `rake-compiler` aware of `gem.build_complete`).

hsbt:
@deivid-rodriguez I see. I'll add the test at least.

kou:
@hsbt Do you remember a warning message with erb?

PR #6446, opened 2023-03-08
Added only missing extensions option into pristine command

hsbt:

The current `gem pristine --extensions` try all of C-ext gems. But I sometimes need to pristine(rebuild) only broken C ext gems.

In background: I have over 100+ C-ext gems each Ruby versions. `gem pristine --extensions` spend while long time.

I added `--only-missing-extensions` option into `gem pristine`. It will rebuild only gems that `Gem::Specification#missing_extensions?` returns `true`.


deivid-rodriguez:
I'm good with adding this.

deivid-rodriguez:
By the way, I just found about `gem pristine --extensions` 😅

PR #6462, opened 2023-03-10
Write test for GH-6444

hsbt:

https://github.com/rubygems/rubygems/pull/6444 has no test because my laziness.

Write a test-case for simple scenario. I'm working this issue with `rake-compiler` until next stable version of RubyGems.

### DIFF
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -3674,7 +3674,16 @@ end
     FileUtils.mkdir_p File.join(@ext.gem_dir, "lib")
 
     # ext_spec used empty extconf.rb, so we need to create dummy extension for rake-compiler case.
+    # Ex. lib/gemname.so
     FileUtils.touch File.join(@ext.gem_dir, "lib", "#{@ext.name}.#{RbConfig::CONFIG['DLEXT']}")
+
+    refute @ext.missing_extensions?
+
+    # Try to another case of extconf.rb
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Test looks good @hsbt.

I think I understand the problem better now. 

Something I don't like is that in #6444 we are adding logic that will run for every single `require` call, just to fix a very small warning edge case (running `bundle exec` in the context of developing a gem with extensions).

I think it'd be better to:

* Fix `Bundler::Source::Path` to return a correct `Specification#extension_dir` so that we don't need to loop through all `full_require_paths` and can look at `Specification#extension_dir` directly.
* Once the above is fixed, change `rake-compiler` to drop a sentinel `gem_build.complete` file at `Specification#extension_dir` and thus be compatible with the current mechanism.

hsbt:
@deivid-rodriguez Are you say about https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/source/path.rb#L171 ?

deivid-rodriguez:
I mean something similar to what we do to "fix" `Specification#full_gem_path` for `Bundler::Source::Path` sources:

https://github.com/rubygems/rubygems/blob/6a62e461cd648285c10c4c9b3b07fcfb2708f9ad/bundler/lib/bundler/rubygems_ext.rb#L47-L53

hsbt:
I see. It's reasonable for me too. I'll move #6444 to bundler side.

Just I merge this test until I finish to move your suggestion.

PR #6470, opened 2023-03-13
Reduce method call in missing_extensions?

hsbt:

https://github.com/rubygems/rubygems/pull/6444 is heavily called in RubyGems.

I moved search logic to bundler side. This logic called only git source repo.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -82,9 +82,10 @@ module Gem
       # TODO: Gem::Specification couldn't access extension name from extconf.rb
       #       so we find them with heuristic way. We should improve it.
       if source.respond_to?(:root)
-        return false if (full_require_paths - [extension_dir]).any? do |path|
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
@hsbt Can you reproduce anymore the problem you mentioned in https://github.com/rubygems/rubygems/pull/6444? Was this problem fixed in the rake-compiler side in the end? I tried and I couldn't reproduce the original issue.

I'm worried that releasing this may cause some accidental regression, and it's unclear what situation this is fixing, since there's no test with this PR.

Do you mind if I revert this, and if the problem shows up again, we release the fix as a patch version after reevaluating it?

hsbt:
I'll look again.

hsbt:
@deivid-rodriguez Sorry, I couldn't look this today. But this issue is only for Ruby committers especially I and k0kubun. It's ok to revert this for RG 3.5 and Bundler 2.5 👍 

I will rise this if I could reproduce this.

deivid-rodriguez:
Thank you! I'll be happy to help if we see the issue reoccur!

PR #6472, opened 2023-03-13
Allow simulating "Bundler 3 mode" more easily

deivid-rodriguez:

Currently to test Bundler 3 mode we have to actually edit the version file to simulate we're running a future version. This is inconvenient.
    
Instead, allow passing an environment variable, `BUNDLER_EFFECTIVE_VERSION`, to set the "working mode" Bundler should use.
    
This can now be set easily by end users to enable them to try out the changes in the future version and give us feedback.
    
Theirs is one small change with respect to the previous way. When running something like `bundle update --bundler 2.3.0`, what Bundler does is install Bundler 2.3.0 and then re'exec using that. Previously, the version file would of course not be changed in the 2.3.0 version just installed, so "the simulated future Bundler" would start behaving as the standard Bundler after re'exec. Now, since the behavior is controlled by an ENV variable, you also get "Bundler 3 mode" after re'excing. I think this is fine and I only mention it because I needed to change a couple of specs due to this change (I added the `--verbose` flag to them so that the output they check is not dependant on the version of Bundler, since Bundler 3 is less verbose that Bundler 2 by default).


bundler/lib/bundler/version.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: false
 
 module Bundler
-  VERSION = "2.7.0.dev".freeze
+  VERSION = (ENV["BUNDLER_3_MODE"] == "true" ? "3.0.0" : "2.7.0.dev").freeze
```
simi: Not sure how beneficial that could be, but what about to make it possible to override version at all with like `BUNDLER_OVERRIDE_VERSION=3.1.0`?

deivid-rodriguez: That was my first approach, but I ended up leaning towards not allowing that and just enable testing exactly what we would like end users to test.

deivid-rodriguez:
I added some changes to make sure the next major is actually Bundler 4, so now the environment variable is `BUNDLER_4_MODE`.

deivid-rodriguez:
I think this is good enough to allow easily testing Bundler 4.

I considered to expose a setting but that complicates things because we'd need to add some restart mechanism, since we'd read the setting using the installed version, and once we find "bundler 4" is configured we would have to restart in "Bundler 4" mode.

I think an ENV variable is simple and good enough, so I'll set this as ready.

bundler/lib/bundler/version.rb
deivid-rodriguez: @simi Planning on merging this now. Is my response satisfactory or you'd still like to go with allowing setting an arbitrary "virtual version"?

deivid-rodriguez: I'll assume it's ok, please do let me know if you believe this need more tweaks!

simi: All good!

PR #6488, opened 2023-03-15
Modify shebang to use settings

### LINKED ISSUE #6478: bundler install shebang
When installing gems via `gem install` the resulting binstubs have a shebang line with `#!/path/to/embedded/bin/ruby` but when doing via `bundle install` it creates as `#!/usr/bin/env ruby`.

Bundler doesn't have a `--no-env-shebang` option like `gem install` does and I tried to specify `bundle config set BUNDLE_SHEBANG /path/to/embedded/bin/ruby` but the installed gems are still with the `env ruby` shebang.

How can I get the `bundler install` to use the full path and not the shebang?

I'd expect both `gem install` and `bundle install` of the same gems to behave the same.
Currently the default for `gem install` provides the full path to the interpreter in the shebang whereas `bundle install` defaults to the `/usr/bin/env ruby` option.

I would expect the behavior to be the same as well as option to disable that option in both. (like `--[no-]env-shebang`)

cdenneen:
Fixes #6478

### DIFF
--- a/bundler/lib/bundler/source/rubygems.rb
+++ b/bundler/lib/bundler/source/rubygems.rb
@@ -176,7 +176,7 @@ module Bundler
           :bin_dir => bin_path.to_s,
           :ignore_dependencies => true,
           :wrappers => true,
-          :env_shebang => true,
+          :env_shebang => Bundler.settings[:shebang],
           :build_args => options[:build_args],
           :bundler_extension_cache_path => extension_cache_path(spec)
         )


bundler/lib/bundler/source/rubygems.rb
```diff
@@ -172,7 +172,7 @@ def install(spec, options = {})
           :bin_dir => bin_path.to_s,
           :ignore_dependencies => true,
           :wrappers => true,
-          :env_shebang => true,
+          :env_shebang => Bundler.settings[:shebang],
```
simi, line 175: :thinking: I see there was some discussion for this at #6478 already. Looking at this again, `Bundler.settings[:shebang]` is `nil` by default. To keep this backward compatible, would it make sense to keep the default `true`, but respect `Bundler.settings[:shebang]` when set (aka when not `nil`)? Also some test would be great to cover this.

simi:
If I understand it well, this current code change changes the behaviour, but there are no failing specs. Would it be possible to cover this better in test suite?

cdenneen:
@simi I created this PR based on @deivid-rodriguez suggestion. I wouldn't know how to do the requested changes. Please feel free to update the PR or replace with new one that includes this fix.

Thanks

deivid-rodriguez:
@cdenneen Can you confirm that you have `--no-env-shebang` in your `~/.gemrc`. After a second thought, I think the right fix for this is to check that RubyGems setting.

Unifying Bundler & RubyGems shebang behavior would be a separate thing.

cdenneen:
> @cdenneen Can you confirm that you have `--no-env-shebang` in your `~/.gemrc`. After a second thought, I think the right fix for this is to check that RubyGems setting.
> 
> Unifying Bundler & RubyGems shebang behavior would be a separate thing.

I'm not putting that in a `.gemrc` as these are builds on multiple windows machines being built with packer. 
I can have packer create a `.gemrc` but think the issue is it's behavior being different is more of the issue.
I can't see any reason why a `gem install` and a `bundle install` of a gem creates binstubs but one has env-shebang and the other doesn't.

How do you propose to check the RubyGems setting to fix this?

So your suggestion is to create a `.gemrc` that specifies:

```
---
install: --no-document --no-env-shebang
update: --no-document --no-env-shebang
```

cdenneen:
> @cdenneen Can you confirm that you have `--no-env-shebang` in your `~/.gemrc`. After a second thought, I think the right fix for this is to check that RubyGems setting.
> 
> Unifying Bundler & RubyGems shebang behavior would be a separate thing.

@deivid-rodriguez also it's been a bit since I opened this but if I recall the issue here isn't the `gem install` but the `bundle install`. `gem install` produces the correct shebang on the binstubs whereas bundler was putting in the env-shebang. 
I ended up having to stop using Bundler in our packer builds because of this but it's lead to issues where I can't scale `--jobs 4` during install. I also can't lock with `Gemfile.lock` all the dependencies that get installed. It's really been frustrating not being able to leverage Bundler for these builds to ensure consistency vs getting an updated dependency that now might break something. I agree with @simi there isn't a test around this but feel it's useful change vs forced to `true` as it is now.

So doesn't `Bundler.settings[:shebang]` currently defaultsto `true` so this change doesn't actually affect anything unless you specify `Bundler.settings[:shebang]` to be false or something?

cdenneen:
I've tried adding this `gemrc` file to `C:\Users\Administrator\.gemrc`,`C:\Users\Administrator\gemrc`, `C:\ProgramData\gemrc` and `bundle install` still results in:
```
C:\Program Files\Puppet Labs\Puppet\puppet\bin>more r10k
#! ruby
```
Where as when I did the `gem install`
```
C:\Program Files\Puppet Labs\Puppet\puppet\bin>more r10k
#!"C:/Program Files/Puppet Labs/Puppet/puppet/bin/ruby.exe"
```

cdenneen:
This will be bigger issue for me since the embedded ruby is 2.5 so latest ruby gem I can upgrade to is the 3.3.x releases, would need to patch this into 3.3.27 or something

deivid-rodriguez:
Hei!

So what I meant to say is that I realized that the `shebang` Bundler setting is meant to control the project binstubs `bundle binstubs` generates, not global binstubs generated by RubyGems.

So, instead of respecting that setting here, I think we should respect the RubyGems setting.

That RubyGems setting is controlled by the `--[no-]env-shebang` RubyGems flag, and defaults to `false`.

Regardless of this, handling backwards compatibility seems hard here 🤔.

cdenneen:
@deivid-rodriguez do you know of any workaround/hack for this? 
I'd like to do the install with a Gemfile/Gemfile.lock but just need the shebang line for all the resulting binaries to be #!/PATH/TO/EMBEDDED/ruby instead of #!env ruby since that seems to break constantly on my Windows setup.

If I install the gems via `gem install` the shebang is fine but I keep running into dependency update issues breaking on me so if I can lock down with Gemfile.lock and can also leverage the `bundle install --jobs 4` parallel option, it would make things easier.

deivid-rodriguez:
Given the value is currently hardcoded, there's no workaround at the moment.

Regarding backwards compatibility, it should be very rare for someone to want conflicting settings for Bundler & RubyGems (if we go with using the value of `--env-shebang` configured at `~/.gemrc`), or conflicting settings for project and global binstubs (if we go with `Bundler.settings[:shebang]`).

So I think I'd go with @simi's suggestion at https://github.com/rubygems/rubygems/pull/6488#discussion_r1145296810.

If we go that way, we need to:

* Make this PR more compatible by implementing what @simi suggested.
* Add a spec for this feature.
* Update `shebang` setting documentation to mention that it applies to any shebang generated by Bundler (for both project binstubs and global binstubs).

zacheryph:
`settings[:shebang]` and `env-shebang` are related, but are not the same at all.

The shebang setting is a string, where as `env-shebang` is a boolean.

_Adding_ `env-shebang` as a flag would require addition of handling that flag itself, as well as documentation.

jeromedalbert:
In order to implement the right solution, let me summarize the problem first, to the best of my understanding.

### Comparison of binstub shebangs

- Binstubs generated by `gem install`:
    - Default shebang: `full/path/to/ruby`
    - There is a `--env-shebang` boolean option to change the shebang to `/usr/bin/env`.

- Binstubs generated by `bundle install`:
    - Default shebang: `/usr/bin/env ruby`
    - **(\*)** There is no boolean option to change the shebang to `full/path/to/ruby`, which in most cases is fine, because the ruby locations to run a project can vary widely from environment to environment and even within the same environment types (for example developers' ruby locations can vary a lot as [mentioned](https://github.com/rubygems/rubygems/issues/6478#issuecomment-2087939997) by @zacheryph). So `/usr/bin/env ruby` makes sense most of the time because it works for any ruby location.
    - There is a `--shebang` (or `BUNDLE_SHEBANG` aka `Bundler.settings[:shebang]`) string option to change the "ruby" in `/usr/bin/env ruby` to another value, e.g. "jruby" would result in `/usr/bin/env jruby`.

### Problem

@cdenneen wants to install gems globally with `bundle install` instead of the usual `gem install` command, for easier and automatic dependency management, as he is having dependency update issues and manually managing dependencies is a headache. As a bonus he can use Bundler's parallel `--jobs` option.

But when installing gems via `bundle install`, the generated binstub shebangs are `/usr/bin/env ruby`, and this is an issue because that shebang seems to [break constantly](https://github.com/rubygems/rubygems/pull/6488#issuecomment-1591182154) on his Windows setup.

He would like to change the generated shebang to `full/path/to/ruby`, but there is currently no option to do that with `bundle install`.

### Possible solution

This seems like a valid reason to introduce some sort of new option for `bundle install`, like `--no-env-shebang` or `--full-path-shebang`. Providing this option would trigger the new behavior that @cdenneen is interested in, i.e. generating binstubs with a `full/path/to/ruby` shebang.

In terms of defaults, I think that bundler binstub shebangs should continue to be `/usr/bin/env ruby` by default, not only to preserve backwards compatibility mentioned by @simi, but more importantly because of **(\*)**. This means that gem binstub shebangs and bundler binstub shebangs would continue to be different, as it arguably should be, because I think that gem global installations and bundler local/project installations have different use cases for the majority of people. So this is a case where consistency between `gem install` and `bundle install` is not necessarily desired IMO, but let me know if you have a different opinion. In any case the more pressing issue to fix is not about consistency but just allowing a different shebang.

As shown in my binstub comparison above and as [hinted at](https://github.com/rubygems/rubygems/pull/6488#issuecomment-2087943648) by @zacheryph, `--shebang`/`BUNDLE_SHEBANG`/`Bundler.settings[:shebang]` is mostly unrelated to the problem. It does become slightly related if we introduce a `--no-env-shebang`/`--full-path-shebang` option though, as the naming would be similar to the existing `--shebang` option, and this might confuse people. This may not be a pressing issue, but a way to fix this could be to deprecate/alias `--shebang` in favor of a new name like `--shebang-executable` or `shebang-ruby` or similar.

### Alternate solution

An alternate solution is to do nothing, i.e. not add any new option to Bundler if the problem is considered too niche. I am not saying that Puppet Enterprise is niche, but maybe the Windows issue is, as it is hard to know without more details. If the true root cause of the issue is that `/usr/bin/env ruby` breaks on @cdenneen's Windows setup, this could maybe be fixed by debugging the error messages a bit more, or worked around with a hacky sed one-liner going over every binstub and setting `full/path/to/ruby` as the shebang, something like `find full/path/to/ruby/bin -type f -exec sed -i '1c #!/full/path/to/ruby' {} +` (untested).

A counter-argument is that it is currently already possible to install gems globally via `bundle install`, and global installations can benefit from using a full path shebang to avoid all sorts of issues including @cdenneen's. Using a full path shebang is [more predictable](https://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my/29620#29620) in that sense, and I guess this is why `gem install` defaults to a full path shebang to begin with.

---

What are your thoughts @deivid-rodriguez and @simi? Once we agree on the approach I can make a PR if needed.

deivid-rodriguez:
Hi @jeromedalbert, thanks for the great summary.

My initial thought was that, since Bundler uses RubyGems gem installer under the hood, it should make sure to respect the RubyGems setting that controls the generated shebang. So ideally we respect `install: --env-shebang` in ~/.gemrc`.

However:
* RubyGems configuration is explicitly tailored for the `gem` CLI, so it's not obvious that Bundler should respect it.
* RubyGems and Bundler have different defaults, so we can't make this backwards compatible.

I think maybe I'd go with some form of option 1. Maybe add an `env_shebang` boolean setting to Bundler analogous to RubyGems one. We could consider later to clarify or rename the existing `shebang` string Bundler setting if it's causing confusion.

jeromedalbert:
> Maybe add an env_shebang boolean setting to Bundler analogous to RubyGems one.

Adding a new `--env-shebang` option sounds good to me. 

If we want to be fully analogous to the Rubygem behavior though, should we also change the default shebang for `bundle install` global binstubs [here](https://github.com/rubygems/rubygems/blob/238eedc5376240954676a893b631f7def7ed1efd/bundler/lib/bundler/source/rubygems.rb#L179), or is that too much @deivid-rodriguez ? Otherwise we are not truly analogous because in order to get `full/path/to/ruby` in global binstubs, with `gem install` it is done by default, but with `bundle install` you'd have to provide `--no-env-shebang`.

deivid-rodriguez:
@jeromedalbert I think that'd be nice but for now I'd just give users the option of making things consistent by explicitly configuring `env_shebang` to `false`. Then we can consider changing the default in the next major version and make things consistent by default.

Also, I believe `--shebang` as a CLI flag is deprecated, so we should be adding `env_shebang` only as configuration, not as a CLI flag.

jeromedalbert:
> Also, I believe --shebang as a CLI flag is deprecated, so we should be adding env_shebang only as configuration, not as a CLI flag.

I am unclear on this setting vs CLI part; `env_shebang` seems unrelated to `--shebang` as mentioned in my summary, so `--shebang` being potentially deprecated doesn't seem related to the decision of using a setting and/or flag for `env_shebang`, but maybe I am misunderstanding and starting a PR might clear things up. I do not see any mention of deprecation in the [`--[no-]env-shebang`](https://github.com/rubygems/rubygems/blob/c40def1658d4d7b75d38ecf7c7aed39e76c437db/lib/rubygems/install_update_options.rb#L67-L71) gem install/update option either.

deivid-rodriguez:
Basically, `bundle install --shebang` is a "remembered setting". That means that once you use it once, it's recorded in your configuration file and used in every subsequent `bundle install` invocation.

That was a trick used years to be able to "configure Bundler and run a command at the same time". However, the silent saving of install options was very confusing to users, so we deprecated all "sticky flags" and made them configuration options only, because their main purpose was to stay for all Bundler invocations, not to affect just one Bundle command. 

A new `env_shebang` setting feels the same kind of setting to me. Something that you want set for all Bundler invocations, not some option that you'd want just for a single `bundle install` command. Hence, my suggestion of making it a configuration option rather than a CLI flag.

deivid-rodriguez:
Let's close this in favor of https://github.com/rubygems/rubygems/pull/7849. @cdenneen please do follow up in there if still interested!

PR #6490, opened 2023-03-16
Fix `$LOAD_PATH` in rake and ext_conf builder

ntkme:

Installing gem with rake based extension which depends on a native gem (e.g. psych) that has more than one copies on the system (one comes default with ruby installation and one user installation) with the system rubygems can fail in a very strange way.

Reproduction: https://github.com/sinatra/sinatra/pull/1911#issuecomment-1471165928
Explanation: https://github.com/sinatra/sinatra/pull/1911#issuecomment-1471182649

Per previous PR related to this issue: https://github.com/rubygems/rubygems/pull/4165

> I'm not sure using this -I parameter here is a good idea, so since I'm not sure I'm just keeping things as they are for now and fixing the bug.

Removing this flag caused one of the existing tests to be broken: https://github.com/rubygems/rubygems/actions/runs/4433225457/jobs/7778058872

Therefore, this PR removes it conditionally.

- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes (existing tests should be enough)

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -54,16 +54,20 @@ class Gem::Ext::Builder
     end
   end
 
-  def self.rubygems_load_path
+  def self.ruby
+    require "shellwords"
+    # Gem.ruby is quoted if it contains whitespace
+    cmd = Gem.ruby.shellsplit
+
     # This load_path is only needed when running rubygems test without a proper installation.
     # Prepending it in a normal installation will cause problem with order of $LOAD_PATH.
     # Therefore only add load_path if it is not present in the default $LOAD_PATH.
     load_path = File.expand_path("../..", __dir__)
     case load_path
     when RbConfig::CONFIG["sitelibdir"], RbConfig::CONFIG["vendorlibdir"], RbConfig::CONFIG["rubylibdir"]
-      []
+      cmd
     else
-      ["-I#{load_path}"]
+      cmd << "-I#{load_path}"
     end
   end
 
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -54,6 +54,19 @@ class Gem::Ext::Builder
     end
   end
 
+  def self.rubygems_load_path
+    # This load_path is only needed when running rubygems test without a proper installation.
+    # Prepending it in a normal installation will cause problem with order of $LOAD_PATH.
+    # Therefore only add load_path if it is not present in the default $LOAD_PATH.
+    load_path = File.expand_path("../..", __dir__)
+    case load_path
+    when RbConfig::CONFIG["sitelibdir"], RbConfig::CONFIG["vendorlibdir"], RbConfig::CONFIG["rubylibdir"]
+      []
+    else
+      ["-I#{load_path}"]
+    end
+  end
+
   def self.run(command, results, command_name = nil, dir = Dir.pwd, env = {})
     verbose = Gem.configuration.really_verbose
 
--- a/lib/rubygems/ext/ext_conf_builder.rb
+++ b/lib/rubygems/ext/ext_conf_builder.rb
@@ -21,22 +21,10 @@ class Gem::Ext::ExtConfBuilder < Gem::Ext::Builder
 
     destdir = ENV["DESTDIR"]
 
-    # This is only needed when running rubygems test without a proper installation.
-    # Prepending it in a normal installation can cause problem with order of $LOAD_PATH.
-    # Therefore only add rubygems_load_path if it is not present in the default $LOAD_PATH.
-    rubygems_load_path = File.expand_path("../..", __dir__)
-    load_path =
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
Hi! Thanks so much! You could've just moved on by upgrading RubyGems, but you fixed the root cause 😍.

Could you add a test for this?

ntkme:
> Could you add a test for this?

I'd be happy to. The case is so specific that I'm not sure how to write it / where to put it:

1. It has to be a system installation of rubygems, as `gem update --system` will change the behavior so the bug is no longer reproducible.
2. It has to install a second copy of a dependency gem (e.g. psych) that already comes with ruby and the version of second installation needs to be different enough so that the linking between the ruby code and native ext could break.
3. The Rakefile in the gem being installed needs to use that the dependency gem.

@deivid-rodriguez If you can give me some pointer to maybe a similar test case we have currently, I can try come up with a test case.

deivid-rodriguez:
I was thinking of something similar to the one added by https://github.com/rubygems/rubygems/pull/4165, but this seems much more specific, indeed.

For this kind tricky tests I normally use realworld testing like the commands run in `.github/workflows/install-rubygems.yml`. But that requires manually installing RubyGems first, which does not have the same structure as the RubyGems bundled with a pristine version of Ruby.

How are you testing that the fix works? Directly editing the RubyGems copy included with Ruby?

Anyways, yeah, too hard to get this tested, I think it's fine as is.

ntkme:
> How are you testing that the fix works? Directly editing the RubyGems copy included with Ruby?

Exactly, I start with a ruby 3.0 container which I used in the reproduction, manually apply the patch to the default rubygems, and confirmed that it fixed the problem.

deivid-rodriguez:
This is a ruby-core specific thing, not a RubyGems thing, so I found a little discrepancy. On my laptop:

```
$ find /Users/deivid/.asdf/installs/ruby/3.2.1/lib/ruby/3.2.0/ -type d -depth 1 -name 'arm*'
/Users/deivid/.asdf/installs/ruby/3.2.1/lib/ruby/3.2.0//arm64-darwin22

$ ruby -e 'puts Gem::Platform.local'                               
arm64-darwin-22

$ ruby -e 'puts RUBY_PLATFORM'                                     
arm64-darwin22
```

So it sounds like we want the latter.

ntkme:
Oh wow. On linux it was the same. Let me update the PR.

ntkme:
Umm, there is actually two different possibilities:

https://github.com/ruby/ruby/blob/5825d7d4a1a53ba84909a01679f4dfab63fa9739/loadpath.c#L25-L27
https://github.com/ruby/ruby/blob/5825d7d4a1a53ba84909a01679f4dfab63fa9739/loadpath.c#L48-L50
https://github.com/ruby/ruby/blob/5825d7d4a1a53ba84909a01679f4dfab63fa9739/loadpath.c#L86-L90

Depends on whether `RUBY_PLATFORM_CPU` is defined, it will be either RUBY_ARCH or RUBY_THINARCH...

ntkme:
@deivid-rodriguez How here is the problem:

For an upgraded rubygems, because it is in site_ruby, it should be:

``` ruby
[ "-I#{RbConfig::CONFIG['sitelibdir']}", "-I#{RbConfig::CONFIG['sitearchdir']}" ]
```

For a default rubygems installation it should be:

``` ruby
[ "-I#{RbConfig::CONFIG['rubylibdir']}", "-I#{RbConfig::CONFIG['rubyarchdir']}" ]
```

Not sure what's the best way to detect what kind of rubygems installation we have...

ntkme:
@deivid-rodriguez I updated the solution and added comment in the code to explain it. It might be an overkill, but should work.

ntkme:
@deivid-rodriguez Sorry for lots of back and forth changing the PR. I think now I have the most clean and robust solution:

If rubygems' install location is not present in `$LOAD_PATH`, we add it. - This happens in the rubygems development and test.
If rubygems' install location is present in `$LOAD_PATH`, just don't add it. - This happens for normal users.

hsbt:
@ntkme Wait to push for re-run tests. We reached to limit of macOS jobs.

hsbt:
Can you rebase this from https://github.com/rubygems/rubygems/commit/7e5fbea7dfd523714ff19a3a2bfe4a0d14fdffda? Thanks.

deivid-rodriguez:
@ntkme Nice, your solution makes sense to me! Should we move this common logic to the base ext builder?

deivid-rodriguez (approved):
Awesome, thanks for your work!

ntkme:
Looks like merge queue check failed due to flakiness of #6512. Not sure if any action is required.

deivid-rodriguez:
I'm merging the PR fixing the flaky failure and then will merge this one. No action needed @ntkme, thanks for your work!

PR #6493, opened 2023-03-16
Don't recommend `--full-index` on errors

deivid-rodriguez:

The `--full-index` flag is slow since the resolver needs to fetch the remote gemspec every time it needs to figure out dependencies of a gem, which happens a lot during resolution.

Plus it's still potentially vulnerable to dependency [confusion issues](#4694).

Also, some of this recommendations did not even make sense as far as I understand.

Finally these days the dependency APIs are reliable enough, and there are implicit fallbacks in place anyways.

Stop recommending `--full-index`.

### DIFF
--- a/bundler/lib/bundler/fetcher/dependency.rb
+++ b/bundler/lib/bundler/fetcher/dependency.rb
@@ -34,13 +34,9 @@ module Bundler
 
         returned_gems = spec_list.map(&:first).uniq
         specs(deps_list, full_dependency_list + returned_gems, spec_list + last_spec_list)
-      rescue MarshalError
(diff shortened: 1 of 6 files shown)


simi (approved):

deivid-rodriguez:
I created a separate PR to bump the `ruby/setup-ruby` action with Dependabot, but I now realized I think this is actually fixed already because in the case of mswin we use the `ruby/setup-ruby-pkgs` action which depends on a "rolling v1 version" of `ruby/setup-ruby`, so the fix has been picked up automatically I think. Let me try it though.

PR #6476, opened 2023-03-14
Improve TarHeader#calculate_checksum

Maumagnaguagno:
String#unpack("C*") is equivalent to String#bytes and Array#inject {|a,b| a + b } is equivalent to Array#sum.
This is a minor enhancement in terms of speed and readability.

## Benchmark

This benchmark uses header data from related [tests](https://github.com/rubygems/rubygems/blob/master/test/rubygems/test_gem_package_tar_header.rb):

```ruby
PACK_FORMAT = "a100" + # name
              "a8"   + # mode
              "a8"   + # uid
              "a8"   + # gid
              "a12"  + # size
              "a12"  + # mtime
              "a7a"  + # chksum
              "a"    + # typeflag
              "a100" + # linkname
              "a6"   + # magic
              "a2"   + # version
              "a32"  + # uname
              "a32"  + # gname
              "a8"   + # devmajor
              "a8"   + # devminor
              "a155"   # prefix

def header(checksum)
  header = [
    "x",
    oct(0644, 7),
    oct(1000, 7),
    oct(10000, 7),
    oct(100, 11),
    oct(12345, 11),
    checksum,
    " ",
    "0",
    "link",
    "ustar",
    oct("00", 2),
    "user",
    "group",
    oct(1, 7),
    oct(2, 7),
    "y"
  ]
  header = header.pack PACK_FORMAT
  header << ("\0" * ((512 - header.size) % 512))
end

def oct(num, len)
  "%0#{len}o" % num
end

h = header(' ' * 8)
p a = h.unpack("C*").inject {|a, b| a + b }
p b = h.bytes.sum
p oct(a,6)
p oct(b,6)

require 'benchmark'
n = 100_000
Benchmark.bm do |benchmark|
  benchmark.report("old checksum") do
    n.times do
      h.unpack("C*").inject {|a, b| a + b }
    end
  end
  benchmark.report("new checksum") do
    n.times do
      h.bytes.sum
    end
  end
end
```

The result is ~8 times faster for 100_000 runs with Ruby 3.0.5p211.
```
5431
5431
"012467"
"012467"
       user     system      total        real
old checksum  3.966327   0.003138   3.969465 (  3.971703)
new checksum  0.496551   0.000001   0.496552 (  0.496906)
```

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write tests => **tests already exist**
- [x] Write code to solve the problem
[...]

PR #6494, opened 2023-03-16
Avoid intermediate array in TarHeader#calculate_checksum

Maumagnaguagno:

Avoid the call to String#bytes which generates an intermediate array of character values for checksum.

String#sum(0) sums the character bytes without a modulo, much faster.
Follow-up of #6476 based on comment from @nobu.

```
old checksum  3.954847   0.000451   3.955298 (  3.957550)
new checksum  0.490703   0.000003   0.490706 (  0.490972) #6476
sum checksum  0.018343   0.000000   0.018343 (  0.018351) #This PR
```

- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes **=> tests already exist**

### DIFF
--- a/lib/rubygems/package/tar_header.rb
+++ b/lib/rubygems/package/tar_header.rb
@@ -208,7 +208,7 @@ class Gem::Package::TarHeader
   private
 
   def calculate_checksum(header)
-    header.bytes.sum
+    header.sum(0)
   end
 
   def header(checksum = @checksum)
--- a/test/rubygems/package/tar_test_case.rb
+++ b/test/rubygems/package/tar_test_case.rb
@@ -67,7 +67,7 @@ class Gem::Package::TarTestCase < Gem::TestCase
   end
 
   def calc_checksum(header)
-    sum = header.bytes.sum
+    sum = header.sum(0)
     SP(Z(to_oct(sum, 6)))
   end
 


simi (approved):

martinemde (approved):

lib/rubygems/package/tar_header.rb
```diff
@@ -208,7 +208,7 @@ def update_checksum
   private
 
   def calculate_checksum(header)
-    header.bytes.sum
+    header.sum(0)
```
nobu, line 211: Although I don't remember the spec of tar well, as this value is formatted as `"%06o"`, upper bits should be dropped?

```suggestion
    header.sum
```

martinemde: Let's double check just to be sure we're following spec. [Wikipedia](https://en.wikipedia.org/wiki/Tar_(computing)#:~:text=The%20checksum%20is%20calculated%20by,NUL%20and%20then%20a%20space.) says:

> The [checksum](https://en.wikipedia.org/wiki/Checksum) is calculated by taking the sum of the unsigned byte values of the header record with the eight checksum bytes taken to be ASCII spaces (decimal value 32). It is stored as a six digit octal number with leading zeroes followed by a NUL and then a space. Various implementations do not adhere to this format. In addition, some historic tar implementations treated bytes as signed. Implementations typically calculate the checksum both ways, and treat it as good if either the signed or unsigned sum matches the included checksum.

The chars are unsigned, so the sum of the chars should be the correct number, then it's stored as a 0 padded octal. I reproduced the calculation here using an example gem and it produces the same checksum as when I let the old version calculate the header.

```

martinemde:
@deivid-rodriguez This should be ready to merge. I unqueue to wait for your confirmation, but I do believe it's ready and have confirmed the behavior.

lib/rubygems/package/tar_header.rb
martinemde: Thanks for challenging me to think about all this. I've learned a bunch just because of this.

Since the header is limited to 512bytes and they're not all used (about 500 are used and a big portion of them are \0.) I think it's impossible for the checksum to exceed the 6 character octal, so any argument other than 0 runs more code than necessary in `String#sum` (if I understand the admittedly cryptic implementation well enough).

You're completely right that 2**18-1 is perfect for the space (777777 in octal). I just don't think it's necessary.

PR #6510, opened 2023-03-17
Cleanup CI

deivid-rodriguez:

Just nitpicking. I don't like keeping commented out code, it kind of defeats the point of source control, and it makes me think there's something to fix here to uncomment the code.

Instead, I prefer explaining why we're only running a single job on macOS.

Also cleaned up commented out stuff in the scorecard workflow since I was at it.


hsbt (approved):

simi (approved):

deivid-rodriguez:
I also added an extra commit to fix more instance of s/MacOS/macOS. In particular I think one of them was making the job name not match some jobs show as pending.

PR #6486, opened 2023-03-15
Removed needless condition about RUBY_PATCHLEVEL

hsbt:
## What was the end-user or developer problem that led to this PR?

RUBY_PATCHLEVEL is always provided from Ruby 1.8.5. But RubyGems has a condition about this.

## What is your fix for the problem, implemented in this PR?

Removed needless condition.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6511, opened 2023-03-17
Remove constants condition related ruby releases.

hsbt:

This is similar with https://github.com/rubygems/rubygems/pull/6486. `RUBY_REVISION` is always provided in modern Ruby versions. We don't need to consider this constant condition in our code.

I removed `defined?(RUBY_REVISION)` from `lib/rubygems/request.rb`. And refactor this test code. We have some of needless `defined?` check globally.

### DIFF
--- a/test/rubygems/test_gem_request.rb
+++ b/test/rubygems/test_gem_request.rb
@@ -479,8 +479,7 @@ ERROR:  Certificate  is an invalid CA certificate
 
   def util_restore_version
     Object.send :remove_const, :RUBY_ENGINE
-    Object.send :const_set,    :RUBY_ENGINE, @orig_RUBY_ENGINE if
-      defined?(@orig_RUBY_ENGINE)
+    Object.send :const_set,    :RUBY_ENGINE, @orig_RUBY_ENGINE
 
     Object.send :remove_const, :RUBY_PATCHLEVEL
     Object.send :const_set,    :RUBY_PATCHLEVEL, @orig_RUBY_PATCHLEVEL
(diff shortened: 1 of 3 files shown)


test/rubygems/test_gem_request.rb
```diff
@@ -338,20 +338,6 @@ def test_user_agent_revision
     util_restore_version
   end
 
-  def test_user_agent_revision_missing
```
hsbt, line 341: I removed this test because there is no platform and versions that didn'T have `RUBY_REVISION` today.

martinemde (approved):
RUBY_VERSION looks like it has existed since well before ruby 2.5, which was [EOL 2 years ago](https://www.ruby-lang.org/en/downloads/branches/). I used 2.5 as a reference because it's [the oldest version supported by a Rails that is not also EOL (not 1st party source)](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html). My system has ruby 2.6 by default which also has RUBY_VERSION and can install the latest rubygems still.

test/rubygems/test_gem_request.rb
```diff
@@ -491,22 +477,20 @@ def test_verify_certificate_message_UNABLE_TO_VERIFY_LEAF_SIGNATURE
   end
 
   def util_restore_version
-    Object.send :remove_const, :RUBY_ENGINE
-    Object.send :const_set,    :RUBY_ENGINE, @orig_RUBY_ENGINE if
-      defined?(@orig_RUBY_ENGINE)
+    Object.send :remove_const, :RUBY_ENGINE if defined?(RUBY_ENGINE)
```
deivid-rodriguez, line 480: Why do we need this check (and similar below) now?

hsbt: Thanks. This is debug code before removing `test_user_agent_revision_missing` method. There is no reason we should check them. I removed all of guard conditions.

PR #6520, opened 2023-03-20
add some test suites under the Gem namespace, so as not to collide

luke-gru:
To avoid collisions especially in ruby/ruby

### DIFF
--- a/test/rubygems/test_bundled_ca.rb
+++ b/test/rubygems/test_bundled_ca.rb
@@ -15,7 +15,7 @@ require "rubygems/request"
 # The tested hosts are explained in detail here: https://github.com/rubygems/rubygems/commit/5e16a5428f973667cabfa07e94ff939e7a83ebd9
 #
 
-class TestBundledCA < Gem::TestCase
+class TestGemBundledCA < Gem::TestCase
   def bundled_certificate_store
     store = OpenSSL::X509::Store.new
 
--- a/test/rubygems/test_config.rb
+++ b/test/rubygems/test_config.rb
@@ -4,7 +4,7 @@ require_relative "helper"
 require "rubygems"
 require "shellwords"
 
-class TestConfig < Gem::TestCase
+class TestGemConfig < Gem::TestCase
   def test_datadir
     util_make_gems
     spec = Gem::Specification.find_by_name("a")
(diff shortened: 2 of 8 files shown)


luke-gru:
Yes, ideally all the tests would have proper namespacing as this gem is used in ruby/ruby, but these test classes in particular are not named /^TestGem.+/, so things like TestKernel, TestConfig, TestExit, could possibly be names when running make test-all in the ruby/ruby repository in the future. I say "in the future" because these classes don't currently exist when testing ruby, but they might be added later (I'm honestly surprised `TestKernel` isn't there already). Running make test-all in ruby/ruby tests all ruby functionality and all libraries in a single process, so class names must not clash with each other (including in other default gems). For instance, when ruby tests net/http, if they created a `HTTPResponseFactory` class, it would clash with the one in rubygems. To be a good testing citizen, especially if you're a default gem, namespacing under your module (`Gem`) is the best way to go 👍 If you want me to namespace all of the test classes I can do that, I was just trying to keep the diff small and to address the potential problems first.

simi:
Thanks for detailed explanation. Would it make sense to "namespace" all test files?

luke-gru:
Yes I think so, just for consistency's sake. I'll do that and push an update soon.

hsbt:
It's good to separate namespaces. But We should use `TestGemDeprecate` instead of `Gem::TestDeprecate` with consistency of other tests.

luke-gru:
@hsbt Okay fixed, I changed all of the test classes to be `TestGem*`. The windows build is failing at the moment, I think unrelated to my changes.

simi:
Any particular reason why changes in `util` folder are needed? Those files are not required in RubyGems code (including tests), just used directly to ease maintenance if I understand it well.

luke-gru:
In `test/test_changelog_generator.rb` the file is used, so that's why I changed it.

MSP-Greg:
> The windows build is failing at the moment

Seems similar to https://bugs.ruby-lang.org/issues/19568 and https://github.com/oxidize-rb/rb-sys/actions/runs/4599045589.

I suspect it's a Rust issue.

hsbt:
> In `test/test_changelog_generator.rb` the file is used, so that's why I changed it.

This test only run with this repo, not `ruby/ruby` repo. This file and files under `util` directory are unnecessary to separate namespace.

luke-gru:
Okay, fixed 👍

hsbt:
@luke-gru Can you also revert `Gem::Changelog` namespace? Other changes are okay to me.

luke-gru:
Sorry I forgot a few. Done.

PR #6523, opened 2023-03-20
Add message for otp bypass

ericherscovich:
As described in https://github.com/Shopify/ruby-dependency-security/issues/258, we want a fallback for OTP if OTP and WebAuthN are enabled. By default WebAuthN is chosen. 

After reviewing the `gem signin` flow, we already offer an OTP option. Also, as discussed in the issue linked above, the chosen approach for this was to only offer webauthn, and offer OTP as a fallback, which would be a flag. So, no new logic needed to be introduced. Instead, I added a line to the WebAuthN logic to ensure the user knows they have this option.

Screenshot 2023-03-20 at 2 51 11 PM

This is the OTP flow for signin:
image

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -46,7 +46,7 @@ jobs:
       RGV: ..
       RUBYOPT: --disable-gems
     steps:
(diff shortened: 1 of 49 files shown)


indirect (approved):

aellispierce (approved):
🎉

PR #6491, opened 2023-03-16
Merge rubocop configuration: 2nd PR

hsbt:
## What was the end-user or developer problem that led to this PR?

This is second migration work of https://github.com/rubygems/rubygems/pull/6487. 

## What is your fix for the problem, implemented in this PR?

I migrate bundler's cops to rubygems with following conditions:

* Without effect of Bundler
* Without test failure
* Without change of logic meanings

Finally, I choose and apply safety Cops with auto-correct. 

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6487, opened 2023-03-15
Merge rubygems and bundler rubocop rules 

hsbt:
## What was the end-user or developer problem that led to this PR?

We have two rubocop rules between rubygems and bundler. We should move to merge them.

## What is your fix for the problem, implemented in this PR?

I merge `.rubocop_bundler.yml` into `.rubocop.yml` only non-effect cops at first in this PR. After that, I'll triage `rubocop_bundler.yml` cops and modify rubygems or bundler files step by step.

@deivid-rodriguez @simi Do you have any concerns about this merger plan?

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6529, opened 2023-03-22
Merge rubocop rules 3rd try

hsbt:

This PR is 3rd try of merging rubocop rules.

1st: https://github.com/rubygems/rubygems/pull/6487
2nd: https://github.com/rubygems/rubygems/pull/6491

This PR contained some of logic changes under the `lib` directory. I confirmed to pass test suite. But does anyone review this too?


hsbt:
The remaining rules are:

* Naming/ConstantName and Naming/MethodName
  * These are trivial. But we need to consider compatibility.
* Naming/VariableName and Naming/VariableNumber
  * These are also trivial. 
* Style/ClassVars
  * I need to consider how change RubyGems code for enabling this cop.
* Style/HashSyntax
  * Should we remove this cop? It seems only for Ruby 1.8 compatibility. I have no strong opinion that we should enable Ruby 1.9 style or hash-rocket style.
* Style/FormatString, Style/LineEndConcatenation, Style/NumericLiteralPrefix and Style/RegexpLiteral
  * They are simply applied. But Many of code are affected.
* Performance/RegexpMatch
  * We need to deprecate `Gem::Specification#match?` or disable this cop for under the affected code of RubyGems.

lib/rubygems/basic_specification.rb
```diff
@@ -175,11 +175,13 @@ def to_fullpath(path)
...
           suffixes = Gem.suffixes
-          suffixes.find do |suf|
+          if suffixes.find do |suf|
             full_require_paths.find do |dir|
               File.file?(fullpath = "#{dir}/#{path}#{suf}")
             end
-          end ? fullpath : nil
+          end
+            fullpath
+          end
```
deivid-rodriguez, line 184: This code is a bit crazy, no?

I think if we extract the logic to a separate method we can make it simpler, like

```ruby
def find_full_path(path)
  Gem.suffixes.each do |suf|
    full_require_paths.each do |dir|
      fullpath = "#{dir}/#{path}#{suf}"
      return fullpath if File.file?(fullpath)
    end
  end

  nil
end
```

lib/rubygems/specification.rb
```diff
@@ -2674,7 +2674,7 @@ def version=(version)
     end
     invalidate_memoized_attributes
 
-    return @version
+    return
```
deivid-rodriguez, line 2674: Like you did in bb7094a2d4246cf586ba55e1e0b645574936ab59, maybe better to completely remove this line. That would also allow us to enable https://docs.rubocop.org/rubocop/cops_style.html#styleredundantreturn.

deivid-rodriguez (commented):
Small comments, but looks good!

simi (approved):

lib/rubygems/specification.rb
hsbt: 👍  I enabled `Style/RedundantReturn` cop.

lib/rubygems/basic_specification.rb
hsbt: Maked sense. I rewrite it to this:

```ruby
  def to_fullpath(path)
    if activated?
      @paths_map ||= {}
      Gem.suffixes.each do |suf|
        full_require_paths.each do |dir|
          fullpath = "#{dir}/#{path}#{suf}"
          next unless File.file?(fullpath)
          @paths_map[path] ||= fullpath
        end
      end
      @paths_map[path]
    end
  end
```

lib/rubygems.rb
```diff
@@ -1012,7 +1012,7 @@ def self.load_plugin_files(plugins) # :nodoc:
 
       begin
         load plugin
-      rescue ::Exception => e
+      rescue ::Exception => e # rubocop:disable Lint/RescueException
```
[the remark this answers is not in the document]
hsbt: Ah, I disabled this line because test was failed. I update this line and test fixture of rubygems plugin now.

lib/rubygems/basic_specification.rb
deivid-rodriguez: Oh, that looks great! 👍

PR #6417, opened 2023-02-27
Backport test improvements from ruby/ruby

hsbt:
## What is your fix for the problem, implemented in this PR?

Backport the following commits from ruby/ruby

* https://github.com/ruby/ruby/pull/7394
* <del>https://git.ruby-lang.org/ruby.git/commit/?id=b112ae9971ea63b9f0ba3df049ac4c3bea89bf66</del>

## Make sure the following tasks are checked

- [ ] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [ ] Write code to solve the problem
- [ ] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #6533, opened 2023-03-23
remove test skips/pends on mswin platform

MSP-Greg:

Three tests in `test/test_gem_ext_builder.rb` are currently skipped on the mswin platform.  A long time ago, Actions did not enable the msvc build tools with mswin builds.  setup-ruby now does this, and ruby/ruby's mswin CI does something similar.

remove the `pend if` lines.

- [x] **Enable** [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes

### DIFF
--- a/test/rubygems/test_gem_commands_exec_command.rb
+++ b/test/rubygems/test_gem_commands_exec_command.rb
@@ -216,7 +216,7 @@ class TestGemCommandsExecCommand < Gem::TestCase
 
   def test_gem_with_platform_and_platform_dependencies
     pend "extensions don't quite work on jruby" if Gem.java_platform?
-    pend "terminates on mswin" if Gem.win_platform? && ruby_repo?
+    pend "terminates on mswin" if vc_windows? && ruby_repo?
 
     spec_fetcher do |fetcher|
       fetcher.download "a", 2 do |s|
--- a/test/rubygems/test_gem_ext_builder.rb
+++ b/test/rubygems/test_gem_ext_builder.rb
@@ -105,7 +105,7 @@ install:
   end
 
   def test_build_extensions
-    pend if RUBY_PLATFORM.include?("mswin") && ENV.key?("GITHUB_ACTIONS") # not working from the beginning
+    pend "terminates on mswin" if vc_windows? && ruby_repo?
     @spec.extensions << "ext/extconf.rb"
 
     ext_dir = File.join @spec.gem_dir, "ext"
@@ -141,7 +141,7 @@ install:
   end
 
   def test_build_extensions_with_gemhome_with_space
-    pend if RUBY_PLATFORM.include?("mswin") && ENV.key?("GITHUB_ACTIONS") # not working from the beginning
+    pend "terminates on mswin" if vc_windows? && ruby_repo?
     new_gemhome = File.join @tempdir, "gem home"
     File.rename(@gemhome, new_gemhome)
     @gemhome = new_gemhome
@@ -162,7 +162,7 @@ install:
         false
       end
     end
-    pend if RUBY_PLATFORM.include?("mswin") && ENV.key?("GITHUB_ACTIONS") # not working from the beginning
+    pend "terminates on mswin" if vc_windows? && ruby_repo?
 
     @spec.extensions << "ext/extconf.rb"
 
--- a/test/rubygems/test_gem_installer.rb
+++ b/test/rubygems/test_gem_installer.rb
@@ -1496,7 +1496,7 @@ gem 'other', version
 
   def test_install_extension_and_script
     pend "Makefile creation crashes on jruby" if Gem.java_platform?
-    pend if RUBY_PLATFORM.include?("mswin") && ENV.key?("GITHUB_ACTIONS") # not working from the beginning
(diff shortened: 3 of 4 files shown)


MSP-Greg:
I was specifically looking at test/test_gem_ext_builder.rb.  There may be more tests can have mswin skips removed.  Can't look at immediately, but should be today.

test/rubygems/test_gem_commands_exec_command.rb
```diff
@@ -216,7 +216,6 @@ def test_gem_with_platform_dependencies
 
   def test_gem_with_platform_and_platform_dependencies
     pend "extensions don't quite work on jruby" if Gem.java_platform?
-    pend "terminates on mswin" if Gem.win_platform? && ruby_repo?
```
deivid-rodriguez, line 219: I'm not sure we can remove this one, since this was a ruby/ruby specific issue that I'm not sure it's related to this? See https://github.com/rubygems/rubygems/pull/6417#discussion_r1119573694 and comments below.

deivid-rodriguez (commented):
Thanks! A minor comment about one of the removals that I'm not sure it was related to this. The test was actually running here already and the skip was specific to ruby/ruby repo.

test/rubygems/test_gem_commands_exec_command.rb
deivid-rodriguez: It is confusing indeed. Thanks for double checking this!

MSP-Greg: It's running now...

MSP-Greg: I feel like posting a NSFW message, but...

Bunch of failures in ruby/ruby, all because they're being run from the build folder.  JFYI, a long time ago I decided to run all the tests in ruby-loco from the install folder, and I've suggested that ruby/ruby do the same.  Or, there's very little testing that verifies that `make install` makes a usable installation.  And there's plenty of 'hacky' code to make tests run both ways.  Anyway...

Question - many of the `pends` in the existing code used `ENV.key?("GITHUB_ACTIONS")`.  That will skip in any CI, not just ruby/ruby.  The run here verified that the tests work from RubyGems CI.  Also, `ruby_repo?` uses:
```
!ENV["GEM_COMMAND"].nil?
```
Are you confident that the above is never true in the CI here?  I don't know, but I'm can't think of any good way to determine if tests are being run from ruby/ruby.  Ideally, it should work for both local & in GHA.  I'm going to check a few things, do you have any thoughts?

If we want to ignore local testing, `ENV['GITHUB_REPOSITORY']` in GHA here would be 'rubygems/rubygems', in ruby/ruby it would (obviously) be 'ruby/ruby'.  We could use ENV['GITHUB_REPOSITORY'].end_with? "/ruby".  Would that work?  People could set it for local work.  If you're okay with that, should I change `ruby_repo?` or just the ENV test in the pend statements?

Sorry about the mess, but I hate skipping tests when they don't need to be...

deivid-rodriguez: > Are you confident that the above is never true in the CI here? I don't know, but I'm can't think of any good way to determine if tests are being run from ruby/ruby. Ideally, it should work for both local & in GHA. I'm going to check a few things, do you have any thoughts?

I'm pretty sure we're not setting `ENV["GEM_COMMAND"]` here, yes. So I think that works fine.

I think this PR is fine as is, just remove this particular change so we don't break ruby/ruby CI when this change is backported.

MSP-Greg: Thanks.  Updated everything.  Almost all (or maybe all) the tests fail in ruby/ruby, so rather than the following, which is skipped in **all GHA**:
```ruby
pend if RUBY_PLATFORM.include?("mswin") && ENV.key?("GITHUB_ACTIONS")
```

we're now using:
```ruby
pend "terminates on mswin" if vc_windows? && ruby_repo?
```

MSP-Greg: @deivid-rodriguez

I thought I'd take the updated version from here and apply it to ruby/ruby.  It did not work, so please don't merge.

I may fix tonite, or possibly tomorrow.  I'm -0500.

MSP-Greg: Fixed.  Passsing both here and ruby/ruby

deivid-rodriguez (approved):
Awesome thank you!

hsbt:
👍

hsbt:
@deivid-rodriguez I'd like to backport this and https://github.com/rubygems/rubygems/commit/dadedeb515b3528cbff21a2769d924261f5191c3 to `3.4` branch for Ruby 3.2.

deivid-rodriguez:
Weird, they should have been backported already since they are labeled as `rubygems: backport`. In any case, I will backport them to the next release 👍.

hsbt:
🙏 

>they are labeled as rubygems: backport

confirmed. I also not sure this PR is not backported.

PR #6537, opened 2023-03-24
Add gem version promoter specs

siegfault:

Pulled out some spec changes as a result of https://github.com/rubygems/rubygems/pull/5258#pullrequestreview-1214895689 (

> However, the specs added in the last commit raise some questions, since the current behavior feels quite surprising and I suspect we may want to tweak the implementation instead of codifying the current behavior in specs. I also have some minor comments about wording.

) and a bug fix mentioned in https://github.com/rubygems/rubygems/pull/5258#issuecomment-1366962554.

### Spec changes
Add tests for pre, move more of the setup into a helper method, and restructure tests.

There seem to be five considerations for these tests (level, pre, strict, locked, and whether the current version is a prerelease version, though the last one overlaps with pre and didn't seem to behave how I expected under test). Rather than write out the 16 (/32 if the last consideration is real) combinations, I wrote most with independent tests for each value. The existing combined tests were maintained (level vs strict) because these seem the most interrelated.

### DIFF
--- a/bundler/spec/bundler/gem_version_promoter_spec.rb
+++ b/bundler/spec/bundler/gem_version_promoter_spec.rb
@@ -1,175 +1,162 @@
 # frozen_string_literal: true
 
 RSpec.describe Bundler::GemVersionPromoter do
-  context "conservative resolver" do
-    def versions(result)
-      result.flatten.map(&:version).map(&:to_s)
-    end
+  let(:gvp) { described_class.new }
 
+  # Rightmost (highest array index) in result is most preferred.
+  # Leftmost (lowest array index) in result is least preferred.
+  # `build_candidates` has all versions of gem in index.
+  # `build_spec` is the version currently in the .lock file.
+  #
+  # In default (not strict) mode, all versions in the index will
+  # be returned, allowing Bundler the best chance to resolve all
+  # dependencies, but sometimes resulting in upgrades that some
+  # would not consider conservative.
+
+  describe "#sort_versions" do
     def build_candidates(versions)
       versions.map do |v|
         Bundler::Resolver::Candidate.new(v)
       end
     end
 
-    def build_spec_set(name, v)
-      Bundler::SpecSet.new(build_spec(name, v))
+    def build_package(name, version, locked = [])
+      Bundler::Resolver::Package.new(name, [], :locked_specs => Bundler::SpecSet.new(build_spec(name, version)), :unlock => locked)
     end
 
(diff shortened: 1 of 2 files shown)


siegfault:
@deivid-rodriguez the tests are more or less the same commit from that original PR. I made sure all the cases are still covered, and fixed some new kwargs that have been added since. Re: your comment, I wonder whether the current behavior feeling surprising is due to how I worded some of the tests w.r.t. 'favors'. Ultimately, what the GemVersionPromoter returns is flipped (I believe here: https://github.com/rubygems/rubygems/blob/849342b11d208e14fe0369b8f34891b44749d6e1/bundler/lib/bundler/resolver.rb#L313), so what's favored here is ultimately the opposite of what's selected/used. Very open to better wording to make this less confusing! As well as any other suggestions you had. Thank you!

bundler/spec/bundler/gem_version_promoter_spec.rb
```diff
@@ -1,175 +1,162 @@
...
-    end
+      context "when level is major" do
+        before { gvp.level = :major }
 
-    context "filter specs (strict) level minor" do
-      let(:gvp) do
-        Bundler::GemVersionPromoter.new.tap do |gvp|
-          gvp.level = :minor
-          gvp.strict = true
+        it "keeps downgrades" do
```
deivid-rodriguez, line 52: I think this is what was surprising to me. The other levels remove downgrades, but this one (the default), does not. I think this complicates things for Bundler outside of the gem version promoter because it needs to explicit add lower bound requirements to dependencies and I think it may be to workaround this.

Nothing to fix in this PR, but I think we may be able to remove a bit of complexity but making `level=major` behave like the other levels.

deivid-rodriguez (approved):
Looks awesome, thanks for following up on this!

simi (approved):

bundler/spec/bundler/gem_version_promoter_spec.rb
siegfault: Ah, gotcha. Yeah that had tripped me up a couple times. I'm not currently at the computer, but I wonder if changing https://github.com/rubygems/rubygems/blob/3586d716d0f8b6707afdafab37a6eb26033fffd2/bundler/lib/bundler/gem_version_promoter.rb#L80 with
```diff
-       return specs if locked_version.nil? || major?
+       return specs if locked_version.nil?
```
would fix that. Happy to try it either on this branch or a follow up!

PR #6544, opened 2023-03-24
Onboard Rubocop Naming/MemoizedInstanceVariableName rule

simi:

It is followup of https://github.com/rubygems/rubygems/pull/6420 to make it easier to prevent similar problems in the future.

Onboard related Rubocop rule with two exceptions.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2236,7 +2236,7 @@ class Gem::Specification < Gem::BasicSpecification
   # The platform this gem runs on.  See Gem::Platform for details.
 
   def platform
-    @new_platform ||= Gem::Platform::RUBY
+    @new_platform ||= Gem::Platform::RUBY # rubocop:disable Naming/MemoizedInstanceVariableName
   end
 
   def pretty_print(q) # :nodoc:
@@ -2711,6 +2711,8 @@ class Gem::Specification < Gem::BasicSpecification
     end
 
     @installed_by_version ||= nil
+
+    nil
   end
 
   def flatten_require_paths # :nodoc:
--- a/lib/rubygems/stub_specification.rb
+++ b/lib/rubygems/stub_specification.rb
@@ -182,7 +182,7 @@ class Gem::StubSpecification < Gem::BasicSpecification
   ##
   # The full Gem::Specification for this gem, loaded from evalling its gemspec
 
-  def to_spec
+  def spec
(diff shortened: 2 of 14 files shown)


lib/rubygems/specification.rb
```diff
@@ -2239,7 +2239,7 @@ def original_platform # :nodoc:
   # The platform this gem runs on.  See Gem::Platform for details.
 
   def platform
-    @new_platform ||= Gem::Platform::RUBY
+    @new_platform ||= Gem::Platform::RUBY # rubocop:disable Naming/MemoizedInstanceVariableName
```
simi, line 2242: There is already @platform used for different purpose. It could be refactored if needed (or later in other PR).

lib/rubygems/specification.rb
```diff
@@ -2718,6 +2718,8 @@ def reset_nil_attributes_to_default
     end
 
     @installed_by_version ||= nil
+
+    nil
```
simi, line 2722: `@installed_by_version` is just "initialized" in here, return value of this method is never used.

lib/rubygems/stub_specification.rb
```diff
@@ -182,14 +182,15 @@ def full_name
...
-  def to_spec
+  def spec
     @spec ||= if @data
       loaded = Gem.loaded_specs[name]
       loaded if loaded && loaded.version == version
     end
 
     @spec ||= Gem::Specification.load(loaded_from)
   end
+  alias_method :to_spec, :spec
```
simi, line 193: `@to_spec` var name seems strange to me, is it ok to add this alias instead?

util/release.rb
```diff
@@ -47,7 +47,7 @@ def create_for_github!
     end
 
     def previous_version
-      @latest_release ||= latest_release.tag_name.gsub(/^#{@tag_prefix}/, "")
+      @previous_version ||= latest_release.tag_name.gsub(/^#{@tag_prefix}/, "")
```
simi, line 50: Seems additional problem was located thanks to `Naming/MemoizedInstanceVariableName`. :muscle:

lib/rubygems/specification.rb
deivid-rodriguez: I'm happy either way!

deivid-rodriguez:
Looks good to me @simi!

deivid-rodriguez (approved):

Issue #6242, opened 2023-01-06
Add a new entrypoint for bundler that automatically bundle installs


ngan:
### Describe the problem as clearly as you can

We work in a large monolith and gems get bumped all the time. This means I find myself `bundle install`ing many times a day. And oftentimes, this is done after seeing a bunch of errors because, well my bundle is out of date.

Would it make sense to introduce another entry similar to `bundler/setup`? Eg...
```ruby
# instead of...
require "bundler/setup"
# you can do...
require "bundler/prepare"
```

This would automatically install your bundle if it's out of date. We can also take `BUNDLE_DEPLOYMENT` and `BUNDLE_FROZEN` into consideration. Maybe if deployment is enabled, it wouldn't auto install, and if frozen, it'd err if lockfile changes.

Separately, maybe Rails could use this in `config/boot.rb` such that `rails s` or `rails c` would just work. Binstubs would benefit from this as well? We have githooks that use the bundle and we would switch to `bundler/prepare`.

Happy to put up PR, but wanted to get feedback first.

indirect:
Is there a reason to build this into Bundler itself, rather set up a tool like [tpope/hookup](https://github.com/tpope/hookup), which handles more than just `bundle install`?

deivid-rodriguez:
By the way, we actually have a `bundle config auto_install true` setting that does this, but it only applies to selected CLI commands: https://github.com/rubygems/rubygems/blob/131085f1afeea7b76ec5c9fb3b37776d0163ac0e/bundler/lib/bundler/cli.rb#L11.

That doesn't fit your use case, since you want it for direct require of `bundle/exec`, but I wanted to mention it in case it's helpful and also ask how do you feel about @indirect's suggestion.

ngan:
Hey! Sorry for the silence. I forgot about this. Hookup and other tools like this is…okay. In our experience, githooks are generally difficult to deal with. I was hoping bundle would consider taking care of this out of the box since it’s a very easy process. People who need more functionality can reach for githook tools. But this would satisfy the needs of most projects I suspect.

indirect:
Can you say more about how you would want to use this? As David pointed out, we do already support auto-install for bundle exec if you set the config or env var. Do you need more than that?

ngan:
[...]

PR #6561, opened 2023-03-28
Add auto_install support to require "bundler/setup"

technicalpickles:
We have some places that already use `bundle config auto_install true`, ie:

https://github.com/technicalpickles/rubygems/blob/7a144f3374f6a400cc9832f072dc1fc0bca8c724/bundler/lib/bundler/cli.rb#L11

This applies the same logic (copy and pasted) to happen when you `require "bundler/setup"`.

As described in https://github.com/rubygems/rubygems/issues/6242

> We work in a large monolith and gems get bumped all the time. This means I find myself bundle installing many times a day. And oftentimes, this is done after seeing a bunch of errors because, well my bundle is out of date.

@indirect suggested adding `auto_install` support to `require 'bundler/setup', so this is an attempt at that.

_edit_ updated to reflect current implementation

This PR adds a `Bundle.auto_install` to handle the logic of:
- checking that auto_install is configured
- try loading the specs to see if install is needed
- delegate to the `Bundler::CLI::Install` to do the actual installation

Existing uses of auto_install are updated to use this

The reason we settled on using `Bundler::CLI::Install` is because it has quite a bit of logic, including things like [updating the app cache](https://github.com/rubygems/rubygems/blob/7316186db83b87eaa1d6f81531515a5abb5b1f69/bundler/lib/bundler/cli/install.rb#L65-L67)

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -40,6 +40,7 @@ module Bundler
   SUDO_MUTEX = Thread::Mutex.new
 
   autoload :Checksum,               File.expand_path("bundler/checksum", __dir__)
+  autoload :CLI,                    File.expand_path("bundler/cli", __dir__)
   autoload :CIDetector,             File.expand_path("bundler/ci_detector", __dir__)
   autoload :Definition,             File.expand_path("bundler/definition", __dir__)
   autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
@@ -162,6 +163,25 @@ module Bundler
         @setup = load.setup
       else
         load.setup(*groups)
+      end
+    end
+
+    # Automatically install dependencies if Bundler.settings[:auto_install] exists.
+    # This is set through config cmd `bundle config set --global auto_install 1`.
+    #
+    # Note that this method `nil`s out the global Definition object, so it
+    # should be called first, before you instantiate anything like an
+    # `Installer` that'll keep a reference to the old one instead.
+    def auto_install
+      return unless settings[:auto_install]
+
+      begin
+        definition.specs
+      rescue GemNotFound, GitError
+        ui.info "Automatically installing missing gems."
+        reset!
+        CLI::Install.new({}).run
+        reset!
       end
     end
 
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -5,6 +5,7 @@ require_relative "vendored_thor"
 module Bundler
   class CLI < Thor
     require_relative "cli/common"
+    require_relative "cli/install"
 
     package_name "Bundler"
 
@@ -69,7 +70,7 @@ module Bundler
       Bundler.settings.set_command_option_if_given :retry, options[:retry]
 
       current_cmd = args.last[:current_command].name
-      auto_install if AUTO_INSTALL_CMDS.include?(current_cmd)
+      Bundler.auto_install if AUTO_INSTALL_CMDS.include?(current_cmd)
     rescue UnknownArgumentError => e
       raise InvalidOption, e.message
     ensure
@@ -736,26 +737,6 @@ module Bundler
     end
 
     private
-
-    # Automatically invoke `bundle install` and resume if
-    # Bundler.settings[:auto_install] exists. This is set through config cmd
-    # `bundle config set --global auto_install 1`.
-    #
-    # Note that this method `nil`s out the global Definition object, so it
-    # should be called first, before you instantiate anything like an
-    # `Installer` that'll keep a reference to the old one instead.
-    def auto_install
-      return unless Bundler.settings[:auto_install]
-
-      begin
-        Bundler.definition.specs
-      rescue GemNotFound, GitError
-        Bundler.ui.info "Automatically installing missing gems."
-        Bundler.reset!
-        invoke :install, []
-        Bundler.reset!
-      end
-    end
 
     def current_command
       _, _, config = @_initializer
--- a/bundler/lib/bundler/setup.rb
+++ b/bundler/lib/bundler/setup.rb
@@ -5,6 +5,9 @@ require_relative "shared_helpers"
 if Bundler::SharedHelpers.in_bundle?
   require_relative "../bundler"
 
+  # try to auto_install first before we get to the `Bundler.ui.silence`, so user knows what is happening
+  Bundler.auto_install
+
   if STDOUT.tty? || ENV["BUNDLER_FORCE_TTY"]
     begin
       Bundler.ui.silence { Bundler.setup }
--- a/bundler/spec/runtime/setup_spec.rb
+++ b/bundler/spec/runtime/setup_spec.rb
@@ -1599,4 +1599,19 @@ end
     sys_exec "#{Gem.ruby} #{script}", raise_on_error: false
     expect(out).to include("requiring foo used the monkeypatch")
   end
+
+  it "performs an automatic bundle install" do
+    gemfile <<-G
+      source "#{file_uri_for(gem_repo1)}"
+      gem "rack", :group => :test
+    G
+
+    bundle "config set auto_install 1"
+
+    ruby <<-RUBY
+      require 'bundler/setup'
+    RUBY
+    expect(err).to be_empty
+    expect(out).to include("Installing rack 1.0.0")
+  end
 end


bundler/lib/bundler/cli.rb
```diff
@@ -802,16 +802,7 @@ def self.deprecated_ext_value?(arguments)
     # should be called first, before you instantiate anything like an
     # `Installer` that'll keep a reference to the old one instead.
     def auto_install
-      return unless Bundler.settings[:auto_install]
```
deivid-rodriguez, line 805: The description of this method still mentions `Bundler.settings[:auto_install]`, so I think it makes sense to keep this guard clause and remove it from `Bundler.auto_install`.

deivid-rodriguez (commented):
This is looking very good now! I think this will be a neat feature for big repos with many users working simultaneously.

Can we add some specs?

bundler/lib/bundler/setup.rb
```diff
@@ -5,6 +5,10 @@
 if Bundler::SharedHelpers.in_bundle?
   require_relative "../bundler"
 
+  if Bundler.settings[:auto_install]
+    Bundler.auto_install
+  end
```
technicalpickles, line 10: Was chatting with @ngan and he suggested this should go into `Bundler.setup`. That would mean you get that anytime you use `Bundler.setup`, not just `require "bundler/setup"`

Does that seem reasonable @deivid-rodriguez ?

deivid-rodriguez: That seems fine to me, yes.

bundler/lib/bundler.rb
```diff
@@ -165,6 +169,22 @@ def setup(*groups)
       end
     end
 
+    def install
+      Bundler.definition.specs
+    rescue GemNotFound
+      Bundler.ui.info "Automatically installing missing gems."
```
technicalpickles, line 175: While testing, we were surprised to not see this output when using `require 'bundler/setup'`, but not when calling `bundle exec` or other subcommand that support `auto_install`.

Turns out `require 'bundler/setup'` explicitly does a `Bundler.ui.silence` to silence everything.

I think ideally, this should be shown.

technicalpickles (commented):
Did some pairing with @ngan on this, and feel pretty good on the implementation. That said, some issues we ran into:

- we're seeing `auto_install: true` being output when it's enabled. that is causing the realworld tests to fail
- The "Automatically installing gems" is shown when running `bundle exec`, but not in `bundle/setup`. I think that is okay for now, but ideally would
- We found a separate bug while testing against our app, when a bundle is locked to an different version of a builtin gem, that bundler also uses... psych in our case

Still need to add / move some specs around.

technicalpickles:
This is ready for review! I'm feeling pretty good about the implementation, and was able to add a spec to cover it.

technicalpickles:
Not sure why `system-rubygems-bundler / Bundler 2 against system Rubygems (ruby-2.6)` is failing, but I'm able to reproduce it locally 🤔  

The paths I'm seeing kinda make it seem like it's running against my system bundler, vs the code in the repo, though.

```
     RuntimeError:


       Commands:
       $ /Users/josh.nichols/.rbenv/versions/2.6.10/bin/ruby -I/Users/josh.nichols/workspace/rubygems/bundler/spec -r/Users/josh.nichols/workspace/rubygems/bundler/spec/support/artifice/fail.rb -r/Users/josh.nichols/workspace/rubygems/bundler/spec/support/hax.rb /Users/josh.nichols/workspace/rubygems/bundler/tmp/1/gems/system/bin/bundle config set auto_install 1
       # $? => 0

       Invoking `/Users/josh.nichols/.rbenv/versions/2.6.10/bin/ruby -r/Users/josh.nichols/workspace/rubygems/bundler/spec/support/hax.rb -w -e \ \ \ \ \ \ require\ \'bundler\''
       '\ \ \ \ \ \ Bundler.setup'
       '` failed with output:
       /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/resolver.rb:287:in `block in verify_gemfile_dependencies_are_found!': Could not find gem 'rack' in any of the gem sources listed in your Gemfile. (Bundler::GemNotFound)
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/resolver.rb:255:in `each'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/resolver.rb:255:in `verify_gemfile_dependencies_are_found!'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/resolver.rb:49:in `start'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/resolver.rb:22:in `resolve'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/definition.rb:258:in `resolve'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/definition.rb:170:in `specs'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/definition.rb:237:in `specs_for'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/definition.rb:226:in `requested_specs'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/runtime.rb:108:in `block in definition_method'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler/runtime.rb:20:in `setup'
       	from /Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/bundler.rb:107:in `setup'
       	from -e:2:in `<main>'
```

I would have expected to see `/Users/josh.nichols/workspace/rubygems/bundler/` in the last part of the stack trace rather than `/Users/josh.nichols/.rbenv/versions/2.6.10/lib/ruby/2.6.0/`

technicalpickles:
Trying to pick this up again, so I just merged master and resolved the merge conflict. It's only failing on [Bundler 2 against system Rubygems (ruby-2.6)](https://github.com/rubygems/rubygems/actions/runs/5872732044/job/15924799212#logs), but passes everywhere else 🤔 

It is reproducible locally fortunately. I'm not sure I understand the differences between the different jobs. I get the ruby version, but I don't know what the 2 vs 3 refers too, and why it could contribute to this failure.

bundler/lib/bundler.rb
```diff
@@ -167,6 +171,22 @@ def setup(*groups)
       end
     end
 
+    def install
+      Bundler.definition.specs
+    rescue GemNotFound, GitError
+      Bundler.ui.info "Automatically installing missing gems."
```
technicalpickles, line 177: I was thinking through if this would ever show up when it wasn't "automatic", and I don't think so. `bundle install` calls `Bundler::Installer.install` the same as this method does, so this should be the only time it happens automatically.

bundler/lib/bundler.rb
deivid-rodriguez: Yeah! I was actually surprised by this in the past too, particularly when expecting to see "Resolving gems..." output during `require "bundler/setup"` and it was not there.

If I recall correctly, we couldn't figure out the best path forward, because while it does make sense to me that `require "bundler/setup"` is more silent and does not show messages like "Resolving gems...", I agree we would not want to hide output indicating that it's actually installing gems.

Should we only silence `bundler/setup` output only up to a certain log level?

bundler/lib/bundler.rb
deivid-rodriguez: I think you're right!

bundler/lib/bundler.rb
technicalpickles: That is an option, but at the same time, I'm not sure what level different log messages are at, or if there would be a level to represent what we want shown 🤔 

I do have this comment calling out that it would normally be silenced:

```
  # this happens in Bundler.setup too, but we call it here early
  # so it can happen without being Bundler.ui.silence'd
```

deivid-rodriguez: Oh, sorry! So you ended up resolving this in the end. Then this is fine, and we could iterate on a better solution and the comment you left there will be useful for that 👍. Should we add an assertion to the test that checks that the message is printed?

technicalpickles: Maybe could do something like... remove the `rescue` logic in the method that does the installation, and hoist it to call sites. Or make it a conditional that can be checked

```
if auto_install_needed?
  install
end


def auto_install_needed?
  return false unless settings[:auto_install]
  begin
    definition.specs
    true
  rescue GemNotFound, GitError
     false
  end
end
```

deivid-rodriguez: That seems fine to me, but at this point would be just cleaner code but would not avoid having to do this both before and during `Bundler.setup`?

deivid-rodriguez: As minor additional suggestion (besides checking that the log message is printed in the spec), you could remove `Bundler.` prefixes since this now lives inside the `Bundler` module.

technicalpickles:
I started moving the logic for auto-install around like I suggested [on slack](https://bundler.slack.com/archives/C08V1RPAP/p1702408823566069?thread_ts=1697726138.574379&cid=C08V1RPAP) for `bundler/setup`.

```ruby
    retried = false
    begin
      Bundler.ui.silence { Bundler.setup }
    rescue Bundler::BundlerError => e
      if !retried && Bundler.settings[:auto_install] && (e.is_a?(Bundler::GemNotFound) || e.is_a?(Bundler::GitError))
        retried = true
        Bundler.auto_install
​
        retry
      end
​
      Bundler.ui.error e.message
      Bundler.ui.warn e.backtrace.join("\n") if ENV["DEBUG"]
      if e.is_a?(Bundler::GemNotFound)
        suggested_cmd = "bundle install"
        original_gemfile = Bundler.original_env["BUNDLE_GEMFILE"]
        suggested_cmd += " --gemfile #{original_gemfile}" if original_gemfile
        Bundler.ui.warn "Run `#{suggested_cmd}` to install missing gems."
      end
      exit e.status_code
    end
```

When my `spec/runtime/setup_spec.rb` started failing, I realized I this spec uses `Bundler.setup` rather than `require 'bundler/setup'`

I suspect this 

---

Stepping back a second, the things I'm thinking about are:

- `require 'bundler/setup'` will install missing dependencies with `settings[:auto_install]`
- `Bundler.setup` will install missing dependencies with `settings[:auto_install]`
- when auto-install happens, it's shown to the user so they know why something is taking awhile

I believe this PR accomplishes that. The only kinda questionable thing is in `bundler/setup`, we try to auto-install before calling `Bundler.setup` because `Bundler.setup` is wrapped with `Bundler.ui.silence`, which would hide the install output. `Bundler.setup` then also tries to auto-install, because users may do that instead of bundler/setup, and I want them to get it. Technically, we have this in two places, but it should be a no-op when it gets to the second one.

What do y'all think?

technicalpickles:
Also, not sure what this new failure is about for [Bundler under a ruby-core setup](https://github.com/rubygems/rubygems/actions/runs/7196790378/job/19602410433#logs) 🤔  

```
  No such file or directory @ rb_sysopen - exe/bundle
```

deivid-rodriguez:
Here's what I'm thinking:

You initially aimed to add support for this only to `require "bundler/setup"`, but then @ngan pointed out that it should live inside `Bundler.setup` and I said it seemed fine at https://github.com/rubygems/rubygems/pull/6561#discussion_r1248114146.

Now we have detected that in order for this feature to work fine in both cases we need to try to auto-install both before and during `Bundler.setup` which indeed feels a bit questionable although I agree it should not cause any problems.

That said, I think there's no strong motivation for us to support this also inside `Bundler.setup` (at least not yet?). The way I currently see these implemented is that `require "bundler/setup"` is just a wrapper around `Bundler.setup` which some UX improvements like hiding some info messages, or showing a friendly error when gems are missing. I think this feature could fit just fine as one of those UX enhancements over `Bundler.setup`, which just deals with setting up the bundle.

In any case, I don't have a strong opinion either way, and I'm willing to accept both approaches.

I wonder if other maintainers have some opinions here?

Regarding that failure you mentioned, a rebased should do the trick there. CI in master was temporarily broken due to that today, but it's fine now.

technicalpickles:
>  I think there's no strong motivation for us to support this also inside Bundler.setup (at least not yet?). The way I currently see these implemented is that require "bundler/setup" is just a wrapper around Bundler.setup which some UX improvements like hiding some info messages, or showing a friendly error when gems are missing.

One of the motivations was handling binstubs that are called during a pre-commit hook, ie `bin/rubocop`. I confirmed that they are using `require "bundler/setup"` rather than `Bundler.setup`.

If we do move it to just bundler/setup, I'm not sure if `spec/runtime/setup_spec.rb` is the right place to be testing.

technicalpickles:
> If we do move it to just bundler/setup, I'm not sure if spec/runtime/setup_spec.rb is the right place to be testing.

I looked more closely, and there are places calling `require 'bundler/setup'`, so this is fine.

I've also added `spec/runtime/auto_install` which is largely the same setup / assertions, except it is calling `Bundler.auto_install` explicitly.

My thought is that a user wanted to have direct control, they could do `Bundler.auto_install` then `Bundler.setup` on their own.

@deivid-rodriguez with that, I think we are good for another review 😅

deivid-rodriguez:
@technicalpickles I was having one more look here and I believe this is actually changing behavior by no fully invoking the install command as before. For example, it will no longer [keep app cache up to date ](https://github.com/rubygems/rubygems/blob/7316186db83b87eaa1d6f81531515a5abb5b1f69/bundler/lib/bundler/cli/install.rb#L65-L67)?

I think we could replace the previous `invoke :install, []` with directly instantiating and running `CLI::Install`? Something like this?

```diff
diff --git a/bundler/lib/bundler.rb b/bundler/lib/bundler.rb
index 7e84fb1892..59a1107bb7 100644
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -165,25 +165,6 @@ def setup(*groups)
       end
     end
 
-    # Automatically installs missing gems from Gemfile.
-    #
-    # To setup the Bundler runtime, see +Bundler.setup+.
-    def auto_install
-      definition.specs
-    rescue GemNotFound, GitError
-      ui.info "Automatically installing missing gems."
-      reset!
-
-      settings.temporary(no_install: false) do
-        Plugin.gemfile_install(default_gemfile) if feature_flag.plugins?
-
-        definition = self.definition
-        definition.validate_runtime!
-
-        Installer.install(root, definition, {})
-      end
-    end
-
     # Setups Bundler environment (see Bundler.setup) if it is not already set,
     # and loads all gems from groups specified. Unlike ::setup, can be called
     # multiple times with different groups (if they were allowed by setup).
diff --git a/bundler/lib/bundler/cli.rb b/bundler/lib/bundler/cli.rb
index dcdba1277f..e3b2c94f03 100644
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
 
 require_relative "vendored_thor"
+require_relative "../bundler"
 
 module Bundler
   class CLI < Thor
     require_relative "cli/common"
+    require_relative "cli/install"
 
     package_name "Bundler"
 
@@ -55,6 +57,25 @@ def self.aliases_for(command_name)
       COMMAND_ALIASES.select {|k, _| k == command_name }.invert
     end
 
+    # Automatically install dependencies if Bundler.settings[:auto_install] exists.
+    # This is set through config cmd `bundle config set --global auto_install 1`.
+    #
+    # Note that this method `nil`s out the global Definition object, so it
+    # should be called first, before you instantiate anything like an
+    # `Installer` that'll keep a reference to the old one instead.
+    def self.auto_install
+      return unless Bundler.settings[:auto_install]
+
+      begin
+        Bundler.definition.specs
+      rescue GemNotFound, GitError
+        Bundler.ui.info "Automatically installing missing gems."
+        Bundler.reset!
+        Bundler::CLI::Install.new({}).run
+        Bundler.reset!
+      end
+    end
+
     def initialize(*args)
       super
 
@@ -69,7 +90,7 @@ def initialize(*args)
       Bundler.settings.set_command_option_if_given :retry, options[:retry]
 
       current_cmd = args.last[:current_command].name
-      auto_install if AUTO_INSTALL_CMDS.include?(current_cmd)
+      self.class.auto_install if AUTO_INSTALL_CMDS.include?(current_cmd)
     rescue UnknownArgumentError => e
       raise InvalidOption, e.message
     ensure
@@ -735,18 +756,6 @@ def self.deprecated_ext_value?(arguments)
 
     private
 
-    # Automatically install dependencies if Bundler.settings[:auto_install] exists.
-    # This is set through config cmd `bundle config set --global auto_install 1`.
-    #
-    # Note that this method `nil`s out the global Definition object, so it
-    # should be called first, before you instantiate anything like an
-    # `Installer` that'll keep a reference to the old one instead.
-    def auto_install
-      return unless Bundler.settings[:auto_install]
-
-      Bundler.auto_install
-    end
-
     def current_command
       _, _, config = @_initializer
       config[:current_command]
diff --git a/bundler/lib/bundler/setup.rb b/bundler/lib/bundler/setup.rb
index 917336061a..020e900c4d 100644
--- a/bundler/lib/bundler/setup.rb
+++ b/bundler/lib/bundler/setup.rb
@@ -3,12 +3,10 @@
 require_relative "shared_helpers"
 
 if Bundler::SharedHelpers.in_bundle?
-  require_relative "../bundler"
+  require_relative "cli"
 
   # try to auto_install first before we get to the `Bundler.ui.silence`, so user knows what is happening
-  if Bundler.settings[:auto_install]
-    Bundler.auto_install
-  end
+  Bundler::CLI.auto_install
 
   if STDOUT.tty? || ENV["BUNDLER_FORCE_TTY"]
     begin
diff --git a/bundler/spec/runtime/auto_install_spec.rb b/bundler/spec/runtime/auto_install_spec.rb
deleted file mode 100644
index ab817c561c..0000000000
--- a/bundler/spec/runtime/auto_install_spec.rb
+++ /dev/null
@@ -1,24 +0,0 @@
-# frozen_string_literal: true
-
-RSpec.describe "Bundler.auto_install" do
-  describe "with a gemfile" do
-    before do
-      gemfile <<-G
-        source "#{file_uri_for(gem_repo1)}"
-        gem "rack", :group => :test
-      G
-    end
-
-    it "installs the gems" do
-      # NOTE: only require bundler to make the method calls. spec/runtime/setup_spec.rb covers requiring bundler/setup
-      ruby <<-RUBY
-        require 'bundler'
-        Bundler.auto_install
-        Bundler.setup
-      RUBY
-
-      expect(err).to be_empty
-      expect(out).to include("Installing rack 1.0.0")
-    end
-  end
-end
```

deivid-rodriguez:
@technicalpickles Just wanted to check back on this. If you didn't have time yet, it's ok, just wanted to make sure you did not forget about it :)

technicalpickles:
I did not actually see it, so thanks for the ping 😅

bundler/lib/bundler/setup.rb
```diff
@@ -3,7 +3,10 @@
 require_relative "shared_helpers"
 
 if Bundler::SharedHelpers.in_bundle?
-  require_relative "../bundler"
+  require_relative "cli"
+
+  # try to auto_install first before we get to the `Bundler.ui.silence`, so user knows what is happening
+  Bundler::CLI.auto_install
```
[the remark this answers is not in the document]
technicalpickles: That is how it was d93dfb2 😬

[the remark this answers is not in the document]
deivid-rodriguez: Yes, that's what I meant. It does feel slightly better than before to me.

bundler/lib/bundler.rb
```diff
@@ -40,6 +40,7 @@ module Bundler
   SUDO_MUTEX = Thread::Mutex.new
 
   autoload :Checksum,               File.expand_path("bundler/checksum", __dir__)
+  autoload :CLI,                    File.expand_path("bundler/cli", __dir__)
```
technicalpickles, line 43: I have a hunch this was intentionally excluded 😨 

An alternative to this would be adding `require "bundler/cli"` down in `auto_install`

technicalpickles: @deivid-rodriguez do you have an opinion either way?

- a) add `CLI` to the list of autoloaded consts
- b) `require 'bundler/cli'` when it's needed during the autoinstall

deivid-rodriguez: I think I'm good with `autoload` 👍.

technicalpickles: Awesome, sounds good. Marking this comment resolved, but let me know if there's anything else needed to get in shape to merge!

deivid-rodriguez (approved):
This looks good to me, but I'll leave it opened for some other potential reviews. Particularly, in case there's some better code organization approach that I'm not seeing!

technicalpickles:
I've updated the PR body to describe the current implementation and decisions 🙇🏻

deivid-rodriguez:
Great! One potential improvement that I was thinking of is that `Bundler::CLI::Install` is "just a class" that does not really depend on `thor` or actually loading `Bundler::CLI`. The problem is that this class is defined under the `Bundler::CLI` class, so can't be loaded without `Bundler::CLI` being loaded first. To allow loading `Bundler::CLI::Install` without loading the full CLI, we could:

* Make sure explicitly that `Bundler::CLI` is exists before loading `Bundler::CLI::Install` (just opening the class as empty).
* Move all commands to a separate module `Bundler::Commands`, so that they don't require `Bundler::CLI` defined.

But it feels a bit overkill just to make this solution feel a bit better.

technicalpickles:
> One potential improvement that I was thinking of is that Bundler::CLI::Install is "just a class" that does not really depend on thor or actually loading Bundler::CLI

Ah! I missed that, and it's good to know. I was concerned loading the constant might trigger something else.

> Move all commands to a separate module Bundler::Commands, so that they don't require Bundler::CLI defined.
This feels like a better approach. Maybe `Bundler::Actions`? Thinking of actions as something a user does, etc.

> But it feels a bit overkill just to make this solution feel a bit better.

Agreed. I don't mind doing more work around this, but I would get this improvement out first.

simi (approved):

simi:
Looks good. I'm just wondering why `require_relative "../bundler"` is needed in other Bundler modules/classes, but that's not related to this change at all and used widely already in the codebase.

deivid-rodriguez:
Actually, is `require_relative "../bundler"` necessary here? I think so far we only use it in entrypoints (`bundler/setup`, `bundler/inline`, etc) that depend on stuff defined directly under the `Bundler` module.

technicalpickles:
> Actually, is require_relative "../bundler" necessary here? 

I don't think so anymore. Checked blame, and it came in with d93dfb2 ... and it was needed because of when how it was loaded before. Pushed up removing it to see how it looks.

simi:
@technicalpickles change seems effective, would you mind to rebase before merge?

technicalpickles:
@simi took a few tries, but cleanly rebased 😁

PR #6564, opened 2023-03-29
Use TestCase#util_spec instead of custom code.

voxik:
This is just small refactoring. I don't think this really changes that much.

This is extracted from #2909 to make it lighter.


deivid-rodriguez (approved):
Thanks!

simi (approved):

Issue #6403, opened 2023-02-23
Can we add support for running generic Ruby scripts?


ioquatix:
https://github.com/rubygems/rubygems/blob/451ac56078a3deaacef640906070d3d8f7502570/lib/rubygems/ext/builder.rb#L122-L138

Can we add `/.rb/` or something that matches scripts ending in `.rb` and just run them with the Ruby interpreter? The current behaviour is inflexible.

simi:
I have seen `extconf.rb` used to generate `Makefile` to just run custom script. Something like https://stackoverflow.com/a/38912965/319233.

What is your usecase @ioquatix?

jchestershopify:
I am not a fan. Install scripts are a recurring cause of security issues, particularly credential harvesting.

ioquatix:
@jchestershopify I agree with you, but according to the current interface, that is already possible. The only solution to that would be to use a sandbox for building native extensions. Which, I think is probably a good idea, but would be a totally separate issue.

Assuming that we just create a `makefile` which invokes a Ruby script, that's not practically different from what I'm proposing here, just that it's a simplification and it's more explicit. Otherwise, as @simi said, I can just use a hack to execute the code.

My use case is to use custom build tools, i.e. code which doesn't use any of the existing mechanisms.

simi:
@ioquatix what kind of build tools? `cmake`, `make`, `cargo` are used to build native extensions. Is it possible to build native extension using `nix` or `npm`? I would like to prevent easy way of coupling multiple package managers together.

PR #6569, opened 2023-04-03
Add support for executing a script directly.

ioquatix:

I don't want to use any of the existing builders for compiling a native extension, because I want to use a build tool which is not in the list and I don't want to write a rakefile to wrap it either.

This enables `spec.extensions = ["my_script.rb"]`. That script will be executed with the appropriate environment variables.


ioquatix:
The original issue was moved to a discussion: https://github.com/orgs/rubygems/discussions/6422

lib/rubygems/ext/builder.rb
```diff
@@ -146,6 +146,7 @@ def initialize(spec, build_args = spec.build_args, target_rbconfig = Gem.target_
 
   def builder_for(extension) # :nodoc:
     case extension
+    # File that ends with .rb
```
ioquatix, line 149:
```suggestion
```

ioquatix:
@simi you closed the issue #6403, don't reply on the discussion #6422, what's the path forward here?

I've got a smaller to address a subset of this issue: https://github.com/rubygems/rubygems/pull/8879

But it still doesn't address the full scope of the problem.

hsbt:
I decide to not support this feature. Please use `Gem::Ext::RakeBuilder`.

PR #6575, opened 2023-04-04
Don't need to use dtolnay/rust-toolchain for rustfmt for Windows

hsbt:

Runner images of GitHub Actions have rustfmt. We shouldn't install ourselves if it works.

https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#packages

Removed 3rd party setup actions.


hsbt:
I'm not sure why we need to install rustfmt. There is no code to use it in current rubygems.

ianks:
`rustfmt` was previously needed, but is not anymore ✅

hsbt:
@ianks Thanks for additional information 🙏

PR #6578, opened 2023-04-04
Remove one fallback to full indexes on big gemfiles

deivid-rodriguez:

If Gemfile has a lot of dependencies, we have an optimization that uses the full index in that case, assuming it's going to be faster.
    
I think this is an old optimization that predates compact index API times, I believe we no longer need it these days.
    
Also, since a few releases ago we [check for circular dependencies](https://github.com/rubygems/rubygems/pull/6330) when resolving by looping through all versions of each name and removing those that have circular dependencies that would trip up the resolver.

This loop becomes actually very slow when full indexes are used because to find dependencies of a gemspec, we need to explicitly fetch the marshaled gemspec (`gemspec.rz` endpoint) for it, so the optimization has the opposite effect of making things very slow.

Consistently use dependency APIs without special casing Gemfiles with a lot of dependencies.


stanhu:
Thanks for this! I was just investigating why our `bundle install` invocations were taking so long and saw that we had exceeded the 500 gems. When I bumped up `API_REQUEST_LIMIT` to 1000, I didn't see any issues. 

Would you be the right person to release an update for this?

deivid-rodriguez:
Currently preparing releases 👍.

stanhu:
Thank you!

PR #6581, opened 2023-04-05
Extract Windows CI of bundler for reducing waiting time of merge queue

hsbt:

Windows CI of bundler needs to execution time over the two hours. We should avoid waiting 4 hours for single pull-request.

I extracted Windows platform from `bundler.yml` and removed `merge_group` tag. I know this has some of duplicated code of YAML. But this is simple matrix rather than current one.

----

@deivid-rodriguez @simi 

How about this? Should we remove merge queue instead of this?

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -8,7 +8,7 @@ concurrency:
   group: ci-${{ github.ref }}-${{ github.workflow }}
   cancel-in-progress: true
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I'd remove the merge queue instead, indeed. I think it's solving a problem (semantic conflicts on consecutive merges) that we didn't have in practice, and creating others (slower development pace) that do affect us.

hsbt:
👌 I'll disabled `merge_queue` instead of this.

PR #6596, opened 2023-04-10
Prepare RubyGems 3.4.11 and Bundler 2.4.11

deivid-rodriguez:
It's release day!


deivid-rodriguez:
There was an issue with the changelog generation script. I fixed it at #6597 and cherry-picked the fix here so that I could generate correct changelogs.

Drmostafa20 (commented):
Nice

PR #6605, opened 2023-04-12
Show webauthn warning if default browser is Safari

jenshenny:

From: https://github.com/rubygems/rubygems.org/pull/3674#issuecomment-1494447858
Enhancement of: https://github.com/rubygems/rubygems/pull/6560

After discovering that Safari does not allow connection to localhost https://bugs.webkit.org/show_bug.cgi?id=186039, we're displaying browser warnings to say that Safari isn't support with the feature (https://github.com/rubygems/rubygems.org/pull/3674).

It would be nice to have a warning in the command line mentioning that the user's default browser doesn't support this feature.

## What is your fix for the problem, implemented in this PR

Checked the browser using LaunchServices plist by using `plutil` to retrieve the list and parse it to json. If the application assigned to the https scheme is safari, an error is displayed.

The alternative would be read the LSHandlers and to grab the browser using sed eg.
```
defaults read com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers \
  | sed -n -e '/LSHandlerURLScheme = https;/{x;p;d;}' -e 's/.*=[^"]"\(.*\)";/\1/g' -e x
```
## Testing
1. Add a security device to the user
2. Set your default browser to Safari
3. When `gem signin` is run, a warning is displayed that the default browser isn't supported
4. Set your default browser to another browser and the warning won't be displayed

Screenshot 2023-04-06 at 9 53 42 AM

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -275,6 +275,11 @@ module Gem::GemcutterUtilities
     url_with_port = "#{webauthn_url}?port=#{port}"
     say "You have enabled multi-factor authentication. Please visit #{url_with_port} to authenticate via security device. If you can't verify using WebAuthn but have OTP enabled, you can re-run the gem signin command with the `--otp [your_code]` option."
 
+    if default_browser_safari?
+      say "\n[WARNING] It looks like your default browser is Safari. Due to limitations within Safari, you will be unable to " \
+        "authenticate using this browser. Please visit the link in another browser."
+    end
+
     thread.join
     if error = thread[:error]
       alert_error error.message
@@ -294,6 +299,18 @@ module Gem::GemcutterUtilities
       end
     end
     response.is_a?(Net::HTTPSuccess) ? response.body : nil
+  end
+
+  def default_browser_safari?
+    return false unless RUBY_PLATFORM.include?("darwin")
+
+    require "json"
+
+    launch_services = JSON.parse(`defaults read com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers \
+      | plutil -convert json -o - -`)
+    default_browser = launch_services.find {|handler| handler["LSHandlerURLScheme"] == "https" }&.[]("LSHandlerRoleAll")
+
+    default_browser.nil? ? false : default_browser.include?("safari")
   end
 
   def pretty_host(host)
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -1491,6 +1491,23 @@ Also, a list:
     "#{__dir__}/#{key_name}_key.pem"
   end
 
+  ##
+  # Sets default browser to Safari in macOS by saving current configuration and overwriting LSHandlers
+
+  def set_default_browser_safari
+    @launch_services = `defaults read com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers`
+    default_browser_setting = "{ LSHandlerURLScheme = \"https\"; LSHandlerRoleAll = \"com.apple.safari\"; }"
+
+    `defaults write com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers -array '#{default_browser_setting}'`
+  end
+
+  ##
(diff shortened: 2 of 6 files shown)


test/rubygems/test_gem_commands_owner_command.rb
```diff
@@ -425,6 +425,42 @@ def test_with_webauthn_enabled_failure
     refute_match response_success, @stub_ui.output
   end
 
+  def test_with_webauthn_enabled_with_safari_as_default_browser
+    omit("Safari is only available on MacOS") unless RUBY_PLATFORM.include?("darwin")
```
hsbt, line 429: nit-pick: s/MacOS/macOS/

test/rubygems/test_gem_commands_owner_command.rb
```diff
@@ -425,6 +425,42 @@ def test_with_webauthn_enabled_failure
...
+      Gem::WebauthnListener.stub(:wait_for_otp_code, "Uvh6T57tkWuUnWYo") do
+        use_ui @stub_ui do
+          @cmd.add_owners("freewill", ["user-new1@example.com"])
+        end
+      end
+    ensure
+      server.close
+    end
+
+    restore_default_browser
```
hsbt, line 454: Should we restore default browser in `ensure` block? If exception was happend after `set_default_browser_safari`, developer configuration is no longer restored.

jenshenny:
cc: @simi @segiddins

test/rubygems/test_gem_commands_owner_command.rb
jenshenny: Ah good catch, will do!

segiddins (approved):

test/rubygems/helper.rb
```diff
@@ -1491,6 +1491,23 @@ def self.key_path(key_name)
...
   end
 
+  ##
+  # Sets default browser to Safari in macOS by saving current configuration and overwriting LSHandlers
+
+  def set_default_browser_safari
+    @launch_services = `defaults read com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers`
+    default_browser_setting = "{ LSHandlerURLScheme = \"https\"; LSHandlerRoleAll = \"com.apple.safari\"; }"
+
+    `defaults write com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers -array '#{default_browser_setting}'`
```
simi, line 1501: Does this really temporarily change default browser? It seems too annoying for users running tests. In troubles it can leave system in different state.

What about to relax a little in here and check if default browser is Safari and if not skip the test? Would that be covered by current CI setup? Alternatively, what about to just stub the method?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -296,6 +301,18 @@ def webauthn_verification_url(credentials)
     response.is_a?(Net::HTTPSuccess) ? response.body : nil
   end
 
+  def default_browser_safari?
+    return false unless RUBY_PLATFORM.include?("darwin")
+
+    require "json"
+
+    launch_services = JSON.parse(`defaults read com.apple.LaunchServices/com.apple.launchservices.secure LSHandlers \
```
simi, line 309: :thinking: What about to wrap this, rescue any exception raised and return false on troubles? It would not make sense to break RubyGems just because system will change those internal commands.

test/rubygems/helper.rb
simi: Unwritten rule is to set `ENV['CI'] == 'true'` on CIs. Would it make sense to check if browser is Safair and run test (for local usage). And force switch to Safari on CI.

Alternatively we can run this step on MacOS CI workflow, so no need to set this from within the code itself. It seems probably the best. Also it would be great to fail on CI if this is not set properly to prevent this test being silently skipped in the future.

https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

jenshenny:
Hey I'm going to close this since a fallback has been implemented: https://github.com/rubygems/rubygems/pull/6774

PR #6615, opened 2023-04-17
Use pure-ruby YAML parser for loading configuration at RubyGems

hsbt:

Related https://github.com/rubygems/rubygems/issues/6611

RubyGems uses Psych(libyaml) for loading/storing configuration that is `.gemrc`. So, Psych is always activate in RubyGems. We can't use specified version of Psych. 

`libyaml` used by Psych have risk for vulnerability today. We should leave depends on Psych in long term.

I replaced Psych dependency to `Bundler::YAMLSerializer` at `Gem::ConfigFile`. It has incompatibility that is `Bundler::YAMLSerializer` can parse invalid yaml like `invalid: key: val`. We can't raise it after this.

I have no changes bundler side in this PR. We can replace `Psych` only `Gem::ConfigFile` changes.

### DIFF
--- a/bundler/spec/bundler/friendly_errors_spec.rb
+++ b/bundler/spec/bundler/friendly_errors_spec.rb
@@ -5,29 +5,6 @@ require "bundler/friendly_errors"
 require "cgi"
 
 RSpec.describe Bundler, "friendly errors" do
-  context "with invalid YAML in .gemrc" do
-    before do
-      File.open(home(".gemrc"), "w") do |f|
-        f.write "invalid: yaml: hah"
-      end
-    end
-
-    after do
-      FileUtils.rm(home(".gemrc"))
-    end
-
-    it "reports a relevant friendly error message" do
-      gemfile <<-G
-        source "#{file_uri_for(gem_repo1)}"
-        gem "rack"
-      G
-
-      bundle :install, :env => { "DEBUG" => "true" }
-
-      expect(err).to include("Failed to load #{home(".gemrc")}")
-    end
-  end
-
   it "calls log_error in case of exception" do
     exception = Exception.new
     expect(Bundler::FriendlyErrors).to receive(:exit_status).with(exception).and_return(1)
--- a/test/rubygems/test_gem_config_file.rb
+++ b/test/rubygems/test_gem_config_file.rb
@@ -465,21 +465,6 @@ if you believe they were disclosed to a third party.
     assert_equal %w[http://even-more-gems.example.com], Gem.sources
   end
 
-  def test_ignore_invalid_config_file
-    File.open @temp_conf, "w" do |fp|
-      fp.puts "invalid: yaml:"
-    end
-
-    begin
-      verbose = $VERBOSE
-      $VERBOSE = nil
-
-      util_config_file
-    ensure
-      $VERBOSE = verbose
-    end
-  end
-
   def test_load_ssl_verify_mode_from_config
     File.open @temp_conf, "w" do |fp|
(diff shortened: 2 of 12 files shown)


deivid-rodriguez:
In general, I think we should try to not require Bundler functionality from RubyGems, but instead move the functionality down to RubyGems if it's necessary there. That's what we've been doing in other cases and it has worked fine.

So, I'd copy `Bundler::YAMLSerializer` to RubyGems and then make Bundler use that if available, and otherwise its own copy of the functionality.

hsbt:
Fmm, I want to not maintain same code while RubyGems and Bundler because I implement this at PR only RubyGems working with old Bundler version.

hsbt:
okay, I will copy YAMLSerializer from Bundler.

We should start to consider this duplicated code and codebase.

bundler/spec/bundler/friendly_errors_spec.rb
```diff
@@ -5,29 +5,6 @@
...
-      gemfile <<-G
-        source "#{file_uri_for(gem_repo1)}"
-        gem "rack"
-      G
-
-      bundle :install, :env => { "DEBUG" => "true" }
-
-      expect(err).to include("Failed to load #{home(".gemrc")}")
-    end
-  end
```
deivid-rodriguez, line 29: Why was this test removed? What happens now when having an invalid `.gemrc` file?

test/rubygems/test_gem_config_file.rb
```diff
@@ -465,21 +465,6 @@ def test_write_from_hash
...
-
-    begin
-      verbose = $VERBOSE
-      $VERBOSE = nil
-
-      util_config_file
-    ensure
-      $VERBOSE = verbose
-    end
-  end
```
deivid-rodriguez, line 481: Same here, what happens now when loading an invalid file? It seems like previous behavior was ignoring it? Has that changed?

bundler/spec/bundler/friendly_errors_spec.rb
hsbt: I restored this test and behavior at https://github.com/rubygems/rubygems/pull/6624

jeremy:
This introduced a regression when the gem sources list is empty:
```bash
$ gem sources | grep http | while read -r source; do gem sources --remove $source; done
$ gem sources
    undefined method `each' for "[]":String

    other.each do |x|
         ^^^^^
```

This is because the new serializer parses `[]` as the `"[]"` String instead of as an empty Array:
```bash
ruby -rrubygems/yaml_serializer -e 'puts Gem::YAMLSerializer.load(":sources: []")'
{":sources"=>"[]"}
```

hsbt:
@jeremy Thanks. I filed it to https://github.com/rubygems/rubygems/issues/7098

williantenfen:
will this be extracted as a gem ? `rb-yaml`?

simi:
@williantenfen per my knowledge, there is no plan to do this for now. But feel free to extract if you consider this useful.

voiski:
To refer here in case someone had same issue, this PR changes created a regression on our code due the way we setup `s3_source`.

Sample with error
```yaml
:sources:
  - https://rubygems.org/
  - s3://my-gems/
s3_source:
  :my-gems:  # it will not map as before when symbol matches the host
    ....
```

To fix, remove the `:`, `:my-gems:` to `my-gems:`:
```yaml
s3_source:
  my-gems:
    ....
```

I'm not sure if it was desired. However, since it will work with older and newer Rubygem, it's better to change it to a normal string key.

PR #6624, opened 2023-04-19
Warn invalid gemrc with Psych version

hsbt:

I removed test and change behavior when RubyGems handled invalid gemrc at https://github.com/rubygems/rubygems/commit/cfcfde04c783bfd090c718978f9899fe261135eb.

I restored test and warn message when invalid yaml like `invalid: foo: bar`. But we can't cover all of invalid pattern same as libyaml.

### DIFF
--- a/bundler/spec/bundler/friendly_errors_spec.rb
+++ b/bundler/spec/bundler/friendly_errors_spec.rb
@@ -5,6 +5,29 @@ require "bundler/friendly_errors"
 require "cgi"
 
 RSpec.describe Bundler, "friendly errors" do
+  context "with invalid YAML in .gemrc" do
+    before do
+      File.open(home(".gemrc"), "w") do |f|
+        f.write "invalid: yaml: hah"
+      end
+    end
+
+    after do
+      FileUtils.rm(home(".gemrc"))
+    end
+
+    it "reports a relevant friendly error message" do
+      gemfile <<-G
+        source "#{file_uri_for(gem_repo1)}"
+        gem "rack"
+      G
+
+      bundle :install, :env => { "DEBUG" => "true" }
+
+      expect(err).to include("Failed to load #{home(".gemrc")}")
+    end
+  end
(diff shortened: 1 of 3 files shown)


simi (approved):

ericherscovich:
Hi I think this PR may be preventing the auth flow for MFA for working locally. The auth flow fails with the warning:
```
Failed to load /Users/ericherscovich/.local/share/gem/credentials because it doesn't contain valid YAML hash
```
I believe the cause of this is from this commit, specifically the line 352 in `config_file.rb` :
```ruby
 if config.keys.any? {|k| k.to_s.gsub(%r{https?:\/\/}, "").include?(":") }
```

Locally, this returns true which triggers the warning. I believe this is because the config file contains `"http://localhost:3000/"` in it. The `k.to_s.gsub(%r{https?:\/\/}, "")` turns the contents from 
```
{"http://localhost:3000/"=>"rubygems_aHash"}
```
to
```
"{\"localhost:3000/\"=>\"rubygems_aHash\"}"
```
but because of the port, the `includes?(":")` is still true. I am not sure if this is intentional or not, but this seems to be why it's failing locally. If this isn't intentional, can we add some kind of condition where this doesn't fail if port 3000 is used?

indirect:
That sounds good to me.

ericherscovich:
@hsbt @indirect Would it actually be better to have the include check for `: ` instead of just `:`? It appears the YAML requires those spaces, so checking for those would be more accurate. Doing this would also solve the `localhost` problem.

indirect:
I think you are describing a check for valid YAML (which should have one `: ` in the line, for single line keys and values). Ultimately I would suggest building a set of examples to test against, both valid and invalid, and then writing whatever code makes the test pass.

jenshenny:
This PR should fix the issues with loading localhost api keys https://github.com/rubygems/rubygems/pull/6683

PR #6631, opened 2023-04-21
Add a Metadata DSL Method

jessedoyle:

We were looking for a mechanism to annotate dependencies with additional metadata to support internal tooling to assist with application maintenance.

This PR adds a `meta` DSL method that works as follows:

```ruby
meta team: :backend do
  gem "rails"
end

meta team: :frontend do
  gem "react-rails"
end
```

The `meta` method accepts a hash of key/value pairs and yields the data to a block. All gems defined within the block are annotated with the metadata provided.

Metadata is not written to the lockfile, nor does it influence installation or dependency resolution.

One use case for metadata is to enable internal tooling that tracks the long term maintenance of dependencies across many development teams.

Given the definition above, tooling can be implemented as follows:

```ruby
builder = Bundler::Dsl.new
dependencies = builder.eval_gemfile("Gemfile")

dependencies.each do |dependency|
  # do something important...
  puts "#{dependency.name}: maintained by #{dependency.meta[:team]}"
end
```

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -621,7 +621,7 @@ module Bundler
     method_option "major", type: :boolean, banner:       "If updating, prefer updating to next major version (default)"
     method_option "pre", type: :boolean, banner: "If updating, always choose the highest allowed version, regardless of prerelease status"
     method_option "strict", type: :boolean, banner: "If updating, do not allow any gem to be updated past latest --patch | --minor | --major"
-    method_option "conservative", type: :boolean, banner:       "If updating, use bundle install conservative update behavior and do not allow shared dependencies to be updated"
+    method_option "conservative", type: :boolean, banner: "If updating, use bundle install conservative update behavior and do not allow shared dependencies to be updated"
     method_option "bundler", type: :string, lazy_default: "> 0.a", banner: "Update the locked version of bundler"
     def lock
       require_relative "cli/lock"
--- a/bundler/lib/bundler/dependency.rb
+++ b/bundler/lib/bundler/dependency.rb
@@ -7,7 +7,7 @@ require_relative "rubygems_ext"
 module Bundler
   class Dependency < Gem::Dependency
     attr_reader :autorequire
-    attr_reader :groups, :platforms, :gemfile, :path, :git, :github, :branch, :ref
+    attr_reader :groups, :platforms, :gemfile, :path, :git, :github, :branch, :ref, :meta
 
     ALL_RUBY_VERSIONS = ((18..27).to_a + (30..33).to_a).freeze
     PLATFORM_MAP = {
@@ -41,6 +41,7 @@ module Bundler
       @ref            = options["ref"]
       @platforms      = Array(options["platforms"])
       @env            = options["env"]
+      @meta           = options.fetch("meta") { Hash.new }
       @should_include = options.fetch("should_include", true)
       @gemfile        = options["gemfile"]
       @force_ruby_platform = options["force_ruby_platform"] if options.key?("force_ruby_platform")
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -15,7 +15,7 @@ module Bundler
 
     VALID_PLATFORMS = Bundler::Dependency::PLATFORM_MAP.keys.freeze
 
-    VALID_KEYS = %w[group groups git path glob name branch ref tag require submodules
+    VALID_KEYS = %w[group groups git path glob name branch ref tag require submodules meta
                     platform platforms type source install_if gemfile force_ruby_platform].freeze
 
     GITHUB_PULL_REQUEST_URL = %r{\Ahttps://github\.com/([A-Za-z0-9_\-\.]+/[A-Za-z0-9_\-\.]+)/pull/(\d+)\z}
@@ -33,6 +33,7 @@ module Bundler
       @optional_groups      = []
       @platforms            = []
       @env                  = nil
+      @meta                 = nil
       @ruby_version         = nil
       @gemspecs             = []
       @gemfile              = nil
@@ -264,6 +265,14 @@ module Bundler
       @env = old
     end
 
+    def meta(data = {})
+      old = @meta
+      @meta = data
+      yield @meta
+    ensure
+      @meta = old
+    end
+
     def plugin(*args)
       # Pass on
     end
@@ -389,6 +398,7 @@ module Bundler
 
       opts["source"]         ||= @source
       opts["env"]            ||= @env
+      opts["meta"]           ||= @meta
       opts["platforms"]      = platforms.dup
       opts["group"]          = groups
       opts["should_include"] = install_if
--- a/bundler/spec/bundler/dsl_spec.rb
+++ b/bundler/spec/bundler/dsl_spec.rb
@@ -219,6 +219,25 @@ RSpec.describe Bundler::Dsl do
       dep = subject.dependencies.last
       expect(dep.name).to eq "foo"
     end
+
+    it "allows specifying metadata" do
+      subject.gem("foo", meta: { bar: "baz" })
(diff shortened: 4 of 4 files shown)


hsbt:
I'm +1 to this feature. But I'm not sure what `meta` is best name for this usecase.

segiddins:
I wonder if this is a bit of a heavy-handed change. If you're looking to do this just in your own gemfile, you should be able to monkey patch the `gem` method to collect this metadata for your own use, without requiring a change to bundlers public api

simi:
@jessedoyle is there anything similar in other packaging managers?

jessedoyle:
> is there anything similar in other packaging managers?

I'm not aware of any package managers that offer this functionality out of the box. Haven't done an exhaustive search though!

One primary difference between Bundler and many other package managers is that the `Gemfile` is parsed/evaluated as Ruby code.

For example with node dependencies, the `package.json` file is structured. Annotations can be added in the `package.json` file as new JSON properties that the package manager doesn't reference. Tooling can then be written to correlate the annotations with the package definitions.

It's more difficult to achieve the same thing with a `Gemfile`, as it would have to be parsed and evaluated first. And the `gem` method is defined in Bundler itself, so there's not a great mechanism to be able to pass annotations through the whole flow.

jessedoyle:
Just a note on the CI failures. Looks like there's a warning being generated:

> warning: instance variable @meta not initialized

Going to make sure this is addressed! Just hoping to align on whether this is something we think is mergeable first?

jessedoyle:
Thanks invoking CI! Looks like there was an [issue](https://github.com/rubygems/rubygems/actions/runs/5028989922/jobs/9097769249?pr=6631) with the check against Ruby 2.7. 

Is this something we think is related to this PR?

deivid-rodriguez:
No problem, I don't think it's related to this PR. I restarted that job.

Generally I think this PR is fine, although I understand the reluctances from other maintainers. It does look simple enough from an implementation point of view.

Same concern would be fixed by having a non-evaluated static manifest but that's a very big feature that probably won't get implemented in a long time.

deivid-rodriguez:
So, should we do this?

jessedoyle:
> So, should we do this?

Ultimately that's up to the RubyGems maintenance team, but I'm hoping we can agree that this would be a useful feature to land!

This change is quite flexible. While our use case is to track dependencies for maintenance reasons, I think metadata can be used in some novel ways. For example, it might be useful if writing a specialized bundler plugin.

While this change does add to the public API, it's a rather low impact addition. The implementation is trivial, and it really doesn't add a ton of complexity to the mix.

deivid-rodriguez:
Yeah, it was more of a question for the rest of maintainers, given what has been pointed out so far :)

simi:
I was thinking about this for a long time and recently more questions and issues popped out sharing the same root cause -> "custom" Gemfile/Gemfile.lock formats. @jessedoyle would it work for you to use let's say alternative `Gemfile.json` file format and add any metadata in there? It should be possible to vendor ruby-only JSON library into Bundler and add alternative json versions of Gemfile/Gemfile.lock files which could be useful for those custom additions.

WDYT @deivid-rodriguez? Would it make sense to explore this approach?

deivid-rodriguez:
@simi Sure, having a static manifest would be beneficial not only for this feature. For example, we could save a gemfile checksum in the lockfile and save some work when checking whether the gemfile has changed with respect to the lockfile, since we'd only have to compare two checksums.

I'm just not sure when something like that would be implemented, and this seemed like a tiny feature that we may want to add regardless.

martinemde:
This reminds me of the metadata on S3. It's your own space to put things that you can reference later. I like the idea as long as it stays simple and bundler explicitly stays uninvolved in the data stored within.

jessedoyle:
Thanks for the feedback @martinemde!

Exactly - the intent here is to keep things simple from bundler's perspective, but allow for a place for the implementor to put additional contextual information that is passed through.

> Should there also be a meta: key per gem?

Yup, that's supported as well! There's a [spec](https://github.com/jessedoyle/rubygems/blob/fa182a3dfdcc6706ee09b7c24c05f32a6231f2d7/bundler/spec/bundler/dsl_spec.rb#L223-L227) for this variant.

i.e.

```ruby
gem 'react-rails', meta: { team: :frontend }
```

> would it work for you to use let's say alternative Gemfile.json file format and add any metadata in there?

Honestly that's a pretty large change for most folks maintaining larger Ruby applications. Not sure it would end up getting utilized too often.

jessedoyle:
Hey team - we'd *love* to see this feature upstream. Is there anything we can do to help move this along?

If this is something that's a hard-no, happy to close the PR.

indirect:
@jessedoyle Thanks for checking back in! I am +1 on this, especially as a solution for this need in the near future.

We have also been discussing ideas internally about some kind of non-executable Gemfile format for quite some time, and I think that is a good idea we should pursue, but I don't think that (good) idea is a reason to block this feature that's already built and validated by actual use.

indirect (approved):

simi:
I'm still not sure this is complete feature.

What about dependencies of dependencies in your use case? How are you going to handle those? Do you need to know also version of the dependency for your usecase (in that case, so you plan to parse lockfile as well and pair with parsed Gemfile)?
What's the benefit of this over making a unified comment which can be also parsed using common tools?
Is there any benefit of this (or any plan to benefit from this) for Bundler itself like `bundler meta` command to output all related info?

jessedoyle:
Hey @simi - these are some great questions!

The implementation in this PR is intentionally minimal from Bundler's perspective. It puts the onus on the implementor to do the heavy lifting. This allows implementors the flexibility to develop a solution for their unique workflow.

With respect to transitive dependencies, yes we do parse the lockfile and marry this data with information from the Gemfile. Our tooling effectively cascades metadata down to nested dependencies.

> What's the benefit of this over making a unified comment which can be also parsed using common tools?

Fair point - a few thoughts here. 

First, using a comment lacks structure. The implementation in this PR allows for multiple gems to be defined in a block (similar to groups). With a comment, it's difficult to achieve the same functionality.

Second, using a comment requires static analysis of the Gemfile to derive information. We define our dependencies as Ruby code. Our tooling is written as Ruby code. In this scenario, we can leverage the evaluation context already present from the Ruby interpreter and avoid static analysis.

> Is there any benefit of this (or any plan to benefit from this) for Bundler itself like bundler meta command to output all related info?

This type of command could be implemented in the future! We have no immediate need for this functionality to be built into Bundler, but perhaps others may want that?

bundler/lib/bundler/cli.rb
```diff
@@ -621,7 +621,7 @@ def inject(name, version)
     method_option "major", type: :boolean, banner:       "If updating, prefer updating to next major version (default)"
     method_option "pre", type: :boolean, banner: "If updating, always choose the highest allowed version, regardless of prerelease status"
     method_option "strict", type: :boolean, banner: "If updating, do not allow any gem to be updated past latest --patch | --minor | --major"
-    method_option "conservative", type: :boolean, banner:       "If updating, use bundle install conservative update behavior and do not allow shared dependencies to be updated"
+    method_option "conservative", type: :boolean, banner: "If updating, use bundle install conservative update behavior and do not allow shared dependencies to be updated"
```
jessedoyle, line 624: Looks like RuboCop autocorrect fixed this whitespace via `bundle exec rubocop -A`. Happy to revert if necessary!

simi:
@jessedoyle looking at this again after some time, I don't think `Gemfile` is the best place to define team responsibilities, it is place to list dependencies only. There were also some plans to introduce non-evaluated Gemfile (like JSON or YAML variant). Keeping Gemfile as simple as possible would make it easier to do this transition. On the other side I'm not sure what's current state on this.

Wouldn't be in your case enough to split `Gemfile` into `Gemfile.frontend` and `Gemfile.backend` and parse each individually for analysis and in `Gemfile` just include those?

Anyway I'm not going to block this.

PR #6639, opened 2023-04-26
Update CodeClimate config

duckinator:

CodeClimate still lists too many problems to be easily actionable.

In an effort to surface the worst problems, I configured CodeClimate to:
- ignore vendored libraries (`optparse`, `tsort`)
- use v2 of the config instead of v1
- allow methods to have up to 7 arguments (instead of 4)
- allow methods to have a Cognitive Complexity of up to 30 (instead of 5)
- allow classes/modules to have 30 methods (instead of 20)
- allow files to have 400 lines (instead of 250)
- not care about the number of return statements

This dropped the number of issues reported by CodeClimate **from 777 to 148**.

The other problems are still there, but silencing the other issues highlights things like:
- a class that has 178 methods
- a file with 1,382 lines of code
- a method with a Cognitive Complexity of **52**(?!)
- many instances of duplicate code

None of which I was able to find before, because they were buried in 30+ pages of issues that are fairly minor in comparison (like functions with 6 or 7 arguments).

- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes (N/A)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -24,3 +24,6 @@ exclude_paths:
 - bundler/spec/**/*
 - bundler/Rakefile
 - test/**/*
+- lib/rubygems/optparse.rb
+- lib/rubygems/optparse/**/*
+- lib/rubygems/tsort/**/*
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -15,7 +15,7 @@ checks:
       # - @duckinator
       #
       # Default: 5.
-      threshold: 20
+      threshold: 30
   method-count:
     config:
       # Default: 20
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,6 +1,10 @@
 ---
 version: "2"
 checks:
+  argument-count:
+    config:
+      # Default: 4
+      threshold: 7
   method-complexity:
     config:
       # We have multiple methods where CodeClimate measures the
@@ -40,6 +44,11 @@ checks:
       #
       # Default: 250.
       threshold: 400
+  return-statements:
+    # This was adding flagging a lot of functions where having many
+    # return statements made sense.
+    # - @duckinator
+    enabled: false
 plugins:
   duplication:
     enabled: true
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,5 +1,6 @@
 ---
-engines:
+version: "2"
+plugins:
   duplication:
     enabled: true
     config:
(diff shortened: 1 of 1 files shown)


duckinator:
This PR has the down side of making the codebase look like it's in better shape than it is, if you take CodeClimate's "Codebase summary" at face value.

But I think this is outweighed by CodeClimate now finally being _useful_ for RubyGems and Bundler, which is not a thing we've had before.

You can see the changes by comparing these two URLs:

- https://codeclimate.com/github/rubygems/rubygems (before these changes)
- https://codeclimate.com/github/duckinator/rubygems (after these changes)

simi:
Strange, I thought I left comment on this, but I can't find it.

Is codeclimate still related in Rubocop era? What's the benefit of keeping Codeclimate? Wouldn't be easier to just remove it? I see a lot of duplication with Rubocop setup in Codeclimate config.

duckinator:
It may make sense to disable redundant features, and enable the Rubocop plugin. CodeClimate's main benefits are the nice UI and tracking information over time.

If we can make it surface useful information, but avoid duplicating functionality that Rubocop already has (by having CodeClimate use Rubocop behind the scenes), that might be a good option.

Do you think I should experiment with that sometime?

simi:
I tried to setup CC for RubyGems.org to be inline with Rubocop, but I failed and almost removed it in the end. The only useful missing feature was code coverage track, but that was replaced with `codecov.io` with super simple setup.

There is nice web UI for CC indeed, but locally during development it is impossible to check metrics status if I understand it well. This is well addressed by Rubocop.

But until CC is not going to fail builds I think there is no real need to remove it.

deivid-rodriguez (approved):
@duckinator I think you're the only maintainer actively looking at Code Climate, so I'm happy for you to maintain it as you like best!

duckinator:
I've been thinking on this + what @simi said, and the conclusions that I came to are:
- Like @simi said, CodeClimate is redundant with RuboCop.
- If CodeClimate is configured to just be a nice interface for RuboCop, there's little point to using it.

So I think the best course of action would be this:

1. Close this PR.
2. Remove the CodeClimate badge in the README, or replace it with a CodeCov.io badge.
3. Disable CodeClimate for the RubyGems repository.
4. Possibly write/find documentation for how to get RuboCop integrated with different text editors/IDEs.

I think documenting or linking to other folks' documentation for integrating RuboCop will provide more real-world benefit.

deivid-rodriguez:
Alright @duckinator, let me close this then, and you can follow up with the steps detailed when you get some time!

Issue #6630, opened 2023-04-20
Bundle keep "installing" plugins from Gemfile


dfop02:
### Describe the problem as clearly as you can

I started building a very simple plugin using as reference this [guide](https://bundler.io/guides/bundler_plugins.html), after push and test I notice that when we bundle install them from Gemfile (without git or path), it keep saying "installing" every run.

### Did you try upgrading rubygems & bundler?

Yes, I tested on Bundler 2.4.12 and Rubygems 3.4.12

### Post steps to reproduce the problem

Just add any plugin to your Rails App gemfile as specified on the early guide, ex.:
Gemfile
```
plugin 'bundle_filter'
plugin 'extended_bundler-errors'
```

### Which command did you run?

`bundle` or `bundle install`

### What were you expecting to happen?

plugins are already installed, but the bundler tries install again

### What actually happened?

"Installing plugin x" even if is already installed

### Where is the problem?

I did a small search and found the problem probably [here](https://github.com/rubygems/bundler/blob/master/lib/bundler/plugin.rb#L88) where Bundler keep installing plugins if they're already installed, what makes the bug.

### How solve the problem?

I actually tried few changes on my own version of Bundler gem 2.4.12, looks very simple to fix but since I don't know if could have any impact, prefer wait you guys to take a look. I solved the problem just checking if there is plugins not installed yet before actually install someone, here:

[bundler/plugin.rb#L88](https://github.com/rubygems/bundler/blob/master/lib/bundler/plugin.rb#L88)
```
plugins = definition.dependencies.map(&:name).reject { |p| index.installed? p }
return if plugins.empty?
installed_specs = Installer.new.install_definition(definition)
```
Please take a look and say if makes sense :)

### If not included with the output of your command, run `bundle env` and paste the output below

```
Bundler             2.4.12
  Platforms         ruby, x86_64-darwin-20
Ruby                2.7.2p137 (2020-10-01 revision 5445e0435260b449decf2ac16f9d09bae3cafe72) [x86_64-darwin-20]
  Full Path         /Users/dfop/.rvm/rubies/ruby-2.7.2/bin/ruby
  Config Dir        /Users/dfop/.rvm/rubies/ruby-2.7.2/etc
RubyGems            3.4.12
  Gem Home          /Users/dfop/.rvm/gems/ruby-2.7.2@my-test-rails
[...]

PR #6643, opened 2023-04-28
[Plugin] Fix "Installing" bug when install plugin from Gemfile

dfop02:

The problem was initially described [here](https://github.com/rubygems/rubygems/issues/6630). When you install a plugin from a Gemfile (without git or path options), the bundle keep installing it every bundle install run.

There is 2 problems to solve. 

The first problem is block an installed plugin to install again, it was fixed the by adding a condition [here](https://github.com/dfop02/rubygems/blob/fix-bundle-plugin-bug/bundler/lib/bundler/plugin.rb#L117), where I confirm if there is any plugin not installed yet before continue to install dependences.

The second problem is when you already have an installed plugin, but add other into your Gemfile, then it keep installing the old one. For this case, I need go a bit far to check if the plugin is installed and if the plugin installed version is the same. For this, I create a `cannot_install?` method on Index class for this propose. You can check the code [here](https://github.com/dfop02/rubygems/blob/fix-bundle-plugin-bug/bundler/lib/bundler/plugin/installer.rb#L103).

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -114,6 +114,9 @@ module Bundler
         return if definition.dependencies.empty?
 
         plugins = definition.dependencies.map(&:name).reject {|p| index.installed? p }
+
+        return if plugins.empty?
+
         installed_specs = Installer.new.install_definition(definition)
 
         save_plugins plugins, installed_specs, builder.inferred_plugins
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -39,8 +39,9 @@ module Bundler
       raise InvalidOption, "You cannot specify `--branch` and `--ref` at the same time." if options["branch"] && options["ref"]
 
       specs = Installer.new.install(names, options)
+      plugins = plugins_to_save(names, specs)
 
-      save_plugins names, specs
+      save_plugins(plugins, specs)
     rescue PluginError
       specs_to_delete = specs.select {|k, _v| names.include?(k) && !index.commands.values.include?(k) }
       specs_to_delete.each_value {|spec| Bundler.rm_rf(spec.full_gem_path) }
@@ -113,10 +114,10 @@ module Bundler
 
         return if definition.dependencies.empty?
 
-        plugins = definition.dependencies.map(&:name).reject {|p| index.installed? p }
         installed_specs = Installer.new.install_definition(definition)
+        plugins = plugins_to_save(definition.dependencies.map(&:name), installed_specs)
 
-        save_plugins plugins, installed_specs, builder.inferred_plugins
+        save_plugins(plugins, installed_specs, builder.inferred_plugins)
       end
     rescue RuntimeError => e
       unless e.is_a?(GemfileError)
@@ -139,6 +140,15 @@ module Bundler
         local_root
       else
         global_root
+      end
+    end
+
+    # Return array of plugins that we need save or update the path
+    # based if was not installed with that version yet
+    def plugins_to_save(plugins, installed_specs)
+      plugins.reject do |p|
+        installed_spec = installed_specs[p]
+        installed_spec.nil? || index.version_already_installed?(p, installed_spec.version)
       end
     end
 
@@ -253,7 +263,7 @@ module Bundler
     # @param [Array<String>] names of inferred source plugins that can be ignored
     def save_plugins(plugins, specs, optional_plugins = [])
       plugins.each do |name|
-        next if index.installed?(name)
+        next if index.version_already_installed?(name, specs[name].version)
 
         spec = specs[name]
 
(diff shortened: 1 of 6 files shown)


bundler/lib/bundler/plugin.rb
```diff
@@ -113,6 +113,9 @@ def gemfile_install(gemfile = nil, &inline)
         return if definition.dependencies.empty?
 
         plugins = definition.dependencies.map(&:name).reject {|p| index.installed? p }
+
+        return if plugins.empty?
```
deivid-rodriguez, line 117: If I have

```
plugin 'extended_bundler-errors', '0.3.2'
```

in my Gemfile, and I change it to

```
plugin 'extended_bundler-errors', '0.3.3'
```

I think this is going to change things to not install the new version?

bundler/lib/bundler/plugin/index.rb
```diff
@@ -111,6 +111,17 @@ def command_plugin(command)
         @commands[command]
       end
 
+      # Plugin cannot be installed and updating to install
+      def cannot_install?(name, version)
+        installed?(name) && !updating?(name, version)
```
deivid-rodriguez, line 116: The `cannot_install?` makes me think of same problem when trying to install the gem. Maybe `version_already_installed?` or something like that would be better?

deivid-rodriguez (commented):
Thanks for fixing this, I added a couple of comments.

Also, the previous CI timeout is unrelated to this PR and now fixed on master, so a rebased CI run should not run into that.

bundler/lib/bundler/plugin.rb
dfop02: Yeah, you're right, I'll fix it.

bundler/lib/bundler/plugin/index.rb
dfop02: Really sounds better, I changed the def name. Also want ask about the plugin version, do you have any suggest to how get it without using the regex, maybe a build-in way that I missed... or this way is fine? I'm talking about plugin index [here](https://github.com/dfop02/rubygems/blob/fix-bundle-plugin-bug/bundler/lib/bundler/plugin/index.rb#L121).

deivid-rodriguez: I had the same thought but I could not find a better way that the regex based approach you used.

dfop02: My initial objective was separate the methods to looks simple, because already exists the `installed?` so I just added the `!updating?`, maybe change to `not_updating?` or `same_version?` feels more readable. But since still sounds confuse, I can try your approach. Also, good idea about the "regex issue", I'll test :)

dfop02: While I was testing your approach about the "regex issue", I found few more things to fix, and now looks like *almost everything works fine. I notice that the index file was not updating the path correctly after change versions updating/downgrading so I needed fix it too.

Almost because before, the plugin was suppose to appears "Using foo x.y.z" when already installed, but now is suppressed by default (only for plugins), I'm not sure if is a big problem. That happens because when you cli bundle install, the [installed_specs](https://github.com/dfop02/rubygems/blob/fix-bundle-plugin-bug/bundler/lib/bundler/source/rubygems.rb#L451-L453) returns the plugin while bundle by gemfile doesnt and I couldnt understand why that happens.

deivid-rodriguez: Thanks for all the work, I'll loop back and have a look at that last issue.

deivid-rodriguez: @dfop02 Sorry I didn't loop back to this yet. Did you look into it? We recently removed unnecessary "Using..." messages at #6804, not sure if that changes anything here.

dfop02: Hello @deivid-rodriguez !

I take a look again today on currently bundler version, looks like the updates solve the "using..." messages that I told before, what is nice, but installing messages and index file issue continues. For now I create a separate matcher because I thought that was causing the failures on specs, let's see if solves it :)

ccutrer (commented):
you haven't done anything that prevents downloading metadata from rubygems if the plugin is already installed

bundler/lib/bundler/plugin/index.rb
```diff
@@ -111,6 +111,15 @@ def command_plugin(command)
         @commands[command]
       end
 
+      # Plugin version is already installed?
+      def version_already_installed?(name, version)
+        plugin_path = @plugin_paths[name]
+        return false unless plugin_path
+
+        File.basename(plugin_path) == "#{name}-#{version}"
```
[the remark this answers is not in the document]
dfop02: Hello @ccutrer ! Indeed, you're right, I was focusing my attention on the "avoid unnecessary installs" problem and forgot about more things that I could fix. I saw your approach, better solving the problem at the root, congrats.

deivid-rodriguez:
Since everyone seems to agree #6957 is a better approach, let's close this in favor of that other PR.

PR #6673, opened 2023-05-11
Load plugin immediately

kou:

We can install RubyGems plugin by "gem install XXX". The installed plugin is used from the NEXT "gem ...".

For example, "gem install gem-src kaminari" doesn't use gem-src plugin for kaminari. "gem install gem-src && gem install kaminari" uses gem-src plugin for kaminari.

How about loading a plugin immediately when the plugin is installed? If this proposal is implemented, "gem install gem-src kaminari" works like "gem install gem-src && gem install kaminari".

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -343,6 +343,8 @@ class Gem::Installer
 
     Gem::Specification.add_spec(spec)
 
+    load_plugin
+
     run_post_install_hooks
 
     spec
@@ -1005,5 +1007,18 @@ TEXT
     else
       ""
     end
+  end
+
+  def load_plugin
+    specs = Gem::Specification.find_all_by_name(spec.name)
+    # If old version already exists, this plugin isn't loaded
+    # immediately. It's for avoiding a case that multiple versions
+    # are loaded at the same time.
+    return unless specs.size == 1
+
+    plugin_files = spec.plugins.map do |plugin|
+      File.join(@plugins_dir, "#{spec.name}_plugin#{File.extname(plugin)}")
+    end
+    Gem.load_plugin_files(plugin_files)
   end
 end
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1449,6 +1449,8 @@ class TestGem < Gem::TestCase
   def test_load_plugins
     plugin_path = File.join "lib", "rubygems_plugin.rb"
 
+    foo1_plugin_path = nil
+    foo2_plugin_path = nil
     Dir.chdir @tempdir do
       FileUtils.mkdir_p "lib"
       File.open plugin_path, "w" do |fp|
@@ -1458,17 +1460,22 @@ class TestGem < Gem::TestCase
       foo1 = util_spec "foo", "1" do |s|
         s.files << plugin_path
       end
+      foo1_plugin_path = File.join(foo1.gem_dir, plugin_path)
 
       install_gem foo1
 
       foo2 = util_spec "foo", "2" do |s|
         s.files << plugin_path
       end
+      foo2_plugin_path = File.join(foo2.gem_dir, plugin_path)
 
       install_gem foo2
     end
 
     Gem::Specification.reset
+    PLUGINS_LOADED.clear
+    $LOADED_FEATURES.delete(foo1_plugin_path)
+    $LOADED_FEATURES.delete(foo2_plugin_path)
 
     gem "foo"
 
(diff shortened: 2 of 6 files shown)


kou:
Rebased.

deivid-rodriguez:
If I have version X of plugin `gem-src` installed, then that will be loaded right before `gem install gem-src:X+1 kaminari` runs, but also version `X+1` will be loaded right after `gem-src` is installed, correct? Won't that cause two different versions of the same gem to be loaded? That sounds not good.

kou:
> If I have version X of plugin `gem-src` installed, then that will be loaded right before `gem install gem-src:X+1 kaminari` runs, but also version `X+1` will be loaded right after `gem-src` is installed, correct?

Correct.

> Won't that cause two different versions of the same gem to be loaded?

Yes. Two different versions of the same gem are loaded in the situation.

> That sounds not good.

I agree with you.

How about preventing loading `gem-src:X+1` in the situation? `gem install gem-src:X+1 kaminari` loads `gem-src:X` before it's executed and `gem-src:X+1` isn't loaded while `gem install gem-src:X+1 kaminari` is executing.
`gem-src:X+1` is used (and `gem-src:X` isn't used) from the next `gem install ...`.

Is this behavior acceptable?

kou:
I've implemented the behavior.

What do you think about this?

deivid-rodriguez:
@kou I'm fine with the updated behavior, thanks. I will review the PR in a few days, just to give other maintainers time to leave their feedback or raise their objections (if any).

kou:
Thanks!

lib/rubygems/installer.rb
```diff
@@ -345,6 +345,8 @@ def install
 
     run_post_install_hooks
 
+    load_plugin
```
deivid-rodriguez, line 348: If `run_post_install_hooks` happened before `load_plugin`, then any post install hooks potentially registered by the gem would be run immediately too, right?

For example, say the `rdoc` gem installs a plugin that generates documentation after gems are installed, then `gem install rdoc` would immediately generate docs for `rdoc` itself.

Did you consider this? Do you think it would be desirable?

deivid-rodriguez (commented):
There seem to be some error output when running tests now, can you have a look? Could those point to some bad side effect on real usage?

```
/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-9o50ny/gemhome/gems/a-2/lib/rubygems_plugin.rb
Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-gehefu/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-oovia/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-rpqs7e/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
```

lib/rubygems/installer.rb
kou: Ah, sorry. I didn't consider this.

`rdoc` (and `yard`) use `done_installing` hook not `post_install` hook but I can understand what you want to say.

I'll move `load_plugin` BEFORE `run_post_install_hooks`.

kou: Done.

BTW, I think that we can remove https://github.com/rubygems/rubygems/blob/master/lib/rubygems/rdoc.rb and and implement this by RubyGems plugin in RDoc after we merge this pull request.

kou:
> There seem to be some error output when running tests now, can you have a look? Could those point to some bad side effect on real usage?
> 
> ```
> /home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-9o50ny/gemhome/gems/a-2/lib/rubygems_plugin.rb
> Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-gehefu/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
> Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-oovia/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
> Error loading RubyGems plugin "/home/runner/work/rubygems/rubygems/tmp/test_rubygems_20230516-2057-rpqs7e/gemhome/plugins/a_plugin.rb": cannot load such file --  (LoadError)
> ```

Sorry. I missed them.
I've fixed them.

deivid-rodriguez (approved):
Thank you! This looks good to me, let's hope it doesn't create any bad side effects. But it something unexpected comes up, we can consider it later.

lib/rubygems/installer.rb
kou: Oh, I didn't notice #6021.
I'll work on it after we merge this.

kou:
Please mention me when we found any unexpected side effect.

lib/rubygems/installer.rb
kou: Implementation: https://github.com/ruby/rdoc/pull/1171

Issue #4412, opened 2021-02-28
Make it easier to run bundler specs from the root of the repository


deivid-rodriguez:
### Describe the problem as clearly as you can

People are understandably confused by our mono repository setup. We should make it easier to run bundler specs from the root of the repository, by doing `bin/rspec` or `bin/parallel_rspec`, or specifying specific files like `bin/rspec bundler/spec/commands/new_gem_spec.rb`.

We should create `bin/rspec` and `bin/parallel_rspec` binstubs at the root of the repo that automatically switch to the `bundler` folder, strip off `bundler/` from command line arguments, and run `bin/rspec`  or `bin/parallel_rspec` from the `bundler` folder with the modified arguments.

PR #6691, opened 2023-05-23
Unify Rakefile for RubyGems and Bundler

hsbt:

We have two `Rakefile` in one repository.

Part of fixes https://github.com/rubygems/rubygems/issues/4412

I merge `bundler/Rakefile` into `Rakefile` that is for RubyGems. After that, we can invoke Bundler examples with same Rakefile.

`bin/*` of bundler directory is out of scope with this PR. In this time, only integrate rake task.


Rakefile
```diff
@@ -556,14 +556,208 @@ task :update_licenses do
...
+    end
+  end
+end
+
+desc "Check RVM integration"
+task :check_rvm_integration do
+  # The rubygems-bundler gem is installed by RVM by default and it could easily
+  # break when we change bundler. Make sure that binstubs still run with it
+  # installed.
+  sh("RUBYOPT=-Ilib gem install rubygems-bundler rake && RUBYOPT=-Ilib rake -T")
```
deivid-rodriguez, line 638: The second RUBYOPT should now be `-Ibundler/lib`, given before it was `-Ilib`.

Rakefile
```diff
@@ -556,14 +556,208 @@ task :update_licenses do
...
+end
+
+desc "Check RubyGems integration"
+task :check_rubygems_integration do
+  # Bundler monkeypatches RubyGems in some ways that could potentially break gem
+  # activation. Run a non trivial binstub activation, with two different
+  # versions of a dependent gem installed.
+  sh("ruby -Ilib -S gem install reline:0.3.0 reline:0.3.1 irb && ruby -Ilib -rbundler -S irb --version")
```
deivid-rodriguez, line 646: Same with the second `-Ilib`.

deivid-rodriguez (commented):
Thanks for doing this.

Something else that needs to be changed is the `RGV` environment variable. Since now tests run from root, I think it needs to be `RGV=.` rather than `RGV=..`. I think that's most likely why realworld specs were failing, and should fix a bunch of warnings alerting that the specs are running using an old RubyGems version.

Rakefile
hsbt: Ah, Thanks!

deivid-rodriguez:
Oh, I was wrong, since `bin/parallel_rspec` still runs from `bundler/` folder. But it would need to be `RGV=.` for `rake spec:deps` and `rake spec:parallel_deps` since those now run from the root folder.

.github/workflows/realworld-bundler.yml
```diff
@@ -51,14 +51,15 @@ jobs:
...
-        run: BUNDLER_SPEC_SUB_VERSION=${{ matrix.bundler.value }} bin/rake override_version
+        run: BUNDLER_SPEC_SUB_VERSION=${{ matrix.bundler.value }} rake override_version
         if: matrix.bundler.value != ''
-        working-directory: ./bundler
       - name: Run Test
-        run: bin/rake spec:realworld
+        # rake spec:realworld is failing with Ruby 2.6. We should use `bin/rspec` until dropping Ruby 2.6 support.
+        run: BUNDLER_SPEC_PRE_RECORDED=1 bin/rspec --tag realworld
+        env:
+          RGV: ..
```
deivid-rodriguez, line 62: I think it should pass now as it was before? Can you go back to `bin/rake spec:realworld` to verify? I'm curious about those failures.

hsbt: see https://github.com/rubygems/rubygems/actions/runs/5075333460/jobs/9116438594?pr=6691

I have no idea to investigation for next step.

deivid-rodriguez: I think the reason may be that the `bin/rake` binstub has been removed, and that's what ensured that the version of RubyGems in the git repository is being used (according to the `RGV` env variable), instead of the system installed version of RubyGems. Does it work if instead of removing it, you move the binstub to the repo root `bin` folder, and use `bin/rake` instead of `rake`?

hsbt: Thanks. I try it with tomorrow.

hsbt: `bin/rake` seems working fine.

hsbt:
@deivid-rodriguez Thank you. I extracted `bin` directory renaming to #6704.

PR #6706, opened 2023-05-29
Use released version of turbo_tests

hsbt:

The latest version(2.1.0) of `turbo_tests` contained our forked feature. We have no strong reason vendor `turbo_tests` now.

Replaced released version of `turbo_tests`


hsbt:
`turbo_tests` didn't support Ruby 2.6 and Windows platform. I added fallback for single rspec when running with their environment.

deivid-rodriguez:
Nice that someone took the time to extract this code into a gem. I was planning to do this but never got to it!

deivid-rodriguez:
Nice job @hsbt by the way!

deivid-rodriguez:
> turbo_tests didn't support Ruby 2.6 and Windows platform. I added fallback for single rspec when running with their environment.

Can we open an issue upstream about this? Particularly for Windows support. Our initial implementation supported this just fine so it's unclear why this was lost when the gem was extract. It's now quite inconvenient to inspect CI failures on Windows, for example, see https://github.com/rubygems/rubygems/actions/runs/5247046724/jobs/9476701221?pr=6736.

bundler/bin/parallel_rspec
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env ruby
 # frozen_string_literal: true
 
-require_relative "../spec/support/switch_rubygems"
+require_relative "../spec/support/rubygems_ext"
 
-require_relative "../tool/turbo_tests/cli"
-
-TurboTests::CLI.new(ARGV).run
+if RUBY_VERSION.start_with?("2.6") || RUBY_PLATFORM =~ /mingw|mswin/
```
ilyazub, line 6: The turbo_tests gem stopped supporting Ruby 2.6 after https://github.com/serpapi/turbo_tests/pull/29. The reason for that was Dependabot's concern: https://github.com/serpapi/turbo_tests/issues/28#issuecomment-1484682085

The parallel_tests dropped support for Ruby 2.6 in v4: https://github.com/grosser/parallel_tests/blob/v4.2.2/CHANGELOG.md#breaking-changes-1

I think we can revert parallel_tests to version 3 in the turbo_tests repository and unify the Bundler code for Ruby 2.6 against other versions if that helps.

@hsbt @deivid-rodriguez What do you think?

deivid-rodriguez: That Dependabot concern does not make a lot of sense to me, maybe it's a bug in Dependabot. Assuming one is using `turbo_tests` as a direct dependency, and `parallel_tests` just as an indirect dependency of it, Dependabot should never downgrade a top level dependency in order to bump an indirect dependency.

It would indeed be great for us if you restore Ruby 2.6 support, yes, and we can unify around `turbo_tests`, yes!

PR #6717, opened 2023-06-01
Prepare to release RubyGems 3.3.27 and Bundler 2.3.27

hsbt:
https://bugs.ruby-lang.org/issues/19576


deivid-rodriguez:
Thanks both for this, it seems that the backports need some work for old Ruby support?

jackorp:
> it seems that the backports need some work for old Ruby support?

Looking at Ruby 2.4, the message when it fails
```
/home/runner/work/rubygems/rubygems/bundler/tmp/1/gems/system/gems/bundler-2.4.13/lib/bundler/rubygems_ext.rb:352:in `glob': no implicit conversion of Hash into Integer (TypeError)
```

Seems to  point me towards the difference between bundler 2.4.13 and the v2.3.27 (this PR's) one.

The old Rubies were dropped from workflows in https://github.com/rubygems/rubygems/commit/77e7b79d657c20ecb0dccc73d38d180accf90aac

And the same commit caused the error above. See https://github.com/rubygems/rubygems/commit/77e7b79d657c20ecb0dccc73d38d180accf90aac#diff-dd13cb568c0fc396fbad7bc3665465e0fc81ce91042afc2d0d537d08a31c649e

In short, I don't think we can expect that upgrade tests using the latest versions of Bundler and Rubygems will work moving forward.

hsbt:
@jackorp I'm considering to release RubyGems-3.3.27 and Bundler-2.3.27 from this PR. Can you confirm to resolve your issue from this branch? Thanks.

jackorp:
I confirm that the issue is no longer present with this branch, thanks.

Issue #6568, opened 2023-04-03
uninitialized constant Gem::Ext::Builder::Shellwords


ioquatix:
Not every usage of `Shellwords` is prepended by `require 'shellwords'`.

```
Exception `Gem::Ext::BuildError' at /home/samuel/.rubies/ruby-3.2.1/lib/ruby/site_ruby/3.2.0/rubygems/ext/builder.rb:172 - ERROR: Failed to build gem native extension.

    uninitialized constant Gem::Ext::Builder::Shellwords
```

Do we want to add more "require 'shellwords'` in that file? Or do we want to use some kind of autoload? Or remove it completely? (avoid using shellwords).

Shellwords generally feels like a hack to me. Better to use proper array of arguments.

simi:
Any hint how to reproduce?

I guess it is missing before this call.

https://github.com/rubygems/rubygems/blob/f8ead4f5f0ea48c7a5ab2947167f66d3c3c2b1fe/lib/rubygems/ext/builder.rb#L30

ioquatix:
I can't remember how it was caused, but any case where we have `Shellwords` without an immediately preceeding `require 'shellwords'` is probably suspect given the current mechanism we are using. Maybe better just to have a top level autoload?

deivid-rodriguez:
Any of the suggested solutions sounds good to me.

ioquatix:
@deivid-rodriguez your original PR here only made it lazy for a subset of usage: https://github.com/rubygems/rubygems/pull/4783

Was that deliberate? What about other files? `require "shellwords"` is at the top of lots of files in `rubygems` project.

deivid-rodriguez:
No, I was probably just focused on fixing the particular issue I was seeing, and didn't audit every usage.

ioquatix:
Okay, I made a PR, hopefully that is helpful.

PR #6734, opened 2023-06-09
Autoload shellwords when it's needed.

ioquatix:
Attempt to fix <https://github.com/rubygems/rubygems/issues/6568>.

Some code paths weren't loading `shellwords` and I experienced some errors because of this. So, adding a top level autoload avoids this problem.

@fxn is it safe to add an autoload like this? i.e. are we at risk of clobbering anything if another project decided to do the same?

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -508,6 +508,7 @@ lib/rubygems/security/policy.rb
 lib/rubygems/security/signer.rb
 lib/rubygems/security/trust_dir.rb
 lib/rubygems/security_option.rb
+lib/rubygems/shellwords.rb
 lib/rubygems/source.rb
 lib/rubygems/source/git.rb
 lib/rubygems/source/installed.rb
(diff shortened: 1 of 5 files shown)


fxn:
@ioquatix I believe it should be mostly safe, but let me elaborate a bit to give elements to decide.

Adding an autoload is in practice like defining the constant, except that you defer the file evaluation. So this would be as "safe" as issuing a `require` call that defines `Shellwords` normally, in this respect.

On paper, this would conflict with code bases that define their own `Shellwords` top-level constant and also load RubyGems. However, who could do that before without loading "shellwords" as a side-effect one way or another. I believe it is really, really unlikely, if at all possible.

ioquatix:
Thanks so much for your great clarification and assurances.

ioquatix:
@deivid-rodriguez @simi I think this is good to go.

deivid-rodriguez:
Thanks!

PR #6739, opened 2023-06-13
Workaround for `turbo_tests` with Windows failure

hsbt:

`turbo_tests` used `File.mkfifo` that supports only UNIX environment. So, we shouldn't use `turbo_tests` with Windows platform yet.

I applied our forked version of [turbo_tests](https://github.com/rubygems/rubygems/blob/3552c064c1c9a44f2c7f70d77f35dfa00bd886e4/bundler/tool/turbo_tests/runner.rb) when it's running with Windows.

### DIFF
--- a/bundler/bin/parallel_rspec
+++ b/bundler/bin/parallel_rspec
@@ -3,7 +3,7 @@
 
 require_relative "../spec/support/rubygems_ext"
 
-if RUBY_VERSION.start_with?("2.6") || RUBY_PLATFORM =~ /mingw|mswin/
+if RUBY_VERSION.start_with?("2.6")
   Spec::Rubygems.gem_load("parallel_tests", "parallel_rspec")
 else
   Spec::Rubygems.gem_load("turbo_tests", "turbo_tests")
--- a/bundler/bin/parallel_rspec
+++ b/bundler/bin/parallel_rspec
@@ -6,5 +6,13 @@ require_relative "../spec/support/rubygems_ext"
 if RUBY_VERSION.start_with?("2.6")
   Spec::Rubygems.gem_load("parallel_tests", "parallel_rspec")
 else
-  Spec::Rubygems.gem_load("turbo_tests", "turbo_tests")
+  require_relative "../spec/support/switch_rubygems"
+  require "turbo_tests"
+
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@hsbt I think the problem is that `turbo_tests` uses a special FIFO file:

https://github.com/serpapi/turbo_tests/blob/2f0f152ed13981f02b8459c345422e6882a120ce/lib/turbo_tests/runner.rb#L175-L182

while we parsed stdout directly:

https://github.com/rubygems/rubygems/blob/3552c064c1c9a44f2c7f70d77f35dfa00bd886e4/bundler/tool/turbo_tests/runner.rb#L111-L123

I think I precisely did this, deviating from what `discourse` used, because it worked on Windows.

hsbt:
@deivid-rodriguez Yes, I filed https://github.com/serpapi/turbo_tests/issues/38 without `File.mkfifo` issues of Windows.

deivid-rodriguez:
We could probably workaround this by monkeypatching `JsonRowsFormatter#output_row` and `Runner#start_subprocess`, but better to wait for an upstream answer.

ilyazub:
From what I remember, an extra FIFO was needed to catch `errors outside of examples`: https://github.com/serpapi/turbo_tests/pull/10.

If it'll work with `stdout` instead of `tmp/test-pipes`, let's do this in `turbo_tests`.

hsbt:
@deivid-rodriguez I also override `Runner#run` and `Runner#handle_messages` for Windows. It may resolve stuck or flaky results.

@ilyazub Thanks, I'll investigate to support Windows platform of `turbo_tests` continuously.

deivid-rodriguez:
@hsbt Thanks for working on this and trying to bring back this feature!

The CI log on Windows seems still not helpful though, since it's now bloated with all the json messages being passed around. Ideally we should get just dots or actual failures like it did before.

deivid-rodriguez:
Just to close the loop in here. The monkeypatch has not really been working, but I only discovered that recently due to the verbose output and the job still having a success status even if no tests were not being run. Monkeypatch was fixed at #6988. Happy to contribute the patch upstream once it's exercised here for a while, either specifically for Windows, or for all platforms.

PR #6744, opened 2023-06-15
Mark Git source as cached when initialized from lockfile

### LINKED ISSUE #6743: Git sources sorting isn't stable

Yes

The Gemfile looks like this:
```ruby
source "https://rubygems.org"

```

```sh-session
bundle
git diff --name-only
# Gemfile.lock
bundle
git diff --name-only
# nothing
```

```sh-session
bundle update test
git diff --name-only
# Gemfile.lock
bundle
git diff --name-only
# nothing
```

I can also repro with main:

```
 Gemfile.lock | 2 +-
```

---

etiennebarrie:
This allows to get the current branch which ensures the identifier for a Git source is stable whether installing gems or not.

Fixes #6743 

I'm not sure it's the correct fix, I'm not totally clear what the `@allow_remote` and `@allow_cached` flags represent for `Bundler::Source`, `Bundler::Source::Path` and `Bundler::Source::Git`.

All I know is that `allow_git_ops?` use them to decide whether getting the current branch is ok or not.
https://github.com/etiennebarrie/rubygems/blob/f81cc96e33a445597e19a2e5c53017544922f2c5/bundler/lib/bundler/source/git.rb#L238-L240
https://github.com/etiennebarrie/rubygems/blob/f81cc96e33a445597e19a2e5c53017544922f2c5/bundler/lib/bundler/source/git/git_proxy.rb#L314-L315

The only other place that uses `allow_git_ops?` is `requires_checkout?`:
https://github.com/etiennebarrie/rubygems/blob/f81cc96e33a445597e19a2e5c53017544922f2c5/bundler/lib/bundler/source/git.rb#L272-L274

Returning `true` for that will now cause `local?` to be called (just an instance variable get), and `cached_revision_checked_out?`. This last one uses the `GitProxy`, but since it's initialized with the `cached_revision`, and we ask for the `revision`, it's just returned without using Git at all (could that be an issue?). It also checks that the install_path exists, which should add a system call.

Also not sure how to test this yet.

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -32,7 +32,9 @@ module Bundler
       end
 
       def self.from_lock(options)
-        new(options.merge("uri" => options.delete("remote")))
+        git = new(options.merge("uri" => options.delete("remote")))
+        git.cached!
+        git
       end
 
       def to_lock
--- a/bundler/spec/install/deploy_spec.rb
+++ b/bundler/spec/install/deploy_spec.rb
@@ -467,7 +467,7 @@ RSpec.describe "install in deployment or frozen mode" do
       expect(err).to include("frozen mode")
       expect(err).not_to include("You have deleted from the Gemfile")
       expect(err).not_to include("You have added to the Gemfile")
-      expect(err).to include("You have changed in the Gemfile:\n* rack from `#{lib_path("rack-1.0")}` to `no specified source`")
+      expect(err).to include("You have changed in the Gemfile:\n* rack from `#{lib_path("rack-1.0")} (at main@#{revision_for(lib_path("rack-1.0"))[0..6]})` to `no specified source`")
     end
 
     it "explodes if you change a source" do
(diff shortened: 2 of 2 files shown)


duckinator:
I confirmed locally that this fixes the issue you reported in #6743. The testcase you provided there seems like it could be a good starting point for an actual test.

I'm not familiar with this part of the codebase, though, so I'm unsure what exactly `@allow_cached` is for and what the implications of setting it (via `.cached!`) are.

I'll see if I can get someone more familiar with it to give it a look. This looks good to me (if a test is added to avoid regressions), though, barring any objections from someone more familiar with this part of the codebase.

segiddins:
So, thinking about this, I'm not sure this is the right place to call `cached!` (or `remote!` similarly), because it would enable `allow_git_ops?` even when the lock file is being loaded, say, due to `bundle exec`, instead of a `bundle install/update`.

Regarding the linked issue, I think overriding `Git#identifier` to not include the discovered ref, and only include what is given explicitly in the options, might be another approach?

duckinator:
I'm working on a spec for this problem, then I'm going to try @segiddins' idea of overriding `Git#identifier`.

deivid-rodriguez:
I actually had run into this issue myself a while ago and got a commit locally fixing it. I resurrected it to verify whether it was correct or not at #6782.

deivid-rodriguez:
@duckinator When you have that spec, can you add it to #6782? CI seems happy and local testing also works, but it'd be nice to get that spec added too!

duckinator:
Closing in favor of #6786. Thank you for getting the ball rolling on this, @etiennebarrie!

PR #6755, opened 2023-06-19
Update rdiscount gem version from 2.2.7 to 2.2.7.1

andrykonchin:

Rubygems CI builds (daily-rubygems) fail on TruffleRuby ([example](https://github.com/rubygems/rubygems/actions/runs/5306405480/jobs/9603964418)).

Error message from a job output:

```
gethopt.c:48:13: error: parameter 'val' was not declared, defaults to 'int'; ISO
C99 and later do not support implicit int [-Wimplicit-int]
hopterr(ctx,val)
            ^
```

CI fails on TruffleRuby on the `bundle install` step because building a native extension of `rdiscount` gem fails. A new version of `rdiscount` (2.2.7.1) was released recently with a [fix](https://github.com/davidfstr/rdiscount/issues/155).

In this PR `rdiscount` is upgraded to the latest version.

### DIFF
--- a/tool/bundler/dev26_gems.rb.lock
+++ b/tool/bundler/dev26_gems.rb.lock
@@ -10,7 +10,7 @@ GEM
     power_assert (2.0.3)
     rake (13.0.6)
     rb_sys (0.9.78)
-    rdiscount (2.2.7)
+    rdiscount (2.2.7.1)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Thank you @andrykonchin!

PR #6766, opened 2023-06-22
Small refactor for Gem::Resolver

hsbt:

`Gem::Resolver` have some of unused or unnecessary variables.

I removed them.

* `@missing` is unused variable in RubyGems.
* The default 2nd argument of `Molinillo::Resolver#resolve` is instance of `Molinillo::DependencyGraph`. So, we don't need to pass it from `Gem::Resolver`.

### DIFF
--- a/lib/rubygems/resolver.rb
+++ b/lib/rubygems/resolver.rb
@@ -38,8 +38,6 @@ class Gem::Resolver
   ##
   # List of dependencies that could not be found in the configured sources.
 
-  attr_reader :missing
-
   attr_reader :stats
 
   ##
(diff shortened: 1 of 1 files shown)


simi (approved):

lib/rubygems/resolver.rb
```diff
@@ -38,8 +38,6 @@ class Gem::Resolver
   ##
   # List of dependencies that could not be found in the configured sources.
```
segiddins, line 39: This comment should also be removed, since it was the documentation for `#missing`

PR #6769, opened 2023-06-24
Add ability to silence Gem::Package#build

segiddins:
For use in RubyGems.org

No way to build a `.gem` without printing to the console or changing global state

Add a kwarg to silence the output

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -290,7 +290,7 @@ class Gem::Package
   ##
   # Builds this package based on the specification set by #spec=
 
-  def build(skip_validation = false, strict_validation = false)
+  def build(skip_validation = false, strict_validation = false, silent: false)
     raise ArgumentError, "skip_validation = true and strict_validation = true are incompatible" if skip_validation && strict_validation
 
     Gem.load_yaml
@@ -312,12 +312,14 @@ class Gem::Package
       end
     end
 
-    say <<-EOM
+    unless silent
+      say <<-EOM
   Successfully built RubyGem
   Name: #{@spec.name}
   Version: #{@spec.version}
   File: #{File.basename @gem.path}
-EOM
+      EOM
+    end
   ensure
     @signer = nil
   end


simi:
For cli usage there is `--silent` and for "script" usage you should be able to temporary switch ui to silent one. Have you tried something like this?

```ruby
package = Gem::Package.new(...)
package.use_ui(Gem::SilentUI.new) do
  package.build(...)
end
```

segiddins:
That requires changing global state, which isn't ideal for a multiprocess web server, and it also doesn't prevent the string from being constructed

simi:
> That requires changing global state, which isn't ideal for a multiprocess web server, and it also doesn't prevent the string from being constructed

Yes, but that works the same for almost everything inside gem command. Any reason to make this one method special? Any details on the usage of this new argument? Maybe it is worth it to make an exception or explore instance based "ui".

Also if I understand this well, this code change prevents only one specific output of the `build` method, but spec validation can output other (warning, error) messages. :thinking:

segiddins:
Because this is the only one I needed at this timeOn Jun 26, 2023, at 3:23 PM, Josef Šimánek ***@***.***> wrote:﻿

That requires changing global state, which isn't ideal for a multiprocess web server, and it also doesn't prevent the string from being constructed

Yes, but that works the same for almost everything inside gem command. Any reason to make this one method special? Any details on the usage of this new argument? Maybe it is worth it to make an exception or explore instance based "ui".
Also if I understand this well, this code change prevents only one specific output of the build method, but spec validation can output other (warning, error) messages. 🤔

lib/rubygems/package.rb
```diff
@@ -290,7 +290,7 @@ def add_metadata(tar) # :nodoc:
   ##
   # Builds this package based on the specification set by #spec=
 
-  def build(skip_validation = false, strict_validation = false)
+  def build(skip_validation = false, strict_validation = false, silent: false)
```
deivid-rodriguez, line 293: I also end up avoiding switching from positional to kwargs, just to keep signatures consistent, but in retrospective, we should start to move on at some point 👍.

deivid-rodriguez (approved):
I'm good with this!

simi:
IMHO this is non-systematic and I still have no clue about your usecase @segiddins to evaluate if that's worth it to merge. :thinking:

segiddins:
https://github.com/rubygems/rubygems.org/pull/3875/files#diff-2252ed8e85781cc8ef901a0b3038fe197bab29ebde207b8b989a53a5af267227R37-R41

PR #6774, opened 2023-06-27
Add poller to fetch WebAuthn OTP

jenshenny:

Fixes: https://github.com/rubygems/rubygems.org/issues/3826
Discussed in the issue, we decided to rely on a polling solution for Safari users

RubyGems.org PR: https://github.com/rubygems/rubygems.org/pull/3873

If the user is using Safari, instead of redirecting the user to localhost after successful webauthn verification in the browser, [we'll show the success page](https://github.com/shopify/rubygems.org/blob/5cc347e37c8fcee61d9f463cba4f4c62fde46a6c/app/controllers/webauthn_verifications_controller.rb#L27) (since the auth on the browser was successful).

On the CLI, a poller is added that would poll `/api/v1/webauthn_verification/<token>/status` every 5 seconds.

Since we can't be sure which browser the user will be using (finding the default browser won't be fool proof since a user can paste the link into any browser), so we'll have a socket and poller thread open. 

Whichever thread terminates successfully with the otp will be used. This is done via a loop checking the status of the two threads.

## Testing

- Use this RubyGems.org branch https://github.com/rubygems/rubygems.org/pull/3873
- Have a Webauthn credential associated with your user.
- Run `RUBYGEMS_HOST=http://localhost:3000 ruby -Ilib exe/gem signin` and webauthn verify using Safari


test/rubygems/test_gem_gemcutter_utilities.rb
```diff
@@ -345,4 +348,75 @@ def test_verify_missing_api_key
       @cmd.verify_api_key :missing
     end
   end
+
+  class SignInFetcher < Gem::FakeFetcher
```
jenshenny, line 352: `util_gem_signin` was getting too gnarly to add fetcher data so I extracted a helper class to setup the fetcher. After writing the rest of the tests for the other commands, I'm intending to follow up with another refactor to have a universal child fetcher for webauthn verifications since they all stub the same requests.

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -253,17 +253,40 @@ def request_with_otp(method, uri, &block)
...
+        elsif error = otp_thread[:error]
+          alert_error error.message
+          terminate_interaction(1)
+        end
+      end
+      sleep 0.1
+    end
+  ensure
+    threads.each(&:exit)
+  end
```
jenshenny, line 287: Is there a better way to terminate once the poller or the listener succeeds?

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -272,17 +295,34 @@ def wait_for_otp(webauthn_url)
...
+          sleep 5
+        when "success"
+          Thread.current[:otp] = parsed_response["code"]
+          break
+        else
+          raise Gem::WebauthnVerificationError, parsed_response["message"]
+        end
+      end
+    rescue Gem::WebauthnVerificationError => e
+      Thread.current[:error] = e
```
jenshenny, line 320: I wanted to move this polling logic to outside this file to reduce bloat. However, it relies on `GemcutterUtilities#rubygems_api_request`. Is there a way to extract this logic while being able to still use `rubygems_api_request`? 🤔

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -253,36 +253,82 @@ def request_with_otp(method, uri, &block)
...
+          case parsed_response["status"]
+          when "pending"
+            sleep 5
+          when "success"
+            Thread.current[:otp] = parsed_response["code"]
+            break
+          else
+            raise Gem::WebauthnVerificationError, parsed_response["message"]
+          end
+        end
```
jenshenny, line 323: Thinking of extracting this logic out into a class as a follow up

jenshenny: Mostly to get the current functionality out as soon as possible and wanted to get people's thoughts on it. If you feel that extracting would be good to do in this PR, I can do that 👍

jenshenny: commit to extract the logic to a class https://github.com/rubygems/rubygems/pull/6774/commits/78a8b8a32e8baf7d9969e810e0686e36be7aa642

I also moved the webauthn listener class to the gemcutter utilities namespace https://github.com/rubygems/rubygems/commit/fd92d0d510d985daad9806a9af3634691594e372

test/rubygems/test_gem_gemcutter_utilities.rb
jenshenny: Done in https://github.com/rubygems/rubygems/pull/6774/commits/f49fb73cac9b6700b7de07fd977a5c21578046cf

lib/rubygems/gemcutter_utilities.rb
```diff
@@ -253,36 +254,39 @@ def request_with_otp(method, uri, &block)
...
   def fetch_otp(credentials)
     options[:otp] = if webauthn_url = webauthn_verification_url(credentials)
-      wait_for_otp(webauthn_url)
+      server = TCPServer.new 0
+      port = server.addr[1].to_s
+
+      url_with_port = "#{webauthn_url}?port=#{port}"
+      say "You have enabled multi-factor authentication. Please visit #{url_with_port} to authenticate via security device. If you can't verify using WebAuthn but have OTP enabled, you can re-run the gem signin command with the `--otp [your_code]` option."
+
+      threads = [WebauthnListener.listener_thread(host, server), WebauthnPoller.poll_thread(options, host, webauthn_url, credentials)]
```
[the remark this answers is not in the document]
jenshenny: Good question, I toiled with this a bit. The original reason sockets were chosen was that the connection between the browser and the CLI cannot be phished + responses would be immediate. With polling, while very limited, there's still a risk that someone else can retrieve an OTP code if they have access to credentials + your unique link while also having up to a 5 second delay. With those reasons, I still think it would be worth to have polling as just a fallback.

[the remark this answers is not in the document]
indirect: Yeah, I agree that the socket is preferred, and if Safari ever lets us remove polling, we should.

indirect (approved):

indirect:
Nothing from me!

simi (approved):

lib/rubygems/gemcutter_utilities/webauthn_poller.rb
```diff
@@ -0,0 +1,79 @@
...
+      thread
+    end
+
+    def poll_for_otp(webauthn_url, credentials)
+      Timeout.timeout(TIMEOUT_IN_SECONDS) do
+        loop do
+          response = webauthn_verification_poll_response(webauthn_url, credentials)
+          raise Gem::WebauthnVerificationError, response.message unless response.is_a?(Net::HTTPSuccess)
+
+          require "json"
```
simi, line 52: @deivid-rodriguez is it safe in here to require JSON? I did some experiment and we can also vendor pure Ruby JSON if needed. It is possible to also use YAML in API, but I'm not sure that will make any difference in here.

deivid-rodriguez: My understanding is that this is only used for `gem push`, `gem yank`, and `gem owner`. If that's the case I think it's fine, at least for now, since there's no user code involved in these commands, so there should not be any conflicts.

jenshenny:
Thanks folks for taking a look at this!

Hmm after rebasing, the tests appear to be a bit flaky now with the changes 🤔 
https://github.com/rubygems/rubygems/actions/runs/5600005188/jobs/10241758022
https://github.com/rubygems/rubygems/actions/runs/5600328937/jobs/10242557190

Going to dig into this before this gets merged in.

jenshenny:
So to deal with the possibly flaky nature of using real TCPServers to test the commands (signin, owner add/remove, yank), I created a mock server object to mock that logic. I say possibly flaky since running `ruby-core` on the same test seed passes + I couldn't find a way to reproduce the errors. 

I think it's better to mock the server as we aren't testing the WebAuthn listener logic (which interacts the TCPServer instance) in those tests but more on the output (we are stubbing the result of the WebAuthn listener anyways). We are still using real TCPServer instances on the Webauthn Listener tests to test requests coming in and out of the socket as realistically as possible.

Let me know if you all think differently! Open to other suggestions.

indirect:
That sounds like a totally reasonable approach to me. 👍🏻

deivid-rodriguez:
Happy to merge and release this whenever needed 👍.

jenshenny:
Thanks @deivid-rodriguez 🙌  I tested everything again and it all works well. If this can be a part of the next rubygems release, that would be perfect!

lib/rubygems/gemcutter_utilities/webauthn_poller.rb
```diff
@@ -0,0 +1,79 @@
...
+      @host = host
+    end
+
+    def self.poll_thread(options, host, webauthn_url, credentials)
+      thread = Thread.new do
+        Thread.current[:otp] = new(options, host).poll_for_otp(webauthn_url, credentials)
+      rescue Gem::WebauthnVerificationError, Timeout::Error => e
+        Thread.current[:error] = e
+      end
+      thread.abort_on_exception = true
```
segiddins, line 40: nit: set these in the `Thread.new` block using `Thread.current`

segiddins (approved):

lib/rubygems/gemcutter_utilities/webauthn_listener/response.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      def add_access_control_headers
+        @http_response["access-control-allow-origin"] = @host
+        @http_response["access-control-allow-methods"] = "POST"
+        @http_response["access-control-allow-headers"] = %w[Content-Type Authorization x-csrf-token]
+      end
+
+      def add_body
+        return unless body
+        @http_response["content-type"] = "text/plain"
```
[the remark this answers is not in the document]
jenshenny: Oh good catch, done ✅

PR #6782, opened 2023-06-30
Sort by hash when expiring sources

### LINKED ISSUE #6743: Git sources sorting isn't stable

deivid-rodriguez:

Sometimes git source ordering is not stable.

Currently, git sources have a `#to_s` method that may leave side effects, since it actually "materialize" the source (it checks it out and figures out HEAD revision, to include it when printing the source).

This caused some surprising breakage when I was moving code around, and it's not necessary as long as the sorting is consistent.

Fixes #6743.

### DIFF
--- a/bundler/lib/bundler/source_list.rb
+++ b/bundler/lib/bundler/source_list.rb
@@ -105,6 +105,10 @@ module Bundler
       get(source) || default_source
     end
 
+    def unsorted_lock_sources
+      other_sources + lock_rubygems_sources
+    end
+
     def lock_sources
       lock_other_sources + lock_rubygems_sources
     end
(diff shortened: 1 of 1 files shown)


bundler/lib/bundler/source_list.rb
```diff
@@ -221,7 +225,7 @@ def warn_on_git_protocol(source)
     end
 
     def equivalent_sources?(lock_sources, replacement_sources)
-      lock_sources.sort_by(&:identifier) == replacement_sources.sort_by(&:identifier)
+      lock_sources.sort_by(&:hash) == replacement_sources.sort_by(&:hash)
```
segiddins, line 228: sorting by hash is almost never stable

deivid-rodriguez: Good catch, the idea still works though I think? We want to sort by something that's based on the keys used to compute this hash, but that is stable?

deivid-rodriguez: Actually, all we care about here is that sorting is stable within the same invocation. And that's guaranteed, right?

deivid-rodriguez:
Closing since this does not fix the original issue.

PR #6783, opened 2023-06-30
Deprecate Gem::Platform.match?

hsbt:

This is extracted from https://github.com/rubygems/rubygems/pull/6238. We should deprecate `Gem::Platform.match?` because this method is ambiguous name.

I marked `Gem::Platform.match?` as deprecate status.

### DIFF
--- a/lib/rubygems/platform.rb
+++ b/lib/rubygems/platform.rb
@@ -22,6 +22,11 @@ class Gem::Platform
     match_platforms?(platform, Gem.platforms)
   end
 
+  class << self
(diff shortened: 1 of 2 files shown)


simi (approved):

deivid-rodriguez (approved):
This will cause very noisy warnings when using Bundler 2.1.4 with latest RubyGems.

I believe that's fine, Bundler 2.1.4 is from 3 and a half years ago, but I thought it was worth calling that out in case we want to wait a bit more for this.

PR #6786, opened 2023-06-30
Fix git source lockfile instability

### LINKED ISSUE #6743: Git sources sorting isn't stable

deivid-rodriguez:

Sometimes Bundler may print inconsistent git source ordering to the lockfile.

We have some flags that limit running git commands under certain situations, for example, when running under `--local`. However, those should only affect remote git operations, not local read-only operations like `git --version`, or `git rev-parse --abbrev-ref HEAD`.

This commit refactors things to achieve that.

By doing this, the `#to_s` representation of a source is more consistent, since we don't get any errors when reading the checked out branch, and we avoid some flip-flop lockfile issues.

Fixes #6743.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -394,8 +394,8 @@ module Bundler
         lock_source = lock_dep.source || sources.default_source
         next if lock_source.include?(gemfile_source)
 
-        gemfile_source_name = dep.source ? gemfile_source.identifier : "no specified source"
-        lockfile_source_name = lock_dep.source ? lock_source.identifier : "no specified source"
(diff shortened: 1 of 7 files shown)


duckinator:
Added a test. Currently waiting for CI to finish. If CI is happy, I'll go ahead and merge it.

PR #6790, opened 2023-07-02
`gemfile.5`: Code format the default glob to escape Markdown

ParadoxV5:
The default glob as rendered (`{,,/*}.gemspec`) doesn’t make sense.

Quote the default glob in an inline code rather than double-quotes so the `*`s don’t and will never leak out as Markdown for emphasis.

* [x] Describe the problem/feature
* [ ] ~~Write tests for features and bug fixes~~
* [x] Write code to solve the problem
* [x] Make sure you follow the current code style and write meaningful commit messages without tags

### DIFF
--- a/bundler/lib/bundler/man/gemfile.5.ronn
+++ b/bundler/lib/bundler/man/gemfile.5.ronn
@@ -518,7 +518,7 @@ paths.
 
 The `gemspec` method supports optional `:path`, `:glob`, `:name`, and `:development_group`
 options, which control where bundler looks for the `.gemspec`, the glob it uses to look
-for the gemspec (defaults to: "{,*,*/*}.gemspec"), what named `.gemspec` it uses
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Thanks!

deivid-rodriguez:
@ParadoxV5 Would you be able to run `rake man:build` from the branch with your changes and commit the result? Our `gemfile.5` file also needs this update, and that's what `rake man:build` does.

ParadoxV5:
> @ParadoxV5 Would you be able to run `rake man:build` from the branch with your changes and commit the result?

Not quite… I just patched the file on the GitHub web editor 😁. I’ll clone my fork locally rq.

deivid-rodriguez (approved):
Looks good @ParadoxV5, thank you!

PR #6793, opened 2023-07-04
Vendor `net-http`, `net-protocol`, `resolv`, and `timeout` to reduce conflicts between Gemfile gems and internal dependencies

deivid-rodriguez:

Problems with our internal usage of default gems getting in the middle of our users.

Running the following script on Ruby 3.2 without any `timeout` version other than the default install reproduces the issue:

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'
  gem 'timeout'
end
```

Vendor `net-http`, `net-protocol`, `resolv`, and `timeout`.


hsbt:
Thank you for considering that. I understood this problem and your concern.

I worried that is not only for code management. The official ruby package size is increased in recent years. 

```
ruby-2.7.8.tar.gz 16950365
ruby-3.0.6.tar.gz 21315725
ruby-3.1.4.tar.gz 20917933
ruby-3.2.2.tar.gz 20467023
ruby-3.3.0-preview3.tar.gz 21550473
```

It's not only caused vendoring code. Especially, Ruby 3.3 has Prism parser and related code. Ruby core team are also negative to manage duplicated code like `net-http`, `timeout` on ruby repo.

I have no idea to resolve both problem that are specified any versions and not vendoring them.

deivid-rodriguez:
I don't think increased ruby package size strong reason to not do this, to be honest. This PR "just" increases about 70K the rubygems package size.

martinemde (approved):
I thought about this a bunch and talked with @deivid-rodriguez. I don't think this is the very best long term solution. I'd imagine rg&b playing nicely with any version of a stdlib that is loaded. However, this is not where this project is at right now, so I think we should stick with the plan in the RFC until we have a better plan. This is consistent with that plan.

deivid-rodriguez:
Since I got someone else to back me up, and @hsbt said privately he's ok with vendoring if we decide so, I set this PR to be merged after CI.

Rakefile
```diff
@@ -208,12 +189,13 @@ if File.exist?("tool/automatiek.rake")
...
   # We currently include the following changes over the official version:
   # * Avoid requiring the optional `net-http-pipeline` dependency, so that its version can be selected by end users.
+  # * Require vendored net/http version RubyGems if available, otherwise the stdlib version.
   desc "Vendor a specific version of net-http-persistent to bundler"
   Automatiek::RakeTask.new("net-http-persistent") do |lib|
     lib.version = "v4.0.2"
     lib.download = { github: "https://github.com/drbrain/net-http-persistent" }
     lib.namespace = "Net::HTTP::Persistent"
-    lib.prefix = "Bundler::Persistent"
+    lib.prefix = "Gem"
```
voxik, line 198: Should it really be `Gem` prefix instead of `Bundler`?

deivid-rodriguez: It was not only Bundler loading `net/http`, but also rubygems. The only way I managed to make net-http-persistent use the vendored copy in RubyGems when available was to put it under the `Gem` namespace.

It feels cleaner to me anyways, since it's a `net/http` extension after all and I believe this code organization makes that more clear.

voxik: It got me by surprise, because it is located under `bundler/lib/bundler/vendor/` and therefore it should logically have the `Bundler` prefix. So I wanted to double check.

But since I struggled with the namespaces a bit to fix our [test case in Fedora](https://src.fedoraproject.org/rpms/ruby/blob/rawhide/f/ruby.spec#_949-956), I understand that this was compromise.

Thx for elaborating.

PR #6797, opened 2023-07-06
Extract `Base64.encode64` method into test case

hsbt:

I have a plan to promote `base64` as bundled gems at `ruby/ruby`. see https://bugs.ruby-lang.org/issues/19351

After that, the test suite of rubygems is broken because it couldn't load `base64`. So, we need to avoid to use Base64 library at our tests.

`Base64.encode64` is simple pure ruby method. I ported it to `test_gem_request.rb`.

### DIFF
--- a/test/rubygems/test_gem_request.rb
+++ b/test/rubygems/test_gem_request.rb
@@ -3,7 +3,6 @@
 require_relative "helper"
 require "rubygems/request"
 require "ostruct"
-require "base64"
 
 unless Gem::HAVE_OPENSSL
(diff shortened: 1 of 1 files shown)


simi (approved):

simi:
Would be an alternative to add `base64` gem into test Gemfile?

hsbt:
>Would be an alternative to add base64 gem into test Gemfile?

Yes, but It's only for this repository. rubygems tests of `ruby/ruby` repository is failed after extracting `base64` from its repository.

Issue #6613, opened 2023-04-17
Option to disable "is running, but your lockfile was generated with 2.2.28. Installing Bundler 2.2.28 and restarting using that version"


mjobin-mdsol:
### Describe the problem as clearly as you can

I understand this functionality is to prevent some issues on deployed environment. And it's probably a safe thing to do...
But on my local, I don't really care to install an older version of bundler just because the Gemfile.lock of a random app is outdated.

We work with many apps and I hope it would only try to fetch newer versions, not older ones.

### Did you try upgrading rubygems & bundler?

Yes Always but ... N/A

### Post steps to reproduce the problem

Download an app with an old version of bundler listed in the Gemfile.lock

### Which command did you run?

`bundle install`

### What were you expecting to happen?

Don't install older version, (behavior could be configurable)

### What actually happened?

bundler automatically downgraded itself.

Thanks

simi:
Hello @mjobin-mdsol! Per my understanding there is currently no way to disable this behaviour. The original idea behind this is to keep it consistent. This feature actually doesn't downgrade installed bundler, just install bundler locked in Gemfile.lock and use that one for predictable result of bundler behaviour according to the lockfile.

In theory it could be possible to add config to disable this by modifying following method. But I'm not sure that change would be welcomed.

https://github.com/rubygems/rubygems/blob/51d407c2bf6af941b63a2413469b4a69a8fdd391/bundler/lib/bundler/self_manager.rb#L85-L90

deivid-rodriguez:
For what it's worth, I think this has been requested enough time, so I won't block making it configurable. Or maybe support to give `BUNDLE_VERSION` a requirement instead of an exact version.

duckinator:
I'm also okay with making it configurable.

**Workaround:** Run `bundle --version`, take the version number from that, and use `bundle _VERSION_ ...`.

(Using `bundle _VERSION_ ...` to override the version used was existing functionality, and we made sure to respect it when adding this.)

E.g.,

```console
puppy@orthrus.fox:~/dev/duckinator/okay$ tail -n 2 Gemfile.lock 
BUNDLED WITH
   2.2.15
puppy@orthrus.fox:~/dev/duckinator/okay$ bundle --version
Bundler version 2.5.0.dev
puppy@orthrus.fox:~/dev/duckinator/okay$ bundle _2.5.0.dev_ install
Fetching gem metadata from https://rubygems.org/.
Fetching rake 12.3.3
Installing rake 12.3.3
[...]

PR #6808, opened 2023-07-10
Don't switch locked version when using prerelease Bundler

hsbt:

I used pre-released version of Bundler like `2.5.0.dev` every day for testing development version. But auto-switching feature change released version of `BUNDLED WITH`.

So, I can't test prereleased version with production application. I always set `BUNDLER_VERSION=2.5.0.dev bundle ...`. 

I disabled auto-switching feature when we use prerelease version of Bundler. So, the users who used prereleased version always want to test prereleased version of Bundler with their application.

### DIFF
--- a/bundler/lib/bundler/self_manager.rb
+++ b/bundler/lib/bundler/self_manager.rb
@@ -76,7 +76,8 @@ module Bundler
     end
 
     def needs_switching?
-      autoswitching_applies? &&
+      !Gem::Version.new(Bundler::VERSION).prerelease? &&
+        autoswitching_applies? &&
         released?(lockfile_version) &&
         !running?(lockfile_version) &&
         !updating?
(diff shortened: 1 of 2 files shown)


hsbt:
I couldn't stub-out `Bundler::VERSION` in `bundle` methods. It spawn another process on rspec runner. I'm considering to change `version.rb` each affected examples.

deivid-rodriguez:
I would image some people with dev Bundler installed still want the locked version for certain apps?

How about implementing #6613 instead?

We could add a new `version` setting for this that could take either a specific version, `global`, or `local` (`bundle config version <local|global|<version>>`, or `BUNDLE_VERSION=<global|local|<version>>`).

Where the `local` value keeps the current behavior, the `global` value chooses the highest version installed locally, and a specific version would be equivalent to current `BUNDLER_VERSION=<version>`.

With that we would essentially reimplement `BUNDLER_VERSION` feature, but in a more flexible way, where you can easily configure the desired behavior globally or per app.

hsbt:
@deivid-rodriguez It's useful for me. I'll implement it.

Issue #6807, opened 2023-07-07
gem install is sloooow


jsirex:
### Describe the problem as clearly as you can

I have use-case with only locally available gem, which depends on some public gems.
I built gem locally (e.g. pkg/my-private-gem-0.1.0.gem)
When I try to install it via `gem install pkg/my-private-gem-0.1.0.gem` it takes really long time and consumes lots of CPU.

After debugging I noticed, that when gem is not found, it tries to re-download whole specs.4.8.gz and parses it each time.
I tried to reproduce it with local gem:

```bash
time gem install --debug --verbose my-private-gem-0.3.0.gem 
NOTE:  Debugging mode prints all exceptions even when rescued
HEAD https://index.rubygems.org/
200 OK
GET https://index.rubygems.org/info/my-private-gem
404 Not Found
Exception `Gem::RemoteFetcher::FetchError' at /usr/local/lib/site_ruby/3.1.0/rubygems/remote_fetcher.rb:232 - bad response Not Found 404 (https://index.rubygems.org/info/my-private-gem)
GET https://index.rubygems.org/prerelease_specs.4.8.gz
304 Not Modified
GET https://index.rubygems.org/specs.4.8.gz
304 Not Modified
# >>>>>>> HANG HERE <<<<<<<<
....
1 gem installed

real    0m54.307s
user    0m51.660s
sys     0m1.390s
```
or with just non-exist gem:

```bash
time sudo gem install --debug --verbose pony-gem
NOTE:  Debugging mode prints all exceptions even when rescued
HEAD https://index.rubygems.org/
200 OK
GET https://index.rubygems.org/info/pony-gem
404 Not Found
Exception `Gem::RemoteFetcher::FetchError' at /usr/local/lib/site_ruby/3.1.0/rubygems/remote_fetcher.rb:232 - bad response Not Found 404 (https://index.rubygems.org/info/pony-gem)
GET https://index.rubygems.org/prerelease_specs.4.8.gz
200 OK
GET https://index.rubygems.org/specs.4.8.gz
200 OK
# >>>>>>>>>>> IT HANGS HERE FOR A MINUTE <<<<<<<<<<<<
Exception `Gem::UnsatisfiableDependencyError' at /usr/local/lib/site_ruby/3.1.0/rubygems/resolver/installer_set.rb:82 - Unable to resolve dependency: user requested 'pony-gem (>= 0)'
ERROR:  Could not find a valid gem 'pony-gem' (>= 0) in any repository
[...]

PR #6810, opened 2023-07-10
Avoid unnecessary work for private local gem installation

deivid-rodriguez:

When installing a local private gem, we were still checking for remote versions of it. That involved checking whether the gem is available in the compact index, and if not, use the full index for resolution instead.

This is slow, and unnecessary, since in the end, it's the local gem version that will be chosen. We can skip this check and as a result, use the faster Compact Index API in this case.

Avoid fetching remote gem information for resolution if the gem is locally available.

Fixes part (the most important part) of #6807. 

### DIFF
--- a/lib/rubygems/resolver/installer_set.rb
+++ b/lib/rubygems/resolver/installer_set.rb
@@ -148,6 +148,8 @@ class Gem::Resolver::InstallerSet < Gem::Resolver::Set
       res << Gem::Resolver::InstalledSpecification.new(self, gemspec)
     end unless @ignore_installed
 
+    matching_local = []
+
     if consider_local?
       matching_local = @local.values.select do |spec, _|
         req.match? spec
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I simulated this by renaming a local copy of https://github.com/activeadmin/activeadmin, building the renamed version, and installing the locally built gem. The results are:

### Before

```
$ time gem install activeadminaaaaaaaa-2.14.0.gem
Successfully installed activeadminaaaaaaaa-2.14.0
1 gem installed
gem install activeadminaaaaaaaa-2.14.0.gem  15,93s user 0,95s system 25% cpu 1:05,92 total
```

### After

```
$ time gem install activeadminaaaaaaaa-2.14.0.gem
Successfully installed activeadminaaaaaaaa-2.14.0
1 gem installed
gem install activeadminaaaaaaaa-2.14.0.gem  0,70s user 0,14s system 27% cpu 3,031 total
```

indirect:
Nice! 🎉

indirect (approved):

hsbt:
👍

PR #6814, opened 2023-07-11
Fix `bundler/setup` unintendedly writing to the filesystem

### LINKED ISSUE #6743: Git sources sorting isn't stable

### LINKED ISSUE #6813: Issues with Bundler 2.4.16 Cache & Write Only Filesystems

During some recent container deployments we hit an issue where Bundler changed its behavior with regard to the cache. 

This may be off the beaten path, but we remove the bundler path since we are cloning a lot of repos during our Bundle install and said repos are rather large. For example prior to doing a Docker `COPY . .` we do this in our build process.

```
rm -rf vendor/bundle/ruby/*/cache
```

Prior to Bundler v2.4.16 this was just fine. I can show you an example of a before and after below. The fact that `bundler/setup` now writes to disk was the source of our issue and kept our application from booting.

2.4.15

```
rm -rf vendor/bundle/ruby/*/cache
BUNDLE_GEMFILE=/aoo/Gemfile ruby -e "require 'bundler/setup'"
> tree vendor/bundle/ruby/3.1.0/cache
vendor/bundle/ruby/2.7.0/cache [error opening dir]
0 directories, 0 files
```

2.4.16

```
rm -rf vendor/bundle/ruby/*/cache
BUNDLE_GEMFILE=/app/Gemfile ruby -e "require 'bundler/setup'"
tree vendor/bundle/ruby/3.1.0/cache
vendor/bundle/ruby/2.7.0/cache
└── bundler
    └── git
```

Yes, that is the issue. When we build our containers, we always install the latest version of ruby gems and bundler.

See description above and usage of `bundler/setup`.

I was expecting that Bundler would not try to write to the file system when the cache was removed. 

### What actually happened?

It created an empty `bundler/git` directory in the vendor cache.

deivid-rodriguez:

After #6786, `bundler/setup` would write to the filesystem when dealing with git sources.

When sorting and comparing git Gemfile vs lockfile sources during `bundler/setup` to figure out whether we need to re-resolve or not, we would try to find the default branch if nothing more specific was specified in the Gemfile.
    
If the git cache has been deleted though, that would fail.

The error would still be swallowed (and the branch would simply not be displayed), but trying to clone would still generate the side effect of creating the parent folder for the clone.
    
That could affect non-writable systems that don't expect `bundler/setup` to write to the filesystem at all.
    
To fix this, override `Bundler::Source::Git#identifier` to use exclusively static information, so it does not even try to clone the repo nor generate any side effects.

This is essentially a rework of the [original attempt](https://github.com/rubygems/rubygems/pull/6782) to fix #6743, but without the problems discussed in that PR.

Fixes #6813.

### DIFF
--- a/bundler/lib/bundler/source/git.rb
+++ b/bundler/lib/bundler/source/git.rb
@@ -69,19 +69,7 @@ module Bundler
 
       def to_s
         begin
-          at = if local?
-            path
-          elsif user_ref = options["ref"]
-            if /\A[a-z0-9]{4,}\z/i.match?(ref)
-              shortref_for_display(user_ref)
-            else
-              user_ref
-            end
-          elsif ref
-            ref
-          else
-            current_branch
-          end
+          at = humanized_ref || current_branch
 
           rev = "at #{at}@#{shortref_for_display(revision)}"
         rescue GitError
@@ -89,6 +77,10 @@ module Bundler
         end
 
         uri_with_specifiers([rev, glob_for_display])
+      end
+
+      def identifier
+        uri_with_specifiers([humanized_ref, cached_revision, glob_for_display])
       end
 
       def uri_with_specifiers(specifiers)
@@ -255,6 +247,20 @@ module Bundler
       end
 
       private
+
+      def humanized_ref
+        if local?
+          path
+        elsif user_ref = options["ref"]
+          if /\A[a-z0-9]{4,}\z/i.match?(ref)
+            shortref_for_display(user_ref)
+          else
+            user_ref
+          end
+        elsif ref
+          ref
+        end
+      end
 
       def serialize_gemspecs_in(destination)
         destination = destination.expand_path(Bundler.root) if destination.relative?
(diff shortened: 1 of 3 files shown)


metaskills:
Thank you for doing this! I like the new test case. Were the other test failures random or from this change/fix?

deivid-rodriguez:
@metaskills They were directly related to this change. I pushed one more tweak that I expect to fix them 🤞.

metaskills:
YUP! Looks green now. I really appreciate this followup. Is there anything I can do to help test/QA/etc?

simi:
> YUP! Looks green now. I really appreciate this followup. Is there anything I can do to help test/QA/etc?

You can try to checkout related branch and add to path.

```
git clone git@github.com:rubygems/rubygems.git -b fix-git-cache /tmp/rubygems
ruby -I/tmp/rubygems/bundler/lib -r bundler/setup -e "puts Bundler::VERSION" 
2.5.0.dev
```

metaskills:
Thanks! Confirmed working from my end.

simi (approved):

metaskills:
@simi !!!! How are you doing? Nice to see you around.

PR #6817, opened 2023-07-12
Introduce bundle config set version feature

hsbt:

Fixes https://github.com/rubygems/rubygems/issues/6613
Ref https://github.com/rubygems/rubygems/pull/6808

I introduced `bundle config --local set version global` for disabling to auto-install locked version of Bundler. And I also introduced `bundle config --local set version 2.4.0` for specifying to use version of Bundler.


deivid-rodriguez:
I'm wondering if `global` and `local` are the best names here. It could get confusing in combination with `--local` and `--global` flags to `bundle config`. Perhaps `lockfile` instead of `local`, and `default` instead of `global` would be better?

Also I'd like to hear opinions from other maintainers before releasing this.

hsbt:
>Perhaps `lockfile` instead of `local`

+1

And I prefer to use `system` instead of `global`.

deivid-rodriguez:
I also prefer `system` to both `global` and `default` 👍.

PR #6827, opened 2023-07-19
Improve require

nobu:

In my experiments extending `Gem.find_unresolved_default_spec` dead locks happened in a few tests, and I suspect this method may exit leaving the monitor locked.

Instead `rescue` and `exit` for each time, just surround by `mon_synchronize`, mainly requested gem activation part, and `LoadError` part.

Additionally, the outer loop that iterates until the inner loop find an element can be simplified by using `find` (or `any?`) instead of `each` and `break`.

### DIFF
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -36,9 +36,7 @@ module Kernel
   def require(path) # :doc:
     return gem_original_require(path) unless Gem.discover_gems_on_require
 
-    begin
-      RUBYGEMS_ACTIVATION_MONITOR.enter
-
+    RUBYGEMS_ACTIVATION_MONITOR.synchronize do
       path = path.to_path if path.respond_to? :to_path
 
       if spec = Gem.find_unresolved_default_spec(path)
@@ -49,13 +47,8 @@ module Kernel
           Gem.suffixes.each do |s|
             $LOAD_PATH[0...load_path_check_index].each do |lp|
               safe_lp = lp.dup.tap(&Gem::UNTAINT)
-              begin
-                if File.symlink? safe_lp # for backward compatibility
-                  next
-                end
-              rescue SecurityError
-                RUBYGEMS_ACTIVATION_MONITOR.exit
-                raise
+              if File.symlink? safe_lp # for backward compatibility
+                next
               end
 
               full_path = File.expand_path(File.join(safe_lp, "#{path}#{s}"))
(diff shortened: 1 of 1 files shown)


deivid-rodriguez (approved):
Nice!

nobu:
Only on mswin, `mon_synchronize` is not found.
I’ll investigate it today.

MSP-Greg:
@nobu

> Only on mswin, mon_synchronize is not found.

Haven't really looked at this, but I copied the file to my local copy of the ruby-loco mswin build, and required both a default and a bundled gem, and they both loaded (I've got code to download all the ruby-loco builds & logs).

Also, hacked the `daily-rubygems` workflow so I could run it with 'workflow dispatch', and mswin passed [here](https://github.com/MSP-Greg/rubygems/actions/runs/5604899782/jobs/10253419462), which ran on the https://github.com/MSP-Greg/rubygems/commits/00-pr-6827 branch.

nobu:
@MSP-Greg Thank you, also I thought there should be no differences depending platforms.
But it failed as https://github.com/nobu/ruby/actions/runs/5605156505.

MSP-Greg:
@nobu  Sorry, I checked your rubygems fork, but didn't look at your ruby fork.

Strange error, as btest rarely fails.  I'll see if I can find anything, not feeling confident that I will...

nobu:
The failing program is `.\ruby.exe`, so it should be the just-built mswin.
Also appveyor similarly.
https://ci.appveyor.com/project/nobu/ruby/builds/47590800#L2103

MSP-Greg:
Yes, I'd agree.  Can you run Windows locally?

MSP-Greg:
I added the changes here to my Ruby mswin folder, changed to `ruby/bootstraptest`, and ran the following:
```
ruby.exe --disable=gems runner.rb --ruby="C:/ruby-mswin/bin/ruby --disable=gems"
```
Tests passed?

MSP-Greg:
The test logs for the run show `-- --disable-gems`.  Is the beginning `--` intended?

I added some debug output to the file, and `kernel_require.rb` shouldn't be loaded with `--disable-gems`.  Or, I see it loaded with:
```
ruby -ropenssl -e"puts OpenSSL::VERSION"
```

But not with:
```
ruby --disable=gems -ropenssl -e"puts OpenSSL::VERSION"
```

nobu:
It passed by using `synchronize` instead of `mon_synchronize`.
A possibility may be that `BASERUBY` is old.

nobu:
`--` separates arguments into for runruby.rb and for the ruby invoked from runruby.rb.

nobu:
`Monitor#mon_synchronize` seems defined in 1.8 already, I have no idea what happened on mswin...

hsbt (approved):

PR #6831, opened 2023-07-21
Warn bundled gems

hsbt:

When users load `rexml` after upgrading Ruby version, they got `LoadError` of `rexml`. Because `rexml` is now bundled gems, not default gems that is shipped with under `RbConfig::CONFIG["rubylibdir"]`.

I want to eliminate this surprised situation.

I added warning feature for bundled gems and default gems. When users did load default gems that are listed on our plan of bundled gems, we can notice to them.

Notification message is like this.

```
>> require "csv"
csv is not part of the default gems since Ruby 3.4.0. Add it to your Gemfile.
=> true
```

and I also notice bundled gems like this:

```
>> require "net-imap"
net-imap is not part of the default gems since Ruby 3.1.0. Add it to your Gemfile.
/Users/hsbt/.local/share/rbenv/versions/master/lib/ruby/3.3.0+0/bundler/rubygems_integration.rb:261:in `require': cannot load such file -- net-imap (LoadError)
```

I believe this feature is helpful for Bundler users.

also see https://bugs.ruby-lang.org/issues/19776

### TODO

I didn't implement to notice about gemspec. If `foo` gem depends on `csv` intenally, This PR warns only about `csv` and `Gemfile`. We should also warn about `foo.gemspec`.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -31,6 +31,14 @@ end
 module Gem
 
   BUNDLED_GEMS = {
+    "rexml" => "3.0.0",
+    "rss" => "3.0.0",
+    "webrick" => "3.0.0",
+    "matrix" => "3.1.0",
+    "net-ftp" => "3.1.0",
+    "net-imap" => "3.1.0",
+    "net-pop" => "3.1.0",
+    "net-smtp" => "3.1.0",
     "abbrev" => "3.4.0",
     "observer" => "3.4.0",
     "getoptlong" => "3.4.0",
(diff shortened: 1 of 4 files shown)


hsbt:
`Gem::BUNDLED_GEMS` is here: https://github.com/ruby/ruby/blob/98bd5a71f2a01db450e2a2722d1938fdb9e784e9/lib/bundled_gems.rb

headius:
@hsbt I think we could just include the same `bundled_gems.rb` in JRuby stdlib and get this functionality yes?

Of the 3.4 gems, only two have extensions: nkf, which we've fixed in ruby/nkf#15, and syslog, which JRuby provides with FFI (we should get that into the gem). I think they all would be good for warnings.

Issue #6674, opened 2023-05-12
Should `gem generate_index` create a `versions` file? 


rmosolgo:
### Describe the problem as clearly as you can

Hi there! For many years I have distributed some custom gems on a static fileserver, using `gem generate_index` to prepare files on the server for use with bundler. It's been a wonderful workflow because the server requirements are so low. Recently, bundle times for my custom server got much slower, and I discovered that it's because bundler is looking for `/versions`, not finding it, and falling back to downloading _every_ version of the library on the server. 

<details>
<summary>`bundle install --verbose` with my gem server (without a Gemfile.lock) </summary>

<p>

```
bundle install --verbose
Running `bundle install --verbose` with bundler 2.4.13
[DEPRECATED] This Gemfile does not include an explicit global source. Not using an explicit global source may result in a different lockfile being generated depending on the gems you have installed locally before bundler is run. Instead, define a global source in your Gemfile like this: source "https://rubygems.org".
Found changes from the lockfile, re-resolving dependencies because the dependencies in your gemfile changed, you added a new platform to your gemfile
HTTP GET https://1b0890259d8@gems.graphql.pro/versions
HTTP 404 Not Found https://1b0890259d8@gems.graphql.pro/versions
Bundler::Fetcher::FallbackError: Net::HTTPNotFound: https://1b0890259d8@gems.graphql.pro/versions
/Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/fetcher/downloader.rb:45:in `fetch'
  /Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/fetcher/compact_index.rb:124:in `call'
  /Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/compact_index_client/updater.rb:47:in `update'
  /Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/compact_index_client.rb:92:in `update'
  /Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/compact_index_client.rb:79:in `update_and_parse_checksums!'
  /Users/rmosolgo/.rbenv/versions/3.2.0/lib/ruby/site_ruby/3.2.0/bundler/fetcher/compact_index.rb:68:in `available?'
[...]

PR #6853, opened 2023-08-01
Add support for generating compact index files

segiddins:

Fixes https://github.com/rubygems/rubygems/issues/6674

`gem generate_index` did not output compact index files, making those self-hosted indexes slower to install gems from than necessary

Added compact index file generation, behind a `--compact` CLI flag

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -10,6 +10,7 @@ AllCops:
     - lib/rubygems/resolver/molinillo/**/*
     - lib/rubygems/tsort/**/*
     - lib/rubygems/optparse/**/*
+    - lib/rubygems/indexer/compact_index/**/*
     - bundler/lib/bundler/vendor/**/*
   CacheRootDirectory: tmp/rubocop
   MaxFilesInCache: 5000
--- a/Rakefile
+++ b/Rakefile
@@ -148,6 +148,16 @@ if File.exist?("tool/automatiek.rake")
     lib.license_path = "COPYING"
   end
 
+  desc "Vendor a specific version of compact_index to rubygems"
+  Automatiek::RakeTask.new("compact_index") do |lib|
+    lib.version = "v0.14.0"
+    lib.download = { :github => "https://github.com/rubygems/compact_index" }
+    lib.namespace = "CompactIndex"
+    lib.prefix = "Gem::Indexer"
+    lib.vendor_lib = "lib/rubygems/indexer/compact_index"
+    lib.license_path = "LICENSE.txt"
+  end
+
   desc "Vendor a specific version of pub_grub to bundler"
   Automatiek::RakeTask.new("pub_grub") do |lib|
     lib.version = "main"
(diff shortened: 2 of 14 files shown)


simi (approved):
Wow, this is super cool! :muscle:

indirect (approved):
awesome! can we turn this on by default and use the flag to turn it off?

deivid-rodriguez:
Yeah, this is awesome!

hsbt:
Is there plan to optional for using `compact_index` like `gem server`? It's a not strong opinion, But ruby-core would like to avoid the additional vendored code.

indirect:
My recollection is that the separate gem was intended to be temporary, and the compact index code was always intended to eventually end up inside RubyGems so that it could be use by `gem serve` and by `gem generate_index` and by RubyGems.org. I would be ok with moving `gem generate_index` out of the RubyGems repo and into a separate gem if we want to make this repo smaller, though.

hsbt:
>I don't see why we'd want to reimplement the logic and diverge from how rubygems.org generates the compact index files

I didn't say that. How only enable `--compact` option when user installed `compact_index` gem in their environment?

In background, vendored code is giving me trouble because It makes difficult to handle vulnerability in ruby core like URI.

Again, I'm okay to add `compact_index` to rubygems.

martinemde:
Do I understand this correctly? @hsbt would like to _not_ vendor compact_index. Then when running `gem generate_index`, if compact_index can be required then it should also build the compact index?

Is this the only blocker (besides test failures) for getting this code added?

segiddins:
@hsbt would you mild clarifying for me what your desired outcome is here? I would really like to ship compact index support by default, since I think it will make it significantly more likely it gets used. Otherwise, no one is going to just have the gem sitting around, so I'm afraid a weak dependency on it would practically speaking mean `gem generate_index` would never generate the compact index, which is the index format that bundler & rubygems perform best with

indirect:
I also think that `gem generate_index` should include compact index files by default.

What do the rest of you think about moving the entire `generate_index` command to a separate gem so we don't have to keep the server-side stuff inside RubyGems core? (After extracting to a gem, we could even make `gem generate_index` install the gem if needed when the command is run, so there is no functional change from an end-user perspective.)

indirect:
@segiddins are you cool with completing this by extracting `generate_index` into a gem that gets auto-installed if needed, which is something I think we want to do regardless?

segiddins:
Yeah, I can take a stab at that. Would we want to rip the code for `generate_index` out entirely & move it into a plugin?

indirect:
I think so, yes. Just leave some stub code behind that installs the gem if it’s not installed and then delegates to it?

deivid-rodriguez:
I think vendoring this would be fine, and I believe there's even opportunity to also reuse the gem in Bundler for parsing compact index responses? The gem seems also quite simple so I don't think it will add a lot of overhead.

indirect:
My understanding is that Samuel has been working on optimized parsing code just for Bundler, but I am in favor of the parsing code living inside Bundler.

Personally, I think it's a good idea for us to shrink the core RubyGems package whenever we are able, both to ease the maintenance burden and to streamline the install/upgrade/load process. This seems like a good place to do that to me, since so few users ever generate indexes of any kind.

If we can shrink RubyGems, satisfy ruby-core's request, and make it easier to ship updates to the index-generation code (since we can update when the command is run rather than requiring a full rubygems upgrade), that sounds compelling to me.

All that said, if there are other reasons to keep it vendored, I'm not opposed.

deivid-rodriguez:
My only concern is that it may be surprising that a rubygem gets autoinstalled when `gem generate_index` is run, but other than that this seems fine and in any case auto installing as suggested seems like the only way to make this extraction fully transparent to users, without actually vendoring the code.

segiddins:
Re-implemented in https://github.com/rubygems/rubygems-generate_index/pull/10 !

hsbt:
@segiddins @indirect Thank you for your consideration.

PR #6863, opened 2023-08-07
Stop building docs for rubygems-update gem

### LINKED ISSUE #6860: Default to no docs generation during `gem update --system`?
Upgrading RubyGems is very slow compared to upgrading Bundler:

```
❯ gem update --system
Updating rubygems-update
Fetching rubygems-update-3.4.18.gem
Successfully installed rubygems-update-3.4.18
Parsing documentation for rubygems-update-3.4.18
Installing ri documentation for rubygems-update-3.4.18
Done installing documentation for rubygems-update after 28 seconds
Parsing documentation for rubygems-update-3.4.18
Done installing documentation for rubygems-update after 0 seconds
Installing RubyGems 3.4.18
  Successfully built RubyGem
  Name: bundler
  Version: 2.4.18
  File: bundler-2.4.18.gem
Bundler 2.4.18 installed
RubyGems 3.4.18 installed
Regenerating binstubs
Regenerating plugins
Parsing documentation for rubygems-3.4.18
Installing ri documentation for rubygems-3.4.18
```

It's pretty frustrating to have to wait 30 seconds for a RubyGems upgrade when upgrading Bundler by itself is basically instantaneous.

I noticed that `gem update --system` generates rdoc and ri documentation for the rubygems-update gem. Since we don't actually use that gem for anything, could we skip that step to speed up upgrading rubygems itself, even if we continue to generate documentation for other gems by default?

deivid-rodriguez:

Upgrading RubyGems is painfully slow unless `--no-document` is explicitly set in `.gemrc`. It takes like 30 seconds.

This is just a temporary gem to help upgrading RubyGems. I don't think it makes sense to build docs for it.

Most of the time is spent building documentation. We setup a bunch of documentation files to be parsed by rdoc, and that's slow. In particular, two of these files (`CHANGELOG.md` and `bundler/CHANGELOG.md`) are very big and doc is super slow when parsing them.

If we remove those two files from the `extra_rdoc_files` in the `.gemspec` file, things are back to fast, but I went a step further and completely removed building documentation for the ruby gems-update gem, since I don't think it's really useful?

Furthermore, in the future we should probably cleanup after ourselves and remove this gem once we've successfully updated the installation?

Fixes #6860.
 
### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -292,7 +292,7 @@ command to remove old versions.
     args = []
     args << "--silent" if options[:silent]
     args << "--prefix" << Gem.prefix if Gem.prefix
-    args << "--no-document" unless options[:document].include?("rdoc") || options[:document].include?("ri")
+    args << "--no-document"
     args << "--no-format-executable" if options[:no_format_executable]
     args << "--previous-version" << Gem::VERSION if
       options[:system] == true ||
--- a/rubygems-update.gemspec
+++ b/rubygems-update.gemspec
@@ -21,15 +21,6 @@ Gem::Specification.new do |s|
   s.bindir = "exe"
   s.executables = ["update_rubygems"]
   s.require_paths = ["hide_lib_for_update"]
-  s.rdoc_options = ["--main", "README.md", "--title=RubyGems Update Documentation"]
-  s.extra_rdoc_files = [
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
`gem update --system` takes explicit `--documentation` option, so I think this would require a more elaborated plan to deprecated that.

So I went with the easier fix of removing the slow files from the list at #6864.

Issue #6212, opened 2022-12-28
Support `bundle info <name>-<version>`, or better document that `bundle info` takes a gem name


fulldecent:
### Describe the problem as clearly as you can

This is also known as the:

> Could not find XXXX in locally installed gems

problem

### Did you try upgrading rubygems & bundler?

Yes

### Post steps to reproduce the problem

Use this build environment:

<img width="572" alt="Screenshot 2022-12-28 at 12 39 00" src="https://user-images.githubusercontent.com/382183/209869760-9c72120f-5ed8-438d-9097-5e3effc64f7a.png">

Use this Gemfile

<img width="715" alt="Screenshot 2022-12-28 at 15 42 13" src="https://user-images.githubusercontent.com/382183/209869794-1df57105-d4d9-42c3-ab10-1d4a44078908.png">

Type

bundle install 

then

bundle info json-2.6.3

### Which command did you run?

As above

### What were you expecting to happen?

Immediately after installing JSON, JSON is installed

### What actually happened?

<img width="510" alt="Screenshot 2022-12-28 at 15 36 22" src="https://user-images.githubusercontent.com/382183/209869912-15c853c9-db97-4a25-ac18-29c67b101bab.png">

### If not included with the output of your command, run `bundle env` and paste the output below

vscode ➜ /workspaces/911coned.com (911coned.com ✗) $ bundle env
## Environment

```
Bundler       2.4.1
  Platforms   ruby, aarch64-linux
Ruby          3.1.3p185 (2022-11-24 revision 1a6b16756e0ba6b95ab71a441357ed5484e33498) [aarch64-linux]
  Full Path   /usr/local/rvm/rubies/ruby-3.1.3/bin/ruby
  Config Dir  /usr/local/rvm/rubies/ruby-3.1.3/etc
RubyGems      3.4.1
  Gem Home    /usr/local/rvm/gems/default
  Gem Path    /usr/local/rvm/gems/default:/usr/local/rvm/gems/default@global
  User Home   /home/vscode
  User Path   /home/vscode/.local/share/gem/ruby/3.1.0
  Bin Dir     /usr/local/rvm/gems/default/bin
Tools         
  Git         2.39.0
  RVM         1.29.12 (latest)
  rbenv       rbenv 1.2.0
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2022-12-24
Git SHA           f3175f033c
Released Version  true
```

## Bundler settings

```
app_config
[...]

PR #6875, opened 2023-08-12
Update bundle info man description

manuraj17:

Relates to https://github.com/rubygems/rubygems/issues/6212

`bundle info [GEM]` command's description didn't mention that only gem name is supported and adding the version was not required. The issue reported was around that confusion. One the suggestion in the discussion was to update the documentation which seemed like a fast thing to do for now.

Update the man info.

## Screenshot
image

## Note

This is my first PR in the repo, do let me know if I have missed any steps in the guidelines.
Also, do point out if I need to update in any other place other than this file.


bundler/lib/bundler/man/bundle-info.1
```diff
@@ -10,7 +10,7 @@
 \fBbundle info\fR [GEM] [\-\-path]
```
deivid-rodriguez, line 10: Maybe we can change to synopsis to use `GEM_NAME` instead? An also the description to be worded with something like

> Given a gem name present in your bundle, print the basic information about it such a homepage, version, path and summary\.

With thise clarifications in place, do you think we still need the explicit example?

manuraj17: @deivid-rodriguez I did think of this. Although, I was thinking that people may still try `GEM-VERSION` even if we say name. Hence, I added the example. Though - you folks would know better than me.
Thinking on it a bit more I think `GEM_NAME` is good enough. We can let the users catchup and learn that `GEM_NAME` means without the version. I will go ahead and make the changes once you reply and if you have no other objections.

manuraj17: Okay, I agree you with here. I felt there were more reports. I will make changes as you suggested.

Synposis from`GEM` to `GEM_NAME`
Description to 
```
Given a gem name present in your bundle, print the basic information about it such a homepage, version, path and summary.
```

deivid-rodriguez: There's a lint failure about man pages being out of sync. Can you run `rake man:build` and commit any differences? That should get the build back to green.

manuraj17: @deivid-rodriguez Have pushed the same. There were a bunch of changes as all the man pages got updated with current month. Is that what is expected?

deivid-rodriguez: Yes, that's exactly what was expected. The first documentation update each month also bumps the month in the man pages 👍.

deivid-rodriguez:
Rebased, squashed, and scheduled to auto-merge.

Thanks so much!

Issue #6742, opened 2023-06-14
Add file support to `ruby` method in Gemfile


indirect:
I use the `.ruby-version` file standard/convention to track the ruby version of my applications, since that file is respected by many version managers and other tools. Unfortunately, a different set of tools (mostly Heroku, buildpacks, and anything descended from that tradition) uses the Gemfile method `ruby` to track the Ruby version instead. This means I have to update the ruby version string in two places, and I am constantly missing one or the other.

To solve this today, I use this line in my Gemfile:

```ruby
ruby Pathname.new(__dir__).join(".ruby-version").read.strip
```

That technically works, but is pretty ugly. I have to think about what is happening every time I read it, and I would like to instead understand instantly "oh the ruby version is coming from .ruby-version". To improve that situation I propose adding an (optional) `file:` argument to the `ruby` method:

```ruby
ruby file: ".ruby-version"
```

How does that sound to everyone else?

deivid-rodriguez:
@indirect I think that'd be useful and more readable than the above hack indeed!

hsbt:
I'm also +1 for this. How about like this?

```
ruby :version_file # => handle `.ruby-version`, `.tool-version` and `.rbenv-version` or something.
```

We may prevent to pass arbitrary filename.

indirect:
Can you say more about how it helps us to prevent arbitrary filenames? I feel like we tried something similar with `source :rubygems` but had to change to strings because users needed to choose http or https. So if there is a future new file, we would have to edit the Bundler source.

I would like to be able to see which file to look in based on reading the Gemfile--if I see `ruby :version_file` I don't know what file to open, so if we limit arbitrary values maybe we could keep a string but have a list of allowed filenames?

clairity:
to add to the discussion, here's my version of using `.ruby-version` as the canonical ruby version declaration:

```ruby
# Gemfile
ruby File.read('.ruby-version').strip rescue '~>3.2.0'
```

which i learned from https://andycroll.com/ruby/read-ruby-version-in-your-gemfile/, and modified to add a good-enough guard clause in case the file gets deleted for some reason.

hsbt:
>Can you say more about how it helps us to prevent arbitrary filenames? 

In security team of Ruby core, we will handle implicitly code evaluation on like configuration file or dependency list file. I didn't consider the attack surface and vector for `file: '.ruby-version'` yet.

Maybe I worry too much. `:verson_file` symbol is not strong opinion.

[...]

PR #6876, opened 2023-08-13
Add `file` option to `ruby` method in Gemfile

ngan:
Fixes https://github.com/rubygems/rubygems/issues/6742. Please read the original issue for full context.

You can now do this in your Gemfile:
```
ruby file: '.ruby-version'
```

### DIFF
--- a/bundler/lib/bundler/man/gemfile.5
+++ b/bundler/lib/bundler/man/gemfile.5
@@ -85,6 +85,19 @@ ruby "3\.1\.2"
 .
 .IP "" 0
 .
+.P
+If you wish to derive your Ruby version from a version file (ie \.ruby\-version), you can use the \fBfile\fR option instead\.
+.
+.IP "" 4
+.
+.nf
+
+ruby file: "\.ruby\-version"
+.
+.fi
+.
+.IP "" 0
+.
 .SS "ENGINE"
 Each application \fImay\fR specify a Ruby engine\. If an engine is specified, an engine version \fImust\fR also be specified\.
 .
--- a/bundler/lib/bundler/man/gemfile.5.ronn
+++ b/bundler/lib/bundler/man/gemfile.5.ronn
@@ -69,6 +69,11 @@ should be the Ruby version that the engine is compatible with.
 
     ruby "3.1.2"
 
+If you wish to derive your Ruby version from a version file (ie .ruby-version),
+you can use the `file` option instead.
+
+    ruby file: ".ruby-version"
+
 ### ENGINE
 
 Each application _may_ specify a Ruby engine. If an engine is specified, an
--- a/bundler/lib/bundler/ruby_dsl.rb
+++ b/bundler/lib/bundler/ruby_dsl.rb
@@ -5,8 +5,14 @@ module Bundler
     def ruby(*ruby_version)
       options = ruby_version.last.is_a?(Hash) ? ruby_version.pop : {}
       ruby_version.flatten!
+
       raise GemfileError, "Please define :engine_version" if options[:engine] && options[:engine_version].nil?
       raise GemfileError, "Please define :engine" if options[:engine_version] && options[:engine].nil?
+
+      if options[:file]
+        raise GemfileError, "Cannot specify version when using the file option" if ruby_version.any?
+        ruby_version << Bundler.read_file(options[:file]).strip
+      end
 
       if options[:engine] == "ruby" && options[:engine_version] &&
          ruby_version != Array(options[:engine_version])
--- a/bundler/spec/bundler/ruby_dsl_spec.rb
+++ b/bundler/spec/bundler/ruby_dsl_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe Bundler::RubyDsl do
 
   let(:dsl) { MockDSL.new }
   let(:ruby_version) { "2.0.0" }
+  let(:ruby_version_arg) { ruby_version }
   let(:version) { "2.0.0" }
   let(:engine) { "jruby" }
   let(:engine_version) { "9000" }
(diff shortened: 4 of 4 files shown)


deivid-rodriguez:
Thanks!

The original specification by indirect seems better to me though? I think it could get confusing if `ruby '3.1.0', file: '.ruby-version'` ends up using a different ruby than 3.1.0? What the use case for the extra flexibility?

martinemde:
I know @hsbt brought up something similar, but I'll ask anyway. Could `ruby :file` be a shortcut for the very most common filename of `.ruby-version`. 99% of all gemfiles will have `ruby file: '.ruby-version'` once this change is widely adopted.

ngan:
> Could ruby :file be a shortcut for the very most common filename of .ruby-version.

I personally think it's too much implicitness. Typing out `.ruby-version` makes it very clear where it's coming from for new Ruby users. Similarly, `source 'https://rubygems.org'` is at the top of 99% of Gemfiles out there...but I'm guessing it's there for the same reason. 🤷

martinemde (approved):

martinemde:
One last thought: should we add `ruby version: '3.2.2'` for consistency?

deivid-rodriguez (approved):

deivid-rodriguez (commented):
Actually, this should be documented in the `gemfile.5` man page. Can we add a note about the new option?

ngan:
@deivid-rodriguez done!

deivid-rodriguez (approved):
Awesome! Rebased and squashed, and going in!

simi:
> Does this feature work with ruby 3.3.0.preview1? I'm getting an error message: Your Ruby version is 3.3.0.preview1, but your Gemfile specified

It is not part of Ruby 3.3.0.preview1 per my understanding, but it was released in Bundler 2.4.19 version. Just ensure you have right version.

jasnow:
This issue is "rvm list" shows "3.0.0.preview1" (period before preview1) where .ruby-versions and bundler want "-" before preview1. Now have a "fix".

alec-c4:
What about support of [asdf](https://asdf-vm.com/)? Because they have slightly different syntax.

simi:
> What about support of [asdf](https://asdf-vm.com/)? Because they have slightly different syntax.

@alec-c4 support for `.tool-versions` was added at https://github.com/rubygems/rubygems/pull/6898

alec-c4:
> > What about support of [asdf](https://asdf-vm.com/)? Because they have slightly different syntax.
> 
> @alec-c4 support for `.tool-versions` was added at #6898

Yay!!!!!! Such an awesome news!!!!

pboling:
This doesn't quite work right.
```
[!] There was an error parsing `Gemfile`: Illformed requirement ["ruby 2.7.8\ndirenv 2.32.2\nmongodb 6.2.1"]. Bundler cannot continue.

 #  from /Volumes/🐉🩸/src/apps/counting_sheep/Gemfile:1
 >  ruby file: ".tool-versions"
 #  
```
where my `.tool-versions` file has:
```
ruby 2.7.8
direnv 2.32.2
mongodb 6.2.1
```

hsbt:
@pboling Because we didn't backport https://github.com/rubygems/rubygems/pull/6898 into stable version yet.

pboling:
The problem is `Gem::Requirement.parse` is not sufficient for either `.ruby-version` or `.tool-versions`, and the spec added in this PR is too mocky to detect the issue.
```
before { allow(Bundler).to receive(:read_file).with("foo").and_return("#{version}\n") }
```

martinemde:
It seems we need better tests for this. Thank you for reporting.

pboling:
You could argue that it is not mocky enough. 👍  Perhaps a test that mocks the parse result for each distinct standard format of `.ruby-version`.

martinemde:
I'm looking into fixing this now. FWIW this might be [the original proposal for .ruby-version from 2012](https://gist.github.com/fnichol/1912050) and has received running updates along the way as things were decided.

I'm of two minds for this behavior. One the one hand, Bundler could be opinionated about the format. We could simply say that if you put something into .ruby-version that is not a simple version, then Bundler won't read it.

On the other hand, this format is not up to bundler and has existed for a long time. It's used by rbenv, chruby, rvm and we have chosen to attempt to read asdf's format.

It's trivial to trim ruby-  off the front and `@gemset` off the end, but if users of bundler can just tweak their .ruby-version file with no ill effect, then keeping this simple would also produce a narrowing of usage of `.ruby-version`, and dare I even say it, a standard format.

pboling:
For those who rely on the `@gemset` to provide encapsulaiton between multiple installations on a single VPS, as I used to many years ago, not supporting it would mean they can't use the `file:` feature at all... but maybe that's not the end of the world?  I do think that at minimum the `ruby-` can be stripped, because even those who are not using gemsets may be using it, and supporting it will prevent confused bug reports from users attempting to try out `file:` and having it blow up.

martinemde:
Thanks for taking up this discussion and for making sure we solve this as well as possible.

For `@gemset`, I see that RVM [documents](https://rvm.io/workflow/projects#project-file-ruby-version)  that `@gemset` is not compatible with other version managers. They suggest that users should rely on the `.ruby-gemset` file for that feature. I feel good excluding that behavior since it's definitely non-standard and the work around is easy and recommended.

I am patching this to support removing `ruby-` since that seems to be fairly common. I'll update with a ticket number when I push.

Also for jruby/truffleruby folks out there, you will end up with something like this:

```ruby
ruby file: ".ruby-version", engine: "jruby", engine_version: "9.4.3.0"
```

```sh
$ cat .ruby-version
3.1.4
```

Issue #6724, opened 2023-06-07
Suppress 'licenses is empty' warning from `gem build` for private gems


dgholz:
### Describe the problem as clearly as you can

RubyGems specification reference says to leave license empty for commercial/proprietary/all rights reserved gem, but `gem build` shows a warning:
```
WARNING:  licenses is empty, but is recommended.  Use a license identifier from
http://spdx.org/licenses or 'Nonstandard' for a nonstandard license.
```

The spec says
> Not specifying a license means all rights are reserved; others have no right to use the code for any purpose.

This exactly matches the requirement for the gem (an internal one). Is it possible to provide a value for `license` that means the same as not specifying a value? Or another way to avoid the warning while leaving `license` unspecified?

I didn't understand how `Nonstandard` should be interpreted from reading the specification, nor from the list on SPDX.

I see #3665 & #5253, they didn't help me understand how to:
1. Avoid the warning about a missing `license`
2. Note that I want all rights reserved in the same way as not specifying `license`

### Did you try upgrading RubyGems?

Yes, am on version `3.4.13`

### Post steps to reproduce the problem

1. Make a gemspec like
```
Gem::Specification.new do |s|
  s.name = 'pincushion'
  s.version = '0.1.0'
  s.authors = ['dgholz']
  s.summary = "A real gem for trying things out"
  s.homepage = "https://github.com/dgholz/pincushion"
  s.files = `git ls-files`.split("\n")
end
```
3. Run `gem build`

### Which command did you run?

`gem build`

### What were you expecting to happen?

The gem is built without warnings.

### What actually happened?

```
WARNING:  licenses is empty, but is recommended.  Use a license identifier from
http://spdx.org/licenses or 'Nonstandard' for a nonstandard license.
WARNING:  See https://guides.rubygems.org/specification-reference/ for help
  Successfully built RubyGem
  Name: pincushion
  Version: 0.1.0
  File: pincushion-0.1.0.gem
```

### Run `gem env` and paste the output below

```
RubyGems Environment:
  - RUBYGEMS VERSION: 3.4.13
  - RUBY VERSION: 3.1.2 (2022-04-12 patchlevel 20) [arm64-darwin21]
  - INSTALLATION DIRECTORY: /Users/danielholz/.gem/ruby/3.1.2
  - USER INSTALLATION DIRECTORY: /Users/danielholz/.gem/ruby/3.1.0
[...]

PR #6879, opened 2023-08-14
Extend license spec to accept `nil` 

jhong97:
Closes issue #6724 

- We would like to have a way to avoid the missing license warning for an all-rights-reserved gem
   - However, the license warning should still appear when the license spec is not specified, in case of unintentional omission
- Discussions preferred to have `nil` to represent all-rights-reserved and prevent the missing license warning from triggering

- Make `nil` a valid value in the `licenses` spec array by updating checks in the `specification_policy`
- Raise an error when attempts to add 2 valid licenses of different classes (a.k.a a `nil` and a `string`) are made.
  - Reserving all rights while having other licenses seems contradictory


test/rubygems/test_gem_specification.rb
```diff
@@ -3090,6 +3090,28 @@ def test_validate_license
...
+
+  def test_validate_licenses_containing_multiple_types
+    util_setup_validate
+
+    e = assert_raise Gem::InvalidSpecificationException do
+      @a1.licenses = [nil, "MIT"]
+      @a1.validate
+    end
+
+    assert_match "licenses must be of the same class, detected 2 classes", e.message
```
martinemde, line 3112: This message is somewhat confusing to me. Could we say something more clear and specific. Something like "Licenses field cannot include nil with other licenses."

I haven't thought about the phrasing very much but I'd like it to be more specific about nil not being allowed in the licenses array unless it is the only element.

lib/rubygems/specification.rb
```diff
@@ -350,7 +350,8 @@ def authors=(value)
   # You should specify a license for your gem so that people know how they are
   # permitted to use it and any restrictions you're placing on it.  Not
   # specifying a license means all rights are reserved; others have no right
-  # to use the code for any purpose.
+  # to use the code for any purpose. You can also specifically set the license
+  # to nil to specify all rights are reserved.
```
martinemde, line 354: I missed this discussion so forgive me if this was already discussed.

I don't think we should give a meaning to nil. Nil is nil, no license. The meaning of that is unknown without more information. I would therefore remove this comment and any other text that might imply some (legal-like) meaning from `nil`.

If you want an all right reserved license, would it be reasonable to instead add that specifically?

lib/rubygems/specification_policy.rb
```diff
@@ -367,8 +369,14 @@ def validate_licenses_length
   def validate_licenses
     licenses = @specification.licenses
 
+    num_classes = licenses.map(&:class).uniq.size
+
+    if !licenses.empty? && num_classes != 1
+      error "licenses must be of the same class, detected #{num_classes} classes"
+    end
+
```
martinemde, line 377: Given my other comment, can we simply ignore nil if it is present, but not print the unwanted warning if it is explicitly set to nil? There is no conflict having nil in an array if we don't try to overload nil with meaning.

lib/rubygems/specification.rb
deivid-rodriguez: Discussion is at https://github.com/rubygems/rubygems/issues/6724. My understanding from the error message is that license is supposed to be filled with an opensource license. There for absence of a license means the gem could be potentially private. The whole point is to allow the warning about missing license to be avoided in this case.

Thinking about it again, you're right that adding a special value for this case may be better. Maybe the whole problem is that the current "Nonstandard" special value is not good enough for private gem, since "Nonstandard" was probably meant for Open Source licenses that have not yet been standardized. So we could add a different special value for private gems.

jhong97: Thanks for the review folks! 

To play devil's advocate, since the convention has been for no license representing all rights reserved, wouldn't `nil` keep with that convention compared to a new specific value for private gems? As in the `nil` case, the devs still haven't specified a license.

martinemde: [The first google result](https://choosealicense.com/no-permission/#:~:text=When%20you%20make%20a%20creative,shake%2Ddowns%2C%20or%20litigation.) says similar to what we've arrived at here. No license just means you get whatever happens when you don't specify a license.

The simplest solution here is "nil removes the error and makes no claim about any license, non license, or otherwise." I suggest simply removing the extra checking and allowing nil. If you specify nil and MIT, then you get MIT because nil makes no claim about licensing.

test/rubygems/test_gem_specification.rb
jhong97: Based on conversations below, removed this check and associated test

martinemde (approved):

deivid-rodriguez (approved):

deivid-rodriguez:
I have one more little comment. Right now this feature is completely hidden in the source code/specs. Can we make the warning a little bit more helpful, while still avoiding to give a meaning to a `nil` license? Like

> licenses is empty, but is recommended.  Use a license identifier from
http://spdx.org/licenses or '#{Gem::Licenses::NONSTANDARD}' for a nonstandard license, or set it to `nil` if you don't want to specify a license.

Would that be acceptable?

martinemde (approved):

jhong97:
Was missing a few more tests that needed updating after updating the warning

martinemde:
I looked through everything else mentioning licenses and I think you covered it all. Merging.

dgholz:
@jhong97 amazing, thank you so much

PR #6884, opened 2023-08-16
Improve memory usage in Bundler::Settings, and thus improve boot time

technicalpickles:

I was running [memory_profiler](https://github.com/SamSaffron/memory_profiler) on our Rails app boot-time, and saw some opportunities.

I used this code to scope memory allocations in our rails app to just bundler:

```ruby
require 'memory_profiler'

MemoryProfiler.report(allow_files: 'bundler/settings.rb') do
  require_relative 'config/environment'
end.pretty_print
```

And then compared it with these changes by adding adding calling it with `ruby -Ipath/to/checkout/rubygems/bundler/lib`.

### Memory

Before:

```
Total allocated: 4883872 bytes (65211 objects)
Total retained:  22644 bytes (226 objects)


allocated memory by file
-----------------------------------
   4883872  /Users/josh.nichols/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/bundler-2.4.18/lib/bundler/settings.rb

allocated objects by file
-----------------------------------
     28606  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb

retained memory by file
-----------------------------------
     21333  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb

retained objects by file
-----------------------------------
       226  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb
```

After:

```
Total allocated: 2326771 bytes (28606 objects)
Total retained:  20207 bytes (221 objects)

allocated memory by file
-----------------------------------
   2326771  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb

allocated objects by file
-----------------------------------
     28606  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb


retained memory by file
-----------------------------------
     20207  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb

retained objects by file
-----------------------------------
       221  /Users/josh.nichols/workspace/rubygems/bundler/lib/bundler/settings.rb
```

### Runtime

For runtime, I with the same file above, but without the memory_profiling:

```ruby
require_relative 'config/environment'
```

I used [hyperfine](https://github.com/sharkdp/hyperfine) to compare before/after: `hyperfine "ruby boot-time-benchmark.rb" "ruby -I/Users/josh.nichols/workspace/rubygems/bundler/lib boot-time-benchmark.rb"`:

CleanShot 2023-08-16 at 15 25 27

### Summary

This reduces the amount of memory allocated during Rails startup in our app by 52%, from 4_883_872 bytes to 2_326_771 bytes.

It also reduces startup time by 6%, from 8.807s to 8.256s.

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -340,7 +340,8 @@ module Bundler
     end
 
     def is_bool(name)
-      BOOL_KEYS.include?(name.to_s) || BOOL_KEYS.include?(parent_setting_for(name.to_s))
+      name = name.to_s
+      BOOL_KEYS.include?(name) || BOOL_KEYS.include?(parent_setting_for(name))
     end
 
     def is_string(name)
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -90,8 +90,10 @@ module Bundler
     def initialize(root = nil)
       @root            = root
       @local_config    = load_config(local_config_file)
+
       @env_config      = ENV.to_h
-      @env_config.select! {|key, _value| key.start_with?("BUNDLE_") }
+      @env_config.select! {|key, _value| key.start_with?("BUNDLE_")}
+      @env_config.delete("BUNDLE_")
 
       @global_config   = load_config(global_config_file)
       @temporary       = {}
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -412,6 +412,8 @@ module Bundler
     end
 
     def converted_value(value, key)
+      key = key.to_s
+
       if is_array(key)
         to_array(value)
       elsif value.nil?
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -354,11 +354,7 @@ module Bundler
     def to_bool(value)
       case value
       when String
-        if value.match?(/\A(false|f|no|n|0|)\z/i)
-          false
-        else
-          true
-        end
+        value.match?(/\A(false|f|no|n|0|)\z/i) ? false : true
       when nil, false
         false
       else
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -161,7 +161,7 @@ module Bundler
     def local_overrides
       repos = {}
       all.each do |k|
-        repos[$'] = self[k] if k.start_with?("local.")
+        repos[k.delete_prefix("local.")] = self[k] if k.start_with?("local.")
       end
       repos
     end
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -351,7 +351,13 @@ module Bundler
 
     def to_bool(value)
       case value
-      when nil, /\A(false|f|no|n|0|)\z/i, false
+      when String
+        if value.match?(/\A(false|f|no|n|0|)\z/i)
+          false
+        else
+          true
+        end
+      when nil, false
         false
       else
         true
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -148,9 +148,14 @@ module Bundler
     def all
       keys = @temporary.keys.union(@global_config.keys, @local_config.keys, @env_config.keys)
 
-      keys.map do |key|
-        key.sub(/^BUNDLE_/, "").gsub(/___/, "-").gsub(/__/, ".").downcase
-      end.sort
+      keys.map! do |key|
+        key = key.delete_prefix("BUNDLE_")
+        key.gsub!("___", "-")
+        key.gsub!("__", ".")
+        key.downcase!
+        key
+      end.sort!
+      keys
     end
 
     def local_overrides
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -99,7 +99,11 @@ module Bundler
 
     def [](name)
       key = key_for(name)
-      value = configs.values.map {|config| config[key] }.compact.first
+
+      values = configs.values
+      values.map! {|config| config[key] }
+      values.compact!
+      value = values.first
 
       converted_value(value, name)
     end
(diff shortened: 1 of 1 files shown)


indirect:
Nice! A boon for the massively-environmented users everywhere 😂

technicalpickles:
I think I got over-ambitious with this PR, and now I'm getting tons of failures that aren't super obvious where to begin (ie `expected some big string, but got empty string`) so will break some of these out.

bundler/lib/bundler/settings.rb
```diff
@@ -90,14 +90,20 @@ class Settings
     def initialize(root = nil)
       @root            = root
       @local_config    = load_config(local_config_file)
-      @env_config      = ENV.to_h.select {|key, _value| key =~ /\ABUNDLE_.+/ }
+      @env_config      = ENV.to_h
+      @env_config.select! {|key, _value| key.start_with?("BUNDLE_") }
```
segiddins, line 94: this is technically different behavior, `BUNDLE_` would not match the regexp but would match the `start_with?`, but I doubt it matters

bundler/lib/bundler/settings.rb
```diff
@@ -140,17 +146,22 @@ def set_global(key, value)
...
 
     def all
-      keys = @temporary.keys | @global_config.keys | @local_config.keys | @env_config.keys
+      keys = @temporary.keys.union(@global_config.keys, @local_config.keys, @env_config.keys)
 
-      keys.map do |key|
-        key.sub(/^BUNDLE_/, "").gsub(/___/, "-").gsub(/__/, ".").downcase
-      end.sort
+      keys.map! do |key|
+        key = key.delete_prefix("BUNDLE_")
```
segiddins, line 152: should we handle when this resolves to an empty key?

bundler/lib/bundler/settings.rb
```diff
@@ -140,17 +146,22 @@ def set_global(key, value)
...
+        key
+      end.sort!
+      keys
     end
 
     def local_overrides
       repos = {}
       all.each do |k|
-        repos[$'] = self[k] if k =~ /^local\./
+        repos[$'] = self[k] if k.start_with?("local.")
```
martinemde, line 164: Does the `$'` global get set when you didn't match a regexp?

I'm not sure what the simplest replacement is. Maybe...

```ruby
repos[k.delete_prefix("local.")] = self[k] if k.start_with?("local.")
```

bundler/lib/bundler/settings.rb
```diff
@@ -329,16 +340,24 @@ def split_specific_setting_for(name)
...
 
     def to_bool(value)
       case value
-      when nil, /\A(false|f|no|n|0|)\z/i, false
+      when String
+        if value.match?(/\A(false|f|no|n|0|)\z/i)
+          false
+        else
+          true
+        end
```
martinemde, line 359:
```suggestion
         value !~ /\A(false|f|no|n|0|)\z/i
```

I'm not sure if that's equally performant, but it saves a few lines.

segiddins: could also do `!value.match?(/\A(false|f|no|n|0|)\z/i)`

bundler/lib/bundler/settings.rb
segiddins: no it does not, good catch!

technicalpickles: ah-ha! that might be causing the test failures I'm seeing (I'm still bisecting to track down which introduced them)

bundler/lib/bundler/settings.rb
martinemde: I like your solution better. Docs say that match? doesn't update the globals, so it's probably slightly more performant.

bundler/lib/bundler/settings.rb
martinemde:
```suggestion
        repos[k.delete_prefix("local.")] = self[k] if k.start_with?("local.")
```

martinemde: I committed this change so we can see if the tests pass.

martinemde: Looks like they're passing now. yay!

technicalpickles: Awesome, thanks!

bundler/lib/bundler/settings.rb
technicalpickles: I did think about doing it on a single line. I know it's a few lines, but I find it a bit easier to read. could do as a ternary operator too

```ruby
value.match?(/\A(false|f|no|n|0|)\z/i) ? false : true
```

Checked w/ benchmark-ips and they are the same, so I'm open to whatever style y'all prefer.

bundler/lib/bundler/settings.rb
technicalpickles: Good catch. If that is the case, we can reject/delete it explicitly here. Or handle it in the other spot you mentioned down in `all`

bundler/lib/bundler/settings.rb
deivid-rodriguez: I also like @segiddins' version!

bundler/lib/bundler/settings.rb
technicalpickles: Some ideas:

```ruby
@env_config.select! {|key, _value| key.start_with?("BUNDLE_") && key != "BUNDLE_" }
# or
@env_config.select! {|key, _value| key.start_with?("BUNDLE_")}
@env_config.delete("BUNDLE_")
```

martinemde: Second one look good. Then you're not checking each loop.

bundler/lib/bundler/settings.rb
martinemde: I prefer one line, but don't care between ternary and `!` and sorry for starting the nitpick thread here. 😆

technicalpickles:
@martinemde pushed up those changes!

segiddins (approved):

deivid-rodriguez:
Hei, I didn't fin time to chime in here, but THIS IS SO AWESOME! Thank you so much!

PR #6889, opened 2023-08-17
Ensure that loading multiple gemspecs with legacy YAML class references does not warn

segiddins:
Before this, you would get constant redefinition warnings on Psych::DefaultKey

Additionally, ensure the retries wont continue infinitely in the case of the ArgumentError not being caused by Marshal trying to load the undefined classes

I was getting many constant redefinition warnings when attempting to load every gemspec on RubyGems.org in a single process

Check for constants before defining them

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -1302,10 +1302,20 @@ class Gem::Specification < Gem::BasicSpecification
     Gem.load_yaml
 
     yaml_set = false
+    retry_count = 0
 
     array = begin
       Marshal.load str
     rescue ArgumentError => e
+      # Avoid an infinite retry loop when the argument error has nothing to do
+      # with the classes not being defined.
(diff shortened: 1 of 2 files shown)


lib/rubygems/specification.rb
```diff
@@ -1302,10 +1302,20 @@ def self._load(str)
...
 
     array = begin
       Marshal.load str
     rescue ArgumentError => e
+      # Avoid an infinite retry loop when the argument error has nothing to do
+      # with the classes not being defined.
+      # 1 retry each allowed in case all 3 of
+      # - YAML
+      # - YAML::Syck::DefaultKey
+      # - YAML::Private type
```
deivid-rodriguez, line 1315:
```suggestion
      # - YAML::PrivateType
```

deivid-rodriguez (approved):
Make sense!

You must've dug really deep to find these things! 💪

PR #6896, opened 2023-08-18
Replace Marshal.load with a fully-checked safe gemspec loader

segiddins:

Ruby has historically had many security issues while loading serialized YAML or Marshal data, with the worst outcome being remote code execution. RubyGems and Bundler both load gemspecs from gem source URLs provided by users. To prevent exploits by malicious gem servers against developer machines installing gems, we need to load only a safe subset of Ruby classes and objects.

Previously, we implemented `YAML.safe_load` to prevent exploits by malicious gem servers via crafted YAML files. Until now, we have not been able to provide the same defenses for marshaled gemspecs due to the lack of a `Marshal.safe_load`.

This PR implements `Marshal.safe_load`, a loader for marshaled data that excludes unknown objects and classes. The loader prevents exploits by malicious gem servers, or crafted marshaled data provided to RubyGems or Bundler.

Implement a new reader / AST visitor for Marshal documents, following https://rubyreferences.github.io/rubyref/builtin/marshal.html and using the `Psych` `ToRuby` visitor for inspiration.

As a followup, will conditionally use the new module in Bundler

### DIFF
--- a/test/rubygems/test_gem_safe_marshal.rb
+++ b/test/rubygems/test_gem_safe_marshal.rb
@@ -6,6 +6,120 @@ require "date"
 require "rubygems/safe_marshal"
 
 class TestGemSafeMarshal < Gem::TestCase
+  define_method("test_safe_load_marshal Date #<Date: 1994-12-09 ((2449696j,0s,0n),+0s,2299161j)>") { assert_safe_load_marshal "\x04\bU:\tDate[\vi\x00i\x03 a%i\x00i\x00i\x00f\f2299161" }
+  define_method("test_safe_load_marshal Float 0.0") { assert_safe_load_marshal "\x04\bf\x060" }
+  define_method("test_safe_load_marshal Float -0.0") { assert_safe_load_marshal "\x04\bf\a-0" }
+  define_method("test_safe_load_marshal Float Infinity") { assert_safe_load_marshal "\x04\bf\binf" }
+  define_method("test_safe_load_marshal Float -Infinity") { assert_safe_load_marshal "\x04\bf\t-inf" }
+  define_method("test_safe_load_marshal Float NaN") { assert_safe_load_marshal "\x04\bf\bnan", equality: false }
+  define_method("test_safe_load_marshal Float 1.1") { assert_safe_load_marshal "\x04\bf\b1.1" }
+  define_method("test_safe_load_marshal Float -1.1") { assert_safe_load_marshal "\x04\bf\t-1.1" }
+  define_method("test_safe_load_marshal Float 30000000.0") { assert_safe_load_marshal "\x04\bf\b3e7" }
+  define_method("test_safe_load_marshal Float -30000000.0") { assert_safe_load_marshal "\x04\bf\t-3e7" }
+  define_method("test_safe_load_marshal Gem::Version #<Gem::Version \"1.abc\">") { assert_safe_load_marshal "\x04\bU:\x11Gem::Version[\x06I\"\n1.abc\x06:\x06ET" }
+  define_method("test_safe_load_marshal Hash {}") { assert_safe_load_marshal "\x04\b}\x00[\x00" }
+  define_method("test_safe_load_marshal Hash {:runtime=>:development}") { assert_safe_load_marshal "\x04\bI{\x06:\fruntime:\x10development\x06:\n@type[\x00", permitted_ivars: { "Hash" => %w[@type] } }
+  define_method("test_safe_load_marshal Integer -1") { assert_safe_load_marshal "\x04\bi\xFA" }
+  define_method("test_safe_load_marshal Integer -1048575") { assert_safe_load_marshal "\x04\bi\xFD\x01\x00\xF0" }
+  define_method("test_safe_load_marshal Integer -122") { assert_safe_load_marshal "\x04\bi\x81" }
+  define_method("test_safe_load_marshal Integer -123") { assert_safe_load_marshal "\x04\bi\x80" }
+  define_method("test_safe_load_marshal Integer -124") { assert_safe_load_marshal "\x04\bi\xFF\x84" }
+  define_method("test_safe_load_marshal Integer -127") { assert_safe_load_marshal "\x04\bi\xFF\x81" }
+  define_method("test_safe_load_marshal Integer -128") { assert_safe_load_marshal "\x04\bi\xFF\x80" }
+  define_method("test_safe_load_marshal Integer -2") { assert_safe_load_marshal "\x04\bi\xF9" }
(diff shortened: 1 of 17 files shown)


test/rubygems/test_gem_safe_marshal.rb
```diff
@@ -0,0 +1,144 @@
...
+  end
+
+  def assert_safe_load_as(x, additional_methods: [])
+    dumped = Marshal.dump(x)
+    loaded = Marshal.load(dumped)
+    safe_loaded = Gem::SafeMarshal.safe_load(dumped)
+
+    # NaN != NaN, for example
+    if x == x # rubocop:disable Lint/BinaryOperatorWithIdenticalOperands
+      # assert_equal x, safe_loaded, "should load #{dumped.inspect}"
```
indirect, line 133: intentional comment?

indirect (approved):
Thanks for doing this! Ensuring that we can load marshaled specs safely has been on our list of good ideas that seem hard for several years. :tada:

test/rubygems/test_gem_safe_marshal.rb
segiddins: whoops, no.

segiddins:
```suggestion
      assert_equal x, safe_loaded, "should load #{dumped.inspect}"
```

segiddins:
Specs here finally passed! @deivid-rodriguez if you want to take a look

indirect:
Updated the PR title and description with an explanation of the problem and how this PR solves the problem. 👍🏻

test/rubygems/test_gem_safe_marshal.rb
```diff
@@ -0,0 +1,336 @@
...
+    end
+  end
+
+  def assert_safe_load_as(x, **kwargs)
+    dumped = Marshal.dump(x)
+    equality = x == x # rubocop:disable Lint/BinaryOperatorWithIdenticalOperands
+    assert_safe_load_marshal(dumped, equality: equality, **kwargs)
+  end
+
+  def with_const(mod, name, new_value, &block)
```
simi, line 323: This is used to temporarily modify `PERMITTED_IVARS` of `Gem::SafeMarshal`, since that is used in `safe_load` method. Wouldn't be easier using `Gem::SafeMarshal.load` with `permitted_ivars` attribute for such a purpose directly?

test/rubygems/test_gem_safe_marshal.rb
```diff
@@ -0,0 +1,336 @@
...
+
+    assert_safe_load_as Rational(1, 3)
+  end
+
+  [
+    0, 1, 2, 3, 4, 5, 6, 122, 123, 124, 127, 128, 255, 256, 257,
+    2**16, 2**16 - 1, 2**20 - 1,
+    2**28, 2**28 - 1,
+    2**32, 2**32 - 1,
+    2**63, 2**63 - 1
```
simi, line 260: Is there any specific reason for those numbers? Maybe comment would be handy.

simi (approved):
This is HUUUUUGE! Big applause for reporting found issues in JRuby and Truffleruby, Truffleruby one seems already fixed :muscle: (unreleased).

I can imagine packing this as separate gem at some point and vendoring back using Automatiek, since it can be super useful outside of RubyGems/Bundler.

Is there any plan on testing this on real gems in bulk?

segiddins:
I tested it on every gemspec in rubygems.org on ruby 2.7, 3.2, jruby 9.4, and truffleruby 23. only a few gemspecs could not load because they used some random symbols, all were very old (plus another few that can't properly be downloaded)

simi:
> I tested it on every gemspec in rubygems.org on ruby 2.7, 3.2, jruby 9.4, and truffleruby 23. only a few gemspecs could not load because they used some random symbols, all were very old (plus another few that can't properly be downloaded)

I was thinking to test this against top 100/1000 gems by download (all versions, to include also old releases) in bulk somehow.

indirect:
@simi It seems like a lot of work to dynamically calculate the most popular gems and download their gemspecs in a test... since we have tested against existing gemspecs, maybe we add a few existing marshaled gemspecs to the system tests, and add some code that verifies gemspecs can go through `Marshal.dump`/`Marshal.safe_load` as one of the steps during `gem push`?

simi:
> @simi It seems like a lot of work to dynamically calculate the most popular gems and download their gemspecs in a test... since we have tested against existing gemspecs, maybe we add a few existing marshaled gemspecs to the system tests, and add some code that verifies gemspecs can go through `Marshal.dump`/`Marshal.safe_load` as one of the steps during `gem push`?

I meant this for one off test before release to ensure everything works properly. Clearly this is not good idea to add to test suite or CI. I can craft something locally and report back.

indirect:
@simi ohh ok! I think we are good then, since @segiddins has already tested every gemspec for every gem?

headius:
I have fixed one of the issues reported. Thanks for letting us know! Obviously there's still gaps in Marshal specs and tests.

rondales:
> ## What was the end-user or developer problem that led to this PR?
> Ruby has historically had many security issues while loading serialized YAML or Marshal data, with the worst outcome being remote code execution. RubyGems and Bundler both load gemspecs from gem source URLs provided by users. To prevent exploits by malicious gem servers against developer machines installing gems, we need to load only a safe subset of Ruby classes and objects.
> 
> Previously, we implemented `YAML.safe_load` to prevent exploits by malicious gem servers via crafted YAML files. Until now, we have not been able to provide the same defenses for marshaled gemspecs due to the lack of a `Marshal.safe_load`.
> 
> This PR implements `Marshal.safe_load`, a loader for marshaled data that excludes unknown objects and classes. The loader prevents exploits by malicious gem servers, or crafted marshaled data provided to RubyGems or Bundler.
> 
> ## What is your fix for the problem, implemented in this PR?
> Implement a new reader / AST visitor for Marshal documents, following https://rubyreferences.github.io/rubyref/builtin/marshal.html and using the `Psych` `ToRuby` visitor for inspiration.
> 
> As a followup, will conditionally use the new module in Bundler
> 
> ## Make sure the following tasks are checked
> * [x]  Describe the problem / feature
> * [x]  Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
> * [x]  Write code to solve the problem
> * [x]  Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

Issue #6895, opened 2023-08-18
Bundler support .tool-versions (ASDF)


gvkhna:
Although slim chance this will be accepted/considered. I’d like to propose it anyway. With Bundler 2.4.19 supporting .ruby-version. It would be great if it could take the version from .tool-versions. Which is the file ASDF uses. The current workaround of File.read is sufficient but thought I would throw it out there anyway.

martinemde:
I see that the `.tool-versions` file has multiple lines with one of them being, e.g. `ruby 3.2.2`. Any idea for a simple algorithm that would work for this case too?

gvkhna:
I use this in my Gemfile

`source "https://rubygems.org"
ruby(/^ruby\s(.*)$/.match(File.read(".tool-versions"))[1].strip)
`

The .tool-versions is very simple format doesn’t need to be over complicated.

martinemde:
This is [the line here](https://github.com/ngan/rubygems/blob/6d47ee98b9c6971222edb4e24ae5ecde3327c0a6/bundler/lib/bundler/ruby_dsl.rb#L14) that resolves the version in the file. I think if there's a simple version that's not error prone, then we'd accept it:

gvkhna:
I see, what i would do is if the file is named .tool-versions i would instead run the ruby regex match as shown. That seems simple enough to me.

gvkhna:
@martinemde Please check

martinemde:
@gvkhna can you set or reopen the PR to use rubygems:master as the base branch?

gvkhna:
Whoops sorry, thank you! Done.

ngan:
I have a feeling this is going to get really hairy really fast. We should draw the line somewhere. That logic and regex might get more complicated over time.

If we _really_ want to support known version specifiers, maybe we can extract it to symbols mapping to their own logic. For example:

```ruby
ruby from: :ruby_version # This would delegate to Bundler::RubyVersionResolvers::RubyVersion
ruby from: :tool_versions # This would delegate to Bundler::RubyVersionResolvers::ToolVersions
ruby from: :rvm # etc...
```

gvkhna:
That was my initial concern yes and valid reasoning. But without over complicating the situation the proposed PR still seems reasonable. I've never seen an rvm version file in the wild?

ngan:
> I've never seen an rvm version file in the wild?

Hehe, yeah, I have used RVM in years...but IIRC there's a `.rvmrc` file.

[...]

PR #6898, opened 2023-08-18
Support ruby file: ".tool-versions" in Gemfile

gvkhna:
Add support for .tool-versions via simple detect a line begins with "ruby"

Not supporting ASDF .tool-versions

Discussed more in the relevant issue: https://github.com/rubygems/rubygems/issues/6895

### DIFF
--- a/bundler/lib/bundler/man/gemfile.5.ronn
+++ b/bundler/lib/bundler/man/gemfile.5.ronn
@@ -74,6 +74,11 @@ you can use the `file` option instead.
 
     ruby file: ".ruby-version"
 
+The version file should conform to any of the following formats:
+
+  - `3.1.2` (.ruby-version)
+  - `ruby 3.1.2` (.tool-versions, read: https://asdf-vm.com/manage/configuration.html#tool-versions)
+
 ### ENGINE
 
 Each application _may_ specify a Ruby engine. If an engine is specified, an
--- a/bundler/lib/bundler/ruby_dsl.rb
+++ b/bundler/lib/bundler/ruby_dsl.rb
@@ -12,7 +12,7 @@ module Bundler
       if options[:file]
         raise GemfileError, "Cannot specify version when using the file option" if ruby_version.any?
         file_content = Bundler.read_file(Bundler.root.join(options[:file]))
-        matched_ruby_version = /^ruby\s(.*)$/.match(file_content)
+        matched_ruby_version = /^ruby\s+([\d.]+)(?:\s*#\s*.*)?$/.match(file_content)
         # if a line in the file begins with "ruby" use that else use the whole file
         ruby_version << matched_ruby_version.present? ? matched_ruby_version[1].strip : file_content.strip
       end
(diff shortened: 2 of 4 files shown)


martinemde:
Can you add another test just like [this one](https://github.com/rubygems/rubygems/pull/6876/files#diff-6bb6062f69412b844c9df6557a1bbdad24a327a96575098d260e3d146ec4ddceR107) that returns a typical .tool-versions file with multiple different tool versions to make sure it works?

gvkhna:
@martinemde Thank you for guiding on this!

bundler/lib/bundler/ruby_dsl.rb
```diff
@@ -11,7 +11,10 @@ def ruby(*ruby_version)
 
       if options[:file]
         raise GemfileError, "Cannot specify version when using the file option" if ruby_version.any?
-        ruby_version << Bundler.read_file(Bundler.root.join(options[:file])).strip
+        file_content = Bundler.read_file(Bundler.root.join(options[:file]))
+        matched_ruby_version = /^ruby\s+([\d.]+)(?:\s*#\s*.*)?$/.match(file_content)
+        # if a line in the file begins with "ruby" use that else use the whole file
+        ruby_version << matched_ruby_version ? matched_ruby_version[1].strip : file_content.strip
```
martinemde, line 17: I was thinking about how to handle comments too. I think it might be better to take the whole line, then split at `#` and use whatever comes before it. Here's an approach that hopefully isn't _too_ clever.

```suggestion
        if /^ruby\s+(.*)$/.match(file_content)
          ruby_version << $1.split("#", 2).first.strip
        else
          ruby_version << file_content.strip
        end
```

 Trying too hard to match the version number is going to cause problems since ruby versions can be specified in ways that aren't just standard version numbers.

gvkhna: I really like this solution, extremely reasonable and I would say nice and simple. Should work great and we don't need tons of code here for this simple task.

martinemde (approved):
Great! Thank you for contributing. I think it makes this feature much more complete.

bundler/lib/bundler/man/gemfile.5.ronn
```diff
@@ -74,6 +74,11 @@ you can use the `file` option instead.
 
     ruby file: ".ruby-version"
 
+The version file should conform to any of the following formats:
+
+  - `3.1.2` (.ruby-version)
```
pboling, line 79: It would make sense to support the same formats that `.ruby-version` supports, which includes `ruby-3.1.2` (note the dash).

bundler/lib/bundler/ruby_dsl.rb
```diff
@@ -11,7 +11,12 @@ def ruby(*ruby_version)
 
       if options[:file]
         raise GemfileError, "Cannot specify version when using the file option" if ruby_version.any?
-        ruby_version << Bundler.read_file(Bundler.root.join(options[:file])).strip
+        file_content = Bundler.read_file(Bundler.root.join(options[:file]))
```
pboling, line 14: Perhaps it is not intended to support the variability of `.ruby-version`, but this doesn't currently support the standard `ruby-3.1.2` syntax that is an option for `.ruby-version`.

PR #6899, opened 2023-08-18
Ruby version resolvers

ngan:
Introduce the concept of Ruby version resolvers. This is just a proof of concept. I haven't run this code nor written tests. It's just for discussion.

The idea is that Bundler can now support known Ruby version managers. The API is like so:

`Gemfile`
```ruby
ruby from: :ruby_version
# or
ruby from: :tool_versions
# or
ruby from: :rvmrc
```

FWIW, I'm not a fan of this...I think the entire ecosystem should agree on one way, and I think it's `.ruby-version`.


martinemde:
Thanks for submitting your ideas. I think it's elegant, but I'm inclined to wait for another request for any other format before we add more code here. It seems we have solved for the immediate needs that were previously requiring work arounds.

Let's revisit when we find the current solution lacking. Ideally, any other formats can conform to the standard we've created here, if possible.

PR #6915, opened 2023-08-24
Specify -c .rubocop.yml for bin/rubocop (performance)

martinemde:
Speeds up rubocop by 13% based on hyperfine, presumably because it doesn't need to look for the right config file.

Another attempt at #6913 which addresses this as a performance improvement, in addition to a temporary big fix.

Hyperfine shows a 13% improvement over the identical command and environment without `-c .rubocop.yml`.

Explicit .rubocop.yml config

### DIFF
--- a/bin/rubocop
+++ b/bin/rubocop
@@ -6,4 +6,7 @@ $LOAD_PATH.unshift(File.expand_path("../bundler/lib", __dir__))
 ENV["BUNDLE_GEMFILE"] = File.expand_path("../tool/bundler/lint_gems.rb", __dir__)
 require "bundler/setup"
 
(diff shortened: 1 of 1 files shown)


simi (approved):

simi:
@martinemde just out of my curiosity, does it save time if you have no config in parent directories (including home one)?

martinemde:
@simi yes. My ~/.rubocop.yml won't load, so I renamed it when doing this perf test. This is just comparing `-c .rubocop.yml` with no config files to find. Docs indicate that `-c` skips all the config loading behavior and just uses the specified file; probably where the performance gain comes from.

PR #6917, opened 2023-08-25
Refactor bundler lockfile checksums

martinemde:
Hey @segiddins, I was having trouble giving comments correctly on the PR, so I just made another branch with my ideas. Feel free to take what you like. Test are passing.

### DIFF
--- a/bundler/lib/bundler/checksum.rb
+++ b/bundler/lib/bundler/checksum.rb
@@ -2,6 +2,80 @@
 
 module Bundler
   class Checksum
+    DEFAULT_BLOCK_SIZE = 16_384
+    private_constant :DEFAULT_BLOCK_SIZE
+
+    class << self
+      def from_gem_source(source, digest_algorithms: %w[sha256])
+        raise ArgumentError, "not a valid gem source: #{source}" unless source.respond_to?(:with_read_io)
+
+        source.with_read_io do |io|
+          checksums = from_io(io, "#{source.path || source.inspect} gem source hexdigest", :digest_algorithms => digest_algorithms)
+          io.rewind
+          return checksums
+        end
+      end
+
+      def from_io(io, source, digest_algorithms: %w[sha256])
+        digests = digest_algorithms.to_h do |algo|
+          [algo.to_s, Bundler::SharedHelpers.digest(algo.upcase).new]
+        end
+
+        until io.eof?
+          ret = io.read DEFAULT_BLOCK_SIZE
+          digests.each_value {|digest| digest << ret }
+        end
+
+        digests.map do |algo, digest|
+          Checksum.new(algo, digest.hexdigest!, source)
+        end
+      end
+    end
+
+    attr_reader :algo, :digest, :sources
+    def initialize(algo, digest, source)
+      @algo = algo
+      @digest = digest
+      @sources = [source]
+    end
+
+    def ==(other)
+      match?(other) && other.sources == sources
+    end
+
+    alias_method :eql?, :==
+
+    def match?(other)
+      other.is_a?(self.class) && other.digest == digest && other.algo == algo
+    end
+
+    def hash
+      digest.hash
+    end
+
+    def to_s
+      "#{to_lock} (from #{sources.first}#{", ..." if sources.size > 1})"
+    end
+
+    def to_lock
+      "#{algo}-#{digest}"
+    end
+
+    def merge!(other)
+      raise ArgumentError, "cannot merge checksums of different algorithms" unless algo == other.algo
+
+      unless digest == other.digest
+        raise SecurityError, <<~MESSAGE
+          #{other}
+          #{to_lock} from:
+          * #{sources.join("\n* ")}
+        MESSAGE
+      end
+
+      @sources.concat(other.sources).uniq!
+      self
+    end
+
     class Store
       attr_reader :store
       protected :store
@@ -10,42 +84,61 @@ module Bundler
         @store = {}
       end
 
-      def initialize_copy(o)
+      def initialize_copy(other)
         @store = {}
-        o.store.each do |k, v|
-          @store[k] = v.dup
+        other.store.each do |full_name, checksums|
+          store[full_name] = checksums.dup
         end
       end
 
-      def [](spec)
-        sums = @store[spec.full_name]
-
-        Checksum.new(spec.name, spec.version, spec.platform, sums&.transform_values(&:digest))
+      def checksums(full_name)
+        store[full_name]
       end
 
-      def register(spec, checksums)
-        register_full_name(spec.full_name, checksums)
-      end
-
-      def register_triple(name, version, platform, checksums)
-        register_full_name(GemHelpers.spec_full_name(name, version, platform), checksums)
-      end
-
-      def delete_full_name(full_name)
-        @store.delete(full_name)
-      end
-
-      def register_full_name(full_name, checksums)
-        sums = (@store[full_name] ||= {})
-
-        checksums.each do |checksum|
-          algo = checksum.algo
-          if multi = sums[algo]
-            multi.merge(checksum)
-          else
-            sums[algo] = Multi.new [checksum]
-          end
+      def register_gem_package(spec, source)
+        new_checksums = Checksum.from_gem_source(source)
+        new_checksums.each do |checksum|
+          register spec.full_name, checksum
         end
+      rescue SecurityError
+        expected = checksums(spec.full_name)
+        gem_lock_name = GemHelpers.lock_name(spec.name, spec.version, spec.platform)
+        raise SecurityError, <<~MESSAGE
+          Bundler cannot continue installing #{gem_lock_name}.
+          The checksum for the downloaded `#{spec.full_name}.gem` does not match \
+          the known checksum for the gem.
+          This means the contents of the downloaded \
+          gem is different from what was uploaded to the server \
+          or first used by your teammates, and could be a potential security issue.
+
+          To resolve this issue:
+          1. delete the downloaded gem located at: `#{source.path}`
+          2. run `bundle install`
+
+          If you are sure that the new checksum is correct, you can \
+          remove the `#{gem_lock_name}` entry under the lockfile `CHECKSUMS` \
+          section and rerun `bundle install`.
+
+          If you wish to continue installing the downloaded gem, and are certain it does not pose a \
+          security issue despite the mismatching checksum, do the following:
+          1. run `bundle config set --local disable_checksum_validation true` to turn off checksum verification
+          2. run `bundle install`
+
+          #{expected.map do |checksum|
+            next unless actual = new_checksums.find {|c| c.algo == checksum.algo }
+            next if actual.digest == checksum.digest
+
+            "(More info: The expected #{checksum.algo.upcase} checksum was #{checksum.digest.inspect}, but the " \
+            "checksum for the downloaded gem was #{actual.digest.inspect}. The expected checksum came from: #{checksum.sources.join(", ")})"
+          end.compact.join("\n")}
+        MESSAGE
+      end
+
+      def register(full_name, checksum)
+        return unless checksum
+
+        sums = (store[full_name] ||= [])
+        sums.find {|c| c.algo == checksum.algo }&.merge!(checksum) || sums << checksum
       rescue SecurityError => e
         raise e.exception(<<~MESSAGE)
           Bundler found multiple different checksums for #{full_name}.
@@ -69,191 +162,15 @@ module Bundler
         MESSAGE
       end
 
-      def use(other)
-        other.store.each do |k, v|
-          register_full_name k, v.values
+      def replace(full_name, checksum)
+        store[full_name] = checksum ? [checksum] : nil
+      end
+
+      def register_store(other)
+        other.store.each do |full_name, checksums|
+          checksums.each {|checksum| register(full_name, checksum) }
         end
       end
-    end
-
-    class Single
-      attr_reader :algo, :digest, :source
-      def initialize(algo, digest, source)
-        @algo = algo
-        @digest = digest
-        @source = source
-      end
-
-      def ==(other)
-        other.is_a?(Single) && other.digest == digest && other.algo == algo && source == other.source
-      end
-
-      def hash
-        digest.hash
-      end
-
-      alias_method :eql?, :==
-
-      def to_s
-        "#{algo}-#{digest} (from #{source})"
-      end
-    end
-
-    class Multi
-      attr_reader :algo, :digest, :checksums
-      protected :checksums
-
-      def initialize(checksums)
-        @checksums = checksums
-
-        unless checksums && checksums.size > 0
-          raise ArgumentError, "must provide at least one checksum"
-        end
-
-        first = checksums.first
-        @algo = first.algo
-        @digest = first.digest
-      end
-
-      def initialize_copy(o)
-        @checksums = o.checksums.dup
-        @algo = o.algo
-        @digest = o.digest
-      end
-
-      def merge(other)
-        raise ArgumentError, "cannot merge checksums of different algorithms" unless algo == other.algo
-        unless digest == other.digest
-          raise SecurityError, <<~MESSAGE
-            #{other}
-            #{self} from:
-            * #{sources.join("\n* ")}
-          MESSAGE
-        end
-
-        case other
-        when Single
-          @checksums << other
-        when Multi
-          @checksums.concat(other.checksums)
-        else
-          raise ArgumentError
-        end
-        @checksums.uniq!
-
-        self
-      end
-
-      def sources
-        @checksums.map(&:source)
-      end
-
-      def to_s
-        "#{algo}-#{digest}"
-      end
-    end
-
-    attr_reader :name, :version, :platform, :checksums
-
-    SHA256 = %r{\Asha256-([a-z0-9]{64}|[A-Za-z0-9+\/=]{44})\z}.freeze
-    private_constant :SHA256
-
-    def initialize(name, version, platform, checksums = {})
-      @name     = name
-      @version  = version
-      @platform = platform || Gem::Platform::RUBY
-      @checksums = checksums || {}
-
-      # can expand this validation when we support more hashing algos later
-      if !@checksums.is_a?(::Hash) || (@checksums.any? && !@checksums.key?("sha256"))
-        raise ArgumentError, "invalid checksums (#{@checksums.inspect})"
-      end
-      if @checksums.any? {|_, checksum| !checksum.is_a?(String) }
-        raise ArgumentError, "invalid checksums (#{@checksums})"
-      end
-    end
-
-    def self.digests_from_file_source(file_source, digest_algorithms: %w[sha256])
-      raise ArgumentError, "not a valid file source: #{file_source}" unless file_source.respond_to?(:with_read_io)
-
-      digests = digest_algorithms.map do |digest_algorithm|
-        [digest_algorithm.to_s, Bundler::SharedHelpers.digest(digest_algorithm.upcase).new]
-      end.to_h
-
-      file_source.with_read_io do |io|
-        until io.eof?
-          block = io.read(16_384)
-          digests.each_value {|digest| digest << block }
-        end
-
-        io.rewind
-      end
-
-      digests
-    end
-
-    def full_name
-      GemHelpers.spec_full_name(@name, @version, @platform)
-    end
-
-    def match_spec?(spec)
-      name == spec.name &&
-        version == spec.version &&
-        platform.to_s == spec.platform.to_s
-    end
-
-    def to_lock
-      out = String.new
-      out << "  #{GemHelpers.lock_name(name, version, platform)}"
-      checksums.sort_by(&:first).each_with_index do |(algo, checksum), idx|
-        out << (idx.zero? ? " " : ",")
-        out << algo << "-" << checksum
-      end
-      out << "\n"
-
-      out
-    end
-
-    def match?(other)
-      return false unless match_spec?(other)
-      match_digests?(other.checksums)
-    end
-
-    def match_digests?(digests)
-      return true if checksums.empty? && digests.empty?
-
-      common_algos = checksums.keys & digests.keys
-      return true if common_algos.empty?
-
-      common_algos.all? do |algo|
-        checksums[algo] == digests[algo]
-      end
-    end
-
-    def merge!(other)
-      raise ArgumentError, "can't merge checksums for different specs" unless match_spec?(other)
-
-      merge_digests!(other.checksums)
-    end
-
-    def merge_digests!(digests)
-      if digests.any? {|_, checksum| !checksum.is_a?(String) }
-        raise ArgumentError, "invalid checksums (#{digests})"
-      end
-      @checksums = @checksums.merge(digests) do |algo, ours, theirs|
-        if ours != theirs
-          raise ArgumentError, "Digest mismatch for #{algo}:\n\t* #{ours.inspect}\n\t* #{theirs.inspect}"
-        end
-        ours
-      end
-
-      self
-    end
-
-    private
-
-    def sha256
-      @checksums.find {|c| c =~ SHA256 }
     end
   end
 end
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -738,7 +738,7 @@ module Bundler
         # has to be done separately, because we want to keep the locked checksum
         # store for a source, even when doing a full update
         if @locked_gems && locked_source = @locked_gems.sources.find {|s| s == source }
-          source.checksum_store&.use(locked_source.checksum_store)
+          source.checksum_store.register_store(locked_source.checksum_store)
         end
         # If the source is unlockable and the current command allows an unlock of
         # the source (for example, you are doing a `bundle update <foo>` of a git-pinned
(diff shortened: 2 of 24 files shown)


bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+          algo, digest = c.split("-", 2)
+          new(algo, digest, source)
+        end
+      end
+
+      def to_lock(checksums)
+        checksums.map(&:to_lock).sort.join(",")
+      end
+
+      def match_digests?(checksums, digests)
```
martinemde, line 36: This should move to Store.

bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+    attr_reader :algo, :digest, :sources
+    def initialize(algo, digest, source)
+      @algo = algo
+      @digest = digest
+      @sources = Set.new
+      @sources << source
+    end
+
+    def ==(other)
+      other.is_a?(Single) && other.digest == digest && other.algo == algo && sources == other.sources
```
martinemde, line 57: Oops
```suggestion
      other.is_a?(self.class) && other.digest == digest && other.algo == algo && sources == other.sources
```

bundler/lib/bundler/checksum.rb
```diff
@@ -17,34 +99,23 @@ def initialize_copy(o)
...
-          if multi = sums[algo]
-            multi.merge(checksum)
-          else
-            sums[algo] = Multi.new [checksum]
-          end
+      def register(full_name, checksums)
+        Array(checksums).each do |checksum|
+          @store[full_name][checksum.algo] ||= checksum
+          @store[full_name][checksum.algo].merge(checksum)
         end
```
martinemde, line 119: This is a little awkward. I think something better is possible, but the merge with self is only a little stupid and completely harmless.

bundler/lib/bundler/fetcher.rb
```diff
@@ -148,14 +148,11 @@ def specs(gem_names, source)
       specs.each do |name, version, platform, dependencies, metadata|
         spec = if dependencies
           EndpointSpecification.new(name, version, platform, self, dependencies, metadata).tap do |es|
-            unless index.local_search(es).empty?
```
martinemde, line 151: I'm not sure about removing this, but it seems to me that << just overwrites so we can too. The case not covered here is replacing one that has a checksum with one that doesn't.

bundler/lib/bundler/fetcher.rb
```diff
@@ -148,14 +148,11 @@ def specs(gem_names, source)
...
-              # here, make sure to also override the checksum, otherwise downloading the
-              # specs (even if that version is completely unused) will cause a SecurityError
-              source.checksum_store.delete_full_name(es.full_name)
-            end
-            source.checksum_store.register(es, [es.checksum]) if source && es.checksum
+            # Duplicate spec.full_names, different spec.original_names
+            # index#<< ensures that the last one added wins, so if we're overriding
+            # here, make sure to also override the checksum, otherwise downloading the
+            # specs (even if that version is completely unused) will cause a SecurityError
+            source.checksum_store[es.full_name] = es.checksum if source && es.checksum
```
martinemde, line 155: Maybe it would work if we didn't guard on es.checksum here and allowed nil checksum to delete without setting.

bundler/lib/bundler/lockfile_generator.rb
```diff
@@ -69,10 +69,12 @@ def add_dependencies
     def add_checksums
       out << "\nCHECKSUMS\n"
 
-      empty_store = Checksum::Store.new
-
       definition.resolve.sort_by(&:full_name).each do |spec|
-        out << (spec.source.checksum_store || empty_store)[spec].to_lock
+        next unless spec.source.checksum_store
+        next if (checksums = spec.source.checksum_store[spec.full_name]).empty?
```
martinemde, line 74: This is different than the existing specs. Probably causing some of the failures. If we still want gem names with empty checksums then we can move this guard down to only skipping the checksum part.

bundler/lib/bundler/rubygems_gem_installer.rb
```diff
@@ -122,9 +122,15 @@ def validate_bundler_checksum(checksum_store)
       return true unless source.respond_to?(:with_read_io)
       digests = Bundler::Checksum.digests_from_file_source(source).transform_values(&:hexdigest!)
 
-      checksum = checksum_store[spec]
-      unless checksum.match_digests?(digests)
-        expected = checksum_store.send(:store)[spec.full_name]
+      Checksum.match_digests?(checksum_store[spec.full_name], digests)
+
+      if checksum_store[spec.full_name].transform_values(&:digest)
```
martinemde, line 127: Oops

```suggestion
      if Checksum.match_digests?(checksum_store[spec.full_name], digests)
```

bundler/lib/bundler/rubygems_gem_installer.rb
```diff
@@ -147,24 +153,16 @@ def validate_bundler_checksum(checksum_store)
           1. run `bundle config set --local disable_checksum_validation true` to turn off checksum verification
           2. run `bundle install`
 
-          #{expected.map do |algo, multi|
+          #{expected.map do |algo, checksum|
             next unless actual = digests[algo]
-            next if actual == multi
+            next if actual == checksum.digest
```
martinemde, line 158: I think this was a bug in the original.

bundler/spec/support/checksums.rb
```diff
@@ -14,12 +14,16 @@ def repo_gem(gem_repo, gem_name, gem_version, platform = nil, empty: false)
           "#{gem_repo}/gems/#{gem_name}-#{gem_version}.gem"
         end
 
-        checksum = { "sha256" => sha256_checksum(gem_file) } unless empty
-        @checksums << Bundler::Checksum.new(gem_name, gem_version, platform, checksum)
+        checksum = Bundler::Checksum.new("sha256", sha256_checksum(gem_file), 'Checksum builder') unless empty
```
martinemde, line 17:
```suggestion
        checksum = Bundler::Checksum.new("sha256", sha256_checksum(gem_file), 'ChecksumsBuilder') unless empty
```

martinemde (commented):
Mistakes and fixes I need to make.

bundler/lib/bundler/checksum.rb
```diff
@@ -17,34 +99,23 @@ def initialize_copy(o)
...
-          algo = checksum.algo
-          if multi = sums[algo]
-            multi.merge(checksum)
-          else
-            sums[algo] = Multi.new [checksum]
-          end
+      def register(full_name, checksums)
+        Array(checksums).each do |checksum|
+          @store[full_name][checksum.algo] ||= checksum
+          @store[full_name][checksum.algo].merge(checksum)
```
martinemde, line 118: LOL. Atrocious but no extra merge. (Sorry)

```suggestion
          @store[full_name][checksum.algo]&.merge(checksum)
          @store[full_name][checksum.algo] ||= checksum

```

bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+
+    def to_s
+      "#{algo}-#{digest} (from #{sources.first}#{", ..." if sources.size > 1})"
+    end
+
+    def to_lock
+      "#{algo}-#{digest}"
+    end
+
+    def merge(other)
```
martinemde, line 74:
```suggestion
    def merge!(other)
```

bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+      raise ArgumentError, "cannot merge checksums of different algorithms" unless algo == other.algo
+      unless digest == other.digest
+        raise SecurityError, <<~MESSAGE
+          #{other}
+          #{self} from:
+          * #{sources.join("\n* ")}
+        MESSAGE
+      end
+      @sources.merge(other.sources)
```
martinemde, line 83:
```suggestion
      @sources.merge!(other.sources)
```

bundler/lib/bundler/definition.rb
```diff
@@ -738,7 +738,7 @@ def converge_sources
         # has to be done separately, because we want to keep the locked checksum
         # store for a source, even when doing a full update
         if @locked_gems && locked_source = @locked_gems.sources.find {|s| s == source }
-          source.checksum_store&.use(locked_source.checksum_store)
+          source.checksum_store&.merge(locked_source.checksum_store)
```
martinemde, line 741: This should be merge!, but also this seems weird because if source.checksum_store is nil then we should make it not nil by setting the locked_source checksums into it.

```suggestion
          source.checksum_store&.merge!(locked_source.checksum_store)
```

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -235,19 +235,14 @@ def parse_checksum(line)
...
-        algo, digest = c.split("-", 2)
-        Checksum::Single.new(algo, digest, source)
-      end
-
+      checksums = Checksum.from_lock(checksums, "#{@lockfile_path}:#{@pos} in the CHECKSUMS lockfile section")
       full_name = GemHelpers.spec_full_name(name, version, platform)
 
       # Don't raise exception if there's a checksum for a gem that's not in the lockfile,
       # we prefer to heal invalid lockfiles
       return unless spec = @specs[full_name]
```
martinemde, line 243: This should change order so we don't parse checksums for a not-found spec. Just save the time and return early.

bundler/lib/bundler/checksum.rb
```diff
@@ -69,191 +140,11 @@ def register_full_name(full_name, checksums)
...
-        @digest = first.digest
-      end
-
-      def initialize_copy(o)
-        @checksums = o.checksums.dup
-        @algo = o.algo
-        @digest = o.digest
-      end
-
       def merge(other)
```
martinemde, line 143: Should be `merge!` or go back to `use`

bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+    def merge!(other)
+      raise ArgumentError, "cannot merge checksums of different algorithms" unless algo == other.algo
+      unless digest == other.digest
+        raise SecurityError, <<~MESSAGE
+          #{other}
+          #{self} from:
+          * #{sources.join("\n* ")}
+        MESSAGE
+      end
+      @sources.merge!(other.sources)
```
martinemde, line 83:
```suggestion
      @sources.merge(other.sources)
```
This is what I get for working from my phone.

bundler/lib/bundler/checksum.rb
```diff
@@ -2,12 +2,94 @@
...
+    def to_lock
+      "#{algo}-#{digest}"
+    end
+
+    def merge!(other)
+      raise ArgumentError, "cannot merge checksums of different algorithms" unless algo == other.algo
+      unless digest == other.digest
+        raise SecurityError, <<~MESSAGE
+          #{other}
+          #{self} from:
```
martinemde, line 79:
```suggestion
          #{to_lock} from:
```

bundler/lib/bundler/source/git.rb
```diff
@@ -11,6 +11,7 @@ class Git < Path
 
       def initialize(options)
         @options = options
+        @checksum_store = Checksum::Store.new
```
martinemde, line 14: I don't know if this is right, but the cases where store was nil on source.checksum_store.merge! were when the source was Git.

martinemde:
This is working, but ruby-core is failing to install so those tests are failing.

bundler/lib/bundler/source/git.rb
martinemde: @segiddins Do you understand this part enough to know if this is fine? There a place in the code where it tries to merge lock checksums into this git source and there's no checksum store. Original code used checksum_store&.use to avoid the problem, but it seems weird to lose checksums that we have. Thoughts?

segiddins: yeah this is fine. I dont think we'd lose checksums in practice because we would have checksums ever for git source gems, but probably better to be consistent here 👍🏻

segiddins:
So in general, I think the argument for including gems in the CHECKSUMS section that we dont have a checksum for is that it makes it clear that there is an un-checksummed gem in the bundle. Otherwise, if you have a massive gemfile.lock, it would be really easy to miss adding an uncheck summed gem

martinemde:
Should we call that out in the UI instead? I don't think it's common practice to examine your Gemfile.lock.

segiddins:
It easily reviewed in a PR, whereas `bundle install` output is immediately lost

bundler/lib/bundler/fetcher.rb
```diff
@@ -148,14 +148,11 @@ def specs(gem_names, source)
...
-              # here, make sure to also override the checksum, otherwise downloading the
-              # specs (even if that version is completely unused) will cause a SecurityError
-              source.checksum_store.delete_full_name(es.full_name)
-            end
-            source.checksum_store.register(es, [es.checksum]) if source && es.checksum
+            # Duplicate spec.full_names, different spec.original_names
+            # index#<< ensures that the last one added wins, so if we're overriding
+            # here, make sure to also override the checksum, otherwise downloading the
+            # specs (even if that version is completely unused) will cause a SecurityError
+            source.checksum_store.replace(es.full_name, es.checksum) if source
```
segiddins, line 155: now that I think about this... we should have a spec for when that gem has a checksum in the lockfile, and make sure we're not silently overriding that -- just the one from the API

bundler/lib/bundler/lockfile_generator.rb
```diff
@@ -69,10 +69,11 @@ def add_dependencies
     def add_checksums
       out << "\nCHECKSUMS\n"
 
-      empty_store = Checksum::Store.new
-
       definition.resolve.sort_by(&:full_name).each do |spec|
-        out << (spec.source.checksum_store || empty_store)[spec].to_lock
+        checksums = spec.source.checksum_store&.checksums(spec.full_name)
```
segiddins, line 73: is checksum_store always non-nil now?

martinemde:
@segiddins I split out the code for not including gems without checksums. I have the changes in another branch that I can push if we decide we want it.

bundler/lib/bundler/lockfile_generator.rb
martinemde: I removed the `&` and running the test suite locally now to find out.

bundler/lib/bundler/checksum.rb
```diff
@@ -2,6 +2,71 @@
...
+          until io.eof?
+            block = io.read(16_384)
+            digests.each_value {|digest| digest << block }
+          end
+
+          io.rewind
+        end
+
+        digests.map do |algo, digest|
+          Checksum.new(algo, digest.hexdigest!, file_source.path)
```
martinemde, line 23:
```suggestion
          Checksum.new(algo, digest.hexdigest!, "#{file_source.path} calculated #{algo.upcase} hexdigest")
```

martinemde (commented):
My last few edits have been focusing on getting the messaging right when there is a mismatch. Improving the source descriptions could help find and identify exactly which checksum came from where. For example we must include the platform and not just name (version) since there may be multiple.

bundler/lib/bundler/checksum.rb
```diff
@@ -10,42 +75,71 @@ def initialize
...
-      def delete_full_name(full_name)
-        @store.delete(full_name)
+      def replace(full_name, checksum)
+        delete(full_name)
+        register(full_name, checksum)
       end
 
-      def register_full_name(full_name, checksums)
-        sums = (@store[full_name] ||= {})
+      # accepts one or an Array of Checksum objects
```
martinemde, line 137:
```suggestion
```

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -235,19 +235,15 @@ def parse_checksum(line)
...
       full_name = GemHelpers.spec_full_name(name, version, platform)
-
       # Don't raise exception if there's a checksum for a gem that's not in the lockfile,
       # we prefer to heal invalid lockfiles
       return unless spec = @specs[full_name]
 
-      spec.source.checksum_store.register_full_name(full_name, checksums)
+      checksums.split(",").each do |c|
+        algo, digest = c.split("-", 2)
+        spec.source.checksum_store.register(full_name, Checksum.new(algo, digest, "#{@lockfile_path}:#{@pos} CHECKSUMS #{GemHelpers.lock_name(spec.name, spec.version, spec.platform}"))
```
martinemde, line 245:
```suggestion
        spec.source.checksum_store.register(full_name, Checksum.new(algo, digest, "#{@lockfile_path}:#{@pos} CHECKSUMS #{GemHelpers.lock_name(spec.name, spec.version, spec.platform)}"))
```

bundler/spec/install/gems/compact_index_spec.rb
```diff
@@ -893,15 +893,15 @@ def start
...
+          Bundler cannot continue installing rack-1.0.0.
           The checksum for the downloaded `rack-1.0.0.gem` does not match the known checksum for the gem.
           This means the contents of the downloaded gem is different from what was uploaded to the server or first used by your teammates, and could be a potential security issue.
 
           To resolve this issue:
           1. delete the downloaded gem located at: `#{gem_path}`
           2. run `bundle install`
 
-          If you are sure that the new checksum is correct, you can remove the `rack (1.0.0)` entry under the lockfile `CHECKSUMS` section and rerun `bundle install`.
+          If you are sure that the new checksum is correct, you can remove the `rack-1.0.0` entry under the lockfile `CHECKSUMS` section and rerun `bundle install`.
```
segiddins, line 904: this shouldn't have changed, because `rack (1.0.0)` is the format used in the lock file

martinemde: ok, good point. Let's me see if I can seed the error message correctly.

bundler/lib/bundler/checksum.rb
```diff
@@ -2,6 +2,71 @@
...
+      def from_file_source(file_source, digest_algorithms: %w[sha256])
+        raise ArgumentError, "not a valid file source: #{file_source}" unless file_source.respond_to?(:with_read_io)
+
+        digests = digest_algorithms.map do |digest_algorithm|
+          [digest_algorithm.to_s, Bundler::SharedHelpers.digest(digest_algorithm.upcase).new]
+        end.to_h
+
+        file_source.with_read_io do |io|
+          until io.eof?
+            block = io.read(16_384)
```
simi, line 15: Can we make 16_384 constant and add comment?

bundler/lib/bundler/checksum.rb
```diff
@@ -10,42 +75,70 @@ def initialize
...
+          the known checksum for the gem.
+          This means the contents of the downloaded \
+          gem is different from what was uploaded to the server \
+          or first used by your teammates, and could be a potential security issue.
+
+          To resolve this issue:
+          1. delete the downloaded gem located at: `#{file_source.path}`
+          2. run `bundle install`
+
+          If you are sure that the new checksum is correct, you can \
```
simi, line 113: I think we should not suggest to edit lockfile manually and also to disable checksum validation for whole project (Gemfile.lock).

bundler/lib/bundler/checksum.rb
```diff
@@ -2,6 +2,71 @@
...
+    def hash
+      digest.hash
+    end
+
+    def to_s
+      "#{to_lock} (from #{sources.first}#{", ..." if sources.size > 1})"
+    end
+
+    def to_lock
+      "#{algo}-#{digest}"
```
simi, line 54: What about to use format used by [Digest Fields](https://datatracker.ietf.org/doc/draft-ietf-httpbis-digest-headers/, https://www.rfc-editor.org/rfc/rfc8941.html#name-serializing-a-byte-sequence)? I know this is not directly related to HTTP protocol, but it could make it easier later to re-use.

Or at least unify algo naming (use `sha-256` insetad of `sha256`) as use at https://github.com/rubygems/rubygems/compare/master...opaque-etag-http-digest#diff-603a4693e8a95163d56f6e250d52ec0c280378829f2f81c967385c38ca8a6a64R9.

bundler/lib/bundler/checksum.rb
martinemde: This is just cargo-culted from somewhere else. Other than being the "ultimate goal" in the game 2048, I suspect this is intended to use a common default block size, 16KiB. I'll call it `DEFAULT_BLOCK_SIZE`.

bundler/lib/bundler/checksum.rb
martinemde: I like the idea of using something standardized. The first few links suggest `":#{base64}:"`. We currently use most of our checksums as hexdigests, but there's code to accept either from the API.

Also, looking at your last link I see there's this [existing error class](https://github.com/rubygems/rubygems/compare/master...opaque-etag-http-digest#diff-603a4693e8a95163d56f6e250d52ec0c280378829f2f81c967385c38ca8a6a64R11-R22) for checksum mismatches.

martinemde: Reading into it, I'm thinking about: `sha-256=FEEDDADC0FFEE111DECAF15BAD`

Hex is slightly easier, though it takes a few extra bytes. The command line output of shasum or openssl can output hex by default instead of having to pass it through as binary to `openssl base64`. I don't feel strongly here though, because base64 is more compact and it's not exactly uncommon. I'd be happy to convert or even support both out of the box (easy because string size differs)

For the algorithm, you're definitely right that "sha-256" is the more common formatting. We use `Digest()` in that code you linked, so maybe it can find the class even with the dash there. I'll try.

simi: > We use Digest() in that code you linked, so maybe it can find the class even with the dash there. I'll try.

sadly it can't :(

simi: Feel free to extract that one into some generic namespace and use, I'll update my branch to use the same.

martinemde: Well, [all](https://ruby-doc.org/stdlib-3.0.1/libdoc/digest/rdoc/Digest.html) of [ruby](https://ruby-doc.org/stdlib-3.0.1/libdoc/digest/rdoc/Digest.html) seems to use `SHA256` without the dash. Since we're using ruby, I'll just change to an `=` and call it good. Some future algorithm may want to use the dash (e.g. "SHA3-256")

martinemde: I think they mean different things. One is used to retry partial downloads of compact index when Etag doesn't match. The other is a security error. I'll keep them separate.

bundler/lib/bundler/checksum.rb
martinemde: I brought this up in the [RFC](https://github.com/rubygems/rfcs/pull/50/files#diff-e042d284f6944e04fe33f6cd94cf0305182593c6e368c19ee039eaa094c64860R82) as an unknown. Since it was not originally added as part of this patch PR, we can solve it on the main PR.

bundler/lib/bundler/checksum.rb
```diff
@@ -2,258 +2,174 @@
...
+        return unless checksum
 
-    def match?(other)
-      return false unless match_spec?(other)
-      match_digests?(other.checksums)
-    end
+        sums = (store[full_name] ||= [])
+        sums.find {|c| c.algo == checksum.algo }&.merge!(checksum) || sums << checksum
+      rescue SecurityError => e
+        raise e.exception(<<~MESSAGE)
```
segiddins, line 142: can we extract out this message construction to a method, since we do it 2x in this file?

bundler/spec/support/checksums.rb
```diff
@@ -3,49 +3,81 @@
...
 
-      def repo_gem(gem_repo, gem_name, gem_version, platform = nil, empty: false)
-        gem_file = if platform
-          "#{gem_repo}/gems/#{gem_name}-#{gem_version}-#{platform}.gem"
-        else
-          "#{gem_repo}/gems/#{gem_name}-#{gem_version}.gem"
+      def repo_gem(repo, name, version, platform = Gem::Platform::RUBY)
+        gem_file = File.join(repo, "gems", "#{Bundler::GemHelpers.spec_full_name(name, version, platform)}.gem")
+
+        File.open(gem_file) do |f|
```
segiddins, line 14: nit:

```suggestion
        File.open(gem_file, "rb") do |f|
```

segiddins (approved):

bundler/lib/bundler/checksum.rb
martinemde: I was working towards this. They are quite different. In the RFC I think I covered the cases where it will happen and wrote a simpler version. We could leave this as a task to do because it's linked to the entire way we verify checksums and how we message mismatches. (in other words, it's not staying like this, but the diff is so big it's hard to read and I'd like to make focused PRs rather than add commits on this one)

bundler/lib/bundler/checksum.rb
```diff
@@ -2,258 +2,174 @@
...
+          digests.each_value {|digest| digest << ret }
+        end
 
-          If you wish to continue installing the downloaded gem, and are certain it does not pose a \
-          security issue despite the mismatching checksum, do the following:
-          1. run `bundle config set --local disable_checksum_validation true` to turn off checksum verification
-          2. run `bundle install`
-        MESSAGE
-      end
+        io.rewind
```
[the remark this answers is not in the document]
martinemde: ah, good point.

Issue #6273, opened 2023-01-13
`bundle install --standalone` broken with stdlibs for ruby 3.2


AlfonsoUceda:
### Describe the problem as clearly as you can

When using ruby 3.2 and bundler with `--standalone` option I receive a problem related to a stdlib, e.g. json or io-console.

Running `bundle install --standalone --path=./gems` install gems but doesn't finish because I get a warning like this:
```
json-2.6.3 is built in to Ruby, and can't be cached because your Gemfile doesn't have any sources that contain it.
```
and at the end I see this error:
```
Bundler::GemNotFound: Could not find json-2.6.3.gem for installation
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/source/rubygems.rb:159:in `install'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/installer/gem_installer.rb:54:in `install'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/installer/gem_installer.rb:16:in `install_from_spec'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/installer/parallel_installer.rb:155:in `do_install'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/installer/parallel_installer.rb:146:in `block in worker_pool'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/worker.rb:62:in `apply_func'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/worker.rb:57:in `block in process_queue'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/worker.rb:54:in `loop'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/worker.rb:54:in `process_queue'
  /Users/alfonsouceda/.rubies/ruby-3.2.0/lib/ruby/site_ruby/3.2.0/bundler/worker.rb:90:in `block (2 levels) in create_threads'

[...]

PR #6925, opened 2023-08-29
Prevent gem activation in standalone mode

composerinteralia:

As discussed in https://github.com/rubygems/rubygems/issues/6273#issuecomment-1449176658

The `gem` method behaves awkwardly in standalone mode. Assuming bundler isn't loaded at all, a call to gem might activate a gem that is not part of the bundle (because it's the gem method defined in lib/rubygems/core_ext/kernel_gem.rb and not
lib/bundler/rubygems_integration.rb). And when running with `--disable-gems`, the gem method won't be defined at all so we'll get a NoMethodError.

Calls to `gem` can appear in dependencies outside an application's control. To work around this at GitHub we defined our own `Kernel#gem` that no-ops.

I agree with https://github.com/rubygems/rubygems/issues/6273#issuecomment-1440755882

> people using standalone mode don't want to activate gems like Kernel.gem

This commit redefines `Kernel#gem` in the standalone script to no-op.

### DIFF
--- a/bundler/lib/bundler/installer/standalone.rb
+++ b/bundler/lib/bundler/installer/standalone.rb
@@ -12,6 +12,7 @@ module Bundler
       end
       File.open File.join(bundler_path, "setup.rb"), "w" do |file|
         file.puts "require 'rbconfig'"
+        file.puts prevent_gem_activation
         file.puts define_path_helpers
         file.puts reverse_rubygems_kernel_mixin
         paths.each do |path|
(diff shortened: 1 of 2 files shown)


simi (approved):

hsbt:
>And when running with --disable-gems, the gem method won't be defined at all so we'll get a NoMethodError.

Question: Why do you use `--disable-gems` in your application? It's only for Ruby core developper, not users.

composerinteralia:
> Question: Why do you use --disable-gems in your application? It's only for Ruby core developper, not users.

We mostly don't. It looks like GitHub started running some things with `--disable-gems` back in 2014 or so. I think the way GitHub did git operations back then involved spawning a new Ruby process each time, and so `--disable-gems` was used to make spawning those processes a tiny bit fast. (I'm not certain about that—I didn't work here at the time and I don't think that's how things work nowadays anyway.) There's a few old scripts that still use `--disable-gems`, but they might not need to.

I think this PR makes sense anyway, since even without `--disable-gems` the `gem` method in standalone mode may activate a gem that's not part of the bundle.

deivid-rodriguez (approved):

deivid-rodriguez:
This looks great, but I just noticed there's a merge conflict now. Could you rebase this @composerinteralia :pray:?

deivid-rodriguez:
Thank you @composerinteralia!

PR #6926, opened 2023-08-30
Update SPDX list and warn on deprecated identifiers.

simi:
- updates SPDX list
- split licence identifiers list into valid and deprecated ones
- provide detailed message of SPDX identifier validity (invalid or deprecated identifier or exception part)
- update SPDX link to https

closes https://github.com/rubygems/rubygems/issues/2153, https://github.com/rubygems/rubygems/pull/2259

/cc @voxik

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -371,10 +371,21 @@ duplicate dependency on #{dep}, (#{prev.requirement}) use:
 
     licenses.each do |license|
       next if Gem::Licenses.match?(license) || license.nil?
+      license_id_deprecated = Gem::Licenses.deprecated_license_id?(license)
+      exception_id_deprecated = Gem::Licenses.deprecated_exception_id?(license)
       suggestions = Gem::Licenses.suggestions(license)
+
+      if license_id_deprecated
+        main_message = "License identifier '#{license}' is deprecated"
+      elsif exception_id_deprecated
+        main_message = "Exception identifier at '#{license}' is deprecated"
+      else
+        main_message = "License identifier '#{license}' is invalid"
+      end
+
       message = <<-WARNING
-license value '#{license}' is invalid.  Use a license identifier from
-http://spdx.org/licenses or '#{Gem::Licenses::NONSTANDARD}' for a nonstandard license,
+#{main_message}. Use an identifier from
+https://spdx.org/licenses or '#{Gem::Licenses::NONSTANDARD}' for a nonstandard license,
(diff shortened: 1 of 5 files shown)


segiddins (approved):

hsbt (approved):
👍

tool/generate_spdx_license_list.rb
```diff
@@ -7,21 +7,29 @@
 def generate_spdx_license_list(dest = "lib/rubygems/util/licenses.rb")
```
martinemde, line 7: This method is gnarly. How often do we need to run this? I want to avoid paying off this technical debt if this is almost never run, but I did want to file a formal complaint about the RUBY heredocs in the middle.

martinemde (approved):
Thanks for digging into this. 💪

tool/generate_spdx_license_list.rb
segiddins: Hey, this is a massive improvement over when there was no script to be ran every two years at all

simi: What do you mean @martinemde? Migrate this into erb template to render instead of inline heredoc?

martinemde: I don't think it's a big deal right now, but if this was a piece of code we had to regularly run and maintain, I would spend time trying to make it easier to read.

If I had to refactor it, just off the top of my head I'd avoid any interpolation at all and just write the data to files, then read them in for use in the script. Still, I think there's probably more important work than this, I was just noting how hard that code was to read the first time through.

lib/rubygems/specification_policy.rb
```diff
@@ -371,19 +371,30 @@ def validate_licenses
...
+      elsif exception_id_deprecated
+        main_message = "Exception identifier at '#{license}' is deprecated."
+      else
+        main_message = "License identifier '#{license}' is invalid."
+      end
+
       message = <<-WARNING
-license value '#{license}' is invalid.  Use a license identifier from
-http://spdx.org/licenses or '#{Gem::Licenses::NONSTANDARD}' for a nonstandard license,
+#{main_message} Use an identifier from
```
deivid-rodriguez, line 387: So small, but I think the composed message would read slightly better if we keep the final `.` here, instead of as part of `main_message`.

deivid-rodriguez (approved):
Looks great, thanks for also moving the link to https!

lib/rubygems/specification_policy.rb
simi: I'll fix this and merge.

PR #6928, opened 2023-08-30
Handle BUNDLER_VERSION being set to an empty string

ccutrer:
This is useful, in case you're using Docker, and an upstream Dockerfile sets BUNDLER_VERSION to something you don't want. It's impossible to unset it... only override to be the empty string.

I have a Dockerfile that has an upstream Dockerfile that sets BUNDLER_VERSION, but I don't want that set. I instead want bundler to automatically use the correct version from my Gemfile.lock. Unfortunately, once an env var is set in Docker, it can't be unset. Only overridden (and set to the empty string). When I do that, I get this:

```
% BUNDLER_VERSION= bundle --version                                
/Users/cody/.rubies/ruby-3.1.3/lib/ruby/site_ruby/3.1.0/rubygems/requirement.rb:107:in `parse': Illformed requirement [""] (Gem::Requirement::BadRequirementError)
	from /Users/cody/.rubies/ruby-3.1.3/lib/ruby/site_ruby/3.1.0/rubygems/requirement.rb:139:in `block in initialize'
	from /Users/cody/.rubies/ruby-3.1.3/lib/ruby/site_ruby/3.1.0/rubygems/requirement.rb:139:in `map!'
	from /Users/cody/.rubies/ruby-3.1.3/lib/ruby/site_ruby/3.1.0/rubygems/requirement.rb:139:in `initialize'
	from /Users/cody/src/canvas-lms/bin/bundle:114:in `new'
	from /Users/cody/src/canvas-lms/bin/bundle:114:in `activate_bundler'
	from /Users/cody/src/canvas-lms/bin/bundle:102:in `load_bundler!'
	from /Users/cody/src/canvas-lms/bin/bundle:128:in `<main>'
```

Treat an empty BUNDLER_VERSION env var the same as not present at all.


bundler/spec/commands/binstubs_spec.rb
```diff
@@ -150,6 +150,12 @@
           sys_exec "bin/bundle install", :env => { "BUNDLER_VERSION" => "999.999.998", "DEBUG" => "1" }, :raise_on_error => false
           expect(out).to include %(Using bundler 999.999.998\n)
         end
+
+        it "runs correctly even if empty" do
+          skip "does not work on old rubies" if RUBY_VERSION < "2.7.0"
```
segiddins, line 155: we are now 3.0+, this line should be deletable

segiddins (approved):

bundler/spec/commands/binstubs_spec.rb
```diff
@@ -150,6 +150,12 @@
           sys_exec "bin/bundle install", env: { "BUNDLER_VERSION" => "999.999.998", "DEBUG" => "1" }, raise_on_error: false
           expect(out).to include %(Using bundler 999.999.998\n)
         end
+
+        it "runs correctly even if empty" do
+          skip "does not work on old rubies" if RUBY_VERSION < "2.7.0"
+
+          sys_exec "bin/bundle install", :env => { "BUNDLER_VERSION" => "" }
```
hsbt, line 157:
```suggestion
          sys_exec "bin/bundle install", env: { "BUNDLER_VERSION" => "" }
```

simi:
Instead of trying to update all usage of `ENV["BUNDLER_VERSION"]`, wouldn't make sense to do the opposite and update the `ENV["BUNDLER_VERSION"]` to `nil` if set to empty string somewhere in initialisation code? That way it will not be possible to forget handling `""` as `nil` in the future.

jeromedalbert:
I cannot repro the issue:

```
$ BUNDLER_VERSION= bundle --version
Bundler version 2.5.16
```

Can you confirm @ccutrer?

jeromedalbert:
Thanks, that was it! Here the steps I used to repro:

```
$ bundle gem mygem; cd mygem
$ bundle binstubs bundler
$ BUNDLER_VERSION= bin/bundle --version
/Users/jerome/.asdf/installs/ruby/3.3.2/lib/ruby/site_ruby/3.3.0/rubygems/requirement.rb:107:in `parse': Illformed requirement [""] (Gem::Requirement::BadRequirementError)

      raise BadRequirementError, "Illformed requirement [#{obj.inspect}]"
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        from /Users/jerome/.asdf/installs/ruby/3.3.2/lib/ruby/site_ruby/3.3.0/rubygems/requirement.rb:139:in `block in initialize'
        from /Users/jerome/.asdf/installs/ruby/3.3.2/lib/ruby/site_ruby/3.3.0/rubygems/requirement.rb:139:in `map!'
        from /Users/jerome/.asdf/installs/ruby/3.3.2/lib/ruby/site_ruby/3.3.0/rubygems/requirement.rb:139:in `initialize'
        from bin/bundle:92:in `new'
        from bin/bundle:92:in `activate_bundler'
        from bin/bundle:81:in `load_bundler!'
        from bin/bundle:105:in `<main>'

```

PR #6931, opened 2023-09-01
Improve Bundle::Index efficiency by removing unnecessary creation and dups

martinemde:

`Bundler::Source::Rubygems#fetch_names` was always called with override_dupes set to false.
`Bundler::Index` had two versions of `#use` that did were semantically different enough to split into two methods.

Refactor these methods to make them easier to understand.

Change the name of Index @all_specs to be called `@duplicates` because it does not contain all specs, but only specs that have been found as duplicates.

### DIFF
--- a/bundler/lib/bundler/index.rb
+++ b/bundler/lib/bundler/index.rb
@@ -10,8 +10,8 @@ module Bundler
       i
     end
 
-    attr_reader :specs, :all_specs, :sources
-    protected :specs, :all_specs
+    attr_reader :specs, :duplicates, :sources
+    protected :specs, :duplicates
 
     RUBY = "ruby"
     NULL = "\0"
@@ -20,20 +20,20 @@ module Bundler
       @sources = []
       @cache = {}
       @specs = Hash.new {|h, k| h[k] = {} }
-      @all_specs = Hash.new {|h, k| h[k] = EMPTY_SEARCH }
+      @duplicates = {}
     end
 
     def initialize_copy(o)
       @sources = o.sources.dup
       @cache = {}
       @specs = Hash.new {|h, k| h[k] = {} }
-      @all_specs = Hash.new {|h, k| h[k] = EMPTY_SEARCH }
+      @duplicates = {}
 
       o.specs.each do |name, hash|
         @specs[name] = hash.dup
       end
-      o.all_specs.each do |name, array|
-        @all_specs[name] = array.dup
+      o.duplicates.each do |name, array|
+        @duplicates[name] = array.dup
       end
     end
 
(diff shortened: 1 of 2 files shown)


bundler/lib/bundler/index.rb
```diff
@@ -47,7 +47,8 @@ def empty?
     end
 
     def search_all(name)
-      all_matches = local_search(name) + @all_specs[name]
+      all_matches = specs_by_name(name) # always returns a new Array
+      all_matches.concat(@duplicates[name]) if @duplicates[name]
```
segiddins, line 51: nit: pull it out into a single hash lookup

bundler/lib/bundler/source/rubygems.rb
```diff
@@ -380,7 +380,7 @@ def installed_specs
 
       def cached_specs
         @cached_specs ||= begin
-          idx = @allow_local ? installed_specs.dup : Index.new
+          idx = Index.new
```
segiddins, line 383: is this a change in behavior here?

segiddins (approved):

bundler/lib/bundler/source/rubygems.rb
martinemde: When `@allow_local` is true, [`installed_specs` is always merged on top of `cached_specs`](https://github.com/rubygems/rubygems/pull/6931/files#diff-cae021dce97344756b9347e262ca14ff32d50fdc3b48930398f37a68515afe2eR135), replacing every installed_spec from this duped version with the exact same spec from installed_specs. Before this change it was like the following pseudo code:
```
# @allow_cached = true && allow_local = true
  specs = installed.merge(cached).merge(installed)
# @allow_remote = true && @allow_local = false
  specs = cached
```
The symptom I was tracing was "why do we run into so many duplicate specs? Why is there a list of specs in (formerly) `@all_specs` that contains multiple copies of the same spec and source?" It's because this was causing a guaranteed dup of every installed spec every time allow_local was enabled.

segiddins: are there any places that are using `installed_specs` independently? because I think it's possible for a spec to be cached but not installed, and iirc those are public methods on `Source::Rubygems`? It's possible (maybe event likely) that the answer is no, I'm just worried about something not being caught by the specs that someone might be relying on

martinemde: `cached_specs` is only called once by `specs` in the way I described. It should be private imho.

martinemde: To put it another way, you could write `specs` like so without this fix:

```ruby
      def specs
        @specs ||= begin
          index = @allow_remote ? remote_specs.dup : Index.new
          index.merge!(installed_specs) if @allow_local
          index.merge!(cached_specs) if @allow_cached || @allow_remote
          index.merge!(installed_specs) if @allow_local
          index
        end
```

Since `cached_specs` is never used outside of this context, the only thing that matters is it's result for this `specs`.

segiddins: Sweet, thanks for diving in & verifying!

martinemde: Oh, I was wrong, `cached_specs` is already `protected`, as is `installed_specs`. They aren't public.

PR #6933, opened 2023-09-01
Unify LockfileParser loading of SPECS section

martinemde:

Tiny refactor that was begging to be made.

This was almost following a very common refactoring pattern, so I just took the last step.

### DIFF
--- a/bundler/lib/bundler/lockfile_parser.rb
+++ b/bundler/lib/bundler/lockfile_parser.rb
@@ -110,21 +110,9 @@ module Bundler
     def parse_source(line)
       case line
       when SPECS
-        case @type
-        when PATH
-          @current_source = TYPES[@type].from_lock(@opts)
-          @sources << @current_source
-        when GIT
-          @current_source = TYPES[@type].from_lock(@opts)
-          @sources << @current_source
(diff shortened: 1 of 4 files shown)


bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -110,21 +110,8 @@ def may_include_redundant_platform_specific_gems?
...
-        when GEM
-          @opts["remotes"] = Array(@opts.delete("remote")).reverse
-          @current_source = TYPES[@type].from_lock(@opts)
-          @sources << @current_source
-        when PLUGIN
-          @current_source = Plugin.source_from_lock(@opts)
-          @sources << @current_source
-        end
+        @current_source = TYPES[@type]&.from_lock(@opts)
+        @sources << @current_source if @current_source
```
martinemde, line 114: Since there's no else on the former  case statement, this should be the following so the instance vars are not overwritten on an unrecognized type.

```suggestion
        return unless TYPES.key?(@type)
        @current_source = TYPES[@type].from_lock(@opts)
        @sources << @current_source
```

segiddins (approved):

PR #6936, opened 2023-09-02
Fix force_ruby_platform: when the lockfile only locks the ruby platform

segiddins:

I was unable to force ruby platform for google-protobuf in the rubygems.org bundle

Respect `force_ruby_platform:` when materializing by transferring it over to the lazy spec from the gemfile dep

### DIFF
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -202,6 +202,7 @@ module Bundler
       specs_for_name = lookup[dep.name]
       target_platform = dep.force_ruby_platform ? Gem::Platform::RUBY : (platform || Bundler.local_platform)
       matching_specs = GemHelpers.select_best_platform_match(specs_for_name, target_platform)
+      matching_specs.each {|s| s.force_ruby_platform = true } if dep.force_ruby_platform
       matching_specs.map!(&:materialize_for_installation).compact! if platform.nil?
       matching_specs
     end
--- a/bundler/spec/install/gemfile/force_ruby_platform_spec.rb
+++ b/bundler/spec/install/gemfile/force_ruby_platform_spec.rb
@@ -114,5 +114,48 @@ RSpec.describe "bundle install with force_ruby_platform DSL option", :jruby do
       expect(the_bundle).to include_gems "depends_on_platform_specific 1.0.0 RUBY"
       expect(the_bundle).to include_gems "platform_specific 1.0.0 #{Bundler.local_platform}"
     end
+
+    it "uses ruby variants for the explicit transitive dependency with a lockile that has only ruby platform" do
+      lockfile <<-L
+        GEM
+          remote: #{file_uri_for(gem_repo4)}
+          specs:
+            dep_level_2 (1.0)
+              depends_on_platform_specific
+            depends_on_platform_specific (1.0)
+              platform_specific
+            platform_specific (1.0)
+
+        PLATFORMS
+          ruby
+
+        DEPENDENCIES
+          dep_level_2
+          platform_specific
+
+        BUNDLED WITH
+            #{Bundler::VERSION}
+      L
+
+      install_gemfile <<-G, :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }, :artifice => "compact_index", :verbose => true
+        source "#{file_uri_for(gem_repo4)}"
+
+        gem "platform_specific"
+        gem "depends_on_platform_specific"
+      G
+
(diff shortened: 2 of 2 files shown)


segiddins:
(I don't love this fix but I got really fed up after a few hours and figured it was worth getting something up that (a) had a failing spec and (b) fixed that one spec)

bundler/lib/bundler/spec_set.rb
```diff
@@ -202,6 +202,7 @@ def specs_for_dependency(dep, platform)
       specs_for_name = lookup[dep.name]
       target_platform = dep.force_ruby_platform ? Gem::Platform::RUBY : (platform || Bundler.local_platform)
       matching_specs = GemHelpers.select_best_platform_match(specs_for_name, target_platform)
+      matching_specs.each {|s| s.force_ruby_platform = true } if dep.force_ruby_platform
```
martinemde, line 205: Hmm, maybe a little refactor could make this solution feel more elegant to you?

```
matching _specs =
  if dep.force_ruby_platform
    GemHelpers.force_ruby_platform(specs_for_name)
  else
    GemHelpers.select_best_platform_match(specs_for_name, platform || Bundler.local_platform)
  end
```

Then you don't need to iterate through matching_specs twice since it can be handled in GemHelpers more cleanly.

martinemde (commented):
The fix itself makes sense to me, though I don't fully understand the implications from the change.

bundler/lib/bundler/spec_set.rb
segiddins: honestly, feels the same. matching specs should always be a very small array, so the double iteration doesn't feel _that_ bad

bundler/spec/install/gemfile/force_ruby_platform_spec.rb
```diff
@@ -114,5 +114,48 @@
...
+
+      install_gemfile <<-G, :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }, :artifice => "compact_index", :verbose => true
+        source "#{file_uri_for(gem_repo4)}"
+
+        gem "platform_specific"
+        gem "depends_on_platform_specific"
+      G
+
+      expect(the_bundle).to include_gems "depends_on_platform_specific 1.0.0 #{Bundler.local_platform}"
+      expect(the_bundle).to include_gems "platform_specific 1.0.0 #{Bundler.local_platform}"
```
deivid-rodriguez, line 148: This bug seems to happen when a platform specific variant is pre-installed, the Gemfile is using `:force_ruby_platform => true`, and the lockfile includes "RUBY" platform.

In this case, the locally installed platform specific variant will be selected, and unless we propagate the `force_ruby_platform` flag to it from the Gemfile, it will be happily picked as already installed, and used for materialization.

So, I believe transitive dependencies are unrelated to the fix, and I think the whole spec can be simplified to better illustrate this case like this:

```ruby
it "reinstalls the ruby variant when a platform specific variant is already installed, the lockile has only RUBY platform, and :force_ruby_platform is used in the Gemfile" do
  lockfile <<-L
    GEM
      remote: #{file_uri_for(gem_repo4)}
      specs:
        platform_specific (1.0)

    PLATFORMS
      ruby

    DEPENDENCIES
      platform_specific

    BUNDLED WITH
       #{Bundler::VERSION}
  L

  system_gems "platform_specific-1.0-#{Gem::Platform.local}", :path => default_bundle_path

  install_gemfile <<-G, :env => { "BUNDLER_SPEC_GEM_REPO" => gem_repo4.to_s }, :artifice => "compact_index", :verbose => true
    source "#{file_uri_for(gem_repo4)}"

    gem "platform_specific", :force_ruby_platform => true
  G

  expect(the_bundle).to include_gems "platform_specific 1.0.0 RUBY"
end
```

deivid-rodriguez (commented):
Looks great, nice catch!

I only added a comment to simplify the spec to better illustrate the bug.

bundler/spec/install/gemfile/force_ruby_platform_spec.rb
segiddins: maybe just add that as a second spec, since we also want to test the transitive case, and the behavior could very well end up different with a different implementation

deivid-rodriguez: Sure, I just don't see it as related to this PR.

I'm also not sure which role `dep_level2` intends to play since it's in the lockfile but not in the gemfile. I guess kit's testing replacing one dependency in the Gemfile with a subdependency of it? If we want to keep it, I'd at least clarify the wording.

deivid-rodriguez (approved):
Approving this to clarify that my comment on the spec is definitely not blocking. This is a great fix!

PR #6945, opened 2023-09-07
Stop bundler eagerly loading all specs with exts

segiddins:
We were setting the wrong `extension_dir` for git specs stubs

Additionally, the call to `self.extension_dir` was loading the remote spec, which was avoidable since the stub had an extension dir (and in fact its #gem_build_complete_path does exactly what we want anyway)

Finally, now set the base_dir when loading the remote_spec from a stub specification, since the git source sets the base dir for stubs based on where the spec _will_ be installed to, and we want to preserve that so the base_dir for the loaded spec & the stub are the same

Bundler was eagerly loading all gemspecs that had extensions in their stubs

Avoid calling methods that force loading the remote spec inside of `#missing_extensions?`

### DIFF
--- a/bundler/lib/bundler/stub_specification.rb
+++ b/bundler/lib/bundler/stub_specification.rb
@@ -16,7 +16,8 @@ module Bundler
       # Stub has no concept of source, which means that extension_dir may be wrong
       # This is the case for git-based gems. So, instead manually assign the extension dir
       return unless source.respond_to?(:extension_dir_name)
-      path = File.join(stub.extensions_dir, source.extension_dir_name)
+      unique_extension_dir = [source.extension_dir_name, File.basename(full_gem_path)].uniq.join("-")
+      path = File.join(stub.extensions_dir, unique_extension_dir)
       stub.extension_dir = File.expand_path(path)
     end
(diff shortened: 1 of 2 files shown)


indirect (approved):

hanyang-tony:
For universal rubies, the change to `def gem_build_complete_path` is incorrect. Previously, `extension_dir` is invoked on the `Bundler::StubSpecification`, which is a child class of `Specification`. After the change to `stub.gem_build_complete_path`, @stub is actually an instance of `Gem::StubSpecification`, thus `gem_build_complete_path` is from `Gem::BasicSpecification`.

The problem arise when in bundler, we replace the 'universal' part of `Gem::Platform` with the arch. But we still use the universal name for `Gem::Specification`, this is defined in 'rubygems_ext.rb'. Now the stub is using `extension_dir` defined in `Gem::BasicSpecification`, so the 'universal' part ended up being swapped by arch.

So, we install native extensions to 'extensions/universal-darwin-22/', but finding them in 'extensions/x86_64-darwin-22/'. Leading to native extension gems can never be installed on universal ruby.

Issue #2375, opened 2018-08-04
ENOENT when updating from 2.7.6 to 2.7.7


duckinator:
**UPDATE:** Workaround: `gem update --system --install-dir=/usr/local/share/gems --bindir /usr/local/bin`

---

My current problem is upgrading RubyGems 2.7.6 to 2.7.7 results in an `Errno::ENOENT` error when attempting due to system-default `--install-dir`.

## Short Version

- Default gem flags for root, according to `gem env`, are `--install-dir=/usr/local/share/gems --bindir /usr/local/bin`.
- `sudo gem update --system` expects `/usr/share/gems/gems/rubygems-update-2.7.7` to exist.
- `sudo gem update --system` actually puts it in `/usr/local/share/gems/gems/rubygems-update-2.7.7`.
- RubyGems thus fails to find it, and the entire upgrade fails.

I did not set this default, and I am unsure where my system is finding it.

## System Information

- Fedora 28
- Ruby 2.5.1p57 (installed from the official repos)
- RubyGems 2.7.6 (installed from the official repos)
- No gemrc anywhere in /root or /etc.
- No environment variables including the words `RUBY` or `GEM`.

## Details

Output from the attempt:

```
~$ sudo gem update --system
Updating rubygems-update
Successfully installed rubygems-update-2.7.7
Parsing documentation for rubygems-update-2.7.7
Done installing documentation for rubygems-update after 0 seconds
ERROR:  While executing gem ... (Errno::ENOENT)
    No such file or directory @ dir_chdir - /usr/share/gems/gems/rubygems-update-2.7.7
~$
```

It appears to originate in the `Gem::Commands::UpdateCommand#install_rubygems` function:

https://github.com/rubygems/rubygems/blob/5971b486d4dbb2bad5d3445b3801c456eb0ce183/lib/rubygems/commands/update_command.rb#L154-L169

According to the error, it wants: `/usr/share/gems/gems/rubygems-update-2.7.7`
However, it winds up installing it to: 

This issue is related to:

  - [ ] Network problems
  - [ ] Installing a library
  - [ ] Publishing a library
  - [x] The command line `gem`
  - [ ] Other

Here are my current environment details:

```
$ gem env version
2.7.6
```

<details>
<summary>Full `gem env` output.</summary>

```
$ sudo gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 2.7.6
[...]

PR #6946, opened 2023-09-09
Fix the LoadError for requiring a new gem

samfrench:

When creating a new gem of type rust, this worked locally and the specs passed. When adding the gem to another application using bundler it then couldn't find the module referenced from the the lib file in the gem. For this, the gem is called mustachers. The application was inside a Docker container, just in case that adds anything different to this problem.
```
Failure/Error: require 'mustachers'

LoadError:
  cannot load such file -- /code/my_app/src/.gems/gems/mustachers-0.1.1/lib/mustachers/mustachers
# ./.gems/gems/mustachers-0.1.1/lib/mustachers.rb:4:in `require_relative'
# ./.gems/gems/mustachers-0.1.1/lib/mustachers.rb:4:in `<top (required)>'
# ./lib/my_app/run/simple.rb:1:in `<top (required)>'
# ./lib/my_app/run.rb:8:in `require_relative'
# ./lib/my_app/run.rb:8:in `<top (required)>'
# ./lib/my_app.rb:7:in `<top (required)>'
# ./spec/spec_helper.rb:16:in `<top (required)>'
# ./spec/my_app_spec.rb:1:in `<top (required)>'
```

When running the gem from the application directory inside Docker I get the same LoadError:
```
bundle exec ruby -r mustachers -e 'puts Mustachers::Renderer.render("<h1>{{text}}</h1>", {text: "abc123"})'
/code/my_app/src/.gems/gems/mustachers-0.1.1/lib/mustachers.rb:4:in `require_relative': cannot load such file -- /code/my_app/src/.gems/gems/mustachers-0.1.1/lib/mustachers/mustachers (LoadError)
	from /code/my_app/src/.gems/gems/mustachers-0.1.1/lib/mustachers.rb:4:in `<top (required)>'
	from <internal:/usr/local/share/ruby/site_ruby/rubygems/core_ext/kernel_require.rb>:38:in `require'
	from <internal:/usr/local/share/ruby/site_ruby/rubygems/core_ext/kernel_require.rb>:38:in `require'
```

Possibly I could look at paths and variables inside Docker but all other gems worked fine, it was only the new gem created which was type rust. I would expect these to be able to be referenced in any environment, so think it was more a problem of the gem setup which was mostly from the template provided on creation.

I have tried locally on the Mac and seems fine there with require and require_relative inside the gem. The only difference for the Docker setup is Ruby is installed using an rpm and that seems to have a difference in how it loads the gem. To upgrade rubygems in that setup it needed very specific paths which this comment solved. It wasn't the same version but seemed to be have a problem finding the paths to update: https://github.com/rubygems/rubygems/issues/2375#issuecomment-451584753

This Dockerfile replicates the problem. When running the command inside this container, the same error is produced. Using a "require" instead of a "require_relative" in the gem allows the gem to work as expected.
```
FROM rockylinux:8

ENV RUBY_VERSION=3.0

RUN dnf module enable -y ruby:${RUBY_VERSION} && \
    dnf install -y gcc-c++ ruby-devel clang cargo && \
    gem install --install-dir=/usr/share/gems --no-user-install rubygems-update && \
    gem update --system --install-dir=/usr/local/share/gems --bindir /usr/local/bin

RUN gem install mustachers -v 0.1.1

CMD [ "/bin/bash" ]
```

I looked at rubygems and noticed other gems seem to require without relative. I tried this and this fixed it locally for me and allowed me to build the application.

This was the fix in the gem which solved it: https://github.com/samfrench/mustachers/commit/cd4a025c0fae91eefc8336d2b7d0cc567aff3f8d


martinemde:
I'm sorry it took so long for someone to reply to you here. This just slipped by.

Unfortunately this doesn't look correct to me. I'm not sure how the gem ended up with "mustachers/mustachers" but typically the namespace is going to be generated by a gem that uses dashes. So "gem new rails-patch" would generate "lib/rails/patch.rb" where "rails" is the namespace in this context.

I'm not quite sure why `require "mustachers/mustachers"` is working for you because it doesn't look like there is a file there. We want to keep this as a require_relative.

I will close this for now because I don't think this is the right fix for this problem. Please open a new issue if you can reproduce this with a freshly generated gem (and please include the command you used). I really appreciate that you took the time to put together a patch and we'd be happy to keep working on a fix if you encounter this again or can reproduce it.

martinemde:
Ok... I felt like I was missing something so I kept looking into this. I was confused about what part of this was caused by the rust/ext gem and what part is just plain rubygems.

I think the problem here is that with C extensions the .so gets dropped in lib/gemname/gemname.so so this require_relative is right. The question is why this isn't happening the same for rust. I don't think we should change it since this is correct for other extension gems but if it's not working, we should obviously fix it.

Please reopen this if I'm more confused than I realize or if there's a different solution that will keep the template for C extensions if the rust one needs to be different.

@simi, looks like you wrote the article on rust rubygems. Want to double check my thinking here?

samfrench:
I know it may not be an issue here, but the issue is shown in the Dockerfile above. I don't think it's an issue directly with ruby or rust, but seems an issue with the way ruby is available on rocky linux. On other installations it seems fine, just the one provided on the Dockerfile seems to have load problems.

@martinemde, is it a valid issue somewhere using the Dockerfile in the description and the "bundle exec ruby" commands which show the load error?

martinemde:
It certainly does seem like a bug somewhere, but I don't know if it's the docker container or ruby gems.

The most useful thing, if you want to help us out, is if you could reproduce this with a simple set of commands from start (making the gem) to finish (the require_relative fails) without being in a docker container. We don't run individual tests in docker containers like that so it would be hard to write a test case for this as is.

Thanks for your help!

PR #6947, opened 2023-09-10
Fixed false positive SymlinkError in symbolic link directory

negi0109:

False positive SymlinkError is detected when all of the following conditions are met.
- Symlink is included in the gem.
- Symlink is included in the installation location of the gem.

The code that verifies that the symlink points to a location in the gem is changed.

The `real_destination` is a realpath, whereas the `destination_dir` is the path containing the symlink.
Changed to compare after converting both to realpath.

Below is the previous corresponding code.
https://github.com/rubygems/rubygems/blob/fa06cbff5fd5b61de6afdb454367bf38a01b0993/lib/rubygems/package.rb#L424-L434

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -413,6 +413,8 @@ EOM
   # extracted.
 
   def extract_tar_gz(io, destination_dir, pattern = "*") # :nodoc:
+    real_destination_dir = File.realpath(destination_dir)
+
     directories = []
     symlinks = []
 
@@ -428,7 +430,7 @@ EOM
           real_destination = link_target.start_with?("/") ? link_target : File.expand_path(link_target, File.dirname(destination))
 
           raise Gem::Package::SymlinkError.new(full_name, real_destination, destination_dir) unless
-            normalize_path(real_destination).start_with? normalize_path(destination_dir + "/")
+            normalize_path(real_destination).start_with? normalize_path(real_destination_dir + "/")
 
           symlinks << [full_name, link_target, destination, real_destination]
         end
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -413,7 +413,7 @@ EOM
   # extracted.
 
   def extract_tar_gz(io, destination_dir, pattern = "*") # :nodoc:
-    real_destination_dir = File.realpath(destination_dir)
+    destination_dir = File.realpath(destination_dir)
 
     directories = []
     symlinks = []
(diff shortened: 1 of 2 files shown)


lib/rubygems/package.rb
```diff
@@ -413,6 +413,8 @@ def extract_files(destination_dir, pattern = "*")
   # extracted.
 
   def extract_tar_gz(io, destination_dir, pattern = "*") # :nodoc:
+    real_destination_dir = File.realpath(destination_dir)
```
martinemde, line 416: I'd suggest replacing `destination_dir` with the realpath rather than only using this for the symlink comparison. Personally, I think it makes sense to report the errors with the resolved path so it's clear that the destination is a symlink that is being resolved before install.

```suggestion
    destination_dir = File.realpath(destination_dir)
```

In the case where you found this, would this error still be helpful if the symlink was resolved first and any error message showed that realpath?

martinemde (approved):
Thank you for submitting this. This symlink code can be tricky (I have spent a lot of time looking at it also.) Glad you caught this.

lib/rubygems/package.rb
martinemde: In fact, looking at `install_location` we clearly replace the `destination_dir` with its realpath, then use that realpath in the almost identical error generated therein if a file is trying to be installed outside of the destination dir. If we resolve the realpath of `destination_dir` at the top here then we can do it just once and remove the extra realpath in install_location.

negi0109: > In the case where you found this, would this error still be helpful if the symlink was resolved first and any error message showed that realpath?

Since the path in the error message revealed that symlink was the cause, it would be better to output the realpath that is actually used for comparison as the error message.


> If we resolve the realpath of destination_dir at the top here then we can do it just once and remove the extra realpath in install_location.

It certainly looks good, Replace `destination_dir` with realpath.

negi0109: I am concerned about the processing differences because `install_location` is a public method, but the symlink validation works well enough, and the naming of `install_location` does not require it to be a realpath, so I do not see a major problem.

However, There was one problem.
In the test code for `install_location`, the `destination_dir` contains `..` in the test code.
This may be a reason to convert it to a realpath when calling install_location, so this process should not be removed.
(https://github.com/rubygems/rubygems/blob/fa06cbff5fd5b61de6afdb454367bf38a01b0993/test/rubygems/test_gem_package.rb#L756C1-L770C6 )

If allow it to become a destructive change,It looks good to remove this test case `test_install_location_dots` and remove the conversion from install_location to realpath, but I can't decide.

martinemde: Thank you for digging into this.

I'd say replacing destination_dir with realpath is good, but don't remove realpath from install_location yet. Calling realpath twice won't hurt. The second time might be slightly more efficient anyway.

We can worry about optimizing and maybe breaking install_location after the bug is fixed. The tests clearly access it as if it is an intentional public method.

negi0109: > We can worry about optimizing and maybe breaking install_location after the bug is fixed. 

Yes, it is. This PR will only fix bugs. Thank you for your response.

PR #6954, opened 2023-09-13
Gemfile `ruby file:` supports `ruby-3.2.2` but not `3.2.2@gemset`

martinemde:

Follow up comments in issue #6876 brought up missing support in `.ruby-version`.

- Increase test coverage to explicitly test what is and is not supported.
- Add support for `ruby-3.2.2` with the `ruby-` prefix, which is common and easy to support.

### DIFF
--- a/bundler/lib/bundler/ruby_dsl.rb
+++ b/bundler/lib/bundler/ruby_dsl.rb
@@ -10,13 +10,8 @@ module Bundler
       raise GemfileError, "Please define :engine" if options[:engine_version] && options[:engine].nil?
 
       if options[:file]
-        raise GemfileError, "Cannot specify version when using the file option" if ruby_version.any?
-        file_content = Bundler.read_file(Bundler.root.join(options[:file]))
-        if /^ruby\s+(.*)$/.match(file_content) # match .tool-versions files
-          ruby_version << $1.split("#", 2).first.strip # remove trailing comment
-        else
-          ruby_version << file_content.strip
-        end
+        raise GemfileError, "Do not pass version argument when using :file option" unless ruby_version.empty?
+        ruby_version << normalize_ruby_file(options[:file])
       end
 
       if options[:engine] == "ruby" && options[:engine_version] &&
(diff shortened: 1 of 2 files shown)


bundler/spec/bundler/ruby_dsl_spec.rb
```diff
@@ -21,11 +21,13 @@ class MockDSL
...
       :engine_version => engine_version }
   end
+  let(:project_root) { Pathname.new("/path/to/project") }
+  before { allow(Bundler).to receive(:root).and_return(project_root) }
 
   let(:invoke) do
     proc do
       args = []
-      args << Array(ruby_version_arg) if ruby_version_arg
+      args << ruby_version_arg if ruby_version_arg
```
martinemde, line 30: Changed this to be slightly more accurate in tests because the arg is not always given as an array in the Gemfile.

bundler/spec/bundler/ruby_dsl_spec.rb
```diff
@@ -97,53 +99,67 @@ class MockDSL
...
-      let(:engine) { "ruby" }
-      let(:project_root) { Pathname.new("/path/to/project") }
+      context "with a @gemset" do
+        let(:file_content) { "ruby-#{version}@gemset\n" }
 
-      before do
-        allow(Bundler).to receive(:read_file).with(project_root.join("foo")).and_return("nodejs 18.16.0\nruby #{version} # This is a comment\npnpm 8.6.12\n")
-        allow(Bundler).to receive(:root).and_return(Pathname.new("/path/to/project"))
+        it "raises an error" do
+          expect { subject }.to raise_error(Gem::Requirement::BadRequirementError, "Illformed requirement [\"#{version}@gemset\"]")
```
martinemde, line 136: I covered the reason for rejecting `@gemset` in the linked ticket. TL;DR rvm says it's incompatible with other version managers and you should use `.ruby-gemset`.

pboling (commented):
LGTM

bundler/lib/bundler/ruby_dsl.rb
```diff
@@ -25,5 +20,26 @@ def ruby(*ruby_version)
       end
       @ruby_version = RubyVersion.new(ruby_version, options[:patchlevel], options[:engine], options[:engine_version])
     end
+
+    # Support the various file formats found in .ruby-version files.
+    #
+    #     3.2.2
+    #     ruby-3.2.2
```
segiddins, line 27: What about other ruby implementations, like jruby or truffleruby?

segiddins (approved):

bundler/lib/bundler/ruby_dsl.rb
martinemde: This is a bit weird since specifying jruby or truffleruny require 3 values: version, engine, and engine version. I tested a bunch of ways to allow this. Bundler needs version, engine version and engine. .ruby-version doesn't support syntax like that.

The only solution I see is putting engine: and engine_version: in the options with file:, and having file be the underlying ruby version that is given by the engine. Bundler doesn't like getting the jruby's engine version as the ruby version, as far as I found in testing, so we'd have to load the ruby version from the environment if .ruby-version had jruby-0.9.2.0.

pboling: With `asdf` only MRI rubies are the simple version.
Alternate values are the same as with `rbenv`, or `ruby-build` (which is what `asdf-ruby` uses under the hood).
Many of those are not numeric at all, or are numeric but the numbers are not based on MRI versions:
```
artichoke-dev
jruby-dev
jruby-1.7.27
jruby-9.4.3.0
maglev-2.0.0-dev
mruby-dev
mruby-3.2.0
picoruby-3.0.0
rbx-5.0
ree-1.8.7-2012.02
truffleruby-dev
truffleruby-23.0.0
truffleruby+graalvm-dev
truffleruby+graalvm-23.0.0
```

Issue #6951, opened 2023-09-11
bundle outdated uses incorrect git command when branch and ref specified


pboling:
### Describe the problem as clearly as you can

```
$ bundle outdated        
Fetching https://github.com/bmallred/spork-rails
Fetching https://github.com/pboling/simplecov-rcov.git
Fetching https://github.com/seielit/aws-sdk-v1-ruby24.git
Git error: command `git rev-parse --verify aws-sdk-v1-ruby24` in directory /Users/pboling/.asdf/installs/ruby/2.7.8/lib/ruby/gems/2.7.0/cache/bundler/git/aws-sdk-v1-ruby24-16fca015a7035a21a66dda9f530d389ba25367a2 has failed.
Revision aws-sdk-v1-ruby24 does not exist in the repository https://github.com/seielit/aws-sdk-v1-ruby24.git. Maybe you misspelled it?
```

### Did you try upgrading rubygems & bundler?

Yes.

### Post steps to reproduce the problem

In a gemfile:
```
# Version 1 of the AWS SDK for Ruby.
# Available as both `aws-sdk` and `aws-sdk-v1`.
# Use `aws-sdk-v1` if you want to load v1 and v2 of the Ruby SDK in the same application
# gem 'aws-sdk-v1', '~> 1.67'
# We use a fork that only updates the json dependency, because we aren't cave dwellers.
# See Thread: https://github.com/aws/aws-sdk-ruby/issues/1372#issuecomment-1434167290
gem "aws-sdk-v1-ruby24",
  "1.67.0",
  github: "seielit/aws-sdk-v1-ruby24",
  branch: "aws-sdk-v1-ruby24",
  ref: "41b1829a12d020a3f7880de958d7b3f4eebc8436",
  require: "aws-sdk-v1"
```

### Which command did you run?

```
bundle outdated
```

### What were you expecting to happen?

Information about which gems are out of date.

### What actually happened?

One of the git commands failed, and that failure bubbled up to bundler.  The problem is that the git command which ran and failed was the wrong command.

If I run the command which failed by hand it does fail, making it clear why it failed for the `bundle outdated` command:
```
/Users/pboling/.asdf/installs/ruby/2.7.8/lib/ruby/gems/2.7.0/cache/bundler/git/aws-sdk-v1-ruby24-16fca015a7035a21a66dda9f530d389ba25367a2 git(master) ruby-2.7.8 
$ git rev-parse --verify aws-sdk-v1-ruby24
[...]

PR #6959, opened 2023-09-14
Fix GitProxy for branch and ref

pboling:
Fixes https://github.com/rubygems/rubygems/issues/6951 to honor `ref` > `branch`.  This in turn fixes `bundle outdated` for Gemfiles with dependencies specifying both `branch` and `ref`.

Also:
- Raises an error when `ref` or `branch` are specified with `tag`, as that's an ambiguous state.
- Improves spec for GitProxy

1. When specifying both `branch` and `ref` on a git source, some bundler commands break (e.g. `bundle outdated` see #6951)
2. If it were to work, `branch` had higher precedence than `ref`, which is incorrect.
3. `tag` was in between `branch` and `ref` in precedence, which doesn't make any sense, and in fact, `tag` should never be specified with `branch` or `ref`, as it creates an ambiguous reference that has until now been arbitrarily decided by the code in bundler (should it be the tag, or the branch/ref?).
4. Last commit is strictly RuboCop lint fixes, which, if undesired as unrelated, I'm happy to pop off of this PR to keep it focused.
5. First commit just synchronizes the platforms for latest MacOS/Darwin across some lockfiles.

This implementation raises an error, and I expect it can affect users who unwittingly specified ambiguous references in their Gemfile, but for whom `bundle` commands currently work.  As such this might be considered a breaking change, but it is also definitely a bugfix.  In many cases this would only affect users for whom the bundle / git commands are already broken, such as the case in #6951.

### DIFF
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -43,6 +43,13 @@ module Bundler
         end
       end
 
+      class AmbiguousGitReference < GitError
+        def initialize(options)
+          msg = "Specification of branch or ref with tag is ambiguous. You specified #{options.inspect}"
+          super msg
+        end
+      end
+
       # The GitProxy is responsible to interact with git repositories.
       # All actions required by the Git source is encapsulated in this
       # object.
@@ -53,10 +60,15 @@ module Bundler
         def initialize(path, uri, options = {}, revision = nil, git = nil)
           @path     = path
           @uri      = uri
-          @branch   = options["branch"]
           @tag      = options["tag"]
+          @branch   = options["branch"]
           @ref      = options["ref"]
-          @explicit_ref = branch || tag || ref
+          if @tag
+            raise AmbiguousGitReference.new(options) if @branch || @ref
+            @explicit_ref = @tag
+          else
+            @explicit_ref = @ref || @branch
+          end
           @revision = revision
           @git      = git
           @commit_ref = nil
--- a/bundler/spec/bundler/source/git/git_proxy_spec.rb
+++ b/bundler/spec/bundler/source/git/git_proxy_spec.rb
@@ -3,34 +3,84 @@
 RSpec.describe Bundler::Source::Git::GitProxy do
   let(:path) { Pathname("path") }
   let(:uri) { "https://github.com/rubygems/rubygems.git" }
-  let(:ref) { "HEAD" }
+  let(:ref) { nil }
+  let(:branch) { nil }
+  let(:tag) { nil }
+  let(:options) { { "ref" => ref, "branch" => branch, "tag" => tag }.compact }
   let(:revision) { nil }
   let(:git_source) { nil }
   let(:clone_result) { double(Process::Status, :success? => true) }
   let(:base_clone_args) { ["clone", "--bare", "--no-hardlinks", "--quiet", "--no-tags", "--depth", "1", "--single-branch"] }
-  subject { described_class.new(path, uri, ref, revision, git_source) }
+  subject(:git_proxy) { described_class.new(path, uri, options, revision, git_source) }
+
+  context "with explicit ref" do
+    context "with branch only" do
+      let(:branch) { "main" }
+      it "sets explicit ref to branch" do
+        expect(git_proxy.explicit_ref).to eq(branch)
+      end
+    end
+
+    context "with ref only" do
+      let(:ref) { "HEAD" }
+      it "sets explicit ref to ref" do
+        expect(git_proxy.explicit_ref).to eq(ref)
+      end
+    end
+
+    context "with tag only" do
+      let(:tag) { "v1.0" }
+      it "sets explicit ref to ref" do
+        expect(git_proxy.explicit_ref).to eq(tag)
+      end
+    end
+
+    context "with tag and branch" do
+      let(:tag) { "v1.0" }
+      let(:branch) { "main" }
+      it "raises error" do
+        expect { git_proxy }.to raise_error(Bundler::Source::Git::AmbiguousGitReference)
+      end
+    end
+
+    context "with tag and ref" do
+      let(:tag) { "v1.0" }
+      let(:ref) { "HEAD" }
+      it "raises error" do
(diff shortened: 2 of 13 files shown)


bundler/spec/bundler/source/git/git_proxy_spec.rb
```diff
@@ -3,127 +3,177 @@
...
-  let(:ref) { "HEAD" }
+  let(:ref) { nil }
+  let(:branch) { nil }
+  let(:tag) { nil }
+  let(:options) { { "ref" => ref, "branch" => branch, "tag" => tag }.compact }
   let(:revision) { nil }
   let(:git_source) { nil }
   let(:clone_result) { double(Process::Status, :success? => true) }
   let(:base_clone_args) { ["clone", "--bare", "--no-hardlinks", "--quiet", "--no-tags", "--depth", "1", "--single-branch"] }
-  subject { described_class.new(path, uri, ref, revision, git_source) }
```
pboling, line 11: Note that the signature for GitProxy.new is not followed here, sending `'HEAD'` as `ref` instead of the actual expected argument of `options` as a `Hash`.  This is the case for all tests in this file, clearly indicating that these tests did zero testing of any `options`.  It was working only because of ducky handling:
```
>> opt = "HEAD"
=> "HEAD"
>> opt["tag"]
=> nil
```

bundler/spec/bundler/source/git/git_proxy_spec.rb
```diff
@@ -3,127 +3,177 @@
...
+  subject(:instance) { described_class.new(path, uri, options, revision, git_source) }
```
segiddins, line 14: is it necessary to change the name of the `subject` here?  it makes the diff harder to read

bundler/lib/bundler/source/git/git_proxy.rb
```diff
@@ -53,10 +60,16 @@ class GitProxy
...
+          @branch   = options["branch"]
           @ref      = options["ref"]
-          @explicit_ref = branch || tag || ref
+          @explicit_ref = if @tag
+            raise AmbiguousGitReference.new(options) if @branch || @ref
+
+            @tag
+          else
+            @ref || @branch
+          end
```
martinemde, line 72: Style tweak. This branch is complicated. How about something more like: 

```suggestion
          raise AmbiguousGitReference.new(options) if @tag && (@branch || @ref)
          @explicit_ref = @tag || @ref || @branch
```

bundler/spec/bundler/source/git/git_proxy_spec.rb
martinemde, line 14: Nit: I know this is rubocop's preference but `instance` feels equally generic. Maybe `git_proxy` or just leave it as `subject` since it's already quite common in these tests. I don't mind merging this as is but if we're going to change it at all, maybe we can make it more clear.

martinemde (approved):
This is great. Thank you for this. I had made a quick pass at flipping the order, saw that not one single test failed, and decided I would have to come back to it. Thanks for taking the time to increase the test coverage and make more than zero tests cover this behavior, and for raising the error when the nonsensical is given.

martinemde:
By the way, I'm fine with the multiple unrelated commits. We just won't squash this when we merge it so the commit messages can stay coupled to their changes even though they are not themselves related.

bundler/spec/bundler/source/git/git_proxy_spec.rb
pboling: I am going to change `instance` to `git_proxy`, as that is a more clear name.  Using subject directly is "bad rspec practice" (according to rubocop-rspec), particularly when overriding the subject inside a test , as had been done on [old line 41](https://github.com/rubygems/rubygems/pull/6959/files#diff-9ede4bd79279fea44d1c52a45f234c711d280867436ec01be6c948a17d5a3c93L41).

bundler/lib/bundler/source/git/git_proxy.rb
martinemde: ok, I'm fine with that, but can you move the assignment into the `if`? I'll update my suggestion above to clarify what I mean. Something about the nested raise/if within an nested assignment if. The logical purity is there, but the assignment purity is obscured.

bundler/spec/bundler/source/git/git_proxy_spec.rb
pboling: Amended the commit, and it is now named `git_proxy`.

bundler/lib/bundler/source/git/git_proxy.rb
pboling: Another way of explaining my discomfort with the "simple" way is that these two lines "look" like a bug to someone unfamiliar with the reasoning behind the logic.
```
          raise AmbiguousGitReference.new(options) if @tag && (@branch || @ref)
          @explicit_ref = @tag || @ref || @branch
```
If I saw this in business logic app code I'd ask for a comment to explain, or a refactor.  The logic as-is needs no explanation because the branching logic isn't collapsed.

bundler/spec/bundler/source/git/git_proxy_spec.rb
pboling: Done, it is now `git_proxy`.

bundler/lib/bundler/source/git/git_proxy.rb
martinemde: You make a good point. I felt a twinge of the same feeling when I collapsed it to the above. My real beef is with the assignment outside a complicated if. I'd rather duplication than obscuration.

pboling: Though I do know that with some RuboCop configs that'll be flagged, we'll see about this one.

pboling: Amended commit. 😄

pboling: RuboCop did not flag on either of the rules I thought it might. 👍

pboling:
Rebased on latest upstream/master

bundler/lib/bundler/source/git/git_proxy.rb
martinemde: Rubucop has shown us compassion today.

PR #6960, opened 2023-09-14
Allow installing plugins from path via CLI

### LINKED ISSUE #6953: `bundle plugin uninstall` wipes out your code if using a path

In my Gemfile, I have something like

`plugin "my_plugin", path: "../my_plugin"`

I was testing some installation behavior of my plugin, and so needed to re-install it. First I ran `bundle plugin uninstall my_plugin` to clean it out of the plugin index. Then to my horror, I realized that the entire my_plugin directory was gone!

Yes. I'm on rubygems 3.4.19, and bundler 2.4.19.

Because it's a `path:` reference in the gemfile, I expected it to only get removed from .bundle/plugins/index, but the actual plugin (which is not cached in .bundle) would remain untouched.

### What actually happened?

My local code was removed!

## Environment

```
Bundler       2.4.19
Ruby          3.1.3p185 (2022-11-24 revision 1a6b16756e0ba6b95ab71a441357ed5484e33498) [arm64-darwin-21]
  Full Path   /Users/cody/.rubies/ruby-3.1.3/bin/ruby
  Config Dir  /Users/cody/.rubies/ruby-3.1.3/etc
RubyGems      3.4.19
  Gem Home    /Users/cody/.gem/ruby/3.1.3
  Gem Path    /Users/cody/.gem/ruby/3.1.3:/Users/cody/.rubies/ruby-3.1.3/lib/ruby/gems/3.1.0
  User Home   /Users/cody
  User Path   /Users/cody/.gem/ruby/3.1.0
  Bin Dir     /Users/cody/.gem/ruby/3.1.3/bin
Tools         
  Git         2.39.2 (Apple Git-143)
  RVM         not installed
  rbenv       not installed
  chruby      0.3.9
```

## Bundler Build Metadata

```
Built At          2023-08-17
Git SHA           86f98098e3
Released Version  true
```

## Bundler settings

```
gem.changelog
  Set for the current user (/Users/cody/.bundle/config): false
gem.ci
  Set for the current user (/Users/cody/.bundle/config): false
gem.coc
  Set for the current user (/Users/cody/.bundle/config): false
gem.linter
  Set for the current user (/Users/cody/.bundle/config): "rubocop"
gem.mit
  Set for the current user (/Users/cody/.bundle/config): false
gem.test
  Set for the current user (/Users/cody/.bundle/config): false
gemfile
  Set via BUNDLE_GEMFILE: "/Users/cody/src/canvas-lms/Gemfile"
unset
  Set for the current user (/Users/cody/.bundle/config): "frozen"
```

## Gemfile

```
<snip>
```

ccutrer:
Refs #3319

You can install a local plugin gem in the Gemfile via `plugin 'plugin', path: '../plugin'`, but you cannot do the analogous `bundle plugin install plugin --path=../plugin` from the CLI.

Refactor how sources are passed to the plugin installer, support the path source, remove duplicate code for git sources, and update docs.

### DIFF
--- a/bundler/lib/bundler/plugin/installer.rb
+++ b/bundler/lib/bundler/plugin/installer.rb
@@ -10,6 +10,7 @@ module Bundler
     class Installer
       autoload :Rubygems, File.expand_path("installer/rubygems", __dir__)
       autoload :Git,      File.expand_path("installer/git", __dir__)
+      autoload :Path, File.expand_path("installer/path", __dir__)
 
       def install(names, options)
         check_sources_consistency!(options)
@@ -18,6 +19,8 @@ module Bundler
 
         if options[:git]
           install_git(names, version, options)
+        elsif options[:path]
+          install_path(names, version, options[:path])
         else
           sources = options[:source] || Gem.sources
           install_rubygems(names, version, sources)
@@ -50,8 +53,8 @@ module Bundler
           options[:git] = options.delete(:local_git)
         end
 
-        if (options.keys & [:source, :git]).length > 1
-          raise InvalidOption, "Only one of --source, or --git may be specified"
+        if (options.keys & [:source, :git, :path]).length > 1
+          raise InvalidOption, "Only one of --source, --git, or --path may be specified"
         end
 
         if (options.key?(:branch) || options.key?(:ref)) && !options.key?(:git)
@@ -64,10 +67,19 @@ module Bundler
       end
 
       def install_git(names, version, options)
-        uri = options.delete(:git)
-        options["uri"] = uri
+        source_list = SourceList.new
+        source = source_list.add_git_source({ "uri" => options[:git],
+                                              "branch" => options[:branch],
+                                              "ref" => options[:ref] })
 
-        install_all_sources(names, version, options, options[:source])
+        install_all_sources(names, version, source_list, source)
+      end
+
+      def install_path(names, version, path)
+        source_list = SourceList.new
+        source = source_list.add_path_source({ "path" => path })
+
+        install_all_sources(names, version, source_list, source)
       end
 
       # Installs the plugin from rubygems source and returns the path where the
@@ -79,16 +91,15 @@ module Bundler
       #
       # @return [Hash] map of names to the specs of plugins installed
       def install_rubygems(names, version, sources)
-        install_all_sources(names, version, nil, sources)
-      end
-
-      def install_all_sources(names, version, git_source_options, rubygems_source)
         source_list = SourceList.new
 
-        source_list.add_git_source(git_source_options) if git_source_options
-        Array(rubygems_source).each {|remote| source_list.add_global_rubygems_remote(remote) } if rubygems_source
+        Array(sources).each {|remote| source_list.add_global_rubygems_remote(remote) }
 
-        deps = names.map {|name| Dependency.new name, version }
+        install_all_sources(names, version, source_list)
+      end
+
+      def install_all_sources(names, version, source_list, source = nil)
+        deps = names.map {|name| Dependency.new(name, version, { "source" => source }) }
 
         Bundler.configure_gem_home_and_path(Plugin.root)
 
--- a/bundler/spec/plugins/install_spec.rb
+++ b/bundler/spec/plugins/install_spec.rb
@@ -212,6 +212,16 @@ RSpec.describe "bundler plugin install" do
     end
   end
 
+  it "installs from a path source" do
+    build_lib "path_plugin" do |s|
+      s.write "plugins.rb"
+    end
+    bundle "plugin install path_plugin --path #{lib_path("path_plugin-1.0")}"
+
+    expect(out).to include("Installed plugin path_plugin")
+    plugin_should_be_installed("path_plugin")
+  end
+
   context "Gemfile eval" do
     before do
       allow(Bundler::SharedHelpers).to receive(:find_gemfile).and_return(bundled_app_gemfile)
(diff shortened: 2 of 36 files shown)


ccutrer:
I rebased, and removed the accidental change to the rubocop tools lockfile.

deivid-rodriguez:
You just need a couple more tweaks to get CI green. Run `rake update_manifest`, `rake man:build`, and amend the latest commit with the changes generated.

Also, I would probably extract the bug fix to a separate PR and leave this to the `--path` CLI flag addition?

ccutrer:
Done. Bugfix split off to #7001 (For some reason I was thinking it depended on this commit, so put both commits in the same PR).

ccutrer:
@deivid-rodriguez: any further comments on this PR?

bundler/lib/bundler/plugin/installer.rb
```diff
@@ -45,20 +46,35 @@ def check_sources_consistency!(options)
         if options.key?(:git) && options.key?(:local_git)
           raise InvalidOption, "Remote and local plugin git sources can't be both specified"
         end
+        options[:git] = options.delete(:local_git) if options.key?(:local_git)
```
deivid-rodriguez, line 49: It the `--local_git` option now a loop? I guess we should keep it working as before for now.

deivid-rodriguez (commented):
Looks really good, just have some comments on the `--local_git` option.

bundler/lib/bundler/plugin/installer.rb
ccutrer: it simply renames the option to :git here, before choosing the install method. the old install_git and install_local_git methods were identical (except choosing a different option key). so actual functionality has not changed

deivid-rodriguez: Oh, my bad, I read the code incorrectly, thanks for clarifying!

bundler/spec/plugins/install_spec.rb
```diff
@@ -92,16 +92,18 @@
     expect(out).to include("Using foo 1.1")
   end
 
-  it "installs when --branch specified" do
-    bundle "plugin install foo --branch main --source #{file_uri_for(gem_repo2)}"
+  it "raises an error when when --branch specified" do
+    bundle "plugin install foo --branch main --source #{file_uri_for(gem_repo2)}", :raise_on_error => false
```
deivid-rodriguez, line 96: This is another different bug fix / change included with this PR, right? Previously we would interpret `--source` as a `git` source if `branch` or `ref` were specified. Now we're more strict about it.

I'm in general in favor of this since it makes plugin sources more consistent with Gemfile sources, but since this was coded as specs, maybe extract a PR & discuss/review there?

deivid-rodriguez (commented):
The PR looks good in general, but there's a few different things going on, and while I like all of them, I think perhaps the change about `--source` in combination with `--branch`, or `--ref` could be extracted as a separate PR. Particularly since it was codified in specs so maybe someone wanted this for a reason.

deivid-rodriguez:
@ccutrer Apologies for that extra nit where I had basically approved this earlier. Just noticed this in the last review pass when I was going to merge it. I'm also happy to do this extraction myself if you prefer!

ccutrer:
Not a problem. I'll hopefully get around to it on Monday or Tuesday. I've been super busy updating lots of gems since the Rails 7.1 release. I wanted to to re-look at this to understand if it actually made sense to do `--source ... --branch ...`.

bundler/spec/plugins/install_spec.rb
ccutrer: I've actually broken this into 4 commits now, but they all depend on each other. I don't know if you have a preference to just have 4 commits in this PR, or one PR per commit. #7045 is the first commit in the stack if you prefer separate PRs; I'll create the next PR once that's merged, etc.

ccutrer:
I actually split this into 4 atomic commits (that are all dependent on each other to avoid conflicts). I'm not sure if you have a preference for 1-commit-per-PR or not. If so, #7045 is the first commit in the stack, and as each one is submitted, I'll open the next PR. For now the branch this PR is from has all the commits, and I'll rebase it once its predecessors have been submitted separately.

deivid-rodriguez:
Wow, that's so nice of you. Generally I don't mind small extra changes sneaked into PR, but since you split this already, separate PRs are indeed better because they will show up as different improvements in our release notes. Thanks for doing this, I'll start with the docs PR!

ccutrer:
#7048 is the second in the stack

ccutrer:
Rebased and re-worked spec to use gemfile installation, so this PR no longer has any dependencies.

deivid-rodriguez:
I merged the deprecation of `bundle plugin install --local_git`. https://github.com/rubygems/rubygems/pull/6960/commits/f44322cad11e5d2507b367a35f7eb78f6e564440 can be extracted now to a separate PR.

deivid-rodriguez:
I extracted last PR into #7529.

deivid-rodriguez:
I merged #7529 and gave this PR one last rebase!

deivid-rodriguez (approved):
Alright, let's wrap this up. Thanks so much for all the improvements @ccutrer!

ccutrer:
This has caused a regression in Bundler 2.5.8 that plugin ..., path: ... in a Gemfile can no longer find the plugin. Investigating now...

deivid-rodriguez:
Thanks for the super fast report and initial investigation. I can get a quick release out once we fix the problem.

PR #6961, opened 2023-09-14
Add bundler before/after eval hooks for plugins

ccutrer:

My plugin would have a much cleaner implementation if it could hook both before and after the Gemfile is evaluated (the former to easily add additional DSL methods, the latter to do verification that the new methods were properly used).

Add the hooks and call them from `Definition.build` (not `Dsl#evaluate` or `Dsl#eval_gemfile`, so as not to interfere with the plugin-parsing pass of the Gemfile).

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -35,7 +35,10 @@ module Bundler
 
       raise GemfileNotFound, "#{gemfile} not found" unless gemfile.file?
 
-      Dsl.evaluate(gemfile, lockfile, unlock)
+      Plugin.hook(Plugin::Events::GEM_BEFORE_EVAL, gemfile, lockfile, unlock)
+      Dsl.evaluate(gemfile, lockfile, unlock).tap do |definition|
+        Plugin.hook(Plugin::Events::GEM_AFTER_EVAL, definition)
+      end
     end
 
     #
--- a/bundler/lib/bundler/plugin/events.rb
+++ b/bundler/lib/bundler/plugin/events.rb
@@ -56,6 +56,18 @@ module Bundler
       #   Includes an Array of Bundler::Dependency objects
       #   GEM_AFTER_INSTALL_ALL = "after-install-all"
       define :GEM_AFTER_INSTALL_ALL,  "after-install-all"
+
+      # @!parse
+      #   A hook called before the Gemfile is evaluated
+      #   Includes the Gemfile name, the Lockfile name, and the unlock options
+      #   GEM_BEFORE_EVAL = "before-eval"
+      define :GEM_BEFORE_EVAL, "before-eval"
+
+      # @!parse
+      #   A hook called after the Gemfile is evaluated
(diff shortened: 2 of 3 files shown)


ccutrer:
This has caused a regression in Bundler 2.5.8 that `plugin ..., path: ...` in a Gemfile can no longer find the plugin. Investigating now...

binford2k:
This hook would GREATLY help our situation. In the Puppet ecosystem, modules are tested via a `Gemfile` that declares all the ruby code required to run the tests. One of the requirements is `puppet` which may now be provided by:
- a gem named `openvox`
- a gem named `puppet` from ruby gems
- a gem named `puppet` from a private authenticated gem server

Each gem is very nearly equivalent and the real difference comes from transitive dependencies. But in order to test that a module works on all three implementations, we need to allow the site operator to select which implementation to install and use.

It's infeasible to require every author of every module to add boilerplate code to their Gemfile to select a gem to install.

My https://github.com/overlookinfra/bundler-alias plugin now handles all but one case, and that would be enabled by this contribution.

bundler/lib/bundler/definition.rb
```diff
@@ -35,7 +35,10 @@ def self.build(gemfile, lockfile, unlock)
 
       raise GemfileNotFound, "#{gemfile} not found" unless gemfile.file?
 
-      Dsl.evaluate(gemfile, lockfile, unlock)
+      Plugin.hook(Plugin::Events::GEM_BEFORE_EVAL, gemfile, lockfile, unlock)
```
Edouard-chin, line 38: I think the hook should be in the `dsl` file. If the intent is to let plugin extend the dsl, then we should pass the dsl class to the hook.
So the plugin would be able to do something like:

```ruby
Bundler::Plugin::API.hook(Plugin::Events) do |dsl|
  dsl.define_method(:overrides!) {}
end
```

We'd also need to passthrough method_missing https://github.com/ruby/rubygems/blob/43f15a14019d8c0046392133d36e542314285da3/bundler/lib/bundler/plugin/dsl.rb#L35 in the DSL plugin, otherwise users won't be able to use ours custom dsl methods.
If a user accidentaly call a methods that doesn't exist (either in the vanilla Bundler dsl or our custom plugins) it will be caught afterwards when Bundler evaluates the DSL from the DSL class
https://github.com/ruby/rubygems/blob/43f15a14019d8c0046392133d36e542314285da3/bundler/lib/bundler/dsl.rb#L236-L239

Edouard-chin (commented):
Having a hook to extend the DSL is also something I'm looking for. I was about to propose a patch but saw this PR.

For the maintainers that will review this PR, having an official way to extend the dsl would allow Rubist to more easily build tooling, and implement Bundler features that could make it upstream if they are popular in the community. 
Right now, it's possible but your Gemfile would look like this:


```ruby
plugin 'foo'

if Plugin.installed?(:foo)
  Plugin.send(:load_plugins, :foo)

  my_cool_new_dsl_method
end
```

PR #6962, opened 2023-09-15
Avoid allocating empty hashes in Index

segiddins:
Since the hashes have a default proc that returns a (new) empty hash, we
can avoid allocating those empty hashes when we are only doing lookups.

Test from running `bundle update --bundler` against a rails app I have
lying around:

```
==> memprof.after.txt <==
Total allocated: 9.71 MB (68282 objects)
Total retained:  4.87 MB (33791 objects)

==> memprof.before.txt <==
Total allocated: 10.83 MB (100596 objects)
Total retained:  5.02 MB (34721 objects)
```

### DIFF
--- a/bundler/lib/bundler/index.rb
+++ b/bundler/lib/bundler/index.rb
@@ -19,14 +19,14 @@ module Bundler
     def initialize
       @sources = []
       @cache = {}
-      @specs = Hash.new {|h, k| h[k] = {} }
+      @specs = {}
       @duplicates = {}
     end
 
     def initialize_copy(o)
       @sources = o.sources.dup
       @cache = {}
-      @specs = Hash.new {|h, k| h[k] = {} }
+      @specs = {}
       @duplicates = {}
 
       o.specs.each do |name, hash|
@@ -46,14 +46,11 @@ module Bundler
       true
     end
 
-    def search_all(name)
-      all_matches = specs_by_name(name) # always returns a new Array
-      dupes = @duplicates[name]
-      all_matches.concat(dupes) if dupes
-      @sources.each do |source|
-        all_matches.concat(source.search_all(name))
-      end
-      all_matches
+    def search_all(name, &blk)
+      return enum_for(:search_all, name) unless blk
+      specs_by_name(name).each(&blk)
+      @duplicates[name]&.each(&blk)
+      @sources.each {|source| source.search_all(name, &blk) }
     end
 
     # Search this index's specs, and any source indexes that this index knows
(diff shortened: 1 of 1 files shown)


martinemde (approved):
Great find, thanks @segiddins!

bundler/lib/bundler/index.rb
```diff
@@ -170,31 +160,41 @@ def add_source(index)
 
     private
 
+    def search_by_query(query)
```
martinemde, line 163: There were no callers except the one here. Moved to private and changed to return nil when nothing found so calling code can more easily reduce array usage.

bundler/lib/bundler/source/rubygems.rb
```diff
@@ -314,10 +314,11 @@ def credless_remotes
       end
 
       def remotes_for_spec(spec)
-        specs.search_all(spec.name).inject([]) do |uris, s|
+        uris = []
+        specs.with_all_matching(spec.name) do |s|
```
martinemde, line 318: this was the only caller of search_all, and in just injected the result to make a list of arrays. It's way less overhead if we directly yield each search result.

bundler/lib/bundler/index.rb
```diff
@@ -170,31 +160,41 @@ def add_source(index)
...
+      end
+    end
+
     def add_duplicate(spec)
       (@duplicates[spec.name] ||= []) << spec
     end
 
     def specs_by_name_and_version(name, version)
-      specs_by_name(name).select {|spec| spec.version == version }
+      specs_by_name(name)&.select {|spec| spec.version == version }
```
segiddins, line 178: I think this can `select!` since specs by name always returns a new array

bundler/lib/bundler/index.rb
```diff
@@ -46,42 +46,32 @@ def empty?
...
 
       @sources.each do |source|
-        results.concat(source.search(query))
+        if results.empty?
+          results = source.search(query)
+        else
+          results.concat(source.search(query))
+        end
       end
       results.uniq(&:full_name)
```
segiddins, line 68: Can this be `uniq!` ?

martinemde: Let me see. Tracking the EMPTY_SEARCH as it potentially falls through the nested search determines whether we can modify the result.

segiddins: we could wait to return `EMPTY_SEARCH` until the end, and have `results = results&.concat(...) || ...` (or something like that)

martinemde: I think I figured out a way around it. One moment.

bundler/lib/bundler/source/rubygems.rb
martinemde: I figured out a way to preserve most of the old code while still avoiding extra concats. Reverted this change.

PR #6963, opened 2023-09-15
Stop allocating the same settings keys repeatedly

segiddins:
Running `bundle update --bundler` on a rails app locally:

```
==> memprof.after.txt <==
Total allocated: 301.90 kB (3794 objects)
Total retained:  73.24 kB (698 objects)

==> memprof.before.txt <==
Total allocated: 14.47 MB (196378 objects)
Total retained:  25.93 kB (202 objects)
```

So for a slight increase in retained memory (all keys are now retained),
we go from about 200k allocations in the settings file to under 4k

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -97,6 +97,8 @@ module Bundler
 
       @global_config   = load_config(global_config_file)
       @temporary       = {}
+
+      @key_cache = {}
     end
 
     def [](name)
@@ -312,7 +314,7 @@ module Bundler
     end
 
     def key_for(key)
-      self.class.key_for(key)
+      @key_cache[key] ||= self.class.key_for(key)
     end
 
     private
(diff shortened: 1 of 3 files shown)


segiddins:
@technicalpickles you might like this one too :)

martinemde (approved):
🚀

indirect:
Nice work!

bundler/lib/bundler/settings.rb
```diff
@@ -536,5 +537,34 @@ def self.normalize_uri(uri)
...
     end
+
+    # This is a hot method, so avoid respond_to? checks on every invocation
+    if :read.respond_to?(:name)
+      def self.key_to_s(key)
+        case key
+        when String
+          key
+        when Symbol
+          key.name
```
technicalpickles, line 548: Oh interesting! So this does not end up allocating any extra strings.

technicalpickles (commented):
Love to see it 🚀 

I did come across part of this, ie the `key_for` and those `key_to_s`, but didn't have a good solution for it yet.

bundler/lib/bundler/settings.rb
segiddins: yup!

PR #6964, opened 2023-09-15
Add Bundler::Plugin.loaded? helper

ccutrer:

I have a bundler plugin that adds to the bundler DSL. Right now users of the plugin must do something like `return unless defined?(...)` to see if the plugin has loaded. A simpler, more direct helper was desired.

Add a Plugin.loaded? helper. Use it internally, also making Plugin.load_plugin idempotent (many plugins I've looked at have a tendency to call that method directly, and use a `unless defined?(...)` guard.

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-require "set"
-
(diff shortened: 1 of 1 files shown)


duckinator:
This is a very nice improvement, thank you for the PR!

PR #6965, opened 2023-09-15
Index#first_result last_result empty? any? to short circuit queries

martinemde:

Index#search constructs arrays of all results before returning. Sometimes the return value is reduced to `search(query).empty?` which wastes the time constructing the results.

Add specialized methods for queries that only need to know empty/any or fetch the first or last result.


martinemde:
Fun to write. Maybe it will be useful someday 😆

PR #6970, opened 2023-09-18
Optimize allocations in Gem::Version

segiddins:
From running in a random rails app I have locally, here are the changes

1) for `bundle lock --update --bundler` (forcing Bundler to go through
dependency resolution)

```
==> memprof.after.txt <==
Total allocated: 2.98 MB (48307 objects)
Total retained:  1.21 MB (16507 objects)

==> memprof.before.txt <==
Total allocated: 12.62 MB (198506 objects)
Total retained:  1.30 MB (23133 objects)
```

2) for `bin/rails runner true` (essentially only bundler/setup)

```
==> memprof.after.txt <==
Total allocated: 59.50 kB (1017 objects)
Total retained:  25.08 kB (362 objects)

==> memprof.before.txt <==
Total allocated: 561.82 kB (8575 objects)
Total retained:  27.28 kB (513 objects)
```

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -162,7 +162,7 @@ class Gem::Version
   # A string representation of this Version.
 
   def version
-    @version.dup
+    @version
   end
 
   alias_method :to_s, :version
@@ -173,7 +173,7 @@ class Gem::Version
   def self.correct?(version)
     nil_versions_are_discouraged! if version.nil?
 
-    !!(version.to_s =~ ANCHORED_VERSION_PATTERN)
+    ANCHORED_VERSION_PATTERN.match?(version.to_s)
   end
 
   ##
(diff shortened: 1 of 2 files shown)


lib/rubygems/version.rb
```diff
@@ -381,10 +389,26 @@ def <=>(other)
   end
 
   def canonical_segments
-    @canonical_segments ||=
-      _split_segments.map! do |segments|
```
martinemde, line 385: This was previously mutated. Is it ok to no longer mutate it?

martinemde (approved):
I like storing a frozen version better than trying to protect against it getting modified. I had a few questions, though none are blockers if it's ready to merge.

lib/rubygems/version.rb
```diff
@@ -395,17 +419,13 @@ def freeze
...
-  end
-
   def _segments
     # segments is lazy so it can pick up version values that come from
     # old marshaled versions, which don't go through marshal_load.
     # since this version object is cached in @@all, its @segments should be frozen
 
-    @segments ||= @version.scan(/[0-9]+|[a-z]+/i).map do |s|
-      /^\d+$/.match?(s) ? s.to_i : s
+    @segments ||= @version.scan(/[0-9]+|[a-z]+/i).map! do |s|
```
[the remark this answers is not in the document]
martinemde: If you're curious, this code works but it's less memory efficient. `scan` creates extra arrays from captures.

```ruby
@version.scan(/([0-9]+)|([a-z]+)/i).map! { |i, s| i ? i.to_i : -s }
```

Resolving this comment.

lib/rubygems/version.rb
deivid-rodriguez: There was an issue with one gem relying being able to mutate this array. See https://github.com/rubygems/rubygems/pull/4247. I think that issue shows that it's best to keep this frozen to avoid those kind of issues, but also shows this could potentially break something.

I think it's ok after but just wanted to mention it!

martinemde: Wow. That's a subtle interaction. It reinforces the choice to make these frozen. Thanks for replying @deivid-rodriguez.

PR #6971, opened 2023-09-18
Reduce allocations when parsing compact index

segiddins:
This still allocates a ton (a string for each line, plus a bunch of
splits into arrays), but it helps a bit when Bundler has to go through
dependency resolution.

```
==> memprof.after.txt <==
Total allocated: 194.14 MB (2317172 objects)
Total retained:  60.81 MB (593164 objects)

==> memprof.before.txt <==
Total allocated: 211.97 MB (2404890 objects)
Total retained:  62.85 MB (640342 objects)
```

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -95,7 +95,12 @@ module Bundler
         # because we need to preserve \n line endings on windows when calculating
         # the checksum
         SharedHelpers.filesystem_access(path, :read) do
-          SharedHelpers.digest(:MD5).hexdigest(File.read(path))
+          File.open(path, "rb") do |f|
+            digest = SharedHelpers.digest(:MD5).new
+            buf = String.new(:capacity => 16_384, :encoding => Encoding::BINARY)
+            digest << buf while f.read(16_384, buf)
+            digest.hexdigest
+          end
         end
       end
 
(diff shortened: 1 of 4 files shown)


martinemde (approved):

martinemde:
I fixed the linting errors so this can be merged.

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -95,7 +95,12 @@ def checksum_for_file(path)
         # because we need to preserve \n line endings on windows when calculating
         # the checksum
         SharedHelpers.filesystem_access(path, :read) do
-          SharedHelpers.digest(:MD5).hexdigest(File.read(path))
+          File.open(path, "rb") do |f|
```
simi, line 98: What about to move this into shared helper directly and just call it here? There is no need to complicate this class with logic for "streamed" MD5 IMHO.

martinemde: That makes sense to me.

...I wonder if we could use [Gem::Package::DigestIO](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/package/digest_io.rb) here to calculate the checksum as we write the file rather than performing an additional step. It would replace `copy_file` so might cancel out the benefit.

martinemde: @simi @segiddins I got a DigestIO based solution working over at #7039.

martinemde: I think since it's only used here, it's better if we leave it. There's nothing else doing this pattern in bundler that I could find.

martinemde:
Based on testing in #7039, this is a perfectly fine solution that shaves about 30MB of memory off a larger bundle lock run. We can follow up if there's additional improvements to be made.

PR #6974, opened 2023-09-19
Freeze more strings in generated gemspecs

segiddins:
Specifically, this will have frozen string literals for:
- Gem platform tuple entries
- Gem::Version strings
- Gem::Specification#installed_by_version
- Dependency requirement strings

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2362,12 +2362,12 @@ class Gem::Specification < Gem::BasicSpecification
     when Hash               then
       seg = obj.keys.sort.map {|k| "#{k.to_s.dump} => #{obj[k].to_s.dump}" }
       "{ #{seg.join(", ")} }"
-    when Gem::Version       then obj.to_s.dump
+    when Gem::Version       then ruby_code(obj.to_s)
     when DateLike           then obj.strftime("%Y-%m-%d").dump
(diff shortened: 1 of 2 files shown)


martinemde (approved):

martinemde:
Will this change the checksums of the gemspecs or is it outside of that process? I think we call to_ruby when storing the plain text gemspec for gem content storage?

segiddins:
> Will this change the checksums of the gemspecs or is it outside of that process

it's outside of that process, since the gemspec we upload to `quick/` is the marshaled gemspec (and also those don't get regenerated, so changing what _would_ be generated is kosher, and we've done that in the past)

segiddins:
> Ok cool. Now I'm wondering which gemspec version should be [uploaded with gem content](https://github.com/rubygems/rubygems.org/blob/master/app/models/version_manifest.rb#L104).

IMO it should be the yaml gemspec that's inside the .gem, but that's mostly since it's what'd be most helpful to me from a rubygems.org security perspective

martinemde:
> IMO it should be the yaml gemspec that's inside the .gem, but that's mostly since it's what'd be most helpful to me from a rubygems.org security perspective

That seems most reasonable to me. I can't imagine what else we'd want besides the most useful thing 😆

PR #6975, opened 2023-09-19
Fewer allocations in gem installation

segiddins:
For now, on a small rails app I have hanging around:

```
==> memprof.after.txt <==
Total allocated: 872.51 MB (465330 objects)
Total retained:  40.48 kB (326 objects)

==> memprof.before.txt <==
Total allocated: 890.79 MB (1494026 objects)
Total retained:  40.40 kB (328 objects)
```

Not a huge difference in memory usage, but it's a drastic improvement
in total number of allocations.

Additionally, this will pay huge dividends once
https://github.com/ruby/zlib/pull/61 is merged, as it will allow us to
completely avoid allocations in the repeated calls to readpartial,
which currently accounts for most of the memory usage shown above.


lib/rubygems/package.rb
```diff
@@ -447,12 +447,12 @@ def extract_tar_gz(io, destination_dir, pattern = "*") # :nodoc:
...
 
         unless directories.include?(mkdir)
-          FileUtils.mkdir_p mkdir, **mkdir_options
+          FileUtils.mkdir_p mkdir, mode: dir_mode ? 0o755 : (entry.header.mode if entry.directory?)
           directories << mkdir
         end
 
         if entry.file?
-          File.open(destination, "wb") {|out| out.write entry.read }
+          File.open(destination, "wb") {|out| IO.copy_stream entry, out }
```
martinemde, line 455: Yep, truffleruby failing in your tests too. I haven't been able to extract a distilled case yet, but @eregon could maybe take a look anyway? I think the issue is that copy_stream is not reading and writing byte for byte identically. 

There's layers and layers here, the src entry is a TarReader wrapping a zlib reader, wrapping a TarReader. This change alone causes the bug, if I'm not mistaken.

You may have to revert this part until it's fixed in truffleruby, or found not to be broken?

lib/rubygems/package.rb
```diff
@@ -708,7 +708,11 @@ def verify_files(gem)
 
   def verify_gz(entry) # :nodoc:
     Zlib::GzipReader.wrap entry do |gzio|
-      gzio.read 16_384 until gzio.eof? # gzip checksum verification
+      buf = String.new(capacity: 16_384, encoding: Encoding::BINARY)
+      begin
+        gzio.readpartial(16_384, buf) until gzio.eof? # gzip checksum verification
```
martinemde, line 713: `gzio` read partial returns arbitrarily different length strings on some gzips. Check the pathological gem that I mentioned in the zlib bug. It extracts validly when using read but not readpartial. I don't know how common this problem is.

lib/rubygems/package/tar_header.rb
```diff
@@ -127,7 +127,8 @@ def self.from(stream)
   end
 
   def self.strict_oct(str)
-    return str.strip.oct if /\A[0-7]*\z/.match?(str.strip)
+    str.strip!
```
martinemde, line 130: I don't feel strongly about this, but I'd rather see this public class method replaced with a strict_octal! method that modifies its argument than simply changed to do so. Better to totally break someone using this method by renaming it than subtly break things by mutating the argument.

lib/rubygems/package/tar_header.rb
```diff
@@ -243,4 +244,12 @@ def header(checksum = @checksum)
...
     format("%0#{len}o", num)
   end
+
+  eval <<-RUBY, binding, __FILE__, __LINE__ + 1
+# frozen_string_literal: true
+
+  def copy_vals(vals)
+    #{FIELDS.map {|name| "@#{name} = vals[:#{name}]" }.join("; ")}
+  end
+  RUBY
```
martinemde, line 254: This definitely needs a comment about why. This is not your grandma's self documenting code. 👵🏻

martinemde: Or, alternatively, why convert to instance variables at all? Why not store `vals` directly and iterate FIELDS one time to create attr readers that just lookup the key in the vals hash. vals is mutable and is already allocated.

test/rubygems/test_gem_package_tar_reader_entry.rb
```diff
@@ -185,6 +185,13 @@ def test_readpartial
     assert_equal @contents[0...100], @entry.readpartial(100)
   end
 
+  def test_readpartial_outbuf
+    buffer = String.new(capacity: 100, encoding: Encoding::BINARY)
+    partial = @entry.readpartial(100, buffer)
+    assert_equal @contents[0...100], partial
+    assert_same buffer, partial
```
martinemde, line 192: Good test. I wonder if it's worth doing one using the gzip setup lower in this test file. I'm not sure it will catch the failure case though, since I think you need a few files so that it can overread one of them.

martinemde:
Blocked until code is updated to not use `readpartial` or Zlib is updated.

lib/rubygems/package.rb
eregon: It seems likely to be https://github.com/oracle/truffleruby/issues/3280 which is now fixed in dev builds

segiddins:
My git scripts messed up, will clean up tomorrow

lib/rubygems/package.rb
segiddins: @martinemde thoughts on introducing a utility method that is `IO.copy_stream` on all rubies except truffleruby before 23.1.2? Could go into https://github.com/rubygems/rubygems/pull/6958

martinemde: That's a great idea. Even fixed it's still broken on older trufflerubies, even if people don't usually update rubygems separately.

martinemde: I can add it to #6958.

lib/rubygems/package/tar_header.rb
martinemde: After coming back to this 3 months later, I think I'd prefer we just write out all the assignments just like in `==`. The eval feels really complex compared to simple repetition.

lib/rubygems/package/tar_header.rb
```diff
@@ -161,9 +163,7 @@ def initialize(vals)
     vals[:devmajor] ||= 0
     vals[:devminor] ||= 0
```
martinemde, line 164: Instead of the lines like this, just write directly to the instance variable like:

```ruby
@devminor = val[:devminor] || 0
```

Then you avoid strings, loops, method calls and a hash assingment.

martinemde (approved):
Awesome!

PR #6976, opened 2023-09-19
Reduce allocations when parsing lockfile

segiddins:
```
==> memprof.after.txt <==
Total allocated: 673.08 kB (7644 objects)
Total retained:  107.35 kB (1018 objects)

==> memprof.before.txt <==
Total allocated: 739.12 kB (9140 objects)
Total retained:  138.61 kB (1695 objects)
```

Savings will scale by the number of lines in the lockfile

### DIFF
--- a/bundler/lib/bundler/lockfile_parser.rb
+++ b/bundler/lib/bundler/lockfile_parser.rb
@@ -23,13 +23,15 @@ module Bundler
       Gem::Version.create("1.13") => [PLUGIN].freeze,
     }.freeze
 
-    KNOWN_SECTIONS = SECTIONS_BY_VERSION_INTRODUCED.values.flatten.freeze
+    KNOWN_SECTIONS = SECTIONS_BY_VERSION_INTRODUCED.values.flatten!.freeze
 
     ENVIRONMENT_VERSION_SECTIONS = [BUNDLED, RUBY].freeze
     deprecate_constant(:ENVIRONMENT_VERSION_SECTIONS)
 
     def self.sections_in_lockfile(lockfile_contents)
-      lockfile_contents.scan(/^\w[\w ]*$/).uniq
+      sections = lockfile_contents.scan(/^\w[\w ]*$/)
+      sections.uniq!
+      sections
     end
 
     def self.unknown_sections_in_lockfile(lockfile_contents)
@@ -38,7 +40,7 @@ module Bundler
 
     def self.sections_to_ignore(base_version = nil)
       base_version &&= base_version.release
-      base_version ||= Gem::Version.create("1.0".dup)
+      base_version ||= Gem::Version.create("1.0")
       attributes = []
       SECTIONS_BY_VERSION_INTRODUCED.each do |version, introduced|
         next if version <= base_version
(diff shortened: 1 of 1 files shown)


bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -84,10 +86,10 @@ def initialize(lockfile)
         elsif /^[^\s]/.match?(line)
           @state = nil
         elsif @state
-          send("parse_#{@state}", line)
+          send(PARSE_METHODS.fetch(@state), line)
```
martinemde, line 89: 👏

This is better code, but what do you think of scrapping `@state` altogether and just storing the `@parser` method name (or keep the name `@state` bit store the method name)?

Personally, I'd save the extra obfuscation and fetching and directly store `@parser = :parse_ruby` etc.

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -149,11 +160,11 @@ def parse_dependency(line)
       return unless line =~ NAME_VERSION
       spaces = $1
       return unless spaces.size == 2
-      name = $2
+      name = -$2
```
martinemde, line 163: Cute. I'm surprised this isn't used more often, enough that it's common practice. First I've seen it and had to look it up.

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -177,11 +188,13 @@ def parse_dependency(line)
...
       return unless line =~ NAME_VERSION
       spaces = $1
-      name = $2
+      name = -$2
       version = $3
-      platform = $4
 
       if spaces.size == 4
+        # no platform for a dep line
+        platform = $4
```
martinemde, line 196: So it doesn't allocate until you call the global, or just better organization?

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -177,11 +188,13 @@ def parse_dependency(line)
     def parse_spec(line)
       return unless line =~ NAME_VERSION
       spaces = $1
-      name = $2
+      name = -$2
       version = $3
-      platform = $4
 
       if spaces.size == 4
+        # no platform for a dep line
```
martinemde, line 195: This comment is confusing because it's saying "no platform" right where you set platform. Maybe "only load platform for non-dependency line"

martinemde (approved):
Feel free to merge anyway if you don't think the `@state` change is worth it.

bundler/lib/bundler/lockfile_parser.rb
indirect: I don't think I've seen this before, but I failed when I tried to look it up. What is it?

martinemde: Unary minus on String. It freezes the string. Unary + thaws.

indirect: Huh, how about that. I guess I should have checked the API docs! I did finally find a writeup: https://metaredux.com/posts/2019/05/10/weird-ruby-positive-and-negative-strings.html

bundler/lib/bundler/lockfile_parser.rb
segiddins: yeah that's a good idea

bundler/lib/bundler/lockfile_parser.rb
segiddins: `@-` is an alias for `#dedup`, which not only returns a frozen string, but will deduplicate it, leading to reduced retained memory usage

nobu: FYI: The unary operator is `-@`.

PR #6980, opened 2023-09-20
Fix `bundle install` when older revisions of git source

### LINKED ISSUE #6929: `bundle install` fails with git error `Could not parse object`

When running `bundle install`, it may fail with the error `fatal: Could not parse object 'commit-sha1'.`.

Our CI uses a cache to install dependencies. ref. https://circleci.com/docs/caching/
And we are using dependabot to update gems, but if we use the cache of the updated gem with git sources for another gem update PR, we may get this error.

Yes

Run the following script.

```rb
require 'tmpdir'

Dir.mktmpdir do |dir|
  commits = []

  foo_path = "#{dir}/foo"
  Dir.mkdir(foo_path)
  Dir.chdir(foo_path) do
    system 'git init'
    File.write("#{foo_path}/foo.gemspec", <<~GEMSPEC)
      Gem::Specification.new do |spec|
        spec.name = "foo"
        spec.version = "0.0.1"
        spec.authors = ["sample"]
        spec.summary = "Foo"
      end
    GEMSPEC
    system 'git add foo.gemspec'
    system 'git commit -m "First commit"'
    commits << `git rev-parse HEAD`.chomp
  end

  app_path = "#{dir}/app"
  Dir.mkdir(app_path)
  File.write("#{app_path}/Gemfile", <<~GEMFILE)
    source 'https://rubygems.org'
    gem 'foo', git: '#{foo_path}', branch: 'main'
  GEMFILE
  Dir.chdir(app_path) do
    system 'bundle config --local path vendor/bundle'
    system 'bundle config --local clean true'
    system 'bundle install'
  end

  Dir.chdir(foo_path) do
    system 'git commit --allow-empty -m "Second commit"'
    commits << `git rev-parse HEAD`.chomp
  end

  Dir.chdir(app_path) do
    system 'bundle update'
  end

  File.write(
    "#{app_path}/Gemfile.lock",
    File.read("#{app_path}/Gemfile.lock").gsub(commits[1], commits[0])
  )

  Dir.chdir(app_path) do
    system 'bundle install' # => Git error
  end
end
```

bundle install

I expect `bundle install` to succeed.

### What actually happened?

The following error

```
Git error: command `git reset --hard 41982d4703e4a89eaeaa1c90ae4982ff574e48eb` in directory
/tmp/path/to/app/vendor/bundle/ruby/3.1.0/bundler/gems/foo-41982d4703e4 has failed.
fatal: Could not parse object '41982d4703e4a89eaeaa1c90ae4982ff574e48eb'.
```

## Environment

```
Bundler       2.4.19
  Platforms   ruby, arm64-darwin-22
Ruby          3.1.4p223 (2023-03-30 revision 957bb7cb81995f26c671afce0ee50a5c660e540e) [arm64-darwin-22]
  Full Path   /Users/krororo/.rbenv/versions/3.1.4/bin/ruby
  Config Dir  /Users/krororo/.rbenv/versions/3.1.4/etc
RubyGems      3.4.19
  Gem Home    /Users/krororo/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0
  Gem Path    /Users/krororo/.gem/ruby/3.1.0:/Users/krororo/.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0
  User Home   /Users/krororo
  User Path   /Users/krororo/.gem/ruby/3.1.0
  Bin Dir     /Users/krororo/.rbenv/versions/3.1.4/bin
Tools
  Git         2.42.0
  RVM         not installed
  rbenv       rbenv 1.2.0
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-08-17
Git SHA           86f98098e3
Released Version  true
```

## Bundler settings

```
clean
  Set for your local app (/tmp/path/to/app/.bundle/config): true
path
  Set for your local app (/tmp/path/to/app/.bundle/config): "vendor/bundle"
```

krororo:
Fixes #6929.

`bundle install` may fail if an old revision is specified for a gem of git source.
At that time, the following error occurs.

```
Git error: command `git reset --hard 41982d4703e4a89eaeaa1c90ae4982ff574e48eb` in directory
/tmp/path/to/app/vendor/bundle/ruby/3.1.0/bundler/gems/foo-41982d4703e4 has failed.
fatal: Could not parse object '41982d4703e4a89eaeaa1c90ae4982ff574e48eb'.
```

This implementation is to execute `fetch` after copying to the installation directory. Because the target commit may exist in the repository in cache, but not in the destination repository.

### DIFF
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -130,7 +130,8 @@ module Bundler
             end
           end
 
-          git "fetch", "--force", "--quiet", *extra_fetch_args, :dir => destination if @commit_ref
+          ref = @commit_ref || (full_sha_revision? && @revision)
+          git "fetch", "--force", "--quiet", *extra_fetch_args(ref), :dir => destination if ref
 
           git "reset", "--hard", @revision, :dir => destination
 
@@ -248,6 +249,10 @@ module Bundler
 
         def pinned_to_full_sha?
           ref =~ /\A\h{40}\z/
+        end
+
+        def full_sha_revision?
+          @revision.match?(/\A\h{40}\z/)
         end
 
         def git_null(*command, dir: nil)
@@ -411,9 +416,9 @@ module Bundler
           ["--depth", depth.to_s]
         end
 
-        def extra_fetch_args
+        def extra_fetch_args(ref)
           extra_args = [path.to_s, *depth_args]
-          extra_args.push(@commit_ref)
+          extra_args.push(ref)
           extra_args
         end
 
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -130,7 +130,7 @@ module Bundler
             end
           end
 
-          ref = @commit_ref || (full_sha_revision? && @revision)
+          ref = @commit_ref || (locked_to_full_sha? && @revision)
           git "fetch", "--force", "--quiet", *extra_fetch_args(ref), :dir => destination if ref
 
           git "reset", "--hard", @revision, :dir => destination
@@ -248,11 +248,15 @@ module Bundler
         end
 
         def pinned_to_full_sha?
-          ref =~ /\A\h{40}\z/
+          full_sha_revision?(ref)
         end
 
-        def full_sha_revision?
-          @revision.match?(/\A\h{40}\z/)
+        def locked_to_full_sha?
+          full_sha_revision?(@revision)
+        end
+
+        def full_sha_revision?(ref)
+          ref&.match?(/\A\h{40}\z/)
         end
 
         def git_null(*command, dir: nil)
--- a/bundler/spec/install/git_spec.rb
+++ b/bundler/spec/install/git_spec.rb
@@ -170,5 +170,36 @@ RSpec.describe "bundle install" do
 
       expect(out).to include("Bundle complete!")
     end
+
+    it "allows older revisions of git source when clean true" do
+      build_git "foo", "1.0", :path => lib_path("foo")
+      rev = revision_for(lib_path("foo"))
+
+      bundle "config set path vendor/bundle"
+      bundle "config set clean true"
+      install_gemfile <<-G, :verbose => true
+        source "#{file_uri_for(gem_repo1)}"
+        gem "foo", :git => "#{file_uri_for(lib_path("foo"))}"
+      G
+
+      expect(out).to include("Using foo 1.0 from #{file_uri_for(lib_path("foo"))} (at main@#{rev[0..6]})")
+      expect(the_bundle).to include_gems "foo 1.0", :source => "git@#{lib_path("foo")}"
+
+      old_lockfile = lockfile
+
+      update_git "foo", "2.0", :path => lib_path("foo"), :gemspec => true
+      rev2 = revision_for(lib_path("foo"))
+
+      bundle :update, :all => true, :verbose => true
+      expect(out).to include("Using foo 2.0 (was 1.0) from #{file_uri_for(lib_path("foo"))} (at main@#{rev2[0..6]})")
+      expect(out).to include("Removing foo (#{rev[0..11]})")
+      expect(the_bundle).to include_gems "foo 2.0", :source => "git@#{lib_path("foo")}"
+
+      lockfile(old_lockfile)
+
+      bundle :install, :verbose => true
+      expect(out).to include("Using foo 1.0 from #{file_uri_for(lib_path("foo"))} (at main@#{rev[0..6]})")
+      expect(the_bundle).to include_gems "foo 1.0", :source => "git@#{lib_path("foo")}"
+    end
   end
 end


bundler/lib/bundler/source/git/git_proxy.rb
```diff
@@ -250,6 +251,10 @@ def pinned_to_full_sha?
           ref =~ /\A\h{40}\z/
         end
 
+        def full_sha_revision?
+          @revision.match?(/\A\h{40}\z/)
+        end
```
deivid-rodriguez, line 256: This logic seems mainly duplicated with `pinned_to_full_sha?`. I would refactor this logic to a private method that takes a `ref` and does this regexp match, and then rename this to `locked_to_full_sha?`, and make both `pinned_to_full_sha?` and `locked_to_full_sha?` use the private method.

deivid-rodriguez (commented):
Looks great, just a small comment!

bundler/lib/bundler/source/git/git_proxy.rb
krororo: Thank you for the review.

I fixed f0d8255ebd89f2dd47a53fb400c5ed6a61f72c28

deivid-rodriguez: Thank you!

igorpeshansky:
This seems to break with really old versions of `git`, e.g.:
```
$ git --version
git version 2.13.7
$ grep omnibus-software Gemfile
gem 'omnibus-software', :github => 'chef/omnibus-software'
$ bundle install
Git error: command `git fetch --force --quiet /tmp/path/to/app/vendor/bundle/ruby/3.1.0/cache/bundler/git/omnibus-software-338444b7bc50110d52bb35857ad8f946482db0ef --depth 1 4fef36786ead96e7f2f23e6f01f09efb71894992` in directory /tmp/path/to/app/vendor/bundle/ruby/3.1.0/bundler/gems/omnibus-software-4fef36786ead has failed.
error: Server does not allow request for unadvertised object 4fef36786ead96e7f2f23e6f01f09efb71894992
```
One fix is to set `git config uploadpack.allowAnySHA1InWant true` in bundler's git cache repositories, but I'm not sure how to make bundler do that… My attempts to add
```ruby
          git "config", "uploadpack.allowAnySHA1InWant", "true", :dir => destination if ref
```
[here](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/source/git/git_proxy.rb#L133) didn't seem to actually change the cache repository config. If someone can point me at the right place in the code, I can send a PR.

## Environment

```
Bundler             2.4.21
  Platforms         ruby, x86_64-linux
Ruby                3.1.2p20 (2022-04-12 revision 4491bb740a9506d76391ac44bb2fe6e483fec952) [x86_64-linux]
  Full Path         /usr/local/rvm/rubies/ruby-3.1.2/bin/ruby
  Config Dir        /usr/local/rvm/rubies/ruby-3.1.2/etc
RubyGems            3.3.7
  Gem Home          /usr/local/rvm/gems/ruby-3.1.2
  Gem Path          /usr/local/rvm/gems/ruby-3.1.2:/usr/local/rvm/rubies/ruby-3.1.2/lib/ruby/gems/3.1.0
  User Home         …
  User Path         …/.local/share/gem/ruby/3.1.0
  Bin Dir           /usr/local/rvm/gems/ruby-3.1.2/bin
Tools               
  Git               2.13.7
  RVM               1.29.12 (latest)
  rbenv             not installed
  chruby            not installed
  rubygems-bundler  (1.4.5)
```

## Bundler Build Metadata

```
Built At          2023-10-17
Git SHA           d10b46bd15
Released Version  true
```

deivid-rodriguez:
Hei! Interesting!

So this used to work for you and only broke after the latest release including this PR?

My understanding is that `uploadpack.allowAnySHA1InWant` is a server setting and most git servers have it set. Github should definitely allow it. So this would seem unrelated to `git` version to me. Did you verify it works with newer `git`?

I googled the error and found [this thread](https://stackoverflow.com/questions/51001919/how-to-solve-git-error-server-does-not-allow-request-for-unadvertised-object-3a) which points to submodule issues? Is a git submodule involved here?

bkabrda:
:wave: I just wanted to report we're seeing the similar happen when using https://github.com/DataDog/omnibus-software as a git source in our Gemfile (no git submoduled involved). We see this with git 2.7 and 2.10, but not with 2.40.

deivid-rodriguez:
Thank you, since that's an open source repo, it should be easily reproducible, right? What do I need to do to see the issue for myself?

deivid-rodriguez:
Since this sounds fixed in `git` a fix for us would be a matter of reverting this change for older git versions. But it'd be nice to bisect which particular `git` version started behaving just fine.

igorpeshansky:
I'm in the same boat as @bkabrda — no submodules involved, also erroring out on an `omnibus-software` clone. FWIW, I was looking at https://stackoverflow.com/a/61492323, referenced from the thread you posted. The setting needs to be on the local clone of the repo, which is in the cache (`/tmp/path/to/app/vendor/bundle/ruby/3.1.0/cache/bundler/git/<gem>-<hash>`). I've verified this by manually setting `uploadpack.allowAnySHA1InWant` in the cache directory, which allowed `bundle install` to proceed. I just couldn't figure out how to do this automatically. I'll respond soon with repro instructions.

Unfortunately, upgrading git is not an option for us — we build on a variety of older platforms in Docker containers, and we're stuck with the system git.

igorpeshansky:
### Repro instructions

Save the following files in a new directory: \
`6980.Dockerfile`:
```Dockerfile
FROM opensuse/leap:42.3 as base
RUN zypper -n install git curl which tar
RUN mv /var/lib/ca-certificates/pem/DST_Root_CA_X3.pem /etc/pki/trust/blacklist/ \
 && update-ca-certificates
RUN gpg2 --keyserver hkp://keyserver.ubuntu.com --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
RUN curl -sSL -o get-rvm-io.sh https://get.rvm.io \
 && /bin/bash get-rvm-io.sh stable
RUN ["/bin/bash", "-lc", "rvm requirements && rvm install 3.1.2"]

ADD "https://www.random.org/cgi-bin/randbyte?nbytes=10&format=h" skipcache

ARG BUNDLER_VERSION=2.4.21
RUN ["/bin/bash", "-lc", "gem install bundler --no-document --version ${BUNDLER_VERSION}"]
COPY 6980.rb /6980.rb
RUN ["/bin/bash", "-lc", "rvm use 3.1.2 && bundle --version && ruby /6980.rb"]
```
`6980.rb`:
```ruby
require 'tmpdir'

Dir.mktmpdir do |dir|
  app_path = "#{dir}/app"
  Dir.mkdir(app_path)
  File.write("#{app_path}/Gemfile", <<~GEMFILE)
    source 'https://rubygems.org'
    gem 'omnibus-software', :github => 'chef/omnibus-software'
  GEMFILE
  Dir.chdir(app_path) do
    system 'bundle config --local path vendor/bundle'
    system 'bundle config --local clean true'
  end

  Dir.chdir(app_path) do
    system 'bundle install' # => Git error
  end
end
```
Invoke as follows:
```
docker build -f 6980.Dockerfile --build-arg BUNDLER_VERSION=2.4.21 .
```
(or just `docker build -f 6980.Dockerfile .`).
Observe the git error in the output:
```
Git error: command `git fetch --force --quiet /tmp/d20231019-204-elytvg/app/vendor/bundle/ruby/3.1.0/cache/bundler/git/omnibus-software-338444b7bc50110d52bb35857ad8f946482db0ef --depth 1 4fef36786ead96e7f2f23e6f01f09efb71894992` in directory /tmp/d20231019-204-elytvg/app/vendor/bundle/ruby/3.1.0/bundler/gems/omnibus-software-4fef36786ead has failed.
error: Server does not allow request for unadvertised object 4fef36786ead96e7f2f23e6f01f09efb71894992
```
Invoke to see it succeed with `bundler` 2.4.20:
```
docker build -f 6980.Dockerfile --build-arg BUNDLER_VERSION=2.4.20 .
```
@deivid-rodriguez

deivid-rodriguez:
Nice investigation and repro @igorpeshansky :heart_eyes:

I think you are basically on the right track but you were running the git command in the wrong folder, it should be:

```
git "config", "uploadpack.allowAnySHA1InWant", "true", :dir => path.to_s
```

Besides that, this setting is not [the default](https://git-scm.com/docs/git-config#Documentation/git-config.txt-uploadpackallowAnySHA1InWant) in git, so I'd feel more confortable keeping it like that on git versions without this bug. So I'd go with something like:

```diff
diff --git a/bundler/lib/bundler/source/git/git_proxy.rb b/bundler/lib/bundler/source/git/git_proxy.rb
index 93625145c..0b507abc7 100644
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -131,7 +131,11 @@ def copy_to(destination, submodules = false)
           end
 
           ref = @commit_ref || (locked_to_full_sha? && @revision)
-          git "fetch", "--force", "--quiet", *extra_fetch_args(ref), :dir => destination if ref
+          if ref
+            git "config", "uploadpack.allowAnySHA1InWant", "true", :dir => path.to_s if @commit_ref.nil? && needs_allow_any_sha1_in_want?
+
+            git "fetch", "--force", "--quiet", *extra_fetch_args(ref), :dir => destination
+          end
 
           git "reset", "--hard", @revision, :dir => destination
 
@@ -434,6 +438,10 @@ def supports_minus_c?
           @supports_minus_c ||= Gem::Version.new(version) >= Gem::Version.new("1.8.5")
         end
 
+        def needs_allow_any_sha1_in_want?
+          @needs_allow_any_sha1_in_want ||= Gem::Version.new(version) <= Gem::Version.new("2.13.7")
+        end
+
         def supports_fetching_unreachable_refs?
           @supports_fetching_unreachable_refs ||= Gem::Version.new(version) >= Gem::Version.new("2.5.0")
         end
```

I'm not sure which git version fixed this bug, but I guess we workaround it for everyone who has noticed it here and move on, that's why I set 2.13.7. We could also do a proper bisection by manually building from source different git versions, or try to rescue the error message and retry configuring that setting right before, but feels overkill to me.

PR #6983, opened 2023-09-21
Aggressively optimize allocations in SafeMarshal

segiddins:
Reduces allocations in a bundle install --full-index by an order of magnitude

Main wins are (a) getting rid of exessive string allocations for exception message stack

(b) Avoiding hash allocations caused by kwargs for #initialize

(c) avoid using unpack to do bit math, its easy enough to do by hand

(d) special case the most common elements so they can be read without an allocation

(e) avoid string allocations every time a symbol->string lookup is done by using symbol#name

### DIFF
--- a/lib/rubygems/safe_marshal/reader.rb
+++ b/lib/rubygems/safe_marshal/reader.rb
@@ -49,19 +49,19 @@ module Gem
         when 0x00
           0
         when 0x01
-          @io.read(1).unpack1("C")
+          read_byte
         when 0x02
-          @io.read(2).unpack1("S<")
+          read_byte | (read_byte << 8)
         when 0x03
-          (@io.read(3) + "\0").unpack1("L<")
+          read_byte | (read_byte << 8) | (read_byte << 16)
         when 0x04
-          @io.read(4).unpack1("L<")
+          read_byte | (read_byte << 8) | (read_byte << 16) | (read_byte << 24)
         when 0xFC
-          @io.read(4).unpack1("L<") | -0x100000000
+          read_byte | (read_byte << 8) | (read_byte << 16) | (read_byte << 24) | -0x100000000
         when 0xFD
-          (@io.read(3) + "\0").unpack1("L<") | -0x1000000
+          read_byte | (read_byte << 8) | (read_byte << 16) | -0x1000000
         when 0xFE
-          @io.read(2).unpack1("s<") | -0x10000
+          read_byte | (read_byte << 8) | -0x10000
         when 0xFF
           read_byte | -0x100
         else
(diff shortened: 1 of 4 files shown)


indirect (approved):
Awesome. Also *sigh* at kwargs.

lib/rubygems/safe_marshal/visitors/to_ruby.rb
```diff
@@ -23,39 +24,61 @@ def inspect # :nodoc:
...
-          visit(e)
-        end)
+        array = register_object([])
+
+        elements = a.elements
+        size = elements.size
+        idx = 0
+        # not idiomatic, but there's a huge number of IMEMOs allocated here, so we avoid the block
+        # because this is such a hot path when doing a bundle install with the full index
+        until idx == size
```
martinemde, line 48: Maybe `for idx in (0..size)` maintains the performance while avoiding the bounds check every iteration? Making the range is hopefully cheap.

lib/rubygems/safe_marshal/reader.rb
```diff
@@ -124,67 +144,123 @@ def read_false
...
+        data = read_element
+        Elements::UserMarshal.new(name, data)
+      end
+
+      # profiling bundle install --full-index shows that
+      # offset 6 is by far the most common object link,
+      # so we special case it to avoid allocating a new
+      # object a third of the time.
+      # the following are all the object links that
+      # appear more than 10000 times in my profiling
```
martinemde, line 188: This is interesting. Is it possible that this is "overfitting" to the data as it exists right now and this will change, or is this something intrinsically connected?

lib/rubygems/safe_marshal/reader.rb
```diff
@@ -88,31 +88,51 @@ def read_element
...
+      STRING_E_SYMBOL = Elements::Symbol.new("E").freeze
+      private_constant :STRING_E_SYMBOL
+
       def read_symbol
-        Elements::Symbol.new name: @io.read(read_integer)
+        len = read_integer
+        if len == 1
+          byte = read_byte
+          if byte == 69 # ?E
+            STRING_E_SYMBOL
```
martinemde, line 118: I saw A and at least one other mentioned in tests. Are the others just test cases and E is the only really common one?

martinemde (approved):
"640K is more memory than anyone will ever need."

lib/rubygems/safe_marshal/visitors/to_ruby.rb
segiddins: the range unfortunately requires an allocation

lib/rubygems/safe_marshal/reader.rb
segiddins: `E` is the only one that is allowlisted, but I wanted to be comprehensive in my tests.

`E` is the symbol used for string encoding, so it shows up very often

segiddins:
Hmmm @indirect why did this merge with the jruby tests failing?

segiddins:
https://github.com/rubygems/rubygems/pull/6984 fixes the test

PR #6987, opened 2023-09-22
Run CI on pushes to master again

deivid-rodriguez:

We currently have no easy way to verify if a semantic conflict between two consecutively merged PRs appears.

I think making sure CI is green on pushes to master is reassuring.

My fix is to revert https://github.com/rubygems/rubygems/pull/6369, because we ended up disabling the merge queue feature because it didn't quite work for us.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -2,7 +2,6 @@ name: bundler
 
 on:
   pull_request:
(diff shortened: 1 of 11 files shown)


deivid-rodriguez:
Happy to change this policy if we decide so, but since this change was unintentional, I will merge this.

PR #6988, opened 2023-09-22
Fix turbo tests Windows monkeypatch

deivid-rodriguez:

We haven't been running Bundler specs on Windows since https://github.com/rubygems/rubygems/pull/6739. That PR introduced a monkeypatch for Windows that made tests take a long time (like now) and still report success. So we did not notice that no specs were actually run.

This can be noticed in the last two lines of the current output:

```
Finished in 93 minutes 10 seconds (files took 0 seconds to load)
0 examples, 0 failures
```

My fix is to correct the patch to properly apply our previous solution. We should upstream something but for now I just want the tests running again.

### DIFF
--- a/tool/turbo_tests_windows_patch.rb
+++ b/tool/turbo_tests_windows_patch.rb
@@ -1,42 +1,7 @@
 # frozen_string_literal: true
 
 module TurboTests
-  class JsonRowsFormatter
-    private
-
-    def output_row(obj)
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I just merged #6989, so this should be green now.

hsbt:
Thank you @deivid-rodriguez 🙏

PR #6989, opened 2023-09-22
Allow standalone mode to work on a Windows edge case

deivid-rodriguez:

If a gem is located in a different drive than the Gemfile, standalone mode will fail to generate the `bundler/setup` script, failing with an error like

```
ArgumentError: different prefix: "C:/" and "D:/a/rubygems/rubygems/bundler/tmp/2/bundled_app/bundle/bundler"
  C:/hostedtoolcache/windows/Ruby/3.0.5/x64/lib/ruby/3.0.0/pathname.rb:528:in `relative_path_from'
  D:/a/rubygems/rubygems/bundler/tmp/2/gems/system/gems/bundler-2.4.20/lib/bundler/installer/standalone.rb:58:in `gem_path'
  D:/a/rubygems/rubygems/bundler/tmp/2/gems/system/gems/bundler-2.4.20/lib/bundler/installer/standalone.rb:33:in `block (2 levels) in paths'
  D:/a/rubygems/rubygems/bundler/tmp/2/gems/system/gems/bundler-2.4.20/lib/bundler/installer/standalone.rb:32:in `map'
  D:/a/rubygems/rubygems/bundler/tmp/2/gems/system/gems/bundler-2.4.20/lib/bundler/installer/standalone.rb:32:in `block in paths'
```

I'm fixing this by falling back to using a full path in this case.

This was caught by a failing spec, so I'm not adding new specs.

### DIFF
--- a/bundler/lib/bundler/installer/standalone.rb
+++ b/bundler/lib/bundler/installer/standalone.rb
@@ -55,11 +55,18 @@ module Bundler
       if spec.source.instance_of?(Source::Path) && spec.source.path.absolute?
         full_path
       else
-        Pathname.new(full_path).relative_path_from(Bundler.root.join(bundler_path)).to_s
+        relative_path_from(Bundler.root.join(bundler_path), :to => full_path) || full_path
       end
     rescue TypeError
       error_message = "#{spec.name} #{spec.version} has an invalid gemspec"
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I expect #6988 to make this spec failure surface. But it's already appearing in #6985, since the stable branch doesn't have the issue.

PR #7002, opened 2023-09-26
Avoid duplicates -rbundler/setup in RUBYOPT with Ruby preview

intrip:

When using a Ruby preview version the require path of `bundler/setup` is similar to `-r/opt/ruby3.3.0-preview2/lib/ruby/3.3.0+0/bundler/setup`. The special character `+` in the string makes the Regexp matching fail leading to the addition of duplicate require statements to RUBYOPT (e.g. server reloading).

Example:

```ruby
>> RUBY_DESCRIPTION
=> "ruby 3.3.0preview2 (2023-09-14 master e50fcca9a7) [arm64-darwin22]"
>> ENV["RUBYOPT"]
=> "-r/Users/jacopobeschi/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup"
>> Bundler::SharedHelpers.send(:set_rubyopt)
=> "-r/Users/jacopobeschi/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup -r/Users/jacopobeschi/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup"
>> ENV["RUBYOPT"]
=> "-r/Users/jacopobeschi/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup -r/Users/jacopobeschi/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup"
```

Escaping the characters in the string ensure a correct match with all the different Ruby versions.

Example:

```ruby
>> s
=> "gems/3.3.0+0/gems/bundler-2.3.24/lib/bundler/setup"
>> /#{s}/.match?(s)
=> false
>> /#{Regexp.escape(s)}/.match?(s)
=> true
```

### DIFF
--- a/bundler/spec/bundler/shared_helpers_spec.rb
+++ b/bundler/spec/bundler/shared_helpers_spec.rb
@@ -242,7 +242,7 @@ RSpec.describe Bundler::SharedHelpers do
     shared_examples_for "ENV['RUBYOPT'] gets set correctly" do
       it "ensures -rbundler/setup is at the beginning of ENV['RUBYOPT']" do
         subject.set_bundle_environment
-        expect(ENV["RUBYOPT"].split(" ")).to start_with("-r#{source_lib_dir}/bundler/setup")
+        expect(ENV["RUBYOPT"].split(" ")).to start_with("-r#{install_path}/bundler/setup")
       end
     end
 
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
I think there are specs for this at `bundler/spec/bundler/shared_helpers_spec.rb`, could you add one to catch this issue?

intrip:
> I think there are specs for this at `bundler/spec/bundler/shared_helpers_spec.rb`, could you add one to catch this issue?

Of course! Added a test.

intrip:
Just noticed the failure, will fix it

deivid-rodriguez:
Thank you!

bundler/spec/bundler/shared_helpers_spec.rb
```diff
@@ -385,6 +385,22 @@
       it_behaves_like "ENV['RUBYOPT'] gets set correctly"
     end
 
+    context "ENV['RUBYOPT'] exists and contains -rbundler/setup with special characters in the path" do
+      source_lib_dir = "/opt/ruby3.3.0-preview2/lib/ruby/3.3.0+0"
```
deivid-rodriguez, line 389: It's a bit non standard to set a variable in this context. Maybe you can group all specs that use the shared specs under a context

```ruby
context "when bundler install path is standard" do
  let(:install_path) { source_lib_dir }

  ### current specs
end

context "when bundler install path contains special characters" do
  let(:install_path) { "/opt/ruby3.3.0-preview2/lib/ruby/3.3.0+0" }

  ### new spec that you added
end
```

And change the shared spec to use `install_path`.

deivid-rodriguez (approved):
Super minor comment on the spec. Feel free to address it if you think it's a improvement!

bundler/spec/bundler/shared_helpers_spec.rb
intrip: Thanks, followed your suggestion.

intrip:
I'm having trouble reproducing the failure locally because It's triggered only in some combination of the test matrix 😅. I've added a stub which "should" help but let's see CI.

intrip:
@deivid-rodriguez CI is green now

deivid-rodriguez:
Looks great, thank you!

PR #7003, opened 2023-09-26
Raise an error when lockfile is invalid

etiennebarrie:
Follow-up to https://github.com/rubygems/rubygems/pull/4459

Instead of printing a warning, raise a handled error to make sure `bundle install` exits with a non-zero status code.

An invalid lockfile passed CI when I wish it would have failed.

To add some detail to the linked PR, this can happen relatively easily because Git will happily merge lockfiles, so if one commit changes the top of the lockfile, and another changes the bottom, Git can end up merging those, resulting in an invalid lockfile (in our case it was a commit updating gem to a new version that was limiting the requirements for another dependency, and a commit updating said dependency to the now forbidden version). Once the second commit was merged, the warning was printed, but did not result in CI failing.

A demonstration is available in https://github.com/etiennebarrie/invalid-gemfile.lock.

The fix raises a `InstallError` instead of printing a warning, resulting in a non-zero status code.

### DIFF
--- a/bundler/lib/bundler/installer/parallel_installer.rb
+++ b/bundler/lib/bundler/installer/parallel_installer.rb
@@ -108,19 +108,19 @@ module Bundler
       end.reject {|a| a.last.empty? }
       return if unmet_dependencies.empty?
 
-      warning = []
-      warning << "Your lockfile doesn't include a valid resolution."
-      warning << "You can fix this by regenerating your lockfile or manually editing the bad locked gems to a version that satisfies all dependencies."
-      warning << "The unmet dependencies are:"
+      error = []
+      error << "Your lockfile doesn't include a valid resolution."
+      error << "You can fix this by regenerating your lockfile or manually editing the bad locked gems to a version that satisfies all dependencies."
+      error << "The unmet dependencies are:"
 
       unmet_dependencies.each do |spec, unmet_spec_dependencies|
         unmet_spec_dependencies.each do |unmet_spec_dependency|
           found = @specs.find {|s| s.name == unmet_spec_dependency.name && !unmet_spec_dependency.matches_spec?(s.spec) }
-          warning << "* #{unmet_spec_dependency}, dependency of #{spec.full_name}, unsatisfied by #{found.full_name}"
+          error << "* #{unmet_spec_dependency}, dependency of #{spec.full_name}, unsatisfied by #{found.full_name}"
         end
       end
 
-      Bundler.ui.warn(warning.join("\n"))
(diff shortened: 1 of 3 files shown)


etiennebarrie:
So there are specs checking that installing a Gemfile with invalid dependencies should work.

0ae2a168fdeb5801f0f6d22a85e44d6474ff2b7b
abe73a4706c07ac1917a44b2756b31de9d2ae6d0

I tentatively removed them but it's not something that should be merged, I'm just trying to make sure nothing is else is impacted.
I'm not sure why that is a desirable feature though in either case. Do you remember @segiddins? That second spec in particular shouldn't be removed as it's the only test testing the Bundler/RubyGems integration monkey-patching of `Gem.finish_resolve`.

The changelog for those specs/features is:
https://github.com/rubygems/rubygems/blob/d750e7050153a2e8f8b40fcbd347ea7252a6d0b6/bundler/CHANGELOG.md?plain=1#L1880
https://github.com/rubygems/rubygems/blob/d750e7050153a2e8f8b40fcbd347ea7252a6d0b6/bundler/CHANGELOG.md?plain=1#L2209

segiddins:
Correct, our preference is healing broken lockfiles. This comes from https://github.com/rubygems/bundler/pull/4445 & https://github.com/rubygems/bundler/pull/5637

segiddins:
The reason this exists for `gemspec` dependencies in particular is otherwise you end up with a circularity when trying to bump a version:

e.g. rails tests use a version of a dep that pins to a rails major, rails wants to bump major versions... with this feature, rails can do that, since rails is coming from a `gemspec` dependency

deivid-rodriguez:
The special thing about this particular situation is that this lockfile is not healable. So indeed, this ends up installing gems that are not compatible with each other, which seems not good.

In my opinion the best thing to do here would be to _try_ to heal it by re-resolving gems and end up with a resolution error.

deivid-rodriguez:
Working on my proposed approach at https://github.com/rubygems/rubygems/pull/7020.

With that PR, you'll now get a proper explanation when trying to install this lockfile:

```
$ bundle install
Resolving dependencies...

Because zzz >= 0, < 0.2.0 could not be found in source at `.`
  and every version of aaa depends on zzz < 0.2.0,
  every version of aaa requires zzz < 0.
So, because Gemfile depends on aaa >= 0
  and Gemfile depends on zzz >= 0,
  version solving has failed.
```

deivid-rodriguez:
Closing since this was superseded by #7020.

Thanks for raising this issue to us @etiennebarrie!

Issue #7013, opened 2023-09-29
Can't use Ruby's preview version with `ruby file: '.ruby-version'`


harshalbhakta:
### Describe the problem as clearly as you can

I got below error when using a preview version of Ruby in `.ruby-version` file

#### `bundle install` error

````ruby
Your Ruby version is 3.3.0.preview2, but your Gemfile specified 3.3.0.pre.preview2
````

#### .ruby-version

````ruby
3.3.0-preview2
````

#### Gemfile

````ruby
ruby file: '.ruby-version'
````

### Did you try upgrading rubygems & bundler?

Yes. Using the latest version of rubygems and bundler.

### Post steps to reproduce the problem

Use `3.3.0-preview2` in `.ruby-version` file with `rbenv`

### Which command did you run?

`bundle install`

### What were you expecting to happen?

`bundle install` should have executed without any error

### What actually happened?

Got below error

````ruby
Your Ruby version is 3.3.0.preview2, but your Gemfile specified 3.3.0.pre.preview2
````

### If not included with the output of your command, run `bundle env` and paste the output below

````ruby
## Environment

```
Bundler       2.5.0.dev
  Platforms   ruby, x86_64-darwin-23
Ruby          3.3.0p-1 (2023-09-14 revision e50fcca9a79d8e25b33ad3611df6bf4627faafbf) [x86_64-darwin-23]
  Full Path   /Users/harshal/.rbenv/versions/3.3.0-preview2/bin/ruby
  Config Dir  /Users/harshal/.rbenv/versions/3.3.0-preview2/etc
RubyGems      3.5.0.dev
  Gem Home    /Users/harshal/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0
  Gem Path    /Users/harshal/.local/share/gem/ruby/3.3.0+0:/Users/harshal/.rbenv/versions/3.3.0-preview2/lib/ruby/gems/3.3.0+0
  User Home   /Users/harshal
  User Path   /Users/harshal/.local/share/gem/ruby/3.3.0+0
  Bin Dir     /Users/harshal/.rbenv/versions/3.3.0-preview2/bin
Tools         
  Git         2.42.0
  RVM         not installed
  rbenv       rbenv 1.2.0-17-ga8ecfc7
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-09-29
Git SHA           unknown
Released Version  false
````

hsbt:
[...]

PR #7016, opened 2023-09-30
Support Ruby's preview version format (Ex: 3.3.0-preview2) in Gemfile

harshalbhakta:

Fixes https://github.com/rubygems/rubygems/issues/7013

Support Ruby's preview versions in Gemfile.

Replaced '-' with '.' before passing to `Gem::Requirement.parse` to support preview version format (Ex: 3.3.0-preview2) in Gemfile.

### DIFF
--- a/bundler/lib/bundler/ruby_version.rb
+++ b/bundler/lib/bundler/ruby_version.rb
@@ -23,7 +23,7 @@ module Bundler
       #   specified must match the version.
 
       @versions = Array(versions).map do |v|
-        op, v = Gem::Requirement.parse(v)
+        op, v = Gem::Requirement.parse(normalize_version(v))
         op == "=" ? v.to_s : "#{op} #{v}"
       end
 
@@ -111,6 +111,13 @@ module Bundler
     end
 
     private
+
+    # Ruby's official preview version format uses a `-`: Example: 3.3.0-preview2
+    # However, RubyGems recognizes preview version format with a `.`: Example: 3.3.0.preview2
+    # Returns version string after replacing `-` with `.`
+    def normalize_version(version)
+      version.tr("-", ".")
+    end
 
     def matches?(requirements, version)
       # Handles RUBY_PATCHLEVEL of -1 for instances like ruby-head
--- a/bundler/spec/bundler/ruby_dsl_spec.rb
+++ b/bundler/spec/bundler/ruby_dsl_spec.rb
@@ -65,6 +65,15 @@ RSpec.describe Bundler::RubyDsl do
       it_behaves_like "it stores the ruby version"
     end
 
+    context "with a preview version" do
+      let(:ruby_version) { "3.3.0-preview2" }
+
+      it "stores the version" do
+        expect(subject.versions).to eq(Array("3.3.0.preview2"))
+        expect(subject.gem_version.version).to eq("3.3.0.preview2")
+      end
+    end
+
     context "with two requirements in the same string" do
       let(:ruby_version) { ">= 2.0.0, < 3.0" }
       it "raises an error" do


bundler/lib/bundler/ruby_version.rb
```diff
@@ -23,7 +23,7 @@ def initialize(versions, patchlevel, engine, engine_version)
       #   specified must match the version.
 
       @versions = Array(versions).map do |v|
-        op, v = Gem::Requirement.parse(v)
+        op, v = Gem::Requirement.parse(v.tr("-", "."))
```
simi, line 26: Would it be possible to move this into `PATTERN` instead? Whole parsing logic is currently in there.

harshalbhakta: Not really sure if modifying `PATTERN` will help. `3.3.0-preview2` does not violate any regex matching for `PATTERN`. Preview version format is `3.3.0-preview2` while `rubygems` repository works correctly with `3.3.0.preview2`. I think `3.3.0-preview2` does not work beacuse of below line in [version.rb](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L229).

````ruby
@version = version.to_s.strip.gsub("-",".pre.")
```` 

This line converts `3.3.0-preview2` to `3.3.0.pre.preview2`, which is an incorrect Ruby version and causes below error while executing `bundle install`.

````ruby
Your Ruby version is 3.3.0.preview2, but your Gemfile specified 3.3.0.pre.preview2
````

Will dig a little deeper to see how this can be fixed using `PATTERN`.

Comment (simi) [reply bundler/lib/bundler/ruby_version.rb:26]: If PATTERN is not the best place to do this, just please wrap this into own method with documentation. This way it is just replacing `-` with `.` inline in constructor with no explanation.

Comment (simi) [reply bundler/lib/bundler/ruby_version.rb:26]: :thinking: changing ruby version part of PATTERN `(\d+\.\d+\.\d+(?:\.\S+)?)` into `(\d+\.\d+\.\d+(?:\.?-?\S+)?)` seems properly parsing the `3.3.0-preview2` versions. No idea if that breaks other parts.

Comment (simi) [reply bundler/lib/bundler/ruby_version.rb:26]: > I think 3.3.0-preview2 does not work beacuse of below line in [version.rb](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L229).

I'll check with this as well, thanks for the info.

Comment (harshalbhakta) [reply bundler/lib/bundler/ruby_version.rb:26]: I created new branch and just changed the PATTERN to what you suggested. That did not resolve the issue.

image

However, using `v.tr("-", ".")` resolved the issue. Also, all tests passed with this change here (https://github.com/harshalbhakta/rubygems/pull/1) so I don't think it will break anything else.

image

Should definitely add this to a method with documentation. Just wanted to be sure that this is the correct way of solving this issue.

Comment (harshalbhakta) [reply bundler/lib/bundler/ruby_version.rb]: > If PATTERN is not the best place to do this, just please wrap this into own method with documentation. This way it is just replacing `-` with `.` inline in constructor with no explanation.

Moved it to `normalize_version` method

**Review by deivid-rodriguez — APPROVED** [review]
I believe this PR is the right solution.

Unfortunately RubyGems does not accept versions with the `-` character in them, so it uses something that looks similar.

In the RubyGems side, we do the same kind of trick by [extracting the prerelease identifier from `RUBY_DESCRIPTION`, and putting it after `RUBY_VERSION`](https://github.com/rubygems/rubygems/blob/20d38e27a44b91e81d58b750b023ac73c8bb5242/lib/rubygems.rb#L866-L871), so that

```
$ ruby -e 'puts Gem.ruby_version'
3.3.0.preview2
```

So we should make sure we do the same thing with the Ruby version in `Gemfile` to avoid this kind of mismatch.

File: bundler/lib/bundler/ruby_version.rb [hunk 2]
```diff
@@ -112,6 +112,13 @@ def self.system
 
     private
 
+    # Ruby's official preview version format uses a `-`: Example: 3.3.0-preview2
+    # However, rubygems recognizes preview version format with a `.`: Example: 3.3.0.preview2
```
simi, line 116: _nit: RubyGems_

harshalbhakta: Fixed this

simi (approved):
Thanks a lot for patience with my review @harshalbhakta! :muscle:

hsbt (approved):

schneems:
Thanks for working on this. It bit me big time in https://github.com/heroku/docker-heroku-ruby-builder/pull/34

PR #7017, opened 2023-09-30
Fix typo: eglible -> eligible.

simi:
Sadly, I have introduced this mistake in original PR.


deivid-rodriguez (approved):
Too bad aspell did not catch it!

PR #7020, opened 2023-09-30
Re-resolve when lockfile is invalid

deivid-rodriguez:

Sometimes, even if we give a warning about it, we may end up installing gems incompatible with each other.

My fix is to move the check for unmet dependencies in lockfile just in time to be able to re-resolve if unmet dependencies are found.

Still missing a spec.


martinemde (approved):
This looks like a much better approach.

deivid-rodriguez:
Thanks @martinemde!

etiennebarrie:
Tested this and it works great, I get a non-zero exit status code (6 for some reason, I don't think that's publicly documented), the error message is really good. Thank you!

I don't see anything checking the return value explicitly in the spec though, the `raise_on_error: false` allows the command to fail but does not make sure it does.

Also while looking into this, I found this, which sounds 6 shouldn't go all the way to the exit status of the command:
https://github.com/rubygems/rubygems/blob/956ecb81d62a6b723b5942c4f457bfdc196cb78b/bundler/lib/bundler/errors.rb#L23-L24

deivid-rodriguez:
You're right, happy to take a PR to add that assertion. We have helpers for checking that no gems were installed, and also for checking that the last command returned a failed status.

PR #7036, opened 2023-10-05
Use latest patch level Ruby releases in CI

deivid-rodriguez:

Not much, just making sure we're on latest.

Upgrade all the things.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -30,7 +30,7 @@ jobs:
 
         ruby:
           - { name: ruby-2.6, value: 2.6.10 }
-          - { name: ruby-2.7, value: 2.7.7 }
+          - { name: ruby-2.7, value: 2.7.8 }
           - { name: ruby-3.0, value: 3.0.5 }
           - { name: ruby-3.1, value: 3.1.3 }
           - { name: ruby-3.2, value: 3.2.1 }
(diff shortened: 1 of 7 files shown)


martinemde:
Windows builds besides 3.2 failed. I'm rerunning just to see if it's transient, but I'm not hopeful.

simi (approved):
no idea about the CI failures, but :+1: on update

martinemde:
CI on windows is failing because `etc` is a default gem. Here's one of the errors:

```

**Comment by martinemde** [comment]
Aha, I think etc needs to be added to the exclusions. I'll push an update.

**Comment by martinemde** [comment]
I don't know if the change I pushed should specifically target windows since they are the only failing runs, or if my patch is fine excluding it always. It would seem to be a no-op on any other build since they were already passing.

**Comment by deivid-rodriguez** [comment]
Thanks for this. From https://github.com/rubygems/rubygems/actions/runs/6441869064/job/17492104237 I see that 3.2.2 was also passing?

Yes, I normally prefer to limit these exclusions to the environments are versions where the feature is not working, so that we can keep testing other combinations still work fine, and so that we can track how the feature evolves eventually remove the exclusion.

**Comment by martinemde** [comment]
> I see that 3.2.2 was also passing?
I tried to keep 3.2.2 in the tests. I tried to keep 3.2 in the tests. Did I mess up the conditional?

> Yes, I normally prefer to limit these exclusions to the environments are versions where the feature is not working

We could constrain to only windows (not sure why it only shows on windows)

**Comment by deivid-rodriguez** [comment]
Oh god, I actually read your message and assumed the exact content of your extra commit without even looking at it 😅. Yeah, the current patch is exactly what I meant, sorry about that.

**Comment by deivid-rodriguez** [comment]
> We could constrain to only windows (not sure why it only shows on windows)

Yep, that's my suggestion. Ruby On Windows adds some customisations through a `operating_system.rb` file that RubyGems loads when it's required. It could be that this file is now loading `etc` for some reason.

**Comment by martinemde** [comment]
Updated!

**Comment by deivid-rodriguez** [comment]
I can't approve my own PR but looks great!

**Review by simi — APPROVED** [review]
@deivid-rodriguez let's use my approval on your behalf :)

**Comment by martinemde** [comment]
Well you don't see this everyday (but I don't work with Ruby on Windows everyday):

```
       D:/a/rubygems/rubygems/bundler/tmp/1/gems/system/gems/bundler-2.5.0.dev/lib/bundler/injector.rb:90: [BUG] Segmentation fault
       ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x64-mingw-ucrt]
```

deivid-rodriguez:
Never seen that before, and indeed rare to see Ruby segfault these days.

I'd say this may have been surfaced by the recent optimizations merged (maybe #6970?), since the [last build on master](https://github.com/rubygems/rubygems/actions/runs/6456478707/job/17526067804) also run into that, and also [a separate PR](https://github.com/rubygems/rubygems/actions/runs/6458388573/job/17532028630?pr=7042) is having issues with that spec.

PR #7037, opened 2023-10-05
Lock Rails version installed in our Windows CI

deivid-rodriguez:

Broken CI after Rails 7.1 release.

The point of this realworld test was that the `bundle install` commands run by `rails new` succeed. I don't think the Rails version matters much, and Rails 7.1 does not work with JRuby.


martinemde (approved):

martinemde:
We may want to revert this if and when it does work.

PR #7038, opened 2023-10-06
Add missing platform to tool/bundler/(dev|standard|rubocop)26_gems.rb

martinemde:

Avoid re-resolving gems for new platform on ruby 2.6 CI runs.

Add platforms used in CI to the development gemfiles in `tool/bundler` for ruby 26.

### DIFF
--- a/tool/bundler/standard26_gems.rb.lock
+++ b/tool/bundler/standard26_gems.rb.lock
@@ -65,7 +65,12 @@ GEM
     unicode-display_width (2.4.2)
 
 PLATFORMS
+  arm64-darwin-22
   arm64-darwin-23
+  x86_64-darwin-20
+  x86_64-darwin-21
+  x86_64-darwin-22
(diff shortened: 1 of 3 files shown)


tool/bundler/standard26_gems.rb.lock
```diff
@@ -78,4 +83,4 @@ DEPENDENCIES
   test-unit
 
 BUNDLED WITH
-   2.4.13
+   2.5.0.dev
```
simi, line 86: Do we need dev bundler here?

martinemde: It has to switch bundler version during install if it's not the same as the current version, but it's probably not important either way.

martinemde: I pushed using 2.4.20 instead.

deivid-rodriguez (approved):

PR #7039, opened 2023-10-07
Use DigestIO to reduce I/O when parsing compact index file digests

martinemde:

When reviewing #6971, @simi asked about moving the MD5 digest code to its own helper. When looking at the code it seemed like we could calculate the digest as the files were moved and saved. I decided to see if it would work.

This builds on #6971 using `Gem::Package::DigestIO` to read the compact index file digests while the other file movement operations are being performed, saving a few extra file reads.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "../vendored_fileutils"
+require "rubygems/package"
 
 module Bundler
   class CompactIndexClient
@@ -29,11 +30,13 @@ module Bundler
         local_temp_path = local_temp_path.sub(/$/, ".retrying") if retrying
         local_temp_path = local_temp_path.sub(/$/, ".tmp")
 
+        digests = { :MD5 => SharedHelpers.digest(:MD5).new }
+
         # first try to fetch any new bytes on the existing file
         if retrying.nil? && local_path.file?
-          copy_file local_path, local_temp_path
+          copy_file local_path, local_temp_path, digests
 
-          headers["If-None-Match"] = etag_for(local_temp_path)
+          headers["If-None-Match"] = etag_for(digests)
           headers["Range"] =
             if local_temp_path.size.nonzero?
               # Subtract a byte to ensure the range won't be empty.
@@ -52,13 +55,14 @@ module Bundler
         etag = (response["ETag"] || "").gsub(%r{\AW/}, "")
         correct_response = SharedHelpers.filesystem_access(local_temp_path) do
           if response.is_a?(Net::HTTPPartialContent) && local_temp_path.size.nonzero?
-            local_temp_path.open("a") {|f| f << slice_body(content, 1..-1) }
+            local_temp_path.open("a") {|f| digest_io(f, digests).write slice_body(content, 1..-1) }
 
-            etag_for(local_temp_path) == etag
+            etag_for(digests) == etag
           else
-            local_temp_path.open("wb") {|f| f << content }
+            digests[:MD5] = SharedHelpers.digest(:MD5).new
+            local_temp_path.open("wb") {|f| digest_io(f, digests).write content }
 
-            etag.length.zero? || etag_for(local_temp_path) == etag
+            etag.length.zero? || etag_for(digests) == etag
           end
         end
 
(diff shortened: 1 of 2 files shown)


martinemde:
I switched the base to @segiddins branch so that we could separate his changes from what I just wrote here so it's less confusing.

segiddins:
What's the comparison of allocations here?

martinemde:
@segiddins Ok, this was closer than I thought. The memory usage is barely different, but DigestIO is slightly higher (I even tried making our own DigestIO that doesn't take a hash). Overall, the usage here pales in comparison to the rest of the app.

Tested by running `bundle lock --update --bundler` on a client app I have locally.

```
==> memprof.before.txt <==
Total allocated: 398.62 MB (5878673 objects)
Total retained:  107.41 MB (1469380 objects)
==> memprof.custom_digest_io.txt <==
Total allocated: 370.14 MB (5634274 objects)
Total retained:  104.96 MB (1412709 objects)
==> memprof.rubygems_digestio.txt <==
Total allocated: 370.15 MB (5634247 objects)
Total retained:  104.96 MB (1412706 objects)
==> memprof.segiddins.txt <==
Total allocated: 369.68 MB (5629483 objects)
Total retained:  104.90 MB (1412106 objects)
```

Not much, still better than before, but your version does allocate less memory. 

But that's not the whole story. I couldn't believe there was no advantage to avoiding extra reads of the same file. The advantage is that it's much faster.

```
           digest_io     1.781k i/100ms
    copy then digest     1.193k i/100ms
           digest_io     17.251k (± 8.9%) i/s -     85.488k in   5.004996s
    copy then digest     11.590k (± 6.7%) i/s -     58.457k in   5.072384s

Comparison:
           digest_io:    17251.3 i/s
    copy then digest:    11590.4 i/s - 1.49x  slower
```

<details>
  <summary>benchmark.rb</summary>
    # frozen_string_literal: true

    require 'digest'
    require 'benchmark/ips'

    class DigestIO
      def initialize(io, digest)
        @io = io
        @digest = digest
      end

      def write(data)
        @digest << data
        @io.write(data)
      end
    end

    def copy_file(source, dest, digest)
      File.open(source, "r") do |s|
        File.open(dest, "wb", s.stat.mode) do |f|
          out = digest ? DigestIO.new(f, digest) : f
          IO.copy_stream(s, out)
        end
      end
    end

    def calcdigest(path)
      File.open(path, "rb") do |f|
        digest = Digest::MD5.new
        buf = String.new(:capacity => 16_384, :encoding => Encoding::BINARY)
        digest << buf while f.read(16_384, buf)
        digest.hexdigest
      end
    end

    source = 'tmp/1/gems/remote1/specs.4.8'
    dest = 'tmp/deleteme'

    Benchmark.ips do |x|
      x.report("digest_io") do
        digest = Digest::MD5.new
        copy_file(source, dest, digest)
        digest.hexdigest
      end

      x.report("copy then digest") do
        copy_file(source, dest, nil)
        calcdigest(dest)
      end

      x.compare!
    end
</details>

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -80,38 +84,30 @@ def update(local_path, remote_path, retrying = nil)
...
+
+      def digest_io(io, digests)
+        Gem::Package::DigestIO.new(io, digests)
       end
 
       def slice_body(body, range)
         body.byteslice(range)
       end
 
       def checksum_for_file(path)
```
segiddins, line 99: can this method be removed now that `etag_for` uses the output from `digest_io` ?

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative "../vendored_fileutils"
+require "rubygems/package"
```
segiddins, line 4: can we instead use `gem/package/digest_io` ?

segiddins (approved):
will defer to your judgement on which version you prefer, @martinemde

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: unfortunately that causes the error `NameError: uninitialized constant Gem::Package`

martinemde:
I'm continuing this in #7122 and closing this version. This only really makes sense when merged with the opaque etag work.

PR #7041, opened 2023-10-09
BUG_TRIAGE: Refer to CHANGELOG with a full URL for the link in the website to render right

olleolleolle:

We are using this in the bundler-site repo, and this would make the link work.

I assume that CHANGELOG.md is not present in bundler-site, so the link becomes a 404.

Fixes https://github.com/rubygems/bundler-site/issues/846

Use a full URL.


deivid-rodriguez (approved):
Make sense to me!

Alternatively we could also display the changelog on the bundler-site side.

olleolleolle:
@deivid-rodriguez I _think_ I prefer to refer to canonical place for the CHANGELOG.

deivid-rodriguez:
I think that's fine!

My only concern is that someone sees that there are only relative links here and updates this one to make things consistent in the future too and we get a 404 again. Displaying the changelog in the docs site could be handy for users and wouldn't have that potential problem.

deivid-rodriguez:
https://github.com/rubygems/bundler-site/pull/1239 feels better to me, so let me close this and I'll get to the other PR now!

PR #7046, opened 2023-10-09
fix pre-commit hook failing rubocop on non-ruby files

ccutrer:

If you changed a man page, it was force-feeding that to rubocop when committing, and it would fail every line as unparseable.

Add the rubocop flag that tells it to apply file-type filters to the list of files passed on the command line.

### DIFF
--- a/.githooks/pre-commit-rubocop
+++ b/.githooks/pre-commit-rubocop
@@ -5,7 +5,7 @@ length=${#files}
 
 # Run rubocop hook only when analyzable files are present
 if [[ $length -ne 0 ]]; then
-  ./bin/rubocop $files
+  ./bin/rubocop --only-recognized-file-types $files
 else
   echo "No files to analyze"
 fi


.githooks/pre-commit-rubocop
```diff
@@ -5,7 +5,7 @@ length=${#files}
 
 # Run rubocop hook only when analyzable files are present
 if [[ $length -ne 0 ]]; then
-  ./bin/rubocop $files
+  ./bin/rubocop --only-recognized-file-types $files
```
martinemde, line 8: Should this go directly into `bin/rubocop`? I'm not sure I'd ever want to run rubocop on an unrecognized file.

ccutrer:
At that point... why does Rubocop allow it at all??

martinemde:
Good point, haha. Fixing it here makes sense then.

martinemde:
The test failure is the windows segfault. I'm ignoring that for now since it's happening on master too.

PR #7047, opened 2023-10-09
Refactor lockfile generation and deprecate `Definition#lock` with explicit lockfile

deivid-rodriguez:

No user problem, but came up with a refactoring that I believe makes the code better and easier to customize.

I think this may help [bootboot](https://github.com/Shopify/bootboot) remove some monkey patches.

This PR deprecates passing a file to `Definition#lock`. Instead, it will always lock the lockfile the definition was built for.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -334,7 +334,9 @@ module Bundler
           "Instead, instantiate a new definition passing `#{target_lockfile}`, and call `lock` without a file argument on that definition"
         end
 
-        warn "Passing a file to `Definition#lock` is deprecated. #{suggestion}"
+        msg = "`Definition#lock` was passed a target file argument. #{suggestion}"
+
+        Bundler::SharedHelpers.major_deprecation 2, msg
       end
 
       write_lock(target_lockfile, preserve_unknown_sections)
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -320,38 +320,24 @@ module Bundler
       dependencies.map(&:groups).flatten.uniq
     end
 
-    def lock(file, preserve_unknown_sections = false)
-      return if Definition.no_lock
+    def lock(file_or_preserve_unknown_sections = false, preserve_unknown_sections_or_unused = false)
+      if [true, false, nil].include?(file_or_preserve_unknown_sections)
+        target_lockfile = lockfile || Bundler.default_lockfile
+        preserve_unknown_sections = file_or_preserve_unknown_sections
+      else
+        target_lockfile = file_or_preserve_unknown_sections
+        preserve_unknown_sections = preserve_unknown_sections_or_unused
 
-      contents = to_lock
+        suggestion = if target_lockfile == lockfile
+          "To fix this warning, remove it from the `Definition#lock` call."
+        else
+          "Instead, instantiate a new definition passing `#{target_lockfile}`, and call `lock` without a file argument on that definition"
+        end
 
-      # Convert to \r\n if the existing lock has them
-      # i.e., Windows with `git config core.autocrlf=true`
-      contents.gsub!(/\n/, "\r\n") if @lockfile_contents.match?("\r\n")
-
-      if @locked_bundler_version
-        locked_major = @locked_bundler_version.segments.first
-        current_major = bundler_version_to_lock.segments.first
-
-        updating_major = locked_major < current_major
+        warn "Passing a file to `Definition#lock` is deprecated. #{suggestion}"
       end
 
-      preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
-
-      if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
-        return if Bundler.frozen_bundle?
-        SharedHelpers.filesystem_access(file) { FileUtils.touch(file) }
-        return
-      end
-
-      if Bundler.frozen_bundle?
-        Bundler.ui.error "Cannot write a changed lockfile while frozen."
-        return
-      end
-
-      SharedHelpers.filesystem_access(file) do |p|
-        File.open(p, "wb") {|f| f.puts(contents) }
-      end
+      write_lock(target_lockfile, preserve_unknown_sections)
     end
 
     def locked_ruby_version
@@ -518,7 +504,45 @@ module Bundler
     end
 
     def lockfile_exists?
-      lockfile && File.exist?(lockfile)
+      file_exists?(lockfile)
+    end
+
+    def file_exists?(file)
+      file && File.exist?(file)
+    end
+
+    def write_lock(file, preserve_unknown_sections)
+      return if Definition.no_lock
+
+      contents = to_lock
+
+      # Convert to \r\n if the existing lock has them
+      # i.e., Windows with `git config core.autocrlf=true`
+      contents.gsub!(/\n/, "\r\n") if @lockfile_contents.match?("\r\n")
+
+      if @locked_bundler_version
+        locked_major = @locked_bundler_version.segments.first
+        current_major = bundler_version_to_lock.segments.first
+
+        updating_major = locked_major < current_major
+      end
+
+      preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
+
+      if file_exists?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+        return if Bundler.frozen_bundle?
+        SharedHelpers.filesystem_access(file) { FileUtils.touch(file) }
+        return
+      end
+
+      if Bundler.frozen_bundle?
+        Bundler.ui.error "Cannot write a changed lockfile while frozen."
+        return
+      end
+
+      SharedHelpers.filesystem_access(file) do |p|
+        File.open(p, "wb") {|f| f.puts(contents) }
+      end
     end
 
     def resolver
--- a/bundler/lib/bundler/installer.rb
+++ b/bundler/lib/bundler/installer.rb
@@ -260,8 +260,8 @@ module Bundler
       true
     end
 
-    def lock(opts = {})
-      @definition.lock(Bundler.default_lockfile, opts[:preserve_unknown_sections])
+    def lock
+      @definition.lock(Bundler.default_lockfile)
     end
   end
 end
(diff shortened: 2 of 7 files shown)


bundler/lib/bundler/definition.rb
```diff
@@ -310,34 +310,24 @@ def groups
...
+        preserve_unknown_sections = file_or_preserve_unknown_sections
+      else
+        target_lockfile = file_or_preserve_unknown_sections
+        preserve_unknown_sections = preserve_unknown_sections_or_unused
 
-      return if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+        suggestion = if target_lockfile == lockfile
+          "To fix this warning, remove it from the `Definition#lock` call."
+        else
+          "Instead, instantiate a new definition passing `#{target_lockfile}`, and call `lock` without a file argument on that definition"
```
martinemde, line 324: Maybe a wild idea, but in this case, we could simplify Definition by literally re-creating the Definition on this branch and calling lock, just as we're advising. Then definition would only ever write to the lockfile given at the start. It would be a performance hit but hopefully no one fixes the warning and it would maybe allow us to commit to a "final" version for the rest of the Definition API! Too crazy?

bundler/lib/bundler/installer.rb
```diff
@@ -260,8 +260,8 @@ def resolve_if_needed(options)
       true
     end
 
-    def lock(opts = {})
-      @definition.lock(Bundler.default_lockfile, opts[:preserve_unknown_sections])
+    def lock
+      @definition.lock
```
martinemde, line 264: Why does the `Bundler::Runtime` preserve the opts, but this one doesn't?

bundler/spec/bundler/definition_spec.rb
```diff
@@ -10,34 +10,34 @@
       allow(Bundler).to receive(:ui) { double("UI", :info => "", :debug => "") }
     end
     context "when it's not possible to write to the file" do
-      subject { Bundler::Definition.new(nil, [], Bundler::SourceList.new, []) }
+      subject { Bundler::Definition.new("Gemfile.lock", [], Bundler::SourceList.new, []) }
```
martinemde, line 13: Your other caller [converts to Pathname](https://github.com/rubygems/rubygems/pull/7047/files#diff-f6ec8889d95d366c2ac4d3b18240fc23eb9486bce5e4a999991ed557ff961d7aR37) before calling, however, this is inconsistent across initializations of Definition. Should we test for receiving a Pathname too?

martinemde (approved):
Good refactor. It makes a lot more sense to me this way. Just a few question and maybe a crazy idea that you can probably ignore.

bundler/lib/bundler/definition.rb
deivid-rodriguez: Not crazy at all! Neat idea on the contrary. Will update 👍.

bundler/lib/bundler/installer.rb
deivid-rodriguez: I removed `opts` because they were not being passed by any callers, so they felt misleading.

My understanding of why we're not passing `preserve_unknown_sections` here is the following:

* When `lock`ing from a `Bundler::Runtime` we never want lockfile updates, because that happens from `bundle exec` or a `bundler/setup` require and that causing lockfile modifications is generally unexpected. So we explicitly pass `preserve_unknown_sections: true`.

* When `lock`ing from `Bundler::Installer`, however, we want to respect [the rules defined in `Bundler::Definition`](https://github.com/rubygems/rubygems/blob/ed65279100234a17d65d71fe26de5083984ac5b8/bundler/lib/bundler/definition.rb#L329) for deciding whether to remove unknown sections or not.  So we don't pass anything.

bundler/lib/bundler/definition.rb
martinemde: Nice. When I reread my comment I probably meant "hopefully no one ignores the warning". I'm guessing the performance hit won't matter because it will get fixed, then we can eventually remove this code branch.

deivid-rodriguez: Yes, I was confused too by that part, but I assumed you meant exactly the opposite :) I agree performance is not a concern in deprecated code paths.

bundler/spec/bundler/definition_spec.rb
deivid-rodriguez: I looked into this and I believe all our callers actually use `Pathname`'s (and method headers explicitly mention `Pathname` as the expected type). So what I did was to change these unit specs to also recieve `Pathname`'s.

I think this is enough but let me know if we should explicitly be covering passing strings.

bundler/spec/bundler/definition_spec.rb
```diff
@@ -5,39 +5,36 @@
...
     end
     context "when Bundler::Definition.no_lock is set to true" do
-      subject { Bundler::Definition.new(nil, [], Bundler::SourceList.new, []) }
       before { Bundler::Definition.no_lock = true }
       after { Bundler::Definition.no_lock = false }
 
       it "does not create a lock file" do
-        subject.lock("Gemfile.lock")
+        subject.lock
         expect(File.file?("Gemfile.lock")).to eq false
```
martinemde, line 38: Could this be?

```suggest
      expect(bundled_app_lockfile).not_to be_file
```

martinemde (approved):

bundler/spec/bundler/definition_spec.rb
deivid-rodriguez: Yeah, I can update those since I'm at it!

bundler/lib/bundler/definition.rb
```diff
@@ -320,38 +320,26 @@ def groups
...
-
-      if @locked_bundler_version
-        locked_major = @locked_bundler_version.segments.first
-        current_major = bundler_version_to_lock.segments.first
-
-        updating_major = locked_major < current_major
-      end
+    def lock(file_or_preserve_unknown_sections = false, preserve_unknown_sections_or_unused = false)
+      if [true, false, nil].include?(file_or_preserve_unknown_sections)
+        target_lockfile = lockfile || Bundler.default_lockfile
```
jeremy, line 325: Regression: this causes bundler/inline, which had a `nil` lockfile previously, to fall back to the default lockfile—which may exist—and overwrite it with the inline definition's lock.

This is a common pattern with bin/some-executable that executes with an inline bundle distinct from the main bundle.

To reproduce:
```bash
mkdir inline-lockfile-bug
cd inline-lockfile-bug
touch Gemfile.lock
chmod u-w Gemfile.lock  # to demonstrate that the inline bundle tries to overwrite
ruby -v; gem -v; bundle -v
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin24]
3.3.22
Bundler version 2.5.19
ruby -rbundler/inline -e 'gemfile {}'
/Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/definition.rb:547:in `initialize': Permission denied @ rb_sysopen - /Users/jeremy/inline-lockfile-bug/Gemfile.lock (Errno::EACCES)
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/definition.rb:547:in `open'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/definition.rb:547:in `block in write_lock'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/shared_helpers.rb:105:in `filesystem_access'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/definition.rb:546:in `write_lock'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/definition.rb:337:in `lock'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/runtime.rb:98:in `lock'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/runtime.rb:35:in `setup'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/inline.rb:64:in `block (2 levels) in gemfile'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/settings.rb:158:in `temporary'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/inline.rb:50:in `block in gemfile'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler.rb:429:in `block in with_unbundled_env'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler.rb:684:in `with_env'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler.rb:429:in `with_unbundled_env'
	from /Users/jeremy/.local/share/mise/installs/ruby/3.3.5/lib/ruby/gems/3.3.0/gems/bundler-2.5.19/lib/bundler/inline.rb:42:in `gemfile'
	from -e:1:in `<main>'
```

deivid-rodriguez: Hei @jeremy!

You're totally right about this regression. It did not manifest until Bundler 2.5.19 because `bundler/inline` used to make `Definition#lock` a noop until https://github.com/rubygems/rubygems/commit/0b7be7bb770538fc26238dbaf055be9b06062ef8.

We got this reported last week and will be fixing it through https://github.com/rubygems/rubygems/pull/8055. I hope to do a quick release to patch this today or tomorrow.

deivid-rodriguez: I released Bundler 2.5.20 with this fix!

jeremy: Nice!

PR #7048, opened 2023-10-10
Deprecate `bundle plugin install --local-git=`

ccutrer:

Adding the --path param, I noticed --git and --local-git have identical implementations.

Keep --local-git as an undocumented alias for --git. Use the same actual implementation now. No functionality change.

### DIFF
--- a/bundler/lib/bundler/cli/plugin.rb
+++ b/bundler/lib/bundler/cli/plugin.rb
@@ -5,12 +5,12 @@ module Bundler
   class CLI::Plugin < Thor
     desc "install PLUGINS", "Install the plugin from the source"
     long_desc <<-D
-      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with --git (for remote repos) or --local_git (for local repos). If no sources are provided, it uses Gem.sources
+      Install plugins either from the rubygems source provided (with --source option) or from a git source provided with --git. If no sources are provided, it uses Gem.sources
    D
     method_option "source", type: :string, default: nil, banner: "URL of the RubyGems source to fetch the plugin from"
     method_option "version", type: :string, default: nil, banner: "The version of the plugin to fetch"
     method_option "git", type: :string, default: nil, banner: "URL of the git repo to fetch from"
-    method_option "local_git", type: :string, default: nil, banner: "Path of the local git repo to fetch from"
+    method_option "local_git", type: :string, default: nil, banner: "Path of the local git repo to fetch from (deprecated)"
     method_option "branch", type: :string, default: nil, banner: "The git branch to checkout"
     method_option "ref", type: :string, default: nil, banner: "The git revision to check out"
     def install(*plugins)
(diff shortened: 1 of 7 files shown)


bundler/lib/bundler/cli/plugin.rb
```diff
@@ -14,7 +14,7 @@ class CLI::Plugin < Thor
     method_option "git", :type => :string, :default => nil, :banner =>
       "URL of the git repo to fetch from"
     method_option "local_git", :type => :string, :default => nil, :banner =>
-      "Path of the local git repo to fetch from"
+      "Path of the local git repo to fetch from (deprecated)"
```
deivid-rodriguez, line 17: Can we also add a `SharedHelpers.major_deprecation` call, similarly to other flag deprecations we have in place in this file?

deivid-rodriguez (commented):
Looks great, just added a comment to add an explicit runtime deprecation.

I think this option is mostly discoverable through CLI help, and you already included a note about deprecation there, but we print a runtime message for other deprecated flags, so I think it does not hurt to also do it here.

deivid-rodriguez:
@ccutrer Sorry for my slow reviews, the plugin work you were doing is awesome. Whenever you're ready to continue, I am too!

ccutrer:
👍  I was on vacation for a week, then got Covid, so I'm a bit backlogged, but I promise to get back to it as soon as I can.

deivid-rodriguez:
Hope you enjoyed your vacation, and are feeling better now after Covid. Thanks so much for the work!

bundler/lib/bundler/cli/plugin.rb
ccutrer: Done (though over in plugin/installer.rb, where we implement the backwards compatibility)

deivid-rodriguez:
Since the flag is deprecated in Bundler 2 and will make commands using it fail in Bundler 3, this is making some specs fail. If they are unrelated to the setting itself, they can be changed to use `--git`.

ccutrer:
The failing specs are all ones I left in to ensure the deprecated behavior still works. I've add spec filters to them so they won't run at all under bundler 3.

deivid-rodriguez:
Sorry for the lack of feedback here.

By adding the filters, we'll lose the specs on Bundler 3, right? I think it's better to migrate the specs to use `--git` instead? We could limit usage of `--local-git` in specs to an explicit test at `bundler/spec/other/major_deprecation_spec.rb`. That's what we do for other deprecations.

ccutrer:
Ahhh, yes, I could move the specs for deprecated features there. I didn't know it existed. The specs that have the filters are already duplicates of ones that use `--git`, so we won't lose coverage on Bundler 3.

deivid-rodriguez:
Are we testing that `--git` works with local git repos? I checked superficially and didn't find those specs.

deivid-rodriguez:
I addressed what I think was left to to here.

PR #7051, opened 2023-10-10
Gem::Version avoid mutating temporary Array in attempt to fix Windows

martinemde:

Windows specs are failing with invalid version string `false.1.pre`. I don't thin this is caused by an actual bug in our code, but some inconsistency in the Windows Ruby 3.2.2 implementation.

Windows is sometimes producing `false.1.3.pre`, with the first segment being the literal `false`, when given the version string `1.1.3.pre`. My guess is that this is caused by a malfunction of map! which was recently introduced into `Gem::Version#_segments` when it is used to modify the temporary Array resulting from of scan (but only on Windows with ruby version 3.2.2).

This is an attempt at a harmless alternate version of the same code that I suspect is causing this problem. This allocates the same number Arrays and blocks, but avoids mutating the Array with `map!`, instead adding each element one at a time.

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -417,8 +417,10 @@ class Gem::Version
   end
 
   def partition_segments(ver)
-    ver.scan(/\d+|[a-z]+/i).map! do |s|
-      /\A\d/.match?(s) ? s.to_i : -s
-    end.freeze
+    segments = []
+    ver.scan(/\d+|[a-z]+/i) do |s|
+      segments << (/\A\d/.match?(s) ? s.to_i : -s)
+    end
+    segments.freeze
   end
 end


bundler/lib/bundler/injector.rb
```diff
@@ -86,7 +86,7 @@ def conservative_version(spec)
       segments = version.segments
       seg_end_index = version >= Gem::Version.new("1.0") ? 1 : 2
 
-      prerelease_suffix = version.to_s.gsub(version.release.to_s, "") if version.prerelease?
+      prerelease_suffix = version.to_s.sub(version.release.to_s, "") if version.prerelease?
       "#{version_prefix}#{segments[0..seg_end_index].join(".")}#{prerelease_suffix}"
```
martinemde, line 90: This is the triggering code in the failing test. The output here produces `false.1.pre, which means the first element of segments is `false`. This eliminates many of the possible code paths for this malfunction since it directly relies on `version.segments`.

bundler/lib/bundler/injector.rb
```diff
@@ -86,7 +86,7 @@ def conservative_version(spec)
       segments = version.segments
       seg_end_index = version >= Gem::Version.new("1.0") ? 1 : 2
 
-      prerelease_suffix = version.to_s.gsub(version.release.to_s, "") if version.prerelease?
+      prerelease_suffix = version.to_s.sub(version.release.to_s, "") if version.prerelease?
```
segiddins, line 89: I think this is going to be a change in behavior -- worth adding a test for when the release string is duplicated

martinemde:
It's only on Windows 3.2 so I expect it's a real bug. It is returning an array but with false as the first element. This is part of the diagnostic process to see if this fixes it here since I don't have a Windows machine available.

bundler/lib/bundler/injector.rb
segiddins: `1.0.0.pre.1.0.0` would be an example, I think

martinemde: But that would he incorrect to replace both.

segiddins: yes, but im saying we should test it since you're changing the code around it

martinemde: Got it. Agree.

martinemde: Ok, added a test that fails with the old version with `gsub` but not with `sub`:
```diff
-gem "lemur", "~> 3.1.pre.2023.1.1"
+gem "lemur", "~> 3.1.pre.202"
```

segiddins: another idea: `prerelease_suffix = version.to_s.delete_prefix(version.release.to_s) if version.prerelease?` or even `version._split_segments[1].join(".")`

lib/rubygems/version.rb
```diff
@@ -422,11 +422,18 @@ def freeze
   def _segments
     # segments is lazy so it can pick up version values that come from
     # old marshaled versions, which don't go through marshal_load.
+    return @segments if @segments
+
+    @segments = []
+    @version.scan(/\d+|[a-z]+/i) do |s|
```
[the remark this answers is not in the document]
martinemde: I think dynamically growing the array is taking up more memory :(

```
==> memprof.ThisPR.txt <==
Total allocated: 29.24 MB (279674 objects)
Total retained:  3.97 MB (40456 objects)
==> memprof.HEAD.txt <==
Total allocated: 29.18 MB (278954 objects)
Total retained:  3.97 MB (40456 objects)
```

martinemde:
What happened with Windows? Only windows is coming up with a totally different version number. This is clearly a Ruby bug, but I need help from a windows computer. (@simi maybe you can look at this test and see if it replicates locally?)

lib/rubygems/version.rb
```diff
@@ -420,13 +420,23 @@ def freeze
...
+    raise "How is this not nil? #{@segments.inspect}" unless @segments.nil?
+
+    @segments = Array.new
+
+    raise "How is a new array frozen? #{@segments.inspect}" if @segments.frozen?
+
+    @version.scan(/\d+|[a-z]+/i) do |s|
+      raise "How did this become frozen? #{@segments.inspect}" if @segments.frozen?
+
+      @segments << (/^\d+$/.match?(s) ? s.to_i : -s)
```
martinemde, line 437: Ok, riddle me this. How is this printing the error: 

```
/home/runner/.rubies/truffleruby-22.3.0/lib/ruby/site_ruby/3.0.3.22.3.0/rubygems/version.rb:437:in `<<': can't modify frozen Array (FrozenError)
```

on both [jruby](https://github.com/rubygems/rubygems/actions/runs/6486538501/job/17614928473?pr=7051#step:11:12) and [truffleruby](https://github.com/rubygems/rubygems/actions/runs/6486538501/job/17614928971?pr=7051#step:11:33)

I checked if it's frozen before each iteration. What Array could possibly be frozen?

martinemde: I seed that truffleruby has Gem::Version in it, but is it using this code? It can't be using both.

https://github.com/oracle/truffleruby/blob/21ad763c46422459aba15838b0168d89d91c1bef/lib/mri/rubygems/version.rb#L230

segiddins: what happens if you assign the new array to a local variable instead of of the ivar, and only assign to the ivar after freezing? im wondering if it could be concurrent modification of the version instance

martinemde: Good idea. That would also make sense about why it's intermittent. Seems this class should be atomic in its modifications, or immutable. (I did run a test with freezing on create, it does allocate a little more memory, but it's not dramatic).

martinemde: This is the branch where I immediately freeze at the end of initialize compared to the exact same code not doing that:

```
==> memprof.nofreeze.txt <==
Total allocated: 24.85 MB (206507 objects)
Total retained:  3.71 MB (38062 objects)
head: : No such file or directory
==> memprof.2.4.20.txt <==
Total allocated: 28.57 MB (255984 objects)
Total retained:  4.09 MB (42873 objects)
==> memprof.frozen.txt <==
Total allocated: 24.95 MB (207562 objects)
Total retained:  3.72 MB (38216 objects)
```

martinemde:
I wonder if #7052 will make this PR moot since it reduces usage of _segments. I'm closing this for now and we'll watch for windows to fail again.

PR #7052, opened 2023-10-11
Gem version segments optimizations

martinemde:

I was trying to solve the problem that Windows is encountering and, instead of solving it, I wondered if a regexp based solution could improve canonical sections so we didn't need to use intermediary arrays or reverse/join.

Slightly improve memory usage (fewer objects, less retained, but slightly higher total usage in my example app).
Greatly reduce the size of the code, but compress a lot of complexity into a regexp.

```
==> memprof.ThisPR.txt <==
Total allocated: 29.22 MB (279000 objects)
Total retained:  3.97 MB (40389 objects)
==> memprof.2.4.20.txt <==
Total allocated: 38.18 MB (394563 objects)
Total retained:  4.25 MB (43634 objects)
==> memprof.master.txt <==
Total allocated: 29.21 MB (279009 objects)
Total retained:  3.97 MB (40472 objects)
```

### DIFF
--- a/lib/rubygems/version.rb
+++ b/lib/rubygems/version.rb
@@ -388,31 +388,21 @@ class Gem::Version
     0
   end
 
+  # remove trailing zeros segments before first letter or at the end of the version
   def canonical_segments
     @canonical_segments ||= begin
-      numeric_segments, string_segments = _split_segments
-      canonical_segments = []
-
-      seen_non_zero = false
-      string_segments.reverse_each do |segment|
-        if seen_non_zero || (seen_non_zero = (segment != 0))
-          canonical_segments << segment
-        end
-      end
-      seen_non_zero = false
-      numeric_segments.reverse_each do |segment|
-        if seen_non_zero || (seen_non_zero = (segment != 0))
-          canonical_segments << segment
-        end
-      end
-
-      canonical_segments.reverse!
-      canonical_segments.freeze
+      # remove trailing 0 segments, using dot or letter as anchor
+      # may leave a trailing dot which will be ignored by partition_segments
+      canonical_version = @version.sub(/(?<=[a-zA-Z.])[.0]+\z/, "")
+      # remove 0 segments before the first letter in a prerelease version
+      canonical_version.sub!(/(?<=\.|\A)[0.]+(?=[a-zA-Z])/, "") if prerelease?
+      partition_segments(canonical_version)
     end
   end
 
   def freeze
     prerelease?
+    _segments
     canonical_segments
     super
   end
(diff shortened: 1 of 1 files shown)


lib/rubygems/version.rb
```diff
@@ -388,51 +392,29 @@ def <=>(other)
     0
   end
 
+  # remove trailing zeros segments before first letter or at the end of the version
   def canonical_segments
     @canonical_segments ||= begin
-      numeric_segments, string_segments = _split_segments
```
martinemde, line 393: If this is too aggressive, there is an intermediate solution where we just inline the split segments code here which saves an array for the return value and sometimes avoids slicing entirely.

lib/rubygems/version.rb
```diff
@@ -388,51 +392,29 @@ def <=>(other)
...
+      canonical_version = @version.sub(/(?<=[a-zA-Z.])[.0]+$/, "")
+      # remove 0 segments before the first letter in a prerelease version
+      canonical_version.sub!(/(?<=\.|^)[0.]+(?=[a-zA-Z])/, "") if prerelease?
+      partition_segments(canonical_version)
     end
   end
 
   def freeze
     prerelease?
+    segments
```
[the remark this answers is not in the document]
martinemde: good catch. I had to add this because canonical_segments no longer uses _segments, but forgot about the dup. Fixed.

lib/rubygems/version.rb
```diff
@@ -423,16 +412,12 @@ def _segments
     # segments is lazy so it can pick up version values that come from
     # old marshaled versions, which don't go through marshal_load.
     # since this version object is cached in @@all, its @segments should be frozen
+    @segments ||= partition_segments(@version)
+  end
 
-    @segments ||= @version.scan(/[0-9]+|[a-z]+/i).map! do |s|
+  def partition_segments(ver)
+    ver.scan(/\d+|[a-z]+/i).map! do |s|
       /^\d+$/.match?(s) ? s.to_i : -s
```
segiddins, line 420:
```suggestion
      /\A\d/.match?(s) ? s.to_i : -s
```

segiddins (approved):
nice!

lib/rubygems/version.rb
```diff
@@ -388,31 +388,20 @@ def <=>(other)
...
-      numeric_segments.reverse_each do |segment|
-        if seen_non_zero || (seen_non_zero = (segment != 0))
-          canonical_segments << segment
-        end
-      end
-
-      canonical_segments.reverse!
-      canonical_segments.freeze
+      # remove trailing 0 segments, using dot or letter as anchor
+      canonical_version = @version.sub(/(?<=[a-zA-Z.])[.0]+$/, "")
```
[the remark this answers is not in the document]
martinemde: Done.

lib/rubygems/version.rb
martinemde: This is a little slower than /\d/ which still achieves the same purpose. (it can't be anything other than a string of all digits or a string of all alphas).

segiddins: hm, I have matching an anchored `\d` being faster (which makes sense, `/\d/` is O(n) to fail matching, whereas the anchored version is constant time)

martinemde: weird, my synthetic test must not be as good as yours. Let's go with the anchored one.

Issue #6979, opened 2023-09-19
`gem uninstall` can't handle multiple versions of the same gem


deivid-rodriguez:
Specification works fine, installation as well, but uninstallation can't handle several versions of the same gem.

## Installing
```
$ gem install simplecov:0.19.0 simplecov:0.22.0
Fetching simplecov-0.19.0.gem
Successfully installed simplecov-0.19.0
Parsing documentation for simplecov-0.19.0
Installing ri documentation for simplecov-0.19.0
Done installing documentation for simplecov after 0 seconds
Fetching simplecov-0.22.0.gem
Successfully installed simplecov-0.22.0
Parsing documentation for simplecov-0.22.0
Installing ri documentation for simplecov-0.22.0
Done installing documentation for simplecov after 0 seconds
2 gems installed
```

## Uninstalling after
```
~$ gem uninstall simplecov:0.19.0 simplecov:0.22.0
Successfully uninstalled simplecov-0.22.0
~$ gem uninstall simplecov:0.19.0 simplecov:0.22.0
Gem 'simplecov' is not installed
Successfully uninstalled simplecov-0.19.0
```

Looks like it is uniqued by name at some point

_Originally posted by @hurricup in https://github.com/rubygems/rubygems/issues/6345#issuecomment-1718745111_

kstevens715:
If I use `gems_to_uninstall[dep.full_name]` on lines 171 and 172 in the code below, I can get the command `gem uninstall simplecov:0.19.0 simplecov:0.22.0` to work. But the tests fail because it breaks something else.

https://github.com/rubygems/rubygems/blob/7c5ba7152f504e72a9e83b47623023a25aac30c1/lib/rubygems/commands/uninstall_command.rb#L170-L179

deivid-rodriguez:
@kstevens715 Thanks for working on this. I suggest you open a WIP PR with your proposed changes anyways, maybe we can fix it!

kstevens715:
@deivid-rodriguez Sounds good. If I find time, I could also look into the failing tests sometime this weekend or next week.

deivid-rodriguez:
Fixed by #7063. Thanks @kstevens715!

PR #7063, opened 2023-10-14
Allow explicitly uninstalling multiple versions of same gem

kstevens715:

As described in #6979, you can install multiple versions of the same gem just fine:

```
$ gem install simplecov:0.19.0 simplecov:0.22.0
Fetching simplecov-0.19.0.gem
Successfully installed simplecov-0.19.0
Parsing documentation for simplecov-0.19.0
Installing ri documentation for simplecov-0.19.0
Done installing documentation for simplecov after 0 seconds
Fetching simplecov-0.22.0.gem
Successfully installed simplecov-0.22.0
Parsing documentation for simplecov-0.22.0
Installing ri documentation for simplecov-0.22.0
Done installing documentation for simplecov after 0 seconds
2 gems installed
```

But to uninstall both of them, you need to run the equivalent uninstall command twice:

```
~$ gem uninstall simplecov:0.19.0 simplecov:0.22.0
Successfully uninstalled simplecov-0.22.0
~$ gem uninstall simplecov:0.19.0 simplecov:0.22.0
Gem 'simplecov' is not installed
Successfully uninstalled simplecov-0.19.0
```

When just a gem name is specified without a version, there's code so the uninstall only happens once for a gem when there's multiple versions installed which makes sense because the next step lets the user choose which specific version or if they want to uninstall all versions. This PR bypasses the skip logic if a specific gem version was given so that all can be uninstalled.

### DIFF
--- a/lib/rubygems/commands/uninstall_command.rb
+++ b/lib/rubygems/commands/uninstall_command.rb
@@ -168,10 +168,10 @@ that is a dependency of an existing gem.  You can use the
     gems_to_uninstall = {}
 
     deps.each do |dep|
-      next if gems_to_uninstall[dep.name]
-      gems_to_uninstall[dep.name] = true
-
-      unless original_gem_version[dep.name] == Gem::Requirement.default
+      if original_gem_version[dep.name] == Gem::Requirement.default
+        next if gems_to_uninstall[dep.name]
+        gems_to_uninstall[dep.name] = true
+      else
         options[:version] = dep.version
       end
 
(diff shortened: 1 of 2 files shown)


kstevens715:
This PR is in draft, because although it makes the actual command succeed it causes related tests to fail. I still need to figure out why that's happening and write a new test for the issue that was fixed.

deivid-rodriguez:
Thanks so much, just approved CI to run for you, let's see how it goes!

deivid-rodriguez (approved):
Looks great!

PR #7077, opened 2023-10-19
Fix regression when installing native extensions on universal rubies

hanyang-tony:
Since #6945 the extension dir changed to Gem::BasicSpecification's implementation, we didn't hook that in rubygems_ext.rb. So for universal rubies, we ended up using the universal platform name when installing, but arch replaced platform name when checking. This lead to native extensions can never be correctly installed on universal rubies.

Hook Gem::BasicSpecifications so the behavior is consistent on installing and checking.

For universal rubies, the change from #6945 to def gem_build_complete_path is incorrect. Previously, extension_dir is invoked on the Bundler::StubSpecification, which is a child class of Specification. After the change to stub.gem_build_complete_path, @Stub is actually an instance of Gem::StubSpecification, thus gem_build_complete_path is from Gem::BasicSpecification.

The problem arise when in bundler, we replace the 'universal' part of Gem::Platform with the arch. But we still use the universal name for Gem::Specification, this is defined in 'rubygems_ext.rb'. Now the stub is using extension_dir defined in Gem::BasicSpecification, so the 'universal' part ended up being swapped by arch.

So, we install native extensions to 'extensions/universal-darwin-22/', but finding them in 'extensions/x86_64-darwin-22/'. Leading to native extension gems can never be installed on universal ruby.

Hook Gem::BasicSpecifications so the behavior is consistent on installing and checking.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -338,7 +338,7 @@ module Gem
   end
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
-  class Specification
+  class BasicSpecification
     if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
       local_platform = Platform.local
       if local_platform.cpu == "universal"
--- a/bundler/spec/bundler/specifications/foo.gemspec
+++ b/bundler/spec/bundler/specifications/foo.gemspec
@@ -1,0 +1,13 @@
+# rubocop:disable Style/FrozenStringLiteralComment
+# stub: foo 1.0.0 ruby lib
+
+# The first line would be '# -*- encoding: utf-8 -*-' in a real stub gemspec
+
+Gem::Specification.new do |s|
+  s.name = "foo"
+  s.version = "1.0.0"
+  s.loaded_from = __FILE__
+  s.extensions = "ext/foo"
+  s.required_ruby_version = ">= 2.6.0"
+end
+# rubocop:enable Style/FrozenStringLiteralComment
--- a/bundler/spec/bundler/stub_specification_spec.rb
+++ b/bundler/spec/bundler/stub_specification_spec.rb
@@ -19,6 +19,17 @@ RSpec.describe Bundler::StubSpecification do
     end
   end
 
+  describe "#gem_build_complete_path" do
+    it "StubSpecification should have equal gem_build_complete_path as Specification" do
+      spec_path = File.join(File.dirname(__FILE__), "specifications", "foo.gemspec")
+      spec = Gem::Specification.load(spec_path)
+      gem_stub = Gem::StubSpecification.new(spec_path, File.dirname(__FILE__),"","")
+
+      stub = described_class.from_stub(gem_stub)
+      expect(stub.gem_build_complete_path).to eq spec.gem_build_complete_path
+    end
+  end
+
   describe "#manually_installed?" do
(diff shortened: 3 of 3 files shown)


segiddins:
Looks good, mind adding a spec for the scenario that this addresses?

hanyang-tony:
I realized we can avoid code duplication by change `class Specification` to `class BasicSpecification`, since the former is a subclass of the latter. Also added a spec to test if `StubSpecification`'s `extensions_dir` equals `Specification`s.

hanyang-tony:
> Looks good, mind adding a spec for the scenario that this addresses?

Done :)

deivid-rodriguez:
@hanyang-tony I tried reverting the fix here, and the spec still passes. Can you have a look?

hanyang-tony:
> @hanyang-tony I tried reverting the fix here, and the spec still passes. Can you have a look?

Sorry, I was testing the wrong method. I've fixed it now. Due to the nature of how the `Gem::StubSpecification::StubLine` is loaded, I have to include a separate gemspec file instead of inline it in spec test. Please check it out.

deivid-rodriguez:
Maybe I'm crazy but the new spec is still passing without this fix?

hanyang-tony:
> Maybe I'm crazy but the new spec is still passing without this fix?

This is a fix for 'universal ruby', `Gem::Platform.local.cpu` should be 'universal'. Other versions of ruby don't have this issue.

deivid-rodriguez:
Ideally we'd fine a way to test this in our CI, to avoid regressions, even if by mocking the platform somehow. Let's ship the bug fix though for now.

hanyang-tony:
> Ideally we'd find a way to test this in our CI, to avoid regressions, even if by mocking the platform somehow. Let's ship the bug fix though for now.

It's possible to hook `Platform.local` into universal ruby in this spec. Or add a workflow that hook it first then run the specs.

hanyang-tony:
I've fixed the linting error, can we re-run the actions?

hanyang-tony:
Can we re-run the failed test? Seems like it was not caused by the changes in this PR.

deivid-rodriguez:
Ok, let's ship this as is.

> It's possible to hook Platform.local into universal ruby in this spec. Or add a workflow that hook it first then run the specs.

Would happily accept such a test to cover this 👍.

Bo98:
> It's possible to hook `Platform.local` into universal ruby in this spec

You could take inspiration from https://github.com/rubygems/rubygems/commit/11229b16c3bd400c70294855d7e9de7bdd84a2c5 which uses `simulate_platform` and `simulate_ruby_platform`.

hanyang-tony:
> > It's possible to hook `Platform.local` into universal ruby in this spec
> 
> You could take inspiration from [11229b1](https://github.com/rubygems/rubygems/commit/11229b16c3bd400c70294855d7e9de7bdd84a2c5) which uses `simulate_platform` and `simulate_ruby_platform`.

I tried to use `simulate_ruby_platform`, but it doesn't work in this case. Because the RUBY_PLATFORM isn't used directly by this spec, it's the hook in `rubygems_ext.rb` that make use of it. The hook only run once, before the spec is run, so change the platform in spec doesn't do anything.

Bo98:
Yeah it's designed to work when running a Bundle process (`install_gemfile`) rather than direct method calls. So if the problem can be translated to a real world steps to reproduce then we could test that. Was the problem revealed from doing a `bundle exec` after an install? The existing tests only tested the install.

hanyang-tony:
The regression caused bundler to search gem extension in a different directory than the one which it was installed. The existing tests could work if the tested gem has native extension.

PR #7078, opened 2023-10-19
Update vendored thor to v1.3.0

segiddins:
See https://github.com/rails/thor/releases/tag/v1.3.0


deivid-rodriguez:
You probably need to add the new files to `Manifest.txt` with `rake update_manifest`.

Issue #7076, opened 2023-10-18
Connections not reused for gem downloads


rye-stripe:
Years ago @tenderlove in [PR #869](https://github.com/rubygems/rubygems/pull/869) introduced network connection reuse, which improved the gem installation times significantly. Unfortunately the reuse does not apply to `bundler install` command downloading gems - for each gem a new tcp/tls connection is established. That's a significant amount of overhead, especially when many gems need to be fetched and the network latency to the gem source is high.

Looking at the source code of bundler, it's rather clear why reuse is not applied. The [download_gem](https://github.com/rubygems/rubygems/blob/a3696cb7fc7794a4c727db39a0c7313c050fe46e/bundler/lib/bundler/rubygems_integration.rb#L491C29-L491C29) method in bundler creates a new instance of [Gem::RemoteFetcher](https://github.com/rubygems/rubygems/blob/a3696cb7fc7794a4c727db39a0c7313c050fe46e/bundler/lib/bundler/rubygems_integration.rb#L525) for every gem that's downloaded. And it is the RemoteFetcher object that [holds](https://github.com/rubygems/rubygems/blob/a3696cb7fc7794a4c727db39a0c7313c050fe46e/lib/rubygems/remote_fetcher.rb#L84-L85) the connection pools. Throwing away the RemoteFetcher object implies throwing away the established network connections.

I put together a quick [hack](https://github.com/rubygems/rubygems/compare/master...rye-stripe:rubygems:connection-reuse?expand=1) to see how big of a difference reuse would make. Additionally, I added a test script which installs a gem with many dependancies (aws-sdk-v3) several times with the hack enabled/disabled and prints out the timings. 

```
 $ ./run.sh
Testing with REUSE=false and JOBS=1
       15.61 real         7.73 user         2.61 sys
Testing with REUSE=true and JOBS=1
        8.85 real         3.90 user         2.32 sys
Testing with REUSE=false and JOBS=10
       11.13 real         8.66 user         4.84 sys
Testing with REUSE=true and JOBS=10
        6.83 real         4.64 user         4.37 sys
```

The results suggest a close to a 2x reduction in wall-time and cpu-time.

When the latency to the gem source is high (~100ms), the difference is even more pronounced when parallel jobs are disabled.
```
 $ ./run.sh
Testing with REUSE=false and JOBS=1
      269.89 real        11.59 user         3.96 sys
Testing with REUSE=true and JOBS=1
      102.78 real         4.56 user         3.14 sys
[...]

PR #7079, opened 2023-10-19
Reuse Gem::RemoteFetcher instance in bundler

segiddins:
Closes https://github.com/rubygems/rubygems/issues/7076

Bundler will now use the same (shared) remote fetcher instance that
RubyGems uses.

This will allow installs to use a shared connection pool, which
represents a significant performance improvement on a clean install.

### DIFF
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -521,8 +521,7 @@ module Bundler
 
     def gem_remote_fetcher
       require "rubygems/remote_fetcher"
-      proxy = Gem.configuration[:http_proxy]
-      Gem::RemoteFetcher.new(proxy)
+      Gem::RemoteFetcher.fetcher
     end
 
     def build(spec, skip_validation = false)


martinemde (approved):
LOL. I had to double take... "how could this be that simple?" I didn't even notice this method on RemoteFetcher but that looks like exactly the right solution.

deivid-rodriguez (approved):
Hard to beat such a simple patch!

rye-stripe:
Thank you for the fix @segiddins! 

When I was thinking about how to solve this, I thought sharing the same fetcher object between multiple threads (when using `--jobs` is set to flag) could lead to issues. In particular, each thread would clobber the http headers set by other threads on [this line](https://github.com/rubygems/rubygems/blob/cd87b40fe18c218c3484c7a85254298e6123e9fa/bundler/lib/bundler/rubygems_integration.rb#L494-L495). Furthermore, the headers are never reset after the requests, so they may "leak" to further users of `GemFetcher`. Are these not valid concerns?

segiddins:
https://github.com/rubygems/rubygems/pull/7092

rye-stripe:
That PR addresses my concerns, thanks!

PR #7081, opened 2023-10-20
Gem::NameTuple equality ignores Gem::Platform or String variation

martinemde:

When implementing bundler checksums (#6374), I needed `NameTuple` to be equivalent whether the platform came from a `String` or a `Gem::Platform`.

My fix is based on the assumption that `NameTuple` is only for generating names. It's not intended to hold the `Gem::Platform` unaltered so that it can be used later.

This should be safe because:
* I can't find any calls to `NameTuple#platform` in the code. It's only used in the tests I modified here.
* Existing platforms could be either `String` or `Gem::Platform` object so all callers will already handle either case.
* I'm naturally optimistic.

Despite the fix, I will need to patch over this in Bundler for the foreseeable future. Is there anything you can think of to improve that situation?

### DIFF
--- a/lib/rubygems/name_tuple.rb
+++ b/lib/rubygems/name_tuple.rb
@@ -6,14 +6,12 @@
 # wrap the data returned from the indexes.
 
 class Gem::NameTuple
-  def initialize(name, version, platform="ruby")
+  def initialize(name, version, platform=Gem::Platform::RUBY)
     @name = name
     @version = version
 
-    unless platform.is_a? Gem::Platform
-      platform = "ruby" if !platform || platform.empty?
-    end
-
+    platform &&= platform.to_s
+    platform = Gem::Platform::RUBY if !platform || platform.empty?
     @platform = platform
   end
 
@@ -49,7 +47,7 @@ class Gem::NameTuple
 
   def full_name
     case @platform
-    when nil, "ruby", ""
+    when nil, "", Gem::Platform::RUBY
       "#{@name}-#{@version}"
     else
       "#{@name}-#{@version}-#{@platform}"
--- a/test/rubygems/test_gem_name_tuple.rb
+++ b/test/rubygems/test_gem_name_tuple.rb
@@ -19,14 +19,31 @@ class TestGemNameTuple < Gem::TestCase
   end
 
   def test_platform_normalization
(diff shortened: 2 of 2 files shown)


test/rubygems/test_gem_name_tuple.rb
```diff
@@ -40,4 +46,26 @@ def test_spaceship
...
+    assert_equal a, b
+    assert_equal a.hash, b.hash
+
+    a = Gem::NameTuple.new "a", Gem::Version.new(0), "ruby"
+    b = Gem::NameTuple.new "a", Gem::Version.new(0), Gem::Platform.new("ruby")
+    assert_equal a, b
+    assert_equal a.hash, b.hash
+
+    a = Gem::NameTuple.new "a", Gem::Version.new(0), "universal-darwin-23"
+    b = Gem::NameTuple.new "a", Gem::Version.new(0), Gem::Platform.new("universal-darwin-23")
```
segiddins, line 67: Please also add a test case where the to_s for platform does normalization

segiddins (approved):

lib/rubygems/name_tuple.rb
```diff
@@ -10,10 +10,8 @@ def initialize(name, version, platform="ruby")
     @name = name
     @version = version
 
-    unless platform.is_a? Gem::Platform
-      platform = "ruby" if !platform || platform.empty?
-    end
-
+    platform &&= platform.to_s
+    platform = "ruby" if !platform || platform.empty?
```
segiddins, line 14:
```suggestion
    platform = Gem::Platform::RUBY if !platform || platform.empty?
```

segiddins (approved):

lib/rubygems/name_tuple.rb
```diff
@@ -49,7 +47,7 @@ def self.null
 
   def full_name
     case @platform
-    when nil, "ruby", ""
+    when nil, "", Gem::Platform::RUBY
```
martinemde, line 50: this is mostly impossible given `initialize`. This should just be...
```suggestion
    if @platform == Gem::Platform::RUBY
```

martinemde: Seems it's possible that Marshal could load a NameTuple with a nil/empty platform, so I'll leave this as is.

PR #7086, opened 2023-10-21
Lockfile base64 checksum support. Use = as checksum algo separator.

martinemde:

Follow-up to #6374 that makes this version of Bundler a bit more future proof. I was influenced by the `Content Digest` http header which uses = signs and base64 digests. This just covers us if we choose to change it since the base64 stuff was already here and the `=` is easy now before this is released.

Handle base64 checksums and use the `=` character to separate the algo and digest.

Some checksum algorithm names are written with a hypen in them, so using `=` is less likely to cause future problems if we adopt, for example, "SHA3-512".


indirect (approved):

bundler/lib/bundler/checksum.rb
```diff
@@ -15,20 +15,24 @@ def from_gem(io, pathname, algo = DEFAULT_ALGORITHM)
...
-        Checksum.new(algo, digest, Source.new(:lock, lockfile_location))
+        algo, digest = lock_checksum.strip.split(ALGO_SEPARATOR, 2)
+        Checksum.new(algo, to_hexdigest(digest, algo), Source.new(:lock, lockfile_location))
+      end
+
+      def to_hexdigest(digest, algo = DEFAULT_ALGORITHM)
+        return digest unless algo == DEFAULT_ALGORITHM
+        return digest if digest.match?(/\A[0-9a-f]{64}\z/i)
+        if digest.match?(%r{\A[-0-9a-z_+/]{43}={0,2}\z}i)
+          digest.tr("-_", "+/") # fix urlsafe base64
```
martinemde, line 31:
```suggestion
          digest = digest.tr("-_", "+/") # fix urlsafe base64
```

martinemde (commented):
Oops

martinemde:
Added a test enforcing that the compact index may return base64 so that at least one test says that we are preserving base64 forwards compatibility.

martinemde:
As an aside here, I found myself wishing that something in the compact index response said what algorithm it was using for the checksums. If it was a header, for example, then we could almost transparently handle any algorithm from this point onwards. I would just need to make sure the parser knew to convert all base64 hashes to hex, or vice versa, before comparing them (or just compare them as binary strings preserving whatever format they are in).

At this point, if we ever wanted to support another algorithm we could just tack on another attribute at the end of the line on compact index in the format of `algo=digest` and it would compare them string to string without any processing for base64. Interesting to think about, but probably far off. SHA256 is quite adequate for what we do.

PR #7092, opened 2023-10-22
Use bundler UA when downloading gems

segiddins:
Gem::RemoteFetcher uses Gem::Request, which adds the RubyGems UA.
Gem::RemoteFetcher is used to download gems, as well as the full index.
We would like the bundler UA to be used whenever bundler is making
requests.

This PR also avoids unsafely mutating the headers hash on the shared
`Gem::RemoteFetcher.fetcher` instance, which could cause corruption or
incorrect headers when making parallel requests. Instead, we create one
remote fetcher per rubygems remote, which is similar to the connection
segregation bundler is already doing

### DIFF
--- a/bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
+++ b/bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require "rubygems/remote_fetcher"
+
+module Bundler
+  class Fetcher
+    class GemRemoteFetcher < Gem::RemoteFetcher
+      def request(*args)
+        super do |req|
+          req.delete("User-Agent") if headers["User-Agent"]
+          yield req if block_given?
+        end
+      end
+    end
+  end
+end
--- a/bundler/lib/bundler/source/rubygems.rb
+++ b/bundler/lib/bundler/source/rubygems.rb
@@ -255,11 +255,15 @@ module Bundler
         end
       end
 
-      def fetchers
-        @fetchers ||= remotes.map do |uri|
+      def remote_fetchers
+        @remote_fetchers ||= remotes.to_h do |uri|
           remote = Source::Rubygems::Remote.new(uri)
-          Bundler::Fetcher.new(remote)
-        end
+          [remote, Bundler::Fetcher.new(remote)]
+        end.freeze
+      end
+
+      def fetchers
+        @fetchers ||= remote_fetchers.values.freeze
       end
 
       def double_check_for(unmet_dependency_names)
@@ -480,7 +484,8 @@ module Bundler
       def download_gem(spec, download_cache_path, previous_spec = nil)
         uri = spec.remote.uri
         Bundler.ui.confirm("Fetching #{version_message(spec, previous_spec)}")
-        Bundler.rubygems.download_gem(spec, uri, download_cache_path)
+        gem_remote_fetcher = remote_fetchers.fetch(spec.remote).gem_remote_fetcher
+        Bundler.rubygems.download_gem(spec, uri, download_cache_path, gem_remote_fetcher)
       end
 
       # Returns the global cache path of the calling Rubygems::Source object.
--- a/bundler/spec/support/artifice/endpoint_mirror_source.rb
+++ b/bundler/spec/support/artifice/endpoint_mirror_source.rb
@@ -4,7 +4,7 @@ require_relative "helpers/endpoint"
 
 class EndpointMirrorSource < Endpoint
   get "/gems/:id" do
-    if request.env["HTTP_X_GEMFILE_SOURCE"] == "https://server.example.org/"
+    if request.env["HTTP_X_GEMFILE_SOURCE"] == "https://server.example.org/" && request.env["HTTP_USER_AGENT"].start_with?("bundler")
       File.binread("#{gem_repo1}/gems/#{params[:id]}")
     else
       halt 500
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -71,6 +71,7 @@ bundler/lib/bundler/fetcher/base.rb
 bundler/lib/bundler/fetcher/compact_index.rb
 bundler/lib/bundler/fetcher/dependency.rb
 bundler/lib/bundler/fetcher/downloader.rb
+bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
 bundler/lib/bundler/fetcher/index.rb
 bundler/lib/bundler/force_platform.rb
 bundler/lib/bundler/friendly_errors.rb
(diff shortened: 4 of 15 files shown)


martinemde (approved):
Makes sense. Glad this was caught.

bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
```diff
@@ -0,0 +1,16 @@
...
+    class GemRemoteFetcher < Gem::RemoteFetcher
+      def request(*args)
+        super do |req|
+          req.delete("User-Agent") if headers["User-Agent"]
+          yield req if block_given?
+        end
+      end
+    end
+  end
+end
```
deivid-rodriguez, line 16: Why are we adding this class? If we use a dedicated fetcher instance per remote and no longer use `Gem::RemoteFetcher.fetcher`, do we still need the cleanup?

bundler/lib/bundler/source/rubygems.rb
```diff
@@ -506,6 +506,11 @@ def extension_cache_slug(spec)
         return unless remote = spec.remote
         remote.cache_slug
       end
+
+      def fetcher_for_remote(remote)
+        fetchers.find {|f| f.remote?(remote) } ||
+          raise(ArgumentError, "No remote found")
+      end
```
deivid-rodriguez, line 513: It felt a bit off having to expose this `Fetcher#remote?` method just for this. How about

```diff
diff --git a/bundler/lib/bundler/fetcher.rb b/bundler/lib/bundler/fetcher.rb
index 844a3fa94..46ee68fed 100644
--- a/bundler/lib/bundler/fetcher.rb
+++ b/bundler/lib/bundler/fetcher.rb
@@ -214,10 +214,6 @@ def gem_remote_fetcher
       end
     end
 
-    def remote?(remote)
-      @remote == remote
-    end
-
     private
 
     def available_fetchers
diff --git a/bundler/lib/bundler/source/rubygems.rb b/bundler/lib/bundler/source/rubygems.rb
index e72369be8..428ddad24 100644
--- a/bundler/lib/bundler/source/rubygems.rb
+++ b/bundler/lib/bundler/source/rubygems.rb
@@ -255,11 +255,16 @@ def unmet_deps
         end
       end
 
-      def fetchers
-        @fetchers ||= remotes.map do |uri|
+      def remote_fetchers
+        @remote_fetchers ||= remotes.map do |uri|
           remote = Source::Rubygems::Remote.new(uri)
-          Bundler::Fetcher.new(remote)
-        end
+
+          [remote, Bundler::Fetcher.new(remote)]
+        end.to_h
+      end
+
+      def fetchers
+        @fetchers ||= remote_fetchers.values
       end
 
       def double_check_for(unmet_dependency_names)
@@ -480,7 +485,7 @@ def cache_path
       def download_gem(spec, download_cache_path, previous_spec = nil)
         uri = spec.remote.uri
         Bundler.ui.confirm("Fetching #{version_message(spec, previous_spec)}")
-        Bundler.rubygems.download_gem(spec, uri, download_cache_path, fetcher_for_remote(spec.remote).gem_remote_fetcher)
+        Bundler.rubygems.download_gem(spec, uri, download_cache_path, remote_fetchers[spec.remote].gem_remote_fetcher)
       end
 
       # Returns the global cache path of the calling Rubygems::Source object.
@@ -506,11 +511,6 @@ def extension_cache_slug(spec)
         return unless remote = spec.remote
         remote.cache_slug
       end
-
-      def fetcher_for_remote(remote)
-        fetchers.find {|f| f.remote?(remote) } ||
-          raise(ArgumentError, "No remote found")
-      end
     end
   end
 end
```

bundler/spec/support/artifice/endpoint_mirror_source.rb
```diff
@@ -4,7 +4,7 @@
 
 class EndpointMirrorSource < Endpoint
   get "/gems/:id" do
-    if request.env["HTTP_X_GEMFILE_SOURCE"] == "https://server.example.org/"
+    if request.env["HTTP_X_GEMFILE_SOURCE"] == "https://server.example.org/" && request.env["HTTP_USER_AGENT"].start_with?("bundler")
```
deivid-rodriguez, line 7: This is tied to `spec/realworld/gemfile_source_header_spec.rb` spec, right? And you have expanded it to also check that a proper Bundler user agent is set. It's still not a very straighforward spec, but we I would add this to the spec description: 

https://github.com/rubygems/rubygems/blob/d45b23d886e7eaa351d7214d51e595d505e8f0e8/bundler/spec/realworld/gemfile_source_header_spec.rb#L20

and maybe also rename the file.

deivid-rodriguez (commented):
I added a question and some suggestions!

bundler/lib/bundler/source/rubygems.rb
duckinator: I second this suggestion. The current code is basically using an array as a hash, so using an actual hash should simplify the code considerably.

bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
segiddins: needed because `Gem::RemoteFetcher` only `add`s the supplied headers, and we need to be able to completely _set_ the UA header, otherwise all the parsing we do of bundler UAs won't work

duckinator (approved):
This looks good to me, with your most recent changes. 🙂

bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
deivid-rodriguez: I don't fully understand. Previously we were using a single instance and were using the `headers` setter to set headers, right? Was that incorrect already? Is there any test that fails if we use don't add this class, in case that illustrates its need?

bundler/lib/bundler/source/rubygems.rb
deivid-rodriguez: This was fixed, thanks @segiddins!

bundler/lib/bundler/fetcher/gem_remote_fetcher.rb
segiddins: Yes, the artifice test. The header needs to be cleared so the custom UA is set rather than appended to the rubygems UA

martinemde: I don't know for sure that this class is the best approach, but I feel confident that it won't be difficult to refactor later, if needed, since it adds no new interface. We can revisit, but I think this fix should go in so the parallel fetching code is not broken on master.

deivid-rodriguez: Cool, cool, I'm just curious but did not mean to block this fix.

Issue #7058, opened 2023-10-13
With auto-user-install feature, no warning about install directory not being in PATH


duckinator:
### Describe the problem as clearly as you can

If the installation directory isn't in my `$PATH`, RubyGems doesn't warn me -- but I thought it previously did that.

### Did you try upgrading RubyGems?

Using master, as of 8072d99.

### Post steps to reproduce the problem

- Install RubyGems from source (I know 8072d99 is affected).
- _Do not_ set `GEM_HOME`.
- Ensure you don't have write access to the normal install directory.
- `gem install boop`

### Which command did you run?

`gem install boop`

### What were you expecting to happen?

I expected it to say it defaulted to user installation, _then_ say the bin path isn't writable.

### What actually happened?

```
puppy@orthrus:~$ gem install boop
Defaulting to user installation because default installation directory (/usr/local/lib/ruby/gems/3.1) is not writable.
Fetching boop-0.5.0.gem
Successfully installed boop-0.5.0
1 gem installed
puppy@orthrus:~$ echo $PATH
/home/puppy/.cargo/bin:/home/puppy/bin:/home/puppy/.local/bin:/home/puppy/.gem/ruby/bin:/home/puppy/.npm-global/bin:/home/puppy/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
puppy@orthrus:~$
```

### Run `gem env` and paste the output below

```
Defaulting to user installation because default installation directory (/usr/local/lib/ruby/gems/3.1) is not writable.
RubyGems Environment:
  - RUBYGEMS VERSION: 3.5.0.dev
  - RUBY VERSION: 3.1.4 (2023-03-30 patchlevel 223) [amd64-freebsd13]
  - INSTALLATION DIRECTORY: /usr/home/puppy/.gem/ruby/3.1
  - USER INSTALLATION DIRECTORY: /home/puppy/.gem/ruby/3.1
  - RUBY EXECUTABLE: /usr/local/bin/ruby31
  - GIT EXECUTABLE: /usr/local/bin/git
  - EXECUTABLE DIRECTORY: /usr/home/puppy/.gem/ruby/3.1/bin
  - SPEC CACHE DIRECTORY: /home/puppy/.local/share/gem/specs
  - SYSTEM CONFIGURATION DIRECTORY: /usr/local/etc
  - RUBYGEMS PLATFORMS:
     - ruby
     - amd64-freebsd-13
[...]

Issue #7083, opened 2023-10-20
--build-root option is not respected anymore


voxik:
This is command we regularly use on Fedora during build of PRM packages:

~~~
$ gem install -V --local --build-root . --force --document=ri,rdoc json-2.6.3.gem
Defaulting to user installation because default installation directory (/usr/share/gems) is not writable.
WARNING:  You build with buildroot.
  Build root: /builddir/build/BUILD/json-2.6.3
  Bin dir: /builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/bin
  Gem home: /builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby
  Plugins dir: /builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/plugins
/builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/gems/json-2.6.3/CHANGES.md
/builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/gems/json-2.6.3/LICENSE
/builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/gems/json-2.6.3/README.md
/builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/gems/json-2.6.3/VERSION
/builddir/build/BUILD/json-2.6.3/builddir/.local/share/gem/ruby/gems/json-2.6.3/ext/json/ext/fbuffer/fbuffer.h

... snip ...
~~~

However, this should not install anything into `~/.local`, but is should create the regular `/usr/share/gems` in the directory specified by the `--build-root` option.

@duckinator This is caused by #5327

Unless there is some straight forward fix, I'd suggest to revert the PR temporary (or I'll need to revert this feature for Fedora until this is fixed :/ )

voxik:
Hm, OTOH, maybe we should change stuff on Fedora side. Because this really changes defaults. But how?

voxik:
This might be the biggest difference actually:

https://github.com/rubygems/rubygems/blob/54a7a7abc6bbb5b9c0528ad4144467be865506a1/lib/rubygems/path_support.rb#L38

The thing is that the writability of the directory was never checked at this place. And while the `Gem.default_dir` is not writable, the `@build_root + Gem.default_dir` would be writable (in theory, because the directory structure does not likely exists at this stage, not sure).

duckinator:
[...]

PR #7100, opened 2023-10-24
Less exceptional user install

voxik:
This is an attempt to make the user install less exceptional. It does several things:

1) It removes the arbitrary limitation that the `--install-dir` and `--user-install` can't be specified at once on command line. This was not the case internally anyway. Now the `--install-dir` simple overrides the `--user-install`.
2) It refactors the code in an attempt to make it more readable. Therefore now if `--build-root` option is specified, it is applied to every possible path, including the user install case. I am not 100% convinced that this really is the right thing, OTOH it should not harm anything, because using `--build-root` is very specific and use it together with `--user-install` would be exceptional

This is an attempt to address some concerns I have risen in #7083

### DIFF
--- a/lib/rubygems/commands/install_command.rb
+++ b/lib/rubygems/commands/install_command.rb
@@ -136,13 +136,6 @@ You can use `i` command instead of `install`.
     "#{program_name} [options] GEMNAME [GEMNAME ...] -- --build-flags"
   end
 
-  def check_install_dir # :nodoc:
-    if options[:install_dir] && options[:user_install]
-      alert_error "Use --install-dir or --user-install but not both"
-      terminate_interaction 1
-    end
-  end
-
   def check_version # :nodoc:
     if options[:version] != Gem::Requirement.default &&
        get_all_gem_names.size > 1
@@ -162,7 +155,6 @@ You can use `i` command instead of `install`.
 
     ENV.delete "GEM_PATH" if options[:install_dir].nil?
 
-    check_install_dir
     check_version
 
     load_hooks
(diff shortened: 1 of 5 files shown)


deivid-rodriguez (approved):
This makes sense to me, I think!

duckinator:
In principle this makes sense to me and I'm in favor of it. Looking at the code, I don't see any obvious issues. 

I'll poke at it locally later and see if everything seems good.

duckinator:
@deivid-rodriguez if you think this is good to go, feel free to merge. I'm just looking at it from my phone atm, so couldn't try running it myself, is all.

voxik:
Funnily enough, I have discovered `--vendor` option today (after a decade :see_no_evil:) and it also clashes with `--user-install` [in non transparent way](https://github.com/rubygems/rubygems/blob/9bdad37f184deb1e099c6c35c94f2ccdee0acdae/lib/rubygems/install_update_options.rb#L59):

~~~
$ gem install gem2rpm --user-install --vendor
ERROR:  Use --install-dir or --user-install but not both
~~~

This PR should also address this issue.

deivid-rodriguez:
@duckinator I just looked at this superficially, I'd appreciate if you have a closer look 👍.

duckinator:
I can find no difference in behavior between 3.4.21 (latest release) and 313b1c5 (latest commit on this branch), so I think we're good to go. 👍

duckinator:
Thanks for the PR, @voxik. This is quite a nice improvement. ^.^

duckinator:
@voxik turns out you also fixed #7058 with this. 🙂

PR #7115, opened 2023-10-26
Improve recent `force_ruby_platform` fix

deivid-rodriguez:

The developer problem is small:

* Potential to make code a bit simpler.
* Potential to make both spec wording and implementation easier to understand.

This PR iterates on #6936.

Feedback in the PR was not blocking, but I think it has merit to be introduced.

### DIFF
--- a/bundler/lib/bundler/gem_helpers.rb
+++ b/bundler/lib/bundler/gem_helpers.rb
@@ -48,6 +48,13 @@ module Bundler
     end
     module_function :select_best_platform_match
 
+    def force_ruby_platform(specs)
+      matching = specs.select {|spec| spec.match_platform(Gem::Platform::RUBY) && spec.force_ruby_platform! }
+
+      sort_best_platform_match(matching, Gem::Platform::RUBY)
+    end
+    module_function :force_ruby_platform
+
     def sort_best_platform_match(matching, platform)
       exact = matching.select {|spec| spec.platform == platform }
       return exact if exact.any?
--- a/bundler/lib/bundler/lazy_specification.rb
+++ b/bundler/lib/bundler/lazy_specification.rb
@@ -133,6 +133,10 @@ module Bundler
       " #{source.revision[0..6]}"
     end
 
+    def force_ruby_platform!
+      @force_ruby_platform = true
+    end
+
     private
 
     def use_exact_resolved_specifications?
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -200,9 +200,11 @@ module Bundler
 
     def specs_for_dependency(dep, platform)
       specs_for_name = lookup[dep.name]
-      target_platform = dep.force_ruby_platform ? Gem::Platform::RUBY : (platform || Bundler.local_platform)
-      matching_specs = GemHelpers.select_best_platform_match(specs_for_name, target_platform)
-      matching_specs.each {|s| s.force_ruby_platform = true } if dep.force_ruby_platform
+      matching_specs = if dep.force_ruby_platform
+        GemHelpers.force_ruby_platform(specs_for_name)
+      else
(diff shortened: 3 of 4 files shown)


bundler/lib/bundler/gem_helpers.rb
```diff
@@ -48,6 +48,13 @@ def select_best_platform_match(specs, platform)
     end
     module_function :select_best_platform_match
 
+    def force_ruby_platform(specs)
+      matching = specs.select {|spec| spec.force_ruby_platform = true if spec.match_platform(Gem::Platform::RUBY) }
```
segiddins, line 52: using a `select` here where the return value is coming from a setter feels a bit non-obvious to me -- maybe expand to

```
select do |spec|
  next unless spec.match_platform(Gem::Platform::RUBY)
  spec.force_ruby_platform = true
  true
end
```

segiddins (approved):

bundler/lib/bundler/gem_helpers.rb
deivid-rodriguez: Agreed, it was feeling too cryptic when I was writing it. I like your approach better, will switch to it.

deivid-rodriguez: I'm changing this but now it feels pretty neat 😆, let's ask for a third opinion. @martinemde?

martinemde: Oooh, count me in!

The more I look at it though, the more I lean towards:

```ruby
select do |spec|
  spec.match_platform(Gem::Platform::RUBY) && spec.force_ruby_platform = true
end
```

What I like about it is that it preserves this as a "positive" case match which is what select is all about. "Select specs that match platform ruby && where force_ruby_platform = true returns true".

Select should contain a logical boolean operation in the block, so this preserves that characteristic.

deivid-rodriguez: Love it!

deivid-rodriguez: That does have the potential confusion of `==` vs `=`, so maybe RuboCop will complain, let's see.

deivid-rodriguez: Ugg, RuboCop does not complain but I'm on the fence again because of the potential assignment vs comparison confusion 😅.

martinemde: That's fair. I wish there was a declarative `force_ruby_platform!` (and I would make it only apply to matching ruby platforms).

I like both Sam's and your solutions just fine though. I'm always game to debate code styles but any of them will be fine. The most verbose and intentional is maybe this, but it's inefficient.:

```ruby
specs
  .select {|spec| spec.match_platform(Gem::Platform::RUBY) }
  .each {|spec| spec.force_ruby_platform = true }
```

Absolutely no confusion there, but it's a trade-off for sure.

martinemde: Just add another checkmark to *bundler should ship with rubygems and be upgraded in lockstep* because all these little compromises would go away when you can change rubygems code as needed.

martinemde: oh my, I didn't realize the context here. This is for spec support and no code uses this yet. I'd just be verbose here with the clearest one since it runs once, haha.

deivid-rodriguez: Mmmm, no this is not only for specs, it's part of the bug fix in #6936. I just took [your suggestion](https://github.com/rubygems/rubygems/pull/6936#discussion_r1314320125) from the original PR since I liked it :)

I went with your declarative idea in the end.

martinemde (approved):
Neat, I forgot all about that older comment that I made on this code. Glad it could help revived. As long as the modified test still covers the case we're trying to fix, then this looks good.

simi (approved):

PR #7122, opened 2023-10-29
Opaque ETag using DigestIO to calculate checksums

martinemde:

Building @simi's branch for opaque etags (related to #6632) and merging in my work with DigestIO in the CompactIndex client at #7039 (which can be closed if this merges).

I merged the two branches so we could make Content-Digest header support automatic based on supported digests. I also formalized some of the logic for Content-Digest headers to conform to [Structured Fields in HTTP Headers RFC 8941](https://www.rfc-editor.org/rfc/rfc8941#name-introduction).


bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,46 +1,58 @@
...
+      def update(local_path, remote_path, local_etag_path, retrying = nil)
         headers = {}
 
         local_temp_path = local_path.sub(/$/, ".#{$$}")
         local_temp_path = local_temp_path.sub(/$/, ".retrying") if retrying
+        local_etag_temp_path = local_temp_path.sub(/$/, ".etag.tmp")
+        local_etag_temp_path = local_temp_path.sub(/$/, ".etag.tmp.retrying") if retrying
         local_temp_path = local_temp_path.sub(/$/, ".tmp")
 
+        headers["Want-Content-Digest"] = SUPPORTED_DIGESTS.keys.join(",")
```
martinemde, line 38: I added this if we want to support telling the server which content digests we will be able to parse and compare, which could make future updates easier.

https://webconcepts.info/concepts/http-header/Want-Content-Digest

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,46 +1,60 @@
 # frozen_string_literal: true
 
 require_relative "../vendored_fileutils"
+require "rubygems/package/digest_io"
```
simi, line 4: I'm not sure it is currently possible to rely on this being present since combination of Bundler and RubyGems could be almost anything supported.

martinemde: DigestIO has been around for at least 11 years. It's supported, though requiring it this way does not work. We have to `require "rubygems/package"`.

martinemde: Since we don't normally have to require Gem::Package, I wonder if it would be better to make our own DigestIO here. It's an extremely simple class.

bundler/lib/bundler/compact_index_client.rb
```diff
@@ -76,36 +76,37 @@ def dependencies(names)
...
       end
-      @updater.update(local_path, url(remote_path))
+      @updater.update(local_path, url(remote_path), local_etag_path)
     end
 
     def update_info(name)
       Bundler::CompactIndexClient.debug { "update_info(#{name})" }
       path = @cache.info_path(name)
-      checksum = @updater.checksum_for_file(path)
+      etag_path = @cache.info_etag_path(name)
```
segiddins, line 98: nit: this is unused until the last line of the method, move down

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,46 +1,60 @@
...
+
         # first try to fetch any new bytes on the existing file
-        if retrying.nil? && local_path.file?
-          copy_file local_path, local_temp_path
+        if retrying.nil? && local_path.file? && local_etag_path.file?
+          copy_file local_path, local_temp_path, digests
+          copy_file local_etag_path, local_etag_temp_path
 
-          headers["If-None-Match"] = etag_for(local_temp_path)
+          headers["If-None-Match"] = local_etag_temp_path.read
```
segiddins, line 50: should we `strip!` here?

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,46 +1,60 @@
...
+      def update(local_path, remote_path, local_etag_path, retrying = nil)
         headers = {}
 
         local_temp_path = local_path.sub(/$/, ".#{$$}")
         local_temp_path = local_temp_path.sub(/$/, ".retrying") if retrying
+        local_etag_temp_path = local_temp_path.sub(/$/, ".etag.tmp")
+        local_etag_temp_path = local_temp_path.sub(/$/, ".retrying") if retrying
         local_temp_path = local_temp_path.sub(/$/, ".tmp")
 
+        headers["Want-Content-Digest"] = WANT_DIGEST
```
segiddins, line 40: I'm not seeing `Want-Content-Digest` listed at all on MDN, and it says `Want-Digest` is deprecated https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Want-Digest

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -50,73 +64,134 @@ def update(local_path, remote_path, retrying = nil)
...
+          algorithm.downcase!
+          response_digests[algorithm] = value
+        end
+        response_digests
+      end
+
+      # Unwrap surrounding colons (byte sequence) or quotes (string).
+      # The wrapping characters must be matched or we return nil.
+      # Converts hex to base64 based on field type assumptions.
+      def parse_structured_value(value)
```
segiddins, line 178: this definitely needs some tests

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -50,73 +64,134 @@ def update(local_path, remote_path, retrying = nil)
...
             FileUtils.mv(local_temp_path, local_path)
+            FileUtils.mv(local_etag_temp_path, local_etag_path)
           end
+
           return nil
         end
 
         if retrying
-          raise MisMatchedChecksumError.new(remote_path, etag, etag_for(local_temp_path))
+          # TODO
```
segiddins, line 100: TODO

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: It's only a [draft](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers) 🤷‍♂️

bundler/lib/bundler/shared_helpers.rb
```diff
@@ -193,6 +193,21 @@ def digest(name)
       Digest(name)
     end
 
+    def checksum_for_file(path, digest)
```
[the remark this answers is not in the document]
martinemde: We only use it in one place only for the MD5 of the version info files for the compact index. The server side of the compact index is hard coded to MD5 as well. The most correct thing is probably to move this completely into the compact index client and only generalize if we need to later.

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: `strip!` and `chomp!` both return nil if there's nothing to strip or chomp, which makes this pretty ugly:
```ruby
etag = local_etag_path.read
etag.chomp!
headers["If-None-Match"] = etag
```
this is one line but the to_proc might negate the benefit of using `chomp!` over `chomp`:
```ruby
headers["If-None-Match"] = local_etag_path.read.tap(&:chomp!)
```

martinemde:
Hey @segiddins & @simi, I refactored this pretty significantly. I mostly replaced the giant method since two smaller methods. The file is the same length as it was before this commit. I also cut down on the pending temporary files, clearly separated what was called "retrying" logic, and removed the nested called to update which made the file management weird.

The behavior changes a little, but I think for the best: In the case where a range request is ignored or not available, it no longer attempts the full request twice (it was requesting a range `0-` which is the same as a full request, then it would do the same thing again, whereas the intention, I think, was to try a partial request, then try the full request).

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: We could totally skip this header. It's just future proofing.

simi:
@martinemde this is on my list, I'll do some local testing first.

martinemde:
I'm still working on fixing and adding tests. I think the mocked ones are a good way to test the more subtle behavior.

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -76,6 +84,16 @@ def info_path(name)
         end
       end
 
+      def info_etag_path(name)
+        name = name.to_s
+        if /[^a-z0-9_-]/.match?(name)
```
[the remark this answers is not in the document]
martinemde: What is the purpose of this logic? @simi?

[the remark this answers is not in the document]
simi: I was about to keep it the same as other data we store https://github.com/rubygems/rubygems/blob/f8e87eba76526d2609d543eb7774711d345edc61/bundler/lib/bundler/compact_index_client/cache.rb#L71. Not every gem name is unique when transformed into file name, this is to bandaid this AFAIK.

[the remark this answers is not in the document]
martinemde: So then is @segiddins suggesting that we use MD5 of every name instead of having both? Could we _just_ do the MD5 of the name without the prefix?

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: I've now removed this feature. We could re-add later.

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+
+      # set the digests that will be verified at the end
+      def digests=(expected_digests)
+        @expected_digests = expected_digests
+
+        @digests = if @expected_digests.nil?
+          nil
+        elsif @digests
+          @digests.slice(*@expected_digests.keys)
```
martinemde, line 62:
```suggestion
        if @expected_digests.nil?
          @digests = nil
        elsif @digests
          @digests = @digests.slice(*@expected_digests.keys)
```

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,129 @@
...
+        new(path) do |file|
+          file.initialize_digests
+          file.copy
+          yield file
+        end
+      end
+
+      def self.write(path, data, digests = nil)
+        return unless data
+        new(path, digests) do |file|
```
martinemde, line 27:
```suggestion
        new(path) do |file|
          file.digests = digests
```

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,129 @@
...
+          file.write!(data)
+        end
+      end
+
+      attr_reader :original_path, :path
+
+      def initialize(original_path, digests = nil, &block)
+        @original_path = original_path
+        @path = original_path.sub(/$/, ".#{$$}.tmp")
+        self.digests = digests
```
martinemde, line 37:
```suggestion
      def initialize(original_path, &block)
        @original_path = original_path
        @path = original_path.sub(/$/, ".#{$$}.tmp")
```

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,129 @@
...
+          initialize_digests(@expected_digests.keys)
+        end
+      end
+
+      def digests?
+        @digests&.any?
+      end
+
+      def open(*args)
+        raise "File is closed" if @closed
```
martinemde, line 73:
```suggestion
        raise "Cannot reopen closed file" if @closed
```

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,129 @@
...
+      end
+
+      def verify
+        return true unless @expected_digests && digests?
+        @base64digests = @digests.transform_values!(&:base64digest)
+        @digests = nil
+        @base64digests.all? {|algo, digest| @expected_digests[algo] == digest }
+      end
+
+      def commit
```
martinemde, line 115:
```suggestion
      def commit
        raise "Cannot commit closed file" if @closed
```

bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -1,121 +1,103 @@
...
-        def initialize(path, server_checksum, local_checksum)
-          @path = path
-          @server_checksum = server_checksum
-          @local_checksum = local_checksum
-        end
-
-        def message
-          "The checksum of /#{@path} does not match the checksum provided by the server! Something is wrong " \
-            "(local checksum is #{@local_checksum.inspect}, was expecting #{@server_checksum.inspect})."
+      class MismatchedChecksumError < Error
```
martinemde, line 6:
```suggestion
      class MismatchedChecksum < Error
```

martinemde (commented):
Nitpicks of my own code that I want to remember to fix.

bundler/lib/bundler/compact_index_client/cache_file.rb
```diff
@@ -0,0 +1,134 @@
...
+        raise "Cannot commit closed file" if @closed
```
segiddins, line 120: nit: raise specific error

segiddins (approved):
Nice! took me a while to read through the changes, but I think this is _both_ good future-proofing, as well as fully backwards compatible. well done!

bundler/lib/bundler/compact_index_client.rb
```diff
@@ -5,7 +5,13 @@
 
 module Bundler
   class CompactIndexClient
+    # NOTE: MD5 is here not because we expect a server to respond with it, but
```
simi, line 8: Should this be TODO instead?

simi:
Thanks a lot for taking over @martinemde. Well done :muscle:.

bundler/lib/bundler/compact_index_client.rb
martinemde: I'm not sure. I mentioned this temporary md5 thing in a bunch of places so maybe we'll remember to get it.

simi: I mean I do rarely scan codebase for `NOTE`, but I do often for `TODO`. I'll try to keep this in my mind as well for now as a candidate to update at some point.

martinemde:
@simi and @segiddins thanks for reviewing and supporting this, and especially thanks @simi for jumping on a call to go over this code. Now let's just hope it rolls out smoothly.

bundler/lib/bundler/compact_index_client.rb
martinemde: Ah, I don't usually leave TODO in PRs, but that's a good point if you actually search for it. We could update.

PR #7123, opened 2023-10-30
Relax matching pattern for rake version

hsbt:

I released [Rake 13.1.0](https://github.com/ruby/rake/releases/tag/v13.1.0) in last weekend. After that, some of bundler example is failing because they used `13.0` pattern for checking version number.

I replaced it with `\d` regex.


simi:
@deivid-rodriguez uhh, would be good to change the test to not rely on rack latest version somehow to not break those tests again with rake 14 :pray:

deivid-rodriguez:
Sure, or even better make those tests don't rely on the release schedule of real gems. Any improvements there super welcome!

PR #7128, opened 2023-11-01
Support again Pathname objects in the replaced require

mtasaka:
So I am trying to try shoulda-matches ( https://github.com/thoughtbot/shoulda-matchers/commit/a5e3133296b5443340e92f06039f751a92935e38 ) test suite and it fails a lot like:

```
An error occurred while loading ./spec/unit/shoulda/matchers/routing/route_matcher_spec.rb.
Failure/Error: require 'unit_spec_helper'

NoMethodError:
  undefined method `tr' for an instance of Pathname
# /builddir/build/BUILD/spec/support/unit/rails_application.rb:170:in `load_environment'
# /builddir/build/BUILD/spec/support/unit/rails_application.rb:27:in `load'
# /builddir/build/BUILD/spec/support/unit/load_environment.rb:8:in `<top (required)>'
# /builddir/build/BUILD/spec/unit_spec_helper.rb:1:in `require_relative'
# /builddir/build/BUILD/spec/unit_spec_helper.rb:1:in `<top (required)>'
# ./spec/unit/shoulda/matchers/routing/route_matcher_spec.rb:1:in `<top (required)>'
# /usr/share/gems/gems/bundler-2.5.0.dev/libexec/bundle:37:in `block in <top (required)>'
# /usr/share/gems/gems/bundler-2.5.0.dev/libexec/bundle:29:in `<top (required)>'
```

`unit/rails_application.rb: 170` is  https://github.com/thoughtbot/shoulda-matchers/blob/a5e3133296b5443340e92f06039f751a92935e38/spec/support/unit/rails_application.rb#L170

So it seems that it is expected that replaced require supports Pathname objects, once it was modified so:
https://github.com/rubygems/rubygems/pull/6837

But again it seems this support got broken by:
https://github.com/rubygems/rubygems/pull/7056

This patch enables support again for Pathname objects in the replaced require, as is done as https://github.com/rubygems/rubygems/pull/6837 , which is a bit modified as https://github.com/rubygems/rubygems/pull/6839

### DIFF
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -244,6 +244,7 @@ module Bundler
       [::Kernel.singleton_class, ::Kernel].each do |kernel_class|
         kernel_class.send(:alias_method, :no_warning_require, :require)
         kernel_class.send(:define_method, :require) do |name|
+          name = File.path(name)
           if message = ::Gem::BUNDLED_GEMS.warning?(name, specs: specs) # rubocop:disable Style/HashSyntax
             warn message, :uplevel => 1
           end


voxik:
> The question is if this really is the right place for fix, because I'd say that plain Ruby `require` does not support `Pathname`:
> 
> ```
> $ ruby --disable-gems -rpathname -e '
> pa = Pathname.new("test")
> require pa
> '
> -e:3:in `require': no implicit conversion of Pathname into String (TypeError)
> 	from -e:3:in `<main>'
> ```
> 
> So why RubyGems/Bundler should change it? I don't find the arguments in #6837 well founded.

voxik:
OTOH, RubyGems likely keeps accepting Pathname:

https://github.com/rubygems/rubygems/blob/f8e87eba76526d2609d543eb7774711d345edc61/lib/rubygems/core_ext/kernel_require.rb#L40

But then this deserve better test case then this removed:

https://github.com/rubygems/rubygems/pull/7056/commits/3ba4259e79527c35ab0b2339e9d401dd86e9f48e#diff-40efe173e5c278f1610cc996fe00bbfb5ac63b4bd5fa598572ae94ad153e25e5L1676

martinemde:
Arguably Ruby should be updated to support Pathname. The argument to require is semantically a "pathname".

However, given that RubyGems has supported this in the past and that a special argument to ruby is needed to avoid loading rubygems, I think we should both fix it here and request Ruby support.

Ruby with rubygems is the standard;  pure Ruby is the exception. I think they should both act the same. Regardless, we shouldn't break previously expected behavior unintentionally.

martinemde (changes requested):
This needs a test so that it doesn't get broken again.

hsbt:
@mtasaka Thanks. I fixed at https://github.com/ruby/ruby/commit/e34e8b93f8fac3ef40ab5ed8672fa003f3b4d9c0.

>Arguably Ruby should be updated to support Pathname. The argument to require is semantically a "pathname".

@martinemde Feel free to open issue at https://bugs.ruby-lang.org/.

voxik:
https://bugs.ruby-lang.org/issues/19985

rubyFeedback:
I think the big problem voxik pointed out is that the same ruby code would behave differently, which creates an inconsistency, and on that note I concur - that inconsistency should not exist. I can't tell which variant is the correct one as I stopped using Pathname a long time ago, due to it confusing me quite a lot.

PR #7133, opened 2023-11-04
Devcontainer / Github Codespace

colby-swandale:

Next week, contributors from the RubyGems/RubyGems.org/Bundler team will be attending the [RubyConf community day](https://rubyconf.org/faq#community-day), a day where attendees can meet face-to-face with Ruby Open-Source maintainers and learn the ins-and-outs of projects and possibly look to make some contributions.

To cut down on setup time and having maintainers fidgeting with various differing environments, I want to provide a way to  create environments super easily and have people onboarded with as little hassle as possible. 

This Pull Request adds [https://containers.dev](devcontainer ) / [GitHub Codespace](https://docs.github.com/en/codespaces/overview) configuration that automates:
* Setting up an isolated development environment
* Installs the necessary OS / Language / Library packages
* Setups the project to be ready for development.

After setup has been completed, users can then start writing code and testing their changes as if it were running on a MacBook / Linux Laptop.

## Notes

The devcontainer setup is fairly straightforward, but I made a trade-off with the Docker image the devcontainer is loaded into. Microsoft ships a [Ruby devcontainer](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/ruby) image, which unfortunately uses the RVM tool to install and manage the Ruby Environment. We could manipulate the environment to not involve RVM, but I felt it was too confusing, and instead we should opt to use the well-known [base Ruby image](https://hub.docker.com/_/ruby/).

The trade-off for this means that we need to set up the devcontainer ourselves by:
* installing the required packages
* create a non-root user

But I'm confident this does not add much complexity that places a burden on maintainers and any issues couldn't be easily resolved.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,27 @@
+ARG RUBY_VERSION=3.2
+
+FROM ruby:${RUBY_VERSION}-alpine
+
+ENV USERNAME=vscode \
+    UID=1000 \
+    GID=1000
+
+# Install the git-credential-manager package via the dotnet tooling to 
+RUN apk update && apk add --no-cache \
+    github-cli \
+    git \
+    build-base \
+    bash \
+    mandoc \
+    man-pages \
+    sudo
+
+# create non-root group and user
+RUN addgroup -g $GID $USERNAME \
+    && adduser -s /bin/bash -u $UID -G $USERNAME $USERNAME --disabled-password --gecos ""
+
+# set sudo permissions for vscode user
+RUN echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME
+RUN chmod 0440 /etc/sudoers.d/$USERNAME
+
+USER $USERNAME
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+{
+	"name": "RubyGems",
+	"build": {
+		"dockerfile": "Dockerfile"
+	},
+	"remoteUser": "vscode",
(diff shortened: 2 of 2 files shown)


colby-swandale:
Anyone looking to test, you can create one from this branch via this link:

* https://github.com/codespaces/new?skip_quickstart=true&machine=basicLinux32gb&repo=614070&ref=colby%2Fdevcontainer&geo=UsWest&devcontainer_path=.devcontainer%2Fdevcontainer.json

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+{
```
simi, line 2: Is it common to use this indentation in `devcontainer.json` file?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+FROM ruby:${RUBY_VERSION}-alpine
+
+ENV USERNAME=vscode \
+    UID=1000 \
+    GID=1000
+
+# Install the git-credential-manager package via the dotnet tooling to 
+RUN apk update && apk add --no-cache \
+    github-cli \
```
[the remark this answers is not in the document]
colby-swandale: It's not strictly needed per se. The container will work fine with it, but it's handy to have around when working with Github Repositories & Codespaces. It doesn't feel harmful to include it by default.

.devcontainer/devcontainer.json
colby-swandale: This was generated from the VSCode Devcontainer extension, we can change it to match the rest of the RubyGems project.

colby-swandale: Updated the indentation to be 2 spaces.

simi (approved):
Code-wise looks ok, but I have not tried. If you can run rubygems tests and random bundler spec in this container it is good to go from my side.

PR #7137, opened 2023-11-05
bundle lock/install --checksums

martinemde:

Users need to be able to opt-in to checksums.

I'm slowly fixing all the specs so that they generate checksums depending on if checksums exist already in the current lockfile so that we don't throw away all the specs. When checksums become enabled by default, we can just change the checksums_section method to ignore existing lockfile.

Note: I was starting to solve the problem of detecting when changes happened to the lockfile checksums, so some of this code is a bit excessive. It turns out that's a very difficult problem since some of the checksums are added for use in the lockfile _after_ installing the gems (when the checksums are calculated from the .gem), which means we can't detect modification of frozen lockfile before installation.

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -214,6 +214,8 @@ module Bundler
       "Install using defaults tuned for deployment environments"
     method_option "frozen", :type => :boolean, :banner =>
       "Do not allow the Gemfile.lock to be updated after this install"
+    method_option "checksums", :type => :boolean, :default => false, :banner =>
+      "Enable gem checksum verification"
     method_option "full-index", :type => :boolean, :banner =>
       "Fall back to using the single-file index of all gems"
     method_option "gemfile", :type => :string, :banner =>
@@ -658,6 +660,8 @@ module Bundler
       "do not attempt to fetch remote gemspecs and use the local gem cache only"
     method_option "print", :type => :boolean, :default => false, :banner =>
       "print the lockfile to STDOUT instead of writing to the file system"
+    method_option "checksums", :type => :boolean, :default => false, :banner =>
+      "Enable gem checksum verification"
     method_option "gemfile", :type => :string, :banner =>
       "Use the specified gemfile instead of Gemfile"
     method_option "lockfile", :type => :string, :default => nil, :banner =>
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -56,7 +56,7 @@ module Bundler
 
       Plugin.gemfile_install(Bundler.default_gemfile) if Bundler.feature_flag.plugins?
 
-      definition = Bundler.definition
+      definition = Bundler.definition(checksums: options["checksums"])
       definition.validate_runtime!
 
       installer = Installer.install(Bundler.root, definition, options)
@@ -166,6 +166,10 @@ module Bundler
       Bundler.settings.set_command_option_if_given :no_install, options["no-install"]
 
       Bundler.settings.set_command_option_if_given :clean, options["clean"]
+
+      if Bundler.settings[:disable_checksum_validation] && options["checksums"]
+        raise InvalidOption, "Cannot specify --checksums with bundle config disable_checksum_validation true"
+      end
 
       normalize_groups if options[:without] || options[:with]
 
--- a/bundler/lib/bundler/cli/lock.rb
+++ b/bundler/lib/bundler/cli/lock.rb
@@ -31,7 +31,14 @@ module Bundler
         update = { :conservative => conservative }
       elsif update && bundler
         update = { :bundler => bundler }
+      else
+        update = { :update => update }
       end
+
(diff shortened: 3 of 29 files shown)


deivid-rodriguez:
Would it make it easier or harder to only provide the new flag with `bundle lock`? I was thinking of naming this `--add-checksums` for consistency with the existing `--add-platform` flag to `bundle lock`.

martinemde:
Maybe easier, but @indirect mentioned that it has confused people in the past when they don't support the same options. Also, you wouldn't be able to add checksums on first install.

The biggest pain is fixing all the tests.

I also foresee a problem detecting when a Gemfile.lock will change from checksums. Everything else in the lockfile is generated before install, but checksums is not. I think maybe the best answer is "the lockfile always says it will change if checksums is enabled and any gem from a rubygems source doesn't have checksums".

deivid-rodriguez:
> Maybe easier, but @indirect mentioned that it has confused people in the past when they don't support the same options. Also, you wouldn't be able to add checksums on first install.

I see. I don't recall that confusion but @indirect has been around as a maintainer much longer than myself :) Regarding first install, will we make it so that checksums are automatically added on first install when there's no previous lockfile? If that's the idea I'd say there's no need for a `bundle install --checksums` flags, particularly since we'll probably want `bundle install` to add checksums by default in the future.

Regarding the pain point from detecting the lockfile will change, you mean for `Definition#nothing_changed?`? Do we need to add any conditions related to checksums in there? That's mainly used to decided whether we need to re-resolve gems, and I don't think checksums affect that?

martinemde:
Oh, it's used to determine if we need to resolve specifically so we can download checksums. Theoretically we should resolve if checksums are missing that we could fill in, but not resolve if we have all checksums and resolving would do nothing.

deivid-rodriguez:
Regarding, no checksums by default on new lockfiles, I'm good with being extra cautious 👍, there's no rush in my opinion.

> Oh, it's used to determine if we need to resolve specifically so we can download checksums. Theoretically we should resolve if checksums are missing that we could fill in, but not resolve if we have all checksums and resolving would do nothing.

I think it's fine to keep the current logic for figuring if we need to resolve gems. Even if we skip resolving we can still fill in missing checksums after resolution, no?

By the way, what are the cases where we would miss checksums? `bundle install` would install gems, so will always have the package checksums available. In the case of `bundle lock`, even if we're talking to an old API that does not provide checksums, we could still download the packages and lock their checksums?

martinemde:
> By the way, what are the cases where we would miss checksums? bundle install would install gems, so will always have the package checksums available. In the case of bundle lock, even if we're talking to an old API that does not provide checksums, we could still download the packages and lock their checksums?

Bundle install usually does not get most checksums since we can only get the checksum for gems we're installing fresh from the .gem. Unless the install is pristine, we don't get checksums as currently designed. We could maybe check the .gem cache anyway, but they may not be available even if the gem is installed. Maybe `install --checksums` (or `--add-checksum`) could purposely do this?

As for lock, we don't download gems to calculate checksums. We also don't have a process for fetching checksums outside of the resolving process, so only gems fetched from compact index during resolve can have checksums recorded. If we run a resolve we get all available checksums that the compact index can supply but we don't have a direct "fetch this info file for the checksum".

deivid-rodriguez:
> Bundle install usually does not get most checksums since we can only get the checksum for gems we're installing fresh from the .gem. Unless the install is pristine, we don't get checksums as currently designed. We could maybe check the .gem cache anyway, but they may not be available even if the gem is installed. Maybe install --checksums (or --add-checksum) could purposely do this?

> As for lock, we don't download gems to calculate checksums. We also don't have a process for fetching checksums outside of the resolving process, so only gems fetched from compact index during resolve can have checksums recorded. If we run a resolve we get all available checksums that the compact index can supply but we don't have a direct "fetch this info file for the checksum".

Aaaah, I see. That makes it all clear and I understand the need to re-resolve if checksums are missing now 👍.

In principle, it feels like it should be a separate task decoupled from resolution what fills in missing checksums by looking at the different available sources (If compact index is available, look at local compact index cache or fetch remotely if not present. If compact index is not available, look at local gem cache or fetch package remotely if not present).

But I can see how triggering a full reresolve if checksums are missing would be easier for now even if more costly. Won't fix however when gems from private servers not implementing the compact index are used, and those cases would be bound to never use the lockfile and always re-resolve until they actually do a pristine install, right?

martinemde:
> In principle, it feels like it should be a separate task decoupled from resolution what fills in missing checksums by looking at the different available sources (If compact index is available, look at local compact index cache or fetch remotely if not present. If compact index is not available, look at local gem cache or fetch package remotely if not present).

You're right and this would solve a lot of problems honestly. I will add that. I just rewrote the compact index fetcher in #7122 so I should be able to figure out a good way to access that cache.

martinemde:
Current status: chipping away at test failures caused by the opt-in approach. I think the actual code needs more work so that lock/install work correctly when adding checksums. I will probably make another pass at the code to simplify this down to the minimum necessary for opt-in without any other changes.

I want to add explicit fetching of checksums as an operation rather than relying on the resolve process. This could come in a different PR.

PR #7141, opened 2023-11-08
Fix daily Bundler CI

deivid-rodriguez:

[Daily Bundler CI](https://github.com/rubygems/rubygems/actions/runs/6791813468/job/18464010210#step:6:163) against ruby-head is failing because ruby-head now uses bigdecimal 3.1.5, so that gets locked by this spec.

Bump fake big decimal version used by this test to 99.1.4, so that the test is stable until bigdecimal 99.1.5 is bundled with Ruby :)

### DIFF
--- a/bundler/spec/commands/lock_spec.rb
+++ b/bundler/spec/commands/lock_spec.rb
@@ -387,7 +387,7 @@ RSpec.describe "bundle lock" do
         build_gem "sequel", "5.72.0" do |s|
           s.add_dependency "bigdecimal", ">= 0"
         end
-        build_gem "bigdecimal", %w[1.4.4 3.1.4]
+        build_gem "bigdecimal", %w[1.4.4 99.1.4]
       end
 
       gemfile <<~G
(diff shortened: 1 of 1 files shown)


simi (approved):

simi:
> Bump fake big decimal version used by this test to 99.1.4, so that the test is stable until bigdecimal 99.1.5 is bundled with Ruby :)

That should give us enough time IMHO. :muscle: :smile:

bundler/spec/commands/lock_spec.rb
```diff
@@ -387,7 +387,7 @@
         build_gem "sequel", "5.72.0" do |s|
           s.add_dependency "bigdecimal", ">= 0"
         end
-        build_gem "bigdecimal", %w[1.4.4 3.1.4]
+        build_gem "bigdecimal", %w[1.4.4 99.1.4]
```
martinemde, line 390: May we live long enough to see this test fail.

martinemde (approved):

PR #7142, opened 2023-11-08
Improve specs

deivid-rodriguez:

Easier troubleshooting of CI issues like the one that happens in the last CI run in master: https://github.com/rubygems/rubygems/actions/runs/6780542603/job/18449283187

We only notice the issue when checking the lockfile, but I think the issue happened earlier:

* The `bundle install` command did touch the network, which is what the failing test tries to check that's not happening. We can raise immediately if this test touches the network.
* I think the test did touch the network because the local bundler spec was ignored, and I think that happened because of a permissions issue with the test command that installed the local bundler. I think we can also raise immediately if that kind of permission issue happens.

### DIFF
--- a/bundler/spec/commands/update_spec.rb
+++ b/bundler/spec/commands/update_spec.rb
@@ -1471,7 +1471,7 @@ RSpec.describe "bundle update --bundler" do
       gem "rack"
     G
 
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
Since this affects specs only and I hope it will help stabilizing CI, I'll merge it.

Issue #7105, opened 2023-10-24
Ability to provide custom error message from private gem server


adrianthedev:
### Describe the problem as clearly as you can

I run a private rubygems server (https://packager.dev) to serve private gems of [my product](https://avohq.io).

For my product, I have two license keys. One for the rubygems server and one for a runtime check.
Users often get those mixed and try to use the runtime key instead of the gems key. 
They get the standard 403 message which most of the time doesn't help them very much.
Can I add a custom message to that response? 
If that's not possible, would you be open to a PR with this addition?

duckinator:
I went ahead and added a title to this issue. Can you provide some example output of what the current error looks like? I'm mainly hoping it mentions a class name (e.g. `Bundler::HTTPError`), because that'll narrow down the relevant part of the codebase quite a bit.

adrianthedev:
The response looks like this.
We'd love to be able to respond with a URL to an article or something that will help the user understand what's happening and how they can get self-help.

![CleanShot 2023-10-30 at 21 22 33@2x](https://github.com/rubygems/rubygems/assets/1334409/b78e9e4a-6a7c-44f4-9537-a94e26223b62)

duckinator:
Looks like this is from `‎Bundler::RubygemsIntegration#download_gem‎`.

https://github.com/rubygems/rubygems/blob/b6fe6dd386dbfc3b0b009b10987e74650ff15232/bundler/lib/bundler/rubygems_integration.rb#L491-L520

duckinator:
Closed by #7148.

PR #7148, opened 2023-11-09
improvement: include response body on fetch_http error

Paul-Bob:

We would like to show custom error messages on our private rubygems server (https://packager.dev) that serves private [Avo](https://avohq.io) gems. 

Right now the error message looks like (using local-host for testing purposes):

`Bundler::HTTPError: Could not download gem from http://localhost:4567/avo-hq/ due to underlying error <bad response Forbidden 403 (http://localhost:4567/avo-hq/gems/avo-menu-0.1.18.gem)>`

It would be helpful to us to make this message more customized so we can inform our users about the reason that download failed.

@adrianthedev opened [this issue](https://github.com/rubygems/rubygems/issues/7105) and @duckinator pointed us in the right direction in [this comment](https://github.com/rubygems/rubygems/issues/7105#issuecomment-1785971507).

Here is where we can fully customize the message [lib/bundler/rubygems_integration.rb L519](https://github.com/rubygems/rubygems/blob/b6fe6dd386dbfc3b0b009b10987e74650ff15232/bundler/lib/bundler/rubygems_integration.rb#L519C7-L519C7) but I noticed that the error that is rescued on that line it's already pre-processed [here (remote_fetcher.rb L237)](https://github.com/rubygems/rubygems/blob/4aa1dd06544aaa32a157dddda2fc8637c7b0976f/lib/rubygems/remote_fetcher.rb#L237) and it passes only the `response.message` and the `response.code`.

With this PR the error message also includes the response's body and looks like:

`Bundler::HTTPError: Could not download gem from http://localhost:4567/avo-hq/ due to underlying error <bad response Forbidden 403 Invalid License tier for 'avo-menu'. (http://localhost:4567/avo-hq/gems/avo-menu-0.1.18.gem)>`

Notice that the error now includes the response's body `Invalid License tier for 'avo-menu'.`.

Not sure if is the best approach, open for any kind of improvement.

# PS:

## Alternative approach

What if we add an attribute on `FetchError` that will store the original `response` and we would be able to manipulate it on  [lib/bundler/rubygems_integration.rb L519](https://github.com/rubygems/rubygems/blob/b6fe6dd386dbfc3b0b009b10987e74650ff15232/bundler/lib/bundler/rubygems_integration.rb#L519C7-L519C7) ?

This way the error message can be fully customizable.


adrianthedev:
@duckinator does this look good?

duckinator:
At a glance, this looks good. I'll test it locally when I'm at my computer later today (mainly to see what errors from rubygems.org look like with it), but don't expect any problems. 👍

adrianthedev:
Thank you! We appreciate it!
Let us know if we need to update anything else.

duckinator:
The test failures don't _seem_ to be related to this PR; I asked the other maintainers if they know what's up. I'll get back to you when I have more info. 🙂

duckinator:
Looks like that issue is resolved by #7156.

duckinator:
Thanks for the PR, @Paul-Bob!

lib/rubygems/remote_fetcher.rb
```diff
@@ -234,7 +234,7 @@ def fetch_http(uri, last_modified = nil, head = false, depth = 0)
 
       fetch_http(location, last_modified, head, depth + 1)
     else
-      raise FetchError.new("bad response #{response.message} #{response.code}", uri)
+      raise FetchError.new("bad response #{response.message} #{response.code} #{response.body.strip}", uri)
```
simi, line 237: This can be suuuuuuuuuuuuper messy IMHO when remote provides like 404 HTML page. Would it make sense to limit this to first X characters at least or scan for friendly response and report back only when recognized format is returned? What's returned by RubyGems.org in here?

adrianthedev:
Thanks for the help!
This be available in newer versions of bundler, right?

lib/rubygems/remote_fetcher.rb
duckinator: Referencing a gem that doesn't exist doesn't result in this message being displayed.

I'm not opposed to adding a character limit, though.

```
puppy@orthrus:~/test$ ls -a
./       ../      Gemfile
puppy@orthrus:~/test$ cat Gemfile
source 'https://rubygems.org'

gem 'asdfasdfsadfihgjklsadkjfhakljfhsaf-this-gem-does-not-exist'
puppy@orthrus:~/test$ bundle install
Could not find gem 'asdfasdfsadfihgjklsadkjfhakljfhsaf-this-gem-does-not-exist' in rubygems repository
https://rubygems.org/ or installed locally.
puppy@orthrus:~/test$ 
```

simi: There are other gem servers which can return super long 500 HTML page for example. I would suggest to revert this and instead build some kind of "protocol" to let the error message propagate to client. IMHO it is not good idea to propagate licensing error back to clients using HTTP message on gem install. There is no guarantee how RubyGems handles HTTP errors and IMHO it should not be mis-used to handing over messages to client.

simi: I would be happy to keep this output, but only in kind of "verbose" mode and that is most likely against the original need of @adrianthedev.

indirect: We have also sometimes needed this in Bundler, for example when rate-limiting certain requests. I like the idea of this being officially supported, but I agree that an unlimited message size is messy and unhelpful.

simi: I was thinking how to get this done with raw HTTP without inventing custom thingies like special payload or headers. What about to check for custom content-type (per https://en.wikipedia.org/wiki/Media_type#Vendor_tree like `application/vnd.rubygems.text`) and print the content on non 4xx and 5xx responses?

indirect: That sounds good for "success", but I would also like a way to print errors for rate limiting (429) and server errors (5xx), so we can provide additional information or explanations if desired.

simi: > That sounds good for "success", but I would also like a way to print errors for rate limiting (429) and server errors (5xx), so we can provide additional information or explanations if desired.

I'm not sure I do follow. That is the idea, if response has special recognized content-type, print out the response content, but only for 4xx and 5xx responses.

indirect: Oh, sorry, I misunderstood! I thought you were saying only responses _not_ 4xx or 5xx. 🙃 Sounds good to me.

simi: @adrianthedev would that work for you? ^ Are you in control of returned HTTP headers for 403 on your gemserver?

adrianthedev: Yes @simi, we are in control of the returned HTTP headers. Our gem server is operated by a sinatra app.

Not completely sure how to send the message. 
Is it something like this?

```ruby
halt 403, {'Content-Type' => 'application/vnd.rubygems.text'}, { message: 'some text' }.to_json
```

Our message is something like this. That's 164 characters. We might play around with it so the number of chars might increase.

```
It seems that you don't have a valid GEM_SERVER_TOKEN for Avo (https://bundler.dev/). Please read this for more information
https://docs.avohq.io/3.0/some-path-here
```

deivid-rodriguez: Did I understand correctly that we'd rather revert this PR and go with the cleaner solution that @simi proposed?

simi: Yes, if possible, don't release this. 🙏

PR #7150, opened 2023-11-09
Provide more insightful test_realworld_upgraded_default_gem(TestGemRequire) error output

voxik:
Original output:

~~~
Failure: test_realworld_upgraded_default_gem(TestGemRequire): <false> is not true. /mnt/test/rubygems/test_require.rb:474:in `test_realworld_upgraded_default_gem'
     471:     File.write(path, code)
     472:
     473:     output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
  => 474:     assert $?.success?
     475:     refute_empty output
     476:     assert_equal "999.99.9", output.lines[0].chomp
     477:     # Make sure only files from the newer json gem are loaded, and no files from the default json gem
~~~

New output:

~~~
Failure: test_realworld_upgraded_default_gem(TestGemRequire) /mnt/test/rubygems/test_require.rb:474:in `test_realworld_upgraded_default_gem'
     471:     File.write(path, code)
     472:
     473:     output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
  => 474:     assert_equal "999.99.9", output.lines[0].chomp
     475:     # Make sure only files from the newer json gem are loaded, and no files from the default json gem
     476:     assert_equal ["#{@gemhome}/gems/json-999.99.9/lib/json.rb"], output.lines.grep(%r{/gems/json-}).map(&:chomp)
     477:     assert $?.success?
<"999.99.9"> expected but was
<"/mnt/tmp/test_rubygems_20231109-27890-kd68d4/test_realworld_upgraded_default_gem.rb:3:in `<main>': undefined method `version' for nil:NilClass (NoMethodError)">

diff:
?                               9    99                                           .99.9
? /mnt/tmp/test_rubygems_2023110 -278 0-kd68d4/test_realworld_upgraded_default_gem rb:3:in `<main>': undefined method `version' for nil:NilClass (NoMethodError)
? ++++++++++++++++++++++++++++++ ++++ ? ????
~~~

It is more valuable to check the command output then the error code. If the command fails for some reason, the output probably contains some detail, while checking the return code tells not much.

Also, removing the excessive output content check, when the output is later inspeced in more detail.

### DIFF
--- a/test/rubygems/test_require.rb
+++ b/test/rubygems/test_require.rb
@@ -471,11 +471,11 @@ class TestGemRequire < Gem::TestCase
     File.write(path, code)
 
     output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
-    assert $?.success?
     refute_empty output
     assert_equal "999.99.9", output.lines[0].chomp
     # Make sure only files from the newer json gem are loaded, and no files from the default json gem
     assert_equal ["#{@gemhome}/gems/json-999.99.9/lib/json.rb"], output.lines.grep(%r{/gems/json-}).map(&:chomp)
+    assert $?.success?
   end
 
   def test_default_gem_and_normal_gem


voxik:
Just for the context, I have hit this error on Fedora, where JSON is for "reasons" not 100% default gem. Would you be open to PR switching this test case (and also `test_realworld_default_gem(TestGemRequire)`) from JSON to e.g. date?

voxik:
> Just for the context, I have hit this error on Fedora, where JSON is for "reasons" not 100% default gem. Would you be open to PR switching this test case (and also `test_realworld_default_gem(TestGemRequire)`) from JSON to e.g. date?

Since I am already on tangent, using e.g. date or some less prominent gem would also possible allow to revert 731999e970d3c39a0d716e4e025094799e0131ba, because I don't think the commit is doing any good to the test case. It makes it ambiguous what it actually loads and it might fail with empty string again.

test/rubygems/test_require.rb
```diff
@@ -471,11 +471,10 @@ def test_realworld_upgraded_default_gem
     File.write(path, code)
 
     output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
-    assert $?.success?
-    refute_empty output
     assert_equal "999.99.9", output.lines[0].chomp
     # Make sure only files from the newer json gem are loaded, and no files from the default json gem
     assert_equal ["#{@gemhome}/gems/json-999.99.9/lib/json.rb"], output.lines.grep(%r{/gems/json-}).map(&:chomp)
+    assert $?.success?
```
voxik, line 477: Maybe it would be better to drop this `assert` altogether, not sure.

voxik:
The Bundler test failure does not seem to be related

test/rubygems/test_require.rb
```diff
@@ -471,11 +471,10 @@ def test_realworld_upgraded_default_gem
     File.write(path, code)
 
     output = Gem::Util.popen({ "GEM_HOME" => @gemhome }, *ruby_with_rubygems_in_load_path, path).strip
-    assert $?.success?
-    refute_empty output
```
deivid-rodriguez, line 475: Why removing this assertion? Would provide a clear message if output is empty, no? I think moving the assertion of the status code to the end gets you the better message that you want.

voxik: Hm :thinking: My thinking was that if the string is empty, the following assert would fail anyway. But with empty string, it would actually fail unexpectedly with `NoMethodError: undefined method 'chomp' for nil:NilClass`. This could be fixed by `output.lines(chomp: true)[0]`, which provides better error message: `<"999.99.9"> expected but was <nil>`. But all in all, you are probably right that it won't hurt to keep it around.

voxik: Updated. Thx for the nudge :blush:

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks!

Issue #7119, opened 2023-10-27
RubyGems test suite does not play well with defaults overrides


voxik:
RubyGems provides a way to customize their configuration via:

https://github.com/rubygems/rubygems/blob/1a1948424aca90013b37cb8a78f5d1d5576023f1/lib/rubygems.rb#L1325

or

https://github.com/rubygems/rubygems/blob/1a1948424aca90013b37cb8a78f5d1d5576023f1/lib/rubygems.rb#L1342

Since Fedora ships own `operating_system.rb` file, I was always struggling a bit with executing the test suite. But since it was probably just me and there was relative simple workaround:

~~~
$ mkdir lib/rubygems/defaults/
$ touch lib/rubygems/defaults/operating_system.rb
~~~

I was not bothered to find better solution.

Nevertheless, I have prepared [this](https://bugs.ruby-lang.org/issues/19972) Ruby proposal + [POC PR](https://github.com/ruby/ruby/pull/8761) and because this includes some custom configuration in `rubygems/defaults/ruby.rb`, the test suite [fails](https://github.com/ruby/ruby/actions/runs/6653479202/job/18079551870). I have tried to improve the situation [a bit](https://github.com/ruby/ruby/pull/8761/commits/2347cae771c283745b86a5583e026b026be86518), but then it [fails](https://github.com/ruby/ruby/actions/runs/6653945271/job/18080981045) on different places differently.

Now I wonder what would be the best way to fix this.

The workaround is valid, but I believe that Ruby executes RubyGems test suite in a slightly different way than Ruby itself. So that would probably needed fix in Ruby as well as in RubyGems (and as you can see, I was locally always executing test suite in a yet another way :laughing:, which does not make this too attractive to me).

I have tried to modify `test/rubygems/helper.rb` to provide fake directory structure with such files, but it loads RubyGems as a first step, which does not make this very viable.

Or maybe all the tests which fails with the operating_system.rb should be made more robust? Maybe this kind of failure is really caused just by some wrong assumptions:

~~~
    7) Failure:
  TestGem#test_default_path [/home/runner/work/ruby/ruby/src/test/rubygems/test_gem.rb:582]:
  <["/home/runner/work/ruby/ruby/build/tmp/test_rubygems_20231026-59729-7r82or/gemhome"]> expected but was
  <["/home/runner/work/ruby/ruby/build/tmp/test_rubygems_20231026-59729-7r82or/gemhome",
[...]

PR #7153, opened 2023-11-10
Explicitly define `@default_specifications_dir` for tests

voxik:
Resetting `@default_specifications_dir` to `nil` means that the `Gem.default_specifications_dir` needs to be invoked. However, given that this method might be overridden by operating_system.rb and similar, this might lead to various test failures. Providing the default value makes the issues go away.

Please note that this single line reduced the number of test failures for ruby/ruby/pull/8761 from ~80 to just ~4 and fixes most of my concerns raised in #7119


deivid-rodriguez (approved):
Seems harmless to me!

PR #7156, opened 2023-11-13
Don't check `LIBRUBY_RELATIVE` in truffleruby to signal a bash prelude in rubygems binstubs

deivid-rodriguez:

`gem install rails` on a truffleruby install with upgraded RubyGems hangs with the following command prompt:

```
racc's executable "racc" conflicts with /Users/deivid/.asdf/installs/ruby/truffleruby-dev/bin/racc
Overwrite the executable? [yN]
```
 
My fix is to upstream TruffleRuby monkey patch to RubyGems, where they force RubyGems to skip the checking whether a binstub was generated by RubyGems. Apparently truffle ruby always ships binstubs with a bash prelude, regardless of `LIBRUBY_RELATIVE`, so I think this makes sense.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -226,7 +226,8 @@ class Gem::Installer
       line = io.gets
       shebang = /^#!.*ruby/
 
-      if load_relative_enabled?
+      # TruffleRuby uses a bash prelude in default launchers
+      if load_relative_enabled? || RUBY_ENGINE == "truffleruby"
         until line.nil? || line =~ shebang do
           line = io.gets
         end


simi (approved):
I'm not fan of those conditions everywhere, but if there's nothing better to do for now @eregon, let's ship this.

eregon (commented):
Right, it's the same change as in https://github.com/rubygems/rubygems/pull/6238/files for the first file, LGTM.

eregon:
An alternative would be to just remove the `if` entirely and let the loop condition do its work in all cases.
That would be more general.

deivid-rodriguez:
Happy to take improvements, let ship this for now.

Issue #4895, opened 2021-09-02
Bundle Install Error: 403 Bad Response, Fetch Error


ababar77:
### Describe the problem as clearly as you can

# Error Report

## Questions

Please fill out answers to these questions, it'll help us figure out
why things are going wrong.

- **What did you do?**

  I ran the command `/Users/ababar/.rvm/gems/ruby-3.0.0@venice/bin/bundle install`

- **What did you expect to happen?**

  I expected Bundler to... install my Ruby Gems from my Jfrog Repos.

- **What happened instead?**

  Instead, what happened was... I got this Gem: Fetch error, Bad Response 403. 

- **Have you tried any solutions posted on similar issues in our issue tracker, stack overflow, or google?**

  I tried... making sure my jfrog rest api is authenticated correctly by downloading some gems from my Jfrog repo. That worked, and I was able to authenticate and download multipe gems from my Jfrog Repo in question. So, it appears I can authenticate into my Jfrog via rest api but I am getting this bad response 403 error when I try to run bundle install.

- **Have you read our issues document, https://github.com/rubygems/rubygems/blob/master/bundler/doc/contributing/ISSUES.md?**
yes
  ...
yes

Bundle install isnt downloading ruby gems from Jfrog Artifactory. Getting a fetch error 403 bad response.

### Which command did you run?

I ran the command `/Users/ababar/.rvm/gems/ruby-3.0.0@venice/bin/bundle install`
### What were you expecting to happen?

Bundle install successful

### What actually happened?

Gem::RemoteFetcher::FetchError: bad response  403 (https://thinknear.jfrog.io/artifactory/api/gems/ruby-virtual/gems/aws-partitions-1.431.0.gem)### If not included with the output of your command, run `bundle env` and paste the output below

 bundle env
## Environment

```
Bundler             2.2.17
  Platforms         ruby, x86_64-darwin-20
Ruby                3.0.0p0 (2020-12-25 revision 95aff214687a5e12c3eb57d056665741e734c188) [x86_64-darwin20]
  Full Path         /Users/ababar/.rvm/rubies/ruby-3.0.0/bin/ruby
  Config Dir        /Users/ababar/.rvm/rubies/ruby-3.0.0/etc
RubyGems            3.2.26
  Gem Home          /Users/ababar/.rvm/gems/ruby-3.0.0@venice
[...]

Issue #7169, opened 2023-11-17
Incompatible gems in optional groups is more difficult since Bundler 2.4.21


Bo98:
### Describe the problem as clearly as you can

I have a Gemfile that looks like this: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/Gemfile. Ignore the Dependabot-specific hack at the top. The lockfile is here: https://github.com/Homebrew/brew/blob/master/Library/Homebrew/Gemfile.lock.

The basic scenario is I have:

```rb
group :typecheck, optional: true do
  gem "sorbet-static-and-runtime", require: false
end
```

and in the lockfile:

```
...
    sorbet-static (0.5.10461-universal-darwin-14)
    sorbet-static (0.5.10461-universal-darwin-15)
    sorbet-static (0.5.10461-universal-darwin-16)
    sorbet-static (0.5.10461-universal-darwin-17)
    sorbet-static (0.5.10461-universal-darwin-18)
    sorbet-static (0.5.10461-universal-darwin-19)
    sorbet-static (0.5.10461-universal-darwin-20)
    sorbet-static (0.5.10461-universal-darwin-21)
    sorbet-static (0.5.10461-universal-darwin-22)
    sorbet-static (0.5.10461-x86_64-linux)
...

PLATFORMS
  aarch64-linux
  arm-linux
  arm64-darwin
  x86_64-darwin
  x86_64-linux
```

This is perhaps unconventional as technically speaking aarch64-linux and arm-linux are invalid for sorbet-static as there is no compatible gem. And indeed if you are on those platforms and try to run `bundle update`/`bundle lock --update` or try to install the `typecheck` group, you will indeed get resolver issues.

However, sorbet-static is in an optional group and `bundle install` without the group works perfectly fine on arm64 Linux. For us the group of people modifying and updating the Gemfile and the people installing from the Gemfile are two distinct groups. The latter group doesn't even need to know about the concept of a group: we pass the relevant groups as necessary. Notably, it is acceptable for us to be unable to maintain the Gemfile on arm64 Linux but still be able to install from it.

This all however worked ok until today we saw Dependabot open a PR that started removing one of the bad platforms: https://github.com/Homebrew/brew/pull/16225/files (I'll send a PR that fixes it not removing both platforms). This was because Dependabot now uses Bundler 2.4.22 and it seems 2.4.21 changed the behaviour to start force removing invalid platforms (https://github.com/rubygems/rubygems/pull/7035).

[...]

PR #7172, opened 2023-11-17
lock_platform directive in Gemfile

martinemde:

There are two problems I aim to solve:

#### 1. First deploys can fail because the platform is not added

When a new user creates an application from a generator or template, they get a Gemfile that does not make it easy to see or declare the platform they use for deployment. It's not currently possible to add default platforms without additional steps. Deployment platform is not obvious, not immediately visible and not declarative within the most obvious place to declare dependencies, the Gemfile.

Many options have been proposed, including [always locking linux](https://github.com/rubygems/rubygems/pull/5700) or [locking available platforms](https://github.com/ruby/ruby/commit/435eb56f6175b7c9a16121ec8441f7492fa9aec5) which have resulted in [at least one problem](https://github.com/rubygems/rubygems/issues/7159).

This is notably different than a previous attempt, way back in 2016, to add a similar DSL (#4895). This proposed option is purely additive and does not restrict the user to only installing on certain platforms. It simply ensures that a given platform is always present. The original solution 7 years ago resulted in force ruby platform.

My assumptions are as follows:

* Most people deploy to `x86_64-linux` or know when they are not.
* Most people will look at their Gemfile but almost never look at their whole Gemfile.lock
* People not deploying to `x86_64-linux` would recognize and correct the platform if it is at the top of their Gemfile, but will mostly not notice when we automatically add a platform to their Gemfile.lock.

#### 2. The deployment platform is important, and yet it is ephemeral

You could theoretically completely regenerate your Gemfile.lock by deleting and reinstalling (generating a fully new resolution), and it might work. However, you would _always_ lose your locked platforms because they are not specified anywhere outside of the generated lockfile. Although this sort of regeneration is not recommended, it points out that locked platforms are ephemeral.

This is fine for most platforms. It's ok if we auto-add different darwin versions when a developer installs, but the deployment platform _should_ be added since it will be auto-added, or it will fail on frozen installs. Since it is a strict deployment dependency that we auto-add if it's missing, it should be intentionally declared in the Gemfile.

Additionally, when we lock to checksums this will make it possible to lock checksums for the deployment platform more successfully, instead of allowing production to install without checksum verification or to fail when a frozen bundle is used and the deployment platform is not locked.

Let's cut to the chase here. This is part of `rails new`:

```
         run  bundle install
Resolving dependencies...
Fetching public_suffix 5.0.4
Fetching nokogiri 1.15.5 (arm64-darwin)
Fetching nio4r 2.6.0
Installing public_suffix 5.0.4
Installing nio4r 2.6.0 with native extensions
Installing nokogiri 1.15.5 (arm64-darwin)
Fetching loofah 2.22.0
Installing loofah 2.22.0
Bundle complete! 14 Gemfile dependencies, 82 gems now installed.
         run  bundle lock --add-platform=x86_64-linux
Resolving dependencies...
Writing lockfile to /Users/martinemde/Projects/tmp/rails712/Gemfile.lock
         run  bundle lock --add-platform=aarch64-linux
Resolving dependencies...
Writing lockfile to /Users/martinemde/Projects/tmp/rails712/Gemfile.lock
```

This causes 3 resolves and adds two additional platforms by default.

Let's allow rails to generate the following Gemfile and run only one resolve:

```Gemfile
source "https://rubygems.org"

# Update this with your production deployment platforms
lock_platforms "x86_64-linux", "aarch64-linux"

ruby "3.2.2"

# Bundle edge Rails instead: gem "rails", github: "rails/rails", branch: "main"
gem "rails", "~> 7.1.2"
# SNIP...
```

Since Gemfiles are often generated or cargo culted from other Gemfiles, adding an option for specifying locked platforms in the Gemfile makes it more likely that initial deploys will succeed and already have the platformed gems for the deployment platform.

This PR adds a new Gemfile directive, `lock_platform`, that may be included by default in generated Gemfiles or added to declare and ensure the deployment platform (or other desired platforms).

**This pull request is currently a draft to show a proof of concept and not all tangential cases have been considered. Please call out any problems you see that may result from this approach.**

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -1058,6 +1058,7 @@ module Bundler
       platforms.reverse_each do |platform|
         next if local_platform == platform ||
                 @new_platforms.include?(platform) ||
+                @locked_platforms.include?(platform) ||
                 @path_changes ||
                 @dependency_changes ||
                 @locked_spec_with_invalid_deps ||
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -56,7 +56,7 @@ module Bundler
     #   to be updated or true if all gems should be updated
     # @param ruby_version [Bundler::RubyVersion, nil] Requested Ruby Version
     # @param optional_groups [Array(String)] A list of optional groups
-    def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [], gemfiles = [])
+    def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, optional_groups = [], gemfiles = [], platforms = [])
       if [true, false].include?(unlock)
         @unlocking_bundler = false
         @unlocking = unlock
@@ -89,6 +89,7 @@ module Bundler
         @locked_gems = LockfileParser.new(@lockfile_contents)
         @locked_platforms = @locked_gems.platforms
         @platforms = @locked_platforms.dup
+        platforms.each {|p| add_platform(p) }
         @locked_bundler_version = @locked_gems.bundler_version
         @locked_ruby_version = @locked_gems.ruby_version
         @originally_locked_deps = @locked_gems.dependencies
@@ -107,7 +108,6 @@ module Bundler
         end
       else
         @unlock         = {}
-        @platforms      = []
         @locked_gems    = nil
         @locked_deps    = {}
         @locked_specs   = SpecSet.new([])
@@ -116,6 +116,8 @@ module Bundler
         @locked_sources = []
         @locked_platforms = []
         @locked_checksums = Bundler.feature_flag.bundler_3_mode?
+        @platforms = platforms
+        @new_platform = true unless @platforms.empty?
       end
 
       locked_gem_sources = @locked_sources.select {|s| s.is_a?(Source::Rubygems) }
@@ -149,7 +151,7 @@ module Bundler
         @gems_to_unlock = @explicit_unlocks.any? ? @explicit_unlocks : @dependencies.map(&:name)
       else
         eager_unlock = @explicit_unlocks.map {|name| Dependency.new(name, ">= 0") }
-        @gems_to_unlock = @locked_specs.for(eager_unlock, false, platforms).map(&:name).uniq
+        @gems_to_unlock = @locked_specs.for(eager_unlock, false, @platforms).map(&:name).uniq
       end
 
       @dependency_changes = converge_dependencies
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -33,6 +33,7 @@ module Bundler
       @install_conditionals = []
       @optional_groups      = []
       @platforms            = []
+      @lock_platforms       = []
       @env                  = nil
       @ruby_version         = nil
       @gemspecs             = []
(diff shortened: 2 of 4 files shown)


bundler/lib/bundler/dsl.rb
```diff
@@ -249,6 +250,19 @@ def platforms(*platforms)
     end
     alias_method :platform, :platforms
 
+    # Name collision, working around it for a proof
+    def lock_platform(*platforms)
+      platforms.each do |p|
+        gem_platform = Gem::Platform.new(p)
+        # TODO: this is copied from lock
+        if gem_platform.to_s == "unknown"
+          Bundler.ui.warn "The platform `#{platform_string}` is unknown to RubyGems " \
```
segiddins, line 259: IMO this should raise here

bundler/lib/bundler/definition.rb
```diff
@@ -89,6 +89,7 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
         @locked_gems = LockfileParser.new(@lockfile_contents)
         @locked_platforms = @locked_gems.platforms
         @platforms = @locked_platforms.dup
+        platforms.each {|p| add_platform(p) }
```
segiddins, line 92: does it work to call `platforms.each {|p| add_platform(p) }` unconditionally, at line 118?

bundler/lib/bundler/dsl.rb
martinemde: makes sense. I copied it from the other place we add_platform, in lock.rb.

deivid-rodriguez:
> Many options have been proposed, including https://github.com/rubygems/rubygems/pull/5700 or [locking available platforms](https://github.com/ruby/ruby/commit/435eb56f6175b7c9a16121ec8441f7492fa9aec5) which have resulted in https://github.com/rubygems/rubygems/issues/7159.

Just noticing that I already have a local fix for this. I don't think a bug should "drive" this discussion. I certainly expected some issues to come up from that change.

martinemde:
> Just noticing that I already have a local fix for this. I don't think a bug should "drive" this discussion. I certainly expected some issues to come up from that change.

I agree. Let's not let this drive the discussion, but maybe we can decide if this is a preferred solution for the problems we see related to locked platforms.

bundler/lib/bundler/definition.rb
martinemde: Probably. This was a premature optimization when we know that `@locked_platforms` is empty when there's no lockfile.

deivid-rodriguez:
I don't think these are mutually exclusive! I like being able to see target platforms in the lockfile, but I also like that Gemfiles are usable in many platforms by default. Locking multiple platforms seems to be a trend across several package managers.

indirect:
This seems reasonable to me! I am hoping that most of the current reasons to use this will be removed by universal lockfiles (🤞🏻) but this seems like a good option for the other cases.

martinemde:
#7169 might just be the specific case that benefits from this feature. I think right now resolving might fail if the gem is unavailable on a locked platform. The issue above seems like a good test case if we can get this to support their needs.

martinemde:
Updated to prevent the lock_platforms from being removed even if they would be invalid normally, ensuring that an error is caused rather than produce a Gemfile.lock that cannot be deployed to a locked platform.

segiddins:
@martinemde are there remaining blockers here? would love to get this included in 2.6

martinemde:
@segiddins that'd be great. I think it just plain didn't work correctly last I tried.

martinemde:
@segiddins rebased, let's see...

martinemde:
Closing this to reduce clutter. Please feel free to revisit.

PR #7184, opened 2023-11-22
Use GitHub Discussions with report suggestion

hsbt:

`Bundler::FriendlyErrors` still suggests to report our issues. But their report is not helpful for developpers and mostly there is no response from first reporter. We should switch this template to discussions from issues. If we decide to handle it as issues, we can easily convert it to issues.

replaced GH issues to GH discussions.

### DIFF
--- a/bundler/lib/bundler/friendly_errors.rb
+++ b/bundler/lib/bundler/friendly_errors.rb
@@ -78,7 +78,7 @@ module Bundler
       Bundler.ui.error <<~EOS, nil, :yellow
 
         First, try this link to see if there are any existing issue reports for this error:
-        #{issues_url(e)}
+        #{discussions_url(e)}
 
         If there aren't any reports for this error yet, please fill in the new issue form located at #{new_issue_url}, and copy and paste the report template above in there.
       EOS
@@ -99,12 +99,12 @@ module Bundler
       EOS
     end
 
-    def issues_url(exception)
+    def discussions_url(exception)
       message = exception.message.lines.first.tr(":", " ").chomp
       message = message.split("-").first if exception.is_a?(Errno)
       require "cgi"
       "https://github.com/rubygems/rubygems/search?q=" \
-        "#{CGI.escape(message)}&type=Issues"
+        "#{CGI.escape(message)}&type=Discussions"
     end
 
     def new_issue_url
--- a/bundler/lib/bundler/friendly_errors.rb
+++ b/bundler/lib/bundler/friendly_errors.rb
@@ -108,7 +108,7 @@ module Bundler
     end
 
     def new_issue_url
-      "https://github.com/rubygems/rubygems/issues/new?labels=Bundler&template=bundler-related-issue.md"
+      "https://github.com/rubygems/rubygems/discussions/new?category=q-a"
     end
   end
 
--- a/bundler/lib/bundler/friendly_errors.rb
+++ b/bundler/lib/bundler/friendly_errors.rb
@@ -77,7 +77,7 @@ module Bundler
 
       Bundler.ui.error <<~EOS, nil, :yellow
 
-        First, try this link to see if there are any existing issue reports for this error:
+        First, try this link to see if anyone has reported this error:
         #{discussions_url(e)}
 
         If there aren't any reports for this error yet, please fill in the new issue form located at #{new_issue_url}, and copy and paste the report template above in there.
--- a/bundler/lib/bundler/friendly_errors.rb
+++ b/bundler/lib/bundler/friendly_errors.rb
@@ -80,7 +80,7 @@ module Bundler
         First, try this link to see if anyone has reported this error:
         #{discussions_url(e)}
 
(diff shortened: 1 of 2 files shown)


bundler/lib/bundler/friendly_errors.rb
```diff
@@ -78,7 +78,7 @@ def request_issue_report_for(e)
       Bundler.ui.error <<~EOS, nil, :yellow
 
         First, try this link to see if there are any existing issue reports for this error:
```
duckinator, line 80: Maybe this?:

```suggestion
        First, try this link to see if anyone has reported this error:
```

bundler/lib/bundler/friendly_errors.rb
```diff
@@ -99,16 +99,16 @@ def serialized_exception_for(e)
...
+    def discussions_url(exception)
       message = exception.message.lines.first.tr(":", " ").chomp
       message = message.split("-").first if exception.is_a?(Errno)
       require "cgi"
       "https://github.com/rubygems/rubygems/search?q=" \
-        "#{CGI.escape(message)}&type=Issues"
+        "#{CGI.escape(message)}&type=Discussions"
     end
 
     def new_issue_url
```
duckinator, line 110: Given the link now creates a new Discussion, not a new Issue, we should update this function name accordingly.

```suggestion
    def new_discussion_url
```

bundler/lib/bundler/friendly_errors.rb
```diff
@@ -78,7 +78,7 @@ def request_issue_report_for(e)
       Bundler.ui.error <<~EOS, nil, :yellow
 
         First, try this link to see if there are any existing issue reports for this error:
-        #{issues_url(e)}
+        #{discussions_url(e)}
 
         If there aren't any reports for this error yet, please fill in the new issue form located at #{new_issue_url}, and copy and paste the report template above in there.
```
duckinator, line 83: This changes `new_issue_url` to `new_discussion_url`, to match my other suggestion.

```suggestion
        If there aren't any reports for this error yet, please fill in the new issue form located at #{new_discussion_url}, and copy and paste the report template above in there.
```

duckinator (changes requested):
I don't have strong feelings other way about this change, but _if_ this does get merged I'd like the message to be simplified a bit and for `new_issue_url` to be renamed to `new_discussion_url`.

Specifically, I think this phrasing is a bit easier to understand:

> First, try this link to see if anyone has reported this error:

I provided these as "suggested changes" you can apply directly from GitHub, but feel free to make the changes locally yourself if you prefer! 🙂

hsbt:
@duckinator Thank you for your review. Just update them.

duckinator (approved):
Looks good to me.

deivid-rodriguez:
I disagree with this. Whenever the bug report template is printed to end users, that's a bug in Bundler and it's helpful to get it reported. We still get bad quality reports where users only copy the template without providing any more information, but I don't think recommending opening a discussion improves that.

deivid-rodriguez:
I think I've never converted a "Bug report template" issue into a discussion? 🤔

hsbt:
It's not for improving report quality. Discussions is helpful with community support. In fact, https://github.com/rbenv/ruby-build/discussions is handled by many users.

deivid-rodriguez:
As I explained I'm against this because bug reports should get reported to the bug tracker, not as discussion. I'm happy to get wider community help with triaging issues, but the right place for that is the issue tracker.

hsbt:
I don't think so.

ok, I withdraw this because I didn't get your approval.

deivid-rodriguez:
I'm sorry! I'm not the only maintainer, if other maintainers prefer this way I can accept the change. I can also triage discussions so it's not a big deal. We can reopen if there's more support for this change.

hsbt:
👌 @indirect @simi How about this? 

https://github.com/rubygems/rubygems/discussions/6760 and https://github.com/rubygems/rubygems/discussions/7183 are good example for this PR.

indirect:
I’m ok with either one, since we have to triage either way.

I think for common copy and paste errors we should change Bundler to not suggest an error report. 😅 For example, if there is a permission error we could print “Bundler cannot continue because it does not have permission to write to /path.” instead of suggesting a new issue/discussion.

deivid-rodriguez:
Absolutely!

That's exactly the direction I've been taking for a while for addressing these issues. See for example https://github.com/rubygems/rubygems/pull/7113, https://github.com/rubygems/rubygems/pull/5726, https://github.com/rubygems/rubygems/pull/4965.

My take is that the bug report being printed is always a Bundler bug, because either:
* It shows a real bug in Bundler and provides helpful info to address it.
* It does NOT show a real bug in Bundler, but an expected error condition. In that case, printing the bug report template is also a bug, since we should be printing instead a proper informative error to the user and not the bug report template.

hsbt:
okay, I'll reopen this when we will get many of not helpful issues.

PR #7189, opened 2023-11-23
Simplify remembered flags deprecation message

deivid-rodriguez:

This deprecation message recommends `bundle config set --local`, however `bundle config` is now local by default. It also quotes some values unnecessarily.

Simplify the deprecation message.


deivid-rodriguez:
The "risk" of this is that users may copy the deprecation message to an app still using a old Bundler version where `bundle config` was not local, so maybe not worth changing.

indirect (approved):
I think it is fine to print messages that only apply to the running version of Bundler, but thank you for calling out that risk!

PR #7191, opened 2023-11-23
Stop remembering cli flags like `--jobs` or `--retry` in configuration

### LINKED ISSUE #7181: Bundler project specific options

When installing `gems` for GitLab, it is easy to break other projects. To fix that by installing GitLab gems into project subdir, I've committed `.bundle/config` with the following contents into the repo.

```yaml
---
BUNDLE_PATH: ".gems"
```

That, however, was met with resistance, as people tell me that `.bundle/config` automatically saves user options, even implicit ones, such as `bundle install --jobs`.

So what are the project level settings file that could be committed to repository without conflicting with user preferences for the same project?

1. Config location for global user preferences for `bundler`
2. Config location for project specific preferences that override global user preferences
3. Config location for project specific user preferences that override project preferences

deivid-rodriguez:

We're still remembering CLI flags for backwards compatibility. However, this is only necessary for a few selected flags.

Stop remembering flags like `--jobs`, or `--retry`.

This is stacked on top of #7189 and #7190.

Closes #7181.

### DIFF
--- a/bundler/lib/bundler/settings.rb
+++ b/bundler/lib/bundler/settings.rb
@@ -46,6 +46,20 @@ module Bundler
       update_requires_all_flag
     ].freeze
 
+    REMEMBERED_KEYS = %w[
+      bin
+      cache_all
+      clean
+      deployment
+      frozen
+      no_prune
+      path
+      shebang
+      path.system
+      without
+      with
+    ].freeze
+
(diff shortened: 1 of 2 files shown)


indirect (approved):
Awesome, thanks for doing this!

martinemde (approved):
Code looks good.

PR #7193, opened 2023-11-25
Extracting table/json formatting from bundle-outdated to Bundle::UI

nevinera:
## Motivation
Follow-on suggestion from #7167 - factoring the table-construction code out of the bundle-outdated cli class into Bundle::UI::Shell, and then extracting the _json_ construction code into a new Bundle::UI::JsonShell class to mirror it.

# Implementation
* Add a `Bundler::Ui::Shell#table` method, which takes `headers`, `rows`, and ignores the `_options` parameter.
  - it uses the headers to define what columns should exist, and to print the titles on those columns
  - it iterates the rows to construct the contents of those columns
  - it determines the width of each column based on the widest entry in the column (header included)
  - it separates columns by two spaces
  - it is _not_ terminal-width-aware; it'll just wrap. Which is very ugly, so don't make wide tables.
  - It's just the ported version that was in Cli::Outdated - this could be improved later by someone that wants to spend some time researching terminal UI building - this is a _very_ solved problem, it's just all the solutions are from the 80s and written in ugly languages -.-
* Add a new `Bundler::UI::JsonShell`, which inherits from `UI::Shell`, overrides the existing methods to all write to stderr (instead of stdout), and overrides `table` to write a flat json blob (array of objects) to stdout.
  - support a `:pretty => true` option to use `JSON.pretty_generate` (if available)
  - If a row is _missing_ any entry that is present in the header, fill it in (with a nil value)
  - If a row includes entries that are _not_ present in the header, leave them in. There is definite value to letting tools include more data in their json-output, where formatting constraints don't limit the data that can be rendered.
  - Ignore the _values_ of the headers - I wavered on this, but I decided that the keys used for the data are likely to be more json-key-like than the table headers supplied for the terminal table. Even mild argument would convince me otherwise though :-)
* Update the `Bundler::Cli` to instantiate the `JsonShell` instead of the `Shell` if passes `--json`.
  - note that this _does_ actually produce a behavioral change in other tools - if you supply `--json` to bundle install for example, anything that would have printed on stdout now prints on stderr. Just as a side-effect of the fact that it needs to get out of the way for the json output that that tool doesn't actually (yet) write. If anyone has a suggestion that would make this less awkward let me know, because it kind of bugs me.
* Update the `Bundler::Cli::Outdated` tool to remove the code dedicated to constructing tables and json data, and replace it with calls to `Bundler.ui.table`.
  - Also rename `print_gem`/`print_gems` to `parseably_print_gem`/`parseably_print_gems`, since that's all it's for (and already was).
  - Remove some dead code from `parseably_print_gem` - it used to be the only printing method a long while ago, and some of the bits that were there to support other types of printing were left around, despite not being reachable.


nevinera:
@simi: I've started on the follow-up work, and I'm running into several interesting questions I'd like your input on:

1. The json format I added here has multiple levels (each spec gets an object full of data). I don't actually _need_ that, and it would probably be simpler to just mirror the existing structure of data in the json format, but with a few extra fields (since there's no formatting constraints), which I've already got a nice solution for. But.. this `--json` code already exists now; how sensitive should I be to yanking around the shape of the resulting data in the short term?
2. The `print_gem` method (which I've got renamed locally to `parseably_print_gem`) is not just a spacing-change - it includes the data in a somewhat different structure, and seems _older_ than the other printing mechanisms, based on some git-spelunking. Given the goal of such a formatter, I'm concerned about changing this output _at all_, and thereby breaking something external that depends on parsing that structure, so I think I'll need to leave this manual string-building in place just for the `--porcelain` option. I don't see a nice way to map that onto the `Bundler::UI` abstraction, but maybe something will occur to one of you :-) (There is some dead code in here that I can remove while I'm at it though)
3. I couldn't see any way to trigger the `Bundler::UI::Silent` (at all), but given its implementation's location and structure, I wanted to double-check that I shouldn't add an empty `table` method to it?
4. How do we want UI::JsonShell to _behave_? I was initially thinking "Oh, we'll just write out the output in json", but .. error messages and the like are still important. And that all can still work fine. But it also occurred to me that we can even handle _info_ messages if we want to, we'd just need to write them to stderr.. Should I perhaps even subclass UI::Shell?

martinemde:
1. Easier to add more later. I'd go with the minimum needed at first.
2. Imho, just leave it as is.
3. I would add a blank table method. I think maybe tests use it, but not sure it will ever hit table. Seems harmless.
4. I agree with subclassing. JsonShell should probably still print errors and such to stderr since a process would watch for stderr as a sign of failure. If json shell is just the same old shell except for things we convert to json, it starts as a clean slate and we can add on more as we find what we need.

martinemde:
Hey, I've been following along and it's looking good. Thanks for tackling this. 


I'm really curious to see if we get a bunch of requests after this is merged for json versions of all of the output.

nevinera:
In a normal engineering team, I'd probably divide this into two pieces (extracting the `table` method into UI::Shell` as one PR, and then adding the `JsonShell` as another) to keep the PRs as easily reviewable as possible. I suspect that larger PRs implementing features in full are preferable here though? (I can keep things readable by commit instead)

bundler/lib/bundler/cli/outdated.rb
```diff
@@ -166,46 +201,18 @@ def by_group(gems, filter: nil)
       end.compact
     end
 
-    def print_gems(gems_list)
+    def parseably_print_gems(gems_list)
```
martinemde, line 204: I can't quite get used to parsing this word "parseably". Does `parseable_print_gems` seem acceptable? Or if you have another idea.

bundler/lib/bundler/ui/json_shell.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Bundler
+  module UI
+    class JsonShell < UI::Shell
+      def table(header, rows, options = {})
+        return unless info?
+        empty_row = Hash[header.keys.map {|k| [k, nil] }]
```
martinemde, line 8: Avoid one intermediate Array.

```suggestion
        empty_row = header.keys.to_h {|k| [k, nil] }
```

bundler/lib/bundler/ui/json_shell.rb
```diff
@@ -0,0 +1,39 @@
...
+        data = rows.map {|row| empty_row.merge(row) }
+
+        require "json"
+        json_text =
+          if options[:pretty] && JSON.respond_to?(:pretty_generate)
+            JSON.pretty_generate(data)
+          else
+            data.to_json
```
martinemde, line 16: For parity with pretty

```suggestion
            JSON.generate(data)
```

bundler/spec/bundler/cli_spec.rb
```diff
@@ -161,16 +161,16 @@ def out_with_macos_man_workaround
...
+        expect(out_data.length).to eq(1)
+
+        expect(out_data.first).to eq({
+          "gem_name" => "rack",
+          "current_version" => "0.9.1",
+          "latest_version" => "1.0.0",
+          "latest_required_ruby" => ">= 0",
+          "latest_required_rubygems" => ">= 0",
+          "requested_version" => "= 0.9.1",
+          "groups" => "default",
```
martinemde, line 173: What's this do if there's more than 1 group? There should at least be a test.

I'm guessing we lose the array because of the to_s, and I'm thinking it gets concatenated. Would it be better to allow whatever comes in for types and expect the caller to convert types to string if necessary?

bundler/spec/bundler/ui/json_shell_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    end
+
+    it "constructs and prints the table to stdout properly" do
+      # * includes a nil-values entry for the missing :bar
+      # * includes the non-header-ed :huh key
+      expanded_json =
+        <<~JSON
+          [
+          {"foo":"Entry 1","bar":5,"baz":"a few words"},
+          {"foo":"two","bar":"five","baz":"less","huh":"extra data"},
```
martinemde, line 79: I like that it prints extra keys for extended data compared to `table`. Can we change `huh` to something like `"extra_keys" => "display extended data beyond table"`. I initially skipped the comment so I thought it was a mistake that `huh` was expected.

martinemde (approved):
This is really clean and looks great. I added a few nitpicks mostly to save an array here and there. I think maybe the most important thing is checking the groups output for more than 1 group and ideally making that an array again.

martinemde:
> Just as a side-effect of the fact that it needs to get out of the way for the json output that that tool doesn't actually (yet) write. If anyone has a suggestion that would make this less awkward let me know, because it kind of bugs me.

You're right. 🤔

We could turn info into json, even if it's ugly.

```json
{ "info": "messages that was sent." }
```

bundler/spec/bundler/cli_spec.rb
nevinera: So, the original presentation glues them together with commas, and my initial `--json` implementation split them back apart. The way it would have to work if we want them to be an array in the json output is that the `Shell#table` implementation would need to detect that it got an array, and `map(&:to_s).join(", ")` them, which .. isn't awful, but I wasn't sure was really generalizable enough.

I think I'd like it a little better though, and if you're also game I'll go ahead and add it :-)

bundler/spec/bundler/ui/json_shell_spec.rb
nevinera: That's definitely clearer (and yeah, now that it'll support them I'll add that as a case)

nevinera:
> > Just as a side-effect of the fact that it needs to get out of the way for the json output that that tool doesn't actually (yet) write. If anyone has a suggestion that would make this less awkward let me know, because it kind of bugs me.
> 
> You're right. 🤔
> 
> We could turn info into json, even if it's ugly.
> 
> ```json
> { "info": "message that was sent." }
> ```
> 
> This makes me think pretty printing is a bit of a problem because the assumption of line delimited json objects is required to parse this ugly format.
> 
> I still think err should be text.

That would make the actual _output_ of bundle-outdated not very parseable - ndjson is an irritating format in some ways, and it's tough to deal with if the json itself contains any newlines, which iirc the json-writer is

bundler/lib/bundler/cli/outdated.rb
nevinera: My fingers had trouble typing it correctly too >.<

`parseable_print_gems` doesn't have the right .. conjugation? I may be using that word wrong. Naming is hard though - `print_gems` is called for its side-effects, so it probably should be named with a verb-phrase, but I'm not too picky which one - `print_gems_carefully`, `print_gems_in_legacy_style`, `print_gems_manually`, `print_gems_locally`, `print_gems_for_parsing`? I'll use `print_gems_as_porcelain` for now, and update to whatever you prefer when you respond :-)

(My actual instinct was to extract that behavior into an inner class called `LegacyOutputGenerator` which would suffice to contextualize the printing instead, but it felt like too large a change for a PR that isn't purely refactoring. And I'm not sure how the maintainers feel about that sort of change yet)

bundler/spec/bundler/cli_spec.rb
nevinera: Added, and it cleaned up the handling of `groups` in Cli::Outdated a bit overall.

bundler/lib/bundler/ui/json_shell.rb
martinemde: We just dropped support for 2.6 and 2.7 on master. Luxurious!

bundler/spec/bundler/cli_spec.rb
martinemde: Ah ok, I thought it received an array already so I was incorrect in my review. I see it's not so simple now because of how Shell handles it. 

So the choice is between preformatted csv in the json or detecting the array in Shell.

I agree with detecting the array.

nevinera:
(resolving a legit problem the tests spotted, but I have to run for a few hours)

nevinera:
I'm not sure what to do about the "bundler 2 on windows" checks, if they don't self-resolve on the next run - they failed after more than an hour, with [no output](https://github.com/rubygems/rubygems/actions/runs/6997859587/job/19035290183). I had a mysterious (quite different) problem with one of them earlier too [here](https://github.com/rubygems/rubygems/actions/runs/6993275701/job/19025770159) that seemed to go away on the next push. Are they just kind of flaky?

martinemde:
I've spent a bunch of time trying to track down why the windows tests are flakey, with some success. The worst part about them is that Windows is much much slower because the tests are very I/O heavy.

martinemde (approved):

martinemde:
@nevinera Could you rebase squash this? I don't mind too much normally but we're at 14 commit, hehe. I'll merge after the squash.

nevinera:
> @nevinera Could you rebase squash this? I don't mind too much normally but we're at 14 commit, hehe. I'll merge after the squash.

Totally - I try to avoid rebasing commits once they've been reviewed, cause it can make re-reviewing irritating, but I'd prefer not to release such a pile of commits, for sure. I think it would be a fair amount of work to collapse them into 2-3 isolated commits like I'd want them to be ideally, so I'll just squash them like you suggest :-)

martinemde:
This is great work and solid code, @nevinera. I appreciate how thorough you were and I enjoyed working with you on this. Hopefully there's more to do soon.

nevinera:
> This is great work and solid code, @nevinera. I appreciate how thorough you were and I enjoyed working with you on this. Hopefully there's more to do soon.

I'm grateful for your willingness to help a project-newbie :-D

simi:
Hello @nevinera and thanks for opening this. I would like to take a look as well before merging. I had some ideas and questions around for this.

Btw. feel free to join use at Slack, we're ready to help you in there if needed. Link is at the bottom of https://bundler.io/.

nevinera:
> Hello @nevinera and thanks for opening this. I would like to take a look as well before merging. I had some ideas and questions around for this.
> 
> Btw. feel free to join us at Slack, we're ready to help you in there if needed. Link is at the bottom of https://bundler.io/.

Totally, let me know what I can add or improve :-)

Thanks for pointing me at the slack, I somehow missed it when I was digging in initially.

martinemde:
@simi Sorry to merge before you could take a look. I'm glad you caught it.

nevinera:
Oh, what's the release-cadence like? If there might be a bundler release before I can get this in, I probably should revert the other PR, since this does end up changing the shape of the resulting json pretty substantially (to make it match the data used by UI::Shell better).

simi:
@nevinera no worries, AFAIK features are cherry-picked from main branch to each release candidate. If needed, some PRs can be opted out. If changes are expected, we can mark the new `--json` option to `outdated` command as "experimental" and subject to change in the future. WDYT @deivid-rodriguez?

nevinera:
I'm not sure what the typical cadence is for feedback - you're clearly _incredibly_ busy @simi (you're in every conversation I look at), but I just wanted to make sure you didn't forget about me. Happy to chat in slack instead if you have something more in-depth you wanted to talk through?

nevinera:
(updated to use modern hash syntax, which is such a relief. Training my fingers to use hash-rockets again in just this project was quite a chore)

martinemde (approved):
Still looks great to me.

martinemde:
Hey @nevinera, we're planning on rubygems & bundle release really soon and we had the idea to revert the first PR that added this feature, then combine adding the original feature with this PR.

I'm letting you know so you can re-add the revert-revert to the beginning of this PR to keep it from conflicting with master. I'll push the revert to master soon.

nevinera:
> Hey @nevinera, we're planning on rubygems & bundle release really soon and we had the idea to revert the first PR that added this feature, then combine adding the original feature with this PR.
> 
> I'm letting you know so you can re-add the revert-revert to the beginning of this PR to keep it from conflicting with master. I'll push the revert to master soon.

Makes total sense to me - it's what I intended when I first got this thing set up. Better not to let anyone build something one what's in there now if I'm going to rearrange it shortly!

nevinera:
@simi - will you have a chance to look at this soon?

nevinera:
> I had some ideas and questions around for this.

@simi: I'd like to move on to other bundler work, but I don't like to carry many open PRs at once. Do you think you'll have time to look at this soon, or is there a way I can re-slice this to be more incrementally releasable?

simi:
Hello @nevinera and thanks for your patience with me in here. I was looking at this already few times, but indeed this seems a little tricky to review since it is composed of fixes for few different problems coupled together.

1. `--parseable` seems actually harder to parse than original output
2. original tabular output will be easier to parse if turned into actual TSV format
3. original output outputs also definition resolution info (aka `Fetching gem metadata...`) breaking parsing
4. newly introduced `JsonShell` actually doesn't output JSON for all messages, but only for table and redirects the rest for stderr and is not reusable IMHO
5. What kind of sense would make JsonShell (or JSON output in general) for other commands like bundle install?

Looking at those, what about to pick a little different strategy? Would it be possible to deprecate `--parseable` and replace it with `--tsv` option (or even replace original output with tsv one) and use `Bundler.ui.silence` for definition resolution. It could be enough to keep the logic in `Outdated` class for now. Outputing this tabular data as JSON doesn't make a much sense to me since there is nothing like "table" in JSON and you need to define your own structure (like expose values as array of key/value pairs or different structure). It could be also possible to move TSV "rendering" into `Shell` class `table` method.

If you would like to keep JSON output over TSV, I'm ok with that as well, but since JSON most likely doesn't make sense for table, I think there is no need to pretend JSON table is being rendered. Would it make sense to reuse `Bundler.ui.silence` to get clean output, create some kind of `OutdatedSet` class keeping mostly list of `OudatedGem` classes (could be simple inline struct) and ensure it provides data needed for current rendering (of table and --parseable), but also it implements simple `to_json` method called conditionally for `--json` flag? Clearly there are better names for those classes. This JSON approach would make this outdated class a little more simpler, since the logic for getting the difference in between current lock and updated lock would be "hidden" in there.

WDYT?

If you would like to iterate on this in multiple steps, I think for now it would be ok to remove JsonShell (it seems it is not good idea, sorry for this misdirection), migrate json table into Shell as `json_table` method (or keep it inline in `Outdated`). I can help with following PRs.

PS: This PR revealed to me some interesting ideas, like should be progress report send to stderr by default in general? IMHO it makes sense, but at some places it would be controversial what is actual progress report. @deivid-rodriguez WDYT?

nevinera:
> Hello @nevinera and thanks for your patience with me in here. I was looking at this already few times, but indeed this seems a little tricky to review since it is composed of fixes for few different problems coupled together.

It's not really intended to be - there was some structural cleanup (there was some dead code and minor refactoring needed in CLI::Outdated to support the functional changes), but there are really fixes for two problems here that I see - adding `json` support and decomposing the `table` output into `Bundler::UI::Shell`. I did originally have the former in isolation, but you requested that the latter be bundled with it (and I agree that it's a better solution).

>     1. `--parseable` seems actually harder to parse than original output
> 
>     2. original tabular output will be easier to parse if turned into actual TSV format
> 
>     3. original output outputs also definition resolution info (aka `Fetching gem metadata...`) breaking parsing
> 
>     4. newly introduced `JsonShell` actually doesn't output JSON for all messages, but only for table and redirects the rest for stderr and is not reusable IMHO
> 
>     5. What kind of sense would make JsonShell (or JSON output in general) for other commands like bundle install?

I'm not sure how to interpret this list - given the later items, it looks like a list of problems with the changeset, but the first item throws me, because I don't believe I changed the format of the `--parseable` output at all (I treated it as a legacy format that needed to be kept consistent for compatibility reasons). 

> Looking at those, what about to pick a little different strategy? Would it be possible to deprecate `--parseable` and replace it with `--tsv` option (or even replace original output with tsv one) and use `Bundler.ui.silence` for definition resolution. It could be enough to keep the logic in `Outdated` class for now. Outputing this tabular data as JSON doesn't make a much sense to me since there is nothing like "table" in JSON and you need to define your own structure (like expose values as array of key/value pairs or different structure). It could be also possible to move TSV "rendering" into `Shell` class `table` method.

I think deprecating `--parseable` and adding `--tsv` (or `--csv` instead/as well?) isn't a bad idea, but I suspect there are a lot of people using `--parseable. We'd probably want to add `--tsv` for a version or two before deprecating `--parseable`, right? I think I'd want to tackle `--tsv` in a separate PR, but I suspect it would be a small one. 

> If you would like to keep JSON output over TSV, I'm ok with that as well, but since JSON most likely doesn't make sense for table, I think there is no need to pretend JSON table is being rendered. Would it make sense to reuse `Bundler.ui.silence` to get clean output, create some kind of `OutdatedSet` class keeping mostly list of `OudatedGem` classes (could be simple inline struct) and ensure it provides data needed for current rendering (of table and --parseable), but also it implements simple `to_json` method called conditionally for `--json` flag? Clearly there are better names for those classes. This JSON approach would make this outdated class a little more simpler, since the logic for getting the difference in between current lock and updated lock would be "hidden" in there.
> 
> WDYT?

My initial goal was to have json (or yaml) output for bundle-outdated, so I could get extra data out of `bundle-outdated`, and avoid writing parsers for data with no structural api contract. Your approach sounds like a cleaner version of my original PR, really, and I'd be happy to take a crack at it - decomposing some of the gem-version comparison logic out of CLI::Outdated sounds like an excellent idea, if that's what you mean.

> If you would like to iterate on this in multiple steps, I think for now it would be ok to remove JsonShell (it seems it is not good idea, sorry for this misdirection), migrate json table into Shell as `json_table` method (or keep it inline in `Outdated`). I can help with following PRs.
> 
> PS: This PR revealed to me some interesting ideas, like should be progress report send to stderr by default in general? IMHO it makes sense, but at some places it would be controversial what is actual progress report. @deivid-rodriguez WDYT?

I generally consider progress reporting to be "logging", and to belong on stderr. But it doesn't really _matter_ except in tools that are intended to produce output as their primary purpose. Most of the tools _do_ just need a UI::Shell abstraction, wrapping "doing things while logging what has been done", but a few of them could use a separate abstraction, because they want to "produce data and give it to you". Maybe a `Bundler::UI::OutputShell`, with some kind of `OutputFormatter` class hierachy that does the actual output dumping? `bundle outdated` is one of those, but there are a few others that could benefit from such a system - `bundle info`, `bundle list`, `bundle platform`, and maybe `bundle version` could all use that abstraction.

-----

I'm going to close this PR and try to scrape together two more (one for `--tsv` and one for `--json` as described - I'll probably find time later this week. Thanks for your insight @simi  - I know it's been a really busy season for rubygems in general, and I appreciate your hard work.

So, closing questions here:

1. TSV, CSV, or both? (separate PR for that when I get to it, either way)
2. Should I keep the `UI::Shell#table` extraction? It's reusable, though not fancy, and it does decompose some gross _out_ of CLI::Outdated..
3. Decompose an OutdatedGems data structure out of CLI::Outdated and drop JsonShell in favor of letting OutdatedGems handle `to_json`. Does that sound about right? CLI::Outdated still does the fetching; OutdatedGems is just a data structure with good access and setup methods.

simi:
> TSV, CSV, or both? (separate PR for that when I get to it, either way)

Pick whatever is useful for you. Adding TSV just because it is possible makes no sense if there's no known usecase.

> Should I keep the UI::Shell#table extraction? It's reusable, though not fancy, and it does decompose some gross out of CLI::Outdated..

Yup, I like that a lot.

> Decompose an OutdatedGems data structure out of CLI::Outdated and drop JsonShell in favor of letting OutdatedGems handle to_json. Does that sound about right? CLI::Outdated still does the fetching; OutdatedGems is just a data structure with good access and setup methods.

It sounds as an good abstraction to me, but better to confirm with other maintainers first. @deivid-rodriguez @martinemde @segiddins IMHO those CLI classes should just get input, pass it somewhere to Bundler classes and print the output. Today there is a lot of complex stuff happening inside and Outdated is good example and ideal candidate to extract the logic outside.

---

> Thanks for your insight @simi - I know it's been a really busy season for rubygems in general, and I appreciate your hard work.

_I'm again super sorry for long time without my response. I'm back currently and I'll be more responsive :pray:. Thanks for your contributions and patience._

PR #7199, opened 2023-11-27
Reduce array allocations when loading definition

segiddins:
The same array was being re-created in a loop (as well as the `generic_local_platform`), which is avoidable by hoisting it to a frozen array created once

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -245,8 +245,9 @@ module Bundler
     end
 
     def filter_relevant(dependencies)
+      platforms_array = [generic_local_platform].freeze
       dependencies.select do |d|
-        d.should_include? && !d.gem_platforms([generic_local_platform]).empty?
+        d.should_include? && !d.gem_platforms(platforms_array).empty?
       end
     end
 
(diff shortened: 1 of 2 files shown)


martinemde (approved):

bundler/lib/bundler/definition.rb
```diff
@@ -270,9 +271,15 @@ def specs_for(groups)
...
       groups.map!(&:to_sym)
-      current_dependencies.reject do |d|
-        (d.groups & groups).empty?
+      deps = current_dependencies # always returns a new array
+      deps.select! do |d|
+        if RUBY_VERSION >= "3.1"
+          d.groups.intersect?(groups)
+        else
+          !(d.groups & groups).empty?
+        end
```
deivid-rodriguez, line 280: Does `reject!` work?

deivid-rodriguez: Oh so using `intersect?` is also for performance? I thought the performance was just modifying the array in place, and while intersect is cool, didn't seem worth the extra complexity of having two code branches, etc. But it's fine either way, specially if this way is faster.

deivid-rodriguez (approved):

bundler/lib/bundler/definition.rb
segiddins: `intersect?` saves the array allocation from `&`

martinemde: Is it worth calculating which version to use outside the block? Is the RUBY_VERSION check is doing anything complex like splitting strings?

PR #7200, opened 2023-11-28
Do not pollute the global namespace

BertramScharpf:
Inside a Gem executable you get:

    TOPLEVEL_BINDING.local_variables    #=> [:version, :str]

This is not fair.

I decided to propose an `Object#tap` call because defining a function would pollute `TOPLEVEL_BINDING.private_methods`.

Maybe an even better way to solve the problem is to put the variable assignments and usage into a `module Gem ; ... ; end` block.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -766,23 +766,23 @@ class Gem::Installer
 
 require 'rubygems'
 #{gemdeps_load(spec.name)}
-version = "#{Gem::Requirement.default_prerelease}"
+"#{Gem::Requirement.default_prerelease}".tap { |version|
 
-str = ARGV.first
-if str
-  str = str.b[/\\A_(.*)_\\z/, 1]
-  if str and Gem::Version.correct?(str)
-    #{explicit_version_requirement(spec.name)}
-    ARGV.shift
+  if ARGV.first =~ /\\A_(.*)_\\z/
+    str = $1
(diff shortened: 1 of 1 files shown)


hsbt:
What's your issue?

BertramScharpf:
> What's your issue?

It's not a technical problem. It's a matter of style. My understanding of good programming style is to leave with a clean global name space.

lib/rubygems/installer.rb
```diff
@@ -766,23 +766,23 @@ def app_script_text(bin_file_name)
...
-if str
-  str = str.b[/\\A_(.*)_\\z/, 1]
-  if str and Gem::Version.correct?(str)
-    #{explicit_version_requirement(spec.name)}
-    ARGV.shift
+"#{Gem::Requirement.default_prerelease}".tap { |version|
+
+  if ARGV.first =~ /\\A_(.*)_\\z/
+    str = $1
+    if Gem::Version.correct?(str)
```
martinemde, line 773: This code is not strictly equivalent to the original code. I don't know with certainty that it would cause any problem, but this is a critical path for almost everything in ruby.

martinemde (approved):
I understand your concern is that a gem executable could not rely on str or version being unaltered in the top level namespace. You would need to compensate for this if you had a very strict need. However given that this is a critical path for every gem executable, I would suggest we need a better reason for changing it than style alone.

PR #7203, opened 2023-11-28
Use globals variables for standard input/output

voxik:
Replace use of `STDIN`, `STDOUT` and `STDERR` constants by their `$stdin`, `$stdout` and `$stderr` global variable equivalents.

This enables easier testing via standard means, such as `assert_output` for minitest or `expect { print 'foo' }.to output.to_stdout` for RSpec test suites.

In eaa315b95ee4887f1980a5819a05dc2d3d49f4c1 the `Gem::MockGemUi` was hidden from external use, breaking other [project](https://github.com/enkessler/childprocess/issues/190) test cases. I thought it should be pretty straight forward to not use the `Gem::MockGemUi`, so the test case could look like:

~~~
expect { gemspec.validate }.not_to output(/warn/i).to_stderr
~~~

However, RubyGems using `STDERR`, that did not work.

I can't see why RubyGems should use `STDERR`, when it could use `$stderr` instead, enabling usage of standard test frameworks methods to check the output.

However, I have not tried to execute the test suite, so I might be surprised.

Please note that there are also other places where the constants are still used, especially in Bundler.

### DIFF
--- a/lib/rubygems/user_interaction.rb
+++ b/lib/rubygems/user_interaction.rb
@@ -193,7 +193,7 @@ class Gem::StreamUI
   # then special operations (like asking for passwords) will use the TTY
   # commands to disable character echo.
 
-  def initialize(in_stream, out_stream, err_stream=STDERR, usetty=true)
+  def initialize(in_stream, out_stream, err_stream=$stderr, usetty=true)
     @ins = in_stream
     @outs = out_stream
     @errs = err_stream
@@ -591,8 +591,8 @@ class Gem::StreamUI
 end
 
 ##
-# Subclass of StreamUI that instantiates the user interaction using STDIN,
-# STDOUT, and STDERR.
+# Subclass of StreamUI that instantiates the user interaction using $stdin,
+# $stdout, and $stderr.
 
 class Gem::ConsoleUI < Gem::StreamUI
   ##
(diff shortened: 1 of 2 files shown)


voxik:
I have also included test case. Not sure about the new file name, but it follows the `test/rubygems/test_gem_stream_ui.rb` example. It actually used to be `test/rubygems/test_user_interaction.rb` file, but it was [renamed ~15 years ago](https://github.com/rubygems/rubygems/commit/8c054fdc9676f912b6464961854ff10fe1c9940c#diff-bf2d940d6d9da9f9d6d61594746f2b722eb76382c173668e8ca324ca9a768c20) 🤷‍♂️

martinemde (approved):
While this is best practice in most ruby programs, it is a potential behavior change. STDOUT should always be constant (at least a warning is printed if it changes), $stdout can and does get changed.

The only complication I can imagine is cases where rubygems is relying on $stderr to print feedback and there is a possibility of running other code that could redefine $stdout, preventing rubygems from giving user feedback.

For example, is it possible for a gem exec call to load a gem that redefines stdout/stderr causing gem exec's own output to get suppressed?

In my opinion this isn't a big concern. Redefining stdout is rare and usually leakage is annoying rather than useful. However, rubygems is such a low level library, I would not be surprised if it's expected to ignore this (like how ruby should still show a crash or interpreter warnings even if stdout is redefined)

voxik:
I have the same concerns :innocent: However, what might be the situation when application redefines `$stderr` while it would still like to keep RubyGems error output somewhere else is beyond me. Right, it is behavior change in a way, that suddenly RubyGems output will be mixed with the application output, but that might be concern e.g. for test suite and not for real life situation. Anyway, I still consider this change for good.

martinemde:
@deivid-rodriguez did I understand you right about merging this? I'm happy with this change myself and we'll see if it causes any issues 🫣

PR #7205, opened 2023-11-29
Consolidate ci detection envs

### LINKED ISSUE #6038: Consolidate ENV list for CI detection
We currently have two different list of environment variables for CI detection. One used by Bundler to include the CI where it's running with request headers, so that we can later on extract metrics. Another one introduced in RubyGems to be able to avoid needless call for updates messages on CI environments.

Since these lists are used for the same thing (detecting whether we are running on a CI environment), we should probably unify them and keep them in the same place.

nevinera:
There were two separate lists of environment variables to maintain that represented our ability to determine the CI environment, which made maintaining and updating them difficult.

I've introduced the `Gem::CIDetector`, a single abstraction responsible for "detecting the CI environment (if any)". It exposes `ci?` and `ci_strings` methods, which represent the usages the two existing sites have, and it clearly documents the provenance and expectations it makes about the presence of specific environment variables for various CI providers.

It also introduces a few effective behavioral _changes_ (though I could slice those off into another PR if preferred):

* We'll consider ourselves to be on a CI in more cases - if CI_NAME or TASKCLUSTER_ROOT_URL are set specifically, since those represent two cases that were either overlooked or are no longer covered by the existing implementation. (Or possibly TaskCluster still does provide RUN_ID, but failed to document it)
* We will notice/track a few additional services in ci_strings (cirrus, dsari, taskcluster), stop tracking 'snap' (they went under in 2017), and update buildbox to buildkite (they've been called that for 8 years, and the BUILDBOX envs have been deprecated for 3).
* We'll also sort/uniq/downcase the values (all of which only matter because of the special case of CI_NAME - probably only the `uniq` will matter in practice, as some of the providers supply `CI_NAME` alongside their own named ENV.

I also found [this](https://github.com/watson/ci-info) more complete list, and would be happy to build CIDetector out a bit more (here or in another PR) if you'd like.

(Resolves #6038)

### DIFF
--- a/bundler/lib/bundler/ci_detector.rb
+++ b/bundler/lib/bundler/ci_detector.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+module Bundler
+  module CIDetector
+    # NOTE: Any changes made here will need to be made to both lib/rubygems/ci_detector.rb and
+    # bundler/lib/bundler/ci_detector.rb (which are enforced duplicates).
+    # TODO: Drop that duplication once bundler drops support for RubyGems 3.4
+    #
+    # ## Recognized CI providers, their signifiers, and the relevant docs ##
+    #
+    # Travis CI   - CI, TRAVIS            https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
+    # Cirrus CI   - CI, CIRRUS_CI         https://cirrus-ci.org/guide/writing-tasks/#environment-variables
+    # Circle CI   - CI, CIRCLECI          https://circleci.com/docs/variables/#built-in-environment-variables
+    # Gitlab CI   - CI, GITLAB_CI         https://docs.gitlab.com/ee/ci/variables/
+    # AppVeyor    - CI, APPVEYOR          https://www.appveyor.com/docs/environment-variables/
+    # CodeShip    - CI_NAME               https://docs.cloudbees.com/docs/cloudbees-codeship/latest/pro-builds-and-configuration/environment-variables#_default_environment_variables
+    # dsari       - CI, DSARI             https://github.com/rfinnie/dsari#running
+    # Jenkins     - BUILD_NUMBER          https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#using-environment-variables
+    # TeamCity    - TEAMCITY_VERSION      https://www.jetbrains.com/help/teamcity/predefined-build-parameters.html#Predefined+Server+Build+Parameters
+    # Appflow     - CI_BUILD_ID           https://ionic.io/docs/appflow/automation/environments#predefined-environments
+    # TaskCluster - TASKCLUSTER_ROOT_URL  https://docs.taskcluster.net/docs/manual/design/env-vars
+    # Semaphore   - CI, SEMAPHORE         https://docs.semaphoreci.com/ci-cd-environment/environment-variables/
+    # BuildKite   - CI, BUILDKITE         https://buildkite.com/docs/pipelines/environment-variables
+    # GoCD        - GO_SERVER_URL         https://docs.gocd.org/current/faq/dev_use_current_revision_in_build.html
+    # GH Actions  - CI, GITHUB_ACTIONS    https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+    #
+    # ### Some "standard" ENVs that multiple providers may set ###
+    #
+    # * CI - this is set by _most_ (but not all) CI providers now; it's approaching a standard.
+    # * CI_NAME - Not as frequently used, but some providers set this to specify their own name
+
+    # Any of these being set is a reasonably reliable indicator that we are
+    # executing in a CI environment.
+    ENV_INDICATORS = [
+      "CI",
+      "CI_NAME",
+      "CONTINUOUS_INTEGRATION",
+      "BUILD_NUMBER",
+      "CI_APP_ID",
+      "CI_BUILD_ID",
+      "CI_BUILD_NUMBER",
+      "RUN_ID",
+      "TASKCLUSTER_ROOT_URL",
+    ].freeze
+
+    # For each CI, this env suffices to indicate that we're on _that_ CI's
+    # containers. (A few of them only supply a CI_NAME variable, which is also
+    # nice). And if they set "CI" but we can't tell which one they are, we also
+    # want to know that - a bare "ci" without another token tells us as much.
+    ENV_DESCRIPTORS = {
+      "TRAVIS" => "travis",
+      "CIRCLECI" => "circle",
+      "CIRRUS_CI" => "cirrus",
+      "DSARI" => "dsari",
+      "SEMAPHORE" => "semaphore",
+      "JENKINS_URL" => "jenkins",
+      "BUILDKITE" => "buildkite",
+      "GO_SERVER_URL" => "go",
+      "GITLAB_CI" => "gitlab",
+      "GITHUB_ACTIONS" => "github",
+      "TASKCLUSTER_ROOT_URL" => "taskcluster",
+      "CI" => "ci",
+    }.freeze
+
+    def self.ci?
+      ENV_INDICATORS.any? {|var| ENV.include?(var) }
+    end
+
+    def self.ci_strings
+      matching_names = ENV_DESCRIPTORS.select {|env, _| ENV[env] }.values
+      matching_names << ENV["CI_NAME"].downcase if ENV["CI_NAME"]
+      matching_names.reject(&:empty?).sort.uniq
+    end
+  end
+end
--- a/bundler/lib/bundler/fetcher.rb
+++ b/bundler/lib/bundler/fetcher.rb
@@ -95,6 +95,7 @@ module Bundler
     self.max_retries    = Bundler.settings[:retry] # How many retries for the API call
 
     def initialize(remote)
+      @cis = nil
       @remote = remote
 
       Socket.do_not_reverse_lookup = true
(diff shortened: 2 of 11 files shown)


bundler/lib/bundler/fetcher.rb
```diff
@@ -5,6 +5,7 @@
 require "securerandom"
 require "zlib"
 require "rubygems/request"
+require "rubygems/ci_detector"
```
simi, line 8: It is possible to run old RubyGems with new Bundler and this file could be missing. What to do in that case? Should we keep the current logic around as a fallback for some time? @deivid-rodriguez any idea what's common practice?

nevinera: That.. actually probably explains the problem I'm trying to resolve locally and in CI right now >.<

nevinera: Does make any shared-code plan kind of awkward though. I'm sure it's further than *I* ought to go, but it seems like there might be space for a shared library of some sort, which would get packaged with both systems?

nevinera: I guess keeping the existing bundler/fetcher code for a while makes sense - should I check `Gem.autoload?` for that?

martinemde: This is where the solutions to this type of problem live:

https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/rubygems_ext.rb

We have to be careful not to make a huge mess though. I'm not sure if duplicating the code into here is better or worse.

nevinera: This doesn't look.. _central_ enough to belong in there, I think. What's the cross-version compatibility policy? If it's a question of keeping an extra copy of one of these methods for a few releases, that seems fine to me, but if we're talking about a long term situation then I'm not sure this ticket is worth resolving :-/

Though I think the research/documentation and the data fixes are worth keeping anyway, so I'd prefer to just split it into parts and keep the parts with the projects that use them. And possibly extract the big comment into a markdown file they both reference?

The other idea that occurs to me is to drive the behaviour from a config file, and include that in both builds. But the logic for memoizing the parsed config would be larger than what we're replacing, so it would seem silly.

bundler/lib/bundler/fetcher.rb
```diff
@@ -4,6 +4,7 @@
 require "cgi"
 require "securerandom"
 require "zlib"
+require "rubygems"
```
martinemde, line 7: I don't think it's possible that this isn't loaded, is it? I could totally be wrong.

lib/rubygems/update_suggestion.rb
```diff
@@ -62,8 +53,4 @@ def eligible_for_update?
   rescue StandardError # don't block install command on any problem
     false
   end
-
-  def ci?
```
martinemde, line 66: Any chance someone uses this? I'm thinking that even if someone does, it's better to break it and have them discover the better named CIDetector.

martinemde (approved):
This looks like a clear improvement, and the maintenance cost of duplication is no worse than before (with an eventual offramp to a single source). I'm for merging this.

martinemde:
Heheh, test failures on CI. Who could have predicted 😆

bundler/lib/bundler/fetcher.rb
nevinera: it's not - there's a `require "rubygems/request"` right below, if nothing else. I'll yank it.

lib/rubygems/update_suggestion.rb
nevinera: oh wow, you mean externally? That didn't occur to me - I checked for internal usage, but I guess it's a public method >.< `UpdateSuggestion.ci?` doesn't seem like a very _natural_ thing to call, at least..

I can

```
def ci?
  Gem::CIDetector.ci?
end
```

if you think there's any chance though.

nevinera:
> Heheh, test failures on CI. Who could have predicted 😆

.. Right. Because `ENV["CI"]` _IS_ set here, and so are the github-actions ones.
The test-helper I wrote in the other PR overrode the entire env, but the one we're using here doesn't do that. I'll update these specs to handle ENV a little more thoroughly.

bundler/lib/bundler/fetcher.rb
nevinera: A similar thought occurred to me - my idea was just to _enforce_ that they match in CI, rather than to generate one from the other, since I feel like machine-generated commits make for complicated workflows. Would that appeal to you? It feels less complex than an actual generator, and the file's not really large enough that it'll hurt much to have to duplicate changes across on the occasion that they're made..

martinemde: I like your idea of writing a test that makes sure we keep them synced. If someone is editing one, the test will help make sure they fix the other one.

nevinera: Cool, I'll get a version of that up in a bit :-)

nevinera: Alright, took a couple cycles to understand how the path helpers worked (`ruby_core?`), but that should do it - there are now `Gem::CIDetector` and `Bundler::CIDetector`; the spec for the later doesn't test functionality, just textual equivalence.

bundler/lib/bundler/ci_detector.rb
```diff
@@ -0,0 +1,75 @@
...
+      "TASKCLUSTER_ROOT_URL" => "taskcluster",
+      "CI" => "ci",
+    }.freeze
+
+    def self.ci?
+      ENV_INDICATORS.any? {|var| ENV.include?(var) }
+    end
+
+    def self.ci_strings
+      matching_names = ENV_DESCRIPTORS.find_all {|env, _| ENV[env] }.map {|_, ci| ci }
```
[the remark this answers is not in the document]
nevinera: too late! And this is definitely safe now that I bothered checking the support for the methods - I was mostly worried about Hash#select, which _felt_ recent to me (and this switch would stop supporting the enumerable-based version). But it looks like it was actually ruby 2.4, and I'm just old -.-

bundler/spec/bundler/fetcher_spec.rb
```diff
@@ -142,19 +142,20 @@
     end
 
     describe "include CI information" do
+      # We actually run these tests _on_ github actions, so these may be set during CI
+      let(:hide_vars) { { "CI" => nil, "GITHUB_ACTIONS" => nil, "CI_NAME" => nil } }
```
martinemde, line 146: This approach may be brittle to rubygems running on other CIs. What do you think of calculating the output before and after you add JENKINS_URL and checking if it includes the same set plus jenkins (possibly allowing it to already have jenkins)?

Or you could just revert to the match based test.

martinemde (commented):
One last little change. I think this should merge before the end of year release, if possible. Agree, @deivid-rodriguez?

bundler/spec/bundler/fetcher_spec.rb
nevinera: Oh, it would be - I didn't realize that was a concern!

I could just replace all of ENV for those tests, instead of setting individual values with the helper.. but your tests do really on ci ENV quite a bit. I can't _think_ of any way these particular tests could be using one, but your suggestion will be safer.

I'll get that in shortly :-)

deivid-rodriguez:
> One last little change. I think this should merge before the end of year release, if possible. Agree, @deivid-rodriguez?

Yes, this is good to ship!

bundler/spec/bundler/fetcher_spec.rb
nevinera: Sorry, that rebase also pulled the master changes in, which makes the diff hard to see imo. Here's what I did:

```patch
diff --git a/bundler/spec/bundler/fetcher_spec.rb b/bundler/spec/bundler/fetcher_spec.rb
index 332ea81c7..f764c44fa 100644
--- a/bundler/spec/bundler/fetcher_spec.rb
+++ b/bundler/spec/bundler/fetcher_spec.rb
@@ -142,20 +142,21 @@
     end
 
     describe "include CI information" do
-      # We actually run these tests _on_ github actions, so these may be set during CI
-      let(:hide_vars) { { "CI" => nil, "GITHUB_ACTIONS" => nil, "CI_NAME" => nil } }
-
       it "from one CI" do
-        with_env_vars(hide_vars.merge("JENKINS_URL" => "foo")) do
+        with_env_vars("CI" => nil, "JENKINS_URL" => "foo") do
           ci_part = fetcher.user_agent.split(" ").find {|x| x.start_with?("ci/") }
-          expect(ci_part).to eq("ci/jenkins")
+          cis = ci_part.split("/").last.split(",")
+          expect(cis).to include("jenkins")
+          expect(cis).not_to include("ci")
         end
       end
 
       it "from many CI" do
-        with_env_vars(hide_vars.merge("TRAVIS" => "foo", "GITLAB_CI" => "gitlab", "CI_NAME" => "MY_ci")) do
+        with_env_vars("CI" => "true", "SEMAPHORE" => nil, "TRAVIS" => "foo", "GITLAB_CI" => "gitlab", "CI_NAME" => "MY_ci") do
           ci_part = fetcher.user_agent.split(" ").find {|x| x.start_with?("ci/") }
-          expect(ci_part).to eq("ci/gitlab,my_ci,travis")
+          cis = ci_part.split("/").last.split(",")
+          expect(cis).to include("ci", "gitlab", "my_ci", "travis")
+          expect(cis).not_to include("semaphore")
         end
       end
     end
```

martinemde:
Thank you for this improvement, @nevinera.

PR #7209, opened 2023-11-30
Update `SharedHelpers#major_deprecation` to support a `removed_message` parameter

### LINKED ISSUE #3662: Improve `SharedHelpers.major_deprecation`

nevinera:
#3662 - Essentially, the `major_deprecation` helper is mostly used for _deprecation warnings_, but those deprecation warnings get automatically promoted into _errors_ after one major version, and the messages in many cases are either no longer exactly accurate or no longer make sense at all (as they were written about future plans, but are being read after that future has come to pass).

* Add an optional `removed_message` keyword argument to the method - if supplied, it will be used instead of `message` in the "current version is greater than the deprecating version" case. This allows the caller of the method to supply two messages, one for "You should stop doing that, cause it's going away" and another for "you can't do that Dave".
* Walk all of the _existing_ call-sites - for each of them:
  - if it is already written to make sense in each context, just leave it alone. They might be made clearer without the constraint of needing to be written that way, but I lack the context to do a good job of it (but feel free to supply replacement copy-pairs if any of those are bothering you and I'll add em in here)
  - otherwise, duplicate the message text (usually by extracting it to a pair of variables and literally copying the text), and update the `replaced_message` to be appropriately past-tense and refer to removals instead of deprecations, etc.

(Fixes #3662)


deivid-rodriguez (approved):
I think this is simple and gets the job done 👍. I think I'd maybe name the kwarg `removed_message`, what do you think?

nevinera:
Sure, that'll be a bit clearer - I'll update it this evening.

PR #7212, opened 2023-11-30
Fix `--build-root` option incorrectly installing to user dir inside build root

### LINKED ISSUE #7083: --build-root option is not respected anymore

deivid-rodriguez:

~I was hoping that this changes would fix #7083, but I can't reproduce the issue either in current master.~

Changes in #5327 made `--build-root` start installing into user dir inside the build root.

It builds the fix on a change in the approach to #5327 that makes it easier to fix this kind of issue and should also be more robust, since it does not require to keep a list of commands that the warning applies to, and to save whether we fallback or not to check the result somewhere else.

With the new approach in place, it gets simpler to fix the problem, since `build-root` is prepended in the exact same place where the fallback to installing to the user directory happens.

Fixes #7083.


lib/rubygems/installer.rb
```diff
@@ -676,7 +676,14 @@ def process_options # :nodoc:
...
     @gem_home = @install_dir
-    @gem_home ||= options[:user_install] ? Gem.user_dir : Gem.dir
+    @gem_home ||= if options[:user_install]
+      Gem.user_dir
+    elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
+      say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
+      Gem.user_dir
+    else
+      Gem.dir
+    end
```
duckinator, line 686: Adding this change will fix `--build-root` as well.

```suggestion
    @gem_home = @install_dir || Gem.dir

    # `--build-root` overrides `--user-install` and auto-user-install
    if @build_root.nil?
      if options[:user_install]
        @gem_home = Gem.user_dir
      elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
        say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
        @gem_home = Gem.user_dir
      end
    end
```

deivid-rodriguez: Yep, that makes sense to me! I'll work on complementing your fix with a test tomorrow!

duckinator: Made a slight edit to my suggestion (added `!options[:install_dir] && `) because I had accidentally broken `--install-dir` taking priority over `--user-install`.

duckinator:
@deivid-rodriguez here's a patch that takes this PR and both fixes `--build-root` and adds a test for it: [0001-Make-build-root-disable-auto-user-install.patch](https://github.com/rubygems/rubygems/files/13519506/0001-Make-build-root-disable-auto-user-install.patch)

If you'd like, I can push the changes myself.

Otherwise, you can download the patch file and run `git am 0001-Make-build-root-disable-auto-user-install.patch`.

(Since this is your PR, I didn't want to push changes without your permission.)

lib/rubygems/installer.rb
duckinator: Marking as resolved, in favor of the patch (with a test) I mentioned at <https://github.com/rubygems/rubygems/pull/7212#issuecomment-1835268767>.

lib/rubygems/installer.rb
```diff
@@ -675,8 +675,17 @@ def process_options # :nodoc:
 
     @build_args = options[:build_args]
 
-    @gem_home = @install_dir
-    @gem_home ||= options[:user_install] ? Gem.user_dir : Gem.dir
+    @gem_home = @install_dir || Gem.dir
+
+    # `--build-root` overrides `--user-install` and automatically falling back to installing into user home when default home is not writable
+    if @build_root.nil? && @install_dir.nil?
```
voxik, line 681: Honestly, this code makes me uneasy. How is user supposed to figure this out if I can't read it?This IMHO just underlines that the design is wrong. I can't imagine I'd like to see every time installing gem the message Defaulting to user installation ....

deivid-rodriguez: I don't understand how this implementation makes you uneasy and you were fine with the previous one. This PR essentially fixes a bug that you reported yourself. 

> I can't imagine I'd like to see every time installing gem the message Defaulting to user installation ....

Ideally you'll fix your permissions, add `--user-install` to your .gemrc, set GEM_HOME, or whatever option to make the warning go away. But my question is, does this PR change the current existing behavior in the master branch regarding this? 🤔

deivid-rodriguez: > Can we remove something? Can we drop the --build-root option and use --install-dir instead?

I don't understand your suggestion. Is it related to this PR?

deivid-rodriguez: Regarding what to do with `--build-root` in combination with the fallback to user installation, I discussed it with @duckinator and we decided the easiest way would be to ignore the fallback. Normally `--build-root` is a directory that doesn't yet have gem home or user gem home created inside it, so I think it would not apply anyways to most situations since none of the involved directories yet exist?

But I'm happy to iterate or change the behavior if this causes problems for OS packagers (main users of this flag). Let's keep this PR though focused on fixing this bug and keeping existing behavior otherwsise.

duckinator: If you don't want the message every time you can fix permissions, use `--user-install`, or use `--install-dir`. The auto user install feature is intended to make RubyGems work out of the box basically everywhere while respecting user and OS-provided configuration, and leaned heavily on existing practices of other package managers (primarily pip). 

However, this PR is about fixing a breaking bug. I agree with @deivid-rodriguez -- let's keep this PR focused on that.

duckinator: To clarify: I'm open to the idea of removing `--build-root` if it's redundant, but we'd definitely want to coordinate with OS/distro packagers and make sure there's documentation for about migrating away from it. 

We're trying to make it easier for end users, but we don't want that to come at the cost of everyone else (like you and other OS/distro packagers) having a harder time. So I think getting this PR in place so everything is at least working, document how it works, _then_ discussing more dramatic changes like removing `--build-root` is the way to go.

voxik: > > Can we remove something? Can we drop the --build-root option and use --install-dir instead?
> 
> I don't understand your suggestion. Is it related to this PR?

This was the change needed to workaround the broken `--build-root` option:

https://src.fedoraproject.org/fork/vondruch/rpms/ruby/c/c48e6174fc8c9c79dc525e57b1bc56373260f8db?branch=ruby-3.3

Is is worth of keeping such option, which does the implementation just more complex? Should we just drop it?

deivid-rodriguez: > Come on, this is not about "my permission". This is about standards such as FHS. On Fedora, Ruby is installed into system /usr/ directory. This directory is managed by RPM and it is good idea if only RPM manages content of this system location.

Sorry, this "you" wasn't personal, I know this would not be a permissions issue in the Fedora's case. I just meant to enumerate a few ways to turn off the warning for the different situations where it could be triggered.

> It changes what was done by https://github.com/rubygems/rubygems/pull/7100. It makes buildroot again more exceptional then it deserves. There was no exception for --user-install, where as there is with this PR.

I'm really confused, I asked for behavioral changes. If we add no exception here, we still simplify the code a lot but we don't fix the problem that you reported yourself. Do you prefer your code from #7100 to be left untouched, or the issue fixed? I explained already why I think `--build-root` deserves an exception here and I also think it's fundamentally different from `--user-install` so I'm not sure removing it would make sense. But happy to discuss as a separate issue.

> And the logic is so unreadable, that there must be comment, which does not explain a lot. Maybe the @build_root.nil? should not be here at this place if the installation should be into @build_root + Gem.dir. But that was not the status prior this PR.

I'm not sure what you mean, but I'll note that your feedback in this thread is coming as unnecessarily hostile, and quite unhelpful to me.

deivid-rodriguez:
@duckinator This is ready from my side. I'd just like to give this a final rebase & some commit squashing, but I'll wait for a final review pass from you in case you have some comments!

duckinator:
@deivid-rodriguez rebase and squash away. 👍

PR #7214, opened 2023-12-01
Make `rake` just run RubyGems tests

deivid-rodriguez:

This is just something very minor with a very easy workaround (just get used to running `rake test` instead), so happy to close if there's any disagreement.

My take is that the default `rake` task should be for something you run often. And I've never wanted to run both the RubyGems and Bundler test suites in sequence. Also my muscle memory keeps running this for RubyGems tests since that's how it worked for years until we integrated Rake tasks of Bundler and RubyGems. So the end result is that every time I run `rake`, I cancel Bundler tests once they start.

I know this does bring some nice symmetry, but I don't think it's useful. Again, happy to close if other maintainers disagree.

Only run RubyGems tests as the default task.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -83,7 +83,7 @@ namespace "test" do
   end
 end
 
-task :default => [:test, :spec]
+task :default => [:test]
 
 spec = Gem::Specification.load(File.expand_path("rubygems-update.gemspec", __dir__))
 v = spec.version


martinemde:
I don't feel strongly about this, but I usually expect rake to be the first thing I run on a fresh project to verify that everything is up and running.

Is bundler a separate project that happens to be in the same repository and it has a separate setup and doesn't need to work right away? Is bundler a critical piece of this repo that must work?

I imagine we're someday koving twards bundler shipping with rubygems, locked to the same version and not upgrade able separately. In this future the tests must all work. If this is not the future then separate is fine.

deivid-rodriguez:
Yeah, I was not really convinced about this. Let me just close this, I think it's reasonable to consider everything together.

PR #7217, opened 2023-12-01
Convert Bundler lockfile checksum validation to manual opt-in

martinemde:

I've not yet been able to finish the full-featured opt-in for checksums.

This makes bundler checksums a hidden opt-in feature. There are no commands which add the CHECKSUMS section to the lockfile and the feature will only be enabled if that section is found.

The actual behavior change is as follows:
* When checksums come from either the API or from a gem, we add them to the checksum store unless the settings disabled checksum validation.
* When checksums come from the lockfile, we always store them in the checksum store regardless of the disabled checksums setting.
* The presence of a checksum section, even empty, in the lockfile indicates that we should write a checksum section to the lockfile when we generate.
* When the disable checksums setting is set, we still read and write checksums to the lockfile, but since all other checksums won't be recorded, we will never compare them or raise an error.
* When the checksums section is not in the lockfile, the api and gem checksums will continue to be stored and compared, perpetuating the existing behavior of checksum verification with the server, but not writing any of them to the lockfile.
* Adding an empty CHECKSUMS section to the lockfile will enable the feature manually. We will follow up with more user-friendly commands to add and populate this new section.

### DIFF
--- a/bundler/lib/bundler/checksum.rb
+++ b/bundler/lib/bundler/checksum.rb
@@ -9,6 +9,18 @@ module Bundler
     private_constant :DEFAULT_BLOCK_SIZE
 
     class << self
+      def from_gem_package(gem_package, algo = DEFAULT_ALGORITHM)
+        return if Bundler.settings[:disable_checksum_validation]
+        return unless source = gem_package.instance_variable_get(:@gem)
+        return unless source.respond_to?(:with_read_io)
+
+        source.with_read_io do |io|
+          from_gem(io, source.path)
+        ensure
+          io.rewind
+        end
+      end
+
       def from_gem(io, pathname, algo = DEFAULT_ALGORITHM)
         digest = Bundler::SharedHelpers.digest(algo.upcase).new
         buf = String.new(:capacity => DEFAULT_BLOCK_SIZE)
@@ -17,6 +29,7 @@ module Bundler
       end
 
       def from_api(digest, source_uri, algo = DEFAULT_ALGORITHM)
+        return if Bundler.settings[:disable_checksum_validation]
         Checksum.new(algo, to_hexdigest(digest, algo), Source.new(:api, source_uri))
       end
 
@@ -177,7 +190,6 @@ module Bundler
       # This ensures a mismatch error where there are multiple top level sources
       # that contain the same gem with different checksums.
       def replace(spec, checksum)
-        return if Bundler.settings[:disable_checksum_validation]
         return unless checksum
 
         name_tuple = spec.name_tuple
@@ -193,7 +205,6 @@ module Bundler
       end
 
       def register(spec, checksum)
-        return if Bundler.settings[:disable_checksum_validation]
         return unless checksum
         register_checksum(spec.name_tuple, checksum)
       end
--- a/bundler/lib/bundler/lockfile_parser.rb
+++ b/bundler/lib/bundler/lockfile_parser.rb
@@ -24,7 +24,15 @@ module Bundler
       end
     end
 
-    attr_reader :sources, :dependencies, :specs, :platforms, :bundler_version, :ruby_version, :checksums
+    attr_reader(
+      :sources,
+      :dependencies,
+      :specs,
+      :platforms,
+      :bundler_version,
+      :ruby_version,
+      :checksums,
+    )
 
     BUNDLED      = "BUNDLED WITH"
     DEPENDENCIES = "DEPENDENCIES"
@@ -111,6 +119,9 @@ module Bundler
         elsif line == DEPENDENCIES
           @parse_method = :parse_dependency
         elsif line == CHECKSUMS
+          # This is a temporary solution to make this feature disabled by default
+          # for all gemfiles that don't already explicitly include the feature.
+          @checksums = true
           @parse_method = :parse_checksum
         elsif line == PLATFORMS
           @parse_method = :parse_platform
@@ -228,8 +239,6 @@ module Bundler
       version = Gem::Version.new(version)
       platform = platform ? Gem::Platform.new(platform) : Gem::Platform::RUBY
       full_name = Gem::NameTuple.new(name, version, platform).full_name
-      # Don't raise exception if there's a checksum for a gem that's not in the lockfile,
-      # we prefer to heal invalid lockfiles
       return unless spec = @specs[full_name]
 
       checksums.split(",") do |lock_checksum|
--- a/bundler/spec/bundler/definition_spec.rb
+++ b/bundler/spec/bundler/definition_spec.rb
@@ -56,6 +56,11 @@ RSpec.describe Bundler::Definition do
         s.add_dependency "rack", "1.0"
       end
 
+      checksums = checksums_section_when_existing do |c|
+        c.no_checksum "foo", "1.0"
+        c.checksum gem_repo1, "rack", "1.0.0"
+      end
+
       bundle :install, :env => { "DEBUG" => "1" }
 
       expect(out).to match(/re-resolving dependencies/)
@@ -76,11 +81,7 @@ RSpec.describe Bundler::Definition do
 
         DEPENDENCIES
           foo!
-
-        CHECKSUMS
-          #{gem_no_checksum "foo", "1.0"}
-          #{checksum_for_repo_gem gem_repo1, "rack", "1.0.0"}
-
+        #{checksums}
         BUNDLED WITH
            #{Bundler::VERSION}
       G
@@ -110,6 +111,11 @@ RSpec.describe Bundler::Definition do
         s.add_development_dependency "net-ssh", "1.0"
       end
 
+      checksums = checksums_section_when_existing do |c|
+        c.no_checksum "foo", "1.0"
+        c.checksum gem_repo1, "rack", "1.0.0"
+      end
+
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
         gem "foo", :path => "#{lib_path("foo")}"
@@ -135,17 +141,17 @@ RSpec.describe Bundler::Definition do
 
         DEPENDENCIES
           foo!
-
-        CHECKSUMS
-          #{gem_no_checksum "foo", "1.0"}
-          #{checksum_for_repo_gem gem_repo1, "rack", "1.0.0"}
-
+        #{checksums}
         BUNDLED WITH
            #{Bundler::VERSION}
       G
     end
 
     it "for a locked gem for another platform" do
+      checksums = checksums_section_when_existing do |c|
+        c.no_checksum "only_java", "1.1", "java"
+      end
+
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
         gem "only_java", platform: :jruby
@@ -166,16 +172,17 @@ RSpec.describe Bundler::Definition do
 
         DEPENDENCIES
           only_java
-
-        CHECKSUMS
-          only_java (1.1-java)
-
+        #{checksums}
         BUNDLED WITH
            #{Bundler::VERSION}
       G
     end
 
     it "for a rubygems gem" do
+      checksums = checksums_section_when_existing do |c|
+        c.checksum gem_repo1, "foo", "1.0"
+      end
+
       install_gemfile <<-G
         source "#{file_uri_for(gem_repo1)}"
         gem "foo"
@@ -195,10 +202,7 @@ RSpec.describe Bundler::Definition do
 
         DEPENDENCIES
           foo
-
-        CHECKSUMS
-          #{checksum_for_repo_gem gem_repo1, "foo", "1.0"}
-
+        #{checksums}
         BUNDLED WITH
            #{Bundler::VERSION}
       G
--- a/bundler/spec/cache/gems_spec.rb
+++ b/bundler/spec/cache/gems_spec.rb
@@ -289,10 +289,23 @@ RSpec.describe "bundle cache" do
       expect(cached_gem("rack-1.0.0")).to exist
     end
 
-    it "raises an error when the gem file is altered and produces a different checksum" do
+    it "raises an error when the gem is altered and produces a different checksum" do
       cached_gem("rack-1.0.0").rmtree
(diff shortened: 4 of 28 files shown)


bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -226,10 +229,7 @@ def parse_checksum(line)
       platform = $4
 
       version = Gem::Version.new(version)
-      platform = platform ? Gem::Platform.new(platform) : Gem::Platform::RUBY
```
segiddins, line 229: why is this line removed?

bundler/lib/bundler/lockfile_parser.rb
```diff
@@ -111,6 +111,9 @@ def initialize(lockfile)
         elsif line == DEPENDENCIES
           @parse_method = :parse_dependency
         elsif line == CHECKSUMS
+          # This is a temporary solution to make this feature disabled by default
+          # for all gemfiles that don't already explicitly include the feature.
+          Bundler::Checksum.enable!
```
segiddins, line 116: this feels super brittle to me, given bundler could have read a different lockfile and then have stored the ivar.

I would feel a lot better if this got hoisted into the definition

bundler/lib/bundler/checksum.rb
```diff
@@ -9,6 +9,23 @@ class Checksum
...
     class << self
+      # This is used by the lockfile parser to enable checksum validation
+      # when checksums are found in the lockfile.
+      # We don't enable if the user has explicitly disabled checksum validation.
+      # TODO: This can be removed when checksums are enabled by default.
+      # The enable_checksum_validation setting may also be removed once this
+      # is opt-out instead of opt-in
+      def enable!
+        return if Bundler.settings[:disable_checksum_validation]
+        @enabled = true
```
segiddins, line 20: will this ivar ever get cleared?

bundler/lib/bundler/lockfile_parser.rb
martinemde: oh, it's duplicating the behavior of Gem::NameTuple.new so it's unnecessary. Maybe it can happen in another PR, but it's something I noticed when going through the code.

bundler/lib/bundler/lockfile_parser.rb
martinemde: ok, I'll shift it there. Would you still use the same ivar approach? I toyed with adding a Bundler.setting for "enabled", but it makes it more complex. Do you prefer that?

bundler/lib/bundler/lockfile_parser.rb
martinemde: I should say, not duplicating exactly, it's making a Gem::Platform, then to_s on the platform. The only reason this would matter is if there's an inconsistent platform name saved into the checksums section of the lockfile that gets resolved by passing through `Gem::Platform.new#to_s`. I'll revert it just based on this slim case.

bundler/lib/bundler/checksum.rb
```diff
@@ -9,6 +9,29 @@ class Checksum
...
+      # TODO: This can be removed when checksums are enabled by default.
+      # The enable_checksum_validation setting may also be removed once this
+      # is opt-out instead of opt-in
+      def enabled?
+        !Bundler.settings[:disable_checksum_validation] &&
+          Bundler.settings[:enable_checksum_validation]
+      end
+
+      def from_gem_package(gem_package, algo = DEFAULT_ALGORITHM)
+        return nil unless enabled?
```
segiddins, line 24: this is a behavior change - we should still be validating the checksums from the server on install by default (unless `:disable_checksum_validation` is set)

martinemde: well, I suppose it wasn't lost until we began disabling this feature. Now we need to add a special case for it.

bundler/spec/cache/gems_spec.rb
```diff
@@ -289,22 +289,49 @@
       expect(cached_gem("rack-1.0.0")).to exist
     end
 
-    it "raises an error when the gem file is altered and produces a different checksum" do
+    it "raises an error when the gem is altered and produces a different checksum" do
+      pending "This test depends on lockfile checksums being enabled by default"
```
martinemde, line 293: I don't understand why this one won't trigger the error. I enable checksums in the lockfile.

bundler/spec/install/gemfile/platform_spec.rb
```diff
@@ -224,21 +233,16 @@
...
+      #{checksums}
       BUNDLED WITH
          #{Bundler::VERSION}
     L
 
     bundle "lock --add-platform ruby"
 
+    good_checksums = checksums.dup
+    good_checksums.no_checksum("pry", "0.11.3")
+
```
martinemde, line 245:
```suggestion
```

This isn't needed any more.

bundler/spec/install/gemfile/sources_spec.rb
```diff
@@ -379,14 +396,11 @@
                   rack
 
             PLATFORMS
-              #{lockfile_platforms}
+              #{lockfile_platforms("ruby")}
```
martinemde, line 399: Locally I had to add ruby to get this passing, now it's failing on windows.

bundler/spec/install/gemfile/sources_spec.rb
```diff
@@ -422,14 +436,11 @@
...
```
martinemde, line 439: Same problem here.

bundler/spec/install/gemfile/sources_spec.rb
deivid-rodriguez: In most situations `lockfile_platforms` should include `ruby` by default since #5700. What's the error if you revert the change? I can't reproduce it locally.

bundler/spec/cache/gems_spec.rb
deivid-rodriguez: I think it may be because we currently don't produce the error either if the checksum section is there, but there was no locked checksum for the gem that was altered (like in the lockfile you added to the spec).

If I modify the spec like this, it passes:

```diff
diff --git a/bundler/spec/cache/gems_spec.rb b/bundler/spec/cache/gems_spec.rb
index 2f5da4e7e..bb07bfc08 100644
--- a/bundler/spec/cache/gems_spec.rb
+++ b/bundler/spec/cache/gems_spec.rb
@@ -289,15 +289,32 @@
       expect(cached_gem("rack-1.0.0")).to exist
     end
 
-    it "raises an error when the gem file is altered and produces a different checksum" do
+    it "raises an error when the gem is altered and produces a different checksum" do
       cached_gem("rack-1.0.0").rmtree
       build_gem "rack", "1.0.0", :path => bundled_app("vendor/cache")
+
+      expected_checksums = checksums_section do |c|
+        c.checksum gem_repo1, "actionpack", "2.3.2"
+        c.checksum gem_repo1, "activesupport", "2.3.2"
+        c.checksum gem_repo1, "rack", "1.0.0"
+      end
+
       simulate_new_machine
 
+      lockfile <<~L
+        GEM
+          remote: #{file_uri_for(gem_repo2)}/
+          specs:
+            rack (1.0.0)
+
+        CHECKSUMS
+          #{expected_checksums}
+      L
+
       bundle :install, :raise_on_error => false
-      expect(exitstatus).to eq(37)
       expect(err).to include("Bundler found mismatched checksums.")
       expect(err).to include("1. remove the gem at #{cached_gem("rack-1.0.0")}")
+      expect(exitstatus).to eq(37)
 
       expect(cached_gem("rack-1.0.0")).to exist
       cached_gem("rack-1.0.0").rmtree
```

Conversely, if I add the lockfile that you added here to the spec in current master branch, it no longer raises the error.

martinemde:
@deivid-rodriguez Updated with your suggestions. I don't know what happened with the ruby platform there, but if I remove it, it works now. This should be updated and ready to go, if you approve.

bundler/spec/cache/gems_spec.rb
martinemde: Thanks for figuring this out. It makes sense that it needs the checksum stored in the lockfile to trigger this. It probably was inserting the checksum automatically when they were enabled by default so I didn't realize it was necessary for the feature to work.

martinemde:
I updated the description of this PR to include the new behavior. I'm happy with this approach because it means the feature is present and ready to use, even can be enabled to work exactly as merged initially, but it is disabled. We can then add new commands that add the section and populate it and those can go out in point releases because they don't change any behavior without explicit interaction from the user. Any 2.5 series bundler will be able to handle lockfile checksum and the future commands will just be a bonus.

bundler/spec/bundler/definition_spec.rb
```diff
@@ -56,6 +56,11 @@
         s.add_dependency "rack", "1.0"
       end
 
+      checksums = checksums_section_when_existing do |c|
+        c.no_checksum "foo", "1.0"
+        c.checksum gem_repo1, "rack", "1.0.0"
+      end
+
```
deivid-rodriguez, line 63: This should go _after_ the `bundle install` command below. Otherwise it will generate an empty section once we generate checksums by default and will make the spec fail.

bundler/spec/lock/lockfile_spec.rb
```diff
@@ -78,9 +79,6 @@
       DEPENDENCIES
         rack
 
-      CHECKSUMS
-        #{checksum_for_repo_gem(gem_repo2, "rack", "1.0.0")}
-
```
deivid-rodriguez, line 83: Should this include the checksums section so that specs keep passing once generated by default?

bundler/spec/lock/lockfile_spec.rb
```diff
@@ -214,9 +210,6 @@
       DEPENDENCIES
         rack (> 0)
 
-      CHECKSUMS
-        #{checksum_for_repo_gem(gem_repo2, "rack", "1.0.0")}
-
```
deivid-rodriguez, line 219: Should this include the checksums section so that specs keep passing once generated by default?

deivid-rodriguez (approved):
Looks great, this is fantastic work and leaves master in a much better place towards confidently releasing.

I tried one experiment, which is applying this diff:

```diff
diff --git a/bundler/lib/bundler/definition.rb b/bundler/lib/bundler/definition.rb
index 3493f0732..6018d1b82 100644
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -93,7 +93,7 @@ def initialize(lockfile, dependencies, sources, unlock, ruby_version = nil, opti
         @locked_bundler_version = @locked_gems.bundler_version
         @locked_ruby_version = @locked_gems.ruby_version
         @originally_locked_specs = SpecSet.new(@locked_gems.specs)
-        @locked_checksums = @locked_gems.checksums
+        @locked_checksums = true
 
         if unlock != true
           @locked_deps    = @locked_gems.dependencies
```

I _think_ the way you designed this PR is so that this kind of change would trigger just a few selected spec failures, but I'm seeing quite a few specs fail.

I started commenting on some of them, but then I thought maybe you prefer to merge and then work on those? If we fix most of them, would it make sense to potentially run a CI job under an ENV setting that enables the setting and make sure we don't regress any thing related to checksum generation. Maybe overkill?

martinemde:
That definitely makes sense as a part of CI, imho. It was my intention to leave the tests in a state where we could enable or disable the feature with a switch. I had planned to revisit this after I got the test passing with opt-in. The thing I most wamted to avoid was ruining all the work to add checksums to all the lockfiles in tests 🙀

I will revisit after merge so that we're less blocked than with the feature enabled by default. I suspect failures are not actually the feature being broken, but tests that assert incorrectly or setup incorrectly.

bundler/spec/support/checksums.rb
```diff
@@ -3,53 +3,82 @@
...
-    def checksum_section(&block)
-      ChecksumsBuilder.new(&block).to_lock
+    def checksums_section(enabled = true, &block)
+      ChecksumsBuilder.new(enabled, &block)
     end
 
-    def checksum_for_repo_gem(*args)
-      checksum_section do |c|
-        c.repo_gem(*args)
+    def checksums_section_when_existing(&block)
```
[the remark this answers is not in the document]
martinemde: That's the intention, though as we discussed I think in practice there's some differences. I'll revisit in the next PR.

bundler/spec/commands/lock_spec.rb
```diff
@@ -65,42 +62,61 @@
...
-    expect(read_lockfile).to eq(remove_checksums_from_lockfile(@lockfile))
+    expect(read_lockfile).to eq(@lockfile)
+  end
+
+  it "prints a lockfile without fetching new checksums if the existing lockfile had no checksums" do
+    lockfile remove_checksums_from_lockfile(@lockfile)
+
+    bundle "lock --print"
+
+    expect(out).to eq(remove_checksums_from_lockfile(@lockfile).chomp)
```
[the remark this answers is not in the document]
martinemde: Oh probably, they used to express different behaviors like auto-adding checksums, but now they've flattened to be the same.

bundler/spec/install/gemfile/platform_spec.rb
```diff
@@ -264,16 +265,7 @@
       DEPENDENCIES
         empyrean (= 0.1.0)
         pry
-
-      CHECKSUMS
-        #{checksum_for_repo_gem gem_repo4, "coderay", "1.1.2"}
-        #{checksum_for_repo_gem gem_repo4, "empyrean", "0.1.0"}
-        #{checksum_for_repo_gem gem_repo4, "ffi", "1.9.23", "java"}
-        #{checksum_for_repo_gem gem_repo4, "method_source", "0.9.0"}
-        pry (0.11.3)
```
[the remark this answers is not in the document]
martinemde: Oops, this is a missed conversion.

deivid-rodriguez:
Yep, the failures I looked at all seemed like a problem with the specs, not with the feature. Good to merge away and iterate 👍.

bundler/spec/commands/lock_spec.rb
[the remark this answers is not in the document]
martinemde: I fixed this one and re-pushed.

PR #7220, opened 2023-12-02
bundle lock: Always touch the lockfile

### LINKED ISSUE #7120: give bundler lock an option to force touching Gemfile.lock
Hi, 

that's a little, but strong wish: 

bundler lock 
creates the Gemfile.lock from Gemfile. But it doesn't touch Gemfile.lock if there's nothing to do, thus not updating the file mod time of the Gemfile.lock. 

That comes with a problem. If there is a Makefile or Rakefile, where Gemfile.lock depends from Gemfile and calls bundler lock to recreate Gemfile.lock once Gemfile has been changed (i.e. newer than Gemfile.lock), it calls bundler lock with every run again and again and again, since Gemfile.lock never becomes newer than Gemfile if there's nothing new to do. 

Not even a bundler lock --update  would update the file if there's nothing new (e.g. because of just minor changes in Gemfile such as whitespace and comments)

Would be nice if bundler lock would always (or at least with a given option) touch the Gemfile.lock to have the file date reflect the time of the last run of bundler lock, and be make/rake-friendly

regards

franzliedke:
Fixes #7120.

In #7120, @hadmut mentioned problems with Make-based build pipelines: `Gemfile.lock` would not be "touched" (receive an updated file modification date) when running `bundle lock` is a no-op (i.e. the lockfile already matches the `Gemfile`).

An explicit `FileUtils.touch` call. 🤓 

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -342,7 +342,10 @@ module Bundler
 
       preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
 
-      return if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+      if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+        FileUtils.touch(file)
+        return
+      end
 
       if Bundler.frozen_bundle?
         Bundler.ui.error "Cannot write a changed lockfile while frozen."
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -343,7 +343,7 @@ module Bundler
       preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
 
       if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
-        FileUtils.touch(file)
+        SharedHelpers.filesystem_access(file) {|p| FileUtils.touch(p) }
         return
       end
 
--- a/bundler/lib/bundler/cli/lock.rb
+++ b/bundler/lib/bundler/cli/lock.rb
@@ -64,6 +64,7 @@ module Bundler
           file = file ? File.expand_path(file) : Bundler.default_lockfile
           puts "Writing lockfile to #{file}"
           definition.lock(file)
+          FileUtils.touch(file)
         end
       end
 
--- a/bundler/lib/bundler/cli/lock.rb
+++ b/bundler/lib/bundler/cli/lock.rb
@@ -64,7 +64,6 @@ module Bundler
           file = file ? File.expand_path(file) : Bundler.default_lockfile
           puts "Writing lockfile to #{file}"
           definition.lock(file)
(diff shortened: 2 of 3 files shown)


bundler/lib/bundler/definition.rb
```diff
@@ -339,7 +339,10 @@ def lock(file, preserve_unknown_sections = false)
 
       preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
 
-      return if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+      if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+        FileUtils.touch(file)
```
franzliedke, line 343: Should I wrap this in a `SharedHelpers.filesystem_access` block as well? 🤔

martinemde: Yes, that helper unifies error rescuing for filesystem access problems. Touch is a write to the filesystem.

martinemde: Tell me what you think. Should we copy the frozen bundle check to guard the touch? I know for `lock` it temporarily unfreezes because it's nonsensical otherwise, but this code path will run during install and update, I believe. We should not otherwise change the lockfile if frozen, right?

franzliedke: You're referring to the little "Cannot write a changed lockfile while frozen" guard directly below, right? Yes, I also caught that and it seemed sensible not to touch the lockfile when frozen, but I did not add an explicit test about that.

franzliedke: Alright, I applied the wrapper. Anything else?

bundler/lib/bundler/definition.rb
```diff
@@ -339,7 +339,10 @@ def lock(file, preserve_unknown_sections = false)
 
       preserve_unknown_sections ||= !updating_major && (Bundler.frozen_bundle? || !(unlocking? || @unlocking_bundler))
 
-      return if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+      if file && File.exist?(file) && lockfiles_equal?(@lockfile_contents, contents, preserve_unknown_sections)
+        SharedHelpers.filesystem_access(file) {|p| FileUtils.touch(p) }
+        return
```
martinemde, line 344: I would specifically return without touching the lockfile if the bundle is frozen, like so:

```suggestion
        return if Bundler.frozen_bundle?
        SharedHelpers.filesystem_access(file) {|p| FileUtils.touch(p) }
        return
```

martinemde: I'd likewise add a test that it is not touched when frozen. Maybe [here](https://github.com/rubygems/rubygems/blob/master/bundler/spec/install/deploy_spec.rb#L262), since the lock command always temporarily unfreezes.

martinemde:
Hey @franzliedke, I decided to go ahead and merge this now since I want to see it make it into the 2.5.0 bundler release. I'll update with the test I was imagining. Thanks for your contribution!

franzliedke:
@martinemde Thanks! To be honest, I was hoping this would happen. 😅 I wasn't 100% sure about the implications in these other commands that I wasn't aware of.

PR #7226, opened 2023-12-07
Remove non-transparent requirement added to prerelease gems

deivid-rodriguez:

I recall this requirement confusing me while working in the resolver, since it could lead to different resolution process when working in master (2.5.0.dev currently) than when working with a released version.

It feels non transparent and too magical. I think we can safely assume these days that all RubyGems and Bundler versions that will ever bundle a new gem created in 2023 support prereleases.

So this non transparent requirement is not necessary. It should be the gem author to explicitly add this kind of constraint, not RubyGems. Setting the version of a gem should do just that, set the version.

This has also caused Circular require issues in the past, and has special conditions attached to workaround that.

Remove the extra constraint.

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2674,19 +2674,12 @@ class Gem::Specification < Gem::BasicSpecification
   rubygems_deprecate :validate_permissions
 
   ##
-  # Set the version to +version+, potentially also setting
-  # required_rubygems_version if +version+ indicates it is a
-  # prerelease.
+  # Set the version to +version+.
 
   def version=(version)
     @version = Gem::Version.create(version)
     return if @version.nil?
 
-    # skip to set required_ruby_version when pre-released rubygems.
-    # It caused to raise CircularDependencyError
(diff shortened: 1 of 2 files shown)


hsbt (approved):

lib/rubygems/specification.rb
```diff
@@ -2682,11 +2682,6 @@ def version=(version)
     @version = Gem::Version.create(version)
     return if @version.nil?
 
-    # skip to set required_ruby_version when pre-released rubygems.
-    # It caused to raise CircularDependencyError
```
martinemde, line 2686: Need to update the comment on `def version=` above as well.

martinemde (approved):
It seems extremely unlikely that this would still be necessary. That is an old RubyGems. Just update the comment before merging.

martinemde:
Ready to merge after the comment is fixed. (just a note to self, since I almost came back and merged this)

lib/rubygems/specification.rb
deivid-rodriguez: Thanks for catching that!

PR #7229, opened 2023-12-07
Figure out why we need to override `Gem::NameTuple.new`

deivid-rodriguez:

Just trying to figure out why this is needed because it's very non standard.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -352,14 +352,6 @@ module Gem
   require "rubygems/name_tuple"
 
   class NameTuple
-    def self.new(name, version, platform="ruby")
-      if Gem::Platform === platform
-        super(name, version, platform.to_s)
-      else
-        super
-      end
-    end
-
     def lock_name
       @lock_name ||=
         if platform == Gem::Platform::RUBY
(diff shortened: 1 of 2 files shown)


martinemde:
For old rubygems that don't to_s their platform arg (which we can't easily test here).

martinemde:
Ok, I have a few more minutes to look at this.

Original PR is #7081.

Useful to know: [Gem::Platform also overrides `self.new` without problems](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb#L64-L73).

If we didn't override new, we'd need to override initialize (which I had a lot of problems with when I first time, but we could try again) or override a few of the other instance methods.

We could also be better about removing the patch when it's not necessary. I'll test that approach by pushing to this branch.

deivid-rodriguez:
> We could also be better about removing the patch when it's not necessary. I'll test that approach by pushing to this branch.

Yes, that's the approach we normally follow, feature detect if RubyGems is missing the feature/fix that needs patching (or hardcode the RubyGems version that first introduces it) and only apply it when necessary. I feel that also makes it easier to remember cleaning up when dropping support for old versions.

Also I think removing these patches should make some test fail? We test Bundler with old rubygems so some matrix entry should catch the issue?

deivid-rodriguez:
Even if there's no end to end test that catches it, this is needed, so closing!

PR #7233, opened 2023-12-07
Test with Ruby 3.3

deivid-rodriguez:

As we approach final Ruby releases, make sure everything works with Ruby 3.3.

Start testing with Ruby 3.3 early. Will update this PR once the rc is released.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -32,6 +32,7 @@ jobs:
           - { name: ruby-3.0, value: 3.0.6 }
           - { name: ruby-3.1, value: 3.1.4 }
           - { name: ruby-3.2, value: 3.2.2 }
+          - { name: ruby-3.3, value: 3.3.0-rc1 }
 
         bundler:
           - { name: 2, value: '' }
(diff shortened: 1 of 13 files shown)


MSP-Greg:
@deivid-rodriguez

JFYI, Actions Windows Ruby builds haven't been available for Ruby master preview releases.

For Windows, see:
https://github.com/ruby/setup-ruby/blob/master/windows-versions.json

For Ubuntu & macOS, see
https://github.com/ruby/setup-ruby/blob/master/ruby-builder-versions.json

deivid-rodriguez:
Cool, thanks! I just added it everywhere without thinking too much. Will undo those changes tomorrow.

deivid-rodriguez:
I did it now in the end :)

deivid-rodriguez:
Green with preview3, trying rc1 now!

deivid-rodriguez:
https://github.com/rubygems/rubygems/pull/7233/commits/95711846f093f6a0d5eada4234aa8e19741c7b3f is not particularly pretty. Other alternatives:

* Always auto-install our locked version when running `bin/rake`. I find this too surprising.
* Upgrade our locked rake to 13.1.0 (what Ruby 3.3 will provide) and move on.
* Keep using 3.2.2 for the unused cassettes job for now.

martinemde (approved):
Looks like this is passing and ready to merge. The fix for version of rake seems fine and we revisit if it ever causes any issues.

martinemde:
I'm merging right away now that this is passing so all other PRs will need to stay passing on 3.3. We could update to use rake 3.1.0. I think that's also a good solution. There's unlikely to be a problem using the newest rake.

PR #7235, opened 2023-12-08
Don't memoize in NameTuple lock_name

martinemde:

Seems unnecessary. `LazySpecification` already memoizes `lock_name`. It seems like it will be created 2 times at most, which doesn't seem worth the memory usage. No other `NameTuple` instance method memoizes.

Remove memoization from `NameTuple#lock_name`.


hsbt (approved):
Good catch 👍

PR #7236, opened 2023-12-08
Make the "auto-user-install" logic easier to follow

voxik:

I had some assertion about ease of following the logic in the "auto-user-install"

This is and attempt to make the logic easier to follow.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -675,16 +675,20 @@ class Gem::Installer
 
     @build_args = options[:build_args]
 
-    @gem_home = @install_dir || Gem.dir
+    @gem_home = @install_dir
 
-    # `--build-root` overrides `--user-install` and auto-user-install
-    if @build_root.nil? && @install_dir.nil?
-      if options[:user_install]
-        @gem_home = Gem.user_dir
-      elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
-        say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
-        @gem_home = Gem.user_dir
+    unless @gem_home
+      # `--build-root` overrides `--user-install` and auto-user-install
+      if @build_root.nil?
+        if options[:user_install]
+          @gem_home = Gem.user_dir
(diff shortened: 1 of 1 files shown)


voxik:
I don't think the Bundler test failure is related to my changes

lib/rubygems/installer.rb
```diff
@@ -675,16 +675,20 @@ def process_options # :nodoc:
...
+      if @build_root.nil?
+        if options[:user_install]
+          @gem_home = Gem.user_dir
+        elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
+          say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
+          @gem_home = Gem.user_dir
+        end
       end
+
+      @gem_home ||= Gem.dir
```
skipkayhil, line 691:
```suggestion
    @gem_home = @install_dir || if @build_root.nil?
      # `--build-root` overrides `--user-install` and auto-user-install
      if options[:user_install]
        Gem.user_dir
      elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
        say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
        Gem.user_dir
      end
    end || Gem.dir
```

Would this make it even more clear? (and maybe the outer if statement should be extracted to a method...)

voxik: Good attempt 👍 But to me, it might not be clear what is the return value of the `if` block. Of course it would be more natural, if the block was extracted into the method, but the logic is a bit tricky.

BTW I have another PR in the works, which actually extracts the `!ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))` part into method and make this more readable and even more configurable (I hope)

martinemde (approved):
Nice and simple. Thanks!

lib/rubygems/installer.rb
martinemde: I think there's room for improvement still, but this change moves in the right direction. We can follow up with more suggestions in another PR. I appreciate the simple refactors like this.

PR #7239, opened 2023-12-08
Override initialize in bundle rubygems_ext for NameTuple

martinemde:

In PR #7229 we were discussing alternative ways to add this back-ported feature.

This avoids overriding `.new` by aliasing initialize to see if it makes any difference on segfaults. Guards tho feature better against rubygems that already have the feature.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -334,11 +334,16 @@ module Gem
   require "rubygems/name_tuple"
 
   class NameTuple
-    def self.new(name, version, platform="ruby")
-      if Gem::Platform === platform
-        super(name, version, platform.to_s)
-      else
-        super
+    # Versions of RubyGems before about 3.5.0 don't to_s the platform.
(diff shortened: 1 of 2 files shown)


martinemde:
It's hard to tell if this code is fine or it's just always circumvented. I did run the suite without the guard before pushing and it passed, but I wonder if we should even bother with the guard because the .to_s is such a cheap operation.

deivid-rodriguez (approved):

martinemde:
Sorry, I meant that because there's a guard in place, that unless there's a test suite that runs new bundler against an old version of rubygems, it won't ever patch initialize.

I'm sure we have tests that use a Gem::Platform arg in a NameTuple.

deivid-rodriguez:
> Edit: uh oh, if you go look at that PR we aren't getting any failures other than the rubygems test for the behavior. I was so confident because I added this literally because of the places in bundler tests where this caused failures. Hmm. I'm happy to merge this as a better approach anyway.

Yeah! That's why when I opened #7229 completely removing this override, I was surprised that everything passed. Probably worth looking into why.

martinemde:
This did not prevent segfaults. I linked a recent one in the ruby bug tracker should have included this commit.

PR #7240, opened 2023-12-09
Stream output from ext builds when --verbose

osyoyu:
Uses Open3.popen2e in place of Open3.capture2e in Gem::Ext::Builder. This change aims to stream stdout/stderr of ext builds when in verbose mode, instead of printing everything at once when the build completes.

Nice for debugging gem builds that consumes longer times.

Encountered a significant slowdown following the update of a gem with native extensions (`grpc`, to be specific).
To gain insights about the bottleneck in the build process (which involves multiple library builds in `grpc`'s case), I tried running `gem install --verbose grpc` and examining the build log flowing in real-time. However, this approach wasn't much helpful, as the entire output gets printed at once after the build completes.

Connect the build's stdout/stderr to `gem install`'s stdout, and let the log flow in real-time.
The original code uses Open3.capture2e which buffers output in a String, so I have switched to Open3.popen2e, which provides a IO object connected to the build's stdout/stderr.

This should affect behavior only when `--verbose` is specified.

### DIFF
--- a/lib/rubygems/ext/builder.rb
+++ b/lib/rubygems/ext/builder.rb
@@ -88,13 +88,20 @@ class Gem::Ext::Builder
       # Set $SOURCE_DATE_EPOCH for the subprocess.
       build_env = { "SOURCE_DATE_EPOCH" => Gem.source_date_epoch_string }.merge(env)
       output, status = begin
-                         Open3.capture2e(build_env, *command, chdir: dir)
+                         Open3.popen2e(build_env, *command, chdir: dir) do |_stdin, stdouterr, wait_thread|
+                           output = String.new
+                           while line = stdouterr.gets
+                             output << line
+                             if verbose
+                               print line
+                             end
+                           end
+                           [output, wait_thread.value]
+                         end
                        rescue StandardError => error
                          raise Gem::InstallError, "#{command_name || class_name} failed#{error.message}"
                        end
-      if verbose
-        puts output
-      else
+      unless verbose
         results << output
       end
     ensure


lib/rubygems/ext/builder.rb
```diff
@@ -88,13 +88,20 @@ def self.run(command, results, command_name = nil, dir = Dir.pwd, env = {})
...
-                         Open3.capture2e(build_env, *command, chdir: dir)
+                         Open3.popen2e(build_env, *command, chdir: dir) do |_stdin, stdouterr, wait_thread|
+                           output = String.new
+                           while line = stdouterr.gets
+                             output << line
+                             if verbose
+                               print line
+                             end
+                           end
+                           [output, wait_thread.value]
```
martinemde, line 99: Since `results` is in scope here would this make sense?

```suggestion
                           while line = stdouterr.gets
                             results << line
                             if verbose
                               print line
                             end
                           end
                           wait_thread.value
```

martinemde (approved):
Seems like a really useful improvement. Thank you!

lib/rubygems/ext/builder.rb
osyoyu: Since `results` is an Array, this will make a change in structure (`["output\noutput\noutput\n"]` vs `["output\n", "output\n", "output\n"]`).

While there might not be much practical impact, I'm not sure removing a String allocation is worth changing `#build`'s result. What do you think?

martinemde: Ah, ok. I didn't look high enough up the method. That makes sense then.

deivid-rodriguez (approved):
Thank you!

PR #7241, opened 2023-12-09
Support dynamic library loading with extension .so or .o

hogelog:

Ruby's native `require` can load dynamic libraries with .so or .o extensions that are not actually filename.

For example, on macOS the socket library name is "socket.bundle" and `require "socket.so"` will load successful.

```console
$ ruby -e 'p RbConfig::CONFIG["DLEXT"]; p require "socket.bundle"'
"bundle"
true
$ ruby -e 'p RbConfig::CONFIG["DLEXT"]; p require "socket.so"'
"bundle"
true
```

However, with the `require` provided by RubyGems, such loading does not work.

```console
$ gem install nokogiri
Successfully installed nokogiri-1.15.5
Parsing documentation for nokogiri-1.15.5
Done installing documentation for nokogiri after 0 seconds
1 gem installed
$ ruby -e 'p RbConfig::CONFIG["DLEXT"]; p require "nokogiri/nokogiri.bundle"'
"bundle"
true
$ ruby -e 'p RbConfig::CONFIG["DLEXT"]; p require "nokogiri/nokogiri.so"'
"bundle"
<internal:/Users/hogelog/.rbenv/versions/3.3.0-dev/lib/ruby/3.3.0+0/rubygems/core_ext/kernel_require.rb>:127:in `require': cannot load such file -- nokogiri/nokogiri.so (LoadError)
        from <internal:/Users/hogelog/.rbenv/versions/3.3.0-dev/lib/ruby/3.3.0+0/rubygems/core_ext/kernel_require.rb>:127:in `require'
        from -e:1:in `<main>'
```

In Ruby native require, `require "foo.so"` is a role process that "loads the dynamic library of foo (not the plain ruby .rb file)".
But such a thing is not possible in RubyGems require.

Since RubyGems overrides Ruby's native require, this functionality may be necessary as well.

### Reference 
The following documentation describes the native require behavior in Ruby.

https://github.com/ruby/ruby/blob/88d9a4d58af5c41a3258761dd6d0ea405fe47c07/load.c#L947-L990

However, this document also seems to describe something slightly different from Ruby's behavior, so I also suggest modifying this document.
https://github.com/ruby/ruby/pull/9180

I changed the behavior so that when checking if a file with a .so or .o extension is included in a gem, the extension is rewritten to DLEXT.
This behavior is consistent with Ruby native require.

### DIFF
--- a/lib/rubygems.rb
+++ b/lib/rubygems.rb
@@ -943,6 +943,13 @@ An Array (#{env.inspect}) was passed in from #{caller[3]}
   end
 
   ##
+  # Suffixes for dynamic library require-able paths.
+
+  def self.dynamic_library_suffixes
+    @dynamic_library_suffixes ||= suffixes - [".rb"]
+  end
+
+  ##
   # Prints the amount of time the supplied block takes to run using the debug
   # UI output.
 
--- a/lib/rubygems/basic_specification.rb
+++ b/lib/rubygems/basic_specification.rb
@@ -84,7 +84,13 @@ class Gem::BasicSpecification
       return false
     end
 
-    have_file? file, Gem.suffixes
+    is_soext = file.end_with?(".so", ".o")
+
+    if is_soext
+      have_file? file.delete_suffix(File.extname(file)), Gem.dynamic_library_suffixes
+    else
+      have_file? file, Gem.suffixes
+    end
   end
 
   def default_gem?
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -1561,6 +1561,17 @@ dependencies: []
     assert_empty err
   end
 
+  def test_contains_requirable_file_extension_soext
+    ext_spec
+    dlext = RbConfig::CONFIG["DLEXT"]
+    @ext.files += ["lib/ext.#{dlext}"]
+
+    FileUtils.mkdir_p @ext.extension_dir
+    FileUtils.touch File.join(@ext.extension_dir, "ext.#{dlext}")
+    FileUtils.touch File.join(@ext.extension_dir, "gem.build_complete")
+    assert @ext.contains_requirable_file? "ext.so"
+  end
+
   def test_date
     assert_date @a1.date
   end


deivid-rodriguez (approved):
This looks good to me.

simi:
Is this related to fact some extensions do have both `rb` and `so` (or other dynamic lib) with same name in same directory and basic `require` prefers `rb`?

deivid-rodriguez:
My understanding is that this is just a subtle difference between RubyGems require and standard require. Standard require accepts `.so` as a "generic" marker to signal "load the compiled extension for this feature, whatever the actual filename suffix is". But RubyGems require does not implement that.

This was preventing native `readline` from being required on Windows, and was causing `reline` to always be used as a fallback when requiring `deadline`.

simi:
> My understanding is that this is just a subtle difference between RubyGems require and standard require. Standard require accepts `.so` as a "generic" marker to signal "load the compiled extension for this feature, whatever the actual filename suffix is". But RubyGems require does not implement that.
> 
> This was preventing native `readline` from being required on Windows, and was causing `reline` to always be used as a fallback when requiring `deadline`.

Thanks for the info, that's clear. And is the original require handling "so" in special way related to the rb/so/dylib in one directory? I'm trying to understand it properly. This PR looks ok, it is good to mirror original `require` as much as possible.

simi (approved):

deivid-rodriguez:
Standard require behavior is explained [here](https://github.com/ruby/ruby/pull/9180/files), as researched by @hogelog.

From reading that, my understanding is that in presence of both `.rb` and `.so` files in the same directory, `.rb` wins.

By the way, I'm observing that while https://rubyapi.org/3.2/o/kernel#method-i-require looks good, the same link for [Ruby 3.3](https://rubyapi.org/3.3/o/kernel#method-i-require) does not show the same useful information. Maybe something is messed up somewhere.

segiddins (approved):

MSP-Greg:
JFYI, try https://msp-greg.github.io/ruby_trunk/Core/Kernel.html#require-class_method

Maybe some might consider this self-promotion, but a long time ago I wanted to view doc on my iPad, so I modified YARD.

hogelog:
Thank you for your kind review!
@deivid-rodriguez's description is exactly what I wanted to describe.

> From reading that, my understanding is that in presence of both .rb and .so files in the same directory, .rb wins.

Yes, Ruby's native require method prioritizes searching for .rb files first, followed by .so files.

> By the way, I'm observing that while https://rubyapi.org/3.2/o/kernel#method-i-require looks good, the same link for Ruby 3.3 does not show the same useful information. Maybe something is messed up somewhere.

I also noticed the discrepancy in the Ruby 3.3 documentation.

However, since `require` is in fact basically overridden by RubyGems' require, I am thinking that it might not be a bad idea to show this documentation.

If you think the documentation should reflect the native Ruby require method for Kernel.require, I am willing to work on correcting this.

deivid-rodriguez:
@hogelog I'm honestly not sure. I think both descriptions are useful but I'm not sure they can easily be displayed side by side in docs.

Thanks for your patch!

hogelog:
> I'm honestly not sure.

I agree with you. I will leave it as it is for now.

Thanks for merge!

PR #7242, opened 2023-12-11
Improve default gem handling by treating default gems as any other gem

deivid-rodriguez:

Bundler behaves inconsistently in presence of default gems, and that can be confusing. See for example the situation in https://github.com/ruby/ruby/pull/9163#issuecomment-1846994447.

My fix is that, once a default gem is specified directly in the Gemfile, or resolved as a transitive dependency, it becomes part of the bundle and gets treated as a "regular gem", so it's cached, explicitly installed in the configured location, etc.

### DIFF
--- a/bundler/lib/bundler/cli/common.rb
+++ b/bundler/lib/bundler/cli/common.rb
@@ -57,6 +57,9 @@ module Bundler
         specs << spec if regexp && spec.name =~ regexp
       end
 
+      default_spec = default_gem_spec(name)
+      specs << default_spec if default_spec
+
       case specs.count
       when 0
         dep_in_other_group = Bundler.definition.current_dependencies.find {|dep|dep.name == name }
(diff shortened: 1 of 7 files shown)


deivid-rodriguez:
First run of CI here made fail all Ruby 3.1 jobs. I think it was a temporary actions issue because it passed fine after retry.

I'm setting this as ready!

deivid-rodriguez:
There's some potential breakage that could happen here. For example, if someone is currently using `gem "bigdecimal", "3.1.5"` in their Gemfile, that currently works as long as you're using current master branch of Ruby. This PR will break that, because version 3.1.5 of bigdecimal does not exist in `rubygems.org` even if master branch of ruby installs that as a default gem.

Bacause of this, I'm tagging this break as breaking. I still this is a good change to ship with Bundler 2.5.

martinemde (approved):
Yeah, I like this change. Let's get it into 2.5.0.

simi (approved):

segiddins (approved):

deivid-rodriguez:
Thinking about this again, the breakage will be bigger than just some situations involving ruby-head.

For example, if `json` was previously part of your bundle on a machine where it's a default gem, and you upgrade Bundler without reinstalling gems, then `require "bundler/setup"` will run into

```
Could not find json-2.6.3 in locally installed gems
Run `bundle install` to install missing gems.
```

A reinstallation of gems should do the trick to get back on track.

deivid-rodriguez:
Example of this kind of breakage in our own repo at https://github.com/rubygems/rubygems/pull/7265.

martinemde:
That failure seems ok. Easy to fix.

The only thing I can imagine doing about it is printing a note when a default gem is missing. I don't think it's a big deal though.

PR #7244, opened 2023-12-11
Fix installing from source with same default bundler version already installed

deivid-rodriguez:

Extracting this bug fix from #7233. When installing bundler from source code, installation will fail if the same Bundler version is already the default Bundler gem. This is happens, for example, when installing from source using Ruby 3.3.0-preview3.

The problem is caused by the discrepancy between bindir folder in the pristine ruby 3.3.0-preview3 tarball, and the bindir folder in bundler's upstream source code.

This discrepancy is getting normalized at https://github.com/ruby/ruby/pull/9165, but it's also a bug that the bindir used to loop through upstream executables is the one of the previously installed default gem, and not the one from Bundler sources themselves.

This PR fixes that.

### DIFF
--- a/.github/workflows/install-rubygems.yml
+++ b/.github/workflows/install-rubygems.yml
@@ -25,6 +25,7 @@ jobs:
           - { name: "3.0", value: 3.0.6 }
           - { name: "3.1", value: 3.1.4 }
           - { name: "3.2", value: 3.2.2 }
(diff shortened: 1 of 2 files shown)


martinemde (approved):

indirect:
Great catch, thanks for fixing this!

PR #7255, opened 2023-12-12
Make sure --no-user-install is respected for auto user installation

### LINKED ISSUE #7237: `--no-user-install` should disable "auto-user-install"
Currently, the "auto-user-install" feature is written in a way, that if the GEM_HOME is not writable, gems are installed into user directory. However, I don't think that "auto-user-install" feature plays well with the `--no-user-install` option. I think that `--no-user-install` should do what it says and that is preventing installation into user directory. But that is not the case:

~~~
$ gem install gem2rpm --no-user-install
Defaulting to user installation because default installation directory (/usr/share/gems) is not writable.
WARNING:  You don't have /builddir/bin in your PATH,
	  gem executables will not run.
Successfully installed gem2rpm-1.0.2
Parsing documentation for gem2rpm-1.0.2
Done installing documentation for gem2rpm after 0 seconds
1 gem installed
~~~

The thing is that the "user_install" might have 3 states:

1) `nil` when the option was not specified
2) `true`, when the `--user-install` is specified and
3) `false` for `--no-user-install` case.

However, the current code treats the cases (1) and (3) the same.

voxik:
The `options[:user_install]` might have three states:
* `true`: `--user-install`
* `false`: `--no-user-install` and
* `nil`: option was not specified

However, this had not been respected previously and the `false` and `nil` were treated the same. This could lead to auto user installation despite `--no-user-install` being specified on the command line.

Fixes #7237

BTW I have added explicit comment about the 3 states. It is kept in separate commit in case somebody considers that redundant.

### DIFF
--- a/lib/rubygems/installer.rb
+++ b/lib/rubygems/installer.rb
@@ -682,7 +682,7 @@ class Gem::Installer
       if @build_root.nil?
         if options[:user_install]
           @gem_home = Gem.user_dir
-        elsif !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
+        elsif options[:user_install].nil? && !ENV.key?("GEM_HOME") && (File.exist?(Gem.dir) && !File.writable?(Gem.dir))
           say "Defaulting to user installation because default installation directory (#{Gem.dir}) is not writable."
           @gem_home = Gem.user_dir
         end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez (approved):
Nice, thanks for fixing this! Comment seems fine to me too!

deivid-rodriguez:
Merging to make sure we don't release any known issues with the "automatic fallback to user installation" feature that we're about to release. Thanks @voxik!

PR #7256, opened 2023-12-12
Bump rake version used by tests

deivid-rodriguez:

None, just keeping things up to date and making them easier to update in the future.

Make it explicit that some tests depend on a specific version of rake being installed, and update that version.


segiddins (approved):

k0kubun:
This failed ruby/ruby's `make test-bundler-parallel` https://github.com/ruby/ruby/actions/runs/7184961196/job/19567221887, despite "ruby-core / Bundler under a ruby-core setup" passing here.

Do you have any idea how we can fix this?

deivid-rodriguez:
Yeah! The automatic imported commit failed to import changes to `tool/bundler` files, so it's still using the old version. Those would need to be imported manually.

deivid-rodriguez:
https://github.com/ruby/ruby/pull/9206 should fix it!

k0kubun:
Thanks for the PR :+1:

> The automatic imported commit failed to import changes to tool/bundler files, so it's still using the old version. Those would need to be imported manually.

I just pushed https://github.com/ruby/ruby/commit/fd7decbf04f58438c04fa09fb325b16fbbec40c4, hoping it will not happen next time.

deivid-rodriguez:
Yeah, I was just looking into that same fix! 👍

PR #7258, opened 2023-12-12
More cleanups

deivid-rodriguez:

Trying to do some more spring clean.

Delete all the things and check CI.

### DIFF
--- a/bundler/lib/bundler/cli/common.rb
+++ b/bundler/lib/bundler/cli/common.rb
@@ -79,7 +79,6 @@ module Bundler
     end
 
     def self.default_gem_spec(name)
-      return unless Gem::Specification.respond_to?(:find_all_by_name)
       gem_spec = Gem::Specification.find_all_by_name(name).last
       gem_spec if gem_spec&.default_gem?
     end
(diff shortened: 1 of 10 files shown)


deivid-rodriguez:
If CI passes, this would be ready from my side. So setting this as ready in case CI passes :)

hsbt (approved):

bundler/lib/bundler/shared_helpers.rb
```diff
@@ -48,15 +48,11 @@ def in_bundle?
     end
 
     def chdir(dir, &blk)
-      Bundler.rubygems.ext_lock.synchronize do
-        Dir.chdir dir, &blk
-      end
+      Dir.chdir dir, &blk
```
segiddins, line 51: i think this is a change in behavior -- it might lead to conflicting `chdir`s when loading gemspecs. i think it still needs to be guarded by a mutex (same thing with `pwd` below)

deivid-rodriguez: You're right. The mutex name was a bit unfortunately because it suggested that we only chdir when building extensions, which we no longer do. But we indeed still chdir when loading gemspecs, and this removes syncronizing that. I'll restore the monitor.

deivid-rodriguez: I restored the monitor, but left any naming improvements for another time.

segiddins (approved):

PR #7261, opened 2023-12-13
Add missing resolv, timeout, and net-protocol licenses

### LINKED ISSUE #7260: Please fix the license of the project to include information about bundled libraries
As far as I can tell, there is LICENSE.txt file, which claims the project to be licensed under MIT or Ruby. There is also MIT.txt. That would be reasonable.

But recently, there was bundled e.g. timeout gem and as far as I can tell, it is licensed under [BSD-2-Clause license](https://github.com/ruby/timeout/blob/master/LICENSE.txt). But this information is not provided anywhere. Maybe the [PR](https://github.com/rubygems/rubygems/pull/6793/commits/cfc5018c54ecc2e0c771ec12baded5e3f7bf5fd5#diff-ee98e028c59b193d58fde56ab4daf54d43c486ae674e63d50ddf300b07943e0fR216) was done just carelessly, because there actually is something like `subsublib.license_path = "License.txt"`. But I'm not really sure about purpose of this construct.

In any case, I don't think the license information of the project is precise anymore and therefore it should be updated. This is unfortunately one danger of bundling.

deivid-rodriguez:

I made typos when specifying licenses for timeout and resolv at #6793.

Fix typos.

NOTE: I know it's not ideal that we currently have to specify the same info every time a vendored gem appears as a subdependency, but I did not care much about it since this is just support code to help us vendor the gems. 

Closes #7260.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -438,6 +438,7 @@ lib/rubygems/net-http/lib/net/http/response.rb
 lib/rubygems/net-http/lib/net/http/responses.rb
 lib/rubygems/net-http/lib/net/http/status.rb
 lib/rubygems/net-http/lib/net/https.rb
+lib/rubygems/net-protocol/LICENSE.txt
 lib/rubygems/net-protocol/lib/net/protocol.rb
 lib/rubygems/net/http.rb
 lib/rubygems/openssl.rb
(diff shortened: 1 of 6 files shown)


simi (approved):

deivid-rodriguez:
I added a change to automatic so that it raises if the license file is missing, and that detected that the net-protocol license was also missing, so I fixed that too.

segiddins (approved):

PR #7263, opened 2023-12-13
Use match? when regexp match data is unused

segiddins:
Improved performance / reduced allocations

### DIFF
--- a/bundler/lib/bundler/shared_helpers.rb
+++ b/bundler/lib/bundler/shared_helpers.rb
@@ -334,7 +334,7 @@ module Bundler
     def set_rubyopt
       rubyopt = [ENV["RUBYOPT"]].compact
       setup_require = "-r#{File.expand_path("setup", __dir__)}"
-      return if !rubyopt.empty? && rubyopt.first =~ /#{Regexp.escape(setup_require)}/
+      return if !rubyopt.empty? && rubyopt.first.include?(setup_require)
       rubyopt.unshift setup_require
       Bundler::SharedHelpers.set_env "RUBYOPT", rubyopt.join(" ")
     end
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -127,8 +127,8 @@ module Bundler
 
       if man_pages.include?(command)
         man_page = man_pages[command]
-        if Bundler.which("man") && man_path !~ %r{^file:/.+!/META-INF/jruby.home/.+}
-          Kernel.exec "man #{man_page}"
+        if Bundler.which("man") && !man_path.match?(%r{^file:/.+!/META-INF/jruby.home/.+})
+          Kernel.exec("man", man_page)
         else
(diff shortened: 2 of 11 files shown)


deivid-rodriguez (approved):
Surprised `Performance/RegexpMatch` does not catch this!

bundler/lib/bundler/rubygems_integration.rb
```diff
@@ -119,7 +119,7 @@ def inflate(obj)
     def correct_for_windows_path(path)
       if Gem::Util.respond_to?(:correct_for_windows_path)
         Gem::Util.correct_for_windows_path(path)
-      elsif path[0].chr == "/" && path[1].chr =~ /[a-z]/i && path[2].chr == ":"
+      elsif path[0].chr == "/" && path[1].chr.match?(/[a-z]/i) && path[2].chr == ":"
```
deivid-rodriguez, line 122: I'm planning to delete this code at #7258, which suggests that we can probably apply the same fix in the rubygems side.

bundler/lib/bundler/shared_helpers.rb
```diff
@@ -329,7 +329,7 @@ def set_path
     def set_rubyopt
       rubyopt = [ENV["RUBYOPT"]].compact
       setup_require = "-r#{File.expand_path("setup", __dir__)}"
-      return if !rubyopt.empty? && rubyopt.first =~ /#{Regexp.escape(setup_require)}/
+      return if !rubyopt.empty? && rubyopt.first.include?(setup_require)
```
deivid-rodriguez, line 332: Not sure this is correct, there may be other stuff in RUBYOPT.

segiddins: matching against the regexp in this case is only checking if the string includes what's in the regexp, especially since the regexp is an escaped version of the string

bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -282,7 +282,7 @@ def match_platforms?(platform, platforms)
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
   class BasicSpecification
-    if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
+    if /^universal\.(?<arch>.*?)-/.match?(CROSS_COMPILING || RUBY_PLATFORM)
```
segiddins, line 285: whoops this one is incorrect, `?<arch>` is capturing a variable here

martinemde (approved):

martinemde:
One of the methods ... moved, I think? Want to correct that here or just resolve the conflict?

deivid-rodriguez:
Yeah, the method was deleted at #7258, we'd want to apply the same patch, but here: https://github.com/rubygems/rubygems/blob/c09c923440b149af46aefe447fc4dd7f504f8365/lib/rubygems/util.rb#L107-L116.

deivid-rodriguez:
I resolved the conflict 👍.

PR #7266, opened 2023-12-13
Revert treating default gems as regular gems

deivid-rodriguez:

I merged #7242 recently but on second thought I think there's probably a better compatibility story here. We can make default gems be installed and cached by default, but still use them when missing during `Bundler.setup`, possibly with a warning that advices to rebundle so that the default gem is included with the rest of the bundle.

I don't have time to work on the above now, so I'd like to revert the PR for now to make it for a smoother release.

### DIFF
--- a/bundler/lib/bundler/rubygems_integration.rb
+++ b/bundler/lib/bundler/rubygems_integration.rb
@@ -469,7 +469,7 @@ module Bundler
     end
 
     def all_specs
-      Gem::Specification.stubs.reject(&:default_gem?).map do |stub|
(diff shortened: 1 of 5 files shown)


martinemde (approved):

martinemde:
Rebased to fix conflict.

deivid-rodriguez:
Actually I just noticed that the original PR has three commits, maybe the first two are an improvement that could stay?

martinemde:
Yeah, that totally makes sense.

PR #7274, opened 2023-12-14
Fix Bundler daily CI

deivid-rodriguez:

Daily Bundler CI (which probably got fixed after we added 3.0.0-rc1 to our regular CI), got broken again after new vendored libraries landed into ruby-core.

A default Bundler copy with vendored net-http is causing a Bundler spec to end up loading that copy instead of the copy of Bundler under test. 

This is because of vcr, but we don't really need vcr in the command that's failing, so I avoided loading it.
 

bundler/spec/runtime/self_management_spec.rb
```diff
@@ -28,7 +28,7 @@
       expect(out).to include("Bundler #{Bundler::VERSION} is running, but your lockfile was generated with #{previous_minor}. Installing Bundler #{previous_minor} and restarting using that version.")
 
       # It uninstalls the older system bundler
-      bundle "clean --force"
+      bundle "clean --force", artifice: nil
```
deivid-rodriguez, line 31: I love being able to use modern hashes 😄

segiddins (approved):

PR #7283, opened 2023-12-14
Checksum::Store mutex

martinemde:

Although ruby/ruby#9240 may fix the underlying problem, the hash access errors in Checksum::Store are cause for some concerns.

Use a mutex to synchronize checksum store hash access.

### DIFF
--- a/bundler/lib/bundler/checksum.rb
+++ b/bundler/lib/bundler/checksum.rb
@@ -63,6 +63,10 @@ module Bundler
 
     alias_method :eql?, :==
 
+    def same_source?(other)
+      sources.include?(other.sources.first)
+    end
+
     def match?(other)
       other.is_a?(self.class) && other.digest == digest && other.algo == algo
     end
@@ -161,6 +165,7 @@ module Bundler
 
       def initialize
         @store = {}
+        @store_mutex = Mutex.new
       end
 
       def inspect
@@ -168,8 +173,9 @@ module Bundler
       end
 
       # Replace when the new checksum is from the same source.
-      # The primary purpose of this registering checksums from gems where there are
+      # The primary purpose is registering checksums from gems where there are
       # duplicates of the same gem (according to full_name) in the index.
+      #
       # In particular, this is when 2 gems have two similar platforms, e.g.
       # "darwin20" and "darwin-20", both of which resolve to darwin-20.
       # In the Index, the later gem replaces the former, so we do that here.
@@ -179,16 +185,14 @@ module Bundler
       # that contain the same gem with different checksums.
       def replace(spec, checksum)
         return unless checksum
-
         lock_name = spec.name_tuple.lock_name
-        checksums = (store[lock_name] ||= {})
-        existing = checksums[checksum.algo]
-
-        # we assume only one source because this is used while building the index
-        if !existing || existing.sources.first == checksum.sources.first
-          checksums[checksum.algo] = checksum
-        else
-          register_checksum(lock_name, checksum)
+        @store_mutex.synchronize do
+          existing = fetch_checksum(lock_name, checksum.algo)
+          if !existing || existing.same_source?(checksum)
+            store_checksum(lock_name, checksum)
+          else
+            merge_checksum(lock_name, checksum, existing)
+          end
         end
       end
 
@@ -207,7 +211,8 @@ module Bundler
 
       def to_lock(spec)
         lock_name = spec.name_tuple.lock_name
-        if checksums = store[lock_name]
+        checksums = @store[lock_name]
+        if checksums
           "#{lock_name} #{checksums.values.map(&:to_lock).sort.join(",")}"
         else
           lock_name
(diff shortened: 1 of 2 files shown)


bundler/lib/bundler/checksum.rb
```diff
@@ -161,26 +165,17 @@ class Store
 
       def initialize
         @store = {}
-      end
-
-      def initialize_copy(other)
```
segiddins, line 166: Why are you removing this?

bundler/lib/bundler/checksum.rb
```diff
@@ -228,16 +223,32 @@ def to_lock(spec)
...
+        if existing
+          merge_checksum(lock_name, checksum, existing)
         else
-          raise ChecksumMismatchError.new(lock_name, existing, checksum)
+          store_checksum(lock_name, checksum)
+        end
+      end
+
+      def merge_checksum(lock_name, checksum, existing)
+        existing.merge!(checksum) || raise(ChecksumMismatchError.new(lock_name, existing, checksum))
```
segiddins, line 236: Should this also be protected by the mutex?

bundler/lib/bundler/checksum.rb
martinemde: It's never used, but if you think there's value to it. I couldn't think of a case where we'd dup it legitimately and there's no test for it.

bundler/lib/bundler/checksum.rb
martinemde: `merge!` just compares the digest/algo, concats the `source` onto `sources`, then runs `sources.uniq!`. Even run concurrently without global lock, I think that's still safe.

bundler/lib/bundler/checksum.rb
martinemde: In actuality, when I was adding the mutex I had to duplicate the mutex into here, then also lock the copy process and it seemed excessive for something that we don't use.

segiddins: ah if it is never used then we're good, I remember it was used at some point

bundler/lib/bundler/checksum.rb
segiddins: if it's modifying any shared state, it is not safe -- remember not all Rubies have a GIL

martinemde: How can you tell which operations are safe on global state, and which aren't? This is tue code:

```
@sources.concat(other.sources).uniq!
```

This brings up another situation in here that isn't locked. I'll add a comment there.

bundler/lib/bundler/checksum.rb
```diff
@@ -228,16 +223,32 @@ def to_lock(spec)
...
+      end
+
+      def store_checksum(lock_name, checksum)
+        @store_mutex.synchronize do
+          (@store[lock_name] ||= {}).store(checksum.algo, checksum)
+        end
+      end
+
+      def fetch_checksum(lock_name, algo)
+        fetch_checksums(lock_name)&.fetch(algo, nil)
```
martinemde, line 246: @segiddins This is a fetch on a Hash (the nested one). We're syncing other reads, but is this one safe? Note the second half of this method is not synced.

segiddins: any read or write on data that is (at the time of access) shared across threads (or fibers, actors, etc), needs to be synchronized. typically, this means not building upon primitive access methods directly, but rather exposing a set of public methods that each synchronizes access to the shared data

segiddins (approved):

bundler/lib/bundler/checksum.rb
```diff
@@ -217,18 +222,27 @@ def to_lock(spec)
       private
 
       def register_checksum(lock_name, checksum)
-        return unless checksum
```
deivid-rodriguez, line 231: You realized we were always passing non nil here, correct?

bundler/lib/bundler/checksum.rb
```diff
@@ -179,16 +185,14 @@ def inspect
...
-        existing = checksums[checksum.algo]
-
-        # we assume only one source because this is used while building the index
-        if !existing || existing.sources.first == checksum.sources.first
-          checksums[checksum.algo] = checksum
-        else
-          register_checksum(lock_name, checksum)
+        @store_mutex.synchronize do
+          existing = fetch_checksum(lock_name, checksum.algo)
+          if !existing || existing.same_source?(checksum)
```
deivid-rodriguez, line 191: The implementation of this method sounds slightly different than before? Also should the comment be carried over?

bundler/lib/bundler/checksum.rb
```diff
@@ -179,16 +185,14 @@ def inspect
       # that contain the same gem with different checksums.
       def replace(spec, checksum)
         return unless checksum
-
```
deivid-rodriguez, line 193: We do generally keep a blank line after guards, I think.

deivid-rodriguez (commented):
Looks very good, just some minor comments on stuff unrelated to the "mutexing".

bundler/lib/bundler/checksum.rb
martinemde: Yes, this was always guarded by the callers, and the method is private.

bundler/lib/bundler/checksum.rb
martinemde: I short-circuited. `register_checksum` was re-fetching the existing checksum when we already have it here.

bundler/lib/bundler/checksum.rb
martinemde: Yeah, this is a bit claustrophobic. Will fix.

bundler/lib/bundler/checksum.rb
martinemde: The comment is no longer correct. I made it so same_source checks for any matching source on existing, and the incoming one should always have only 1. I think the method name explains it well enough now.

deivid-rodriguez: I meant the one you extracted, `same_source?`.

deivid-rodriguez: I get it now!

PR #7285, opened 2023-12-15
Refactor vendoring to allow validating vendoring is reproducible

segiddins:
Helps ensure that unsuspecting diffs to the vendored code arent accidentally introduced

Changes to vendored code are hard to review, and it would be easy for a change to be slipped in that isn't actually from upstream. I deem this to be an attack vector, and one that we don't need to have.

Make the vendoring process idempotent, so we can (in CI) check that re-vendoring from upstream does not introduce any diffs.

As a part of that, use bundler to do the fetching & unpacking, so we can piggy-back off of its caching & parallel installation & checksum validation.


deivid-rodriguez (approved):
Wow. This is freaking awesome, I'm speechless :)

simi (approved):
Massive!

PR #7116, opened 2023-10-26
Drop ruby 2.6 and 2.7 support

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Time to move on. These are EOL.

## What is your fix for the problem, implemented in this PR?

Delete all the things.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [ ] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7291, opened 2023-12-15
Handle realworld specs lost when dropping Ruby 2.7

deivid-rodriguez:

Unfortunately I couldn't find a non-realworld scenario when introducing these specs to prove fixes, and they were specific to ruby 2.7 versions.

After dropping Ruby 2.7 support, coverage was lost.

For one of them, I'm assuming the coverage loss because the Gemfile was very particular to Ruby 2.7, and I wouldn't know how to do better now, the fix happened a long time ago. We could try patching the current ruby to fake it's Ruby 2.7, but not sure if it's worth the effort.

For the other one, I moved it to a spec that runs for all rubies and just makes sure that the Gemfile can be resolved in "reasonable" time.

- [x] ~Write~ Delete [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes


deivid-rodriguez:
Also included a small change to improve output of specs that assert for "reasonable resolution time".

Happy to get suggestions on what's best here.

martinemde (approved):
image

I read the actual code changes. They seem good.

deivid-rodriguez:
Thanks! Yeah, it's a big diff, but technically it's files that should've not been removed with https://github.com/rubygems/rubygems/pull/7116 if we wanted to keep the coverage. I guess it's good for now!

PR #7292, opened 2023-12-15
Use faster macos-13 in CI

deivid-rodriguez:

Slow CI.

Checking if this speeds up macOS.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -61,9 +61,6 @@ jobs:
       - name: Install graphviz (Ubuntu)
         run: sudo apt-get install graphviz -y
         if: matrix.bundler.value == '' && matrix.os.name == 'Ubuntu'
-      - name: Install graphviz (macOS)
-        run: brew install graphviz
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
macos-13 is definitely faster than macos-12. Trying macos-arm-oss now.

martinemde:
Seems like this version of mac os stays queued much longer.

deivid-rodriguez:
Yep, we should stick with macos-13!

MSP-Greg:
> Seems like this version of mac os stays queued much longer.

Friday afternoon US is often slow across all platforms.  Also, I think macOS also has lower concurrency limits for most GitHub 'membership' plans.  Or, in Puma, often waiting on macOS CI.  We've got 16 macOS jobs.

deivid-rodriguez:
macOS has trouble installing graphviz for deprecated `bundle viz` specs. I went ahead and started skipping those on macOS. I think ubuntu is enough to avoid regressions there, it's code that hardly ever regresses or gets any updates.

martinemde (approved):

PR #7296, opened 2023-12-16
Implement SafeMarshal support for hash compare by identity

segiddins:
So there is not a lingering pending test

Intentionally do not support other User Class cases for now

### DIFF
--- a/lib/rubygems/safe_marshal/visitors/to_ruby.rb
+++ b/lib/rubygems/safe_marshal/visitors/to_ruby.rb
@@ -247,6 +247,30 @@ module Gem::SafeMarshal
         end
       end
 
+      def visit_Gem_SafeMarshal_Elements_UserClass(r)
+        if resolve_class(r.name) == ::Hash && r.wrapped_object.is_a?(Elements::Hash)
+
+          hash = register_object({}.compare_by_identity)
+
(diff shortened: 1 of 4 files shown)


martinemde (approved):
Ok, so Marshal considers a Hash set to compare by identity as a user class, but since it's a safe loadable class we specifically allow only that class through the user class loader? Looks correct to me but I'm no Marshal pro.

segiddins:
Yeah, there are even comments in marshal.c about it being a hack

lib/rubygems/safe_marshal/visitors/to_ruby.rb
```diff
@@ -247,6 +247,30 @@ def visit_Gem_SafeMarshal_Elements_Bignum(b)
         end
       end
 
+      def visit_Gem_SafeMarshal_Elements_UserClass(r)
+        if resolve_class(r.name) == ::Hash && r.wrapped_object.is_a?(Elements::Hash)
+
```
deivid-rodriguez, line 252: Is this blank line intentional?

deivid-rodriguez (approved):
No Marshal expert either but makes sense to me!

deivid-rodriguez:
I'll merge this since a blank line is no reason to block it :)

PR #7300, opened 2023-12-17
Ensure File.open applies default umask on gem extract

martinemde:
**This is an alternate to #7299**

Global and group write permissions are possible in gems.

Allow File.open to handle applying the default umask.

Non-backwards-compatible change: This drops previously tested persistence of extended file modes.

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -448,13 +448,15 @@ EOM
           end
 
         unless directories.include?(mkdir)
-          FileUtils.mkdir_p mkdir, mode: dir_mode ? 0o755 : (entry.header.mode if entry.directory?)
+          mkdir_mode = 0o755 if dir_mode
+          mkdir_mode ||= entry.header.mode if entry.directory?
+          mkdir_mode &= ~File.umask if mkdir_mode
+          FileUtils.mkdir_p mkdir, mode: mkdir_mode
           directories << mkdir
         end
 
         if entry.file?
-          File.open(destination, "wb") {|out| out.write entry.read }
-          FileUtils.chmod file_mode(entry.header.mode), destination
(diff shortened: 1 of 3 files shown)


segiddins (approved):

deivid-rodriguez (commented):
This looks good to me!

The only thing intentionally being reverted is 3211b16425be488b5d0998270c1d1d84e34f377f. I agree it's best to respect umask by default, and I would treat this as a security fix, but let me ask @nobu in case there was some reason we are missing for that change.

@nobu Is it ok for us to revert 3211b16425be488b5d0998270c1d1d84e34f377f? Do you remember of any issues with the default umask being respected? Perhaps it was just for consistency with it not being respected in other places?

martinemde:
@deivid-rodriguez @nobu If there's a problem with this PR, #7299 would be the right one to switch to, with maybe an intentional `mode & 0o700` added.

deivid-rodriguez:
mmm so in few words, #7299 is more backwards compatible (defining that by our existing tests) but less secure?

martinemde:
Yes. On my system #7299 doesn't clear setuid/setgid bits on files. Allowing File.open to do its default clears those bits. this implies the File.umask (usually `0o022` or `0o077`) is a limited version (9 bits?) of the actual umask that is applied by File.open (which is something more like `0o7022` but is probably much more platform dependent)

deivid-rodriguez:
Thanks for explaining. Are you comfortable then with shipping this as a security fix? I think I am.

martinemde:
I prefer this version, but there's an untested implied behavior in the original commit by @nobu that may mean some gems on some systems don't get their 0700 set.

deivid-rodriguez:
I see. Let's wait for @nobu's feedback then.

hsbt:
This change break one of our CI https://app.travis-ci.com/github/ruby/ruby/jobs/615527743

PR #7303, opened 2023-12-18
Replace `object_id` comparison with identity Hash

amomchilov:

Found this while auditing uses of `object_id` in `ruby/ruby`. Pulled out from ruby/ruby#9276

Replace a `Hash` keyed by `object_id`s with an identity Hash. See `Hash#compare_by_identity`](https://ruby-doc.org/3.2.2/Hash.html#method-i-compare_by_identity), which has the same semantics, but is faster and doesn't trigger allocation of IDs for these objects.

- [ ] ~Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes~
- [ x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

### DIFF
--- a/lib/rubygems/request.rb
+++ b/lib/rubygems/request.rb
@@ -30,7 +30,7 @@ class Gem::Request
     @uri = uri
     @request_class = request_class
     @last_modified = last_modified
-    @requests = Hash.new 0
+    @requests = Hash.new(0).compare_by_identity
     @user_agent = user_agent
 
     @connection_pool = pool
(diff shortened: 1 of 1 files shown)


martinemde (approved):

deivid-rodriguez (approved):

amomchilov:
Thanks for merging! @martinemde

martinemde:
@amomchilov I did the easy part :grinning: Thank you for finding and contributing this back to rubygems!

PR #7309, opened 2023-12-19
Upgrade *-artifact actions together

deivid-rodriguez:

These actions sound like they'd better be upgraded together.

Configure Dependabot to do that!


martinemde:
I wonder if this will trigger the update or do we need to do it ourselves.

It's also going to need changes to the action, I think.

deivid-rodriguez:
It should open a PR updating both dependencies that we can use as a base to work on the upgrade.

hsbt:
👍

PR #7324, opened 2023-12-22
Fix resolution when different platform specific gems have different dependencies

### LINKED ISSUE #7321: google-protobuf won't build on Linux after adding MinGW platform
### Discussed in https://github.com/rubygems/rubygems/discussions/7315

<div type='discussions-op-text'>

<sup>Originally posted by **AnyOldName3** December 20, 2023</sup>
I had a working Jekyll project that built fine under both real Linux and WSL2, but for reasons not worth getting into, needed to switch from WSL2 to Windows. I ran `bundle lock --add-platform x64-mingw-ucrt01`, the results of which can be seen as this commit: https://github.com/AnyOldName3/osg-website-replacement/commit/a72328e722c0ab38b0914794e767443a14210fb6. The line I believe to be most important changes `google-protobuf (3.25.1-x86_64-linux)` to `google-protobuf (3.25.1)`, and looking at the contents of the two packages, it appears the former has prebuilt `.so`s whereas the latter is source-only, so the shared library needs building.

As intended, this made the project work with Windows Ruby, however, it also produced the following errors on Linux:
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /mnt/c/vsg/osg-website-replacement/vendor/bundler/ruby/3.0.0/gems/google-protobuf-3.25.1/ext/google/protobuf_c
/usr/bin/ruby3.0 -I /usr/lib/ruby/vendor_ruby -r ./siteconf20231220-407-i7czk0.rb extconf.rb
creating Makefile

current directory: /mnt/c/vsg/osg-website-replacement/vendor/bundler/ruby/3.0.0/gems/google-protobuf-3.25.1/ext/google/protobuf_c
make DESTDIR\= clean

current directory: /mnt/c/vsg/osg-website-replacement/vendor/bundler/ruby/3.0.0/gems/google-protobuf-3.25.1/ext/google/protobuf_c
make DESTDIR\=
compiling convert.c
convert.c: In function ‘Msgval_IsEqual’:
convert.c:295:58: warning: passing argument 4 of ‘shared_Msgval_IsEqual’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  295 |                                             type_info.def.msgdef, &status);
      |                                             ~~~~~~~~~~~~~^~~~~~~
In file included from convert.c:21:
shared_convert.h:19:60: note: expected ‘upb_MessageDef *’ but argument is of type ‘const upb_MessageDef *’
   19 |                            upb_CType type, upb_MessageDef* msgdef,
      |                                            ~~~~~~~~~~~~~~~~^~~~~~
convert.c:299:5: error: format not a string literal and no format arguments [-Werror=format-security]
  299 |     rb_raise(rb_eRuntimeError, upb_Status_ErrorMessage(&status));
      |     ^~~~~~~~
convert.c: In function ‘Msgval_GetHash’:
convert.c:308:41: warning: passing argument 3 of ‘shared_Msgval_GetHash’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  308 |       val, type_info.type, type_info.def.msgdef, seed, &status);
      |                            ~~~~~~~~~~~~~^~~~~~~
In file included from convert.c:21:
shared_convert.h:23:48: note: expected ‘upb_MessageDef *’ but argument is of type ‘const upb_MessageDef *’
   23 |                                upb_MessageDef* msgdef, uint64_t seed,
      |                                ~~~~~~~~~~~~~~~~^~~~~~
convert.c:312:5: error: format not a string literal and no format arguments [-Werror=format-security]
  312 |     rb_raise(rb_eRuntimeError, upb_Status_ErrorMessage(&status));
      |     ^~~~~~~~
cc1: some warnings being treated as errors
make: *** [Makefile:245: convert.o] Error 1

make failed, exit code 2
```
This log is from using WSL2, but it also happens in CI: https://github.com/AnyOldName3/osg-website-replacement/actions/runs/7266847247/job/19829238652. I believe that's Arch running in Docker (on an Ubuntu host) whereas WSL2 is running Ubuntu, so dependency packages from the OS should be different - they're at least using different versions of Ruby.

`-Werror=format-security` is coming from `rbconfig.rb` on the WSL2 system, and I imagine that's where it's coming from on the probably-Arch VM, too. Conceivably, other compiler flags requested by that file could be responsible for the other errors. I don't know whether this is a case of a broken gem (i.e. that these flags should work), a packaging error for Ruby itself (i.e. that these flags shouldn't have ended up in `rbconfig.rb`, or some other kind of problem. I also don't know how I'd go about working around this, e.g. by somehow forcing `Gemfile.lock` to use the prebuilt Linux variant of the package on Linux.</div>

deivid-rodriguez:

If two platform specific variants of the same gem version have different dependencies, our resolver may fallback to picking only the RUBY variant. However, _some_ platform specific gems may still be compatible with the resolution.

Do one pass after resolution completing platforms specific gems that may be compatible.

Fixes #7321.
 
### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -312,10 +312,6 @@ module Bundler
       end
     end
 
-    def should_complete_platforms?
-      !lockfile_exists? && generic_local_platform_is_ruby? && !Bundler.settings[:force_ruby_platform]
-    end
-
     def spec_git_paths
       sources.git_sources.map {|s| File.realpath(s.path) if File.exist?(s.path) }.compact
     end
@@ -516,6 +512,10 @@ module Bundler
     end
 
     private
+
+    def should_complete_platforms?
+      !lockfile_exists? && generic_local_platform_is_ruby? && !Bundler.settings[:force_ruby_platform]
+    end
 
     def lockfile_exists?
       lockfile && File.exist?(lockfile)
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -602,6 +602,8 @@ module Bundler
       @resolved_bundler_version = result.find {|spec| spec.name == "bundler" }&.version
       @platforms = result.add_extra_platforms!(platforms) if should_add_extra_platforms?
 
+      result.complete_platforms!(platforms)
+
       SpecSet.new(result.for(dependencies, false, @platforms))
     end
 
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -513,7 +513,7 @@ module Bundler
 
     private
 
-    def should_complete_platforms?
+    def should_add_extra_platforms?
       !lockfile_exists? && generic_local_platform_is_ruby? && !Bundler.settings[:force_ruby_platform]
     end
 
@@ -600,7 +600,7 @@ module Bundler
       result = SpecSet.new(resolver.start)
 
       @resolved_bundler_version = result.find {|spec| spec.name == "bundler" }&.version
-      @platforms = result.complete_platforms!(platforms) if should_complete_platforms?
+      @platforms = result.add_extra_platforms!(platforms) if should_add_extra_platforms?
 
       SpecSet.new(result.for(dependencies, false, @platforms))
     end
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I added a spec for this but turns out our own lockfile was suffering from this and caught the bug too!

Issue #6616, opened 2023-04-17
BUNDLE_GLOBAL_GEM_CACHE breaks natively compiled libraries


brodock:
### Describe the problem as clearly as you can

When setting up `BUNDLE_GLOBAL_GEM_CACHE=1` it installs the gems correctly. In a CI environment you want to carry that cache over to a next build. When the next CI run happens (which contains only the cache folder from previous execution) the gems "install" correctly but when trying to execute the code it claims the gems with native extensions are missing the compiled extension/library.

You can see an example here: 

https://gitlab.com/gitlab-org/gitlab-development-kit/-/jobs/4117614795

```
LoadError: cannot find 'llhttp-ext' library
/builds/gitlab-org/gitlab-development-kit/gdk/gitlab/vendor/bundle/ruby/3.0.0/gems/ffi-compiler-1.0.1/lib/ffi-compiler/loader.rb:21:in `find'
/builds/gitlab-org/gitlab-development-kit/gdk/gitlab/vendor/bundle/ruby/3.0.0/gems/llhttp-ffi-0.4.0/lib/llhttp.rb:13:in `<module:LLHttp>'
/builds/gitlab-org/gitlab-development-kit/gdk/gitlab/vendor/bundle/ruby/3.0.0/gems/llhttp-ffi-0.4.0/lib/llhttp.rb:6:in `<top (required)>'
<internal:/home/gdk/.asdf/installs/ruby/3.0.5/lib/ruby/site_ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
<internal:/home/gdk/.asdf/installs/ruby/3.0.5/lib/ruby/site_ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:37:in `require'
```

### Did you try upgrading rubygems & bundler?

Yes, I've tried using latest versions of both:

```
Bundler version 2.4.12

$ gem --version
3.4.11
```

### Post steps to reproduce the problem

I've tested this locally by setting:

```
export BUNDLE_GLOBAL_GEM_CACHE=1
export BUNDLE_USER_CACHE=~/.cache/bundle
```

And doing the following:
1. On the original project: `bundle install`
2. Go back to another project with fewer gems: `bundle clean --force`
3. Go back to the original project: `bundle install`
4. Try to run rails console (and get a lot of missing libraries errors, including `pg`)

### Which command did you run?

* `bundle install`
* `rails console`

### What were you expecting to happen?

* It should load the code correctly

[...]

PR #7325, opened 2023-12-22
Use a global bundler cache by default

deivid-rodriguez:

### DIFF
--- a/bundler/lib/bundler/feature_flag.rb
+++ b/bundler/lib/bundler/feature_flag.rb
@@ -32,7 +32,6 @@ module Bundler
     settings_flag(:cache_all) { bundler_3_mode? }
     settings_flag(:default_install_uses_path) { bundler_3_mode? }
     settings_flag(:forget_cli_options) { bundler_3_mode? }
-    settings_flag(:global_gem_cache) { bundler_3_mode? }
     settings_flag(:path_relative_to_cwd) { bundler_3_mode? }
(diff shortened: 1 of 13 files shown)


MSP-Greg:
@deivid-rodriguez

I haven't had time to look at the code, and I think this is a good idea.  But, is there a way to switch back to the current behavior?

I think this will affect setup-ruby's caching when doing `bundle install`.

deivid-rodriguez:
I was not expecting this to cause issues, just maybe the cache to not be hit the first time new bundler runs on an app. But I did not really give this a lot of thought and in particular did not think of `ruby/setup-ruby`. I will consider it!

MSP-Greg:
I was working on some other things earlier.  I was mistaken, as I don't think this change will affect CI for most users of setup-ruby.

Some CI may run `bundle install` with different `Gemfile` files and bundle paths, in which case they may need to download a few more gems, but they could also cache the global bundler cache independent of setup-ruby.

Sorry for the erroneous post...

deivid-rodriguez:
As a small note, there's one existing issue with the global shared gem cache that we should probably address before enabling this: https://github.com/rubygems/rubygems/issues/6616.

From a first look at that issue, I noticed that `BUNDLE_GLOBAL_GEM_CACHE` not enables global caching of `.gem` packages but also global caching of compiled extensions.

So another question of whether it would make sense to decouple these two. If it makes sense, then global caching of `.gem` packages could be shipped independently without #6616 interfering.

PR #7331, opened 2023-12-26
Use cache_home instead of data_home in default_spec_cache_dir

mrkn:

This isn't a pragmatic problem for end-users and developers.
I found just an inconsistency in using XDG's directories.

Replace the base directory of `default_spec_cache_dir` from `Gem.data_dir` to `Gem.cache_dir`.
I guess `Gem.cache_dir` is appropriate for the base dir of `default_spec_cache_dir`.

I don't think we need to add any new tests for this change.

### DIFF
--- a/lib/rubygems/defaults.rb
+++ b/lib/rubygems/defaults.rb
@@ -24,7 +24,7 @@ module Gem
     default_spec_cache_dir = File.join Gem.user_home, ".gem", "specs"
 
     unless File.exist?(default_spec_cache_dir)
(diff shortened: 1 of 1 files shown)


hsbt (approved):

hsbt:
Thanks!

deivid-rodriguez:
Are there any backwards compatibility concerns here? Just trying to figure out how/when this change should be released and which changelog section it should go in (enhancement, breaking change, not included since it's not user visible, etc).

hsbt:
Cache directory is not impact for compatibility.

PR #7334, opened 2023-12-26
Make `gem update --system` respect ruby version constraints

### LINKED ISSUE #7329: `gem update --system` does not respect RubyGems constraints on current ruby

`gem update --system` fails on ruby 2.7, or at least on ruby 2.7 as installed by `ruby/setup-ruby`.  Specifically, this is causing CI failures for gems that still support older rubies.

See e.g: https://github.com/ruby/net-imap/actions/runs/7273131765/job/19916818198?pr=251

This is, in fact, the problem. 😉 

I assume this is happening for other gems supporting older versions of ruby as well.  The workflow file is fairly standard.

```yml
- name: Set up Ruby
      uses: ruby/setup-ruby@v1
      with:
        bundler-cache: true
        rubygems: latest
```

```
/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/gem --version
  3.1.6
  Default RubyGems version is 3.1.6
  Updating RubyGems to latest version
 
/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/gem update --system
  ERROR:  Error installing rubygems-update:
  	There are no versions of rubygems-update (= 3.5.3) compatible with your Ruby & RubyGems
  	rubygems-update requires Ruby version >= 3.0.0. The current ruby version is 2.7.8.[22](https://github.com/ruby/net-imap/actions/runs/7273131765/job/19916818198?pr=251#step:3:27)5.
      undefined method `version' for nil:NilClass
  Updating rubygems-update
  Took  14.85 seconds
Error: The process '/opt/hostedtoolcache/Ruby/2.7.8/x64/bin/gem' failed with exit code 1
```

The latest (compatible) version would be installed.

### What actually happened?

The command failed.

### Misc

Perhaps this is related to https://github.com/rubygems/rubygems/pull/7310?  For purposes of passing CI, the simple workaround is to set a version constraint rather than use `latest`.  Still, it would be nice if that were unnecessary. 🙂

deivid-rodriguez:

I thought `gem update --system` used the same `gem install` logic under the hood, so after `gem install` was fixed to respect ruby version constraints, it would also respect them.

However, that's not the case.

Remove code to resolve `rubygems-update` version manually, and instead let the resolver do its job.

Fixes #7329.

### DIFF
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -282,7 +282,7 @@ command to remove old versions.
     check_oldest_rubygems version
 
     installed_gems = Gem::Specification.find_all_by_name "rubygems-update", requirement
-    installed_gems = update_gem("rubygems-update", version) if installed_gems.empty? || installed_gems.first.version != version
+    installed_gems = update_gem("rubygems-update", requirement) if installed_gems.empty? || installed_gems.first.version != version
     return if installed_gems.empty?
 
     install_rubygems installed_gems.first
--- a/lib/rubygems/commands/update_command.rb
+++ b/lib/rubygems/commands/update_command.rb
@@ -294,9 +294,7 @@ command to remove old versions.
     args << "--prefix" << Gem.prefix if Gem.prefix
     args << "--no-document" unless options[:document].include?("rdoc") || options[:document].include?("ri")
     args << "--no-format-executable" if options[:no_format_executable]
-    args << "--previous-version" << Gem::VERSION if
-      options[:system] == true ||
-      Gem::Version.new(options[:system]) >= Gem::Version.new(2)
+    args << "--previous-version" << Gem::VERSION
     args
   end
 
--- a/test/rubygems/test_gem_commands_update_command.rb
+++ b/test/rubygems/test_gem_commands_update_command.rb
@@ -130,6 +130,34 @@ class TestGemCommandsUpdateCommand < Gem::TestCase
     assert_empty err
   end
 
+  def test_execute_system_when_latest_does_not_support_your_ruby_but_previous_one_does
+    spec_fetcher do |fetcher|
+      fetcher.download "rubygems-update", 9 do |s|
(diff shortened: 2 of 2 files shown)


simi (approved):
Nice! Tested locally, works as intended.

```
[retro@retro  rubygems (fix-gem-update-system *%=)]❤ ruby -Ilib exe/gem --version                                                                                                             
3.2.3.dev # manually updated in rubygems.rb
[retro@retro  rubygems (fix-gem-update-system *%=)]❤ ruby -v                                                                                                                                  
ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux]                                                                                                                                
[retro@retro  rubygems (fix-gem-update-system *%=)]❤ ruby -Ilib exe/gem update --system                                                                                                       
Fetching rubygems-update-3.4.22.gem                                                                                                                                                           
Successfully installed rubygems-update-3.4.22                                                                                                                                                 
Building YARD (yri) index for rubygems-update-3.4.22...                                                                                                                                       
Done installing documentation for rubygems-update after 0 seconds                                                                                                                             
Done installing documentation for rubygems-update after 0 seconds                                                                                                                             
Installing RubyGems 3.4.22                                                                                                                                                                    
  Successfully built RubyGem                                                                                                                                                                  
  Name: bundler                                                                                                                                                                               
  Version: 2.4.22                                                                              
  File: bundler-2.4.22.gem                                                                                                                                                                    
Bundler 2.4.22 installed                                                                                                                                                                      
RubyGems 3.4.22 installed                                                                                                                                                                     
```

deivid-rodriguez:
Actually, let me double check something because I moved `check_latest_rubygems` to a method that's also used by plain `gem update`, so maybe I broke something there.

deivid-rodriguez:
I found a way simpler solution!

deivid-rodriguez:
Also there was an enhancement not related to this bug fix, so I split it to #7335.

PR #7348, opened 2024-01-02
Default to gmake for BSD and Solaris

stanhu:
The BSD `make` is not compatible with many directives in GNU make. For FreeBSD and Solaris systems, default to `gmake` so that native gem builds work without having to specify `MAKE=gmake` in the environment.

Relates to:
* https://github.com/oxidize-rb/rb-sys/issues/299
* https://github.com/redis/hiredis-rb/issues/23


jeremyevans:
This should not be done generally.  The majority of ruby gems with C extensions build fine with BSD make (as Ruby itself does).  If a particular gem does not build with BSD make, maybe there can be a way for the gem to signal that.  However, I think it is unacceptable to change the default make program on BSD and Solaris systems, just because a small number of gems use GNU-specific make syntax.

PR #7352, opened 2024-01-03
Fix ETag quoting regression in If-None-Match header of compact index request

martinemde:

HTTP spec requires that ETags in the If-None-Match header must be quoted. Requests to the rubygems.org index without quoted ETag are never matched for a 304 response, always falling back to a partial content request at best.

Fix unintentional removal of quotes around the ETag in the compact index `If-None-Match` header. Make sure that tests cover this requirement.

I also changed a method name that was being shadowed by local variables to make the code a bit more clear.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/updater.rb
+++ b/bundler/lib/bundler/compact_index_client/updater.rb
@@ -42,7 +42,7 @@ module Bundler
           else
             file.write(response.body)
           end
-          CacheFile.write(etag_path, etag(response))
+          CacheFile.write(etag_path, etag_from_response(response))
           true
         end
       end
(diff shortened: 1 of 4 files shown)


bundler/lib/bundler/compact_index_client/updater.rb
```diff
@@ -53,13 +53,13 @@ def replace(remote_path, local_path, etag_path)
...
         CacheFile.write(local_path, response.body, parse_digests(response))
-        CacheFile.write(etag_path, etag(response))
+        CacheFile.write(etag_path, etag_from_response(response))
       end
 
       def request_headers(etag, range_start = nil)
         headers = {}
         headers["Range"] = "bytes=#{range_start}-" if range_start
-        headers["If-None-Match"] = etag if etag
+        headers["If-None-Match"] = %("#{etag}") if etag && !etag.empty?
```
deivid-rodriguez, line 62: In which case the passed etag will be an empty string?

deivid-rodriguez (approved):
Looks good to me, sounds like this just need some test updates?

martinemde:
> Looks good to me, sounds like this just need some test updates?

New laptop and my tests weren't working yet so I had to rely on CI 😬  I'll get these this morning.

bundler/lib/bundler/compact_index_client/updater.rb
martinemde: Guarding against an empty file on write error 🤷‍♂️ Maybe there's no harm in sending "" and we can avoid the check in that rare case.

martinemde: I removed the empty guard. It should have no negative effect, and it should almost never happen.

PR #7356, opened 2024-01-05
Remove tip on how to disable documentation generation

adam12:

We should be encouraging more Ruby documentation, and having a very visible guide on how to disable documentation generation is counterproductive.

Removed the suggestion on how to disable documentation generation during gem install.

### DIFF
--- a/lib/rubygems/commands/setup_command.rb
+++ b/lib/rubygems/commands/setup_command.rb
@@ -222,9 +222,6 @@ By default, this RubyGems will install gem as:
         say "  gem server"
         say "and then peruse beautifully formatted documentation for your gems"
         say "with your web browser."
-        say "If you do not wish to install this documentation in the future, use the"
-        say "--no-document flag, or set it as the default in your ~/.gemrc file. See"
-        say "'gem help env' for details."
         say
       end
 
@@ -234,9 +231,6 @@ By default, this RubyGems will install gem as:
         say "  ri Classname"
         say "  ri Classname.class_method"
         say "  ri Classname#instance_method"
-        say "If you do not wish to install this documentation in the future, use the"
-        say "--no-document flag, or set it as the default in your ~/.gemrc file. See"
-        say "'gem help env' for details."
         say
       end
     end


indirect:
This message was added in response to complaints from users who don't want to spend the CPU time or disk space to generate local rdoc that they will never use.

Can you say more about why you think disabling local rdoc generation discourages documentation generally?

adam12:
Hi @indirect!

That message was added a long time ago (2011), when rdoc was slower and computers were not as powerful. During that time, it was common to see people suggesting `--no-ri --no-rdoc` as options as default values in their gemrc.

Unfortunately, I believe this further pushes developers to not care about documentation, as they expect it to never be looked at. I think this period of `--no-ri --no-rdoc` put us back as a community in general.

I've been trying to foster more documentation in Ruby since it's the number one complaint I hear from new developers about the language. To further this, I created gemdocs.org, which (currently) generates and stores in perpetuity YARD documentation.

I suspect most people are already benefiting from `--no-document` in places where they desire it most, such as their [production Docker containers](https://github.com/docker-library/ruby/blob/52e176c1dee983d991d9cff0c6371f5c50bfd636/3.3/bookworm/Dockerfile#L13-L14).

If they still want it disabled locally, they likely already have it configured or are a quick Google away.

This message is very rarely seen (I think it might be the first time I've _ever_ seen it, after all these years), but still think there is an impact.

deivid-rodriguez:
Even if generating docs is now faster, I'm not sure if documenting how to disable local rdoc and ri generation discourages people from writing more docs. I don't mind removing it but I'm not sure it's an improvement.

indirect:
It seems like gemdocs.org is actually an argument to completely remove local generation of documentation, isn't it? Why should we force every user to generate and store local rdoc that they won't use when they could simply view the documentation online instead?

adam12:
> It seems like gemdocs.org is actually an argument to completely remove local generation of documentation, isn't it?

I wouldn't say that. I think having local and remote documentation is an expectation for any modern programming language.

> Why should we force every user to generate and store local rdoc that they won't use when they could simply view the documentation online instead?

This seems like a different argument since it's already what's happening.

n-rodriguez:
IMHO it should be kept for users who don't want doc in their CI or docker images.

adam12:
> IMHO it should be kept for users who don't want doc in their CI or docker images.

As mentioned above, this is almost guaranteed to be happening automatically for users who are using official Ruby images.

See these lines in the [Ruby Dockerfile from Docker](https://github.com/docker-library/ruby/blob/52e176c1dee983d991d9cff0c6371f5c50bfd636/3.3/bookworm/Dockerfile#L13-L14) and these lines in the [Ruby Dockerfile from Ruby-lang](https://github.com/ruby/ruby-docker-images/blob/3f482ea23977e6d69a03a29e2e775631167ac46f/Dockerfile#L60-L61).

MSP-Greg:
> We should be encouraging more Ruby documentation, and having a very visible guide on how to disable documentation generation is counterproductive.

> Unfortunately, I believe this further pushes developers to not care about documentation, as they expect it to never be looked at. I think this period of --no-ri --no-rdoc put us back as a community in general.

You're assigning an awful lot of blame for a tip.  I think it's a bit more complicated. 

1. Re 'developers', I think it's more a matter of the additional time to doc code, rather than the effect of any 'tip' on their attitude re docs.  Good docs for all code is a goal, but let's take RubyGems as an example.  A very high percentage of users always use RubyGems from the command line.  Hence, normal users wouldn't benefit from complete docs on every module/class/constant/method, etc.

2. For many gems, the 'public' API is a very small part of the code.  Should everything be doc'd?  For 'users', there may not be much benefit.  For (possible) contributors, good and complete docs are very helpful.

3. Ruby has two main doc systems, YARD and RDoc & derivates.  Having two different standards isn't helpful.

JFYI, I'd prefer that all Ruby code was well doc'd.  But considering much of it is OSS, resources are limited.

I have maintained https://msp-greg.github.io/ for several years, so I do see a benefit to docs.

n-rodriguez:
> As mentioned above, this is almost guaranteed to be happening automatically for users who are using official Ruby images

I use my own Ruby images

adam12:
> I'm not sure if documenting how to disable local rdoc and ri generation discourages people from writing more doc

Instead of publishing it as a message post-setup/install, how do you feel about moving it to the `gem install` documentation page and making it a subsection of some sort?

indirect:
@adam12 right now, it seems like that will take us straight back to people who notice that it takes a long time to generate docs, and then open issues here to complain that it takes a long time to generate docs. which is why, iirc, we added that message in the first place.

adam12:
> and then open issues here to complain that it takes a long time to generate docs

@indirect I understand. In this case, perhaps the bigger picture should be to leave the existing comment here for now but look at _not_ generating documentation by default in the future, instead making it easily accessible (locally or remote) on demand.

indirect:
I love that idea. We have a bit of ongoing work in that direction, and plan to host generated documentation for every gem on rubygems.org in the future. Once we have that set up, perhaps we can default to simply fetching already-generated static doc files on demand, rather than spend so much time generating them.

adam12:
Alright, I'm going to close this then.



> we can default to simply fetching already-generated static doc files on demand

I've been musing about this for a few years now since I'm partly doing it already. But as expressed above by @MSP-Greg, we have conflicting documentation tooling already and it would be unfortunate if we ended up with a third (XKCD-927 and all that).

PR #7367, opened 2024-01-09
Add .ruby-version to .gitignore

martinemde:

The [discussion](https://github.com/rubygems/rubygems/discussions/7074) about `.ruby-version` made me realize one of the minor annoyances with this repo is that it doesn't auto-switch to the ruby I want to use for development.

Allow `.ruby-version` to exist, ignored by git, so that chruby and the like can switch automatically but not have it impact anything besides local dev.

The only reason not to do this is if we thought it would encourage some bad behavior. Personally I always switch to the same ruby for working or rubygems anyway, or use the default. This just makes it automatic if you choose to do so, and doesn't require your default to be the same as what you use for rubygems development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.rbc
 *.swp
+.ruby-version
 .DS_Store
 /.idea
 /.rdoc


hsbt (approved):
👍

segiddins (approved):

deivid-rodriguez (approved):
No issues with this either!

PR #7371, opened 2024-01-10
Update dependencies and correct rubocop failure

hsbt:

In Ruby 3.3/3.4, `bin/rubocop` show the following warning.

```
$ bin/rubocop -A
/Users/hsbt/.local/share/gem/gems/rubocop-1.52.1/lib/rubocop/formatter/html_formatter.rb:3: warning: base64 was loaded from the standard library, but is not part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of rubocop-1.52.1 to add base64 into its gemspec.
```

It fixed at recent released version of rubocop.

I updated `tool/bundler/*.lock` and run `bin/rubocop -A` each cop rule.

### DIFF
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -282,7 +282,7 @@ module Gem
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
   class BasicSpecification
-    if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
+    if /^universal\.(?<arch>.*?)-/ =~ CROSS_COMPILING || RUBY_PLATFORM
       local_platform = Platform.local
       if local_platform.cpu == "universal"
         ORIGINAL_LOCAL_PLATFORM = local_platform.to_s.freeze
--- a/bundler/lib/bundler/rubygems_ext.rb
+++ b/bundler/lib/bundler/rubygems_ext.rb
@@ -282,7 +282,7 @@ module Gem
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
   class BasicSpecification
-    if /^universal\.(?<arch>.*?)-/ =~ CROSS_COMPILING || RUBY_PLATFORM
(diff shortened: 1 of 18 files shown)


bundler/lib/bundler/rubygems_ext.rb
```diff
@@ -282,7 +282,7 @@ def match_platforms?(platform, platforms)
 
   # On universal Rubies, resolve the "universal" arch to the real CPU arch, without changing the extension directory.
   class BasicSpecification
-    if /^universal\.(?<arch>.*?)-/ =~ (CROSS_COMPILING || RUBY_PLATFORM)
+    if /^universal\.(?<arch>.*?)-/ =~ CROSS_COMPILING || RUBY_PLATFORM
```
koic, line 285: This is a false positive in RuboCop. I've opened PR https://github.com/rubocop/rubocop/pull/12607, but until that is resolved, it'll need to prevent false detections by using `# rubocop:disable Style/RedundantParentheses`.

hsbt: Thanks for your input!

lib/rubygems/commands/rdoc_command.rb
```diff
@@ -87,7 +87,7 @@ def execute
       begin
         doc.generate
       rescue Errno::ENOENT => e
-        match = / - /.match(e.message)
+        match = e.message.include?(" - ")
         alert_error "Unable to document #{spec.full_name}, " \
                     " #{match.post_match} is missing, skipping"
```
segiddins, line 92: I think this will now raise a no method error

deivid-rodriguez: Dealing with this in #7392.

koic:
This is a piece of trivia and can be remembered in the same way as `git branch -d` and `git branch -D` :-)
https://github.com/rubocop/rubocop/issues/8156#issuecomment-644545129

deivid-rodriguez:
Makes sense, I guess the meaning was changed in v1, and that's why I though `-a` corrected everything?

koic:
I see. The concept of safety was not present initially, and there was only `-a`, which indeed is what `-A` is now 😅

PR #7375, opened 2024-01-10
Refactor setting ENV and ruby flags when shelling out during specs

deivid-rodriguez:

Setting ENV and ruby flags when shelling out during specs is quite messy.

Use `ENV` consistently. This way it's more clear to understand what's going on and since we don't use any ruby CLI flags, we can test restarts more reliably without the workaround we were using previously.


deivid-rodriguez:
Finally got this one green!

Issue #7369, opened 2024-01-09
Bundler unable to lock new gem version even after publication


benlangfeld:
### Describe the problem as clearly as you can

With a specific gem (rails) locked to version 7.0.8, I am unable to upgrade to 7.1.2 despite this being available on rubygems.org. Doing `gem install rails --version 7.1.2` works.

### Did you try upgrading rubygems & bundler?

Yes, latest versions.

Bundler: 2.5.4
Rubygems: 3.5.4

### Post steps to reproduce the problem

https://gist.github.com/benlangfeld/0cc8fa8b0e441cbd6427cf1f1bc35832

### Which command did you run?

`bundler lock --patch --strict --update rails`

Leaving off `--patch` works.

### What were you expecting to happen?

The `Gemfile.lock` should be updated with the new version.

### What actually happened?

```
> bundler lock --patch --strict --update rails:
Fetching gem metadata from https://rubygems.org/..........
Could not find gems matching 'rails (= 7.1.2)' valid for all resolution
platforms (aarch64-linux, x86_64-linux) in rubygems repository
https://rubygems.org/ or installed locally.

The source contains the following gems matching 'rails (= 7.1.2)':
  * rails-7.1.2
```

### If not included with the output of your command, run `bundle env` and paste the output below

https://gist.github.com/benlangfeld/3c608ad0ba53c934194477491f9f6b9e

deivid-rodriguez:
Yeah, `bundle lock --update --patch` is not really prepared to be run on a situation where the Gemfile has been edited with a requirement incompatible with the lockfile. I'll try to fix it!

benlangfeld:
> Yeah, `bundle lock --update --patch` is not really prepared to be run on a situation where the Gemfile has been edited with a requirement incompatible with the lockfile. I'll try to fix it!

Just in case anyone is curious, this [is in use by Renovate](https://github.com/renovatebot/renovate/blob/4fc7554db9cda626f5e566140b082df10d1f2b81/lib/modules/manager/bundler/artifacts.ts#L100) which is how I got to this point and why I can't simply remove that parameter for my use case. From the docs I couldn't find any mention of it being an inappropriate combination of options for this case, though. Making it work would be awesome, but if that's not possible, maybe a more specific error would be useful.

deivid-rodriguez:
[...]

PR #7376, opened 2024-01-10
(Wrong) fix for `bundle lock` edge case

deivid-rodriguez:

The problem is #7369. However, after further thought I don't think this is a good solution and we should only aim to provide a better error message here.

Bring lockfile up to date when `bundle lock --update foo --patch --strict` is given with an out of date lockfile.

I don't think we want to do that, really.


deivid-rodriguez:
I'm replacing this wrong approach with #7527.

PR #7379, opened 2024-01-11
Fix require issue

deivid-rodriguez:

TODO.

TODO.


lib/rubygems/core_ext/kernel_require.rb
```diff
@@ -95,7 +95,7 @@ def require(path) # :doc:
       if found_specs.empty?
         found_specs = Gem::Specification.find_in_unresolved_tree path
 
-        found_specs.each(&:activate)
+        found_specs.each(&:activate) unless default_spec
```
stanhu, line 98: Can we skip line 96 if `default_spec` is present then?

deivid-rodriguez: Good catch!

deivid-rodriguez: I pushed a simpler patch. Once we activate the default gem, we can skip the rest of the stuff!

lib/rubygems/core_ext/kernel_require.rb
```diff
@@ -59,6 +62,8 @@ def require(path) # :doc:
 
         Kernel.send(:gem, spec.name, Gem::Requirement.default_prerelease) unless
           resolved_path
+
+        next
```
stanhu, line 66: This change didn't seem to fix the issue for me. Doesn't this `next` just break out of the `RUBYGEMS_ACTIVATION_MONITOR.synchronize` block and continue as before?

deivid-rodriguez: Yes, that's what it does!

Ooopps! Since I got the same error to reproduce with a spec and this change fixed it, I assumed it was also fixing your case. But your real world scenario seems to have even more complexity, I'll look more tomorrow.

stanhu: Thanks, that did indeed solve it, though should `find_unresolved_default_spec` be renamed to `find_default_spec`? 😄

deivid-rodriguez: Good point, once again! I'll turn on paranoid mode and add `Gem.find_default_spec` and then deprecate `Gem.find_unresolved_default_spec` in a follow up PR.

stanhu: Can `next` be dropped?

deivid-rodriguez: Which `next`? I think both fixes here are necessary.

stanhu: The tests still pass if I apply this diff to this branch:

```diff
diff --git a/lib/rubygems/core_ext/kernel_require.rb b/lib/rubygems/core_ext/kernel_require.rb
index 073966b696..ac652e5ed6 100644
--- a/lib/rubygems/core_ext/kernel_require.rb
+++ b/lib/rubygems/core_ext/kernel_require.rb
@@ -66,8 +66,6 @@ def require(path) # :doc:
 
         Kernel.send(:gem, name, Gem::Requirement.default_prerelease) unless
           resolved_path
-
-        next
       end
 
       # If there are no unresolved deps, then we can use just try
```

deivid-rodriguez: Both tests added here fail for me without that line 🤔. In any case, even if tests passed, seems nice to save unnecessary work on first require of a default gem.

stanhu: @deivid-rodriguez I might be doing something wrong, but I'm seeing the same results with or without the `next` line (see below). I'm not sure whether `0 failures, 2 errors` means pass, but I think this is a test failure?

<details>

```
% bundle exec ruby test/rubygems/test_require.rb
Loaded suite test/rubygems/test_require
Started
http://example.com
.http://example.com
http://example.com
http://example.com
.http://example.com
http://example.com
http://example.com
http://example.com
.http://example.com
http://example.com
..http://example.com
http://example.com
http://example.com
http://example.com
.http://example.com
http://example.com
.http://example.com
..http://example.com
..http://example.com
http://example.com
http://example.com
http://example.com
E
==================================================================================================================================================================
Error: test_default_gem_required_circulary_with_unresolved_gems_depending_on_it(TestGemRequire): Gem::LoadError: can't activate net-http-0.1.1, already activated net-http-0.3.0
/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/specification.rb:2307:in `check_version_conflict'
/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/specification.rb:1446:in `activate'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:98:in `each'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:98:in `block in require'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:39:in `synchronize'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:39:in `require'
test/rubygems/test_require.rb:29:in `assert_require'
test/rubygems/test_require.rb:574:in `block in test_default_gem_required_circulary_with_unresolved_gems_depending_on_it'
/Users/stanhu/github/rubygems/test/rubygems/helper.rb:1253:in `silence_warnings'
test/rubygems/test_require.rb:573:in `test_default_gem_required_circulary_with_unresolved_gems_depending_on_it'
==================================================================================================================================================================
http://example.com
http://example.com
http://example.com
http://example.com
E
==================================================================================================================================================================
Error: test_default_gem_with_unresolved_gems_depending_on_it(TestGemRequire): Gem::LoadError: can't activate net-http-0.1.1, already activated net-http-0.3.0
/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/specification.rb:2307:in `check_version_conflict'
/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/specification.rb:1446:in `activate'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:98:in `each'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:98:in `block in require'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:39:in `synchronize'
<internal:/Users/stanhu/.asdf/installs/ruby/3.2.2/lib/ruby/site_ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:39:in `require'
test/rubygems/test_require.rb:29:in `assert_require'
test/rubygems/test_require.rb:549:in `test_default_gem_with_unresolved_gems_depending_on_it'
==================================================================================================================================================================
.http://example.com
http://example.com
http://example.com
http://example.com
http://example.com
http://example.com
......http://example.com
..http://example.com
.http://example.com
http://example.com
.http://example.com
.http://example.com
.http://example.com
.http://example.com
.http://example.com
http://example.com
http://example.com
.http://example.com
http://example.com
http://example.com
.http://example.com
http://example.com
....http://example.com
..http://example.com
http://example.com
http://example.com
http://example.com
http://example.com
.
Finished in 2.115785 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------
37 tests, 115 assertions, 0 failures, 2 errors, 0 pendings, 0 omissions, 0 notifications
94.5946% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------
17.49 tests/s, 54.35 assertions/s
</details>

Comment (deivid-rodriguez) [reply lib/rubygems/core_ext/kernel_require.rb:70]: It is! I'm not sure what's causing that but since current implementation passes CI (and locally for me) and fixes your realworld issue, I'd suggest to deal with issues in your testing environment as a separate issue?

File: test/rubygems/test_require.rb [hunk 3]
```diff
@@ -529,6 +529,52 @@ def test_default_gem_prerelease
...
+
+    faraday_2 = util_spec "faraday", "2", { "net-http" => ">= 0" }
+    install_gem faraday_2
+
+    chef = util_spec "chef", "1", { "faraday" => [">= 1", "< 3"] }, "lib/chef.rb"
+    install_gem chef
+
+    assert_require "chef"
+
+    # Ignore circular require warnings for clean test output
```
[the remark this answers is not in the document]
deivid-rodriguez: Great idea! I tweaked the test to hopefully catch that.

lib/rubygems/core_ext/kernel_require.rb
deivid-rodriguez: Actually, from looking at your test run output, I think the answer may be that tests are running against your system RubyGems, and you may not have manually applied the patch to that?

deivid-rodriguez:
The issue I was running into when releasing gems was actually not fixed by https://github.com/rubygems/rubygems/pull/7384, and I now think it's actually the same issue being fixed here.

So planning to include this PR with the next release to unblock that.

test/rubygems/test_require.rb
```diff
@@ -529,6 +529,65 @@ def test_default_gem_prerelease
     assert_equal %w[default-3.0.0.rc2], loaded_spec_names
   end
 
+  def test_default_gem_with_unresolved_gems_depending_on_it
```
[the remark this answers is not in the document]
deivid-rodriguez: Can you show a real example? My understanding is that non default gems cannot shadow default gems (unless they are put in front of the `$LOAD_PATH` manually), and default gems take priority?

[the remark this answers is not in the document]
deivid-rodriguez: Oh, maybe you're precisely suggesting that with this change the default gem may start not getting activated, in favor of the non default one?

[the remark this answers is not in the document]
deivid-rodriguez: Oh, or maybe you mean when a higher non default version of the same gem needs to shadow the default version of that gem. In that case, we already have tests for that. See https://github.com/rubygems/rubygems/blob/271c9226589baacebfa18da939c05b4ae4f8a434/test/rubygems/test_require.rb#L481-L491.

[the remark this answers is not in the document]
segiddins: > Oh, maybe you're precisely suggesting that with this change the default gem may start not getting activated, in favor of the non default one?

potentially?

> My understanding is that non default gems cannot shadow default gems

do we have tests covering that behavior? if not, im worried it might have changed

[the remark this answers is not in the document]
deivid-rodriguez: Default gems are already in the default LOAD_PATH while normal gems are not and require explicit activation. I think that's normally done by rescuing `LoadError` from standard `require`. I think this PR should not regress that.

I'm adding a new test to make sure 👍.

hsbt:
This PR is broken with ruby/ruby CI and https://rubyci.org/.

* https://github.com/ruby/ruby/actions/runs/7578672002/job/20641640821
* https://github.com/ruby/ruby/actions/runs/7578672002/job/20641641212
* https://github.com/ruby/ruby/actions/runs/7578672002/job/20641642031
* https://rubyci.s3.amazonaws.com/freebsd14/ruby-master/log/20240119T023002Z.fail.html.gz

deivid-rodriguez:
Interesting. RubyGems should be resetting `$LOADED_FEATURES` after each test, so I'm not sure why that's happening. I can take a closer look when I get time but for now I think it's fine to skip that test given it does not show a regression since it's a newly added test.

stanhu:
I can reproduce the issue if I check out that commit, build `ruby`, and run:

```shell
./miniruby -I./lib -I. -I.ext/common  ./tool/runruby.rb --extout=.ext  -- --disable-gems ./test/runner.rb test/net/http/test_httpheader.rb test/rubygems/test_require.rb
```

It doesn't fail if `test/rubygems/test_require.rb` is run by itself.

stanhu:
@deivid-rodriguez Maybe https://github.com/rubygems/rubygems/pull/7405 will work here?

PR #7395, opened 2024-01-17
Fix rake runtime dependency warning for rake based extension

ntkme:

`rake` should be explicitly listed as runtime dependency for `Rakefile` based extensions, as it's not guaranteed to be available. For example, when using `bundle` with non-default `BUNDLE_PATH`.

Currently `gem` command prints a warning about this, however, the warning is not printed when `rake` is added as a development dependency, which is not sufficient as it has to be a runtime dependency.

Also see discussions here: https://github.com/protocolbuffers/protobuf/pull/15203#issuecomment-1897134421   

Print the warning unless rake is added as runtime dependency.

### DIFF
--- a/lib/rubygems/specification_policy.rb
+++ b/lib/rubygems/specification_policy.rb
@@ -497,10 +497,10 @@ You have specified rust based extension, but Cargo.lock is not part of the gem f
 
   def validate_rake_extensions(builder) # :nodoc:
     rake_extension = @specification.extensions.any? {|s| builder.builder_for(s) == Gem::Ext::RakeBuilder }
-    rake_dependency = @specification.dependencies.any? {|d| d.name == "rake" }
+    rake_dependency = @specification.dependencies.any? {|d| d.name == "rake" && d.type == :runtime }
 
     warning <<-WARNING if rake_extension && !rake_dependency
-You have specified rake based extension, but rake is not added as dependency. It is recommended to add rake as a dependency in gemspec since there's no guarantee rake will be already installed.
+You have specified rake based extension, but rake is not added as runtime dependency. It is recommended to add rake as a runtime dependency in gemspec since there's no guarantee rake will be already installed.
     WARNING
   end
 
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -2727,7 +2727,7 @@ duplicate dependency on c (>= 1.2.3, development), (~> 1.2) use:
         @a1.validate
       end
 
-      assert_match(/add rake as a dependency/, @ui.error)
+      assert_match(/add rake as a runtime dependency/, @ui.error)
     end
   end
 
@@ -2743,7 +2743,7 @@ duplicate dependency on c (>= 1.2.3, development), (~> 1.2) use:
         @a1.validate
       end
 
-      refute_match(/add rake as a dependency/, @ui.error)
+      refute_match(/add rake as a runtime dependency/, @ui.error)
     end
   end
 


JasonLunn:
For the Rubygems maintainers:

Is it really the case that Rubygems expects `rake` to be a **_runtime_** dependency for every gem that uses a `Rakefile` to build native extensions? The existence of [`Gem::Ext::RakeBuilder`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/ext/rake_builder.rb) suggests that native extension building using Rake has first class support, but I think most end users would be surprised to find that they had to include `rake` in a `vendor`'ed production deployment just because Rake was needed during the compilation phase of initial gem installation.

If the assertion change in this PR does indeed align with Rubygems' intentions, then I recommend that documentation changes make this explicit in the rubygems/guides as a follow up.

deivid-rodriguez:
Yeah, rubygems does currently expect `rake` to be a runtime dependency in this situation. We have no concept of "build dependency" so the only way of ensuring it's actually there when the gem gets installed is making it a runtime dependency.

We could drop this requirement if we ever add a concept of "build dependencies" that get installed to support building extensions but don't affect applications' runtime.

This PR makes sense to me by the way!

simi:
@deivid-rodriguez is it good idea to rely on fact `rake` is going to be installed before gem itself?

deivid-rodriguez:
We've tried to enforce that and special cased rake for a long time. See https://github.com/rubygems/rubygems/commit/597d620a08f87c499dbd427c6d87bb80b1d1209d, https://github.com/rubygems/rubygems/commit/d962a980f00cab657b1fafbe9f4b12f13d90561c, or https://github.com/rubygems/rubygems/commit/dbfff84bf44f4536cdd0051ca7a7032eb7ad9adf. I think it works fine.

What are you thinking, any ideas for improvement?

JasonLunn:
The ideal would be a way to express a build dependency that is distinct from a runtime dependency.

Currently, users think of development dependencies as "the dependencies needed to build the gem from source", even though it includes the larger set of dependencies like debuggers that one would want for actual development, not just compilation.

Given that the introduction of a build dependency concept in Gemspec is a heavy lift, in the short term, it would be great to to have a `bundle` equivalent of `gem install --dev`, so that the tens of [thousands of places](https://github.com/search?q=%22add_development_dependency+%5C%22rake%22&type=code) where Gems specify `rake` as a development dependency will just work.

deivid-rodriguez:
What would `bundle install --dev` do? Install all development dependencies of all dependencies? That seems overkill and hardly useful to me 🤔. Rakefile based extensions are also quite rare, but if you get an error while building one of them because it does not declare `rake` as a runtime dependency, and you don't depend on `rake` either (nor directly nor through other dependency), then I think adding `rake` to your Gemfile is a good enough workaround.

If we want to fix this right, we should allow gem authors to declare "build only dependencies", and let those be installed before each application dependency if not already available.

JasonLunn:
> What would `bundle install --dev` do? Install all development dependencies of all dependencies?

Yes. Or perhaps only the development dependencies of any gem without a platform and with platform `ruby`?

> That seems overkill and hardly useful to me 🤔. 

I understand that the utility is limited, as the majority of users will have rake installed anyway and most user won't override their installation path, but both choices are reasonable independently and in combination with one another.

> Rakefile based extensions are also quite rare, but if you get an error while building one of them because it does not declare `rake` as a runtime dependency, and you don't already depend on `rake` either (nor directly nor through other dependency), then I think adding `rake` to your Gemfile is a good enough workaround.

There are at least [hundreds](https://github.com/search?q=path%3Aext%2F*%2FRakefile&type=code) of gems that have Rakefiles in their ext directories. I have suggested adding `rake` to the top level `Gemfile` as well, but that would still result in users ending up with the `rake` gem in their production deployments unnecessarily, right?

> If we want to fix this right, we should allow gem authors to declare "build only dependencies", and let those be installed before each application dependency if not already available.

Agreed. Do you think that's a feature that Rubygems would accept?

deivid-rodriguez:
Yeah, I think if we decided to consider something like this, it'd need be to be a carefully thought out plan to preserve backwards compatibility. I guess we could make it so that only gems using the new feature would also need to specify `required_rubygems_version` to be at least the version that introduces the feature, so that older rubygems never try to install those gems. But I bet there will be many edges.

I think the impact is very low for such a hassle to be honest. Only a minority of gems using Rakefile extensions, and `rake` is so common anyways. Also the consequences of this problem seem minor to me (extra cached gem in `vendor`, extra entry in your `$LOAD_PATH, etc). That said, if someone wants to elaborate a plan and write an RFC, I think it's still an interesting investigation for any future extension of the gemspec API, regardless of whether it's accepted or not.

ntkme:
> I think the impact is very low for such a hassle to be honest.

Totally agree. I think the amendment from this PR is the right thing to do for now, in order to align with the current expectations.

Other things are out of scope here, but maybe up for another separate discussion.

simi:
> There are at least [hundreds](https://github.com/search?q=path%3Aext%2F*%2FRakefile&type=code) of gems that have Rakefiles in their ext directories.

I did quick manual check of top 10 entries, and there was no "classic RubyGem" related to this problem in the result. It was often related to mruby. Doing raw lookup using [gem-codesearch](https://github.com/akr/gem-codesearch) (thanks @hsbt) there is cca 145 individual gem versions (out of current total of 1591877 versions) using `rake` as a extension builder (aka https://github.com/rubygems/rubygems/blob/64eab6ad3ec316fd41e3962ed6bf37545f67af7c/lib/rubygems/ext/rake_builder.rb).

deivid-rodriguez (approved):

simi (approved):

deivid-rodriguez:
Thanks @ntkme!

simi:
@deivid-rodriguez here is list of gems released in last 2 years with Rake extension and not having rake as runtime dependency.

```
libmspack, https://gitlab.com/davispuh/ruby-libmspack
rusty_json_schema, https://github.com/driv3r/rusty_json_schema
vips, https://github.com/ioquatix/vips
unnatural, https://github.com/bjmllr/unnatural
geohash_int, https://github.com/TheoremOne/ruby-geohash_int
blake2b_rs, https://github.com/wuminzhe/blake2b_rs
schnorrkel_rb, https://github.com/wuminzhe/schnorrkel_rb
ffi-hydrogen, https://github.com/minond/ffi-hydrogen
red-plasma, https://arrow.apache.org/
temporalio, https://github.com/temporalio/sdk-ruby
red-datafusion, https://github.com/datafusion-contrib/datafusion-ruby
ko, https://github.com/zhulik/ko
webp-ffi, http://leopard.in.ua/webp-ffi
adwaita, https://ruby-gnome.github.io/
atk, https://ruby-gnome2.osdn.jp/
clutter, https://ruby-gnome2.osdn.jp/
clutter-gstreamer, https://ruby-gnome2.osdn.jp/
clutter-gtk, https://ruby-gnome2.osdn.jp/
gdk3, https://ruby-gnome2.osdn.jp/
gdk4, https://ruby-gnome2.osdn.jp/
gdk_pixbuf2, https://ruby-gnome2.osdn.jp/
gegl, https://ruby-gnome2.osdn.jp/
gnumeric, https://ruby-gnome2.osdn.jp/
goffice, https://ruby-gnome2.osdn.jp/
gsf, https://ruby-gnome2.osdn.jp/
gtksourceview3, https://ruby-gnome2.osdn.jp/
gtksourceview4, https://ruby-gnome2.osdn.jp/
gtksourceview5, https://ruby-gnome.github.io/
libhandy, https://ruby-gnome.github.io/
libsecret, https://ruby-gnome2.osdn.jp/
poppler, https://ruby-gnome2.osdn.jp/
rsvg2, https://ruby-gnome2.osdn.jp/
vte3, https://ruby-gnome2.github.io/
vte4, https://ruby-gnome.github.io/
webkit-gtk, https://ruby-gnome2.osdn.jp/
webkit2-gtk, https://ruby-gnome2.osdn.jp/
wnck3, https://ruby-gnome2.osdn.jp/
chef, https://www.chef.io
cslrb, https://github.com/saschaknobloch/cslrb
xot, https://github.com/xord/xot
red-arrow-cuda, https://arrow.apache.org/
red-arrow-dataset, https://arrow.apache.org/
red-arrow-flight, https://arrow.apache.org/
red-arrow-flight, https://arrow.apache.org/
red-gandiva, https://arrow.apache.org/
red-parquet, https://arrow.apache.org/
eyes_universal, https://www.applitools.com
whichlang, https://github.com/bendangelo/whichlang-rb
beeps, https://github.com/xord/beeps
rays, https://github.com/xord/rays
reflexion, https://github.com/xord/reflex
rucy, https://github.com/xord/rucy
whatlang, https://gitlab.com/KitaitiMakoto/whatlang-rb
```

Do you think it is worth to try to ping them and fix this problem?

deivid-rodriguez:
Hopefully the warning or their users will ping them for us?

One problem I think should be in my opinion more worth looking at is that Bundler's gem authoring tasks, if I recall correctly, silence `gem build` warnings, so these warnings may go unnoticed to some gem authors.

I believe we should audit all `gem build` warnings, and let them be exposed by Bundler's `rake build` task. If we build some of them are going to get too noisy (say, open ended dependency warnings), that problaby means we should also remove those from `gem build`.

PR #7405, opened 2024-01-19
Make tests play with upstream Ruby tests

stanhu:

CI broke in https://github.com/ruby/ruby/pull/9604 because if any Ruby tests run `require 'net/http'`, they will pollute the `$LOADED_FEATURES` for the RubyGems tests. We can fix this by renaming the test default gem from `net-http` to `my-http`. 

See https://github.com/rubygems/rubygems/pull/7379#issuecomment-1901241299 for more details.


deivid-rodriguez (commented):
Thanks for figuring this out @stanhu!

PR #7413, opened 2024-01-23
Skip to load commented out words

### LINKED ISSUE #7393: YAML comments not respected when loading gemrc

RubyGems v3.4.10 parses gemrc contents in a way that it strips away the part of the value after `#` (YAML comment syntax).

RubyGems v3.5.4 doesn't strip values after `#` and loads them verbatim.

Most likely a regression due to https://github.com/rubygems/rubygems/pull/6615 /cc @hsbt 

Yes, but issue still persists in RubyGems 3.5.4.

Notice that `--no-install` is behind YAML comment:
```
$ cat ~/.gemrc
gem: --no-rdoc --no-ri # --user-install
```

`gem install lolcat`

I was expecting that the `lolcat` executable gets placed in EXECUTABLE DIRECTORY as reported by `gem env`.

### What actually happened?

The `lolcat` executable got placed USER INSTALLATION DIRECTORY even though `--user-install` was commented out from gemrc.

```
$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.4.10
  - ...
  - GEM CONFIGURATION:
     - ...
     - "gem" => "--no-rdoc --no-ri"
 
$ gem update --system

$ gem env
RubyGems Environment:
  - RUBYGEMS VERSION: 3.5.4
  - ...
  - GEM CONFIGURATION:
     - ...
     - "gem" => "--no-rdoc --no-ri # --user-install"
```

hsbt:

Fixes #7393

The current YAML serializer couldn't handle commented out words.

Skip to load after `#` symbol.


williantenfen:
@hsbt sorry If I'm getting on your way... I'm just trying to contribute with something.. 

does it make sense to use a regex for this? or update the current ones?

I think I manage to pass some specs with something:

```ruby
REMOVE_COMMENT_REGEX = /
      ^
      (["']?)  # optional opening quote
      (.*?)  # match for characters between quotes
      \1     # matching closing quote
      \s*    # Optional whitespace characters after the closing quote
      (?:\#(.*))?  # optional comment starting with '#'
      $
    /xo

  ...
  # and then call it when needed:


         if val == "[]" # empty array
              val = []
            else
              val = val.match(REMOVE_COMMENT_REGEX)[2]
            end
```

hsbt:
@williantenfen Thanks for your support! I only focused configuration file for rubygems. If you interest to improve this PR, feel free to submit your idea and changes 👍

bundler/lib/bundler/yaml_serializer.rb
```diff
@@ -80,6 +84,14 @@ def load(str)
       res
     end
 
+    def strip_comment(val)
+      if val.include?("#") && !val.start_with?("#")
```
segiddins, line 88: given this is operating on a new string, we can avoid the array & string allocations by using `val.sub!(/#.*/, "")`

PR #7417, opened 2024-01-24
Revert "Ensure File.open applies default umask on gem extract"

hsbt:

@martinemde

388fb83e339143feca66b2f2f8733396890c379e break CI of `ruby/ruby` without GitHub Actions.

* https://app.travis-ci.com/github/ruby/ruby/jobs/616536410#L2983
* http://ci.rvm.jp/logfiles/brlog.trunk.20240124-050253#L3470

https://github.com/rubygems/rubygems/pull/7300 is merged without nobu's opinion. I don't care that. We should revert it.

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -448,15 +448,13 @@ EOM
           end
 
         unless directories.include?(mkdir)
-          mkdir_mode = 0o755 if dir_mode
-          mkdir_mode ||= entry.header.mode if entry.directory?
-          mkdir_mode &= ~File.umask if mkdir_mode
-          FileUtils.mkdir_p mkdir, mode: mkdir_mode
+          FileUtils.mkdir_p mkdir, mode: dir_mode ? 0o755 : (entry.header.mode if entry.directory?)
           directories << mkdir
         end
 
         if entry.file?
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
The PR that's being reverted here has never been released (we were waiting on @nobu's feedback). So better to not tag this one for backport either since it will generate conflicts.

hsbt:
Thanks. I misunderstood to release that commits for RG 3.5.

martinemde:
Thanks for solving this @hsbt. It was unexpected that it would break things upstream.

deivid-rodriguez:
Can we clarify whether there was something wrong with the original PR? @nobu are you ok with #7300?

If the only problem was some test failures in ruby/ruby, should we work on addressing them instead? @hsbt Any easy way to reproduce the original problem? My understanding is that the issue was not reproducible through a standard GitHub PR but happened only on ruby/ruby's CI infrastructure that runs after commits are merged to master, correct?

hsbt:
>Any easy way to reproduce the original problem?

I didn't investigate that. It's hard to reproduce in local because Travis and ci.rvm.jp are specific environment.

deivid-rodriguez:
No problem. Did the new fix work fine this time? :crossed_fingers:

PR #7422, opened 2024-01-25
Change gem login message to clear up that username can be also used

VitaliySerov:

Right now login question looks like this:
```
Don't have an account yet? Create one at https://rubygems.org/sign_up
   Email:   VitaliySerov
Password:  
```

But according to my try it also support login via username, not only via email

Change wording

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -130,12 +130,12 @@ module Gem::GemcutterUtilities
 
     say "The existing key doesn't have access of #{scope} on #{pretty_host}. Please sign in to update access."
 
-    email    = ask "Username/email: "
-    password = ask_for_password "      Password: "
+    identifier = ask "Username/email: "
+    password   = ask_for_password "      Password: "
 
     response = rubygems_api_request(:put, "api/v1/api_key",
                                     sign_in_host, scope: scope) do |request|
-      request.basic_auth email, password
+      request.basic_auth identifier, password
       request["OTP"] = otp if otp
       request.body = Gem::URI.encode_www_form({ api_key: api_key }.merge(update_scope_params))
     end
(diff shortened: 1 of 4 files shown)


lib/rubygems/gemcutter_utilities.rb
```diff
@@ -129,8 +129,8 @@ def update_scope(scope)
 
     say "The existing key doesn't have access of #{scope} on #{pretty_host}. Please sign in to update access."
 
-    email    = ask "   Email: "
-    password = ask_for_password "Password: "
+    email    = ask "Username/email: "
```
deivid-rodriguez, line 132: Should we name the variable `id` or something then?

deivid-rodriguez (commented):
Thanks for improving this. I didn't know username was good here too and that's usually a few characters saved!

lib/rubygems/gemcutter_utilities.rb
VitaliySerov: @deivid-rodriguez Quite frankly not sue, because for me 'id' is also not clear enough

But I totally agree that current variant is super bulky and not elegant

Maybe we can find out some more synonims :)

How about `account`?

williantenfen: maybe `email_or_username` ... ?

deivid-rodriguez: Thank you. I'm good with either alternative proposed, also `identifier` if that's more clear? Change it to your liking and I'm happy to approve!

VitaliySerov:
I will fix the failing tests, as I expected there is places that depends on wording

VitaliySerov:
As I understand test suite in `ruby` repo also expect old wording

https://github.com/ruby/ruby/blob/master/test/rubygems/test_gem_commands_yank_command.rb#L294

It's hard coded:

```ruby
    access_notice = "The existing key doesn't have access of yank_rubygem on http://example. Please sign in to update access."
    assert_match access_notice, @ui.output
    assert_match "Email:", @ui.output
    assert_match "Password:", @ui.output
```

Not sure what flow I should follow. Also create a PR in `ruby` and sync those two somehow?

deivid-rodriguez:
It's fine, changes will be backported automatically to ruby/ruby once they land here.

Regarding the Windows failures, it seems they are caused by recent changes in https://github.com/msys2/MINGW-packages, and I would expect them to get sorted out eventually.

VitaliySerov:
@deivid-rodriguez so everything seems in order regarding my changes? I don't think I can do anything more

deivid-rodriguez:
Yes, I'll merge shortly, thanks for this!

deivid-rodriguez (approved):

Issue #7427, opened 2024-01-27
release requirement loads prerelease


kimesf:
### Describe the problem as clearly as you can

It seems a release requirement is loading a prerelease. I need help understanding if this is expected behavior.

### Did you try upgrading rubygems & bundler?

After running `gem update --system` and `bundle update --bundler`, I bundled with latest version.

### Post steps to reproduce the problem

Run `bundle install` for this setup:

```ruby
source 'https://rubygems.org'

git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '3.3.0'

gem 'rails', '~> 7.1.2'
gem 'validates_timeliness', '>= 6.0.1'
```

Gem 'validates_timeliness' resolves to `7.0.0.beta2`.

### Which command did you run?

`bundle install`

### What were you expecting to happen?

I guess the expected behavior was for Bundler to be unable to solve dependencies.

### What actually happened?

Gem 'validates_timeliness' resolved to `7.0.0.beta2`.

Because of Rails and it's dependencies, validates_timeliness can't resolve to 6.0.1. The next available version for it is 7.0.0.beta1 and 7.0.0.beta2.

In this case, is it expected for it to solve to a prerelease version? If that's the case, I misunderstood [this test](https://github.com/rubygems/rubygems/blob/0a69ce21ca3e8bfbc3ff7a6cec6d36ba2c434128/test/rubygems/test_kernel.rb#L55).

## Environment

Ruby version is being managed with [asdf](https://asdf-vm.com/).

```
Bundler       2.5.5
  Platforms   ruby, x86_64-linux
Ruby          3.3.0p0 (2023-12-25 revision 5124f9ac7513eb590c37717337c430cb93caa151) [x86_64-linux]
RubyGems      3.5.5
Tools
  Git         2.43.0
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

### Gemfile

```ruby
source 'https://rubygems.org'

git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '3.3.0'

gem 'rails', '~> 7.1.2'
gem 'validates_timeliness', '>= 6.0.1'
```

### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (7.1.3)
      actionpack (= 7.1.3)
      activesupport (= 7.1.3)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
      zeitwerk (~> 2.6)
[...]

PR #1755, opened 2016-10-20
Ensure `to_spec` falls back on prerelease specs

indirect:
The `Gem::Dependency#to_spec` method is documented to never return nil, and raise `Gem::MissingSpecError` if there are no gems that meet the dependency. However, if the only gem that meets the dependency is a prerelease gem, then `to_spec` would return `nil` instead of raising.

I've written a test that explicitly checks this case, and modified `to_spec` to fall back on the first prerelease spec if there are no full-release specs available.

PR #7428, opened 2024-01-27
fix Gem::Dependency#to_spec returning nil when prerelease is the only…

kimesf:

Gem `ruby-lsp` calls `Gem::Dependency#to_spec` expecting it can't return `nil`. That caused the lsp to break because sometimes `nil` is actually being returned.

Some PRs were/are open to handle the issue on their side ([see ruby-lsp PR for more details](https://github.com/Shopify/ruby-lsp/pull/1326)), but that prompt me to investigate if there was a problem in the Bundler.

A [comment](https://github.com/Shopify/ruby-lsp/issues/1141#issuecomment-1781812598) there pointed me to [this line of code](https://github.com/rubygems/rubygems/blob/0a69ce21ca3e8bfbc3ff7a6cec6d36ba2c434128/lib/rubygems/dependency.rb#L333).

I found PR #1755 with a similar problem in the past and that stated `#to_spec` must not return `nil`.

The issue with the lsp happened for this bundle setup:

```ruby
source 'https://rubygems.org'

git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '3.3.0'

gem 'rails', '~> 7.1.2'
gem 'validates_timeliness', '>= 6.0.1'
```

Calling `#to_spec` for `<Bundler::Dependency type=:runtime name="validates_timeliness" requirements=">= 6.0.1">`  returns `nil` (see `ruby-lsp` PR linked above for reproducing steps).

Because of Rails dependencies, validates_timeliness can't resolve to 6.0.1. The next available versions for it are 7.0.0.beta1 and 7.0.0.beta2, only prereleases.

This is - to my understanding (be aware this is my first contribution here) - where the problem happens:

```ruby
  def to_spec
    matches = to_specs.compact

    active = matches.find(&:activated?)
    return active if active

    unless prerelease?
      # Move prereleases to the end of the list for >= 0 requirements
      pre, matches = matches.partition {|spec| spec.version.prerelease? }
      matches += pre if requirement == Gem::Requirement.default
    end

    matches.first
  end
```

This moves all matches (all prereleases) to variable `pre`:

```ruby
pre, matches = matches.partition {|spec| spec.version.prerelease? }
```

Because there is a requirement, this prevents adding them to `matches` again, leaving matches empty:
```ruby
matches += pre if requirement == Gem::Requirement.default
```

I honestly don't quite get why we want only change the order `if requirement == Gem::Requirement.default`.

The solution was to simply add `matches.empty?` to the condition.

```ruby
matches += pre if requirement == Gem::Requirement.default || matches.empty?
```

## Failing Test

The fix I tried broke `/test/rubygems/test_kernel.rb:57`.

```ruby
def test_gem_prerelease
    quick_gem "d", "1.1.a"
    refute gem("d", ">= 1"),   "release requirement must not load prerelease"
    assert gem("d", ">= 1.a"), "prerelease requirement may load prerelease"
end
```

That prompt me to question wether `#to_spec` should indeed never return `nil`. It also prompt me to think if a release requirement loading a prerelease is the actual expected behaviour (I opened #7427 for it).

So before moving on, I thought it was a good idea to ask, because maybe I am looking for a fix on the wrong place.

On the other hand, If the expected behavior is to indeed load the prerelease if it's the only available version, is this test wrong?

I'm most likely missing some context here.

### DIFF
--- a/lib/rubygems/dependency.rb
+++ b/lib/rubygems/dependency.rb
@@ -328,9 +328,9 @@ class Gem::Dependency
     return active if active
 
     unless prerelease?
-      # Move prereleases to the end of the list for >= 0 requirements
+      # Consider prereleases only as a fallback
       pre, matches = matches.partition {|spec| spec.version.prerelease? }
-      matches += pre if requirement == Gem::Requirement.default || matches.empty?
+      matches = pre if matches.empty?
     end
 
     matches.first
--- a/lib/rubygems/dependency.rb
+++ b/lib/rubygems/dependency.rb
@@ -330,7 +330,7 @@ class Gem::Dependency
     unless prerelease?
       # Move prereleases to the end of the list for >= 0 requirements
       pre, matches = matches.partition {|spec| spec.version.prerelease? }
-      matches += pre if requirement == Gem::Requirement.default
+      matches += pre if requirement == Gem::Requirement.default || matches.empty?
     end
 
     matches.first
--- a/test/rubygems/test_gem_dependency.rb
+++ b/test/rubygems/test_gem_dependency.rb
@@ -394,6 +394,16 @@ class TestGemDependency < Gem::TestCase
     assert_match "Could not find 'b' (= 2.0) among 1 total gem(s)", e.message
   end
 
+  def test_to_spec_with_only_prereleases
+    a_2_a_1 = util_spec "a", "2.a1"
+    a_2_a_2 = util_spec "a", "2.a2"
+    install_specs a_2_a_1, a_2_a_2
+
+    a_dep = dep "a", ">= 1"
+
+    assert_equal a_2_a_2, a_dep.to_spec
+  end
+
   def test_identity
     assert_equal dep("a", "= 1").identity, :released
     assert_equal dep("a", "= 1.a").identity, :complete
--- a/test/rubygems/test_gem_specification.rb
+++ b/test/rubygems/test_gem_specification.rb
@@ -3801,6 +3801,13 @@ end
     assert Gem::Specification.find_by_name "q"
   end
 
+  def test_find_by_name_with_only_prereleases_with_requirements
+    q = util_spec "q", "2.a"
+    install_specs q
+
+    assert Gem::Specification.find_by_name "q", ">= 1"
+  end
+
   def test_find_by_name_prerelease
     b = util_spec "b", "2.a"
 
--- a/test/rubygems/test_kernel.rb
+++ b/test/rubygems/test_kernel.rb
@@ -52,10 +52,20 @@ class TestGemKernel < Gem::TestCase
     assert_equal 1, $:.count {|p| p.include?("a-1/lib") }
   end
 
-  def test_gem_prerelease
+  def test_gem_prerelease_is_the_only_available
     quick_gem "d", "1.1.a"
-    refute gem("d", ">= 1"),   "release requirement must not load prerelease"
-    assert gem("d", ">= 1.a"), "prerelease requirement may load prerelease"
+
+    assert gem("d", ">= 1"), "release requirement may load prerelease when sole option"
+    assert $:.one? {|p| p.include?("1.1.a/lib") }
+  end
+
+  def test_release_favored_over_prerelease
+    quick_gem "d", "1.1.a"
+    quick_gem "d", "1.2"
+    gem("d", ">= 1")
+
+    refute $:.any? {|p| p.include?("1.1.a/lib") }
+    assert $:.one? {|p| p.include?("1.2/lib") }
   end
 
   def test_gem_env_req
--- a/test/rubygems/test_kernel.rb
+++ b/test/rubygems/test_kernel.rb
@@ -56,7 +56,7 @@ class TestGemKernel < Gem::TestCase
     quick_gem "d", "1.1.a"
 
     assert gem("d", ">= 1"), "release requirement may load prerelease when sole option"
-    assert $:.one? {|p| p.include?("1.1.a/lib") }
+    assert $:.one? {|p| p.include?("/d-1.1.a/lib") }
   end
 
   def test_release_favored_over_prerelease
@@ -64,8 +64,8 @@ class TestGemKernel < Gem::TestCase
     quick_gem "d", "1.2"
     gem("d", ">= 1")
 
-    refute $:.any? {|p| p.include?("1.1.a/lib") }
-    assert $:.one? {|p| p.include?("1.2/lib") }
+    refute $:.any? {|p| p.include?("/d-1.1.a/lib") }
+    assert $:.one? {|p| p.include?("/d-1.2/lib") }
   end
 
   def test_gem_env_req


martinemde:
Hey, thank you for taking the time to figure this out and contribute.

Unfortunately, we can't accept this because it would break current behavior. Bundler must not consider prereleases unless explicitly requested in the gemfile. This is crashing LSP because the bundle can't be resolved as is. Running bundle install should say this.

Selecting prereleases when not explicitly requested would break the contract implied by the gem being a prerelease: a prerelease gem is not ready to be automatically selected by Bundler to satisfy a requirement.

[Relevant spec](https://github.com/rubygems/rubygems/blob/d479be01abbdcdd6f6996dc75c228d4b07e213bf/bundler/spec/install/prereleases_spec.rb#L33)

kimesf:
Thank you for the explanation!

Running `bundle install` for the given `Gemfile` doesn't say anything though. Even though I think it solves `validate_timeliness` to a prerelease. I opened an issue for it (#7427).

I apologize if I'm missing something.

Thanks, again!

deivid-rodriguez:
Yes, actually Bundler only _prefers_ stable releases, but will resolve the prereleases if that's the only possibility.

I think the patch here would be about making RubyGems behave in the same way as Bundler, which in my opinion makes sense.

martinemde:
@kimesf Aha, very sorry. I misread this as being about bundler. I had been primed to see this wrong because we were talking about similar behavior in Bundler just last night. Thank you for coming in to make sense of this @deivid-rodriguez. 😀

kimesf:
> It is indeed confusing, since the gem itself will never resolve to prerelease without deps.
> 
> ```ruby
> source 'https://rubygems.org'
> 
> git_source(:github) { |repo| "https://github.com/#{repo}.git" }
> 
> ruby '3.3.0'
> 
> gem 'validates_timeliness', '>= 6.0.1'
> ```
> 
> never uses pre-release

It only solves to prerelease because of a conflict with Rails dependency of ActiveModel.

If you change the Gemfile to accept only `6.0.1`, `bundle install` breaks with this:

```console

Because rails >= 7.1.2, < 7.1.3 depends on activemodel = 7.1.2
  and rails >= 7.1.3 depends on activemodel = 7.1.3,
  rails >= 7.1.2 requires activemodel = 7.1.2 OR = 7.1.3.
And because validates_timeliness >= 6.0.0.beta1, < 7.0.0.beta1 depends on activemodel >= 6.0.0, < 7,
  rails >= 7.1.2 is incompatible with validates_timeliness >= 6.0.0.beta1, < 7.0.0.beta1.
So, because Gemfile depends on rails ~> 7.1.2
  and Gemfile depends on validates_timeliness = 6.0.1,
  version solving has failed.
```

After `6.0.1`, the gem only have `7.0.0.beta1` and `7.0.0.beta2` available.

kimesf:
> Yes, actually Bundler only _prefers_ stable releases, but will resolve to prereleases if that's the only possibility.
> 
> I think the patch here would be about making RubyGems behave in the same way as Bundler, which in my opinion makes sense.

If that's the case them, I can safely change the current breaking test, correct? I'll also probably add at least one test to differentiate the case `when the prerelease is the only available one` from the case `when the prerelease is not the only available one`.

deivid-rodriguez:
We could change never resolving to prereleases unless explicitly requested, but the current behavior is intentional, so we'd need to do it in a major release in my opinion.

deivid-rodriguez:
I personally think it's fine as is, but just saying that if we decide to change it, I think we should do it in a major version.

The same thing could be argued for this (opposite) change in RubyGems, of course. I thought this was unintentional behavior, but since there's a test backing this up, I guess it's actually intended too!

kimesf:
In case you decide to merge this, I changed the breaking test in order to elucidate the case where pre-releases are the only available ones.

Running `bundler/bin/parallel_spec` locally broke 4 tests, but they were already broken before I did any changes, so I guess it's something related to my environment. It certainly looks like so.

If you want any changes, regardless of having decided anything, please let me know :)

indirect:
IMO the most important thing is that we should give Bundler and RubyGems the same behavior. I think either of these two options is a reasonable behavior to choose, but it's surprising and confusing to discover that our two overlapping tools do something different.

Moving on to which behavior we should keep, I would vote to keep Bundler's behavior, since it most widely used today. Many existing Gemfiles that successfully resolve now would stop resolving if we changed the requirement to demand explicit pre-release version numbers. In my opinion, it is enough that 1) Bundler only chooses pre-releases as a last resort, 2) the prerelease version is printed out during the install message, and 3) the prerelease version is written into the lockfile. If a user truly does not want a prerelease, that gives them multiple chances to notice and make changes.

lib/rubygems/dependency.rb
```diff
@@ -330,7 +330,7 @@ def to_spec
     unless prerelease?
       # Move prereleases to the end of the list for >= 0 requirements
       pre, matches = matches.partition {|spec| spec.version.prerelease? }
-      matches += pre if requirement == Gem::Requirement.default
+      matches += pre if requirement == Gem::Requirement.default || matches.empty?
```
deivid-rodriguez, line 333: There's actually another existing inconsistency.

RubyGems already falls back to prereleases even when the requirement is not a prerelease requirement, as long as the requirement is specifically ">=0" (which is also the default for dependencies that don't specify any requirement).

After realizing about this, I think it'd be fine to call this a bug and fix the inconsistency.

Also, I think it can be simplified to

```suggestion
      matches = pre if matches.empty?
```

and also we can reword the comment two lines above to something like "Consider prereleases only as a fallback".

kimesf: Yes, I can see how `if requirement == Gem::Requirement.default` is a bug now.

Thanks for the explanation.

kimesf:
I guess the failing test is flaky because I'm not specifying the gem name.

Once the current check pipeline is done, I'm changing this:

```ruby
assert $:.one? {|p| p.include?("1.1.a/lib") }
refute $:.any? {|p| p.include?("1.1.a/lib") }
assert $:.one? {|p| p.include?("1.2/lib") }
```

To this:

```ruby
assert $:.one? {|p| p.include?("d-1.1.a/lib") }
refute $:.any? {|p| p.include?("d-1.1.a/lib") }
assert $:.one? {|p| p.include?("d-1.2/lib") }
```

kimesf:
I went with `"/d-1.2/lib"` (prefixed with `/`) to be sure.

But there was another fail for the [windows ruby 3.2 check](https://github.com/rubygems/rubygems/actions/runs/7732674050/job/21083191115?pr=7428) that I don't get it.

kimesf:
Rebased from master.

martinemde (approved):
Looks correct. Thank you for figuring this out.

deivid-rodriguez (approved):
Yeah, sorry for late feedback. Also looks good to me!

kimesf:
Awesome! Thanks for your time!

kimesf:
Rebased from master.

PR #7430, opened 2024-01-27
Bump `required_ruby_version` to be used in `bundle gem` template

ohbarye:

`bundle gem` command generates `.gemspec`, `rubocop.yml` and `standard.yml` with outdated Ruby version 2.6.0. As a gem creator, I expect the version should be any supported one.

I just bumped `required_ruby_version` to 3.0.0 that is still supported.

I think it's reasonable since 2.6 and 2.7 are EOL and bundler dropped their support by https://github.com/rubygems/rubygems/pull/7116.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -437,7 +437,7 @@ module Bundler
     end
 
     def required_ruby_version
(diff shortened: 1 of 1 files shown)


martinemde (approved):
Yes of course! Thank you for spotting this.

PR #6150, opened 2022-12-19
Remove Travis CI from gem skeleton.

simi:
- it doesn't provide free program for OSS anymore
- make it friendly

PR #7431, opened 2024-01-27
Remove Travis related code from `bundle gem` completely

ohbarye:

This is just internal code clean-up.

I removed `travis_removal_info`.

`travis_removal_info` is added by https://github.com/rubygems/rubygems/pull/6150, at [v2.4.0](https://github.com/rubygems/rubygems/blob/0a69ce21ca3e8bfbc3ff7a6cec6d36ba2c434128/bundler/CHANGELOG.md#breaking-changes-1). According to the comment, it's supposed to be removed at bundler v2.5.0 but it still exists.

### DIFF
--- a/bundler/lib/bundler/cli/gem.rb
+++ b/bundler/lib/bundler/cli/gem.rb
@@ -32,7 +32,6 @@ module Bundler
 
       validate_ext_name if @extension
       validate_rust_builder_rubygems_version if @extension == "rust"
-      travis_removal_info
     end
 
     def run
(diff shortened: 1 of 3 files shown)


segiddins (approved):

martinemde (approved):
Blame says this was "2 years ago" so I think we're good to remove.

Thank you!

ohbarye:
Thanks for your review!

bundler/lib/bundler/cli/gem.rb
```diff
@@ -448,19 +447,6 @@ def standard_version
...
     end
 
-    # TODO: remove at next minor release
-    def travis_removal_info
-      if options[:ci] == "travis"
-        Bundler.ui.error "Support for Travis CI was removed from gem skeleton generator."
-        exit 1
-      end
-
-      if Bundler.settings["gem.ci"] == "travis"
```
simi, line 458: This can cause troubles for people having this in config file (not passing travis as an argument on cli) breaking commands. :thinking: Are we sure this is ok to do now? Was this deprecation present already in bundler included with Ruby 3.2?

ohbarye: >Was this deprecation present already in bundler included with Ruby 3.2?


According to [Ruby 3.2.0 release note](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/#:~:text=bigdecimal%203.1.3-,bundler%202.4.1,-cgi%200.3.6), Ruby 3.2.0 includes bundler 2.4.1 that include the deprecation. 👀

martinemde: @simi When I checked, the code was 2 years old, so it's been through at least 2 minor bundler releases.

simi: OK, all good. Thanks for check!

PR #7439, opened 2024-01-30
feature: deep copy requirements in Gem::Specification and Gem::Requirement

flavorjones:

While drafting https://github.com/rake-compiler/rake-compiler/pull/236, I needed to modify the `required_rubygems_version` of a `Gem::Specification` that was a copy of the original.

While doing so, I discovered two things:

- modifying the `required_rubygems_version` of a `Gem::Specification` copy also mutates the original's
- modifying the requirements of a `Gem::Requirements` copy also mutates the original's

That is:

```ruby
requirement2 = requirement.dup

# this modifies original `requirement`
requirement2.concat([">= 3.3.22"])
```

and

```ruby
spec2 = spec.dup

# this modifies original `spec`'s `required_rubygems_version`
spec2.required_rubygems_version.concat([">= 3.3.22"])
```

This was confusing and led to more complex code and tests.

This PR implements what I believe to be the common-sense behavior of "copy" for these classes to avoid accidental mutation of the originals:

- deep-copies the `Gem::Requirement` attributes of a `Gem::Specification`
- deep-copies the `@requirements` attribute of a `Gem::Requirement`

With this new behavior, my use case described above can be solved by:

```ruby
spec2 = spec.dup
spec2.required_rubygems_version.concat([">= 3.3.22"])
```

### DIFF
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -2075,7 +2075,8 @@ class Gem::Specification < Gem::BasicSpecification
   end
 
   ##
-  # Duplicates array_attributes from +other_spec+ so state isn't shared.
+  # Duplicates Array and Gem::Requirement attributes from +other_spec+ so state isn't shared.
+  #
 
   def initialize_copy(other_spec)
     self.class.array_attributes.each do |name|
@@ -2097,6 +2098,9 @@ class Gem::Specification < Gem::BasicSpecification
         raise e
       end
     end
+
+    @required_ruby_version = other_spec.required_ruby_version.dup
+    @required_rubygems_version = other_spec.required_rubygems_version.dup
   end
 
   def base_dir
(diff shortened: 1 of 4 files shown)


lib/rubygems/specification.rb
```diff
@@ -2097,6 +2097,9 @@ def initialize_copy(other_spec)
         raise e
       end
     end
+
+    @required_ruby_version = other_spec.required_ruby_version.dup
+    @required_rubygems_version = other_spec.required_rubygems_version.dup
```
deivid-rodriguez, line 2102: We should probably tweak the documentation of this method now to match the new implementation (it's not only "array attributes" now).

deivid-rodriguez (commented):
Looks good to me. Definitely more common sense! Just left a small comment.

lib/rubygems/specification.rb
flavorjones: Done!

deivid-rodriguez (approved):
Thanks!

PR #7442, opened 2024-01-30
Make slow perf specs more stable

deivid-rodriguez:

Specs that measure how long resolution takes sometimes fail, and it's hard to choose a number that never fails but it's still useful for detecting performance degradation.

I chose an alternative that's not as good for the task, but at least should not have many false positives.


deivid-rodriguez:
Like with #7440, merging since this is test only and should make CI more stable, but happy to iterate on anything.

PR #7449, opened 2024-02-01
Add `gitlab:` git source shorthand

### LINKED ISSUE #7210: Add Codeberg and GitLab shorthands
## Description

For nearly a decade, Bundler has offered [`bitbucket:`](https://bundler.io/v2.4/man/gemfile.5.html#BITBUCKET), [`github:`](https://bundler.io/v2.4/man/gemfile.5.html#GITHUB), and [`gist:`](https://bundler.io/v2.4/man/gemfile.5.html#GIST) shorthands for use in `Gemfile` when defining a gem's source.

As of RubyGems 3.4.22, those are configured in the `Bundler::Dsl` class' `#add_git_sources` method:

https://github.com/rubygems/rubygems/blob/ec2089640dbbb0de136c77190cc20d78aadada15/bundler/lib/bundler/dsl.rb#L275-L299

At one point (see 924cd58), `bitbucket:` and `gist:` were marked for deprecation, but that seems to have been removed at some point along the way. I couldn't locate anything regarding the inclusion of `bitbucket:` and `gist:` (both seem curious to me, but then, there's a lot of history in this codebase).

## Proposed Solution

If it's accurate that `bitbucket:` and `gist:` are _not_ deprecated, **I propose adding similar shorthands for two other prominent Git providers: [GitLab](https://gitlab.com) and [Codeberg](https://codeberg.org)**.

On the 👍🏻 side, this proposal would add convenience for users of those platforms (myself included!).

On the 👎🏻 side, this proposal would add more (seemingly) arbitrary shorthands to the codebase.

If this proposal is acceptable, I'd be happy to write the change and necessary tests and submit a Pull Request.

Thanks for your consideration!

jgarber623:
First off, thank you for considering this Pull Request! I appreciate your time.

In #7210, I noted the existence of the `github:` and `bitbucket:` shorthands for installing gems from Git sources. I proposed adding two new providers, `gitlab:` and `codeberg:`, for the notable Git services [Gitlab](https://gitlab.com) and [Codeberg](https://codeberg.org). The proposal to add Gitlab was accepted.

The changes proposed in this PR would provide a small quality-of-life improvement for users with Ruby dependencies whose source repositories exist somewhere other than GitHub.

Resolves #7210.

This PR adds `gitlab:` shorthand with similar functionality to the existing `github:` shorthand:

```ruby
gem "gitlab-sdk", gitlab: "gitlab-org/analytics-section/product-analytics/gl-application-sdk-rb"
```

Merge Request URLs are also supported for each new shorthand:

```ruby
gem "gitlab-sdk", gitlab: "https://gitlab.com/gitlab-org/analytics-section/product-analytics/gl-application-sdk-rb/-/merge_requests/27"
```

### DIFF
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -19,6 +19,7 @@ module Bundler
                     platform platforms type source install_if gemfile force_ruby_platform].freeze
 
     GITHUB_PULL_REQUEST_URL = %r{\Ahttps://github\.com/([A-Za-z0-9_\-\.]+/[A-Za-z0-9_\-\.]+)/pull/(\d+)\z}
+    GITLAB_MERGE_REQUEST_URL = %r{\Ahttps://gitlab\.com/([A-Za-z0-9_\-\./]+)/-/merge_requests/(\d+)\z}
 
     attr_reader :gemspecs, :gemfile
     attr_accessor :dependencies
@@ -307,6 +308,20 @@ module Bundler
         user_name, repo_name = repo_name.split("/")
         repo_name ||= user_name
         "https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git"
+      end
+
+      git_source(:gitlab) do |repo_name|
+        if repo_name =~ GITLAB_MERGE_REQUEST_URL
+          {
+            "git" => "https://gitlab.com/#{$1}.git",
+            "branch" => nil,
+            "ref" => "refs/merge-requests/#{$2}/head",
+            "tag" => nil,
+          }
+        else
+          repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
+          "https://gitlab.com/#{repo_name}.git"
+        end
       end
     end
 
--- a/bundler/spec/bundler/dsl_spec.rb
+++ b/bundler/spec/bundler/dsl_spec.rb
@@ -32,6 +32,13 @@ RSpec.describe Bundler::Dsl do
       expect(subject.dependencies.first.source.ref).to eq("refs/pull/5/head")
     end
 
+    it "converts :gitlab PR to URI using https" do
+      subject.gem("sparks", gitlab: "https://gitlab.com/indirect/sparks/-/merge_requests/5")
+      gitlab_uri = "https://gitlab.com/indirect/sparks.git"
+      expect(subject.dependencies.first.source.uri).to eq(gitlab_uri)
+      expect(subject.dependencies.first.source.ref).to eq("refs/merge-requests/5/head")
+    end
+
     it "rejects :github PR URI with a branch, ref or tag" do
       expect do
         subject.gem("sparks", github: "https://github.com/indirect/sparks/pull/5", branch: "foo")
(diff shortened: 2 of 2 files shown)


deivid-rodriguez:
Thanks!

I think in the related issue there was pretty much consensus on adding Gitlab. I think codeberg was more unclear, though. In particular, thresholds proposed by @indirect at https://github.com/rubygems/rubygems/issues/7210#issuecomment-1839714042 would exclude it.

So we can proceed in 2 ways:
* Add Gitlab only for now.
* Wait a bit until we make a final decision.

Regarding the implementation, it's pretty confusing, but the tests you added would be for an implementation of the Bundler DSL that actually lives in RubyGems and it's not really used by Bundler at the moment. The proper place for these tests would be in `bundler/spec/bundler/dsl_spec.rb`.

indirect:
I think we should definitely add GitLab. I am less clear on whether we should add Codeberg (and Sourcehut, and Beanstalk, and CodeCommit, and Cloud Source Repositories, and Launchpad, and Keybase, and Planio, and Tuleap). My main hesitation adding smaller repository sources is churn, having to constantly do work to update our shortcuts if URLs change or services shut down or move.

If we see consistent user requests for other sources, I am fine with adding them. So far I think we have only seen that for GitLab.

deivid-rodriguez:
@jgarber623 Could you limit this PR to adding support for the `gitlab` shorthand?

jgarber623:
Thanks for the feedback, y'all! I've reverted the commit adding the Codeberg shortcut. Let me know if you'd rather avoid the revert commit and I can force push. Not sure which of GitHub's many merge approaches you prefer.

@deivid-rodriguez My apologies for that first pass at the tests. I've added a bunch of `gitlab`-specific ones in 98905e6, basing them on the existing `github` specs (since the two shortcuts are mostly the same). Let me know if there's anything else worth testing.

Thanks again and appreciate your consideration!

deivid-rodriguez:
No problem at all! Not your fault, it's a bit confusing :sweat_smile: 

Yes, if you can squash everything into one commit, I do prefer that :+1:.

Thanks for the PR!

deivid-rodriguez (approved):
Thank you!

bundler/lib/bundler/dsl.rb
```diff
@@ -308,6 +309,20 @@ def add_git_sources
         repo_name ||= user_name
         "https://#{user_name}@bitbucket.org/#{user_name}/#{repo_name}.git"
       end
+      
+      git_source(:gitlab) do |repo_name|
```
simi, line 313: Since "gitlab" is name of the project and it could be also self-hosted and this points to public instance of gitlab.com, should this be `gitlab_com`?

simi: > Isn't that also true about GitHub enterprise?

GitHub EE is not so common per my experience, but gitlab is often self-hosted by organizations. I'm also not happy about github shortcut btw.

martinemde: > I'm also not happy about github shortcut btw. 

I agree. We were moving in the way of declaring it within the Gemfile.

This PR is a small ask overall. We've now drawn a clear line and we're enforcing it. I think this will be the last addition for a while baring a major shift.

We should think about, and even purposely deprecate the others that are smaller. That will give us a pattern for deprecating these when they are too small. I'd accept one no deprecating bitbucket, provided it printed the exact code to add to your Gemfile to fix the deprecation.

martinemde: Sorry, I misremembered the bitbucket traffic numbers. Scratch that. Anything smaller should be removed though.

martinemde:
Waiting for tests to be fixed.

jgarber623:
Formatting issues addressed. 👍🏻

Both `rspec ./spec/quality_spec.rb` and `rake rubocop` are reporting all green.

martinemde:
Thank @jgarber623 !

bundler/lib/bundler/dsl.rb
simi: > I think this will be the last addition for a while barring a major shift.

:shrug:

PR #7446, opened 2024-01-31
Add missing Windows job to CI

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Found one missing job in CI.

## What is your fix for the problem, implemented in this PR?

Add the missing job.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7450, opened 2024-02-01
Fix relative path bug in ext_conf_builder

composerinteralia:

https://github.com/rubygems/rubygems/pull/6137 introduced a bug in `get_relative_path` that will sometimes cause gems to get installed to the wrong location.

We found this while trying to install the google-protobuf gem, which has its extension in `ext/google/protobuf_c`. The problem is that `zip` will only zip up to the length of `path`, but in the case of google-protobuf `base` has more components than `path`.

So with this base and path:

```
/home/.rubies/ruby-master/lib/ruby/gems/3.4.0+0/extensions/x86_64-darwin-23/3.4.0+0-static/google-protobuf-3.25.2
/home/.rubies/ruby-master/lib/ruby/gems/3.4.0+0/gems/google-protobuf-3.25.2/ext/google/protobuf_c
```

That will `zip` to:

```
[["home", ..., ["google-protobuf-3.25.2", "google"]]
```

rather than:

```
[["home", ..., ["google-protobuf-3.25.2", "google"], [nil, "protobuf_c"]
```

As a result, the extension gets built into the wrong directory (with four "../" instead of five). The gem appears to install just fine, but then fails to load because the file isn't where it should be.

We think there was also a theoretical (although unlikely) bug with `next result if path_component == base_component`, since components that aren't part of the shared base path could also match. If a gem had an extension dir like "ext/google-protobuf-3.25.2", for example, we'd end up stripping that away incorrectly.

This commit reworks `get_relative_path` to avoid these bugs. We backtrack from the gem base one component at a time until we find the shared base, and keep track of how far we had to backtrack. Then we strip the shared base from the path and add the right number of ".." based on the backtracks.

We were unsure whether the "/" separator in the previous implementation would have worked for all environments. This commit uses `File` helpers so that is no longer a concern.


composerinteralia:
Looks like maybe the truffleruby build crashed:

```
free(): double free detected in tcache 2
Aborted (core dumped)
```

Should we re-run it? (I don't think I have access to do that)

lib/rubygems/ext/ext_conf_builder.rb
```diff
@@ -50,11 +50,16 @@ def self.build(extension, dest_path, results, args=[], lib_dir=nil, extension_di
...
-      result.prepend("../") if base_component
-      result.concat("#{path_component}/") if path_component
-      result
+    absolute_path = File.expand_path(path)
+    absolute_base = File.expand_path(base)
+
+    backtracks = 0
+    prefix = absolute_base
+    until absolute_path.start_with?(prefix)
+      prefix = File.dirname(prefix)
```
martinemde, line 59: Can you add a raise in here if prefix becomes "." so that it doesn't construct a wrong path or loop forever if this, somehow, gets called with paths that don't have a common ancestor?

Maybe overly paranoid since the original was no better, but looking at [Pathname's implementation](https://docs.ruby-lang.org/en/master/Pathname.html#method-i-relative_path_from) maybe it doesn't hurt to exit explicitly in that case?

martinemde (commented):
I think it looks good with a fix to protect against bad paths in unexpected cases. It could just ArgumentError if the prefix never satisfies.

@deivid-rodriguez maybe you have more context about the actual usage here and can approve.

lib/rubygems/ext/ext_conf_builder.rb
deivid-rodriguez: It seems strange that `File.dirname` ever returns `.`, but it does return that on Windows when reaching the root drive, not sure why. If `base` and `path` were on different drives, then we could potentially end up looping forever, indeed? That said, I'm not sure  `base` and `path` can ever be on different drives?

composerinteralia: Fun! I thought the calls to `expand_path` would take care of everything, but I guess not 😅.

composerinteralia: Pushed up a change to raise if `File.dirname` doesn't change anything. That would mean we've gone as far as we can go and didn't find a match. I think that covers both the "." case and the different drives case. 🤞🏻

composerinteralia:
Hm, looks like those windows jobs are failing on master too after https://github.com/rubygems/rubygems/pull/7446

deivid-rodriguez:
Yeah, working on a workaround for those now. Will merge this shortly.

lib/rubygems/ext/ext_conf_builder.rb
```diff
@@ -50,11 +50,21 @@ def self.build(extension, dest_path, results, args=[], lib_dir=nil, extension_di
...
-      next result if path_component == base_component
-      result.prepend("../") if base_component
-      result.concat("#{path_component}/") if path_component
-      result
+    absolute_path = File.expand_path(path)
+    absolute_base = File.expand_path(base)
+
+    backtracks = 0
+    prefix = absolute_base
+    until absolute_path.start_with?(prefix)
```
segiddins, line 58: Won't this be incorrect for eg `base=/a/b, path=/a/bc` ?

composerinteralia: Blah, yeah probably.

"This should be easy enough to fix" he though...

Maybe we should go back to splitting on "/" after all. The approach we had started with was, roughly:

```
path_parts = path.split("/")
base_parts = base.split("/")

while path_parts[0] == base_parts[0]
  path_parts.unshift
  base_parts.unshift
end

backtracks = [".."] * base_parts.size
File.join(backtracks, path_parts)
```

composerinteralia: I guess that's getting kinda close to the Pathname implementation, so maybe we want something pretty close to that

deivid-rodriguez: Yeah, sure! I could've also used `pathname` directly, but introducing stdlib dependencies usually gives us trouble, so I avoided doing it just for this.

composerinteralia: Alright, pushed up something new. 🤞🏻 

Is it fine to split on "/"? That seems like it wouldn't necessarily work on windows, but I'm not too familiar with how that all works. @gabriel-amaral came up with another approach that doesn't involve splitting on "/", but I don't know if it's necessary:

```rb
  def self.get_parts(path)
    parts = []
    prefix = path
    until File.dirname(prefix) == prefix
      dirname, basename = File.split(prefix)
      parts << basename
      prefix = dirname
    end
    parts.reverse
  end
```

deivid-rodriguez: I guess let's go with the alternative approach just in case. We could split by `File::SEPARATOR`, but there's also `File::ALT_SEPARATOR` so we'd need to check both and it seems it would get messy?

deivid-rodriguez:
So last question is:

Is what we had before #6137 simpler than what we have now, because if not the whole point of #6137 is defeated... :sweat_smile: 

I do think the new way is better overall but it does bring this extra complexity that we did not have before because `path` was relative to `base`.

composerinteralia:
> Is what we had before https://github.com/rubygems/rubygems/pull/6137 simpler than what we have now, because if not the whole point of https://github.com/rubygems/rubygems/pull/6137 is defeated... 😅

That is a really good point. I'm not totally sure. Another thing we started to explore was whether it'd be possible to adjust how we're doing the shell escaping in such a way that spaces are no longer a problem (that seems to be the main reason we are using a relative path in the first place), but it seems like we'd have to rework quite a bit and we were worried about getting it wrong.

gabriel-amaral:
> So last question is:
> 
> Is what we had before #6137 simpler than what we have now, because if not the whole point of #6137 is defeated... 😅
> I do think the new way is better overall but it does bring this extra complexity that we did not have before because `path` was inside `base`.

That is a good point. If I got https://github.com/rubygems/rubygems/pull/6137 correctly, it eliminated the third directory (tmp) creation and then moving files from it to the dest. As you mentioned, the relative path calc with the temp directory was simpler. I think it was a good trade; the disk operation for processing cost. The file separator complexity is still a question to me. Is it enough to go back to create and move files around? I think the answer is no.

Ideally, we could work with the absolute paths and do not need to calculate relative paths. As @composerinteralia said, we pursued that for a bit, but it seemed too complex to get to the bottom.

lib/rubygems/ext/ext_conf_builder.rb
segiddins: We could add a method next to `Gem.path_separator` that's like `Gem.file_separator`, and if ALT_SEPARATOR is defined then cache a `Regexp.union` of the two?

deivid-rodriguez:
I guess since it's unclear which one is better, I'll revert #6137 for now and reopen the 2 PRs stacked together as a WIP. Indeed ideally we'd just work with absolute paths and remove both complexities.

PR #7452, opened 2024-02-02
Remove no longer necessary `ruby/setup-ruby-pkgs` action

deivid-rodriguez:

RubyGems CI on Windows is broken.

Stop using `ruby/setup-ruby-pkgs`. I was only added as a [workaround](https://github.com/rubygems/rubygems/pull/5175#discussion_r794873977) originally but it seems no longer necessary.

### DIFF
--- a/.github/workflows/rubygems.yml
+++ b/.github/workflows/rubygems.yml
@@ -49,14 +49,6 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby.value }}
           bundler: none
-        if: matrix.os.name != 'Windows'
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
Interestingly, we're not running rust tests on mingw since https://github.com/rubygems/rubygems/pull/6573.

Also, `ruby/setup-ruby-pkgs` is back to working now.

However, it introduces some brittleness and it's not necessary since rust tests are broken anyways (and I verified after removing it they are still broken in the same way).

So I will merge this.

segiddins (approved):

PR #7460, opened 2024-02-06
Warn if RubyGems version explicitly set in gemspec does not match running version

deivid-rodriguez:

This gets in the middle if we ever start allowing to build as if using a different RubyGems version than the one being run.

This could be useful to make `gem rebuild` a little more usable, and it's already done by Bundler specs which already make this method a noop when they need this.

I'm not sure forcefully setting this, even if user explicitly specified something else is helpful.

Since this could potentially prevent gems explicitly setting a constant RubyGems version from building, I changed the error of incorrect RubyGems version from a hard error to a warning, since it will start happening in those cases if we stop overwriting the version.

### DIFF
--- a/bundler/spec/support/builders.rb
+++ b/bundler/spec/support/builders.rb
@@ -518,7 +518,6 @@ module Spec
 
         if options[:rubygems_version]
           @spec.rubygems_version = options[:rubygems_version]
-          def @spec.mark_version; end
 
           def @spec.validate(*); end
         end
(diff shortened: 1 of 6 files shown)


deivid-rodriguez:
@duckinator Does this make sense to you? The idea is to remove any hurdles to eventually allow mocking the RubyGems version while building gems. We could use this for gem rebuild by using the RubyGems version used to build the original gem for the gem being rebuilt to hopefully eliminate the diff caused by the version mismatch.

duckinator:
@deivid-rodriguez this seems reasonable, and I think it can be extended like you said to help `gem rebuild`. Sorry for the delay!

PR #7471, opened 2024-02-12
Use full path for `instance_eval` in `Bundler::DSL#eval_gemfile`

ccutrer:

When using `eval_gemfile("../gemfile_prefix.rb")` with a Gemfile that's not in the current directory, `__dir__` and `__FILE__`
end up being `".."` and `"gemfile_prefix.rb"` respectively, which is not usable from within gemfile_prefix.rb in order to find other relative files, since `Dir.pwd` is not the Gemfile's directory.

`eval_gemfile` already calculates the absolute path to the gemfile fragment; just pass that to `instance_eval`. This makes it behave the same wrt `__dir__` and `__FILE__` as `require_relative`.

### DIFF
--- a/bundler/lib/bundler/dsl.rb
+++ b/bundler/lib/bundler/dsl.rb
@@ -46,7 +46,7 @@ module Bundler
       @gemfile = expanded_gemfile_path
       @gemfiles << expanded_gemfile_path
       contents ||= Bundler.read_file(@gemfile.to_s)
-      instance_eval(contents, gemfile.to_s, 1)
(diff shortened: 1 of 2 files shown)


segiddins:
Looks good with an added test!

ccutrer:
:+1: I'll get some tests written. I wanted to push up (as a draft) just to make sure it didn't break anything else before I invested time in writing a test.

ccutrer:
Spec added.

simi (approved):

PR #7478, opened 2024-02-18
Remove redundant rubocop configs

dduugg:

When creating a new gem with `bundle gem`, the specified rubocop version is `"~> 1.21"`, which already has several of the configs values set:

- The default `Max` value of `LineLength` has been 120 since [v0.84.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#changes-80)
- `Style/StringLiterals` and `Style/StringLiteralsInInterpolation` have been enabled by default for even [longer](https://github.com/rubocop/rubocop/pull/6293/files#diff-3653213e3a153018a75cc7b2536b1d0217a6d8d75188d71cf45441812f789698R3986)

I've removed the redundant config settings.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/rubocop.yml.tt
+++ b/bundler/lib/bundler/templates/newgem/rubocop.yml.tt
@@ -2,12 +2,7 @@ AllCops:
   TargetRubyVersion: <%= ::Gem::Version.new(config[:required_ruby_version]).segments[0..1].join(".") %>
 
 Style/StringLiterals:
-  Enabled: true
   EnforcedStyle: double_quotes
 
 Style/StringLiteralsInInterpolation:
-  Enabled: true
   EnforcedStyle: double_quotes
-
-Layout/LineLength:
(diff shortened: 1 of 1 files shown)


simi:
Thanks @dduugg! Is there any reason to keep it `~> 1.21`? Maybe we can also update rubocop constraint.

dduugg:
> Thanks @dduugg! Is there any reason to keep it `~> 1.21`? Maybe we can also update rubocop constraint.

We could, but the soft pin means this should still resolve to the latest release (`1.60.2`). Using `1.50` could work, it's the last release that runs on Ruby 2.6 (the `TargetRubyVersion` setting, and the `spec.required_ruby_version` minimum, though both Ruby 2.6 and 2.7 have reached end-of-support status).

simi:
Ahh, good point @dduugg about the Ruby support.

simi (approved):

hsbt (approved):

PR #7484, opened 2024-02-20
Use a constant empty tar header to avoid extra allocations

segiddins:


martinemde (approved):
Nice.

lib/rubygems/package/tar_header.rb
```diff
@@ -172,6 +172,15 @@ def initialize(vals)
     @empty = vals[:empty]
   end
 
+  EMPTY = new({ # :nodoc:
+    mode: 0,
+    name: "",
+    prefix: "",
+    size: 0,
+    empty: true,
+  }).freeze
```
martinemde, line 181: To really be identical you need the following. The default values are not the same as what an empty header outputs.

```suggestion
  EMPTY = new({ # :nodoc:
    checksum: 0,
    gname: "",
    linkname: "",
    magic: "",
    mode: 0,
    name: "",
    prefix: "",
    size: 0,
    uname: "",
    version: 0,
    empty: true,
  }).freeze
```
Here's my irb doublecheck:

```ruby
irb(main):033> EMPTY = Gem::Package::TarHeader.new({ checksum: 0, gname: "", linkname: "", magic: "", mode: 0, name: "", prefix: "", size: 0, uname: "", version: 0, empty: true,  })
=> #<Gem::Package::TarHeader:0x00000001203e60e0 @checksum=0, @devmajor=0, @devminor=0, @empty=true, @gid=0, @gname="", @linkname="", @magic="", @mode=0, @mtime=0, @name="", @prefix="", @size=0, @typeflag="0", @uid=0, @uname="", @version=0>
irb(main):034> Gem::Package::TarHeader.from(StringIO.new(Gem::Package::TarHeader::EMPTY_HEADER))
=> #<Gem::Package::TarHeader:0x00000001204cde90 @checksum=0, @devmajor=0, @devminor=0, @empty=true, @gid=0, @gname="", @linkname="", @magic="", @mode=0, @mtime=0, @name="", @prefix="", @size=0, @typeflag="0", @uid=0, @uname="", @version=0>
```


We should also add a test.

martinemde (commented):
I approved it before double checking, here's the additional I think it needs.

simi (approved):
:thinking: *wondering what's the impact of this change*

martinemde:
@simi are you thinking the tar header is expected to be mutable? That's worth considering...

simi:
> @simi are you thinking the tar header is expected to be mutable? That's worth considering...

No, I was just wondering what's memory footprint impact.

lib/rubygems/package/tar_header.rb
```diff
@@ -241,7 +257,7 @@ def header(checksum = @checksum)
 
     header = header.pack PACK_FORMAT
 
-    header << ("\0" * ((512 - header.size) % 512))
+    header.ljust 512, "\0"
```
deivid-rodriguez, line 260: This seems like a neat change regardless of the rest :)

deivid-rodriguez (approved):
I also wonder about impact. Do we instantiate an empty tar reader once per gem, and this would reuse it for `gem install` invocations that install many gems?

Anyways, code changes look good to me!

segiddins:
There is one empty tar header per tar file read, in my experience. each gem requires reading 2 tar files.

deivid-rodriguez (approved):

deivid-rodriguez:
Should we squash into a single commit? Right now there's three commits, two of them with the same commit message, which seems a bit confusing! Other than that, seems great!

deivid-rodriguez:
@segiddins I squashed commits myself and will be merging this, hope it's ok.

PR #7485, opened 2024-02-21
Get rid of too long paths

nobu:

This is a fix-up of #7483.

Test method names in Rubygems are often very long, and the path under the temporary directory generated from them can easily exceed system limits.

Even without including the method name in the path name, by saving the method and path name pair, it is possible to find the method from the remaining path name.


hsbt:
`test_build_cdylib` and `test_rubygems_cfg_passed_to_rustc` are leaked from temp directory.

nobu:
Probably, seems that `assert_ffi_handle` and `refute_ffi_handle` leave `Fiddle` handles open.

PR #7488, opened 2024-02-25
Use `IO.popen` to list files

nobu:

When building by `gem`, `__FILE__` is the path name given in the command line, or the gemspec file name in the current directory.  In that case, comparison it and expanded path never equal.

List file names by using `chdir` option of `IO.popen`, and compare them with the base name of `__FILE__` instead.

Also redirect stderr `git ls-files` to null without shelling out.

### DIFF
--- a/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
+++ b/bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
@@ -27,9 +27,10 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files = Dir.chdir(__dir__) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      (File.expand_path(f) == __FILE__) ||
+  gemspec = File.basename(__FILE__)
+  spec.files = IO.popen(%w[git ls-files -z], chdir: __dir__, err: IO::NULL) do |ls|
+    ls.readlines("\x0", chomp: true).reject do |f|
+      (f == gemspec) ||
         f.start_with?(*%w[bin/ test/ spec/ features/ .git <%= config[:ci_config_path] %>appveyor Gemfile])
     end
   end


bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
```diff
@@ -27,9 +27,10 @@ Gem::Specification.new do |spec|
 
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
-  spec.files = Dir.chdir(__dir__) do
-    `git ls-files -z`.split("\x0").reject do |f|
-      (File.expand_path(f) == __FILE__) ||
+  gemspec = File.basename(__FILE__)
+  spec.files = IO.popen(%w[git ls-files -z], chdir: __dir__, err: IO::NULL) do |ls|
+    ls.readlines("\x0", chomp: true).reject do |f|
+      (f == gemspec) ||
```
martinemde, line 33: This was already a complicated line to drop into a new gem. The change makes it even more complicated.

Should we add a helper to `spec` that does this? Maybe:

`spec.files = spec.git_files_excluding(__FILE__, "bin", "spec")`

I know this is a more involved change. We could make a temporary fix while we wait for improved syntax.

martinemde: > BTW, what version of git is required in the current rubygems?

I think we must support the version of git that ships with OSX. We've never made a real requirement that I know of.

@hsbt, what do you think? I don't feel strongly myself and I would be fine merging this if others are ok with it.

segiddins (approved):

martinemde (approved):

bundler/lib/bundler/templates/newgem/newgem.gemspec.tt
nobu: On macOS 14.3.1:

```sh-session
$ /usr/bin/git --version
git version 2.39.3 (Apple Git-145)
```

Not sure about older versions.

hsbt: I investigate git versions of debian and ubuntu.

* Ubuntu 22.04: 2.34.1
* Debian 12: 2.39.2

We should support git-2.34.x at least.

nobu: @hsbt Thanks, git 2.x is enough for the "magic signatures".

I found this description.
> [New Features](https://developer.apple.com/documentation/xcode-release-notes/xcode-12-release-notes#New-Features)
> You can now specify the default branch name for new repositories in the Source Control pane of Xcode’s Preferences. This preference uses the init.defaultBranch Git config option available in Git version 2.28. (66232985)

deivid-rodriguez: We don't have an official "lowest supported version" but we do ship code to make sure to play nice with git [older than 1.8.5](https://github.com/rubygems/rubygems/blob/8e52109990f729b25a5aed07e807002a791c9513/bundler/lib/bundler/source/git/git_proxy.rb#L437-L440).

If I recall correctly, the motivation was that CentOS 7 shipped with a very old git version. CentOS 7 reaches its end of life by the 30th of June, I think it'd be ok to drop support after that an officially declare an official lowest supported version like 2.x.

PR #7517, opened 2024-03-17
Add `/vendor/bundle/` to `gitignore.tt`

joshuay03:

When contributing to (or authoring) open source gems, it's not uncommon to configure bundler to use `vendor/bundle` as the local gem directory. However, I've noticed that most gems do not have this path ignored in `.gitignore` by default, resulting in installed dependencies showing up as changed files unless `vendor/bundle` is `gitignore`'d.

Add `/vendor/bundle` to `.gitignore.tt` so new gems ignore locally installed dependencies by default. This already exists in [Github's Ruby `.gitignore` template](https://github.com/github/gitignore/blob/4488915eec0b3a45b5c63ead28f286819c0917de/Ruby.gitignore#L43).

### DIFF
--- a/bundler/lib/bundler/templates/newgem/gitignore.tt
+++ b/bundler/lib/bundler/templates/newgem/gitignore.tt
@@ -1,4 +1,5 @@
 /.bundle/
+/vendor/bundle/
 /.yardoc
 /_yardoc/
 /coverage/


indirect:
We recommend using `.bundle` rather than `vendor/bundle`. I believe that `.bundle` is already included in the `.gitignore`, isn't it?

joshuay03:
> We recommend using `.bundle` rather than `vendor/bundle`.

I see. I've personally only used `vendor/bundle`. I think I've picked up the habit while navigating in environments that use `bunde install --deployment` and liked the consistency in using the same setup locally as well.

>  I believe that `.bundle` is already included in the `.gitignore`, isn't it?

Yes it is. Is there a downside to including both?

simi:
@joshuay03 this file is used for gem skeleton and there is no configuration in gem skeleton to use `vendor/bundle`. The path of bundle install path is decided on client side and it can be anything and for that reason there is no way to cover all cases in default gitignore file.

joshuay03:
> @joshuay03 this file is used for gem skeleton and there is no configuration in gem skeleton to use `vendor/bundle`. The path of bundle install path is decided on client side and it can be anything and for that reason there is no way to cover all cases in default gitignore file.

That's fair. I understand that covering all cases doesn't make sense, but I was assuming that `vendor/bundle` is a fairly common path to configure. I'm happy to close this however, thanks for replying.

PR #7518, opened 2024-03-18
Respect global umask when writing regular files

deivid-rodriguez:

RubyGems does not respect global umask when writing regular files.

Always apply the default umask when writing.

This tries a more focused alternative to #7300, that hopefully plays nice with ruby/ruby CI.
 
### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -454,7 +454,7 @@ EOM
 
         if entry.file?
           File.open(destination, "wb") {|out| copy_stream(entry, out) }
-          FileUtils.chmod file_mode(entry.header.mode), destination
+          FileUtils.chmod file_mode(entry.header.mode) & ~File.umask, destination
         end
 
         verbose destination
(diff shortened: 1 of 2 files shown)


martinemde (approved):

martinemde:
@deivid-rodriguez Thank you for taking this on.

I looked more closely and I remember making changes to the tests in #7299 and #7300 (same changes in both) because there wasn't a test that would fail in the global permissions case. I'm concerned we won't have a test that covers the reason for this fix.

deivid-rodriguez:
@hsbt linked to ruby-core CI logs in https://github.com/rubygems/rubygems/pull/7417. In TravisCI both an old test and the new test failed. In ci.rvm.jp, only the new test failed.

Looking at the logs, I think the previous approach caused issues with directory permissions. Because of that, I decided to split the fix and focus for now on umask being respected for regular files.

My understanding is that my change, and the test that failed because of it, illustrate well what's being fixed here, and that changing the expectation of the failing test is enough. Maybe once we get this in (and ruby-core CI is good with it), we can add focus on directory permissions and try adding your extra test?

martinemde:
Ok, that sounds good to me. Thank you again @deivid-rodriguez

martinemde (approved):

PR #7494, opened 2024-03-01
Fix: vendor_gem takes a block

martinemde:
## What was the end-user or developer problem that led to this PR?

Fixes #7493

## What is your fix for the problem, implemented in this PR?

Pass the block through vendor_gem to save_gemspec.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7537, opened 2024-03-25
Fix resolver to properly intersect Arrays of `Gem::Resolver::Activation` objects

deivid-rodriguez:

Iterating on #7494. I think the issue is that:

* Resolver uses `Array#&` [here](https://github.com/rubygems/rubygems/blob/59bc67f40d6f98fb823d2356e5c2835a82a39447/lib/rubygems/vendor/molinillo/lib/molinillo/resolution.rb#L709), which compares items using `eql?`.
* `eql?` also needs an implementation of `hash` that guarantees that two equal objects return the same hash.
* The objects that are compared by the resolver are `Gem::Resolver::ActivationRequest` objects which implement `eql?` and `hash` [like this](https://github.com/rubygems/rubygems/blob/59bc67f40d6f98fb823d2356e5c2835a82a39447/lib/rubygems/resolver/activation_request.rb#L38-L45).
* In particular `hash` delegates to `@spec.hash`, which is usually a `Gem::Resolver::SpecSpecification` which does not override `hash` itself.

I think the above causes incorrect array intersection in JRuby & Truffleruby (I'd say it just works by chance in CRuby).

My attempted fix is to implement `Gem::Resolver::SpecSpecification#hash`.

### DIFF
--- a/lib/rubygems/resolver/spec_specification.rb
+++ b/lib/rubygems/resolver/spec_specification.rb
@@ -66,4 +66,11 @@ class Gem::Resolver::SpecSpecification < Gem::Resolver::Specification
   def version
     spec.version
   end
+
+  ##
+  # The hash value for this specification.
+
+  def hash
+    spec.hash
+  end
 end
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -1367,12 +1367,12 @@ Also, a list:
   #
   # Yields the +specification+ to the block, if given
 
-  def vendor_gem(name = "a", version = 1)
+  def vendor_gem(name = "a", version = 1, &block)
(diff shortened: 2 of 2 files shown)


martinemde (approved):
Looks like you really dug into this. Thank you for forgiving it out. It seemed like such a simple fix and I was surprised when it produced this failure.

deivid-rodriguez:
Yeah, I was really surprised when I saw the issue and associated PR because it felt really strange that both JRuby & TruffleRuby didn't like such a simple change.

In the end, the ruby-core change that caught this seems like a very good idea because it uncovered a real bug in our resolver that was supposed to have been tested but it actually was not because of the missing block.

PR #7539, opened 2024-03-25
Allow FormatError to take either String or Gem for source

jez:

Most of the calls to `FormatError.new` pass `@gem` for the second argument, which has a `path` method.

But in one case—on package.rb:691 in `verify_gz`, the `source` argument is a `String`.

So if there's ever a GZip decode error when attempting to read the contents of the `data.tar.gz` file, instead of reporting the underlying GZip error (which might be something like "unexpected end of file"), we would report instead a NoMethodError coming from package.rb

```
Exception while verifying sorbet-0.5.11301.gem
    undefined method `path' for "data.tar.gz":String

        @path = source.path
                      ^^^^^
```

There are two ways to fix this:

1. Make `FormatError#initialize` aware of the fact that `source` might sometimes be a `String`
2. Make the call to `FormatError.new` in `verify_gz` pass `@gem` instead of `entry.full_name`.

I've chosen 1 because I think it's more useful to see "unexpected end of file in data.tar.gz" instead of "unexpected end of file in sorbet-0.5.11301.gem." The end of file **is actually** in data.tar.gz, not in the gem file itself, which was decoded successfully.

I have tested this by forcibly untar'ing, corrupting, and re-tar'ing a `*.gem` file. Without this change, I see the NoMethodError on that tar file. With this change, I see the underlying gzip library's error. I've attached the tar file that I used for this test.
[test-case.tar.gz](https://github.com/rubygems/rubygems/files/14750347/test-case.tar.gz)

To reproduce:

```
tar xf test-case.tar.gz
gem push sorbet-0.5.11301.gem
```

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -59,7 +59,7 @@ class Gem::Package
 
     def initialize(message, source = nil)
       if source
-        @path = source.path
+        @path = source.is_a?(String) ? source : source.path
 
         message += " in #{path}" if path
       end


jez:
Another point here:

I [used sed](https://github.com/sorbet/sorbet/pull/7782/files#diff-54e25afd8f5d3edc378b93ad0eac18f6a751454cce518692c708258969b66e7aR37) to apply this patch to Sorbet's CI, and you can see what the behavior looks like when there's a GZip error in our CI pipeline:

https://buildkite.com/sorbet/sorbet/builds/30749#018e7e55-afe3-458a-93eb-3652dca5f399/127-161

jez:
I'm also attaching another `*.gem` archive that exhibits `undefined method path` error as an example
[crc-error.tar.gz](https://github.com/rubygems/rubygems/files/14779524/crc-error.tar.gz)

deivid-rodriguez (approved):
This seems like a strict improvement to me. Did you try to add a test? Not sure how hard it'd be, @martinemde knows this part of the code more, maybe has insights on that.

jez:
Hey I thought about writing a test, but when I went to write one, I saw that one with a similar name already existed, and it wasn't clear to me why it didn't catch this bug

https://github.com/jez/rubygems/blob/master/test/rubygems/test_gem_package.rb#L1027

So then I gave up and left the PR untested.

martinemde (approved):
Thank you for pushing this upstream and for the thorough investigation. Let's merge as is.

Unrelated to merging this, if anyone feels up for it, my slight preference is for the error to say "unexpected end of file in data.tar.gz in sorbet-0.5.11301.gem." Or something similar. It makes tracking down the error easier.

(This isn't to say we should not have a test for this failure, but that I'd rather the fix go in than get buried and follow up with a test using a similar corrupt gem.)

jez:
Thanks!

PR #7543, opened 2024-03-29
Allow string keys with gemrc

hsbt:

I sometimes confused why it's not working correctly with the following `gemrc`.

```
gem: "--no-document"
verbose: false
```

The correct case is `:verbose:` for that key. It's not useful for users.

I convert them to symbol keys if specified at gemrc.

### DIFF
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -210,6 +210,17 @@ class Gem::ConfigFile
       @hash = @hash.merge environment_config
     end
 
+    @hash.transform_keys! do |k|
+      # gemhome and gempath are not working with symbol keys
+      if %w[backtrace bulk_threshold verbose update_sources cert_expiration_length_days
+            install_extension_in_lib ipv4_fallback_enabled sources disable_default_gem_server
+            ssl_verify_mode ssl_ca_cert ssl_client_cert].include?(k)
+        k.to_sym
+      else
+        k
+      end
+    end
+
     # HACK: these override command-line args, which is bad
     @backtrace                   = @hash[:backtrace]                   if @hash.key? :backtrace
     @bulk_threshold              = @hash[:bulk_threshold]              if @hash.key? :bulk_threshold
--- a/lib/rubygems/config_file.rb
+++ b/lib/rubygems/config_file.rb
@@ -213,19 +213,20 @@ class Gem::ConfigFile
     # HACK: these override command-line args, which is bad
     @backtrace                   = @hash[:backtrace]                   if @hash.key? :backtrace
     @bulk_threshold              = @hash[:bulk_threshold]              if @hash.key? :bulk_threshold
-    @home                        = @hash[:gemhome]                     if @hash.key? :gemhome
-    @path                        = @hash[:gempath]                     if @hash.key? :gempath
-    @update_sources              = @hash[:update_sources]              if @hash.key? :update_sources
     @verbose                     = @hash[:verbose]                     if @hash.key? :verbose
-    @disable_default_gem_server  = @hash[:disable_default_gem_server]  if @hash.key? :disable_default_gem_server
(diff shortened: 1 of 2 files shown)


lib/rubygems/config_file.rb
```diff
@@ -210,22 +210,32 @@ def initialize(args)
...
     @backtrace                   = @hash[:backtrace]                   if @hash.key? :backtrace
     @bulk_threshold              = @hash[:bulk_threshold]              if @hash.key? :bulk_threshold
-    @home                        = @hash[:gemhome]                     if @hash.key? :gemhome
-    @path                        = @hash[:gempath]                     if @hash.key? :gempath
-    @update_sources              = @hash[:update_sources]              if @hash.key? :update_sources
     @verbose                     = @hash[:verbose]                     if @hash.key? :verbose
-    @disable_default_gem_server  = @hash[:disable_default_gem_server]  if @hash.key? :disable_default_gem_server
-    @sources                     = @hash[:sources]                     if @hash.key? :sources
+    @update_sources              = @hash[:update_sources]              if @hash.key? :update_sources
+    # concurrent_downloads
```
deivid-rodriguez, line 227: Is the above comment intentional?

lib/rubygems/config_file.rb
```diff
@@ -210,22 +210,32 @@ def initialize(args)
       @hash = @hash.merge environment_config
     end
 
+    @hash.transform_keys! do |k|
+      if %w[backtrace bulk_threshold verbose update_sources cert_expiration_length_days
+            install_extension_in_lib ipv4_fallback_enabled].include?(k)
```
deivid-rodriguez, line 215: How did you choose the list of configuration keys? I see there are a few more than this.

deivid-rodriguez (commented):
Great idea, I always found this very confusing.

lib/rubygems/config_file.rb
hsbt: Yes. I'm not sure why `concurrent_download` is not handled this place. We should fix that for `Gem.configuration.concurrent_downloads`.

But It's not this PR scope.

lib/rubygems/config_file.rb
hsbt: Make sensed. It's not strong reason.

I update that to deny list.

lib/rubygems/config_file.rb
deivid-rodriguez: So basically this comment is acting as a TODO note? If that's the case, I'd make that a bit more clear like `TODO: Review why concurrent_downloads is not handled here`.

lib/rubygems/config_file.rb
```diff
@@ -210,22 +210,32 @@ def initialize(args)
       @hash = @hash.merge environment_config
     end
 
+    @hash.transform_keys! do |k|
+      # These keys didn't work as symbols
+      if %w[gem install gemhome gempath].include?(k)
```
deivid-rodriguez, line 215: From the comment in lines 15-16 in this file:

> gemrc is a YAML file that uses strings to match gem command arguments and
> symbols to match RubyGems options.

My understanding is that right now, confusingly, you can specify any command name (or  `gem` for all commands), and that's specified as string. For RubyGems options, they are specified as keys.

So, I think we should:

* Include all commands here, not just `install`, otherwise configurations for other commands are going to be symbolized and possible break?
* Remove or reword the comment above to no longer document this inconsistency.

Makes sense?

lib/rubygems/config_file.rb
hsbt: I see. I did update that.

lib/rubygems/config_file.rb
hsbt: I changed my proposal. I only convert `Gem::ConfigFile` accessors. I don't want to work this simple issue to `Gem.configuration` entirely.

Issue #7540, opened 2024-03-26
Bundle updates are moving dependencies from global to secondary source


agrobbin:
Bundler is moving a bunch of dependencies from our primary source (`https://rubygems.org/`) within `Gemfile.lock` to a secondary source, when there is an overlap of sub-dependencies between sources. That secondary source includes a gem that has one of the same sub-dependencies as the dependencies that were moved, which I _think_ is the likely cause.

![image](https://github.com/rubygems/rubygems/assets/46724/a751a15d-b715-4bc0-b662-e79a88cdc065)

The removals are within the `https://rubygems.org/` source, while the additions are within a private source (in our case, a GItHub Packages RubyGems registry).

```ruby
source 'https://rubygems.org'

gem 'rails'

source 'https://rubygems.pkg.github.com/[redacted]' do
  gem 'foo' # <= some private dependency that includes an overlapping dependency with `rails`, such as `activesupport`
end
```

I originally thought this was a Dependabot-specific issue (dependabot/dependabot-core#9326), but I've now seen it happen when doing gem updates locally, so I think it's actually a Bundler issue.

This has been happening at least since Bundler v2.5.4, and still on v2.5.6.

deivid-rodriguez:
Does the secondary source include `activesupport` and the other gems that were moved? If those exist in the non global source, Bundler does prefer that, although I'm not sure it should be changing the lockfile if they're already locked in the global source section.

agrobbin:
Here is a snippet of the secondary source in the `Gemfile.lock`, when no additional shared dependencies are included:

```
GEM
  remote: https://rubygems.pkg.github.com/[redacted]
  specs:
    [redacted] (1.4.0)
      activesupport (>= 5.2, < 7.2)
      google-cloud-pubsub (~> 2.0)
      hc_metrics (~> 0.3)
      sidekiq (>= 5.0, < 8.0)
    [redacted] (0.4.1)
    [redacted] (0.1.1)
      iterable-api-client (~> 0.5)
      mail (~> 2.8)
```

The secondary source does include `activesupport` as a dependency, but it is not actually stored in the secondary source repository itself (at least, it doesn't show up in GitHub's Packages list!).

Let me know if I can provide any other details here to help look into the issue.

deivid-rodriguez:
Yeah, I meant if there was a gem called `activesupport` stored in the secondary source. If there's no such gem, then this seems like a bug, yeah. Very intriguing.

[...]

PR #7544, opened 2024-03-29
add test case to ensure updating with multiple sources + caching maintains the right lockfile

agrobbin:
@deivid-rodriguez as requested in #7540, here's my test case to ensure `bundle update` with caching enabled and multiple sources creates the correct lockfile.


deivid-rodriguez (approved):
Thank you!

PR #7557, opened 2024-04-04
Add `--glob` flag to `bundle add`

thedavemarshall:
Bundler online documentation says that if the gem is located within a subdirectory of a git repository, you can use the `:glob` option to specify the location of its .gemspec, [docs](https://bundler.io/guides/git.html)

`gem 'cf-copilot', git: 'https://github.com/cloudfoundry/copilot', glob: 'sdk/ruby/*.gemspec'`

This change allows for equivalent functionality from the bundler CLI

`bundle add cf-copilot --git=https://github.com/cloudfoundry/copilot --glob='sdk/ruby/*.gemspec'`

I'm working on some projects that require gems from remote git sources, and sometimes these projects don't specify the `.gemspec` file at the root of the repo. Instead of manually editing the `Gemfile`, I'd like to use bundler's CLI where possible.

My fix is to pass-through the `glob` option from `Bundler::CLI#add(*gems)`, to `Bundler::Dependency#glob` to be used in `Bundler::Injector#build_gem_lines()`

### DIFF
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -347,6 +347,7 @@ module Bundler
     method_option "github", type: :string
     method_option "branch", type: :string
     method_option "ref", type: :string
+    method_option "glob", type: :string
     method_option "skip-install", type: :boolean, banner: "Adds gem to the Gemfile but does not install it"
     method_option "optimistic", type: :boolean, banner: "Adds optimistic declaration of version to gem"
     method_option "strict", type: :boolean, banner: "Adds strict declaration of version to gem"
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -347,7 +347,7 @@ module Bundler
     method_option "github", type: :string
     method_option "branch", type: :string
     method_option "ref", type: :string
-    method_option "glob", type: :string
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (approved):
I don't think this option gets a lot of use, but seems nice to have all options mapped to `bundle add` flags :+1:

I think people will need to run this with quotes, otherwise most shells will automatically expand `*` if run from a directory with gemspecs matching the glob. But there's not much we can do about that. Maybe change tests examples to use quotes, explicitly document it? Not sure.

thedavemarshall:
@deivid-rodriguez good points, thanks for the feedback!

> I think people will need to run this with quotes, otherwise most shells will automatically expand `*` if run from a directory with gemspecs matching the glob. But there's not much we can do about that. Maybe change tests examples to use quotes, explicitly document it? Not sure.

I've changed the specs to use single quotes around the `--glob` value when invoking `bundle add`, and I've added `banner:` text for the CLI explaining that the globs will be expanded with Ruby and recommending single quotes.

deivid-rodriguez:
Nice, thanks! This looks good to me! There's a failed spec unrelated to your changes, don't worry about that. I'll merge this next week :+1:

deivid-rodriguez (approved):
Thanks!

Issue #7176, opened 2023-11-20
`bundle update GEMNAME` fails to use latest version


andyw8:
### Describe the problem as clearly as you can

`bundle update GEMNAME` is not installing the latest version of a dependency, even when the dependencies are satisfied.

Demo repo: https://github.com/andyw8/temp-bundle-update-issue

### Did you try upgrading rubygems & bundler?

Yes:
```
% gem -v; bundle -v
3.4.22
Bundler version 2.4.22
```

### Which command did you run?

`bundle update tapioca`

### What were you expecting to happen?

Should update `tapioca` to `v0.11.12`.

### What actually happened?

`tapioca` version in `Gemfile.lock` did not change, remains at `v0.11.10`.

If I explicitly specify `gem "tapioca", "= 0.11.12"` and run `bundle update tapioca` then it updates successfully.

## Environment

```
Bundler       2.4.22
  Platforms   ruby, arm64-darwin-21
Ruby          3.1.1p18 (2022-02-18 revision 53f5fc4236a754ddf94b20dbb70ab63bd5109b18) [arm64-darwin-21]
  Full Path   /opt/rubies/3.1.1/bin/ruby
  Config Dir  /opt/rubies/3.1.1/etc
RubyGems      3.4.22
  Gem Home    /Users/andyw8/.gem/ruby/3.1.1
  Gem Path    /Users/andyw8/.gem/ruby/3.1.1:/opt/rubies/3.1.1/lib/ruby/gems/3.1.0
  User Home   /Users/andyw8
  User Path   /Users/andyw8/.gem/ruby/3.1.0
  Bin Dir     /Users/andyw8/.gem/ruby/3.1.1/bin
Tools
  Git         2.41.0
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

## Bundler Build Metadata

```
Built At          2023-11-09
Git SHA           ec2089640
Released Version  true
```

## Bundler settings

```
gem.changelog
  Set for the current user (/Users/andyw8/.bundle/config): false
gem.ci
  Set for the current user (/Users/andyw8/.bundle/config): "github"
gem.coc
  Set for the current user (/Users/andyw8/.bundle/config): false
gem.linter
  Set for the current user (/Users/andyw8/.bundle/config): "rubocop"
gem.mit
  Set for the current user (/Users/andyw8/.bundle/config): true
gem.test
  Set for the current user (/Users/andyw8/.bundle/config): "minitest"
pkgs.shopify.io
[...]

Issue #5463, opened 2022-04-10
Allow easier upgrade of Rails components (and transitive dependencies in general)


deivid-rodriguez:
### Describe the problem as clearly as you can

When Rails release a security fix in, say, `activerecord`, Bundler users are not able to pickup the fix by doing `bundle update activerecord`, if they have a `Gemfile.lock` file locked to an insecure version.

This is because `bundle update activerecord` only "unlocks" (ignores locked version in the `Gemfile.lock` file) the `activerecord` gem and its dependencies. However, gems that have `activerecord` as a dependency, namely `rails`, are not unlocked and thus stay at the previous insecure version, making the upgrade impossible.

So `bundle update activerecord` ends up logging a message like

> Bundler attempted to update activerecord but its version stayed the same

I think if Bundler is unable to update the requested dependency, it could try to also unlock gems that have the target gem as a dependency, in order to try to make the upgrade succeed.

### Did you try upgrading rubygems & bundler?

Yes.

#### Gemfile

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

gem "rails"
```

#### Gemfile.lock

```
GEM
  remote: https://rubygems.org/
  specs:
    actioncable (7.0.2.2)
      actionpack (= 7.0.2.2)
      activesupport (= 7.0.2.2)
      nio4r (~> 2.0)
      websocket-driver (>= 0.6.1)
    actionmailbox (7.0.2.2)
      actionpack (= 7.0.2.2)
      activejob (= 7.0.2.2)
      activerecord (= 7.0.2.2)
      activestorage (= 7.0.2.2)
      activesupport (= 7.0.2.2)
      mail (>= 2.7.1)
      net-imap
      net-pop
      net-smtp
    actionmailer (7.0.2.2)
      actionpack (= 7.0.2.2)
      actionview (= 7.0.2.2)
      activejob (= 7.0.2.2)
      activesupport (= 7.0.2.2)
      mail (~> 2.5, >= 2.5.4)
      net-imap
      net-pop
      net-smtp
      rails-dom-testing (~> 2.0)
    actionpack (7.0.2.2)
      actionview (= 7.0.2.2)
      activesupport (= 7.0.2.2)
      rack (~> 2.0, >= 2.2.0)
      rack-test (>= 0.6.3)
      rails-dom-testing (~> 2.0)
      rails-html-sanitizer (~> 1.0, >= 1.2.0)
    actiontext (7.0.2.2)
[...]

Issue #3310, opened 2019-08-27
"bundle update" doesn't update gem to latest patch version


leomao10:
Here is the repo I created to reproduce the issue:
https://github.com/leomao10/test_bundler

## What happen
When I use `bundle update stackdriver`, it only update its depending gem `google-cloud-logging` to latest minor version which is `1.6.0`, but not the latest patch version which is `1.6.7`. 

Run `bundle update stackdriver` again will update `google-cloud-logging` to latest patch version.

Also, remove `jbuilder` from the Gemfile and run `bundle update stackdriver`  would upgrade
`google-cloud-logging` straight to the latest patch level(`1.6.7`)

I assume `jbuilder` somehow constraining `google-cloud-logging` to upgrade to latest patch level, but I could not spot the constraint when checking `Gemfile.lock`.

## Expected Behavior
`bundle update stackdriver` would update all depending gems to latest patch level.

## Affected version:
I tested it in bundler `1.6.7`, `1.7.2` under ruby `2.5.1`  and also `2.0.2` under ruby `2.6.3`.

deivid-rodriguez:
Hei!

So the underlying problem here is that `bundler` is not properly unlocking the `concurrent-ruby` dependency.

In the first lockfile, `concurrent-ruby` is not a dependency of the gem being updated (`stackdriver`) or any of its locked dependencies. So `bundler` is not unlocking it, and keeping the `1.0.5` version locked in the `Gemfile.lock` file instead. The newest version of `google-cloud-logging` that satisfies that is [1.6.0](https://rubygems.org/gems/google-cloud-logging/versions/1.6.0), since [1.6.1](https://rubygems.org/gems/google-cloud-logging/versions/1.6.1) depends on `concurrent-ruby ~> 1.1`. That's why bundler is stopping there.

The second time you run `bundle update`, the lockfile has been updated and now `concurrent-ruby` is a transitive dependency of `stackdriver` (because some updated dependecies of `stackdriver` started depending on it). So now `concurrent-ruby` is properly unlocked and updating all the way to `google-cloud-logging` is now possible.

The problem here is that `bundler` is only unlocking the specified gem and its _current_ dependencies (the dependencies of the currently locked version). Instead, it should unlock the specified gem and its _current and future_ dependencies (the dependencies of the current locked version and higher versions of it).

I'll try to provide a fix for this.

PR #5520, opened 2022-05-09
Automatically unlock dependants released in lockstep

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Sometimes users might run `bundle update <transitive-dependency>`, for example, because `<transitive-dependency>` has a security issue and they want to pick up a release with a fix. This will only work if there are no dependents holding the previous version. A very common case where it does not work is where dependants are released in lockstep with the gem being updated.

## What is your fix for the problem, implemented in this PR?

Smartly detecting dependants holding updates and retrying resolution with those unlocked is a general improvement that could be worth adding to the resolver. However, exact dependencies is a very easy case that can be detected in advance without having to retry at all.

Once dependencies to be unlocked have been figured out, have one more pass to check whether any gems have exact dependencies on them, and unlock those too.

This PR allows things like `bundle update actionview` just work, by updating the whole "Rails family" at once. It also makes it easier for dependabot to create automated PRs to fix this kind of issue.

Fixes #5463.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7558, opened 2024-04-04
Make sure `bundle update <specific_gems>` can always update to the latest resolvable version of each requested gem

deivid-rodriguez:

Sometimes `bundle update foo` won't actually update foo, even if a clean `bundle install` will resolve to a higher version of foo than the one currently locked.

The problem is that other locked dependencies may be setting some constraints on dependencies of foo, preventing it from being upgraded. These locked dependencies are not present on a clean `bundle install`.

My fix is, if `bundle update foo` is requested, and initial try did not update foo, try unlocking other gems, even if they are not dependencies of foo. Keep doing this until foo updates, or it's not possible to unlock more gems.

Fixes https://github.com/rubygems/rubygems/issues/3310.
Fixes https://github.com/rubygems/rubygems/issues/5463.
Fixes https://github.com/rubygems/rubygems/issues/7176.

Supersedes https://github.com/rubygems/rubygems/pull/5520.

### DIFF
--- a/bundler/lib/bundler/definition.rb
+++ b/bundler/lib/bundler/definition.rb
@@ -570,7 +570,9 @@ module Bundler
         last_resolve = converge_locked_specs
         remove_invalid_platforms!(current_dependencies)
         packages = Resolver::Base.new(source_requirements, expanded_dependencies, last_resolve, @platforms, locked_specs: @originally_locked_specs, unlock: @gems_to_unlock, prerelease: gem_version_promoter.pre?)
-        additional_base_requirements_for_resolve(packages, last_resolve)
+        packages = additional_base_requirements_to_prevent_downgrades(packages, last_resolve)
+        packages = additional_base_requirements_to_force_updates(packages)
+        packages
       end
     end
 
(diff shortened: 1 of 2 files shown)


bundler/lib/bundler/definition.rb
```diff
@@ -671,14 +674,18 @@ def add_current_platform
...
       if unlocking?
-        unlock_reason = @unlock.reject {|_k, v| Array(v).empty? }.map do |k, v|
-          if v == true
-            k.to_s
-          else
-            v = Array(v)
-            "#{k}: (#{v.join(", ")})"
-          end
-        end.join(", ")
+        stuff_to_unlock = if @gems_to_unlock.any?
```
simi, line 677: _:smiley_cat:  on the naming, but I have no better idea :fearful:_

simi (approved):

bundler/lib/bundler/definition.rb
deivid-rodriguez: Haha, yeah, maybe `unlock_targets`?

deivid-rodriguez: I went with `unlock_targets` since it seemed a bit more "formal" 😅

simi: sounds good

Issue #7185, opened 2023-11-22
Source credentials present in both configuration and the Gemfile are removed when writing the lock file


tt:
### Describe the problem as clearly as you can

If you have a Gemfile with a `source` block or a `:source` option and that remote contains credentials, having the same credentials set in your Bundler configuration file will cause `bundle install` to remove the credentials from the lock file which will then prevent `bundle exec` from finding the gem.

Turning on `allow_deployment_source_credential_changes` can be used as a workaround but as far as I interpret the documentation, that is not its intended behavior. (I'm not running in deployment mode and credentials are removed rather than added.)

### Did you try upgrading rubygems & bundler?

Yes. I'm using the latest versions and have tried other versions to check for recent regressions.

### Post steps to reproduce the problem

I'm running Ruby 3.2.2 built with ruby-install on macOS 14.1.1. I'm using Bundler 2.4.22.

As the reproduction requires access to a protected gem server, I can't share complete instructions.

Any Gemfile with a block like below should allow reproducing the bug:

```ruby
source "https://username:password@example.com" do
  gem "gem"
end
```

You then need to set the credentials in your configuration file with `bundle config https://example.com username:password`.

Running `bundle install` will then cause the lock file to contain the following lines (notice the missing credentials):

```
GEM
  remote: https://example.com/
  specs:
    ...
```

### Which command did you run?

`bundle install --quiet && bundle exec irb`

### What were you expecting to happen?

I expected `irb` to launch.

I think there are two different approaches for how to solve this:

- `bundle install` could write credentials in the lock file if the credentials are present in the Gemfile even if they're also present in the Bundler configuration.
- `bundle exec` could remove configured credentials from sources before comparing them.

The first option seems best to me because it also prevents changes to the lock file.

### What actually happened?

`bundle exec` told me I had to first run `bundle install` which I had done and which completed successfully:

```shell
$ bundle install --quiet && bundle exec irb
Could not find gem '<gem>' in locally installed gems
Run `bundle install` to install missing gems.
```

This happens due to the code in `Bundler::Source::Rubygems#suppress_configured_credentials`:

https://github.com/rubygems/rubygems/blob/f0e8dacdecaa39aaee0b43450d92cc08f5b84d4c/bundler/lib/bundler/source/rubygems.rb#L358-L365

[...]

Issue #3234, opened 2018-04-16
Credentials are not properly filtered out in 'bundle install'


perlun:
rubygems/bundler#4704 fixed some of this, but apparently there still seems to be an issue under certain circumstances.

Here is my Gemfile, running with bundler `1.16.1`:

```ruby
git_source(:acme) { |repo_name| "https://#{ENV['GITHUB_TOKEN']}:x-oauth-basic@github.com/acme/#{repo_name}" }

gem 'foobar', acme: 'foobar'
```

When I run `GITHUB_TOKEN=my-secret-token bundle install`, this is part of the output I get:

```
Using foobar 0.1.0 from https://my-secret-token:x-oauth-basic@github.com/acme/foobar (at master@abc1234)
```

This is clearly less than optimal, especially if being used on Travis or other CI services. (In our case, we are using private repos so the criticality is "less severe" _for us_, but it's still an issue since potential credential leakage is never good.)

(Rationale for using https git sources rather than SSH - this is being used in Dockerfiles, where passing in an ENV variable for a GitHub OAuth token is somewhat more convenient than having to pass in a custom SSH key.)

perlun:
(Technically, this relates to a slightly different problem than the one fixed in #7560 if I get it right @deivid-rodriguez - it wasn't just that the private token was written to the lockfile, but it was also printed to stdout. Has this also been fixed by now?)

deivid-rodriguez:
Nice catch @perlun. Indeed my focus in #7560 was not to fix credentials being leaked to log messages, but to the lockfile. However, I suspect #7560 has fixed this problem too, because by the time the log message is printed, the lockfile source (without credentials) has already replaced the corresponding Gemfile source (with credentials). However, worth giving this a try and reopening if not really fixed.

perlun:
Sounds good @deivid-rodriguez. I am only a very marginal Ruby user these days, so if anyone else who is following this thread (@olleolleolle, do you copy?) has a chance to try, give it a go to see how it behaves now with the latest fixes in #7560.

Issue #1240, opened 2015-05-01
Gemfile.lock should NOT contain private credentials in source urls


joelpresence:
Hi,

I use GemFury for private gems.  In my `Gemfile` I have code like:
```
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end
```
The environment variable `FURY_AUTH` has value `_MY_SECRET_FURY_TOKEN_`.  When I run `bundle install` or `bundle update`, bundler updates `Gemfile.lock` to include:
```
# In Gemfile.lock
remote: https://_MY_SECRET_FURY_TOKEN_@gem.fury.io/devpresence/
```
This is a risk because now `_MY_SECRET_FURY_TOKEN_` risks getting committed to source control.  I took great care to *not* list my secret token in plain text in `Gemfile` (e.g. by specifying it as an env var) so bundler/gems should not now just write the secret token to the `Gemfile.lock`.

When I deploy this, say to Heroku, I further specify an env var as in:
```
bash> heroku config:set FURY_AUTH=_MY_SECRET_FURY_TOKEN_
```
so even Heroku deploys don't contain my secret token (except in private runtime config).

Further, when I run tests, bundler seems to periodically update `Gemfile.lock`.  So even if I manually edit `Gemfile.lock` to remove `_MY_SECRET_FURY_TOKEN_`, bundler seems to periodically re-add it, without me even running bundler manually (maybe rspec is doing this?).

I would propose that if a source url in the Gemfile contains valid HTTP credentials, and the bundler-created `Gemfile.lock` remote url contains credentials as in:
```
remote: https://token@host/path
remote: https://user:password@host/path
```
then bundler should automatically strip these when it's creating the `Gemfile.lock` from `Gemfile`.

So
```
# CASE 1: CURRENT BEHAVIOR
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end

# Becomes in Gemfile.lock
remote: https://_MY_SECRET_FURY_TOKEN_@gem.fury.io/devpresence/

# CASE 2: DESIRED BEHAVIOR
# In Gemfile
source "https://#{ENV['FURY_AUTH']}@gem.fury.io/devpresence/" do
  gem 'presence_common_web', '~> 0.1.4'
end

# Becomes in Gemfile.lock
[...]

PR #7560, opened 2024-04-05
Never write credentials to lockfiles

deivid-rodriguez:

Bundler will sometimes write credentials to lockfiles, which is NOT very secure since, for example, they may get pushed to a public repo by mistake.

Never write credentials to lockfiles.

Previously, if they were present, they'd be considered for lockfile expiration. I think it's fine to never consider them and never write them in the lockfile. If there are credentials present in the Gemfile (through ENV, for example), complete lockfile sources with those so that deployment mode (that only considers the lockfile) still works.

There was previously a setting to enable this behavior but I think it should be the default, so I removed the setting.

Fixes https://github.com/rubygems/rubygems/issues/1240.
Fixes https://github.com/rubygems/rubygems/issues/3234.
Closes https://github.com/rubygems/rubygems/issues/7185.

### DIFF
--- a/bundler/lib/bundler/man/bundle-add.1
+++ b/bundler/lib/bundler/man/bundle-add.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-ADD" "1" "March 2024" ""
+.TH "BUNDLE\-ADD" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-add\fR \- Add gem to the Gemfile and run bundle install
 .SH "SYNOPSIS"
--- a/bundler/lib/bundler/man/bundle-binstubs.1
+++ b/bundler/lib/bundler/man/bundle-binstubs.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-BINSTUBS" "1" "March 2024" ""
+.TH "BUNDLE\-BINSTUBS" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-binstubs\fR \- Install the binstubs of the listed gems
 .SH "SYNOPSIS"
--- a/bundler/lib/bundler/man/bundle-cache.1
+++ b/bundler/lib/bundler/man/bundle-cache.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-CACHE" "1" "March 2024" ""
+.TH "BUNDLE\-CACHE" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-cache\fR \- Package your needed \fB\.gem\fR files into your application
 .SH "SYNOPSIS"
--- a/bundler/lib/bundler/man/bundle-check.1
+++ b/bundler/lib/bundler/man/bundle-check.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-CHECK" "1" "March 2024" ""
+.TH "BUNDLE\-CHECK" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-check\fR \- Verifies if dependencies are satisfied by installed gems
 .SH "SYNOPSIS"
--- a/bundler/lib/bundler/man/bundle-clean.1
+++ b/bundler/lib/bundler/man/bundle-clean.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-CLEAN" "1" "March 2024" ""
+.TH "BUNDLE\-CLEAN" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-clean\fR \- Cleans up unused gems in your bundler directory
 .SH "SYNOPSIS"
--- a/bundler/lib/bundler/man/bundle-config.1
+++ b/bundler/lib/bundler/man/bundle-config.1
@@ -1,6 +1,6 @@
 .\" generated with nRonn/v0.11.1
 .\" https://github.com/n-ronn/nronn/tree/0.11.1
-.TH "BUNDLE\-CONFIG" "1" "March 2024" ""
+.TH "BUNDLE\-CONFIG" "1" "April 2024" ""
 .SH "NAME"
 \fBbundle\-config\fR \- Set bundler configuration options
 .SH "SYNOPSIS"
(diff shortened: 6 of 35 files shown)


technicalpickles:
I'm still gathering information and getting a reproducible case, but I believe this change is causing problems during CI. The `bundle install` succeeds, but then we get missing gems when actually trying to activate, similar to the errors on https://github.com/rubygems/rubygems/issues/7185 ie

```
Could not find gem 'our-gem' in locally installed gems
Run `bundle install` to install missing gems.
```

Where `our-gem` uses a `source` with a credential in the gemfile.

deivid-rodriguez:
Thanks letting me know so quickly. I will try to figure out on Monday what's happening here and and make a patch release with a revert if needed.

gdubicki:
Hi @deivid-rodriguez!

Thanks for your work on this but I think we see an issue with it too: when using Bundler 2.5.10, although we DO use credentials in our Gemfile for some gems, with a syntax like:

```ruby
source "https://<redacted>@gems.contribsys.com/" do
  gem "sidekiq-pro", "~> 7.2"
end
```

...in the generated Gemfile.lock the credentials were stripped.

deivid-rodriguez:
That's exactly the feature implemented by this PR? Did it cause any issues or is it that you actually want credentials written to the lockfile?

gdubicki:
> That's exactly the feature implemented by this PR. Did it cause any issues or is it that you actually want credentials written to the lockfile?

Oops, sorry! I actually just assumed that it would cause an issue. After actually testing it, I have none. So nvm!

deivid-rodriguez:
@technicalpickles Since we haven't got any more reports, I'm actually holding on the revert, but I'll wait till you find out more and investigate then.

deivid-rodriguez:
@technicalpickles Can you check if the lockfile is correct after `bundle install` (same as before, except it's removing credentials from it) or if it is somehow mis-assigning gems to incorrect sources?

ngan:
Hey @deivid-rodriguez, our Gemfile.lock diff only contains removal of credentials and bundler version change.

I spent a few minutes trying to reproduce this locally but couldn’t. The only difference between local and CI is that we use deployment mode. I’m going to give that a try next when I get a chance.

nertzy:
@deivid-rodriguez I have reproduced the issue.

Our application uses Karafka Pro, following the setup from @mensfeld in the [Karafka docs](https://karafka.io/docs/Pro-Getting-Started/):
```ruby
source 'https://USERNAME:PASSWORD@gems.karafka.io' do
  gem 'karafka-license', 'LICENSE_ID'
end
```


We upgraded our Bundler version via `bundle update --bundler` and checked in the resulting diff to the `Gemfile.lock` that updated the Bundler version and removed those credentials.

When deploying to Heroku, we see:
```
-----> Installing dependencies using bundler 2.5.6
       Running: BUNDLE_WITHOUT='development:test' BUNDLE_PATH=vendor/bundle BUNDLE_BIN=vendor/bundle/bin BUNDLE_DEPLOYMENT=1 bundle install -j4
       The list of sources changed, but the lockfile can't be updated because frozen
       mode is set
       
       * karafka-license from `rubygems repository [https://gems.karafka.io/`](https://gems.karafka.io/%60) to
       `rubygems repository ***gems.karafka.io/`
       
       Run `bundle install` elsewhere and add the updated Gemfile to version control.
       If this is a development machine, remove the Gemfile.lock freeze by running
       `bundle config set frozen false`.
       Bundler Output: The list of sources changed, but the lockfile can't be updated because frozen
       mode is set
       
       * karafka-license from `rubygems repository [https://gems.karafka.io/`](https://gems.karafka.io/%60) to
       `rubygems repository ***gems.karafka.io/`
       
       Run `bundle install` elsewhere and add the updated Gemfile to version control.
       If this is a development machine, remove the Gemfile.lock freeze by running
       `bundle config set frozen false`.

 !
 !     Failed to install gems via Bundler.
 !
 !     Push rejected, failed to compile Ruby app.

 !     Push failed
 ▸    Build failed
```

For now I will roll back to our previous Bundler version as a workaround.

nertzy:
I noticed there is a backtick getting escaped in the output `https://gems.karafka.io/%60` so I wonder if that is related to the underlying issue?

deivid-rodriguez:
Oh your logs made me realize.

Normally a lockfile generated with Bundle x.y.z will be bundled with Bundler x.y.z. Bundler has builtin behavior to ensure that. If that happens things work fine.

However, Heroku is special and will forcefully use its own version (2.5.6 at the moment, apparently) and also does not provide any options to customize that. If Bundler older than 2.5.10 bundles a 2.5.10 lockfile that stripped credentials, and does that in "frozen mode", which Heroku also enables by default, then indeed things won't work.

If you run `bundle config --local allow_deployment_source_credential_changes true` and commit the resulting configuration, does that make things work again?

deivid-rodriguez:
Or if that's easier, set the `BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES` env variable to `true`.

nertzy:
> Or if that's easier, set the `BUNDLE_ALLOW_DEPLOYMENT_SOURCE_CREDENTIAL_CHANGES` env variable to `true`.

That didn't work, because that variable only has an effect in Bundler 2.5.10 and later, and we're still given 2.5.6 by the Heroku buildpack, if I understand correctly.

I think another workaround would be to explicitly set the credentials via `BUNDLE_GEMS__KARAFKA__IO` (or its equivalent).

deivid-rodriguez:
> That didn't work, because that variable only has an effect in Bundler 2.5.10 and later, and we're still given 2.5.6 by the Heroku buildpack, if I understand correctly.

Oh, too bad. That variable was present in 2.5.6 and I suggested it because I thought the reporter of https://github.com/rubygems/rubygems/issues/7185 suggested that it had the effect of preventing credentials to be added, but now I see it's actually the other way around (it forced credentials to be added when credentials were specified in configuration).

> I think another workaround would be to explicitly set the credentials via BUNDLE_GEMS__KARAFKA__IO (or its equivalent).

Yes, that'd be the preferred workaround, don't reference credentials directly in Gemfile either 👍.

veganstraightedge:
Hi 👋🏻 while I understand and support the idea of keeping credentials out of checked in lockfiles, it feels to me that changing the behavior and format of the lockfile is a _major_ change. A contract breaking change.

IMO, this kind of change should be warned about with deprecation warnings in a minor version first. Then in only in a major version, make this change. 

This change broke all Dependabot PRs with private gems using this pattern:

```ruby
# Gemfile
source "https:/USERNAME:PASSWORD@example.com" do
  gem "private_gem"
end
```

For example, [Sidekiq Enterprise](https://github.com/rubygems/rubygems/issues/7715).

Can this change be urgently reverted please? 
And then rolled out with deprecation warnings and major version release please? 

🖤

cc: @deivid-rodriguez

deivid-rodriguez:
Hello. Thanks for the ping.

This was supposed to change only a lockfile detail, but I was not expecting it to break anything other than that. I don't consider a somewhat surprising diff in a lockfile to be a breaking change.

That said, it sounds like this has caused real problems, as reported in #7715. That's of course not ok. I will investigate that issue and decide what to do next.

broksonic21:
As a user, my thoughts:

in #7715, they are running an older bundler vs a 2.5.11 generated lockfile. Like the dependabot related tickets, that's ultimately what I am noticing too -> anytime you use a 2.5.10+ bundler (whether locally or via a tool like dependabot), it writes out a syntax that doesn't work if you also use a lower-bundler for any reason (heroku, vendor'd, etc)

In that sense, this is a breaking change, but I also think there's a bit of user (or tooling issues) - when you upgrade bundler in one spot, you should update in all.

That's hard, I admit, if:
* Heroku won't respect the bundler version in the Gemfile.lock - rubygems has code to autoupdate you locally on recent rubygems, so unfortunate Heroku doesn't
* Dependabot like tools should respect the Gemfile.lock listed bundler version. I know dependabot just updated to use 2.5.11 always, which helps _my case_ since we want to use 2.5.10+ bundlers with the security fix + without constant lockfile churn by dependabot putting creds back in if older bundler and local devs taking it back out with the newer bundler, but probably harms the _reverse_ case

deivid-rodriguez:
Thanks for the helpful feedback @broksonic21. I agree. I thought only issues were with Heroku and Dependabot, which in my opinion should eventually respect auto-switching. I also thought issues were only about constant lockfile churn from Dependabot/Heroku and local Bundler competing to add & remove credentials, and I expected those to be gone once both tools upgrade their internal version of Bundler.

However, after thinking more about this today and looking at the new issue opened, I realize there are more cases where Bundler won't autoswitch (expectedly, or unexpectedly), and things can break in that case. Also I realize there are way more people putting credentials in their Gemfiles than I thought 😅.

So I think we should restore a better default behavior that keeps credentials in lockfile if they are already there. Working on it!

deivid-rodriguez:
I created two PRs to alleviate these issues:

* https://github.com/rubygems/rubygems/pull/7719 will make Bundler switch versions properly when using binstubs, preventing #7715 from happening.
* https://github.com/rubygems/rubygems/pull/7720 keeps credentials in lockfiles that already had credentials in them, allowing them to keep working on older versions of Bundler.

As I follow up change, I will add deprecation warnings when writing lockfiles with credentials, so that people can gradually remove credentials from them.

PR #7561, opened 2024-04-05
Add missing ostruct require until upstream fixes it

deivid-rodriguez:

json dropped its dependency on ostruct in ruby-core. That surfaced that turbo_tests was using ostruct without requiring it first, since it was their json dependency actually doing that.

That broke our daily bundler specs. 

Add the missing require ourselves, for now.

### DIFF
--- a/bundler/bin/parallel_rspec
+++ b/bundler/bin/parallel_rspec
@@ -4,6 +4,8 @@
 require_relative "../spec/support/rubygems_ext"
 
 require_relative "../spec/support/switch_rubygems"
+
+require "ostruct" # Backfill missing require in turbo_tests. Remove when added upstream
 require "turbo_tests"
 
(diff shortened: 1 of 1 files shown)


bundler/bin/parallel_rspec
```diff
@@ -4,6 +4,8 @@
 require_relative "../spec/support/rubygems_ext"
 
 require_relative "../spec/support/switch_rubygems"
+
+require "ostruct" # Backfill missing require in turbo_tests. Remove when added upstream
```
ilyazub, line 8: This is fixed in `turbo_tests` v2.2.1 (ref: https://github.com/serpapi/turbo_tests/pull/49#issuecomment-2046576834).

I'm sorry about this issue.

deivid-rodriguez: No problem at all! Thanks for the quick release!

PR #7563, opened 2024-04-06
Honor a specified path as the temporary diretory if given

nobu:
## The problem

Currently the tests are executed in the fixed name directory "tmp" under the top source directory.  However it makes the tests fail when the source path contains symlinks.  Or unable to even start if the top source directory is read-only, e.g., remote-mounting in read-only mode from virtual machines.

Also, with the fixed directory, there is no way to avoid `pend` in `TestGemPackage#test_extract_symlink_parent_doesnt_delete_user_dir`.

## The fix

Instead of creating the fixed name directory, this PR allows to use a different path given with the environment variable "GEM_TEST_TMPDIR". The default path is, as well as the current behavior, "tmp" from the top source directory.

### Caveat

It is the caller's responsibility to make the directory safe (owned, world unwritable, or sticky) when setting the environment variable.

### DIFF
--- a/test/rubygems/helper.rb
+++ b/test/rubygems/helper.rb
@@ -285,9 +285,12 @@ class Gem::TestCase < Test::Unit::TestCase
   def setup
     @orig_hooks = {}
     @orig_env = ENV.to_hash
-    @tmp = File.expand_path("../../tmp", __dir__)
 
-    FileUtils.mkdir_p @tmp
+    top_srcdir = __dir__ + "/../.."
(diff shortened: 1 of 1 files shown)


nobu:
Thanks, of course!

nobu:
When suggested simply using `Dir.mktmpdir` some years ago, the objection was that the files left in "tmp" would be useful for debugging, if I remember correctly.
However now it is checked that the directory becomes empty at the end.
It seems like there is little point in using the fixed path.

PR #7568, opened 2024-04-11
Add a limit to the size of the metadata and checksums files in a gem package.

segiddins:
This is to prevent a malicious gem from causing a denial of service by
including a very large metadata or checksums file,
which is then read into memory in its entirety just by opening the gem package.

This is guaranteed to limit the amount of memory needed, since
gzips (which use deflate streams for compression) have a maximum compression
ratio of 1032:1, so the uncompressed size of the metadata or checksums file
will be at most 1032 times the size of the (limited) amount of data read.

This prevents a gem from causing 500GB of memory to be allocated
to read a 500MB metadata file.

### DIFF
--- a/lib/rubygems/package.rb
+++ b/lib/rubygems/package.rb
@@ -528,12 +528,13 @@ EOM
   # Loads a Gem::Specification from the TarEntry +entry+
 
   def load_spec(entry) # :nodoc:
+    limit = 10 * 1024 * 1024
     case entry.full_name
     when "metadata" then
-      @spec = Gem::Specification.from_yaml entry.read
+      @spec = Gem::Specification.from_yaml limit_read(entry, "metadata", limit)
     when "metadata.gz" then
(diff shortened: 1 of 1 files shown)


indirect (approved):
seems like a good plan to me

lib/rubygems/package.rb
```diff
@@ -528,12 +528,13 @@ def normalize_path(pathname)
   # Loads a Gem::Specification from the TarEntry +entry+
 
   def load_spec(entry) # :nodoc:
+    limit = 10*1024*1024
```
martinemde, line 531: Could become a constant, like `MAX_METADATA_BYTES`.

martinemde (approved):

lib/rubygems/package.rb
simi: :pray:

deivid-rodriguez (approved):
Seems like a good solution to me too! Happy to wait for that constant to be extracted or merge this directly.

PR #7571, opened 2024-04-11
Fix installing plugins via relative paths

ccutrer:
This affected both CLI and Gemfile installs. It fixes a regression caused by #6960.

```
% bundle plugin install --path vendor/gems/bundler-multilock 
The path `/Users/cody/src/canvas-lms/.bundle/plugin/vendor/gems/bundler-multilock` does not exist.
```

 * Explicitly pass Dir.pwd as the root when installing a plugin from the CLI
 * Make sure that Bundler::Plugin::Source::Path#root still returns Bundler.root if we're within a bundle
 * Still return Plugin.root when we're not in the bundle (actual value doesn't matter; we just can't call Bundler.root when not in a bundle -- but the first bullet point will make the path calculation not actually use this value)

### DIFF
--- a/bundler/lib/bundler/plugin/installer.rb
+++ b/bundler/lib/bundler/plugin/installer.rb
@@ -77,7 +77,7 @@ module Bundler
 
       def install_path(names, version, path)
         source_list = SourceList.new
-        source = source_list.add_path_source({ "path" => path })
+        source = source_list.add_path_source({ "path" => path, "root_path" => SharedHelpers.pwd })
 
         install_all_sources(names, version, source_list, source)
       end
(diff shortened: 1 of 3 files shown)


bundler/lib/bundler/plugin/installer.rb
```diff
@@ -77,7 +77,7 @@ def install_git(names, version, options)
 
       def install_path(names, version, path)
         source_list = SourceList.new
-        source = source_list.add_path_source({ "path" => path })
+        source = source_list.add_path_source({ "path" => path, "root_path" => Pathname.new(Dir.pwd) })
```
deivid-rodriguez, line 80: I think

```suggestion
        source = source_list.add_path_source({ "path" => path, "root_path" => SharedHelpers.pwd })
```
gives you a Pathname already.

deivid-rodriguez (approved):
Minor suggestion, but looks good regardless!

bundler/lib/bundler/plugin/installer.rb
ccutrer: Done.

deivid-rodriguez (approved):

PR #7574, opened 2024-04-12
Don't upcase Windows ENV before backing it up

### LINKED ISSUE #7530: bundler/inline changes ENV keys to upper case

After the call to `gemfile do ... end` of _bundler/inline_ the keys of `ENV` are all changed to only upper case characters. This problem I found only with a Windows PC, not on a Mac.

Yes, I upgraded to the latest available versions.

The context for my tests was as follows:
* Windows 10 Build 19045
* Ruby 3.2.2 from [rubyinstaller-devkit-3.2.2-1-x64.exe](https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.2.1-1/rubyinstaller-devkit-3.2.1-1-x64.exe)
* Bundler 2.5.7
* Rubygems 3.5.7

``` Ruby
require 'bundler/inline'

ENV['Test_Variable'] = 'value string'
puts("before: #{ENV.each_key.select { |key| key.match?(/test_variable/i) }}")

gemfile do
  # pass
end

puts("after: #{ENV.each_key.select { |key| key.match?(/test_variable/i) }}")
```

Printing all key-value-pairs of ENV also shows that all keys are changed to upper case.

The output should be:
```
before: ["Test_Variable"]
after: ["Test_Variable"]
```

### What actually happened?

The output I've got was:
```
before: ["Test_Variable"]
after: ["TEST_VARIABLE"]
```

#### Environment

```
Bundler       2.5.7
  Platforms   ruby, x64-mingw-ucrt
Ruby          3.2.2p53 (2023-03-30 revision e51014f9c05aa65cbf203442d37fef7c12390015) [x64-mingw-ucrt]
  Full Path   C:/Develop/Ruby32-x64/bin/ruby.exe
  Config Dir  C:/ProgramData
RubyGems      3.5.7
  Gem Home    C:/Develop/Ruby32-x64/lib/ruby/gems/3.2.0
  Gem Path    C:/Users/<user-name>/.gem/ruby/3.2.0;C:/Develop/Ruby32-x64/lib/ruby/gems/3.2.0
  User Home   C:/Users/<user-name>
  User Path   C:/Users/<user-name>/.gem/ruby/3.2.0
  Bin Dir     C:/Develop/Ruby32-x64/bin
Tools
  Git         2.27.0.windows.1
  RVM         not installed
  rbenv       not installed
  chruby      not installed
```

#### Bundler Build Metadata

```
Built At          2024-03-22
Git SHA           199531f621
Released Version  true
```

deivid-rodriguez:

People find it strange that their ENV case is changed after `bundler/inline`. See https://github.com/rubygems/rubygems/issues/7530.

I originally started upcasing it in https://github.com/rubygems/rubygems/pull/3525, but I wasn't before specific about the problem and did not add a spec.

Checking if this is really needed.

Potentially fixes #7530.

In addition, I'm also cleaning up a Windows workaround that's no longer necessary.

### DIFF
--- a/bundler/lib/bundler/environment_preserver.rb
+++ b/bundler/lib/bundler/environment_preserver.rb
@@ -39,18 +39,7 @@ module Bundler
 
     # Replaces `ENV` with the bundler environment variables backed up
     def replace_with_backup
-      unless Gem.win_platform?
-        ENV.replace(backup)
-        return
-      end
-
-      # Fallback logic for Windows below to workaround
-      # https://bugs.ruby-lang.org/issues/16798. Can be dropped once all
-      # supported rubies include the fix for that.
-
-      ENV.clear
-
-      backup.each {|k, v| ENV[k] = v }
+      ENV.replace(backup)
     end
 
     # @return [Hash]
--- a/bundler/lib/bundler/environment_preserver.rb
+++ b/bundler/lib/bundler/environment_preserver.rb
@@ -19,14 +19,7 @@ module Bundler
     BUNDLER_PREFIX = "BUNDLER_ORIG_"
 
     def self.from_env
-      new(env_to_hash(ENV), BUNDLER_KEYS)
-    end
-
-    def self.env_to_hash(env)
-      to_hash = env.to_hash
-      return to_hash unless Gem.win_platform?
-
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I don't see any reason why the original fix was introduced, so I'd say this is good to go with a test!

Issue #7005, opened 2023-09-26
cannot load such file -- /Users/jedrek/.rvm/rubies/ruby-2.7.7/lib/ruby/gems/2.7.0/gems/bundler-2.4.19/exe/bundle (LoadError)


jedrekdomanski:
### Describe the problem as clearly as you can
On clean ruby 2.7.7 installed by RVM a default `bundler` v. 2.1.4 gem is installed. I need more recent version of the gem so I install it by running `gem install bundler:2.4.19 --default`. Then I try to install the gems `bundle isntall` and I get an error
```
cannot load such file -- /Users/jedrek/.rvm/rubies/ruby-2.7.7/lib/ruby/gems/2.7.0/gems/bundler-2.4.19/exe/bundle (LoadError)
```
I check the bundler versions and it looks like I have two versions installed both set as default gem. How can two gems be set as default? Why is the old one not unset as default automatically?
```
gem info bundler

*** LOCAL GEMS ***

bundler (2.4.19, 2.1.4)
    Authors: André Arko, Samuel Giddins, Colby Swandale, Hiroshi
    Shibata, David Rodríguez, Grey Baker, Stephanie Morillo, Chris
    Morris, James Wen, Tim Moore, André Medeiros, Jessica Lynn Suttles,
    Terence Lee, Carl Lerche, Yehuda Katz
    Homepage: https://bundler.io
    License: MIT
    Installed at (2.4.19, default): /Users/jedrek/.rvm/rubies/ruby-2.7.7/lib/ruby/gems/2.7.0
                 (2.1.4, default): /Users/jedrek/.rvm/rubies/ruby-2.7.7/lib/ruby/gems/2.7.0
```

### Did you try upgrading RubyGems?
Yes, but it failed
```
gem update --system
ERROR:  While executing gem ... (Gem::RemoteFetcher::FetchError)
    bad response Forbidden 403 (https://jedrekdomanski:REDACTED@rubygems.pkg.github.com/REDACTED/specs.4.8.gz)
```

hsbt:
>I need more recent version of the gem so I install it by running gem install bundler:2.4.19 --default

You don't need to add `--default` option for your use-case. Use just `gem install bundler` for recent stable version.

duckinator:
@jedrekdomanski did you ever get this sorted out?

[...]

Issue #5951, opened 2022-09-25
rubygems test fail with latest dependencies


Segaja:
Running `rake test` on version `v3.3.21` I get this result:

```
Loaded suite /usr/lib/ruby/gems/3.0.0/gems/rake-13.0.6/lib/rake/rake_test_loader
Started
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...............................................................................
...........................................................F
===============================================================================
Failure: test_build_extensions(TestGemExtBuilder): <"/build/rubygems/src/rubygems-3.3.21/tmp/test_rubygems_20220925-1122-f7jhor/gemhome/gems/a-2/lib/a.rb"> was expected to exist
/build/rubygems/src/rubygems-3.3.21/test/rubygems/test_gem_ext_builder.rb:140:in `test_build_extensions'
     137:     assert_path_exist @spec.gem_build_complete_path
     138:     assert_path_exist File.join @spec.extension_dir, "gem_make.out"
     139:     assert_path_exist File.join @spec.extension_dir, "a.rb"
  => 140:     assert_path_exist File.join @spec.gem_dir, "lib", "a.rb"
     141:     assert_path_exist File.join @spec.gem_dir, "lib", "a", "b.rb"
     142:   end
     143:
===============================================================================
......F
===============================================================================
Failure: test_build_extensions_with_gemhome_with_space(TestGemExtBuilder): <"/build/rubygems/src/rubygems-3.3.21/tmp/test_rubygems_20220925-1122-4fif2k/gem home/gems/a-2/lib/a.rb"> was expected to exist
/build/rubygems/src/rubygems-3.3.21/test/rubygems/test_gem_ext_builder.rb:140:in `test_build_extensions'
/build/rubygems/src/rubygems-3.3.21/test/rubygems/test_gem_ext_builder.rb:153:in `test_build_extensions_with_gemhome_with_space'
     150:     @spec = util_spec "a"
     151:     @builder = Gem::Ext::Builder.new @spec, ""
     152:
  => 153:     test_build_extensions
     154:   end
     155:
     156:   def test_build_extensions_install_ext_only
===============================================================================
.....P
===============================================================================
Pending: test_build_cdylib(TestGemExtCargoBuilder): cargo not present
[...]

PR #7588, opened 2024-04-18
Deprecate `--default` option from install command

hsbt:

The current `--default` option is confused for RubyGems users.

* https://github.com/rubygems/rubygems/issues/7005
* https://github.com/rubygems/rubygems/issues/5951#issuecomment-1273879356

This option is special behavior for default gems. But I'm not sure why it only generate executable, not copy library files and other processes. 

It leads complicated environment and there is no merit for them. I don't allow that users install gem as default gems from core maintainer's view. 

Finally, I'll show deprecated message and removed code about this option from rubygems. 


deivid-rodriguez:
@hsbt I agree with this approach. The `--default` option is problematic and causes broken installs. The easiest option would be to remove it and consider default gems fixed. Is this ready for review?

hsbt:
👌 It's ready for review now.

lib/rubygems/commands/setup_command.rb
```diff
@@ -400,16 +400,18 @@ def install_default_bundler_gem(bin_dir)
...
           force: options[:force],
-          install_as_default: true,
           bin_dir: bin_dir,
           install_dir: default_dir,
           wrappers: true
-        ).install
+        )
+        installer.install
+        File.delete installer.spec_file
+        installer.write_default_spec
```
deivid-rodriguez, line 414: I think this will install a full copy of bundler in the standard `$GEM_HOME`. I don't think it's a problem per se, but it makes Bundler inconsistent with other default gems with executables provided with Ruby (for example, irb, which includes a placeholder gem directory with just the executable).

We could try to iterate on this but perhaps for now it's better to not use `Gem::Installer` at all here and create the files necessary (executables, and gemspec) manually, to keep the exact same folder structure `gem update --system` creates now.

lib/rubygems/install_update_options.rb
```diff
@@ -158,10 +158,9 @@ def add_install_update_options
       options[:without_groups].concat v.map(&:intern)
     end
 
-    add_option(:"Install/Update", "--default",
+    add_option(:Deprecated, "--default",
                "Add the gem's full specification to",
                "specifications/default and extract only its bin") do |v,_o|
-      options[:install_as_default] = v
     end
```
deivid-rodriguez, line 164: Should we use `deprecate_option` here, to also give a runtime deprecation message? The way it is right now, I think it only appears as deprecated in documentation.

deivid-rodriguez (commented):
Right now this PR is deprecating the flag, but changing the behavior by making the flag a noop and install the gem normally, correct?

I believe this may be acceptable, because I don't see how the current implementation is helpful. However, we currently document exactly what the flag does, so may be potentially useful for someone, and the more conservative approach would be to first deprecate the flag without changing behavior, and then remove everything.

I think the latter option is preferable, but open to do it differently.

simi:
:thinking: It seems it was never finished feature (according to https://github.com/rubygems/rubygems/pull/566).

simi:
@headius is it ok to remove? Isn't that something JRuby specific? :thinking:

headius:
Wow, this goes back a ways.

When I originally added default gem support, I had intended `--default` to basically perform an install directly into the Ruby standard library location. Unfortunately I never finished that work and others who were more aware of how CRuby organizes its libraries were not available to help me fill in the blanks.

As it is now – a partially-implemented feature – it's obviously not very useful.

It would have been nice to get it fully working, but even then it's not something that should be user-facing; what we really want is something that can be used by Ruby implementers to install default gems into the standard library. CRuby and JRuby both have their own custom scripts to do this. TruffleRuby just copies out of the CRuby repository. None of these approaches are great.

I think it's fine to remove `--default` from RubyGems, but perhaps in its stead @hsbt and @nobu and I need to work on making our default gem installation logic more standard and shared.

headius:
> it's not something that should be user-facing

This may not entirely be true; there are probably users that would like to alter or update the "default" gems in their Ruby installs, and currently there's no way to do that without manually manipuating those gems' files. But I'm not sure what that functionality should look like.

Issue #3305, opened 2020-03-14
Is bundler fetching a gem **before** showing the "Fetching gem" message on screen?


jrodrigosm:
I recently found that, when deploying to Heroku (which is currently using Bundler 1.17.3), bundler's ouput was stalling for several minutes:

    (... messages here ...)
    remote: Installing devise-security 0.14.3
    (... output **stopped for 4 minutes 45 seconds** - and then all the messages below showed immediately)
    remote: Fetching bootstrap-sass 3.4.1
    remote: Fetching sassc-rails 2.1.2
    remote: Installing sassc-rails 2.1.2
    remote: Installing bootstrap-sass 3.4.1
    remote: Bundle complete! 56 Gemfile dependencies, 119 gems now installed.
    remote: Gems in the groups development and test were not installed.
    (... more messages here ...)

@schneems was very helpful in finding the source of the problem. At the beginning we thought it might be a problem with the `devise-security` gem, but we soon found that the culprits were `bootstrap-sass` and `sassc`, which were taking a long time to fech/install.

The thing is that bundler did not show the "Fetching bootstrap-sass" message until it was downloaded/installed - so we could not know why the output was stalling, as the last message shown was "Fetching devise-security" (that's why we though there was an issue with `devise-security`).

Could this somehow be an issue with bundler ordering, and how messages are shown on screen? Maybe something could be done so that bundler does not mislead us about which gem is taking so long to download/install?

bronzdoc:
@jrodrigosm Is this still an issue for you?

schneems:
I'm seeing this behavior with another app still. One option: could we add some debug output when gems are installing, record the amount of time the app is spending to install each individual gem and then output that to the shell?

deivid-rodriguez:
Yeah, very confusing indeed :(

The problem is that we have several concurrent threads (I guess you were using the `--jobs` flag in these cases?) that all print the log messages to the screen, so if `sassc` and `devise-security` get installed on separate threads, and `sassc` starts earlier, then the log message for `devise-security` is printed last, and it's indeed pretty confusing.

One possibility is that each thread saves the cursor location of the end of the last message it printed, and when done, it prints :heavy_check_mark: right there. But it will require quite a bit of terminal magic.

schneems:
Brainstorming here, on other possible fixes to mitigate this issue. 

[...]

PR #7622, opened 2024-05-02
Bundler: Show "Installed" message after installing a gem

jordan-brough:
Add an "Installed XXX in Y.YYYs" message after installing a gem. 

Example output:
```
Fetching buffering_logger 3.1.1
Installing buffering_logger 3.1.1
Installed buffering_logger 3.1.1 in 0.004s
```
☝️ the last line is new.

When installing gems in parallel (e.g. `bundle install --jobs 4`), I wanted it to be easier to tell when each gem finished installing and how long each gem took to install.

Currently, especially when reviewing logs on production servers, I have timestamps on log lines but I still can't tell which gems are taking a long time to install because I can't get a time diff between starting a particular gem and finishing the install (if I'm installing gems in parallel). Gems with native extensions are often ones that are slow that I'd like to identify.

Measure the time elapsed while installing (not including fetching) and print it out.

If this looks OK I will add some tests.

### DIFF
--- a/bundler/lib/bundler/source/rubygems.rb
+++ b/bundler/lib/bundler/source/rubygems.rb
@@ -202,7 +202,10 @@ module Bundler
         message += " with native extensions" if spec.extensions.any?
         Bundler.ui.confirm message
 
+        start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
         installed_spec = installer.install
+        end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        Bundler.ui.confirm "Installed #{version_message(spec, options[:previous_spec])} in %.3fs" % (end_time - start_time)
 
         spec.full_gem_path = installed_spec.full_gem_path
         spec.loaded_from = installed_spec.loaded_from


deivid-rodriguez:
This is great, and fixes an issue other people have experienced! See https://github.com/rubygems/rubygems/issues/3305.

One thought I'm having is that in https://github.com/rubygems/rubygems/issues/3305#issuecomment-997851919, people also complained about output being complicated to parse with two separate (possibly not consecutive) lines per gem.

So even if this change will help clarify which gem took longer to install, it will make the output even more complicated with 3 lines per gem.

I wonder if we could fix this problem (unclarity of which gems took long to install) while keeping the out as verbose as it is now, by instead of adding the "Installed " line to the output, make it replace the "Installing" line?

jordan-brough:
> instead of adding the "Installed " line to the output, make it replace the "Installing" line?

Personally I don't mind extra output, and the difference between fetching & installing could be important in some cases (network issues?). But that's an much less important edge case. I'm happy as long as I get a "start" and "stop" marker!

I'll update to remove the "Installing" line and I'll add some tests.

deivid-rodriguez:
> Personally I don't mind extra output, and the difference between fetching & installing could be important in some cases (network issues?). But that's an much less important edge case. I'm happy as long as I get a "start" and "stop" marker!

You're right, and I'm not totally sold on my own suggestion. It was mostly brainstorming to try add the useful information being added here without increasing the number of lines logged. Maybe, iterating on the suggestion, we could change everything to "after the fact" info, like

```
Fetched buffering_logger 3.1.1 in <0.xs>
Installed buffering_logger 3.1.1 in 0.004s
(...)
```

jordan-brough:
Hm, if we want to limit to 2 lines, personally I'd prefer to have "start" and "finished" lines, so if e.g. fetching is hung at any point I can still understand what the system is trying to do right now (I can see that it started but never finished).

From a "debug stuff" perspective I'd prefer to have all of these:
* Fetching
* Fetched in XXX
* Installing
* Installed in XXX

But if we want to be concise then the first and last seem the best to pick imo.

deivid-rodriguez:
All good points!

Keeping only "Fetching" and "Installed" may have issues too, since "Fetching" does not show up if the gem is already cached.

I'm not fully convinced but I guess all 4 would be best, indeed. If users are bothered enough about multiple lines in output per gem, they could open a new issue.

Ideally we would fix this "the right way" by properly parallelizing the input so that it just shows one line per gem that's updated with final execution time and status once done. But that's much more tricky to implement.

I wonder what @simi thinks since he participated in the linked issue too.

jordan-brough:
@deivid-rodriguez should I go ahead and implement displaying all 4 messages, or do we want to wait to hear back from @simi or others?

deivid-rodriguez:
Since I wouldn't want you to waste effort going in a direction we may end up changing, I'd say let's wait a bit for more input.

jordan-brough:
@deivid-rodriguez how long would you like to wait to see if someone else wants to chime in?

simi:
Sorry for chiming in late. Change looks good. Finding the balance is hard here to not overhelm users with a lot of info. It is wise to show something as early as possible (thus the starting message of Fetching...) and info about the end (Installed...). The only problem I do see is in the different verb used - Fetching abc -> Installed abc. Is that acceptable? @jordan-brough would you mind to rebase?

jordan-brough:
Fetching/Installed sounds good to me 👍 
I'll get this rebased. Thanks!

hsbt:
This feature request is impremented with https://github.com/ruby/rubygems/pull/9066. I'm closing this.

/cc @Edouard-chin

PR #7628, opened 2024-05-04
Add `--target-rbconfig` option to `gem install` and `gem update` commands

kateinoigakukun:

Today, CRuby runs on many platforms. But not all platforms are capable of running build tools (e.g. WebAssembly/WASI), so cross-target compilation against extensions libraries is essential for those platforms.

Currently, there is no way to cross-compile extension libraries in gems at _install-time_.

This patch adds `--target-rbconfig` option to specify the rbconfig.rb file for the deployment target platform. It basically pass through it to mkmf. The underlying `--target-rbconfig` support in mkmf has been introduced recently https://bugs.ruby-lang.org/issues/20345

At the moment, this option is only available when the following requirements are met:
- `extconf.rb`-based extensions
- requires mkmf in the latest master branch.

But we can relax the first requirement for Cargo builder by adding a support in rb-sys later.

```console
$ GEM_HOME=/tmp/gems-wasm32-wasi gem install nokogiri --target-rbconfig path/to/wasm32-wasi/rbconfig.rb
```

### DIFF
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -19,6 +19,10 @@ module Bundler
       # Disable color in deployment mode
       Bundler.ui.shell = Thor::Shell::Basic.new if options[:deployment]
 
+      if target_rbconfig_path = options["target-rbconfig"]
+        Bundler.rubygems.set_target_rbconfig(target_rbconfig_path)
+      end
+
       check_for_options_conflicts
 
       check_trust_policy
--- a/bundler/lib/bundler/cli/install.rb
+++ b/bundler/lib/bundler/cli/install.rb
@@ -19,7 +19,7 @@ module Bundler
       # Disable color in deployment mode
       Bundler.ui.shell = Thor::Shell::Basic.new if options[:deployment]
 
-      if target_rbconfig_path = options["target-rbconfig"]
+      if target_rbconfig_path = options[:"target-rbconfig"]
         Bundler.rubygems.set_target_rbconfig(target_rbconfig_path)
       end
 
--- a/lib/rubygems/ext/cargo_builder.rb
+++ b/lib/rubygems/ext/cargo_builder.rb
@@ -16,9 +16,14 @@ class Gem::Ext::CargoBuilder < Gem::Ext::Builder
     @profile = :release
   end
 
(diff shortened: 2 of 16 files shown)


lib/rubygems/ext/cargo_builder.rb
```diff
@@ -16,10 +16,15 @@ def initialize
...
   end
 
-  def build(extension, dest_path, results, args = [], lib_dir = nil, cargo_dir = Dir.pwd)
+  def build(extension, dest_path, results, args = [], lib_dir = nil, cargo_dir = Dir.pwd,
+    target_rbconfig=Gem.target_rbconfig)
     require "tempfile"
     require "fileutils"
 
+    if target_rbconfig.path
+      warn "--target-rbconfig is not yet supported for Rust extensions. Ignoring"
```
danini-the-panini, line 25: what would it take to implement this for rust?

kateinoigakukun: Basically, it should be done by replacing `RbConfig` reference with the corresponding value of `target_rbconfig` in `cargo_builder.rb`.  Maybe we need some change in rb-sys side, but wouldn't be big.

bundler/lib/bundler/cli/install.rb
```diff
@@ -19,6 +19,10 @@ def run
       # Disable color in deployment mode
       Bundler.ui.shell = Thor::Shell::Basic.new if options[:deployment]
 
+      if target_rbconfig_path = options["target-rbconfig"]
```
hsbt, line 22:
```suggestion
      if target_rbconfig_path = options[:target-rbconfig]
```

hsbt:
After resolving https://github.com/rubygems/rubygems/pull/7628/files#r1642315631, Let's merge this.

bundler/lib/bundler/cli/install.rb
```diff
@@ -19,6 +19,10 @@ def run
       # Disable color in deployment mode
       Bundler.ui.shell = Thor::Shell::Basic.new if options[:deployment]
 
+      if target_rbconfig_path = options[:"target-rbconfig"]
+        Bundler.rubygems.set_target_rbconfig(target_rbconfig_path)
```
deivid-rodriguez, line 23: This should probably fail more gracefully when the underlying RubyGems method is not available?

PR #7630, opened 2024-05-06
Remove unnecessary compatibility code

deivid-rodriguez:

I noticed a note about Ruby 2.5 compatibility which we should no longer need.

Remove compatibility code.


deivid-rodriguez:
Merging since it's straightforward.

PR #7633, opened 2024-05-07
Clarify Bundler support policy

deivid-rodriguez:

There's no clear place to check which versions of Bundler are supported, and which versions of Ruby and RubyGems are supported by a given version of Bundler.

Policy was last updated when we were about to start tracking upstream Ruby policy. Now we're already doing that, so we can clarify it.

### DIFF
--- a/bundler/doc/POLICIES.md
+++ b/bundler/doc/POLICIES.md
@@ -15,9 +15,13 @@ These policies are intended to be examples of how to apply these goals, and we r
 
 Bundler tries for perfect backwards compatibility. That means that if something worked in version 1.x, it should continue to work in 1.y and 1.z. That thing may or may not continue to work in 2.x. We may not always get it right, and there may be extenuating circumstances that force us into choosing between different kinds of breakage, but compatibility is very important to us. Infrastructure should be as unsurprising as possible.
 
-As of August, 2022, the Bundler team ships features and bugfixes as 2.3.x versions, and may exceptionally backport security fixes to older series shipped with supported Rubies by default, namely, Bundler 2.1 (shipped with Ruby 2.7) and Bundler 2.2 (shipped with Ruby 3.0).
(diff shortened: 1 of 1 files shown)


bundler/doc/POLICIES.md
```diff
@@ -15,9 +15,13 @@ These policies are intended to be examples of how to apply these goals, and we r
 
 Bundler tries for perfect backwards compatibility. That means that if something worked in version 1.x, it should continue to work in 1.y and 1.z. That thing may or may not continue to work in 2.x. We may not always get it right, and there may be extenuating circumstances that force us into choosing between different kinds of breakage, but compatibility is very important to us. Infrastructure should be as unsurprising as possible.
 
-As of August, 2022, the Bundler team ships features and bugfixes as 2.3.x versions, and may exceptionally backport security fixes to older series shipped with supported Rubies by default, namely, Bundler 2.1 (shipped with Ruby 2.7) and Bundler 2.2 (shipped with Ruby 3.0).
+Bundler supports any Ruby versions supported by the Ruby core team, as [listed here](https://www.ruby-lang.org/en/downloads/branches/). Shortly after Ruby core team officially drops support for a Ruby release, the Bundler team will do the same.
```
segiddins, line 18:
```suggestion
Bundler supports any MRI Ruby versions supported by the Ruby core team, as [listed here](https://www.ruby-lang.org/en/downloads/branches/). Shortly after Ruby core team officially drops support for a Ruby release, the Bundler team will do the same.
```

should also add a paragraph about non-MRI ruby engines IMO

deivid-rodriguez: Good point! I'll add it. For non MRI engines, I'd say we support only the latest version, does that sound ok?

segiddins: latest stable versions of truffleruby & jruby, and we are constantly evaluating changes in ruby implementation usage?

deivid-rodriguez:
Updated @segiddins!

simi (approved):

PR #7638, opened 2024-05-08
Rakefile is failing to load without requiring shared helpers.

martinemde:

Somehow, a fresh copy of rubygems is failing when running rake.

Require bundler shared helpers high enough up that it fixes the issue. If someone else can figure out a better way, I'm open to that. This patch at least solves the issue for now.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -659,6 +659,7 @@ task :override_version do
 end
 
 namespace :bundler do
+  require_relative "bundler/lib/bundler/shared_helpers"
   chdir(File.expand_path("bundler", __dir__)) do
     require_relative "bundler/lib/bundler/gem_tasks"
   end


MSP-Greg:
A recent CI run in ruby openssl had a possibly similar issue (NameError: uninitialized constant Bundler::SharedHelpers).

See https://github.com/ruby/openssl/actions/runs/8999647687, the three failing Windows jobs, which all installed Bundler 2.5.10.

segiddins (approved):

simi:
This seems to be locally problem only on Ruby 3.2. Is that possible? :thinking:

martinemde:
I was using 3.3.1.

hsbt:
@martinemde Can you share more details? I could use `gem_task` without this.

```
$ git clean -fxd
Removing .ruby-lsp/
Removing bundler/.rspec_status
Removing bundler/pkg/
Removing bundler/spec/support/artifice/used_cassettes.txt
Removing bundler/tmp/

$ rake bundler:install
bundler 2.6.0.dev built to pkg/bundler-2.6.0.dev.gem.
bundler (2.6.0.dev) installed.
```

I tried with fresh repository with `git clone git@github.com:rubygems/rubygems.git` and `rake bundler:install`. It's also working.

deivid-rodriguez:
By looking at logs posted by @MSP-Greg, it seems like the more natural place to add this require is right before https://github.com/rubygems/rubygems/blob/936827c1171a23773d0643e8fcb207bf67ec8397/bundler/lib/bundler/rubygems_ext.rb#L151.

It also explains why we never got this problem reported before, since that code was just introduced in Bundler 2.5.10.

hsbt:
Thanks, I got it. Should we require `shared_helpers` on `gem_tasks`?

MSP-Greg:
Thanks for looking at this.

Not sure how many have looked at how the error occurred.  `setup-ruby` updates Bundler because some Windows Ruby builds do not have a bash bin-stub for Bundler.  Note that the ruby-loco builds do have the bin-stub.  So, with the RubyGems/Bundler combination in Ruby master, there is no error, as you can see in the ruby/openssl CI Actions logs mentioned above.

I'm not sure how @martinemde noticed the issue, as the OpenSSL logs only occurred with a Bundler upgrade.  I apologize for not digging into the code to find the cause, but it does seem to only occur when Bundler 2.5.10 is used with a 'previously released' RubyGems.

Lastly, since RubyGems & Bundler are in the same repo, and testing new Bundler releases against previous RubyGems releases must be quite a PITA, maybe consideration should be given to no longer releasing Bundler gems?

Or, if one wants to update Bundler, one must do so with a RubyGems update...

martinemde:
I just pulled and then ran rake. We were helping people get setup to hack on rubygems at RailsConf and multiple people on OSX with fresh copies of the repo had this happen. My copy was not fresh but this error started happening recently.

I pushed this up to unblock people quickly so they could build off this branch. I'm not certain it's the right fix.

deivid-rodriguez:
I think the better fix is to require `bundler/shared_helpers` in `bundler/lib/bundler/rubygems_ext.rb`, right before it's used.

I was expecting `Bundler::SharedHelpers` be autoloaded, not sure why it's happening but at least adding the `require` right before usage is more explainable.

By the way, sorry for the bad timing of this regression! I wish we had found it before Rails Conf!

deivid-rodriguez:
I figured it out. It's unrelated to the version of RubyGems and/or Ruby. It can be reproduced regardless of those by having two different versions of `psych` installed. If this is the case, the RubyGems `require` logic that runs to resolve the proper `psych` version to activate will end up using the monkeypatched `StubSpecification` functionality _before_ the `SharedHelpers` autoload in Bundler has the chance to be run.

deivid-rodriguez:
I'll open a separate PR for this :+1:.

deivid-rodriguez:
I added what I think it's a better fix at #7647.

Issue #7636, opened 2024-05-08
Invalid plugin configuration should have a more useful error message when running rake setup


sarahsehr:
### Describe the problem as clearly as you can

Running `rake setup` as part of the dev setup errors out if previously installed bundler plugin no longer exists.

### Did you try upgrading rubygems & bundler?

Updated rubygems and bundler, but it's irrelevant to this STR

### Post steps to reproduce the problem

1) Install a bundler plugin
e.g.
```
gem install bundler-symlink
cd ~ && bundle plugin install bundler-symlink
```

2) Find the path for the plugin
```
vim ~/.bundle/plugin/index
```

Look for `plugin_paths:` and grab the path for the relevant plugin (e.g. `bundler-symlink: "/Users/sarah.sehr/.asdf/installs/ruby/2.7.7/lib/ruby/gems/2.7.0/gems/bundler-symlink-0.4.0"`)

3) Delete the plugin
e.g.
```
rm -rf /Users/sarah.sehr/.asdf/installs/ruby/2.7.7/lib/ruby/gems/2.7.0/gems/bundler-symlink-0.4.0
```

4) setup bundler for development
```
rake setup
```

### Which command did you run?

`rake setup`

### What were you expecting to happen?

Error indicating that it came from the plugin configuration, but the configuration isn't right.

### What actually happened?

```
LoadError: cannot load such file -- /Users/sarah.sehr/.asdf/installs/ruby/2.7.5/lib/ruby/gems/2.7.0/gems/bundler-symlink-0.3.0/plugins.rb
  /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/plugin.rb:347:in `load'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/plugin.rb:347:in `load_plugin'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/plugin.rb:231:in `block in hook'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/plugin.rb:231:in `each'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/plugin.rb:231:in `hook'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/installer.rb:24:in `install'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/cli/install.rb:63:in `run'
          /Users/sarah.sehr/src/rubygems/bundler/lib/bundler/cli.rb:247:in `block in install'
[...]

PR #7639, opened 2024-05-08
Add useful error message for plugin load

sarahsehr:

When starting to work on my first issue, I found that `rake setup` failed for me. This was due to having previously installed a plugin, whose files had since been deleted (presumably when I uninstalled that version of Ruby).

Fixes https://github.com/rubygems/rubygems/issues/7636.

Check to see whether each load path for the plugin is a directory. If it is not, notify the user in a friendly message which plugin in the problem and how to resolve the issue.

### DIFF
--- a/bundler/lib/bundler/plugin.rb
+++ b/bundler/lib/bundler/plugin.rb
@@ -342,7 +342,26 @@ module Bundler
       # done to avoid conflicts
       path = index.plugin_path(name)
 
-      Gem.add_to_load_path(*index.load_paths(name))
+      paths = index.load_paths(name)
+      invalid_paths = paths.reject {|p| File.directory?(p) }
+
+      if invalid_paths.any?
+        Bundler.ui.warn <<~MESSAGE
+          The following plugin paths don't exist: #{invalid_paths.join(", ")}.
+
+          This can happen if the plugin was installed with a different version of Ruby that has since been uninstalled.
+
+          If you would like to reinstall the plugin, run:
+
+          bundler plugin uninstall #{name} && bundler plugin install #{name}
+
+          Continuing without installing plugin #{name}.
+        MESSAGE
+
+        return
+      end
+
+      Gem.add_to_load_path(*paths)
 
       load path.join(PLUGIN_FILE_NAME)
 
--- a/bundler/spec/bundler/plugin_spec.rb
+++ b/bundler/spec/bundler/plugin_spec.rb
@@ -333,5 +333,28 @@ RSpec.describe Bundler::Plugin do
         end.to output("win\n").to_stdout
       end
     end
+
+    context "the plugin load_path is invalid" do
+      before do
+        allow(index).to receive(:load_paths).with("foo-plugin").
+          and_return(["invalid-file-name1", "invalid-file-name2"])
+      end
+
+      it "outputs a useful warning" do
+        msg =
+          "The following plugin paths don't exist: invalid-file-name1, invalid-file-name2.\n\n" \
+          "This can happen if the plugin was " \
+          "installed with a different version of Ruby that has since been uninstalled.\n\n" \
+          "If you would like to reinstall the plugin, run:\n\n" \
+          "bundler plugin uninstall foo-plugin && bundler plugin install foo-plugin\n\n" \
+          "Continuing without installing plugin foo-plugin.\n"
+
+        expect(Bundler.ui).to receive(:warn).with(msg)
+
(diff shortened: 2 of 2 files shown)


simi:
Hello @sarahsehr. Code looks good, but some specs are failing, which seems not really related to your change, since I can see the same failure on main branch. I'm looking into. :eyes:

bundler/lib/bundler/plugin.rb
```diff
@@ -342,7 +342,26 @@ def load_plugin(name)
...
+          This can happen if the plugin was installed with a different version of Ruby that has since
+          been uninstalled. To fix this, run:
+
+          bundler plugin uninstall #{name}
+
+          If you would like to reinstall the plugin, run:
+
+          bundler plugin install #{name}
+        MESSAGE
+        raise PluginError.new(message)
```
simi, line 361: Have you considered just warn in here and skip plugin + continue? :thinking:

bundler/lib/bundler/plugin.rb
```diff
@@ -342,7 +342,26 @@ def load_plugin(name)
...
+      end
+
+      Gem.add_to_load_path(paths)
```
simi, line 364: @sarahsehr this seems to fails the test (and breaks some test gem installation at the same time). `add_to_load_path` is defined at https://github.com/rubygems/rubygems/blob/c1933bfd4fc7579f679e80599b9ed28791619e24/lib/rubygems.rb#L576-L584 and it expects multiple arguments. To achieve this by passing `paths` array, you need to add splat (*) operator. Changing this to `Gem.add_to_load_path(*paths)` should fix this. :pray:

deivid-rodriguez:
Since this a `bundle install` run after all, should we automatically reinstall any missing plugins if necessary, instead of erroring out?

simi:
@sarahsehr friendly ping, is there still interest to finish this? Happy to help or take over to fix the one blocking comment in plugin.rb.

sarahsehr:
> @sarahsehr friendly ping, is there still interest to finish this? Happy to help or take over to fix the one blocking comment in plugin.rb.

Thanks for the friendly ping! Coincidentally, I was planning to work on this today. :)

deivid-rodriguez:
We're almost done with a big rework of how plugins currently work at #6957, I wonder if that could actually fix/improve this problem.

sarahsehr:
> We're almost done with a big rework of how plugins currently work at #6957, I wonder if that could actually fix/improve this problem.

Thanks for pointing this out! I pulled the branch from that PR and was able to reproduce the issue, so unfortunately it doesn't.

sarahsehr:
PluginLoadFailedWarning

@simi I have changed the output to warning and continue the install. The output looks like above (ignore the black line, that's debug output from another gem).

bundler/lib/bundler/plugin.rb
```diff
@@ -342,7 +342,30 @@ def load_plugin(name)
...
+          bundler plugin uninstall #{name} && bundler plugin install #{name}
+
+          Continuing without installing plugin #{name}.
+        MESSAGE
+        Bundler.ui.warn message
+        all_paths_valid = false
+      end
+      return unless all_paths_valid
+
+      Gem.add_to_load_path(*paths)
```
deivid-rodriguez, line 368: If there are two non existing paths in the same plugin, we'll end up printing this big error block twice. I think it will read slightly better to find all invalid paths first, and then print an error including all of them. Like

```ruby
      paths = index.load_paths(name)
      invalid_paths = paths.reject {|p| File.directory?(p) }

      if invalid_paths.any?
        Bundler.ui.warn <<~MESSAGE
          The following plugin paths don't exist: #{invalid_paths.join(", ")}.

          This can happen if the plugin was installed with a different version of Ruby that has since been uninstalled.

          If you would like to reinstall the plugin, run:
          
          bundler plugin uninstall #{name} && bundler plugin install #{name}

          Continuing without installing plugin #{name}.
        MESSAGE

        return
      end

      Gem.add_to_load_path(*paths)
```

sarahsehr: Ya, I agree that makes sense. I updated code/test. This is what it looks like now after running `rake setup`:

Screenshot 2024-09-13 at 12 48 29 PM

deivid-rodriguez:
This seems like a strict improvement, let's merge it!

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks so much @sarahsehr!

PR #7644, opened 2024-05-09
Fix double colons in questions asked from GemcutterUtilities

technicalpickles:
At some point, this is using Thor, and it automaticlaly adds a trailing `:` with space, so we don't need to do it ourselves.

I was using https://github.com/rubygems/configure_trusted_publisher and got minorly annoyed by the double `:` prompts:

```
Don't have an account yet? Create one at https://rubygems.org/sign_up

Username/email: : pickles@example.com

      Password: : 

Code: : 12345678
```

Update these `ask` calls to not include the `:`

### DIFF
--- a/lib/rubygems/gemcutter_utilities.rb
+++ b/lib/rubygems/gemcutter_utilities.rb
@@ -130,8 +130,8 @@ module Gem::GemcutterUtilities
 
     say "The existing key doesn't have access of #{scope} on #{pretty_host}. Please sign in to update access."
 
-    identifier = ask "Username/email"
-    password   = ask_for_password "      Password"
+    identifier = ask "Username/email: "
+    password   = ask_for_password "      Password: "
 
     response = rubygems_api_request(:put, "api/v1/api_key",
                                     sign_in_host, scope: scope) do |request|
@@ -159,8 +159,8 @@ module Gem::GemcutterUtilities
     say "Don't have an account yet? " \
         "Create one at #{sign_in_host}/sign_up"
 
-    identifier = ask "Username/email: "
-    password   = ask_for_password "      Password: "
+    identifier = ask "Username/email"
+    password   = ask_for_password "      Password"
     say "\n"
 
     key_name     = get_key_name(scope)
@@ -275,7 +275,7 @@ module Gem::GemcutterUtilities
       otp_thread[:otp]
     else
       say "You have enabled multi-factor authentication. Please enter OTP code."
-      ask "Code"
+      ask "Code:"
     end
   end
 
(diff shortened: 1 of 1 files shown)


technicalpickles:
The tests are failing because there isn't a colon 😅  I wonder if it has to do with the thor version?

deivid-rodriguez:
I don't think RubyGems uses thor. Bundler uses thor, but uses a fixed version so even then thor version should be irrelevant. So maybe it's something else? I'm unsure about what could cause this.

technicalpickles:
Well, I can reproduce the test failure locally at least.

The place I observed it was in https://github.com/rubygems/configure_trusted_publisher ... my output is identical to the readme, ie:

```
$ gem exec configure_trusted_publisher rubygem
Configuring trusted publisher for rubygem0 in /Users/segiddins/Development/github.com/rubygems/configure_trusted_publisher for rubygems/configure_trusted_publisher

Enter your https://rubygems.org credentials.
Don't have an account yet? Create one at https://rubygems.org/sign_up
Username/email: : gem-author
      Password: :

  1) Automatically when a new tag matching v* is pushed
  2) Manually by running a GitHub Action

How would you like releases for rubygem0 to be triggered? (1, 2) [2]: 2

Successfully configured trusted publisher for rubygem0:
  https://rubygems.org/gems/rubygem0/trusted_publishers
```

Specifically, `Username/email` and `Password` have extra `: ` in them.

simi:
@technicalpickles I can confirm this seems not actually fixing the problem you see. Can you reopen at https://github.com/rubygems/configure_trusted_publisher maybe?

```ruby
irb(main):001> require 'rubygems/user_interaction'
=> true
irb(main):002> Gem::DefaultUserInteraction.ui.ask "Hello"
Hello  yes
=> "yes"
irb(main):003> Gem::DefaultUserInteraction.ui.ask "Hello:"
Hello:  no
=> "no"
```

deivid-rodriguez:
Yes, this seems definitely an issue in https://github.com/rubygems/configure_trusted_publisher. It uses https://github.com/postmodern/command_kit.rb/blob/2f8c485eca112d06bd3fbd301f5659d4967a3ac7/lib/command_kit/interactive.rb#L75 for asking questions, which adds ": " automatically. And it tries to remove the ":" from rubygems question at https://github.com/rubygems/configure_trusted_publisher/blob/4245c026d3de2dd63b92d714dfc7c35a7723ade6/lib/configure_trusted_publisher/cli.rb#L163 to avoid the duplication. However, I think the `chomp` is probably not working because there are spaces after the ":". I think `strip.chomp(":")` would probably do the trick!

@technicalpickles Since you're already using the trusted publishers CLI, I'll let you verify that the suggestion works and upstream it, but let me know if you'd prefer that I do that!

technicalpickles:
Thanks for looking into this! Filed https://github.com/rubygems/configure_trusted_publisher/issues/11 for it

Issue #6457, opened 2023-03-09
`gem uninstall --user-install GEM_NAME` throws InstallError, even if the gem is installed


duckinator:
### Describe the problem as clearly as you can

If you run `gem install --user-install SOME_GEM`, then run `gem uninstall --user-install SOME_GEM`, RubyGems throws an error because the gem is not in `GEM_HOME` and tells you to use `--install-dir` instead.

There is a test which seems relevant, and passes. It does not catch this problem, however.

https://github.com/rubygems/rubygems/blob/625682996020c0d192cfacc0799fce2a86f005b0/test/rubygems/test_gem_uninstaller.rb#L371-L394

I suspect we need something closer to an integration test, possibly in [test/rubygems/test_gem_commands_uninstall_command.rb](https://github.com/rubygems/rubygems/blob/master/test/rubygems/test_gem_commands_uninstall_command.rb).

### Did you try upgrading RubyGems?

Yes. This occurs both on RubyGems 3.4.8 and master (as of https://github.com/rubygems/rubygems/commit/625682996020c0d192cfacc0799fce2a86f005b0)

### Post steps to reproduce the problem

Using Ruby 3.0.5p211, via the FreeBSD `lang/ruby30` package.

```
$ gem --version
3.4.8
$ gem install --user-install boop
Fetching boop-0.5.0.gem
WARNING:  You don't have /home/puppy/.gem/ruby/3.0/bin in your PATH,
          gem executables will not run.
Successfully installed boop-0.5.0
1 gem installed
$ gem uninstall --user-install boop
ERROR:  While executing gem ... (Gem::InstallError)
    boop is not installed in GEM_HOME, try:
        gem uninstall -i /usr/home/puppy/.gem/ruby/3.0 boop
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/uninstaller.rb:124:in `uninstall'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:195:in `uninstall'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:181:in `uninstall_gem'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:175:in `block in uninstall_specific'
        /usr/local/lib/ruby/site_ruby/3.0/rubygems/commands/uninstall_command.rb:167:in `each'
[...]

PR #7645, opened 2024-05-09
Fix `gem uninstall --user-install`  for symlinked HOME

deivid-rodriguez:

`gem uninstall --user-install` won't work if HOME is symlinked, which is the default on FreeBSD.

Make sure to expand home to its realpath.

Fixes https://github.com/rubygems/rubygems/issues/6457.

### DIFF
--- a/lib/rubygems/commands/uninstall_command.rb
+++ b/lib/rubygems/commands/uninstall_command.rb
@@ -184,7 +184,7 @@ that is a dependency of an existing gem.  You can use the
   rescue Gem::GemNotInHomeException => e
     spec = e.spec
     alert("In order to remove #{spec.name}, please execute:\n" \
-          "\tgem uninstall #{spec.name} --install-dir=#{spec.installation_path}")
+          "\tgem uninstall #{spec.name} --install-dir=#{spec.base_dir}")
   rescue Gem::UninstallError => e
     spec = e.spec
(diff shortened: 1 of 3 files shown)


lib/rubygems/defaults.rb
```diff
@@ -79,7 +79,7 @@ def self.default_specifications_dir
   #++
 
   def self.find_home
-    Dir.home.dup
+    File.realpath(Dir.home)
```
voxik, line 82: What is the reason for the `realpath`? That is IMHO almost always wrong thing to do.

deivid-rodriguez: We `realpath` all the other paths when comparing paths in `Gem::Uninstaller`. So in order for the comparisons to work as expected, the fix is to `realpath` all of them.

This is the whole point of the fix.

deivid-rodriguez: Alternatively I can expand it in `Gem::Uninstaller` directly when comparing paths. Maybe that's less invasive.

duckinator: @deivid-rodriguez I think it makes sense to move the `File.realpath(...)` call into `Gem::Uninstaller`. It avoids having to assume the input has already been normalized.

deivid-rodriguez: Done!

PR #7603, opened 2024-04-23
Show better error when installed gemspecs are unreadable

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

Bug report template is printed when an installed gemspec is unreadable.

## What is your fix for the problem, implemented in this PR?

Make sure we raise an error class that results in a nice error being printed instead of the error report template. Now the error is like this:

```
$ bundle install
Fetching gem metadata from https://rubygems.org/.
There was an error while trying to read from `/path/to/ruby/3.3.0/lib/ruby/gems/3.3.0/specifications/debug-1.9.1.gemspec`. It is likely that you need to grant read permissions for that path.
```

Fixes #7586.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7647, opened 2024-05-10
Fix crash caused by RubyGems `require` gem activation logic running before Bundler can properly register its own monkeypatches

deivid-rodriguez:

In https://github.com/rubygems/rubygems/pull/7603, we added a monkeypatch to `Gem::StubSpecification#data` so that when we fail to read the underlying gemspec due to not being readable, we can print a friendly error message. In order to do that, we reused a Bundler helper that knows how to rescue filesystem errors and re-raise them in way that Bundler knows how to handle in a friendly way.

This usage of `Bundler::SharedHelpers` was added without an explicit `require` because Bundler is supposed to setup an `autoload` for `Bundler::SharedHelpers`. However, the `autoload` does not seem to be triggering under some situations, such as when loading our own Rakefile. But that also does not happen always, but only when you have two different versions of `psych` installed in your system.

This is why:

* Our Rakefile first loads `rdoc/task` which activates `rdoc`, which depends on `psych >= 4.0`.
* Because there are multiple matching `psych` versions in the system, RubyGems `require` does not automatically activate `psych` but adds it to `Gem::Specification.unresolved_deps` and then further `require` calls will check in the potential paths for those `unresolved_deps` to see if any unresolved dependency can be activated. This enables logic that ends up using `Gem::StubSpecification#data`.
* Then `bundler/gem_tasks` is required that eventually requires `bundler/rubygems_ext` and enables the monkeypatch to `Gem::StubSpecification#data`.
* Then, before `bundler/gem_tasks` eventually loads `bundler.rb` and sets up an autoload for `Bundler::SharedHelpers`, other `require` calls are run that eventually use `Gem::Specification#data` for gem activation without `Bundler::SharedHelpers` being available yet.

The fix to `require "bundler/shared_helpers"` before using it. However, that's not enough because that `require` itself ends up running into the same error. So on top of that, I made sure no standard `require` calls are used during loading of `bundler/shared_helpers`.

Closes #7638.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -589,10 +589,13 @@ end
 
 desc "Check RubyGems integration"
 task :check_rubygems_integration do
-  # Bundler monkeypatches RubyGems in some ways that could potentially break gem
-  # activation. Run a non trivial binstub activation, with two different
-  # versions of a dependent gem installed.
+  # Bundler monkeypatches RubyGems in some ways that could potentially break gem activation.
+
+  # Run a non trivial binstub activation, with two different versions of a dependent gem installed.
   sh("ruby -Ilib -S gem install reline:0.3.0 reline:0.3.1 irb && ruby -Ibundler/lib -rbundler -S irb --version")
+
+  # With two psych versions installed, load a gem that depends on pysch and then load rubygems extensions.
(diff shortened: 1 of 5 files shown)


simi (approved):

martinemde:
Thanks for the thorough solution! I'll close my quick patch.

PR #7648, opened 2024-05-10
Recommend `bin/rake` over `rake` in contributing docs

deivid-rodriguez:

Our `bin/rake` script automatically installs `rake` if missing when running "setup tasks" and also makes sure to activate a consistent version of `rake`.

On top of that, it seems less prone to gem activation issues since it does not run into the same problem reported at #7638.

Consistently recommend and use `bin/rake`.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,9 +7,9 @@
   "remoteUser": "vscode",
   "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
   // Use 'postCreateCommand' to run commands after the container is created.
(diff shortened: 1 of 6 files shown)


simi (approved):

jasonkarns:
Should the `bundle gem` templates also be recommending the binstubs over `bundle exec`?

deivid-rodriguez:
That's a good point @jasonkarns! I'm not sure. My idea for the future is that you can eventually skip both "bundle exec" and binstubs (see https://github.com/rubygems/rubygems/pull/7752), but we're not there yet. So maybe creating binstubs by default is a good idea, I wonder what others think.

Issue #7547, opened 2024-03-31
2.4 breaks dumb (http) git repo support (for no good reason?)


costa:
### Describe the problem as clearly as you can

With 2.3, say, `gem 'ffmprb', :git => 'http://git/ffmprb/.git'`
works within my setup (`git` being a dumb [work-dir mirror] http server),
and with 2.4,
`fatal: dumb http transport does not support shallow capabilities`

### Did you try upgrading rubygems & bundler?

The problem was introduced in recent versions.

### Post steps to reproduce the problem

Find yourself in an older ruby environment (e.g. `$ docker run -it --rm --name bundler-test ruby:3.1 sh`)
Clone any gem (e.g. `$ git clone https://github.com/costa/ffmprb.git`),
Update server info (e.g. `$ cd ffmprb; git update-server-info`),
Serve the directory via HTTP locally (e.g. `$ gem install webrick; ruby -run -e httpd . -p 8080`),
In another term, in a temporary directory (e.g. `$ docker exec -it bundler-test sh`, `$ mkdir test-proj; cd test-proj`),
Create a `Gemfile` (e.g. `$ echo "source 'https://rubygems.org'" > Gemfile; echo "gem 'ffmprb', :git => 'http://localhost:8080/.git'" >> Gemfile`),
See that it succeeds (e.g. `$ bundle`) and exit the first term's environment (Ctrl-C, Ctrl-D).

Now, try the same with a newer ruby environment (e.g. `$ docker run -it --rm --name bundler-test ruby:3.2 sh`),
See, after all the steps above, instead of the big success, you get:
`fatal: dumb http transport does not support shallow capabilities`

### Which command did you run?

See above

### What were you expecting to happen?

See above

### What actually happened?

See above

### If not included with the output of your command, run `bundle env` and paste the output below

See above, it's your standard official Docker env.

hsbt:
>fatal: dumb http transport does not support shallow capabilities

The above messages was displayed by `git`, not RubyGems and Bundler.

costa:
> > fatal: dumb http transport does not support shallow capabilities
> 
> The above messages was displayed by `git`, not RubyGems and Bundler.

Correct, yet I believe the message is caused by some options Bundler is passing to `git`.

deivid-rodriguez:
I think https://github.com/rubygems/rubygems/pull/4475 may have broken that, since it indeed started cloning swallow repositories by default for performance.

[...]

PR #7649, opened 2024-05-10
Fix for gems not downloading from git via http

llenk:

https://github.com/rubygems/rubygems/issues/7547

The depth arguments in the git command were causing the problem. If we get the error on the first try, we remove the depth arguments from the command and try again. 

### DIFF
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -181,15 +181,15 @@ module Bundler
 
             if err.include?("Could not find remote branch")
               raise MissingGitRevisionError.new(command_with_no_credentials, nil, explicit_ref, credential_filtered_uri)
-            elsif err.include?("dumb http transport does not support shallow capabilities")
+            else
               idx = command.index("--depth")
               if idx
                 command.delete_at(idx)
                 command.delete_at(idx)
+                command_with_no_credentials = check_allowed(command)
+
                 err += "Retrying without --depth argument."
               end
-              raise GitCommandError.new(command_with_no_credentials, path, err)
-            else
               raise GitCommandError.new(command_with_no_credentials, path, err)
             end
           end
--- a/bundler/lib/bundler/source/git/git_proxy.rb
+++ b/bundler/lib/bundler/source/git/git_proxy.rb
@@ -181,6 +181,14 @@ module Bundler
 
             if err.include?("Could not find remote branch")
               raise MissingGitRevisionError.new(command_with_no_credentials, nil, explicit_ref, credential_filtered_uri)
+            elsif err.include?("dumb http transport does not support shallow capabilities")
+              idx = command.index("--depth")
+              if idx
+                command.delete_at(idx)
+                command.delete_at(idx)
+                err += "Retrying without --depth argument."
+              end
+              raise GitCommandError.new(command_with_no_credentials, path, err)
             else
               raise GitCommandError.new(command_with_no_credentials, path, err)
             end
--- a/bundler/spec/bundler/source/git/git_proxy_spec.rb
+++ b/bundler/spec/bundler/source/git/git_proxy_spec.rb
@@ -197,4 +197,18 @@ RSpec.describe Bundler::Source::Git::GitProxy do
 
     expect(Pathname.new(bundled_app("canary"))).not_to exist
   end
+
+  context "URI is HTTP" do
+    let(:uri) { "http://github.com/rubygems/rubygems.git" }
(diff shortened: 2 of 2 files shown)


martinemde (approved):

martinemde:
@llenk thank you! It was great working with you at RailsConf.

bundler/lib/bundler/source/git/git_proxy.rb
```diff
@@ -181,6 +181,14 @@ def clone_needs_extra_fetch?
 
             if err.include?("Could not find remote branch")
               raise MissingGitRevisionError.new(command_with_no_credentials, nil, explicit_ref, credential_filtered_uri)
+            elsif err.include?("dumb http transport does not support shallow capabilities")
```
simi, line 184: If the problem is with `http` protocol, wouldn't be possible to detect this earlier in `depth_args` based on the URI provided and don't even try with `--depth` for http ones?

martinemde: We tried this. GitHub works with http, but the example in the issue fails. It has to do with how it's being hosted. This allows it to try first and then remove if it fails.

simi: So does this rely on host not on the git itself? If so, isn't there better way how to react to this (like exit code of the command) to prevent vendoring 3rd party error message (which could change at any time)?

simi: https://github.com/CocoaPods/cocoapods-downloader/pull/65/files looks similar, seems this is good approach for this :white_check_mark:

martinemde: I like their regex. Let's borrow it.

```
/^fatal:.*does not support (--depth|shallow capabilities)$/
```

deivid-rodriguez: Maybe this can be completely moved into the `else` clause, so that we don't duplicate the `raise GitCommandError.new(command_with_no_credentials, path, err)` line.

deivid-rodriguez (commented):
Added a minor style comment but looks good to me other than that.

bundler/lib/bundler/source/git/git_proxy.rb
llenk: I think moving it to the else branch makes sense. That way it's also not dependent on the text of the error message. In the git source code, that error message is marked for translation, so it may not be consistent for all users. Also, even if some other error cause the code to go into the else branch, my understanding is that removing the depth argument wouldn't cause problems for the next retries. 

Added that change.

martinemde (approved):
I like the revisions. Looks more general now. Maybe even a pattern to follow if there's other ways to back off the command when it doesn't work.

bundler/lib/bundler/source/git/git_proxy.rb
deivid-rodriguez: Actually my suggestion was just to move the regexp check on the error message to inside the `else` branch:

```ruby
else
  if err.include?("dumb http transport does not support shallow capabilities")
    idx = command.index("--depth")
    if idx
      command.delete_at(idx)
      command.delete_at(idx)
      command_with_no_credentials = check_allowed(command)
      err += "Retrying without --depth argument."
    end
  end
  
  raise GitCommandError.new(command_with_no_credentials, path, err)
```

But maybe retrying every time fetching with `--depth` fails makes sense too!

llenk: Ah, I see what you meant. I think with the translation issue it would make sense to remove the `--depth` argument for any failure that hits the `else` branch, but I can make that change if it's preferred.

simi (approved):
:thinking: IMHO best would be to be able to check if remote supports "shallow capabilities", but I wasn't able to find out generic way to find out.

simi:
@llenk would you mind to squash commits please before merge?

deivid-rodriguez:
> 🤔 IMHO best would be to be able to check if remote supports "shallow capabilities", but I wasn't able to find out generic way to find out.

I was considering that, but (in addition to not being obvious how to do it) since most servers these days support them, it felt better to optimize for that case, and react when servers don't support it, instead of detecting support upfront.

bundler/lib/bundler/source/git/git_proxy.rb
deivid-rodriguez: I'm happy to try the current way, the potential translation issue makes sense to me, I'll set this as resolved!

deivid-rodriguez (approved):
Thanks for the great work here!

martinemde (approved):

martinemde:
Oops, I didn't notice it wasn't squashed yet. It's only 1 extra commit so I won't feel too bad 😉

PR #7659, opened 2024-05-14
Fix regression when caching gems from secondary sources

deivid-rodriguez:

If `cache_all_platforms` setting is enabled, the secondary source was no longer considering cached gems.

That means that if the remote secondary source has removed its gems, then this was now resulting in an error while before the previously cached gem from the source would still be used.

This is a regression from https://github.com/rubygems/rubygems/pull/7516.

This commit restores previous behavior by explicit enabling cached gems when materializing gems for all platforms so that they can be cached in vendor/cache.

### DIFF
--- a/bundler/lib/bundler/spec_set.rb
+++ b/bundler/lib/bundler/spec_set.rb
@@ -129,6 +129,7 @@ module Bundler
     def materialized_for_all_platforms
       @specs.map do |s|
         next s unless s.is_a?(LazySpecification)
+        s.source.cached!
         s.source.remote!
         spec = s.materialize_for_installation
         raise GemNotFound, "Could not find #{s.full_name} in any of the sources" unless spec
--- a/bundler/spec/commands/cache_spec.rb
+++ b/bundler/spec/commands/cache_spec.rb
@@ -386,6 +386,66 @@ RSpec.describe "bundle install with gem sources" do
       expect(the_bundle).to include_gems "rack 1.0.0"
     end
 
+    it "uses cached gems for secondary sources when cache_all_platforms configured" do
+      build_repo4 do
+        build_gem "foo", "1.0.0" do |s|
+          s.platform = "x86_64-linux"
+        end
+
+        build_gem "foo", "1.0.0" do |s|
+          s.platform = "arm64-darwin"
+        end
+      end
+
+      gemfile <<~G
+        source "https://gems.repo2"
+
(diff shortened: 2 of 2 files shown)


honeyankit:
@deivid-rodriguez Can we ship this tomorrow?

deivid-rodriguez:
No problem! Yes, my idea is to ship a new release this week.

That said, I think upgrading Dependabot may be a bit of work. In particular, while looking into this issue, I noticed Dependabot relies on the [`@unlock[:gems]`](https://github.com/dependabot/dependabot-core/blob/22e6cbe35cd1e8e46a53cb2e798f82ecba0ac45d/bundler/helpers/v2/lib/functions/lockfile_updater.rb#L192) value in `Bundler::Definition`, which is now removed at `initialize` time since https://github.com/rubygems/rubygems/pull/7558.

sachin-sandhu:
@deivid-rodriguez , do you have a release  

> No problem! Yes, my idea is to ship a new release this week.
> 
> That said, I think upgrading Dependabot may be a bit of work. In particular, while looking into this issue, I noticed Dependabot relies on the [`@unlock[:gems]`](https://github.com/dependabot/dependabot-core/blob/22e6cbe35cd1e8e46a53cb2e798f82ecba0ac45d/bundler/helpers/v2/lib/functions/lockfile_updater.rb#L192) value in `Bundler::Definition`, which is now removed at `initialize` time since #7558.

Hi @deivid-rodriguez , 
Is the release complete, if not, can you please share the # for us to keep track.
 
Thanks

deivid-rodriguez:
Hei @sachin-sandhu! Sorry, I delayed last week release a bit because I saw some strange hang in CI and I was afraid it could be due to some recent change. But I haven't seen it anymore, either locally or in CI, so I'll move the release forward as planned. I'll try to wrap it up tomorrow.

sachin-sandhu:
> Sorry once again for the delay here 🙏. I saw that hang again yesterday and I'm looking into it. Since this fix is a one liner, I would advice to monkeypatch the `Bundler::SpecSet#materialize_for_all_platforms` to get the fix for now. That way you don't need to block on the next Bundler release, nor on upgrading dependabot to use it, and can benefit from it inmediately.

Thanks @deivid-rodriguez for the inputs, we are working on the monkey patch fix 👍🏻

deivid-rodriguez:
No problem, sounds good, let me know if you find any issues!

PR #7672, opened 2024-05-19
Use string searching instead of parsing the versions file

martinemde:

I was curious if I could find more efficiency improvements for reading the versions file.

Directly read and search the versions file with string methods rather than indexing into separate strings, arrays, and hashes, and then using 0.5% of the data.

Updated metrics:

```
==> memprof.after.txt <==
Total allocated: 625.81 MB (8981279 objects)
Total retained:  202.57 MB (2449040 objects)

==> memprof.before.txt <== (bundler 2.5.11 with the new memory improvement)
Total allocated: 696.48 MB (9712558 objects)
Total retained:  237.74 MB (2992313 objects)
```

### DIFF
--- a/bundler/lib/bundler/compact_index_client.rb
+++ b/bundler/lib/bundler/compact_index_client.rb
@@ -4,6 +4,29 @@ require "pathname"
 require "set"
 
 module Bundler
+  # The CompactIndexClient is responsible for fetching and parsing the compact index.
+  #
+  # The compact index is a set of caching optimized files that are used to fetch gem information.
+  # The files are:
+  # - names: a list of all gem names
+  # - versions: a list of all gem versions
+  # - info/[gem]: a list of all versions of a gem
+  #
+  # The client is instantiated with:
+  # - `directory`: the root directory where the cache files are stored.
+  # - `fetcher`: (optional) an object that responds to #call(uri_path, headers) and returns an http response.
+  # If the `fetcher` is not provided, the client will only read cached files from disk.
+  #
+  # The client is organized into:
+  # - `Updater`: updates the cached files on disk using the fetcher.
+  # - `Cache`: calls the updater, caches files, read and return them from disk
+  # - `Parser`: parses the compact index file data
+  # - `CacheFile`: a concurrency safe file reader/writer that verifies checksums
+  #
+  # The client is intended to optimize memory usage and performance.
+  # It is called 100s or 1000s of times, parsing files with hundreds of thousands of lines.
+  # It may be called concurrently without global interpreter lock in some Rubies.
+  # As a result, some methods may look more complex than necessary to save memory or time.
   class CompactIndexClient
     # NOTE: MD5 is here not because we expect a server to respond with it, but
     # because we use it to generate the etag on first request during the upgrade
--- a/bundler/spec/bundler/compact_index_client/parser_spec.rb
+++ b/bundler/spec/bundler/compact_index_client/parser_spec.rb
@@ -20,7 +20,9 @@ RSpec.describe Bundler::CompactIndexClient::Parser do
 
   let(:compact_index) { TestCompactIndexClient.new(names, versions, info_data) }
   let(:names) { "a\nb\nc\n" }
-  let(:versions) { <<~VERSIONS }
+  let(:versions) { <<~VERSIONS.dup }
+    created_at: 2024-05-01T00:00:04Z
+    ---
     a 1.0.0,1.0.1,1.1.0 aaa111
     b 2.0.0,2.0.0-java bbb222
     c 3.0.0,3.0.3,3.3.3 ccc333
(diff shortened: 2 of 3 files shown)


bundler/lib/bundler/compact_index_client.rb
```diff
@@ -82,10 +82,9 @@ def dependencies(names)
 
     def update_and_parse_checksums!
       Bundler::CompactIndexClient.debug { "update_and_parse_checksums!" }
-      return @info_checksums_by_name if @parsed_checksums
+      return if @parsed_checksums
       update("versions", @cache.versions_path, @cache.versions_etag_path)
-      @info_checksums_by_name = @cache.checksums
-      @parsed_checksums = true
+      @parsed_checksums = !@cache.info_checksum_for("bundler").nil?
```
martinemde, line 87: This is the stand in for "is the file parseable?".

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -54,15 +54,14 @@ def versions_etag_path
         directory.join("versions.etag")
       end
 
-      def checksums
-        checksums = {}
+      def info_checksum_for(name)
+        @version_data ||= versions_path.read.freeze
```
martinemde, line 58: I think this is fine, but I want to note that the Cache instance is not retaining data. This is consistent with how we use the class now, but since cache is not responsible for invalidating or refetching `versions`, it can't bust its cache if the file is updated after being read. I think this is actually a design flaw with how the client is implemented but I wanted to keep this PR simple.

martinemde:
Note: test failures are in tests that mock `update_and_parse_checksums!` which was removed. I'll need to update the mocks.

deivid-rodriguez (commented):
Happy to merge this after we reevaluate it on top of #7637, but it seems a great idea because it's even more performant and also easy to understand 👍

bundler/lib/bundler/compact_index_client.rb
segiddins: i dont think this works, there can be other gem sources that don't include bundler

bundler/lib/bundler/compact_index_client/cache.rb
segiddins: since the client is what currently holds state, I think it makes sense to move this caching up into the compact index client, and keep the Cache responsible only for reading from disk

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -54,15 +54,14 @@ def versions_etag_path
...
-        lines(versions_path).each do |line|
-          name, _, checksum = line.split(" ", 3)
-          checksums[name] = checksum
-        end
+        line_start = @version_data.rindex("\n#{name} ")
+        return if line_start.nil?
 
-        checksums
+        checksum_end = @version_data.index("\n", line_start + 1)
+        @version_data[checksum_end - 32, 32] # MD5 checksum is 32 characters long
```
segiddins, line 64: could also make this future-proof by find the `" "` before the checksum end via `rindex`

bundler/lib/bundler/compact_index_client.rb
martinemde: Good point.

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: I had that at first, I'll put it back.

martinemde:
Based on some of the annoyances I ran into here, I created #7678 which will make this PR easier to merge.

martinemde:
Again, trying to get this merged has led me to #7705 which organizes the concurrency logic so that I don't have to deal with it in the client. This made the tests a lot easier to get passing.

This PR also quietly solves #7698 bit without explicitly allowing blank lines, which still do not fit [the spec](https://guides.rubygems.org/rubygems-org-compact-index-api/).

bundler/lib/bundler/compact_index_client.rb
martinemde: My new approach is to parse the last line, expecting it to end with a newline and have at least one space on the line. If the versions file is malformed after all the work we put in to make sure it's not, then I think it's just going to have to crash and fallback.

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: Parser holds state now after the refactor, which is where state was held when we parsed the entire thing into a big hash.

martinemde:
Now that #7705 is merged, this is ready to merge if approved.

bundler/spec/bundler/compact_index_client/parser_spec.rb
```diff
@@ -52,10 +55,50 @@ def set_info_data(name, value)
...
+      compact_index.versions = +"a 1.0.0 aaa1\n"
+      expect(parser).to be_available
+    end
+
+    it "returns true when versions has a gem and a header" do
+      compact_index.versions = +"---\na 1.0.0 aaa1\n"
+      expect(parser).to be_available
+    end
+
+    it "returns true when versions has only the header with header data" do
```
deivid-rodriguez, line 68:
```suggestion
    it "returns true when versions has a gem and a header with header data" do
```

deivid-rodriguez (approved):

PR #7677, opened 2024-05-21
Remove client-side MD5 ETag transition from compact index client

martinemde:

This was an offramp for generated etags to allow existing caches to be served until people could upgrade. It has been about 6 months since the transitional version was released, so we can remove this transitional code now.

Remove CompactIndexClient code that generated ETags using MD5.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/cache_file.rb
+++ b/bundler/lib/bundler/compact_index_client/cache_file.rb
@@ -86,11 +86,6 @@ module Bundler
         end
       end
 
-      # remove this method when we stop generating md5 digests for legacy etags
-      def md5
-        @digests && @digests["md5"]
-      end
-
       def digests?
         @digests&.any?
       end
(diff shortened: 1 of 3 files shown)


indirect:
I'm not super confident if 6 months is enough time for people to upgrade--the last thing we deprecated was still being used by 20% of our traffic more than 10 years after we released new clients. 😬

martinemde:
The only thing that happens is that people first running bundler after upgrading from 2.4 to 2.5. If you have an perfectly up-to-date versions file, then you upgrade to 2.5.x, then you `bundle update`, you will receive a ranged request response instead of a "not modified" response.

indirect (approved):
lol oh no I misread this as removing the support from the _server_ and not the client. ignore me, this is good 🤦🏻

martinemde:
Sorry for the confusion. I updated the title so it's more obvious what I changed.

segiddins (approved):

deivid-rodriguez (approved):
🎉

PR #7678, opened 2024-05-22
Reorganize and refactor CompactIndexClient

martinemde:

The way the cache and client interact, and the way the parsing is split between the two, plus some of the hacks we use made me want to refactor the compact index client.

Oh jeeze... what didn't I do. I layered all of the bits of the compact index so each can focus on one thing and then pass down to the next step without exposing its implementation details. The cache only does cache and not parsing. I extracted a parser. I wonder if the parser is really the same as the client. It could just move back into the cilent, but I like that it's a very basic class that just expects a simple client.

The cache is now responsible for updating so that only the updater knows about the files that get written and from outside the cache you only deal with strings. This makes the cache and the updating completely transparent. The cache now takes care of tracking which endpoints have been fetch too, just like a cache ought to.

My goal was really just to have some fun and understand the compact index better, but the result is nice. I particularly like pulling the Parser out of the cache/client.

I think I have it close to working now. It tries very hard not to change any actual behavior or approaches. For example, the parsing is just as efficient/inefficient as it was before, so #7637 is still applicable (though it moved, so this should probably merge after).

### DIFF
--- a/bundler/lib/bundler.rb
+++ b/bundler/lib/bundler.rb
@@ -42,6 +42,7 @@ module Bundler
   autoload :Checksum,               File.expand_path("bundler/checksum", __dir__)
   autoload :CLI,                    File.expand_path("bundler/cli", __dir__)
   autoload :CIDetector,             File.expand_path("bundler/ci_detector", __dir__)
+  autoload :CompactIndexClient,     File.expand_path("bundler/compact_index_client", __dir__)
   autoload :Definition,             File.expand_path("bundler/definition", __dir__)
   autoload :Dependency,             File.expand_path("bundler/dependency", __dir__)
   autoload :Deprecate,              File.expand_path("bundler/deprecate", __dir__)
--- a/bundler/lib/bundler/cli.rb
+++ b/bundler/lib/bundler/cli.rb
@@ -767,13 +767,10 @@ module Bundler
 
       return unless SharedHelpers.md5_available?
 
-      latest = Fetcher::CompactIndex.
-               new(nil, Source::Rubygems::Remote.new(Gem::URI("https://rubygems.org")), nil, nil).
-               send(:compact_index_client).
-               instance_variable_get(:@cache).
-               dependencies("bundler").
-               map {|d| Gem::Version.new(d.first) }.
-               max
+      require_relative "vendored_uri"
+      remote = Source::Rubygems::Remote.new(Gem::URI("https://rubygems.org"))
+      cache_path = Bundler.user_cache.join("compact_index", remote.cache_slug)
+      latest = Bundler::CompactIndexClient.new(cache_path).latest_version("bundler")
       return unless latest
 
       current = Gem::Version.new(VERSION)
--- a/bundler/lib/bundler/compact_index_client.rb
+++ b/bundler/lib/bundler/compact_index_client.rb
@@ -21,24 +21,17 @@ module Bundler
 
     require_relative "compact_index_client/cache"
     require_relative "compact_index_client/cache_file"
+    require_relative "compact_index_client/parser"
     require_relative "compact_index_client/updater"
 
-    attr_reader :directory
-
-    def initialize(directory, fetcher)
-      @directory = Pathname.new(directory)
-      @updater = Updater.new(fetcher)
-      @cache = Cache.new(@directory)
-      @endpoints = Set.new
-      @info_checksums_by_name = {}
-      @parsed_checksums = false
-      @mutex = Thread::Mutex.new
+    def initialize(directory, fetcher = nil)
+      @cache = Cache.new(directory, fetcher)
+      @parser = Parser.new(@cache)
     end
 
     def execution_mode=(block)
       Bundler::CompactIndexClient.debug { "execution_mode=" }
-      @endpoints = Set.new
-
+      @cache.reset!
       @execution_mode = block
     end
 
@@ -60,67 +53,28 @@ module Bundler
     end
 
     def names
-      Bundler::CompactIndexClient.debug { "/names" }
-      update("names", @cache.names_path, @cache.names_etag_path)
-      @cache.names
+      Bundler::CompactIndexClient.debug { "names" }
+      @parser.names
     end
 
     def versions
-      Bundler::CompactIndexClient.debug { "/versions" }
-      update("versions", @cache.versions_path, @cache.versions_etag_path)
-      versions, @info_checksums_by_name = @cache.versions
-      versions
+      Bundler::CompactIndexClient.debug { "versions" }
+      @parser.versions
     end
 
     def dependencies(names)
       Bundler::CompactIndexClient.debug { "dependencies(#{names})" }
-      execution_mode.call(names) do |name|
-        update_info(name)
-        @cache.dependencies(name).map {|d| d.unshift(name) }
-      end.flatten(1)
+      execution_mode.call(names) {|name| @parser.info(name) }.flatten(1)
     end
 
-    def update_and_parse_checksums!
-      Bundler::CompactIndexClient.debug { "update_and_parse_checksums!" }
-      return @info_checksums_by_name if @parsed_checksums
-      update("versions", @cache.versions_path, @cache.versions_etag_path)
-      @info_checksums_by_name = @cache.checksums
-      @parsed_checksums = true
+    def latest_version(name)
+      Bundler::CompactIndexClient.debug { "latest_version(#{name})" }
+      @parser.info(name).map {|d| Gem::Version.new(d[1]) }.max
     end
 
-    private
-
-    def update(remote_path, local_path, local_etag_path)
-      Bundler::CompactIndexClient.debug { "update(#{local_path}, #{remote_path})" }
-      unless synchronize { @endpoints.add?(remote_path) }
-        Bundler::CompactIndexClient.debug { "already fetched #{remote_path}" }
-        return
-      end
-      @updater.update(url(remote_path), local_path, local_etag_path)
-    end
-
-    def update_info(name)
-      Bundler::CompactIndexClient.debug { "update_info(#{name})" }
-      path = @cache.info_path(name)
-      unless existing = @info_checksums_by_name[name]
-        Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since it is missing from versions" }
-        return
-      end
-      checksum = SharedHelpers.checksum_for_file(path, :MD5)
-      if checksum == existing
-        Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since the versions checksum matches the local checksum" }
-        return
-      end
-      Bundler::CompactIndexClient.debug { "updating info for #{name} since the versions checksum #{existing} != the local checksum #{checksum}" }
-      update("info/#{name}", path, @cache.info_etag_path(name))
-    end
-
-    def url(path)
-      path
-    end
-
-    def synchronize
-      @mutex.synchronize { yield }
+    def available?
+      Bundler::CompactIndexClient.debug { "available?" }
+      @parser.available?
     end
   end
 end
--- a/bundler/lib/bundler/compact_index_client/cache.rb
+++ b/bundler/lib/bundler/compact_index_client/cache.rb
@@ -7,123 +7,89 @@ module Bundler
     class Cache
       attr_reader :directory
 
-      def initialize(directory)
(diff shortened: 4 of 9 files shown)


martinemde:
I'm not surprised that I broke this code:

https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/cli.rb#L770-L776

I am surprised that it seems to be almost the only thing I broke. 🤯

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -54,67 +63,41 @@ def versions_etag_path
...
-        ]
+      def fetch(remote, path, etag_path, &blk)
+        @updater.update url(remote), path, etag_path
+        read(path)
       end
 
-      def info_etag_root
-        directory.join("info-etags")
+      def url(remote)
+        remote
```
segiddins, line 100: is this necessary?

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,16 +7,46 @@ class CompactIndexClient
...
+          Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since the versions checksum matches the local checksum" }
+          read(path)
+        else
+          Bundler::CompactIndexClient.debug { "updating info for #{name} since the versions checksum #{remote_checksum} != the local checksum #{local_checksum}" }
+          fetch("info/#{name}", path, info_etag_path(name))
+        end
+      end
+
+      def available?
+        versions_path.file? && versions_path.size.nonzero?
```
martinemde, line 45: Probably needs to try fetching to be useful, but currently the parser doesn't call this anyway.

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,16 +7,46 @@ class CompactIndexClient
...
+        fetch("versions", versions_path, versions_etag_path)
+      end
+
+      def info(name, remote_checksum)
+        unless remote_checksum
+          Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since it is missing from versions index" }
+          return read(info_path(name))
+        end
+
+        path = info_path(name)
```
martinemde, line 33: Move to top of method

bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -0,0 +1,77 @@
...
+            version = version.split("-", 2).unshift(name)
+            if delete
+              @versions_by_name[name].delete(version)
+            else
+              @versions_by_name[name] << version
+            end
+          end
+        end
+
+        @parsed_checksums = true
```
martinemde, line 35: Unused

bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Bundler
+  class CompactIndexClient
+    class Parser
+      # `compact_index` - an object responding to #names, #versions, #info(name, checksum),
+      #                   returning the file contents as a string
+      def initialize(compact_index)
+        @compact_index = compact_index
+        @parsed_checksums = false
```
martinemde, line 10: Unused

bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def available?
+        return @info_checksums_by_name if @info_checksums_by_name
+
+        @info_checksums_by_name = {}
+        lines(@compact_index.versions).each do |line|
+          name, _, checksum = line.split(" ", 3)
+          @info_checksums_by_name[name] = checksum
+        end
```
martinemde, line 54: Will be replaced by new version

bundler/lib/bundler/compact_index_client/synchronized_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+        @mutex = Thread::Mutex.new
+        @endpoints = Set.new
+      end
+
+      def update(remote_path, *args)
+        if @mutex.synchronize { @endpoints.add?(remote_path) }
+          super
+        else
+          Bundler::CompactIndexClient.debug { "already fetched #{remote_path}" }
+          true
```
martinemde, line 19: I don't think true is needed.

bundler/lib/bundler/compact_index_client/synchronized_updater.rb
```diff
@@ -0,0 +1,28 @@
...
+
+require "set"
+
+module Bundler
+  class CompactIndexClient
+    class SynchronizedUpdater < Updater
+      def initialize(fetcher)
+        super(fetcher)
+        @mutex = Thread::Mutex.new
+        @endpoints = Set.new
```
martinemde, line 11:
```suggestion
        reset!
```

bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -0,0 +1,77 @@
...
+              @versions_by_name[name].delete(version)
+            else
+              @versions_by_name[name] << version
+            end
+          end
+        end
+
+        @parsed_checksums = true
+
+        versions
```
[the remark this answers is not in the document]
martinemde: This is the preserved old behavior, but luckily this method is entirely unused (yes really)

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: Preserved only because it looked so weird. I figured it was a plugin attach point. This was a pure refactor so I tried to change nothing.

martinemde (commented):
Notes to self.

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: Removed.

bundler/lib/bundler/fetcher/compact_index.rb
```diff
@@ -61,7 +59,7 @@ def available?
           return nil
         end
         # Read info file checksums out of /versions, so we can know if gems are up to date
-        compact_index_client.update_and_parse_checksums!
+        compact_index_client.available?
```
deivid-rodriguez, line 62: Love it!

bundler/lib/bundler.rb
```diff
@@ -42,6 +42,7 @@ module Bundler
   autoload :Checksum,               File.expand_path("bundler/checksum", __dir__)
   autoload :CLI,                    File.expand_path("bundler/cli", __dir__)
   autoload :CIDetector,             File.expand_path("bundler/ci_detector", __dir__)
+  autoload :CompactIndexClient,     File.expand_path("bundler/compact_index_client", __dir__)
```
deivid-rodriguez, line 45: Agree the `autoload` makes more sense here with the rest of the stuff.

bundler/lib/bundler/cli.rb
```diff
@@ -767,13 +767,10 @@ def warn_on_outdated_bundler
...
-               new(nil, Source::Rubygems::Remote.new(Gem::URI("https://rubygems.org")), nil, nil).
-               send(:compact_index_client).
-               instance_variable_get(:@cache).
-               dependencies("bundler").
-               map {|d| Gem::Version.new(d.first) }.
-               max
+      require_relative "vendored_uri"
+      remote = Source::Rubygems::Remote.new(Gem::URI("https://rubygems.org"))
+      cache_path = Bundler.user_cache.join("compact_index", remote.cache_slug)
+      latest = Bundler::CompactIndexClient.new(cache_path).latest_version("bundler")
```
deivid-rodriguez, line 773: Nice, much less hacky with a new public method 👍.

bundler/lib/bundler/compact_index_client.rb
```diff
@@ -21,24 +21,17 @@ class Error < StandardError; end
...
+    def initialize(directory, fetcher = nil)
+      @cache = Cache.new(directory, fetcher)
+      @parser = Parser.new(@cache)
     end
 
     def execution_mode=(block)
       Bundler::CompactIndexClient.debug { "execution_mode=" }
-      @endpoints = Set.new
-
+      @cache&.reset!
```
deivid-rodriguez, line 34: Is `@cache` ever nil?

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,114 +7,107 @@ class CompactIndexClient
...
-      def initialize(directory)
+      def initialize(directory, fetcher = nil)
         @directory = Pathname.new(directory).expand_path
-        info_roots.each {|dir| mkdir(dir) }
-        mkdir(info_etag_root)
+        @updater = Updater.new(fetcher) if fetcher
+        @mutex = Thread::Mutex.new
+        @endpoints = Set.new
+
+        SharedHelpers.filesystem_access(@directory) do
```
deivid-rodriguez, line 16: Each `mkdir` already wraps the directory creation with this method, is this one necessary?

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,114 +7,107 @@ class CompactIndexClient
...
-            delete = version.delete_prefix!("-")
-            version = version.split("-", 2).unshift(name)
-            if delete
-              versions_by_name[name].delete(version)
-            else
-              versions_by_name[name] << version
-            end
-          end
+        unless remote_checksum
+          Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since it is missing from versions index" }
```
deivid-rodriguez, line 35: "versions index" rather than "versions" makes the message more clear 👍.

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,114 +7,107 @@ class CompactIndexClient
...
+          return read(path)
         end
 
-        [versions_by_name, info_checksums_by_name]
+        local_checksum = SharedHelpers.checksum_for_file(path, :MD5)
+        if local_checksum == remote_checksum
+          Bundler::CompactIndexClient.debug { "skipping updating info for #{name} since the versions checksum matches the local checksum" }
```
deivid-rodriguez, line 41: Perhaps using "versions index" in these debug messages makes sense as well.

bundler/lib/bundler/compact_index_client/cache.rb
```diff
@@ -7,114 +7,107 @@ class CompactIndexClient
...
-          versions_string.split(",") do |version|
-            delete = version.delete_prefix!("-")
-            version = version.split("-", 2).unshift(name)
-            if delete
-              versions_by_name[name].delete(version)
-            else
-              versions_by_name[name] << version
-            end
-          end
+        unless remote_checksum
```
deivid-rodriguez, line 34: Purely style/taste: previously we had two early returns for the "update skipped" cases, and then the "update performed" case followed. Now we have one early return for the first skipped case and then symmetric logic for the others. I would either keep both early returns as they were before, or remove all of them.

deivid-rodriguez (approved):
I added a few comments, most of them requiring no action, but just commenting on the good things the I observed. The others are very minor.

I love how you managed to split pure parsing logic to its own class 🥇.

bundler/lib/bundler.rb
martinemde: Yeah, it now fails without this change so I had to move it. However, I agree it makes more sense here.

bundler/lib/bundler/compact_index_client.rb
martinemde: oh! leftover from when this interacted with `@updater` which could be nil. I'll fix now that the `@cache` abstracts this situation.

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: hmm, good point. I pushed this down from Fetcher::CompactIndex. I think maybe it's just entirely unnecessary.

bundler/lib/bundler/cli.rb
deivid-rodriguez: Didn't think a lot about it, but avoiding things like `instance_variable_get` in code we do control feels good!

bundler/lib/bundler/compact_index_client/cache.rb
martinemde: good idea. fixed.

PR #7682, opened 2024-05-23
Rework `bundle gem` specs so that they work on Windows

deivid-rodriguez:

I was experimenting with migrating specs to use a "shelling out" library that handles timeouts well, like https://github.com/piotrmurach/tty-command.

If we do this, we can set a maximum time for commands to run, and get the test to abort otherwise. That should give us better behavior tests hang for whatever reason.

While experimenting with this, I noticed that these tests were hanging with the new library. I think it's the same problem that explain why we skip them on Windows. Most of the commands run by these specs actually prompt for user input but for some reason `Open3` outside of Windows ignores that and goes on with no user input. Windows and tty-command don't like this.

This commit refactors these specs so that:

* They don't prompt for user input when not necessary.
* They pass proper user input when necessary.

As long as no user input is needed, most specs pass, so it allows to remove the `:readline` tag from most of them.

### DIFF
--- a/bundler/spec/commands/newgem_spec.rb
+++ b/bundler/spec/commands/newgem_spec.rb
@@ -36,22 +36,25 @@ RSpec.describe "bundle gem" do
     sys_exec("git config --global user.name 'Bundler User'")
     sys_exec("git config --global user.email user@example.com")
     sys_exec("git config --global github.user bundleuser")
+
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
On Windows CI, before:

```
3182 examples, 0 failures, 120 pending
```

After:

```
3462 examples, 0 failures, 122 pending
```

I will merge since it affects specs only.

Issue #7693, opened 2024-05-29
bundle gem --ext=rust does not install well as a gem out of the box


gl-yziquel:
Hi.

### Describe the problem as clearly as you can

I've been trying to scaffold mixed rust/ruby code with magnus. So I tried the `bundle gem --ext=rust` command. And tweaked it a bit to make it work, but very minimally. Here is the result:

https://github.com/gl-yziquel/skeleton_rust_gem.git

Tagged: https://github.com/gl-yziquel/skeleton_rust_gem/releases/tag/gh-rubygems-7693

I added a Justfile to automate some tasks.

Bottom line: `just build` works fine and creates a gem in `pkg/`. However, `just bug-1` and `just bug-2` show that the scaffolded code is not designed, out of the box, to get the gem installed. I believe this is a shortcoming of the current scaffolding of rust gems.

```console
mini-me@virtucon ~/h/s/skeleton_rust_gem (master)> bundle exec rake install
cd tmp/x86_64-linux-gnu/skeleton_rust_gem/3.1.2
/usr/bin/ruby3.1 -I. ../../../../ext/skeleton_rust_gem/extconf.rb
checking for x86_64-linux-gnu-gcc... yes
checking for x86_64-linux-gnu-g++... yes
checking for x86_64-linux-gnu-gcc-ar... yes
cd -
cd tmp/x86_64-linux-gnu/skeleton_rust_gem/3.1.2
/usr/bin/gmake
generating /home/mini-me/home/stove/skeleton_rust_gem/target/release/libskeleton_rust_gem.so (release)
cargo rustc  --manifest-path /home/mini-me/home/stove/skeleton_rust_gem/ext/skeleton_rust_gem/Cargo.toml --target-dir /home/mini-me/home/stove/skeleton_rust_gem/target --lib --profile release -- -C linker=x86_64-linux-gnu-gcc -L native=/usr/lib -C link-arg=-Wl,-Bsymbolic-functions -C link-arg=-Wl,-z,relro -C link-arg=-Wl,-z,now -C link-arg=-lm
    Finished release [optimized] target(s) in 0.09s
cd -
/usr/bin/gmake install sitearchdir=../../../../lib/skeleton_rust_gem sitelibdir=../../../../lib/skeleton_rust_gem target_prefix=
generating /home/mini-me/home/stove/skeleton_rust_gem/target/release/libskeleton_rust_gem.so (release)
[...]

PR #7610, opened 2024-04-25
Update `--ext=rust` to support compiling the native extension from source

karreiro:
## What was the end-user or developer problem that led to this PR?

I needed to apply some tiny changes to the skeleton generated by the `--ext=rust` option to compile my native extension from source.

## What is your fix for the problem, implemented in this PR?

When gem consumers install a gem with a native extension and their platform is not included the pre-compiled bundles, they need to compile it locally and rely on the `extconf.rb` file.

Referencing the `extconf.rb` file in the gemspec is required  to make that work. We also need to include `rb_sys` as a dependency because the Rust `extconf.rb` depends on it.

Here's a demo of this change in action :)

https://github.com/rubygems/rubygems/assets/1079279/54e543a8-0827-4fce-bab6-1bb8c0740b42

Closes #7693.

PR #7695, opened 2024-05-29
Rust gems should install if extension is extconf.rb and not Cargo.toml.

gl-yziquel:

Try to install a rust gem scaffolded with bundle gem --ext=rust.

This is a quick fix. I do not expect it to be merged as is.

I fixed the templating of the gemspec file to refer to extconf.rb instead of Cargo.toml.

I'm not sure it is the right way to proceed. Would appreciate advice or feedback.


gl-yziquel:
Should fix: https://github.com/rubygems/rubygems/issues/7693

deivid-rodriguez:
@gl-yziquel Thanks for this 🙏. It seems #7610 is also fixing the same problem, but in a more robust way, so let's close this PR in favor of #7610. Feel free to leave any comments in there 👍.

PR #7698, opened 2024-05-30
Allow Artifactory to add blank lines to versions file

martinemde:
I see no reason for these lines to exist, but, like an appendix, they do exist and, for some people, they explode.

After we released #7637, some Artifactory users found that they were getting NoMethodErrors when parsing the versions file.

I reached out to JFrog to see if Artifactory will fix the issue, but in the meantime we will handle blank lines more gracefully.

### DIFF
--- a/bundler/lib/bundler/compact_index_client/parser.rb
+++ b/bundler/lib/bundler/compact_index_client/parser.rb
@@ -22,6 +22,7 @@ module Bundler
 
         lines(@compact_index.versions).each do |line|
           name, versions_string, checksum = line.split(" ", 3)
+          next unless versions_string # Artifactory has a bug where blank lines exist in the versions file
           @info_checksums[name] = checksum || ""
           versions_string.split(",") do |version|
             delete = version.delete_prefix!("-")
@@ -69,15 +70,16 @@ module Bundler
       # This is mostly the same as `split(" ", 3)` but it avoids allocating extra objects.
       # This method gets called at least once for every gem when parsing versions.
       def parse_version_checksum(line, checksums)
-        line.freeze # allows slicing into the string to not allocate a copy of the line
         name_end = line.index(" ")
+        return if name_end.nil? # Artifactory has a bug where blank lines exist in the versions file
         checksum_start = line.index(" ", name_end + 1) + 1
+        return if checksum_start.nil?
         checksum_end = line.size - checksum_start
-        # freeze name since it is used as a hash key
-        # pre-freezing means a frozen copy isn't created
-        name = line[0, name_end].freeze
+
+        line.freeze # allows slicing into the string to not allocate a copy of the line
+        name = line[0, name_end]
         checksum = line[checksum_start, checksum_end]
-        checksums[name] = checksum
+        checksums[name.freeze] = checksum # pre-freezing name means a frozen copy isn't created
       end
     end
   end
--- a/bundler/spec/bundler/compact_index_client/parser_spec.rb
+++ b/bundler/spec/bundler/compact_index_client/parser_spec.rb
@@ -56,6 +56,22 @@ RSpec.describe Bundler::CompactIndexClient::Parser do
       compact_index.versions = nil
       expect(parser).not_to be_available
     end
+
+    context "with Artifactory" do
+      # Artifactory is adding seemingly random blank lines to the versions file
+      let(:versions) { <<~VERSIONS }
+        a 1.0.0,1.0.1,1.1.0 aaa111
+
+        b 2.0.0,2.0.0-java bbb222
+        c 3.0.0,3.0.3,3.3.3 ccc333
+        c -3.0.3 ccc333yanked
+      VERSIONS
+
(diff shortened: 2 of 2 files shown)


bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -69,15 +69,16 @@ def gem_parser
...
         checksum_end = line.size - checksum_start
-        # freeze name since it is used as a hash key
-        # pre-freezing means a frozen copy isn't created
-        name = line[0, name_end].freeze
+
+        line.freeze # allows slicing into the string to not allocate a copy of the line
+        name = line[0, name_end]
         checksum = line[checksum_start, checksum_end]
-        checksums[name] = checksum
+        checksums[name.freeze] = checksum # pre-freezing name means a frozen copy isn't created
```
martinemde, line 81: Just reshuffled a bit here:
* I moved `line.freeze` down so we don't freeze before returning.
* I also moved `name.freeze` into a spot that allowed me to explain with code, shortening the comment.
* I scratched my pedantic code nitpicking itch

bundler/spec/bundler/compact_index_client/parser_spec.rb
```diff
@@ -56,6 +56,21 @@ def set_info_data(name, value)
...
+      let(:versions) { <<~VERSIONS }
+        a 1.0.0,1.0.1,1.1.0 aaa111
+
+        b 2.0.0,2.0.0-java bbb222
+        c 3.0.0,3.0.3,3.3.3 ccc333
+        c -3.0.3 ccc333yanked
+      VERSIONS
+
+      it "parses the versions file correctly even with blank lines in the file" do
+        parser.available?
```
segiddins, line 71:
```suggestion
        expect(parser).to be_available
```

segiddins (approved):

bundler/lib/bundler/compact_index_client/parser.rb
```diff
@@ -22,6 +22,7 @@ def versions
 
         lines(@compact_index.versions).each do |line|
           name, versions_string, checksum = line.split(" ", 3)
+          next unless versions_string # Artifactory has a bug where blank lines exist in the versions file
```
martinemde, line 25: I reluctantly added this, even though there's no previous expectation that it should handle blank lines in the file.

deivid-rodriguez (approved):

simi:
I'm not sure this is good approach. We should not bandaid troubles caused by "wrongly" implemented gem servers. If we onboard this, is there plan to offboard this as well? If not, we will randomly change "specification" of compact index to allow empty lines. Those tiny details can block later (per my experience) block optimizations etc... I would rather focus energy to provide tools to make it easier for other gemserver implementors to lint their output according to RubyGems expectations..

deivid-rodriguez:
We could add a warning, but I think we should definitely patch this back and get Artifactory back on track, even if blank lines were not intentionally part of the specification.

martinemde:
@simi This was exactly my concern... We don't have a good offramp and Artifactory is not following spec.

To @deivid-rodriguez's point, we used to support this, if only accidentally.

I put in a technical support ticket with jfrog about this and also asked for them to reach out to me to coordinate. If anyone knows someone there who we could talk to, I assume this would be as easy fix.

Here's the possible ways forward:
1. Fix it, supporting it indefinitely or until we hear from jfrog that it's fixed.
2. Don't fix it, tell artifactory users to downgrade to 3.5.10 until Artifactory doesn't have the bug.
3. Switch to the string search approach I wrote in #7672 which doesn't care about empty lines.

deivid-rodriguez:
Could we print a warning if we find an empty line, so people can fix their API implementations, and we can eventually remove the ad-hoc code to support it?

If we were going to move on with #7672 anyways, I guess it's not that much of an issue.

pascalbetz:
> 2. Don't fix it, tell Artifactory users to downgrade to 3.5.10 until Artifactory doesn't have the bug.

I work in a big company, we use JFrog.
Even if JFrog releases a fixed version right now it would take months or years for it to land in my environment 🥹

So I'd much prefer 1) or 3) (or 1) and then 3) ).

Thanks for the work on this, it is much appreciated.

simi:
@pascalbetz do you mean you'll get stucked with Bundler 3.5.10 for long time? Is it not common to upgrade Artifactory anytime soon after release?

pascalbetz:
@simi 
yes, we'd probably be stuck with an not-up-tp-date bundler for a very long time.
As mentioned: BigCompany. Artifactory/JFrog is operated by a different department and not only used for Ruby development.

Whats more is that we use Base Images with ruby pre-installed. So it's hard to switch to a different bundler version - as there is no straightforward way to replace the system bundler (that i know of).

martinemde:
We decided to merge #7672 which addresses this problem indirectly by not parsing the entire file (and therefore ignore the invalid newlines). This problem still needs to be fixed by Artifactory since it may still cause parsing problems.

PR #7664, opened 2024-05-16
Fix `gem pristine` sometimes failing to pristine user installed gems

deivid-rodriguez:
## What was the end-user or developer problem that led to this PR?

If the same version of a gem is installed in the default gem home, and in the user home, `gem pristine` will fail to pristine the user installed one.

## What is your fix for the problem, implemented in this PR?

Make sure all copies are properly pristined by not removing duplicates from the stubs list.

## Make sure the following tasks are checked

- [x] Describe the problem / feature
- [x] Write [tests](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#tests) for features and bug fixes
- [x] Write code to solve the problem
- [x] Make sure you follow the [current code style](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#code-formatting) and [write meaningful commit messages without tags](https://github.com/rubygems/rubygems/blob/master/bundler/doc/development/PULL_REQUESTS.md#commit-messages)

PR #7700, opened 2024-05-30
Fix Windows CI that got silently broken recently

deivid-rodriguez:

After https://github.com/rubygems/rubygems/pull/7664, `Gem::StubSpecification` sorting is no longer stable, which causes spec failures under some platforms.

Restore a stable criteria for sorting specifications, that keeps the current tests green.

Also, make sure Windows can actually run our tests properly and does not give back a green status without running anything.

### DIFF
--- a/bin/rake.cmd
+++ b/bin/rake.cmd
@@ -1,0 +1,2 @@
+@ECHO OFF
+@"ruby.exe" -x "%~dpn0" %*
--- a/bundler/spec/support/rubygems_ext.rb
+++ b/bundler/spec/support/rubygems_ext.rb
@@ -117,7 +117,14 @@ module Spec
     def gem_activate_and_possibly_install(gem_name)
       gem_activate(gem_name)
     rescue Gem::LoadError => e
-      Gem.install(gem_name, e.requirement)
+      # Windows 3.0 puts a Windows stub script as  `rake` while it should be
+      # named `rake.bat`. RubyGems does not like that and avoids overwriting it
+      # unless explicitly instructed to do so with `force`.
+      if RUBY_VERSION.start_with?("3.0") && Gem.win_platform?
+        Gem.install(gem_name, e.requirement, force: true)
+      else
+        Gem.install(gem_name, e.requirement)
+      end
       retry
     end
 
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -830,7 +830,9 @@ class Gem::Specification < Gem::BasicSpecification
       next names if names.nonzero?
       versions = b.version <=> a.version
       next versions if versions.nonzero?
-      Gem::Platform.sort_priority(b.platform)
+      platforms = Gem::Platform.sort_priority(b.platform) <=> Gem::Platform.sort_priority(a.platform)
+      next platforms if platforms.nonzero?
+      b.base_dir == Gem.path.first ? 1 : -1
     end
   end
 
--- a/test/rubygems/test_gem.rb
+++ b/test/rubygems/test_gem.rb
@@ -1615,9 +1615,11 @@ class TestGem < Gem::TestCase
 
     Gem.use_paths Gem.dir, [Gem.dir, Gem.user_dir]
 
+    spec = Gem::Dependency.new("m", "1").to_spec
+
     assert_equal \
       File.join(Gem.dir, "gems", "m-1"),
-      Gem::Dependency.new("m","1").to_spec.gem_dir,
+      spec.gem_dir,
       "Wrong spec selected"
   end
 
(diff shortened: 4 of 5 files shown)


MSP-Greg:
As mentioned in [#7664](https://github.com/rubygems/rubygems/pull/7664#issuecomment-2138352085), the CI running for this PR is generating a false positive for all Windows workflow jobs.

Open the job log for any Windows run and look at the 'Install Dependencies' and 'Run Tests' step logs, and no code is running.

I'll open an issue in the PowerShell repo, as this should be throwing an error...

deivid-rodriguez:
Nice catch @MSP-Greg, let me cherry-pick your commit in #7701 here so I can see if CI is fixed.

And thanks for also reporting in powershell, it's not good that these are not reported as failures.

MSP-Greg:
@deivid-rodriguez

Thanks for your work on this.  Glad to see it passing.

The only thing I'm not sure of is whether `activation_priority` should set to calculate once, as it might be called often when a large number of specs exist, with a lot of created arrays.  You probably know better than anyone, and I've never debugged `Array#sort` calls...

JFYI, Locally, I believe I've got the same Rubies as GHA.  The 3.0 and 3.1 rake 'bin' files seem identical, so the issue may RubyGems 3.2.33.  If you want me to check (installing rake in the Ruby 3.0.z folder), I can.

Normally, I do everything in 'user-install', as I like to keep my Ruby installs 'clean'.  My normal Rubies on Ubuntu & Windows are head/master builds...

deivid-rodriguez:
I extracted the method to make this sort a bit more clear, but indeed it's a few more arrays. Not sure if I should care or not. I guess I'll make some tests to see if it's noticiable.

Regarding the bin files, I'm not sure what to check. I thought what GHA provides is a pristine Windows Ruby install. In any case, since it's something specific to an old Ruby, and our tests, I would not give it more attention.

MSP-Greg:
> Not sure if I should care or not.

What, me worry?   I have no idea how often or when the specs are sorted, so whatever.

In any case, since it's something specific to an old Ruby, and our tests, I would not give it more attention.

Not sure, it might have to do with overwriting the original rake install, but as you said, both the Ruby & RubyGems are old.  We've got better things to work on...

deivid-rodriguez:
On my system with 105 gems, the difference is minimal:

### On master

Total allocated: 3.09 MB (23064 objects)
Total retained:  289.20 kB (2834 objects)

### With this PR

Total allocated: 3.09 MB (23072 objects)
Total retained:  289.20 kB (2834 objects)

We could back to the previous allocations with this alternative patch

```diff
diff --git a/lib/rubygems/specification.rb b/lib/rubygems/specification.rb
index a51a498158..da55a2e6d3 100644
--- a/lib/rubygems/specification.rb
+++ b/lib/rubygems/specification.rb
@@ -828,7 +828,11 @@ def self._resort!(specs) # :nodoc:
     specs.sort! do |a, b|
       names = a.name <=> b.name
       next names if names.nonzero?
-      b.activation_priority <=> a.activation_priority
+      versions = b.version <=> a.version
+      next versions if versions.nonzero?
+      platforms = Gem::Platform.sort_priority(b.platform) <=> Gem::Platform.sort_priority(a.platform)
+      next platforms if platforms.nonzero?
+      b.base_dir == Gem.path.first ? 1 : -1
     end
   end
```

I'll revert to that alternative, and I'll consider introducing the `activation_priority` method in the next PR.

deivid-rodriguez:
Merging to get our Windows CI back. Thanks @MSP-Greg!

simi:
:muscle: @MSP-Greg

simi (approved):

deivid-rodriguez:
Linking to the PowerShell ticket for visibility: https://github.com/PowerShell/PowerShell/issues/23874. Subscribed!

MSP-Greg:
The Actions 'minimum-time' feature:

https://github.com/orgs/community/discussions/126651

I may open an issue in https://github.com/actions/runner/ if the discussion leads nowhere

deivid-rodriguez:
I think the right place to fix this is in powershell, I'm not sure about `mininum-time` actions feature, particularly if it's just driven by having found a bug somewhere else.

MSP-Greg:
I'd agree.  Conversely, CI false positives are dangerous.  This one may have gone on for a bit longer, as it was noticed only because the commits are pushed to ruby/ruby.  I noticed it when I saw a ruby-loco failure, and others noticed it in ruby/ruby.

`minimum-time` should be optional, just like `timeout-minutes`.  If it existed, and was used in the workflows here, the problem would have been identified immediately.

'a bug somewhere else' is messy, in this case we're considering something to be a bug that's related to error handling of a unique situation.  PowerShell is a well-used, mature product (like the repos we contribute to), and you can bury yourself handling 'unique situation' errors...

PR #7705, opened 2024-05-31
Move compact index concurrency to fetcher

martinemde:

Finishing #7672 was made significantly harder because of the way the tests for the Fetcher::CompactIndex couldn't encapsulate CompactIndexClient behavior as a simple mock. This is because the concurrency model was split between the fetcher and the client only for the benefit of a single method on the client. 

It's much easier if the caller can just use the client concurrently if desired. Nothing else changes except pushing the concurrency model completely into Fetcher::CompactIndex. This makes sense especially after the refactor in #7678 which made `execution_mode` seem out of place. The only method using it was `dependencies` (which I consider soft deprecated now since it just maps and then calls info.)


deivid-rodriguez (approved):
So much better! 😍

